PR #1, opened 2019-10-06
Fix typo in wasm example button text

nico-abram:


emilk:
This one slipped by, sorry for missing it. Fixed in ec93f20.

PR #2, opened 2020-05-01
Mention miniquad backend in readme

not-fl3:
Hi! 

Made emigui backend for `good-web-game` and apparently it works! 

https://github.com/not-fl3/good-web-game/blob/master/examples/emigui.rs this code should work on linux, mac, windows, android and wasm

Web demo: https://not-fl3.github.io/miniquad-samples/emigui.html


emilk:
Very cool!

Please note that emigui is still in heavy flux. I will change a lot in the interface the coming weeks, and maybe even change the name, so be ready for that :)

PR #9, opened 2020-07-24
Fix master

tomassedovic:
This fixes build errors when using the latest checkout of egui as a dependency.

The first one is caused by entries in `Cargo.lock` referencing a report that no longer seems to exist in the source repo and the second is a compile error when the with_serde feature is disabled (which seems to be the default when using the crate as a library).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -427,7 +427,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 [[package]]
 name = "egui"
 version = "0.1.2"
-source = "git+https://github.com/emilk/emigui#e3d1d6c99ce37090d496b4f210aaef700235dd9b"
+source = "git+https://github.com/emilk/emigui#554e6e7120fa0d3d052b9802146c5dcf003f46ae"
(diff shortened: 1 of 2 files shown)


tomassedovic:
Greetings @emilk! I've pulled from master yesterday (I've added egui to my game) and saw the breakages above.

I'm not super familiar with the codebase yet so if there's a better way to fix this, please let me know! I'll be happy to update the pull request.

Also, thank you for for this crate! I'm really like the way it works and I'm happy with the results in my game. I needed to tweak a few things that I'd love to send back if you're interested. Are you open to receiving PRs?

emilk:
Hi!

Thanks to this PR I decided to finally set up [Github Actions CI](https://github.com/emilk/emigui/actions) for `egui`, and after some work it is now all green.

> Also, thank you for for this crate! I'm really like the way it works and I'm happy with the results in my game. I needed to tweak a few things that I'd love to send back if you're interested. Are you open to receiving PRs?

Yes I am!

tomassedovic:
\o/ thank you!

PR #14, opened 2020-08-04
FIX(container): add menu containers public re-export

liketechnik:
Really like the concept of this library and was eager to try some stuff out.

Was really happy to find the menu containers in the documentation, but the compiler complained they were private in the module.

I'm not totally sure if my fix is the correct/planned way for adding the menu containers, but it seems to work :D

(P.S.: Sorry for the push spam. Apparently I should really start looking at the details of my work before pushing, because I'm too stupid to get them right the first time...)

### DIFF
--- a/egui/src/containers.rs
+++ b/egui/src/containers.rs
@@ -8,6 +8,6 @@ pub(crate) mod scroll_area;
 pub(crate) mod window;
 
 pub use {
-    area::Area, collapsing_header::CollapsingHeader, frame::Frame, popup::*, resize::Resize,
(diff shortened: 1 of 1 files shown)


emilk:
Menus are barely half-done, but feel free to give them a spin!

I've exposed them now in 56352c26a492915ef9aba24cea204b825f63469f

Use it as:

``` rust
use egui::menu;

menu::bar(ui, |ui| {
    menu::menu(ui, "File", |ui| {
        if ui.add(Button::new("Open")).clicked {
            ...
        }
    });
});
```

PR #15, opened 2020-08-05
Add `collapsible` to Window

tomassedovic:
This lets the user decide whether the Window can be collapsed or not.
The default is `true` (window is collapsible), but calling
`window.collapsible(false)` will hide the collapsing triangle icon and
prevent the window from being collapsed by clicking on the title.

### DIFF
--- a/egui/src/containers/window.rs
+++ b/egui/src/containers/window.rs
@@ -12,6 +12,7 @@ pub struct Window<'open> {
     pub frame: Option<Frame>,
     pub resize: Resize,
     pub scroll: Option<ScrollArea>,
+    pub collapsible: bool,
 }
 
 impl<'open> Window<'open> {
@@ -36,6 +37,7 @@ impl<'open> Window<'open> {
                     .always_show_scroll(false)
                     .max_height(f32::INFINITY),
             ), // As large as we can be
+            collapsible: true,
         }
     }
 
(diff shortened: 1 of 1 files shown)


egui/src/containers/collapsing_header.rs
```diff
@@ -19,6 +19,11 @@ pub(crate) struct State {
 
     /// Height of the region when open. Used for animations
     open_height: Option<f32>,
+
+    /// Whether collapsing controls and behavior is visible and
+    /// active. Set to `false` if you don't want to collapse your
+    /// windows.
+    pub enabled: bool,
```
emilk, line 26: This is a weird place to store this boolean, considering it is not used in this file. In particular, if you just set up a `CollapsingHeader` and set `enabled = false` on it, it is still collapsible. I think it would make more sense to NOT have it here, but to keep `collapsible` to only `window.rs`.

egui/src/containers/window.rs
```diff
@@ -541,7 +551,7 @@ fn show_title_bar(
         let item_spacing = ui.style().item_spacing;
         let button_size = ui.style().start_icon_width;
 
-        {
+        if collapsing.is_enabled() {
```
emilk, line 554: replace with `if collapsible {`

egui/src/containers/window.rs
```diff
@@ -206,6 +215,7 @@ impl<'open> Window<'open> {
                 collapsing_id,
                 default_expanded,
             );
+            collapsing.enabled = collapsible;
```
emilk, line 218: remove

egui/src/containers/window.rs
```diff
@@ -624,6 +634,7 @@ impl TitleBar {
         if ui
             .interact(self.rect, title_bar_id, Sense::click())
             .double_clicked
+            && collapsing.is_enabled()
```
emilk, line 637: replace with `&& collapsible`

tomassedovic:
Thanks! I've removed the changes from `collapsing_header~ and made this a `Window`-only change. Please let me know, if you'd like to do this differently.

PR #16, opened 2020-08-05
Add fixed_pos to Window

tomassedovic:
This allows changing the position of a window by opening the window
with a different value.

### DIFF
--- a/egui/src/containers/window.rs
+++ b/egui/src/containers/window.rs
@@ -91,6 +91,7 @@ impl<'open> Window<'open> {
         self.default_pos(rect.min).default_size(rect.size())
     }
 
+    /// Positions the window and prevents it from being moved
     pub fn fixed_pos(mut self, pos: impl Into<Pos2>) -> Self {
         self.area = self.area.fixed_pos(pos);
         self
--- a/egui/src/containers/window.rs
+++ b/egui/src/containers/window.rs
@@ -91,6 +91,11 @@ impl<'open> Window<'open> {
         self.default_pos(rect.min).default_size(rect.size())
     }
 
+    pub fn fixed_pos(mut self, pos: impl Into<Pos2>) -> Self {
+        self.area = self.area.fixed_pos(pos);
+        self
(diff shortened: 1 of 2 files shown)


tomassedovic:
In my game, I want to have more control over how the windows work, e.g. control their size and position (e.g. cover the whole screen minus some padding), or disable collapsing (#15).

I'm not sure you want to go with the direction of extending the Window widget. I guess the alternative would be to write a custom widget, but since everything else would be the same, here's a couple PRs to see what you think.

egui/src/containers/window.rs
```diff
@@ -91,6 +91,11 @@ impl<'open> Window<'open> {
         self.default_pos(rect.min).default_size(rect.size())
     }
 
+    pub fn fixed_pos(mut self, pos: impl Into<Pos2>) -> Self {
```
emilk, line 94: Could do with a docstring, e.g. `/// Positions the window and prevents it from being moved`.

I should have added one to `Area::fixed_pos` but I guess I was lazy :P

emilk:
Thanks for describing your use case!

I think it makes sense to add helpers like these to disable parts of the `Window` features. Fixed position and fixed size windows makes a lot of sense.

If you don't even want the title bar you could consider just combining `Frame` and `Area` directly. Something like this:


``` rust
Area::new("my_area").fixed_pos([32.0, 16.0]).show(ctx, |ui| {
    Frame::window(ui.style()).show(|ui| {
        ui.label("Hello world!");
    });
});
```

egui/src/containers/window.rs
tomassedovic: Thanks, I've added it to`Window` and `Area` both.

tomassedovic:
Thanks for showing the example of Area+Frame! In this particular usecase, I really want the UI, but I did have another one where I basically wanted a statically-positioned borderless window and wasn't sure how to do it nicely (ended up with a child `Ui` and a bunch of extra code).

PR #18, opened 2020-08-17
Mutable ui borrows

### LINKED ISSUE #17: No lightweight way to adjust ui style
I could be wrong about the following. I'm just starting to learn Rust, so I could have misunderstood something. I believe this is the only way to change style elements given the current API:

    let mut style = ui.style().clone();
    style.item_spacing = egui::vec2(x, y);
    ui.set_style(style);

The `ui`'s `style` member is private and the `ui.style()` method returns an immutable reference. So the only way to change a single style element is to clone the whole struct out, make the change, and pass ownership back into the `ui`. This is awkward to implement, and forces needless copies.

The patterns in the `Style::ui` widget hints that some kind of public access is safe and may have been intended, but I can't find a pathway for it.

vkkoskie:
resolves #17

### DIFF
--- a/egui/src/ui.rs
+++ b/egui/src/ui.rs
@@ -257,6 +257,8 @@ impl Ui {
         &self.layout
     }
 
+    /// Mutably borrow internal `Layout`.
+    /// Changes apply to this `Ui` and its subsequent children.
     pub fn layout_mut(&mut self) -> &mut Layout {
         &mut self.layout
     }
--- a/egui/src/ui.rs
+++ b/egui/src/ui.rs
@@ -93,6 +93,10 @@ impl Ui {
         &self.style
     }
 
+    pub fn style_mut(&mut self) -> &mut Style {
+        &mut self.style
+    }
+
     pub fn set_style(&mut self, style: Style) {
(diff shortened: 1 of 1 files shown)


egui/src/ui.rs
```diff
@@ -251,6 +257,12 @@ impl Ui {
         &self.layout
     }
 
+    /// Mutably borrow internal `Layout`.
+    /// Changes apply to this `Ui` and its subsequent children.
+    pub fn layout_mut(&mut self) -> &mut Layout {
```
emilk, line 262: The design of how `Layout` works is not yet complete. In particular, when changing `ui.layout` one must also take care that `ui.cursor` is still valid for that new layout. The `Ui::set_layout` tries to do this, but I'm not quite happy with the design yet (which is why `Ui::set_layout` is marked with `// TODO: remove`). I think `cursor` may be removed from `Ui` and moved to some Layout-specific layout state.

Long story short: if you want to change layout you should use helpers like `ui.horizontal(...)` or `ui.set_layout`, but accessing the `Layout` directly via a `layout_mut()` is not a good idea.

PR #20, opened 2020-09-04
Kinetic scrolling

1HPorange:
I implemented kinetic scrolling to the extent that it makes sense to me right now:

- Kinetic scrolling is only active when you drag the background, not when you use the scroll wheel or drag the scroll handle. I feel like people would prefer precise scrolling when using those methods.
- Kinetic velocity (like offset) is stored and updated as a Vec2, event though right now we only use the y component
- I basically copy-pasted the kinetic part from area.rs; Maybe you would prefer to factor it out, but I'm not sure where to


emilk (approved):
Works nicely; thanks!

PR #21, opened 2020-09-12
Fix continuous RunMode in glium backend

parasyte:
The primary change here is that the `winit` control flow is set according to `RunMode`, and not assumed to be `Wait`. I also had to change the event handler to `RedrawEventsCleared`.

Prior to this patch, the `Continuous` run mode didn't work correctly in the glium backend. But it does work as expected in the web backend.

### DIFF
--- a/egui_glium/src/backend.rs
+++ b/egui_glium/src/backend.rs
@@ -102,33 +102,44 @@ pub fn run(
     let mut clipboard = init_clipboard();
 
     event_loop.run(move |event, _, control_flow| {
-        *control_flow = glutin::event_loop::ControlFlow::Wait;
+        let mut redraw = || {
+            let egui_start = Instant::now();
+            raw_input.time = start_time.elapsed().as_nanos() as f64 * 1e-9;
+            raw_input.seconds_since_midnight = Some(local_time_of_day());
 
-        match event {
-            glutin::event::Event::RedrawRequested(_) => {
-                let egui_start = Instant::now();
-                raw_input.time = start_time.elapsed().as_nanos() as f64 * 1e-9;
-                raw_input.seconds_since_midnight = Some(local_time_of_day());
+            let mut ui = ctx.begin_frame(raw_input.take());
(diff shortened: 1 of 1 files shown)


emilk (changes requested):
> Prior to this patch, the `Continuous` run mode didn't work correctly in the glium backend

Can you elaborate what was incorrect? For me `master` works as it should, i.e. repainting at 60 FPS on my 60 Hz display..

I ask, because trying this PR the `Reactive` mode is now definitely broken on my Mac: it repaints the GUI even when the window is in the background (the point of the `Reactive` mode is to avoid unnecessary repaints).

parasyte:
On Windows 10, both modes act completely identically. No updates occur unless an event is delivered (especially moving the mouse cursor). I believe this is caused by a bug in winit. See https://github.com/rust-windowing/winit/issues/987 and https://github.com/rust-windowing/winit/issues/1619

Continuous mode _does_ work correctly on macOS. And this PR also breaks Reactive mode on macOS. So I'll have to go back to the drawing board on this.

parasyte:
@emilk I force-pushed a new patch for this. Tested on Windows and macOS. I haven't tested on Linux (X11 or Wayland).

Neither event handler `RedrawRequested` or `RedrawEventsCleared` worked on both platforms, so I decided to use the latter for Windows and the former (original event) for all other platforms. I wasn't able to come up with a good pattern for platform-specific event handling, so I moved all of the code into a closure. It hurts diff readability, but I don't know of another way to solve that.

I also had to add another `request_redraw()` call for `Continuous` mode to work on macOS.

parasyte:
And rebased on master.

emilk (approved):
Great, nice work!

bel8z:
A bit late here, but this fix doesn't work as expected for me.

RedrawEventsCleared is raised at the vsync frequency, causing the Reactive mode to behave as Continuous. Removing the platform-dependent handling and responding to RedrawRequested only does the work on my machine. Is this fix still needed?

Using master on Windows 10 btw.

PR #22, opened 2020-09-12
Replace default font

parasyte:
This saves about 46 KB overall, and improves the text quality (IMHO).

Before: 
egui-original

After: 
egui-oxygen

As you can see, Oxygen is not the most ideal font. It only supports the Latin character set. Still feels worth the increase in readability and decrease in executable size.


emilk:
Thanks for the PR!

I agree Comfortaa has some legibility issues (especially lowercase `a` vs `o`), but I like the friendly roundness of it. I also like how it gives Egui a distinct feel. In comparison Oxygen looks kind of vanilla and boring (dare I say "corporate"?). 

I would like to try out more alternatives. Ideally the font should be both legible and distinct, but that may turn out to be a tall order.

Supporting non-latin script should probably be an opt-in for the size reasons.

Anyway, I am gonna leave the PR open to solicit more opinions, and to mull it over a bit more.

arirawr:
Agree that Oxygen is considerably easier to read.

parasyte:
Here's a short list of readable fonts that have "rounded features":

- [Titillium Web](https://fonts.google.com/specimen/Titillium+Web)
- [Maven Pro](https://fonts.google.com/specimen/Maven+Pro)
- [Antic](https://fonts.google.com/specimen/Antic)
- [Carme](https://fonts.google.com/specimen/Carme)

All of these are available under the terms of the OFL.

emilk:
Having tested a few alternatives now, I am leaning towards [Ubuntu-Light](https://fonts.google.com/specimen/Ubuntu). One benefit for it is that it also supports a few more characters than most of the proposed alternatives

**Ubuntu-Light (proposed)**:
Ubuntu-Light


**Comfortaa-Regular (current)**:
Comfortaa-Regular

parasyte:
Keeping this simple, Ubuntu-Light looks good to me.

For supporting more of the Unicode standard and non-Latin languages in general, there is a wealth of in-depth information available. Here are a few selections, which are all written by Rust enthusiasts:

- https://raphlinus.github.io/rust/skribo/text/2019/02/27/text-layout-kickoff.html
- https://yeslogic.com/blog/allsorts-rust-font-shaping-engine.html
- https://kas-gui.github.io/blog/why-kas-text.html

That is to say, yes this ticket should just move forward with a simple font change. Focus on advanced text rendering can happen later independently.

emilk:
Thanks for those links @parasyte - I've created a tracking issue for further improvements to the text rendering at https://github.com/emilk/egui/issues/56

emilk:
Font changed to Ubuntu-Light in https://github.com/emilk/egui/commit/428e2373e553a627e1e156f4be7ce957c7cb731e

PR #25, opened 2020-10-01
build_web.sh: fix script

yjh0502:
It looks `build_web.sh` is not working as intended, so here's a fix.


emilk (approved):
Thanks!

Both versions work for me, so if only the latter works for you, let's go for it! I'm not a bash expert, so can't really tell why one would chose one over the other.

PR #34, opened 2020-10-28
Make `from_get_set` public

Hizoul:
Hi @emilk,

thank you very much for creating a pure rust immediate mode GUI!
I am currently playing around with it, and it works perfectly :)
I did come along a small limitation that would be fixable by making the `Slider::from_get_set` public.
I am storing my state in an Arc<Mutex<T>> or Atomic value that is accessed by a background thread and would like to avoid the additional lifetime for `&'a mut`, but I can easily clone the Arc and give it to the closure that the `from_get_set` method receives.

In case this is intended and what I am trying to do is bad I am happy to be taught better :)

Kind Regards,
Hizoul

### DIFF
--- a/egui/src/widgets/slider.rs
+++ b/egui/src/widgets/slider.rs
@@ -49,7 +49,7 @@ pub struct Slider<'a> {
 }
 
 impl<'a> Slider<'a> {
-    fn from_get_set(
+    pub fn from_get_set(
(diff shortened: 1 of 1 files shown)


emilk (approved):
Thank you Hizoul!

That is a good use case to consider, and I think it makes perfect sense to make `Slider::from_get_set` public. Good call!

PR #35, opened 2020-10-29
Downgrade ahash for wasm build

samsamai:
When running `./build_web.sh`, I ran into the following errors:

```
error: target is not supported, for more information see: https://docs.rs/getrandom/#unsupported-targets
   --> /Users/sss/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.0/src/lib.rs:224:9
    |
224 | /         compile_error!("target is not supported, for more information see: \
225 | |                         https://docs.rs/getrandom/#unsupported-targets");
    | |_________________________________________________________________________^

error[E0433]: failed to resolve: use of undeclared type or module `imp`
   --> /Users/sss/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.0/src/lib.rs:246:5
    |
246 |     imp::getrandom_inner(dest)
    |     ^^^ use of undeclared type or module `imp`

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0433`.
error: could not compile `getrandom`.

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed
````

Seems like getrandom doesn't see `wasm32-unknown-unknown` as a valid target.
Dropping ahash, which has the getrandom dependency, to version 0.4.6 fixes this for now.

### DIFF
--- a/egui/Cargo.toml
+++ b/egui/Cargo.toml
@@ -10,12 +10,17 @@ readme = "../README.md"
 repository = "https://github.com/emilk/egui"
 categories = ["gui", "graphics"]
 keywords = ["gui", "imgui", "immediate", "portable", "gamedev"]
-include = [ "**/*.rs", "Cargo.toml", "fonts/ProggyClean.ttf", "fonts/Comfortaa-Regular.ttf"]
+include = [
+  "**/*.rs",
+  "Cargo.toml",
+  "fonts/ProggyClean.ttf",
+  "fonts/Comfortaa-Regular.ttf",
+]
 
 [lib]
 
 [dependencies]
-ahash = { version = "0.5", features = ["std"], default-features = false }
+ahash = { version = "0.4", features = ["std"], default-features = false }
 parking_lot = "0.11"
 rusttype = "0.9"
 serde = { version = "1", features = ["derive"], optional = true }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: egui/Cargo.toml [hunk 1]
```diff
@@ -10,12 +10,17 @@ readme = "../README.md"
...
+  "Cargo.toml",
+  "fonts/ProggyClean.ttf",
+  "fonts/Comfortaa-Regular.ttf",
+]
 
 [lib]
 
 [dependencies]
-ahash = { version = "0.5", features = ["std"], default-features = false }
+ahash = { version = "0.4", features = ["std"], default-features = false }
```
emilk, line 23: It would be good with a note here about why we are not using the latest version

emilk:
Thanks for noticing this!

Seems to me like two other things should also happen:

* Open an issue for `ahash`
* Run the web build on the CI so this doesn't happen again

I have also been thinking about making `ahash` an optional dependency (I only use it for a speed boost anyway).... just me thinking out loud :)

emilk (changes requested):
Please add a comment to Cargo.toml about why we aren't using the latest `ahash` version (maybe link to this PR)

emilk (approved):

emilk:
Issue filed to https://github.com/tkaitchuck/aHash/issues/59

emilk:
Upgraded to ahash 0.5 in 8393d1df4714f82c813fe24242a51c8220038d2e

PR #36, opened 2020-10-30
Check demo_web on CI

samsamai:
Catch possible wasm target issues by checking the demo_web app on CI


emilk (approved):
Awesome, thank you so much for this!

PR #38, opened 2020-11-03
Full widget culling

emilk:
Experiment with how we could accomplish full widget culling, to early-out from a lot of widget painting code.

It adds some complexity, and the benefits are small (10-15% in the `demo_windows_full` benchmark).


emilk:
Not worth it

PR #41, opened 2020-11-10
Rename menues to menus

gamecubate:
Minor error : menues -> menus. Updated menu.rs and memory.rs.


emilk (approved):
Nice! Spelling was never my strength :)

PR #52, opened 2020-11-26
Fix the accessor for desired rows.

phoglund:


emilk (approved):
Oh, good catch! The builder-pattern leads to a lot of boilerplate which leads to copy-pasta which leads to dumb mistakes like this.

Thanks for fixing it!

PR #61, opened 2020-12-10
actually take path for `FileStorage::from_path`

tgolsson:
Otherwise; if one has a userland pathbuf an unsafe conversion back to String has to be made:

```rust
let mut user_config_directory = dirs::config()?;
user_config_directory.push("foo");
user_config_directory.push("bar.txt");

let egui_filepath = user_config_directory.to_str()?.to_owned(); // fallible but both sides want to use a path.
FileStorage::from_path(egui_filepath);
```

Tests (`cargo test`) unfortunately seem to be broken on main already; but example_glium works as expected with the removal of `.into()` below.

### DIFF
--- a/demo_glium/src/main.rs
+++ b/demo_glium/src/main.rs
@@ -6,7 +6,7 @@ fn main() {
     let title = "Egui glium demo";
 
     // Persist app state to file:
-    let storage = egui_glium::storage::FileStorage::from_path(".egui_demo_glium.json".into());
+    let storage = egui_glium::storage::FileStorage::from_path(".egui_demo_glium.json");
 
     // Alternative: store nowhere
     // let storage = egui::app::DummyStorage::default();
--- a/egui_glium/CHANGELOG.md
+++ b/egui_glium/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Changed
+
+* FileStorage::from_path now takes `Into<Path>` instead of `String`
+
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed 🔧
 
 * Changed default font to [Ubuntu-Light](https://fonts.google.com/specimen/Ubuntu).
+* FileStorage::from_path now takes `Into<Path>` instead of `String`
```
emilk, line 17: This changelog is only for the `egui` crate. This is not very well explained, and very confusing since `egui_glium` doesn't even have it's own changelog :) So I just added one in https://github.com/emilk/egui/commit/8c82eb1970a00aa65506aed102141dabbd3a3f8b

You can move the changelog line to `egui_glium/CHANGELOG.md`, but it's also fine to just remove this (since it is a very minor change for a crate that is not that well used ... yet)

emilk (changes requested):
Thanks for the PR! I was lazy not to do this in the first place :)

> Tests (`cargo test`) unfortunately seem to be broken on main already

Pure `cargo test` fails only because of `egui_web` requires experimental features to work. I should probably fix that.

`cargo test -p egui_glium` (and the more important `cargo test -p egui` ) works fine, as does running `./check.sh` which sets the `RUSTFLAGS` requires for the experimental features.

tgolsson:
@emilk Updated per for your feedback, thanks!

emilk (approved):
Thanks!

Issue #31, opened 2020-10-25
Move keyboard focus with tab to any interactive widget


emoon:
It would be great to have support for keyboard only navigation (i.e a concept of focused widget and the ability to move focus to another one) that would allow "tab" key for example to move to the next widget and "return" would act as press, etc. 

This in return would make it quite easy to add gamepad support as it would just behave as a subset of a regular keyboard.

emilk:
Switching between input fields with tab and shift-tab has now been implemented on `master`. Pressing enter in a singleline text edit box now sets the `kb_focus_lost` flag in the `Response` object, so you can use that.

emoon:
Tested this today and it seems that I wonder if it would be possible to support for all widgets (such the way it works in a regular gui) and not just for input fields?

emilk:
That would be nice yes, but it is not very high on my priority list right now

emoon:
I think I will try to implement this unless it's too complicated. Any pointers before I start?

emilk:
Wow, that would be great!

The feature that handles focus of `TextEdit`:s is called "kb_focus".

I think first of all, let's split this into two tasks:

A) tab also moves kb_focus to buttons, checkboxes, sliders etc
B) gamepad support for moving kb_focus up/down/left/right

Let's start with A, and postpone B until A is done.

---

First we have some design problems:

When clicking a `TextEdit` we get a cursor in it, and it is obvious it has kb_focus. If we then press tab and move focus to a checkbox, how do we indicate it has focus? We need a uniform language for focus. To start, the easiest would probably be some sort of thing colored rectangle around the widget with kb_focus (including `TextEdit`:s). That or changing its fg_color. This accent could perhaps become confusing when it comes to `SelectableLabel` which uses color to indicate "this is selected".

We can give kb_focus to a TextEdit by clicking it. Can we give kb_focus to a checkbox or slider by clicking it? I strongly vote no on this one, as it would give add some weird accent to the last button you pressed all the time. Instead I propose that the only way to give kb_focus to a button is by using tab/shift-tab.

What happens when a slider has kb_focus? Arrows to move the slider one pixel? Enter to edit the value?

How do we toggle/click buttons? Enter? Space?

---

As for the implementation:

[...]

PR #65, opened 2020-12-13
Allow user to check for Key::Space

emoon:
This is part of #31 to allow a user (or internally) to check for `Key::Space` to press on widgets to toggle on/off/press/etc. I didn't want to do everything in one large PR as this one may be useful otherwise as well.


emilk (approved):
I'm glad you are doing many small PR:s instead of one big one!

PR #69, opened 2020-12-18
Make it possible to set Glium windows as not resizable.

phoglund:
My devs are complaining about resize flickering etc, and the UI I'm building shouldn't resize, so I want to be able to turn it off.

I considered putting this into the storage, but it doesn't really make sense to save on app exit since it's probably something
you want to control as an app maker.

I could also make an enum so you say

egui_glium::run("Title", egui_glium::Resizable, ...)

, which is a bit more readable than true or false.


emilk (changes requested):
I think having a fixed sized glium window is probably a good idea, but first:

> My devs are complaining about resize flickering etc

The problem is probably [this line](https://github.com/emilk/egui/blob/master/example_glium/src/example_app.rs#L49). It is nice to have an auto-sized window (I think), but it would be nice to have it auto-size once (first frame), and then have an option to stay that size. That's a separate issue though.

OK - onto the problem at hand!

I am planning to move options such as title into [`trait App`](https://github.com/emilk/egui/blob/master/egui/src/app.rs#L14). So to that end, I think it would be good to add this function:

``` rust
/// If this returns `Some`, the backend will use this as the fixed size (in points) of the window.
fn fixed_size(&self) -> Option<Vec> { None }
```

Then your app can return a size and we can have `egui_glium` use that for the window size, and `egui_web` use it for canvas size (I can implement that part).

phoglund:
All right, you got it! I did something similar to what you suggested:

trait App {
  fn is_resizable(&self) -> bool { true}
}

I didn't implement default_size yet. It would not be useful to me, but I guess you're thinking of resizable window users here? And you want it to be used if there is no window size in storage already?

In my latest version of the code I'm doing this (I think you suggested something similar above):

```
   // Grab the size on the first frame and then don't allow resizing.
        match self.window_size {
            Some(size) => integration_context.output.window_size = Some(size),
            None => self.window_size = Some(ctx.used_size()),
        };
```

I could also have hard-coded the win size, but this way I don't have to when I change things in the UI. So my use case is fixed size after the first frame :)

emilk (changes requested):
Looks good! It need a rebase on `master` to fix some merge conflicts. Please also add a line to `egui_glium/CHANGELOG.md`

phoglund:
> Looks good! It need a rebase on `master` to fix some merge conflicts. Please also add a line to `egui_glium/CHANGELOG.md`

Done

emilk (approved):
Thank you!

Issue #75, opened 2020-12-21
Making a scrollable combo box doesn't work as expected.


ArjunNair:
![egui issue](https://user-images.githubusercontent.com/1013419/102754628-40b83680-4393-11eb-81aa-a3fc4d7c96db.png)

I'm trying to list all the files in the combo box (as buttons). This will over spill the window area because the drop down is bigger than the window height as can be seen above. Since combo box itself doesn't seem to support scrolling out of the box, I thought of adding a ScrollArea within the combo box (since it's a container presumably). 

This kind of works, in that I can scroll by moving the mouse up and down, but couple of things are broken:
1) The scroll area width is the size of the Window width which is the parent container ui. I'm not sure how I can constrain it to the combo box dimensions. 
2) The vertical scroll bar doesn't appear (perhaps because it falls outside the window width).

Frankly, I expected the combo box to automatically take care of this. That is, it has a scroll area built into it, so that if the item dimensions > default combo box dimension, the comb box shows appropriate scroll bars (vertical/horizontal). 

Failing the above, what would be the idiomatic way to do this in egui? I've included the source code for how I'm doing things now:
```
&egui::Window::new("Chipper")
            .fixed_pos(Pos2::new(0f32,0f32))
            //.default_size(vec2(WINDOW_WIDTH as f32, WINDOW_HEIGHT as f32))
            .resizable(false)
            .collapsible(false)
            .show(&mut egui_ctx, |ui| {
                if !is_paused {
                    ui.label(format!("FPS: {} ({} ms/frame)", fps, avg_frame_time));
                }
                else {
                    ui.label(format!("PAUSED"));
                }
                ui.add(Image::new(chip8_tex_id, vec2((CHIP8_DISPLAY_WIDTH * DISPLAY_SCALE) as f32, (CHIP8_DISPLAY_HEIGHT * DISPLAY_SCALE) as f32)));
                ui.label("");
                combo_box_with_label(ui, "ROM files", selected_rom, |ui| {
                    ScrollArea::from_max_height(50f32).show(ui, |ui| {
                        for (f, _p) in &rom_files {
                            if ui.button(f).clicked {
                                selected_rom = f;
                                chip8.boot_rom(rom_files.get(selected_rom).expect("No rom files to load!")).expect("Failed to load rom!");
                            };
                        }
                    })
                });
                //There is probably a better way to add line breaks in egui....
                ui.label("");
[...]

PR #77, opened 2020-12-23
Feature: ComboBox now shows scrollbars with scrollable content.

ArjunNair:
With regards to https://github.com/emilk/egui/issues/75.

I've tried to make minimum changes to ComboBox to make it work as a scrollable container.  I also had to change one line in scroll_area.rs to make the scroll area fit the content width. Without it, the container exceeds the parent UI width causing the scroll bars to disappear.

I believe the changes are trivial enough that they don't break anything while providing some useful functionality out of the box (so to speak). I've tested the functionality with egui_demo (glium version) and it seems to be in order.


emilk:
Thanks for the PR! It is a good start, but the changes required were a bit more subtle. I went ahead and made them myself in https://github.com/emilk/egui/commit/38e36fd806120e2914649268bd72adbe1add9c13

PR #81, opened 2020-12-25
Add manual scrolling

### LINKED ISSUE #74: [Feature Request] Manually control ScrollArea scroll offset
I have the currently scenario:

https://user-images.githubusercontent.com/35241085/102737405-1714f600-4326-11eb-8cec-d0c3b1305856.mp4

that i want to control the scroll when moving the selected line keeping the focus at the active line.

This is what I got trying to achieve this:

Currently the offset state that controls the scroll position is not public. I tried creating an simple public fn to set the offset manually, but since we can't get the Id of the `ScrollArea.source_id`, we can't change the state. We only have access to the Id of the child_ui (`Prepared.content_ui`) inside the `show()` function.

Also since the Prepared state is passed from the `begin()` to the `end()`, we can't change it inside the `add_contents()`. We would need to get the state from the memory again in the `end()` if we changed it in the `add_contents()`. 

Maybe the way out is to pass the Id used to save the state in the return of the show:  `ScrollArea.show() -> ScrollResponse<R>` similar to the CollapsingResponse. 

There is probably better solution for this

lucaspoffo:
Fixes #74


egui/src/ui.rs
```diff
@@ -526,6 +526,14 @@ impl Ui {
         let clip_rect = self.clip_rect().intersect(rect); // Make sure we don't paint out of bounds
         Painter::new(self.ctx().clone(), self.layer_id(), clip_rect)
     }
+
+    pub fn scroll_to_here(&mut self, center_ratio: f32) {
+        // This is the imgui aproach to set the scroll relative to the position 
+        // 0.0: top of widget / 0.5: midle of widget / 1.0: bottom of widget
+        assert!(center_ratio >= 0.0 && center_ratio <= 1.0);
+        let offset = self.min_rect().height() - self.clip_rect().height() * center_ratio;
+        self.ctx().frame_state().set_scroll_offset_y(offset);
```
emilk, line 535: Instead of storing offset, store the screen-space location of the desired scroll area:

`let offset = lerp(self.min_rect().y_range(), center_ratio);`

This will allow it to work for any `Ui` (if `ScrollArea` takes care to interpret it correctly).

egui/src/ui.rs
```diff
@@ -526,6 +526,14 @@ impl Ui {
         let clip_rect = self.clip_rect().intersect(rect); // Make sure we don't paint out of bounds
         Painter::new(self.ctx().clone(), self.layer_id(), clip_rect)
     }
+
+    pub fn scroll_to_here(&mut self, center_ratio: f32) {
```
emilk, line 530: Make sure you add this to `Response` as well.  Scrolling to the top/center/bottom of a `Ui` is less useful that scrolling to a specific widget. The `struct Response` represents the place where a widget is located.

lucaspoffo:
Added the scroll_to_me inside Response,
Needed to pass the `center_ratio` to the `Context, so we can correctly centralize inside the `ScrollArea`.

Still need to document the methods, 
improve the demo, maybe do something similar to imgui.

lucaspoffo:
Update the scroll demo (made it similar to `imgui`) and moved it to its own module.
Added documentation to the methods.
Added `ScrollArea::scroll_offset()`.

egui/src/demos/demo_window.rs
```diff
@@ -68,9 +70,7 @@ impl DemoWindow {
         CollapsingHeader::new("Scroll area")
             .default_open(false)
             .show(ui, |ui| {
-                ScrollArea::from_max_height(200.0).show(ui, |ui| {
-                    ui.label(LOREM_IPSUM_LONG);
-                });
```
emilk, line 73: I think it would be good to keep a simple scroll demo to demonstrate how to create a simple `ScrollArea`.

emilk (changes requested):
Very nice! Thanks for taking your time to work on this.

I left a lot of comments. Most contentious will probably be `center_factor` vs `Align`. This is basically a tradeoff between flexibility and simplicity, and I'm keen to hear your thoughts on it!

emilk:
Another minor addition: when clicking a button to scroll to the top of that button, I think it would make sense to add a small offset (e.g. `item_spacing.y`) to the scroll. Otherwise the top of the button frame will be clipped by the scroll areas clip rect. In other words, it will look better if the top and bottom scrolls always came with an offset of `item_spacing.y`.

egui/src/containers/scroll_area.rs
```diff
@@ -155,6 +171,27 @@ impl Prepared {
 
         let content_size = content_ui.min_size();
 
+        let scroll_target = content_ui.ctx().frame_state().scroll_target();
+        if let Some(scroll_target) = scroll_target {
+            let center_ratio = content_ui.ctx().frame_state().scroll_target_center_ratio();
+            let height_offset = content_ui.clip_rect().height() * center_ratio;
+            let top = content_ui.min_rect().top();
+            let offset_y = scroll_target - top - height_offset;
```
emilk, line 179:
```suggestion
            let offset_y = scroll_target - lerp(content_ui.clip_rect().y_range(), center_ratio);
```

egui/src/types.rs
```diff
@@ -159,6 +159,19 @@ impl Response {
         self.ctx
             .interact_with_hovered(self.layer_id, self.id, self.rect, sense, self.hovered)
     }
+
+    /// Move the scroll to this UI.
+    /// The scroll centering is based on the `center_factor`:
+    /// * 0.0 - top  
+    /// * 0.5 - middle
+    /// * 1.0 - bottom
+    pub fn scroll_to_me(&self, center_factor: f32) {
```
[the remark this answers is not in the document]
lucaspoffo: I think for now using `Align` is good enough. Keep it simple and consistent, if the necessity for more control is needed it can be changed later. :+1:

egui/src/ui.rs
```diff
@@ -584,6 +584,19 @@ impl Ui {
         let painter = Painter::new(self.ctx().clone(), self.layer_id(), clip_rect);
         (response, painter)
     }
+
+    /// Move the scroll to this position.
+    /// The scroll centering is based on the `center_factor`:
+    /// * 0.0 - top  
+    /// * 0.5 - middle
+    /// * 1.0 - bottom
```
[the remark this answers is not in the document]
lucaspoffo: I changed the code example from the `scroll_to_cursor` to use  the `scroll_bottom` example similar from the demo. Makes more sense for this case. Added the button example for the `scroll_to_me`.

egui/src/ui.rs
```diff
@@ -584,6 +584,19 @@ impl Ui {
...
+    pub fn scroll_to_here(&mut self, center_factor: f32) {
+        let scroll_target = self.min_rect().bottom();
```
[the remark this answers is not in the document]
lucaspoffo: Good point :+1: Didn't knew about that.

egui/src/containers/scroll_area.rs
lucaspoffo: This does not work, as the `clip_rect` is always the same, but I updated the code to be more easy to understand. I hope.

egui/src/align.rs
```diff
@@ -29,6 +29,14 @@ impl Align {
...
         Self::Max
     }
+
+    pub(crate) fn scroll_center_factor(&self) -> f32 {
+        match self {
+            Self::Min => 0.0,
+            Self::Center => 0.5,
+            Self::Max => 1.0,
+        }
+    }
```
lucaspoffo, line 39: I don't know if this is the best place to add this.

emilk: This is the right place! Maybe could be called simply `to_factor`

egui/src/ui.rs
```diff
@@ -584,6 +584,28 @@ impl Ui {
...
+    ///     let scroll_bottom = ui.button("Scroll to bottom.").clicked;
+    ///     for i in 0..1000 {
+    ///         ui.label(format!("Item {}", i));
+    ///     }
+    ///
+    ///     if scroll_bottom {
+    ///         ui.scroll_to_cursor(Align::bottom());
+    ///     }
+    /// });
+    /// ```
```
emilk, line 603: Good example :+1:

emilk (approved):
Awesome - this works great!

PR #83, opened 2020-12-27
Fix nested scrolling.

### LINKED ISSUE #59: Nested scrolling behaves weirdly
OS: MacOS Catalina
Browser: Safari 13.1.3

When multiple scrollable elements are nested, I would expect a deepest one to scroll first, until it will be scrolled completely and then next one kicks in. In the demo, nested scroll views are scrolled simultaneously instead.

Screen Recording 2020-12-07 at 16 32 03 2020-12-07 16_40_06

lucaspoffo:
Fixes #59

Added the scroll_delta to the FrameState so we use this value instead of the actual scroll input, so when an inner ScrollArea is scrolling it will clear the value and the Parent can't use it. When no actual scroll is happening the value is not cleared and the parent scroll may use it.


egui/src/context.rs
```diff
@@ -97,6 +100,14 @@ impl FrameState {
...
         self.used_by_panels = self.used_by_panels.union(panel_rect);
     }
+
+    pub(crate) fn scroll_delta(&self) -> Vec2 {
+        self.scroll_delta
+    }
+
+    pub(crate) fn clear_scroll_delta(&mut self) {
+        self.scroll_delta = Vec2::zero();
+    }
```
emilk, line 110: I'd prefer if `scroll_delta` was just made `pub(crate)` to make the use of it more transparent (and less code!).

egui/src/containers/scroll_area.rs
```diff
@@ -197,9 +197,19 @@ impl Prepared {
...
+        let max_offset = content_size.y - inner_rect.height();
         if ui.rect_contains_mouse(outer_rect) {
-            state.offset.y -= ui.input().scroll_delta.y;
+            let mut frame_state = ui.ctx().frame_state();
+            let scroll_delta = frame_state.scroll_delta();
+
+            let scrolling_top = state.offset.y > 0.0 && scroll_delta.y > 0.0;
+            let scrolling_bottom = state.offset.y < max_offset && scroll_delta.y < 0.0;
+
+            if scrolling_top || scrolling_bottom {
```
emilk, line 208:
```suggestion
            let scrolling_up = state.offset.y > 0.0 && scroll_delta.y > 0.0;
            let scrolling_down = state.offset.y < max_offset && scroll_delta.y < 0.0;

            if scrolling_up || scrolling_down {
```

emilk (changes requested):
Very nice! I just have some minor nits

lucaspoffo:
:+1: Thanks for the suggestion, using pub(crate) is something that I always forget is a thing.

emilk (approved):
Works great - thanks!

PR #84, opened 2020-12-27
Scroll based on the mouse start position.

### LINKED ISSUE #57: Scrollbars behave as if clicked when moving the mouse outside the scrollbar handle
1. Have a long scrollbar.
2. Drag from the end of the scrollbar handle in a direction orthogonal to the scrollbar direction (so horizontal for a vertical scrollbar).
3. Once the mouse moves even slightly outside the scrollbar handle the handle will snap so that the mouse is at the center of the scrollbar handle.

This leads to very jarring movements since while dragging a scrollbar you often move the mouse slightly outside the scrollbar handle. The issue also happens if you move the mouse quickly even in the same direction as the scrollbar.

lucaspoffo:
Fixes #57

Since i was already changing the `ScrollArea`, decided to try fix some reported bugs in it.

The mouse position relative to the center of the handle is saved, so we can use it to check if we are scrolling top or bottom. Now it behaves similar to how the scroll in the browser works.

### DIFF
--- a/egui/src/containers/scroll_area.rs
+++ b/egui/src/containers/scroll_area.rs
@@ -12,8 +12,8 @@ pub(crate) struct State {
     /// Momentum, used for kinetic scrolling
     #[cfg_attr(feature = "serde", serde(skip))]
     pub vel: Vec2,
-    /// Mouse position relative to the center of the handle when started moving the handle.
-    scroll_start_handle: Option<Vec2>,
+    /// Mouse offset relative to the top of the handle when started moving the handle.
+    scroll_start_offset_from_top: Option<f32>,
 }
 
 impl Default for State {
@@ -22,7 +22,7 @@ impl Default for State {
             offset: Vec2::zero(),
             show_scroll: false,
             vel: Vec2::zero(),
-            scroll_start_handle: None,
+            scroll_start_offset_from_top: None,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


egui/src/containers/scroll_area.rs
```diff
@@ -239,17 +242,21 @@ impl Prepared {
...
+                        .scroll_start_handle
+                        .get_or_insert_with(|| mouse_pos - handle_rect.center());
+
+                    let mouse_offset_y = mouse_pos.y - handle_rect.center().y;
+                    let scrolling_bottom = mouse_offset_y > scroll_start_handle.y && delta_y > 0.0;
+                    let scrolling_top = mouse_offset_y < scroll_start_handle.y && delta_y < 0.0;
+
+                    if scrolling_top || scrolling_bottom {
+                        state.offset.y += delta_y * content_size.y / inner_rect.height();
                     }
```
emilk, line 256: I think it could be a lot simpler to instead store the offset from the top of the handle:

``` rust
 let scroll_start_offset_from_top = state
    .scroll_start_offset_from_top
    .get_or_insert_with(|| mouse_pos.y - handle_rect.top());
let new_handle_top = mouse_pos.y - scroll_start_offset_from_top;
state.offset.y = remap(new_handle_top, top..=bottom, 0.0..=content_size.y);
```

emilk (changes requested):
Thanks for taking a look at this! I think it could be simplified quite a bit more

egui/src/containers/scroll_area.rs
lucaspoffo: Starting to understand the greatness of `remap` and `remap_clamp`.

egui/src/containers/scroll_area.rs
```diff
@@ -239,17 +242,15 @@ impl Prepared {
...
-                                ui.input().mouse.delta.y * content_size.y / inner_rect.height();
-                        }
-                    } else {
-                        // Center scroll at mouse pos:
-                        let mpos_top = mouse_pos.y - handle_rect.height() / 2.0;
-                        state.offset.y = remap(mpos_top, top..=bottom, 0.0..=content_size.y);
-                    }
+                    let scroll_start_offset_from_top = state
+                        .scroll_start_offset_from_top
+                        .get_or_insert_with(|| mouse_pos.y - handle_rect.top());
```
emilk, line 247:
```suggestion
                    let scroll_start_offset_from_top = state
                        .scroll_start_offset_from_top
                        .get_or_insert_with(|| {
                            if handle_rect.contains(mouse_pos) {
                                mouse_pos.y - handle_rect.top()
                            } else {
                                0.5 * handle_rect.height() // center handle on mouse
                            }
                        });
```

emilk (changes requested):
One final edge-case: clicking outside of the handle rectangle should make the handle jump to that position. See suggested fix

egui/src/containers/scroll_area.rs
lucaspoffo: Good catch with an easy fix :100:

lucaspoffo: Made a more complete fix, that deal with the cases when the handle centering does not occur, ex: clicking near the inner_rect bottom or top.

emilk (approved):
Works great - thanks!

PR #92, opened 2021-01-03
Make `style` module public

LinuCC:
Allow access to types such as `Visual` to configure the Ui style:

```rust
        // ...
            let mut_style = ui.style_mut();
            mut_style.visuals = egui::style::Visuals {
                 widgets: egui::style::Widgets {
                 // ...
            };
```

Have not found a better solution ¯\_(ツ)_/¯


emilk (approved):
Makes sense!

PR #100, opened 2021-01-07
Make `egui_glium::painter::Painter::paint_job` pub

walterpie:
The background is: I'm working on an engine where I would like to draw the gui and the game to the same `glium::Frame`. Haven't found any other obvious, trivial solution. I could of course write my own integration, but I just think this makes sense to be public.


emilk (approved):
Great!

So far I've just been using `egui_glium` as a stand-alone thing, so it is great to have someone use it as a library.

PR #104, opened 2021-01-10
Text wrap correctly on Chinese chraracter

n2:
There are so few spaces in Chinese text, so a new line can happen after any Chinese character.


egui/src/paint/font.rs
```diff
@@ -426,7 +426,10 @@ impl Font {
             }
 
             const NON_BREAKING_SPACE: char = '\u{A0}';
-            if chr.is_whitespace() && chr != NON_BREAKING_SPACE {
+            let is_ch = |c|{
+                (c >= '\u{4E00}' && c <= '\u{9FA5}') || (c >= '\u{3400}' && c <= '\u{4DB5}') //Chinese
+            };
```
emilk, line 431: Please make this a free function:

```
fn is_chinese(c: char) -> bool { … }
```

emilk (changes requested):
Thanks for the PR!

Would this also make sense for something like Korean?

egui/src/paint/font.rs
n2: Ok, it is done. P.S. like Chinese, Japanese and Korean are both hieroglyphs, so the newline rule may be the same. But I am not familliar with either of them. Maybe it's better to leave them to some experts.

emilk (approved):
Thank you!

PR #107, opened 2021-01-11
Implement InputState::key_down

### LINKED ISSUE #105: input: add InputState::key_down method
Hi, this issue was opened to track adding a `InputState::key_down` that will return if a key is currently being held down. This would be useful for e.g. folks making games using pure `egui`.

This issue was created based on a comment from a previous issue:

> Yeah, I'd be open to have `egui::InputState` track which keys are currently down. It is a very simple change. Feel free to open a separate issue on that (or a PR! a simple `HashSet` should do fine)

_Originally posted by @emilk in https://github.com/emilk/egui/issues/102#issuecomment-758091620_

tangmi:
Closes #105

Please let me know if there's any changes (style or otherwise) requested!

### DIFF
--- a/egui/src/input.rs
+++ b/egui/src/input.rs
@@ -1,5 +1,7 @@
 //! The input needed by Egui.
 
+use std::collections::HashSet;
+
 use crate::{math::*, util::History};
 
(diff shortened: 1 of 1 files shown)


emilk (approved):
Looks great - thanks!

PR #115, opened 2021-01-16
Use a cubic ease function for animations

tangmi:
This changes the bool animation being linear to using a cubic ease-in-out curve. This curve is widely used in user interface animations (ease-in and ease-out also are frequently found) and lets objects movement and sizing feel more natural at the start and end of the animation (e.g. mimicking acceleration). This is less noticeable at small changes or `animation_time` values, but becomes very noticeable when an animation travels a far distance or takes a longer amount of time.

The curve itself looks something like:

image
(from https://www.desmos.com/calculator/tdvd2vd4f1)

Note: This is just a proposal for an alternate animation curve with no immediate desire to merge. It's likely `egui` would still want to expose linear (e.g. for opacity animations) and potentially ease-in and ease-out as well (when the end and start of the animation is obscured).

### DIFF
--- a/egui/src/animation_manager.rs
+++ b/egui/src/animation_manager.rs
@@ -1,6 +1,9 @@
 use std::collections::HashMap;
 
-use crate::{math::remap_clamp, Id, InputState};
+use crate::{
+    math::{clamp, remap_clamp},
+    Id, InputState,
+};
 
 #[derive(Clone, Default)]
 pub(crate) struct AnimationManager {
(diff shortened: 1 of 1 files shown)


emilk:
I have actually experimented with this in the past, and I found that because Egui uses so short animation times the smooth transition squishes the transition time to something so small that the eye has problems noticing it. That is to say, with an ease-in-ease-out animation, most change happens just in the middle third of the animation time, which will just be 2 or 3 frames.

Or in other words: collapsing a collapsing header just looks better (to my eyes) with the current linear blend, especially at lower framerates.

PR #126, opened 2021-01-20
[egui] Stop velocity after overscroll

paulshen:
This stops a `ScrollArea`'s velocity after it has hit an edge. Prior to this change, using the mousewheel after hitting an edge via momentum scrolling led to weird behavior (momentum continued)


emilk (approved):

emilk:
👍

PR #127, opened 2021-01-21
Prefer overlapping row in galley cursor_from_pos

paulshen:
This changes `cursor_from_pos` to prefer the row that the cursor is in. Prior to this change, `cursor_from_pos` would bias towards the previous row. The cursor is equidistant from the previous row's `y_max` and next row's `y_min`.

confirmed that this is how the cursor functions in other apps (eg Chrome textarea). video showing the TextEdit component without this change.

https://user-images.githubusercontent.com/2266187/105284424-47d78d80-5b67-11eb-8260-037e46a446a9.mp4


emilk (approved):
This works great - thanks!

PR #139, opened 2021-01-25
Add support for a HTTP body

PauloMelo:
Hi emilk,

here is the PR.

Regards,
Paulo Melo

### DIFF
--- a/egui_glium/src/http.rs
+++ b/egui_glium/src/http.rs
@@ -7,7 +7,10 @@ pub fn fetch_blocking(request: &Request) -> Result<Response, String> {
 
     let resp = match method.as_str() {
         "GET" => ureq::get(url).set("Accept", "*/*").call(),
-        "POST" => ureq::post(url).set("Accept", "*/*").set("Content-Type", "text/plain; charset=utf-8").send_string(body),
+        "POST" => ureq::post(url)
+            .set("Accept", "*/*")
+            .set("Content-Type", "text/plain; charset=utf-8")
+            .send_string(body),
         _ => return Err("method not implemented".to_string()),
         // TODO: PUT only sends in binary form.
     };
--- a/egui_glium/src/http.rs
+++ b/egui_glium/src/http.rs
@@ -5,15 +5,12 @@ pub use epi::http::{Request, Response};
 pub fn fetch_blocking(request: &Request) -> Result<Response, String> {
     let Request { method, url, body } = request;
 
-    let resp = match method.as_str() {
-        "GET" => ureq::get(url).set("Accept", "*/*").call(),
-        "POST" => ureq::post(url)
-            .set("Accept", "*/*")
-            .set("Content-Type", "text/plain; charset=utf-8")
-            .send_string(body),
-        _ => return Err("method not implemented".to_string()),
-        // TODO: PUT only sends in binary form.
-    };
+    let mut req = ureq::request(method, url).set("Accept", "*/*");
+    if !body.is_empty() {
+        req.set("Content-Type", "text/plain; charset=utf-8");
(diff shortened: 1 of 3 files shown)


egui_web/src/http.rs
```diff
@@ -24,6 +24,10 @@ async fn fetch_jsvalue(request: &Request) -> Result<Response, JsValue> {
     opts.method(method);
     opts.mode(web_sys::RequestMode::Cors);
 
+    if method == &"POST".to_string() {
```
emilk, line 27: `method == "POST"` should work fine

emilk, line 27:
```suggestion
    if !body.is_empty() {
```

`POST` is not the only verb that wants to send data (there is also `PUT`, and maybe others).

egui_web/src/http.rs
```diff
@@ -24,6 +24,10 @@ async fn fetch_jsvalue(request: &Request) -> Result<Response, JsValue> {
...
+        opts.body(Some(&JsValue::from_serde(body).unwrap()));
```
emilk, line 28: `JsValue::from_serde` seems wrong. `JsValue::from` or `JsValue::from_string` or `from_str` or something.

emilk (changes requested):
This also needs to be implemented in `egui_glium/src/http.rs`

The CI wasn't configured to actually test the `http` feature, but that has now been fixed, so please rebase on latest `master` to get this PR tested by the CI.

Also: please test this manually with an actual HTTP request. I would assume most HTTP servers would also require a `Content-Type` header?

egui_web/src/http.rs
PauloMelo: I have tested manually with my personal development http server. I use kore from kore.io. 
By default the Content-Type is plain/text. If we would send binary data then we need to change it, but since web_sys only has the body referenced by JsValue (and I am new to Rust - less than a week) I believe that we need only to send it if we want to change. 

I the GLIUM version since it uses ureq it has the separate functions to send binary, text, form or json.

egui_web/src/http.rs
PauloMelo: Yes I've only seen that after the first commit.

egui_glium/src/http.rs
```diff
@@ -3,9 +3,17 @@ pub use epi::http::{Request, Response};
...
-    let resp = ureq::request(method, url).set("Accept", "*/*").call();
+    let resp = match method.as_str() {
+        "GET" => ureq::get(url).set("Accept", "*/*").call(),
+        "POST" => ureq::post(url)
+            .set("Accept", "*/*")
+            .set("Content-Type", "text/plain; charset=utf-8")
+            .send_string(body),
+        _ => return Err("method not implemented".to_string()),
+        // TODO: PUT only sends in binary form.
+    };
```
emilk, line 16:
```suggestion
    let mut req = ureq::request(method, url).set("Accept", "*/*");
    if !body.is_empty() {
        req.set("Content-Type", "text/plain; charset=utf-8");
        req.send_string(body);
    }
    let resp = req.call();
```

this way we will support the other methods too!

emilk (changes requested):
Starting to look good!

egui_web/src/http.rs
emilk: > I have tested manually with my personal development http server.

Great! Web or glium version or both?

> I am new to Rust - less than a week

Wow! Good work on this PR, and welcome to the wonderful world of Rust!

emilk:
well that last commit was meant to have a proper commit message 🙄  sorry about that :)

emilk:
Thank you!

Issue #143, opened 2021-01-26
Color picker closes immediately on click


EmbersArc:
**Describe the bug**
When picking a color, I'd expect to be able to click and drag within the color picker and select a color. It works that way in the online demo. In my native build however, the popup closes immediately when clicking anywhere in the color picker area.
This works fine in the web demo (online, I wasn't able to compile it myself). Dragging the rest of the popup works fine.

**To Reproduce**
1. Build the demo
2. Open any color picker
3. Click in the colored region but don't release the mouse
4. The popup closes

**Expected behavior**
 I'd expect to be able to click and drag within the color picker and select a color.

**Screenshots**
![test](https://user-images.githubusercontent.com/1352472/105896749-7f3ab400-6017-11eb-993a-0af67af58d6e.gif)

**Desktop (please complete the following information):**
 - OS: Manjaro Linux KDE Plasma

emilk:
This seems to have come with https://github.com/emilk/egui/pull/135 - thanks for the report!

aakamenov:
Hello @emilk and EmbersArc ,

I've submitted a PR for this - #144

emilk:
Fixed by https://github.com/emilk/egui/pull/144

PR #144, opened 2021-01-27
Use any_click() instead of any_pressed() in check

aakamenov:
Hello emilk ,

This is a fix for #143 . Calling `any_click()` instead of `any_pressed()` seems to have fixed the issue. I'm sorry, I'm not sure how to run the web version so I haven't tested on that. Although, as mentioned on the issue, the bug doesn't occur there.

Thanks,
Asparuh


emilk (approved):
Thank you!

Issue #103, opened 2021-01-08
[Web Demo] Blurry rendering when browser window has odd width / height


daniel5151:
Consider this side-by-side comparison of the basic demo pane:

![egui-crisp](https://user-images.githubusercontent.com/2389445/104047741-23b39e00-51b0-11eb-8b65-d813f7bfad38.png) ![egui-blurry](https://user-images.githubusercontent.com/2389445/104047742-24e4cb00-51b0-11eb-8864-15ab3ec5c2c8.png)

The left screenshot was taken in a browser window with dimensions 1000px x 1200px, while the one on the right was taken in a browser window with dimensions 1001px x 1200px. i.e: a difference of one pixel resulted in the entire image becoming significantly blurrier.

A similar issue occurs when the window's height is odd, albeit to a lesser extent.

Tested in Google Chrome Version 87.0.4280.88 on Ubuntu 20.10 with NVIDIA Drivers

emilk:
Good find - I can reproduce in Firefox on Mac too. Maybe a simple fix is to just round down the canvas size to the nearest multiple-of-2.

emilk:
I was gonna fix this now, but can no longer reproduce it. Weird :/

bugadani:
I can :)
![image](https://user-images.githubusercontent.com/977627/104850459-8400b900-58ef-11eb-9c86-b650d54ee8e3.png)
MS Edge,  version 87.0.664.75

emilk:
Anyone want to take a stab at this? I can't reproduce on my machine, so I can't.

I'm guessing we need to add some rounding in https://github.com/emilk/egui/blob/master/egui_web/src/lib.rs#L117:L141

My theory is that we need to round (or rather `floor`) the height and width to the next factor-of-2 in pixel or in point size. Or maybe just round to the nearest integers.

bugadani:
Well for me (at least on firefox, which I don't use, LOL) was to turn off webgl antialiasing in about:config. The option is not available in Edge (chromium) though, so this is only food for thought. Also, my head hurts from staring into blurry text :)

On
![image](https://user-images.githubusercontent.com/977627/105632544-f6274f80-5e53-11eb-9a40-53e7cd9cde0d.png)

Off
![image](https://user-images.githubusercontent.com/977627/105632533-e3ad1600-5e53-11eb-8409-384863ad09f3.png)

crumblingstatue:
[...]

PR #149, opened 2021-01-31
Fix blurry rendering on web demo

aakamenov:
Hi emilk ,

This a fix for #103 . As per your suggestion, rounding both values to the nearest even integer did the trick. I don't know if there is a better way to implement this (I don't have much Rust experience). Could add a function to `emath/src/vec2.rs` that works on the struct , but since it would only be used here and this is a hack anyway, I thought that it would be unnecessary. I've compiled the web demo with the latest `master` branch too.

Thanks,
Asparuh


egui_web/src/lib.rs
```diff
@@ -132,8 +132,26 @@ pub fn resize_canvas_to_screen_size(canvas_id: &str) -> Option<()> {
...
+        canvas_size_points.y -= 1f32;
+    }
+
+    if canvas_size_pixels.x % 2f32 != 0f32 {
+        canvas_size_pixels.x -= 1f32;
+    }
+
+    if canvas_size_pixels.y % 2f32 != 0f32 {
+        canvas_size_pixels.y -= 1f32;
+    }
```
emilk, line 154: This will only work for integers.

``` rust
    fn floor_to_even(v: f32) -> f32 {
        (v / 2.0).floor() * 2.0
    }
```

is a more robust method

emilk (commented):
Thanks for taking a stab at this!

Unfortunately your rounding only works for integers. Furthermore I only publish a new wasm every other week or so (on a new egui release) so as not to pollute the git history with too many large files. Not easy for you to know!

Anyway, instead of reworking this PR I just decided to create another one, please try it out at https://github.com/emilk/egui/pull/151

emilk:
Closed in favor of https://github.com/emilk/egui/pull/151

PR #151, opened 2021-02-01
Fix web blur (take 2)

### LINKED ISSUE #103: [Web Demo] Blurry rendering when browser window has odd width / height

emilk:
Fixes #103 according to @aakamenov

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -27,15 +27,15 @@
             padding: 0 !important;
         }
 
-        /* Center canvas vertically and horizontally: */
+        /* Position canvas in center-top: */
         canvas {
             margin-right: auto;
             margin-left: auto;
             display: block;
             position: absolute;
-            top: 50%;
+            top: 0%;
             left: 50%;
-            transform: translate(-50%, -50%);
+            transform: translate(-50%, 0%);
         }
     </style>
 </head>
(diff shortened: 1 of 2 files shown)


emilk:
@aakamenov please try this out to double-check it works as well as your PR!

I noticed the position of the canvas jumping around when resizing the browser window, so I added a CSS change to always position the canvas at the top of the screen.

aakamenov:
Ah, that's my bad! Since I didn't get the whole WASM debugging story to work, I tried my fix in a Rust pad first and then later forgot to call `.round()` before the `%` operator to make it work with floats. Anyways, your solution is more elegant regardless. 

I tested this and it also fixes the issue, so I guess it can be merged. Also, nice catch on noticing the canvas jumping problem!

emilk:
Thanks again @aakamenov !

PR #152, opened 2021-02-02
Speed up fractal_clock painting

xue-blood:
Save painting call time.
`before`
image

`after`
image

### DIFF
--- a/egui/src/painter.rs
+++ b/egui/src/painter.rs
@@ -116,6 +116,9 @@ impl Painter {
             .add(self.clip_rect, shape)
     }
 
+    /// Add many shapes at once.
+    ///
+    /// Calling this once is generally faster than calling [`Self::add`] multiple times.
     pub fn extend(&self, shapes: Vec<Shape>) {
         if !shapes.is_empty() {
             self.ctx
--- a/egui_demo_lib/src/apps/fractal_clock.rs
+++ b/egui_demo_lib/src/apps/fractal_clock.rs
@@ -13,6 +13,7 @@ pub struct FractalClock {
     length_factor: f32,
     luminance_factor: f32,
     width_factor: f32,
+    line_count: usize,
 }
 
 impl Default for FractalClock {
@@ -26,6 +27,7 @@ impl Default for FractalClock {
             length_factor: 0.8,
             luminance_factor: 0.8,
             width_factor: 0.9,
+            line_count: 0,
         }
     }
 }
@@ -79,6 +81,7 @@ impl FractalClock {
         } else {
             ui.label("The fractal_clock clock is not showing the correct time");
         };
+        ui.label(format!("Painted line count: {}", self.line_count));
 
         ui.checkbox(&mut self.paused, "Paused");
         ui.add(Slider::f32(&mut self.zoom, 0.0..=1.0).text("zoom"));
@@ -128,13 +131,16 @@ impl FractalClock {
         ];
 
         let scale = self.zoom * rect.width().min(rect.height());
-        let paint_line = |points: [Pos2; 2], color: Color32, width: f32| {
+        let mut shapes: Vec<Shape> = Vec::new();
+        let mut paint_line = |points: [Pos2; 2], color: Color32, width: f32| {
             let line = [
                 rect.center() + scale * points[0].to_vec2(),
                 rect.center() + scale * points[1].to_vec2(),
             ];
 
-            painter.line_segment([line[0], line[1]], (width, color));
+            if rect.intersects(Rect::from_two_pos(line[0], line[1])) {
+                shapes.push(Shape::line_segment(line, (width, color)));
+            }
         };
 
         let hand_rotations = [
@@ -199,5 +205,7 @@ impl FractalClock {
 
             std::mem::swap(&mut nodes, &mut new_nodes);
         }
+        self.line_count = shapes.len();
+        painter.extend(shapes);
     }
 }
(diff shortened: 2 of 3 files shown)


xue-blood:
If using the `Liang-Barsky` algorithm  to clip the line which out of the rect before render.
The performance will be more efficiency.
I think it will be fine if this function writen in `painter.rs`.
However It can not handle that many line overlay each other
```
            if self.clip_line(line, rect, self.min_line_length_sq) {
                line_data.push(Shape::line_segment(line, (width, color)));
            }

    fn clip_line(&self, input: [Pos2; 2], rect: Rect, min_len_sq: f32) -> bool {
        let mask_a: u8 = self.clip_line_mask(input[0], rect);
        let mask_b: u8 = self.clip_line_mask(input[1], rect);
        if mask_a == 0 && mask_b == 0 {
            if input[0].distance_sq(input[1]) < min_len_sq {
                false
            } else {
                true
            }
        } else if mask_a & mask_b == 0 {
            if input[0].distance_sq(input[1]) < min_len_sq {
                false
            } else {
                true
            }
        } else {
            false
        }
    }

    fn clip_line_mask(&self, p: Pos2, rect: Rect) -> u8 {
        let mut mask: u8 = 0;
        let top = rect.top();
        let btm = rect.bottom();
        let lft = rect.left();
        let rgt = rect.right();
        if p.y < top {
            mask |= 0b1000;
        }
        if p.y > btm {
            mask |= 0b0100;
        }
        if p.x < lft {
            mask |= 0b0010;
        }
        if p.x > rgt {
            mask |= 0b0001;
        }
        mask
    }
```
image

egui/src/painter.rs
```diff
@@ -116,6 +116,17 @@ impl Painter {
             .add(self.clip_rect, shape)
     }
 
+    /// Add shape list to painting in one call
+    /// Speed up shape painting by reduce call of graphics.lock()
+    pub fn add_vec(&self, shapes: Vec<Shape>) {
```
emilk, line 121: No need for this, this is exactly what `fn extend` below it does.

egui_demo_lib/src/apps/fractal_clock.rs
```diff
@@ -129,13 +132,14 @@ impl FractalClock {
...
-        let paint_line = |points: [Pos2; 2], color: Color32, width: f32| {
+        let mut line_data: Vec<Shape> = Vec::new();
+        let mut paint_line = |points: [Pos2; 2], color: Color32, width: f32| {
             let line = [
                 rect.center() + scale * points[0].to_vec2(),
                 rect.center() + scale * points[1].to_vec2(),
             ];
-
-            painter.line_segment([line[0], line[1]], (width, color));
+            //painter.line_segment([line[0], line[1]], (width, color));
```
emilk, line 141: rm dead code

egui_demo_lib/src/apps/fractal_clock.rs
```diff
@@ -129,13 +132,14 @@ impl FractalClock {
         ];
 
         let scale = self.zoom * rect.width().min(rect.height());
-        let paint_line = |points: [Pos2; 2], color: Color32, width: f32| {
+        let mut line_data: Vec<Shape> = Vec::new();
```
emilk, line 135: I usually call this `shapes`

egui/src/painter.rs
```diff
@@ -116,6 +116,17 @@ impl Painter {
...
+        let mut graphics = self.ctx.graphics();
+        let paint_list = graphics.list(self.layer_id);
+
+        for shape in shapes {
+            paint_list.add(self.clip_rect, shape);
+        }
+    }
+
     pub fn extend(&self, shapes: Vec<Shape>) {
```
emilk, line 130:
```suggestion
    /// Add many shapes at once.
    ///
    /// Calling this once is generally faster than calling [`Self::add`] multiple times.
    pub fn extend(&self, shapes: Vec<Shape>) {
```

emilk (changes requested):

egui_demo_lib/src/apps/fractal_clock.rs
xue-blood: got it

egui_demo_lib/src/apps/fractal_clock.rs
```diff
@@ -200,5 +206,33 @@ impl FractalClock {
...
+        if p.y > btm {
+            mask |= 0b0100;
+        }
+        if p.x < lft {
+            mask |= 0b0010;
+        }
+        if p.x > rgt {
+            mask |= 0b0001;
+        }
+        mask
```
emilk, line 236:
```suggestion
```

I rather not make this longer or complex. This is a demo of painting with egui, not a demo of performance.

Also `clip_line` is misleading: this is not doing [line clipping](https://en.wikipedia.org/wiki/Line_clipping) but [view culling](https://en.wikipedia.org/wiki/Hidden-surface_determination#Viewing-frustum_culling).

 If you really want this (and it really improves perf a lot) then please make it dead simple with `rect.instersects(Rect::from_points(input))`

emilk (changes requested):

egui_demo_lib/src/apps/fractal_clock.rs
xue-blood: Yes it is more brief.

emilk (approved):

emilk:
Thanks!

xue-blood:
You are welcome.

PR #160, opened 2021-02-05
Discard obviously absurd window positions.

phoglund:
Looks like winit sometimes writes pos x=-32000.0, y=-32000.0 on multiscreen setups. Clamp this to be at least 0 when we load the window.

### DIFF
--- a/egui_glium/src/window_settings.rs
+++ b/egui_glium/src/window_settings.rs
@@ -72,7 +72,13 @@ impl WindowSettings {
         //         height: size.y as f64,
         //     });
 
-        if let Some(pos) = self.pos {
+        if let Some(mut pos) = self.pos {
+            if pos.x < 0.0 {
+                pos.x = 0.0;
+            }
+            if pos.y < 0.0 {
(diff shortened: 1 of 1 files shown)


emilk:
On some platforms the (0, 0) coordinate is the top-left corner of the *primary* screen. So if you have two screens and the right is the primary one, then the left screen will have negative X coordinates. This is true on my Mac, and maybe other platforms too.

Clamping the window position to the screens makes sense to me, but this is not the way to do it.

I think using https://docs.rs/winit/0.24.0/winit/window/struct.Window.html#method.available_monitors we could get a list of all monitor coordinates and then clamp window position to be inside at least one of them.

I've opened an issue on winit on this topic: https://github.com/rust-windowing/winit/issues/1851

phoglund:
Ok! I wonder if I should use DummyStorage in my app as workaround for now so this doesn't hit my users too much?

emilk:
That sounds reasonable. Persistence and window positioning are both quite under-cooked features, so it is probably wise not to rely on them too much :)

Issue #80, opened 2020-12-25
egui_glium only supports GLSL 1.40


emilk:
As discussed [here](https://github.com/emilk/egui/discussions/79#discussioncomment-241235), `egui_glium` currently only supports GLSL 1.40, which apparently precludes it from working on Windows 7.

The fix would probably include using the [`glium::program!` macro](https://docs.rs/glium/0.29.0/glium/macro.program.html).

emilk:
Fixed in https://github.com/emilk/egui/pull/187

PR #187, opened 2021-02-20
Added shaders on GLSL 1.2

katyo:
- Used `glium::program` to create shaders
- Moved shaders code to its own sources and include it as str
- Added shaders implementation on GLSL which allows run egui on old hardware
  (Raspberry Pi 1/zero in game again)

It should solve #79 #80.
Another shader versions (1.1, 1.3, 1.0 ES and etc.) still missing and can be added later.

### DIFF
--- a/egui_glium/CHANGELOG.md
+++ b/egui_glium/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+* Added GLSL 1.2 shaders (now `egui` works well on old hardware which supports OpenGL 2.1 only like Raspberry Pi 1 and Zero)
+* Added GLSL ES 1.0 and 3.0 shaders (by reusing from webgl backend)
+
 
 ## 0.9.0 - 2021-02-07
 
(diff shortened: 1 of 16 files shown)


katyo:
I'm not an expert in GLSL dialects but seems the shader's code from `webgl1` and `webgl2` can be reused for GLSLES 1.0 and 3.0 respectively.

emilk (approved):
Awesome, thank you! Please also add a line to egui_glium/CHANGELOG.md

emilk (approved):

emilk:
@katyo did you test both `fragment_300es.glsl` and `fragment_100es.glsl` ?

I ask because they seem to be carbon-copied of the WebGL shaders, including statements like "WebGL doesn't come with sRGBA textures" (but does GLES?) and "WebGL doesn't support linear blending in the framebuffer" (but does GLES?).

This is particularly relevant since I recently improved the WebGL shaders: https://github.com/emilk/egui/commit/effd3c7440d9f70c0eff677871a60be539cf1113

katyo:
@emilk Seems the version string is wrong for GLSL ES 1.00 shaders. According to specs it should be `#version 100` without `es` at end. I forget to fix this.
I cannot test GLSL ES 3.0 version just now, but 1.00 works fine (with version fix).

Issue #173, opened 2021-02-11
Multicolor TextEdit widgets (for Syntax Highlighting etc.)


the-drunk-coder:
**Is your feature request related to a problem? Please describe.**
I'm working on a little code editor with egui, and of course it'd be nice to have syntax highlighting ...

**Describe the solution you'd like**
A TextEdit widget that allows to pass a color map.

As a matter of fact I already implemented multicolor methods in `tesselator.rs` and `painter.rs` in my egui fork (which I otherwise keep aligned with the latest version): 

https://github.com/the-drunk-coder/egui/blob/master/epaint/src/tessellator.rs#L730
https://github.com/the-drunk-coder/egui/blob/master/egui/src/painter.rs#L324

Works fine, but I'm not sure if that's how you'd do it ...

emilk:
Interesting! I think this could be a very useful feature with not much work. So do you envision something like `ui.add(TextEdit::multiline(my_text).colormap(my_highlighting));` then? If the user enters text then the colormap will be wrong that frame, but maybe having the colors be off by a few characters for a frame is fine.

I wonder if such a feature should be extended to also cover things like background color, underline, _italics_ and ~strikethrough~, though arguably that is a lot more work for something a lot less useful.

Buy the way, I have an opt-in feature in the demo app which applies syntax-highlighting to the results of HTTP requests using the [`syntect`](https://github.com/trishume/syntect) crate: <https://github.com/emilk/egui/blob/master/egui_demo_lib/src/apps/http_app.rs#L197-L251>. What are you using for syntax highlighting?

the-drunk-coder:
Well the final color map is calculated every frame and maps character positions to text color changes. It's then passed to the painter method mentioned above. 

The color map on the higher level that is passed to the text edit widget are mappings from function names, keywords etc. to their respective colors.

To generate the final map so far i just look for the positions of the keywords etc, and calculate the color changes from that.

emilk:
I think having the colormap as a `BTree<usize, Color32>` works, but and ordered `Vec<(usize, Color32)>` might be more efficient, but also harder to use. In any case a wrapper type is in order: `struct Colormap(…)` and pass that on to the tessellator and painter like in your fork. Feel free to open a PR!

Pebaz:
I'd like to express my interest for a syntax highlighting widget also! I think the example in the demo is perfect, but I would prefer if it were standardized rather than having to custom make it. Either way thanks for the example it's really cool! :)

[...]

PR #195, opened 2021-02-27
Text color map (Issue #173)

the-drunk-coder:
Please read the "Making a PR" section of [`CONTRIBUTING.md`](https://github.com/emilk/egui/blob/master/CONTRIBUTING.md) before opening a Pull Request!


epaint/src/shape.rs
```diff
@@ -49,6 +49,15 @@ pub enum Shape {
         /// If true, tilt the letters for an ugly italics effect
         fake_italics: bool,
     },
+    MulticolorText {
+        /// Top left corner of the first character.
+        pos: Pos2,
+        /// The layed out text
+        galley: Galley,
+	text_style: TextStyle, // TODO: Font?
+	color_map: TextColorMap,
```
emilk, line 58: No need to add a separate `MulticolorText` - just add `color_map: TextColorMap` to `Text`, and set it to `TextColorMap::default()` when there are no color changes

epaint/src/tessellator.rs
```diff
@@ -705,6 +725,74 @@ impl Tessellator {
         }
         assert_eq!(chars.next(), None);
     }
+    
+    pub fn tessellate_multicolor_text(
```
emilk, line 729: I rather have just one `tesselate_text`. With `TextColorMap::default()` they should act the same anyway.

emilk (changes requested):
Thank you for this PR! Hot tip: configure your editor to run `cargo fmt` on save :)

emilk:
I'll convert this to a draft and you can change it to "Ready for review" when it is green

the-drunk-coder:
I removed the additional `tesselate_multicolor_text` etc. I only kept the multicolor galley because there were already some methods in place, so I thought that was in line ...

emilk (changes requested):
This is starting to look good!

What I think is missing is some way to use the color map, and a demonstration of it in the demo app. I think at least `Label`, and maybe `TextEdit`, should have an option to set a `TextColorMap`. It would be great to demonstrate it with some simple demo, at least for `Label`. It can be added to `egui_demo_lib/src/apps/demo/demo_window.rs`.

...or if you are feeling very ambitious, check out the syntax highlighting in `egui_demo_lib/src/apps/http_app.rs` which currently uses a bunch of labels, but could be just a single `Label` with a `TextColorMap`. To test that use `cargo run --all-features -p egui_demo_app` (you need to rebase on `master` first). Totally optional though!

emilk:
Oh, and please add a line to `CHANGELOG.md`!

Sorry for the slow review btw

the-drunk-coder:
No worries, got plenty of things on my plate as well. 

So, would you prefer having separate methods for the galleys in `Painter` ? Like, `galley_with_italics` and `galley_with_italics_multicolor` ? Or rather just one ? There's plenty of galley methods, so I'm not quite sure here ... personally I'd make separate ones ...

the-drunk-coder:
Also, now that I think about it, at least for `TextEdit`, I'm not sure whether it makes sense to set a static color map ? I mean, the way I use it it's generated dynamically based on the content.

epaint/src/text/fonts.rs
```diff
@@ -139,7 +142,7 @@ impl Default for FontDefinitions {
             fonts_for_family.insert(
                 FontFamily::Monospace,
                 vec![
-                    "ProggyClean".to_owned(),
+                    "ComicMono".to_owned(),
```
emilk, line 145: please revert!

egui/src/painter.rs
```diff
@@ -3,9 +3,9 @@ use crate::{
     layers::{LayerId, PaintList, ShapeIdx},
     Color32, CtxRef,
 };
-use epaint::{
+use epaint::{
```
emilk, line 6: Configure your editor to trim trailing spaces!

emilk (commented):
This needs a demo!

For instance, if `Label` was given a method to set a color map, then `egui_demo_lib/src/apps/http_app.rs` could be configured to use it for syntax highlighting (requires `syntect` feature) in `ColoredText::ui`.

emilk:
Closed in favor of https://github.com/emilk/egui/pull/682 which is a complete rewrite of the text layout engine which supports color ranges and much, much more.

PR #199, opened 2021-03-01
Texture as free 3d graphics rendering area

t18b219k:
I implemented texture register function that enables image as free 3d graphics drawing area in wgpu backend .
and it extends capability of egui so decided to port this functionality for glium backend .


emilk:
Sorry for not reviewing this earlier, I've been focused on other things. I'll start now!

emilk:
The HTTP changes are quite separate from glium texture changes - I'd prefer if they were two PR:s.

Please also remove the `.wasm` and `.js` build artifacts. `git` is not great at storing large files like these, so we only commit a new web demo after a new egui release.

t18b219k:
ok i will fix it

egui_http_web/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "egui_http_web"
```
emilk, line 2: If possible I'd prefer if you could combine `egui_http_web` and `egui_http_native` to one crate (`epi_http`) instead, with different dependencies and code paths depending on platform.

egui_web/src/webgl1.rs
```diff
@@ -196,11 +205,16 @@ impl WebGlPainter {
...
         for v in &mesh.vertices {
             positions.push(v.pos.x);
             positions.push(v.pos.y);
             tex_coords.push(v.uv.x);
-            tex_coords.push(v.uv.y);
+            if let Some(true) = is_offscreen {
+                tex_coords.push(1.0 - v.uv.y);
+            } else {
+                tex_coords.push(v.uv.y);
+            }
```
emilk, line 217: Why store the egui texture in another direction from other textures? How are you loading these other images that makes them appear upside-down?

Also, do you know you can flip images with https://docs.rs/egui/0.10.0/egui/widgets/struct.Image.html#method.uv ?

emilk (changes requested):

egui_web/src/webgl1.rs
t18b219k: okay i didnt see it.
 remove off screen flag but register function still alive

egui_http_web/Cargo.toml
t18b219k: okay i will change it

t18b219k:
refork your origin and rework it

Issue #185, opened 2021-02-20
Configurable label text style for CollapsingHeader


mvlabat:
**Is your feature request related to a problem? Please describe.**
I'd like to have a possibility to configure `CollapsingHeader` label style.

**Describe the solution you'd like**
Probably, just a new setter method for `text_style`.

**Additional context**
 In my game, I'd like to display some preview data for a component in the `CollapsingHeader` label itself, and in that particular case it's a vector, so I'd like to use the monospace font.

emilk:
Sounds like a good idea, and an easy PR!

PR #200, opened 2021-03-02
Configurable label text style for CollapsingHeader

baysmith:
Fixes emilk/egui#185


emilk (approved):
sorry for the slow review. and thanks!

PR #208, opened 2021-03-06
fix: Modern browsers encode Space as ' '

zeroeightysix:
Related to #65 and #31 

egui currently checks for `"Space"` for spaces, while modern browsers prefer to use `" "`. 
This can be checked [here](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values). The documentation also mentions:

> [1] While older browsers used words like "Add", "Decimal", "Multiply", and so forth modern browsers identify these using the actual character ("+", ".", "*", and so forth).

So, for backwards compat, I've left the `"Space"` check in.


emilk (approved):
Thanks!

PR #213, opened 2021-03-08
Derive (de)serialize for Key

zeroeightysix:
Sorry for another one-line PR.

In my application I'd like to add custom user controls to the persistent state. 
I could use serde's [workaround](https://serde.rs/remote-derive.html) for implementing (de)serialize on foreign types. But, considering egui already depends on serde, and this usecase shouldn't be too uncommon, I figured it's not a bad idea to just derive it here.


emilk (approved):
Makes sense - thanks!

PR #216, opened 2021-03-09
DragValue: fix crash for speed==0.0

### LINKED ISSUE #215: "Sliders" demo crashes when left==right
Open "Sliders" demo
Move left&right slider to the same value
*CRASH*

NNemec:
Fixes #215


emilk (approved):
Thanks!

PR #221, opened 2021-03-10
Replace clipboard with copypasta

### LINKED ISSUE #146: build failures on docs.rs
You're getting build failures on docs.rs for eframe (and maybe others?)  Here is a link to the log file:

https://docs.rs/crate/eframe/0.8.0/builds/338677

I didn't see anything else that jumped out at me, but I'm not as familiar with the project as I wish I was.

parasyte:
- Fixes #146


emilk (approved):
Awesome, thank you!

PR #223, opened 2021-03-10
Restrict window move to title bar

NNemec:
I found it strange that windows can be moved by grabbing dragging anywhere within the window are except those parts covered by widgets that react to dragging. This patch changes the behavior to allow moving only when dragging the title bar, like most other window systems I know of.

Caveat 1: windows without title bar are not movable at all.
Caveat 2: areas that are not part of a window (e.g. popups like the color picker) still have the old behavior.

This was mostly an exercise for me to learn more about the internals of egui. Feel free to bring ideas how the caveats could best be addressed or how the behavior change needs to be otherwise modified/completed.

### DIFF
--- a/egui/src/containers/window.rs
+++ b/egui/src/containers/window.rs
@@ -263,6 +263,12 @@ impl<'open> Window<'open> {
 
         let title_content_spacing = 2.0 * ctx.style().spacing.item_spacing.y;
 
+        let title_bar_height = if with_title_bar {
+            title_label.font_height(ctx.fonts(), &ctx.style()) + title_content_spacing
+        } else {
+            0.0
+        };
+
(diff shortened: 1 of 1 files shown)


optozorax:
I believe this should be a configurable option in the window parameters, not enabled or disabled by default.

NNemec:
Adding an option is the easy part. Far more important is the design decision on the default behavior, which, as I believe, very few people will ever bother to change. Which should it be - start move only on title bar or on entire window?

optozorax:
Only I can say that this is the default behavior for Dear ImGui ([proof](https://pbrfrat.com/post/imgui_in_browser.html)).

emilk:
Hi! I like being able to drag any part of the window to move it. If this is causing some problem, please open an issue on it!

Issue #194, opened 2021-02-27
Custom user textures for egui_glium


SparkyPotato:
Currently, egui_glium supports uploading custom user textures from the CPU side. However, if the user is doing any sort of OpenGL rendering on their own (like a viewport in a scene editor), there is no way to pass on the `glium` textures to the `Painter`.
Currently, the only way is to readback the texture onto the CPU, and then reload it every frame, which is terrible for performance.

What could be done is allowing the user to add an `SrgbTexture2d` to the `Painter`, and then get a reference to it. They can then use the reference to render to.

emilk:
Sounds like a good idea! Feel free to open a PR for it

t18b219k:
some backends support this feature e.g. egui_wgpu_backend , egui_ash_vk_mem

t18b219k:
I wrote function and I'm writing sample

SparkyPotato:
A simple way to make it work for the glium backend would be to just make the vector of textures public.

t18b219k:
create rental function ?

t18b219k:
I wrote sample and test it but rendered image was flipped

t18b219k:
ok i fixed image flipping problem so i'm entering refining

t18b219k:
ok i will make PR

t18b219k:
#199 is that i implemented this issue

emilk:
Closed by https://github.com/emilk/egui/pull/226

PR #226, opened 2021-03-11
Add functions to register textures in egui_web and egui_glium

t18b219k:
#194  related PR but platform dependent feature therefore you need some abstract for glium texture alloc and webgl texture alloc


egui_web/src/webgl2.rs
```diff
@@ -192,7 +192,17 @@ impl WebGl2Painter {
             }
         }
     }
-
```
emilk, line 195: next time please don't remove the blank lines between function. I like them for readability :)

emilk (approved):
thanks

Issue #147, opened 2021-01-29
Text editing on mobile


emilk:
`egui_web` fails to bring up the virtual keyboard when editing text.

You would think a modern browser would have a function call for "show a virtual keyboard!" but of course nothing is ever that simple in web-tech.

Related issue: https://github.com/emilk/egui/issues/8

n2:
This problem also occurs on the browser on PC when input method(IME) is needed, it may be caused by the incompatibility of the IME. The virtual keyboard on the mobile device is also an IME window, too. The input candidate window on PC or virtual keyboard on mobile device will pop up only when the browser’s (or native app's) input control gets focus. As we know, from the perspective of the native app or  browser, Egui's "text widgets" are nothing more than a set of triangles, not input controls, so there will be no input candidate window or virtual keyboard. Egui_glium can communicate with IME correctly (but not yet complete), because it's window handler, winit, can connect to the input method service of the system and exchange data with it. But as far as I know, there is no similar method on the browser.

parasyte:
The only way (that I know of) to open an IME window in a browser is by interacting directly with an input element. Some projects like [CanvasInput](https://github.com/goldfire/CanvasInput) use a hidden input element on the DOM to work.

emilk:
@parasyte thanks for that reference! The relevant code seems to be https://github.com/goldfire/CanvasInput/blob/master/CanvasInput.js#L125 and onwards. The problem for egui_web is that we don't want to always have a hidden `input` element on the canvas, or else we'd get the onscreen keyboard popup where-ever the user clicked. I guess we need something like "if the user is hovering over a text fields, add an invisible input element. If the user stops hovering, remove it" (or alternative move it offscreen or something).

This sounds exactly like the kind of hacky web code that I'd rather not have to write. If we can find some other option that would be great, but otherwise I guess this is the way. Any volunteers?

jgrussell:
@emilk,

Can you give us some idea of your prioritization of this issue?

If not for this issue, egui would be my hands down pick for a Rust GUI framework.

Sadly, I do not currently have the Rust skills to tackle this issue myself.

Thank you very much for all of the work you have already done here.  I am already very impressed with this project and hope to actually be able to contribute back in addition to utilizing it in my own projects.

parasyte:
[...]

PR #234, opened 2021-03-18
epaint: default_fonts replaced by custom_fonts

n2:
Replace epaint: "default_font" to "custom_fonts"(not in default-feature group), 
and expose it to egui, and hence to egui_glium and egui_web.
By doing so, the users can choose font file freely without wasting wasm size.


emilk:
The CI error comes from it using the `--all-features` flag.

It is a bit unfortunate that `cargo run --all-features` won't work with this approach.

Can't this be solved by adding the `default_font` option to all `Cargo.toml`:s instead?

n2:
Aye, I was wondering why CI failed. This PR will be closed. My work on [this issue](https://github.com/emilk/egui/issues/147) is almost done, I would like to commit that first.

Issue #62, opened 2020-12-12
Text editing should be over graphemes


emilk:
As pointed out by @norcalli  in https://github.com/emilk/egui/commit/861573c4975e362535b14cf0555fbcb5b8a6d5aa#r44948551

Egui currently handles text layout and cursors using character boundaries, but using graphemes is the better approach in the long run.

Issue #56, opened 2020-11-30
Improve text


emilk:
Tracking issue for improved text rendering

* [ ] Wrapping text mid-word (https://github.com/emilk/egui/issues/55)
* [ ] Emoji
* [ ] Right-to-left scripts
* [ ] "All" of Unicode

A few links ([thanks](https://github.com/emilk/egui/pull/22#issuecomment-735303161) @parasyte):

* https://raphlinus.github.io/rust/skribo/text/2019/02/27/text-layout-kickoff.html
* https://yeslogic.com/blog/allsorts-rust-font-shaping-engine.html
* https://kas-gui.github.io/blog/why-kas-text.html

msklywenn:
I think storing the text has a signed distance field instead of a simple white on black bitmap should also be considered. (scales better and uses less memory)
https://steamcdn-a.akamaihd.net/apps/valve/2007/SIGGRAPH2007_AlphaTestedMagnification.pdf

It also allows for all sorts of effects (outline, shadows...)

emilk:
@msklywenn SDF text has some benefits, like being able to scale text, but come with the downside that the Egui fragment shader will no longer be a simple texture sampler. Currently the same shader supports showing text as well as images. If we switch to a distance field font we would need a more complicated shader taking in both a distance field texture and a normal texture to conveniently support both use cases (or have two shaders and switch between them which comes with its own downsides). I'd like to keep Egui integration as simple as possible for now.

I would also dispute that SDF fonts use less memory - that is only really true for when one want to write really large characters (which Egui doesn't really need). For small characters SDF:s actually tend to use more texels (margin for the blurry "glow").

msklywenn:
Basic SDF rendering only needs alpha testing. To have one shader compatible with bitmap and SDF rendering, a solution would be to have an offset be sent to the shader and applied before discarding. With an offset of 0, discarding would never happen and bitmap display would occur just like it does now. With a -0.5 offset, negative alpha values would be discarded. That offset could then also be negated and added so that the pixels that pass don't look alpha translucent.

In my experience, SDF looks better even at low resolution as it can be thought of a cheap antialiasing. Also, the scaling of SDF resolves the problem of hi-DPI screens, it is not just to render big texts.

emilk:
On low-DPI screens alpha-tested text will look horrible, but I take your point of having a single texture sampler and the programmatically controlling whether to treat it as an SDF or as a color image. Still, there is a lot of complexity here:

[...]

PR #246, opened 2021-03-22
Tab identation for multiline text edit

### LINKED ISSUE #73: Allow tabs for indentation in text edit widget
Hi, 

it seems like the Tab key is not captured in the text edit widget, could that be ? 

Like, when I put some debug output in the event handling loop, it seems like there's no event emitted at all ... just the outer frame flickers ... 

Is there a way to capture the tab key ?

Best,
N

DrOptix:
With this PR we resolve #73, allowing opt-in tab indentation in `TextEdit`

When we load a string in `TextEdit` we keep the tabs as they are and by default `tab_as_space` is initialized to `false` and `tab_moves_focus` is initialized to `true`. 

To config `tab_as_spaces` and `tab_moves_focus` to the values we want we can use the following helper functions:

- `TextEdit::tab_as_spaces()`
- `TextEdit::tab_moves_focus()`

Example:
```rust
ui.add(egui::TextEdit::multiline(&mut multiline_text_input)
    .tab_as_spaces(self.tab_as_spaces)
    .tab_moves_focus(self.tab_size));
```

Now depending on `tab_as_spaces` and `tab_moves_focus` properties of the `TextEdit` we have the following situations:

- When `tab_as_spaces` is false, then when we press `TAB` the result is a `TAB` char is added to the string with the width of 4 spaces.

- When `tab_as_spaces` is true, then when we press `TAB` the result is a chunk of 4 spaces are added to the string.

To fully understand how to use `TextEdit` with those two opt-in behaviors check the "Misc Demos" demo.

Other two helper functions were added to `Ui` and `TextEdit` to simplify the creation of a `TextEdit` focused on code editing:
- `code_editor`
- `code_editor_with_config`

The following key combinations are supported. Depending on the "tab-as-spaces" config the indentation block will be firstly converted to the appropriate type and the increased or decreased.
- `tab`
  Insert `\t` or needed number of spaces to align to the tabulation size. `\t` char will be rendered so that it will align to the tabulation size
- `shift`+`tab`
  Will remove indentation at the end of the indentation block to align the text to the tabulation size. This works on the current line or with multiple lines selected.
- `shift`+`[` and `shift`+`]`

Remove and add indentation at the end of the indentation block to align the text to the tabulation size.

Another improvement was done to `build_demo_web.sh` script to adapt the "open" command to the platform we work on. This is not covering all possible platforms, but it's a good template to have for when we need to adapt to a new platform:
- `xdg-open` for Linux, Fedora for example
- `start` for Windows, MSYS
- `open` for anything else

egui_code_editor


build_demo_web.sh
```diff
@@ -32,4 +32,13 @@ wasm-bindgen "target/wasm32-unknown-unknown/$BUILD/$TARGET_NAME" \
...
 
-open http://localhost:8888/index.html
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+  # Linux, ex: Fedora
+  xdg-open http://localhost:8888/index.html
+elif [[ "$OSTYPE" == "msys" ]]; then
+  # Windows
+  start http://localhost:8888/index.html
+else
+  # Darmin aka MacOS or something else
```
emilk, line 42: nit:
```suggestion
  # Darwin aka MacOS or something else
```

build_demo_web.sh
```diff
@@ -32,4 +32,13 @@ wasm-bindgen "target/wasm32-unknown-unknown/$BUILD/$TARGET_NAME" \
...
+  open http://localhost:8888/index.html
+fi
```
emilk, line 44: Nice, this is a good improvement ⭐

egui/src/memory.rs
```diff
@@ -128,8 +128,12 @@ pub(crate) struct Focus {
     /// The last widget interested in focus.
     last_interested: Option<Id>,
 
+    /// Set at the beginning of the frame, set to `false` when "used".
```
emilk, line 131:
```suggestion
    /// If `true`, pressing tab will NOT move focus away from the current widget.
```

egui/src/ui.rs
```diff
@@ -883,6 +883,25 @@ impl Ui {
...
+
+    /// A `TextEdit` for code editing with configurable `Tab` management.
+    ///
+    /// Se also [`TextEdit::code_editor_with_config`].
+    pub fn code_editor_with_config(
+        &mut self,
+        text: &mut String,
+        tab_as_spaces: bool,
+        tab_moves_focus: bool,
+    ) -> Response {
```
emilk, line 901: It is bad form to have multiple boolean arguments, as it leads to code that is very hard to decipher:

``` rust
TextEdit::code_editor_with_config(code, true, false) // what does it mean!?
```

I think it is better in this case to just have the opinionated `code_editor`

egui/src/ui.rs
```diff
@@ -883,6 +883,25 @@ impl Ui {
         self.add(TextEdit::multiline(text))
     }
 
+    /// A `TextEdit` for code editing.
+    ///
+    /// Se also [`TextEdit::code_editor`].
```
emilk, line 888:
```suggestion
    /// A `TextEdit` for code editing.
    ///
    /// This will be multiline, monospace, and will insert tabs instead of moving focus.
    ///
    /// See also [`TextEdit::code_editor`].
```

egui/src/widgets/text_edit.rs
```diff
@@ -182,9 +186,71 @@ impl<'t> TextEdit<'t> {
...
         }
     }
 
+    /// Registers if this widget will insert spaces instead of tab char
+    ///
+    /// ```rust, ignore
+    /// ui.add(egui::TextEdit::multiline(&mut self.multiline_text_input)
+    ///     .tab_as_spaces(true));
+    /// ```
+    pub fn tab_as_spaces(mut self, b: bool) -> Self {
```
emilk, line 200: I'm not so sure about this feature, since it is only "half" supported. I think most people would expect that if this is set, then backspace would delete four spaces (it currently doesn't).

Personally I'd be happy to just support tabs, but if we should support spaces let's go all the way and make sure that you can also delete them four at a time!

egui/src/widgets/text_edit.rs
```diff
@@ -182,9 +186,71 @@ impl<'t> TextEdit<'t> {
...
+    /// Build a `TextEdit` focused on code editing with configurable `Tab` management.
+    ///
+    /// Shortcut for:
+    /// ```rust, ignore
+    /// egui::TextEdit::multiline(code_snippet)
+    ///     .code_editor()
+    ///     .tab_as_spaces(tab_as_spaces)
+    ///     .tab_moves_focus(tab_moves_focus);
+    /// ```
+    pub fn code_editor_with_config(self, tab_as_spaces: bool, tab_moves_focus: bool) -> Self {
```
emilk, line 248: Same comment here about multiple boolean arguments. If you want to set multiple values with a single call, it is better to make a config struct and pass that in (`struct CodeConfig { tabs_as_spaces: bool, tab_width: usize, … } `). This has the benefit of being more explicit and more extensible.

egui/src/widgets/text_edit.rs
```diff
@@ -432,12 +502,34 @@ impl<'t> TextEdit<'t> {
...
+                        if multiline {
+                            let mut ccursor = delete_selected(text, &cursorp);
+
+                            if ui.memory().has_lock_focus(id) {
+                                if tab_as_spaces {
+                                    insert_text(&mut ccursor, text, "    ");
+                                } else {
+                                    insert_text(&mut ccursor, text, "\t");
+                                }
+                            }
```
emilk, line 526: It would be great to also implement shift-tab (delete leading tab/spaces from the current line). A bit more work though due to the complicated text cursor system.

egui_demo_lib/src/apps/demo/widgets.rs
```diff
@@ -23,7 +23,11 @@ pub struct Widgets {
     angle: f32,
     color: Color32,
     single_line_text_input: String,
+    tab_size: usize,
```
emilk, line 26: rm: unused

egui_demo_lib/src/apps/demo/widgets.rs
```diff
@@ -35,7 +39,22 @@ impl Default for Widgets {
...
             single_line_text_input: "Hello World!".to_owned(),
-            multiline_text_input: "Text can both be so wide that it needs a line break, but you can also add manual line break by pressing enter, creating new paragraphs.\nThis is the start of the next paragraph.\n\nClick me to edit me!".to_owned(),
+
+            tab_size: 4,
+            tab_as_spaces: true,
+            tab_moves_focus: false,
+            multiline_text_input: r#"Text can both be so wide that it needs a line break, but you can also add manual line break by pressing enter, creating new paragraphs.
+This is the start of the next paragraph.
+
+Existing	tabs	are	kept	as	tabs.
```
emilk, line 49: remove or move to the `code_snippet` demo instead

egui_demo_lib/src/apps/demo/widgets.rs
```diff
@@ -133,7 +152,26 @@ impl Widgets {
...
+
+        ui.separator();
+
+        ui.horizontal(|ui| {
+            ui.label("Code editor:");
+
+            ui.separator();
+
+            ui.checkbox(&mut self.tab_as_spaces, "Tabs as spaces");
+            ui.checkbox(&mut self.tab_moves_focus, "Tabs moves focus");
```
emilk, line 168:
```suggestion
            ui.checkbox(&mut self.tab_moves_focus, "Tabs moves focus");
            if !self.tab_moves_focus {
                ui.checkbox(&mut self.tab_as_spaces, "Tabs as spaces");
            }
```

(the `tab_as_spaces` is meaningless if tab moves focus)

emilk (changes requested):
Nice work!

build_demo_web.sh
DrOptix: Thank you!

egui/src/widgets/text_edit.rs
DrOptix: With this commit https://github.com/emilk/egui/pull/246/commits/abb8b538c4ddbb2c554bc79be491fd9f5ddd66d8 the bulk of the work should be done. I'm going to iron out some things later. Stuff like what to do when we have a selection and we press `shift`+`tab`.

DrOptix: After extensive testing I concluded that what I proposed above is garbage. Now I try to wrap my head around the cursor system while starting kind of from scratch. Nevertheless I won't stop until I get it done.

DrOptix: Last push contains indentation management using `tab`, `shift`+`tab`, `ctrl`+`[` and `shift`+`]`.
What may need to still be polished is selection update when adding or removing indentation or when converting from spaces to tabs.

egui/src/widgets/text_edit.rs
DrOptix: I crated a `CodingConfig` struct for this purpose.
https://github.com/DrOptix/egui/blob/e14c5491b61574647ed837130e0a61f4c1ce98b4/egui/src/widgets/text_edit.rs#L107-L112

egui/src/ui.rs
DrOptix: I crated a `CodingConfig` struct for this purpose.
https://github.com/DrOptix/egui/blob/e14c5491b61574647ed837130e0a61f4c1ce98b4/egui/src/widgets/text_edit.rs#L107-L112

egui/src/widgets/text_edit.rs
DrOptix: It is done

emilk:
Oh wow, this kind of got out of hand didn't it :) It's amazing with a code editor in egui, but I'm a worried about adding 1550 lines of code. The whole `egui` crate is only 14kloc!

I won't have time to review this for another few days, but anything you can think of the reduce the complexity would be appreciated. In particular: do we really need to supports spaces for indentation? Making an editor that only supports tabs would make everything A LOT simpler (just add or remove the first tab on a line). If a user wants spaces in their code when they save to disk, then can just convert to/from spaces on load/save instead. The code to convert to/from tabs/spaces would also be a lot simpler as we wouldn't need to track the cursor position.

Or, going the other direction, if we want a full-fledged code editor for egui, maybe that should be a separate crate (`egui-code` or something) that builds *on* egui rather than being a part of egui.

DrOptix:
Well yes, I went from ecstasy to frustration and back to ecstasy. I will try to think of a way to reduce the complexity. For me at least the biggest complexity was represented by keeping track of remaining tabulation size. I used VS Code as a blueprint for how to add, remove or convert indentation and also how to keep the selection while adding or removing indentation. 
What I can tell you is that I added quite a few tests. A birds eye view gives me around 500 lines of tests in `text_edit.rs`.

emilk:
To explain my last post a bit more: I consider all code to be a liability and a burden. I always strive to express as much as possible with as little as possible. Each line of code is something that needs to be maintained in perpetuity, therefor each line of code must pull its weight.

In this particular case I know that the cursor handling in egui will need to be rewritten in the future, probably several times, to support things like [right-to-left languages](https://github.com/emilk/egui/issues/56) and [moving over graphemes rather than characters](https://github.com/emilk/egui/issues/62). This PR will make such work a lot harder.

If an advanced code editor was a separate crate, then egui can be refactored and the code editor crate can follow, instead of doing both in one swoop (and by the same person).

But in this case I think we can get 90% of the result with 10% of the code by only supporting tabs for indentation, and force users to convert their code to that convention if they want to use egui as a code editor. As stated before, it is rather simple to convert to/from tabs when loading said code, so there is really no need for egui to support both.


PS:
Supporting variable-width tabs would also be great (https://twitter.com/sarah_federman/status/1146544481556033537?lang=en) but can also wait for a separate PR. It is always better with many small PR:s than one huge PR.

lunabunn:
**TL;DR: IMHO this should all be in a separate crate except for the part where you can disable tab moving focus and swap over to tab inserting `'\t'` instead.**

Nice work! However, this has some major flaws:
- 4-space tab is an arbitrary number
    - 2-space tabs and even 3-space tabs are used
    - if we were to have a code editor widget, I'd expect it would at least support those
- tabbing/shift-tabbing keybinds should be customizable
- **tab indents does NOT mean code**
    - selecting and tabbing/shift-tabbing is a coding convention and should not be on by default... or be there at all
    - the whole idea of spaces as tabs is also a coding convention and should not be on by default... or be there at all

Overall, although I like the idea, all the stuff aside from not moving focus on tab should at the very least be separated into a separate PR. Further, IMHO, the code here is too big to be wieldy while also being too "basic" for many use cases anyway. Both of these factors make it unfit for the core library and as such it should _really_ be just a separate crate. egui is a GUI library. It should expose just enough features so the user can build whatever they want to _using_ it. It shouldn't try to build the GUI _for_ the user; that's still the user's job (and for good reason, as it's unreasonable for one library to support every opinionated use case).

DrOptix:
Hello sorry for the inactivity, I was focusing on something else.

I will try to clean the PR this weekend, but my questions is, what should I leave in it?
Only the part where you can capture the focus and insert `\t` and rendering it as 4 spaces as @emilk initially proposed in one of it's commits, before I went and coded all of this?

Now about configurable key bindings, at the moment all key bindings are hard coded and from what I know there is no KeyMapManager available that fires events when some configurable key binding was detected.

I'm OK with both giving up on this PR or cleaning it and leave only what's needed and/or acceptable.
I would appreciate a lot to know what to leave and what to purge from the PR.

Thank you a lot for your feedback!

lunabunn:
>>> Now about configurable key bindings, at the moment all key bindings are hard coded and from what I know there is no KeyMapManager available that fires events when some configurable key binding was detected.

Right, which is why the code editor should be implemented in user code. Along with the user's own key map manager.

I'm not emilk, but I think this PR should implement, as you say, only the part where you can capture the focus and insert \t and rendering it as 4 spaces. However, definitely don't get rid of your current code! You've done a lot of great work and it could be used as a starting point when implementing a code editor in the future on top of egui, probably as an external crate.

lunabunn:
And in any case, even if we decide to add a "code editor" to core egui sometime, that doesn't need to happen in this very PR. Those two basic changes mentioned above would be enough to fix https://github.com/emilk/egui/issues/73. If we want to discuss this more and maybe merge more of your changes, you can always open a new PR.

emilk:
> I will try to clean the PR this weekend, but my questions is, what should I leave in it?
> Only the part where you can capture the focus and insert `\t` and rendering it as 4 spaces as @emilk initially proposed in one of it's commits, before I went and coded all of this?

I think that would be good, yes. I think the main bulk of this PR could be turned into an `egui-code` crate or similar. Such a crate could have a lot more freedom to do things like adding integration with [`syntect`](https://crates.io/crates/syntect).

DrOptix:
Cleanup according to PR reviews

- Removed tab_as_spaces idea.
- Removed identation management

The only thing left is the tab rendering which adaps to the remaining
space inside the current tab block.

We consider the following sequence of characters, separated with
spaces for clarity:

```
\t 1 \t 12 \t 123 \t 1234 \t
```

It will be rendered like this, using ---> to denote the space occupied
by a tab:

```
1234 | 1234 | 1234 | 1234 | 1234
-----+------+------+------+-----
---> | 1--> | 12-> | 123> | --->
```

lunabunn:
I don't really understand the adaptation thing. I don't seem to be able to replicate your example in Google Docs nor OSX Pages, could you clarify what this is actually supposed to be doing?

DrOptix:
Even if I used a mono spaced font in Google Docs I just reproduced it without issues.
From my checks a tab in Google Docs is around 5 spaces and a bit.
image

epaint/src/text/font.rs
```diff
@@ -282,9 +343,34 @@ impl Font {
         let mut cursor_x_in_points = 0.0f32;
         let mut last_glyph_id = None;
 
+        let mut tab_size = super::MAX_TAB_SIZE;
+
         for c in text.chars() {
             if !self.fonts.is_empty() {
-                let (font_index, glyph_info) = self.glyph_info(c);
+                // For tabs it is nice to let them occupy only what's left of the current
```
emilk, line 350: this is very common behavior, but is it "nice"? Does anyone actually like this behavior? It adds a lot of complication for a dubious feature.

If we should do something smart with tabs, we should implement [elastic tabstops](https://nickgravgaard.com/elastic-tabstops/) instead (but not in this PR)

epaint/src/text/font.rs
```diff
@@ -282,9 +343,34 @@ impl Font {
...
+                // -----+------+------+------+-----
+                // ---> | 1--> | 12-> | 123> | --->
+                let (font_index, glyph_info) = if c == '\t' {
+                    let result = self.tab_glyph_info(tab_size);
+                    tab_size = super::MAX_TAB_SIZE;
+                    result
+                } else {
+                    tab_size -= 1;
+                    self.glyph_info(c)
+                };
```
emilk, line 368: A lot simpler implementation of this would be to just modify `glyph_info.advance_width` directly here. No need for `fn tab_glyph_info` and a separate glyph cache for different tab widths

emilk (changes requested):
Let's do the stupid thing of always keeping tabs at width=4. Tabs should only be used for indentation anyway (["indent with tabs, align with spaces"](https://dmitryfrank.com/articles/indent_with_tabs_align_with_spaces))

DrOptix:
All the simplifications should be done now.

epaint/src/text/font.rs
DrOptix: I'm still wrapping my head around it, but it looks really interesting

emilk (approved):

Issue #250, opened 2021-03-25
warning: lint `missing_crate_level_docs` has been removed


jgrussell:
**Describe the bug**
A recent commit to master has resulted in this warning which prevents release builds.
`warning: lint 'missing_crate_level_docs' has been removed: use 'rustdoc::missing_crate_level_docs' instead
`

**To Reproduce**
1. clone this repo locally
2. cargo +nightly run --release // to see build failure
3. cargo +nightly run  // to see warnings but still able to run

**Expected behavior**
A successful release build

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04.3 LTS

**Additional context**
This was a trivial code fix; I am putting together a PR to address it now.

jgrussell:
If this is being handled in #249, please feel free to close this as a duplicate.

PR #251, opened 2021-03-25
fix for lint warnings

jgrussell:
Quick fix for failing release build documented in [warning: lint missing_crate_level_docs has been removed #250](https://github.com/emilk/egui/issues/250#issue-841180785) issue.

### DIFF
--- a/egui/src/lib.rs
+++ b/egui/src/lib.rs
@@ -227,6 +227,7 @@
 //! ui.style_mut().wrap = Some(false);
 //! ```
 
+#![cfg_attr(not(debug_assertions), deny(rustdoc::missing_doc_code_examples))] 
 #![cfg_attr(not(debug_assertions), deny(warnings))] // Forbid warnings in release builds
 #![forbid(unsafe_code)]
 #![warn(
@@ -274,7 +275,6 @@
     clippy::unused_self,
     clippy::verbose_file_reads,
     future_incompatible,
-    rustdoc::missing_crate_level_docs,
     nonstandard_style,
     rust_2018_idioms
 )]
--- a/egui/src/lib.rs
+++ b/egui/src/lib.rs
@@ -274,7 +274,7 @@
     clippy::unused_self,
     clippy::verbose_file_reads,
     future_incompatible,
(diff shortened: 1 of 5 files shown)


jgrussell:
I obviously have more work to do on this before merging.  This did compile and build cleanly on my local machine; I should have remembered to enable checks on my fork of this repository.

egui/src/lib.rs
```diff
@@ -227,6 +227,7 @@
 //! ui.style_mut().wrap = Some(false);
 //! ```
 
+#![cfg_attr(not(debug_assertions), deny(rustdoc::missing_doc_code_examples))]
```
emilk, line 230: Shouldn't this be `rustdoc::missing_crate_level_docs` ?

jgrussell: Yes.  Definitely. Copy and paste error on my part.  Sorry.  Fixing....

emilk: I am having trouble getting any version of `missing_crate_level_docs` to work though. That is, if I remove the crate level docs I don't get any warnings.

jgrussell: I am seeing the same behavior, no warnings if I remove the crate level comments.

Looks like this is not as trivial a fix as I had thought; but, I will continue working on it.

jgrussell: @emilk 

I cannot figure out how to perform this lint without running `cargo doc` or something similar.  But, if I replace the above line of code with `#![deny(rustdoc::missing_crate_level_docs)] ` and remove the crate documentation comments, this will indeed fail:
`cargo +nightly doc --no-deps --package egui_demo_lib `

I have not yet been able to get this to fail only when including the `--release` cargo doc flag.

How would you like me to proceed?

- Do you want me to push this change to make cargo doc fail any time crate level documentation is missing?  
- I can just remove the current missing_crate_level_docs line so it does not become an issue later.
- Something else; if so, please advise.

emilk: Oh, so there is no way to get that warning without nightly. Yeah, let's just remove it.

PR #257, opened 2021-03-27
Arbitrary data in `Memory` using `Any`, fix #255

### LINKED ISSUE #255: Let `Memory` store arbitrary data for custom widgets.
I have this widget:
image image

This is the editor where the result can be previewed. And this widget has two states: `View` or `Edit`, this can be represented as `bool`. I believe this is more convenient to store this data in `egui::Memory` because this whole thing looks like some complicated widget that receives `&mut String`, and using this we can have the same benefits as other widgets have (able to serialize UI memory between runs). Also, creating custom memory and drag it into every function is quite inconvenient.

I looking for the code of `CollapsingHeader`, because it has some bool state too (opened or not). And found that `Memory` [hard-coded](https://github.com/emilk/egui/blob/master/egui/src/memory.rs#L30) all data for collapsing headers and collapsing header [uses this](https://github.com/emilk/egui/blob/master/egui/src/containers/collapsing_header.rs#L27) private access. And, as I see there is no public interface to store data for a custom widget.

First, we can provide an interface to store just a single `bool`, `i32`, `String` using enums for our id.

But a more scalable approach is using `HashMap` with `Box<dyn Any + ...>`. Using it, we can store arbitrary types for our `id`. And default widgets can be painlessly rewritten using it.

We also can require `Serialize` + `Deserialize` traits for all objects that should be stored in this type-map (under feature, of course).

But this type-map can possibly add visible runtime overhead comparing to current hard-coded maps.

Not allowing a user to store data in `egui::Memory`.

**Additional context**
I want to try to implement this feature by myself and open a Pull Request. What do you think?

optozorax:
- [x] read [`CONTRIBUTING.md`](https://github.com/emilk/egui/blob/master/CONTRIBUTING.md)


egui/src/memory.rs
```diff
@@ -373,6 +374,44 @@ impl Memory {
...
+    }
+
+    // Not all can be resetted, if it's not deserialized
+    pub fn reset_all<T: DataElementTrait>(&mut self) {
+        self.data
+            .retain(|_, v| (*v).type_id() == Some(std::any::TypeId::of::<T>()));
+    }
+
+    // Not anything can be counted
+    pub fn count<T: DataElementTrait>(&mut self) -> usize {
```
emilk, line 400: Right now counting is only used for some inspection views, and I struggle to find any real-life usage of it.

egui/src/memory.rs
```diff
@@ -373,6 +374,44 @@ impl Memory {
...
+            .count()
+    }
+
+    pub fn count_unknown(&mut self) -> usize {
+        self.data
+            .iter()
+            .filter(|(_, v)| (*v).type_id().is_some())
+            .count()
+    }
+}
```
emilk, line 413: I think it would be nice if this interface was put on a newtype, e.g.

`pub struct DataStore(HashMap<Id, DataElement>)`

optozorax: Done https://github.com/emilk/egui/pull/257/commits/81ff915cd127bc7864dde58b416a0189b89a65eb

egui/src/memory.rs
optozorax: That's clever! With this approach, we can really count things and reset individual types. Unfortunately, `TypeId` is not de/serializable, so I introduced `SerializableTypeId` using, which has drawbacks because implemented using `Hash`.

optozorax:
Tomorrow I will write more documentation and rewrite all widgets to new API.

egui/src/memory.rs
emilk: It is a shame that `TypeId` cannot be easily serialized. In [the documentation](https://doc.rust-lang.org/std/any/struct.TypeId.html) they make it clear that "the hashes will vary between Rust releases". This means upgrading the rust compiler may void the egui memory. As stated before the memory shouldn't be used for anything important, but I wonder if we can do better.

For instance, we could serialize the hash of the TypeId with each `DataElement` value. Then we could use that for counting and resetting, and if the hash breaks between releases only counting and resetting will break, but the memory will still be intact.

egui/src/memory.rs
```diff
@@ -20,21 +17,19 @@ use epaint::color::{Color32, Hsva};
...
-    pub(crate) collapsing_headers: HashMap<Id, collapsing_header::State>,
-    pub(crate) grid: HashMap<Id, crate::grid::State>,
-    #[cfg_attr(feature = "persistence", serde(skip))]
-    pub(crate) menu_bar: HashMap<Id, menu::BarState>,
-    pub(crate) resize: HashMap<Id, resize::State>,
-    pub(crate) scroll_areas: HashMap<Id, scroll_area::State>,
-    pub(crate) text_edit: HashMap<Id, text_edit::State>,
-
     #[cfg_attr(feature = "persistence", serde(skip))]
     pub(crate) window_interaction: Option<window::WindowInteraction>,
```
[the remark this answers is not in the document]
optozorax: Fixed in https://github.com/emilk/egui/pull/257/commits/5e0186f96d56b07f9fca90e3290490c32c42a0b6 by introducing different types to store by `Id` and store by `TypeId`.

egui/src/memory.rs
```diff
@@ -20,21 +17,19 @@ use epaint::color::{Color32, Hsva};
 pub struct Memory {
     pub options: Options,
 
+    /// This map stores current states for any different widgets. This will be saved between different program runs if you use `persistence` feature.
+    pub data: AnyMap,
+
+    /// Same as `data`, but this data will not be saved between runs.
+    #[cfg_attr(feature = "persistence", serde(skip))]
+    pub data_temp: AnyMap,
```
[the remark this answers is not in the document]
optozorax: Fixed in https://github.com/emilk/egui/pull/257/commits/5e0186f96d56b07f9fca90e3290490c32c42a0b6 by introducing different types for serializable and not serializable storages. Unfortunately, much similiar code, but a nice interface.

egui/src/memory.rs
optozorax: Did that in next commits, also write what could break due to this in the documentation.

optozorax:
I will test this on the web, improve grammar and vocabulary in the documentation and mark this PR as ready for review.

egui/src/any/element.rs
```diff
@@ -0,0 +1,64 @@
+use std::any::{Any, TypeId};
+use std::fmt;
+
+pub(crate) struct AnyMapElement {
```
emilk, line 4:
```suggestion
/// Like [`std::any::Any`], but also implements `Clone`.
pub(crate) struct AnyMapElement {
```

emilk (changes requested):
Excellent work!

I left some comments, but beside that I think it would also be good with a couple of unit tests, especially covering serialization/deserialization of `AnyMapId`. For instance, does the `AnyMapId` handle serializing `struct State { foo: i32 }` and deserializing to `stuct State { foo: i32, #[serde(default)] bar: f64 }` (simulating adding a new field to an existing state)?

I also think you can make your own crate out of all this if you would like! I think it could become a popular alternative to e.g. `anymap` and `std::any::Any` for others who needs serialization!

optozorax:
Okay, thank you! Right now I am realized that I have no more time, so I can continue this PR after 6 April.

Also, thanks about the idea of moving this out as a crate :thinking:

egui/src/any/any_map.rs
```diff
@@ -0,0 +1,64 @@
+use crate::any::element::{AnyMapElement, AnyMapTrait};
+use std::any::TypeId;
+use std::collections::HashMap;
+
+/// Stores object of any type.
+#[derive(Clone, Debug, Default)]
+pub struct AnyMap(HashMap<TypeId, AnyMapElement>);
```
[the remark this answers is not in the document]
bjorn: I like reusing `TypeMap` for consistency, but I'll just add that I know this kind of container as a `MultiAny`.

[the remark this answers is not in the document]
optozorax: Done in https://github.com/emilk/egui/pull/257/commits/bd5d0419307531cef0bc1f7dc1dc15fed7943487

Also, I renamed `AnyMapId` to `AnyMap` in https://github.com/emilk/egui/pull/257/commits/ec8da3a6fa7c5c9ce3b64ff507e143f698e4bae8

egui/src/any/serializable/any_map.rs
```diff
@@ -0,0 +1,75 @@
+use std::collections::HashMap;
+
+#[cfg(feature = "persistence")]
```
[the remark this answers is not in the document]
optozorax: done in https://github.com/emilk/egui/pull/257/commits/177557dd77881fcccc26b8ca74c814e4597702fb

egui/src/any/serializable/element.rs
```diff
@@ -0,0 +1,148 @@
...
+pub(crate) struct AnyMapElement(AnyMapElementInner);
+
+enum AnyMapElementInner {
+    Deserialized {
+        value: Box<dyn Any + 'static>,
+        clone_fn: fn(&Box<dyn Any + 'static>) -> Box<dyn Any + 'static>,
+
+        serialize_fn: fn(&Box<dyn Any + 'static>) -> Result<String, ron::Error>,
+    },
+    ToDeserialize(String),
```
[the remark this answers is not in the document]
optozorax: done in https://github.com/emilk/egui/pull/257/commits/511a37ae41da97fc97df3cb6ca1b35814a1bb4a3

optozorax:
@emilk everything is done.

optozorax:
I don't like all this copy-pasting in `any_map` and `type_map` files. So, I think it's possible to remove this copy-pasting by using `type` syntax and generics. I should try this.

optozorax:
No, I believe this is not possible, because this may require an associated trait inside of a trait to generically handle `AnyMapElement` and `serialiblaze::AnyMapElement`.

optozorax:
However, copy-pasting between `TypeMap`/`serializable::TypeMap` and between `AnyMap`/`serializable::AnyMap` can be eliminated by using `incude!()` macro, but this is kinda ugly. I could do this, if you find this more proper than copy-pasting.

egui/Cargo.toml
```diff
@@ -30,8 +31,11 @@ default = ["default_fonts", "single_threaded"]
...
 
-persistence = ["serde", "epaint/persistence"]
+persistence = ["serde", "epaint/persistence", "ron"]
 
 # Only needed if you plan to use the same egui::Context from multiple threads.
 single_threaded = ["epaint/single_threaded"]
 multi_threaded = ["epaint/multi_threaded"]
+
+[dev-dependencies]
+serde_json = "1"
```
emilk, line 41: we could use `ron` for the tests too

egui/src/any/serializable/type_map.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    let file_string = {
+        let mut map: TypeMap = Default::default();
+        map.insert(State { a: 42 });
+        serde_json::to_string(&map).unwrap()
+    };
+
+    let mut map: TypeMap = serde_json::from_str(&file_string).unwrap();
+    assert!(map.get::<StateNew>().is_none());
```
emilk, line 114: Oh right, because `TypeId` changes. In the future it would be nice to not rely on `TypeId`, but perhaps use a per-type UUID or similar to handle it. That would make persistence work across rustc versions too.

emilk (approved):
Amazing work - thank you! ⭐

egui/src/any/any_map.rs
```diff
@@ -0,0 +1,205 @@
+use crate::any::element::{AnyMapElement, AnyMapTrait};
+use std::any::TypeId;
+use std::collections::HashMap;
+use std::hash::Hash;
+
+/// Stores any object by `Key`.
+#[derive(Clone, Debug)]
+pub struct AnyMap<Key: Hash + Eq>(HashMap<Key, AnyMapElement>);
```
emilk, line 8: Now that this no longer relies on `Id` it would be trivial to move to its own crate - hint, hint, nudge, nudge ;)

Issue #248, opened 2021-03-24
IME Support


lunabunn:
**Update (April 2021):**
Since this issue was initially opened, good progress has been made on IME support.
#258 Implemented proper candidate window positioning for `egui_web`
#278 Implemented IME composition events for `egui` as well as logic to fire them for `egui_web`
This means `egui_web`, as well as  core`egui`, now has IME support. (Thanks @n2 !)

However, `eframe` on native targets still DOES NOT have IME support. And with `egui-glium` (and many other `winit`-based egui backends), it is unlikely support will land any time in the near future. This is not an egui issue but rather an upstream issue. The biggest blocker is https://github.com/rust-windowing/winit/issues/1497. `winit` currently does not have IME support, which means most of the Rust ecosystem does not have IME support. Glium/glutin are no exception.

`egui-miniquad`, despite not relying on `winit`, also has a similar issue because `sokol-app`, `miniquad`'s windowing library of choice, doesn't have IME support either. The only egui backend listed in the README that would be able to support IMEs is `egui_sdl2_gl`. I opened [an issue there](https://github.com/ArjunNair/egui_sdl2_gl/issues/8).

----

**Original issue:**
**Describe the bug**

There are no events available for IME composition, nor do text inputs display IME composition strings. The candidate window is not shown in the appropriate location.

**To Reproduce**
1. See that https://docs.rs/egui/0.10.0/egui/enum.Event.html does not have any IME-related events.
2. Open up any egui program, regardless of the backend impl, and observe the lack of IME support.

**Expected behavior**
There are events available for when the IME composition string/state changes, and text inputs display said string/state correctly. The candidate window should be shown right below the cursor.

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
N/A

**Additional context**
IME support is a(n often overlooked) must for many people around the world in any sort of production GUI.

emilk:
It's being worked on in https://github.com/emilk/egui/pull/247 !

lunabunn:
> It's being worked on in #247 !

As far as I can tell that PR is unrelated? On PC there needs to be a proper hook to pass composition events into egui. That PR seems like a "fix" at the egui_web level.

emilk:
Ah, I read your issue to quickly! Yeah, #247 is IME for web, but you want IME for desktop.

n2:
Please let me do it. I have a plan on it, and, it is important to me, too.

lunabunn:
[...]

PR #258, opened 2021-03-27
Move IME candidate window following text cursor

n2:
Now IME candidate window can move with text cursor, both on web(laptop/desktop) and native app.This is part of the work at #248 , it is relatively simple and independent of the rest of the work, a separate PR is submitted.


emilk:
This looks great! Let me know when it is ready for review.

n2:
Well, it is done. Thera is a little bug in composition events part on mobile devices, so it will be commited latter. @emilk

emilk (approved):
Looks great, thank you!

PR #262, opened 2021-03-31
Fixed some spelling in documentation

bjorn:
Just some spelling errors I encountered while trying to write my first egui program.

Thanks for your work on egui, I think it looks very promising!


emilk (approved):
Thanks!

PR #263, opened 2021-03-31
egui_glium: Use config_dir instead of data_dir

bjorn:
[The documentation](https://docs.rs/epi/0.10.0/epi/trait.App.html#method.save) stated `.config` would be used but since `data_dir` is used here, it really ends up in `.local/share`.

Of course, alternatively the documentation could be updated. I'm personally not sure what is the more suitable location for storing the application state. Does it count as configuration or data?


emilk:
Oops, good catch! In this case I'd say it's the docs that are wrong. The paths are used to store application state, which is more than just "configurations" in my opinion.

PR #274, opened 2021-04-04
DragValue improvements

### LINKED ISSUE #273: Proposed DragValue improvement
At the moment drag values are good, but I believe we can make them better. I'm ready to work on this one.
@emilk if you give me the go ahead I can start working on it or part of it depending on what you think will be good for **egui**.

I would like to propose an implementation similar with the ones from Blender.

You can see more in the GIF below, but the key improvements are in order of priority:

- [x] On hover display ResizeHorizontal cursor
- [x] When `shift` is pressed and we drag we add or subtract `DragValue::speed()` / 10
- [x] While dragging display ResizeHorizontal.
- [ ] ~Add arrows to increase or decrease value, maybe display them only on hover~
- [ ] ~When arrows are clicked we add or subtract the amount equal to the value set by `DragValue::speed()`~
- [ ] ~When `shift` is pressed and we click on the arrows we add or subtract `DragValue::speed()` / 10~
- [ ] ~When we start dragging save cursor position, hide the cursor while dragging, when dragging is done display the cursor at the saved location~

Nothing here as my focus is to do my best to help improve **egui**

**Additional context**
How Blender does it.
blender_drag_value

DrOptix:
- On hover display ResizeHorizontal cursor
- When shift is pressed and we drag we add or subtract DragValue::speed() / 10
- While dragging display ResizeHorizontal.

Resolves #273.


egui/src/widgets/drag_value.rs
```diff
@@ -262,21 +273,41 @@ impl<'a> Widget for DragValue<'a> {
...
             } else if response.dragged() {
+                response.ctx.output().cursor_icon = CursorIcon::ResizeHorizontal;
+
                 let mdelta = response.drag_delta();
                 let delta_points = mdelta.x - mdelta.y; // Increase to the right and up
+
+                let is_slow_speed = self.is_slow_speed
+                    && response.ctx.input().modifiers.shift
+                    && !(response.ctx.input().modifiers.alt
+                        || response.ctx.input().modifiers.command);
```
emilk, line 305: Add a helper to `Modifiers` instead so you can write `ui.input().modifiers.only_shift()` here

egui/src/widgets/drag_value.rs
```diff
@@ -262,21 +273,41 @@ impl<'a> Widget for DragValue<'a> {
...
+            let response = ui.add_sized(ui.spacing().interact_size, button);
+            let response = response
+                .on_hover_cursor(CursorIcon::ResizeHorizontal)
+                .on_hover_text(hover_text);
 
             if response.clicked() {
                 ui.memory().request_focus(kb_edit_id);
                 ui.memory().drag_value.edit_string = None; // Filled in next frame
             } else if response.dragged() {
+                response.ctx.output().cursor_icon = CursorIcon::ResizeHorizontal;
```
emilk, line 297:
```suggestion
                ui.output().cursor_icon = CursorIcon::ResizeHorizontal;
```

just shorter, and also more consistent with the rest of egui codebase

emilk (changes requested):
Thank you for the PR!

If I do this:

```
ui.add(egui::DragValue::new(scalar).speed(1.0).slow_speed(true));
```

Then dragging slowly still rounds to nearest integer (integer because speed=1). This can be easily fixed! The rounding is using `auto_decimals`, so when shift is pressed increase `auto_decimals` with 1.

I also think we should not have `slow_speed` as an option, but always enable it. It is a great feature, so let's not hide it! Or is there a reason why one would not want it?

egui/src/widgets/drag_value.rs
DrOptix: > Thank you for the PR!
> 
> If I do this:
> 
> ```
> ui.add(egui::DragValue::new(scalar).speed(1.0).slow_speed(true));
> ```
> 
> Then dragging slowly still rounds to nearest integer (integer because speed=1). This can be easily fixed! The rounding is using `auto_decimals`, so when shift is pressed increase `auto_decimals` with 1.
> 
> I also think we should not have `slow_speed` as an option, but always enable it. It is a great feature, so let's not hide it! Or is there a reason why one would not want it?

Indeed when we work with integers slow speed is a bit redundant, but for controlling floats I think it is very useful.
I made it optional because I thought it would be a benefit to be able to opt-in for this behavior, but personally I think it should be always active.
I will update the PR first thing in the morning.

DrOptix:
@emilk I think I solved all the issues you mentioned.

emilk (approved):
Very nice, thank you!

PR #293, opened 2021-04-10
Add links to the changelogs for other crates.

follower:
Makes the other changelogs easier to find/access.

I always forget that the other crates aren't in separate repos & waste time when trying to catch up on the changelogs.

Now they're only a click away. :)

(Tested via the GH web interface.)


emilk (approved):
Great idea! ⭐

Issue #279, opened 2021-04-07
Touch gesture support


emilk:
It would be great if `egui` could support pinch-to-zoom, multi-finger swipes and other touch gestures. For a start, the most important place to support it is the web (i.e. in `egui_web`).

(NOTE: full multi-touch support, e.g. dragging two sliders at once, is a separate issue).

I suggest we start by adding touch events to [`egui::Event`](https://docs.rs/egui/0.11.0/egui/enum.Event.html). Let's do something similar to `winit`:

``` rust
enum Event {
    …
    Touch(TouchEvent),
}

pub struct TouchId(pub u64);

struct TouchEvent {
    id: TouchId,
    pos: Pos2,
    phase: TouchPhase,
}

pub enum TouchPhase {
    Started,
    Moved,
    Ended,
    Cancelled,
}
```

`egui::InputState` can then keep track of the touches in a `HashMap<TouchId, Pos2>` and detect gestures. For instance, if there is exactly two touches then it could measure zoom, translation and rotation to support manipulating 2D surfaces. An egui application would then just do something like `self.zoom *= ui.input().zoom();` to update its zoom-state each frame.

quadruple-output:
I plan to spend some time on it on the weekend 🚧

PR #306, opened 2021-04-13
Basic multi touch support (issue #279)

quadruple-output:
I guess I should open a draft PR before this gets too big. It would be great to get some feedback, @emilk .

Done:
- [x] Created an `Event::Touch` variant
- [x] `egui_glium` creates touch events 
(but I am not able to test this as it turns out that winit does not recognize my apple trackpad as a touch device)
- [x] `egui_web` creates touch events
- [x] Added a map of `TouchState`s to the `InputState` struct
(need a map because, technically there can be more than one touch device, identified by individual IDs)
- [x] Implemented `TouchState::ui()` (used in `InputState::ui()`), so touch events can be manually tested with `egui_demo_lib` on a mobile device.
This can be tested with the link mentioned below, but you need a device with a touch surface (otherwise you will see no change). Open the Inspection window and expand the Input collapsible. You should see a collapsible labelled "Touch State [device 0]".
- [x] Implemented the first prototype of a pinch/zoom gesture
~~This is realized with a `Gesture` trait object and a control loop capable of managing an arbitrary number of individual `Gesture` implementations. (Got this idea from studying [ctk::Gesture](https://gtk-rs.org/docs/gtk/struct.Gesture.html))~~
- [x] implement Zoom and Rotate, including a test window for `egui_demo_lib` (can be tested at the demo link below)
- [x] having consumed two gestures in a demo app, decide on the final API for applications
The proposal of having just a simple `input.zoom()` method returning the zoom factor is charming, but we may need more flexibility (for example the starting position (center point) and current position of the gesture).
→ I implemented the API as proposed by emilk in his initial comment (except that `touches.count()` is still missing).
- [x] add support for more than 2 fingers, and for determining the number of fingers on the surface

Out of scope of this PR (should I create issues?):
- [ ] return the touch info in the `Response` of `ui.interact()`, but only if the start position of the touch gesture is inside the rect area. (Not sure if this needs a deeper integration with `Sense`.)
- [ ] discuss whether the synthetic generation of pointer events from touch events, as it is currently implemented in egui_web, could be moved to egui. Advantage: backends don't have to care about this, any more.

Link to online demo for this branch: https://quadruple-output.github.io/egui/ – remember that you need a touch _screen_ (e.g. mobile phone or tablet) for testing.

Closes <https://github.com/emilk/egui/issues/279>.


emilk:
Thanks for starting work on this!

The primary use case I have in mind (and want to use) is simultaneous zoom and pan with two finger, something many UI:s fail to support. Therefor I don't think discreetizing the touches into gestures is the right approach. I think a simpler API could actually be more powerful.

Say we provide an API that allows the user to query the number of fingers, the total transform, and the delta transform (where transform is zoom around a point + translation). Then the user can then use what they need. For most cases, that would be panning and zooming (either the total or the per-frame delta).

If the user want to detect a two-finger swipe, it would be as easy as `touches.count() == 2 && touches.total.translation.len() > 100.0`, and we can add a helpful function for that if we want (`if ui.input().touches.swipe_left()`). But I suggest we don't build that until we need it. Let's keep a minimal-viable-product mindset and build the smallest useful touch support first, and keep the PR as small as possible. Better with many small PR:s than one big one!

quadruple-output:
Yes, this all sounds reasonable to me. Thanks a lot for these comments. I have to admit that I was at first fascinated by the idea of having minimal and therefore very simple implementations for each individual gesture. However, I also noticed later that at least implementing _zoom_ and _rotation_ in a combined gesture is a very simple thing. Therefore, I guess I just needed your feedback to help me getting back on track. I consider deleting code a good thing, so I will be happy to give it another try.

Still, I have one remark: discreetizing the gesture implementations does not mean that multiple "atomic" gestures (like _zoom_ and _rotate_) cannot be active _at the same time._ I think it would just be an implementation detail. Still, I fully agree to focus on an MVP. (I admit that I even thought about offering the ability to implement app-specific gestures by implementing the `Gesture` trait and registering trait objects. ;) But then - how many gestures could there possibly be?)

quadruple-output:
Should the touch API be part of the `Response` from `ui.interact()`? Or is this beyond the MVP scope?

emilk:
> Should the touch API be part of the `Response` from `ui.interact()`? Or is this beyond the MVP scope?

Let's hold off on that and get just the important parts in first

quadruple-output:
> If the user want to detect a two-finger swipe, it would be as easy as touches.count() == 2 && touches.total.translation.len() > 100.0

You are suggesting to offer `touches.total…`.

tl;dr: I suggest _not_ to implement this, at least not in the first step.

I understand the idea, and I have mostly implemented it (see `TouchInfo` [here](https://github.com/emilk/egui/pull/306/commits/dfc4f20aca5db512c3428305ba316c4bdf601d30#diff-7736bd3e2091835b1777837f4d8c3ba338556bf891e45bdea7b872b890a92460)). _However,_ the more I think about it, the more questions come up: It is straight forward to implement this for gestures involving a fixed number of fingers, but it gets complicated when you consider that fingers can be added and removed during the course of a gesture. This means that the `translation`  _(position = average of all fingers?)_ of the guesture would probably need to be adjusted, and `touches.total.translation` would perform a sudden "jump". Similar things can happen to `rotation` and `zoom`.

One solution could be to ignore the positions of fingers other than the first two – but what do you do if the third finger stays while the first gets removed?

These issues can easily be avoided for the _relative_ values by simply not delivering the information in the first frame after an individual touch has started or stopped.

I did not come across anything similar to `touches.total…` in the [MacOS Cocoa docs](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/HandlingTouchEvents/HandlingTouchEvents.html#//apple_ref/doc/uid/10000060i-CH13-SW1) I consulted, or in [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events). I conclude that there is no wide-spread need for the `total` touch data.

What do you think?

emilk:
I agree, scap `total`. `egui` is immediate mode anyway, so users should use the delta transform immediately!

egui_web/src/lib.rs
```diff
@@ -906,11 +952,17 @@ fn install_canvas_events(runner_ref: &AppRunnerRef) -> Result<(), JsValue> {
             let pos = pos_from_touch_event(runner_lock.canvas_id(), &event);
             runner_lock.input.latest_touch_pos = Some(pos);
             runner_lock.input.is_touch = true;
+
+            // TO BE DISCUSSED: todo for all `touch*`-events:
+            // Now that egui knows about Touch events, the backend does not need to simulate
+            // Pointer events, any more.  This simulation could be moved to `egui`.
```
emilk, line 958: I agree - the integrations should either send `Pointer`-events or `Touch`-events, but not both.

egui/src/input_state/touch_state.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+}
+
+fn center_pos(pos_1: Pos2, pos_2: Pos2) -> Pos2 {
+    pos2((pos_1.x + pos_2.x) * 0.5, (pos_1.y + pos_2.y) * 0.5)
+}
+
+fn distance(pos_1: Pos2, pos_2: Pos2) -> f32 {
+    (pos_2 - pos_1).length()
+}
```
emilk, line 268: you can already do `pos_1.distance(pos_2)`

egui/src/data/input.rs
```diff
@@ -123,6 +123,25 @@ pub enum Event {
...
+        phase: TouchPhase,
+        /// Position of the touch (or where the touch was last detected)
+        pos: Pos2,
+        /// Describes how hard the touch device was pressed. May always be `0` if the platform does
+        /// not support pressure sensitivity.
+        /// The value is in the range from 0.0 (no pressure) to 1.0 (maximum pressure).
+        force: f32,
+        // ### Note for review: using f32 forced me to remove `#[derive(Eq)]`
+        //     from the `Event` struct.  Can this cause issues?  I did not get errors,
+        //     so I wonder if `Eq` had been derived on purpose.
```
emilk, line 143: I think it's fine to remove `Eq`. Can't really see any real reason to have it

quadruple-output:
I have no idea why github complains about "1 workflow awaiting approval". I am not aware that my PR contains any such thing. I merged with upstream/master, though. Can this be the reason?

emilk:
@quadruple-output this is a new thing on GitHub - PR:s from first-time contributors do not have the CI run until an approve from a code owner. It is to stop people making PR:s that add bitcoin miners to a project and then have the CI do compute for them, and shady stuff like that. I didn't find any obvious cryptomining in your code, so I hazarded a press on the approve button :)

egui_web/src/lib.rs
quadruple-output: I agree that this should be done in a separate PR. Should I turn this comment into an issue?

egui/src/data/input.rs
quadruple-output: ok, thanks. I removed these code comments.

egui/src/input_state/touch_state.rs
quadruple-output: handy, thanks. The `distance` function has now dissolved into a new fn `calc_averages()`

quadruple-output:
Ready for review according to my judgement. Remember that there is a ready-to-try demo version [here](https://quadruple-output.github.io/egui/) (check out the "Multi Touch" demo window).

egui_web/src/lib.rs
quadruple-output: Code comment is already removed from PR.

egui/src/input_state/touch_state.rs
```diff
@@ -0,0 +1,288 @@
...
+/// minimized.
+fn normalized_angle(current_direction: f32, previous_direction: f32) -> f32 {
+    let mut angle = current_direction - previous_direction;
+    angle %= TAU;
+    if angle > PI {
+        angle -= TAU;
+    } else if angle < -PI {
+        angle += TAU;
+    }
+    dbg!(angle)
```
emilk, line 279: remove `dbg!` please!

egui/src/input_state/touch_state.rs
```diff
@@ -0,0 +1,288 @@
...
+    start_time: f64,
+    start_pointer_pos: Pos2,
+    force: f32,
+    previous: Option<DynGestureState>,
+    current: DynGestureState,
+}
+
+/// Gesture data which can change over time
+#[derive(Clone, Copy, Debug)]
+struct DynGestureState {
```
emilk, line 77: Could perhaps be combined with `TouchAverages` for some shorter code

egui/src/input_state/touch_state.rs
```diff
@@ -0,0 +1,288 @@
...
+    avg_pos: Pos2,
+    avg_distance: f32,
+    direction: f32,
```
emilk, line 80: minor nit: I think `angle` is a better name, since "direction" make me think of a vector

egui/src/input_state/touch_state.rs
```diff
@@ -0,0 +1,288 @@
...
+            // velocity, the shortcomings of this method will not be noticed.  One can see the
+            // issues though, when touching with three or more fingers, and moving only one of them
+            // (it takes two hands to do this in a controlled manner).  A better technique would be
+            // to store the current and previous directions (with reference to the center) for each
+            // touch individually, and then calculate the average of all individual changes in
+            // direction.  But this approach cannot be implemented locally in this method, making
+            // everything a bit more complicated.
+            let first_touch = self.active_touches.values().next().unwrap();
+            let direction_vec = (avg.pos - first_touch.pos).normalized();
+            avg.direction = direction_vec.y.atan2(direction_vec.x);
```
emilk, line 239:
```suggestion
            avg.direction = (avg.pos - first_touch.pos).angle();
```

egui_demo_lib/src/apps/demo/zoom_rotate.rs
```diff
@@ -0,0 +1,117 @@
...
+        ui.label("Try touch gestures Pinch/Stretch, Rotation, and Pressure with 2+ fingers.");
+        Frame::dark_canvas(ui.style()).show(ui, |ui| {
+            // Note that we use `Sense::drag()` although we do not use any pointer events.  With
+            // the current implementation, the fact that a touch event of two or more fingers is
+            // recognized, does not mean that the pointer events are suppressed, which are always
+            // generated for the first finger.  Therefore, if we do not explicitly consume pointer
+            // events, the window will move around, not only when dragged with a single finger, but
+            // also when a two-finger touch is active.  I guess this problem can only be cleanly
+            // solved when the synthetic pointer events are created by egui, and not by the
+            // backend.
```
emilk, line 58: 👍  I think it makes sense to consider multitouch manipulation as a form of dragging.

egui_demo_lib/src/apps/demo/zoom_rotate.rs
```diff
@@ -0,0 +1,117 @@
...
+                    4 => Color32::YELLOW,
+                    _ => Color32::RED,
+                };
+                // for a smooth touch experience (not strictly required, but I had the impression
+                // that it helps to reduce some lag, especially for the initial touch):
+                ui.ctx().request_repaint();
+            } else if let Some(last_time) = self.time_of_last_update {
+                // This has nothing to do with the touch gesture. It just smoothly brings the
+                // painted arrow back into its original position, for a nice visual effect:
+                let dt = ui.input().time - last_time;
```
emilk, line 94: You can use `ui.input().unstable_dt` instead here (not the best name, I know) so you don't need `time_of_last_update`

```suggestion
                let dt = ui.input().unstable_dt;
```

egui_demo_lib/src/apps/demo/zoom_rotate.rs
```diff
@@ -0,0 +1,117 @@
...
+            // but scaled and rotated according to the current translation:
+            let arrow_start = zoom_and_rotate * vec2(-0.5, 0.5);
+            let arrow_direction = zoom_and_rotate * vec2(1., -1.);
+            painter.arrow(
+                to_screen * (Pos2::ZERO + arrow_start),
+                to_screen.scale() * arrow_direction,
+                Stroke::new(stroke_width, color),
+            );
+            self.time_of_last_update = Some(ui.input().time);
+        });
```
emilk, line 115: It would be great if this demo also showed translation; then it would demonstrate all the higher level parts of the multitouch gesture API

egui_glium/src/lib.rs
```diff
@@ -185,6 +190,38 @@ pub fn input_to_egui(
...
+            input_state.raw.events.push(Event::Touch {
+                device_id: hasher.finish(),
+                id: touch.id,
+                phase: match touch.phase {
+                    glutin::event::TouchPhase::Started => egui::TouchPhase::Start,
+                    glutin::event::TouchPhase::Moved => egui::TouchPhase::Move,
+                    glutin::event::TouchPhase::Ended => egui::TouchPhase::End,
+                    glutin::event::TouchPhase::Cancelled => egui::TouchPhase::Cancel,
+                },
+                pos: Pos2::new(touch.location.x as f32, touch.location.y as f32),
```
emilk, line 213: `touch.location` is a `PhysicalPosition` so it should be divided by `pixels_per_point`

https://docs.rs/winit/0.24.0/winit/event/struct.Touch.html

emilk (changes requested):
Really great work! Super clean code. I only had a few very minor suggestions, but feel free to ignore or postpone.

Bonus points for publishing a demo of this for me to test ⭐ 

The only thing I lack is for the multitouch demo to also show translation (dragging with two fingers) to demonstrate, and test, full multitouch manipulation.

...and please remove the `dbg!(…)` call :)

emilk:
There is also an initial translation-jump when using two finger to zoom/pan in the Plot demo.

egui/src/data/input.rs
```diff
@@ -296,3 +312,27 @@ impl RawInput {
...
     }
 }
+
+/// this is a `u64` as values of this kind can always be obtained by hashing
+pub type TouchDeviceId = u64;
+
+/// Unique identifiction of a touch occurence (finger or pen or ...).
+/// A Touch ID is valid until the finger is lifted.
+/// A new ID is used for the next touch.
+pub type TouchId = u64;
```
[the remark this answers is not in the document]
quadruple-output: resolved by https://github.com/emilk/egui/pull/306/commits/ce8f3a8bedde6c67408c78d542e204f57b8e764f

egui/src/input_state/touch_state.rs
quadruple-output: resolved by https://github.com/emilk/egui/pull/306/commits/e3cc56ef6f6f6297312c3ba056d217ef5bea81a9

egui_demo_lib/src/apps/demo/zoom_rotate.rs
quadruple-output: resolved by https://github.com/emilk/egui/pull/306/commits/e3cc56ef6f6f6297312c3ba056d217ef5bea81a9

egui/src/input_state/touch_state.rs
quadruple-output: Oh, clippy seems to have a blind eye although `clippy::dbg_macro` is active.
Resolved by https://github.com/emilk/egui/pull/306/commits/612069a498369a0f59e3ca4259637ec9df625ccb

egui_glium/src/lib.rs
quadruple-output: fixed by https://github.com/emilk/egui/pull/306/commits/e4c9a65d57560385fa35701fe5119a6de210e840

egui/src/input_state/touch_state.rs
quadruple-output: Resolved with <https://github.com/emilk/egui/pull/306/commits/3918653979261912606a2a6d184c0c964fdcc1ce>. I removed `TouchAverages` completely and moved the `force` field from `GestureState` to `DynGestureState`.

egui/src/input_state/touch_state.rs
quadruple-output: Resolved with <https://github.com/emilk/egui/pull/306/commits/3918653979261912606a2a6d184c0c964fdcc1ce>. Decided to use `heading`.

egui_glium/src/lib.rs
```diff
@@ -185,6 +190,40 @@ pub fn input_to_egui(
...
+            input_state.raw.events.push(Event::Touch {
+                device_id: TouchDeviceId(hasher.finish()),
+                id: TouchId(touch.id),
+                phase: match touch.phase {
+                    glutin::event::TouchPhase::Started => egui::TouchPhase::Start,
+                    glutin::event::TouchPhase::Moved => egui::TouchPhase::Move,
+                    glutin::event::TouchPhase::Ended => egui::TouchPhase::End,
+                    glutin::event::TouchPhase::Cancelled => egui::TouchPhase::Cancel,
+                },
+                pos: pos2(touch.location.x as f32 * pixels_per_point_recip,
```
emilk, line 214:
```suggestion
                pos: pos2(touch.location.x as f32 * pixels_per_point_recip,
```

Hot tip: configure your editor to run `cargo fmt` on save. It is super nice!

emilk (changes requested):
`cargo fmt` whites about a single little whitespace.

Tip: configure your editor to run `cargo fmt` on save!

Also: you can use `./check.sh` to run the CI steps locally

egui_glium/src/lib.rs
quadruple-output: Resolved by <https://github.com/emilk/egui/pull/306/commits/b38e2255f347a65d42645ab13ac02223b4aaca1b>.
I actually do have my editor configured to run `cargo fmt` on save, but this particular source file seems to be a bit too much for `cargo fmt`, so it does not format around this peace of code.

quadruple-output:
> There is also an initial translation-jump when using two finger to zoom/pan in the Plot demo.

Resolved by <https://github.com/emilk/egui/pull/306/commits/acbefdac08b420e61b468a5cf7af5c79971c77a6>. I also uploaded an updated version of the [online demo](https://quadruple-output.github.io/egui/) for testing.

Honestly, I feel that the new code I added to `egui_web` should actually move into `egui` itself, so other integrations don't have to re-implement the same stuff. But this would certainly require another PR.

egui_demo_lib/src/apps/demo/zoom_rotate.rs
quadruple-output: OK, I will add this. Maybe tomorrow…

I was a bit reluctant to add translation because, for some devices (like a touch pad), the units for translation would neither be points, nor pixels on the screen, but rather the physical resolution of the touch device itself. Therefore, it feels like 'cheating' to use the translation coordinates just as if they were real screen coordinates – although it would work just fine on touch *screens*.

Unfortunately, I am only able to test all this with an iPhone and iPad. Both have (obviously) touch *screens*, not touch *pads*, so the resolution of the touch device and screen are always identical. I have a track pad, but `winit` does not support touch events for it. Same for `web_sys` in any browser I have tried.

quadruple-output: Resolved with <https://github.com/emilk/egui/pull/306/commits/755ca301a08b3d99fc0cca9dca9c7642c4be5d16>.
The [online demo](https://quadruple-output.github.io/egui/) is up-to-date.

quadruple-output:
Thanks a lot for your detailed feedback. Very appreciated. I think I resolved all comments. You are welcome to request more changes.

quadruple-output:
Actually, there is one thing I would like to propose to change: your method `ui.input().zoom_delta()` (which I adapted to multi touch) returns `1.0` in case there is no Zoom event. I personally would prefer this method to return an `Option<f32>`. I understand the motivation to return `1.0` by default, but even your own code in the implementation of `Plot` checks for the special value `1.0` (and has to silence a clippy warning to do so!).

If the only existing consumer of `zoom_delta()` would be better off with an `Option<f32>`, why do you think that everyone else would prefer the current implementation? If someone does not care whether a Zoom event is ongoing, it is as easy as `ui.input().zoom_delta().unwrap_or(1.0)` to get the current behavior.

emilk:
> I personally would prefer this method to return an `Option<f32>`. I understand the motivation to return `1.0` by default, but even your own code in the implementation of `Plot` checks for the special value `1.0` (and has to silence a clippy warning to do so!).

I really dislike that clippy warning, and should disable it project-wide, but that is beside the point.

I think you make a good argument, but if zoom returns an option, should scroll and mouse delta do so too? I think it is nicer to return the identity value to indicate no change. For additive values (like scroll) the identity is zero, an for zoom it is one.

emilk (approved):
Amazing work - thanks you so much!

PR #307, opened 2021-04-15
Memory usage example in the widget gallery

optozorax:
Related to https://github.com/emilk/egui/issues/255.


egui_demo_lib/src/apps/demo/view_edit.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    // The default state will be set for first call of widget with this id.
+    impl Default for State {
+        fn default() -> Self {
+            State::View
+        }
+    }
+
+    // 2. Create id
+    let id = Id::new(id_source);
```
emilk, line 44: This hinges on that `id_source` is unique in the entire program. It is better to use `ui.make_persistent_id(id_source);`

egui_demo_lib/src/apps/demo/view_edit.rs
```diff
@@ -0,0 +1,96 @@
...
+//! This is meant to be read as a tutorial, hence the plethora of comments.
+
+use egui::Id;
+use std::hash::Hash;
+
+/// Easymarkup text editor with the ability to preview a result.
+///
+/// ## Example:
+/// ``` ignore
+/// toggle_ui(ui, &mut my_text, "description_1");
```
emilk, line 12:
```suggestion
/// view_edit_ui(ui, &mut my_text, "description_1");
```

egui_demo_lib/src/apps/demo/view_edit.rs
```diff
@@ -0,0 +1,96 @@
...
+/// ```
+pub fn view_edit_ui(ui: &mut egui::Ui, text: &mut String, id_source: impl Hash) -> egui::Response {
```
emilk, line 14: It's better to call `text` `easymark`

egui_demo_lib/src/apps/demo/view_edit.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    // 5. Insert changed state back
+    ui.memory().id_data.insert(id, state);
+
+    // All done! Return the interaction response so the user can check what happened
+    // (hovered, clicked, ...) and maybe show a tooltip:
+    result.response
+}
+
+// A wrapper that allows the more idiomatic usage pattern: `ui.add(toggle(&mut my_bool))`
```
emilk, line 83:
```suggestion
// A wrapper that allows the more idiomatic usage pattern: `ui.add(view_edit(&mut my_bool))`
```

egui_demo_lib/src/apps/demo/view_edit.rs
emilk: Or use `ui.make_position_id()` and there is no need for `id_source`, and it will work as long as the number of proceeding widgets in the same UI doesn't change.

emilk (changes requested):
Great tutorial!

I'm a bit hesitant about showing off easymark, as it is not something I want people to use, but more of a demo itself (I should actually move it out from `egui` into `egui_demo_lib`).

In https://github.com/emilk/egui/commit/daf2e1323839d0796ffac2df57e97044a218e21f I added a password field demo with a show/hide toggle. I think something like that is a better subject for a demo, but moved to its own file with your great comments. If a user could type `ui.add(password(&mut password));` to have a password field with show/hide toggle next to it, that would be a realistic and useful thing to demo.

egui_demo_lib/src/apps/demo/view_edit.rs
optozorax: The best thing about contributing is that I better understand how to use the library. Thanks for `make_persistent_id`, now I use this. Previously I didn't know what this function is doing, I only understand that after reading [these comments](https://docs.rs/egui/0.11.0/src/egui/ui.rs.html#28-33), I think they should be added to this function documentation.

`make_position_id` is deprecated and says that `allocate_space` should be used. But I don't think allocation is needed for a composite widget without custom painting and etc.

optozorax:
I agree that the password widget with `hide`/`show` button is a more real-world and suitable example, so the example is changed to show this widget.

image

emilk (approved):

Issue #155, opened 2021-02-02
2D plotting functionality


EmbersArc:
Having a 2D plotting library inside egui would be highly beneficial. It already has most of the capabilities necessary for implementing one. Taking some inspiration from [implot](https://github.com/epezent/implot) for imgui, I put together a prototype of what this feature may look like.

Note that I only have a surface-level understanding of egui, so much of what I do here may not be "best practice" and there may be better ways to do things. I am willing to continue work on this, but would appreciate occasional reviews, advice or even collaboration from someone who is more proficient with the internal workings of egui. So let me know if you want to help.

https://github.com/EmbersArc/eplot
![simplescreenrecorder-2021-02-02_21 22 15](https://user-images.githubusercontent.com/1352472/106660902-de14a600-65a0-11eb-8c9b-347a05d497eb.gif)

emilk:
Very cool and impressive!

I have a vague plan of adding an immediate mode plot library to egui. My initial inclination is to make something very simple: plotting colored lines without any mouse interaction except hover.

I think a more fully-featured plotting library like you are demoing would make sense as a separate crate (`eplot`).

Let me know how `eplot` progresses and how I can help! In particular if there are any features you miss in `egui`.

One thing I've been planning to add is a transform to `Painter` to make it easier to build things like this. The transform would be applied to each [`Shape`](https://docs.rs/epaint/0.8.0/epaint/enum.Shape.html) passed to the `Painter`. I think initially it would only support translation and non-uniform scaling (including mirroring).

EmbersArc:
You're right that this should better be a separate crate since creating/maintaining it would be quite an endeavor. I don't yet know how far I'm going to take this but so far working with egui has been great.

As for missing features, the biggest ones are related to text rendering. I haven't been able to find a way to rotate text (note the missing y-axis label) and text positioning in the plot seems to be quite tricky as well but I probably didn't try hard enough there.

The transformation function you mention would indeed be helpful, especially with axis mirroring. I am currently doing this in my own function. But I'm afraid linear transformations won't be enough for things like plots with log-log scaling anyways, and there might not be a case to add those to egui itself.
I wouldn't recommend scaling whole shapes in this function, just coordinates. There are some cases where I want to transform the position of polygons, but not their individual vertices. But such a function might have other applications that I'm not thinking of right now.

emilk:
There is no way to rotate text at the moment. I think adding `emath::Rot2` to `Shape::Text` should be pretty easy.

[...]

PR #317, opened 2021-04-18
Drag and zoom support for plots

### LINKED ISSUE #316: Add drag and zoom support to the plotting demo
The plotting library is missing drag and zoom support in the plot.

I'd like to drag and zoom in the plot to adjust the plot bounds.

N/A

**Additional context**
The new custom memory will be great for saving the bounds between frames.

I've already started working on this and will open a PR soon.

EmbersArc:
Closes #316.

### DIFF
--- a/egui/src/widgets/plot.rs
+++ b/egui/src/widgets/plot.rs
@@ -3,7 +3,6 @@
 #![allow(clippy::comparison_chain)]
 
 use color::Hsva;
-use serde::{Deserialize, Serialize};
 
 use crate::*;
 
@@ -36,7 +35,8 @@ impl Value {
 
 /// 2D bounding box of f64 precision.
 /// The range of data values we show.
-#[derive(Clone, Copy, PartialEq, Serialize, Deserialize, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug)]
+#[cfg_attr(feature = "persistence", derive(serde::Deserialize, serde::Serialize))]
 struct Bounds {
     min: [f64; 2],
     max: [f64; 2],
--- a/egui/src/widgets/plot.rs
+++ b/egui/src/widgets/plot.rs
@@ -48,6 +48,11 @@ impl Bounds {
         max: [-f64::INFINITY; 2],
     };
 
+    pub const EMPTY: Self = Self {
+        min: [0.0; 2],
+        max: [0.0; 2],
+    };
+
     pub fn new_symmetrical(half_extent: f64) -> Self {
         Self {
             min: [-half_extent; 2],
@@ -274,7 +279,6 @@ pub struct Plot {
 
     symmetrical_x_bounds: bool,
     symmetrical_y_bounds: bool,
-    automatic_bounds: bool,
     margin_fraction: Vec2,
 
     min_size: Vec2,
@@ -282,6 +286,8 @@ pub struct Plot {
     height: Option<f32>,
     data_aspect: Option<f32>,
     view_aspect: Option<f32>,
+
+    bounds: Bounds,
 
     show_x: bool,
     show_y: bool,
@@ -299,7 +305,6 @@ impl Plot {
 
             symmetrical_x_bounds: false,
             symmetrical_y_bounds: false,
-            automatic_bounds: false,
             margin_fraction: Vec2::splat(0.05),
 
             min_size: Vec2::splat(64.0),
@@ -307,6 +312,8 @@ impl Plot {
             height: None,
             data_aspect: None,
             view_aspect: None,
+
+            bounds: Bounds::EMPTY,
 
             show_x: true,
             show_y: true,
@@ -365,9 +372,16 @@ impl Plot {
         self
     }
 
-    /// If true, the bounds will be set based on the data.
-    pub fn automatic_bounds(mut self, enabled: bool) -> Self {
-        self.automatic_bounds = enabled;
+    /// Expand bounds to include the given x value.
+    pub fn include_x(mut self, x: impl Into<f64>) -> Self {
+        self.bounds.extend_with_x(x.into());
+        self
+    }
+
+    /// Expand bounds to include the given y value.
+    /// For instance, to always show the x axis, call `plot.include_y(0.0)`.
+    pub fn include_y(mut self, y: impl Into<f64>) -> Self {
+        self.bounds.extend_with_y(y.into());
         self
     }
 
@@ -438,14 +452,14 @@ impl Widget for Plot {
             view_aspect,
             show_x,
             show_y,
-            automatic_bounds,
+            bounds,
         } = self;
 
         let plot_id = ui.make_persistent_id(name);
         let memory = ui
             .memory()
             .id_data
-            .get_or_default::<PlotMemory>(plot_id)
+            .get_mut_or_insert_with(plot_id, || PlotMemory { bounds })
             .clone();
 
         let PlotMemory { mut bounds } = memory;
@@ -473,7 +487,7 @@ impl Widget for Plot {
 
         let (rect, response) = ui.allocate_exact_size(size, Sense::drag());
 
-        if automatic_bounds || response.double_clicked_by(PointerButton::Primary) {
+        if response.double_clicked_by(PointerButton::Primary) || bounds == Bounds::EMPTY {
             bounds = Bounds::NOTHING;
             hlines.iter().for_each(|line| bounds.extend_with_y(line.y));
             vlines.iter().for_each(|line| bounds.extend_with_x(line.x));
--- a/egui/src/widgets/plot.rs
+++ b/egui/src/widgets/plot.rs
@@ -231,7 +231,8 @@ impl Curve {
 // ----------------------------------------------------------------------------
 
 /// Information about the plot that has to persist between frames.
-#[derive(Clone, PartialEq, Serialize, Deserialize, Debug)]
+#[cfg_attr(feature = "persistence", derive(serde::Deserialize, serde::Serialize))]
+#[derive(Clone)]
 struct PlotMemory {
     bounds: Bounds,
 }
--- a/egui/src/widgets/plot.rs
+++ b/egui/src/widgets/plot.rs
@@ -222,8 +222,8 @@ impl Curve {
     }
 
     /// Name of this curve.
-    pub fn name(mut self, name: impl ToString) -> Self {
-        self.name = name.to_string();
+    pub fn name(mut self, name: impl Into<String>) -> Self {
+        self.name = name.into();
         self
     }
 }
@@ -287,9 +287,9 @@ pub struct Plot {
 }
 
 impl Plot {
-    pub fn new(name: impl ToString) -> Self {
+    pub fn new(name: impl Into<String>) -> Self {
         Self {
-            name: name.to_string(),
+            name: name.into(),
             next_auto_color_idx: 0,
 
             curves: Default::default(),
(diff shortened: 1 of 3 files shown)


egui/src/widgets/plot.rs
```diff
@@ -35,7 +36,7 @@ impl Value {
 
 /// 2D bounding box of f64 precision.
 /// The range of data values we show.
-#[derive(Clone, Copy, PartialEq)]
+#[derive(Clone, Copy, PartialEq, Serialize, Deserialize, Debug)]
```
optozorax, line 39: It's strange that CI does not complain about this line. `persistence` feature is not enabled by default and I get errors when `cargo check` `egui` folder locally. @emilk CI didn't check this.

You should use serde like this:
```suggestion
#[derive(Clone, Copy, PartialEq, Debug)]
#[cfg_attr(feature = "persistence", derive(serde::Deserialize, serde::Serialize))]
```

egui/src/widgets/plot.rs
```diff
@@ -3,6 +3,7 @@
 #![allow(clippy::comparison_chain)]
 
 use color::Hsva;
+use serde::{Deserialize, Serialize};
```
optozorax, line 6: Under feature, it's better to not import this

```suggestion
```

egui/src/widgets/plot.rs
EmbersArc: Thanks for the review!

`cargo check` in the `egui` folder works for me even without the changes...

egui/src/widgets/plot.rs
```diff
@@ -239,18 +287,19 @@ pub struct Plot {
     show_y: bool,
 }
 
-impl Default for Plot {
-    fn default() -> Self {
+impl Plot {
+    pub fn new(name: impl Into<String>) -> Self {
```
emilk, line 291: I think this should be either changed to `id_source: impl Hash`, or we should actually use it as a `String` - i.e. show it on hover. Though then I wonder what the difference would be between a plot name and `on_hover_ui`-tooltip

egui/src/widgets/plot.rs
```diff
@@ -332,6 +365,12 @@ impl Plot {
         self
     }
 
+    /// If true, the bounds will be set based on the data.
+    pub fn automatic_bounds(mut self, enabled: bool) -> Self {
```
emilk, line 369: This is a bit under-documented (what is the default? what does it mean for it to be `false`?).

I wonder if we really need this at all? I like the behavior of automatically calculating bounds at the start and on double-click, but otherwise allow users to scroll and zoom.

egui/src/widgets/plot.rs
```diff
@@ -89,12 +97,34 @@ impl Bounds {
...
 
-    pub fn union_mut(&mut self, other: &Bounds) {
+    pub fn merge(&mut self, other: &Bounds) {
         self.min[0] = self.min[0].min(other.min[0]);
         self.min[1] = self.min[1].min(other.min[1]);
         self.max[0] = self.max[0].max(other.max[0]);
         self.max[1] = self.max[1].max(other.max[1]);
     }
+
+    pub fn shift_x(&mut self, delta: f64) {
```
emilk, line 107: I think [`translate`](https://en.wikipedia.org/wiki/Translation_(geometry)) is a more common word for this

egui/src/widgets/plot.rs
```diff
@@ -300,24 +347,10 @@ impl Plot {
...
-        self
-    }
-
-    /// Expand bounds to include the given y value.
-    /// For instance, to always show the x axis, call `plot.include_y(0.0)`.
-    pub fn include_y(mut self, y: impl Into<f64>) -> Self {
-        self.bounds.extend_with_y(y.into());
-        self
-    }
-
```
emilk, line 320: This is a feature I'd rather not lose. I think it makes sense if these would only apply to the default bounds, so that we would have `default_min_bounds: Bounds` as a member, which starts out `Bounds::EMPTY` and then is potentially expanded with `include_x` and `include_y`. The first frame (and on double-click) the automatic bounds would be computed from all curves in union with the `default_min_bounds`

emilk (changes requested):
Thank you for this PR; it works really well, and nice use of the new memory feature!

egui/src/widgets/plot.rs
s-nie: How about we use the same system that `Window` uses with a constructor that takes the shown name and an `id` method to set a custom id when required? The title will be visible to the user at some point, I just haven't thought about how to show it yet.

egui/src/widgets/plot.rs
s-nie: Changed in b3002ff

s-nie (commented):
Thanks for the review! I share your concerns about how to handle the plot bounds.
I got a bit carried away and started refactoring this whole thing quite a bit and added new features. If it's alright with you then I'll address the bounds issues there so that those changes don't have to be made twice. The bounds need a bit more thought anyway since there is no way for the user to define them manually or to make them fixed yet. I'll open a PR for these changes as soon as possible.

Is it alright with you when the plotting functionality keeps growing? In #155 you mentioned that you'd prefer if it didn't get too extensive. But I really like the foundation that you've built here and see potential for this to grow into something that can be used as much more than just a demo.

EmbersArc:
Thanks for the review! I share your concerns about how to handle the plot bounds.
I got a bit carried away and started refactoring this whole thing quite a bit and added new features. If it's alright with you then I'll address the bounds issues there so that those changes don't have to be made twice. The bounds need a bit more thought anyway since there is no way for the user to define them manually or to make them fixed yet. I'll open a PR for these changes as soon as possible.

egui/src/widgets/plot.rs
EmbersArc: How about we use the same system that Window uses with a constructor that takes the shown name and
an id method to set a custom id when required? The title will be visible to the user at some point,
I just haven't thought about how to show it yet.

egui/src/widgets/plot.rs
EmbersArc: Changed in b3002ff

emilk:
> If it's alright with you then I'll address the bounds issues there so that those changes don't have to be made twice

I'm a bit hesitant to merge a PR that removes a desired feature. Maybe you can open the other PR as a draft so I can take a look?

> Is it alright with you when the plotting functionality keeps growing?

I'm alright with that, but it would be good to sync exactly in what direction it is growing before you do too much work! Perhaps you can open a tracking issue with your plans?

EmbersArc:
>I'm a bit hesitant to merge a PR that removes a desired feature 

Understandable, I added back those methods, and made automatic bounds the default.

I'll open a tracking issue and a PR soon for the coming changes.

Question: Is there a way to tell if a window has been drawn for the first time?

egui/src/widgets/plot.rs
```diff
@@ -192,14 +220,23 @@ impl Curve {
     }
 
     /// Name of this curve.
-    pub fn name(mut self, name: impl Into<String>) -> Self {
-        self.name = name.into();
+    pub fn name(mut self, name: impl ToString) -> Self {
```
emilk, line 223:
```suggestion
    #[allow(clippy::needless_pass_by_value)]
    pub fn name(mut self, name: impl ToString) -> Self {
```

egui/src/widgets/plot.rs
```diff
@@ -235,20 +272,22 @@ pub struct Plot {
...
+    bounds: Bounds,
+
     show_x: bool,
     show_y: bool,
 }
 
-impl Default for Plot {
-    fn default() -> Self {
+impl Plot {
+    pub fn new(name: impl ToString) -> Self {
```
emilk, line 282:
```suggestion
    #[allow(clippy::needless_pass_by_value)]
    pub fn new(name: impl ToString) -> Self {
```

egui/src/widgets/plot.rs
```diff
@@ -425,9 +470,15 @@ impl Widget for Plot {
             vec2(width, height)
         };
 
-        let (rect, response) = ui.allocate_exact_size(size, Sense::hover());
+        let (rect, response) = ui.allocate_exact_size(size, Sense::drag());
 
-        let mut bounds = bounds;
+        if response.double_clicked_by(PointerButton::Primary) || bounds == Bounds::EMPTY {
```
emilk, line 475: There is a potential bug here: the `bounds` set by the builder is constrained by `include_x` and/or `include_y`. If the user calls `include_y(0.0)` to always show the X-axis, then the bounds won't `== EMPTY`, but they will be invalid. So the check should be `|| !bounds.is_valid()` with something like:

``` rust
impl Bounds {
    fn is_valid(&self) -> bool {
        self.is_finite() && self.width() > 0.0 && self.height() > 0.0
    }
}
```

egui/src/widgets/plot.rs
```diff
@@ -440,17 +491,16 @@ impl Widget for Plot {
...
-        bounds.expand_y(margin_fraction.y as f64 * bounds.height());
-
         if let Some(data_aspect) = data_aspect {
             let data_aspect = data_aspect as f64;
             let rw = rect.width() as f64;
             let rh = rect.height() as f64;
             let current_data_aspect = (bounds.width() / rw) / (bounds.height() / rh);
-            if current_data_aspect < data_aspect {
+
+            let margin = 1e-5;
```
emilk, line 500: A more common name for this is `epsilon` (something used to avoid rounding errors)

emilk (changes requested):
I just realized there is an important use-case that this breaks: plotting dynamic data. For instance, plotting the last few seconds of a recorded time series. In this case we want the bounds to automatically follow the data each frame.

I suggest a (hopefully) simple change: add a `auto_bounds: bool` to `PlotMemory` and initialize it to `true`. This is set to `false` on zoom/pan, and reset to `true` on double-tap. Then we will maintain previous behavior (tracking dynamic plots) while still allowing the new awesome zoom/pan behavior.

A second use case that need to work is the `include_x/include_y` (e.g. `.include_y(0.0)` to include the X-axis in a plot). This should be used when computing automatic bounds, but ignored when users pan and zoom. I suggest changing `Plot::bounds` to be called `Plot::min_auto_bounds` and use it as the starting point when computing automatic bounds.

egui/src/widgets/plot.rs
EmbersArc: Fixed in 1b00e7f

egui/src/widgets/plot.rs
EmbersArc: Fixed in 1b00e7f

EmbersArc:
Thanks for the great suggestions! I tried to apply all of them. Might not work in all cases thought, but I don't have time to test it more thoroughly right now.

EmbersArc:
I've opened up a PR based on this one: #331

emilk (approved):
Awesome work!

blagovestb2c2:
Hey guys, awesome library first of all. I found this PR after having problems with zoom. I am not sure how i'm supposed to achieve that on a Plot (mac with magic mouse if it matters). I tried reading the code and zoom seems to require hovering but i only see labels if i hover? And if I hover + track (or slide my finger on my mouse rather), the plot is just dragged. Am I doing something dumb? :D 

egui = "0.12.0"
epi = "0.12.0"
eframe = "0.12.0"

EmbersArc:
@blagovestb2c2 You need to hold control while scrolling to zoom

PR #324, opened 2021-04-19
made drag and drop support selectable on Windows for the eframe glium integration 

jnises:
to avoid issues with crates that use multi-threaded COM apis such as cpal
https://github.com/rust-windowing/winit/pull/1524

### DIFF
--- a/epi/src/lib.rs
+++ b/epi/src/lib.rs
@@ -155,6 +155,12 @@ pub trait App {
     fn transparent(&self) -> bool {
         false
     }
+
+    /// On Windows: enable drag and drop support.
+    /// Set to false to avoid issues with crates such as cpal which uses that use multi-threaded COM API (https://github.com/rust-windowing/winit/pull/1524)
+    fn drag_and_drop_support(&self) -> bool {
+        true
+    }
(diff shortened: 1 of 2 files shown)


epi/src/lib.rs
```diff
@@ -155,6 +155,12 @@ pub trait App {
     fn transparent(&self) -> bool {
         false
     }
+
+    /// On Windows: enable drag and drop support.
+    /// Set to false to avoid issues with crates such as cpal which uses that use multi-threaded COM API <https://github.com/rust-windowing/winit/pull/1524>
+    fn drag_and_drop_support(&self) -> bool {
```
jnises, line 161: perhaps it is better to keep this as a egui_glium-specific setting instead somehow? add arguments to eframe::run_native?

emilk: Yeah, I think this and a few of other things in `trait App` (`is_resizable`, `icon_data`, `decorated`, and `transparent`) should be moved to some options that is passed to `eframe::run_native`, but we can do that in a separate PR

emilk (approved):

PR #331, opened 2021-04-21
Plot refactor

### LINKED ISSUE #330: Plotting library: More features and cleanup
The current plotting library provides a good start for more advanced features. A good first step
towards this would be to split the module into multiple files. I'm thinking:

-  `plot.rs` for the core plot functionality
-  `items.rs` for anything that can be drawn in a plot
-  `transforms.rs` for things related to screen/plot bounds and transforms between them

I have a PR ready that would do those things. There, I also started investigating curve drawing from
user-provided functions. This could be a very useful feature but does not have to be included in
this PR.

EmbersArc:
Closes #330


EmbersArc:
Since I created some new files it's difficult to see what's changed, so here's a quick overview:

- Split plot.rs into multiple files as described in #330 
- Move more functionality from the plot builder into the [`ScreenTransform`](https://github.com/emilk/egui/pull/331/files#diff-ebfb704f132d0e483c73b29fd6adfb34f9cb9423044c854fa1d8b77389e94705R121) struct
- Allow users to create curves based on callback functions [`y=f(x)`](https://github.com/emilk/egui/pull/331/files#diff-c90f68457249d416c065da7421c47f86c4092840d2fba40d922a21037406dbceR113) and [`(x,y)=f(t)`](https://github.com/emilk/egui/pull/331/files#diff-c90f68457249d416c065da7421c47f86c4092840d2fba40d922a21037406dbceR126). The points are generated [here](https://github.com/emilk/egui/pull/331/files#diff-c90f68457249d416c065da7421c47f86c4092840d2fba40d922a21037406dbceR145) once the plot limits are known. This should be the only functional change and it would be great to get some feedback on it.

egui/src/widgets/plotting/items.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// Returns true if there are no data points available and there is no function to generate any.
+    pub(crate) fn no_data(&self) -> bool {
+        self.generator.is_none() && self.values.is_empty()
+    }
+
+    /// If initialized with a generator function, this will generate `n` evenly spaced points in the
+    /// given range.
+    pub(crate) fn generate_points(&mut self, mut x_range: RangeInclusive<f64>) {
+        let range_union = |range1: &RangeInclusive<f64>, range2: &RangeInclusive<f64>| {
+            range1.start().max(*range2.start())..=range1.end().min(*range2.end())
```
emilk, line 147: This is the intersection; not the union

emilk: This would also be better as `fn` instead of a lambda, since it doesn't capture anything.

egui/src/widgets/plotting/items.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    /// Draw a curve based on a function `y=f(x)`, an optional range for x and the number of points.
+    pub fn from_explicit_callback(
+        function: impl Fn(f64) -> f64 + 'static,
+        range: Option<RangeInclusive<f64>>,
+        points: usize,
+    ) -> Self {
+        let generator = Generator::Explicit(Box::new(function), range, points);
+        Self {
+            generator: Some(generator),
```
emilk, line 120: I think the x-axis of `self.bounds` should be initialized with the range (if finite) so that we get a better default bounds

egui/src/widgets/plotting/plot.rs
```diff
@@ -244,7 +22,7 @@ struct PlotMemory {
 ///
 /// ```
 /// # let ui = &mut egui::Ui::__test();
-/// use egui::plot::{Curve, Plot, Value};
+/// use egui::plotting::{Curve, Plot, Value};
```
emilk, line 25: This rename will be a breaking change for users. I think we should just call the module `plot` instead to avoid this.

egui/src/widgets/plotting/plot.rs
```diff
@@ -419,18 +171,59 @@ impl Plot {
...
+        self.min_auto_bounds.extend_with_x(x.into());
+        self
+    }
+
+    /// Expand bounds to include the given y value.
+    /// For instance, to always show the x axis, call `plot.include_y(0.0)`.
+    pub fn include_y(mut self, y: impl Into<f64>) -> Self {
+        self.min_auto_bounds.extend_with_y(y.into());
+        self
+    }
```
emilk, line 211: the diff becomes harder to read when moving around code like this. what is the motivation?

egui/src/widgets/plotting/plot.rs
```diff
@@ -345,33 +124,6 @@ impl Plot {
         self
     }
 
-    /// If true, the x-bounds will be symmetrical, so that the x=0 zero line
-    /// is always in the center.
-    pub fn symmetrical_x_bounds(mut self, symmetrical_x_bounds: bool) -> Self {
```
emilk, line 350: I agree that the new name (`center_x_axis`) is better, but keep the old function with an added `#[deprecated = "Renamed center_x_axis"]`

egui/src/widgets/plotting/plot.rs
```diff
@@ -419,18 +171,59 @@ impl Plot {
         self.show_y = show_y;
         self
     }
+
+    /// Always keep the x-axis centered.
```
emilk, line 175: Please document what the default is for this and the other methods (I know I hadn't, but better late than never)

egui/src/widgets/plotting/plot.rs
```diff
@@ -480,170 +272,86 @@ impl Widget for Plot {
...
-            value.x,
-            self.bounds.min[0]..=self.bounds.max[0],
-            (self.rect.left() as f64)..=(self.rect.right() as f64),
-        );
-        let y = remap(
-            value.y,
-            self.bounds.min[1]..=self.bounds.max[1],
-            (self.rect.bottom() as f64)..=(self.rect.top() as f64), // negated y axis!
+        // Initialize values from functions.
+        // TODO: Let the user pick a resolution?
```
emilk, line 331: The `from_explicit_callback` and `from_parametric_callback` already does, so this `TODO` is done?

egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -143,6 +143,8 @@ impl super::View for PlotDemo {
             .curve(self.circle())
             .curve(self.sin())
             .curve(self.thingy())
+            .center_x_axis(false)
+            .center_y_axis(false)
```
emilk, line 147: Since this is the default

```suggestion
```

egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -108,25 +108,25 @@ impl PlotDemo {
...
-            let t = remap(i as f64, 0.0..=(n as f64), -TAU..=TAU);
-            Value::new(t / 5.0, 0.5 * (self.time + t).sin())
-        });
-        Curve::from_values_iter(circle)
-            .color(Color32::from_rgb(200, 100, 100))
-            .name("0.5 * sin(x / 5)")
+        let time = self.time;
+        Curve::from_explicit_callback(
+            move |x| 0.5 * (2.0 * x).sin() * time.sin(),
+            Some((-2.)..=2.),
```
emilk, line 114: I think it would be a good demo for `sin` to span `-INF..=INF`

egui/src/widgets/plotting/items.rs
```diff
@@ -0,0 +1,214 @@
...
+    }
+}
+
+// ----------------------------------------------------------------------------
+
+enum Generator {
+    /// Describes a function y = f(x) with an optional range for x and a number of points.
+    Explicit(Box<dyn Fn(f64) -> f64>, Option<RangeInclusive<f64>>, usize),
+    /// Describes a function (x,y) = f(t) with a range for t and a number of points.
+    Parametric(Box<dyn Fn(f64) -> (f64, f64)>, RangeInclusive<f64>, usize),
```
emilk, line 71:
```suggestion
    /// Describes a function y = f(x) with a range for x and a number of points.
    Explicit{
        y_from_x: Box<dyn Fn(f64) -> f64>,
        x_range: RangeInclusive<f64>,
        resolution: usize,
    },
    /// Describes a function (x,y) = f(t) with a range for t and a number of points.
    Parametric{
        xy_from_t: Box<dyn Fn(f64) -> (f64, f64)>,
        t_range: RangeInclusive<f64>,
        resolution: usize
    },
```

I see no difference between `None` range and an `-INF..=INF` range, so better to unify to the simpler type

egui/src/widgets/plotting/items.rs
```diff
@@ -0,0 +1,214 @@
...
+        }
+    }
+
+    /// Draw a curve based on a function `(x,y)=f(t)`, a range for t and the number of points.
+    pub fn from_parametric_callback(
+        function: impl Fn(f64) -> (f64, f64) + 'static,
+        t_range: RangeInclusive<f64>,
+        points: usize,
+    ) -> Self {
+        let generator = Generator::Parametric(Box::new(function), t_range, points);
```
emilk, line 131: The parametric function always computes all the points, so there is not reason to defer that computation to later as far as I can tell.

If you compute values here instead and pass to `Self::from_values` we get default bounds and also a speedup, because `function` doesn't need to be boxed (and doesn't need to be `'static` either)

emilk (changes requested):
Very nice improvements and cleanups!

I left some comments, mainly:
* keep the `plot` name so we don't need to introduce a breaking change a
* keep old functions with a deprecation notice
* remove `Parametric` type and compute its values right away

egui/src/widgets/plot/items.rs
```diff
@@ -0,0 +1,223 @@
...
+            bounds,
+            ..Self::empty()
+        }
+    }
+
+    pub fn from_values_iter(iter: impl Iterator<Item = Value>) -> Self {
+        Self::from_values(iter.collect())
+    }
+
+    /// Draw a curve based on a function `y=f(x)`, an optional range for x and the number of points.
```
emilk, line 112:
```suggestion
    /// Draw a curve based on a function `y=f(x)`, a range (which can be infinite) for x and the number of points.
```

EmbersArc:
Thanks for the review! I agree with your suggestions and made all the changes in 2abfb0f. The remaining commits are just the filename changes.

emilk (approved):

emilk:
Thanks!

PR #341, opened 2021-04-30
memory: add `Send + Sync` reqirement, fix #337

### LINKED ISSUE #337: New memory not implementing `Send + Sync`
@mvlabat reported that new memory based on `Box<dyn Any + 'static>` didn't implement `Send` and `Sync`, and he can't send `CtxRef` to different places (maybe he use different threads for GUI and for other things).

This is possible to implement because `typemap` [has](https://docs.rs/typemap/0.3.3/typemap/type.ShareMap.html) this type of map.

But, do we want to restrict a user to only `Send + Sync` types in custom memory? Is it possible to have `!Send` and `!Sync` type that also implements `De-`/`Serialize`? (Sorry, I don't know much about this area of Rust)

If it's ok to require `Send + Sync` types in custom memory, then I will try to implement PR to fix this.

optozorax:
Closes https://github.com/emilk/egui/issues/337.

### DIFF
--- a/egui/src/memory.rs
+++ b/egui/src/memory.rs
@@ -500,3 +500,16 @@ impl Areas {
         wants_to_be_on_top.clear();
     }
 }
+
+// ----------------------------------------------------------------------------
+
+#[cfg(test)]
+#[test]
+fn memory_impl_send_sync() {
+    fn test_send_sync<T: Send + Sync>(t: T) {
+        drop(t);
+    }
+
+    let memory = Memory::default();
+    test_send_sync(memory);
+}
--- a/egui/src/memory.rs
+++ b/egui/src/memory.rs
@@ -506,10 +506,6 @@ impl Areas {
 #[cfg(test)]
 #[test]
 fn memory_impl_send_sync() {
-    fn test_send_sync<T: Send + Sync>(t: T) {
-        drop(t);
-    }
-
-    let memory = Memory::default();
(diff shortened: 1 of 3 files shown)


egui/src/memory.rs
```diff
@@ -500,3 +500,16 @@ impl Areas {
...
+
+#[cfg(test)]
+#[test]
+fn memory_impl_send_sync() {
+    fn test_send_sync<T: Send + Sync>(t: T) {
+        drop(t);
+    }
+
+    let memory = Memory::default();
+    test_send_sync(memory);
```
LPGhatguy, line 514: This is testable without instantiating a `Memory`!

```suggestion
    fn assert_send_sync<T: Send + Sync>() {}
    assert_send_sync::<Memory>();
```

emilk (approved):
Thank you!

egui/src/memory.rs
```diff
@@ -500,3 +500,12 @@ impl Areas {
...
 }
+
+// ----------------------------------------------------------------------------
+
+#[cfg(test)]
+#[test]
+fn memory_impl_send_sync() {
+    fn assert_send_sync<T: Send + Sync>() {}
+    assert_send_sync::<Memory>();
+}
```
mvlabat, line 511: I was suggesting adding a test for `CtxRef` itself :)

mvlabat:
It works great now, cheers!

PR #349, opened 2021-05-04
Add plot legends

### LINKED ISSUE #348: Add legends to the plotting library
I would like the plotting library to support legends, similar to those in implot:

image

Requirements:
* Show a legend of all named curves in the plot
* When hovering over entries, the corresponding curve should be highlighted
* Clicking entries should toggle the visibility of the corresponding curve

I've already started work on this and will open a PR soon.

EmbersArc:
Closes #348

I don't have much experience with designing widgets yet, so much of what I did here might not be "best practice". Would be great if someone could have a look and maybe give suggestions on how do do things differently.

gif

### DIFF
--- a/egui/src/widgets/plot/legend.rs
+++ b/egui/src/widgets/plot/legend.rs
@@ -1,0 +1,124 @@
+use std::string::String;
+
+use crate::*;
+
+pub(crate) struct LegendEntry {
+    pub text: String,
+    pub color: Color32,
+    pub checked: bool,
+    pub hovered: bool,
+}
+
+impl LegendEntry {
+    pub fn new(text: String, color: Color32, checked: bool) -> Self {
+        Self {
+            text,
+            color,
+            checked,
+            hovered: false,
+        }
+    }
+}
+
+struct LegendEntryWidget<'a> {
+    entry: &'a mut LegendEntry,
+}
+
+impl<'a> LegendEntryWidget<'a> {
+    fn new(entry: &'a mut LegendEntry) -> Self {
+        Self { entry }
+    }
+}
+
+impl<'a> Widget for LegendEntryWidget<'a> {
+    fn ui(self, ui: &mut Ui) -> Response {
+        let LegendEntry {
+            checked,
+            text,
+            color,
+            ..
+        } = self.entry;
+        let icon_width = ui.spacing().icon_width;
+        let icon_spacing = ui.spacing().icon_spacing;
+        let padding = vec2(2.0, 2.0);
+        let total_extra = padding + vec2(icon_width + icon_spacing, 0.0) + padding;
+
+        let text_style = TextStyle::Button;
+        let galley = ui.fonts().layout_no_wrap(text_style, text.clone());
+
+        let mut desired_size = total_extra + galley.size;
+        desired_size = desired_size.at_least(ui.spacing().interact_size);
+        desired_size.y = desired_size.y.at_least(icon_width);
+
+        ui.with_layout(Layout::top_down(Align::LEFT), |ui| {
+            let (rect, response) = ui.allocate_exact_size(desired_size, Sense::click());
+            let rect = rect.shrink2(padding);
+
+            response.widget_info(|| {
+                WidgetInfo::selected(WidgetType::RadioButton, *checked, &galley.text)
+            });
+
+            let visuals = ui.style().interact(&response);
+
+            let (small_icon_rect, big_icon_rect) = ui.spacing().icon_rectangles(rect);
+
+            let painter = ui.painter();
+
+            painter.add(Shape::Circle {
+                center: big_icon_rect.center(),
+                radius: big_icon_rect.width() / 2.0 + visuals.expansion,
+                fill: visuals.bg_fill,
+                stroke: visuals.bg_stroke,
+            });
+
+            if *checked {
+                painter.add(Shape::Circle {
+                    center: small_icon_rect.center(),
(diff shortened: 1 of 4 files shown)


egui/src/widgets/plot/legend.rs
```diff
@@ -0,0 +1,124 @@
...
+        let mut desired_size = total_extra + galley.size;
+        desired_size = desired_size.at_least(ui.spacing().interact_size);
+        desired_size.y = desired_size.y.at_least(icon_width);
+
+        ui.with_layout(Layout::top_down(Align::LEFT), |ui| {
+            let (rect, response) = ui.allocate_exact_size(desired_size, Sense::click());
+            let rect = rect.shrink2(padding);
+
+            response.widget_info(|| {
+                WidgetInfo::selected(WidgetType::RadioButton, *checked, &galley.text)
```
emilk, line 58: nit: This is (functionally) a `Checkbox`, since checking and unchecking it is independent from whether or not its siblings are checked

```suggestion
                WidgetInfo::selected(WidgetType::Checkbox, *checked, &galley.text)
```

egui/src/widgets/plot/mod.rs
```diff
@@ -345,8 +414,9 @@ impl Widget for Plot {
                     transform.zoom(zoom_factor, hover_pos);
                     auto_bounds = false;
                 }
-                let scroll_delta = ui.input().scroll_delta;
+                let mut scroll_delta = ui.input().scroll_delta;
                 if scroll_delta != Vec2::ZERO {
+                    scroll_delta.y *= -1.0;
```
emilk, line 419: Why this change? The scroll direction is correct on my mac - if it is wrong for you, please file a separate issue for that (it should then be fixed for more than just the plot!)

egui/src/widgets/plot/mod.rs
```diff
@@ -260,50 +275,104 @@ impl Widget for Plot {
...
+            curves.retain(|curve| !hidden_curves.contains(&curve.name));
+        }
+
+        // ---
+
         auto_bounds |= response.double_clicked_by(PointerButton::Primary);
 
         // Set bounds automatically based on content.
-        if auto_bounds || !bounds.is_valid() {
+        if auto_bounds {
```
emilk, line 375: why this change? seems like good behavior to recompute bounds if they ever become invalid

egui/src/widgets/plot/mod.rs
```diff
@@ -260,50 +275,104 @@ impl Widget for Plot {
...
+                    let text = curve.name.clone();
+                    let color = curve.stroke.color;
+                    LegendEntry::new(text, color, checked)
+                })
+                .collect();
+
+            // Show the legend.
+            if !legend_entries.is_empty() {
+                let legend = PlotLegend::new(&mut legend_entries, rect);
+                let response = legend.ui(ui);
```
emilk, line 342: Instead of doing manual placement of the legends, try this:

``` rust
// Put legend on top of plot:
let mut legend_ui = ui.child_ui(rect, Layout::top_down(Align::LEFT)); 
for entry in &mut legend_entries {
    if legend_ui.add(entry).hovered() {
        show_x = false;
        show_y = false;
    }
}
```

with 


``` rust
impl Widget for &mut LegendEntry {
    fn ui(self, ui: &mut Ui) -> Response {
        // same as before but WITHOUT `ui.with_layout(Layout::top_down(Align::LEFT), |ui| {`
    }
}
```


No need for `struct PlotLegend` or `struct LegendEntryWidget` !

egui/src/widgets/plot/legend.rs
```diff
@@ -0,0 +1,124 @@
...
+    entry: &'a mut LegendEntry,
+}
+
+impl<'a> LegendEntryWidget<'a> {
+    fn new(entry: &'a mut LegendEntry) -> Self {
+        Self { entry }
+    }
+}
+
+impl<'a> Widget for LegendEntryWidget<'a> {
```
emilk, line 33: Simpler alternative: `impl Widget for &mut LegendEntry`

emilk (changes requested):
This looks super nice! I left some comments on how you can simplify the legend layouting

egui/src/widgets/plot/mod.rs
EmbersArc: Interesting, I didn't think that this would be a platform issue. So on Linux, the two-finger scroll delta reported by `ui.input().scroll_delta` is positive to the right and up. I can enable "natural scrolling" in the settings and this reverses the direction on both axes. If this is different for you I can open an issue.

egui/src/widgets/plot/legend.rs
EmbersArc: Ah, makes sense. Changed in 6aaeae7.

egui/src/widgets/plot/mod.rs
EmbersArc: Changed it back in 6aaeae7.

egui/src/widgets/plot/mod.rs
EmbersArc: Ha! So good, I knew there would be a better way but had no idea that it would be _that_ elegant.

egui/src/widgets/plot/mod.rs
emilk: Yeah, it is different for me :( Please open an issue, and we can collect data on what platforms this happens on (native vs web vs os etc)

emilk (approved):
Really nice - thank you! I like the extra touch of highlighting the curve whose legend you hover

PR #352, opened 2021-05-06
implement mint conversions

### LINKED ISSUE #314: Mint Integration!
`egui` has been an awesome solution to getting ui up and running for for my game. For all of my coordinate systems, I'm currently using `glam` and would love an easy way to translate some of my vectors easily into egui's format without having to unpack them every time.

Fortunately, that's why Mint exists: https://github.com/kvark/mint

Mint is a data / representation only linear algebra library meant to be used as a unified intermediate step when converting between linear algebra representations. All each library has to do is implement `from<mint::types>` for their own representations, and users can easily convert between them.

Here's a short list of libs that implement mint conversions:
- `cgmath`
- `euclid`
- `nalgebra`
- `glam`
- `kurbo`
- `vek`

The library itself is fairly lightweight, (7.5kb), has a single (optional) dependency on `serde` and could be enabled / disabled with a feature easily.

To my knowledge `mint` is the best supported interoperability lib for math-related types. If `egui` had consumes another library's types, I would suggest to just re-export them. But `egui` wrote them from scratch, so the next best solution would be to implement conversions for `mint`

This would be a pretty lightweight feature, and would be happy to contribute it, I just wanted to raise the idea first, and gauge how open y'all are to the idea. 

In general, I would just like to thank y'all for the work on this lib. It's been a pleasure to use, and I hope to improve it even more.

LU15W1R7H:
Implement conversions for [mint](https://docs.rs/mint) (math interoperability standard types).

- `impl {From, Into}<mint::Point2> for Pos2`
- `impl {From, Into}<mint::Vector2> for Vec2`

Closes #314

### DIFF
--- a/emath/src/pos2.rs
+++ b/emath/src/pos2.rs
@@ -85,6 +85,23 @@ impl From<&Pos2> for (f32, f32) {
 }
 
 // ----------------------------------------------------------------------------
+// Mint compatibility and convenience conversions
+
+#[cfg(feature = "mint")]
+impl From<mint::Point2<f32>> for Pos2 {
+    fn from(v: mint::Point2<f32>) -> Self {
+        Self::new(v.x, v.y)
+    }
+}
+
+#[cfg(feature = "mint")]
+impl From<Pos2> for mint::Point2<f32> {
+    fn from(v: Pos2) -> Self {
+        Self { x: v.x, y: v.y }
+    }
+}
+
+// ----------------------------------------------------------------------------
 
 impl Pos2 {
     /// The zero position, the origin.
--- a/emath/src/vec2.rs
+++ b/emath/src/vec2.rs
@@ -82,6 +82,23 @@ impl From<&Vec2> for (f32, f32) {
 }
 
 // ----------------------------------------------------------------------------
+// Mint compatibility and convenience conversions
+
+#[cfg(feature = "mint")]
+impl From<mint::Vector2<f32>> for Vec2 {
+    fn from(v: mint::Vector2<f32>) -> Self {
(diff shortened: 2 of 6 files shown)


emath/src/pos2.rs
```diff
@@ -84,6 +84,26 @@ impl From<&Pos2> for (f32, f32) {
     }
 }
 
+// ----------------------------------------------------------------------------
+// Compatibility and convenience conversions to and from (f32, f32):
```
emilk, line 88: rm misleading comment 

```suggestion
```

emath/src/vec2.rs
```diff
@@ -81,6 +81,26 @@ impl From<&Vec2> for (f32, f32) {
     }
 }
 
+// ----------------------------------------------------------------------------
+// Compatibility and convenience conversions to and from (f32, f32):
```
emilk, line 85: rm misleading comment 

```suggestion
```

emilk (changes requested):
Cool!

So an `egui` user would need to add `emath = { version = 0.11, features = ["mint"] }` to turn this on, correct? Would it maybe make sense to add `mint` as a feature in `egui` too, so that users could just change `egui = "0.11"` into `egui = { version = "0.11", features = ["mint"] }` without having to know about `emath`?

emath/src/pos2.rs
LU15W1R7H: Sorry, forgot to adjust the comment 😅

LU15W1R7H:
I fixed the clippy issue.

emilk (approved):
Thanks!

Issue #356, opened 2021-05-07
Native: Scroll direction varies by platform


EmbersArc:
**Describe the bug**
As we noticed in https://github.com/emilk/egui/pull/349#discussion_r627701110:
On Linux, the two-finger scroll delta reported by `ui.input().scroll_delta` is positive to the right and up. I can enable "natural scrolling" in the settings and this reverses the direction on both axes. Apparently this is different on other platforms, namely on Mac.

**To Reproduce**
1. Use `ui.input().scroll_delta` to get the scroll delta when using the touchpad with two fingers
2. The signs of the values may vary depending on the platform that egui is used on

**Expected behavior**
That the reported values are always the same.

**Desktop:**
Operating System: Manjaro Linux
KDE Plasma Version: 5.21.4
KDE Frameworks Version: 5.81.0
Qt Version: 5.15.2
Kernel Version: 5.12.1-2-MANJARO
OS Type: 64-bit
Graphics Platform: X11
Processors: 12 × Intel® Core™ i7-9750H CPU @ 2.60GHz
Memory: 15.3 GiB of RAM
Graphics Processor: GeForce GTX 1650 with Max-Q Design/PCIe/SSE2

emilk:
On my macbook (macOS Mojave) the positive scroll direction is down and to the right (with "natural scrolling" turned on). This is the correct direction, and same as the positive mouse delta.

emilk:
I'd like to get a few more voices on this before taking action. This may be an issue with KDE, Linux, or all non-Mac platforms.

So please test the scroll direction by checking out latest `master`, `cargo run --release` and check the `scroll_delta` values in the "Inspection" window.

follower:
It might be helpful to re-title this issue "...scroll delta *sign/direction*..." given that there ~~is~~ _was_ also a recently existing issue (#159) which is actually about the magnitude of the scroll value.

akhilman:
I tested two finger scrolling on Xfce-4.12 with "Reverse scroll direction" enabled.
Moving fingers up and right gives negative values, left and down - positive values.
Overall the up/down scroll direction is correct (as in any other app). But the speed is low.

emilk:
This should be worked on in winit if anyone has the energy to work on it

aevyrie:
On windows, I'm experiencing _horizontal_ scroll going the wrong direction.
Scrolling anywhere other than egui results in "natural" scrolling in x and y. 
[...]

PR #357, opened 2021-05-08
egui_glium: Flip vertical scroll direction on Linux

emilk:
This is to address https://github.com/emilk/egui/issues/356

Draft until we have more data on exactly what is wrong where


akhilman:
In gnome wayland, this fix causes scrolling in the egui app to go in the opposite direction to all other apps. Without this fix, scrolling goes in the same direction as the entire desktop. Tested with natural scrolling enabled and disabled.

emilk:
Let's leave it to winit to fix this then

PR #363, opened 2021-05-09
Plotting: Add line markers

### LINKED ISSUE #362: Plotting: Add line markers
It would be great to have line markers for curves in the plotting library. It's mostly done but there are still some open questions that I'll have to figure out when it comes to user friendliness. I'll open a draft PR soon.

image

EmbersArc:
Closes #362 

https://user-images.githubusercontent.com/1352472/117843780-6b8e2300-b27f-11eb-9534-485d3d34920b.mp4


egui/src/widgets/plot/items.rs
```diff
@@ -73,13 +73,317 @@ struct ExplicitGenerator {
...
+        self
+    }
+
+    /// Set the maximum extent of the marker around its position.
+    pub fn radius(mut self, radius: f32) -> Self {
+        self.radius = radius;
+        self
+    }
+
+    pub(crate) fn get_shapes(&self, position: &Pos2) -> Vec<Shape> {
```
emilk, line 217: To reduce the number of allocations it would be better to pass in a `out_shapes: &mut Vec<Shape>` and push into that

egui/src/widgets/plot/items.rs
```diff
@@ -73,13 +73,317 @@ struct ExplicitGenerator {
...
+                    vec2(-frac_1_sqrt_2, -frac_1_sqrt_2),
+                    vec2(frac_1_sqrt_2, frac_1_sqrt_2),
+                ]);
+                let diagonal2 = tf(vec![
+                    vec2(frac_1_sqrt_2, -frac_1_sqrt_2),
+                    vec2(-frac_1_sqrt_2, frac_1_sqrt_2),
+                ]);
+                vec![
+                    Shape::line(diagonal1, default_stroke),
+                    Shape::line(diagonal2, default_stroke),
```
emilk, line 296: Use `Shape::line_segment` instead

egui/src/widgets/plot/items.rs
```diff
@@ -73,13 +73,317 @@ struct ExplicitGenerator {
...
+            radius,
+        } = *self;
+
+        if color == Color32::TRANSPARENT {
+            return Vec::new();
+        }
+
+        let stroke_size = radius / 5.0;
+
+        let tf = |offset: Vec<Vec2>| -> Vec<Pos2> {
```
emilk, line 235: this will cause a lot of extra allocations. Just make it a `|Vec2| -> Pos2` and call it per point instead

egui/src/widgets/plot/items.rs
```diff
@@ -73,13 +73,317 @@ struct ExplicitGenerator {
...
+                    Shape::line(diagonal1, default_stroke),
+                    Shape::line(diagonal2, default_stroke),
+                ]
+            }
+            MarkerShape::Plus => {
+                let horizontal = tf(vec![vec2(-1.0, 0.0), vec2(1.0, 0.0)]);
+                let vertical = tf(vec![vec2(0.0, -1.0), vec2(0.0, 1.0)]);
+                vec![
+                    Shape::line(horizontal, default_stroke),
+                    Shape::line(vertical, default_stroke),
```
emilk, line 304: Use `Shape::line_segment` instead

egui/src/widgets/plot/items.rs
```diff
@@ -73,13 +73,317 @@ struct ExplicitGenerator {
...
+                }]
+            }
+            MarkerShape::Asterisk => {
+                let vertical = tf(vec![vec2(0.0, -1.0), vec2(0.0, 1.0)]);
+                let diagonal1 = tf(vec![vec2(-frac_sqrt_3_2, 0.5), vec2(frac_sqrt_3_2, -0.5)]);
+                let diagonal2 = tf(vec![vec2(-frac_sqrt_3_2, -0.5), vec2(frac_sqrt_3_2, 0.5)]);
+                vec![
+                    Shape::line(vertical, default_stroke),
+                    Shape::line(diagonal1, default_stroke),
+                    Shape::line(diagonal2, default_stroke),
```
emilk, line 370: Use `Shape::line_segment` instead

egui/src/widgets/plot/items.rs
```diff
@@ -73,13 +73,317 @@ struct ExplicitGenerator {
...
+
 /// A series of values forming a path.
 pub struct Curve {
     pub(crate) values: Vec<Value>,
     generator: Option<ExplicitGenerator>,
     pub(crate) bounds: Bounds,
-    pub(crate) stroke: Stroke,
+    pub(crate) marker: Option<Marker>,
+    pub(crate) color: Option<Color32>,
+    pub(crate) width: f32,
```
emilk, line 384: It is reasonable to want to use a different marker color from the curve color?

emilk, line 384: Why replace `Stroke` with color + widht?

egui/src/widgets/plot/items.rs
```diff
@@ -223,4 +544,13 @@ impl Curve {
...
         self
     }
+
+    /// Return the color by which the curve can be identified.
+    pub(crate) fn get_color(&self) -> Option<Color32> {
+        self.color.filter(|color| color.a() != 0).or_else(|| {
+            self.marker
+                .map(|marker| marker.color)
+                .filter(|color| *color != Color32::TRANSPARENT)
+        })
```
emilk, line 554: this again makes me think we should only have one color. Likely users will either show markers or a line. If they want to show both, it is likely they will be separate anyway (e.g. show a smooth line and the sparse points they interpolate)

egui/src/widgets/plot/mod.rs
```diff
@@ -100,31 +99,50 @@ impl Plot {
...
+        curves.into_iter().fold(self, Self::curve)
+    }
+
     /// Add a horizontal line.
     /// Can be useful e.g. to show min/max bounds or similar.
     /// Always fills the full width of the plot.
     pub fn hline(mut self, mut hline: HLine) -> Self {
-        self.auto_color(&mut hline.stroke.color);
+        if hline.stroke.color.a() == 0 {
+            hline.stroke.color = self.auto_color();
```
emilk, line 144: this will overwrite additive colors

emilk (changes requested):
The markers look sweet!

I'm not sure it makes sense to couple them to `Curve` though.

It seems likely users will want them to show e.g. a scatterplot, or to show the datapoint to which a smooth curve has been fitted. But right now it is more designed to highlight the corners of a coarse curve (like the demo shows), which is just a weird use case to me (but that may just be me). Such a coarse curve isn't very... well, curvy!think 

Consider an alternative where markers where instead part of a separate `Points` primitive. So you can add a curve and points, or just points, or just a curve. That way we can keep `Curve` for smooth things (and add methods to e.g. show the derivative/tangent on hover) and use the markers for sparse point data.

What do you think?

egui/src/widgets/plot/mod.rs
EmbersArc: Do you mean I shouldn't check if alpha == 0 and instead compare against `Color32::TRANSPARENT`?

EmbersArc:
@emilk Thanks for the review! I'm very much open to having a separate `Points` primitive, in fact I was going to take this approach first. But everything is very much built around `Curve` right now and introducing a separate `Points` struct would lead to a lot of duplicate code. Maybe we could unite them in a `Series` primitive which has a field that defines whether it's a curve or a set of points? Maybe you have an idea?

EmbersArc:
@emilk I've now created a separate `Points` primitive, let me know what you think. It has lead to some duplicate code but made other parts much cleaner and straight forward. Both `Curve`s and `Point`s now have just one constructor, which takes a `ValuesSeries` object, like so:

```rust
let curve = Curve::new(ValueSeries::from_values_iter(sin));
let points = Points::new(ValueSeries::from_values_iter(sin));
ui.add(
    Plot::new("Test Plot").curve(curve).points(points)
);
```

EmbersArc:
One issue with this approach (apart from the duplicate code) is that points are always drawn on top of curves. But I want the user to have full control over the drawing order. One way to fix both issues might be to share an interface between `Points` and `Curve` and then store them in a vector of boxes. I can give that a shot...

EmbersArc:
Done, I'm quite happy with how this turned out. I also took the chance to rename `Curve` to `Line` since it's more general and we're breaking a lot here anyways.

egui/src/widgets/plot/items.rs
EmbersArc: Probably not, irrelevant after ae00424.

EmbersArc: ae00424 allowed me to go back to `Stroke`

egui/src/widgets/plot/items.rs
EmbersArc: Done in ae00424

egui/src/widgets/plot/mod.rs
emilk: yeah, exactly - `hline.stroke.color.a() == 0` is true for transparent colors, but also for additive colors (which is pretty nice to use in a plot with dark background). `color == Color32::TRANSPARENT` is the correct check here

egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -85,7 +65,7 @@ impl PlotDemo {
                 });
             });
 
-            ui.vertical(|ui| {
+            ui.horizontal(|ui| {
```
emilk, line 68: Unfortunately this makes the plot window very wide, which makes it now very nice on phones. Maybe `ui.horizontal_wrapped` ?

egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -85,7 +65,7 @@ impl PlotDemo {
...
                 ui.style_mut().wrap = Some(false);
                 ui.checkbox(animate, "animate");
                 ui.add_space(8.0);
```
emilk, line 71: the space looks weird with the new horizontal layout

egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -116,55 +96,158 @@ impl PlotDemo {
...
+            .show(ctx, |ui| self.ui(ui));
+    }
+}
+
+impl super::View for PlotDemo {
+    fn ui(&mut self, ui: &mut Ui) {
+        ui.vertical_centered(|ui| {
+            egui::reset_button(ui, self);
+            ui.add(crate::__egui_github_link_file!());
+        });
```
emilk, line 249: Maybe the "Pan by dragging…" instructions could be moved here, since it is common for both the lines and markers demo

emilk (changes requested):
Much better!

I only have some minor notes

egui/src/widgets/plot/mod.rs
```diff
@@ -100,31 +100,53 @@ impl Plot {
...
         self
     }
 
     /// Add a horizontal line.
     /// Can be useful e.g. to show min/max bounds or similar.
     /// Always fills the full width of the plot.
     pub fn hline(mut self, mut hline: HLine) -> Self {
-        self.auto_color(&mut hline.stroke.color);
+        if hline.stroke.color != Color32::TRANSPARENT {
+            hline.stroke.color = self.auto_color();
```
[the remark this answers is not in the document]
EmbersArc: Ah good catch!

egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -116,55 +96,158 @@ impl PlotDemo {
...
+                let y_offset = i as f32 * 0.5 + 1.0;
+                let mut points = Points::new(ValueSeries::from_values(vec![
+                    Value::new(1.0, 0.0 + y_offset),
+                    Value::new(2.0, 0.5 + y_offset),
+                    Value::new(3.0, 0.0 + y_offset),
+                    Value::new(4.0, 0.5 + y_offset),
+                    Value::new(5.0, 0.0 + y_offset),
+                    Value::new(6.0, 0.5 + y_offset),
+                ]))
+                .name("Marker Lines")
```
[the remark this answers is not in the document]
EmbersArc: Good idea :+1:

egui/src/widgets/plot/items.rs
```diff
@@ -182,18 +176,81 @@ impl Curve {
...
+    pub(super) stroke: Stroke,
+    pub(super) name: String,
+    pub(super) highlight: bool,
+}
+
+impl Line {
+    pub fn new(series: ValueSeries) -> Self {
+        Self {
+            series,
+            stroke: Stroke::new(1.0, Color32::TRANSPARENT),
```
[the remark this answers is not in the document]
EmbersArc: Yes, looks a bit better in my opinion and 1 seems more "default".

egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -116,55 +96,158 @@ impl PlotDemo {
...
+}
+
+impl super::View for PlotDemo {
+    fn ui(&mut self, ui: &mut Ui) {
+        ui.vertical_centered(|ui| {
+            egui::reset_button(ui, self);
+            ui.add(crate::__egui_github_link_file!());
+        });
+        ui.collapsing("Lines", |ui| ui.add(&mut self.line_demo));
+        ui.collapsing("Markers", |ui| ui.add(&mut self.marker_demo));
```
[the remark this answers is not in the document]
EmbersArc: Ah, I like it.

EmbersArc:
@emilk Thanks for having another look. Much appreciated. I applied all your suggestions.

emilk (approved):
Thanks for this!

PR #364, opened 2021-05-09
Grid: Row/column colors

skairunner:
~~This is a draft that I will hopefully keep working on.~~

I added a "header_rows" option that simply sets the color of the row to a preset if it is true. 

|Dark|Light|
|-|-|
|image|image|

Thoughts:
* This is extremely hacky because I just wanted to figure out how the system works.
* It doesn't look good because the text color is not being appropriately set. The dark mode color isn't sufficiently header-y enough.
* It's hacky to force the user to set the text color appropriately if they just want a header.

Next work:
* I will instead make it so you can color arbitrary cells and/or rows, possibly by providing a predicate
* In a separate PR, I will make a "Table" widget that is specialized for tabular data that will contain auto-formatting for header rows, filtering, and other nice gadgets. It will wrap the Grid layout.


skairunner:
Took a couple attempts, but I think this is a good approach.

I refactored grid a bit to add GuiColor (either a single color, or a pair of colors corresponding to
the light/dark scheme) and ColorSpec (which describes how rows/cols/cells should be colored). I then refactored `striped` and `header_row` to simply add a new predicate to the spec when transforming from Grid to GridLayout. 

I think GuiColor could be a good concept to introduce through the rest of the codebase, since light/dark is a first class citizen here. Future work could possibly generalize it into color schemes?

Next work:
- Add all the other color modes (col, cell, list)
- Nicer ways to use this functionality?

skairunner:
image

Here's cell support.

image

Can also have both rows and columns. Even cell-based, if you want. 

It took me a lot of trial and error and fiddling to get the cells and rows to line up perfectly. Maybe in the future it would be nice to formalize how the grid is laid out, because as of right now any change in the logic will make the grid no longer line up.

Also, the option to add padding to labels would probably be good too.

Next work:
* Figure out how to run check.sh locally -- I installed `speech-dispatcher` but the build script fails to find it, which kills the entire process.

emilk (changes requested):
Sorry for not getting to this earlier!

Custom colors for rows and columns is great!

I think the API and implementation could be made a lot simpler if we change the per-column and per-row coloring to just be a single lambda, like so:

```
grid.with_row_color(|row| {
    if row == 0 { Some(header_color) } else { None  }
});
```

No need for `SinglePredicate`, and no need to loop over multiple predicates. The default predicate would simply be `|_| None`. The type would be `column_color: Fn(usize) -> Option<Color32>`.

---

`GuiColor` is interesting, but I'm not sure it is the right choice. It needs discussion in a separate issue first. If we gonna have something like that, we should have it for every widget. And I'm not 100% it is what the user want. We may have more than two themes, and it would get weird quickly.

--- 

Per-cell colors I think is something that is better accomplished by the user with a `Frame` instead. That is, if you want a grid cell to have a background color you can do:

``` rust
let frame = egui::Frame { fill: my_color, ..Default::default() };
frame.show(|ui| {
    cell_content(ui);
});
```

This is also more flexible, as the user can specify more options (rounding, margin, stroke, ...)

skairunner:
> I think the API and implementation could be made a lot simpler if we change the per-column and per-row coloring to just be a single lambda, like so:

Sounds good to me. I wasn't aware of `Frame` and it seems like a good idea.

> GuiColor is interesting, but I'm not sure it is the right choice. 

The immediate problem is pretty annoying, though. It means that if you want to cleanly support two different UI schemes you need to have conditionals everywhere colors are used -- both in library code and userspace code -- and that frankly seems like a bad solution, especially when the library already has first-class support for light mode.

egui/src/grid.rs
```diff
@@ -182,6 +259,129 @@ impl GridLayout {
...
+    pub(crate) fn paint_column(&self, col: usize, min: Pos2, color: GuiColor, painter: &Painter) {
+        let col_f = col as f32;
+
+        let color = match color {
+            GuiColor::Single(color) => color,
+            GuiColor::Pair { light, dark } => match self.style.visuals.dark_mode {
+                true => dark,
+                false => light,
+            },
+        };
```
emilk, line 292: Instead of duplicating this, make a helper member function:

``` rust
let color = color.pick(&self.style);
```

egui/src/grid.rs
```diff
@@ -32,6 +47,68 @@ impl State {
...
+}
+
+impl GuiColor {
+    /// Create a single color
+    pub fn single(color: Rgba) -> Self {
+        Self::Single(color)
+    }
+
+    /// Create a light/dark scheme pair
+    pub fn pair(light: Rgba, dark: Rgba) -> Self {
```
emilk, line 79:
```suggestion
    pub fn light_dark(light: Rgba, dark: Rgba) -> Self {
```

emilk:
> The immediate problem is pretty annoying, though. It means that if you want to cleanly support two different UI schemes you need to have conditionals everywhere colors are used -- both in library code and userspace code -- and that frankly seems like a bad solution, especially when the library already has first-class support for light mode.

You make an excellent point. I added a comment about how it can be improved a bit.

egui/src/grid.rs
```diff
@@ -182,6 +252,110 @@ impl GridLayout {
...
 
+    /// Paint a row.
+    pub(crate) fn paint_row(&self, min: Pos2, color: GuiColor, painter: &Painter) {
+        let color = match color {
+            GuiColor::Single(color) => color,
+            GuiColor::LightDark { light, dark } => match self.style.visuals.dark_mode {
+                true => dark,
+                false => light,
+            },
+        };
```
emilk, line 263:
```suggestion
        let color = color.pick(&self.style);
```

egui/src/grid.rs
```diff
@@ -182,6 +252,110 @@ impl GridLayout {
...
+    pub(crate) fn paint_column(&self, col: usize, min: Pos2, color: GuiColor, painter: &Painter) {
+        let col_f = col as f32;
+
+        let color = match color {
+            GuiColor::Single(color) => color,
+            GuiColor::LightDark { light, dark } => match self.style.visuals.dark_mode {
+                true => dark,
+                false => light,
+            },
+        };
```
emilk, line 285:
```suggestion
        let color = color.pick(&self.style);
```

egui/src/grid.rs
```diff
@@ -182,6 +251,72 @@ impl GridLayout {
...
+        // Only paint columns when we're on the first row.
+        // They are painted for the entire grid.
+        if self.row == 0 {
+            for p in &self.color_spec.col_pickers {
+                // Iterate over columns
+                for col in 0..self.prev_state.dimensions().y {
+                    if let Some(color) = p(col) {
+                        self.paint_column(col, min, color, painter);
+                    }
+                }
```
emilk, line 315: O(N²) !

Each column sums up the width of the previous columns, and you do that for each column.

Better to keep a running sum here, and then we don't need `struct Dimensions`:

egui/src/grid.rs
```diff
@@ -182,6 +251,72 @@ impl GridLayout {
...
+                3.0,
+            );
+
+        let rect = Rect::from_min_size(min_offset, size);
+
+        painter.rect_filled(rect, 2.0, color);
+    }
+
+    /// Paint the background for the next row & columns.
+    pub(crate) fn do_paint(&mut self, min: Pos2, painter: &Painter) {
```
emilk, line 300:
```suggestion
    pub(crate) fn paint(&mut self, cursor_min: Pos2, painter: &Painter) {
```

egui/src/grid.rs
```diff
@@ -336,16 +491,46 @@ impl Grid {
...
+                    } else {
+                        None
+                    }
+                });
+            }
+
+            // ---
+
             ui.set_grid(grid);
+            ui.start_row();
```
emilk, line 533: Instead of adding two `fn start_row`, just have `Placer::set_grid` call the correct function directly. In this case I *believe* what you are doing is painting the top row?

egui/src/placer.rs
```diff
@@ -178,6 +178,12 @@ impl Placer {
         self.region.expand_to_include_rect(frame_rect); // e.g. for centered layouts: pretend we used whole frame
     }
 
+    pub(crate) fn start_row(&mut self, painter: &Painter) {
+        if let Some(grid) = &mut self.grid {
+            grid.do_paint(self.region.cursor.min, painter)
+        }
+    }
+
```
emilk, line 186: This shouldn't be needed

egui/src/ui.rs
```diff
@@ -1576,6 +1576,10 @@ impl Ui {
             .end_row(self.spacing().item_spacing, &self.painter().clone());
     }
 
+    pub(crate) fn start_row(&mut self) {
+        self.placer.start_row(&self.painter().clone());
+    }
+
```
emilk, line 1582: This shouldn't be needed

egui_demo_lib/src/apps/demo/tests.rs
```diff
@@ -1,3 +1,5 @@
+use egui::Rgba;
```
emilk, line 1: please no top-level `use` statements in the demo lib - I try to be as explicit as possible to make it easier for people to copy-paste the code they want

egui_demo_lib/src/apps/demo/tests.rs
```diff
@@ -205,47 +213,62 @@ impl super::Demo for TableTest {
...
         ui.add(
             egui::Slider::new(&mut self.max_col_width, 0.0..=400.0).text("Maximum column width"),
         );
-        ui.add(egui::Slider::new(&mut self.num_cols, 0..=5).text("Columns"));
+        ui.add(egui::Slider::new(&mut self.num_cols, 0..=8).text("Columns"));
         ui.add(egui::Slider::new(&mut self.num_rows, 0..=20).text("Rows"));
 
+        ui.horizontal(|ui| {
+            ui.add(egui::Checkbox::new(&mut self.striped, "Striped"));
+            ui.add(egui::Checkbox::new(&mut self.with_header, "Headers"));
```
emilk, line 229:
```suggestion
            ui.add(egui::Checkbox::new(&mut self.with_header, "Highlight Header"));
```

egui_demo_lib/src/apps/demo/tests.rs
```diff
@@ -205,47 +213,62 @@ impl super::Demo for TableTest {
...
+            ui.add(egui::Checkbox::new(&mut self.with_columns, "Columns"));
```
emilk, line 230:
```suggestion
            ui.add(egui::Checkbox::new(&mut self.with_columns, "Highlight columns"));
```

egui_demo_lib/src/apps/demo/tests.rs
```diff
@@ -205,47 +213,62 @@ impl super::Demo for TableTest {
...
         );
         ui.add(
             egui::Slider::new(&mut self.max_col_width, 0.0..=400.0).text("Maximum column width"),
         );
-        ui.add(egui::Slider::new(&mut self.num_cols, 0..=5).text("Columns"));
+        ui.add(egui::Slider::new(&mut self.num_cols, 0..=8).text("Columns"));
         ui.add(egui::Slider::new(&mut self.num_rows, 0..=20).text("Rows"));
 
+        ui.horizontal(|ui| {
+            ui.add(egui::Checkbox::new(&mut self.striped, "Striped"));
```
emilk, line 228: Maybe we should rename this to "highlight_rows` or something, and add a similar helper for columns?

"Striped" is ambiguous in the direction (my bad).

egui_demo_lib/src/apps/demo/demo_app_windows.rs
```diff
@@ -82,7 +82,7 @@ impl Default for Tests {
             Box::new(super::tests::InputTest::default()),
             Box::new(super::layout_test::LayoutTest::default()),
             Box::new(super::tests::ManualLayoutTest::default()),
-            Box::new(super::tests::TableTest::default()),
+            Box::new(super::tests::GridTest::default()),
```
emilk, line 85: Good name change, but please sort the list alphabetically by window name

egui/src/grid.rs
```diff
@@ -190,22 +325,7 @@ impl GridLayout {
...
-
-                let color = if self.style.visuals.dark_mode {
-                    Rgba::from_white_alpha(0.0075)
-                } else {
-                    Rgba::from_black_alpha(0.075)
-                };
-                painter.rect_filled(rect, 2.0, color);
-            }
-        }
+        self.do_paint(cursor.min, painter);
```
emilk, line 328: `do_paint` checks if `row == 0` which it never is here. Better to just call `paint_row` here to avoid confusion:

```suggestion
        self.paint_row(cursor, painter);
```

egui/src/grid.rs
```diff
@@ -280,6 +428,11 @@ impl Grid {
         self
     }
 
+    pub fn header_row(mut self, header_row: bool) -> Self {
+        self.header_row = header_row;
```
emilk, line 432: This can just add to `color_spec.row_pickers` directly. Same for `striped`

egui/src/grid.rs
```diff
@@ -263,14 +385,40 @@ impl Grid {
...
+    }
+
+    /// Add a `ColorPickerFn` to the color spec without overwriting existing column colors.
+    pub fn add_column_color(mut self, predicate: ColorPickerFn) -> Self {
+        self.color_spec.col_pickers.push(predicate);
+        self
+    }
+
+    /// Replace the color picker for rows
+    pub fn with_row_color(mut self, predicate: ColorPickerFn) -> Self {
```
emilk, line 411: Do we really need both `add` and `with`? Seems like we can do without the `with` (why would anyone add colors and then replace them later?)

egui/src/grid.rs
```diff
@@ -7,6 +7,13 @@ pub(crate) struct State {
     row_heights: Vec<f32>,
 }
 
+/// Describe the dimensions of a grid state
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct Dimensions {
+    x: usize,
+    y: usize,
+}
+
```
emilk, line 16:
```suggestion
```

emilk (changes requested):
The way the rows and columns are painted is very convoluted.

It should be possible to make it a lot simpler.

I think `Placer::set_grid` can take care of painting the columns and the first row

egui/src/grid.rs
skairunner: 👍

egui/src/grid.rs
skairunner: I don't think `set_grid` has access to the Painter, because that's passed in when calling grid-related functions from `Ui`. I may be mistaken.

egui/src/grid.rs
skairunner: `paint_row` only paints a single row. The `paint` method iterates through all the row pickers and paints for every case that fits. If you prefer, I can break `paint` into `paint_rows` and `paint_cols`, but I personally feel that it's okay to call a single `paint` method that handles everything.

egui/src/grid.rs
skairunner: The problem is that if someone subsequently calls `header_row(false)`, it's not trivial to figure out how to remove the added row coloration from the `row_pickers`, and also kinda weird to be comparing closures/fn pointers/whatever. Would appreciate thoughts.

egui/src/grid.rs
skairunner: Fixed

egui/src/grid.rs
skairunner: Perhaps header_row() can be a method with no arguments that, when called, irreversibly sets the header coloration, and same for stripes?

skairunner:
Turns out that for any practical colorpicker, we really want to be able to capture variables in closures, so I changed ColorPickerFn to be a boxed Fn. I wonder if I could make a method that takes a &Fn and then clones and boxes it. Also, the new `style.visuals.faint_bg_color` doesn't work very well for composing because it's not an alpha color, so I removed that code while merging, unfortunately. Better ideas are welcome.

skairunner:
Seems this PR has fallen off the bus.

imgurbot12:
Hey all, any update on this? Individual row color controls would be amazingly helpful.

lucasmerlin:
Closing this since it was implemented in https://github.com/emilk/egui/pull/2519

PR #386, opened 2021-05-13
Fix BottomUp layout

### LINKED ISSUE #385: BottomUp layout draws widgets on top of each other
**Describe the bug**
I was just working on improved plot legends and noticed that when using `Layout::bottom_up()`, the widgets get stacked on top of each other instead of next to each other.
image

**Expected behavior**
image

I think I've found the issue. Just one missing line. I'll open a PR.

EmbersArc:
Closes #385. Works as expected now, at least for my use case.


emilk (approved):
Good catch - thanks!

Issue #391, opened 2021-05-15
Support horizontal scrolling in ScrollArea


parasyte:
`ScrollArea` currently only supports scrolling vertically. We need a way to allow scrolling horizontally for wide content.

**Describe the solution you'd like**

"Wide content" is a broad definition, so I'll propose that we define the term to mean, roughly, that the content width _needs_ to be wider than the `ScrollArea` width. This is usually because the content area contains UI elements that cannot be made thinner. A very obvious example is an `Image` which has a static width and height.

Draw a horizontal scroll bar at the bottom of the `ScrollArea`, and allow it to be scrolled with <kbd>Shift + Scroll Wheel</kbd>. Otherwise it should support the same interactions as the existing vertical scroll bar, including touch inputs.

emilk:
Having horizontal scrolling as an opt-in part of `ScrollArea` is the right approach. Users should have the choice between vertical, horizontal or both.

PR #392, opened 2021-05-15
Fix scroll bar position

parasyte:
While experimenting with #391, I noticed a bug in the way the vertical scroll bar is positioned. The screens below show the demo app resized to various widths.

A wide window looks nice, the scroll bar is firmly attached to the right side of the window:

vertical-scroll-bar-1

At some point, when the window is made thinner, the scroll bar begins to disappear behind the window border:

vertical-scroll-bar-2

When the window is very thin, you can see contents getting clipped as if the scroll bar should appear in the blank space:

vertical-scroll-bar-3

----

With this very simple patch, the scroll bar correctly sits in that blank space:

vertical-scroll-bar-4

And bonus, wide windows still allow the scroll bar to stick to the right edge (see first screen).

### DIFF
--- a/egui/src/containers/scroll_area.rs
+++ b/egui/src/containers/scroll_area.rs
@@ -133,9 +133,9 @@ impl ScrollArea {
             ),
             *ui.layout(),
         );
-        let mut content_clip_rect = inner_rect.expand(ui.visuals().clip_rect_margin);
-        content_clip_rect = content_clip_rect.intersect(ui.clip_rect());
-        content_clip_rect.max.x = ui.clip_rect().max.x - current_scroll_bar_width; // Nice handling of forced resizing beyond the possible
+        let content_clip_rect = inner_rect
+            .expand(ui.visuals().clip_rect_margin)
+            .intersect(ui.clip_rect());
         content_ui.set_clip_rect(content_clip_rect);
 
         Prepared {
--- a/egui/src/containers/scroll_area.rs
+++ b/egui/src/containers/scroll_area.rs
@@ -187,7 +187,7 @@ impl Prepared {
         }
 
         let width = if inner_rect.width().is_finite() {
-            inner_rect.width().max(content_size.x) // Expand width to fit content
+            inner_rect.width().min(content_size.x) // Position scroll bar correctly
         } else {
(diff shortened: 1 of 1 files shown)


parasyte:
The original patch positioned the scroll bar incorrectly when the content was smaller than the `ScrollArea` width.

parasyte:
I just discovered another bug in the `ScrollArea` with content clipping. This one is easiest to see with "paint clip rectangles (debug)" enabled.

<details><summary>Wall of text, lots of screen shots...</summary>

First, this is what it looks like with a window-with-scroll-bar that is about the same width as its parent:

scrollarea-clipping-rect-1

There is something else interesting to note: The width of the clipping rect on the side bar takes up all available width (except for space for a scroll bar)! This will be important later...

If I start resizing the setting window horizontally to the right, it will eventually become wider than the parent. The scroll bar will be hidden, but the clipping rect is pinned to the parent width:

scrollarea-clipping-rect-2

This improperly clips the contents and leaves a gap that is the same size as the scroll bar that is now going under the fold.

If I keep stretching the window width, you can see the clipping issue quite clearly:

scrollarea-clipping-rect-3

Now with the patch in c080ca94299c242bf3323fa5823681bde61cd620, we can see how different it is when the window gets _just_ too wide for the parent:

scrollarea-clipping-rect-4

The right side of the collapsible headers no longer gets clipped. You can make out the rounded corners.

Also! Notice that now the clipping rect for the sidebar is the expected size.

And if we keep stretching the window width, everything is fine:

scrollarea-clipping-rect-5
</details>

I did a little digging and it looks like the original clipping logic was introduced in 24ce7b41457c4e6d7b72102e4a207994fe828c44, and then fixed for the `Resize` container in 347fdb97d632b14c8e18b56583c5b55776e460c2 (but sadly not fixed for `ScrollArea`).

emilk (approved):
Thanks for the digging into this!

Seems the code not only became more correct, but also shorter! Awesome :)

emilk:
Oh no, this broke resizing of windows: windows can now be shrunk so narrow that they don't fit their contents. This is not good:

Screen Shot 2021-06-04 at 00 02 02

I'll revert

PR #399, opened 2021-05-16
Add support for buffers other than a `String` in `TextEdit`

Zenithsiz:
Adds support for custom buffers to be used with the `TextEdit` widget.

I haven't made an issue because this was relatively simple to implement and it's something I need for one of my projects, so I'd likely re-use it even if you'd prefer to not make this change.

The motivation behind this for me is to use an `AsciiString` from the `ascii` crate as the underlying buffer, and also limit it's size to a certain number of max characters.
In order to achieve this the `TextBuffer` trait hands responsibility of any actual insertions / removals to the type itself, with a default impl for `String`.
Due to the default type of `String` I believe this doesn't break any existing code and a default of `String` is likely the best option either way.
This also modifies most operations to be in-place instead of creating a new string and assigning it to it. This should (?) also increase performance slightly, but I doubt it's too noticeable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ NOTE: [`eframe`](eframe/CHANGELOG.md), [`egui_web`](egui_web/CHANGELOG.md) and [
 
 ### Added ⭐
 * Add features `extra_asserts` and `extra_debug_asserts` to enable additional checks.
+* `TextEdit` now supports edits on a generic buffer using `TextBuffer`.
 
 ## 0.12.0 - 2021-05-10 - Multitouch, user memory, window pivots, and improved plots
 
@@ -26,7 +27,6 @@ NOTE: [`eframe`](eframe/CHANGELOG.md), [`egui_web`](egui_web/CHANGELOG.md) and [
   enabling zoom, rotate, and more. Works with `egui_web` on mobile devices,
   and should work with `egui_glium` for certain touch devices/screens.
 * Add (optional) compatibility with [mint](https://docs.rs/mint).
-* `TextEdit` now supports edits on a generic buffer using `TextBuffer`.
 
 ### Changed 🔧
 * Make `Memory::has_focus` public (again).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ NOTE: [`eframe`](eframe/CHANGELOG.md), [`egui_web`](egui_web/CHANGELOG.md) and [
   enabling zoom, rotate, and more. Works with `egui_web` on mobile devices,
   and should work with `egui_glium` for certain touch devices/screens.
 * Add (optional) compatibility with [mint](https://docs.rs/mint).
+* `TextEdit` now supports edits on a generic buffer using `TextBuffer`.
 
 ### Changed 🔧
 * Make `Memory::has_focus` public (again).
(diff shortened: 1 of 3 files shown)


egui/src/widgets/text_edit.rs
```diff
@@ -108,6 +109,50 @@ impl CCursorPair {
     }
 }
 
+/// Trait contraining what types [`TextEdit`] may use as
+/// an underlying buffer
```
emilk, line 113:
```suggestion
/// Trait constraining what types [`TextEdit`] may use as
/// an underlying buffer.
///
/// Most likely you will use a `String` which implements `TextBuffer`.
```

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ NOTE: [`eframe`](eframe/CHANGELOG.md), [`egui_web`](egui_web/CHANGELOG.md) and [
   enabling zoom, rotate, and more. Works with `egui_web` on mobile devices,
   and should work with `egui_glium` for certain touch devices/screens.
 * Add (optional) compatibility with [mint](https://docs.rs/mint).
+* `TextEdit` now supports edits on a generic buffer using `TextBuffer`.
```
emilk, line 27: This should be under the `Unreleased` heading!

egui/src/widgets/text_edit.rs
```diff
@@ -478,18 +523,24 @@ impl<'t> TextEdit<'t> {
                 let did_mutate_text = match event {
                     Event::Copy => {
                         if cursorp.is_empty() {
-                            copy_if_not_password(ui, text.clone());
+                            copy_if_not_password(ui, text.as_ref().to_owned());
```
emilk, line 526:
```suggestion
                            copy_if_not_password(ui, text.to_string());
```

is a bit shorter (and iirc it should work since `text` implements `Display`)

egui/src/widgets/text_edit.rs
```diff
@@ -108,6 +109,50 @@ impl CCursorPair {
...
+    ///
+    /// # Return
+    /// Returns how many *characters* were successfully inserted
+    fn insert_text(&mut self, text: &str, ch_idx: usize) -> usize;
+
+    /// Deletes a range of text `ch_range` from this buffer.
+    ///
+    /// # Notes
+    /// `ch_range` is a *character range*, not a byte range.
+    fn delete_text_range(&mut self, ch_range: Range<usize>);
```
emilk, line 130:
```suggestion
    fn delete_char_range(&mut self, ch_range: Range<usize>);
```

is a more descriptive name

emilk (changes requested):
Looks good!

CHANGELOG.md
Zenithsiz: Oops, I somehow missed that heading. Should be fixed now.

egui/src/widgets/text_edit.rs
Zenithsiz: Wouldn't calling `to_owned` on a `str` bypass the `std::fmt` formatting, though?
I don't know how it works in general, but there might be a specialization for `String` and `str` so it might not matter.

Zenithsiz:
The branch conflict seems to be because of the changelog, I added the "Added" heading, should I rebase on master?

emilk:
yeah, rebase on or merge in master

Zenithsiz:
Should be rebased now? I had to do a `git push --force`, but not sure if that was what I was supposed to do after rebasing.

emilk (approved):

PR #400, opened 2021-05-16
Add custom margin support on all panels

Mubelotix:
It was not possible to set margin on side panels. 

## Example

```rust
egui::SidePanel::left("my_side_panel", 0.0).margin(vec2(5.0, 3.0)).show(ctx, |ui| {
    ui.label("Hello World!");
});
```

This PR is non-breaking. It keeps all default margin values and does not change signature of any existing public function.

### DIFF
--- a/egui/src/containers/frame.rs
+++ b/egui/src/containers/frame.rs
@@ -19,7 +19,8 @@ impl Frame {
         Self::default()
     }
 
-    /// For when you want to group a few widgets together within a frame.
+    /// For when you want to group a few widgets together within a frame.  
+    /// Default margin is `vec2(8.0, 6.0)`. Use [`group_with_margin`](Self::group_with_margin) to set the margin.
     pub fn group(style: &Style) -> Self {
         Self {
             margin: Vec2::new(8.0, 6.0),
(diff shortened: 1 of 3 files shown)


emilk:
Thaks for the PR!

Did you consider the alternative of having this as a setting in `Spacing` instead? Putting it in `Spacing` makes it convenient when a user wants to change the style of all panels or all groups.

So there are two use cases:
* a user want to change the default `Style` of `egui` to have smaller/larger panels margins everywhere
* a user wants to have different margins for different panels and different groups

The second use case is served by this PR, but I suspect that the first use case is more common. In particular, the second use will lead to an incoherent visual style of the app where different panels/groups have differing amount of margin

Mubelotix:
> Did you consider the alternative of having this as a setting in Spacing instead? Putting it in Spacing makes it convenient when a user wants to change the style of all panels or all groups.

This is a great idea! We could add a `panel_padding` field. Since the default values are currently different depending on the panel location, do you prefer to:
- Add a field with a single value, but some default values would change and that would be breaking,
- Add a field with a single optional value so that this is an opt-in feature,
- Add a field with a struct that has different fields like `side_panel_padding` and `central_panel_padding` so that we can keep the different default values

> The second use case is served by this PR, but I suspect that the first use case is more common. In particular, the second use will lead to an incoherent visual style of the app where different panels/groups have differing amount of margin

That's right, I focused on the second use case. I need a high level of flexibility since I made a custom side panel. As long as we keep visual style coherence by default, I don't think it is a problem to allow the user to break the defaults.

Issue #401, opened 2021-05-16
Zooming with ctrl + MouseWheel does not work as expected on a Mac


quadruple-output:
**Describe the bug**

Holding `CTRL` while turning the mouse wheel is translated to a Zoom event. This is correct on Windows, but on a Mac the correct modifier key would be `Command` rather than `CTRL`.

(On a Mac, the combination `CTRL`+Wheel is captured by the OS and triggers a full-screen zoom.)

**To Reproduce**
1. On a Mac: `cargo run -p egui_demo_app`
2. Open the "Plot" demo
3. try zooming with `Command`+MouseWheel (only moves but does not zoom) or `CTRL`+MouseWheel (triggers hardware full-screen zoom which is not desired)

emilk:
I picked ctrl-scroll because that's how I zoom in Firefox, but I notice cmd-scroll also works. Perhaps egui_glium should treat both as a zoom event?

quadruple-output:
Oh, ok… I am glad you tested this on your machine.

Keybindings on the Mac often surprise me. I think I found the setting that makes the difference for me: System Preferences → Accessibility → Zoom → [/] Use scroll gesture with modifier keys to zoom (can be configured as "Control", "Option", or "Command")

PR #402, opened 2021-05-16
Fix mac scroll modifier

quadruple-output:
Closes <https://github.com/emilk/egui/issues/401>.

### DIFF
--- a/egui_demo_lib/src/apps/demo/plot_demo.rs
+++ b/egui_demo_lib/src/apps/demo/plot_demo.rs
@@ -103,7 +103,7 @@ impl PlotDemo {
                 "Zoom with ctrl + mouse wheel (or ⌘ + wheel on a Mac), or with two-finger pinch.",
             );
         } else if cfg!(target_os = "macos") {
-            ui.label("Zoom with ⌘ + mouse wheel, or ⌘ + two-finger swipe.");
+            ui.label("Zoom with <ctrl or ⌘> + <mouse wheel or two-finger swipe>.");
         } else {
             ui.label("Zoom with ctrl + mouse wheel, or ctrl + two-finger swipe (your experience may vary).");
         }
--- a/egui_demo_lib/src/apps/demo/plot_demo.rs
+++ b/egui_demo_lib/src/apps/demo/plot_demo.rs
@@ -95,17 +95,13 @@ impl PlotDemo {
             });
         });
 
-        ui.label(
-            "Pan by dragging, or with mouse wheel (+ shift = horiz.), or with two-finger swipe.",
-        );
+        ui.label("Pan by dragging, or scroll (+ shift = horizontal).");
         if cfg!(target_arch = "wasm32") {
-            ui.label(
-                "Zoom with ctrl + mouse wheel (or ⌘ + wheel on a Mac), or with two-finger pinch.",
-            );
+            ui.label("Zoom with ctrl / ⌘ + mouse wheel, or with pinch gesture.");
         } else if cfg!(target_os = "macos") {
-            ui.label("Zoom with <ctrl or ⌘> + <mouse wheel or two-finger swipe>.");
+            ui.label("Zoom with ctrl / ⌘ + scroll.");
         } else {
-            ui.label("Zoom with ctrl + mouse wheel, or ctrl + two-finger swipe (your experience may vary).");
+            ui.label("Zoom with ctrl + scroll.");
         }
         ui.label("Reset view with double-click.");
     }
--- a/egui_demo_lib/src/apps/demo/plot_demo.rs
+++ b/egui_demo_lib/src/apps/demo/plot_demo.rs
@@ -95,7 +95,19 @@ impl PlotDemo {
             });
         });
 
-        ui.label("Drag to pan, ctrl + scroll to zoom. Double-click to reset view.");
+        ui.label(
+            "Pan by dragging, or with mouse wheel (+ shift = horiz.), or with two-finger swipe.",
+        );
+        if cfg!(target_arch = "wasm32") {
+            ui.label(
+                "Zoom with ctrl + mouse wheel (or ⌘ + wheel on a Mac), or with two-finger pinch.",
+            );
+        } else if cfg!(target_os = "macos") {
+            ui.label("Zoom with ⌘ + mouse wheel, or ⌘ + two-finger swipe.");
+        } else {
+            ui.label("Zoom with ctrl + mouse wheel, or ctrl + two-finger swipe (your experience may vary).");
+        }
+        ui.label("Reset view with double-click.");
     }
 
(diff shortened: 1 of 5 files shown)


egui_glium/src/lib.rs
```diff
@@ -188,7 +188,7 @@ pub fn input_to_egui(
                 delta.x *= -1.0;
             }
 
-            if input_state.raw.modifiers.ctrl {
+            if input_state.raw.modifiers.command {
```
follower, line 191: Am I correct in understanding we only want this change to apply on macOS?

If so, I'm guessing the modifier key to match presumably needs to be determined via e.g. `cfg!(target_os = "macos")`?

(Also, as a matter of interest is there a similar issue with the web backend?)

quadruple-output: By using `modifiers.command` instead of `modifiers.mac_cmd`, the fix should automatically do the right thing, depending on the OS. Sorry, I guess I should have mentioned this in the PR, but I thought this documentation of the constants would technically be sufficient:

```
pub struct Modifiers {
[…]
    /// Either of the control keys are down.
    /// When checking for keyboard shortcuts, consider using [`Self::command`] instead.
    pub ctrl: bool,
[…]
    /// The Mac ⌘ Command key. Should always be set to `false` on other platforms.
    pub mac_cmd: bool,
    /// On Windows and Linux, set this to the same value as `ctrl`.
    /// On Mac, this should be set whenever one of the ⌘ Command keys are down (same as `mac_cmd`).
    /// This is so that egui can, for instance, select all text by checking for `command + A`
    /// and it will work on both Mac and Windows.
    pub command: bool,
```

quadruple-output: Regarding the web-backend: yes, you are correct. I did not notice it first, since the web backend works for me with a normal pich/zoom gesture (no need for a modifier key). But I tested now by holding CRTL+CMD at the same time (which prevents the hardware zoom from happening), and this reveals that there _would_ we a zoom event if `Command` would be used instead of `CTRL`.

I will update the PR…

quadruple-output: … should be resolved by <https://github.com/emilk/egui/pull/402/commits/16fd0b9b44a893c10c9fcbd114a876dfe3c92bc6>

quadruple-output:
I uploaded a pre-compiled version for the web backend at <https://quadruple-output.github.io/egui/>, so you may compare the behavior with the main branch at <https://emilk.github.io/egui/>. There should not be any noticeable difference on Windows, Linux, or on a tablet. But on Mac, the `Command` key can now be used for zooming with the mouse wheel (or two-finger touch gesture).

egui_glium/src/lib.rs
follower: Ah, thanks for pointing out the docs for `modifiers.command` I wasn't aware that functionality existed. :)

Thanks for investigating & addressing the web backend issue too.

Didn't notice anything obviously broken on Linux/Firefox with the pre-compiled version you uploaded but haven't tested it on Mac myself.

(Only other thing that comes to mind is that we might want to update the UI text in the demo that says to use "Ctrl" to additionally/alternatively mention "Command" key--and could even use the fancy emoji included in egui: https://github.com/emilk/egui/blob/ebb08f87f1adb0b96c6c854bc3a0ed1b7e5cb392/egui_demo_lib/src/apps/demo/font_contents_emoji.rs#L22 :D )

Thanks for your contribution--I imagine Emil will take a look when he has a chance. :)

quadruple-output: Thanks for hinting at the UI text. <https://github.com/emilk/egui/pull/402/commits/16fd0b9b44a893c10c9fcbd114a876dfe3c92bc6> makes an attempt to address your proposal. Not sure if this is a bit _too much,_ though. Don't hesitate to criticize.

I tested on a Mac with native app, and with wasm in Firefox. It would be great if you (and anyone else) could check the correctness of the UI instructions of the Plot demo on other platforms.

emilk, line 191:
```suggestion
            if input_state.raw.modifiers.ctrl || input_state.raw.modifiers.command {
```

So we also keep the old ctrl-scroll behavior. See https://github.com/emilk/egui/issues/401

quadruple-output:
Hi emilk, I just accepted your proposed change, but don't close the PR, yet! I would like to adapt the UI demo instructions to reflect that, on Mac, _both_ CMD and CTRL can be used. However, I would like to know whether you consider the verbose UI texts I proposed to be "too much"? (In particular, I am not sure about the behavior an some Windows touch screens which I cannot test. Therefore, the text is both verbose and vage for native non-mac builds.)

egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -95,7 +95,19 @@ impl PlotDemo {
             });
         });
 
-        ui.label("Drag to pan, ctrl + scroll to zoom. Double-click to reset view.");
+        ui.label(
+            "Pan by dragging, or with mouse wheel (+ shift = horiz.), or with two-finger swipe.",
```
emilk, line 99: I didn't know about shift-scroll 😄

egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -95,7 +95,19 @@ impl PlotDemo {
...
+        );
+        if cfg!(target_arch = "wasm32") {
+            ui.label(
+                "Zoom with ctrl + mouse wheel (or ⌘ + wheel on a Mac), or with two-finger pinch.",
+            );
+        } else if cfg!(target_os = "macos") {
+            ui.label("Zoom with ⌘ + mouse wheel, or ⌘ + two-finger swipe.");
+        } else {
+            ui.label("Zoom with ctrl + mouse wheel, or ctrl + two-finger swipe (your experience may vary).");
```
emilk, line 108: I think just `Zoom with ctrl/⌘ + scroll, or with pinch gesture.` works. Two finger swiping on a touch screen or touch pad is often referred to as scrolling.

quadruple-output: Does the pinch gesture work for you when running natively on MacOS? For me, pinching exclusively works in the browser.
If you like my latest commit (which I committed in parallel to your latest remark), you are welcome to release the PR. Otherwise, I will be happy to get rid of the term "two-finger swipe".

emilk (approved):

egui_demo_lib/src/apps/demo/plot_demo.rs
quadruple-output: <https://github.com/emilk/egui/pull/402/commits/b185d64a19a88add3c2fbc029acd6cc239d4b284> removes the clumsy term "two-finger swipe" as suggested.

emilk (approved):

PR #405, opened 2021-05-17
Value of `DragValue` correctly clamped

### LINKED ISSUE #404: Clamping a `DragValue` clears invalid value visually but not internally
**Describe the bug**
When a `DragValue`'s clamp range is changed, if its previous value ends outside of the new limits, the value that is shown is updated to be on the correct range. However, the internal value remains unchanged, which leads to a discrepancy between the shown and the internal value.

**To Reproduce**
1. Make two `DragValue`s and have the first control the clamping range of the second. 
2. Set the maximum value for the second `DragValue`.
3. Use the first `DragValue` to clamp the second one into a range that makes the previous value invalid.
4. The value shown on the second `DragValue` will be within the desired range, but if we print out its value, we'll see that the internal value isn't.

**Expected behavior**
The internal and shown value should always coincide.

**Additional context**
My guess is that the shown value has a check in place to clamp the value, but the internal value doesn't. So perhaps this check should simply be moved within the code?

OfficialURL:
Closes #404.

### DIFF
--- a/egui/src/widgets/drag_value.rs
+++ b/egui/src/widgets/drag_value.rs
@@ -220,8 +220,11 @@ impl<'a> Widget for DragValue<'a> {
         let is_slow_speed =
             ui.input().modifiers.shift_only() && ui.memory().is_being_dragged(ui.next_auto_id());
 
-        let value = get(&mut get_set_value);
-        let value = clamp_to_range(value, clamp_range.clone());
+        let old_value = get(&mut get_set_value);
+        let value = clamp_to_range(old_value, clamp_range.clone());
+        if old_value != value {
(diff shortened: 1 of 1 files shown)


emilk:
Good catch! There is one additional line that needs fixing. At the end of the function there is this:

`response.changed = get(&mut get_set_value) != value;`

this should probably be changed to

`response.changed = get(&mut get_set_value) != old_value;`

so that we also set the `changed` flag when the value changes due to the clamping.

OfficialURL:
I've implemented this fix in the latest commit.

emilk (approved):

PR #410, opened 2021-05-17
Plot: Legend improvements

### LINKED ISSUE #409: Plot: Make legend more configurable
The plot legend is nice, but I would like to make some improvements:

* Draw it in front of the plot contents instead of behind
* Set its position
* Change the font
* Add an optional background

I have a PR ready for those changes, except for the background. I haven't found a solution for that yet.

EmbersArc:
Closes #409.

https://user-images.githubusercontent.com/1352472/118518207-2e6ed880-b738-11eb-8335-8c4526c5fc20.mp4

This is based on the work in #363, so we should wait for that to be merged first. Sorry about the large PRs.


egui/src/widgets/plot/legend.rs
```diff
@@ -1,81 +1,231 @@
...
-pub(crate) struct LegendEntry {
-    pub text: String,
-    pub color: Color32,
-    pub checked: bool,
-    pub hovered: bool,
+use super::items::PlotItem;
+
+/// Where to place the plot legend.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum LegendPosition {
```
emilk, line 12:
```suggestion
pub enum Corner {
```

maybe?

Alternatively could use `emath::Align2` here: https://docs.rs/emath/0.12.0/emath/align/struct.Align2.html

egui/src/widgets/plot/legend.rs
```diff
@@ -1,81 +1,231 @@
...
+    TopLeft,
+    TopRight,
+    BottomLeft,
+    BottomRight,
```
emilk, line 16:
```suggestion
    LeftTop,
    RightTop,
    LeftBottom,
    RightBottom,
```

egui uses consistent X,Y order for everything. It reads a bit backwards, but one gets used to it :)

See for instance https://docs.rs/emath/0.12.0/emath/struct.Rect.html#method.left_top

egui/src/widgets/plot/legend.rs
```diff
@@ -1,81 +1,231 @@
...
-            text,
             color,
             checked,
             hovered: false,
         }
     }
 }
 
-impl Widget for &mut LegendEntry {
+impl Widget for (&String, &mut LegendEntry) {
```
emilk, line 65: This is a bit weird. Maybe just make a function for it:

``` rust
fn legend_ui(ui: &mut Ui, text: &str, legend: &mut LegendEntry) {
…
}
```

?

egui/src/widgets/plot/legend.rs
```diff
@@ -1,81 +1,231 @@
...
+        let total_extra = vec2(icon_size + icon_spacing, 0.0);
+
+        let desired_size = total_extra + galley.size;
         let (rect, response) = ui.allocate_exact_size(desired_size, Sense::click());
-        let rect = rect.shrink2(padding);
 
         response.widget_info(|| WidgetInfo::selected(WidgetType::Checkbox, *checked, &galley.text));
 
         let visuals = ui.style().interact(&response);
+        let flipped = ui.layout().cross_align() == Align::RIGHT;
```
emilk, line 90:
```suggestion
        let flipped = ui.layout().horizontal_align() == Align::RIGHT;
```

emilk, line 90: flipped vs what? Better to call it something like `label_on_the_left` (or is it `label_on_the_right`?)

egui/src/widgets/plot/legend.rs
```diff
@@ -1,81 +1,231 @@
...
+        items
+            .iter()
+            .filter(|item| !item.name().is_empty())
+            .for_each(|item| {
+                entries
+                    .entry(item.name().to_string())
+                    .and_modify(|entry| {
+                        if entry.color != item.color() {
+                            entry.color = Color32::TRANSPARENT
+                        }
```
emilk, line 167:
```suggestion
                        if entry.color != item.color() {
                            // Multiple items with different colors
                            entry.color = Color32::TRANSPARENT;
                        }
```

egui/src/widgets/plot/items.rs
```diff
@@ -273,7 +274,8 @@ impl Line {
 
     /// Name of this line.
     ///
-    /// This name will show up in the plot legend, if legends are turned on.
+    /// This name will show up in the plot legend, if legends are turned on. Multiple lines may
+    /// share the same name.
```
emilk, line 278: But if two items share the same name, they also share the same legend entry, correct?

If so we should document that here and below

egui/src/widgets/plot/legend.rs
```diff
@@ -1,81 +1,231 @@
...
+}
+
+impl LegendWidget {
+    /// Create a new legend from items, the names of items that are hidden and the style of the
+    /// text. Returns `None` if the legend has no entries.
+    pub(super) fn try_new(
+        rect: Rect,
+        config: Legend,
+        items: &[Box<dyn PlotItem>],
+        hidden_items: &HashSet<String>,
```
emilk, line 153: it feels a bit backwards to have a list of things to hide rather than a list of things to show

emilk (changes requested):
Nice additions!

egui/src/widgets/plot/legend.rs
EmbersArc: Good idea to call it `Corner`. I couldn't use `Align2` unfortunately since top, left, right and bottom center are not supported.

egui/src/widgets/plot/legend.rs
EmbersArc: Made it a method of `LegendEntry.

egui/src/widgets/plot/legend.rs
EmbersArc: :+1: Changed

egui/src/widgets/plot/items.rs
EmbersArc: Added the comment.

egui/src/widgets/plot/legend.rs
EmbersArc: Yes, but it seems to be the only way. I can't pass things that are shown between frames. Because if a new curve is added, it will not be shown because it's not already in the "shown curves".

egui/src/widgets/plot/legend.rs
EmbersArc: Seems to be private unfortunately...

EmbersArc:
Thanks for the review! I made the changes.

I also tried to give the legend a background, but couldn't find a way to do it without doing all the size calculation myself. Do you happen to know a way to do this?

egui/src/widgets/plot/legend.rs
emilk: Oops! fixed in 3c603c5

emilk:
> Thanks for the review! I made the changes.
> 
> I also tried to give the legend a background, but couldn't find a way to do it without doing all the size calculation myself. Do you happen to know a way to do this?

Sure, just add it to a [`Frame`](https://docs.rs/egui/0.12.0/egui/containers/struct.Frame.html):

``` rust
let frame = egui::Frame { … };
frame.show(ui, |ui| { /* legend ui */ });
```

EmbersArc:
> Sure, just add it to a Frame

I tried that and it looks great when the legend is on the top left:

Screenshot_20210607_213424

But it doesn't work for all the other cases:

Screenshot_20210607_213444
Screenshot_20210607_213500

Is there a way to work around this issue? Maybe something like choosing an anchor point for the frame?

emilk:
> > Sure, just add it to a Frame
> 
> I tried that and it looks great when the legend is on the top left:
> 
> …
> 
> Is there a way to work around this issue? Maybe something like choosing an anchor point for the frame?

I think this is a bug in `Frame`. Does this fix it:

```
diff --git a/egui/src/containers/frame.rs b/egui/src/containers/frame.rs
index 36b272c..6b41e31 100644
--- a/egui/src/containers/frame.rs
+++ b/egui/src/containers/frame.rs
@@ -175,10 +175,7 @@ impl Frame {
 
 impl Prepared {
     pub fn outer_rect(&self) -> Rect {
-        Rect::from_min_max(
-            self.outer_rect_bounds.min,
-            self.content_ui.min_rect().max + self.frame.margin,
-        )
+        self.content_ui.min_rect().expand2(self.frame.margin)
     }
 
     pub fn end(self, ui: &mut Ui) -> Response {
```

?

(that fix may introduce other problems, so please don't commit it)

emilk:
I think we should get this in even without the legend background

EmbersArc:
That was it! Works in all corners now. I'll add that in the next PR then. Thanks!

image

PR #412, opened 2021-05-18
Additional accessibility support

ndarilek:
This is an initial stab at broadening egui's accessibility support. It isn't at all complete, but I'm hoping to merge what I can as I go, and to get some feedback. I'd particularly like to merge regularly so branches don't get out of sync, and particularly because I think something is better than nothing where accessibility is concerned. Currently supported:

* Focus changes
* Value updates when clicking on things (I.e. checkboxes)
* Reporting of changed text in `TextEdit`, as well as reporting when the selection changes. Still not perfect, but it's something.

[bevy_egui_a11y](https://github.com/ndarilek/bevy_egui_a11y) has been updated to show off these changes. Note that you can tab through all fields, get meaningful feedback, type into the text field, check the checkbox, arrow around, etc.

I've already started using my fork in my game and it seems to work well, but I really don't want to maintain a fork going forward. Please let me know if you'd rather not go down this road, and I'll stop work and back out the integration in my game.

Thanks.

### DIFF
--- a/egui/src/data/output.rs
+++ b/egui/src/data/output.rs
@@ -242,7 +242,7 @@ pub struct WidgetInfo {
     /// The text on labels, buttons, checkboxes etc.
     pub label: Option<String>,
     /// The contents of some editable text (for `TextEdit` fields).
-    pub text_value: Option<String>,
+    pub current_text_value: Option<String>,
     // The previous text value.
     pub prev_text_value: Option<String>,
     /// The current value of checkboxes and radio buttons.
@@ -261,7 +261,7 @@ impl std::fmt::Debug for WidgetInfo {
             typ,
             enabled,
             label,
-            text_value,
+            current_text_value: text_value,
             prev_text_value,
             selected,
             value,
@@ -306,7 +306,7 @@ impl WidgetInfo {
             typ,
             enabled: true,
             label: None,
-            text_value: None,
+            current_text_value: None,
             prev_text_value: None,
             selected: None,
             value: None,
@@ -353,7 +353,7 @@ impl WidgetInfo {
     #[allow(clippy::needless_pass_by_value)]
     pub fn text_edit(text_value: impl ToString, prev_text_value: impl ToString) -> Self {
         Self {
-            text_value: Some(text_value.to_string()),
+            current_text_value: Some(text_value.to_string()),
             prev_text_value: Some(prev_text_value.to_string()),
             ..Self::new(WidgetType::TextEdit)
         }
@@ -368,7 +368,7 @@ impl WidgetInfo {
         Self {
             primary_cursor: Some(primary_cursor),
             secondary_cursor: Some(secondary_cursor),
-            text_value: Some(text_value.to_string()),
+            current_text_value: Some(text_value.to_string()),
             ..Self::new(WidgetType::TextEdit)
         }
     }
@@ -379,7 +379,7 @@ impl WidgetInfo {
             typ,
             enabled,
             label,
-            text_value,
+            current_text_value: text_value,
             prev_text_value: _,
             selected,
             value,
--- a/egui/src/data/output.rs
+++ b/egui/src/data/output.rs
@@ -40,6 +40,12 @@ impl Output {
         // only describe last event:
         if let Some(event) = self.events.iter().rev().next() {
             match event {
+                OutputEvent::Clicked(widget_info) => {
+                    return widget_info.description();
+                }
+                OutputEvent::DoubleClicked(widget_info) => {
+                    return widget_info.description();
+                }
                 OutputEvent::FocusGained(widget_info) => {
                     return widget_info.description();
                 }
@@ -204,6 +210,10 @@ impl Default for CursorIcon {
 /// In particular, these events may be useful for accessability, i.e. for screen readers.
 #[derive(Clone, PartialEq)]
 pub enum OutputEvent {
+    // A widget was clicked.
+    Clicked(WidgetInfo),
+    // A widget was double-clicked.
+    DoubleClicked(WidgetInfo),
     /// A widget gained keyboard focus (by tab key).
     FocusGained(WidgetInfo),
 }
@@ -211,6 +221,8 @@ pub enum OutputEvent {
 impl std::fmt::Debug for OutputEvent {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
+            Self::Clicked(wi) => write!(f, "Clicked({:?})", wi),
+            Self::DoubleClicked(wi) => write!(f, "DoubleClicked({:?})", wi),
             Self::FocusGained(wi) => write!(f, "FocusGained({:?})", wi),
         }
     }
--- a/egui/src/data/output.rs
+++ b/egui/src/data/output.rs
@@ -40,13 +40,10 @@ impl Output {
         // only describe last event:
         if let Some(event) = self.events.iter().rev().next() {
             match event {
-                OutputEvent::Clicked(widget_info) => {
-                    return widget_info.description();
-                }
-                OutputEvent::DoubleClicked(widget_info) => {
-                    return widget_info.description();
-                }
-                OutputEvent::FocusGained(widget_info) => {
+                OutputEvent::Clicked(widget_info)
+                | OutputEvent::DoubleClicked(widget_info)
+                | OutputEvent::FocusGained(widget_info)
+                | OutputEvent::ValueChanged(widget_info) => {
                     return widget_info.description();
                 }
             }
@@ -216,6 +213,8 @@ pub enum OutputEvent {
     DoubleClicked(WidgetInfo),
     /// A widget gained keyboard focus (by tab key).
     FocusGained(WidgetInfo),
+    // A widget's value changed.
+    ValueChanged(WidgetInfo),
 }
 
 impl std::fmt::Debug for OutputEvent {
@@ -224,6 +223,7 @@ impl std::fmt::Debug for OutputEvent {
             Self::Clicked(wi) => write!(f, "Clicked({:?})", wi),
             Self::DoubleClicked(wi) => write!(f, "DoubleClicked({:?})", wi),
             Self::FocusGained(wi) => write!(f, "FocusGained({:?})", wi),
+            Self::ValueChanged(wi) => write!(f, "ValueChanged({:?})", wi),
         }
     }
 }
@@ -236,7 +236,9 @@ pub struct WidgetInfo {
     /// The text on labels, buttons, checkboxes etc.
     pub label: Option<String>,
     /// The contents of some editable text (for `TextEdit` fields).
-    pub edit_text: Option<String>,
+    pub text_value: Option<String>,
+    // The previous text value.
+    prev_text_value: Option<String>,
     /// The current value of checkboxes and radio buttons.
     pub selected: Option<bool>,
     /// The current value of sliders etc.
@@ -248,7 +250,8 @@ impl std::fmt::Debug for WidgetInfo {
         let Self {
             typ,
             label,
-            edit_text,
+            text_value,
+            prev_text_value,
             selected,
             value,
         } = self;
@@ -260,8 +263,11 @@ impl std::fmt::Debug for WidgetInfo {
         if let Some(label) = label {
             s.field("label", label);
         }
-        if let Some(edit_text) = edit_text {
-            s.field("edit_text", edit_text);
+        if let Some(text_value) = text_value {
+            s.field("text_value", text_value);
+        }
+        if let Some(prev_text_value) = prev_text_value {
+            s.field("prev_text_value", prev_text_value);
         }
         if let Some(selected) = selected {
             s.field("selected", selected);
@@ -279,7 +285,8 @@ impl WidgetInfo {
         Self {
             typ,
             label: None,
-            edit_text: None,
+            text_value: None,
+            prev_text_value: None,
             selected: None,
             value: None,
         }
@@ -321,9 +328,10 @@ impl WidgetInfo {
     }
 
     #[allow(clippy::needless_pass_by_value)]
-    pub fn text_edit(edit_text: impl ToString) -> Self {
+    pub fn text_edit(text_value: impl ToString, prev_text_value: impl ToString) -> Self {
         Self {
-            edit_text: Some(edit_text.to_string()),
+            text_value: Some(text_value.to_string()),
+            prev_text_value: Some(prev_text_value.to_string()),
             ..Self::new(WidgetType::TextEdit)
         }
     }
@@ -333,7 +341,8 @@ impl WidgetInfo {
         let Self {
             typ,
             label,
-            edit_text,
+            text_value,
+            prev_text_value: _,
             selected,
             value,
         } = self;
@@ -370,9 +379,18 @@ impl WidgetInfo {
             description = format!("{}: {}", label, description);
         }
 
-        if let Some(edit_text) = edit_text {
-            description += " ";
-            description += edit_text;
+        if typ == &WidgetType::TextEdit {
+            let text;
+            if let Some(text_value) = text_value {
+                if text_value.is_empty() {
+                    text = "blank".into();
+                } else {
+                    text = text_value.to_string();
+                }
+            } else {
+                text = "blank".into();
+            }
+            description = format!("{}: {}", text, description);
         }
 
         if let Some(value) = value {
--- a/egui/src/data/output.rs
+++ b/egui/src/data/output.rs
@@ -43,6 +43,7 @@ impl Output {
                 OutputEvent::Clicked(widget_info)
                 | OutputEvent::DoubleClicked(widget_info)
                 | OutputEvent::FocusGained(widget_info)
+                | OutputEvent::TextSelectionChanged(widget_info)
                 | OutputEvent::ValueChanged(widget_info) => {
                     return widget_info.description();
                 }
@@ -213,6 +214,8 @@ pub enum OutputEvent {
     DoubleClicked(WidgetInfo),
     /// A widget gained keyboard focus (by tab key).
     FocusGained(WidgetInfo),
+    // Text selection was updated.
+    TextSelectionChanged(WidgetInfo),
     // A widget's value changed.
     ValueChanged(WidgetInfo),
 }
@@ -223,6 +226,7 @@ impl std::fmt::Debug for OutputEvent {
             Self::Clicked(wi) => write!(f, "Clicked({:?})", wi),
             Self::DoubleClicked(wi) => write!(f, "DoubleClicked({:?})", wi),
             Self::FocusGained(wi) => write!(f, "FocusGained({:?})", wi),
+            Self::TextSelectionChanged(wi) => write!(f, "TextSelectionChanged({:?})", wi),
             Self::ValueChanged(wi) => write!(f, "ValueChanged({:?})", wi),
         }
     }
@@ -238,11 +242,15 @@ pub struct WidgetInfo {
     /// The contents of some editable text (for `TextEdit` fields).
     pub text_value: Option<String>,
     // The previous text value.
-    prev_text_value: Option<String>,
+    pub prev_text_value: Option<String>,
     /// The current value of checkboxes and radio buttons.
     pub selected: Option<bool>,
     /// The current value of sliders etc.
     pub value: Option<f64>,
+    // Location of primary cursor.
+    pub primary_cursor: Option<usize>,
+    // Location of secondary cursor.
+    pub secondary_cursor: Option<usize>,
 }
 
 impl std::fmt::Debug for WidgetInfo {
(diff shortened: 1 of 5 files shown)


egui/src/widgets/text_edit.rs
```diff
@@ -138,6 +138,7 @@ impl CCursorPair {
 #[derive(Debug)]
 pub struct TextEdit<'t> {
     text: &'t mut String,
+    prev_text: String,
```
emilk, line 141: I don't think this should be a member. `TextEdit` is just a builder pattern - you can declare `prev_text` in the `content_ui` function.

egui/src/widgets/text_edit.rs
```diff
@@ -331,6 +334,7 @@ impl<'t> Widget for TextEdit<'t> {
 impl<'t> TextEdit<'t> {
     fn content_ui(self, ui: &mut Ui) -> Response {
         let TextEdit {
+            mut prev_text,
```
emilk, line 337: If we declare `prev_text` here as `let mut prev_text: Option<String> = None` then (`None`)  will mean "no change". We can then later check

``` rust
if let Some(prev_text) = prev_text {
    let info = WidgetInfo::text_changed(prev_text, text);
    self.ctx.output().events.push(OutputEvent::TextChanged(info));
}
````

File: egui/src/widgets/text_edit.rs [hunk 3]
```diff
@@ -741,7 +761,13 @@ fn byte_index_from_char_index(s: &str, char_index: usize) -> usize {
...
 }
 
-fn insert_text(ccursor: &mut CCursor, text: &mut String, text_to_insert: &str) {
+fn insert_text(
+    ccursor: &mut CCursor,
+    text: &mut String,
+    prev_text: &mut String,
+    text_to_insert: &str,
+) {
+    *prev_text = text.clone();
```
emilk, line 770:
```suggestion
    prev_text: &mut Option<String>,
    text_to_insert: &str,
) {
    *prev_text = prev_text.or_else(|| text.clone());
```

This will make it keep the oldest value if there are multiple changes during the frame. The oldest + latest value is what we we want to emit each frame there is a change.

emilk: This means changing `prev_text: &mut String` to `prev_text: &mut Option<String>` in all the arguments.

egui/src/widgets/text_edit.rs
```diff
@@ -656,7 +665,18 @@ impl<'t> TextEdit<'t> {
...
+        } else if let Some(text_cursor) = text_cursor {
+            response.has_widget_info = true;
+            response.widget_info(|| {
+                WidgetInfo::text_selection_changed(
+                    text_cursor.primary.ccursor.index,
+                    text_cursor.secondary.ccursor.index,
+                    &*text,
+                )
+            });
+        }
```
emilk, line 679: this can be simplified to something like this:

``` rust
if let Some(prev_text) = prev_text {
    let info = WidgetInfo::text_changed(prev_text, text);
    self.ctx.output().events.push(OutputEvent::TextChanged(info));
} else if text_cursor_changed {
    let char_range = text_cursor.primary.ccursor.index..text_cursor.secondary.ccursor.index;
    let info = WidgetInfo::text_cursor_changed(char_range, text);
    self.ctx.output().events.push(OutputEvent::TextCursorChanged(info));
}

// Response::widget_info will emit events if we are clicked, doublt-clicked etc:
response.widget_info(|| WidgetInfo::text_edit(&*text));
```

egui/src/response.rs
```diff
@@ -65,6 +65,8 @@ pub struct Response {
     /// e.g. the slider was dragged, text was entered in a `TextEdit` etc.
     /// Always `false` for something like a `Button`.
     pub(crate) changed: bool,
+    /// Has a `WidgetInfo` but isn't covered by some other case (E.g. `changed`, `clicked`.)
+    pub(crate) has_widget_info: bool,
```
emilk, line 69: as noted elsewhere, I don't think this is necessary

egui/src/response.rs
```diff
@@ -428,10 +432,26 @@ impl Response {
...
+            Some(OutputEvent::FocusGained(make_info()))
+        } else if self.changed {
+            Some(OutputEvent::ValueChanged(make_info()))
+        } else if self.has_widget_info {
+            let info = make_info();
+            if info.primary_cursor.is_some() && info.secondary_cursor.is_some() {
+                Some(OutputEvent::TextSelectionChanged(info))
+            } else {
+                None
+            }
```
emilk, line 450: Remove this and emit `TextSelectionChanged` directly in `TextEdit` instead

egui/src/response.rs
```diff
@@ -428,10 +432,26 @@ impl Response {
...
-            let event = OutputEvent::WidgetEvent(WidgetEvent::Focus, widget_info);
+        use crate::output::OutputEvent;
+        let event = if self.clicked() {
+            Some(OutputEvent::Clicked(make_info()))
+        } else if self.double_clicked() {
+            Some(OutputEvent::DoubleClicked(make_info()))
+        } else if self.gained_focus() {
+            Some(OutputEvent::FocusGained(make_info()))
+        } else if self.changed {
+            Some(OutputEvent::ValueChanged(make_info()))
```
emilk, line 443: this is great :+1:

egui/src/data/output.rs
```diff
@@ -317,9 +351,24 @@ impl WidgetInfo {
...
+            text_value: Some(text_value.to_string()),
+            prev_text_value: Some(prev_text_value.to_string()),
+            ..Self::new(WidgetType::TextEdit)
+        }
+    }
+
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn text_selection_changed(
+        primary_cursor: usize,
+        secondary_cursor: usize,
```
emilk, line 365:
```suggestion
        char_range: std::Ops::Range<usize>,
```

Data that belong together should be clumped together :)

egui/src/data/output.rs
```diff
@@ -317,9 +351,24 @@ impl WidgetInfo {
     }
 
     #[allow(clippy::needless_pass_by_value)]
-    pub fn text_edit(edit_text: impl ToString) -> Self {
+    pub fn text_edit(text_value: impl ToString, prev_text_value: impl ToString) -> Self {
+        Self {
+            text_value: Some(text_value.to_string()),
+            prev_text_value: Some(prev_text_value.to_string()),
```
emilk, line 357: "after, before" is a surprising order

```suggestion
    pub fn text_edit(prev_text_value: impl ToString, current_text_value: impl ToString) -> Self {
        Self {
            prev_text_value: Some(prev_text_value.to_string()),
            current_text_value: Some(current_text_value.to_string()),
```

egui/src/data/output.rs
```diff
@@ -204,67 +208,93 @@ impl Default for CursorIcon {
...
+    // The previous text value.
+    pub prev_text_value: Option<String>,
     /// The current value of checkboxes and radio buttons.
     pub selected: Option<bool>,
     /// The current value of sliders etc.
     pub value: Option<f64>,
+    // Location of primary cursor.
+    pub primary_cursor: Option<usize>,
+    // Location of secondary cursor.
+    pub secondary_cursor: Option<usize>,
```
emilk, line 255:
```suggestion
    // Selected range of characters in [`Self::current_text_value`].
    pub char_cursor: Option<std::ops::Range<usize>>,
```

egui/src/widgets/text_edit.rs
```diff
@@ -656,7 +665,18 @@ impl<'t> TextEdit<'t> {
 
         ui.memory().id_data.insert(id, state);
 
-        response.widget_info(|| WidgetInfo::text_edit(&*text));
+        if response.changed {
+            response.widget_info(|| WidgetInfo::text_edit(&*text, &*prev_text));
+        } else if let Some(text_cursor) = text_cursor {
+            response.has_widget_info = true;
+            response.widget_info(|| {
+                WidgetInfo::text_selection_changed(
```
emilk, line 673: This emits text selection change every frame, no? That's a bit excessive!

We should be able to detect when the cursor changes and only emit it then

emilk (commented):
Thank you for this PR, and I think this is great!

I saw some opportunities for improvements and simplifications:
TextEdit can handle emitting its own special events (`TextChanged` and `TextCursorChanged`).
It should also be possible to easily (and cheaply) only emit events when a change has occurred.

There will also be some unfortunate merge conflicts when I merge https://github.com/emilk/egui/pull/399

egui/src/data/output.rs
```diff
@@ -204,67 +208,93 @@ impl Default for CursorIcon {
...
 pub struct WidgetInfo {
     /// The type of widget this is.
     pub typ: WidgetType,
+    // Whether the widget is enabled.
+    pub enabled: bool,
     /// The text on labels, buttons, checkboxes etc.
     pub label: Option<String>,
     /// The contents of some editable text (for `TextEdit` fields).
-    pub edit_text: Option<String>,
+    pub text_value: Option<String>,
```
emilk, line 245:
```suggestion
    pub current_text_value: Option<String>,
```

maybe to distinguish it from `prev_text_value`? 🤷

egui/src/widgets/text_edit.rs
ndarilek: I just made it a clone of `text` so we could do a straight equality check, and also pass it directly to the various text manipulation functions. Does that work for you too? Otherwise agreed that it should be removed from the widget--that's done.

egui/src/widgets/text_edit.rs
```diff
@@ -513,12 +521,12 @@ impl<'t> TextEdit<'t> {
...
                     } => {
                         if multiline && ui.memory().has_lock_focus(id) {
-                            let mut ccursor = delete_selected(text, &cursorp);
+                            let mut ccursor = delete_selected(text, &mut prev_text, &cursorp);
                             if modifiers.shift {
                                 // TODO: support removing indentation over a selection?
                                 decrease_identation(&mut ccursor, text);
                             } else {
-                                insert_text(&mut ccursor, text, "\t");
+                                insert_text(&mut ccursor, text, &mut prev_text, "\t");
```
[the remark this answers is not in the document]
ndarilek: OK, I'll address those when it lands and I'm not basing a PR on a PR. :) Thanks for the heads-up.

egui/src/widgets/text_edit.rs
ndarilek: You're right, it does. Have to admit to being a bit overwhelmed with all of this. :) How do I get the previous cursor location? Right now I'm stashing that state locally in the screen reader and checking it for changes, but obviously if I can kick that can back down the road to egui then I'm happy to do it.

egui/src/widgets/text_edit.rs
ndarilek: I feel like we should consolidate accessibility event dispatch into as few places as possible, and we should locate the dispatches in the places that make the most sense.

E.g. `Response::widget_info` dispatches a click event when `self.clicked()` is true. That makes a lot of sense and seems maintainable. It currently dispatches `ValueChanged` when `self.changed` is true. Also makes a lot of sense to me. I'm hesitant to have a bunch of different places pushing change events, different places pushing click events, etc. if we can at all avoid it. The fewer choke points for accessibility event dispatch, the better IMO.

That said, something like tracking widget-internal state like the cursor position can absolutely push events directly, so I'll nix `has_widget_info`.

Thoughts?

egui/src/widgets/text_edit.rs
ndarilek: I see `output.text_cursor`. Is that only the cursor position? I'd need start and end values to determine if the value changed.

egui/src/widgets/text_edit.rs
ndarilek: Again, not sure how I feel about this one. If text is never optional in a `TextEdit` then wrapping it in an `Option` adds extra layers of indirection. If an `EditText` contains "foo" and isn't changed in a frame, then `prev_text` isn't `None`, it's "foo".

egui/src/data/output.rs
ndarilek: Done, though I used `RangeInclusive`. When I took a subset of text to speak, I needed an inclusive range anyway. May not work for actually selecting/copying text, in which case we can revisit.

ndarilek:
Hitting an issue I'd appreciate help with. The changes in 87bd38d431aa942ed32ba7c89c723b84ea2f15e5 break focus events on text fields and text fields only. That is, if I tab onto a text field, nothing is spoken and nothing is reported. Every other widget seems to work fine, though.

That particular commit pushes events in `content_ui`, but none are focus-related. I hand-applied a revert on top of master, and with those changes reverted, focus works again. Somehow, pushing `TextSelectionChanged` events blocks focus.

I'm starting to get a bit overwhelmed. When you have time for egui again, any chance you might spare some cycles to help track down why pushing unrelated events breaks focus events? [bevy_egui_a11y](https://github.com/ndarilek/bevy_egui_a11y) is updated and exhibits the behavior. I've committed the hand-revert on the branch [fix-focus-somehow](https://github.com/ndarilek/egui/tree/fix-focus-somehow).

Thanks.

egui/src/widgets/text_edit.rs
emilk: The goal is that `prev_text` can be `None` when there is no change. This is both a useful signal, and also there is no need to clone the text if there is no change.

In the current code `prev_text == ""` can mean "no change" or it can mean "there was a change, and the text field started empty".

With `Option<String>` `None` means "no change", `Some("")` means "there was a change, and the text field started empty"

egui/src/widgets/text_edit.rs
emilk: The simplest is probably to store the text cursor at the start of `TextEdit::content_ui` and then compare against that here.

egui/src/widgets/text_edit.rs
emilk: I'd rather avoid cloning the text each frame, but if the alternative is very complicated it's fine (it can be optimized later)

egui/src/widgets/text_edit.rs
```diff
@@ -656,7 +660,18 @@ impl<'t> TextEdit<'t> {
...
+        } else if let Some(text_cursor) = text_cursor {
+            let char_range =
+                text_cursor.primary.ccursor.index..=text_cursor.secondary.ccursor.index;
+            let info = WidgetInfo::text_selection_changed(char_range, &*text);
+            response
+                .ctx
+                .output()
+                .events
+                .push(OutputEvent::TextSelectionChanged(info));
+        }
```
emilk, line 674: Why the focus event aren't being sent anymore:

It is the call to `response.widget_info(…)` that sends those events (if applicable) and you now only call it when the text has changed.

So it will work again with something like:

``` rust
let text_cursor_changed = old_text_cursor != text_cursor;

if text_was_edited {
    let info = WidgetInfo::text_changed(prev_text, text);
    self.ctx.output().events.push(OutputEvent::TextChanged(info));
} else if text_cursor_changed && text_cursor.is_some() {
    let text_cursor = text_cursor.unwrap();
    let char_range = text_cursor.primary.ccursor.index..text_cursor.secondary.ccursor.index;
    let info = WidgetInfo::text_cursor_changed(char_range, text);
    self.ctx.output().events.push(OutputEvent::TextCursorChanged(info));
}  else {
    // This will emit events on focus, click, etc:
    response.widget_info(|| WidgetInfo::text_edit(&*text));
}
```

the reasoning for putting it in the else is that we don't want to send two events in the same frame for the same widget.

egui/src/widgets/text_edit.rs
emilk: I agree that dispatching all events from one function has a charm to it, and that was my original intent. What break the charm is that we are constructing different content for the `WidgetInfo`:s for different events.

We could have `TextEdit` call `widget_info` just as on `main` but with a `WidgetInfo` that now contains the latest cursor and the before/after text pair (every frame, even if nothing changed). `Response` can then be augmented with the boolean `text_cursor_changed` which `widget_info` would check and dispatch that special event. `OutputEvent::TextChanged` can be removed (`OutputEvent::Changed` would be sent when text changes, and the receiver can figure out that it was text that changed). I think this is the cleanest design.

emilk:
Thank you for taking the time for this! I completely understand that it is a bit overwhelming 😅

ndarilek:
Oh wow, thanks for catching that. I'll fix tomorrow.

egui/src/widgets/text_edit.rs
ndarilek: I certainly won't pick this hill to die on, but I do strongly feel that we should keep these as plain strings for now. I take your point about the extra `clone` and am certainly willing to revisit if there are performance issues, but the alternative is a bunch of checks/unwraps anywhere that needs access to both strings, and I'm not sure that would be any better than a quick clone of what would likely only be a few bytes. I'll make the change if you insist, though.

Your point about `None` being a useful signal is well-taken, though. I did an equality check for `ValueChanged`, and if they're equal then I set the previous value to `None`. Again, if we actually find string equality checks to be a bottleneck then I'll change that to `Option`.

ndarilek:
This is one of the more complex PRs I've worked with, so I tried to close conversations I thought might be resolved. Except for #399, I think everything might be resolved. Please let me know if I missed something, or if there's anything else you might want reworked.

Haven't updated the demo yet, but in my game, everything now works as expected, and I got rid of the crappy selection cache since those changes are dispatched correctly.

There are two more things I need for my game: sliders, and support for entering passwords (I.e. not reading characters as they're entered, or reading the field when it gets focus.) I'll work on those immediately after this PR lands, but I feel like there's a lot in flight with this one, so I'd like to get it merged before I work on those. Hope that's OK.

egui/src/widgets/text_edit.rs
emilk: I'm a bit confused though.

I agree the simpler solution would be to have `let prev_text = text.clone()` at the start of the function, and then compare with that later. Super-simple, though it costs an extra clone. And no need for any book-keeping of when the text changed. If you want to do this, go ahead (we can optimize it later).

But that is now what the PR currently does - instead it passes in a `mut` ref to `prev_text` into all functions that modify text, and manually set it when the text changes. This avoids the clone when there is no change, but it also has bugs: if I press two keys the same frame, the `prev_text` is set to the second value, not the first initial value. So if the start is `"A"` and I press `B` and then `C`, the current code will change `prev_text` first to `"A"` and then to `"AB"`, and the event will emit `Text changed from "AB" to "ABC"` when it in fact changed from "A" to "ABC".

So the code in the PR has all the the book-keeping to track of when the text changes (which is great for avoid clones) but because it isn't using `None` it has this bug.

ndarilek: I guess I'm a little confused too. Is this something you're able to duplicate, or hypothetical?

I'm not the fastest typist, but I placed my fingers over the f and d keys and tried to bring them down simultaneously. I think I succeeded about as well as I could, but either I didn't manage multiple keypresses in a single frame or I received two events. Tried multiple times. I am on a wireless keyboard so perhaps that introduced some delay, but it types quickly enough.

Also, in practical terms, I wouldn't be able to hear two spoken characters at a typical 1/60 framerate.

I'll accept that this isn't technically correct, but in practice it works, and I still maintain that the intent is clearer than it'd be with bunches of unwraps. Your call, though.

bevy_egui_a11y updated to master. Sorry for the delay, stepped away to work on other projects for a bit.

emilk: Since you are cloning the string each frame, there is no need to pass in `&mut prev_string` to all the functions. cherry-pick this and see: https://github.com/emilk/egui/commit/77b3a154b4a10e8cc9be674a7199fa29380e13bb

emilk (changes requested):
Tested this now, and it works really well! Good job!

I did find one bug: while holding down the mouse button in a text edit field the `ValueChanged` event is being spammed each frame. I don't know why though, but it seems to be introduced by this PR.

Also: see https://github.com/emilk/egui/commit/77b3a154b4a10e8cc9be674a7199fa29380e13bb for how to simplify the code by a lot.

egui/src/widgets/text_edit.rs
ndarilek: OK, that does indeed work. Thanks for that--things are still a bit unintuitive but they're starting to click.

ndarilek:
I went ahead and merged upstream today, resolving the couple merge conflicts that popped up.

I think we may be good to go. Please let me know if I missed anything--tracking all the requested changes has been a bit challenging but I think I got them all.

emilk:
> I did find one bug: while holding down the mouse button in a text edit field the `ValueChanged` event is being spammed each frame. I don't know why though, but it seems to be introduced by this PR.

Did you manage to fix this?

emilk:
> Unfortunately I can't fix the mouse issue without help. Sorry. I can't see the field to click in it, and my attempts to do so just click wherever the mouse happens to be. Assistance welcome when you have the time.
> 
> Thanks.

I'll fix after merge! Thank you so much for this PR!

emilk:
> I did find one bug: while holding down the mouse button in a text edit field the `ValueChanged` event is being spammed each frame. I don't know why though, but it seems to be introduced by this PR.

Fixed in f4a95b1e5f45b395d9a08d941f90559ca7230254

Issue #136, opened 2021-01-25
Would you please add localization support


xue-blood:
The text in widget are hard code in English.
Do you have the plan for support different language?
Thanks a lot.

emilk:
There aren't that many strings in egui right now that I can think of. There are a few tooltips when you hover e.g. a slider or a dragvalue. Is it these few strings you would like localized? I guess I can add them to some `Options` struct somewhere.

xue-blood:
I'm glad for your reply.
I had searched the whole by using regex `".*?"[\),]`.
I think this bellow can be localize.

`egui\src\widgets\slider.rs`
- `"{}\nClick to enter a value."`

`egui\src\widgets\mod.rs`
- `.on_hover_text("Extrusion")`
- `.on_hover_text("Width")`
- `if ui.add(Button::new("Reset").enabled(*value != def)).clicked`

`egui\src\widgets\drag_value.rs`
- `"{}{}{}\nDrag to edit or click to enter a value."`

`egui\src\widgets\color_picker.rs`
- `on_hover_text("Click to edit color")`
- `ui.label("Value / Saturation")`
- `ui.label("Value");`
- `ui.label("Saturation")`
- `ui.label("Hue")`
- `ui.label("Selected color")`
- `ui.label("Alpha")`
- `ui.label("Selected color")`
- `ui.radio_value(&mut additive, false, "Normal")`
- `ui.radio_value(&mut additive, true, "Additive")`
- `ui.label("Blending:")`
- `on_hover_text("Click to copy")`
- `"RGBA (premultiplied): rgba({}, {}, {}, {})"`

jgrussell:
What would you think about integrating one of the existing i18n crates into this project?

If you think this is a good idea, which one?  The two leading contenders seem to be [fluent-rs](https://crates.io/crates/fluent) and [gettext-rs](https://crates.io/crates/gettext-rs).

I am very new to Rust and have never contributed to an open source project; but, this seems like an issue that I might be able to successfully tackle with some patience and guidance from this community.

parasyte:
Just my opinion, but both options do not feel like a great fit for egui. Both are overkill for supporting translations of the entire corpus of English text provided in https://github.com/emilk/egui/issues/136#issuecomment-767284569 I can't be sure that this is a complete list, but even if it's a majority of text that appears in the UI that's worth pointing out.

[...]

PR #426, opened 2021-05-21
Localization of Widgets strings

flawphobic:
Closes https://github.com/emilk/egui/issues/136.

I'm in the middle of something right now so my write-up will be limited.

Not sure if I'm doing anything right to be honest but here's the gist of it:

- Added a localization module containing the `Localization` struct and `Language` enum
- `Localization` has methods pertaining to the language to localize into. For example, `malay()` returns a copy of `Localization` with all fields filled with Bahasa Malaysia texts; English is a bit special as it is covered by the `default()` 
- `Context` now has a `localization` field that stores `Localization`
- `Context` can be used to set localization by using it in tandem with the `Language` enum
- Added `new_language` as a field in `Memory`; it works similarly (I think) to `new_font_definitions` and `new_pixels_per_point` wherein if available, it'll set the new language at the start of the next frame
- `Localization` struct supports two localized texts - English and Bahasa Malaysia. However, the extent of localization is only for the color picker widget; I'll add more if the current code is satisfactory
- Updated the demo to include a language switcher (I might add a GIF to illustrate this)

My concerns:
 - I'm not familiar with mutexes, shared references, and the likes so it's very likely that a lot of my codes will be non-idiomatic
 - To overcome some snags with shared references, I used `clone()`; not sure if that's a good idea
 - No documentations/tests for now. I'll add them if this PR has the potential to be merged


egui/src/widgets/color_picker.rs
```diff
@@ -206,14 +206,15 @@ pub enum Alpha {
 
 fn color_text_ui(ui: &mut Ui, color: impl Into<Color32>) {
     let color = color.into();
+    let button_text = &ui.ctx().localization().click_copy.clone();
```
emilk, line 209:
```suggestion
    let button_text = ui.ctx().localization().click_copy;
```

Should work, no? Since `click_copy` is `&'static str` this should just copy that.

flawphobic: That's true. I've removed all the clones and everything still works.

egui/src/localization.rs
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Clone)]
+#[cfg_attr(feature = "persistence", derive(serde::Deserialize, serde::Serialize))]
+/// Specifies the languages currently available for localization and is required by [`crate::Context::set_localization`] as the parameter type.
+pub enum Language {
+    English,
+    BahasaMalaysia,
```
emilk, line 57: Do we need this list? Can't the user just call `context.set_localization(Localization::malay());` instead? That way it is also easy for a user to add a new language without having to make a PR to egui.

emilk: `Localization::lang` could be changed to `&'static str` too (e.g. "english")

egui/src/memory.rs
```diff
@@ -53,6 +56,9 @@ pub struct Memory {
     /// new fonts that will be applied at the start of the next frame
     pub(crate) new_font_definitions: Option<epaint::text::FontDefinitions>,
 
+    /// new language that will be applied at the start of the next frame
+    pub(crate) new_language: Option<Language>,
```
emilk, line 60: No need for this - let's make it instant instead.

egui/src/context.rs
```diff
@@ -950,3 +955,25 @@ impl Context {
         self.set_style(style);
     }
 }
+
+/// ## Localization
+impl Context {
+    /// Sets the current language of the default texts within [`crate::widgets`] and [`crate::containers`].
+    ///
+    /// The languages available for localization can be seen in [`crate::localization::Language`].
+    pub fn set_localization(&self, lang: Language) {
```
emilk, line 964:
```suggestion
    pub fn set_localization(&self, localization: Localization) {
```

emilk (changes requested):

egui/src/localization.rs
Mingun: > `Localization::lang` could be changed to `&'static str` too (e.g. "english")

Agree, some clarification: `<language code>[_<territory code>]` should be used (`en`, `en_US` and so on)

emilk:
I think perhaps we should instead remove strings from egui (like we did in https://github.com/emilk/egui/pull/704) instead of going down this path, or let the user set the egui helper string directly, i.e. something like this, where the user of egui choses what goes where:

* `ctx.help_texts().drag_value = "Click to edit value";`
* `ctx.help_texts().drag_value = "Klicka för att skriva in ett värde"; // if you are Swedish`
* `ctx.help_texts().drag_value = ""; // show nothing`

PR #433, opened 2021-05-24
Detect single-threaded mutex reentry in debug mode

TheCallSign:
I ran up against being unable to use egui from muliple threads in debug mode recently. I gave a shot at allowing for multithreading in debug mode by fixing this TODO:

https://github.com/emilk/egui/blob/196ddff499e7e27c0abf3119ac93fd5df81cb081/epaint/src/mutex.rs#L25

This will have a penalty when running in debug mode but the trade off is that one can actually run egui in multi-threaded mode.


emilk (approved):
Looks great - thanks!

Issue #439, opened 2021-05-27
Limit ordered-float to version 2.4


baysmith:
ordered-float has released version 2.5 which adds a dependency on getrandom which doesn't compile for wasm32.

Perhaps the ordered-float dependency should specify version 2.4? Or maybe certain feature settings would work?

I have used `cargo update -p ordered-float --precise 2.4.0` as a workaround.

emilk:
Please verify that this was indeed fixed by https://github.com/emilk/egui/pull/440 or I will probably remove `ordered-float` as a dependency

PR #440, opened 2021-05-27
Disable default features of `ord-float`

OfficialURL:
This probably closes <https://github.com/emilk/egui/issues/439>. This should disable the `std` features of the `ordered-float` crate, allowing the latest version to compile on WASM. I ran `cargo update` to test this.


emilk (approved):
Good catch - thanks!

PR #442, opened 2021-05-27
Add ability to create a checkbox or radiobutton with label on other side

### LINKED ISSUE #436: Change position of `Checkbox` label
Currently it seems like there's no way to make a checkbox (or radiobox) whose text precedes its button. This would be a useful layout option, as having labels before controls and text after checkboxes clashes a lot.

vv9k:
Closes #436.

### DIFF
--- a/egui/src/widgets/button.rs
+++ b/egui/src/widgets/button.rs
@@ -214,6 +214,7 @@ pub struct Checkbox<'a> {
     text: String,
     text_color: Option<Color32>,
     text_style: Option<TextStyle>,
+    left_sided: bool,
 }
 
 impl<'a> Checkbox<'a> {
@@ -224,6 +225,7 @@ impl<'a> Checkbox<'a> {
             text: text.to_string(),
             text_color: None,
             text_style: None,
+            left_sided: false,
         }
     }
 
@@ -236,6 +238,12 @@ impl<'a> Checkbox<'a> {
         self.text_style = Some(text_style);
         self
     }
+
+    /// If set to `true` the label will be displayed before the checkbox
+    pub fn label_on_left(mut self, is_on_left: bool) -> Self {
+        self.left_sided = is_on_left;
+        self
+    }
 }
 
 impl<'a> Widget for Checkbox<'a> {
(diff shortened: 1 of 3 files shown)


egui/src/widgets/button.rs
```diff
@@ -214,6 +214,7 @@ pub struct Checkbox<'a> {
     text: String,
     text_color: Option<Color32>,
     text_style: Option<TextStyle>,
+    left_sided: bool,
```
emilk, line 217:
```suggestion
    label_on_left: bool,
```

Let's use the same explicit name consistently

egui/src/widgets/button.rs
```diff
@@ -236,6 +238,12 @@ impl<'a> Checkbox<'a> {
         self.text_style = Some(text_style);
         self
     }
+
+    /// If set to `true` the label will be displayed before the checkbox
+    pub fn label_on_left(mut self, is_on_left: bool) -> Self {
+        self.left_sided = is_on_left;
+        self
```
emilk, line 245:
```suggestion
    pub fn label_on_left(mut self, label_on_left: bool) -> Self {
        self.label_on_left = label_on_left;
        self
```

egui_demo_lib/src/apps/demo/widget_gallery.rs
```diff
@@ -124,6 +128,29 @@ impl WidgetGallery {
...
+            };
+            if ui
+                .add(egui::RadioButton::new(*radio == Enum::Third, "Third").label_on_left(true))
+                .clicked()
+            {
+                *radio = Enum::Third
+            };
+        });
+        ui.end_row();
+
```
emilk, line 153: The widget gallery is to showcase one of each widget, not do showcase all variants of each widget.

Please move this to e.g. `egui_demo_lib/src/apps/demo/misc_demo_window.rs` instead!

emilk (changes requested):
Nice!

Is this something you think people will want to tweak on per-checkbox basis? If it is more likely one would set this for an entire `Ui` or for the entire egui context then this options better belongs in `struct Spacing`

emilk:
Is there still interest for this?

PR #444, opened 2021-05-28
Expanded `TextBuffer` interface to allow borrowed values.

Zenithsiz:
Previously the `TextBuffer` trait only allowed owned values, since it relied on `Clone`, `Default` and `'static`.
This allows it to use borrowed values.

A possible use case for this is creating a buffer which has dynamic max length:
```rust
struct LimitedStringBuffer<'a> {
    text: &'a mut String,
    max_len: usize,
}
```

This wasn't possible in the previous api, as `Default` and `Clone` cannot be implemented for this struct, and it wasn't `'static`.

All new trait methods have a default impl using the existing 2 methods, so this is backwards compatible and also suffers from no performance loss on `String`s due to them implementing these new methods.

### DIFF
--- a/egui/src/widgets/text_edit.rs
+++ b/egui/src/widgets/text_edit.rs
@@ -5,11 +5,11 @@ use std::ops::Range;
 #[derive(Clone, Debug, Default)]
 #[cfg_attr(feature = "persistence", derive(serde::Deserialize, serde::Serialize))]
 #[cfg_attr(feature = "persistence", serde(default))]
-pub(crate) struct State<S: TextBuffer> {
+pub(crate) struct State {
     cursorp: Option<CursorPair>,
 
     #[cfg_attr(feature = "persistence", serde(skip))]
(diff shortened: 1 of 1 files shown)


emilk:
I wanted to wait with this until https://github.com/emilk/egui/pull/412 is in, and now it is.

This seems like a good idea, but there is a conflict with the above PR: we now make a clone of the text each frame (not great, not terrible) to easily detect and report changes for screen readers. This would have to change to instead of cloning into an `S` to clone into a `String` (e.g. call `text.to_string()` instead of `text.clone()`)

Zenithsiz:
It should also be possible to reuse the buffer by storing it somewhere, which would avoid an allocation every frame, and instead just do `old_text.clone_from(text.as_ref())`.

I haven't implemented that however, as I'm not sure where I could store the old buffer, nor if it would be a good idea.
Have rebased on master though.

emilk (approved):

PR #445, opened 2021-05-28
Fix scrolling demo

rick68:
fix the item which can't display the yellow color correctly.


emilk:
The current behavior is to only highlight the line when one is dragging the slider, and only track the line when one is dragging the slider. This is the intended behavior, as the highlighting correlates with the behavior. That is, once you let go of the slider, the yellow line has nothing special going for it, so why highlight it?

PR #458, opened 2021-06-03
Revert "Fix scroll bar position"

emilk:
Reverts emilk/egui#392

It turns out the code that was removed was there for a reason:

Screen Shot 2021-06-04 at 00 02 02

### DIFF
--- a/egui/src/containers/scroll_area.rs
+++ b/egui/src/containers/scroll_area.rs
@@ -137,9 +137,9 @@ impl ScrollArea {
             ),
             *ui.layout(),
         );
-        let content_clip_rect = inner_rect
-            .expand(ui.visuals().clip_rect_margin)
-            .intersect(ui.clip_rect());
+        let mut content_clip_rect = inner_rect.expand(ui.visuals().clip_rect_margin);
+        content_clip_rect = content_clip_rect.intersect(ui.clip_rect());
+        content_clip_rect.max.x = ui.clip_rect().max.x - current_scroll_bar_width; // Nice handling of forced resizing beyond the possible
         content_ui.set_clip_rect(content_clip_rect);
 
         let viewport = Rect::from_min_size(Pos2::ZERO + state.offset, inner_size);
@@ -251,7 +251,7 @@ impl Prepared {
         }
 
         let width = if inner_rect.width().is_finite() {
-            inner_rect.width() // Position scroll bar correctly
+            inner_rect.width().max(content_size.x) // Expand width to fit content
         } else {
             // ScrollArea is in an infinitely wide parent
             content_size.x


emilk:
If we had bidirectional scrolling sure. But turning on vertical scrolling should only mean "you can now fit more vertically, so the container can be shorter than the contents, and you can scroll to see it" NOT "you can now make the container more narrow than the contents so that you can no longer see all the contents".

parasyte:
Yep, I understand the reasons. Ultimately I'm trying to introduce horizontal scrolling (#391) But if you want to keep the existing behavior, H-scroll will have to be supported by a separate container type.

emilk:
Horizontal scrolling would be great, but should still be opt-in. I want to be able to keep the existing behavior where windows auto-size based on their content, which I think is a really good feature.

Shfty:
I've been following #391 and #392 since shortly after they opened, and feel like I should throw my hat in here:

I'm building out an async application that can visualize its own call graph using instrumentation from the `tracing` crate, and have written an egui integration to hook into its data / logic setup and drive UI rendering for it.

It uses a traditional tree approach to visualize call graphs, meaning lots of recursively-nested collapsing headers:

image

Their nesting depth isn't known statically as it can change depending on the enabled log level, so horizontal scrolling (or at minimum, window clipping) ends up being a big-ticket feature; the nested layout naturally creates rightward pressure, and only so much of that can be accounted for by forcing the window to enlarge.

As is visible in the screenshot, the hidden vertical scroll bar formerly fixed by #392 also heavily effects this use-case: At deeper levels of nesting, header titles frequently end up far enough to the right that they intersect the edge of the view and cause the early-hiding bug.

> Yep, I understand the reasons. Ultimately I'm trying to introduce horizontal scrolling (#391) But if you want to keep the existing behavior, H-scroll will have to be supported by a separate container type.

Would such an approach still be able to support scroll areas that can move in both axes at once, i.e. by nesting a horizontal area inside a vertical area or vice-versa? Presumably there would also be some extra overhead invoked by having to run two separate scroll area widgets on top of one another.

Thank you for trying to push this forward, by the way!

> Horizontal scrolling would be great, but should still be opt-in. I want to be able to keep the existing behavior where windows auto-size based on their content, which I think is a really good feature.

Coming from OS-level window handling, I'd argue that this behavior is unintuitive, and hurts more than it helps.

Take the Windows / Mac / Linux window managers as an example: No constraints are imposed on a window by default. They can resize independently of their content, and offer `min size` and `max size` options to constrain window geometry to a certain range. There is no concept of scrolling - merely of what is displayed versus how large the view area is.

Using `egui` inside one of these OS-level windows, I'm free to size it however I like based on the available desktop space. If my content simply can't fit on the desktop because it's too big or my screen is too small, I at least have some recourse that prevents the window from forcibly expanding itself offscreen, or over the top of other windows.

Whereas if I wanted to create an equivalent setup inside `egui`, or create an abstraction unifying OS-level `winit` windows and `egui` windows under some trait to allow for dockable / undockable panels, this would not be possible. The `egui` sizing implementation ends up being the lowest common denominator, meaning I either have to force-size OS-level windows for the sake of consistent behavior, or deal with a lop-sided abstraction that needs to denote extra edge cases for a subset of its behavior.

So while I can see the logic behind auto-sizing a more general container primitive based on its contents, a window primitive carries additional expectations of being able to freely resize and clip its children. The concept of scrolling is a distinct layer on top of that, and should not be conflated with, or used as the predicate for, clippable window content. Especially given that `egui` doesn't even cover that case in its current state...

On a slightly broader note, as a user it's disheartening to see so much focus on the assistive 'middleware' strata that now exists around `egui` (such as `epi` and `eframe`) when features so fundamental as freely-resizable windows and horizontal scrolling are still absent. I feel that writing my custom integration took more digging than it reasonably should have on account of this strata, as the idioms around lower-level work seem unclear versus the higher-level plug-and-play app approach.

I think it's great that we have a slick, pure Rust alternative to `imgui`, but please consider a greater focus on the core functionality.

emilk:
If I try to resize my firefox window, it refuses to be smaller than the size of the search bar. Maybe that's a hard-coded constant (min_size) but hopefully it is actually a hard constraint set by the size of the search bar. This is the sort of feature we currently support in egui, and that I want to continue supporting.

This is not at all at odds with adding horizontal scrolling, nor is it necessarily at odds with fixing the bug that was originally fixed in <https://github.com/emilk/egui/pull/392> - it just means we need a smarter fix.

parasyte:
I'm ok with opt-in for horizontal scrolling. My second attempt at this (so far) has been leaning toward an implementation that allows configuring which scroll bars should appear using an enum like:

```rust
enum ScrollBars {
    Vertical,
    Horizontal,
    Both,
}
```

Where the `Vertical` variant acts as it does today and is the default setting; The children choose the width, and the parent chooses the height. `Horizontal` does the opposite (children choose the height and parent chooses the width). And `Both` has the behavior implemented in #392 (parent chooses the width and height).

I do see the value with the `Vertical` default. But at the same time it has so many bugs at the extremes that it ultimately behaves poorly.

Shfty:
> If I try to resize my firefox window, it refuses to be smaller than the size of the search bar. Maybe that's a hard-coded constant (min_size) but hopefully it is actually a hard constraint set by the size of the search bar. This is the sort of feature we currently support in egui, and that I want to continue supporting.

That behavior is the `Core::Widget::nsBaseWidget` for the Firefox window applying minimum size constraints via one of the platform-wrapped WM APIs (such as `Core::Widget::Win32::nsWindow`), yes. It likely infers this from the size of the search bar, but that's a choice made by the user of the UI framework, not the UI framework itself or window primitive.

And regardless of the top bar, the content inside the webview will still extend past the bounds of the window if it's configured with a fixed size.

> This is not at all at odds with adding horizontal scrolling, nor is it necessarily at odds with fixing the bug that was originally fixed in #392 - it just means we need a smarter fix.

Sure, but horizontal scrolling isn't the entire issue. Stripping scrolling away entirely, there's still no apparent way to achieve the behavior in your opening screenshot: A window that can be manually resized to be smaller than the minimum bounds of its content.

In cases where UI space is limited, this is actually desirable behavior. I might want to impose a manual minimum bound (or perhaps none at all) in order to give the user of my program more control over how the interface is arranged.

While responsive layout behavior is desirable as a general rule, some content (such as a non-scaling, fixed-size image) is fundamentally at odds with it. Scrolling the overflowed axis is one way to fix the issue, but isn't sufficient for all cases.

What about:
- A window that only scrolls vertically, but can be resized to be narrower than its widest element
- Vice-versa for a prospective horizontal scroll view with vertically pressured content
- Or even just a window that overflows, doesn't scroll at all, but can still be freely resized

Ideally, the child elements' minimum size would be applied initially, but not be a fixed minimum. Having the current behavior as a default would be fine, but making it a hard rule renders various layout-content combinations impossible.

> I'm ok with opt-in for horizontal scrolling. My second attempt at this (so far) has been leaning toward an implementation that allows configuring which scroll bars should appear using an enum like:
> 
> ```rust
> enum ScrollBars {
>     Vertical,
>     Horizontal,
>     Both,
> }
> ```
> 
> Where the `Vertical` variant acts as it does today and is the default setting; The children choose the width, and the parent chooses the height. `Horizontal` does the opposite (children choose the height and parent chooses the width). And `Both` has the behavior implemented in #392 (parent chooses the width and height).
> 
> I do see the value with the `Vertical` default. But at the same time it has so many bugs at the extremes that it ultimately behaves poorly.

Whether it ends up being opt-in or opt-out doesn't really matter provided that the use-cases are covered. Retaining a vertical default is probably preferable from the API point of view, since it avoids introducing a breaking change and is resonably intuitive for most content.

The horizontal / vertical / both option on one primitive approach is in-line with the way other UI frameworks expose axis options for scroll views, which seems reasonable to me.

And this is probably a point better made in #391, but it might be worth looking at generics (i.e. `ScrollView<Horizontal>, ScrollView<Vertical>, ScrollView<Both>`) to encode the axes rather than a runtime enum value. Delegating axis-specific behavior to its own impl like that would allow the unused axis' logic to be skipped statically, though might introduce some complications with making sure the appropriate state is available for runtime switching.

parasyte:
> And this is probably a point better made in #391, but it might be worth looking at generics (i.e. `ScrollView<Horizontal>, ScrollView<Vertical>, ScrollView<Both>`) to encode the axes rather than a runtime enum value. Delegating axis-specific behavior to its own impl like that would allow the unused axis' logic to be skipped statically, though might introduce some complications with making sure the appropriate state is available for runtime switching.

I might agree on principle, but in practice separating the two appears to be challenging. To start with, the vertical scroll bar (as implemented today) steals space from the children. You can see this easily be giving a resizable window enough height that it doesn't display the scroll bar, then dragging the resize handle from the bottom up until the scroll bar comes into view. Some alternatives would be stealing external space by widening the window (if possible) or just overlaying the scroll bar on top of the content area.

Stealing space from the children means that it necessarily must also scroll horizontally to reveal the hidden part, but that also means stealing vertical space to display the horizontal scroll bar. Which ultimately means the vertical scrolling will be invalidated. Resolving this kind of circular dependency is crucial for UX. It's made even more difficult with independent views that have strict (immediate mode) ordering.

There are similar ordering problems with the existing vertical scroll implementation, like this line which pins the content clipping rectangle to the minimum size required by the children: https://github.com/emilk/egui/blob/02db9ee5835a522ddf04308f259025388abf0185/egui/src/containers/scroll_area.rs#L142

The catch is that this code runs _before_ the children have been added to the UI, so it shouldn't even know what the clip rect size should be! It only works today because the `Resize` container (the parent of `ScrollArea` in a scrollable `Window`) remembers the last size of its children: https://github.com/emilk/egui/blob/02db9ee5835a522ddf04308f259025388abf0185/egui/src/containers/resize.rs#L262 I don't know if this was intentional or serendipitous. But that's how it is.

Changing any of this now while keeping the same (or similar) behavior seems untenable, and I am under the impression that emilk wants to leave it as is. I could argue my own preference, but I would rather make progress with compromise.

Shfty:
> I might agree on principle, but in practice separating the two appears to be challenging.

I'm not suggesting it in terms of separating the behavior so thoroughly that `ScrollArea<Horizontal>` and `ScrollArea<Vertical>` end up like the initially-proposed `HorizontalScrollArea` / `VerticalScrollArea`; more in terms of being a static replacement for the `if` and `match` conditionals used to do things like ignore input for an axis or show/hide each bar.

Ultimately though, it's a saving of one branch / jump table per callsite, along with some mild API strength gains, which isn't make-or-break by comparison to having horizontal scrolling in the first place!

> To start with, the vertical scroll bar (as implemented today) steals space from the children. You can see this easily be giving a resizable window enough height that it doesn't display the scroll bar, then dragging the resize handle from the bottom up until the scroll bar comes into view. Some alternatives would be stealing external space by widening the window (if possible) or just overlaying the scroll bar on top of the content area.

Stealing space from the children is an intuitive approach I think - no UI I've ever used has expanded a window to fit its scroll bar.

Overlaying is its own contextual thing on top of stealing, since it's sometimes necessary if there's not enough space, but otherwise undesirable.

> Stealing space from the children means that it necessarily must also scroll horizontally to reveal the hidden part, but that also means stealing vertical space to display the horizontal scroll bar. Which ultimately means the vertical scrolling will be invalidated. Resolving this kind of circular dependency is crucial for UX. It's made even more difficult with independent views that have strict (immediate mode) ordering.

The bulk of this problem evaporates if windows are given the option to be authoritative over their min-size, rather than having it forced by their children.

You only need to introduce scrolling for an axis to reveal hidden content (and engage in the resulting mind-bending circular dependency resolution) if there's a constraint in place that demands it, as is the case now.

> There are similar ordering problems with the existing vertical scroll implementation, like this line which pins the content clipping rectangle to the minimum size required by the children:
> 
> https://github.com/emilk/egui/blob/02db9ee5835a522ddf04308f259025388abf0185/egui/src/containers/scroll_area.rs#L142
> 
> The catch is that this code runs _before_ the children have been added to the UI, so it shouldn't even know what the clip rect size should be! It only works today because the `Resize` container (the parent of `ScrollArea` in a scrollable `Window`) remembers the last size of its children:
> 
> https://github.com/emilk/egui/blob/02db9ee5835a522ddf04308f259025388abf0185/egui/src/containers/resize.rs#L262
> I don't know if this was intentional or serendipitous. But that's how it is.

Having a `Resize` maintaining a window's size seems intuitive to me, because it implies ownership of that data - that the window is in fact authoritative over its geometry, as is the case in the OS-level window managers `egui` is trying to emulate with its `Window` primitive.

What's not intuitive is the hard-coded behavior pinning it to the minimum size of its children - while useful for initial layout or cases where you do want to constrain, forcing it makes no sense for a UI primitive that the library consumer or end user should in theory have control over. At best that should be a boolean `constrain_min_size_to_children` flag.

> Changing any of this now while keeping the same (or similar) behavior seems untenable, and I am under the impression that emilk wants to leave it as is. I could argue my own preference, but I would rather make progress with compromise.

Ultimately it comes down to the direction emilk wants to take it, but given that `egui` is still pre-1.0 (and disclaimed with potential for breaking changes in the readme), I'd say that now is the best time to push for improvements and illustrate common use-cases that may still be unserved. It's always preferable to avoid breaking changes, but not when the usability of your software hinges on it.

Shfty:
A more visual `imgui`-powered example of what I'm getting at:

imgui-example

Horizontal space is stolen from the responsive text by the vertical scrollbar, but elements that aren't able to shrink simply don't, and the window is still free to resize.

parasyte:
I consider this issue resolved following #475. There are a few remaining issues with the scroll bar implementation, but those are for another ticket.

PR #471, opened 2021-06-07
More plot items

### LINKED ISSUE #468: Add more plot items
I'd like to add the following items:
-  Arrows, also called "Quiver plots" in matplotlib etc.
-  (Convex*) polygons
-  Text
-  Images

Additionally it would be nice to make the following changes to existing plot items:
- Make `HLine`/`VLine` into `PlotItem`s as well.
- Add a "fill" property to `Line` so that we can fill/shade the area between a line and a horizontal reference line.
- Add stems to `Points`, which are lines between the points and a horizontal reference line.
- Allow using `..` when specifying ranges for values generated by explicit callback functions, as an alias for `f64::NEG_INFINITY..f64::INFINITY`
- Allow using ranges with exclusive end bounds for values generated by parametric callback functions to generate values where the first and last value are not the same.

*I also have an implementation of the ear-clipping algorithm lying around that can be used to tessellate non-convex polygons. It's relatively few lines of code, but it's not the most efficient algorithm with O(n²) time complexity. I could add this to the plotting library, but maybe it's something for egui in general?

I've already implemented those things and will open a PR once #410 is merged

https://user-images.githubusercontent.com/1352472/121086246-3542b580-c7e3-11eb-922c-6a7a94e60aa6.mp4

EmbersArc:
Closes #468. See the issue for the changes.

This got a bit out of hand, sorry about that. I'll have to break it up a bit next, I didn't want to do that in this PR to keep the diff relatively manageable.

@niladic Maybe you want to have a look as well, this changed a couple of things about plot items.

https://user-images.githubusercontent.com/1352472/121086246-3542b580-c7e3-11eb-922c-6a7a94e60aa6.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ NOTE: [`eframe`](eframe/CHANGELOG.md), [`egui_web`](egui_web/CHANGELOG.md) and [
 ## Unreleased
 
 ### Added ⭐
+* [More plot items: Arrows, Polygons, Text, Images](https://github.com/emilk/egui/pull/471).
 * [Plot legend improvements](https://github.com/emilk/egui/pull/410).
 * [Line markers for plots](https://github.com/emilk/egui/pull/363).
 * Add right and bottom panels (`SidePanel::right` and `Panel::bottom`).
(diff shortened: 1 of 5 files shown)


niladic:
Thanks for the heads up @EmbersArc this is impressive! I checked your previous work on your repo before beginning, so what I am working on is compatible. I actually did the same as you with the `get_bounds` method, I had to rework the `series` method that you renamed `values`, but it should be an easy rebase. I will try to push something this week.

emilk (approved):
Very impressive, good job!

PR #473, opened 2021-06-09
Grid fixes

spersson:
Three things here:
- fix vertical spacing when nesting a grid within a grid.
- use current state instead of previous for getting row height, used for advancing cursor
- fix horizontal alignment for items following a cell containing a nested layout.

```rust
egui::Grid::new("unique_id")
        .show(ui, |ui| {
            ui.vertical(|ui| {
                ui.label("nest1");
                ui.label("nest2");
            });
            ui.label("First row, second column");
            ui.end_row();
            
            ui.label("Second row, second column");
            ui.label("Second row, first column");
            ui.end_row();
            
            ui.label("Third row, first column");
            ui.label("Third row, second column");
            ui.end_row();

            egui::Grid::new("unique_id_2").show(ui, |ui|{
                ui.label("nest11");
                ui.label("nest12");
                ui.end_row();
                ui.label("nest21");
                ui.label("nest22");
                ui.end_row();
            });
            ui.label("Fourth row, second column");
            ui.end_row();
            
            ui.label("row 5, 1");
            ui.label("row 5, 2");
            ui.end_row();
        });
```
## Before:
Screenshot_before_fix

## After
Screenshot_after_fix

### DIFF
--- a/egui/src/grid.rs
+++ b/egui/src/grid.rs
@@ -184,7 +184,12 @@ impl GridLayout {
 
     pub(crate) fn end_row(&mut self, cursor: &mut Rect, painter: &Painter) {
         cursor.min.x = self.initial_x;
-        cursor.min.y += self.curr_state.row_height(self.row).unwrap_or(self.min_cell_size.y) + self.spacing.y;
+        cursor.min.y += self.spacing.y;
+        cursor.min.y += self
+            .curr_state
+            .row_height(self.row)
+            .unwrap_or(self.min_cell_size.y);
+
         self.col = 0;
         self.row += 1;
 
@@ -332,21 +337,24 @@ impl Grid {
         // If somebody wants to wrap more things inside a cell,
         // then we should pick a default layout that matches that alignment,
         // which we do here:
-        ui.allocate_ui_at_rect(ui.cursor(), |ui|{
-            let id = ui.make_persistent_id(id_source);
-            let grid = GridLayout {
-                striped,
-                spacing,
-                min_cell_size: vec2(min_col_width, min_row_height),
-                max_cell_size,
-                row: start_row,
-                ..GridLayout::new(ui, id)
-            };
+        ui.allocate_ui_at_rect(ui.cursor(), |ui| {
+            ui.horizontal(|ui| {
+                let id = ui.make_persistent_id(id_source);
+                let grid = GridLayout {
+                    striped,
+                    spacing,
+                    min_cell_size: vec2(min_col_width, min_row_height),
+                    max_cell_size,
+                    row: start_row,
+                    ..GridLayout::new(ui, id)
+                };
 
-            ui.set_grid(grid);
-            let r = add_contents(ui);
-            ui.save_grid();
-            r
+                ui.set_grid(grid);
+                let r = add_contents(ui);
+                ui.save_grid();
+                r
+            })
+            .inner
         })
     }
 }
--- a/egui/src/grid.rs
+++ b/egui/src/grid.rs
@@ -148,7 +148,7 @@ impl GridLayout {
         self.align_size_within_rect(size, frame)
     }
 
-    pub(crate) fn advance(&mut self, cursor: &mut Rect, frame_rect: Rect, widget_rect: Rect) {
+    pub(crate) fn advance(&mut self, cursor: &mut Rect, _frame_rect: Rect, widget_rect: Rect) {
         let debug_expand_width = self.style.debug.show_expand_width;
         let debug_expand_height = self.style.debug.show_expand_height;
         if debug_expand_width || debug_expand_height {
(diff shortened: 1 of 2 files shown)


spersson:
Forgot to say that I am unsure about the following comment:
```
       // Each grid cell is aligned LEFT_CENTER.
        // If somebody wants to wrap more things inside a cell,
        // then we should pick a default layout that matches that alignment,
        // which we do here:
        ui.allocate_ui_at_rect(ui.cursor(), |ui|{
```
where it was previously a horizontal() call instead of allocate.
Since I don't understand the comment I don't know if my change breaks something.

emilk:
Please add your test code to `TableTest` in `egui_demo_lib/src/apps/demo/tests.rs` so we don't regress on this again!

egui/src/grid.rs
```diff
@@ -178,15 +178,13 @@ impl GridLayout {
             widget_rect.height().at_least(self.min_cell_size.y),
         );
 
+        cursor.min.x += self.prev_col_width(self.col) + self.spacing.x;
```
emilk, line 181: `frame_rect` here is what is returned by `next_cell`, which currently is the maximum of requested width the the width of the previous column. So afaict the only change here that we advance less than we should in some situations?

egui/src/grid.rs
```diff
@@ -178,15 +178,13 @@ impl GridLayout {
...
         self.col += 1;
-        cursor.min.x += frame_rect.width() + self.spacing.x;
     }
 
     pub(crate) fn end_row(&mut self, cursor: &mut Rect, painter: &Painter) {
-        let row_height = self.prev_row_height(self.row);
-
         cursor.min.x = self.initial_x;
-        cursor.min.y += row_height + self.spacing.y;
+        cursor.min.y += self.curr_state.row_height(self.row).unwrap_or(self.min_cell_size.y) + self.spacing.y;
```
emilk, line 187: This makes sense to me - we know how high the row turned out to be, so we should use that instead of the height from previous frame. 👍

egui/src/grid.rs
```diff
@@ -334,7 +332,7 @@ impl Grid {
         // If somebody wants to wrap more things inside a cell,
         // then we should pick a default layout that matches that alignment,
         // which we do here:
-        ui.horizontal(|ui| {
+        ui.allocate_ui_at_rect(ui.cursor(), |ui|{
```
emilk, line 335: The previous code ensured that the default layout of each grid cell is `horizontal`. i.e. if you call `ui.scope(|ui| { ui.label("foo"); ui.label("bar"); })` from within a grid cell, it would show up as "foo bar".

With the new code we will instead use the same layout as whatever the grid sits in (e.g. vertical if it is in a vertical layout). This could break some existing code.

I think the safe choice is

``` rust
ui.allocate_ui_at_rect(ui.cursor(), |ui| {
    ui.horizontal(|ui| {
```

emilk (changes requested):
Your before/after images are compelling! But please add it as a test.

Also a hot tip: configure your editor to run `cargo fmt` on save. It is wonderful :)

egui/src/grid.rs
spersson: Opposite, kind of. :) The problem I'm trying to fix here is that it was advancing less than the column width when a cell contains a nested ui.
I'm very new to your code base and this is a bit tricky stuff, but just from looking at where Placer advance_after_rects is called it seems many of those call sites use child_ui.min_rect() to pass as the frame_rect. Which would explain why my fix works. My change makes it advance the full width of the current column, not just the minimum width of the child ui.

Do you think there's something that breaks with this change? Should it be changed to be advance by the maximum of frame_rect width and prev col width?

spersson:
Screenshots of the test I added to demo now.
## Test case before
Screenshot_before_fix2

## Test case after
Screenshot_after_fix2

egui/src/grid.rs
emilk: > I'm very new to your code base and this is a bit tricky stuff

I wrote the darn code, and I still don't understand all of it 😅  Doing layout in immediate mode is very tricky, since it is a balancing act of compromises.

> Do you think there's something that breaks with this change? Should it be changed to be advance by the maximum of frame_rect width and prev col width?

Possibly there will be an extra frame of delay when something expands in a grid now, so maybe the maximum would be better. If the grid test would include a slider to change the width of the contents of one cell we could test this.

emilk:
Thanks for adding a test!

emilk (approved):
Thanks for this!

Please run `cargo fmt` and `cargo clippy` and we can get this merged

egui_demo_lib/src/apps/demo/tests.rs
```diff
@@ -246,6 +246,47 @@ impl super::View for TableTest {
                     ui.end_row();
                 }
             });
+            
+        ui.separator();
+        
+        egui::Grid::new("parent grid")
```
emilk, line 252: `.striped(true)` makes it a little more legible

emilk:
Rebase on latest `master` to fix the clippy errors

egui/src/grid.rs
spersson: I added a slider now to the table test, tried with and without at least frame rect width. Playing with the slider I was unable to see any difference.

emilk:
Thank you!

PR #475, opened 2021-06-09
Make sure the scroll bar is never outside the screen rectangle

emilk:
This is an alternative attempt to fix the bug mentioned in https://github.com/emilk/egui/pull/392

egui expects that the container can always be made wider, which is true for all egui Ui:s, but not true for the outer frame/chrome that egui ultimately needs to sit within.

This handles the example of resizing the Color Test too narrow.

PTAL @parasyte


parasyte:
Looks good at first, except the margin isn't right when the max width is exceeded. Also is allows unusual things to happen.

clipping

PR #481, opened 2021-06-11
`Frame` background fix

### LINKED ISSUE #480: `Frame` background scales incorrectly
As we noticed in https://github.com/emilk/egui/pull/410#issuecomment-856236061, the `Frame` background does not behave as expected:

With the fix proposed in the comment, it seems to work fine. At least in the cases I've tried.

EmbersArc:
Closes #480. I just applied the suggested fix. Looking at the demo this does not seem to have any unwanted side effects.


emilk (approved):
Thanks!

PR #482, opened 2021-06-12
Plot: Line styles

### LINKED ISSUE #474: Plot: Dashed and dotted lines
I'd like to add dashed and dotted line styles to the plotting library:

https://user-images.githubusercontent.com/1352472/121424693-e70fec80-c971-11eb-8847-82adb34425d5.mp4

### LINKED ISSUE #524: HLines and VLines cannot have custom stroke colors
With egui 0.13.0, it is impossible to set a custom `Stroke` for the `HLine` and `VLine` plot items, since there is no setter method on either of those. (`Line` _does_ however have setter methods for the stroke, stroke width, and stroke color.)

I'm not sure if this was an oversight as part of the plot widget overhaul or an intended API change - if it was unintentional, I'd be happy to PR the missing setters!

EmbersArc:
Closes #474 and closes #524.

### DIFF
--- a/egui/src/widgets/plot/items.rs
+++ b/egui/src/widgets/plot/items.rs
@@ -37,25 +37,25 @@ impl Value {
 #[derive(Debug, PartialEq, Clone, Copy)]
 pub enum LineStyle {
     Solid,
-    Dotted { px: f32 },
-    Dashed { px: f32 },
+    Dotted { spacing: f32 },
+    Dashed { length: f32 },
 }
 
 impl LineStyle {
     pub fn dashed_loose() -> Self {
-        Self::Dashed { px: 10.0 }
+        Self::Dashed { length: 10.0 }
     }
 
     pub fn dashed_dense() -> Self {
-        Self::Dashed { px: 5.0 }
+        Self::Dashed { length: 5.0 }
     }
 
     pub fn dotted_loose() -> Self {
-        Self::Dotted { px: 10.0 }
+        Self::Dotted { spacing: 10.0 }
     }
 
     pub fn dotted_dense() -> Self {
-        Self::Dotted { px: 5.0 }
+        Self::Dotted { spacing: 5.0 }
     }
 
     fn style_line(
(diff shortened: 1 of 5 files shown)


emilk:
Tested, and it looks so pretty! I'm sorry I'm so slow to review, and I don't have the time right now either.

The dotting and dashing is such a nice feature that it would be nice to expose it outside of the plotter, i.e. add it to `epaint`. Maybe part of a constructor for `Shape` (`Shape::dashed(points, dash_length, stroke)`)? It could internally call the `dashes_from_line` function and return a `Shape::Vec`. In a more distant future the `LineStyle` could even become a member of `Stroke`.

egui/src/widgets/plot/items.rs
```diff
@@ -34,13 +34,151 @@ impl Value {
 
 // ----------------------------------------------------------------------------
 
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum LineStyle {
+    Solid,
+    Dotted { px: f32 },
+    Dashed { px: f32 },
```
emilk, line 41: "px" is a weird name. Call it "radius" and/or "length" if that is what it is.

EmbersArc: Changed in 80936ba

EmbersArc:
@emilk Thanks for the suggestions, I've moved the functions to `epaint`.

epaint/src/shape.rs
```diff
@@ -86,6 +86,36 @@ impl Shape {
...
+        points_from_line(points, spacing, radius, color.into(), &mut shapes);
+        shapes
+    }
+
+    /// Turn a line into dashes.
+    pub fn dashed_line(
+        points: &[Pos2],
+        stroke: impl Into<Stroke>,
+        dash_length: f32,
+        dash_gap_ratio: f32,
```
emilk, line 106: `dash_gap_ratio` deserves an explanation. Length of dash divided by length of the gap between them? it is a bit of a weird API imho. `dash_length` and `gap_length` seems simpler.

egui/src/widgets/plot/items.rs
```diff
@@ -82,20 +82,26 @@ impl LineStyle {
...
-                        points_from_line(&line, *px, radius, stroke.color, shapes)
+                        shapes.extend(Shape::dotted_line(&line, stroke.color, *spacing, radius))
                     }
-                    LineStyle::Dashed { px } => {
+                    LineStyle::Dashed { length } => {
                         if highlight {
                             stroke.width *= 2.0;
                         }
-                        dashes_from_line(&line, *px, stroke, shapes);
+                        let golden_ratio = (5.0_f32.sqrt() - 1.0) / 2.0; // 0.61803398875
```
emilk, line 98: hint: `let golden_ratio = (5.0_f32.sqrt() + 1.0) / 2.0; // 1.61803398875` to avoid the division later :)

emilk (approved):

epaint/src/shape.rs
EmbersArc: Applied in 5418cd7.

egui/src/widgets/plot/items.rs
EmbersArc: True, but made redundant by the other change :)

Issue #467, opened 2021-06-05
More flexible plotting (histogram, box plot...)


niladic:
I am impressed by how nice the plotting is in egui. It would be really useful to have more capabilities, like histograms or box plots.

# My attempt

I was not at all familiar with the codebase, but I tried to create such plots. A quick way I could get it done was: modify the `Points` struct, include something like a hashmap from `Value` to a `dyn Trait` that knows how to draw an element. This `dyn Trait` would contain the actual values of, say, a given box. The `dyn Trait` is to provide maximum flexibility to the end user:
```rust
// The "drawing trait"
pub trait DataPoint {
    // v- here `self` would contain `Value` (in a clean impl, this is not particularly desirable)
    fn get_shapes(&self, transform: &ScreenTransform) -> Vec<Shape>;
}

pub struct Points {
...
    // v- as it is, it would not work, since `Value` does not impl `Hash`
    points: HashMap<Value, Box<dyn DataPoint>>,
...
}
```
Then at the call site (`get_shapes` in `ItemPlot`), one can fetch the correct drawing impl for a given `Value` and create a `Vec<Shape>`.

There are a few problems with this approach:
* `Value` is not `Hash` nor `Eq`, this means hacking arround it
* the drawing trait is `pub` (so end users can create their own drawings), this means putting `pub` on `ScreenTransform` and `Bounds`
* a given `Value` is duplicated in memory, both in the `Values` struct and the hashmap

(I am not sharing code about what I describe because its a mess ;))

# Some ideas for implementing this feature

## Idea 1
Creating a new `impl` for `PlotItem` similar the existing `Points` (like I describe above).
Benefits: minimal code modification. Drawbacks: the ones I cited above.

## Idea 2
Modifying `Value` or `Values`, so they carry more data and the way they can be drawn. I am thinking about something like
```rust
pub trait ValueWrapper {
    fn value(&self) -> Value;
    fn shapes(&self, transform: &ScreenTransform) -> Vec<Shape>;
}
struct BoxPlotValue { ... };
impl ValueWrapper for BoxPlotValue { ... };

// such that (obviously wrong, I am just trying to convey the idea)
pub(super) trait PlotItem<V: ValueWrapper> {
...
    fn series(&self) -> &V;
...
}
impl Plot {
...
    pub fn points<V: ValueWrapper>(mut self, mut points: PlotItem<V>) -> Self {
...
}
```
[...]

PR #499, opened 2021-06-19
Add bar charts and boxplots

niladic:
Following up on #467, this PR adds bar charts and boxplots. Both work and can be tested in the demo app (I am sorry for the amount of code...). I open this PR as a draft because it conflicts with #471, and there might be some debatable choices:

1. I think it would be better to cut `items.rs` into submodules. I don't do it for ease of review here, but can do it before merging.
2. I tried to stay consistent with the existing API, only changing `PlotItem::series()` by `get_bounds()` and `closest()` (because it is only used in the hover behavior). My idea here it to make `PlotItem` sufficiently abstract to allow for more complex drawings like bar charts and boxplots here.
3. I created a `struct HoverElement` which can cause some concerns about performance. I choose this solution because it is simpler, however I am totally open to change it for some kind of reference (I did not see a simple implementation, suggestions are welcome).
4. There are fields named `dummy_values`, it is possible to remove them, but this will conflict with #471 (I can cleanup at some point)
5. I purposefully avoided statistics here: no mention of histogram (I called it bar chart, like many JS libraries do) and no quartiles or whiskers calculation. I guess it can always be added as an additional layer at one point. Note that [plotters](https://crates.io/crates/plotters) took the opposite approach.
6. ~In the plot demo, I put some values that were generated using `rand`, but did not know if it was appropriate to add the crate as a dependency.~ (fixed)

Screen Shot 2021-06-20 at 11 44 14
Screen Shot 2021-06-20 at 11 44 31

Screen Shot 2021-06-19 at 17 26 17
Screen Shot 2021-06-19 at 17 26 24
Screen Shot 2021-06-19 at 17 26 32
Screen Shot 2021-06-19 at 17 26 40


egui/src/widgets/plot/mod.rs
```diff
@@ -628,85 +658,24 @@ impl Prepared {
...
         let mut closest_dist_sq = interact_radius.powi(2);
         for item in items {
-            for value in &item.series().values {
-                let pos = transform.position_from_value(value);
-                let dist_sq = pointer.distance_sq(pos);
-                if dist_sq < closest_dist_sq {
-                    closest_dist_sq = dist_sq;
-                    closest_value = Some(value);
-                    closest_item = Some(item.name());
+            if let Some(closest) = item.closest(ui, pointer, transform, *show_x, *show_y) {
```
EmbersArc, line 664: This seems to create a bunch of shapes for all items and then throws all of them away except for maybe one set right? That is very inefficient.

egui/src/widgets/plot/items.rs
```diff
@@ -568,3 +662,1032 @@ impl PlotItem for Points {
...
+        self.point_at(self.position - self.whisker_width / 2.0, self.lower_whisker)
+    }
+
+    fn lower_whisker_center(&self) -> Value {
+        self.point_at(self.position, self.lower_whisker)
+    }
+
+    fn lower_whisker_right_end(&self) -> Value {
+        self.point_at(self.position + self.whisker_width / 2.0, self.lower_whisker)
+    }
```
EmbersArc, line 1297: Do we need functions for all of those? They are only used once...

egui/src/widgets/plot/items.rs
```diff
@@ -568,3 +662,1032 @@ impl PlotItem for Points {
...
+        };
+        shapes.push(rect);
+        shapes
+    }
+
+    fn default_values_format(&self, transform: &ScreenTransform) -> String {
+        let mut text = String::new();
+        let scale = transform.dvalue_dpos();
+        let y_decimals = ((-scale[1].abs().log10()).ceil().at_least(0.0) as usize).at_most(6);
+        text.push_str(&format!("\n{:.*}", y_decimals, self.height));
```
EmbersArc, line 862: Why not just return the format string here?

EmbersArc (changes requested):
Looks great! I like most of the new design decisions. I gave it a quick look and wrote down some suggestions.

egui/src/widgets/plot/items.rs
EmbersArc: Just noticed: Same goes for lines 1361-1365. Write just one format string since it may reduce allocations.

egui/src/widgets/plot/mod.rs
niladic: I just had an idea about that (commit 912b2f3)

egui/src/widgets/plot/items.rs
```diff
@@ -568,3 +662,1032 @@ impl PlotItem for Points {
...
+    }
+
+    fn bounds(&self) -> Bounds {
+        let mut bounds = Bounds::NOTHING;
+        bounds.extend_with(&self.bounds_min());
+        bounds.extend_with(&self.bounds_max());
+        bounds
+    }
+
+    fn shapes(&self, transform: &ScreenTransform, highlighted: bool) -> Vec<Shape> {
```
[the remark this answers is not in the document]
niladic: oops scala dev habit! I think I got them all

egui/src/widgets/plot/mod.rs
EmbersArc: This looks better, I don't know enough about about the language to estimate the performance impact of creating potentially thousands of closures though. Did you consider splitting this into two functions, one to get the distance to the item and one to then get the hover shapes? Then you could do some `min_by` and then call the trait method to get the shapes just for that object.

niladic: I considered the split with a simple `distance_square` and `hover_shapes` on `PlotItem`, but decided against because it required to calculate the distance twice. I think I can salvage it by putting an index instead of the closure, and having `fn hover_shapes(index: usize)` on `PlotItem`. I think at this point, I need to do some benchmark of the different ideas. I am not sure how to bench a gui app though, I saw some code in the repo. I will try to have something concrete later this week.

niladic: Sorry for the delay, it took me some time to get the benchmarks somewhat clean. I implemented an alternative here https://github.com/niladic/egui/commit/24d6547c9f5dfb1a8e760ad622cd533fb16f32d3. Also, after trying to implement a non boxed closure here, I found out that I was mistaken on the feasibility. It is not possible without existential types (unstable feature).

I benched (code here https://github.com/niladic/egui/commit/fd8e3448ff052404748814160d3b96581e94f46c and https://github.com/niladic/egui/commit/ecf907487be35366483073666e8f0fcca8ffa444)
* code before this PR as baseline [feature/add-barchart-v0-bench](/niladic/egui/tree/feature/add-barchart-v0-bench)
* my first commit with all the allocs [feature/add-barchart-v1-bench](/niladic/egui/tree/feature/add-barchart-v1-bench)
* the closure implementation [feature/add-barchart-v2-bench](/niladic/egui/tree/feature/add-barchart-v2-bench)
* the alternative with indexes [feature/add-barchart-v3-bench](/niladic/egui/tree/feature/add-barchart-v3-bench)

Results are on this branch [feature/add-barchart-bench-results](/niladic/egui/tree/feature/add-barchart-bench-results). I tried to do it as clean as possible on my machine, but the benches are still a bit noisy. I would say they are really close to each other and there are no regressions in terms of perf. The most interesting bits are [the boxed closure vs before this PR as baseline](https://github.com/niladic/egui/blob/feature/add-barchart-bench-results/cargo-output-v2-closure-vs-v0.txt), [the alternative vs before this PR as baseline](https://github.com/niladic/egui/blob/feature/add-barchart-bench-results/cargo-output-v3-with-index-vs-v0.txt) and [the alternative vs the boxed closure as baseline](https://github.com/niladic/egui/blob/feature/add-barchart-bench-results/cargo-output-v3-with-index-vs-v2-with-closure.txt).

I also added some flamegraphs on what I think is high but somewhat realistic load:
* [boxed closure 100 barcharts of 100 bars](https://github.com/niladic/egui/blob/feature/add-barchart-bench-results/v2-with-closure_plot_100x100_bars.svg)
* [boxed closure 100 lines of 100 points](https://github.com/niladic/egui/blob/feature/add-barchart-bench-results/v2-with-closure_plot_100x100_lines.svg)
* [boxed closure 100x100 points](https://github.com/niladic/egui/blob/feature/add-barchart-bench-results/v2-with-closure_plot_100x100_points.svg)
* [alternative 100 barcharts of 100 bars](https://github.com/niladic/egui/blob/feature/add-barchart-bench-results/v3-with-index_plot_100x100_bars.svg)
* [alternative 100 lines of 100 points](https://github.com/niladic/egui/blob/feature/add-barchart-bench-results/v3-with-index_plot_100x100_lines.svg)
* [alternative 100x100 points](https://github.com/niladic/egui/blob/feature/add-barchart-bench-results/v3-with-index_plot_100x100_points.svg)

To sum up, since there are no regressions and both implementations are comparable in terms of perf, I have no strong preference. The alternative with indexes might be simpler and easier to work with down the line. Tell me if you have any preference or other idea, otherwise I will go with the alternative one.

s-nie: Hey thanks a lot for doing those benchmarks! I'll suggest we go with the alternative using indices.

EmbersArc: Hey thanks a lot for doing those benchmarks! I agree that we should go with the alternative using indices.

emilk:
Merged in https://github.com/emilk/egui/pull/863

PR #505, opened 2021-06-22
Use old 1.51 toolchain instead of bleeding edge

emilk:
1.52 and 1.53 has incremental compilation turned off, so some people choose to stay on 1.51 for now.

So let's make sure egui supports 1.51 for a while!


emilk:
CI is green, but locally I get problems with `cint` when I run `cargo +1.51.0 check --workspace --all-targets --all-features`. I've contacted @termhn about possible solutions to this.

PR #519, opened 2021-06-26
Progress bar

### LINKED ISSUE #496: Progress bars
Quick question, does egui contain any kind of functionality to draw a progress bar? I haven't been able to find any.

Ideally it would be kinda similar to this imgui progress bar: https://github.com/ocornut/imgui/issues/1901

EmbersArc:
Closes #496.

https://user-images.githubusercontent.com/1352472/123526073-cb2c7a80-d6d5-11eb-9cb3-e1626a379523.mp4

### DIFF
--- a/egui/src/widgets/progress_bar.rs
+++ b/egui/src/widgets/progress_bar.rs
@@ -53,7 +53,7 @@ impl Widget for ProgressBar {
         let desired_width = desired_width.unwrap_or(ui.available_size_before_wrap().x);
         let height = ui.spacing().interact_size.y;
         let (outer_rect, response) =
-            ui.allocate_exact_size(vec2(desired_width, height), Sense::hover());
+            ui.allocate_exact_size(vec2(desired_width, height), Sense::click());
         let visuals = ui.style().visuals.clone();
         let corner_radius = outer_rect.height() / 2.0;
         ui.painter().rect(
--- a/egui/src/widgets/progress_bar.rs
+++ b/egui/src/widgets/progress_bar.rs
@@ -9,6 +9,7 @@ pub struct ProgressBar {
 }
 
 impl ProgressBar {
+    /// Progress in the `[0, 1]` range, where `1` means "completed".
     pub fn new(progress: f32) -> Self {
         Self {
             progress: progress.clamp(0.0, 1.0),
--- a/egui/src/widgets/progress_bar.rs
+++ b/egui/src/widgets/progress_bar.rs
@@ -1,10 +1,15 @@
 use crate::*;
 
+enum ProgressBarText {
+    Custom(String),
+    Percentage,
+}
+
 /// A simple progress bar.
 pub struct ProgressBar {
     progress: f32,
     desired_width: Option<f32>,
-    text: String,
+    text: Option<ProgressBarText>,
     animate: bool,
 }
 
@@ -14,7 +19,7 @@ impl ProgressBar {
         Self {
             progress: progress.clamp(0.0, 1.0),
             desired_width: None,
-            text: "".to_string(),
+            text: None,
             animate: false,
         }
     }
@@ -28,7 +33,13 @@ impl ProgressBar {
     /// A custom text to display on the progress bar.
     #[allow(clippy::needless_pass_by_value)]
     pub fn text(mut self, text: impl ToString) -> Self {
-        self.text = text.to_string();
+        self.text = Some(ProgressBarText::Custom(text.to_string()));
+        self
+    }
+
+    /// Show the progress in percent on the progress bar.
+    pub fn show_percentage(mut self) -> Self {
+        self.text = Some(ProgressBarText::Percentage);
         self
     }
 
@@ -77,7 +88,7 @@ impl Widget for ProgressBar {
             ui.ctx().request_repaint();
             lerp(dark..=bright, ui.input().time.cos().abs())
         } else {
-            (bright + dark) / 2.0
+            bright
         };
 
         ui.painter().rect(
@@ -109,8 +120,12 @@ impl Widget for ProgressBar {
             });
         }
 
-        if !text.is_empty() {
-            ui.painter().text(
+        if let Some(text_kind) = text {
+            let text = match text_kind {
+                ProgressBarText::Custom(string) => string,
+                ProgressBarText::Percentage => format!("{}%", (progress * 100.0) as usize),
+            };
+            ui.painter().sub_region(outer_rect).text(
                 outer_rect.left_center() + vec2(ui.spacing().item_spacing.x, 0.0),
                 Align2::LEFT_CENTER,
                 text,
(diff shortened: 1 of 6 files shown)


parasyte:
The animation looks awesome! I think I like it a little better with the selection background and text colors. What do you think?

egui-progress-dark

egui-progress-light

The diff is:

```diff
diff --git a/egui_demo_lib/src/apps/demo/progress_bar.rs b/egui_demo_lib/src/apps/demo/progress_bar.rs
index f0cb976..f37120f 100644
--- a/egui_demo_lib/src/apps/demo/progress_bar.rs
+++ b/egui_demo_lib/src/apps/demo/progress_bar.rs
@@ -72,11 +72,7 @@ impl Widget for ProgressBar {
             ),
         );
 
-        let (dark, bright) = if visuals.dark_mode {
-            (0.2, 0.3)
-        } else {
-            (0.8, 0.9)
-        };
+        let (dark, bright) = (0.8, 1.0);
         let color_factor = if animate {
             ui.ctx().request_repaint();
             lerp(dark..=bright, ui.input().time.cos().abs())
@@ -87,7 +83,7 @@ impl Widget for ProgressBar {
         ui.painter().rect(
             inner_rect,
             corner_radius,
-            Color32::from_gray((color_factor * 255.0) as u8),
+            Color32::from(Rgba::from(visuals.selection.bg_fill) * color_factor as f32),
             Stroke::none(),
         );
 
@@ -119,7 +115,9 @@ impl Widget for ProgressBar {
                 Align2::LEFT_CENTER,
                 text,
                 TextStyle::Button,
-                visuals.text_color(),
+                visuals
+                    .override_text_color
+                    .unwrap_or(visuals.selection.stroke.color),
             );
         }
 
```

Also, shouldn't the widget be included in `egui::widgets` and not just the demo lib?

EmbersArc:
@parasyte Much better! I applied the suggestions.

egui/src/widgets/progress_bar.rs
```diff
@@ -0,0 +1,124 @@
...
+/// A simple progress bar.
+pub struct ProgressBar {
+    progress: f32,
+    desired_width: Option<f32>,
+    text: String,
+    animate: bool,
+}
+
+impl ProgressBar {
+    pub fn new(progress: f32) -> Self {
```
emilk, line 12:
```suggestion
    /// Progress in the `[0, 1]` range, where `1` means "completed".
    pub fn new(progress: f32) -> Self {
```

egui/src/widgets/progress_bar.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /// A custom text to display on the progress bar.
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn text(mut self, text: impl ToString) -> Self {
+        self.text = text.to_string();
+        self
+    }
+
+    /// Whether to display a loading animation. Note that this require the UI to be redrawn.
```
emilk, line 34:
```suggestion
    /// Whether to display a loading animation when progress `< 1`.
    /// Note that this require the UI to be redrawn.
```

egui/src/widgets/progress_bar.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// Defaults to `false`.
```
emilk, line 35: It is an interesting question what the default should be here.

It wastes some CPU in reactive mode using `eframe`, but basically anywhere else it is nice with an animation.

In the future we may want to have a global egui option for "do you want animations or not?", but for now... either `false` or `true` work for me to be honest.

egui/src/widgets/progress_bar.rs
```diff
@@ -0,0 +1,124 @@
...
+            text,
+            mut animate,
+        } = self;
+
+        animate &= progress < 1.0;
+
+        let desired_width = desired_width.unwrap_or(ui.available_size_before_wrap().x);
+        let height = ui.spacing().interact_size.y;
+        let (outer_rect, response) =
+            ui.allocate_exact_size(vec2(desired_width, height), Sense::click());
```
emilk, line 56:
```suggestion
            ui.allocate_exact_size(vec2(desired_width, height), Sense::hover());
```

Something that responds to clicks should do on-hover highlights etc.

Clicking progress bars seems like a rare thing to want to do (I know you do it in the widget gallery though), but if we want to support it I think we should do so fully with a `fn sense(mut self, sense: Sense)` in the builder and then using [`Style::interact`](https://docs.rs/egui/0.13.0/egui/style/struct.Style.html#method.interact) to get the right colors on mouse-hover etc.

egui/src/widgets/progress_bar.rs
```diff
@@ -0,0 +1,124 @@
...
+        if !text.is_empty() {
+            ui.painter().text(
+                outer_rect.left_center() + vec2(ui.spacing().item_spacing.x, 0.0),
+                Align2::LEFT_CENTER,
+                text,
+                TextStyle::Button,
+                visuals
+                    .override_text_color
+                    .unwrap_or(visuals.selection.stroke.color),
+            );
```
emilk, line 119: This can easily overflow the size of the progress bar, which would look pretty ugly.

I think we either should either:
* measure the text (https://docs.rs/egui/0.13.0/egui/struct.Painter.html#method.layout_no_wrap) and use it as a minimum width of the widget
* clip the text (https://docs.rs/egui/0.13.0/egui/struct.Painter.html#method.sub_region)

egui_demo_lib/src/apps/demo/widget_gallery.rs
```diff
@@ -157,6 +160,19 @@ impl WidgetGallery {
         ui.add(egui::Slider::new(scalar, 0.0..=360.0).suffix("°"));
         ui.end_row();
 
+        ui.add(doc_link_label("ProgressBar", "ProgressBar"));
+        let progress = *scalar / 360.0;
+        let progress_bar = egui::ProgressBar::new(progress)
+            .text(format!("{}%", (progress * 100.0) as usize))
```
emilk, line 166: This seems like something so common that we should have a helper for it, e.g. `.show_percentage()`

emilk (changes requested):
This looks really good!

egui_demo_lib/src/apps/demo/widget_gallery.rs
EmbersArc: Added in 1d0d86f

egui/src/widgets/progress_bar.rs
EmbersArc: Good point, the text now gets clipped in 1d0d86f.

egui/src/widgets/progress_bar.rs
EmbersArc: The animation now plays on hover, I'll leave it up to you to decide whether it should be animated by default.

emilk (approved):
Wonderful - thank you!

PR #522, opened 2021-06-26
Fix a bug on Windows where minimizing adjusts all of the egui window positions.

### LINKED ISSUE #518: Windows move to top left when restoring from minimized state.
When running the demo app with a bunch of windows open, if you minimize the application and restore it, all windows will have their positions moved to the top left.

Is there any way to prevent this behavior?

before minimizing:
before
after minimizing:
after

parasyte:
- Closes #518
- This bug is caused by an issue in winit where minimized windows will be given 0 width and height on Windows.
- See: https://github.com/rust-windowing/winit/issues/208
- See also: https://github.com/hasenbanck/egui_winit_platform/pull/19

### DIFF
--- a/egui_glium/CHANGELOG.md
+++ b/egui_glium/CHANGELOG.md
@@ -5,6 +5,9 @@ All notable changes to the `egui_glium` integration will be noted in this file.
 
 ## Unreleased
 
+### Fixed 🐛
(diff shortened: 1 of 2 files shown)


emilk (approved):
Very nice!

PR #531, opened 2021-06-29
TextEdit: Add visual clipping for singleline inputs when text is large.

### LINKED ISSUE #97: [Feature request] Fixed TextEdit width
https://user-images.githubusercontent.com/35241085/103775789-51141680-500d-11eb-9005-660db51083ab.mp4

I want to set the text input to a fixed width so it does not change the `SidePanel` width.

lucaspoffo:
Wow,

This PR adds visual clipping for singleline inputs. Currently the input grows with the text, now it will behave like the browser input, will keep the same size and show the text based on the cursor position.

Peek 2021-06-29 01-59

Fixes #97

### DIFF
--- a/egui/src/widgets/text_edit.rs
+++ b/egui/src/widgets/text_edit.rs
@@ -17,7 +17,7 @@ pub(crate) struct State {
 
     // Visual offset when editing singleline text bigger than the width.
     #[cfg_attr(feature = "persistence", serde(skip))]
-    singleline_offset: Option<f32>,
+    singleline_offset: f32,
 }
 
 #[derive(Clone, Copy, Debug, Default)]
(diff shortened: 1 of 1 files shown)


lucaspoffo:
@emilk There is some problems with the multiline input, but the singleline seems to work fine. I will try to fixes the problems another time, maybe tomorroy. But I want to see what you think about the solution I used

lucaspoffo:
Managed to fix most the issues I was having:

- Multiline text input now behaves normally
- Text selection now considers the offset
- Simplified the logic 

Peek 2021-06-29 15-36

emilk (changes requested):
Very nice, works as expected!

I left some suggestions

lucaspoffo:
Thanks for the suggestions :+1: 
Already made the changes

emilk (approved):
Thank you!

Issue #536, opened 2021-07-01
Merge Pos2 and Vec2? Or use linear algebra library?


mankinskin:
Is there a difference between Pos2 and Vec2? It seems like they are basically interchangable.

Also what about using a linear algebra library like nalgebra or cgmath for vector math? This would also give us matrices and all the related implementations immediately.

emilk:
Points and vectors are different things (https://docs.rs/emath/0.13.0/emath/struct.Vec2.html). Some libraries chose to represent them using the same primitive, but `emath` does not. This helps catch some bugs, like the mixing up the parameters to `Rect::from_min_max` and `Rect::from_min_size`.

egui doesn't really need matrices, so right now I'm opting for a more minimalistic and custom approach. `emath` is opinionated and custom made for a GUI library (i.e. calling the max Y edge of a rect "bottom", which nalgebra or cgmath would not, and should not, do)

mankinskin:
Understandable. In this case what do you think about `From` conversion implementations between them? `Pos2` has `.to_vec2()` but `Vec2` does not have the reverse. I think `.to_pos2()` would be consistent and useful, aswell as `From` implementations. It should be a small addition so I can quickly open a PR.

PR #537, opened 2021-07-01
Add conversion functions between Vec2 and Pos2

mankinskin:
#536

### DIFF
--- a/emath/src/vec2.rs
+++ b/emath/src/vec2.rs
@@ -1,5 +1,7 @@
 use std::ops::{Add, AddAssign, Div, Mul, MulAssign, Neg, Sub, SubAssign};
 
+use crate::pos2::{Pos2, pos2};
+
 /// A vector has a direction and length.
 /// A [`Vec2`] is often used to represent a size.
 ///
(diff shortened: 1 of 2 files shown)


emath/src/vec2.rs
```diff
@@ -80,7 +82,18 @@ impl From<&Vec2> for (f32, f32) {
...
+// Compatibility and convenience conversions to and from Pos2:
+impl From<Pos2> for Vec2 {
+    fn from(pos: Pos2) -> Self {
+        vec2(pos.x, pos.y)
+    }
+}
+impl From<Vec2> for Pos2 {
+    fn from(pos: Vec2) -> Self {
+        pos2(pos.x, pos.y)
+    }
```
emilk, line 95: I'm not sure this is a good idea. There are many places where we take `impl Into<Vec>` as an argument for e.g. sizes, and passing in a position there is probably a mistake, and should not compile. I think it is better if we keep conversations between positions and vectors explicit

emilk (changes requested):

emath/src/vec2.rs
mankinskin: sure thing

emilk (approved):

emilk:
Fixed with a smaller diff in 12334adddaabc19f727750a178c241cba5b8f9d7

PR #543, opened 2021-07-03
Context menu

mankinskin:
- `ContextMenuSystem`: Stored in an App, listens for mouse clicks in `App::update` and creates a context menu at pointer on secondary mouse click.
- context menu contents are defined by builder closure, taking a mutable reference to the menu's persistent submenu state:
```rust
fn update(&mut self, ctx: &CtxRef, frame: &mut Frame) {
    self.context_menu.listen(ctx, |ui, menu_state| {
        ...
    });
}
```
- `SubMenu` creates a managed nested sub-menu which is expanded when it's button is hovered in the parent:
```rust
fn update(&mut self, ctx: &CtxRef, frame: &mut Frame) {
    self.context_menu.listen(ctx, |ui, menu_state| {
        ...
        SubMenu::new("SubMenu").show(ui, menu_state, |ui, menu_state| {
            ...
            SubMenu::new("SubMenu").show(ui, menu_state, |ui, menu_state| {
                ...
            });
            ...
        });
        ...
    });
}
```
- `menu_state` can be used to interact with the menu hierarchy, for example for closing the context menu when a button is clicked:
```rust
fn update(&mut self, ctx: &CtxRef, frame: &mut Frame) {
    self.context_menu.listen(ctx, |ui, menu_state| {
        if ui.button("Open...").clicked() {
            // Open some file or window...
            menu_state.close(); // close context menu
        }
    });
}
```

Peek 2021-07-03 16-52

To-Do:

- [ ] respect click context (i.e. clicked element) in builder
- [ ] improve submenu navigation with [more forgiving mouse movement paths](https://css-tricks.com/dropdown-menus-with-more-forgiving-mouse-movement-paths/#get-fancy-with-a-javascript-built-triangle)
- [ ] improve styling

Let me know if you have any suggestions or find any problems :)


emilk:
Interesting! Have you considered how this relates to `egui::menu`? It would be great if all menus could be unified, as there are a lot of similar things to solve (submenus, when to close the main menu, etc).

mankinskin:
> Have you considered how this relates to `egui::menu`?

Yes, I drew inspiration from it, as this was pretty much my first experience with egui, and now they both render the contents inside a `Frame` inside an `Area`, so this part could probably be unified. I would like to keep things separate for now though, until the functionality is complete and we can work on the styling.

I am now trying to make `ctx.context_menu(add_contents)` and `ui.context_menu(add_contents)` work and respect the clicked area to show context related context menus.

mankinskin:
I am pretty happy so far, but it doesn't work perfectly yet. It is now possible to define a context menu for an individual `Ui` using
```rust
ui.context_menu(|ui, menu_state| {
    if ui.button("Add...").clicked() {
        self.add();
        menu_state.close();
    }
});
```
for example. The `ContextMenuSystem` is now stored by the `egui::Context` and uses `Ui` `Id`'s to manage which context menu to show. Basically a context menu will only be shown when a proper click happens in it's `max_rect_finite()`. I am not sure if this is the best Rect to use, but it was the one that worked best so far. I added some examples to the demo:

- top bar
- backend panel
- widget gallery plot
- drag and drop left column

Peek 2021-07-07 00-38


Unfortunately I was not able to attach context menu's to the items in the drag and drop example properly. When right clicking them I only ever get the context menu for the top left item of all columns, and it mostly closes immediately and only stays open a few of the times I click. I suspect it is something with the bounding boxes of the items, where I am probably not using an accurate one to detect clicks on the respective `Ui`.

@emilk do you maybe have a suggestion which Rect I should use to detect clicks on the content of a `Ui`? I am currently using `ui.rect_contains_pointer(ui.max_rect_finite())`.

mankinskin:
Problem solved, the context menu is now applied to Responses, which hold the accurate rect of the relevant Ui and their Id. Next step is to improve submenu navigation.

parasyte:
Some nitpicks from a UX point of view:

The `menu_state` callbacks look unusual and creates flow control that is difficult to reason about. I'm thinking it would make more sense to have the `add_content()` closure return an enum that will be actionable by the caller:

```rust
ui.add(example_plot(plot))
    .context_menu(|ui| {
        if ui.button("Sin").clicked() {
            *plot = Plot::Sin;
            MenuState::Close
        } else if ui.button("Bell").clicked() {
            *plot = Plot::Bell;
            MenuState::Close
        } else if ui.button("Sigmoid").clicked() {
            *plot = Plot::Sigmoid;
            MenuState::Close
        } else {
            MenuState::Continue
        }
    });
```

It could also be an `Option<MenuState>`, so the conditions can return `Some(MenuState::Close)` or `None` when there is no action to perform. This is a matter of personal preference.

Going over the code, there is also a public callback named `get_submenu` which returns a value. I'm not entirely sure what this is for. Shouldn't sub-menus share all (most?) of the same code as the main context menu? There is probably a good reason for this separation, I'm just unfamiliar with the evolution and design of this PR.

mankinskin:
Ah, `MenuState::get_submenu` shouldn't be public, thanks for pointing out. the root menu and sub menus are basically the same, only that the root also has to store the `Id` of the element it was attached to. This is to discriminate two different context menus.

> return an enum

I thought about this too but I kind of liked the state argument better because it felt more flexible. I am not sure how this feature is going to evolve and if there will be other functions you might want to call from the context menu ui. If you wanted to make multiple calls this would probably be more ergonomic.

But I don't see any reason not to return enums by default and have something like `context_menu_with_state` in addition to that if you ever need to manipulate menu state more extensively.

mankinskin:
The submenu navigation works a lot better now, the code checks if the pointer is travelling towards an open submenu and allows it to cross over other items if it is. It also requests a repaint if it is travelling towards a submenu, so that when the pointer stops moving, this can be detected even in reactive mode.

Peek 2021-07-08 16-18

mankinskin:
@parasyte I also just noticed that the menu_state is needed to create SubMenus, as they use the parent state to know if they are open or not.

parasyte:
Isn't that a concern that is internal to the implementation? I'm not sure the caller should be aware of or have access to that information.

mankinskin:
The `MenuState` only allows creating submenus and closing the menu hierarchy. Everything else is private. I don't see another way letting users create submenus in a closure than passing them a handle to the state they need to interact with. Anything else would require probably a much tighter API that would have to mimic `Ui` to manipulate `MenuState` behind the scenes..
The `MenuState` is completely opaque and only exposes `close()` and `submenu()`. I don't see a problem with this.

mankinskin:
Making some progress on the styling, but there seems to be an issue with overlap and hovering of the submenu button. My guess is that the submenu's frame margin or shadow covers a part of the button in the parent menu, and that causes the submenu to close and reopen repeatedly.

Peek 2021-07-09 01-47

I would be very glad if someone could take a look at the styling code and share some thoughts about it. I feel like I am making this harder than it should be.

mankinskin:
I think I will also look into implementing submenus in the existing menus this way. It will pretty much work the same way.

egui/src/context_menu.rs
```diff
@@ -0,0 +1,415 @@
...
+                context_menu.rect = response.rect;
+
+                if context_menu.response.is_close() {
+                    return MenuResponse::Close;
+                }
+            }
+        }
+        MenuResponse::Stay
+    }
+    /// should be called from Context on a Response
```
emilk, line 58:
```suggestion
    
    /// Show a menu if right-clicked.
    /// Should be called from [`Context`] on a [`Response`].
```

egui/src/ui.rs
```diff
@@ -49,7 +49,7 @@ pub struct Ui {
     style: std::sync::Arc<Style>,
 
     /// Handles the `Ui` size and the placement of new widgets.
-    placer: Placer,
```
emilk, line 52: This seems like an unnecessary change?

egui_demo_lib/src/apps/demo/drag_and_drop.rs
```diff
@@ -137,6 +145,15 @@ impl super::View for DragAndDropDemo {
                 })
                 .response;
 
+                if col_idx == 0 {
+                    response.context_menu(|ui, menu_state| {
+                        if menu_state.item("New Item...").show(ui).clicked() {
```
emilk, line 150:
```suggestion
                        if menu_state.item("New Item").show(ui).clicked() {
```

egui_demo_lib/src/apps/demo/drag_and_drop.rs
```diff
@@ -116,18 +119,23 @@ impl super::View for DragAndDropDemo {
...
                     ui.set_min_size(vec2(64.0, 100.0));
-
-                    for (row_idx, &item) in column.iter().enumerate() {
+                    for (row_idx, item) in column.iter().enumerate() {
                         let item_id = Id::new("item").with(col_idx).with(row_idx);
                         drag_source(ui, item_id, |ui| {
-                            ui.label(item);
+                            ui.add(Label::new(item).sense(Sense::click())).context_menu(
+                                |ui, menu_state| {
+                                    if menu_state.item("Remove...").show(ui).clicked() {
```
emilk, line 133:
```suggestion
                                    if menu_state.item("Remove").show(ui).clicked() {
```

An ellipsis (…) is used to signify that additional information will be collected, which is not the case here.

emilk (changes requested):
This is very impressive work, but I have a concern about having to write `menu_state.item("Click me").show(ui).clicked()`. I would prefer if one could simply write `ui.button("Click me").clicked()`, and add any widget at all with just `ui.add(…)` as anywhere else in egui. I understand we need something special for submenus, but I think we should be able to get other widgets to work.

The context menus should support all the widgets egui does, and should do so using the same interface.

mankinskin:
@emilk Thanks for reviewing this! I agree about the API, that should all be hidden in `Ui`. We do need some way to create a submenu and close the menu though, and they need access to some state for the context menu. If this state was managed internally, we would have this interface return `Option`s or `Result`s when the Ui is not of a submenu, i.e. there is no internal menu state. I think this would be kind of unsound and a potential risk, because people would have to call `.unwrap()` on something like `ui.submenu(..)` when there is really no reason there wouldn't be a menu state when they are inside a context_menu builder. So I am thinking about an API like this
```rust
.response
.context_menu(ui, |ui, menu_state| {
    if ui.menu_item("Text").clicked() {
        menu_state.close() 
    }
    ui.sub_menu("Text", &mut menu_state, |ui, menu_state| {
        if ui.menu_item("Nested item").clicked() {
            menu_state.close()
        }
    })
})
```
The benefit here is that there is always a menu_state when there should be and no `unwrap` calls are needed. also the `menu_item` and `sub_menu` builders would fit with the other widget builders in `Ui`. The `menu_item` builder would be needed to have specialized visuals for menu items, instead of using the regular `button` widget, which for example doesn't respect if a submenu is open and the cursor is moving towards it, i.e. there shouldn't be any visuals on other items even when they are hovered shortly. Maybe the API would even benefit from some getters for these conditions so people can integrate their own widgets more easily.

I will try to integrate your suggestions soon, so we can move on with this :)

coderedart:
don't know if this is the right pace for it, but i am waiting on egui menus/submenus for my app. as my app is a rewrite/port of another app that had custom ui framework, i read upon a couple issues they had with menus. thought i should mention them here for consideration. 
1. What happens if menu is too big, like bigger than the height of window/monitor, are the menus beyond the border inaccessible? or will the menu have scrolling? or will there be a new column to show the rest of the items.
2. is it possible to have checkboxes in menu items, and will the menu disappear when clicked on the checkbox to toggle it, so if the user wants to toggle multiple checkboxes, does he/she need to open menu, toggle item, and repeat ? the use case for it is when there's options that are toggleable and when enabled those options themselves have suboptions. 
the following screenshot should explain what i'm trying to say. 
image

mankinskin:
Update:
- applied review suggestions
- add demo window to demo app
- implemented `MenuUi` for hiding menu state

https://github.com/mankinskin/egui/blob/1e9e79df84a243b83d7f18e7060df519c7c85e7c/egui/src/context_menu.rs#L435-L440

New syntax:
```rust
.context_menu(|ui: &mut MenuUi<'_, '_>| {

    // common widgets
    ui.horizontal(|ui| {
        ui.add(
            egui::DragValue::new(&mut self.width)
                .speed(1.0)
                .prefix("Width:"),
        );
        ui.add(
            egui::DragValue::new(&mut self.height)
                .speed(1.0)
                .prefix("Height:"),
        );
    });

    // submenu
    ui.submenu("Plot", |ui| 
        if ui.radio_value(&mut self.plot, Plot::Sin, "Sin").clicked()
            || ui
                .radio_value(&mut self.plot, Plot::Bell, "Gaussian")
                .clicked()
            || ui
                .radio_value(&mut self.plot, Plot::Sigmoid, "Sigmoid")
                .clicked()
        {
            ui.close_menu(); // way to close menus
        }
    );

    // doesn't close on interaction
    ui.horizontal(|ui| {
        ui.checkbox(&mut self.show_axes[0], "x-Axis");
        ui.checkbox(&mut self.show_axes[1], "y-Axis");
    });
    
    // Problem: accessing menu state from inside common widgets requires this
    let menu_state = &mut *ui.menu_state;
    let menu_ui = &mut *ui.ui;
    menu_ui.horizontal(|ui| {
        if ui.checkbox(&mut self.allow_drag, "Drag").changed()
            || ui.checkbox(&mut self.allow_zoom, "Zoom").changed()
        {
            menu_state.close();
        }
    });
});
```
Peek 2021-08-28 11-45

This approach wraps the mutable references to the `Ui` and `MenuState` into a single type, which then implements `Deref` to `&mut Ui`, so it can be used just like a `Ui`, but can also have its own interface to `MenuState`, for creating submenus and closing the menu.
The problem is when we create a child ui using `MenuUi`, the child ui will just be a regular `Ui` and the `MenuUi` variable will be shadowed, unless we use a different name, which we obviously don't want. The `MenuUi` should behave just like a `Ui` but its child uis also need to have access to its menu state. Submenus are a special case, because they carry their own menu state, but that is handled by `MenuUi::submenu`.

My proposal would be to add a `trait Ui` which implements the common `Ui` interface. That would allow us to implement `Ui::child_ui() -> Self::ChildUi` differently for this case, so `ui.horizontal(|ui| {})` would create a `MenuUi` referencing the same `MenuState`. This would mean we should change `struct Ui` to something like `struct BaseUi` and expect `ui: &mut impl Ui` anywhere `&mut Ui` is expected now. Then `BaseUi` and `MenuUi` would both implement `Ui`.

The problem here is that `trait Ui` would need an associated `type ChildUi: Ui` which might complicate things, but maybe not.

emilk:
`MenuUi` is clever, but it may be a bit too clever, as it is trying to create an abstraction that immediately leaks (i.e. need to extract `menu_state` to call `.close()` from an inner closure, as you point out). I'm also not sure how well it works with e.g. code completion using rust analyzer. `trait Ui` is a huge complication which I'd rather avoid. Keep it simple!

I prefer the old `|ui, menu_state|` approach (no need to decompose to pass `menu_state` into child closures). What was the problem with that approach?

An alternative is to let each `Ui` carry an `menu_state: Option<Arc<MenuState>>`. This is a bit more ergonomic: `ui.close()` would close the menu we are in, if any. `ui.menu(…)` could open either a top-level menu or a sub-menu, depending on whether `ui` already has a `menu_state`. I think we can hold off on that to a later time though if the `|ui, menu_state|` approach works instead (it's a good enough start).

emilk:
> items don't adjust to width of larger items when they are added after them

This is a general problem of immediate mode. It has two solutions:

A) the user calls `ui.set_width(…)` at the start (workaround)
B) the implementation stores the size from the last frame and use that at the start of the next frmae

The problem with B is that the menu will then never shrink again. It is also a bit more complicated.

I think finding a solution to this can wait until a later PR.

---

* normal widgets still show hover visuals when cursor moves over them while submenu is open
* sometimes context menus applied to panels trigger when clicking on a window covering the panel

[I just fixed this bug](https://github.com/emilk/egui/commit/56913a9ae9e97dc2e9de7a7e6098b7620642cb17#diff-f4110641005e72cd5fd0d6a673d866c5219a04b04999f7b388befa5e70f19932)! If you rebase on master it should be gone now.

---

> large menus should wrap into multiple columns or become scrollable (should be configurable)

Yes, but we can wait with that for another PR. We don't need to get everything perfect on the first go!

emilk:
Oh and also: the menus work really well now! I'm looking forward to getting this merged once we have settled on the API interface (`|ui|` vs `|ui, menu_state|`).

mankinskin:
> An alternative is to let each `Ui` carry an `menu_state: Option<Arc<MenuState>>`. This is a bit more ergonomic: `ui.close()` would close the menu we are in, if any. `ui.menu(…)` could open either a top-level menu or a sub-menu, depending on whether `ui` already has a `menu_state`. I think we can hold off on that to a later time though if the `|ui, menu_state|` approach works instead (it's a good enough start).

yes, I also like this approach better. Haven't even thought about that without needing to unwrap.

mankinskin:
As far as I am concerned the feature set of this PR is complete. There are some known improvements left, but we will create new Issues and PRs for these.

egui/src/menu.rs
```diff
@@ -59,46 +72,464 @@ pub fn bar<R>(ui: &mut Ui, add_contents: impl FnOnce(&mut Ui) -> R) -> InnerResp
...
     menu_id: impl std::hash::Hash,
-    pos: Pos2,
+    menu_state_arc: Arc<RwLock<MenuState>>,
     mut style: Style,
     add_contents: impl FnOnce(&mut Ui) -> R + 'c,
 ) -> InnerResponse<R> {
+    let mut menu_state = menu_state_arc.write().unwrap();
+    menu_state.entry_count = 0;
+    let pos = menu_state.rect.min;
+    drop(menu_state);
```
emilk, line 508: Alternative:

```suggestion
    let pos = {
        let mut menu_state = menu_state_arc.write().unwrap();
        menu_state.entry_count = 0;
        menu_state.rect.min
    };
```

egui/src/ui.rs
```diff
@@ -1,10 +1,11 @@
 // #![warn(missing_docs)]
 
 use std::hash::Hash;
+use std::sync::{Arc, RwLock};
```
emilk, line 4: Use `epaint::mutex::RwLock` instead!

egui/src/menu.rs
```diff
@@ -15,25 +15,38 @@
...
 //! ```
 
+use super::{
+    style::{Spacing, WidgetVisuals},
+    Align, CtxRef, Id, InnerResponse, PointerState, Pos2, Rect, Response, Sense, Style, TextStyle,
+    Ui, Vec2,
+};
 use crate::{widgets::*, *};
 use epaint::Stroke;
+use std::sync::{Arc, RwLock};
```
emilk, line 25: Use `epaint::mutex::RwLock` instead!

egui/src/ui.rs
```diff
@@ -1631,6 +1637,28 @@ impl Ui {
         self.advance_cursor_after_rect(Rect::from_min_size(top_left, size));
         result
     }
+    /// Close menu (with submenus), if any.
+    pub fn close(&mut self) {
```
emilk, line 1641: I think we should call this `close_menu`, as we may add `close_popup` etc later

egui/src/ui.rs
```diff
@@ -1631,6 +1637,28 @@ impl Ui {
...
+        if let Some(menu_state) = &mut self.menu_state {
+            menu_state.write().unwrap().close();
+        }
+        self.menu_state = None;
+    }
+    pub(crate) fn set_menu_state(&mut self, menu_state: Arc<RwLock<MenuState>>) {
+        self.menu_state = Some(menu_state);
+    }
+    #[inline(always)]
+    pub fn menu<R>(
```
emilk, line 1651: This needs a docstring and a small example!

Also I think `menu_button` is a better name.

egui_demo_lib/src/apps/demo/context_menu.rs
```diff
@@ -0,0 +1,199 @@
...
+                            ui.close();
+                        }
+                    });
+                });
+        });
+        ui.separator();
+        ui.horizontal(|ui| {
+            ui.button("Nested context menu")
+                .context_menu(Self::nested_menus);
+            ui.menu("Top level menu", Self::nested_menus);
```
emilk, line 193:
```suggestion
            ui.menu("Click for menu", Self::nested_menus);
            ui.button("Right-click for menu")
                .context_menu(Self::nested_menus);
```

egui_demo_lib/src/apps/demo/context_menu.rs
```diff
@@ -0,0 +1,199 @@
...
+        });
+        ui.separator();
+        ui.label("Right click to edit items");
+        self.drag_and_drop.ui(ui);
```
emilk, line 197: No need for the "rename window" and the "drag-and-drop" demos imho. The plot and the two buttons is enough for a demo I think!

egui_demo_lib/src/apps/demo/context_menu.rs
```diff
@@ -0,0 +1,199 @@
...
+                    });
+                    ui.horizontal(|ui| {
+                        if ui.checkbox(&mut self.allow_drag, "Drag").changed()
+                            || ui.checkbox(&mut self.allow_zoom, "Zoom").changed()
+                        {
+                            ui.close();
+                        }
+                    });
+                });
+        });
```
emilk, line 188:
```suggestion
        });
        ui.label("Right-click plot to edit it!");
```

egui_demo_lib/src/apps/demo/context_menu.rs
```diff
@@ -0,0 +1,199 @@
...
+                        ui.checkbox(&mut self.show_axes[0], "x-Axis");
+                        ui.checkbox(&mut self.show_axes[1], "y-Axis");
+                    });
+                    ui.horizontal(|ui| {
+                        if ui.checkbox(&mut self.allow_drag, "Drag").changed()
+                            || ui.checkbox(&mut self.allow_zoom, "Zoom").changed()
+                        {
+                            ui.close();
+                        }
+                    });
```
emilk, line 186:
```suggestion
                    egui::Grid::new("button_grid").show(ui, |ui| {
                        ui.checkbox(&mut self.show_axes[0], "x-Axis");
                        ui.checkbox(&mut self.show_axes[1], "y-Axis");
                        ui.end_row();
                        if ui.checkbox(&mut self.allow_drag, "Drag").changed()
                            || ui.checkbox(&mut self.allow_zoom, "Zoom").changed()
                        {
                            ui.close();
                        }
                    });
```

egui_demo_lib/src/apps/demo/demo_app_windows.rs
```diff
@@ -27,6 +27,7 @@ impl Default for Demos {
             Box::new(super::sliders::Sliders::default()),
             Box::new(super::widget_gallery::WidgetGallery::default()),
             Box::new(super::window_options::WindowOptions::default()),
+            Box::new(super::context_menu::ContextMenus::default()),
```
emilk, line 30: Move it to the top. The demos are lexicographically sorted. Also it is a nice demo! ;)

egui_demo_lib/src/apps/demo/drag_and_drop.rs
```diff
@@ -75,24 +75,41 @@ pub fn drop_target<R>(
...
 }
-
+#[derive(Clone, PartialEq)]
+#[cfg_attr(feature = "persistence", derive(serde::Deserialize, serde::Serialize))]
 pub struct DragAndDropDemo {
     /// columns with items
-    columns: Vec<Vec<&'static str>>,
+    columns: Vec<Vec<String>>,
+    /// edit using context menu
+    editable: bool,
```
emilk, line 84: Why not always have it editable?

egui_demo_lib/src/apps/demo/drag_and_drop.rs
```diff
@@ -116,18 +133,24 @@ impl super::View for DragAndDropDemo {
...
+            for (col_idx, column) in self.columns.clone().into_iter().enumerate() {
                 let ui = &mut uis[col_idx];
                 let can_accept_what_is_being_dragged = true; // We accept anything being dragged (for now) ¯\_(ツ)_/¯
                 let response = drop_target(ui, can_accept_what_is_being_dragged, |ui| {
                     ui.set_min_size(vec2(64.0, 100.0));
-
-                    for (row_idx, &item) in column.iter().enumerate() {
-                        let item_id = Id::new("item").with(col_idx).with(row_idx);
+                    for (row_idx, item) in column.iter().enumerate() {
+                        let item_id = Id::new(self.id_source).with(col_idx).with(row_idx);
```
emilk, line 143: Using a single integer as the top-level id source is not idiomatic egui use.

For a demo I think it would suffice with `let id_source = "my_drag_and_drop_demo";` in this function, and then we can remove `Sefl::id_source` and `COUNTER`

emilk (changes requested):
This is amazing work, good job ⭐⭐⭐ 

I'm sorry for keeping you waiting so long for a review.

Overall everything looks great, but there is some things I'd like to get cleared with better naming and/or docstrings, + some minor nits and tweaks.

emilk:
Is there something I can do to help out with this?

If you don't feel you have the time to finish it up, maybe I can take over (though not for another week or so)

mankinskin:
@emilk I have been on a vacation for the past weeks, but I will pick this up in the next few days. Sorry for the delay. If you already have some improvements done, you could probably open a pull request on my branch, but I will resolve this review in the next few days.

emilk:
No worries and no stress @mankinskin, we all need vacation! I'm just excited for this PR!

crumblingstatue:
I just want to say that I'm excited about this too, but take as much time as you need!

egui/src/menu.rs
```diff
@@ -59,46 +72,464 @@ pub fn bar<R>(ui: &mut Ui, add_contents: impl FnOnce(&mut Ui) -> R) -> InnerResp
...
+    Close,
+    Stay,
+    Create(Pos2, Id),
+}
+impl MenuResponse {
+    pub fn is_close(&self) -> bool {
+        *self == Self::Close
+    }
+}
+pub(crate) enum EntryState {
```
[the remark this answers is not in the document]
mankinskin: This was meant to disable hover visuals of menu items when a submenu was open. but because the menu ui may contain any kinds of widgets, this would need a different approach which fixes hover visuals for all widgets when a submenu is open.. The point of this is to prevent hovering highlights when the cursor crosses items while moving to an opened submenu (which may cross the parent menu without closing the submenu).

egui/src/menu.rs
```diff
@@ -59,46 +72,464 @@ pub fn bar<R>(ui: &mut Ui, add_contents: impl FnOnce(&mut Ui) -> R) -> InnerResp
...
+        }
+    }
+    pub fn show<R>(
+        &mut self,
+        ctx: &CtxRef,
+        add_contents: impl FnOnce(&mut Ui) -> R,
+    ) -> InnerResponse<R> {
+        MenuState::show(ctx, &self.menu_state, self.id, add_contents)
+    }
+    fn stationary_interaction(
```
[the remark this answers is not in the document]
mankinskin: A stationary menu is a menu that is not a context menu, basically. It has a button for the top level menu, unlike a context menu, but kind of like a submenu. The menu position is opened below the button though, whereas submenus are opened right of the button.

PR #550, opened 2021-07-07
Fix hover sensing

mankinskin:
Hovers were ignored when sense.click was not enabled. This was due to a piece of code which only allows hover in a response that had also been clicked.

### DIFF
--- a/egui/src/context.rs
+++ b/egui/src/context.rs
@@ -281,10 +281,6 @@ impl CtxRef {
             response.interact_pointer_pos = self.input().pointer.interact_pos();
         }
 
-        if self.input.pointer.any_down() {
-            response.hovered &= response.is_pointer_button_down_on; // we don't hover widgets while interacting with *other* widgets
-        }
-
         if response.has_focus() && response.clicked_elsewhere() {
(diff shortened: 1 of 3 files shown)


emilk:
> Hovers were ignored when sense.click was not enabled.

You can hover things without `sense.click`. For instance: a label can have a tooltip, and it will show it because it detect hovers properly.

Please explain the problem you are attempting to solve in an issue first.

The code you remove does what it is supposed to do (afaict) which is to prevent e.g. tooltips for other widgets to come up when dragging e.g. a slider.

mankinskin:
I think I just realized that hover is not active when dragging.
The issue was that `Plot` (equipped with `Sense::drag`) did not indicate a hover when a mouse button was down. This is how I detect a click on a widget currently, because `response.clicked_by` never really sensed clicks. This may be because the relevant areas were not equipped with `Sense::click`. Now I tried sensing the clicks from the response with `Response::interact` but this doesn't seem to work either.. that is why I still use the workaround with `response.hovered() && response.clicked_by(...)`, but this doesn't work for draggable widgets... I wonder why `Response::interact` doesn't detect the clicks afterwards..

PR #557, opened 2021-07-11
Pass more inner return values

barrowsys:
Fixed all missed opportunities to pass an inner return value.


emilk (approved):
Very thorough - thanks!

PR #562, opened 2021-07-14
Add methods for optionally displaying the background/axes of a `Plot`

mitchmindtree:
These are particularly useful when using the `Plot` widget as an overlay
over an existing grid or some other content.

---

Here's a screenshot of one of my intended use cases - using a `Plot` tool to visualise signals/patterns over a custom `egui_timeline` widget:

egui-plot-no-bg-or-axes-example

---

PS thanks for the awesome work on this lib! I'm the original author of `conrod`, which I made out of necessity back around Rust 0.9 as there were literally zero other GUI libraries out there at the time :) So nice to see a much more serious and well-thought-out immediate mode GUI lib! I'm blown away at how feature-full it is already. I especially appreciate the ease of integrating into different event loops and graphics backends :ok_hand: It has made it a breeze to integrate into `nannou` projects.

### DIFF
--- a/egui/src/widgets/plot/mod.rs
+++ b/egui/src/widgets/plot/mod.rs
@@ -71,6 +71,8 @@ pub struct Plot {
     show_x: bool,
     show_y: bool,
     legend_config: Option<Legend>,
+    show_background: bool,
+    show_axes: bool,
 }
 
 impl Plot {
@@ -98,6 +100,8 @@ impl Plot {
             show_x: true,
             show_y: true,
             legend_config: None,
+            show_background: true,
+            show_axes: true,
         }
     }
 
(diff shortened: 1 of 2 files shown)


egui/src/widgets/plot/mod.rs
```diff
@@ -314,6 +318,22 @@ impl Plot {
...
+    /// Default: `true`.
+    pub fn show_background(mut self, show: bool) -> Self {
+        self.show_background = show;
+        self
+    }
+
+    /// Show the axes.
+    /// Can be useful to disable if the plot is overlaid over an existing grid or content.
+    /// Default: `true`.
+    pub fn show_axes(mut self, show: bool) -> Self {
```
emilk, line 333: It would be nice with a per-axis control here, either with `show_x_axis` + `show_y_axis` or `show_axes(mut self, show: [bool; 2])`. What do you think?

emilk (changes requested):

emilk:
That's a really cool screenshot! I hope you release that app one day :)

mitchmindtree:
I've rebased onto master and updated the new method to take `[bool; 2]`, allowing the user to specify whether or not to show each axis individually.

emilk (approved):
Thank you!

PR #563, opened 2021-07-14
Fix case where `Plot`'s `min_auto_bounds` can be ignored after first instantiation

mitchmindtree:
I ran into an issue using `Plot` within my timeline widget where if I
zoom in and out of the timeline (not the plot), the `Plot` instances
would ignore the necessary changes to the `include_x` calls and in turn
would become skewed and misaligned with the timeline below.

This changes the `Plot` to check whether or not `min_auto_bounds` have
changed and, if so, reset the memory and recalculate the bounds.

See #562 for an image of my current use case.

### DIFF
--- a/egui/src/widgets/plot/mod.rs
+++ b/egui/src/widgets/plot/mod.rs
@@ -358,8 +358,8 @@ impl Widget for Plot {
                 bounds: min_auto_bounds,
                 auto_bounds: !min_auto_bounds.is_valid(),
                 hovered_entry: None,
-                hidden_items: HashSet::new(),
                 min_auto_bounds,
+                ..memory
             };
             ui.memory().id_data.insert(plot_id, memory.clone());
         }
--- a/egui/src/widgets/plot/mod.rs
+++ b/egui/src/widgets/plot/mod.rs
@@ -28,6 +28,7 @@ struct PlotMemory {
     auto_bounds: bool,
     hovered_entry: Option<String>,
     hidden_items: HashSet<String>,
+    min_auto_bounds: Bounds,
 }
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


egui/src/widgets/plot/mod.rs
```diff
@@ -339,22 +340,36 @@ impl Widget for Plot {
...
             })
             .clone();
 
+        // If the min bounds changed, recalculate everything.
+        if min_auto_bounds != memory.min_auto_bounds {
+            memory = PlotMemory {
+                bounds: min_auto_bounds,
+                auto_bounds: !min_auto_bounds.is_valid(),
+                hovered_entry: None,
+                hidden_items: HashSet::new(),
```
emilk, line 361: I think we should carry over the `hidden_items` at least

egui/src/widgets/plot/mod.rs
```diff
@@ -339,22 +340,36 @@ impl Widget for Plot {
...
                 hovered_entry: None,
                 hidden_items: HashSet::new(),
+                min_auto_bounds,
             })
             .clone();
 
+        // If the min bounds changed, recalculate everything.
+        if min_auto_bounds != memory.min_auto_bounds {
+            memory = PlotMemory {
+                bounds: min_auto_bounds,
```
emilk, line 358: Would this make sense:

```suggestion
                bounds: memory.bounds.merge(min_auto_bounds),
```

That is, if the new `min_auto_bounds` are larger, we zoom out to include them, but otherwise we keep the same bounds.

emilk (changes requested):

egui/src/widgets/plot/mod.rs
mitchmindtree: Hmmmm I can see how this approach might make sense as a solution for the behaviour of the `include_*` methods. However, for our case, we want it so that if we decrease the bounds, it appears as though we are "zooming" into the plot. That is, we need the ability to specify the exact bounds of the visible area of the plot each update - if not via the `include_*` methods, perhaps some other method.

mitchmindtree:
OK I've rebased onto master and ensure that we now carry the `hidden_items`.

emilk (approved):

PR #564, opened 2021-07-14
Fix custom font definitions getting replaced when `pixels_per_point` is changed

parasyte:
Fixes a bug where any custom font definitions are replaced with the default font definitions when `pixels_per_point` is changed.

I noticed that all text would disappear on my app when I changed the UI scale. This happens because I use `default-features = false` to disable the `default_fonts` feature. The default font definitions are completely empty when default fonts are disabled.


emilk (approved):
Good catch!

Issue #417, opened 2021-05-19
How to make a tree with selectable lines?


msklywenn:
**Is your feature request related to a problem? Please describe.**
I'd like to display a tree structure like the one made up of `CollapsibleHeaders` in the example, but I want to be able to select a line in the tree instead of opening/closing on every click. For that I would expect only the arrow icon to handle opening/closing and be able to detect clicks on the rest of the header, maybe even had more stuff in there.

**Describe the solution you'd like**
A solution I think would be to have a `CollapsingHeader::show_with_custom_header` with a lambda that handles the UI in the rectangle on the right of the arrow.

**Describe alternatives you've considered**
I don't have much ideas beside the above proposed solution.

**Additional context**
A screenshot from a customized hierarchy done with Unity:
![image](https://user-images.githubusercontent.com/1546678/118832675-1d050800-b8c1-11eb-89b5-0316edb54001.png)

I could implement the above solution if deemed appropriate.

msklywenn:
I managed to hack something by modifying CollapsingHeader like proposed. I probably broke stuff but for my purpose it works. Let me know if you'd be interested in a PR: https://github.com/msklywenn/egui/commit/afb665adf7fe67af05cc1a94ccd074563996e498 (there's also another commit that makes the rectangle of the original CollapsingHeader optional)

https://user-images.githubusercontent.com/1546678/119153066-3046dd80-ba51-11eb-9426-70df4e79b16c.mp4

tgolsson:
I combined the default non-callback header from CollapsingHeader with @msklywenn's implementation for icon as an expander widget here: https://github.com/tgolsson/egui/commit/25e88ef2b4a34383c4dce608afc8a8d34a75ef1d. I'm hoping to clean it up a bit more, but after that I might PR it back if you'd be OK with that @emilk. It's unfortunately a lot of duplicated code at the moment, but I think that these widgets fill different roles and double-dutying will make it bad for both.

msklywenn:
@tgolsson have you checked the subsequent commit? I made it so the header closure returns the content closure so that if both manipulate the same variable, we don't end up with multiple moves of mutable values. https://github.com/msklywenn/egui/commit/bc0c00cdc9c8c5ff168037b3e37fc84b2daa3769

tgolsson:
I didn't, no! I'll take a peek later today. What is the use-case where a label isn't enough and you need to mutate things in the closure? Renaming things in the hierarchy?

msklywenn:
[...]

PR #585, opened 2021-07-26
Add collapsing header select as selectable label AND Adding possibility to have panels inside UI like Windows

gents83:
I've added possibility to have collapsing header with same functionality of selectable label, without breaking previous functionalities.
This is useful to achieve hierarchies or folder dialog with selection.
Here you can find an example:

egui_content_browser

``` rust
    fn populate_with_folders_tree(ui: &mut Ui, root: &Path, data: &mut FolderDialogData) {
        for_each_folder_in(root, |path| {
            let selected = data.selected_folder == path.to_path_buf();
            if is_folder_empty(path) {
                if ui
                    .selectable_label(selected, path.file_stem().unwrap().to_str().unwrap())
                    .clicked()
                {
                    data.selected_folder = path.to_path_buf();
                }
            } else {
                let collapsing = CollapsingHeader::new(path.file_stem().unwrap().to_str().unwrap())
                    .selectable(true)
                    .selected(selected);
                let header_response = collapsing
                    .show(ui, |ui| {
                        Self::populate_with_folders_tree(ui, path, data);
                    })
                    .header_response;
                if header_response.clicked() {
                    data.selected_folder = path.to_path_buf();
                }
            }
        });
    }
```

Closes <https://github.com/emilk/egui/issues/417>.

### DIFF
--- a/egui/src/containers/collapsing_header.rs
+++ b/egui/src/containers/collapsing_header.rs
@@ -140,6 +140,8 @@ pub struct CollapsingHeader {
     default_open: bool,
     id_source: Id,
     enabled: bool,
+    selectable: bool,
+    selected: bool,
 }
 
 impl CollapsingHeader {
@@ -157,6 +159,8 @@ impl CollapsingHeader {
             default_open: false,
             id_source,
             enabled: true,
+            selectable: false,
+            selected: false,
         }
     }
 
@@ -188,6 +192,16 @@ impl CollapsingHeader {
         self.enabled = enabled;
         self
     }
+
+    pub fn selectable(mut self, selectable: bool) -> Self {
+        self.selectable = selectable;
+        self
+    }
+
+    pub fn selected(mut self, selected: bool) -> Self {
+        self.selected = selected;
+        self
+    }
 }
 
 struct Prepared {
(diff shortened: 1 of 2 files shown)


gents83:
Here you can find an example of what we could obtain with Panels that could be used inside other ui elements.
Once integrated I would be glad to add an example inside demo too, if for you is ok @emilk :)


https://user-images.githubusercontent.com/62186646/128388159-c703e2fd-b120-45b1-8593-9536d290c3f8.mp4

egui/src/containers/collapsing_header.rs
```diff
@@ -188,6 +192,16 @@ impl CollapsingHeader {
         self.enabled = enabled;
         self
     }
+
+    pub fn selectable(mut self, selectable: bool) -> Self {
```
emilk, line 196:
```suggestion
    /// Can the header be selected by clicking it? Default: `false`.
    pub fn selectable(mut self, selectable: bool) -> Self {
```

emilk (changes requested):
Selectable `CollapsingHeader` and UI panels are both great ideas, but they are very different. Please split them up into two PR:s and add some demo code for them so they can be tested!

gents83:
Created 2 separated pull requests and closing this one

PR #593, opened 2021-07-29
egui_glium: Allow running in any thread on windows

PepejHa4ker:


epaint/src/color.rs
```diff
@@ -15,7 +15,7 @@
 /// Alpha channel is in linear space.
 #[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq)]
 #[cfg_attr(feature = "persistence", derive(serde::Deserialize, serde::Serialize))]
-pub struct Color32(pub(crate) [u8; 4]);
```
emilk, line 18: This is an unrelated change - please revert! `Color32` is intentionally opaque to common mistakes less likely.

emilk (changes requested):

Issue #571, opened 2021-07-19
Egui will be blocked after adding custom fonts


Itsusinn:
When using egui 0.13, I get a black output after adding custom fonts.
The whole program will block at CtxRef::begin_frame.

To be precise, it will block at this line https://github.com/emilk/egui/blob/7c5a2d60c5d000f193bc7b16c5a53c059fee8259/egui/src/context.rs#L578

In egui 0.12, there is no problem with [this approach](https://github.com/emilk/egui/discussions/373)

Here is the minimized project to reproduce the error
https://github.com/Itsusinn/custom-font-bug

zu1k:
I found that no deadlocks occurred, just that when the font file was large (including too many characters) it was very slow.

Very slow processing here:

https://github.com/emilk/egui/blob/94937a33e8b8f1f21165854ec5426e84849131f9/epaint/src/text/font.rs#L205-L207

zu1k:
> https://github.com/emilk/egui/blob/94937a33e8b8f1f21165854ec5426e84849131f9/epaint/src/text/font.rs#L206

I'm not sure if it's the `iterator` or the BTreeSet that's causing the problem, perhaps a combination of both factors ?

zu1k:
I have created a PR to fix it, maybe you could have a try.  @Itsusinn

PR #594, opened 2021-07-30
Fix blocking when using custom large font files

zu1k:
Fix blocking on startup when using large font files. 
I use the lazy strategy to delay all font character set calculations.

Closes <https://github.com/emilk/egui/issues/571>.

### DIFF
--- a/epaint/src/text/font.rs
+++ b/epaint/src/text/font.rs
@@ -193,7 +193,7 @@ pub struct Font {
     text_style: TextStyle,
     fonts: Vec<Arc<FontImpl>>,
     /// Lazily calculated.
-    characters: Option<std::collections::BTreeSet<char>>,
+    characters: RwLock<Option<std::collections::BTreeSet<char>>>,
     replacement_glyph: (FontIndex, GlyphInfo),
     pixels_per_point: f32,
     row_height: f32,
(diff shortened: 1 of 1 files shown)


epaint/src/text/font.rs
```diff
@@ -192,7 +192,7 @@ type FontIndex = usize;
 pub struct Font {
     text_style: TextStyle,
     fonts: Vec<Arc<FontImpl>>,
-    characters: std::collections::BTreeSet<char>,
+    characters: Option<std::collections::BTreeSet<char>>,
```
emilk, line 195:
```suggestion
    /// Lazily calculated.
    characters: Option<std::collections::BTreeSet<char>>,
```

emilk (changes requested):

epaint/src/text/font.rs
```diff
@@ -258,8 +254,15 @@ impl Font {
     }
 
     /// All supported characters
-    pub fn characters(&self) -> &BTreeSet<char> {
-        &self.characters
+    pub fn characters(&mut self) -> &BTreeSet<char> {
+        if self.characters.is_none() {
```
emilk, line 258: Put `characters` behind a `epaint::RwLock` so that this function can be `&self` again.

emilk (changes requested):

emilk (approved):

PR #599, opened 2021-07-31
Allow adding Vec<Line> instead of one by one

sigaloid:
Lets me add a vec<line> instead of.. well, without this, there's no way to add lines that aren't known at compile time AFAIK.

### DIFF
--- a/egui/src/widgets/plot/mod.rs
+++ b/egui/src/widgets/plot/mod.rs
@@ -123,14 +123,8 @@ impl Plot {
     }
 
     pub fn lines(mut self, lines: Vec<Line>) -> Self {
-        for mut line in lines{
-            if line.series.is_empty(){
-                return self;
-            }
-            if line.stroke.color == Color32::TRANSPARENT {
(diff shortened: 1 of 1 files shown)


egui/src/widgets/plot/mod.rs
```diff
@@ -122,6 +122,19 @@ impl Plot {
...
+    pub fn lines(mut self, lines: Vec<Line>) -> Self {
+        for mut line in lines{
+            if line.series.is_empty(){
+                return self;
+            }
+            if line.stroke.color == Color32::TRANSPARENT {
+                line.stroke.color = self.auto_color();
+            }
+            self.items.push(Box::new(line));
+        }
```
emilk, line 134: No need for code duplication

```suggestion
        for line in lines {
            self = self.line(line);
        }
```

though I would argue the caller could just do this instead?

emilk (changes requested):

emilk:
I don't think there is a pressing issue for this - if so, please reopen when green

PR #608, opened 2021-08-09
fix and pin web-sys breakage in `Navigator::clipboard`

mental32:
About seven days ago [web-sys] released [0.3.52] that contained [this commit](https://github.com/rustwasm/wasm-bindgen/commit/1e4152913ca90f316a1434b395a58c42d8beb774#diff-7f139578692164f44c3a0ff0806ab67b6a3d40f56d42e8ee17417b68d412679a) that changed the return type of `Navigator::clipboard(&self) -> Clipboard` to an `Option<Clipboard>` leading to a failing build.

This patch skips the `set_clipboard_text` code when the clipboard is `None` and pins the web-sys dependency of egui_web to `0.3.52` (although the pinning might be a little unnecessary.)

I ran this for a successful build: `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo build -p egui_web`
and nothing else seems to have broken so I think it doesn't introduce other breakages.

[web-sys]: https://crates.io/crates/web-sys
[0.3.52]: https://crates.io/crates/web-sys/0.3.52

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7,7 +7,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "af0ac006645f86f20f6c6fa4dcaef920bf803df819123626f9440e35835e7d80"
 dependencies = [
  "ab_glyph_rasterizer",
- "owned_ttf_parser 0.12.0",
+ "owned_ttf_parser 0.12.1",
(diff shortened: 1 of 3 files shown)


emilk (approved):
Nice, thanks!

PR #623, opened 2021-08-16
Add collapsing header select as selectable label

gents83:
I've added possibility to have collapsing header with same functionality of selectable label, without breaking previous functionalities.
This is useful to achieve hierarchies or folder dialog with selection.
Here you can find an example:

egui_content_browser

``` rust
    fn populate_with_folders_tree(ui: &mut Ui, root: &Path, data: &mut FolderDialogData) {
        for_each_folder_in(root, |path| {
            let selected = data.selected_folder == path.to_path_buf();
            if is_folder_empty(path) {
                if ui
                    .selectable_label(selected, path.file_stem().unwrap().to_str().unwrap())
                    .clicked()
                {
                    data.selected_folder = path.to_path_buf();
                }
            } else {
                let collapsing = CollapsingHeader::new(path.file_stem().unwrap().to_str().unwrap())
                    .selectable(true)
                    .selected(selected);
                let header_response = collapsing
                    .show(ui, |ui| {
                        Self::populate_with_folders_tree(ui, path, data);
                    })
                    .header_response;
                if header_response.clicked() {
                    data.selected_folder = path.to_path_buf();
                }
            }
        });
    }
```

Closes <https://github.com/emilk/egui/issues/417>.

I've also modified the Tree example in Misc Demos to show its behavior :)

### DIFF
--- a/egui/src/containers/collapsing_header.rs
+++ b/egui/src/containers/collapsing_header.rs
@@ -223,7 +223,7 @@ impl CollapsingHeader {
 
     /// Should the `CollapsingHeader` show a background behind it? Default: `false`.
     ///
-    /// To show it behind all CollapsingHeaders you can just use:
+    /// To show it behind all `CollapsingHeader` you can just use:
     /// ```
     /// # let ui = &mut egui::Ui::__test();
     /// ui.visuals_mut().collapsing_header_frame = true;
--- a/egui/src/containers/collapsing_header.rs
+++ b/egui/src/containers/collapsing_header.rs
@@ -285,8 +285,9 @@ impl CollapsingHeader {
             });
         }
 
-        let selected = self.selected || header_response.hovered() || header_response.has_focus();
-        if self.selectable && selected {
+        if self.selected
+            || self.selectable && (header_response.hovered() || header_response.has_focus())
+        {
             let rect = rect.expand(visuals.expansion);
 
             let corner_radius = 2.0;
--- a/egui/src/containers/collapsing_header.rs
+++ b/egui/src/containers/collapsing_header.rs
@@ -140,6 +140,8 @@ pub struct CollapsingHeader {
     default_open: bool,
     id_source: Id,
     enabled: bool,
+    selectable: bool,
+    selected: bool,
 }
 
 impl CollapsingHeader {
@@ -157,6 +159,8 @@ impl CollapsingHeader {
             default_open: false,
             id_source,
             enabled: true,
+            selectable: false,
+            selected: false,
         }
     }
 
@@ -188,6 +192,16 @@ impl CollapsingHeader {
         self.enabled = enabled;
         self
     }
+
+    pub fn selectable(mut self, selectable: bool) -> Self {
+        self.selectable = selectable;
+        self
+    }
+
+    pub fn selected(mut self, selected: bool) -> Self {
+        self.selected = selected;
+        self
+    }
 }
 
 struct Prepared {
@@ -207,6 +221,8 @@ impl CollapsingHeader {
             default_open,
             id_source,
             enabled: _,
+            selectable: _,
+            selected: _,
         } = self;
 
         label.text_style = label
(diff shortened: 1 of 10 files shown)


egui_demo_lib/src/apps/demo/misc_demo_window.rs
```diff
@@ -374,7 +393,13 @@ impl Tree {
             .into_iter()
             .enumerate()
             .filter_map(|(i, mut tree)| {
-                if tree.ui_impl(ui, depth + 1, &format!("child #{}", i)) == Action::Keep {
+                if tree.ui(
+                    ui,
+                    depth + 1,
+                    &format!("child[{}][{}]", depth, i),
```
emilk, line 399: This naming scheme leads to name clashes:

Screen Shot 2021-08-16 at 21 35 49

A robust alternative is to give the full path to each child, e.g. `root/0/2/3`, with `&format("{}/{}", parent_name, i)`

egui_demo_lib/src/apps/demo/misc_demo_window.rs
```diff
@@ -339,28 +339,47 @@ enum Action {
...
-        self.ui_impl(ui, 0, "root")
+        self.1.ui(ui, 0, "root", &mut self.0)
     }
+}
 
-    fn ui_impl(&mut self, ui: &mut Ui, depth: usize, name: &str) -> Action {
-        CollapsingHeader::new(name)
+#[derive(Clone, Default)]
+#[cfg_attr(feature = "persistence", derive(serde::Deserialize, serde::Serialize))]
+struct SubTree(Vec<SubTree>);
```
emilk, line 361: Why do we need both `Tree` and `SubTree`?

egui/src/containers/collapsing_header.rs
```diff
@@ -188,6 +192,19 @@ impl CollapsingHeader {
         self.enabled = enabled;
         self
     }
+
+    /// Can the `CollapsingHeader` be selected by clicking it? Default: `false`.
+
```
emilk, line 197: I think we should also have some example code, as it is not obvious how to check if the header was selected.

```suggestion
    /// Can the `CollapsingHeader` be selected by clicking it? Default: `false`.
    ///
    /// ``` rust
    /// # let ui = &mut egui::Ui::__test();
    /// let mut selected = false;
    /// let response = CollapsingHeader::new("Select and open me")
            .selectable(true)
            .selected(selected)
            .show(ui, |ui| ui.label("Content"));
    /// if response.header_response.clicked() {
    ///     selected = true;
    /// }
    /// ```
```

egui/src/containers/collapsing_header.rs
```diff
@@ -260,6 +285,15 @@ impl CollapsingHeader {
             });
         }
 
+        let selected = self.selected || header_response.hovered() || header_response.has_focus();
+        if self.selectable && selected {
```
emilk, line 289: From the docstring of `selected` it seems like we should show it as selected even if it is not selectable. This can be useful if the selection is done via some other means (e.g. as a highlight).

```suggestion
        if self.selected || self.selectable && (header_response.hovered() || header_response.has_focus()) {
        }
```

emilk (changes requested):
Looks useful, but let's polish it a bit more!

egui_demo_lib/src/apps/demo/misc_demo_window.rs
gents83: I preferred to split in order to avoid changing too muche the code and having the selected string everywhere, but we can change it if you prefer

egui_demo_lib/src/apps/demo/misc_demo_window.rs
gents83: Ah! I didn't noticed it - my bad - I'll change it

emilk (approved):
Looks good!

emilk:
I found a nicer approach to this in https://github.com/emilk/egui/pull/1538 so I will likely deprecate or remove `CollapsingHeader::selectable`. Please take a look @gents83

PR #624, opened 2021-08-16
Add possibility of panels inside UI (i.e. windows)

### LINKED ISSUE #375: Add a resizeable SplitView panel
Would it please be possible to add a resizeable split-view container, that allows either a vertical or horizontal split, to the standard set of widgets?

gents83:
Here you can find an example of what we could obtain with Panels that could be used inside other ui elements.

https://user-images.githubusercontent.com/62186646/128388159-c703e2fd-b120-45b1-8593-9536d290c3f8.mp4

Could close #375 

I've also added a Window with Panels demo

### DIFF
--- a/egui/src/containers/panel.rs
+++ b/egui/src/containers/panel.rs
@@ -137,9 +137,9 @@ impl SidePanel {
 }
 
 impl SidePanel {
-    pub fn show<R>(
+    pub fn show_inside<R>(
         self,
-        ctx: &CtxRef,
+        ui: &mut Ui,
         add_contents: impl FnOnce(&mut Ui) -> R,
     ) -> InnerResponse<R> {
         let Self {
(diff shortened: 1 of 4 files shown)


egui/src/containers/panel.rs
```diff
@@ -168,67 +166,84 @@ impl SidePanel {
...
         if resize_hover || is_resizing {
             let stroke = if is_resizing {
-                ctx.style().visuals.widgets.active.bg_stroke
+                ui.style().visuals.widgets.active.bg_stroke
             } else {
-                ctx.style().visuals.widgets.hovered.bg_stroke
+                ui.style().visuals.widgets.hovered.bg_stroke
             };
             // draw on top of ALL panels so that the resize line won't be covered by subsequent panels
-            let resize_layer = LayerId::new(Order::PanelResizeLine, Id::new("panel_resize"));
```
emilk, line 224: This makes the resize line go on top of everything else, which is unfortunate but inevitable. Ideally we should have some way of saying "This layer goes directly on top of this other layer", i.e. let each window have it's own foreground layer. But alas, we don't have that yet.

egui_demo_lib/src/apps/demo/window_with_panels.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn ui(&mut self, ui: &mut egui::Ui) {
+        let left_panel_min_width = 100.;
+        let left_panel_max_width = left_panel_min_width * 4.;
+        let bottom_height = 25.;
+
+        ui.expand_to_include_rect(ui.max_rect()); // Expand frame to include it all
+
+        let mut top_rect = ui.available_rect_before_wrap_finite();
+        top_rect.min.y += ui.spacing().item_spacing.y;
+        let mut top_ui = ui.child_ui(top_rect, Layout::top_down(Align::Max));
```
emilk, line 34: Ideally we should have padding options on the Panel directly so we don't need to do this.

emilk (approved):
This is great, thank you!

PR #629, opened 2021-08-18
Make it easy to panels inside of Ui:s

emilk:
This builds on https://github.com/emilk/egui/pull/624 to improve the ergonomics of sticking panels inside of a Ui.

panels

PTAL @gents83


gents83 (commented):
Now it's really nice!!!

gents83 (approved):
New functionalities allowed "hiding" to user all panels positioning - Super appreciated!

Issue #581, opened 2021-07-24
TopBottomPanel resizing doesn't seem to work


UditDey:
I'm new to this library and was experimenting by modifying the template application. If my understanding is right the following code should create a bottom bar that can be resized by dragging:
```
fn update(&mut self, ctx: &egui::CtxRef, frame: &mut epi::Frame<'_>) {
        egui::TopBottomPanel::bottom("bottom_panel")
        .default_height(200.0)
        .resizable(true)
        .show(ctx, |ui| {
            ui.label("Some Text");
        });
}
```
It doesn't seem to work though. The attached video shows what happens:

https://user-images.githubusercontent.com/44467986/126878450-3dd0099c-2a60-420b-8f88-1fe4d012fed9.mp4

The bottom bar stays shrunk to the label's size and when dragged it moves up instead of resizing, and immediately goes back when you release it. Is this a bug or am I using it incorrectly? The online demo has functional resizable panels so I thought I was using it incorrectly. I'm running it via `eframe` like the template application

EDIT:
I added a `CentralPanel` and that had no effect
```
egui::TopBottomPanel::bottom("bottom_panel")
        .resizable(true)
        .show(ctx, |ui| {
            ui.label("Some Text");
        });

        egui::CentralPanel::default().show(ctx, |ui| {
            ui.label("Some Other Text");
        });
```
The bottom panel still doesn't get resized. It drags along with the mouse and immediately returns to its original state when you let go

gents83:
I noticed the same today - the issue is with all widgets that do not "expand" to full available height

UditDey:
So do all the layouts normally shrink down to the size of the child components?

gents83:
> So do all the layouts normally shrink down to the size of the child components?

Not for all of them, but that's what TopBottomPanel does right now :)

UditDey:
Alright thanks, I'm gonna look into creating a custom layout for my use case instead

PR #641, opened 2021-08-23
Fix empty side panels not dragging properly

aclysma:
I noticed when a side panel is empty, dragging the side of it to widen it doesn't work properly. The demo seems to work because there are widgets in the side panels that allocate space (for example the centered header text). Modifying the demo's side panels to be empty reproduces the same problem I was seeing in my app.

I don't know if this is the correct fix, but I think this is the minimal change necessary to make an empty panel act like a panel with contents that do allocate space.

### DIFF
--- a/egui/src/containers/panel.rs
+++ b/egui/src/containers/panel.rs
@@ -208,6 +208,7 @@ impl SidePanel {
         let inner_response = frame.show(&mut panel_ui, |ui| {
             ui.set_min_height(ui.max_rect_finite().height()); // Make sure the frame fills the full height
             ui.set_width_range(width_range);
+            ui.allocate_space(Vec2::new(ui.available_width(), 0.0));
             add_contents(ui)
         });
 


aclysma:
I just ran across the issue describe here: https://github.com/emilk/egui/issues/581

I tried to make a similar change as in the PR to make a bottom bar expandable by dragging, but wasn't able to get anything to work. The intent was to have widgets like scroll areas that can expand to fill it (like a docking tab in other apps, but without the complexity of completely arbitrary docking.)

aclysma:
Thanks for the suggestion, but the TopBottomPanel unfortunately was still not resizable. (I did change show_inside() to show() though.. not sure that it would make a difference)

I think for this particular thing I'm prototyping, using imgui for now in the short-term might make more sense (the goal here was to have a lower-complexity way to mimic docking behavior). That said, I think egui is great and it's progressing forward very quickly! Thanks for all the work you've put into it, I look forward to eventually being able to use egui more than I already do :)

PR #650, opened 2021-08-25
Fix alpha blending in WebGL backends

AsmPrgmC3:
This fixes alpha blending in the WebGL backends by first blending everything in a linear framebuffer, before converting the result to sRGB in a second shader.

The results of the color test are now identical to the egui_glium backend when using the WebGL2 backend.
The WebGL1 backend suffers from reduced precision in dark colors without the `SRGB8_ALPHA8` format and still doesn't have correct texture interpolation.

Color test with the change (I have since removed the warning):
WebGL2
WebGL2
WebGL1
WebGL1

(Sorry for not following the contributing guidelines exactly ("First file an issue" and all that). This was a spontaneous test as I didn't know whether it was even possible and I didn't wan't to file an issue the same minute as closing it (Should I?).)


egui_web/src/webgl1.rs
```diff
@@ -485,10 +495,175 @@ impl crate::Painter for WebGlPainter {
...
+
+        let texture = gl.create_texture().unwrap();
+        gl.bind_texture(Gl::TEXTURE_2D, Some(&texture));
+        gl.tex_parameteri(Gl::TEXTURE_2D, Gl::TEXTURE_WRAP_S, Gl::CLAMP_TO_EDGE as i32);
+        gl.tex_parameteri(Gl::TEXTURE_2D, Gl::TEXTURE_WRAP_T, Gl::CLAMP_TO_EDGE as i32);
+        gl.tex_parameteri(Gl::TEXTURE_2D, Gl::TEXTURE_MIN_FILTER, Gl::NEAREST as i32);
+        gl.tex_parameteri(Gl::TEXTURE_2D, Gl::TEXTURE_MAG_FILTER, Gl::NEAREST as i32);
+        gl.pixel_storei(Gl::UNPACK_ALIGNMENT, 1);
+        // TODO: https://developer.mozilla.org/en-US/docs/Web/API/EXT_sRGB
+        // Dark colors are slightly wrong when not using SRGB8_ALPHA8 format
```
emilk, line 533: This will mean any images shown with the WebGL1 backend (e.g. on iPhones) will have a lot of banding in the dark regions. For instance, in the "HTTP" demo tab you can enter https://picsum.photos/seed/298.393/640/480 and see really ugly banding on an iPhone with this new render-to-texture approach.

It also makes the egui window shadows look really bad in dark mode.

I think we should probably revert the changes to the WebGL1 pipeline, and only do this render-to-texture trick in WebGL2.

egui_demo_lib/src/apps/color_test.rs
```diff
@@ -36,14 +36,6 @@ impl epi::App for ColorTest {
...
         egui::CentralPanel::default().show(ctx, |ui| {
-            if frame.is_web() {
-                ui.colored_label(
-                    RED,
-                    "NOTE: The WebGL backend does NOT pass the color test."
-                );
-                ui.small("This is because WebGL does not support a linear framebuffer blending (not even WebGL2!).\nMaybe when WebGL3 becomes mainstream in 2030 the web can finally get colors right?");
-                ui.separator();
-            }
             ScrollArea::auto_sized().show(ui, |ui| {
```
emilk, line 39: Ideally we should still show a warning about WebGL1 not passing the color test. Right now there is no way to query that, but perhaps a milder warning would fit here:

``` rust
if frame.is_web() {
    ui.label("NOTE: the WebGL1 backend does not pass the color test, but the WebGL2 backend does.");
}
```

epaint/src/texture_atlas.rs
```diff
@@ -17,19 +17,10 @@ impl Texture {
     }
 
     /// Returns the textures as `sRGBA` premultiplied pixels, row by row, top to bottom.
-    ///
-    /// `gamma` should normally be set to 1.0.
-    /// If you are having problems with egui text looking skinny and pixelated, try
-    /// setting a lower gamma, e.g. `0.5`.
-    pub fn srgba_pixels(&'_ self, gamma: f32) -> impl Iterator<Item = super::Color32> + '_ {
+    pub fn srgba_pixels(&'_ self) -> impl Iterator<Item = super::Color32> + '_ {
```
emilk, line 20: As noted elsewhere, I think we should revert the changes to the WebGL1 pipeline, so this code block also needs to be reverted.

emilk (changes requested):
This is wonderful work; thank you so much!

As I noted in my review I think the WebGL1 pipeline should remain as it was, but the improvements to the WebGL2 renderer is massive, and works great! ❤️

egui_web/src/webgl1.rs
AsmPrgmC3: Makes sense, will revert

PR #660, opened 2021-08-27
feat: Set whether to show decorations

zu1k:
Set whether to show decorations when update frame.

```rust
frame.set_decorations(false);
```

I have tested it on my project: https://github.com/zu1k/copy-translator/blob/6eca80ca32593f2d55eda0fc950e1f1ff0cd52eb/src/ui.rs#L252

### DIFF
--- a/eframe/CHANGELOG.md
+++ b/eframe/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to the `eframe` crate.
 
 
 ## Unreleased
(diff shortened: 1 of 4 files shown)


emilk:
I reverted the merge, because this PR removed decorations *by default*, at least on my mac. Not good!

I should have tested it first 🤦

zu1k:
Sorry for that, it's my fault, I will relaunch a PR.

PR #672, opened 2021-08-28
feat: Set whether to show decorations

zu1k:
I am sorry that the previous PR was not tested in the demo, I misunderstood the meaning of `output`, I am very sorry.

In this PR, I added `decorated: Option<bool>` to the `output`, which stands for the `set_decorations` instruction, `None` will not operate.

I have tested the following in the demo:

- Default `true`, updat to `true`, pass
- Default `true`, updat to `false`, pass
- Default `false`, update to `false`, pass
- Default `false`, update to `true`, can show the decorations, but there will be a problem as shown in the figure, I am not sure why, by modifying the window size with the mouse can be restored to normal, need help here

image


emilk:
I tried it using this: https://github.com/emilk/egui/commit/9ae776202942bd2fa924c38a18d857d4eb75bce2

And it works great on my mac! I'm unsure what the problem is on windows. Perhaps one need to prod winit with some special call?

---

On a separate note: what is the user story behind this? i.e., why do you want to toggle decorations on/off at runtime?

zu1k:
I'm developing a translate-on-copy app: https://github.com/zu1k/copy-translator

What I actually need is a tooltop window, so I need to hide the decorations to simulate.

It was difficult to move and change the size when hiding the decorations, so I had to provide a switch to show the decorations.

I think it would be better if egui could provide the ability to move and resize when hiding decorations by default.

zu1k:
> I tried it using this: [9ae7762](https://github.com/emilk/egui/commit/9ae776202942bd2fa924c38a18d857d4eb75bce2)
> 
> And it works great on my mac! I'm unsure what the problem is on windows. Perhaps one need to prod winit with some special call?

I have tested this demo, it works well on windows. I alse test this on my app by using a switch, it also works well.

Could you test this https://github.com/zu1k/egui/commit/ff1daac04c521bca136941e7271c2064b273ed85? Problems can occur in this case.

emilk:
> I'm developing a translate-on-copy app: https://github.com/zu1k/copy-translator

Oh, very cool!

> What I actually need is a tooltop window, so I need to hide the decorations to simulate.
> 
> It was difficult to move and change the size when hiding the decorations, so I had to provide a switch to show the decorations.
> 
> I think it would be better if egui could provide the ability to move and resize when hiding decorations by default.

I agree - have you tried hooking up <https://docs.rs/winit/0.25.0/winit/window/struct.Window.html#method.drag_window>?
For instance, you could have a button in egui ("drag me to move the window") which would call `frame.drag_window()` which in turn would call the winit function above.

zu1k:
> I agree - have you tried hooking up https://docs.rs/winit/0.25.0/winit/window/struct.Window.html#method.drag_window?
> For instance, you could have a button in egui ("drag me to move the window") which would call `frame.drag_window()` which in turn would call the winit function above.

Thanks for your advice, I considered a similar approach to move, but it looks like changing the size is not very easy to implement.

Issue #662, opened 2021-08-27
Support for fullscreen windowing


jeremymeadows:
It doesn't appear that egui has support for fullscreen windowing. You can remove window decorations and set the size to fill the whole screen, but that isn't a great solution.
I added a `fullscreen` field to `NativeOptions` which gets passed through to glutin at compile time, but that isn't as useful as being able to change the mode at runtime. If anybody knows a way to get a reference to the `glutin::window::Window` from the egui context, I'd be happy to try to implement it and make a PR.

emilk:
Closed by https://github.com/emilk/egui/pull/1782

Issue #673, opened 2021-08-28
Refactor epi::Frame


emilk:
[`epi::Frame`](https://docs.rs/epi/0.14.0/epi/struct.Frame.html) contains methods for controlling the surrounding app (i.e. the native winit window or the web browser). At the moment this includes allocating texture, quitting and changing the window size. We may soon also be able to [toggle decorations](https://github.com/emilk/egui/pull/672), [toggle fullscreen](https://github.com/emilk/egui/issues/662) and [request user attention](https://github.com/emilk/egui/issues/622).

At the moment `Frame` is only available in [`App::update`](https://docs.rs/epi/0.14.0/epi/trait.App.html#tymethod.update) which means you cannot call this methods from a separate thread.

I think it would be useful if we could change `App::update` (and `App::setup`) from taking a `&Frame<'_>` to taking a `&Arc<Frame>`. A user could clone this `Arc` and send to a background thread. This would allow users to e.g. start HTTP requests, and when finished request user attention, allocate a texture, or similar.

Perhaps we want something like `struct FrameRef(Arc<Frame>)` or `FrameRef(Arc<Mutex<Frame>>)` to encapsulate this in a nicer way (similar to `egui::CtxRef` vs `egui::Context`).

We could then replace the current [`repaint_signal`](https://docs.rs/epi/0.14.0/epi/struct.Frame.html#method.repaint_signal) with just a normal call (`frame.request_repaint();`, perhaps more appropriately named `request_update`).

cessen:
Just want to drop a note here that I would find this *especially* useful with the applications I'm writing.  I go out of my way to make sure than any long-running jobs aren't on the same thread as the GUI, but often these are jobs computing or updating raster images to display.

Currently I kind-of-sort-of work around this by doing most of the real image processing work in another thread (taking tens of seconds in my case), and then when that work is done marking the image as "ready" for a texture update on a thread-safe bool variable.  Then the GUI code checks that variable on every update, and transfers the image to a texture when `true`.

It works okay, since the long-running work happens on another thread.  But even just transferring the image data to a texture takes a small bit of time (these are very high-res images) and makes the gui hiccup whenever it happens.

Being able to allocate textures from other threads would both simplify the code a lot and eliminate the hiccups.

PR #674, opened 2021-08-28
Fullscreen

jeremymeadows:
The default state can be set with the `NativeOptions` at compile time, and can be changed by using a callback to the UI context during runtime. The fullscreen state is written to the RON file so that the user-set state is restored on each launch.

I added a 'toggle fullscreen' button to the UI too, so that can be previewed.

Closes <https://github.com/emilk/egui/issues/662>.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ NOTE: [`eframe`](eframe/CHANGELOG.md), [`egui_web`](egui_web/CHANGELOG.md) and [
 
 ### Added ⭐
 * Add horizontal scrolling support to `ScrollArea` and `Window` (opt-in).
+* Support for fullscreen windowing.
 
 ### Changed 🔧
 * All `Ui`:s must now have a finite `max_rect`.
--- a/egui/src/context.rs
+++ b/egui/src/context.rs
@@ -719,12 +719,24 @@ impl Context {
 
     // ---------------------------------------------------------------------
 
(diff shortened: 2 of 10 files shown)


emilk:
Thanks for the PR!

`fullscreen` is not something that should be part of the `egui` crate, but instead should be part of the [`epi::Frame`](https://docs.rs/epi/0.14.0/epi/struct.Frame.html) API, just like `set_window_size` is!

jeremymeadows:
Got it, sorry I didn’t see that second comment you made on the issue, but I'll see what I can do about moving that around :)

jeremymeadows:
I got really busy with school stuff and work for several weeks, but finally winding down. I moved those changes around. It was a little tricky to deal with the AppOutput resetting itself every frame; that would undo whatever state was set in the previous frame. I pulled it out to give it slightly broader scope, and I didn't notice any problems it caused.

emilk:
> It was a little tricky to deal with the AppOutput resetting itself every frame

Yeah, [it is in need of a refactor](https://github.com/emilk/egui/issues/673).

emilk:
Any interest in this still?

antouhou:
@emilk Yes! Is there any way I can help get this merged?

emilk:
@antouhou Since a lot in this PR is outdated (a lot of epi/eframe refactor), I'm guessing your best bet it so create a new PR from scratch, perhaps with some inspiration taken from this one.

Issue #680, opened 2021-08-29
Make egui_glium support any Surface


JerzySpendel:
As of now `egui_glium::EguiGlium`'s `paint` function accepts only `glium`'s `Display` structure. In my case I wondered if I can add any post-processing on top of egui's output. Let's say, a shockwave effect. 

In order to do that I would have to execute `paint` on temporary framebuffer, do the postprocessing on the framebuffer, copy frambuffer to a texture, and only then render the texture on Display. But I couldn't do that as the paint function accepts only Display, even though it uses only the assumption that Display is Surface (as framebuffer is) and nothing more.

 So I thought it shouldn't be bad idea to make the paint function generic over some `T: Surface` which is satisfied by Display, SimpleFramebuffer and couple more. Effects below:

![2021-08-29 17-31-39 (1)](https://user-images.githubusercontent.com/1176069/131256588-aa1c1cd0-121d-4b5e-9489-1773cb03c350.gif)

Essential code example:

```rust
let mut texture = glium::texture::SrgbTexture2d::empty(&display, 2000, 1100).unwrap();
let mut framebuffer = glium::framebuffer::SimpleFrameBuffer::new(&display, &texture).unwrap();


framebuffer.clear_color(
    0.1,
    0.2,
    0.3,
    0.,
);

// draw things behind egui here

egui.paint(&display, &mut framebuffer, shapes);


use glium::uniform;
use glium::DrawParameters;

let uniforms = uniform! {
    tex: glium::uniforms::Sampler::new(&texture),
    time: (Instant::now() - now).as_secs_f32() / 5.
};

let mut target = display.draw();
target.clear_color(0.,0., 0., 0.);
target.draw(&vertex_buffer, &index_buffer, &program, &uniforms, &Default::default());
target.finish().unwrap();

```

PR #681, opened 2021-08-29
Make paint function accept anything implementing Surface trait

JerzySpendel:
Closes <https://github.com/emilk/egui/issues/680>.

### DIFF
--- a/egui_glium/src/lib.rs
+++ b/egui_glium/src/lib.rs
@@ -580,10 +580,10 @@ impl EguiGlium {
         (needs_repaint, shapes)
     }
 
-    pub fn paint(
+    pub fn paint<T: glium::Surface>(
         &mut self,
         display: &glium::Display,
-        target: &mut glium::Frame,
+        target: &mut T,
         shapes: Vec<egui::epaint::ClippedShape>,
     ) {
         let clipped_meshes = self.egui_ctx.tessellate(shapes);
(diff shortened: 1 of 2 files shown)


egui_glium/src/lib.rs
```diff
@@ -580,10 +580,10 @@ impl EguiGlium {
         (needs_repaint, shapes)
     }
 
-    pub fn paint(
+    pub fn paint<T: glium::Surface>(
         &mut self,
         display: &glium::Display,
-        target: &mut glium::Frame,
+        target: &mut T,
```
emilk, line 586: Let's replace generics with dynamic dispatch instead to keep down own monomorphization overhead (larger code and slower compilation).

```suggestion
        target: &mut dyn glium::Surface,
```

here and below

emilk (changes requested):

egui_glium/src/lib.rs
JerzySpendel: He @emilk, thanks for reply :) That was also my thought to make it dynamic dispatch, but it's not possible as Surface trait is not object-safe.

emilk: Oh, ok then :)

emilk:
You got some unused import warnings, then LGTM

JerzySpendel:
Done

PR #694, opened 2021-09-03
add fltk-egui as a 3rd party backend crate to the README

MoAlyousef:
Hello

This is a trivial PR that just modifies the README to add fltk-egui as a 3rd party backend crate for egui, which has several advantages like embedding egui as a subview in an FLTK app. It also supports multiwindowing, system menus on macos, dialogs.

Thanks for the great work.


emilk (approved):
Lovely!

Issue #688, opened 2021-09-01
API's for accessing underlying implementation?


sdfgeoff:
This is a weird one, but it represents several hours of frustration.

I have two libraries: egui, and one that I wrote. The library that I wrote requires an openGL handle at initialisation and outputs to an openGL texture. 
I created an interface in egui using the template at https://github.com/emilk/egui_template/ as a model. Now I want to run my library and output the texture to the interface of egui. 

As far as I can tell there is no way to get a reference to the underlying draw implementation from an egui app, and there is no way to coerce a GL texture into an egui texture. This is all for a good reason! Egui is intended to be backend agnostic, so it shouldn't matter if you're running in GL or not!
But it leaves me at a loss to know how to proceed. 
> I did manage to get it running by using the egui_glow port mentioned https://github.com/emilk/egui/issues/93#issuecomment-907607337 and modifying the app.setup() and app.render() functions to include the gl reference. However I then got stuck on how to pass a texture back :) 

So this is a request for documentation that answers questions like:
- How do you display a texture in EGUI where the texture resides on the graphics card? (aka is not an array of pixels in RAM)
- How do you create an application that interacts with the GPU without requiring you to write your own backend for EGUI?

--------------------------------------------------

If the answer to either of those is "you don't" then perhaps it would be worth investigating alternate interfaces to eframe/egui_glium/egui_web that allows you to. As far as I can tell the root of the problem is that the TemplateApp's method of:
```
eframe::run_native(Box::new(app), native_options);
```
does not allow your app to have any reference to the implementation behind egui. Ideally I'd like to avoid vendoring exact clones of these crates just to change a handful of function definitions!

emilk:
The short answer is: don’t use eframe. I suggest using egui_miniquad or waiting for egui_glow. egui_glium should also work well, if you follow https://github.com/emilk/egui/blob/master/egui_glium/examples/pure.rs though an example of user textures is missing there.


If all you need is to allocate a texture, then eframe got you covered: https://docs.rs/epi/0.14.0/epi/struct.Frame.html#method.tex_allocator

t18b219k:
you can register glium texture as egui texture using egui_glium eframe doesn't support offscreen rendering
use egui_glium's register_glium_texture()->egui::TextureId 
and give  this Id to egui::Image initalization

PR #695, opened 2021-09-03
Add api for accessing backend texture via epi

t18b219k:
Closes <https://github.com/emilk/egui/issues/688>.

This PR provide standard way to accessing native texture .
`NativeTexture` trait provides two method
* register_native_texture
  Bind allocated  GPU memory to `egui::TextureId`
* replace_native_texture
  Replacing `egui::TextureId`'s actual pointing texture to given texture without copy.

### use case of this trait
* FFT
* N-body
* Animation
* Drawing
*  any other GPU memory displaying task


egui_glium/src/painter.rs
```diff
@@ -319,3 +319,22 @@ impl Painter {
...
+        if let egui::TextureId::User(id) = id {
+            if let Some(Some(user_texture)) = self.user_textures.get_mut(id as usize) {
+                *user_texture = UserTexture {
+                    pixels: vec![],
+                    gl_texture: Some(replacing),
+                };
+            }
+        }
+    }
+}
```
emilk, line 340: I don't see the point of the `epi::NativeTexture` trait - the point of `epi` is to be able to write code that is the same on native and web, but this trait has a different `Texture` on different platforms, and assumes the user will be using native WebGL/glium functions to create it.

I guess maybe it can be used to write some sort of generic texture manager?

egui_glium/src/painter.rs
```diff
@@ -319,3 +319,22 @@ impl Painter {
         }
     }
 }
+
+impl epi::NativeTexture for Painter {
+    type Texture = glium::texture::srgb_texture2d::SrgbTexture2d;
+
+    fn bind_native_texture(&mut self, native: Self::Texture) -> egui::TextureId {
+        self.register_glium_texture(native)
```
emilk, line 327: I think `register_native_texture` is a better name, and just unify the functions

egui_glium/src/painter.rs
```diff
@@ -319,3 +319,22 @@ impl Painter {
...
+    }
+
+    fn replace_texture_ref(&mut self, id: egui::TextureId, replacing: Self::Texture) {
```
emilk, line 330:
```suggestion
    fn replace_native_texture(&mut self, id: egui::TextureId, replacing: Self::Texture) {
```

emilk (commented):
let's reuse the same example image `eframe/examples/rust-logo-512x512.png` (feel free to move it to a more logical place) instead of adding a new one

egui_glium/Cargo.toml
```diff
@@ -39,6 +39,8 @@ tts = { version = "0.17", optional = true }
 # feature "time"
 chrono = { version = "0.4", optional = true }
 
+[dev-dependencies]
+image="0.23"
```
emilk, line 43:
```suggestion
image = { version = "0.23", default-features = false, features = ["png"] }
```

egui_glium/examples/native_texture.rs
```diff
@@ -0,0 +1,114 @@
...
+    .to_rgba8();
+    let image_dimensions = image.dimensions();
+    //mark as image
+    let image = glium::texture::RawImage2d::from_raw_rgba(image.into_raw(), image_dimensions);
+    //load to gpu memory
+    let native_texture = glium::texture::SrgbTexture2d::new(&display, image).unwrap();
+    //allocate egui's texture id for GL texture
+    let texture_id = egui
+        .ctx_and_painter_mut()
+        .1
```
emilk, line 44:
```suggestion
        .painter_mut()
```

emilk (changes requested):
The new replace texture function seems useful, but I'd like to understand the motivation behind using a common `NativeTexture` trait instead of just adding these functions to the respective painters?

egui_glium/examples/native_texture.rs
t18b219k: ok i will change

egui_glium/Cargo.toml
t18b219k: ok i will change

egui_glium/src/painter.rs
t18b219k: yes i think we can write backend independent texture manager.

egui_glium/src/painter.rs
t18b219k: i changed it

egui_glium/src/painter.rs
t18b219k: i changed it

egui_glium/src/lib.rs
```diff
@@ -40,7 +40,8 @@ use {
     std::hash::{Hash, Hasher},
 };
 
-pub use copypasta::ClipboardContext; // TODO: remove
+pub use copypasta::ClipboardContext;
+use std::borrow::BorrowMut; // TODO: remove
```
emilk, line 44: I don't know why I wrote this `TODO`, but it is now on the wrong line :)

egui_glium/src/painter.rs
```diff
@@ -241,24 +241,6 @@ impl Painter {
...
-            if let Some(Some(user_texture)) = self.user_textures.get_mut(id as usize) {
-                *user_texture = UserTexture {
-                    pixels: vec![],
-                    gl_texture: Some(texture),
-                }
-            }
-        }
-        id
-    }
-
```
emilk, line 261: Please add back the old method with a deprecation notice

``` rust
    #[deprecated = 'Use NativeTexture::register_native_texture` instead]
    pub fn register_glium_texture(
        &mut self,
        texture: glium::texture::SrgbTexture2d,
    ) -> egui::TextureId {
        use NativeTexture as _;
        self.register_native_texture(texture)
    }
```

emilk (changes requested):
Looks good, just a few more comments!

egui_glium/src/painter.rs
t18b219k: ok i will add depecated notice

PR #702, opened 2021-09-07
Increase scroll speed for glium backend

### LINKED ISSUE #461: Scroll is too slow on Gnome/Wayland
Scrolling speed is very slow in Gnome/Wayland using glium backend.
[Screencast](https://user-images.githubusercontent.com/180812/120749843-790d9480-c52f-11eb-95e6-0729af0d579e.mp4)

There must be a way to solve all this scrolling mess.

Scroll event handling happens here: https://github.com/emilk/egui/blob/2af86cd2c7a1a1b275034fa6e3f0e3548384154b/egui_glium/src/lib.rs#L178
Related issue with useful comments: https://github.com/emilk/egui/issues/159#issuecomment-776593734

https://user-images.githubusercontent.com/180812/120749843-790d9480-c52f-11eb-95e6-0729af0d579e.mp4

akhilman:
Each so called "line" of the MuseScrollDelta::LineDelta de-facto represents
a single scroller click which in most cases equals to 15 degrees of wheel
rotation.

Currently, one scroller click scrolls exactly one line of the text,
which is very slow.

This change multiplies the scroll rate by 5. So, with one click of the
scroller, 5 lines scrolled. This number is taken from the "iced" GUI
library and feels quite balanced.

Closes #461

### DIFF
--- a/egui_glium/src/lib.rs
+++ b/egui_glium/src/lib.rs
@@ -252,7 +252,8 @@ pub fn input_to_egui(
             let mut delta = match *delta {
                 glutin::event::MouseScrollDelta::LineDelta(x, y) => {
                     let line_height = 8.0; // magic value!
-                    vec2(x, y) * line_height
+                    let number_of_lines = 5.0;  // number of lines per one scroller click
+                    vec2(x, y) * line_height * number_of_lines
                 }
                 glutin::event::MouseScrollDelta::PixelDelta(delta) => {
                     vec2(delta.x as f32, delta.y as f32) / pixels_per_point


emilk:
Thank you! I did my own commit because I felt we should change egui_web at the same time, and update the changelog

akhilman:
egui_web was totally fine. With 50 as multiplier it scrolls for 300 pixels per click.

emilk:
Haha 🤦  -  thanks for testing @akhilman ! The change has been reverted

Issue #548, opened 2021-07-06
DragValue hover text have default not removable message


vorchunpaul:
when I try to add a message to the widget, I add third-party garbage that my users don't like (because they don't like reading in 2 languages rus + eng)
![image](https://user-images.githubusercontent.com/69152326/124634868-7df79800-de8f-11eb-9ea4-334e7d474d7a.png)

now in the intermediate versions of the program, I just delete the standard message in the widget source code
![image](https://user-images.githubusercontent.com/69152326/124636020-bfd50e00-de90-11eb-978f-a07d9984cebf.png)

probably still worth removing, because this is normally used only in the demo

PR #704, opened 2021-09-09
Tooltip of DragValue can be disabled.

n2:
The DragValue tooltip can be annoying sometimes, it may obscure the widgets below, or degrade the user experience (#548).
Therefore, it is best to allow developers to control whether tooltips are enabled.
Closes <https://github.com/emilk/egui/issues/548>.

### DIFF
--- a/egui/src/widgets/drag_value.rs
+++ b/egui/src/widgets/drag_value.rs
@@ -55,6 +55,7 @@ pub struct DragValue<'a> {
     clamp_range: RangeInclusive<f64>,
     min_decimals: usize,
     max_decimals: Option<usize>,
+    enable_tooltip: bool,
 }
 
 impl<'a> DragValue<'a> {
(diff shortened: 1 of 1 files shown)


sourcebox:
I think a good solution would be to also have some custom text for the tooltip in addition to disable it completely. This would make internationalization much easier.

emilk:
@sourcebox localization of the text is covered by https://github.com/emilk/egui/pull/426 and setting a custom tooltip is already possible with `ui.add(DragValue::new(…)).on_hover_text(…)`

emilk:
@n2 I'm not convinced that having this option per DragValue is the right approach. Isn't it more likely that users would like to disable it for the entire program, e.g. using `Style`?

sourcebox:
@emilk Thanks, I had a look into the source and saw that the text was hard-coded. So I was not sure if it can be changed.

n2:
@emilk Yes, frankly, I think it's better to remove the default tooltip from widgets like DragValue. Even without considering the issue of internationalization, in different scenarios, the same widget may require different tooltips. It is best to leave the tooltip to the developer, as it is so easy to implement. On the other hand, for some widgets, one should delete tooltips with care, as it is not easy to implement (ie the color_picker button) or is necessary (ie a hyperlink). Should I close this pull request and open an issue to carefully organize the tooltips?

PR #708, opened 2021-09-11
Hide DragValue tooltips unless user set Style::explanation_tooltips

### LINKED ISSUE #548: DragValue hover text have default not removable message

emilk:
Closes #548
Closes #704

Please take a look, @n2 and/or @vorchunpaul


n2:
Great job! Also, 
https://github.com/emilk/egui/blob/f2b6edd6db0d59187e66481ef62909eee6f59610/egui/src/ui.rs#L1149
and
https://github.com/emilk/egui/blob/f2b6edd6db0d59187e66481ef62909eee6f59610/egui/src/widgets/color_picker.rs#L341
are suggested to modified simultaneously.

PR #728, opened 2021-09-19
epi: drag native window

zu1k:
I add a `drag_window` function to frame.

```
if ui.button("Drag Window").is_pointer_button_down_on() {
    frame.drag_window();
}
```

### DIFF
--- a/egui_demo_lib/src/backend_panel.rs
+++ b/egui_demo_lib/src/backend_panel.rs
@@ -208,6 +208,10 @@ impl BackendPanel {
             if ui.button("Quit").clicked() {
                 frame.quit();
             }
+
+            if ui.button("Drag Window").is_pointer_button_down_on() {
+                frame.drag_window();
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


epi/src/lib.rs
```diff
@@ -252,6 +252,11 @@ impl<'a> Frame<'a> {
         self.0.output.decorated = Some(decorated);
     }
 
+    /// Drag window
```
emilk, line 255: This needs some documentation, maybe something like

```suggestion
    /// When called, the native window will follow the
    /// movement of the cursor while the primary mouse button is down.
    ///
    /// Does not work on the web, and works badly on Mac.
```

…because it does work very badly on my mac :)

egui_demo_lib/src/backend_panel.rs
```diff
@@ -208,6 +208,10 @@ impl BackendPanel {
             if ui.button("Quit").clicked() {
                 frame.quit();
             }
+
+            if ui.button("Drag Window").is_pointer_button_down_on() {
```
emilk, line 212:
```suggestion
            if ui.button("Drag me to drag window").is_pointer_button_down_on() {
```

or perhaps a tooltip

emilk (changes requested):

PR #748, opened 2021-09-28
Fix disabled textedit blocking focus shift.

### LINKED ISSUE #732: Can't use tab to shift focus past a disabled textedit
```rust
            let mut s = String::from("balalala");
            ui.add(TextEdit::singleline(&mut s).enabled(true));
            ui.add(TextEdit::singleline(&mut s).enabled(false));            
            ui.add(TextEdit::singleline(&mut s).enabled(true));
        }
);
````
By repeatedly pressing tab, it does not work to focus the third textedit (because the second one is disabled). Expected result is that focus simply skips the disabled widget and just moves between the first and third widgets.

I started looking at a fix, hope to get a PR ready in a few days the latest.

### PR DESCRIPTION
**Pull request by spersson**
Fixes #732. Ui::interact was being called twice for the frame rect
regardless of enabled status which was causing problems for kb focus.
Now the interact function is called zero or one time.
#Before
Screenshot 2021-09-28 123605
#After
Screenshot 2021-09-28 124443

The change in looks is because widget visuals now used for the disabled textedit is the noninteractive version, which I'm thinking is actually correct.

Closes <https://github.com/emilk/egui/issues/732>.

### DIFF
--- a/egui/src/widgets/text_edit.rs
+++ b/egui/src/widgets/text_edit.rs
@@ -417,7 +417,7 @@ impl<'t, S: TextBuffer> Widget for TextEdit<'t, S> {
         let frame_rect = response.rect.expand2(margin);
         ui.allocate_space(frame_rect.size());
         if enabled {
-            response = response | ui.interact(frame_rect, id, Sense::click());
+            response |= ui.interact(frame_rect, id, Sense::click())
         }
         if response.clicked() && !response.lost_focus() {
             ui.memory().request_focus(response.id);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: egui/src/widgets/text_edit.rs [hunk 1]
```diff
@@ -406,24 +406,26 @@ impl<'t, S: TextBuffer> TextEdit<'t, S> {
...
-        let response = self.content_ui(&mut content_ui);
+        let mut response = self.content_ui(&mut content_ui);
         let id = response.id;
         let frame_rect = response.rect.expand2(margin);
-        ui.allocate_rect(frame_rect, Sense::hover());
-        let frame_response = ui.interact(frame_rect, id, Sense::click());
-        let response = response | frame_response;
+        ui.allocate_space(frame_rect.size());
+        if enabled {
+            response = response | ui.interact(frame_rect, id, Sense::click());
```
emilk, line 420:
```suggestion
            response |= ui.interact(frame_rect, id, Sense::click())
```

to avoid clippy's wrath

emilk (approved):
Thanks for the fix!

Really the visual should be changed to match that of `{ ui.set_enabled(false); ui.text_edit(…) }`, but that's for another PR.

emilk (approved):

Issue #733, opened 2021-09-21
Separate serde derives from persistence feature


jparoz:
I'm currently working on a small pet project and trying out egui (using eframe) while doing so. I'm making a `Widget` which has a user-configurable `Color32` value, which is working great. I need to serialise this widget, and I am able to do this easily by enabling the `persistence` feature in the dependencies of my `Cargo.toml`.

However, I don't really want all the built-in persistence features. I don't want the window to move to its last recorded place, I don't need my side panel to resize to the previous state, all I want is to turn on the `#[derive(Serialize, Deserialize)]` for `Color32` (and other types). It seems to me that it's currently not possible. I could manually implement `Serialize` and `Deserialize` for my own structs containing an egui type, but this would be inconvenient and ugly compared to deriving.

What I would suggest is to have a new feature named `serde` (or similar), which enables the serde derives for each egui type. Then the existing `persistence` feature can enable the `serde` feature, as well as providing the existing automatic serialisation features.

Alternatively it would be fine to have some sort of eframe configuration to cause egui to not recall different parts of its state. This way I can enable `persistence`, and manually disable the things I don't want to persist.

emilk:
The feature `persistance` exists on both `egui` and `eframe`.

On `egui` it adds `#[derive(Serialize, Deserialize)]` on a bunch of types.
On `eframe` it does the actual persistence (and adds the feature flag to `egui`).

So if your `Cargo.toml` reads:

``` toml
egui = { version = "0.14", features = ["persistence"] }
eframe = { version = "0.14" }
```

It should work as you want it to work.

Reusing the same feature name for the two crates was convenient, but perhaps a bit confusing.

TimonPost:
If I understand the author accurately, he has no difficulties that it does not work just more that the 'persistence' feature flag is used for multiple purposes. 

1. The auto-saving of application state.
2. Implementing serde serialize functionality.

While I understand the initial choice, namely auto-save and persistence functionality occurs with a serialization process, I think it would be better to split the feature flag up. Something like this:

```toml
[features]
persitance = ["serde"]
serialize = []
```

With this, if you want to use serde only with `serialize`, not all persistence logic is immediately activated.

jparoz:
This makes sense. I agree that the naming is confusing, but I now see that the behaviour I wanted is basically already there, just using the one feature name under the two separate crates. I can work with this as a solution, thank you for the help @emilk. This could be closed, although I would be +1 to rename the features to be less confusing.

[...]

PR #753, opened 2021-09-28
Add separate serialize feature and better persitence control

emilk:
Closes https://github.com/emilk/egui/issues/733

* Add separate "serialize" features for `epaint`, `egui` and `egui_demo_lib`
* Use `cfg_attr(feature = "serde")` instead of `cfg_attr(feature = "persistence")` in most of the code
* Allow users control of whether or not egui memory and/or native window position and size gets persisted

PTAL @jparoz


jparoz:
Looks awesome, thanks a bunch!

PR #754, opened 2021-09-29
Collapsing header improvements

DrOptix:
This is a PR without a backing issue.

This PR changes `CollapsingHeader` container like this:
* `CollapsingHeader` is now opening or closing only when clicking the arrow icon
* `CollapsingHeader` is now filling all the available horizontal space
* Emit clicks on `CollapsingHeader`'s header only when NOT clicking the icon

Visual display:
collapsing_header_improvement

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ NOTE: [`eframe`](eframe/CHANGELOG.md), [`egui_web`](egui_web/CHANGELOG.md), [`eg
 * MSRV (Minimum Supported Rust Version) is now `1.54.0`.
 * By default, `DragValue`:s no longer show a tooltip when hovered. Change with `Style::explanation_tooltips`.
 * `CollapsingHeader` is now opening or closing only when clicking the arrow icon
+* `CollapsingHeader` is now filling all the available horizontal space
+
 
 ### Fixed 🐛
 * Fix wrongly sized multiline `TextEdit` in justified layouts.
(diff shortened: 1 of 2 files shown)


emilk:
I think being able to click the entire header should be the default behavior, just as it is the default behavior to be able to click the label of a checkbox or a radiobutton to toggle it. This new behavior is only useful for selectable collapsing headers, so I suggest we restrict it to those.

coderedart:
just a random suggestion: single click could select and double click (or long press/click) could be expand. kinda like reddit comment threads or some tree ui in android apps.

DrOptix:
> 
> 
> I think being able to click the entire header should be the default behavior, just as it is the default behavior to be able to click the label of a checkbox or a radiobutton to toggle it. This new behavior is only useful for selectable collapsing headers, so I suggest we restrict it to those.

Now that you mention it my target is to be able to create something like this for my project.
image

Now I'm not quite sure how to make it API wise so we can add custom controls to the header. In my case I would like to add a checkbox in front of the title and a button on the right side of the label as it can be seen in the screenshot above.

I'm thinking of adding something like this, but the drawback is that you need to paint the title yourself:
```rust
CollapsingHeader::new("title").show_custom_header(|title, ui| {
    // Show checkbox

    // Show title

    // On the right side of the header, show button for actions popup menu
});
```

emilk:
@DrOptix I think the best solution for that screenshot is to refactor `CollapsingHeader` so one can have a `CollapsingSection` without a header, controlled by a button. This is sort of how `Window` implements collapsing.

emilk:
I found a nicer approach in https://github.com/emilk/egui/pull/1538

PR #779, opened 2021-10-09
Code samples in demo

emilk:
I thought something like this would be very nice to show to beginners:

code-samples

However, the clever macro tricks I use to get the source code excerpts have some problems: `file!()` gives us a relative path when running `cargo check`, but an absolute path when running `cargo test`.

I may have to look for another way to accomplish this.


ryo33:
I've opened a topic about the `file!()` behavior at
https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/file!.28.29.20output.20varies.20in.20condition/near/313877507

@emilk If you have any additions or additional discussion, I would be happy if you would join.

PR #784, opened 2021-10-11
Implement `bytemuck` for `Rgba`

Stock84-dev:
Forgot to implement it on `Rgba`.
Relevant PR: #775


emilk:
rebase on latest `master` and run `cargo check` locally and you should find the problem!

PR #791, opened 2021-10-12
egui-winit: Emulate mouse from touch events

TobTobXX:
### About:
Currently, touch hardware can't interact with ui when using `egui-winit`, because there are no `egui::Event::PointerButton` and similar events emitted. This PR emits (additionally to the touch events) also the corresponding mouse events and thus emulates a mouse.

### Implementation details:
At first I just blindly constructed the events and pushed them to the queue, but I realized that a lot of this code already exists in the functions `on_mouse_button_input()` and `on_cursor_moved()`. This creates less code (and more readable IMHO), but creates some hairy situations, like the one with `self.simulate_touch_screen`.

### Considerations:
- **when touchscreen emulation is on:** the functions `on_mouse_button_input()` and `on_cursor_moved()` create additional touch events. Due to the native touch events, we'd have duplicate touch events.
- **Choice of primary mouse button:** Is it a problem if I just assume a touch corresponds to a left-click, or do I need to make sure `PointerButton::Primary` gets emitted (which it does now, but may be considered implementation-specific).

Also this is my first PR, so I have no idea ... :man_shrugging:

### DIFF
--- a/egui-winit/src/lib.rs
+++ b/egui-winit/src/lib.rs
@@ -106,7 +106,14 @@ pub struct State {
 
     /// If `true`, mouse inputs will be treated as touches.
     /// Useful for debugging touch support in egui.
+    ///
+    /// Creates duplicate touches, if real touch inputs are coming.
     simulate_touch_screen: bool,
+
+    /// Is Some(…) when a touch is being translated to a pointer.
+    ///
+    /// Only one touch will be interpreted as pointer at any time.
+    pointer_touch_id: Option<u64>,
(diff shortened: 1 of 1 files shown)


TobTobXX:
Nevermind, it was actually really easy lol.

Rebased & squashed commits. I think it would make sense to leave two commits, but just say if I should squash them.

Touch in the egui_demo_app works as intended on the Pinephone. (Although expect further PRs, window size seems to be incorrect.)

Issue #664, opened 2021-08-27
Consider using different monospace font for demos


athre0z:
Hey there & thanks for creating egui! I've played around with it quite a bit now and came to love it already. 

Recently, I created a small split view application for a toy programming language: half of the screen an editor, the other half an AST view & built it as a wasm application. I then linked it a to a bunch of friends and a common complaint that I got back was about the "terrible monospace font rendering". It appears that, from the (by design) blocky font, people incorrectly inferred that egui simply sucks at rendering monospace. I'm aware of the font-rendering issues in pre-0.14 versions on web, but when asked, people explicitly pointed out that it was the monospace stuff in particular that they disliked. I also remember that this was my first thought when first looking at the markdown editor in the egui demo app.

<img width="1645" alt="Screen Shot 2021-08-28 at 01 48 17" src="https://user-images.githubusercontent.com/6553158/131198761-726b57d1-182d-4255-b91d-9fe9ba632961.png">

While I realize that this is probably a bit controversial and against your personal preferences, I'd suggest changing the default font to a more "standard" looking monospace font.

emilk:
I think this is a good idea - let's collect some suggestions!

The font must have a liberal license (e.g. Open Font License), and the font file should be small (a few hundred kB max).

Nominate some suggestions, ideally with a screenshot, a link and a summary of the license and file size, and then we can pick one!

athre0z:
That's good to hear!vMy personal favorites are these two:

### JetBrains Mono

**License:** SIL Open Font License 1.1
**File size:** 144 KiB (`JetBrainsMonoNL-Regular.ttf`, NL = no ligatures)
**Website:** https://www.jetbrains.com/lp/mono/

<img width="330" alt="Screen Shot 2021-08-28 at 16 44 58" src="https://user-images.githubusercontent.com/6553158/131221816-6695b609-4ae9-44f2-9136-890c4b139b08.png">

### Fira Code

**License:** SIL Open Font License 1.1
**File size:** 299 KiB (`FiraCode-Regular.ttf`)
**Website:** https://github.com/tonsky/FiraCode

<img width="330" alt="Screen Shot 2021-08-28 at 16 47 07" src="https://user-images.githubusercontent.com/6553158/131221870-d33b5fcc-2515-44cf-9129-5d5da0ebda47.png">

[...]

PR #802, opened 2021-10-16
Replace ProggyClean with Hack-Regular as default monospace font

emilk:
Closes https://github.com/emilk/egui/issues/664

<https://github.com/source-foundry/Hack>, [MIT Licence](https://github.com/source-foundry/Hack/blob/master/LICENSE.md)

309 kB, half that when compressed.

Quite a bit more than the 40 kB of ProggyClean, but it looks really nice imho.

Screen Shot 2021-10-16 at 15 56 43
Screen Shot 2021-10-16 at 15 56 21


parasyte:
Looks like this font scales better than ProggyClean, so I think it's worth the added weight in file size.

emilk:
Let's use this for now, at least until someone finds something smaller!

PR #816, opened 2021-10-18
Add clear to webgl1 and webgl2 backends

### LINKED ISSUE #815: egui_web WebGL2 backend Doesn't Clear Internal Buffer
**Describe the bug**

The WebGL2 backend's srgb processing facilities use an internal buffer that it never clears. This doesn't manifest on the web example as it always repaints the entire frame.

**To Reproduce**
1. Use egui-web with an egui load that only uses part of the screen.

I unfortunately can't provide a runnable example, but I'm hoping the fix for this is simple.

**Expected behavior**

It should always clear the backbuffer to (0.0, 0.0, 0.0, 0.0) every frame.

**Screenshots**

image

**Desktop:**
 - OS: Windows
 - Browser Chrome
 - Version 98

cwfitzgerald:
Closes #815.

Added an explicit clear_color in case the user changes it from the default elsewhere (yay gl state machine).

Any chance of getting this out in a patch release? If not it's not a terribly big deal, it'd just be nice to have.

### DIFF
--- a/egui_web/src/webgl1.rs
+++ b/egui_web/src/webgl1.rs
@@ -666,6 +666,8 @@ impl PostProcess {
         }
 
         gl.bind_framebuffer(Gl::FRAMEBUFFER, Some(&self.fbo));
+        gl.clear_color(0.0, 0.0, 0.0, 0.0);
+        gl.clear(Gl::COLOR_BUFFER_BIT);
 
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


emilk:
Thanks!

> Any chance of getting this out in a patch release?

I tend to avoid patch releases, but don't worry - the next release is hopefully coming this weekend.

You can always do manual patching in the meantime by adding this to your `Cargo.toml`:

```toml
[patch.crates-io]
egui_web = { git = "https://github.com/emilk/egui", branch = "master" }
```

PR #822, opened 2021-10-20
Use the same GLSL shader for glium and glow

AlexApps99:
This simplifies the GLSL shaders so there are two total shaders between `egui_glium` and `egui_glow`.

I have tested it on my machine, but as this is affecting the default backend, it would be nice for this to be tested in a variety of configurations before merging.


emilk:
The glium version still works on my intel Mac

emilk:
I'm not sure if it is worth reusing the shader files. It's not that much code, and sharing can make it more brittle. In any case, this PR has been unattended for too long so I will close it for now.

Issue #827, opened 2021-10-20
Add method to set outer window title dynamically


NiceneNerd:
It would be useful to be able to dynamically update the title of the outer (often native) window through the `Frame` API. It could work the same way as `Frame::set_window_size()`. I've already done the work for it, so I'll submit a PR in a second.

PR #828, opened 2021-10-20
Add `Frame::set_window_title()`

NiceneNerd:
Adds the method `Frame::set_window_title()` so outer window titles can be changed dynamically. By consequence `Copy` has been removed from `AppOutput`, which is instead cloned in precisely two places.

Closes <https://github.com/emilk/egui/issues/827>.


NiceneNerd:
Ah, I was indeed uncertain which changelog would be the best to use. Both problems are now sorted.

JPGygax68:
What is the best place to call this from ? I haven't found a better hook than `update()`, is that ok ?

PR #829, opened 2021-10-20
Keep the plot hover text within the plot

piotrek124-1:
Before 

Zrzut ekranu z 2021-10-20 22-54-25_
Zrzut ekranu z 2021-10-20 22-54-34_
After
Zrzut ekranu z 2021-10-20 22-52-35_
Zrzut ekranu z 2021-10-20 22-53-09_
Zrzut ekranu z 2021-10-20 22-53-17_


emilk:
This PR hard-codes the text size, which is not very good. I think it would be nicer to instead have the hover text show up in a separate `Area`, similar to a tooltip (but one that follows the mouse around).

PR #830, opened 2021-10-21
egui_web: constrain the IME text agent to the canvas

sumibi-yakitori:
If you enter a large number of line breaks where the height of `TextEdit` exceeds the height of the browser window, and then enter more non-line breaks, the egui screen will be driven out of the client area

~~Also, [since scrolling is currently disabled in a browser itself](https://github.com/emilk/eframe_template/blob/77626b9ec33e0351ce741c883393195b601f477a/docs/index.html#L25), there doesn't seem to be much benefit in changing the position of the `text_agent`~~

I apologize for my terrible English

### DIFF
--- a/egui_web/src/lib.rs
+++ b/egui_web/src/lib.rs
@@ -1224,41 +1224,39 @@ fn update_text_agent(runner: &AppRunner) -> Option<()> {
     Some(())
 }
 
-// const MOBILE_DEVICE: [&str; 6] = ["Android", "iPhone", "iPad", "iPod", "webOS", "BlackBerry"];
-// /// If context is running under mobile device?
-// fn is_mobile() -> Option<bool> {
-//     let user_agent = web_sys::window()?.navigator().user_agent().ok()?;
-//     let is_mobile = MOBILE_DEVICE.iter().any(|&name| user_agent.contains(name));
-//     Some(is_mobile)
-// }
+const MOBILE_DEVICE: [&str; 6] = ["Android", "iPhone", "iPad", "iPod", "webOS", "BlackBerry"];
+/// If context is running under mobile device?
+fn is_mobile() -> Option<bool> {
+    let user_agent = web_sys::window()?.navigator().user_agent().ok()?;
+    let is_mobile = MOBILE_DEVICE.iter().any(|&name| user_agent.contains(name));
+    Some(is_mobile)
+}
 
 // Move text agent to text cursor's position, on desktop/laptop,
 // candidate window moves following text element (agent),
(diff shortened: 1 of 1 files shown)


sumibi-yakitori:
Screen Shot 2021-10-31 at 12 51 16
Screen Shot 2021-10-31 at 12 51 50

egui_web/src/lib.rs
```diff
@@ -1207,35 +1207,41 @@ fn manipulate_agent(canvas_id: &str, latest_cursor: Option<egui::Pos2>) -> Optio
...
+    //         let y = y + (canvas.scroll_top() + canvas.offset_top()) as f32;
+    //         let x = x + (canvas.scroll_left() + canvas.offset_left()) as f32;
+
+    //         // Canvas is translated 50% horizontally in html.
+    //         let x = x - canvas.offset_width() as f32 / 2.0;
+    //         style.set_property("position", "absolute").ok()?
+    //         style.set_property("top", &(y.to_string() + "px")).ok()?;
+    //         style.set_property("left", &(x.to_string() + "px")).ok()
+    //     })
+    // } else {
```
emilk, line 1242: did you think about what this code was being used for? have you tested that IME still works as expected?

sumibi-yakitori: Oh sorry, I must have misunderstood a bit. I feel that by suppressing the position of the text agent from almost sticking out of the client area of the browser, the position of the IME will also stay in line with the position of the text cursor. Is this a good fix?

emilk: @n2 wrote the original code, perhaps they have some thoughts?

sumibi-yakitori:
@emilk 
I'd like to hear what @n2 thinks, but for now, I've reverted the previous wrong code and added the code to restrict the position of the text agent to within the client area.

I've also checked the IME position tracking. There is a subtle difference in the IME position tracking compared to the browser native text input, but it seems to be a practical problem.

sumibi-yakitori:
Note: For some reason the CI test was failing, so I rebased to the latest master branch. When I did that, the old commits you had linked to the review comments disappeared.

n2:
As mentioned in the comments, move_text_cursor() is used to move IME candidate window following egui's cursor. It is better to constrain the position within the canvas, rather than just delete the function. I think your newly submitted code is suitable. Thank you for fixing the overlooked part of my code. @sumibi-yakitori

emilk (approved):
Nice, that looks like a much better solution to me!

Thanks for your thoughts, @n2 !

PR #832, opened 2021-10-22
Add Button::image_and_text

d10sfan:
Been using egui with the sdl2 backend for a short while now and it's been a great experience, making UI has been nicely easy with it after learning it.

One of the things I was wanted to do which the integration with sdl2 made possible was to have a hybrid mouse and controller UI, where someone could navigate and press buttons using the controller as well.

Part of that I wanted to be able to have a button with a controller icon in it, to show which one could trigger it. I created a "ImageButtonWithText", which tried to merge the ideas of an ImageButton and a Button, similar to how the Radio Button or Checkbox ones work.

This was mainly copying and pasting the Button implementation and adding the necessary spacing and positions for putting the image and text inside the button. It worked well visually, as can be seen in the screenshot below, which was built using this idea, although not sure if it may be better to add this as an additional feature to the "Button" instead of it being its own Button type.

The code to add this type looks like this, assuming that "texture_confirm" was already created (same idea behind adding an Image to the ui).

```
if ui.add(egui::ImageButtonWithText::new("Ok", texture_confirm, egui::vec2(32 as f32, 32 as f32))).clicked() {
    ok = true;
}
```

image

### DIFF
--- a/egui/src/widgets/button.rs
+++ b/egui/src/widgets/button.rs
@@ -166,9 +166,8 @@ impl Widget for Button {
         desired_size = desired_size.at_least(min_size);
 
         if let Some(image) = image {
-            let image_padding = image.size() + 4.0 * button_padding;
-            desired_size.x += image_padding.x;
-            desired_size.y = image_padding.y;
+            desired_size.x += image.size().x + ui.spacing().icon_spacing;
+            desired_size.y = desired_size.y.max(image.size().y + 2.0 * button_padding.y);
         }
 
         let (rect, response) = ui.allocate_at_least(desired_size, sense);
--- a/egui/src/widgets/button.rs
+++ b/egui/src/widgets/button.rs
@@ -29,6 +29,7 @@ pub struct Button {
     small: bool,
     frame: Option<bool>,
     min_size: Vec2,
+    image: Option<widgets::Image>,
 }
 
 impl Button {
@@ -42,6 +43,27 @@ impl Button {
             small: false,
             frame: None,
             min_size: Vec2::ZERO,
+            image: None,
+        }
+    }
+
+    /// Creates a button with an image to the left of the text. The size of the image as displayed is defined by the size Vec2 provided.
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn image_and_text(
+        texture_id: TextureId,
+        size: impl Into<Vec2>,
+        text: impl Into<WidgetText>,
+    ) -> Self {
+        Self {
+            text: text.into(),
+            fill: None,
+            stroke: None,
+            sense: Sense::click(),
+            small: false,
+            frame: None,
+            wrap: None,
+            min_size: Vec2::ZERO,
+            image: Some(widgets::Image::new(texture_id, size)),
         }
     }
 
@@ -123,6 +145,7 @@ impl Widget for Button {
             small,
             frame,
             min_size,
+            image,
         } = self;
 
         let frame = frame.unwrap_or_else(|| ui.visuals().button_frame);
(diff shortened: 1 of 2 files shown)


egui/src/widgets/button.rs
```diff
@@ -55,6 +56,28 @@ impl Button {
...
             min_size: Vec2::ZERO,
+            image: None,
+        }
+    }
+
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn image_and_text(
+        texture_id: TextureId,
+        size: impl Into<Vec2>,
+        text: impl ToString,
```
emilk, line 67:
```suggestion
        text: impl Into<WidgetText>,
```

(please rebase on latest `master`)

egui/src/widgets/button.rs
```diff
@@ -55,6 +56,28 @@ impl Button {
             frame: None,
             wrap: None,
             min_size: Vec2::ZERO,
+            image: None,
+        }
+    }
+
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn image_and_text(
```
emilk, line 64: Add a docstring explaining where the image will be shown (to the left of the text?)

egui/src/widgets/button.rs
```diff
@@ -165,16 +189,30 @@ impl Widget for Button {
...
                 .align_size_within_rect(galley.size(), rect.shrink2(button_padding))
                 .min;
 
+            if let Some(image) = image {
+                let icon_spacing = ui.spacing().icon_spacing;
+                text_pos = pos2(
+                    rect.min.x + button_padding.x + image.size().x + icon_spacing,
+                    rect.center().y - 0.5 * galley.size().y,
+                );
+            }
```
emilk, line 214: `let text_pos = if let Some(image) = image { … } else { … };` is a nicer pattern

emilk (changes requested):
It would be nice to demonstrate this, perhaps in `egui_glium/examples/native_texture.rs` (which has an actual texture loaded). `if ui.button("Quit").clicked() {` could be replaced with the new `Button::image_and_text`

egui/src/ui.rs
```diff
@@ -1108,6 +1108,22 @@ impl Ui {
...
+    #[must_use = "You should check if the user clicked this with `if ui.button_with_image(…).clicked() { … } "]
+    #[inline(always)]
+    pub fn button_with_image(
+        &mut self,
+        texture_id: TextureId,
+        size: impl Into<Vec2>,
+        text: impl ToString,
+    ) -> Response {
+        Button::image_and_text(texture_id, size, text).ui(self)
+    }
```
emilk, line 1125: I'm not sure `Button::image_and_text` will be used enough to warrant a helper in `Ui`. I try to limit that to only the most commonly used things (there isn't even a `ui.slider`!)

d10sfan: Ok, I've removed this one

egui/src/widgets/button.rs
d10sfan: I've changed to match this pattern.

egui/src/widgets/button.rs
d10sfan: Added a string above the function to describe how it works. Let me know if anything additional is needed.

egui/src/widgets/button.rs
d10sfan: Re-based to use the latest master version.

d10sfan:
I re-based with the latest and removed the ui... function.

I added the change to the example glium but not sure how to test it, I didn't see a way to run it like you can with the demo, but it should work.

emilk:
`cargo run --example native_texture`

egui/src/widgets/button.rs
```diff
@@ -142,15 +165,28 @@ impl Widget for Button {
         }
         desired_size = desired_size.at_least(min_size);
 
+        if let Some(image) = image {
+            let image_padding = image.size() + 4.0 * button_padding;
+            desired_size.x += image_padding.x;
+            desired_size.y = image_padding.y;
```
emilk, line 171:
```suggestion
            desired_size.x += image.size().x + ui.spacing().icon_spacing;
            desired_size.y = desired_size.y.max(image.size().y + 2.0 * button_padding.y);
```

emilk:
Screen Shot 2021-11-08 at 21 34 57

The composition is not great. Both the image and the text is too far to the left

egui/src/widgets/button.rs
emilk: why 4x `button_padding` for images?

emilk: You use `icon_spacing` below, but not here - that's inconsistent.

emilk (changes requested):

egui/src/widgets/button.rs
d10sfan: Thanks, I've updated per the suggested change. I believe I was trying to go based off of the padding of the button frame and then extend that, based on some of the ways the radio and other buttons that had text near them, but that didn't work as seen by the shifting too far to the left. The new calculation does look much better.

image

image

emilk (approved):

PR #857, opened 2021-10-30
Context Menu: Return owned Response from Response::context_menu

mankinskin:
Just a minor inconvenience I noticed, `response.context_menu()` returns a `&Response`. This can lead to annoying reference type errors because you may not be able to just move out of the reference.

I think I did this in the first place to avoid having to write `let response = response.context_menu();` but I don't think this is a concern actually.


emilk (approved):
Good catch!

PR #863, opened 2021-11-01
Bar charts + box plots continued

Bromeon:
The addition of box/bar charts in https://github.com/emilk/egui/pull/499 looks very valuable to me. Unfortunately the original PR seems to have been abandoned and has caused several merge conflicts in the meantime. This PR adds commits on top to update to egui's latest API. In case #499 is continued, I'm happy to close this and let my changes be integrated there 🙂 

---

Now there are some open topics:

1. I didn't touch the existing commits to retain the author information; might be worth a squash. In case we rebase rather than merge latest `master` into the branch (as done now), we would probably require @niladic's and my changes to be combined into a single commit (if that's the case, `Co-authored-by` could help).

1. The "closest" algorithm is not yet integrated, as there wasn't yet any consensus on how to do it efficiently, so debug information for hovering over charts won't show. Also, is the idea that every plot type implements that algorithm on their own? Why not return something like an "outline shape" and compute close-ness based on that?

1. I wasn't sure about the `initialize()` methods. The trivial implementation would be the following, but is this meaningful?
   ```rs
   fn initialize(&mut self, x_range: RangeInclusive<f64>) {
       self.dummy_values.generate_points(x_range);
   }
   ```

1. Naming: the original PR uses very mixed `BarChart`, `barchart`, `Boxplot`, `boxplots` symbols. I would probably change this to use `PascalCase`/`snake_case` consistently.

1. Generally, I'm quite new to egui, so let me know where things can be improved!


emilk:
@EmbersArc what do you think?

EmbersArc (commented):
Thanks so much for reviving this PR!

I'm still not sure about the ergonomics of the stacked bar charts, but I'd have to use it to be able to tell really. I think once the dead code (e.g. `rulers`) is either removed or used in some way, we can merge this without issues.

EmbersArc:
>I wasn't sure about the initialize() methods. The trivial implementation would be the following, but is this meaningful?

No it's not. You can just leave this function empty. We can try to think about how to extend the `PlotItem` trait later so that we can make it public and users can create their own.

Bromeon:
Also, any advice regarding the previous commits? Should I squash them, use `Co-authored-by`, or something else?

EmbersArc:
>This is currently disabled, but I could find a way to re-enable it, as it seems very useful.

If I remember correctly, the previous PR would have added something like an `on_hover_shapes` to the `PlotItem` trait. The basic idea was that this would take the mouse coordinate and then return a vector of shapes that would show more information about the hovered item. I think this wasn't a bad design, so you could try adding that again. But feel free to experiment :)

Bromeon:
I managed to find some time to overhaul the existing implementation and clean up a lot. Bar charts + box plots are now working, and the hovering logic is running as well.

Unfortunately, during exactly that time  #766 has landed, and #892 is on the way too, both could create significant merge conflicts.
Let me know what would be a good strategy here.

My implementation is not perfect yet, but I'd rather create a follow-up PR than diverging more and risking even more merge conflicts.

Bromeon:
@EmbersArc Fixed the merge conflicts with current `master` branch.
This pull request should be ready for review now. ✔️ 

Worth mentioning: apart from the plots, I added the (private) `FloatOrd` trait, which allows to use `my_value.ord()` to quickly create a f32 value implementing `Ord`. This is needed for algorithms like `sorted()`, `min_by_key()` etc. Maybe others can benefit from this as well, but let me know if there's already an established way (or dependency) to achieve that.

I would like to express huge thanks to @niladic, who laid out the foundation and much of the work in his original PR. I made sure to attribute him as author of the original commit. 👍

egui/src/widgets/plot/items.rs
```diff
@@ -351,6 +352,14 @@ impl PlotItem for VLine {
     }
 }
 
+pub(super) struct HoverElem {
```
s-nie, line 355: Let's rename it to `HoverElement`

egui/src/widgets/plot/mod.rs
```diff
@@ -694,88 +723,31 @@ impl PreparedPlot {
...
-        }
-        if *show_y {
-            // horizontal line
-            shapes.push(Shape::line_segment(
-                [pos2(rect.left(), pointer.y), pos2(rect.right(), pointer.y)],
-                (1.0, line_color),
-            ));
-        }
+        let closest = candidates
+            .min_by_key(|(_, elem)| elem.dist_sq.ord())
```
s-nie, line 736: I don't think that `float_ord.rs` is really required. I'd just replace all `min_by_key` with `min_by`, for example here:
```rust
            .min_by(|(_, elem1), (_, elem2)| elem1.dist_sq.partial_cmp(&elem2.dist_sq).unwrap())
```

egui/src/widgets/plot/mod.rs
```diff
@@ -562,6 +563,34 @@ impl PlotUi {
...
+        }
+        self.items.push(Box::new(box_plot));
+    }
+
+    /// Add a bar chart.
+    /// You can add multiple such charts.
+    pub fn bar_chart(&mut self, mut chart: BarChart) {
+        if chart.bars.is_empty() {
+            return;
+        }
```
s-nie, line 586:
```suggestion
    /// Add a box plot diagram.
    pub fn box_plot(&mut self, mut box_plot: BoxPlot) {
        if box_plot.boxes.is_empty() {
            return;
        }

        // Give the elements an automatic color if no color has been assigned.
        if box_plot.default_color == Color32::TRANSPARENT {
            box_plot = box_plot.color(self.auto_color());
        }
        self.items.push(Box::new(box_plot));
    }

    /// Add a bar chart.
    pub fn bar_chart(&mut self, mut chart: BarChart) {
        if chart.bars.is_empty() {
            return;
        }
```

s-nie (changes requested):
Gave it a quick look. Looks very good! Just a couple smaller comments.

egui/src/widgets/plot/items.rs
EmbersArc, line 355: Let's rename it to `HoverElement`

egui/src/widgets/plot/mod.rs
```diff
@@ -694,88 +723,31 @@ impl PreparedPlot {
...
-        if *show_y {
-            // horizontal line
-            shapes.push(Shape::line_segment(
-                [pos2(rect.left(), pointer.y), pos2(rect.right(), pointer.y)],
-                (1.0, line_color),
-            ));
-        }
+        let closest = candidates
+            .min_by_key(|(_, elem)| elem.dist_sq.ord())
+            .filter(|(_, elem)| elem.dist_sq <= interact_radius_sq);
```
EmbersArc, line 737: I don't think that `float_ord.rs` is really required. I'd just replace all `min_by_key` with `min_by`, for example here:

` .min_by(|(_, elem1), (_, elem2)| elem1.dist_sq.partial_cmp(&elem2.dist_sq).unwrap())`

egui/src/widgets/plot/mod.rs
```diff
@@ -562,6 +563,34 @@ impl PlotUi {
...
+        // Give the elements an automatic color if no color has been assigned.
+        if box_plot.default_color == Color32::TRANSPARENT {
+            box_plot = box_plot.color(self.auto_color());
+        }
+        self.items.push(Box::new(box_plot));
+    }
+
+    /// Add a bar chart.
+    /// You can add multiple such charts.
+    pub fn bar_chart(&mut self, mut chart: BarChart) {
```
EmbersArc, line 583:
```suggestion
    /// Add a box plot diagram.
    pub fn box_plot(&mut self, mut box_plot: BoxPlot) {
        if box_plot.boxes.is_empty() {
            return;
        }

        // Give the elements an automatic color if no color has been assigned.
        if box_plot.default_color == Color32::TRANSPARENT {
            box_plot = box_plot.color(self.auto_color());
        }
        self.items.push(Box::new(box_plot));
    }

    /// Add a bar chart.
    pub fn bar_chart(&mut self, mut chart: BarChart) {
```

EmbersArc (changes requested):
Gave it a quick look. Looks very good! Just a couple smaller comments.

egui/src/widgets/plot/mod.rs
Bromeon: I was doing that first, but it's _much_ less readable and easier to get wrong, no?
```rs
.min_by(|(_, elem1), (_, elem2)| elem1.dist_sq.partial_cmp(&elem2.dist_sq).unwrap())
```
vs.
```rs
.min_by_key(|(_, elem)| elem.dist_sq.ord())
```
Also, we would need to do that for any sort/min/max related algorithm, so I was thinking it makes sense to abstract it once and reuse it? The idea would be that maybe other parts of egui could benefit as well (now or in the future).

egui/src/widgets/plot/items.rs
Bromeon: Will do 👍

Bromeon:
Thanks for the feedback!

Implemented two suggestions; only [the `ord()` topic](https://github.com/emilk/egui/pull/863#discussion_r751068298) left open until clarified.

egui/src/widgets/plot/items.rs
```diff
@@ -1446,3 +1560,947 @@ impl PlotItem for PlotImage {
...
+        self
+    }
+
+    /// Set the bar width.
+    pub fn width(mut self, width: f64) -> Self {
+        self.bar_width = width;
+        self
+    }
+
+    /// Set orientation of the element as vertical. Key axis is X.
```
emilk, line 1658: Here the word "key" is used, but in other places "argument" is used. It would be nice to be consistent.

egui/src/widgets/plot/items.rs
```diff
@@ -1446,3 +1560,947 @@ impl PlotItem for PlotImage {
...
+    pub fn stack_on(mut self, others: &[&BarChart]) -> Self {
+        for (index, mut bar) in self.bars.iter_mut().enumerate() {
+            if bar.value.is_sign_positive() {
+                let mut max = 0.0;
+                for other_chart in others {
+                    if let Some(other_bar) = other_chart.bars.get(index) {
+                        let other_upper = other_bar.upper();
+                        if other_upper > max {
+                            max = other_upper;
+                        }
```
emilk, line 1931:
```suggestion
                        max = other_bar.upper().max(max);
```

egui/src/widgets/plot/items.rs
```diff
@@ -1446,3 +1560,947 @@ impl PlotItem for PlotImage {
...
+                    bar.base_offset = Some(max);
+                }
+            } else {
+                let mut min = 0.0;
+                for other_chart in others {
+                    if let Some(other_bar) = other_chart.bars.get(index) {
+                        let other_lower = other_bar.lower();
+                        if other_lower < min {
+                            min = other_lower;
+                        }
```
emilk, line 1944:
```suggestion
                        min = other_bar.lower().min(min);
```

egui/src/util/float_ord.rs
```diff
@@ -0,0 +1,40 @@
+//! Total order on floating point types, assuming absence of NaN.
+//! Can be used for sorting, min/max computation, and other collection algorithms.
```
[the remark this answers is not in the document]
emilk: I can also help with this refactor after merge

egui/src/widgets/plot/items.rs
```diff
@@ -1446,3 +1560,947 @@ impl PlotItem for PlotImage {
...
+    pub quartile1: f64,
+
+    /// Value of middle line in box (typically median)
+    pub median: f64,
+
+    /// Value of upper box threshold (typically 75% quartile)
+    pub quartile3: f64,
+
+    /// Value of upper whisker (typically maximum)
+    pub upper_whisker: f64,
```
emilk, line 2035: Perhaps these should be combined into a `[f64; 5]` here and in `new`. This communicates how they are all closely related (on the same dimension, and should be ordered in an increasing order). Or they can be extracted to a `struct BoxSpread` to keep down the complexity of `BoxElem`.

egui/src/widgets/plot/items.rs
```diff
@@ -1446,3 +1560,947 @@ impl PlotItem for PlotImage {
...
+        .into_iter()
+        .enumerate()
+        .map(|(index, bar)| {
+            let bar_rect: Rect = transform.rect_from_values(&bar.bounds_min(), &bar.bounds_max());
+            let dist_sq = point.distance_from_rect_sq(bar_rect);
+
+            HoverElement { index, dist_sq }
+        })
+        .min_by_key(|e| e.dist_sq.ord())
+}
```
emilk, line 2506: There's a lot of new code here. I think it would make sense to add them to new files (maybe `egui/src/widgets/plot/items/bar.rs` and `egui/src/widgets/plot/items/box.rs`)

egui/src/widgets/plot/mod.rs
emilk: I agree with @Bromeon here, especially if we make the `OrderedFloat` also handle `NaN`. Then it is strictly better (avoids panics).

egui/src/widgets/plot/transform.rs
```diff
@@ -217,6 +224,16 @@ impl ScreenTransform {
         Value::new(x, y)
     }
 
+    pub fn rect_from_values(&self, value1: &Value, value2: &Value) -> Rect {
+        let pos1 = self.position_from_value(value1);
+        let pos2 = self.position_from_value(value2);
+        // Note that the transform mirrors rects: top becomes bottom
```
emilk, line 230: I don't understand what this comment is trying to communicate about the surrounding code

emath/src/pos2.rs
```diff
@@ -136,6 +136,25 @@ impl Pos2 {
...
+        };
+        let dy = if rect.min.y > self.y {
+            rect.min.y - self.y
+        } else if self.y > rect.max.y {
+            self.y - rect.max.y
+        } else {
+            0.0
+        };
+        dx * dx + dy * dy
+    }
```
emilk, line 156: This belongs in `Rect` instead: I've structured emath so that lower types (Pos2) never mentions higher types (Rect) in their implementations.

i.e. `rect.distance_sq_to(pos)`

emilk (changes requested):
Looks really good to me! I agree there should be a demo of the "ruler" code.

egui/src/widgets/plot/items.rs
```diff
@@ -372,6 +427,17 @@ struct ExplicitGenerator {
     points: usize,
 }
 
+/// Query the values of the plot, to do e.g. closest checks
+pub(super) enum PlotGeometry<'a> {
```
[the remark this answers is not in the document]
Bromeon: Ok, will also do this for `PlotItem`, which is currently defined after some of its `impl`s.

egui/src/widgets/plot/transform.rs
Bromeon: My guess of the original author's idea would be: the coordinate system for egui has +Y down, but for plots +Y is up. So a rectangle based on `Value`s will be mirrored up-down.

Clarified with Rustdoc:
```rs
/// Transform a rectangle of plot values to a screen-coordinate rectangle.
///
/// This typically means that the rect is mirrored vertically (top becomes bottom and vice versa),
/// since the plot's coordinate system has +Y up, while egui has +Y down. 
pub fn rect_from_values(&self, value1: &Value, value2: &Value) -> Rect { ... }
```

emath/src/pos2.rs
Bromeon: Done.

egui/src/util/float_ord.rs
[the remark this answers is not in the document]
Bromeon: Thanks! I'll gladly open a new PR after this one, but probably makes sense to keep it out of this (already big) changeset.

Bromeon (commented):
Thanks for the feedback!

Not yet all points addressed, will continue tomorrow.

emilk (approved):

egui/src/widgets/plot/items.rs
```diff
@@ -372,6 +427,17 @@ struct ExplicitGenerator {
...
+    None,
+
+    // note: takes the Vec inside 'Values', but already filled after generator has run
```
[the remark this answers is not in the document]
Bromeon: Clarified the comments.

egui/src/widgets/plot/items.rs
Bromeon: Fixed.

Bromeon:
Merged latest master (some conflicts due to `Bounds` -> `PlotBounds` etc. in #892) and implemented requested changes.

@emilk As you noted correctly, the file `items.rs` has grown quite a bit. I turned it into a module folder, with several files for bars/boxes. I also outsourced all the "helper" types which are needed for plots and values -- but not directly related to `PlotItem` -- to their own module `values`. Now, `items` contains mostly the different plot types and their respective `impl PlotItem`.

>  I agree there should be a demo of the "ruler" code.

What do you mean exactly? Also, if it's a feature, would it be possible to do it in another PR? I've just resolved the third big merge conflict due to branch divergence, and with the new file structure there's quite some risk for nasty conflicts 😀

Bromeon:
Thanks a lot to both of you for review and feedback! It's good to have the initial implementation merged, there might be some API refinement once users get their hands on.

In the meantime I moved ahead with OrderedFloat in #918.

Issue #854, opened 2021-10-29
Automatically request repaint when Area size changed?


m-ou-se:
I'm using an `Area` to center some ui, with `.anchor(CENTER_CENTER, ZERO)`, which mostly works very well, except the first frame it shows up in the wrong place. This is expected, as only afterwards can it remember the size of the contents and calculate the right position for the ui to appear centered.

However, this first frame with the wrongly positioned center area stays visible until some kind of input event is generated.

For now I worked around this by using:

```rust
            let r = Area::new("...").anchor(Align2::CENTER_CENTER, Vec2::ZERO).show(...);

            if (r.response.rect.center() - ctx.available_rect().center()).length_sq() > 1.0 {
                ctx.request_repaint();
            }
```

Would it make sense for `egui::Area` to automatically request a repaint when after adding the contents the size turned out to be different than what it assumed while chosing the position? (Or maybe some functionality on `Response` to easliy opt-in to that behaviour?)

emilk:
Related: https://github.com/emilk/egui/issues/843

PR #866, opened 2021-11-01
Area: automatically request a repaint when showing up for the first time

emilk:
Closes https://github.com/emilk/egui/issues/854

@m-ou-se can you please verify that this fixes your issue?

### DIFF
--- a/egui/src/containers/area.rs
+++ b/egui/src/containers/area.rs
@@ -204,6 +204,9 @@ impl Area {
 
         let state = ctx.memory().areas.get(id).cloned();
         let is_new = state.is_none();
+        if is_new {
+            ctx.request_repaint(); // if we don't know the previous size we are likely drawing the area in the wrong place}
+        }
         let mut state = state.unwrap_or_else(|| State {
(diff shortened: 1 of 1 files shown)


m-ou-se:
Hey, sorry for my late reply and thanks for the fix. I'll test it later today!

m-ou-se:
It doesn't seem to change anything for me, and I might have misunderstood the exact problem behind the issue I reported in #854.

One problem appears to be that when using anything like `ui.with_layout(Layout::top_down(Align::center), ..)` inside a window/area, the area directly expands to the full screen width. The jumping I see then after the first redraw seems to be very small, maybe margin/padding on the left of the area disappearing only after it takes up the full screen width.

Whether `ui.separator()` or a centered layout should expand windows like that is more of a fundamental question I suppose, and could only be changed by refactoring the whole `min_rect`/`max_rect` system. (I think?)

And another fundamental question around the problem I saw is whether egui should always keep redrawing until it reaches a steady state/fixed point. (Like LaTeX I suppose? ^^') (E.g. if something needs to be centered but turns out was drawn in the wrong place, always directly redraw with the new width information.)

PR #868, opened 2021-11-02
Make egui_glow painter to work on web

t18b219k:
I made PR #852 but this way affect many points.
These my works
* eliminate glutin dependency from painter
* port PostProcess for sRGB correct color for web from egui_web painter
* eliminate VAO by emulating VAO  
* add feature flag `painter_only` to reuse glow painter
* implement egui_web::Painter trait for wrapped glow painter
* share GLSL ES and GLSL shader
Current tested platforms
* OpenGL 4.6 (Ryzen 5 5500U with AMD proprietary windows driver) (not tested but coded to work 2.0 or above)
* WebGL2 (firefox android , MS edge)
* WebGL1+sRGB (WebGL2 disabled firefox)
* WebGL1 (servo nightly build)


t18b219k:
I monitored by Spector.js and found glow painter 's draw call is half of egui_web.
which means lower CPU usage,lower power consumption,higher frame rate.
* glow painter 132
*  webgl2 painter  269
*

egui_glow/src/misc_util.rs
```diff
@@ -0,0 +1,129 @@
+#![allow(unsafe_code)]
+use glow::HasContext;
+#[cfg(target_arch = "wasm32")]
+use wasm_bindgen::JsValue;
+
+pub(crate) fn srgbtexture2d(
+    gl: &glow::Context,
+    compatibility_mode: bool,
```
emilk, line 8: this needs a docstring (compatibility with what?)

egui_glow/src/painter.rs
```diff
@@ -268,16 +195,26 @@ impl Painter {
         if self.egui_texture_version == Some(texture.version) {
             return; // No change
         }
-
+        let gamma = if self.post_process.is_none() {
+            1.0 / 2.0
```
emilk, line 199: Likely `1.0 / 2.2` will look slightly better (better approximates the sRGB gamma curve)

egui_web/src/backend.rs
```diff
@@ -223,8 +239,12 @@ impl AppRunner {
     }
 
     fn integration_info(&self) -> epi::IntegrationInfo {
+        #[cfg(not(feature = "use_glow_painter"))]
+        const NAME: &str = "egui_web";
+        #[cfg(feature = "use_glow_painter")]
+        const NAME: &str = "egui_web(painted by glow)";
         epi::IntegrationInfo {
-            name: "egui_web",
+            name: NAME,
```
emilk, line 247: we could add a new field `painter: &str` to `IntegrationInfo` and set to either `webgl1`, `webgl2` or `glow` (`painter: self.painter.name()` if we add a `fn name(&self) -> &'static str` to `trait Painter`).

egui_web/src/backend.rs
```diff
@@ -15,14 +15,30 @@ impl WebBackend {
...
 
-        let painter: Box<dyn Painter> =
+        #[cfg(feature = "use_glow_painter")]
+        let painter: Box<dyn Painter> = {
+            let canvas = canvas_element_or_die(canvas_id);
+            let gl_ctx =
+                egui_glow::create_context_for_canvas::init_glow_context_from_canvas(&canvas);
+            let dimension = [canvas.width() as i32, canvas.height() as i32];
+            let painter = egui_glow::Painter::new(&gl_ctx, Some(dimension));
+            Box::new(crate::glow_wrapping::WrappedGlowPainter {
```
emilk, line 25: this would be nicer as a constructor of `WrappedGlowPainter`, so the call here would just be 

`Box::new(crate::WrappedGlowPainter::new(canvas_id))`

Then the fields of `WrappedGlowPainter` could also be private!

egui_web/Cargo.toml
```diff
@@ -44,7 +45,7 @@ default = ["default_fonts"]
 default_fonts = ["egui/default_fonts"]
 persistence = ["egui/persistence", "ron", "serde"]
 screen_reader = ["tts"] # experimental
-
+use_glow_painter =["egui_glow/painter_only","egui_glow/epi"]
```
emilk, line 48: I think just `glow` or `glow_painter` is a better name.

Please don't remove the blank lines - I think they aid in the readability of a file. If you don't like them, configure your editor to don't show them or something :)

egui_glow/src/vao_emulate.rs
```diff
@@ -0,0 +1,59 @@
...
+
+pub(crate) struct BufferInfo {
+    pub location: u32, //
+    pub vector_size: i32,
+    pub data_type: u32, //GL_FLOAT,GL_UNSIGNED_BYTE
+    pub normalized: bool,
+    pub stride: i32,
+    pub offset: i32,
+}
+pub struct EmulatedVao {
```
emilk, line 12: Is there any risk that the "emulated" VAO is slower than a real VAO?

egui_glow/src/vao_emulate.rs
```diff
@@ -0,0 +1,59 @@
...
+    buffer: glow::Buffer,
+    buffer_infos: Vec<BufferInfo>,
+    taught: bool,
+}
+impl EmulatedVao {
+    #[allow(clippy::needless_pass_by_value)]
```
emilk, line 18:
```suggestion
```

I don't think this is needed, is it?

egui_glow/src/vao_emulate.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub(crate) fn add_new_attribute(&mut self, buffer_info: BufferInfo) {
+        self.buffer_infos.push(buffer_info);
+    }
+    pub(crate) fn bind_vertex_array(&self, gl: &glow::Context) {
+        unsafe {
+            gl.bind_buffer(glow::ARRAY_BUFFER, Some(self.buffer));
+        }
+        for attribute in self.buffer_infos.iter() {
+            unsafe {
+                if !self.taught {
```
emilk, line 35: `taught` is a strange word to use. it is also always `false`.

egui_glow/src/shader/post_vertex_100es.glsl
```diff
@@ -0,0 +1,8 @@
+precision mediump float;
+attribute vec2 a_pos;
+varying vec2 v_tc;
+
+void main() {
+    gl_Position = vec4(a_pos * 2. - 1., 0.0, 1.0);
+    v_tc = a_pos;
+}
```
emilk, line 8: missing newline at the end of this and other files

egui_glow/src/post_process.rs
```diff
@@ -0,0 +1,244 @@
...
+            gl,
+            glow::FRAGMENT_SHADER,
+            include_str!("shader/post_fragment_100es.glsl"),
+        )?;
+        let program = link_program(gl, [vert_shader, frag_shader].iter())?;
+
+        let positions = vec![0.0f32, 0.0, 1.0, 0.0, 0.0, 1.0, 1.0, 1.0];
+
+        let indices = vec![0u8, 1, 2, 1, 2, 3];
+        unsafe {
```
emilk, line 124: Instead of having so many this functions has so many `unsafe` blocks I think it just nicer to add one big around the whole function, or mark the entire function as `unsafe`

egui_glow/src/painter.rs
```diff
@@ -581,29 +521,61 @@ impl Painter {
...
+        gl.clear(glow::COLOR_BUFFER_BIT);
+    }
+}
 impl Drop for Painter {
     fn drop(&mut self) {
-        debug_assert!(
+        #[cfg(debug_assertions)]
+        assert!(
             self.destroyed,
-            "Make sure to call destroy() before dropping to avoid leaking OpenGL objects!"
```
emilk, line 606: you seems to have accidentally reverted some cleanup like this that I made in https://github.com/emilk/egui/commit/8accfd9a8f2d21ada20661fda34e76913deef0b6

can you please try reverting these reverts?

egui_glow/src/misc_util.rs
emilk: Maybe `is_webgl_1` is a better name?

egui_glow/src/epi_backend.rs
```diff
@@ -74,9 +97,9 @@ fn integration_info(
 }
 
 // ----------------------------------------------------------------------------
-
+#[cfg(not(feature = "painter_only"))]
 pub use epi::NativeOptions;
-
+#[cfg(not(feature = "painter_only"))]
```
emilk, line 102: Maybe put `#[cfg(not(feature = "painter_only"))]` over `mod epi_backend;` instead of adding it to (almost) everything in this file?

egui_glow/Cargo.toml
```diff
@@ -23,19 +23,26 @@ all-features = true
...
+web-sys = { version = "0.3", features=["HtmlCanvasElement", "WebGl2RenderingContext","WebGlRenderingContext", "Window","console"] }
+wasm-bindgen = { version = "0.2" }
+
 [dev-dependencies]
 image = { version = "0.23", default-features = false, features = ["png"] }
 
 [features]
 default = ["clipboard", "default_fonts", "links", "persistence"]
-
+painter_only = []
```
emilk, line 45: `painter_only` is now a feature that _removes_ features, which is… weird :) 

Better to call it something like `winit` to opt-in to the winit-parts of `egui_glow`

Also: the features are sorted by name to make them easier to find. And there are spaces between them to help readability.

emilk (changes requested):
This looks really promising, nice work!

egui_glow/src/misc_util.rs
t18b219k: Ok i will fix it.

egui_glow/src/painter.rs
t18b219k: maybe i used old base revision .

egui_glow/src/painter.rs
t18b219k: sorry i miss typed it.

egui_web/src/backend.rs
t18b219k: Ok i fixed it.

egui_glow/src/epi_backend.rs
t18b219k: add `winit` feature instead of not(feature="painter_only").

egui_glow/src/vao_emulate.rs
t18b219k: i will remove it

egui_glow/Cargo.toml
t18b219k: add winit feature.

egui_glow/src/misc_util.rs
```diff
@@ -0,0 +1,186 @@
...
+}
+
+impl VAO {
+    pub(crate) unsafe fn new(gl: &glow::Context, is_native_vao: bool) -> Self {
+        if is_native_vao {
+            Self::Native(gl.create_vertex_array().unwrap())
+        } else {
+            Self::Emulated(crate::vao_emulate::EmulatedVao::new())
+        }
+    }
```
parasyte, line 143: This is nitpicky, but I think it would make sense to offer two constructors here instead of one.

```suggestion
    pub(crate) unsafe fn native(gl: &glow::Context) -> Self {
        Self::Native(gl.create_vertex_array().unwrap())
    }
    pub(crate) fn emulated() -> Self {
        Self::Emulated(crate::vao_emulate::EmulatedVao::new())
    }
```

Also, wouldn't the API be nicer if the constructor captured the `&glow::Context` here so it doesn't need to be passed to the other methods? That might also help reduce some confusion about why `VAO::bind_vertex_array` is unsafe, but the others are not?

emilk: Good point about two constructors 👍 

Capturing `&glow::Context` leads to a lot of ugly lifetime issues. Passing it in by reference is also more explicit, i.e. easier to see what is actually doing graphics calls. This was discussed in the original glow PR.

egui_web/src/backend.rs
emilk: …but perhaps in a separate PR

egui_glow/src/misc_util.rs
t18b219k: Ok  i will split constructor.
but is there good way to switch emulated mode?

parasyte: > Capturing `&glow::Context` leads to a lot of ugly lifetime issues. Passing it in by reference is also more explicit, i.e. easier to see what is actually doing graphics calls. This was discussed in the original glow PR.

Ok! Good to know, I didn't follow that conversation closely.

> Ok i will split constructor.
> but is there good way to switch emulated mode?

This question implies there is a way to switch the emulation mode with just one constructor? I don't believe that is the case before or after the suggested patch.

t18b219k: I thought this procedure good enough to switch emulated mode .
Is there any suggestion?
vao_switch_procedure

egui_glow/CHANGELOG.md
```diff
@@ -3,7 +3,7 @@ All notable changes to the `egui_glow` integration will be noted in this file.
 
 
 ## Unreleased
-
+* no glutin dependency in painter .
```
```suggestion
* Make winit/glutin an optional dependency ([#868](https://github.com/emilk/egui/pull/868)).
```

egui_glow/examples/pure_glow.rs
```diff
@@ -35,7 +35,6 @@ fn create_display(
 
     (gl_window, gl)
 }
-
 fn main() {
```
emilk, line 38:
```suggestion

fn main() {
```

egui_glow/src/painter.rs
```diff
@@ -3,247 +3,172 @@
...
+    pub(crate) user_textures: Vec<Option<UserTexture>>,
+    post_process: Option<PostProcess>,
+    vertex_buffer: glow::Buffer,
+    element_array_buffer: glow::Buffer,
 
     // Stores outdated OpenGL textures that are yet to be deleted
-    old_textures: Vec<glow::NativeTexture>,
-    // Only used in debug builds, to make sure we are destroyed correctly.
+    old_textures: Vec<glow::Texture>,
     destroyed: bool,
```
emilk, line 45:
```suggestion
    // Only used in debug builds, to make sure we are destroyed correctly.
    destroyed: bool,
```

egui_glow/src/painter.rs
```diff
@@ -3,247 +3,172 @@
...
-                panic!(
-                    "Failed to compile fragment shader: {}",
-                    gl.get_shader_info_log(frag)
-                );
+    /// create painter
+    ///
+    /// if `pp_fb_extent` is none post process disabled .
+    /// when post process disabled `sRGB` invalid color appeared on OpenGL ES and WebGL .
+    ///
+    /// to enable post process set framebuffer dimension to `pp_fb_extent`.
```
emilk, line 65:
```suggestion
    /// Create painter.
    ///
    /// Set `pp_fb_extent` to the framebuffer size to enable `sRGB` support on OpenGL ES and WebGL.
```

egui_glow/src/painter.rs
```diff
@@ -3,247 +3,172 @@
...
+            )
+            .map_err(|problems| {
+                glow_debug_print(format!("failed to compile fragment shader \n {}", problems));
+            })
+            .unwrap();
+            let program = link_program(gl, [vert, frag].iter())
+                .map_err(|problems| {
+                    glow_debug_print(format!("failed to link shaders \n {}", problems));
+                })
+                .unwrap();
```
emilk, line 128: Instead of doing

``` rust
.map_err(|problems| {
                    glow_debug_print(format!("failed to link shaders \n {}", problems));
                })
                .unwrap();
```

several times, use `?` and let the caller print the errors (once). i.e. return `Result<Self, String>` from `Painter::new`

egui_glow/src/painter.rs
```diff
@@ -268,16 +196,26 @@ impl Painter {
         if self.egui_texture_version == Some(texture.version) {
             return; // No change
         }
-
+        let gamma = if self.post_process.is_none() {
+            1.0 / 2.2
```
emilk, line 200: The glow painter on `main` doesn't need a post-processing step, because the native framebuffer is configured with linear (sRGB-aware) blending. So the gamma should be `1.0` unless we are on web and the post-processing step is missing.

```suggestion
        let gamma = if self.is_web && self.post_process.is_none() {
            1.0 / 2.2
```

…unless I'm missing something

egui_glow/src/painter.rs
```diff
@@ -288,7 +226,7 @@ impl Painter {
 
     unsafe fn prepare_painting(
         &mut self,
-        gl_window: &glutin::WindowedContext<glutin::PossiblyCurrent>,
+        inner_size: [u32; 2],
```
emilk, line 229:
```suggestion
        [width_in_pixels, height_in_pixels]: [u32; 2],
```

makes it clearer what this is

egui_glow/src/painter.rs
```diff
@@ -341,8 +275,7 @@ impl Painter {
     ///
     /// The scissor area and blend parameters will be changed.
     ///
-    /// As well as this, the following objects will be rebound:
-    /// - Vertex Array
```
emilk, line 345: It still changes `Vertex Array`, no?

egui_web/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ All notable changes to the `egui_web` integration will be noted in this file.
 
 
 ## Unreleased
-
+* add feature `glow` if enabled painting switched to `egui_glow::painter::Painter`.
```
emilk, line 7:
```suggestion
*Add feature `glow` to switch to a [`glow`](https://github.com/grovesNL/glow) based painter ([#868](https://github.com/emilk/egui/pull/868)).

emilk (changes requested):
There is a lot of things that needs testing with "Color test":

* Native glow
* Web glow on WebGL2
* Web glow on WebGL1 with sRGBA
* Web glow on WebGL1 without sRGBA (won't pass the color test, but should look as good as the current WebGL painter)

egui_glow/src/painter.rs
t18b219k: Ok i fixed it

egui_glow/src/painter.rs
t18b219k: no but vertex array will be null .

t18b219k:
merging upstream change
Now i post some good photo
スクリーンショット 2021-11-03 230707
.

emilk (approved):

emilk:
Thanks @t18b219k - great work!

AlexApps99:
Sorry I didn't have much feedback for it - busy with exams, I won't be able to contribute much for a bit.

Issue #853, opened 2021-10-29
Allow specifying font index when loading a font file


7sDream:
I try to let my egui app to use system font.

I can use some font query/loading library to find the font location and read it as binary data, however, it's a ttc file(for example, Microsoft YaHei), so egui report the data a invalid when parsing.

It seems ab_glyph do support collection, by using [FontRef::try_from_slice_and_index](https://docs.rs/ab_glyph/0.2.11/ab_glyph/struct.FontRef.html#method.try_from_slice_and_index), so any idea how can I use ttc/otc ffont file as custom font in egui?

## update

Maybe there is some bug in the font loading library I used.

I changed from [font-loader](https://github.com/matprec/rust-font-loader) to [fontdb](https://github.com/RazrFalcon/fontdb), and now it's working, no `invalid font` report.

But, IIRC,  font loading library gives whole file content, and because I can't pass the font face index to egui, current it will always using the font face at index 0, which means when using ttc/otf, the render result may not be as expected.

<details>
<summary>For example, Helvetica Neue(Normal style) is in index 10</summary>
<img width="1301" src="https://user-images.githubusercontent.com/7822577/139451426-57271b62-a023-4d4c-b8ef-124861eb63bc.png">
</details>

emilk:
If you could specify the face index in `epaint::FontDefinitions` (together with the `FontData`), would that solve the problem for you?

7sDream:
Yes, I think add a `font_face_index: HashMap<String, usize>` field in `FontDefinitions` (with the same key in `font_data`, values is the index) is a feasible solution.

Considering compatibility, a default value of 0 can be used when the index cannot be found in this field.

emilk:
It seems `ab_glyph::FontArc` lacks the convenience of `try_from_vec_and_index` and `try_from_slice_and_index`. If you make a PR to `ab_glyph` to add those I can hook them up in egui later!

parasyte:
`FontArc` does implement `From<FontRef>` though. That should be enough?

emilk:
Good point @parasyte - I made a PR; please try it out @7sDream !

7sDream:
Thank you! I can try this pr tonight, after work

PR #873, opened 2021-11-03
When using a custom font you can now specify a font index

emilk:
Closes https://github.com/emilk/egui/issues/853

Please try it out @7sDream


7sDream:
It seems that `FontData` cannot be accessed because it is not re-exported in the `epaint::text` module and `epaint::text::font` is private. 

Maybe it also need to be re-exported in `egui`'s top level, just like `FontDefinitions`, for ease of use.

emilk:
Good catch - fixed!

7sDream:
Works like a charm now, can't wait to see this in next release, thank you!

PR #875, opened 2021-11-04
implement vertical slider orientation

### LINKED ISSUE #589: Vertical Slider
Hello!

I found this project and thought this would really fit my use case so i checked the docs anf the issues page but i cant find anything mentioned about Vertical Slider at all.

So is it possible or is it planned?

B-Reif:
This PR implements a vertical orientation for `Slider`. Closes #589.

# What's new
- Sliders demo now includes a radio button toggle for orientation
- New enum `SliderOrientation` with variants `Horizontal` and `Vertical`
- New mutator function `.orientation` to set the slider orientation
- New constructor functions, `new_vertical` and `from_get_set_vertical` to instantiate with vertical orientation.
- Vertical Sliders use alternate keycodes for increment/decrement (up/down instead of right/left).

# What's changed
- Replaces horizontal-specific Slider logic with some functions matching on `orientation` instead. Some functions and parameters are renamed (e.g. `x_range` -> `position_range`).
- Some helper functions are now methods on `Slider` which match on the Slider's orientation.

# Alternatives
- If API breakage is ok, we could add an orientation argument to existing constructor functions (instead of creating new constructor functions).
- We could expose only the `.orientation` mutator function (instead of changing or creating constructor functions).
- We could expose a vertical slider as a different widget. The logic is similar enough that I think the sharing here is appropriate.

### DIFF
--- a/egui/src/widgets/slider.rs
+++ b/egui/src/widgets/slider.rs
@@ -32,7 +32,13 @@ struct SliderSpec {
     largest_finite: f64,
 }
 
-/// Control a number by a horizontal slider.
+/// Specifies the orientation of a [Slider].
+pub enum SliderOrientation {
+    Horizontal,
+    Vertical,
+}
+
+/// Control a number with a slider.
 ///
 /// The slider range defines the values you get when pulling the slider to the far edges.
 /// By default, the slider can still show values outside this range,
@@ -41,7 +47,7 @@ struct SliderSpec {
 ///
 /// The range can include any numbers, and go from low-to-high or from high-to-low.
 ///
-/// The slider consists of three parts: a horizontal slider, a value display, and an optional text.
+/// The slider consists of three parts: a slider, a value display, and an optional text.
 /// The user can click the value display to edit its value. It can be turned off with `.show_value(false)`.
 ///
 /// ```
@@ -60,6 +66,7 @@ pub struct Slider<'a> {
     clamp_to_range: bool,
     smart_aim: bool,
     show_value: bool,
+    orientation: SliderOrientation,
     prefix: String,
     suffix: String,
     text: String,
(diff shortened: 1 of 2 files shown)


emilk:
I've skimmed through it, and overall it looks good! Will review properly when I find the time.

egui/src/widgets/slider.rs
```diff
@@ -85,9 +93,44 @@ impl<'a> Slider<'a> {
         }
     }
 
+    /// Creates a new vertical slider.
+    pub fn new_vertical<Num: emath::Numeric>(
```
emilk, line 97: It seems unnecessary to have both this and the `.orientation` builder.

Perhaps `Slider::new(value, 0..100).vertical()` is a nicer way to construct a vertical slider anyway? And then we don't need `new_vertical`, `from_get_set_vertical` and `from_get_set_inner`

egui/src/widgets/slider.rs
```diff
@@ -139,6 +183,11 @@ impl<'a> Slider<'a> {
         self
     }
 
+    pub fn orientation(mut self, orientation: SliderOrientation) -> Self {
```
emilk, line 186:
```suggestion
    /// Vertical or horizontal slider? The default is horizontal.
    pub fn orientation(mut self, orientation: SliderOrientation) -> Self {
```

egui/src/widgets/slider.rs
```diff
@@ -346,15 +385,85 @@ impl<'a> Slider<'a> {
...
+    }
+
+    fn position_range(&self, rect: &Rect) -> RangeInclusive<f32> {
+        let handle_radius = self.handle_radius(rect);
+        match self.orientation {
+            SliderOrientation::Horizontal => {
+                (rect.left() + handle_radius)..=(rect.right() - handle_radius)
+            }
+            SliderOrientation::Vertical => {
+                (rect.top() + handle_radius)..=(rect.bottom() - handle_radius)
```
emilk, line 420: I think the default for a vertical slider should be higher values higher up, and lower values lower down:

```suggestion
                (rect.bottom() - handle_radius)..=(rect.top() + handle_radius)
```

emilk (changes requested):

emilk:
Screen Shot 2021-11-07 at 21 05 12

This looks a bit bad to me. I guess most people who use a vertical slider won't want the label text (no shown above), but the slider value would be very nice to show, and should be _below_ the vertical slider.

B-Reif:
Ok, made review changes and moved the value below the slider. Also added a new method `add_contents` to invoke inside of either `ui.horizontal` or `ui.vertical` respectively.

image

This implementation will arrange the slider, value, and then label in a vertical layout. We could further group the value and label onto one horizontal line.

B-Reif:
I'm having some issues with the keyboard step functionality of Slider, but I experience them on the current master branch also. These could probably be resolved in a separate PR.

emilk (approved):

PR #888, opened 2021-11-08
Fix wrong gamma in WebkitGTK

### LINKED ISSUE #794: Everything is super dark in epiphany
**Describe the bug**
The demo web app is super dark in epiphany. See screenshot below.

**To Reproduce**
1. Open the [demo web app](https://emilk.github.io/egui/index.html) in epiphany.

**Expected behavior**
Should look like it does in other web browsers (e.g. Firefox).

**Screenshots**
Screenshot from 2021-10-13 10-49-04

 - OS: Fedora 34
 - Browser: Epiphany
 - Version: 40.3

t18b219k:
Epiphany's incorrect sRGB handling  cause  same result as displaying linear image on other browsers.
I fixed this issue by double sRGB correcting in post process. 
Glitched image
color difference
fixed result
スクリーンショット 2021-11-08 174650

### DIFF
--- a/egui_glow/CHANGELOG.md
+++ b/egui_glow/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to the `egui_glow` integration will be noted in this file.
 
 
 ## Unreleased
+* Add `workarounds` to `Painter::new` to pass GLSL macro to fix platform specific behavior.
 * Make winit/glutin an optional dependency ([#868](https://github.com/emilk/egui/pull/868)).
 * Simplify `EguiGlow` interface ([#871](https://github.com/emilk/egui/pull/871)).
 * Remove `EguiGlow::is_quit_event` ([#881](https://github.com/emilk/egui/pull/881)).
--- a/egui_glow/CHANGELOG.md
+++ b/egui_glow/CHANGELOG.md
@@ -3,7 +3,6 @@ All notable changes to the `egui_glow` integration will be noted in this file.
 
 
 ## Unreleased
-* Add `workarounds` to `Painter::new` to pass GLSL macro to fix platform specific behavior.
 * Make winit/glutin an optional dependency ([#868](https://github.com/emilk/egui/pull/868)).
 * Simplify `EguiGlow` interface ([#871](https://github.com/emilk/egui/pull/871)).
 * Remove `EguiGlow::is_quit_event` ([#881](https://github.com/emilk/egui/pull/881)).
--- a/egui_glow/src/misc_util.rs
+++ b/egui_glow/src/misc_util.rs
@@ -184,7 +184,8 @@ impl VAO {
     }
 }
 
-pub(crate) unsafe fn need_to_emulate_vao(gl: &glow::Context) -> bool {
+/// If returned true no need to emulate vao
+pub(crate) unsafe fn support_vao(gl: &glow::Context) -> bool {
     let web_sig = "WebGL ";
     let es_sig = "OpenGL ES ";
     let version_string = gl.get_parameter_string(glow::VERSION);
@@ -197,10 +198,10 @@ pub(crate) unsafe fn need_to_emulate_vao(gl: &glow::Context) -> bool {
         ));
         if version_str.contains("1.0") {
             //need to test OES_vertex_array_object .
-            !gl.supported_extensions()
+            gl.supported_extensions()
                 .contains("OES_vertex_array_object")
         } else {
-            false
+            true
         }
     } else if let Some(pos) = version_string.rfind(es_sig) {
         //glow targets es2.0+ so we don't concern about OpenGL ES-CM,OpenGL ES-CL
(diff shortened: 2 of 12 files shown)


egui_web/src/webgl1.rs
```diff
@@ -591,7 +592,19 @@ impl PostProcess {
 
         gl.bind_texture(Gl::TEXTURE_2D, None);
         gl.bind_framebuffer(Gl::FRAMEBUFFER, None);
-
+        // currently epiphany only support webgl1
+        let epiphany_workaround = if web_sys::window()
```
emilk, line 596: let's call this `frag_shader_prefix` instead

egui_glow/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ All notable changes to the `egui_glow` integration will be noted in this file.
 
 
 ## Unreleased
+* Add `workarounds` to `Painter::new` to pass GLSL macro to fix platform specific behavior.
```
```suggestion
```

Since this isn't a generally usable thing (you need to also modify the shaders), I think we can omit it from this changelog

emilk (changes requested):

egui_glow/src/painter.rs
```diff
@@ -61,18 +61,33 @@ impl Painter {
...
     /// # Errors
     /// will return `Err` below cases
     /// * failed to compile shader
     /// * failed to create postprocess on webgl with `sRGB` support
     /// * failed to create buffer
-    pub fn new(gl: &glow::Context, pp_fb_extent: Option<[i32; 2]>) -> Result<Painter, String> {
+    pub fn new(
+        gl: &glow::Context,
+        pp_fb_extent: Option<[i32; 2]>,
+        workarounds: &[String],
```
emilk, line 75: I think just calling this `shader_prefix: &str` here and elsewhere would be clearer

emilk (changes requested):

t18b219k:
I made some mistake in diag VAO code .
So i want to fix this issue in this PR.

egui_web/src/glow_wrapping.rs
```diff
@@ -15,10 +16,17 @@ pub(crate) struct WrappedGlowPainter {
...
+            console_log("Enabling epiphany workaround");
+            "#define EPIPHANY_WORKAROUND"
+        } else {
+            ""
+        };
         let canvas = canvas_element_or_die(canvas_id);
         let gl_ctx = init_glow_context_from_canvas(&canvas);
         let dimension = [canvas.width() as i32, canvas.height() as i32];
-        let painter = egui_glow::Painter::new(&gl_ctx, Some(dimension))
+        let painter = egui_glow::Painter::new(&gl_ctx, Some(dimension), epiphany_wr)
```
emilk, line 29: I was wondering if an alternative work-around would work as well: to turn off the `postprocessing` step on epiphany.

So something like:


``` rust
let pp_fb_extents = if user_agent.contains("Epiphany") {
     None // epiphany doesn't support sRGB blending even with render-to-texture
} else {
    Some(dimensions);
};
let painter = egui_glow::Painter::new(&gl_ctx, pp_fb_extents);
```

Did you try this already?

t18b219k: if we disable post process produce not good image in Epiphany.
スクリーンショット 2021-11-10 203714
Epiphany does not

t18b219k: I also tested midori and konqueror .
midori produce same result with epiphany so this issue from webkitGTK implementation.
texture sRGB worked but frame buffer sRGB incorrect.

t18b219k: I'm tuning color correcting.
and found `vec4(pow(gl_FragColor.rgb,vec3(1.0/2.2)),gl_FragColor.a)`is better than double apply.

t18b219k: we can detect WebkitGTK by following checks.
*  webgl unmasked vendor Apple
*  webgl renderer WebKit WebGL
*  user-agent not "Mac OS X"

egui_glow/src/misc_util.rs
```diff
@@ -197,7 +197,7 @@ pub(crate) unsafe fn need_to_emulate_vao(gl: &glow::Context) -> bool {
         ));
         if version_str.contains("1.0") {
             //need to test OES_vertex_array_object .
-            gl.supported_extensions()
+            !gl.supported_extensions()
```
emilk, line 200: easy mistake to make. I think it would be less confusing if the function would be called `supports_vao`. As written, it leads to double-negatives like this.

egui_web/CHANGELOG.md
```diff
@@ -4,7 +4,8 @@ All notable changes to the `egui_web` integration will be noted in this file.
 
 
 ## Unreleased
-*Add feature `glow` to switch to a [`glow`](https://github.com/grovesNL/glow) based painter ([#868](https://github.com/emilk/egui/pull/868)).
+* fix [Everything super dark in epiphany](https://github.com/emilk/egui/issues/794) for webgl1 and glow based painter
```
emilk, line 7:
```suggestion
* Fix [dark rendering in epiphany](https://github.com/emilk/egui/issues/794) for WebGL1 and glow based painter ([#888](https://github.com/emilk/egui/pull/888/)).
```

emilk (approved):

egui_glow/src/misc_util.rs
t18b219k: Ok i fixed it.

PR #892, opened 2021-11-14
Improved plot interaction methods

### LINKED ISSUE #891: Improved plot interaction methods
#766 has just been merged and is usable, I'd like to add the following features next:

* Access to `Ui` (or the context?) inside the closure as well. Either through the `plot_ui` struct or as a second argument.
* A `plot_to_screen` conversion, to enable custom drawing in screen coordinates based on items in the plot.
* A way to get the current plot bounds.
* Fix the `plot_from_screen` function. Pretty sure I messed up hard when writing it and it likely does not work as expected right now.

EmbersArc:
Closes #891 

Changes:
* Added access to the context with `plot_ui.ctx()` inside the plot `show` closure.
* More `PlotUi` methods: `ctx`, `plot_bounds`, `plot_hovered` and `screen_from_plot`.
* Simplified the way plot bounds and the screen transform are initialized.
* Fixed the `plot_from_screen` function.
* `Plot::show` now returns an `InnerResponse` to be consistent with similar functions.
* Improved demo

@emilk I noticed that the plot demo instructions say "Pan by dragging, or scroll (+ shift = horizontal).". This does not work on my system though. Is this a bug or is the note outdated?

### DIFF
--- a/egui/src/widgets/plot/mod.rs
+++ b/egui/src/widgets/plot/mod.rs
@@ -219,11 +219,7 @@ impl Plot {
     }
 
     /// Interact with and add items to the plot and finally draw it.
-    pub fn show<R>(
-        self,
-        ui: &mut Ui,
-        build_fn: impl FnOnce(&mut PlotUi<'_>) -> R,
-    ) -> InnerResponse<R> {
+    pub fn show<R>(self, ui: &mut Ui, build_fn: impl FnOnce(&mut PlotUi) -> R) -> InnerResponse<R> {
         let Self {
             id_source,
             center_x_axis,
@@ -311,7 +307,7 @@ impl Plot {
             next_auto_color_idx: 0,
             last_screen_transform,
             response,
-            ctx: ui.ctx(),
+            ctx: ui.ctx().clone(),
         };
         let inner = build_fn(&mut plot_ui);
         let PlotUi {
@@ -442,15 +438,15 @@ impl Plot {
 
 /// Provides methods to interact with a plot while building it. It is the single argument of the closure
 /// provided to `Plot::show`. See [`Plot`] for an example of how to use it.
-pub struct PlotUi<'ctx> {
+pub struct PlotUi {
     items: Vec<Box<dyn PlotItem>>,
     next_auto_color_idx: usize,
     last_screen_transform: ScreenTransform,
     response: Response,
-    ctx: &'ctx CtxRef,
+    ctx: CtxRef,
 }
 
-impl PlotUi<'_> {
+impl PlotUi {
     fn auto_color(&mut self) -> Color32 {
         let i = self.next_auto_color_idx;
         self.next_auto_color_idx += 1;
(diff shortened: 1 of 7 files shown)


egui/src/widgets/plot/mod.rs
```diff
@@ -446,35 +459,43 @@ impl PlotUi {
         Hsva::new(h, 0.85, 0.5, 1.0).into() // TODO: OkLab or some other perspective color space
     }
 
-    /// The pointer position in plot coordinates, if the pointer is inside the plot area.
+    pub fn ctx(&self) -> &CtxRef {
+        self.ctx
+    }
+
+    /// The plot bounds as they were in the last frame.
```
emilk, line 466: What is its value the first frame? Would it make more sense to return `None` the first frame?

egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -366,45 +371,62 @@ impl Widget for &mut ItemsDemo {
...
+                plot_ui.screen_from_plot(Value::new(0.0, 0.0)),
+                plot_ui.pointer_coordinate(),
+                plot_ui.pointer_coordinate_drag_delta(),
+                plot_ui.plot_bounds(),
+                plot_ui.plot_hovered(),
+            )
+        });
+
+        ui.label(format!(
+            "plot bounds: min: {:.02?}, max: {:.02?}",
```
emilk, line 405: I'm curious: what does the `?` accomplish here? Is the debug-formatting of a `f64` different from its default formatting?

egui/src/widgets/plot/mod.rs
emilk: I guess the same comment applies to `screen_from_plot` and `plot_from_screen` - what is apropiate to return the first frame?

EmbersArc: We made sure that the bounds never become invalid. So if no items have been added to the plot and the bounds were not specified in the builder, they will at least be in the valid range `[-1.0, 1.0]`. The first transform may not be very accurate, but valid and I prefer this to making the user deal with an `Option` just because of one frame. But maybe you know an instance where this could go wrong?

egui_demo_lib/src/apps/demo/plot_demo.rs
EmbersArc: Those are arrays, so the debug printing is needed. I was quite surprised though to see that the float formatting works as well in that case.

emilk: oh right!

emilk (approved):
I would love to see the docs on `plot_bounds` improved a bit, but you can do so in a later PR if you like.

Great work!

egui/src/widgets/plot/mod.rs
```diff
@@ -411,33 +422,35 @@ impl Plot {
...
 /// Provides methods to interact with a plot while building it. It is the single argument of the closure
 /// provided to `Plot::show`. See [`Plot`] for an example of how to use it.
-pub struct PlotUi {
+pub struct PlotUi<'ctx> {
     items: Vec<Box<dyn PlotItem>>,
     next_auto_color_idx: usize,
-    last_screen_transform: Option<ScreenTransform>,
+    last_screen_transform: ScreenTransform,
     response: Response,
+    ctx: &'ctx CtxRef,
```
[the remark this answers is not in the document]
EmbersArc: Ah neat, changed in d1cff07.

egui/src/widgets/plot/mod.rs
EmbersArc: Good point, I added some documentation.

emilk (approved):

PR #902, opened 2021-11-18
eframe: don't prettify stored values

AlexChaplinBraz:
What ultimately ends up stored into a file is the
HashMap<String, String>, which when prettified only nets three
(depending on the settings) lines of "string": "long prettified
string that's really hard to read because of the extra indentation and
literal \n characters that are all just on one single line".

Not prettifying the values in the first place makes it somewhat easier
to read and also saves a bit of space.


AlexChaplinBraz:
The amount of space saved is actually quite significant. I tried getting rid of all the \n plus adjacent whitespace with `sed` on the save file of what I'm working on and it went from 346.6 KiB to 224.4 KiB. That's a third of the size wasted right there. I reckon that's important if targetting the web, right?

The program I'm working on is for native, but I'm expecting the save file to be a few megabytes at least (growing larger the more stuff the user adds). Not sure if that's too big, but I'll see. I _can_ work around it, but would prefer not to if it won't affect performance anyway.

emilk:
We should also consider switching to a binary alternative such as MsgPack

emilk (approved):

PR #907, opened 2021-11-23
Add hovered_anyway() to Response to deal with child widgets events when hovered

imxood:
Add hovered_anyway() to Response, this can help to deal with child widgets events when current widget is hovered

## usage

image

image

### DIFF
--- a/egui/src/response.rs
+++ b/egui/src/response.rs
@@ -185,12 +185,14 @@ impl Response {
         self.hovered
     }
 
+    /// The pointer is hovering above this widget or the widget was clicked/tapped this frame.
+    /// this can help to deal with child widgets events
     pub fn hovered_anyway(&self) -> bool {
         let pointer = &self.ctx.input().pointer;
         if let Some(pos) = pointer.interact_pos() {
             self.rect.contains(pos)
         } else {
-            false
+            self.hovered
         }
     }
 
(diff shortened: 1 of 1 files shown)


egui/src/response.rs
```diff
@@ -185,6 +185,17 @@ impl Response {
         self.hovered
     }
 
+    /// The pointer is hovering above this widget or the widget was clicked/tapped this frame.
+    /// this can help to deal with child widgets events
+    pub fn hovered_anyway(&self) -> bool {
+        let pointer = &self.ctx.input().pointer;
+        if let Some(pos) = pointer.interact_pos() {
+            self.rect.contains(pos)
```
emilk, line 193: this will return true even if another egui window is covering your widget. is this really intentional?

imxood: I didn't think so much ....

emilk: What is your use case? Would `fn hovered_or_clicked(&self) -> bool { self.hovered() || self.clicked() }` handle it?

imxood: I want to realize that when hovering the current component, it can display the sub-components, such as "delete button"

image
image

emilk:
I think something like this is a better solution: https://github.com/emilk/egui/issues/1010

PR #918, opened 2021-11-29
OrderedFloat refactor

Bromeon:
Following https://github.com/emilk/egui/pull/863#discussion_r757770991, this moves `egui/util/float_ord.rs` -> `epaint/util/ordered_float.rs` and implements `Hash` based on `epaint::f32_hash()`.

Open questions:
1. Currently, `f64::ord()` simply truncates to `f32`, which may yield incorrect results for high precision floats. I see three solutions:
   * Make `OrderedFloat` generic on the float type. A bit more complex implementation, as we need to create traits to abstract over `f32` and `f64` for `is_nan()` etc.
   * Use always `f32` (status quo) -- not recommended, can cause really nasty bugs
   * Use always `f64` -- safe, but might incur minor performance penalty
   
2. Should this be part of the public API?
   I moved it to a different crate, so it can't be `pub(crate)` anymore. However, we can mark it `#[doc(hidden)]` if it should stay an implementation detail.

### DIFF
--- a/egui/src/util/mod.rs
+++ b/egui/src/util/mod.rs
@@ -2,7 +2,6 @@
 
 pub mod cache;
 pub(crate) mod fixed_cache;
-pub(crate) mod float_ord;
 mod history;
 pub mod id_type_map;
 pub mod undoer;
--- a/egui/src/widgets/plot/items/mod.rs
+++ b/egui/src/widgets/plot/items/mod.rs
@@ -2,9 +2,9 @@
 
 use std::ops::RangeInclusive;
 
+use epaint::util::FloatOrd;
 use epaint::Mesh;
 
-use crate::util::float_ord::FloatOrd;
 use crate::*;
 
 use super::{PlotBounds, ScreenTransform};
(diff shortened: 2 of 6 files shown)


emilk:
Let's keep it `f32` only until a need for `f64` actually arises. Then we can make it generic.

It's fine if it's part of the public API imho - it is hidden enough in `mod util`

Bromeon:
There's currently one occurrence already: 

https://github.com/emilk/egui/blob/master/egui/src/widgets/plot/items/mod.rs#L1309
The field `Bar::value` is of type `f64`, for consistency with the X/Y values.

```rs
let new_base_offset = if bar.value.is_sign_positive() {
    others
        .iter()
        .filter_map(|other_chart| other_chart.bars.get(index).map(|bar| bar.upper()))
        .max_by_key(|value| value.ord())
} else {
    others
        .iter()
        .filter_map(|other_chart| other_chart.bars.get(index).map(|bar| bar.lower()))
        .min_by_key(|value| value.ord())
};
```

emilk:
Oh right! Then I suggest either `OrderedFloat<T>` or `OrderedF32/OrderedF64` (probably simpler).

Bromeon:
Went for `OrderedFloat<T>`, with `T: Float` as an internal bound. `Float` is a minimalistic generalization of `f32` and `f64` with just `hash()` and `is_nan()` for now; could be extended if other parts of the library need it.

Also added some documentation and a usage example.

emilk (approved):

PR #930, opened 2021-12-03
Update winit to 0.26

parasyte:
- This has some obvious issues at the moment. First, transparent windows are not working on Windows 11. See: https://github.com/rust-windowing/winit/issues/2082 So I've disabled that feature for now.
- Secondly, `glium` needs an update. I'm patching it with an unreleased git revision to make sure things are working together.

### DIFF
--- a/egui-winit/Cargo.toml
+++ b/egui-winit/Cargo.toml
@@ -22,7 +22,7 @@ all-features = true
 
 [dependencies]
 egui = { version = "0.15.0", path = "../egui", default-features = false, features = ["single_threaded"] }
-winit = "0.25"
+winit = "0.26"
 
 epi = { version = "0.15.0", path = "../epi", optional = true }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -635,7 +635,7 @@ checksum = "22813a6dc45b335f9bade10bf7271dc477e81113e89eb251a0bc2a8a81c536e1"
 dependencies = [
  "bstr",
  "csv-core",
- "itoa",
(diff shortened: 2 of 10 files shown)


sourcebox:
The latest release seems to contain the fixes for the freezing dialogs on macOS. So there's finally some progress and we maybe can get rid of the winit fork now.

parasyte:
I think this is as good as I want to get it. Disabling the transparent window to avoid crashing Windows 11 seems like a fair trade.

parasyte:
There is a dependency on `tts` which could be updated, too. But it requires a pinned version of Rust that supports Edition 2021.

egui-winit/Cargo.toml
```diff
@@ -22,7 +22,7 @@ all-features = true
 
 [dependencies]
 egui = { version = "0.15.0", path = "../egui", default-features = false, features = ["single_threaded"] }
-winit = "0.25"
```
emilk, line 25: We should mention the winit version bump in `egui-winit/CHANGELOG.md`

emilk (approved):

egui-winit/Cargo.toml
parasyte: Updated changelogs!

emilk:
Closes https://github.com/emilk/egui/issues/977

emilk:
> There is a dependency on `tts` which could be updated, too. But it requires a pinned version of Rust that supports Edition 2021.

Updated to rust edition 2021 in https://github.com/emilk/egui/pull/998

Issue #933, opened 2021-12-06
Custom hover labels for plot x/y values


isegal:
Currently, it is possible to customize the name of the line graph, but x and y coordinate names for hovering over the plot is not customizable. When hovering over a plot line, the user will always see
```
<name>
x = 123
y = 234
````
This is a request to add a set of api functions to be able to customize how the hover text gets displayd in the plot graph.

For example, the user may want to label x axis as time and y axis as percent or cost or another metric.

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

One option is to allow the user to add a custom closure that takes the plot name and Value as input and returns a formatted string of how the values are to be displayed.

```rust
Plot::new("my_plot")
    .custom_label_func(|name, value| {
        if let Some(name) = name {
            format!("{}: {:.*}%", name, 1, value.y).to_string()
        } else {
            "CPU".to_string()
        }
    })
    .show(ui, |plot_ui| plot_ui.line(line));
```

This would allow relabeling x and y and display a custom formatted string for the hover values.

**Describe alternatives you've considered**
Considered adding x and y names, but this removes some flexibility on how the hover would be displayd.

PR #934, opened 2021-12-06
Add ability to customize the display of hover plot labels

isegal:
This is a feature candidate / RFC for https://github.com/emilk/egui/issues/933

This is a work-in-progress POC to gauge whether there is interest in this feature. Happy to continue to make changes based on feedback.

### DIFF
--- a/egui/src/widgets/plot/items/mod.rs
+++ b/egui/src/widgets/plot/items/mod.rs
@@ -1658,7 +1658,6 @@ pub(super) fn rulers_at_value(
         let x_decimals = ((-scale[0].abs().log10()).ceil().at_least(0.0) as usize).at_most(6);
         let y_decimals = ((-scale[1].abs().log10()).ceil().at_least(0.0) as usize).at_most(6);
         if let Some(custom_label) = custom_label_func {
-            let name = (!name.is_empty()).then(|| name);
             custom_label(name, &value)
         } else if plot.show_x && plot.show_y {
             format!(
--- a/egui/src/widgets/plot/items/mod.rs
+++ b/egui/src/widgets/plot/items/mod.rs
@@ -1658,6 +1658,7 @@ pub(super) fn rulers_at_value(
         let x_decimals = ((-scale[0].abs().log10()).ceil().at_least(0.0) as usize).at_most(6);
         let y_decimals = ((-scale[1].abs().log10()).ceil().at_least(0.0) as usize).at_most(6);
         if let Some(custom_label) = custom_label_func {
+            let name = (!name.is_empty()).then(|| name);
             custom_label(name, &value)
         } else if plot.show_x && plot.show_y {
             format!(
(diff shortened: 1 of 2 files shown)


egui/src/widgets/plot/items/mod.rs
```diff
@@ -61,7 +61,13 @@ pub(super) trait PlotItem {
         }
     }
 
-    fn on_hover(&self, elem: ClosestElem, shapes: &mut Vec<Shape>, plot: &PlotConfig<'_>) {
+    fn on_hover(
+        &self,
+        elem: ClosestElem,
+        shapes: &mut Vec<Shape>,
+        plot: &PlotConfig<'_>,
+        custom_label_func: &CustomLabelFunc,
```
isegal, line 69: We allow the user to pass a custom function that formats the hover string based on input value.

egui/src/widgets/plot/items/mod.rs
```diff
@@ -1365,7 +1371,13 @@ impl PlotItem for BarChart {
         find_closest_rect(&self.bars, point, transform)
     }
 
-    fn on_hover(&self, elem: ClosestElem, shapes: &mut Vec<Shape>, plot: &PlotConfig<'_>) {
+    fn on_hover(
+        &self,
+        elem: ClosestElem,
+        shapes: &mut Vec<Shape>,
+        plot: &PlotConfig<'_>,
+        _: &CustomLabelFunc,
```
isegal, line 1379: Currently implemented only for the line chart. If this is the direction that we are good to go, happy to continue implementing for BarChart and BoxPlot.

egui/src/widgets/plot/mod.rs
```diff
@@ -182,6 +187,35 @@ impl Plot {
...
+    ///     if let Some(name) = name {
+    ///         format!("{}: {:.*}%", name, 1, value.y).to_string()
+    ///     } else {
+    ///         "".to_string()
+    ///     }
+    /// })
+    /// .show(ui, |plot_ui| plot_ui.line(line));
+    /// # });
+    /// ```
+    pub fn custom_label_func<F: 'static + Fn(Option<&str>, &Value) -> String>(
```
emilk, line 211: I think it would be simpler to use `&str` instead of `Option<&str>`. An empty string is a good way to communicate "there is no name for this". The type `Option<&str>` tells users "this could be `Some("")`" which means they should guard against that, which is annoying.

emilk: The `'static` bound may be a bit annoying for some users, but adding a lifetime bond to `Plot` would also be annoying, so perhaps this is a good enough start.

emilk (changes requested):
Looks good to me (except for `Option<&str>` -> `&str`). Perhaps @EmbersArc also has some thoughts.

egui/src/widgets/plot/mod.rs
isegal: > I think it would be simpler to use `&str` instead of `Option<&str>`. An empty string is a good way to communicate "there is no name for this". The type `Option<&str>` tells users "this could be `Some("")`" which means they should guard against that, which is annoying.

No problem. I have reverted the commit so that we switch back to `&str`.

Issue #843, opened 2021-10-25
Investigate a multipass (two-pass) version of egui


emilk:
EDITED 2021-10-30: Expanded and added proposal
EDITED 2021-11-01: Realized input must be done in the _second_ pass

# Introduction
## Problem
One downside of immediate mode is the difficulty in doing layouts where one needs to know the size of a section before positioning it, but to know the size of that section one needs to call the code that creates it - but that also positions it.

Consider the menus in egui: we use a justified layout to make sure the buttons cover the full width of the menu, but how wide should the menu be? We won't know until we put in all the buttons, and we don't know how wide to make the buttons until we know the width of the menu:

```
| Save        |
| Preferences |
| Quit        |
```

How do we know to make the "Save" button wide until we know there is a "Preferences" button? We don't!

Storing the size from the previous frame is also not a solution, because if we then remove the "Preferences" button, the menu will not shrink to the smaller size, but would stay at the widest it has ever been. We see this problem today with some `Window`:s in egui, which will auto-expand but never auto-shrink.

Because of issues like these, egui can feel janky an unreliable today. As a user it is difficult to understand why it is possible to center some widgets, but not other ("composite" widgets, which don't know their sizes beforehand). It also produces frame-delay jank (open a new window and it will often flicker the first frame, before "finding its size").

I want egui to "just work".

## Two-pass
Some immediate mode UI:s take a two-pass approach to solving this a **size pass** and a **layout pass**. In the first pass, sizes of ui:s are calculated and stored, and in the second pass the sizes are used to nicely position things. This can be extended to even more passes, but with diminishing returns.

I believe a two-pass approach could solve most of the layout issues in egui.

It will also make some widgets easier to implement: instead of having to know all the sizes of its part beforehand, it can just have it automatically calculated in the first pass, and then use it in the second pass.

# Things to consider

## Consistency
Both passes should ideally act exactly the same, which will be impossible to guarantee when it comes to user code. For instance, say the user code checks "Is the async download complete?" before deciding to show a widget or not. What if the download completes right in the middle of the two passes? Then the layout pass would not have the stored size of the widget.

So the layout pass need to be robust against missing sizes, falling back to some "okish" behavior (e.g. what egui currently does).

Such inconsistency will always happen with input:

## Input
Inputs events should only be handled in one of the passes, otherwise we would enter text twice in a `TextEdit`, scroll twice as much in `ScrollArea` etc. Inputs must be _consumed_, whereas currently in egui inputs are _read_.

[...]

PR #963, opened 2021-12-17
Dynamic sized strips, tables, and date picker

elwerene:
A (dynamic) grid and table layout widget which allows a mixed use of size constraints:
 * Absolute sizing
 * Relative sizing: relative to available size in dimension
 * Remainder: The remaining space is distributed evenly to all cells using Remainder Sizing
 * Relative/Remainder with minimum size
 
Also contains a date picker button which is based on the dynamic grid.

This is early in development and has definitely lots of bugs and rough edges.

I didn't want to wait longer as I love the results so far and I wanted to show it and open it up for discussion!

## TODO

I added many `TODO` comments in the code which I want as a minimum to get this released, like:
 * [ ] Positioning of date picker popup
 * [ ] Colors on date picker popup
 * [x] Internationalization/locale of date picker button/popup
 * [x] Padding in table/grid should be worked over/switched to a more egui like approach
 * [x] Is having two extra crates really needed? I developed it in extra grades but would not mind having it integrated into the main crate. I would prefer some integration into egui repository over maintaining external crates.


emilk:
This looks amazing 😍

emilk:
There's a lot to review here 😅 The table/grid code could really do with some documentation too, explaining the high level concepts.

Adding two new crates creates a lot of extra work to maintain. However, I have been considering creating a new crate `egui_extras` for stuff like this - useful widgets built on top of egui. I think both the table and datepicker could fit in there. We could later consider moving the table into `egui` proper once it has been battle tested and polished.

kevin406972552:
Looking forward to it. When will it be available?

elwerene:
@emilk I reworked most of the grid and table code to be more readable and with less boilerplate. Also I merged both crates into a new crate `egui_extras`. So maybe you could do another take in reviewing it if you find some time :)

egui_extras/src/sizing.rs
```diff
@@ -0,0 +1,85 @@
...
+pub struct Sizing {
+    sizes: Vec<Size>,
+}
+
+impl Sizing {
+    pub fn new() -> Self {
+        Self { sizes: vec![] }
+    }
+
+    pub fn add_size(&mut self, size: Size) {
```
emilk, line 30:
```suggestion
    pub fn add(&mut self, size: Size) {
```

it's the only thing you can add 🤷  :)

egui_extras/src/table.rs
```diff
@@ -0,0 +1,280 @@
+/// Table view with (optional) fixed header and scrolling body.
+/// Cell widths are precalculated with given size hints so we can have tables like this:
+/// | fixed size | all available space/minimum | 30% of available width | fixed size |
```
emilk, line 3:
```suggestion
//! Table view with (optional) fixed header and scrolling body.
//! Cell widths are precalculated with given size hints so we can have tables like this:
//! | fixed size | all available space/minimum | 30% of available width | fixed size |
```

egui_extras/src/table.rs
```diff
@@ -0,0 +1,280 @@
...
+use crate::{
+    layout::{CellSize, LineDirection},
+    sizing::Sizing,
+    Layout, Padding, Size,
+};
+
+use egui::{Response, Ui};
+use std::cmp;
+
+pub struct TableBuilder<'a> {
```
emilk, line 13: Needs docstring with simple example!

egui_extras/src/table.rs
```diff
@@ -0,0 +1,280 @@
...
+    layout: Layout<'a>,
+    widths: Vec<f32>,
+    striped: bool,
+    odd: bool,
+    start_y: f32,
+    end_y: f32,
+}
+
+impl<'a> TableBody<'a> {
+    /// Add rows with same height
```
emilk, line 157:
```suggestion
    /// Add rows with same height.
    ///
```

Should the height include the spacing?

egui_extras/src/table.rs
```diff
@@ -0,0 +1,280 @@
...
+            .col(|_| ()); // advances the cursor
+        }
+
+        let max_height = self.end_y - self.start_y;
+        let count = (max_height / height).ceil() as usize;
+        let end = cmp::min(start + count, rows);
+
+        if start % 2 != 0 {
+            self.odd = false;
+        }
```
emilk, line 183: surely we should always assign to `self.odd` ?

```suggestion
        self.odd = start % 2 == 1;
```

egui_extras/src/table.rs
```diff
@@ -0,0 +1,280 @@
...
+            });
+        });
+    }
+}
+
+pub struct TableBody<'a> {
+    layout: Layout<'a>,
+    widths: Vec<f32>,
+    striped: bool,
+    odd: bool,
```
emilk, line 151: instead of keeping track of `odd: bool`, perhaps keep track of `row_nr: usize`?

emilk (changes requested):
Very impressive! I think the new dynamic grid could be very useful. I have left a lot of questions though, and have a few more:

What happens if the contents of a cell overflows the width we allocated for it? Everything else in `egui` expands the parent when this happens (parent window gets wider, etc), and I would like for this to happen here as well, if possible. It is very difficult to know beforehand how much space you need for a widget, so to have the parent layout expand as needed is quite important. This is why `egui::Grid` keeps a memory of the sizes from previous frames. Of course, this adds a lot of complexity.

We also have a name-clash: `egui::Grid` and `egui_extras::Grid`. Can we come up with a another name?

Some other things:

* [x] Add `egui_extras` to `.github/workflows/rust.yml`
* [ ] Add `egui_extras` to `sh/check.sh`
* [ ] Please add some labels to explain the grid-demo (just a bunch of rectangles doesn't say much)
* [ ] Add source code links from the grid and table demos

songlinshu:
Looking forward to it. When will it be available?

elwerene:
> Looking forward to it. When will it be available?

I'm away for skiing, will be back in two weeks and continue working on it..

elwerene:
Thanks for the review!
I just started going through your (very valuable) feedback.

> What happens if the contents of a cell overflows the width we allocated for it? Everything else in `egui` expands the parent when this happens (parent window gets wider, etc), and I would like for this to happen here as well, if possible. It is very difficult to know beforehand how much space you need for a widget, so to have the parent layout expand as needed is quite important. This is why `egui::Grid` keeps a memory of the sizes from previous frames. Of course, this adds a lot of complexity.

You are right, that is isn't always easy to know how much space a widget needs and the grid and table use a very different approach. I will try to make this very clear in the documentation but I think it should stay like that as it makes dynamic designs possible. Maybe I can come up with an idea at some point, but for now I think you fiddle around with the sizes until it is good for your use case. Or would this be a show stopper for you?

> We also have a name-clash: `egui::Grid` and `egui_extras::Grid`. Can we come up with a another name?
Maybe egui_extras::Boxes? Or even FixedBoxes? This could help understanding, that they do not grow with its children.
 
>     * [x]  Add `egui_extras` to `.github/workflows/rust.yml`
>     * [x ]  Add `egui_extras` to `sh/check.sh`
>     * [x]  Please add some labels to explain the grid-demo (just a bunch of rectangles doesn't say much)
>     * [x]  Add source code links from the grid and table demos
I don't have permission to tick the checkboxes in your post :)

egui_extras/src/datepicker/popup.rs
```diff
@@ -0,0 +1,360 @@
...
+                    });
+                }
+
+                if self.arrows {
+                    grid.grid(|builder| {
+                        builder.sizes(Size::Remainder, 6).horizontal(|mut grid| {
+                            grid.cell(|ui| {
+                                ui.with_layout(Layout::top_down_justified(Align::Center), |ui| {
+                                    if ui
+                                        .button("<<<")
```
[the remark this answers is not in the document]
elwerene: Only if there's a way to make the dropdowns wider? Otherwise the layout would look off.

egui_extras/src/datepicker/popup.rs
```diff
@@ -0,0 +1,360 @@
...
+                                            Layout::centered_and_justified(Direction::TopDown),
+                                            |ui| {
+                                                ui.add(Label::new("Week"));
+                                            },
+                                        );
+                                    });
+                                }
+
+                                //TODO: Locale
+                                for name in ["Mo", "Di", "Mi", "Do", "Fr", "Sa", "So"] {
```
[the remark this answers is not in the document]
elwerene: I changed it to english for now

egui_extras/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+mod datepicker;
```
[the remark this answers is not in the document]
elwerene: I added one line, but I don't really know what would be a good, longer documentation here.

egui_extras/src/grid.rs
```diff
@@ -0,0 +1,158 @@
...
+    pub fn cell(&mut self, add_contents: impl FnOnce(&mut Ui)) {
+        self._cell(true, add_contents);
+    }
+
+    /// Add cell, content is not clipped
+    pub fn cell_noclip(&mut self, add_contents: impl FnOnce(&mut Ui)) {
+        self._cell(false, add_contents);
+    }
+
+    pub fn _grid(&mut self, clip: bool, grid_builder: impl FnOnce(GridBuilder)) {
```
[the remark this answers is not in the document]
elwerene: It's actually up to the ui which wrap mode happens? I just set a clip rect if clip is set. Do you think I should set the wrap mode in ui inside?

egui_extras/src/layout.rs
```diff
@@ -0,0 +1,164 @@
...
+    pos: Pos2,
+    max: Pos2,
+}
+
+impl<'l> Layout<'l> {
+    pub(crate) fn new(ui: &'l mut Ui, padding: Padding, direction: LineDirection) -> Self {
+        let mut rect = ui.available_rect_before_wrap();
+        rect.set_left(rect.left() + padding.outer + padding.inner);
+        rect.set_top(rect.top() + padding.outer + padding.inner);
+        rect.set_width(rect.width() - 2.0 * padding.outer);
```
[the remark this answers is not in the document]
elwerene: Good find!

egui_extras/src/padding.rs
```diff
@@ -0,0 +1,29 @@
+/// Configure padding of grid or table
+/// TODO: Use padding settings of egui/should we extend egui padding settings for table?
+#[derive(Clone, Debug)]
+pub struct Padding {
+    pub(crate) inner: f32,
+    pub(crate) outer: f32,
```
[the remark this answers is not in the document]
elwerene: I removed the padding in favor of egui item spacing

egui_extras/src/layout.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /// Set the rect so that the scrollview knows about our size
+    fn set_rect(&mut self) {
+        let mut rect = self.rect;
+        rect.set_right(self.max.x);
+        rect.set_bottom(self.max.y);
+
+        self.ui
+            .allocate_rect(rect, Sense::focusable_noninteractive());
```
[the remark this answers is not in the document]
elwerene: good idea!

elwerene:
@emilk It was an absolute joy to go through your suggestions and questions. Thanks for all your effort!

songlinshu:
hi,When will this widget be merged? thank you!

jean-airoldie:
I've been using this WIP widget and I think its pretty good, so here's my two cents.

* The duplicate `Grid` name is kind of annoying, but I can't say I can come up with a much better name. Maybe `DynGrid` or something made up like `Flex`. Another approach would be to have a separate module `table` in the `egui_extra` crate which would solve the duplicates.
* It's somewhat annoying to use the `row` and `rows` method of `TableBody` since it requires knowing the size of the fields ahead of time, which is not always possible. This is also limiting if you want to use the `rows` API, but some of the rows should have greater height. It would be cool if the `row` and `rows` method could take a `Size` as a parameter, but idk what the implications would be.
```rust
pub fn rows(mut self, height: Size, rows: usize, mut row: impl FnMut(usize, TableRow<'_, '_>)) {
```
* I wish there was a `Size` variant that would make a specific column behave like a grid. Specifically I mean that sometimes you just want the colum nto fit the size of the widget it contains, instead of having to find the right size manually by trial an error.

aevyrie:
@elwerene is there anything I can do to help move this along? Looks like you've incorporated most/all of Emil's feedback.

If you still have more work you want to complete with the date picker, would you be open to splitting out the grid and table functionality you've added as a separate PR so we can get that merged first? This is really great work, and I'm eager to use it. 🙂

elwerene:
@aevyrie I'm currently waiting for a new round of feedback of @emilk. I guess we can include the date picker even in the current state and improve it over time. I fixed most of the issues and todos I initially wrote about.

Let's not put too much pressure on @emilk as he seems to maintain this project in his free time. We will get there eventually :)

elwerene:
@jean-airoldie oh sorry, I thought I already answered your post.

You are right, that `Grid` is not a good name as it is already taken in the egui namespace. I proposed `Boxes` in https://github.com/emilk/egui/pull/963#issuecomment-1033642492

I named it `Size` as it can be horizontally or vertically in a `Grid` and I reuse the same struct for the table. In the table it can be a bit to general.

I think it's a good idea to have a table row variant with variable size, but it might not be possible with the current approach as the shading of odd rows would not have the right size. This also matches critique by @emilk that it is not a growing layout like all of egui is currently. I guess for now we might have to live with that problem as I don't have a good idea right now how to do this in one pass and I don't like the flickering of two passes. @emilk Maybe we could have a mode where we render once, discard it and render again with correct sizes?

I have to look into why the separator does not work as intended, what was the issue?

aevyrie:
> Let's not put too much pressure on @emilk as he seems to maintain this project in his free time. We will get there eventually :)

Yup! I was thinking the merge and review process might be made easier if this was split into two PRs - the grid (1) and date picker (2). It's pretty sizeable as it is. 🙂 Totally up to you - thought I would suggest if it made things simpler.

jean-airoldie:
> I have to look into why the separator does not work as intended, what was the issue?

I guess In that case I'm talking more about traditional table separators where there is a clear line between each column and row. The separators don't cover the entire row or column and are kinda hard to use in that context, unless I misunderstood how to use them.

> I don't like the flickering of two passes

Yeah I also dislike that very much so I understand your rationale. Alternatively It would be helpful if there were some utility functions to be able to calculate the size of a given element ahead of time so we could calculate a proper height.

egui_extras/src/grid.rs
```diff
@@ -0,0 +1,189 @@
...
+    layout::{CellSize, Layout, LineDirection},
+    sizing::Sizing,
+    Size,
+};
+use egui::{Response, Ui};
+
+/// The direction in which cells are positioned in the grid.
+///
+/// In a horizontal grid cells are positions from left to right.
+/// In a vertical grid cells are positions from top to bottom.
```
emilk, line 11: This is still ambiguous. Consider this order:

```
0 1 2 3
4 5 6 7
```

This is positioned both from left to right _and_ from top to bottom. Is it "horizontal" or "vertical"?

In linear algebra the above would be called "row-major", while this would be "column-major":

```
0 2 4 6
1 3 5 7
```

Perhaps these are better terms? If this is indeed what `GridDirection` even means.

egui_extras/src/layout.rs
```diff
@@ -0,0 +1,165 @@
...
+
+/// Cells are positioned in two dimensions
+///
+/// In a grid there's only one line which goes into the orthogonal direction of the grid:
+///
+/// In a horizontal grid, a `[Layout]` with vertical `[LineDirection]` is used.
+/// Its cells go from left to right inside this `[Layout]`.
+///
+/// In a table there's a `[Layout]` for each table row with a horizontal `[LineDirection]`.
+/// Its cells go from left to right.
```
emilk, line 19: Perhaps some ascii art could help here

emilk:
The top-level types need some documentation. Running `cargo doc --no-deps -p egui_extras --open` gives this:

Screen Shot 2022-03-12 at 08 15 59

(no summary of what these are)

emilk:
> Maybe we could have a mode where we render once, discard it and render again with correct sizes?

I hope to be able to work on this one day: https://github.com/emilk/egui/issues/843

emilk:
> You are right, that `Grid` is not a good name as it is already taken in the egui namespace. I proposed `Boxes` in [#963 (comment)](https://github.com/emilk/egui/pull/963#issuecomment-1033642492)

`Boxes` can work! I agree that we should avoid name clashes

emilk (changes requested):

Bromeon:
> > You are right, that `Grid` is not a good name as it is already taken in the egui namespace. I proposed `Boxes` in [#963 (comment)](https://github.com/emilk/egui/pull/963#issuecomment-1033642492)
> 
> `Boxes` can work! I agree that we should avoid name clashes

`Boxes` is not ideal, as it's already [heavily used](https://docs.rs/egui/latest/egui/struct.Grid.html?search=box):
* `BoxPlot`, `BoxElem`, `BoxSpread` plot types
* `ComboBox` and `Checkbox` widgets
* Standard `Box` smart pointer

What is the conceptual difference between the `Grid` here and the existing one?
Maybe the answer to this indicates a better term.

egui_extras/src/datepicker/popup.rs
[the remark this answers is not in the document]
elwerene: @emilk I guess you did not see this comment? Github is sometimes very unclear.

egui_extras/src/grid.rs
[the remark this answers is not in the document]
elwerene: @emilk I guess you did not see this comment? Github is sometimes very unclear.

egui_extras/src/grid.rs
elwerene: A grid in my sense is only one-directional and can have sub-grids which have a different direction, so it is either
```
0 1 2 3
```

or

```
0
1
2
3
```

with subgrid it could look like this:

```
0 A 2 3
  B
  C
```

elwerene:
> > > You are right, that `Grid` is not a good name as it is already taken in the egui namespace. I proposed `Boxes` in [#963 (comment)](https://github.com/emilk/egui/pull/963#issuecomment-1033642492)
> > 
> > 
> > `Boxes` can work! I agree that we should avoid name clashes
> 
> `Boxes` is not ideal, as it's already [heavily used](https://docs.rs/egui/latest/egui/struct.Grid.html?search=box):
> 
>     * `BoxPlot`, `BoxElem`, `BoxSpread` plot types
> 
>     * `ComboBox` and `Checkbox` widgets
> 
>     * Standard `Box` smart pointer
> 
> 
> What is the conceptual difference between the `Grid` here and the existing one? Maybe the answer to this indicates a better term.

You are right, `Box` is a used term in rust/egui. Maybe `Strip` as it would reflect more how it works like I explained in https://github.com/emilk/egui/pull/963#discussion_r826700133

egui_extras/src/grid.rs
elwerene: Documentation is hard, I tried to write a better description and changed some struct and variant names to make it easier to understand.

elwerene:
> The top-level types need some documentation. Running `cargo doc --no-deps -p egui_extras --open` gives this:
> 
> Screen Shot 2022-03-12 at 08 15 59
> 
> (no summary of what these are)
> 
> Also some warnings:
> 
> ```
> warning: unresolved link to `count`
>   --> egui_extras/src/grid.rs:65:41
>    |
> 65 |     /// Add size hint for columns/rows [count] times
>    |                                         ^^^^^ no item named `count` in scope
>    |
>    = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
>    = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
> 
> warning: unresolved link to `count`
>   --> egui_extras/src/table.rs:88:35
>    |
> 88 |     /// Add size hint for column [count] times
>    |                                   ^^^^^ no item named `count` in scope
>    |
>    = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
> ```

I added the missing comments.

egui_extras/src/grid.rs
emilk: > A grid in my sense is only one-directional and can have sub-grids which have a different direction, so it is either
> 
> ```
> 0 1 2 3
> ```
> 
> or
> 
> ```
> 0
> 1
> 2
> 3
> ```

Oh, this was definitely surprising to me. I would not call a one-dimensional list a "grid".

elwerene: You could form a grid, but you are right, it was not a good name :roll_eyes:

egui_extras/src/layout.rs
```diff
@@ -0,0 +1,163 @@
...
+/// Its cells go from left to right. And the lines go from top to bottom.
+pub(crate) enum CellDirection {
+    /// Cells go from left to right
+    Horizontal,
+    /// Cells go fromtop to bottom
+    Vertical,
+}
+
+/// Positions cells in `[CellDirection]` and starts a new line on `[Layout::end_line]`
+pub struct Layout<'l> {
```
emilk, line 28: Another potential name clash problem: there is already an `egui::Layout`, though it is not `pub`.

`Layout` is also a very general name. Perhaps `StripLayout` or something would work?

emilk (commented):
I would like to get this in so I can play with it! What is left to do?

elwerene:
> I would like to get this in so I can play with it! What is left to do?

Only one open discussion point which is not a showstopper imho:

https://github.com/emilk/egui/pull/963#discussion_r802574712

The datepicker is more like a demo of the strip, a prototype widget which can be evolved to a full working widget in the future.

egui_extras/src/datepicker/popup.rs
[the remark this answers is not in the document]
emilk: You mean the `ComboBox`:es above? https://docs.rs/egui/latest/egui/containers/struct.ComboBox.html#method.width

elwerene:
@emilk I don't know how to get rid of your requested change as it is outdated?

egui_extras/src/datepicker/popup.rs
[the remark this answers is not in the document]
elwerene: @emilk Yes! I don't know how I could not find `width` :facepalm: previously

elwerene:
I can see you typo fix

emilk:
But can you see the nine other commits at https://github.com/emilk/egui/commits/dynamic-grid ?
Screen Shot 2022-03-31 at 15 12 35

emilk:
In the meantime: there is a bug where a vertical `Strip` doesn't get the width of its child elements. This means it is possible to make the table demo window far more narrow than it should be:

Screen Shot 2022-03-31 at 15 14 53

elwerene:
> But can you see the nine other commits at https://github.com/emilk/egui/commits/dynamic-grid ?
> 
> Screen Shot 2022-03-31 at 15 12 35
> 
> Do you get them if you `pull`?

You are very active! :heart: 
I got them by merging your dynamic-grid into mine :shrug:

elwerene:
> Do you know what is causing the width-resize bug? Seems like it should be a fairly easy fix (I hope)

I was just inspecting it, I'm calling `ui.allocate_rect` for each cell to make them clickable and on the whole Strip/TableHeader/TableBody so the scroll view knows about them. Did something change in that logic?

emilk:
Nothing changed, and also the problem exists in [2ec346e](https://github.com/emilk/egui/pull/963/commits/2ec346e7f5a1c15753958b327dff73cb5b08c586) (before any of my commits).

elwerene:
I introduced the error in ae71f91f8c3b68eed1780e146326bc67a1767c01

time to find the bug!

elwerene:
Oh I guess it was me moving the table inside of the grid :facepalm:

elwerene:
This is a shortcoming of the current implementation. The strip only has one fixed dimension, the other dimension takes whatever there's left. We could add a minimum size here. This could also be added later as an option to the builder. What do you say, @emilk?

emilk (approved):
Awesome work!

emilk:
@elwerene `chrono` needs a version bump, but it has a lot of deprecated functions. Would you like to help out?

elwerene:
@emilk Sure, I'll have a look some time this week :)

Issue #965, opened 2021-12-17
Horizontal scroll reversed in Windows native builds


aevyrie:
**Describe the bug**
Horizontal scrolling does not respect Windows preferences.

**To Reproduce**
1. Build a native demo on windows
2. Open the plot demo
3. Scroll horizontally and vertically
4. Do the same on the web to compare

**Expected behavior**
Horizontal scrolling, like vertical scrolling, should respect the "native scroll" default behavior.

**Desktop (please complete the following information):**
 - OS: Windows 10/11

emilk:
Duplicate: https://github.com/emilk/egui/issues/356

emilk:
Can you please open a winit issue on this?

aevyrie:
> Can you please open a winit issue on this?

Yes, I forgot to link these issues/PRs, my bad. I filed an issue and submitted a PR upstream: https://github.com/rust-windowing/winit/pull/2101

This egui issue and PR (#966) are only useful as workarounds if the winit issue is not merged for some reason.

PR #966, opened 2021-12-17
Reverse horizontal scrolling on windows

aevyrie:
Closes <https://github.com/emilk/egui/issues/965>.

### DIFF
--- a/egui-winit/src/lib.rs
+++ b/egui-winit/src/lib.rs
@@ -459,7 +459,7 @@ impl State {
             delta.x *= -1.0;
         }
         if cfg!(target_os = "windows") {
-            delta.x *= -1.0;
+            delta.x *= -1.0; // until https://github.com/rust-windowing/winit/pull/2101 is merged
         }
 
         if self.egui_input.modifiers.ctrl || self.egui_input.modifiers.command {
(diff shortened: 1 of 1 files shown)


egui-winit/src/lib.rs
```diff
@@ -458,6 +458,9 @@ impl State {
             // https://github.com/rust-windowing/winit/issues/1695 being closed
             delta.x *= -1.0;
         }
+        if cfg!(target_os = "windows") {
+            delta.x *= -1.0;
```
emilk, line 462: if this is wrong on mac and windows, perhaps it is also wrong on linux? i.e perhaps we should just always invert `delta.x`no matter what platform

aevyrie: I agree. However, it's not clear to me how we find the correct answer. As mentioned in #356, it appears that it's not even consistent across Linux desktop environments.

The only dev environment I have readily available is windows. I didn't want to cause problems on systems I can't test, and I wanted to keep this PR tightly scoped.

Most importantly, this is really a workaround for an upstream winit issue. If I can get this https://github.com/rust-windowing/winit/pull/2101 merged, the egui issue and PR for horizontal scrolling on Windows can be closed.

emilk, line 462:
```suggestion
            delta.x *= -1.0; // until https://github.com/rust-windowing/winit/pull/2101 is merged
```

emilk (approved):

PR #971, opened 2021-12-19
Create & show CollapsingHeader with the given Id...

atahabaki:
Hi, I loved your work of art, the `egui`. It's perfect.

In my case I'm using `ui.collapse` with the same label twice, this of course id crashes, I've created two functions in order to solve this problem:

1. `CollapsingHeader::with_id(text: impl Into<WidgetText>, id: impl std::hash::Hash)`: create an Instance of the `CollapsingHeader` with the `id`. 
2. `ui.collapsing_with_id(&mut self, heading: impl Into<WidgetText>, id: impl std::hash::Hash, add_contents: impl FnOnce(&mut Ui))`: Create and show the `CollapsingHeader` with the given id.

Basically, they're almost the same functions as the constructor functions... 

Anyway, you got the point, thanks, take care!

### DIFF
--- a/egui/src/containers/collapsing_header.rs
+++ b/egui/src/containers/collapsing_header.rs
@@ -171,7 +171,7 @@ impl CollapsingHeader {
             show_background: false,
         }
     }
-    
+
     /// By default [`CollapsingHeader`] uses label as [`Id`], here you can set a
     /// custom `Id` for it.
     ///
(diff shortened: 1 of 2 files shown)


emilk:
You can already do this using `CollapsingHeader::new(…).id_source(…)`, as mentioned in the docs: https://docs.rs/egui/latest/egui/containers/struct.CollapsingHeader.html#method.new

As for `ui.collapsing_with_id`: right now I only want the most used things shortcuts in `Ui` so that interface doesn't get too bloated.

PR #990, opened 2021-12-23
docs(readme): mention the discord server

yusdacra:
Adds the discord server invite to the README. This confused me for a while, as I expected such information to be in README rather than the CONTRIBUTING.md file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,9 @@ If you just want to write a GUI application in Rust (for the web or for native),
 
 If you want to integrate egui into an existing engine, go to the [Integrations](#integrations) section.
 
-If you have questions, use [Discussions](https://github.com/emilk/egui/discussions). If you want to contribute to egui, please read the [Contributing Guidelines](https://github.com/emilk/egui/blob/master/CONTRIBUTING.md)
+If you have questions, use [Discussions](https://github.com/emilk/egui/discussions).
+A Discord server is available at https://discord.gg/vbuv9Xan65.
+If you want to contribute to egui, please read the [Contributing Guidelines](https://github.com/emilk/egui/blob/master/CONTRIBUTING.md).
 
 ## Demo
 


coderedart:
discord invite link is not advertised on purpose. 
The developer wanted to discuss the egui development in particular, and not create a place where people will ask questions which could be answered by reading the docs/readme/discussions, especially when these questions are repeated again and again. 
apparently, history has shown how this ends: https://twitter.com/ocornut/status/1364725878907895810?lang=en :)

yusdacra:
Hm, i see, but i don't think thats a valid excuse for completely removing a community (in imgui's case). I think instead actively enforcing rules like "open a thread on discussions, if you get answer on discord post it on your thread" style can be a better solution. A realtime chat platform like Discord offers better hands-on help experience compared to a forum-like platform like GitHub Discussions, so i don't think it's a valid reason to not advertise. But of course, if it's set in stone then i'm fine with it.

coderedart:
yeah, discord is easier to ask/receive help, but its also going to lead to a lot of low effort questions. some people literally just want you to write the code and give it to them :D

It's the developer's decision in the end, but public discord invite link in README would also need moderators to handle the influx of users, enforcing of rules and dealing with bans/mutes drama (of which discord usually has a lot of). so, that would need to be taken care of *before* opening the server for others.
 
there's also the option to link the rust community discord, as it has a gui channel where people already ask bunch of questions related to egui or other gui frameworks. development discord can stay as it is and the users of the library can be directed to the community discord without any effort from the egui developers.

emilk:
@coderedart is completely correct about my hesitation about having an open invitation to the discord server. However, I learned a lot of programming  on IRC chats back in the day, and if we are able to build a friendly community on discord it would be wonderful, so I'm willing to give it a shot! If it turns out bad I can always remove the invites again.

A link and badge has been added in https://github.com/emilk/egui/commit/ee002145117df5be9c8889bfbcf569f3493dec79

Issue #116, opened 2021-01-16
How do I hide the console on Windows?


junknet:

manokara:
Put `#![windows_subsystem = "windows"]` at the top of your `main.rs` file.

junknet:
Thx

ubearline:
mark

Piping:
how can we hide console window on mac? currently on mac the eframe_template opens console window when the app is opened by clicking the icon.

PR #1008, opened 2021-12-28
Hide console on windows for all examples

emilk:
See https://github.com/emilk/egui/discussions/1003 and https://github.com/emilk/egui/issues/116

Please take a look @TicClick @BctfN0HUK7Yg

If this works as expected I will add the same to https://github.com/emilk/eframe_template

### DIFF
--- a/eframe/examples/file_dialog.rs
+++ b/eframe/examples/file_dialog.rs
@@ -1,3 +1,5 @@
+#![windows_subsystem = "windows"] // hide console window on Windows
+
 use eframe::{egui, epi};
 
 #[derive(Default)]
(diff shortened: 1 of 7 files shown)


parasyte:
It is probably best to use something like this:

```rust
#![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]
```

So that any test failures will still print to the console on Windows.

TicClick:
it does work as intended, but I agree with @parasyte here to make it work only for `cargo build --release`

TicClick (approved):

parasyte (approved):

emilk:
Thanks for the feedback!

PR #1011, opened 2021-12-28
Added mappings for - and = keys

blackdragon2447:


emilk (changes requested):
Please update `egui-web` too!

blackdragon2447:
i wasnt aware that web had a seperate implementation, my excuse.

egui_web/src/lib.rs
```diff
@@ -466,6 +466,9 @@ pub fn translate_key(key: &str) -> Option<egui::Key> {
         "y" | "Y" => Some(egui::Key::Y),
         "z" | "Z" => Some(egui::Key::Z),
 
+        "-" => Some(egui::Key::Minus),
+        "=" => Some(egui::Key::Equals),
```
emilk, line 470: Did you test this?

emilk:
Note that not all keyboard have keys for "minus" and "equals" (e.g. Swedish keyboards don't), so using them for keyboard shortcuts will be quite annoying for a lot of people.

Mingun:
I think that you mean that _keyboard layout_ for Swedish does not have "minus" and "equals" keys, but that keys definitely exists on the keyboard, just this symbols are produced in another layout. GUI library should not worry about that, it should use scan codes that identifies _physical key_, not the _symbol_ that that key can produce.

There is a problem, how to get that scan codes from accelerator string, though. Perhaps GUI toolkits use some fixed layout to map a symbol to a correct physical key and do that regardless of the existence of this layout in the user system. It is not expected that accelerators to stop working due to the switching of the layout.

emilk:
> GUI library should not worry about that, it should use scan codes that identifies physical key, not the symbol that that key can produce.

Are you sure? So if I the user presses the button to the right of Tab it should always emit a `Key::Q`, even if the user is using Dvorak?

There are use cases for scancodes (keybindings in games and such), but most gui app code will rather show somerthing like "press Shift-Q to select this tool" and have that map directly to `Key::Q`, and hope the user has a keyboard with a `Q` key on it.

Mingun:
> Are you sure? So if I the user presses the button to the right of Tab it should always emit a `Key::Q`, even if the user is using Dvorak?

Exactly. Because I press the physical key, not a key to input `Q`. The `Q` here just a locator, it can be `Й` in Russian in `ЙЦУКЕН` layout. I use `Ctrl+C`/`Ctrl+V` when I write texts on Russian and I not thinking about that in Russian layout it is a `Ctrl+С`/`Ctrl+М`.

> but most gui app code will rather show somerthing like "press Shift-Q to select this tool"

`Q` here just a locator, because usually keyboard keys a known under their Latin key-codes from QWERTY layout. I think, it is not necessary that symbol `Q` was actually painter on the corresponding key. Of course, sometimes applications can translate those shortcuts and show an actual symbol painted on it, but this is more rare than the rule. Usually QWERTY layout are installed on every computer, so it is a good choice to choose key names from that layout. Of course, I may wrong, need to check behavior of the popular frameworks.

emilk:
@Mingun Does egui:s ctrl-C/V even work as expected for you?

Mingun:
- Web (https://www.egui.rs/#demo): `Ctrl+С` / `Ctrl+V` are worked in both Russian and English layout, `Ctrl+A` (select all) is worked only in English
- Windows 10 (`cargo run`): all those shortcuts are worked in both layouts

emilk:
> I use `Ctrl+C`/`Ctrl+V` when I write texts on Russian and I not thinking about that in Russian layout it is a `Ctrl+С`/`Ctrl+М`.

Am I really understanding you correct that when you press the key marked `M` on your keyboard, winit and egui emits `Key::V`?

emilk:
stale

PR #1019, opened 2021-12-31
0.16.1 patch release: Add back CtxRef::begin_frame/end_frame

emilk:
`begin_frame`, `end_frame` is more convenient when using egui in a game engine. In particular, 0.16.0 was incompatible with <https://github.com/mvlabat/bevy_egui>.

### DIFF
--- a/egui/src/context.rs
+++ b/egui/src/context.rs
@@ -99,19 +99,57 @@ impl CtxRef {
     ///
     /// This will modify the internal reference to point to a new generation of [`Context`].
     /// Any old clones of this [`CtxRef`] will refer to the old [`Context`], which will not get new input.
+    ///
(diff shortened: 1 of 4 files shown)


egui/src/context.rs
```diff
@@ -99,19 +99,57 @@ impl CtxRef {
     ///
     /// This will modify the internal reference to point to a new generation of [`Context`].
     /// Any old clones of this [`CtxRef`] will refer to the old [`Context`], which will not get new input.
+    ///
+    /// In the future [`Self::run`] may be extended to do multi-pass,
+    /// i.e. call the application code twice. This will make
```
crumblingstatue, line 104: This will make what?

emilk: This will make an incomplete sentence!

emilk:
Published!

PR #1023, opened 2022-01-01
Use `instant` crate in egui-winit for WebAssembly support

### LINKED ISSUE #1022: std::time::Instant not supported by WebAssembly
The `std::time::Instant` type used in the egui-winit crate is not supported when running through WebAssembly:

```
panicked at 'time not implemented on this platform', library/std/src/sys/wasm/../unsupported/time.rs:13:9
```

WebAssembly could potentially be supported in egui-winit by using a crate such as https://github.com/sebcrozet/instant instead, which only re-exports the `std::time::Instant` type on native but implements a polyfill for WebAssembly targets.

niklaskorz:
Closes #1022.

### DIFF
--- a/egui-winit/Cargo.toml
+++ b/egui-winit/Cargo.toml
@@ -24,7 +24,7 @@ all-features = true
 [dependencies]
 egui = { version = "0.16.0", path = "../egui", default-features = false, features = ["single_threaded"] }
 winit = "0.26"
-instant = "0.1"
+instant = { version = "0.1", features = ["wasm-bindgen"] }
 
 epi = { version = "0.16.0", path = "../epi", optional = true }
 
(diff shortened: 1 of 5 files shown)


egui-winit/Cargo.toml
```diff
@@ -24,6 +24,7 @@ all-features = true
 [dependencies]
 egui = { version = "0.16.0", path = "../egui", default-features = false, features = ["single_threaded"] }
 winit = "0.26"
+instant = { version = "0.1", features = ["wasm-bindgen"] }
```
emilk, line 27:
```suggestion
instant = { version = "0.1", features = ["wasm-bindgen"] }
winit = "0.26"
```

emilk (approved):

PR #1027, opened 2022-01-01
Faster dashed line generation

lampsitter:
The `Shape::dashed_line` was quite hot in my program, so I decided to optimize it.
In addition to the original function I added an alternative function, which allows reuse of an existing Vec.

Old:
```
running 1 test
test single ... bench:      81,939 ns/iter (+/- 7,938)
```
New:
```
running 2 tests
test single ... bench:      19,690 ns/iter (+/- 1,336)
test many   ... bench:      10,081 ns/iter (+/- 347)
```

Benchmark code:
```rust
#![feature(test)]

extern crate test;
use eframe::egui::{Shape, Stroke};
use test::Bencher;

#[bench]
fn single(b: &mut Bencher) {
    b.iter(|| {
        let mut v = Vec::new();
        use eframe::egui;

        let line = [
            egui::pos2(0.0, 0.0),
            egui::pos2(50.0, 0.0),
            egui::pos2(100.0, 1.0),
        ];

        for _ in 0..100 {
            v.extend(Shape::dashed_line(
                &line,
                Stroke::new(1.5, eframe::egui::Color32::RED),
                10.0,
                2.5,
            ));
        }

        test::black_box(v);
    });
}

// Must be commented out when run against master
#[bench]
fn many(b: &mut Bencher) {
    b.iter(|| {
        let mut v = Vec::new();
        use eframe::egui;

        let line = [
            egui::pos2(0.0, 0.0),
            egui::pos2(50.0, 0.0),
            egui::pos2(100.0, 1.0),
        ];

        for _ in 0..100 {
            Shape::dashed_line_many(
                &line,
                Stroke::new(1.5, eframe::egui::Color32::RED),
                10.0,
                2.5,
                &mut v,
            );
        }

        test::black_box(v);
    });
}

```

### DIFF
--- a/epaint/benches/benchmark.rs
+++ b/epaint/benches/benchmark.rs
@@ -1,0 +1,43 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+
+use epaint::{pos2, Color32, Shape, Stroke};
+
+fn single_dashed_lines(c: &mut Criterion) {
+    c.bench_function("single_dashed_lines", move |b| {
+        b.iter(|| {
+            let mut v = Vec::new();
+
+            let line = [pos2(0.0, 0.0), pos2(50.0, 0.0), pos2(100.0, 1.0)];
+
+            for _ in 0..100 {
+                v.extend(Shape::dashed_line(
+                    &line,
+                    Stroke::new(1.5, Color32::RED),
+                    10.0,
+                    2.5,
+                ));
+            }
+
+            black_box(v);
+        });
+    });
+}
+
+fn many_dashed_lines(c: &mut Criterion) {
+    c.bench_function("many_dashed_lines", move |b| {
+        b.iter(|| {
+            let mut v = Vec::new();
+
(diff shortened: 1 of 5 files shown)


emilk (approved):
Great! Feel free to also add the benchmark, perhaps to a new `epaint/benches/benchmark.rs` file?

epaint/benches/benchmark.rs
```diff
@@ -0,0 +1,40 @@
+#![feature(test)]
```
emilk, line 1: This is an unstable feature - either use [Criterion](https://crates.io/crates/criterion) (which `egui_demo_lib` uses), or if that's too much headache, just remove the benchmarks again :)

emilk (changes requested):

epaint/benches/benchmark.rs
lampsitter: Oh, okay. I considered using criterion but was not sure. I will do that then.

emilk:
thanks!

PR #1028, opened 2022-01-01
added splitter

### LINKED ISSUE #944: [Musthave] Horizontal and vertical splitter
Splits frame in half with access to each frame by two lambdas for each half

Not being able to place elements in one line with different directions

Horizontal and vertical splitter that splits frame in half

-

**Additional context**

Couldn't think of a way to split by n times with access to different frames at once, there is a way to make a different function for each n way split, but it would be messy for sure

LordMZTE:
Closes #944.

image

The splitter currently still has some issues, which I'm not quite sure how to fix:
- Elements can overlap when the splitter is made too small
- The splitter only allocates the current available size, and not the size the containing widgets actually need.

Guidance would be much appreciated!

### DIFF
--- a/egui/src/containers/mod.rs
+++ b/egui/src/containers/mod.rs
@@ -10,6 +10,7 @@ pub mod panel;
 pub mod popup;
 pub(crate) mod resize;
 pub(crate) mod scroll_area;
+pub(crate) mod splitter;
 pub(crate) mod window;
 
 pub use {
@@ -21,5 +22,6 @@ pub use {
     popup::*,
     resize::Resize,
     scroll_area::ScrollArea,
+    splitter::{Splitter, SplitterOrientation},
     window::Window,
 };
--- a/egui/src/containers/splitter.rs
+++ b/egui/src/containers/splitter.rs
@@ -1,0 +1,157 @@
+use epaint::{
+    emath::{lerp, Align},
+    pos2, vec2,
+};
+
+use crate::{egui_assert, Layout, Response, Sense, Ui};
+
+/// A splitter which can separate the UI into 2 parts either vertically or horizontally.
+///
+/// ```
+/// # egui::__run_test_ui(|ui| {
+/// egui::Splitter::vertical().show(ui |ui_left, ui_right| {
+///     ui_left.label("I'm on the left!");
+///     ui_right.label("I'm on the right!");
+/// })
+/// # });
+/// ```
+#[must_use = "You should call .show()"]
+pub struct Splitter {
+    orientation: SplitterOrientation,
+    ratio: f32,
+}
+
+impl Splitter {
+    /// Create a new splitter with the given orientation and a ratio of 0.5.
+    pub fn with_orientation(orientation: SplitterOrientation) -> Self {
+        Self {
+            orientation,
+            ratio: 0.5,
+        }
+    }
+
+    /// Create a new vertical splitter with a ratio of 0.5.
+    #[inline]
+    pub fn vertical() -> Self {
+        Self::with_orientation(SplitterOrientation::Vertical)
+    }
+
+    /// Create a new horizontal splitter with a ratio of 0.5.
+    #[inline]
+    pub fn horizontal() -> Self {
+        Self::with_orientation(SplitterOrientation::Horizontal)
+    }
+
+    /// Set the ratio of the splitter.
+    ///
+    /// The ratio sets where the splitter splits the current UI, where, depending on the
+    /// orientation, 0.0 would mean split at the very top/left and 1.0 would mean split at the very
+    /// bottom/right respectively. The ratio must be in the range 0.0..=1.0.
+    pub fn ratio(mut self, ratio: f32) -> Self {
+        self.ratio = ratio;
+        self
+    }
+
+    #[inline]
+    pub fn show<R>(
+        self,
+        ui: &mut Ui,
+        add_contents: impl FnOnce(&mut Ui, &mut Ui) -> R,
+    ) -> SplitterResponse<R> {
+        self.show_dyn(ui, Box::new(add_contents))
+    }
+
+    pub fn show_dyn<'c, R>(
+        self,
+        ui: &mut Ui,
+        add_contents: Box<dyn FnOnce(&mut Ui, &mut Ui) -> R + 'c>,
+    ) -> SplitterResponse<R> {
+        let Self { orientation, ratio } = self;
+
+        egui_assert!((0.0..=1.0).contains(&ratio));
+
+        let (rect, splitter_response) =
+            ui.allocate_exact_size(ui.available_size_before_wrap(), Sense::hover());
+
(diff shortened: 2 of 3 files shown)


parasyte:
This is awesome. Just tried the demo and I have a few suggestions:

- Pass two closures to the show method, instead of one closure that takes two `Ui`s. Subjective, but this just feels more natural. This is also what was requested in #944.

```rust
Splitter::with_orientation(self.splitter_orientation)
    .ratio(self.splitter_ratio)
    .show(
        ui,
        |ui| {
            ui.label("Left/Top");
            ui.label("Left/Top");
            ui.label("Left/Top");
        },
        |ui| {
            ui.label("Right/Bottom");
            ui.label("Right/Bottom");
            ui.label("Right/Bottom");
        },
    );
```

- Make the splitter line optionally draggable so it can be moved with the cursor. Like the panel containers (`resizable` method).
- I think it makes sense to use the available width for vertical layouts, and the available height for horizontal layouts; let the caller define the remaining dimension.
  - A configurable size would solve the second issue you mention.
- It might be worth wrapping both child UIs in `ScrollArea`s. I tried this, and I really prefer the `Splitter` to do it instead of writing two nested closures for each side of the split. (Or two scroll areas in a single closure with two `Ui`s.)
  - It's hard to consider this and also not want to make these scroll areas configurable. Defaulting to `both()` with auto-shrink disabled is quite nice, though.
  - When I prototyped this, I also had to make the child labels non-wrapping so the scrollbars would show when needed.

LordMZTE:
Thanks! I'm glad you like it!
> * Pass two closures to the show method, instead of one closure that takes two `Ui`s. Subjective, but this just feels more natural. This is also what was requested in [Horizontal and vertical splitter #944](https://github.com/emilk/egui/issues/944).

Well, the reason I originally went with 2 closures is that emilk suggested it as an API in the original issue where this was discussed. I also personally prefer it, as there's no question about which order the closures will be called in and so on.

> * Make the splitter line optionally draggable so it can be moved with the cursor. Like the panel containers (`resizable` method).

Sounds like a useful feature, but to be honest I'm kind of scared at the math needed for this. Also I'm not sure how I'd fit it in alongside the current ratio implementation.

> * I think it makes sense to use the available width for vertical layouts, and the available height for horizontal layouts; let the caller define the remaining dimension.
>   * A configurable size would solve the second issue you mention.

That sounds like a great idea! Although I'm not quite sure if the intended way of doing this kinda thing Isn't just to place the widget inside one that restricts its available space. Also this would make the API a bit less flexible for users who "just want a splitter" and don't really care all that much about precise layout and figuring out dimensions.


> * It might be worth wrapping both child UIs in `ScrollArea`s. I tried this, and I really prefer the `Splitter` to do it instead of writing two nested closures for each side of the split. (Or two scroll areas in a single closure with two `Ui`s.)
>   * It's hard to consider this and also not want to make these scroll areas configurable. Defaulting to `both()` with auto-shrink disabled is quite nice, though.
>   * When I prototyped this, I also had to make the child labels non-wrapping so the scrollbars would show when needed.

That sounds like it would be a good way to go about fixing the clipping issue. I think that's worth doing, although I should probably add a way for the caller to disable this so they can figure out their own layouts and have more freedom.


BTW if you feel like working on this, you'd be more than welcome! Feel free to just open a PR on my fork on the `feature/splitter` branch.

emilk:
> * Pass two closures to the show method, instead of one closure that takes two `Ui`s

This will create borrowing issues, so I don't think it is a good idea.

parasyte:
> This will create borrowing issues, so I don't think it is a good idea.

That is a great point! It can be resolved by proving two `show` methods:

```rust
Splitter::with_orientation(self.splitter_orientation)
    .ratio(self.splitter_ratio)
    .show_left(ui, |ui| {
        ui.label("Left/Top");
        ui.label("Left/Top");
        ui.label("Left/Top");
    })
    .show_right(ui, |ui| {
        ui.label("Right/Bottom");
        ui.label("Right/Bottom");
        ui.label("Right/Bottom");
    });
```

It also resolves the other issue about ordering.

In this case, `show_left` consumes `self` and returns a type that implements `show_right`, which then finally returns a response with results from both closures. (Bikeshedding welcome on the names, but this should be acceptable. and IMHO it looks more reasonable than mixing two `Ui` refs in one closure. The clear delineation is nice.)

parasyte:
> Although I'm not quite sure if the intended way of doing this kinda thing Isn't just to place the widget inside one that restricts its available space. Also this would make the API a bit less flexible for users who "just want a splitter" and don't really care all that much about precise layout and figuring out dimensions.

This sounds completely reasonable.

I would want the demo to include such a wrapper (and of course, documentation) so that it is obvious that the expectation is for the user to allocate space _in some way_. Adding a spitter to some container is a fine way to allocate space without really caring to define dimensions. (This is what the demo currently does in this PR; `Window` is the container. Except it doesn't gracefully handle the situation where all of the collapsing headers are expanded in a small window.)

LordMZTE:
Well, a big part of the reason I'm using only 1 closure is the response type that `Splitter::show` returns:
```rust
/// The response of showing a Splitter
pub struct SplitterResponse<R> {
    /// The return value of the closure passed into show.
    pub body_returned: R,
    /// The response of the top or left UI depending on the splitter's orientation.
    pub top_left_response: Response,
    /// The response of the bottom or right UI depending on the splitter's orientation.
    pub bottom_right_response: Response,
    /// The response of the whole splitter widget.
    pub splitter_response: Response,
}
```

With a show method that takes 2 closures, we'd need 2 fields (and 2 generic parameters) for the return type of both the closures. If we had 2 show functions, I'm not sure how we could make them chainable like you suggested, since that would make this response handling pretty much impossible.

LordMZTE:
I don't really like this. I think this is just more complex than it needs to be. Also this requires calling `show_left` before calling `show_right`, and I don't really understand what the point is in not just using one closure. That would allow for more flexibility too (like declaring a variable when creating one side and using it on the other, and letting the user choose which side to paint first).

parasyte:
I don't think any of those are showstoppers. You can reverse the call order with another intermediate struct. If you want to set a variable in one side and use it in the other, you can just capture it in both closures (like `num` in the prototype).

Mingun:
> That is a great point! It can be resolved by providing two `show` methods to split the lifetimes:

I like that suggestion! Also, that way you can handle scrollables in a graceful way:
```rust
impl Splitter {
  // Only content, without wrapping into a scrollable area
  fn show_first<R>(...) -> R { ... }
  fn show_second<R>(...) -> R { ... }

  // Wraps content into a scrollable area
  fn show_scrollable_first<R>(...) -> R { ... }
  fn show_scrollable_second<R>(...) -> R { ... }
}
```

Also, I think it is better to use neutral words `first`/`second` instead of `left`/`right` because they cover also `top`/`bottom` variant.

LordMZTE:
Closing stale PR.

PR #1031, opened 2022-01-02
eframe: Add `clipboard` and `links` features

tomassedovic:
Rather than enabling them in the backends for all users, introduce
`clipboard` and `links` as eframe's top-level features and make them
default.

That way, people who don't want to use them (e.g. because of
clipboard's extra system dependencies) can disable default features
and bring in only what they need.

### DIFF
--- a/eframe/Cargo.toml
+++ b/eframe/Cargo.toml
@@ -30,8 +30,8 @@ epi = { version = "0.16.0", path = "../epi" }
 # native:
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 egui-winit = { version = "0.16.0", path = "../egui-winit", default-features = false }
-egui_glium = { version = "0.16.0", path = "../egui_glium", default-features = false, features = ["clipboard", "epi", "links"], optional = true }
-egui_glow = { version = "0.16.0", path = "../egui_glow", default-features = false, features = ["clipboard", "epi", "links", "winit"], optional = true }
+egui_glium = { version = "0.16.0", path = "../egui_glium", default-features = false, features = ["epi"], optional = true }
+egui_glow = { version = "0.16.0", path = "../egui_glow", default-features = false, features = ["epi", "winit"], optional = true }
 
 # web:
 [target.'cfg(target_arch = "wasm32")'.dependencies]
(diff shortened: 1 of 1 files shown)


tomassedovic:
Hi! Not sure if this is something you want to bring in, but an app I'm writing doesn't need the clipboard functionality and the default build fails because it's missing some of the system dependencies that feature needs.

This turns `clipboard` and `links` into features of `eframe`. They're on by default (so there shouldn't be any change in behaviour), but one can now disable the default features of `eframe` in their own `Cargo.toml` and enable only those they need.

As far as I can tell, that's not possible for dependencies of dependencies (e.g. I can't disable `clipboard` in `egui_glow` via my own `Cargo.toml`), hence this PR.

eframe/Cargo.toml
```diff
@@ -42,12 +42,16 @@ image = { version = "0.23", default-features = false, features = ["png"] }
...
 # If you plan on specifying your own fonts you may disable this feature.
 default_fonts = ["egui/default_fonts"]
 
+clipboard = ["egui_glium/clipboard", "egui_glow/clipboard"]
+
+links = ["egui_glium/links", "egui_glow/links"]
+
 # Enable saving app state to disk.
 persistence = [
   # we cannot touch egui_glium or egui_glow here due to https://github.com/rust-lang/cargo/issues/8832
```
emilk, line 57: If you follow the above link you'll see why this PR causes a problem. Anyone using `eframe` now would need to compile both `egui_glium` AND `egui_glow`, which is not so good.

I'm not sure what a good solution is to this. I find feature and the cargo dependency system quite confusing.

emilk (changes requested):

eframe/Cargo.toml
tomassedovic: Ooof, I didn't know about that, thanks for the link. At least it looks like this will be fixed at some point in the future.

The only thing I could think of right now is adding a feature for each backend (e.g.: `egui_glow_clipboard` and `egui_glium_clipboard`, etc.), but I'm not happy with that.

It would solve the issue I'm having and I'll be happy to update the PR with that, but it makes for an unpleasant "cargo API". I totally understand if you don't want to go that route.

emilk: I'm hoping that `egui_glow` can take over as the default backend, and that we can remove `egui_glow` support in the future, so in a few months we will hopefully only have `egui_glow` in this file, and the problem will be solved that way.

emilk: In the meantime you can bypass `eframe` and just use `egui_glow` and `egui_web` directly. `eframe` is an extremely small crate – this is all the code in it: https://github.com/emilk/egui/blob/master/eframe/src/lib.rs

tomassedovic: Great suggestion, thanks!

Issue #1005, opened 2021-12-27
CtxRef::clone is a footgun


danielkeller:
The docs for `CtxRef` prominently mention that it's an `Arc` and that it's cloneable. But starting a new frame mutates, not the context, but the `CtxRef`, invalidating prior clones. I'm sure there's a good reason for this, but in most libraries that give cloneable "handle" types (eg winit, wgpu, etc), mutating methods only affect the thing the handle points to, not the handle itself. I think the API and docs shouldn't suggest that `CtxRef` and `Context` are handles, or at least mention that they're handles only for the duration of one frame.

emilk:
You are right in pointing this out - I thought I had documented it, but looking at it now it seems I have forgotten it.

The reason for it is performance and ergonomics. Everything in `Context` that changes needs to be protected by a mutex. Some things (input and fonts) doesn't change over a frame, so by using a "generational" `Context` we can avoid having a mutex on them. A mutex has a performance cost, but also an ergonomic cost, as one needs to take care of the lifetime of the lock and avoid double-locks.

There is perhaps some other design that can get the best of two worlds.

danielkeller:
I was actually thinking about the use of mutexes as well. Specifically, without multi_threaded set, the library is thread-safe in that it doesn't have UB if the `Context` is shared, but it's not semantically correct to share it; any concurrent use causes the `AtomicRefCell` to panic. I think that if the API makes sense with interior mutability, it would be simpler to embrace the interior mutability: Put everything in one big `RefCell`, borrow mutably in the implementation, and be `!Sync`. If the user wants to be thread safe, they just have to wrap it in their own mutex. Alternatively, if a mutable API makes sense, users can "bring their own" interior mutability with RefCell or mutex.
If I have some time, I'll pull the repo and let you know if I have any ideas.

danielkeller:
I tried it out and came up with [this](https://github.com/danielkeller/egui/commit/875efe1ba1c0ff7a2ea63f5a552ea7f7fecf0f08), which makes `Context` a plain struct and moves the interior mutability into `CtxRef`. The result is not terribly un-ergonomic I think. Hopefully, these ideas are helpful in some way.

emilk:
Nice! How does `cargo bench` compare before/after?

danielkeller:
I added another [commit](https://github.com/emilk/egui/compare/master...danielkeller:ctx-handle) that properly integrates it with the thread safety feature flags. The cargo bench result is:
```
[...]

PR #1035, opened 2022-01-03
Put everything in Context behind the same Mutex

emilk:
Closes https://github.com/emilk/egui/issues/1005


emilk:
Closed in favor of https://github.com/emilk/egui/pull/1050

Issue #814, opened 2021-10-18
Graceful shutdown of native eframe


emilk:
Sparked by https://github.com/emilk/egui/discussions/813

There are two user stories we want to support:

When a user closes the native eframe window (e.g. by clicking the X or pressing Ctrl-F4/Cmd-Q)…

…the egui app should continue and show a "Do you want to save your work" popup (using egui).

…the native window should close immediately, but the app should be able to take its time to save data to disk, etc.

For the first case `eframe` would need to communicate the closing as an event to the App rather than shutting down directly, and then have the app tell `eframe` when it is actually time to quit.

For the second case: currently if `App::on_exit` is slow, the native window will linger, frozen and unresponsive (bad). The fix is probably as simple as moving around some code in `egui_glium` and `egui_glow`.

follower:
I was wanting this functionality a while ago (to allow time for network connections to close cleanly while still showing UI progress feedback).

I got as far as determining that AFAICT the reason this wasn't currently possible was due to treating `WindowEvent::CloseRequested` & `WindowEvent::Destroyed` as being identical here:

https://github.com/emilk/egui/blob/19766bfe4cd756e27c07e338075493a2a22f5131/egui-winit/src/lib.rs#L562

I did not get as far as writing this up until now, however. :)

As I understand it, to enable graceful shutdown we'd need to treat [`WindowEvent::CloseRequested`](https://docs.rs/winit/0.25.0/winit/event/enum.WindowEvent.html#variant.CloseRequested) as a separate event type.

jakeisnt:
Is this already implemented? This example: https://github.com/emilk/egui/tree/master/examples/confirm_exit seems to work.

coderedart:
>  This example: https://github.com/emilk/egui/tree/master/examples/confirm_exit seems to work.

the example shows how to close the application using `frame.close()`. developer can close the window anytime using that.

this issue is more about intercepting an external close event (clicking red `X` close button on window's titlebar or system alt+f4 shortcut etc..). then, just forward the close event to the developer. then, dev can decide whether to close the app or show a prompt before calling `frame.close` or  save the state to disk before closing or just ignore it even.

jakeisnt:
It intercepts! The on_close_event function defined when implementing the eframe app is invoked when a user attempts to close the window. In this case, the function flips a boolean that shows the exit dialog to the user. 

When attempting to close the window of that demo on my (Linux/Wayland) machine, instead of the window closing, I see the dialog.

[...]

PR #1038, opened 2022-01-04
Allow aborting an exit event

lampsitter:
This addresses the first point in #814.

I am not sure how to handle the second scenario in the issue as integration is moved into the closure

### DIFF
--- a/egui-winit/src/epi.rs
+++ b/egui-winit/src/epi.rs
@@ -240,7 +240,11 @@ impl EpiIntegration {
         self.app
             .setup(&self.egui_ctx, &self.frame, self.persistence.storage());
         let app_output = self.frame.take_app_output();
-        self.quit |= app_output.quit;
+
+        if app_output.quit {
+            self.quit = self.app.on_exit_event();
+        }
+
         crate::epi::handle_app_output(window, self.egui_ctx.pixels_per_point(), app_output);
     }
 
@@ -260,7 +264,12 @@ impl EpiIntegration {
 
     pub fn on_event(&mut self, event: &winit::event::WindowEvent<'_>) {
         use winit::event::WindowEvent;
-        self.quit |= matches!(event, WindowEvent::CloseRequested | WindowEvent::Destroyed);
+        if *event == WindowEvent::CloseRequested {
+            self.quit = self.app.on_exit_event();
+        } else if *event == WindowEvent::Destroyed {
+            self.quit = true;
+        }
+
         self.egui_winit.on_event(&self.egui_ctx, event);
     }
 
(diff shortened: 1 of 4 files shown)


egui-winit/src/epi.rs
```diff
@@ -305,6 +340,15 @@ impl EpiIntegration {
             .maybe_autosave(&mut *self.app, &self.egui_ctx, window);
     }
 
+    pub fn on_pre_exit(&mut self) -> bool {
```
lampsitter, line 343: Not sure if this needs to request a redraw..

epi/src/lib.rs
```diff
@@ -144,7 +144,14 @@ pub trait App {
     /// where `APPNAME` is what is returned by [`Self::name()`].
     fn save(&mut self, _storage: &mut dyn Storage) {}
 
-    /// Called once on shutdown (before or after [`Self::save`])
+    /// Called before a shutdown that _can_ be aborted. This means that if you are on the web or
+    /// the window is forcefully closed, this method will not be called.
+    fn on_pre_exit(&mut self) -> bool {
+        true
+    }
```
Mingun, line 151: Need to document what return value means and maybe provide an example (from an issue?)

lampsitter: I have added an example and hopefully the documentation is now more clear.

egui_glow/src/epi_backend.rs
```diff
@@ -145,7 +155,12 @@ pub fn run(app: Box<dyn epi::App>, native_options: &epi::NativeOptions) -> ! {
                 }
 
                 integration.on_event(&event);
-                if integration.should_quit() {
+
+                if EpiQuit::Recoverable == integration.should_quit() {
+                    if integration.on_pre_exit() {
+                        *control_flow = glutin::event_loop::ControlFlow::Exit;
+                    }
+                } else if EpiQuit::Unrecoverable == integration.should_quit() {
```
emilk, line 163: Instead of duplicating all this code, I think it should be possible for `EpiIntegration` to handle it and have the `egui_glow` and `egui_glium` code the same as on `main`.

That is: `EpiIntegration::on_event` and `EpiIntegration::update` can just call `app.on_pre_exit` directly. I don't think there is any need for `EpiQuit`.

epi/src/lib.rs
```diff
@@ -144,7 +144,21 @@ pub trait App {
...
+    /// Called before an exit that can be aborted.
+    /// By returning `false` the exit will be aborted. To continue the exit return `true`.
+    ///
+    /// A scenario where this method will be
+    /// run is after pressing the close button on a native window, which allows you to ask the user
+    /// whether they want to do something before exiting. See the example
+    /// `eframe/examples/confirm_exit.rs` for practical usage.
+    ///
+    /// It will _not_ be called on the web or when the window is forcefully closed.
+    fn on_pre_exit(&mut self) -> bool {
```
emilk, line 156: I think `on_exit_event` may be a better name, as in: an exit event arrived, now you can handle it.

emilk (changes requested):
The example and API is very good, but I think the implementation could be made a lot simpler

emilk (approved):

emilk:
thanks!

PR #1039, opened 2022-01-04
Add new function to animate f32 values

BctfN0HUK7Yg:
Add new function animate_value.
It makes the change in value not a step change, but a continuous one.


emilk:
Remember to run `cargo fmt` on the code!

emilk (approved):

PR #1041, opened 2022-01-05
egui_glow: Add function to set the texture filter

justinj:
This commit adds a `set_scale_filter` method to the `glow` painter so that
textures can be set to scale using nearest-neighbour scaling rather than
linear. This is useful for pixel art.

I wasn't entirely sure what kind of API you want for this kind of change so I
went with what seemed least intrusive, I don't mind doing something more
holistic if this isn't what you had in mind.

Linear:
image

Nearest:
image

### DIFF
--- a/egui_glow/CHANGELOG.md
+++ b/egui_glow/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to the `egui_glow` integration will be noted in this file.
 
 
 ## Unreleased
-* Added `set_scale_filter` method to `Painter`.
+* Added `set_texture_filter` method to `Painter`.
 
 ## 0.16.0 - 2021-12-29
 * Made winit/glutin an optional dependency ([#868](https://github.com/emilk/egui/pull/868)).
--- a/egui_glow/CHANGELOG.md
+++ b/egui_glow/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to the `egui_glow` integration will be noted in this file.
 
 
 ## Unreleased
-
+* Added `set_scale_filter` method to `Painter`.
 
 ## 0.16.0 - 2021-12-29
 * Made winit/glutin an optional dependency ([#868](https://github.com/emilk/egui/pull/868)).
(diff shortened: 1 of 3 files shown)


emilk (changes requested):
Looks good to me, but I think `TextureFilter` is a better name!

justinj:
Done, changed to `TextureFilter`.

egui_glow/src/painter.rs
```diff
@@ -35,6 +35,8 @@ pub struct Painter {
     is_embedded: bool,
     vertex_array: crate::misc_util::VAO,
     srgb_support: bool,
+    /// The filter used for subsequent textures.
+    texture_filter: u32,
```
emilk, line 39: wouldn't it be nicer to use `TextureFilter` here and defer the call to `glow_code`? It would keep the code typesafe for longer.

```suggestion
    texture_filter: TextureFilter,
```

justinj: Sure, seems reasonable to me!

egui_glow/CHANGELOG.md
```diff
@@ -3,7 +3,7 @@ All notable changes to the `egui_glow` integration will be noted in this file.
 
 
 ## Unreleased
-
+* Added `set_texture_filter` method to `Painter`.
```
```suggestion
* Added `set_texture_filter` method to `Painter` ((#1041)[https://github.com/emilk/egui/pull/1041]).
```

egui_glow/src/painter.rs
```diff
@@ -387,6 +412,12 @@ impl Painter {
         }
     }
 
+    // Set the filter to be used for any subsequent textures loaded via
+    // `set_texture`.
```
emilk, line 416:
```suggestion
    // [`Self::set_texture`].
```

emilk (changes requested):

egui_glow/src/painter.rs
emilk: Use inter-doc links

emilk (approved):

Issue #1030, opened 2022-01-02
Context menu does not follow light theme


lampsitter:
**Describe the bug**
When using light theme the context menu remains dark.

**To Reproduce**
1. Open the context menu window
2. Change the theme to light
2. Open the context menu

**Expected behavior**
The menu follows the current theme

**Screenshots**
![egui-theme-bug](https://user-images.githubusercontent.com/96946613/147879573-993e0b63-9ee7-4cdd-acce-55a4951bbb8f.png)

**Desktop (please complete the following information):**
 - OS: Linux 

**Additional context**

PR #1043, opened 2022-01-05
Fix context menu styling

lampsitter:
Closes <https://github.com/emilk/egui/issues/1030>.

### DIFF
--- a/egui/src/menu.rs
+++ b/egui/src/menu.rs
@@ -16,9 +16,8 @@
 //! ```
 
 use super::{
-    style::{Spacing, WidgetVisuals},
-    Align, CtxRef, Id, InnerResponse, PointerState, Pos2, Rect, Response, Sense, Style, TextStyle,
-    Ui, Vec2,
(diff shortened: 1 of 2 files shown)


egui/src/menu.rs
```diff
@@ -120,37 +120,44 @@ pub(crate) fn menu_ui<'c, R>(
...
-                    .inner
-            })
-            .inner
+        ui.scope(|ui| {
+            let style = ui.style_mut();
+            style.spacing = Spacing {
+                item_spacing: Vec2::ZERO,
+                button_padding: crate::vec2(2.0, 0.0),
+                ..Default::default()
+            };
```
emilk, line 143: perhaps we should only over-write `item_spacing` and `button_padding` and leave the rest of the fields?

lampsitter: I agree. Though they should probably not be hardcoded at all. At least not the `button_padding`. Would adding a`spacing.menu_button_padding` be appropriate?

PR #1046, opened 2022-01-06
Providing more control over hover label in plots

AlanRace:
I found that the label displayed at the cursor when hovering on a plot is only shown if either `show_x` or `show_y` were set to `true`. I wanted to be able to display the label regardless of whether the line guides were shown or not, and so added in the `show_hover_label` function to control this display.

I also found that `show_x` and `show_y` weren't very descriptive names for what functionality they were controlling (originally I thought it was related to the x-axis), so I have suggested a change in name to `show_hover_line_x`, but feel free to suggest something more fitting. Similarly with `custom_label_func`.

Finally, I modified the definition of `CustomLabelFunc` to enable a custom function to essentially recreate the default (this wasn't previously possible as it didn't have access to `show_x` and `show_y`. 

An example of a custom label
Custom label


egui/src/widgets/plot/mod.rs
```diff
@@ -208,11 +213,43 @@ impl Plot {
...
+    pub fn default_hover_label_func() -> Box<dyn Fn(&HoverConfig, &str, &Value) -> String> {
+        Box::new(|config, name, value| {
+            let mut prefix = String::new();
+
+            if !name.is_empty() {
+                prefix = format!("{}\n", name);
+            }
+
+            let x_decimals = 6 - ((value.x.abs().log10()).ceil().at_least(0.0) as usize).at_most(6);
+            let y_decimals = 6 - ((value.y.abs().log10()).ceil().at_least(0.0) as usize).at_most(6);
```
Bromeon, line 233: This seems to be used in a few places, also in `bar.rs` and `box_elem.rs` -- maybe extract it to a function?
A few similar helper functions are already at the end of `items/mod.rs`.

egui/src/widgets/plot/mod.rs
```diff
@@ -14,12 +14,14 @@ pub use items::{
...
+use self::items::HoverConfig;
+
 mod items;
 mod legend;
 mod transform;
 
-type CustomLabelFunc = dyn Fn(&str, &Value) -> String;
-type CustomLabelFuncRef = Option<Box<CustomLabelFunc>>;
+type CustomLabelFunc = dyn Fn(&HoverConfig, &str, &Value) -> String;
+type CustomLabelFuncRef = Box<CustomLabelFunc>;
```
Bromeon, line 24: Since `CustomLabelFuncRef` seems to be the prevalent type used everywhere, maybe we should think about a shorter/more precise name? `BarChart` and `BoxPlot` have an `element_formatter()` method, which customizes the displayed text when hovering over them, which is essentially the same concept.

Maybe something like `HoverFormatter`?

On the other hand, the type `dyn Fn(&HoverConfig, &str, &Value) -> String` seems to be used only once, so it can either be inlined or use a more verbose name `HoverFormatterFn` or so.

We are going to need a lot of these customization functions, so agreeing on a term like "formatter" could make sense.
Use cases:
* Hovering for bar charts and box plots (currently `element_formatter`, could be renamed to `hover_formatter`)
* Hovering for X/Y values (this PR, maybe `hover_formatter` makes clear it's the same idea)
* Labeling X/Y axis (I'm going to work on this, I'm thinking about `x_axis_formatter` + `y_axis_formatter`)

Bromeon (commented):
Very nice! I also like the internal simplification (better naming + integrating to `PlotConfig`).
I have a few additional ideas to make things more intuitive/consistent 🙂

egui/src/widgets/plot/mod.rs
```diff
@@ -208,11 +213,43 @@ impl Plot {
...
         mut self,
-        custom_lebel_func: F,
+        hover_label_func: F,
     ) -> Self {
-        self.custom_label_func = Some(Box::new(custom_lebel_func));
+        self.hover_label_func = Box::new(hover_label_func);
+        self
+    }
+
+    pub fn default_hover_label_func() -> Box<dyn Fn(&HoverConfig, &str, &Value) -> String> {
```
Bromeon, line 224: Why is this public? Someone who wants the default formatter would simply not call `hover_label_func()`.

emilk:
It would be great if a demo of the new feature was added to `egui_demo_lib/src/apps/demo/plot_demo.rs`, and perhaps a screenshot added to the PR description

egui/src/widgets/plot/mod.rs
AlanRace: That's a good point. Having it public would have made the `plot_demo` code slightly easier to write, but that's probably too much of an edge case.

AlanRace:
I have now updated with the comments above and added a new demo. A screenshot of the demo is also included in the original PR description.

egui/src/widgets/plot/mod.rs
```diff
@@ -108,6 +110,35 @@ impl LinkedAxisGroup {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum HoverLine {
+    None,
+    X,
+    Y,
+    XY,
+}
```
emilk, line 119: Seems like it would make more sense to make this into a

```
pub struct HoverLine {
    /// Highlight values on the X axis by painting a vertical line.
    pub x: bool,
    /// Highlight values on the Y axis by painting a horizontal line.
    pub y: bool
}
```

emilk (changes requested):
Looks great, but I think `HoverLine` makes more sense as a `struct` than an `enum`

Titaniumtown:
This PR looks really cool! Would love to see this feature merged.

emilk:
Coverted to a draft until work resumes!

AlanRace:
Is this still necessary after merging https://github.com/emilk/egui/pull/1235?

alexespencer:
I think it is still needed even though #1235 has merged as this can disable the hover tooltip completely?

emilk:

See also:

PR #1051, opened 2022-01-08
add back repaint_signal in epi::Frame

jnises:
repaint_signal was removed from epi::Frame in version 0.16
I was using that to trigger repaints at a frequency below the screen refresh rate.

Is there a better way to do that these days?

### DIFF
--- a/epi/src/lib.rs
+++ b/epi/src/lib.rs
@@ -319,6 +319,11 @@ impl Frame {
         self.lock().output.drag_window = true;
     }
 
+    /// If you need to request a repaint from another thread, send it to that other thread.
+    pub fn repaint_signal(&self) -> Arc<dyn backend::RepaintSignal> {
+        self.lock().repaint_signal.clone()
+    }
+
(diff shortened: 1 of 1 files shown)


epi/src/lib.rs
```diff
@@ -319,6 +319,11 @@ impl Frame {
         self.lock().output.drag_window = true;
     }
 
+    /// If you need to request a repaint from another thread, send it to that other thread.
+    pub fn repaint_signal(&self) -> Arc<dyn backend::RepaintSignal> {
```
jnises, line 323: a bad idea to expose _backend_ stuff here?
should it be done in some other way?

emilk:
There is a better way to do it - just clone `Frame` and call `Frame::request_repaint`. Did you read the changelog? https://github.com/emilk/egui/blob/master/eframe/CHANGELOG.md#0160---2021-12-29

jnises:
ops. sorry.
i only read the egui changelog.
didn't realize eframe had its own.

PR #1056, opened 2022-01-08
Add some improvements in the documentation

Mingun:


egui/src/ui.rs
```diff
@@ -1055,7 +1055,7 @@ impl Ui {
     /// Add extra space before the next widget.
     ///
     /// The direction is dependent on the layout.
-    /// This will be in addition to the [`Spacing::item_spacing`}.
+    /// This will be in addition to the [`crate::style::Spacing::item_spacing`].
```
emilk, line 1058: Didn't the old link work properly? If so, shouldn't the CI/`check.sh` catch that?

epaint/src/shape.rs
```diff
@@ -15,10 +15,12 @@ pub enum Shape {
     /// For performance reasons it is better to avoid it.
     Vec(Vec<Shape>),
     Circle(CircleShape),
+    /// A line between two points.
     LineSegment {
         points: [Pos2; 2],
         stroke: Stroke,
     },
+    /// A series of lines between points.
```
emilk, line 23:
```suggestion
    /// A series of lines between points.
    /// The path can have a stroke and/or fill (if closed).
```

emilk (approved):
Thanks!

egui/src/ui.rs
Mingun: No, it looks as a monospaced text in the brackets

epaint/src/shape.rs
Mingun: Added

emilk:
There is a weird merge conflict

Mingun:
Rebased

emilk (approved):

PR #1089, opened 2022-01-10
make next_widget_position public

jakobhellermann:
I'm implementing this viewport layout
grafik
In order to provide the 3d camera with the correct viewport I need to know where and how large the inner space is.

```rust
let position = ui.next_widget_position();
let size = ui.available_size();
```
does excactly what I want, sadly the first method is private.

This PR would allow me to use it :)

### DIFF
--- a/egui/src/ui.rs
+++ b/egui/src/ui.rs
@@ -748,7 +748,7 @@ impl Ui {
     }
 
     /// Where do we expect a zero-sized widget to be placed?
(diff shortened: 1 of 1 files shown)


emilk:
`ui.allocate_exact_size(ui.available_size(), Sense::hover())` should also work I think

emilk (approved):

jakobhellermann:
You're right, the method does exactly what I want. I missed it because I didn't notice that it returned a `Rect`. Thanks!

PR #1090, opened 2022-01-10
Fix typo: pupup -> popup

HackerFoo:
Fix typo: pupup -> popup

### DIFF
--- a/egui/src/widgets/color_picker.rs
+++ b/egui/src/widgets/color_picker.rs
@@ -335,19 +335,19 @@ pub fn color_picker_color32(ui: &mut Ui, srgba: &mut Color32, alpha: Alpha) -> b
 }
 
 pub fn color_edit_button_hsva(ui: &mut Ui, hsva: &mut Hsva, alpha: Alpha) -> Response {
-    let pupup_id = ui.auto_id_with("popup");
-    let open = ui.memory().is_popup_open(pupup_id);
+    let popup_id = ui.auto_id_with("popup");
+    let open = ui.memory().is_popup_open(popup_id);
     let mut button_response = color_button(ui, (*hsva).into(), open);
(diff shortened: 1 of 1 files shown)


coderedart:
I like pupup :dog2:

emilk:
Pleas

HackerFoo:
@emilk I suggest not doing this - it gives potential contributors a bad impression. There is also a burden to write small PRs.

I start with small PRs to decide whether it's worth putting effort into merging larger PRs, generally sorting my outstanding changes in increasing size and complexity.

I imagine others do the same to avoid wasting time, or at least to estimate how easy it is to get fixes upstreamed. Typo PRs are actually a good way to welcome new contributors.

For what it's worth, I wasted a few seconds wondering whether `pupup` was a typo, or if it meant something, did a find/replace, and checked that the code still worked. I wasn't looking for typos, I just encountered it and figured I'd save some others a bit of confusion. I made the change while making another functional change (#1091), but I think it is bad practice to bundle unrelated changes together.

emilk:
@HackerFoo You make a good point, and I apologize. I agree that it is good to fix spelling and grammar (code readability is super important), and keeping changes separate is also a good idea. Though I would prefer for the change to be a separate commit in a bigger PR rather than a separate PR. Of course, this is my personal preference (and one that I have gone back and forth on over the years). I also haven't clearly stated my preference in the contributor guidelines, so that's on me.

There is also the context of people making commits with a single minor spelling fix in some unimportant file just to be able to be credited as contributors to a project, a behavior that just adds noise to the open source world.

Anyway, I changed the pupup -> poup in https://github.com/emilk/egui/commit/199bbef77bf59f9f5fa48346c715d827559e00cb#diff-eb4801118508476c267604aae155912ad2f8cae54490b4bc6d6ed9d0a3a38730

PR #1091, opened 2022-01-10
Make `color_picker_hsva_2d` public.

HackerFoo:
Exposes `color_picker_hsva_2d`, so the color picker can be shown without an egui button. A similar public function exists, but it converts the color to `Color32`.

I use this to show the color picker on top of a custom UI.

Screen Shot 2022-01-10 at 11 24 20

### DIFF
--- a/egui/src/widgets/color_picker.rs
+++ b/egui/src/widgets/color_picker.rs
@@ -308,8 +308,10 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsva: &mut HsvaGamma, alpha: Alpha) {
     color_slider_2d(ui, v, s, |v, s| HsvaGamma { s, v, ..opaque }.into());
 }
 
+//// Shows a color picker where the user can change the given Hsva color.
+///
 /// Returns `true` on change.
-fn color_picker_hsva_2d(ui: &mut Ui, hsva: &mut Hsva, alpha: Alpha) -> bool {
+pub fn color_picker_hsva_2d(ui: &mut Ui, hsva: &mut Hsva, alpha: Alpha) -> bool {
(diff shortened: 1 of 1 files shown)


BctfN0HUK7Yg:
This color picker isn't universal. Better way is to make you own widget for this by copying and tuning original code

HackerFoo:
The color picker is useful to me without any further tuning, and the `Color32` version is already public. I could use that if necessary, but I don't need the conversion to `Color32`, and this function was already written.

egui/src/widgets/color_picker.rs
```diff
@@ -308,8 +308,10 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsva: &mut HsvaGamma, alpha: Alpha) {
     color_slider_2d(ui, v, s, |v, s| HsvaGamma { s, v, ..opaque }.into());
 }
 
+//// Shows a color picker where the user can change the given Hsva color.
```
emilk, line 311:
```suggestion
//// Shows a color picker where the user can change the given [`Hsva`] color.
```

egui/src/widgets/color_picker.rs
```diff
@@ -323,7 +325,7 @@ fn color_picker_hsva_2d(ui: &mut Ui, hsva: &mut Hsva, alpha: Alpha) -> bool {
     }
 }
 
-/// Shows a color picker where the user can change the given color.
+/// Shows a color picker where the user can change the given Color32 color.
```
emilk, line 328:
```suggestion
/// Shows a color picker where the user can change the given [`Color32`] color.
```

emilk (approved):

PR #1093, opened 2022-01-11
egui_glow: Optimize Painter::set_texture

dvec:
`Painter::set_texture` for some reason was allocating new Vec for every single pixel in image. For 1280x720 image, there was about a million `malloc` calls. By optimizing this code to allocate Vec only once, I got this function running 20x faster on my machine.

### DIFF
--- a/egui_glow/src/painter.rs
+++ b/egui_glow/src/painter.rs
@@ -432,19 +432,17 @@ impl Painter {
             "Mismatch between texture size and texel count"
         );
 
-        // TODO: optimize
-        let pixels: Vec<u8> = image
-            .pixels
-            .iter()
-            .flat_map(|srgba| Vec::from(srgba.to_array()))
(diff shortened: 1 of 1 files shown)


emilk (approved):
whoa, good catch. I missed that `Vec::from` on my previous review 😓

egui_glow/src/painter.rs
```diff
@@ -432,19 +432,17 @@ impl Painter {
...
+            data.extend_from_slice(&srgba.to_array());
+        }
 
         let gl_texture = srgb_texture2d(
             gl,
             self.is_webgl_1,
             self.srgb_support,
             self.texture_filter,
-            &pixels,
+            &data,
```
emilk, line 445: a further optimization is to do a cast of `image.pixels` here instead of allocating at all (maybe using `bytemuck`?)

Issue #943, opened 2021-12-09
Unexpected behaviour of `drag_window`


feelingnothing:
![result](https://user-images.githubusercontent.com/47575622/145481036-e49d40a3-f52b-4bf5-a638-01a0d65d693e.gif)

feelingnothing:
![result](https://user-images.githubusercontent.com/47575622/145485061-e3b75eb7-4aea-4a53-a5d2-868d82b4eb4d.gif)
After some test i figured that on windows it requires being called only on start, but losing focus after possible only with cursor going out of application

emilk:
Perhaps `winit::Window::drag_window` should only be called once (even if an egui app sets the `drag_window` flag many frames). Since everything works well on my mac it is hard for me to figure out. Perhaps you can take a look at `egui-winit/src/epi.rs` and see if you can fix it? `egui-winit` would need to save some state (`is_dragging_window`) and then only call `drag_window` if `drag_window && !is_dragging_window`.

AlexxxRu:
I've checked this bug and looks like we should call `drag_window()` only after the mouse left click.
It works fine, but I'm not sure how it can work with touch screens. Do we have the same event from winit?

@emik what do you think about it? 
https://github.com/emilk/egui/pull/1108

AlexxxRu:
But this solution is still far from the best. The application should call `drag_window` just after LMB [drag_window](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.drag_window).
For now, `drag_window` is called only during the redraw event. And I'm not sure how we can get frame output just after LMB.

Issue #846, opened 2021-10-26
`drag_window` scuffed even on Windows


CAD97:
**Describe the bug**

https://user-images.githubusercontent.com/5992217/138809615-738f3334-7306-4fab-81c7-1bbabcdddedd.mp4

**Expected behavior**

Either:

- A clean drag, or
- The documentation to note that it's awkward on Windows as well as macOS and impossible on the web

**Desktop (please complete the following information):**
 - OS: Microsoft Windows (10 Education) Version 21H1 (OS Build 19043.1288)

**Additional context**

My original code

```rust
impl epi::App for _ {
    fn update(&mut self, ctx: &egui::CtxRef, frame: &mut epi::Frame<'_>) {
        let InnerResponse { response, .. } =
            egui::TopBottomPanel::top("top_toolbar").show(ctx, |ui| {
                egui::warn_if_debug_build(ui);
            });

        if response.hovered() {
            frame.drag_window();
        }

        egui::CentralPanel::default().show(ctx, |ui| {
            ui.label("Hello, world");
        });
    }
```

was even more scuffed, never leaving the jitter phase. Funnily enough, it's jittery enough that it triggers Windows's "wiggle a window around to minimize all other windows" feature, even if I'm moving my cursor smoothly.

AlexxxRu:
@CAD97 Hi, I have the same trouble on my Windows. 
I've created PR with the fix. Can you check it for your case?

More info here:
https://github.com/emilk/egui/issues/943

PR #1108, opened 2022-01-14
Bugfix/drag window

AlexxxRu:
Fix `drag_window ` behavior on Windows.
I've tested my changes only on Windows 10 and Linux Mint OS. 

Closes https://github.com/emilk/egui/issues/846
Closes https://github.com/emilk/egui/issues/943

### DIFF
--- a/egui-winit/src/epi.rs
+++ b/egui-winit/src/epi.rs
@@ -196,6 +196,7 @@ pub struct EpiIntegration {
     pub app: Box<dyn epi::App>,
     /// When set, it is time to quit
     quit: bool,
+    can_drag_window: bool,
 }
 
 impl EpiIntegration {
@@ -229,6 +230,7 @@ impl EpiIntegration {
             egui_winit: crate::State::new(window),
             app,
             quit: false,
+            can_drag_window: false,
         };
 
         slf.setup(window);
@@ -264,8 +266,18 @@ impl EpiIntegration {
     }
 
     pub fn on_event(&mut self, event: &winit::event::WindowEvent<'_>) {
-        use winit::event::WindowEvent;
+        use winit::event::{ElementState, MouseButton, WindowEvent};
         self.quit |= matches!(event, WindowEvent::CloseRequested | WindowEvent::Destroyed);
+
+        if let WindowEvent::MouseInput {
+            button: MouseButton::Left,
+            state: ElementState::Pressed,
+            ..
+        } = event
+        {
+            self.can_drag_window = true;
+        }
+
         self.egui_winit.on_event(&self.egui_ctx, event);
     }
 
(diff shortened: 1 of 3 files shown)


egui-winit/src/epi.rs
```diff
@@ -289,7 +301,12 @@ impl EpiIntegration {
         self.egui_winit
             .handle_output(window, &self.egui_ctx, egui_output);
 
-        let app_output = self.frame.take_app_output();
+        let mut app_output = self.frame.take_app_output();
+        if app_output.drag_window {
+            app_output.drag_window &= self.can_drag_window;
+            self.can_drag_window = false;
```
emilk, line 307: shouldn't this be reset each frame? Otherwise a single click will set `can_drag_window=true` and keep it set to `true` until `drag_window` is set

emilk (changes requested):

egui-winit/src/epi.rs
AlexxxRu: Yes, you're right. I've tested with a reset on every frame, and it works fine.

emilk (approved):
Please add a line to `eframe/CHANGELOG.md` and `egui-winit/CHANGELOG.md`. Then LGTM!

egui-winit/src/epi.rs
```diff
@@ -264,11 +266,17 @@ impl EpiIntegration {
...
+        use winit::event::{ElementState, MouseButton, WindowEvent};
+
+        match event {
+            WindowEvent::CloseRequested => self.quit = self.app.on_exit_event(),
+            WindowEvent::Destroyed => self.quit = true,
+            WindowEvent::MouseInput {
+                button: MouseButton::Left,
+                state: ElementState::Pressed,
+                ..
+            } => self.can_drag_window = true,
```
emilk, line 278: maybe `mouse_pressed_this_frame` is a better name for this member

AlexxxRu: I was thinking about it, but this name raises other questions:
- What button is pressed (Left, Right, or Middle)? 
- Why do we need to save this in the app context? 

Maybe something like `lmb_pressed_this_frame`or `mouse_left_pressed_this_frame` would be better? At least it will reflect the concrete button. 
But I think it's better to keep this name related to the logic. 

If you want, I can replace it with your proposal.

emilk: This is all very nit-picky :) but:

In general, I find it's better to name things after what they signify rather than what they are used for (especially since some things are used for many things).

The line `app_output.drag_window &= self.can_drag_window;` doesn't say anything useful ("only allow drags if we allow drags"), while `app_output.drag_window &= self.lmb_pressed_this_frame;` explains exactly what is happening ("only allow drags if lmb was pressed this frame"). Anyways, it doesn't matter much, I just wanted to explain how my brain works :P

egui-winit/src/epi.rs
```diff
@@ -291,7 +299,9 @@ impl EpiIntegration {
             .egui_winit
             .handle_output(window, &self.egui_ctx, egui_output);
 
-        let app_output = self.frame.take_app_output();
+        let mut app_output = self.frame.take_app_output();
+        app_output.drag_window &= self.can_drag_window;
```
emilk, line 303:
```suggestion
        app_output.drag_window &= self.can_drag_window; /// Required on windows, see https://github.com/emilk/egui/pull/1108
```

emilk (approved):

egui-winit/src/epi.rs
AlexxxRu: Yeah, I got your point of view :)
As you said, it doesn't matter too much, so feel free to change it in any way. Or if you want I can't make these changes on my side.

emilk:
Thank you!

Issue #1047, opened 2022-01-06
Checkbox without label


voelklmichael:
I would like to have a checkbox without label, since a label like "" also reserves a lot of (empty) space.
This reservation looks bad in my Layout.

Possible solutions:
- change checkbox constructor to use Option<str>
- add new constructors
- add new widget

Is this a good "first contributor" issue? If so, I can try to implement this.

emilk:
I think the simplest approach is just to treat the `text.is_empty()` case differently in `Checkbox::ui`, so that `ui.checkbox(&mut value, "")` takes up minimal space.

Feel free to implement this!

Make sure you also do it for `RadioButon` (the code for `Checkbox` and `RadioButton` are extremely similar)

PR #1113, opened 2022-01-15
checkbox without a text label

voelklmichael:
This fixes the empty-space layout for a checkbox with empty label.
Note that the radio-button is NOT done, since I'm unsure about line 250:
   What is the meaning of 'spacing.interact_size'?
If my adjustment is ok, I will be happy to do the same for radiobutton.

Closes <https://github.com/emilk/egui/issues/1047>.

### DIFF
--- a/egui/src/widgets/button.rs
+++ b/egui/src/widgets/button.rs
@@ -239,7 +239,10 @@ impl<'a> Widget for Checkbox<'a> {
         let icon_width = spacing.icon_width;
         let icon_spacing = ui.spacing().icon_spacing;
         let button_padding = spacing.button_padding;
-        let total_extra = button_padding + vec2(icon_width + icon_spacing, 0.0) + button_padding;
+        let mut total_extra = button_padding;
+        if !text.is_empty() {
+            total_extra += vec2(icon_width + icon_spacing, 0.0) + button_padding;
+        }
 
         let wrap_width = ui.available_width() - total_extra.x;
 
@@ -247,11 +250,11 @@ impl<'a> Widget for Checkbox<'a> {
         let text = if !text.text().is_empty() {
             let text = text.into_galley(ui, None, wrap_width, TextStyle::Button);
             desired_size += text.size();
-            desired_size = desired_size.at_least(spacing.interact_size); //Here I'm unsure - what is interacting size?
             Some(text)
         } else {
             None
         };
+        desired_size = desired_size.at_least(Vec2::splat(spacing.interact_size.y));
         desired_size.y = desired_size.y.max(icon_width);
         let (rect, mut response) = ui.allocate_exact_size(desired_size, Sense::click());
 
--- a/egui/src/widgets/button.rs
+++ b/egui/src/widgets/button.rs
@@ -239,21 +239,18 @@ impl<'a> Widget for Checkbox<'a> {
         let icon_width = spacing.icon_width;
         let icon_spacing = ui.spacing().icon_spacing;
         let button_padding = spacing.button_padding;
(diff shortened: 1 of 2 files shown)


egui/src/widgets/button.rs
```diff
@@ -242,26 +242,34 @@ impl<'a> Widget for Checkbox<'a> {
...
         let wrap_width = ui.available_width() - total_extra.x;
-        let text = text.into_galley(ui, None, wrap_width, TextStyle::Button);
 
-        let mut desired_size = total_extra + text.size();
-        desired_size = desired_size.at_least(spacing.interact_size);
+        let mut desired_size = total_extra;
+        let text = if !text.text().is_empty() {
+            let text = text.into_galley(ui, None, wrap_width, TextStyle::Button);
+            desired_size += text.size();
+            desired_size = desired_size.at_least(spacing.interact_size); //Here I'm unsure - what is interacting size?
```
emilk, line 250: From [the docs](https://docs.rs/egui/latest/egui/style/struct.Spacing.html#structfield.interact_size):

> Minimum size of a DragValue, color picker button, and other small widgets. interact_size.y is the default height of button, slider, etc. Anything clickable should be (at least) this size.

The original idea was that it is bad to have very small buttons, because they are difficult to click (especially on touch screens), but in this case you presumably want a small button.

`desired_size = desired_size.at_least(Vec2::splat(spacing.interact_size.y));` could be a good workaround, and should be applied whether or not there is text.

egui/src/widgets/button.rs
```diff
@@ -242,26 +242,34 @@ impl<'a> Widget for Checkbox<'a> {
         let total_extra = button_padding + vec2(icon_width + icon_spacing, 0.0) + button_padding;
```
emilk, line 242: `icon_spacing` should only be applied if there is text

emilk (changes requested):

egui/src/widgets/button.rs
voelklmichael: I tried to implement this, but I'm unsure about the second button padding.
As far as I can tell, this is fine.

egui/src/widgets/button.rs
voelklmichael: I've removed the original 'at_least' instruction, and replaced it with your suggestion. Is this what you had in your mind?
Or an if/else - if text is empty, then your suggestion, else original code?

voelklmichael (commented):
I've implemented your feedback, but have some more questions.
Questions are responses to your feedback.

HactarCE:
Any update on this?

egui/src/widgets/button.rs
```diff
@@ -239,29 +239,40 @@ impl<'a> Widget for Checkbox<'a> {
         let icon_width = spacing.icon_width;
         let icon_spacing = ui.spacing().icon_spacing;
         let button_padding = spacing.button_padding;
-        let total_extra = button_padding + vec2(icon_width + icon_spacing, 0.0) + button_padding;
+        let mut total_extra = button_padding;
+        if !text.is_empty() {
+            total_extra += vec2(icon_width + icon_spacing, 0.0) + button_padding;
+        }
 
         let wrap_width = ui.available_width() - total_extra.x;
```
emilk, line 247: You only need to calculate `wrap_width` if there is text to wrap, so you only need to calculate `total_extra` if there is text.

voelklmichael: Hi, thank you for the feedback - I still would like to see this merged.
So, I have tried to implement it.
If this is OK, I would copy it to radio-button.

PR #1119, opened 2022-01-16
Fix hard-coded corner radii

juancampa:
# Problem

Two widgets had hard-coded corner radius instead of using the WidgetVisuals value.

# Solution

Remove the hard-coded values and use the style instead.

# Context

I tried to make a style with non rounded corners and noticed that certain widgets kept their roundness. 

It turned out that there were two widgets with hard-coded values. This PR uses the value from the `WidgetVisual` which seems like the right thing to do. Maybe there's a reason not to but it seems to fix my problem with no apparent downsides.

# Check.sh Issues

Note that I tried running `check.sh` but it fails to compile 'speech-dispatcher'. Most likely because the version on libspeechd I installed on my machine (latest on Arch linux) is too new and doesn't work with the existing crate.

### DIFF
--- a/egui/src/containers/collapsing_header.rs
+++ b/egui/src/containers/collapsing_header.rs
@@ -331,9 +331,8 @@ impl CollapsingHeader {
             {
                 let rect = rect.expand(visuals.expansion);
 
-                let corner_radius = 2.0;
                 ui.painter()
-                    .rect(rect, corner_radius, visuals.bg_fill, visuals.bg_stroke);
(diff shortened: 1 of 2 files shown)


emilk:
Look good, but needs `cargo fmt`.

Please file an issue about `check.sh` and what line in it fails, and with what message!

juancampa:
> Look good, but needs `cargo fmt`.
> 
> Please file an issue about `check.sh` and what line in it fails, and with what message!

Done and done

PR #1126, opened 2022-01-17
Re-enable window transparency

parasyte:
- `winit` 0.26.1 fixes the crash on Windows 11
- Cargo.lock was updated with `cargo update`

`winit` version was checked with:

```
$ cargo tree | rg ' winit '
│   └── winit v0.26.1
│   │   └── winit v0.26.1 (*)
```


parasyte:
Ok, that should do it!

emilk (approved):

Issue #1048, opened 2022-01-06
Anchored window tries to fit inside native window


lampsitter:
**Describe the bug**
According to the `Window::anchor` method a window that is anchored is supposed to be immovable.  However when resizing the native window to smaller than the anchored window, it still tries to move the window up in such a manner as witnessed for the memory ui window seen in the screenshot below.
The rendered window is correct, but the red area (interactive area?) is moved upwards. In this state it is impossible to click the `file` menu button.

**To Reproduce**
```rust
egui::TopBottomPanel::top("menu_panel").show(ctx, |ui| {
    ui.menu_button("file", |_| {});
});

egui::Window::new("_").show(ctx, |ui| {
    ctx.memory_ui(ui);
});

egui::Window::new("tool_panel")
    .resizable(false)
    .anchor(egui::Align2::LEFT_TOP, egui::vec2(0.0, 0.0))
    .show(ctx, |ui| {
        ui.scope(|ui| {
            for _ in 0..18 {
                ui.label("Foo");
            }
        });
    });
```

Resize the native window upwards until the window can no longer fit inside the native window.

**Expected behavior**
The anchored window stays in place.

**Screenshots**
![Screenshot_20220106_225247](https://user-images.githubusercontent.com/96946613/148457814-86697382-ffa8-49b4-8338-57eb42f94e3a.png)

**Desktop (please complete the following information):**
 - OS: Linux

**Additional context**

PR #1127, opened 2022-01-17
Don't move fixed windows

lampsitter:
This is a more restricted version of https://github.com/emilk/egui/pull/1049

It only applies to windows that have `fixed_pos` set. The issue still remains for the anchored example provided in https://github.com/emilk/egui/issues/1048 sadly. But doing the same for anchored windows would not be correct as seen with `Align2::CENTER_CENTER`, where moving the window up is correct behaviour.

### DIFF
--- a/egui/src/containers/window.rs
+++ b/egui/src/containers/window.rs
@@ -29,6 +29,7 @@ pub struct Window<'open> {
     resize: Resize,
     scroll: ScrollArea,
     collapsible: bool,
+    fixed_pos: bool,
     with_title_bar: bool,
 }
 
@@ -50,6 +51,7 @@ impl<'open> Window<'open> {
                 .default_size([340.0, 420.0]), // Default inner size of a window
             scroll: ScrollArea::neither(),
             collapsible: true,
+            fixed_pos: false,
             with_title_bar: true,
         }
     }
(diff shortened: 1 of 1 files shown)


egui/src/containers/window.rs
```diff
@@ -117,6 +119,7 @@ impl<'open> Window<'open> {
     /// Set initial position of the window.
     pub fn default_pos(mut self, default_pos: impl Into<Pos2>) -> Self {
         self.area = self.area.default_pos(default_pos);
+        self.fixed_pos = false;
```
emilk, line 122: A window with a default position is not fixed!

egui/src/containers/window.rs
```diff
@@ -29,6 +29,7 @@ pub struct Window<'open> {
     resize: Resize,
     scroll: ScrollArea,
     collapsible: bool,
+    fixed_pos: bool,
```
emilk, line 32: Seems like this should be moved into `Area` instead

egui/src/containers/window.rs
```diff
@@ -399,9 +405,11 @@ impl<'open> Window<'open> {
             content_inner
         };
 
-        area.state_mut().pos = ctx
-            .constrain_window_rect_to_area(area.state().rect(), area.drag_bounds())
-            .min;
+        if !fixed_pos {
+            area.state_mut().pos = ctx
+                .constrain_window_rect_to_area(area.state().rect(), area.drag_bounds())
+                .min;
```
emilk, line 411: Perhaps we should let `area.end` do this instead by just re-setting `area.moveable` to the original value before calling `area.end` ? (area.end already constrains the area iff it is moveable)

emilk (changes requested):

PR #1130, opened 2022-01-17
Formatter for plot axis labels

### LINKED ISSUE #725: Configurable Plot Axis Notation
Hello,

im  currently using egui to write a visualisation for complex fourier transformations. I really love working with it, especially with the excellent plotting module. The only option that I'm missing so far is a way to modify the numbers along along the axes.

E. g. since I'm worling with complex numbers, I would like to display 1i instead of just 1. I'm also writing a module for svg parsing and plotting. Since svg paths are defined in matrix notation, this could also also be useful for inverting the y-axis.
I believe this would also be really nice for interactive web plotting Physics/Chemistry experiments where a unit on the axis is good practice. 

Keep up the great work,
David

Bromeon:
Adds two methods `Plot::x_axis_formatter` and `Plot::y_axis_formatter()`.
Also fixes the bug where `Orientation` was used in public fields, but wasn't public itself.

With them, it's possible to customize the displayed labels along the axes.
```rs
Plot::new("Complex plane")
    .y_axis_formatter(|val| format!("{}i"))
    .show(ui, |plot_ui| plot_ui.bar_chart(chart))
```
It's also possible to return the empty string, in order to not display a label. This can be useful for discrete X/Y domains, when there isn't a meaningful label beyond certain resolution or zoom level. An example can be dates, with only full days.

Here's a screenshot of the stacked bar-chart, with X axis labeled as days and Y axis as percentage.
This example is now implemented in the plot demo.
grafik

Closes #725.


Bromeon:
One open question is positioning of the labels.

In the screenshot above I had to "hide" the Y=0 value `0%`, because it would overlap with the X=0 label `Day 0`. I did this by returning an empty string in the formatter.

Problems like these could be solved by specifying the alignment (left/right/center) and orientation (horizontal/vertical) of the text along the axis. But I think it's an orthogonal problem to text formatting (no pun intended), so it should be possible to do it in a separate PR.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui_w
 ### Added ⭐
 * `Context::load_texture` to convert an image into a texture which can be displayed using e.g. `ui.image(texture, size)` ([#1110](https://github.com/emilk/egui/pull/1110)).
 * Added `Ui::add_visible` and `Ui::add_visible_ui`.
+* Added `Plot::x_axis_formatter` and `Plot::y_axis_formatter()` for custom axis labels ([#1130](https://github.com/emilk/egui/pull/1130)
```
Bromeon, line 13: Just noticed missing closing brace, will fix.

Bromeon: Fixed

Bromeon:
Fixed merge conflicts.

CHANGELOG.md
```diff
@@ -35,6 +36,7 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui_w
 
 ### Fixed 🐛
 * Context menu now respects the theme ([#1043](https://github.com/emilk/egui/pull/1043))
+* Plot `Orientation` was not public, although fields using this type were ([#1130](https://github.com/emilk/egui/pull/1130))
```
emilk, line 39: I wonder why the compiler doesn't stop us from doing that O.o

egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -74,7 +74,7 @@ impl LineDemo {
                 ui.checkbox(animate, "animate");
                 ui.checkbox(square, "square view")
```
emilk, line 75:
```suggestion
                ui.checkbox(animate, "Animate");
                ui.checkbox(square, "Square view")
```

let's keep it consistent

egui/src/widgets/plot/mod.rs
```diff
@@ -216,6 +221,30 @@ impl Plot {
...
     }
 
+    /// Provide a function to customize the labels for the X axis.
+    ///
+    /// This is useful for custom input domains, e.g. date/time.
+    ///
+    /// If axis labels should not appear for certain values or beyond a certain zoom/resolution,
+    /// the formatter function can return empty strings. This is also useful if your domain is
+    /// discrete (e.g. only full days in a calendar).
+    pub fn x_axis_formatter<F: 'static + Fn(f64) -> String>(mut self, func: F) -> Self {
```
emilk, line 231: nit: I prefer this (and it is more consistent with the rest of egui)

```suggestion
    pub fn x_axis_formatter(mut self, func: impl Fn(f64) -> String + 'static) -> Self {
```

CHANGELOG.md
Bromeon: There's a basic check, but it breaks down as soon as you re-export symbols.

So, `pub` doesn't mean "visible outside the crate", but rather "visible outside the crate, if the parent modules allow it" 🙂

emilk (changes requested):
Looks good! Just left a couple of nits

Bromeon:
Applied the suggestions.

---

Sorry for the many CI runs, wasn't aware that previous runs would complete. In case you're interested, it's possible to abort previous runs [as done here](https://github.com/godot-rust/godot-rust/blob/42a6e768a02ed32068ae1dfe5a5e655181ea2e6f/.github/workflows/minimal-ci.yml#L27-L30):
```yaml
# If a new commit is pushed before the old one's CI has completed (on the same branch), abort previous run
concurrency:
  group: ${{ github.head_ref }}
  cancel-in-progress: true
```
This might only work for the `pull_request` and not the `push` trigger however (not sure if you use CI for the latter, too, or if you do all changes via PRs).

emilk (approved):

Issue #1143, opened 2022-01-21
The links for egui_web, egui-winit, egui_glium, and egui_glow changelog inside eframe CHANGELOG.md are broken


LaBatata101:
Here's the [link](https://github.com/emilk/egui/blob/master/eframe/CHANGELOG.md) for the file.

PR #1146, opened 2022-01-21
Fix eframe changelog links to other changelogs

bthehacker:
Links to the changelogs of egui_web, egui-winit, and egui_glium were broken and led to a 404, fixed them by adding "../"

Closes <https://github.com/emilk/egui/issues/1143>.


emilk:
thanks!

PR #1147, opened 2022-01-21
Add ability to override collapsing icons

juancampa:
I wanted custom icons for my tree so I added the ability to override's `CollapsingHeader`'s icon by passing a function that gets called instead of `paint_icon`.

I personally like the pointy triangles but not the fact that they get thicker on hover.

Let's discuss if this is the best approach or if there's something better (e.g. implementing an `Icon` trait or something). I can add a demo if we decide to move forward.

https://user-images.githubusercontent.com/1410520/150612047-357e837e-18f4-4c60-b29a-e5edc3034a61.mp4

### DIFF
--- a/egui/src/containers/collapsing_header.rs
+++ b/egui/src/containers/collapsing_header.rs
@@ -141,6 +141,7 @@ pub struct CollapsingHeader {
     selectable: bool,
     selected: bool,
     show_background: bool,
+    icon: Option<Box<dyn FnOnce(&mut Ui, f32, &Response)>>,
 }
 
 impl CollapsingHeader {
@@ -162,6 +163,7 @@ impl CollapsingHeader {
             selectable: false,
             selected: false,
             show_background: false,
+            icon: None,
         }
     }
 
@@ -236,6 +238,25 @@ impl CollapsingHeader {
         self.show_background = show_background;
         self
     }
+
+    /// Use the provided function to render a different `CollapsingHeader` icon.
+    /// Defaults to a triangle that animates as the `CollapsingHeader` opens and closes.
+    ///
+    /// For example:
+    /// ```
+    /// fn circle_icon(ui: &mut Ui, openness: f32, response: &Response) {
+    ///     let stroke = ui.style().interact(&response).fg_stroke;
+    ///     ui.painter().circle_filled(response.rect.center(), 2.0 + openness, stroke.color);
+    /// }
+    ///
+    /// CollapsingHeader::new("Circles")
+    ///   .icon(circle_icon)
+    ///   .show(ui, |ui| { ui.label("Hi!"); });
+    /// ```
+    pub fn icon(mut self, icon_fn: impl FnOnce(&mut Ui, f32, &Response) + 'static) -> Self {
+        self.icon = Some(Box::new(icon_fn));
+        self
+    }
 }
 
 struct Prepared {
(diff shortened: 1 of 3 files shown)


egui/src/containers/collapsing_header.rs
```diff
@@ -141,6 +141,7 @@ pub struct CollapsingHeader {
     selectable: bool,
     selected: bool,
     show_background: bool,
+    icon: Option<Box<dyn FnOnce(&mut Ui, f32, &Response)>>,
```
emilk, line 144: `pub type IconPainter = Box<dyn FnOnce(&mut Ui, f32, &Response)>;` with a nice docstring would make this type simpler, and silence clippy

egui/src/containers/collapsing_header.rs
```diff
@@ -236,6 +238,28 @@ impl CollapsingHeader {
         self.show_background = show_background;
         self
     }
+
+    /// Use the provided function to render a different `CollapsingHeader` icon.
+    /// Defaults to a triangle that animates as the `CollapsingHeader` opens and closes.
+    ///
+    /// For example:
+    /// ```
+    ///
```
emilk, line 247:
```suggestion
```

egui/src/containers/collapsing_header.rs
```diff
@@ -236,6 +238,28 @@ impl CollapsingHeader {
...
+    /// Use the provided function to render a different `CollapsingHeader` icon.
+    /// Defaults to a triangle that animates as the `CollapsingHeader` opens and closes.
+    ///
+    /// For example:
+    /// ```
+    /// 
+    /// # egui::__run_test_ui(|ui| {
+    /// fn circle_icon(ui: &mut egui::Ui, openness: f32, response: &egui::Response) {
+    ///     let stroke = ui.style().interact(&response).fg_stroke;
+    ///     ui.painter().circle_filled(response.rect.center(), 2.0 + openness, stroke.color);
```
emilk, line 251:
```suggestion
    ///     let radius = egui::lerp(2.0..=3.0, openness);
    ///     ui.painter().circle_filled(response.rect.center(), radius, stroke.color);
```

thanks for adding example code! Perhaps this would make it even more self-documenting!

egui/src/containers/collapsing_header.rs
```diff
@@ -341,7 +366,11 @@ impl CollapsingHeader {
                     ..header_response.clone()
                 };
                 let openness = state.openness(ui.ctx(), id);
-                paint_icon(ui, openness, &icon_response);
+                if let Some(icon) = icon {
+                    icon(ui, openness, &icon_response);
+                } else {
+                    paint_icon(ui, openness, &icon_response);
```
emilk, line 372: perhaps this could be renamed `paint_default_icon` now

emilk (approved):
Very nice, simple idea. I love it!

Only some minor nits and a clippy error to fix.

egui/src/containers/collapsing_header.rs
juancampa: I added the type alias and used it here.

I didn't use it in the `icon` function to keep the API simple (caller doesn't have to use `Box::new`). If [trait aliases](https://github.com/rust-lang/rust/issues/41517) ever get stabilized we could use that, but who knows if that will ever happen.

emilk (approved):

PR #1151, opened 2022-01-23
Reducing glutin dependency in egui_glow

t18b219k:
glutin is just a wrapper for winit and there is no reason to use it where it is not needed.
 The team is not interested in supporting WASM.
Minimize the use of glutin, so that egui_web can be easily retired when it should be retired, e.g. when winit on wasm supports IME.

### DIFF
--- a/egui_glow/src/epi_backend.rs
+++ b/egui_glow/src/epi_backend.rs
@@ -2,7 +2,9 @@ use crate::*;
 
 struct RequestRepaintEvent;
 
-struct GlowRepaintSignal(std::sync::Mutex<egui_winit::winit::event_loop::EventLoopProxy<RequestRepaintEvent>>, );
+struct GlowRepaintSignal(
+    std::sync::Mutex<egui_winit::winit::event_loop::EventLoopProxy<RequestRepaintEvent>>,
+);
 
 impl epi::backend::RepaintSignal for GlowRepaintSignal {
     fn request_repaint(&self) {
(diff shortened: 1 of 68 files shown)


egui_glow/src/epi_backend.rs
```diff
@@ -12,8 +14,8 @@ impl epi::backend::RepaintSignal for GlowRepaintSignal {
 
 #[allow(unsafe_code)]
 fn create_display(
-    window_builder: glutin::window::WindowBuilder,
-    event_loop: &glutin::event_loop::EventLoop<RequestRepaintEvent>,
+    window_builder: egui_winit::winit::window::WindowBuilder,
```
emilk, line 17: I think adding `use egui_winit::winit;` at the top of the file would increase readability of this file

emilk (approved):
Great idea!

Some merge conflicts due to
* https://github.com/emilk/egui/pull/1154

egui_glow/src/epi_backend.rs
t18b219k: ok i will fix it.

egui_glow/src/epi_backend.rs
```diff
@@ -1,8 +1,8 @@
 use crate::*;
-
+use egui_winit::winit;
 struct RequestRepaintEvent;
```
emilk, line 3:
```suggestion

struct RequestRepaintEvent;
```

emilk (approved):

PR #1154, opened 2022-01-23
Choose your own font and size

emilk:
This PR rewrites a lot of the font and `TextStyle` related code so that you can now no longer limited to what's in `TextStyle`. Now you can use any font size and any font family in all widgets and painting.

`TextStyle` still exists, but has been moved from `epaint` to `egui`, with a look-up-table in `egui::Style`. It has also been extended with `TextStyle::Name` so you can create your own named text styles.

`FontFamily` also has a `FontFamily::Name` variant so you can name any of your custom font families (if `FontFamily::Proportional` and `FontFamily::Monospace` are not enough for you).

`FontId` is a new `struct` that contains a `FontFamily` and a size.

# TODO:
* [x] Update changelogs
* [x] Improve docs
* [x] Self-review


coderedart:
What happens if fonts don't fit in a texture within max Tex size?

emilk:
> What happens if fonts don't fit in a texture within max Tex size?

Each font only allocates the glyph it actually uses (*), so a font texture of e.g. 4096x4096 can fit _a lot_ of glyphs. If it gets full, the atlas is cleared and filled in again.

The atlas can still get full if you use a lot of glyphs of different types and font sizes at the same time. For that to realistically happen though, you are filling your screen with all kinds of different text (or have a lot of varying text of varying sizes in a scroll area). Anyway, if this happens: an error is printed to stderr (I will likely start using the `log` crate soon…) and the atlas will be cleared each frame, leading to a big slowdown.

I believe this could happen if you load a big font (which supports tens of thousands of glyphs) and open the Font Book in the demo (which will paint all glyphs in the font), but otherwise I _hope_ this will be extremely rare.

For now I think this is a reasonable limitation, as a more robust solution would require multiple font textures which will complicate text painting (needing to switch textures mid-mesh), or require a completely different way of rendering text.

(*) All of ASCII is currently pre-cached for each font, but this might change

PR #1157, opened 2022-01-24
Add a `StepButtons` widget

### LINKED ISSUE #1156: `DragValue` increment buttons
I'd like to have the option to add step buttons (-/+) to `DragValue`, like they exist in imgui as well:

image

s-nie:
Closes #1156

~~This required a bit more changes than first expected. I removed `Ui::next_auto_id` as it was not needed anymore and only used inside the crate.~~

image

This was developed internally at [embotech AG](https://www.embotech.com). We use egui extensively and would like to contribute changes that could benefit the open-source community.

PS: I'm also @EmbersArc when making private contributions :)

### DIFF
--- a/egui/src/widgets/drag_value.rs
+++ b/egui/src/widgets/drag_value.rs
@@ -56,6 +56,7 @@ pub struct DragValue<'a> {
     clamp_range: RangeInclusive<f64>,
     min_decimals: usize,
     max_decimals: Option<usize>,
+    step: f64,
 }
 
 impl<'a> DragValue<'a> {
@@ -85,6 +86,7 @@ impl<'a> DragValue<'a> {
             clamp_range: f64::NEG_INFINITY..=f64::INFINITY,
             min_decimals: 0,
             max_decimals: None,
+            step: 0.0,
         }
     }
 
@@ -97,6 +99,12 @@ impl<'a> DragValue<'a> {
     /// Clamp incoming and outgoing values to this range.
     pub fn clamp_range<Num: emath::Numeric>(mut self, clamp_range: RangeInclusive<Num>) -> Self {
         self.clamp_range = clamp_range.start().to_f64()..=clamp_range.end().to_f64();
+        self
+    }
+
+    /// Add decrement [-] and increment [+] buttons and set the step value.
+    pub fn step_buttons<Num: emath::Numeric>(mut self, step: Num) -> Self {
+        self.step = step.to_f64();
         self
     }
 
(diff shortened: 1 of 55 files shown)


egui/src/widgets/drag_value.rs
```diff
@@ -157,130 +165,176 @@ impl<'a> Widget for DragValue<'a> {
...
-                        .flatten();
-                    let stored_value = stored_value.unwrap_or(value);
-                    let stored_value = stored_value + delta_value as f64;
-
-                    let aim_delta = aim_rad * speed;
-                    let rounded_new_value = emath::smart_aim::best_in_range_f64(
-                        stored_value - aim_delta,
-                        stored_value + aim_delta,
+        let mut response = ui
+            .horizontal(|ui| {
```
emilk, line 209: This changes `DragValue` from being an atomic widget (which can e.g. be centered in a layout) to a composite widget, which is a bit of a shame. Do the step button really need to be part of the `DragValue` widget? Perhaps it would make the code simpler if there instead was a `DragStepValue` widget which consisted of a `DragValue` and two buttons? Seems more composable than just adding more features into the same widget.

s-nie: Good points, I extracted the buttons into a separate widget. Let me know what you think.

What's your opinion on using the unicode characters for the `-` and `+` buttons? They look nice when the font supports it but won't show correctly if it doesn't.

egui_demo_lib/src/apps/demo/widget_gallery.rs
```diff
@@ -171,7 +171,15 @@ impl WidgetGallery {
...
         ui.add(doc_link_label("DragValue", "DragValue"));
-        ui.add(egui::DragValue::new(scalar).speed(1.0));
+        ui.add(egui::DragValue::new(scalar).speed(1.0).clamp_range(0..=360));
+        ui.end_row();
+
+        ui.add(doc_link_label("StepButtons", "StepButtons"));
+        ui.add(
+            egui::StepButtons::new(scalar)
+                .step(1.0)
+                .clamp_range(0..=360),
```
emilk, line 181: This doesn't actually show the value - isn't it more common for one to want to show the value between the -/+ buttons?

At this point it feels like this is just:

``` rust
if ui.button("+").clicked() { value += 1 }
if ui.button("-").clicked() { value -= 1 }
value = value.clamp(0, 360);
```

with very little extra value added.

See https://www.nngroup.com/articles/input-steppers/ for some examples of input steppers.

I think for this widget to motivate it's own existence, it should also display the value (between the -/+ buttons), and if the user wants finer control than that, they can just write it out themselves (like I did above).

What do you think?

s-nie: Yes, I'm afraid you're right. I still think that step buttons would be useful but maybe this is not the way to do it. Maybe it would be possible to keep the widget atomic and still have the step buttons on it. But I'm closing this issue for now.

egui/src/widgets/drag_value.rs
emilk: I think using the fancy unicode characters makes sense for now - if it causes a problem down-the-road we can think about adding fallbacks.

PR #1180, opened 2022-01-29
Customize grid spacing in plots

Bromeon:
# Description

Adds two `Plot` methods `x_grid_spacer()` and `y_grid_spacer()`, which customize the distance between the plot's background grid. The default is the log-10 grid, but occasionally, it can be useful to have different subdivision than 10. Example use cases:
* date and time
* non-decimal unit systems (imperial)
* other bases (binary, hexadecimal)
* ...

I added a small example just to visualize it. It's definitely not the best, but maybe we can improve this in the future, in a separate PR.
grafik

# Implementation

While providing a customization API, this PR also refactors the hardcoded log-10 logic a bit, making it slightly more modular and easier to understand.

The customization function look as follows:
```rs
fn get_step_sizes(
    bounds: (f64, f64),
    bounds_frame_ratio: f64
) -> [f64; 3]
```
where `bounds` are the boundaries of the visible diagram (plot coordinates), and `bounds_frame_ratio` is a bounds/frame ratio, i.e how much one plot unit amounts to in the visible viewport (scaled with min step size).

The built-in logarithmic spacer is implemented as follows:
```rs
pub fn log_grid_spacer(base: i64) -> GridSpacer {
    let base = base as f64;
    let get_step_sizes = move |_bounds: (f64, f64), bounds_frame_ratio: f64| -> [f64; 3] {
        // The distance between two of the thinnest grid lines is "rounded" up
        // to the next-bigger power of base
        let smallest_visible_unit = next_power(bounds_frame_ratio, base);

        [
            smallest_visible_unit,
            smallest_visible_unit * base,
            smallest_visible_unit * base * base,
        ]
    };

    Box::new(get_step_sizes)
}
```
Now, why an array of exactly 3 elements? In the current implementation, 3 is the number of different thicknesses, with which the grid lines are rendered. The thick line for the largest unit (e.g. 10s), the medium for the middle one (e.g. 1s), and the thin line for the smallest unit (e.g. 0.1s).


Bromeon:
I realized that having a fixed step size may not be flexible enough. Consider e.g. using months as input -- it would be nice if there could be a grid line at every beginning of a month. Now months are 28, 29, 30 or 31 days long, so equidistant grid lines won't cut it.

Instead, _grid spacers_ have the following function signature:
```rs
fn get_step_sizes(input: GridInput) -> Vec<GridMark>;
```
with:
```rs
pub struct GridInput {
    /// Min/max of the visible data range (the values at the two edges of the plot,
    /// for the current axis).
    pub bounds: (f64, f64),

    /// Ratio between the diagram's bounds (in plot coordinates) and the viewport
    /// (in frame/window coordinates), scaled up to represent the minimal possible step.
    /// This is a good value to be used as the smallest of the returned steps.
    pub bounds_frame_ratio: f64,
}

/// One mark (horizontal or vertical line) in the background grid of a plat.
pub struct GridMark {
    /// X or Y value in the plot.
    pub value: f64,

    /// The (approximate) distance to the next value of same thickness.
    /// Determines how thick the grid line is painted at that point.
    pub step_size: f64,
}
```

So for any given range (called `bounds`), the user can decide where exactly the grid lines appear.

Since this may be a bit tedious to implement for more simple cases, I provided two convenience functions:
* **Logarithmic spacing, e.g. `2`, `4`, `8`, `16`...**
  ```rs
   pub fn log_grid_spacer(base: i64) -> Box<dyn Fn(GridInput) -> Vec<GridMark>>
   ```

* **Uniform custom spacing, e.g. `1`, `25`, `100`...**
  This lets the user return 3 step sizes (dependent on bounds) and automatically adds grid lines at multiples of those steps. I used a fixed-size array because 3 is the number of different thicknesses used for grid lines, we could think about allowing to provide only 1 or 2 as well (e.g. by specifying `0.0` or `None`).
  ```rs
   pub fn uniform_grid_spacer(
      spacer: impl Fn(GridInput) -> [f64; 3] + 'static
   ) -> Box<dyn Fn(GridInput) -> Vec<GridMark>>
   ```

The results of those two functions can be passed to `Plot::{x/y}_grid_spacer()`.

Bromeon:
Fixed conflicts, merged latest master.

Bromeon:
Any feedback on this? 🙂

egui/src/widgets/plot/mod.rs
```diff
@@ -376,6 +384,48 @@ impl Plot {
...
+    /// 100 and small ones at 25, you can return:
+    /// ```no_run
+    /// # use egui::plot::GridMark;
+    /// vec![
+    ///    // 100s
+    ///    GridMark { value: 100.0, step_size: 100.0 },
+    ///    GridMark { value: 200.0, step_size: 100.0 },
+    ///
+    ///    // 25s
+    ///    GridMark { value: 125.0, step_size: 25.0 },
```
emilk, line 407: is it `step_size` that makes these smaller (fainter)?

egui/src/widgets/plot/mod.rs
```diff
@@ -897,6 +949,79 @@ impl PlotUi {
...
+/// See [`Plot::x_grid_spacer()`] and [`Plot::y_grid_spacer()`].
+pub struct GridInput {
+    /// Min/max of the visible data range (the values at the two edges of the plot,
+    /// for the current axis).
+    pub bounds: (f64, f64),
+
+    /// Ratio between the diagram's bounds (in plot coordinates) and the viewport
+    /// (in frame/window coordinates), scaled up to represent the minimal possible step.
+    /// This is a good value to be used as the smallest of the returned steps.
+    pub bounds_frame_ratio: f64,
```
emilk, line 966: I think this should be called something like `min_line_spacing` or something.

egui/src/widgets/plot/mod.rs
```diff
@@ -966,43 +1093,34 @@ impl PreparedPlot {
...
-            if value_main > bounds.max[axis] {
-                break;
-            }
+        let bounds = (bounds.min[axis], bounds.max[axis]);
+        let bounds_frame_ratio = transform.dvalue_dpos()[axis] * MIN_LINE_SPACING_IN_POINTS;
+
+        let input = GridInput {
+            bounds,
+            bounds_frame_ratio,
+        };
```
emilk, line 1106: minimize the amount of bindings:

```suggestion
        let input = GridInput {
            bounds: (bounds.min[axis], bounds.max[axis]),
            bounds_frame_ratio: transform.dvalue_dpos()[axis] * MIN_LINE_SPACING_IN_POINTS,
        };
```

egui/src/widgets/plot/mod.rs
```diff
@@ -1094,3 +1212,36 @@ impl PreparedPlot {
...
+/// Fill in all values between [min, max] which are a multiple of `step_size`
+fn fill_marks_between(out: &mut Vec<GridMark>, step_size: f64, (min, max): (f64, f64)) {
+    assert!(max > min);
+    let first = (min / step_size).ceil() as i64;
+    let last = (max / step_size).ceil() as i64;
+
+    for i in first..last {
+        let value = (i as f64) * step_size;
+        out.push(GridMark { value, step_size });
+    }
```
emilk, line 1246: using `out.extend` should be faster than repeatedly calling `.push`

egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -626,6 +626,8 @@ impl ChartsDemo {
             }
         };
 
+        let grid = plot::uniform_grid_spacer(|_input| -> [f64; 3] { [1.0, 5.0, 30.0] });
```
emilk, line 629: Can we come up with a better demo for the usefulness of this new feature? Dividing up days in multiples of 5 and 30 is not very common.

emilk (commented):
It's a nice feature to have, but it could do with a more motivated demo

egui_demo_lib/src/apps/demo/plot_demo.rs
DusterTheFirst: Splitting time into minutes and seconds could be a good demo

egui/src/widgets/plot/mod.rs
Bromeon: Done.

egui/src/widgets/plot/mod.rs
Bromeon: Renamed to `base_step_size`.

egui/src/widgets/plot/mod.rs
Bromeon: Yes, exactly.

I added a line to clarify this:
```rs
    /// `step_size` also determines how thick/faint each line is drawn.
```

egui/src/widgets/plot/mod.rs
```diff
@@ -1094,3 +1212,36 @@ impl PreparedPlot {
         }
     }
 }
+
+/// Returns next bigger power in given base
+/// e.g.
+/// ```ignore
+/// next_power(0.01, 10) == 0.01
+/// next_power(0.02, 10) == 0.1
+/// next_power(0.2,  10) == 1
```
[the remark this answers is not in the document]
Bromeon: Unfortunately, that would require making this function public -- and I'm not sure if that's worth it just for doc tests (and where it would belong in that case).

egui/src/widgets/plot/mod.rs
Bromeon: Done (needed conversion into an iterator).

egui_demo_lib/src/apps/demo/plot_demo.rs
Bromeon: Done.

Bromeon:
Added a new plot category called `Custom Axes`, which showcases multiple features at once:
* Custom axis value formatting
* Custom label formatting (when hovering)
* Custom grid spacing

For this reason, I removed the custom axis formatting from the bar chart I added back in the day, as it seemed too artificial there.

The new demo uses time on the X axis (days, hours, minutes) and percent on the Y axis:

https://user-images.githubusercontent.com/708488/158074012-8604c0b2-9cb8-4e8e-9fea-e7b20425cc20.mp4

egui/src/widgets/plot/mod.rs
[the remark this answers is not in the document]
emilk: Oh right, you cannot doctest private functions 🤦 

But perhaps this could belong in `emath` anyway

egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -309,6 +311,117 @@ impl Widget for &mut LegendDemo {
...
+            } else {
+                // Hours and minutes
+                format!("{h}:{m:02}", h = get_hour(x), m = get_minute(x))
+            }
+        };
+
+        let y_fmt = |y, _range: &RangeInclusive<f64>| {
+            // Display only integer percentages
+            if !is_approx_zero(y) && is_approx_integer(100.0 * y) {
+                format!("{}%", get_percent(y))
```
emilk, line 396:
```suggestion
                format!("{:.0}%", (100.0 * y).round())
```

egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -309,6 +311,117 @@ impl Widget for &mut LegendDemo {
...
+        }
+
+        marks
+    }
+}
+
+impl Widget for &mut CustomAxisDemo {
+    fn ui(self, ui: &mut Ui) -> Response {
+        // Make sure floor() still rounds down, even in presence of arithmetic imprecision
+        const EPS: f64 = 1e-6;
```
emilk, line 365: An alternative is to use minutes (or seconds) as the base unit (instead of days). That would make the precision problems go away since `f32` can express all integers up to 2^24 perfectly.

emilk (changes requested):
Sorry for the slow (re)review! It looks great, but the changelog line needs to move up!

egui_demo_lib/src/apps/demo/plot_demo.rs
Bromeon: `get_percent()` is used twice, not sure if the duplication would make things easier/clearer. Also, pre-formatting in two places would once just produce an intermediate string which is again formatted.

I left the method as-is for now, but open for suggestions:
```rs
fn get_percent(y: f64) -> f64 {
    (100.0 * y + EPS).floor()
}
```

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui_w
 * `Ui::input_mut` to modify how subsequent widgets see the `InputState` and a convenience method `InputState::consume_key` for shortcuts or hotkeys ([#1212](https://github.com/emilk/egui/pull/1212)).
 * Added `Ui::add_visible` and `Ui::add_visible_ui`.
 * Added `CollapsingHeader::icon` to override the default open/close icon using a custom function. ([1147](https://github.com/emilk/egui/pull/1147)).
+* Added `Plot::x_grid_spacer` and `Plot::y_grid_spacer` for custom grid spacing ([#1180](https://github.com/emilk/egui/pull/1180)).
```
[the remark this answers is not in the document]
Bromeon: Done!

egui_demo_lib/src/apps/demo/plot_demo.rs
Bromeon: Ok, I tried that approach, let me know what you think. Maybe some parts can be further simplified...

Bromeon:
Implemented feedback (besides `get_percent()`, see above).

One issue with the new _minute_ base unit is that the default view is very much zoomed in.  
1. I was under the impression egui would adjust the plot bounds according to the data, but maybe I remember this wrong?
2. Is there a way to explicitly set the visible bounds?

egui_demo_lib/src/apps/demo/plot_demo.rs
emilk: Ooh, I missed it was used twice.

`.round()` removed the need for `EPS`, and is more clear (imho), but it is a minor nit.

Bromeon: Sure! Simplified it:
```rs
fn get_percent(y: f64) -> f64 {
    (100.0 * y).round()
}
```

Bromeon:
That was a sneaky merge conflict, a few minutes after my push 😁 

Also removed the colon here in the changelog:
```md
`ComboBox`:es   -->   `ComboBox`es
```
I assume that was a typo? Because further below you also used
```md
`DragValue`s
```
without any colon. If not, I'll revert.

emilk (approved):

emilk:
> One issue with the new _minute_ base unit is that the default view is very much zoomed in.
> 
>     1. I was under the impression egui would adjust the plot bounds according to the data, but maybe I remember this wrong?

Not for "generator" functions (`Values::from_explicit_callback`) - but I'll push a fix for that after merging this PR.

>     2. Is there a way to explicitly set the visible bounds?
I don't think so, but we should add it

nrot:
Hello. 
``` rust 
fn next_power()
``` 
have this 
``` rust 
assert_ne!(value, 0.0); // can be negative (typical for Y axis)
``` 
This create random panic on runtime, when used plot with boxed zoom. I think this happing when boxed zoom create to small zone and plot cant calculate new grid. I don`t shure, because cant stable repeating this bug. Perhaps need some simple protected of zero, but i dont sure where exactly need this protection
```
[2023-05-31][22:30:12][wave_gen::app::waves][DEBUG] Mouse position: Some([355.0 242.0])
[2023-05-31][22:30:12][wave_gen::app::waves][DEBUG] Clicked by plot: 22; Data size: 128
[2023-05-31][22:30:12][wave_gen::app::waves][DEBUG] Mouse position: Some([355.0 242.0])
thread 'main' panicked at 'assertion failed: `(left != right)`
  left: `-0.0`,
 right: `0.0`', github.com-1ecc6299db9ec823\egui-0.22.0\src\widgets\plot\mod.rs:1637:5
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library\std\src\panicking.rs:579
   1: core::panicking::panic_fmt
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library\core\src\panicking.rs:64
   2: core::fmt::Arguments::new_v1
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library\core\src\fmt\mod.rs:405
   3: core::panicking::assert_failed_inner
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library\core\src\panicking.rs:257
   4: core::panicking::assert_failed<f64,f64>
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc\library\core\src\panicking.rs:211
   5: egui::widgets::plot::next_power
             at github.com-1ecc6299db9ec823\egui-0.22.0\src\widgets\plot\mod.rs:1637
   6: egui::widgets::plot::log_grid_spacer::closure$0
             at github.com-1ecc6299db9ec823\egui-0.22.0\src\widgets\plot\mod.rs:1283
   7: alloc::boxed::impl$47::call<tuple$<egui::widgets::plot::GridInput>,dyn$<core::ops::function::Fn<tuple$<egui::widgets::plot::GridInput>,assoc$<Output,alloc::vec::Vec<egui::widgets::plot::GridMark,alloc::alloc::Global> > > >,alloc::alloc::Global>
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc\library\alloc\src\boxed.rs:2001
   8: egui::widgets::plot::PreparedPlot::paint_axis
             at github.com-1ecc6299db9ec823\egui-0.22.0\src\widgets\plot\mod.rs:1458
   9: egui::widgets::plot::PreparedPlot::ui
             at github.com-1ecc6299db9ec823\egui-0.22.0\src\widgets\plot\mod.rs:1340
  10: egui::widgets::plot::Plot::show_dyn<tuple$<> >
             at github.com-1ecc6299db9ec823\egui-0.22.0\src\widgets\plot\mod.rs:971
  11: egui::widgets::plot::Plot::show<tuple$<>,wave_gen::app::waves::impl$2::display::closure$0::closure$0::closure_env$1>
             at github.com-1ecc6299db9ec823\egui-0.22.0\src\widgets\plot\mod.rs:588
  12: wave_gen::app::waves::impl$2::display::closure$0::closure$0
             at .\src\app\waves\mod.rs:115
  13: core::ops::function::FnOnce::call_once<wave_gen::app::waves::impl$2::display::closure$0::closure_env$0,tuple$<ref_mut$<egui::ui::Ui> > >
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc\library\core\src\ops\function.rs:250
  14: alloc::boxed::impl$45::call_once<tuple$<ref_mut$<egui::ui::Ui> >,dyn$<core::ops::function::FnOnce<tuple$<ref_mut$<egui::ui::Ui> >,assoc$<Output,tuple$<> > > >,alloc::alloc::Global>
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc\library\alloc\src\boxed.rs:1987
  15: egui::ui::Ui::allocate_ui_with_layout_dyn<tuple$<> >
             at github.com-1ecc6299db9ec823\egui-0.22.0\src\ui.rs:880
  16: egui::ui::Ui::allocate_ui_with_layout<tuple$<>,wave_gen::app::waves::impl$2::display::closure$0::closure_env$0>
             at github.com-1ecc6299db9ec823\egui-0.22.0\src\ui.rs:865
  17: egui::ui::Ui::allocate_ui<tuple$<>,wave_gen::app::waves::impl$2::display::closure$0::closure_env$0>
             at github.com-1ecc6299db9ec823\egui-0.22.0\src\ui.rs:851
  18: wave_gen::app::waves::impl$2::display::closure$0
  ...
```

PR #1187, opened 2022-01-31
allow `RepaintSignal: ?Sync`

### LINKED ISSUE #1186: `RepaintSignal: Sync`
`epi::backend::RepaintSignal` requires  `Sync` (introduced in #82), which requires wrapping `!Sync` types (`winnit::EventLoopProxy` for example) with a mutex. Not cool IMO.

ZetaNumbers:
Closes #1186.

### DIFF
--- a/egui_web/src/backend.rs
+++ b/egui_web/src/backend.rs
@@ -78,6 +78,10 @@ impl epi::backend::RepaintSignal for NeedRepaint {
     fn request_repaint(&self) {
         self.0.store(true, SeqCst);
     }
+
+    fn dyn_clone(&self) -> Box<dyn epi::backend::RepaintSignal> {
+        Box::new(self.clone())
+    }
 }
 
 // ----------------------------------------------------------------------------
--- a/egui_web/src/backend.rs
+++ b/egui_web/src/backend.rs
@@ -55,11 +55,12 @@ impl WebInput {
 
 use std::sync::atomic::Ordering::SeqCst;
 
-pub struct NeedRepaint(std::sync::atomic::AtomicBool);
+#[derive(Clone)]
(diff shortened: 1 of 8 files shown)


egui_web/src/backend.rs
```diff
@@ -55,11 +55,12 @@ impl WebInput {
 
 use std::sync::atomic::Ordering::SeqCst;
 
-pub struct NeedRepaint(std::sync::atomic::AtomicBool);
+#[derive(Clone)]
+pub struct NeedRepaint(std::sync::Arc<std::sync::atomic::AtomicBool>);
```
Mingun, line 59: Atomic reference counter for an atomic variable seems an overkill.

ZetaNumbers: It has been there already, i just moved it. ¯\\\_(ツ)\_/¯

emilk:
Is all this new code just to avoid the performance impact of locking a mutex when the users requests a repaint? Or is there some other benefit I'm not seeing?

ZetaNumbers:
> Is all this new code just to avoid the performance impact of locking a mutex when the users requests a repaint? Or is there some other benefit I'm not seeing?

This PR firstly fixes logic of the library and only then the performance. There are several of benefits to it:
- Avoids mutex poisoning and panic deadlocks (i. e. if some thread panics while using `request_repaint`, then other threads would panic on a `request_repaint` too);
- Use of DST eliminates an indirection;
- Avoids unnecessary locks (majority of channel `Sender`s (like mpsc) are `Send + !Sync` which are used by window handling libraries);
- There may be other downsides of relying on mutex, like if it's not implemented for some platform or another indirection (posix targets store mutex on the heap IIRC).

Issue #1125, opened 2022-01-16
Check.sh fails to compile speech-dispatcher


juancampa:
**Describe the bug**
check.sh fails to run ith the following compilation error

<details>
  <summary>Click to expand check.sh output</summary>

```
  + cargo check --workspace --all-targets --all-features
      Checking egui v0.16.1 (/home/juan/git/egui/egui)
      Checking speech-dispatcher v0.9.0
  error[E0308]: mismatched types
    --> /home/juan/.cargo/registry/src/github.com-1ecc6299db9ec823/speech-dispatcher-0.9.0/src/lib.rs:35:13
     |
  35 |     Male1 = SPDVoiceType::SPD_MALE1,
     |             ^^^^^^^^^^^^^^^^^^^^^^^ expected `u32`, found `i32`
  
  error[E0308]: mismatched types
    --> /home/juan/.cargo/registry/src/github.com-1ecc6299db9ec823/speech-dispatcher-0.9.0/src/lib.rs:36:13
     |
  36 |     Male2 = SPDVoiceType::SPD_MALE2,
     |             ^^^^^^^^^^^^^^^^^^^^^^^ expected `u32`, found `i32`
  
  error[E0308]: mismatched types
    --> /home/juan/.cargo/registry/src/github.com-1ecc6299db9ec823/speech-dispatcher-0.9.0/src/lib.rs:37:13
     |
  37 |     Male3 = SPDVoiceType::SPD_MALE3,
     |             ^^^^^^^^^^^^^^^^^^^^^^^ expected `u32`, found `i32`
  
  error[E0308]: mismatched types
    --> /home/juan/.cargo/registry/src/github.com-1ecc6299db9ec823/speech-dispatcher-0.9.0/src/lib.rs:38:15
     |
  38 |     Female1 = SPDVoiceType::SPD_FEMALE1,
     |               ^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u32`, found `i32`
  
  error[E0308]: mismatched types
    --> /home/juan/.cargo/registry/src/github.com-1ecc6299db9ec823/speech-dispatcher-0.9.0/src/lib.rs:39:15
     |
  39 |     Female2 = SPDVoiceType::SPD_FEMALE2,
     |               ^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u32`, found `i32`
  
  error[E0308]: mismatched types
[...]

PR #1206, opened 2022-02-04
Per-corner rounding of rectangles

4JX:
Ever got bored of those plain old boring corners? Don't you feel they all look the same? Fear not, I have the solution for you:

Uses a new `Rounding` struct as discussed on the discord. Although maybe redundant, also gave it a `new` function to shorten things a bit when creating more complex layouts.

~~I separated each corner into its own slider for what is shown in the demo app, however I have no idea where the other sliders in `WidgetVisuals` go to.~~ Oh those are for widgets themselves. Seems the labels fit the style so that's good.

**Note:** I was not able to run `check.sh` which seems to have already been reported [here](https://github.com/emilk/egui/issues/1125). Got a bunch of `mismatched type` errors.

### DIFF
--- a/epaint/src/shape.rs
+++ b/epaint/src/shape.rs
@@ -114,12 +114,20 @@ impl Shape {
     }
 
     #[inline]
-    pub fn rect_filled(rect: Rect, corner_radius: f32, fill_color: impl Into<Color32>) -> Self {
+    pub fn rect_filled(
+        rect: Rect,
+        corner_radius: impl Into<Rounding>,
+        fill_color: impl Into<Color32>,
+    ) -> Self {
         Self::Rect(RectShape::filled(rect, corner_radius, fill_color))
     }
 
     #[inline]
-    pub fn rect_stroke(rect: Rect, corner_radius: f32, stroke: impl Into<Stroke>) -> Self {
+    pub fn rect_stroke(
+        rect: Rect,
+        corner_radius: impl Into<Rounding>,
+        stroke: impl Into<Stroke>,
+    ) -> Self {
         Self::Rect(RectShape::stroke(rect, corner_radius, stroke))
     }
 
@@ -321,28 +329,36 @@ impl From<PathShape> for Shape {
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub struct RectShape {
     pub rect: Rect,
-    /// How rounded the corners are. Use `0.0` for no rounding.
-    pub corner_radius: f32,
+    /// How rounded the corners are. Use `Rounding::none()` for no rounding.
+    pub corner_radius: Rounding,
     pub fill: Color32,
     pub stroke: Stroke,
 }
 
 impl RectShape {
     #[inline]
-    pub fn filled(rect: Rect, corner_radius: f32, fill_color: impl Into<Color32>) -> Self {
+    pub fn filled(
+        rect: Rect,
+        corner_radius: impl Into<Rounding>,
+        fill_color: impl Into<Color32>,
+    ) -> Self {
         Self {
             rect,
-            corner_radius,
+            corner_radius: corner_radius.into(),
             fill: fill_color.into(),
             stroke: Default::default(),
         }
     }
 
     #[inline]
-    pub fn stroke(rect: Rect, corner_radius: f32, stroke: impl Into<Stroke>) -> Self {
+    pub fn stroke(
+        rect: Rect,
+        corner_radius: impl Into<Rounding>,
+        stroke: impl Into<Stroke>,
+    ) -> Self {
         Self {
             rect,
-            corner_radius,
+            corner_radius: corner_radius.into(),
             fill: Default::default(),
             stroke: stroke.into(),
         }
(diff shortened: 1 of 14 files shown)


4JX:
Not too sure of whether the changelog for this would go on either `egui`/`epaint` or both.

4JX:
Whoops. Seems github does dont like renaming branches with open PRs

epaint/src/shape.rs
```diff
@@ -362,6 +378,57 @@ impl From<RectShape> for Shape {
...
 }
 
+#[derive(Copy, Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+/// How rounded the corners of things should be
+pub struct Rounding {
+    pub nw: f32,
+    pub ne: f32,
+    pub sw: f32,
+    pub se: f32,
```
emilk, line 388:
```suggestion
    /// north-west = left-top
    pub nw: f32,
    /// north-east = right-top
    pub ne: f32,
    /// south-west = left-bottom
    pub sw: f32,
    /// south-east = right-bottom
    pub se: f32,
```

epaint/src/shape.rs
```diff
@@ -362,6 +378,57 @@ impl From<RectShape> for Shape {
...
+    pub fn same(radius: f32) -> Self {
+        Self {
+            nw: radius,
+            ne: radius,
+            sw: radius,
+            se: radius,
+        }
+    }
+
+    pub fn none() -> Self {
```
emilk, line 422:
```suggestion
    #[inline]
    pub fn none() -> Self {
```

epaint/src/shape.rs
```diff
@@ -362,6 +378,57 @@ impl From<RectShape> for Shape {
...
+        }
+    }
+}
+
+impl Rounding {
+    pub fn new(nw: f32, ne: f32, sw: f32, se: f32) -> Self {
+        Self { nw, ne, sw, se }
+    }
+
+    pub fn same(radius: f32) -> Self {
```
emilk, line 413:
```suggestion
    #[inline]
    pub fn same(radius: f32) -> Self {
```

epaint/src/shape.rs
```diff
@@ -362,6 +378,57 @@ impl From<RectShape> for Shape {
...
+            se: radius,
+        }
+    }
+}
+
+impl Rounding {
+    pub fn new(nw: f32, ne: f32, sw: f32, se: f32) -> Self {
+        Self { nw, ne, sw, se }
+    }
+
```
emilk, line 412: I think this will just lead to confusion - better for users to be explicit imho

```suggestion
```

epaint/src/shape.rs
```diff
@@ -362,6 +378,57 @@ impl From<RectShape> for Shape {
...
+}
+
+impl Default for Rounding {
+    fn default() -> Self {
+        Self::none()
+    }
+}
+
+impl From<f32> for Rounding {
+    fn from(radius: f32) -> Self {
```
emilk, line 398:
```suggestion
    #[inline]
    fn from(radius: f32) -> Self {
```

epaint/src/shape.rs
```diff
@@ -362,6 +378,57 @@ impl From<RectShape> for Shape {
...
+/// How rounded the corners of things should be
+pub struct Rounding {
+    pub nw: f32,
+    pub ne: f32,
+    pub sw: f32,
+    pub se: f32,
+}
+
+impl Default for Rounding {
+    fn default() -> Self {
```
emilk, line 392:
```suggestion
    #[inline]
    fn default() -> Self {
```

epaint/src/tessellator.rs
```diff
@@ -179,33 +179,23 @@ impl Path {
...
         path.clear();
 
         let min = rect.min;
         let max = rect.max;
 
-        let cr = corner_radius
-            .min(rect.width() * 0.5)
-            .min(rect.height() * 0.5);
-
-        if cr <= 0.0 {
```
emilk, line 195: This is an important common-case optimization. Please add it back! (`if cr == Rounding::none() { … }`)

epaint/src/tessellator.rs
```diff
@@ -242,6 +232,35 @@ pub mod path {
             path.push(center + radius * Vec2::angled(angle));
         }
     }
+
+    // Ensures the radius of each corner is within a valid range
+    fn clamp_radius(corner_radius: Rounding, rect: Rect) -> Rounding {
+        let half_width = rect.width() * 0.5;
+        let half_height = rect.height() * 0.5;
```
emilk, line 239:
```suggestion
        let half_height = rect.height() * 0.5;
        let max_cr = half_width.min(half_height);
```

then you only need one call to `.at_most(max_cr)` for each of the corners below (-> faster code)

emilk (changes requested):
Looks great, but I left some comments

epaint/src/shape.rs
4JX: Perhaps something like `/// Top left (north-west)` would be a bit better formatting wise for these?

epaint/src/shape.rs
4JX: I originally added it with the intention of shortening fmt's tendency to have one field per line for structs regardless of the length, but I'll remove it. I agree it might be confusing to some.

emilk (approved):
Spectacular - thanks!

PR #1209, opened 2022-02-04
Remove prevent_default for mousedown events

juancampa:
This fixes the issue of VSCode tabs not focusing when egui is running inside VSCode.

I left a comment just in case someone wants to add it back for completeness.


juancampa:
I tested the demo app on the browser and it seems to work just fine.

emilk (approved):

PR #1212, opened 2022-02-05
Add Ui.input_mut & InputState.ignore_key

cat-state:
This PR pipes through `ctx.input_mut()` through to the Uis, allowing control of what events subsequent widgets see, and adds a utility function to the InputState to do hotkeys/shortcuts.

Helps with https://github.com/emilk/egui/discussions/1124

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui_w
 
 
 ## Unreleased
-
+*  `Ui::input_mut` to modify how subsequent widgets see the `InputState` and a convenience `InputState::ignore_key` for shortcuts or hotkeys. 
 ### Added ⭐
 * Much improved font selection ([#1154](https://github.com/emilk/egui/pull/1154)):
   * You can now select any font size and family using `RichText::size` amd `RichText::family` and the new `FontId`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui_w
 ## Unreleased
 
 ### Added ⭐
-*  `Ui::input_mut` to modify how subsequent widgets see the `InputState` and a convenience method `InputState::ignore_key` for shortcuts or hotkeys ([#1212](https://github.com/emilk/egui/pull/1212)). 
+*  `Ui::input_mut` to modify how subsequent widgets see the `InputState` and a convenience method `InputState::consume_key` for shortcuts or hotkeys ([#1212](https://github.com/emilk/egui/pull/1212)). 
 * Much improved font selection ([#1154](https://github.com/emilk/egui/pull/1154)):
   * You can now select any font size and family using `RichText::size` amd `RichText::family` and the new `FontId`.
   * Easily change text styles with `Style::text_styles`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui_w
 ## Unreleased
 
 ### Added ⭐
-*  `Ui::input_mut` to modify how subsequent widgets see the `InputState` and a convenience `InputState::ignore_key` for shortcuts or hotkeys ([#1212](https://github.com/emilk/egui/pull/1212)). 
+*  `Ui::input_mut` to modify how subsequent widgets see the `InputState` and a convenience nethod `InputState::ignore_key` for shortcuts or hotkeys ([#1212](https://github.com/emilk/egui/pull/1212)). 
 * Much improved font selection ([#1154](https://github.com/emilk/egui/pull/1154)):
   * You can now select any font size and family using `RichText::size` amd `RichText::family` and the new `FontId`.
   * Easily change text styles with `Style::text_styles`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,9 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui_w
 
 
 ## Unreleased
-*  `Ui::input_mut` to modify how subsequent widgets see the `InputState` and a convenience `InputState::ignore_key` for shortcuts or hotkeys. 
+
 ### Added ⭐
+*  `Ui::input_mut` to modify how subsequent widgets see the `InputState` and a convenience `InputState::ignore_key` for shortcuts or hotkeys ([#1212](https://github.com/emilk/egui/pull/1212)). 
 * Much improved font selection ([#1154](https://github.com/emilk/egui/pull/1154)):
   * You can now select any font size and family using `RichText::size` amd `RichText::family` and the new `FontId`.
   * Easily change text styles with `Style::text_styles`.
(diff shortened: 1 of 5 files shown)


egui/src/input_state.rs
```diff
@@ -192,6 +192,23 @@ impl InputState {
         self.pointer.wants_repaint() || self.scroll_delta != Vec2::ZERO || !self.events.is_empty()
     }
 
+    // Ignore a key if it was pressed or released this frame. Useful for hotkeys.
+    // Returns if the key was pressed this frame
+    pub fn ignore_key(&mut self, ignore_key: Key, modifiers: Modifiers) -> bool {
+        self.events = std::mem::take(&mut self.events).into_iter().filter(|event| {
```
emilk, line 198: You can use `Vec::retain` here instead

egui/src/input_state.rs
```diff
@@ -192,6 +192,23 @@ impl InputState {
         self.pointer.wants_repaint() || self.scroll_delta != Vec2::ZERO || !self.events.is_empty()
     }
 
+    // Ignore a key if it was pressed or released this frame. Useful for hotkeys.
+    // Returns if the key was pressed this frame
+    pub fn ignore_key(&mut self, ignore_key: Key, modifiers: Modifiers) -> bool {
```
emilk, line 197: I think `consume_key` might be a better name

```suggestion
    /// Check for a key press. If found, `true` is returned and the key pressed i consumed, so that this will only return `true` once.
    pub fn consume_key(&mut self, modifiers: Modifiers, ignore_key: Key) -> bool {
```

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui_w
 
 
 ## Unreleased
-
+*  `Ui::input_mut` to modify how subsequent widgets see the `InputState` and a convenience `InputState::ignore_key` for shortcuts or hotkeys.
```
emilk, line 9: Put under the Added header, and add a link to this PR

egui/src/ui.rs
```diff
@@ -338,6 +338,23 @@ impl Ui {
...
 
+
+    /// The [`InputState`] of the [`Context`] associated with this [`Ui`].
+    /// Equivalent to `.ctx().input_mut()`.
+    ///
+    /// Note that this locks the [`Context`], so be careful with if-let bindings
+    /// like for [`Self::input()`].
+    /// ```
+    /// # egui::__run_test_ui(|ui| {
+    /// ui.input_mut().time = 0.0;
```
emilk, line 349: Use the new function in the example instead (messing with time is a really bad idea)

egui/src/input_state.rs
```diff
@@ -192,6 +192,23 @@ impl InputState {
...
+        self.events = std::mem::take(&mut self.events).into_iter().filter(|event| {
+            !matches!(
+                event,
+                Event::Key {
+                    key,
+                    modifiers: mods,
+                    ..
+                } if *key == ignore_key && *mods == modifiers
```
emilk, line 205: This will match both on press and release - I think we should only match on `pressed: true`

emilk (changes requested):
Need some polish, but I think this is the right direction.

It would be good to use the new function in `egui_demo_lib/src/easy_mark/easy_mark_editor.rs` (`fn shortcuts`) as a demonstration/test!

egui/src/input_state.rs
cat-state: Hmm, what if code depended on presses always being matched with releases? then you'd have an empty release instead of as if the event "never happened"

CHANGELOG.md
```diff
@@ -6,8 +6,9 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui_w
 
 
 ## Unreleased
-*  `Ui::input_mut` to modify how subsequent widgets see the `InputState` and a convenience `InputState::ignore_key` for shortcuts or hotkeys. 
+
 ### Added ⭐
+*  `Ui::input_mut` to modify how subsequent widgets see the `InputState` and a convenience `InputState::ignore_key` for shortcuts or hotkeys ([#1212](https://github.com/emilk/egui/pull/1212)).
```
Mingun:
```suggestion
*  `Ui::input_mut` to modify how subsequent widgets see the `InputState` and a convenience `InputState::consume_key` for shortcuts or hotkeys ([#1212](https://github.com/emilk/egui/pull/1212)). 
```

egui_demo_lib/src/easy_mark/easy_mark_editor.rs
```diff
@@ -117,59 +117,28 @@ impl EasyMarkEditor {
...
 fn shortcuts(ui: &Ui, code: &mut dyn TextBuffer, ccursor_range: &mut CCursorRange) -> bool {
     let mut any_change = false;
-    for event in &ui.input().events {
-        if let Event::Key {
+    for (key, surrounding) in [
+        (Key::B, "*"),
+        (Key::C, "`"),
+        (Key::I, "/"),
+        (Key::R, "~"),
+        (Key::U, "_"),
```
Mingun: You made a non-equivalent change, is this expected? Also I'd prefer to kept comments:
```suggestion
        (Key::B, "*"), // *bold*
        (Key::C, "`"), // `code`
        (Key::I, "/"), // /italics/
        (Key::L, "$"), // $lowered$
        (Key::R, "^"), // ^raised^
        (Key::S, "~"), // ~strikethrough~
        (Key::U, "_"), // _underline_
```

cat-state: Haha, I just realised and commited this at the same time

egui/src/input_state.rs
emilk: Note also the docstring with `///`

egui_demo_lib/src/easy_mark/easy_mark_editor.rs
```diff
@@ -117,59 +117,30 @@ impl EasyMarkEditor {
...
-                        toggle_surrounding(code, ccursor_range, "_");
-                        any_change = true;
-                    }
-                    _ => {}
-                }
-            }
-        }
+            egui::Modifiers {
+                command: true,
+                ..Default::default()
```
emilk, line 133: We can make this even nicer by switching the argument order and adding some helpers to `Modifiers`:

```rust
            if ui.input_mut().consume_key(egui::Modifiers::command(), key) {
```

emilk (changes requested):

egui_demo_lib/src/easy_mark/easy_mark_editor.rs
cat-state: nice, i added it was a builder api but lmk if you prefer classmethods + operator overloading instead

egui_demo_lib/src/easy_mark/easy_mark_editor.rs
```diff
@@ -117,59 +117,27 @@ impl EasyMarkEditor {
...
-                    // toggle _underline_
-                    Key::U => {
-                        toggle_surrounding(code, ccursor_range, "_");
-                        any_change = true;
-                    }
-                    _ => {}
-                }
-            }
-        }
+            toggle_surrounding(code, ccursor_range, surrounding);
```
emilk, line 133: According to the docs of `consume_key`, it will return true both when the key is pressed and when it is released, so this should toggle twice, which is not intentional. This is why I think `consume_key` should only detect presses and ignore releases (or the other way around)

cat-state: It only returns true when the key was pressed but it will remove the release event too.

emilk: I haven't run in since last review :)

But then the docs for `consume_key` should be changed to indicate it is only triggered on press, and also match for `pressed: true`.

emilk (changes requested):

Issue #1218, opened 2022-02-06
Individual frame margins


lampsitter:
Similarly to  https://github.com/emilk/egui/pull/1206, it can be useful to have individual margins for all four edges of a frame.
Personally I need this to be able to make blender like viewports.
Happy to implement it myself as usual!

PR #1219, opened 2022-02-06
Individual margins

lampsitter:
Closes https://github.com/emilk/egui/issues/1218

### DIFF
--- a/egui/src/containers/frame.rs
+++ b/egui/src/containers/frame.rs
@@ -3,12 +3,49 @@
 use crate::{layers::ShapeIdx, *};
 use epaint::*;
 
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub struct Margin {
+    pub left: f32,
+    pub right: f32,
+    pub top: f32,
+    pub bottom: f32,
+}
+
+impl Margin {
+    /// Margins with the same size on opposing sides
+    #[inline]
+    pub fn symmetric(x: f32, y: f32) -> Self {
+        Self {
+            left: x,
+            right: x,
+            top: y,
+            bottom: y,
+        }
+    }
+
+    /// Total margin in x direction
+    pub fn x(&self) -> f32 {
+        self.left + self.right
+    }
+
+    /// Total margin in y direction
+    pub fn y(&self) -> f32 {
+        self.top + self.bottom
+    }
+}
+
+impl From<Vec2> for Margin {
+    fn from(v: Vec2) -> Self {
+        Self::symmetric(v.x, v.y)
+    }
+}
+
 /// Color and margin of a rectangular background of a [`Ui`].
 #[derive(Clone, Copy, Debug, Default, PartialEq)]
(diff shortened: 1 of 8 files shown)


egui/src/containers/frame.rs
```diff
@@ -23,7 +60,7 @@ impl Frame {
     /// For when you want to group a few widgets together within a frame.
     pub fn group(style: &Style) -> Self {
         Self {
-            margin: Vec2::splat(6.0), // symmetric looks best in corners when nesting
+            margin: Vec2::splat(6.0).into(), // symmetric looks best in corners when nesting
```
emilk, line 63: Perhaps we could have a constructor for this too:

```suggestion
            margin: Margin::same(6.0), // symmetric looks best in corners when nesting
```

egui/src/containers/frame.rs
```diff
@@ -3,12 +3,49 @@
...
+        }
+    }
+
+    /// Total margin in x direction
+    pub fn x(&self) -> f32 {
+        self.left + self.right
+    }
+
+    /// Total margin in y direction
+    pub fn y(&self) -> f32 {
```
emilk, line 32: If I read `margin.y()` I wouldn't guess it is the sum of the top and bottom margins. Perhaps instead we could do `fn sum(&self) -> Vec2` so users could do `margin.sum().y` for the same effect?

egui/src/containers/frame.rs
```diff
@@ -52,7 +89,7 @@ impl Frame {
 
     pub fn window(style: &Style) -> Self {
         Self {
-            margin: style.spacing.window_padding,
+            margin: style.spacing.window_padding.into(),
```
emilk, line 92: `window_padding` should perhaps be renamed `window_margin` and be made into a `Margin` too?

egui/src/containers/frame.rs
```diff
@@ -62,7 +99,7 @@ impl Frame {
 
     pub fn menu(style: &Style) -> Self {
         Self {
-            margin: Vec2::splat(1.0),
+            margin: Vec2::splat(1.0).into(),
```
emilk, line 102:
```suggestion
            margin: Margin::same(1.0),
```

egui/src/containers/frame.rs
```diff
@@ -83,7 +120,7 @@ impl Frame {
     /// dark canvas to draw on
     pub fn dark_canvas(style: &Style) -> Self {
         Self {
-            margin: Vec2::new(10.0, 10.0),
+            margin: Vec2::new(10.0, 10.0).into(),
```
emilk, line 123: I think it is better to be explicit in this file
```suggestion
            margin: Margin:symmetric(10.0, 10.0),
```

emilk (changes requested):
Looks good, but left some comments!

emilk (approved):

emilk:
Thanks!

PR #1225, opened 2022-02-07
Add step parameter to egui::Slider

Gordon01:
PR adds `step_by()` and `disable_step()` methods to egui::Slider.
It allows adjusting the value of the slider with discrete steps, which is useful when we want to control something like the number of bits in a buffer and similar.

The demo app is updated: added one more slider, which controls the step value of another.

### DIFF
--- a/egui/src/widgets/slider.rs
+++ b/egui/src/widgets/slider.rs
@@ -71,6 +71,7 @@ pub struct Slider<'a> {
     suffix: String,
     text: String,
     text_color: Option<Color32>,
+    step: Option<f64>,
     min_decimals: usize,
     max_decimals: Option<usize>,
 }
@@ -113,6 +114,7 @@ impl<'a> Slider<'a> {
             suffix: Default::default(),
             text: Default::default(),
             text_color: None,
+            step: None,
             min_decimals: 0,
             max_decimals: None,
         }
@@ -199,6 +201,28 @@ impl<'a> Slider<'a> {
         self
     }
 
+    /// Sets the minimal change of the value.
+    /// If `clamp_to_range` is enabled, `step` must be less than a maximum between the start and
+    /// the end of the range; otherwise, the step would be left unchanged.
+    /// Without `clamp_to_range` enabled, `step` can be any value.
+    pub fn step_by(mut self, step: f64) -> Self {
+        if self.clamp_to_range {
+            let max = self.range.end().abs().max(self.range.start().abs());
+            if step.abs() <= max {
+                self.step = Some(step);
+            }
+        } else {
+            self.step = Some(step);
+        }
+        self
+    }
+
+    /// Disable the `step_by` feature
+    pub fn disable_step(mut self) -> Self {
+        self.step = None;
+        self
+    }
+
     // TODO: we should also have a "min precision".
     /// Set a minimum number of decimals to display.
     /// Normally you don't need to pick a precision, as the slider will intelligently pick a precision for you.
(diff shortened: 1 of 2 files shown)


egui/src/widgets/slider.rs
```diff
@@ -247,13 +272,20 @@ impl<'a> Slider<'a> {
...
-        }
-        if let Some(max_decimals) = self.max_decimals {
-            value = emath::round_to_decimals(value, max_decimals);
+        if let Some(step) = self.step {
+            let remainer = value % step;
+            if remainer != 0.0 {
+                value -= remainer;
+            };
+        } else {
+            if self.clamp_to_range {
```
emilk, line 281: We should be able to both clamp to range and have stepping (as long as the step size is smaller than the range)

egui/src/widgets/slider.rs
```diff
@@ -199,6 +202,28 @@ impl<'a> Slider<'a> {
...
+        let step = match self.clamp_to_range {
+            true if (step.abs() > max) => None,
+            _ => Some(step),
+        };
+        self.step = step;
+        self
+    }
+
+    /// Disable the `step_by` feature
+    pub fn disable_step(mut self) -> Self {
```
emilk, line 222: No need for this - users should just not call `step_by` if they don't want a step.

egui/src/widgets/slider.rs
```diff
@@ -199,6 +202,28 @@ impl<'a> Slider<'a> {
...
 
+    /// Sets the minimal change of the value.
+    /// If `clamp_to_range` is enabled, `step` must be less than a maximum between the start and
+    /// the end of the range; otherwise, the step would be left unchanged.
+    /// Without `clamp_to_range` enabled, `step` can be any value.
+    ///
+    /// Default: `None`.
+    pub fn step_by(mut self, step: f64) -> Self {
+        let max = self.range.end().abs().max(self.range.start().abs());
+        let step = match self.clamp_to_range {
```
emilk, line 213: This means the order of calling `clamp_to_range` and `step_by` is important - better to just set the step here and resolve this in `set_value` instead imho.

emilk (changes requested):
Good additions, buts needs some polish!

egui_demo_lib/src/apps/demo/widget_gallery.rs
```diff
@@ -166,8 +169,20 @@ impl WidgetGallery {
...
-        ui.add(egui::Slider::new(scalar, 0.0..=360.0).suffix("°"));
+        if *step != 0.0 {
+            ui.add(
+                egui::Slider::new(scalar, 0.0..=360.0)
+                    .suffix("°")
+                    .step_by(*step),
+            );
+        } else {
+            ui.add(egui::Slider::new(scalar, 0.0..=360.0).suffix("°"));
+        }
```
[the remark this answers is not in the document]
Gordon01: Thanks, didn't know about this page :)

egui/src/widgets/slider.rs
Gordon01: I've moved the "stepping code" to the bottom and now it even works better :)
Looks like clamping does not interfere with steps, and because filtering is performed after everything, we can't get a value outside step interval.
In the worst case (clamping and step value out of range), you can't change a value from zero.

egui/src/widgets/slider.rs
Gordon01: Removed.
But one can call steb_by(0.0) and this disables the feature. Turned out, it's convenient and I've used it in demo.

egui/src/widgets/slider.rs
Gordon01: I've removed all clamping logic from here. Turned out, it's overengineering.
Also, I think that the implicit turn-off of the stepping is worse than being stuck at 0.0.

egui/src/widgets/slider.rs
```diff
@@ -255,6 +273,12 @@ impl<'a> Slider<'a> {
         if let Some(max_decimals) = self.max_decimals {
             value = emath::round_to_decimals(value, max_decimals);
         }
+        if let Some(step) = self.step {
+            let remainer = value % step;
+            if remainer != 0.0 {
+                value -= remainer;
+            };
```
emilk, line 280: I believe this will have better numerical properties, and will also round better when clicking:

```suggestion
            value = (value / step).round() * step;
```

emilk (changes requested):

PR #1230, opened 2022-02-09
feat: don't install console panic hook and tracing wasm subscriber by default

### LINKED ISSUE #1229: Don't install console panic hook and tracing subscriber by default (in egui_web)
Libraries should not install these by default, it should be up to the user. Users may want to use other hooks or other tracing subscribers, instead of these. This has broken my app because off trying to set the logger / panic hook twice.

Don't install them by default, instead leave it up to the user https://github.com/emilk/egui/blob/c6ac1827f6eb9ffe0e08d12ce9fbf6a7e65105fc/egui_web/src/backend.rs#L284-L288

None. Feature flags are a possibility, but they add complexity for no reason, when the user can just add the dependencies to their application.

**Additional context**
None.

yusdacra:
This resolves #1229.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -558,16 +558,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "console_error_panic_hook"
-version = "0.1.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a06aeb73f470f66dcdbf7223caeebb85984942f22f1adb2a088cf9668146bbbc"
-dependencies = [
- "cfg-if 1.0.0",
- "wasm-bindgen",
-]
-
-[[package]]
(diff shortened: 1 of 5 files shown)


emilk:
This makes sense, but please move them to `egui_demo_app` so that they are A) present in the demo app and B) act like a good example for other to follow.

yusdacra:
> This makes sense, but I want them to be part of `egui_demo_app` so that they are A) present in the demo app and B) act like a good example for other to follow.

Ah yeah, i totally forgot about that. Added it there now.

> Would it perhaps make sense to put them in `eframe::start_web` instead?

That still has the issues mentioned in the issue. I think a better way is to just include it in the `eframe` template, which would also make it more clear about where the log stuff comes from.

PR #1240, opened 2022-02-11
Make non interactable layers not interact

### LINKED ISSUE #1239: Tooltips flicker when they can't fit on the screen
**Describe the bug**

Tooltips flicker when they can't fit on the screen

**To Reproduce**
1. Resize screen to make it small-ish
2. Hover on a tooltip that's large enough to not fit vertically
3. observe the flickering

**Expected behavior**
The tooltip should not flicker

**Screenshots**

https://user-images.githubusercontent.com/1410520/153644827-71046616-ff87-4724-96fa-f104e9de4d72.mp4

 - OS: Linux

**Additional context**

juancampa:
Areas in the Tooltip layer shouldn't block interactions to layers below them. Currently, layers under a tooltip won't get hovered if there's a tooltip covering it. Causing the flickering seen in #1239 

Closes #1239 - Tooltips flicker when they can't fit on the screen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,12 +44,15 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui_w
 * Replaced Frame's `margin: Vec2` with `margin: Margin`, allowing for different margins on opposing sides ([#1219](https://github.com/emilk/egui/pull/1219)).
 * `Plot::highlight` now takes a `bool` argument ([#1159](https://github.com/emilk/egui/pull/1159)).
 * `ScrollArea::show` now returns a `ScrollAreaOutput`, so you might need to add `.inner` after the call to it ([#1166](https://github.com/emilk/egui/pull/1166)).
+* Tooltips that don't fit the window don't flicker anymore ([#1240](https://github.com/emilk/egui/pull/1240)).
+* `Areas::layer_id_at` ignores non interatable layers (i.e. Tooltips) ([#1240](https://github.com/emilk/egui/pull/1240)).
 
 ### Fixed 🐛
 * Context menus now respects the theme ([#1043](https://github.com/emilk/egui/pull/1043)).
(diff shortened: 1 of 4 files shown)


juancampa:
This currently breaks context menus as they become non-interactable. Do we need a separate layer for those?

juancampa:
I'm not sure why the `Area` used for menus was marked as `.interactable(false)` but flipping that boolean fixed the problem. It makes sense the menus are interactable anyway, right?

emilk:
> I'm not sure why the `Area` used for menus was marked as `.interactable(false)` but flipping that boolean fixed the problem. It makes sense the menus are interactable anyway, right?

I don't know why it was written like that, and it is likely unintentional. However, there is something else wrong, because your PR breaks the `File` menu in `egui_demo_app`

juancampa:
> because your PR breaks the `File` menu in `egui_demo_app`

Hmm. Yeah, I see that, it's interesting that it works fine on the web version but not the native one. I'm investigating

juancampa:
@emilk so I finally figured out what's wrong with that menu. It's a stale `app.ron` file!

The `Areas::areas` `IdMap` is probably persisted on your local machine and the `State` for that particular menu has `interactable` set to `false`.

Deleting that file fixes the issue. Do we have any mechanism to either migrate old save states or delete them if they are incompatible?

emilk:
@juancampa nice debugging! I think the proper fix is to reset `interactable` in `Area::begin` in each frame:

```rust

        let mut state = state.unwrap_or_else(|| State {
            pos: default_pos.unwrap_or_else(|| automatic_area_position(ctx)),
            size: Vec2::ZERO,
            interactable,
        });
        state.pos = new_pos.unwrap_or(state.pos);
        state.interactable = interactable; // <--- ADDED
```

juancampa:
> the proper fix is to reset `interactable` in `Area::begin` in each frame

I had the same thought last night. Fixed!

emilk (approved):
Nice fix, thanks!

PR #1241, opened 2022-02-12
Make v-align and scale of user fonts tweakable

emilk:
Some fonts sit higher on a line, and some lower.

This PR allows users to tweak the vertical alignment and scale of each font to match it better with other fonts and the rest of the UI, like so:

```rust
font_data.insert(
    "emoji-icon-font".to_owned(),
    FontData::from_static(include_bytes!("../../fonts/emoji-icon-font.ttf")).tweak(
        FontTweak {
            scale: 0.8, // make it smaller
            y_offset_factor: 0.1, // move it down slightly
            y_offset: 0.0,
        },
    ),
);
```

--- 

This PR also changes those values slightly for the default fonts.

This moves some text by a pixel or so, especially small and large text, but the difference i minimal:

### Before:
fonts-master

### After:
fonts-tweaked

After plus text bounding boxes (text should be centered vertically in each green box):
fonts-tweaked-bb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui_w
   * Easily change text styles with `Style::text_styles`.
   * Added `Ui::text_style_height`.
   * Added `TextStyle::resolve`.
+* Made v-align and scale of user fonts tweakable ([#1241](https://github.com/emilk/egui/pull/1027)).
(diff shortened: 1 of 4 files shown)


juancampa:
Not sure if just me. But in the "after" screenshot, the top-right heading "egui demos", seems to be less vertically centered than before. Higher, to be specific.

emilk:
> Not sure if just me. But in the "after" screenshot, the top-right heading "egui demos", seems to be less vertically centered than before. Higher, to be specific.

That's because of how the text is placed within the panel, not how the text is centered within the label. I've added another screenshot with the text boxes drawn out - text should now be better centered vertically in those green boxes than before, including the "egui demos" bit.

I will need to tweak the margins for the side panels to make "egui demos" center better.

PR #1242, opened 2022-02-12
Add `move_to_top` and `top_most_layer`

Friz64:
In my application, i want to be able to get and set which window is on the top. This adds a public API for that. Is this the right approach?


egui/src/context.rs
```diff
@@ -889,6 +889,16 @@ impl Context {
         self.memory().layer_id_at(pos, resize_grab_radius_side)
     }
 
+    /// Top-most layer.
```
Mingun, line 892: I would like to see an explanation, what is the top-most layer instead of this useless doc. The doc should answer to question when I should use it? What is its features?

egui/src/memory.rs
```diff
@@ -329,6 +329,11 @@ impl Memory {
         self.areas.layer_id_at(pos, resize_interact_radius_side)
     }
 
+    /// Top-most layer.
```
Mingun, line 332: Same here

egui/src/context.rs
Friz64: What about this?

```
Returns the id of the top-most area. When you click on or interact with an area, it will be moved above all other areas.
```

My personal use case is synchronizing the selected item in a list of open windows with the window that last got focus. This is a very specific use case, but it wasn't possible to do before.

Friz64: I pushed a commit changing it to

> The overall top-most layer. When an area is clicked on or interacted with, it is moved above all other areas.

PR #1243, opened 2022-02-12
Add some macOS emacs keybindings

terhechte:
macOS has support for a subset of Emacs keybindings in all input fields. [Here's an overview of these bindings / shortcuts](https://jblevins.org/log/kbd). A input field that doesn't support at least a subset of this subset feels alien as it doesn't behave as expected. This PR implements some of those bindings:

- Move cursor left: `^B`
- Move cursor right: `^F`
- Beginning of line: `^A`
- End of line: `^E`
- Line up: `^P`
- Line down: `^N`

It doesn't implement move-by-word, selection or deletion, but I thought the above would be a good start.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui_w
 ## Unreleased
 
 ### Added ⭐
-* Support a subset of macOS' emacs input field keybindings in `TextEdit`.
+* Support a subset of macOS' emacs input field keybindings in `TextEdit` ([#1243](https://github.com/emilk/egui/pull/1243).
 * Much improved font selection ([#1154](https://github.com/emilk/egui/pull/1154)):
   * You can now select any font size and family using `RichText::size` amd `RichText::family` and the new `FontId`.
   * Easily change text styles with `Style::text_styles`.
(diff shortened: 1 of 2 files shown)


egui/src/widgets/text_edit/builder.rs
```diff
@@ -1091,11 +1091,31 @@ fn on_key_press(
             None
         }
 
+        Key::P | Key::N | Key::B | Key::F | Key::A | Key::E
+            if cfg!(target_os = "macos") && modifiers.ctrl && !modifiers.shift =>
```
emilk, line 1095: Note that `cfg!(target_os = "macos")` won't work when compiling to `wasm` (i.e. for web).

Perhaps we should add some platform-detection in `egui_web` so we can add `if ctx.options().is_mac` or similar?

Having it only work on native for now might be OK though.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui_w
 ## Unreleased
 
 ### Added ⭐
+* Support a subset of macOS' emacs input field keybindings in `TextEdit`.
```
emilk, line 11: Please add a link to the PR

emilk (changes requested):
I had no idea these key bindings existed on Mac :) Now I know

egui/src/widgets/text_edit/builder.rs
terhechte: I'll take a note and might want to also implement this for web at some point, but for now I think native is ok.

CHANGELOG.md
terhechte: Added :)

PR #1247, opened 2022-02-13
Allow scroll into view without specifying an alignment

juancampa:
# Problem

`response.scroll_to_me` works great but can be jarring if the UI is already in view.

# Solution

Accept an `Option<Align>` instead of just `Align`. If `None` is passed, the minimum scrolling necessary to bring the ui into view will be applied.

:movie_camera: 
https://user-images.githubusercontent.com/1410520/153758767-80630356-756e-49fd-8dd9-1b4743ac3149.mp4

### DIFF
--- a/egui/src/containers/scroll_area.rs
+++ b/egui/src/containers/scroll_area.rs
@@ -489,11 +489,24 @@ impl Prepared {
                 // We take the scroll target so only this ScrollArea will use it:
                 let scroll_target = content_ui.ctx().frame_state().scroll_target[d].take();
                 if let Some((scroll, align)) = scroll_target {
-                    let center_factor = align.to_factor();
-
                     let min = content_ui.min_rect().min[d];
-                    let visible_range = min..=min + content_ui.clip_rect().size()[d];
-                    let offset = scroll - lerp(visible_range, center_factor);
+                    let clip_rect = content_ui.clip_rect();
+                    let visible_range = min..=min + clip_rect.size()[d];
+
+                    let center_factor = if let Some(align) = align {
+                        align.to_factor()
+                    } else {
+                        if *scroll.start() < clip_rect.min[d] {
+                            0.0
+                        } else if *scroll.end() > clip_rect.max[d] {
+                            1.0
+                        } else {
+                            // Ui os already in view, no need to adjust scroll offset.
+                            continue;
+                        }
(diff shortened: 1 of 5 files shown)


egui/src/frame_state.rs
```diff
@@ -40,7 +42,7 @@ impl Default for FrameState {
             used_by_panels: Rect::NAN,
             tooltip_rect: None,
             scroll_delta: Vec2::ZERO,
-            scroll_target: [None; 2],
+            scroll_target: [None, None],
```
juancampa, line 45: Had to change this because `RangeInclusive` is not `Copy`.

egui/src/ui.rs
```diff
@@ -901,15 +901,16 @@ impl Ui {
...
     ///
     ///     if scroll_bottom {
-    ///         ui.scroll_to_cursor(Align::BOTTOM);
+    ///         ui.scroll_to_cursor(Some(Align::BOTTOM));
     ///     }
     /// });
     /// # });
     /// ```
-    pub fn scroll_to_cursor(&mut self, align: Align) {
+    pub fn scroll_to_cursor(&mut self, align: Option<Align>) {
```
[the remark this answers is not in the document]
juancampa: Added docs!

egui/src/containers/scroll_area.rs
```diff
@@ -493,14 +493,17 @@ impl Prepared {
                     let clip_rect = content_ui.clip_rect();
                     let visible_range = min..=min + clip_rect.size()[d];
 
+                    // if the ui is too big to completely fit in the scroll view, align it to the left/top
```
emilk, line 496: Is this the best behavior?

Another alternative is to scroll the minimum amount that makes the scroll view fill with the ui-range. So for instance, if we want to scroll to an image, and that image is very big then we either:

* scroll so that the top of the image is at the top of the scroll view
* scroll so that the bottom of the image is at the bottom of the scroll view (if that is closer)
* don't scroll at all, if all we are currently seeing is the image

juancampa: Yeah, I like the simmetry. I fixed it to have that behavior :point_up:

emilk (approved):

emilk:
Thanks!

Issue #165, opened 2021-02-07
Let scrolling follow the text cursor in a multiline text edit widget


the-drunk-coder:
**Is your feature request related to a problem? Please describe.**
Problem: When editing longer text snippets in a multi-line text edit widget, the cursor just leaves the scroll area, and you have to follow it manually with the mouse.

**Describe the solution you'd like**
Have the scrolling follow the cursor automatically, without having to switch to the mouse.

I've made some experiments with the customized text edit widget I keep in a fork, but when trying to scroll to a certain position the results are somewhat unpredictable ...

emilk:
This might be pretty easy to implement by following the bread-crumbs of the implementation of https://docs.rs/egui/0.9.0/egui/struct.Response.html#method.scroll_to_me and https://docs.rs/egui/0.9.0/egui/struct.Ui.html#method.scroll_to_cursor

the-drunk-coder:
Hmm I tried several variants of that but the results were a little unpredictable ... like, scrolling to a certain y-position never really scrolled to where I needed it ...

But I guess I should give it another look ...

UE2020:
Is there a solution for this yet?

the-drunk-coder:
Haven't worked on this either for a while ...

One problem I've frequently run into is that it seems hard to apply the abovementioned `scroll_to_cursor` and `scroll_to_me` methods selectively when the cursor leaves the visible part of the text field.

I think one thing that would be helpful is to have some method to detect whether the cursor moved out of the visible part of the text field, and in which direction ...

robo-corg:
Is there a way to use to use this with `ScrollArea::show_rows`? Trying to get this working with that but I am not certain what a rect for this would be if the location is not on screen.

Looking at things in the debugger and it seems the clip area never really changes and seems to be the screen coordinates of the `ScrollArea`.

robo-corg:
Ahh I think I figured it out. You need to calculate the position relative to row_range passed in.

PR #1252, opened 2022-02-15
Scroll so that text cursor remains visible

emilk:
Closes https://github.com/emilk/egui/issues/165

This was very easy to implement thanks to https://github.com/emilk/egui/pull/1247


emilk:
Good catch - fixed!

PR #1260, opened 2022-02-17
Add more source code links to demos

urholaukkarinen:
I was looking for an example on how to build context menus and noticed that some demos were missing the `(source code)` link.

Added the link to the following demos:
* Code Example
* Context Menus
* Font Book
* Bezier Curve


emilk (approved):
Thanks!

PR #1276, opened 2022-02-20
Added `scroll_step` for slider

ROCKFAL1:
- Added `scroll_step` for slider. This allows you to change the slider value by the value of `scroll_step` by scrolling the mouse wheel
- Added an example of using this feature.
- Also now, if negative values are passed to 'step_by' or 'scroll_step`, this will mean disabling the function

This feature is strongly related to `step_by`, so I added a little logic in case the user wants to use both at the same time. If `step_by` and `scroll_step` are used, then `scroll_step` = `step`. I'm not sure if this is the right solution, so you can offer your ideas.

### DIFF
--- a/egui/src/widgets/slider.rs
+++ b/egui/src/widgets/slider.rs
@@ -73,6 +73,7 @@ pub struct Slider<'a> {
     text_color: Option<Color32>,
     /// Sets the minimal step of the widget value
     step: Option<f64>,
+    scroll_step: Option<f64>,
     min_decimals: usize,
     max_decimals: Option<usize>,
 }
@@ -116,6 +117,7 @@ impl<'a> Slider<'a> {
             text: Default::default(),
             text_color: None,
             step: None,
+            scroll_step: None,
             min_decimals: 0,
             max_decimals: None,
         }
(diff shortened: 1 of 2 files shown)


egui/src/widgets/slider.rs
```diff
@@ -324,9 +355,20 @@ impl<'a> Slider<'a> {
...
             self.set_value(new_value);
         }
 
+        if response.hover_pos().is_some() {
+            let prev_value = self.get_value();
+            let input = ui.input();
+            let delta = (input.scroll_delta.x + input.scroll_delta.y) as f64;
+            if let Some(scroll_step) = self.scroll_step {
+                // Scroll speed decided by consensus: https://github.com/emilk/egui/issues/461
+                self.set_value(prev_value + (delta / 50_f64) * scroll_step);
```
emilk, line 368: This isn't very robust considering how many backends egui has.

If we want to handle discreet scroll steps (like those from a mouse), then we should consider adding those as separate events in egui. For instance, scrolling on one step on a mouse could produce `scroll_delta.y ≈ 50.0` and `scroll_detla_steps.y == 1`. Scrolling with a touch-pad would produce a smooth `scroll_delta.y` but no `scroll_detla_steps`.

emilk:
What is the use case being solved here? Using the scroll wheel to move a slider is nice, but does it require a separate `scroll_step` setting? Alternatives would include:

* Moving the slider as many points as the user scrolled (would work great for smooth scroll inputs such as touch-pads)
* Moving the slider by the `step_by` size for each discreet scroll tick on discreet scrollers (i.e. mouse wheels)

When would you want `step_by` and `scroll_step` to be different?

egui/src/widgets/slider.rs
ROCKFAL1: I agree, i think it's worth just ignoring input from the touchpad or similar devices, but to make such a change, you will also have to make changes to the backends.

ROCKFAL1:
> When would you want `step_by` and `scroll_step` to be different?

I don't think there are situations where it makes sense to use them together, but I did it from the point of view that an egui user may mistakenly use both of these functions.

PR #1291, opened 2022-02-22
[epaint] Add more text wrapping options

awaken1ng:
Add an ability to limit how many lines the text should have, 
control breaking behavior a little bit, 
and insert arbitrary character, e.g. `…`, to represent clipped text.

Maybe `TextWrapping::overflow_character` should be renamed into something more descriptive, like `overflow_representation`?

https://user-images.githubusercontent.com/20689137/155110673-0dacfab6-cb26-40fb-bf42-14c6124b86f1.mp4


awaken1ng:
> Very good. I want to use it. Can you add wrapping by char (no by word) for small textareas?

For `TextEdit`s? Or you mean a single line `Label`?

If you mean the latter, `TextWrapping::break_anywhere` should do exactly that, if it's set to `true` then wrapping will break at a character, instead of trying to break at a word, you can see it's behaviour in the demo video.

juancampa:
Pulled this into my fork and it works like a charm, at least in my use case. Thank you!

egui_demo_lib/src/apps/demo/misc_demo_window.rs
```diff
@@ -1,12 +1,17 @@
...
 
 /// Showcase some ui code
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 #[cfg_attr(feature = "serde", serde(default))]
 pub struct MiscDemoWindow {
     num_columns: usize,
 
+    break_anywhere: bool,
+    max_lines: usize,
+    overflow_character: String,
```
emilk, line 13:
```suggestion
    overflow_character: char,
```

will make the rest of the code simpler

epaint/src/text/text_layout_types.rs
```diff
@@ -257,6 +257,54 @@ impl TextFormat {
...
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub struct TextWrapping {
+    /// Try to break text so that no row is wider than this.
+    /// Set to [`f32::INFINITY`] to turn off wrapping.
+    /// Note that `\n` always produces a new line.
+    pub max_width: f32,
+
+    /// Maximum amount of rows the text should have.
+    /// Set to `0` to disable this.
+    pub max_lines: usize,
```
emilk, line 270: I think `usize::MAX` is a better way to disable it; then there is no need for any special cases.

epaint/src/text/text_layout_types.rs
```diff
@@ -257,6 +257,54 @@ impl TextFormat {
...
+    }
+}
+
+impl Default for TextWrapping {
+    fn default() -> Self {
+        Self {
+            max_width: f32::INFINITY,
+            max_lines: 0,
+            break_anywhere: false,
+            overflow_character: None,
```
emilk, line 301: Should we perhaps make `Some('…')` the default?

epaint/src/text/text_layout_types.rs
emilk, line 270:
```suggestion
    pub max_rows: usize,
```

so as not to confuse it with a paragraph (that which is ended by the newline character `'\n'`)

epaint/src/text/text_layout.rs
```diff
@@ -170,18 +180,28 @@ fn rows_from_paragraphs(paragraphs: Vec<Paragraph>, wrap_width: f32) -> Vec<Row>
...
     for (i, glyph) in paragraph.glyphs.iter().enumerate() {
         let potential_row_width = glyph.max_x() - row_start_x;
 
-        if potential_row_width > wrap_width {
+        if job.wrap.max_lines > 0 && non_empty_rows >= job.wrap.max_lines {
+            break;
+        }
+
+        if potential_row_width > job.wrap.max_width {
             if first_row_indentation > 0.0 && !row_break_candidates.has_word_boundary() {
```
emilk, line 205: `break_anywhere` should be taken into account here, e.g. with `row_break_candidates.has_good_candidate(job.wrap.break_anywhere)`

epaint/src/text/text_layout.rs
```diff
@@ -225,24 +247,80 @@ fn line_break(paragraph: &Paragraph, wrap_width: f32, out_rows: &mut Vec<Row>) {
...
+
+                    // check if we're still within width budget
+                    let row_end_x = last_glyph.max_x();
+                    let row_start_x = row.glyphs.first().unwrap().pos.x; // if `last_mut()` returned `Some`, then so will `first()`
+                    let row_width = row_end_x - row_start_x;
+                    if row_width <= job.wrap.max_width {
+                        break;
+                    }
+
+                    row.glyphs.pop();
```
emilk, line 301: It would be nice to extract the inside of this `loop` (or most of it) into a separate function

emilk (changes requested):
Left some comments, but looks good!

epaint/src/text/text_layout_types.rs
awaken1ng: Default was initially `usize::MAX`, but I then later realized that if it's set to `0`, then egui crashes in `rows_from_paragraphs` because we end up with no paragraphs, hence the check in `line_break` function.

epaint/src/text/text_layout_types.rs
awaken1ng: Hm, setting `None` is definitely a lot easier than either having an IME setup, or looking for somewhere to copy `…` from.

epaint/src/text/text_layout.rs
awaken1ng: Sorry, I'm being a dum-dum, as in, `potential_row_width > job.wrap.max_width && row_break_candidates.get(job.wrap.break_anywhere).is_some()`? That seemingly breaks the text breaks.

image

epaint/src/text/text_layout_types.rs
awaken1ng: After sleeping on it for a bit, what about `TextWrapping::ellipsis` constructor instead?

epaint/src/text/text_layout_types.rs
```diff
@@ -257,6 +257,54 @@ impl TextFormat {
...
+    }
+}
+
+impl Default for TextWrapping {
+    fn default() -> Self {
+        Self {
+            max_width: f32::INFINITY,
+            max_rows: 0,
+            break_anywhere: false,
+            overflow_character: None,
```
emilk, line 301: I'm suggesting this simple change:

```suggestion
            overflow_character: Some('…'),
```

Users can still pick whatever character they want, but I feel this is a safer default choice when the user sets `max_rows` but doesn't set `overflow_character`, i.e.

```rust
let wrapping = TextWrapping { max_rows: 1, ..Default::default() };
```

epaint/src/text/text_layout.rs
emilk: I meant we should add something like

``` rust
impl RowBreakCandidates {
    fn has_good_candidate(&self, break_anywhere: bool) -> bool {
        if break_anywhere { self.any.is_some() }
        else { self.has_word_boundary() }
    }
}
```

emilk (approved):

Issue #1281, opened 2022-02-21
Awkward API for TexturesDelta in integrations


derivator:
The API for integrations around `TexturesDelta` introduced in #1110 seems a bit awkward to me:  
`egui_winit::State::handle_output` [takes ownership of `egui::Output` and returns `egui::TexturesDelta`](https://github.com/emilk/egui/blob/713917e4819d9c26068aec286ee675f7e3bff907/egui-winit/src/lib.rs#L517), suggesting that egui_winit is somehow involved or needed to get the `TexturesDelta`, when that's not actually the case.  
Perhaps `TexturesDelta` could be moved out of `Output` and into a new struct containing exactly everything needed to render the gui (i.e. the `Vec<ClippedShape>` that egui returns at the moment).  
As an alternative, egui_winit could just not take ownership of `Output` and return nothing.

derivator:
`BackendOutput` sounds ambiguous to me. Isn't rendering also part of the backend? It could also be misunderstood to be output *from* the backend. I can't really think of a better name right now though, but here's some words that might do the trick in the right combination: OS, platform, integration, window manager.  

Maybe something like this?:

```rust
pub struct Output {
    pub for_platform: PlatformOutput, // current `Output`
    pub for_rendering: RenderingOutput, // shapes, texture delta
    pub needs_repaint: bool, // ambiguous items that could go into either category go in top level struct
}
```

emilk:
Take a look: https://github.com/emilk/egui/pull/1292

PR #1292, opened 2022-02-22
Introduce egui::FullOutput, returned from Context::run

emilk:
This adds a new struct:

```
pub struct FullOutput {
    pub output: Output,
    pub needs_repaint: bool,
    pub textures_delta: epaint::textures::TexturesDelta,
    pub shapes: Vec<epaint::ClippedShape>,
}
```

The name is not great, nor the name of the nested `Output`, but structurally it works rather well.

Closes https://github.com/emilk/egui/issues/1281

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,8 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui_w
 * Renamed `Ui::visible` to `Ui::is_visible`.
 * Split `Event::Text` into `Event::Text` and `Event::Paste` ([#1058](https://github.com/emilk/egui/pull/1058)).
 * For integrations:
-  * `FontImage` has been replaced by `TexturesDelta` (found in `Output`), describing what textures were loaded and freed each frame ([#1110](https://github.com/emilk/egui/pull/1110)).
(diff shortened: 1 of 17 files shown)


derivator:
I think I prefer the names I suggested: since this is a breaking change anyway, you might as well rename `Output`... `FullOutput` is a bit awkward.  
It's fine though, at least `handle_egui_output` does not return `TexturesDelta` anymore, I think that was the most confusing thing. :+1:

emilk:
If we rename `Output` we should probably also rename `ctx.output()` which makes it even more of a breaking change :/

derivator:
Ah, I see. I guess the use case there is e.g. for custom widgets to be able to open a URL or set the cursor? Seems like it's not something that most users would encounter often. egui itself only calls `output()` 15 times. Maybe deprecate it with a note to use the new name? That way calls of the form `ui.ctx().output()` would still work even though the name of the return type changed...

This makes me notice another issue though: custom widgets probably want to be able to set `needs_repaint`, right? Possibly even insert textures or shapes into the output? That seems like it has huge footgun potential though...

emilk:
I renamed it `PlatformOutput` but kept the name `ctx.output()` for now. I don't want to spend too much time on this (I want to release egui 0.17.0!)

derivator:
Looks good to me :+1: Thanks for all the work on egui, it really is a fantastic library :)

Issue #1289, opened 2022-02-22
Give users control over the texture filter


mccolljr:
**Is your feature request related to a problem? Please describe.**

I wanted to utilize the `Painter::set_texture_filter` feature that was recently introduced to `egui_glow`, but the only way to do so is by breaking out of `eframe` and calling into `egui_glow` directly. This... involves a lot of manual reconstruction of code that is difficult to get right, especially for someone newer to this type of programming

*Edit:* Per discussion below, it seems more accurate to say that I'd like to be able to specify the texture filter on a per-texture basis.

**Describe the solution you'd like**

Some sort of mechanism to pass backend-specific configuration tweaks via `eframe`. This could be as simple as adding a function alongside `run_native` for that each relevant backend, perhaps `run_native_with_options`? I don't want to prescribe a specific implementation here because I am so new to this library - I don't know what the long term goals are for `eframe`, or how this concept might interact with them.

**Describe alternatives you've considered**
1. Don't do this - just point people towards the examples and let them copy the code that they need. This is the current state of the world and it is functional, but not very ergonimic
2. Add `cfg`-gated options to `NativeOptions` in `epi`. This feels dirty to me
3. Approach this from a different angle and define `PainterOptions` in `epi`, which could provide a backend-agnostic way to set some global properties of the painter implemented by the backend. This seems like it could get complicated and there is a question of what to do if an option is provided that the backend in question can't actually handle.

emilk:
I think texture filtering in particular could be part of the new `epaint::TextureManager` API instead, though we need to be careful not to add too much complexity to what is supposed to be a simple 2D painting API (more advanced painting should always be done with something outside of `egui`, e.g. bevy)

mccolljr:
This seems like it would work, as well!

For context - I am working on some pixel drawing functionality in an app, and I would like to support canvas sizes up to 1024x1024 pixels. The current approach of rendering the image to a mesh where each pixel is a rectangle scaled to the chosen zoom level is not performant once the visible area approaches 512x512 logical "pixels". I have had better results rendering a single rectangle with the appropriate texture uv, but for this to look pixel-perfect I utilize the `NEAREST` filter. So for my purposes, a single global setting is sufficient.

Perhaps this could be provided as a `pub filter: SomeTextureFilterEnum` on `TextureDelta`?

mccolljr:
[...]

PR #1296, opened 2022-02-23
Allow specification of a TextureFilter per texture

mccolljr:
Closes <https://github.com/emilk/egui/issues/1289>.

### DIFF
--- a/egui_demo_lib/src/apps/demo/plot_demo.rs
+++ b/egui_demo_lib/src/apps/demo/plot_demo.rs
@@ -421,8 +421,11 @@ impl Widget for &mut ItemsDemo {
         };
 
         let texture: &egui::TextureHandle = self.texture.get_or_insert_with(|| {
-            ui.ctx()
-                .load_texture("plot_demo", egui::ColorImage::example())
+            ui.ctx().load_texture(
+                "plot_demo",
+                egui::ColorImage::example(),
+                egui::TextureFilter::default(),
+            )
         });
         let image = PlotImage::new(
             texture,
--- a/epaint/src/textures.rs
+++ b/epaint/src/textures.rs
@@ -3,6 +3,20 @@ use ahash::AHashMap;
 
 // ----------------------------------------------------------------------------
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub enum TextureFilter {
+    Linear,
+    Nearest,
+}
+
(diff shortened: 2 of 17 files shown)


egui_demo_lib/src/apps/demo/plot_demo.rs
```diff
@@ -421,8 +421,11 @@ impl Widget for &mut ItemsDemo {
         };
 
         let texture: &egui::TextureHandle = self.texture.get_or_insert_with(|| {
-            ui.ctx()
-                .load_texture("plot_demo", egui::ColorImage::example())
+            ui.ctx().load_texture(
+                "plot_demo",
+                egui::ColorImage::example(),
+                egui::TextureFilter::default(),
```
emilk, line 427:
```suggestion
                Default::default(),
```

this is shorter and more future proof (i.e. will work even if we change the name/type of this argument)

epaint/src/textures.rs
```diff
@@ -27,18 +41,29 @@ impl TextureManager {
...
+                bytes_per_pixel: image.bytes_per_pixel(),
+                retain_count: 1,
+            })
+            .filter;
+
+        self.delta.set.insert(
+            id,
+            TextureDelta {
+                filter,
+                image: ImageDelta::full(image),
```
emilk, line 64: did you consider making `filter` part of `ImageDelta` instead of adding a new `TextureDelta`? The new code is quite onion:y, and I don't think `ImageDelta` is ever used without being part of a `TextureDelta` anyway?

epaint/src/textures.rs
```diff
@@ -178,8 +220,8 @@ impl std::fmt::Debug for TexturesDelta {
         if !self.set.is_empty() {
             let mut string = String::new();
             for (tex_id, delta) in &self.set {
-                let size = delta.image.size();
-                if let Some(pos) = delta.pos {
+                let size = delta.image.image.size();
```
emilk, line 223: example of the onion-problem :)

mccolljr:
@emilk sorry, I have been busy at work recently but I will try to come back to this later this week/weekend

emilk:
No stress @mccolljr !

RainbowCookie32:
I was just looking for something like this, any chance to also get it exposed to eframe?

mccolljr:
I totally lost track of this. I got busy with other stuff. I'm happy to pick this back up - I have some staycation time early next week.

I think some additional changes have been made since I did this work so I may have to refactor a bit

emilk:
Someone beat you to it: https://github.com/emilk/egui/pull/1636

PR #1303, opened 2022-02-25
Fix egui_glow when targeting `wasm32-unknown-unknown`

Jake-Shadle:
This fixes compilation failures when targeting `wasm32-unknown-unknown` with eframe when the `egui_glow` or `persistence` feature flags are enabled. This adds a CI check to prevent this in the future.

See the [puffin action failure](https://github.com/EmbarkStudios/puffin/runs/5333830573?check_suite_focus=true)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -56,6 +56,20 @@ jobs:
           command: check
           args: -p egui_demo_app --lib --target wasm32-unknown-unknown
 
+  check_wasm_eframe_with_features:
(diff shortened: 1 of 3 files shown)


DusterTheFirst:
This compilation failure seems to only arise when using cargo resolver v1 in my experience. If this issue matches the one in #1307, merging this will close #1307

emilk (approved):

emilk:
Thanks!

PR #1304, opened 2022-02-25
Fix combo box misalignment on rtl layout

juancampa:
Combo boxes were hard-coded to be left aligned. Causing them to render in the wrong position within a RTL layout.

Before:
image

After:
image


emilk (approved):

emilk:
Thanks!

Issue #1290, opened 2022-02-22
eframe on the web (egui_web) creates a mess on panic


DusterTheFirst:
**Describe the bug**
When using eframe on the web (wasm), you call `eframe::start_web` to set eframe up. `eframe::start_web` installs many event callbacks into the DOM and a 1 second interval callback which are all used to render/update the application. This is fine and works perfectly when the application is in a valid state.

Because these callbacks call into rust code from the JS engine, when rust code panics from anything (in my example I use an explicit panic to show the behavior), the state of the rust code becomes undefined but the JS engine will continue to call into the wasm code, unaware that it has panicked. This leads to a spam of errors in the console which can be seen in the screencap below: `Uncaught RuntimeError: unreachable executed`.

This causes multiple problems:
1. Any panic message written to the browser's console like one from `console_error_panic_hook` would be hidden away by the error spam
2. The browser will continue to attempt to call the wasm code, leading to the continued `RuntimeError`s which will slow down a browser's devtools and lead to a very unfriendly user experience.
3. The browser can call callbacks with the application in an invalid state, which can be seen in the screencap as the second panic caused by a RefCell's lock not being dropped by the panicking code.

**To Reproduce**
1. Setup an application using eframe running in wasm
2. panic anywhere in your code after initializing eframe

**Expected behavior**
The application would panic, printing the message to the console and would stop responding to user events.

to achieve this eframe could provide a way to de-register the callbacks registered in `eframe::start_web` that can be called in a panic handler, or `eframe_web` could provide/set a panic handler to do this without the need for a programmer to call the function themselves.

**Screenshots**

https://user-images.githubusercontent.com/14093962/155087145-59c85a48-3b9f-49c3-8895-8683d1b60492.mp4

**Desktop (please complete the following information):**
 - OS: Windows 10 Enterprise (19044.1526)
 - Browser: Firefox 97.0.1 (64-bit)
 - egui/eframe Version 0.16.0

**Additional context**
The code for this can be found [here](https://github.com/DusterTheFirst/vocode/tree/d99a2b4eda918a977fc9f89477bea8dff06213ab/loid) and is built using trunk, but trunk should not affect the behavior of eframe.

DusterTheFirst:
You can run this app in your browser using the following link and see the problem yourself.
https://vocode.dusterthefirst.com/

emilk:
[...]

PR #1306, opened 2022-02-25
[egui_web] Prevent event handlers from running if code has panicked

DusterTheFirst:
The way egui_web works involves adding event handlers to a `<canvas>` element, the `<body>`, and a helper `<input>` element. By setting up such callbacks, it creates a position where the JavaScript/WASM runtime will call rust code across a FFI boundary, but that rust code could be in an invalid state, due to a previous panic. More information on the issue can be seen in #1209.

There are 2 ways to prevent this problem from occuring:
1. Provide a way to de-register all of the callbacks that were setup which the user could call in a panic handler, or `std::panic::catch_unwind`
2. Leave callbacks registered, but return early from all of them if the code has encountered a panic.

The first option would be preferable since it would allow for the user to have more control over the de-registering of the callbacks. This was my first attempt but it became hard to near impossible.
1. Lots of the types used in `egui` and `egui_web` are not [`UnwindSafe`](https://doc.rust-lang.org/stable/std/panic/trait.UnwindSafe.html) or [`RefUnwindSafe`](https://doc.rust-lang.org/stable/std/panic/trait.RefUnwindSafe.html) which make using `std::panic::catch_unwind` impossible.
2. None of the types from `wasm_bindgen`, `web_sys`, or `js_sys` are `Send`. This means that there would be no way for a panic handler (`std::panic::set_hook`) to be passed the `js_sys::Function` or even the `web_sys::EventTarget`. both of which are needed to call `remove_event_listener`.

Both of the above limitations mean that the event listeners can not be de-registered from inside of a panic handler. This leaves only option 2.

The way I have implemented this is using a shared `Arc<AtomicBool>` between all of the event handlers and the panic handler. Every time the event handlers are called, they perform a `AtomicBool::load(SeqCst)` and check that the bool is false. If it is, they continue calling the normal rust code. If the bool was true, the event handler returns early. The `AtomicBool` is set to true by the panic handler as soon as it is hit, so all callbacks after will be skipped.

In making this change, I created a [convenience method](https://github.com/emilk/egui/compare/master...DusterTheFirst:web-panic#diff-08ecc98d40fc709c6db305a8848c79dd3018a23007b6dfc3558257c3bab55d84R317-R351) that will wrap the callbacks in a check against the atomic bool as well as reduce boilerplate with closure creation, event listener adding, and closure forgetting.

There is only one event listener I did not touch, and that is the [contextmenu](https://github.com/emilk/egui/compare/master...DusterTheFirst:web-panic#diff-08ecc98d40fc709c6db305a8848c79dd3018a23007b6dfc3558257c3bab55d84R577-R586) event handler. Since this handler does not call into any egui code, it does not contribute to the problem and can be left alone.

The main culprit of the errors is the locking of the mutex guarding the `AppRunner`.

Here is a link to a few important changes:
- [The panic handler](https://github.com/emilk/egui/compare/master...DusterTheFirst:web-panic#diff-df803feb9e2073afc33fd76e71604ec5e6044094886b4b456bd49d05c01fe632R374-R392)
- [Add event listener convenience method](https://github.com/emilk/egui/compare/master...DusterTheFirst:web-panic#diff-08ecc98d40fc709c6db305a8848c79dd3018a23007b6dfc3558257c3bab55d84R317-R351)
- [A check against the atomic bool in `paint_and_schedule`](https://github.com/emilk/egui/compare/master...DusterTheFirst:web-panic#diff-08ecc98d40fc709c6db305a8848c79dd3018a23007b6dfc3558257c3bab55d84R381-R385)

And a before/after comparison:

## Before:

https://user-images.githubusercontent.com/14093962/155794187-cee13e4a-e0b8-4b50-9ee0-780c5852f168.mp4

## After:

https://user-images.githubusercontent.com/14093962/155794214-b8411fba-2fd1-45ec-9a46-a243c33649b4.mp4

As can be seen in the after demo, there is still one `unreachable executed` error, but this is caused by the builtin rust panic handler calling unsupported code so there is nothing I can do about it.

Closes <https://github.com/emilk/egui/issues/1290>.

### DIFF
--- a/egui/src/widgets/plot/mod.rs
+++ b/egui/src/widgets/plot/mod.rs
@@ -1,6 +1,6 @@
 //! Simple plotting library.
 
-use std::{cell::RefCell, ops::RangeInclusive, rc::Rc};
+use std::{cell::Cell, ops::RangeInclusive, rc::Rc};
 
 use crate::*;
 use epaint::ahash::AHashSet;
@@ -92,7 +92,7 @@ impl PlotMemory {
 pub struct LinkedAxisGroup {
     pub(crate) link_x: bool,
     pub(crate) link_y: bool,
-    pub(crate) bounds: Rc<RefCell<Option<PlotBounds>>>,
+    pub(crate) bounds: Rc<Cell<Option<PlotBounds>>>,
 }
 
 impl LinkedAxisGroup {
@@ -100,7 +100,7 @@ impl LinkedAxisGroup {
         Self {
             link_x,
             link_y,
-            bounds: Rc::new(RefCell::new(None)),
+            bounds: Rc::new(Cell::new(None)),
         }
     }
 
@@ -132,11 +132,11 @@ impl LinkedAxisGroup {
     }
 
     fn get(&self) -> Option<PlotBounds> {
-        *self.bounds.borrow()
+        self.bounds.get()
     }
 
     fn set(&self, bounds: PlotBounds) {
-        *self.bounds.borrow_mut() = Some(bounds);
+        self.bounds.set(Some(bounds));
     }
 }
 
--- a/egui_web/CHANGELOG.md
+++ b/egui_web/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to the `egui_web` integration will be noted in this file.
 
 
 ## Unreleased
+* egui code will no longer be called after panic ([#1306](https://github.com/emilk/egui/pull/1306))
 
 
 ## 0.17.0 - 2022-02-22
--- a/egui_web/src/backend.rs
+++ b/egui_web/src/backend.rs
@@ -359,11 +359,37 @@ pub fn start(canvas_id: &str, app: Box<dyn epi::App>) -> Result<AppRunnerRef, Js
 /// Install event listeners to register different input events
 /// and starts running the given `AppRunner`.
 fn start_runner(app_runner: AppRunner) -> Result<AppRunnerRef, JsValue> {
-    let runner_ref = AppRunnerRef(Arc::new(Mutex::new(app_runner)));
-    install_canvas_events(&runner_ref)?;
-    install_document_events(&runner_ref)?;
-    text_agent::install_text_agent(&runner_ref)?;
-    repaint_every_ms(&runner_ref, 1000)?; // just in case. TODO: make it a parameter
-    paint_and_schedule(runner_ref.clone())?;
-    Ok(runner_ref)
+    let runner_container = AppRunnerContainer {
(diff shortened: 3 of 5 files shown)


Titaniumtown (approved):
LGTM works fine on my end. Very useful imo!

Titaniumtown:
Actually. I found an issue:
image

Doesn't occur on the master branch.

DusterTheFirst:
> Actually. I found an issue: 

Can you show the code that causes this to occur?

DusterTheFirst:
@Titaniumtown can you show the code that caused this panic so I can look into it?

emilk:
I'll wait with merging this until @Titaniumtown gets back

Titaniumtown:
Oh I'll send a code snippet in a bit.

Titaniumtown:
Ok. So when interacting with buttons and an input text box like so: https://github.com/Titaniumtown/integral_site/blob/d48b94c1d2a43d0fab4ebba630d315f82cd5553b/src/egui_app.rs#L277 then https://github.com/Titaniumtown/integral_site/blob/d48b94c1d2a43d0fab4ebba630d315f82cd5553b/src/egui_app.rs#L298 and interacting with one, then another. I get that error. This error doesn't occur on the master branch. It seems to be related to the entire thing being wrapped in a `ui.horizontal` as all other widgets work perfectly. Very strange.

Titaniumtown:
It works now! (at least for me)

DusterTheFirst:
@Titaniumtown I think I have fixed your panic issue with the new commits I have pushed. Please try them out on your setup to make sure it works not just on my machine.

The root cause is some very weird wasm/js-engine tomfoolery that must be going on, causing an event handler to interrupt the `text_agent::update_text_agent` call in another event handler which has not dropped its mutex guard yet. It is weird that this did not happen before my changes as this "issue" should have still existed. Weird. You can see the fix here: <https://github.com/DusterTheFirst/egui/commit/ff6f875296f6dac327d55c588931d7ffc2a55f0b?diff=split#diff-b509ec85e00a7dfe010fec7c9542e8bec5b47f877239c25dfd6846a0123fa888R172-R183>

Note: the CI has started failing, but this seems to be caused by [a new advisory for the `regex` crate](https://github.com/rustsec/advisory-db/commit/81e4691173c1506e9b8e7d24f3b191eb27fc1058) being added to rustsec's database

emilk (approved):

emilk:
I've tested it on mobile, and mobile text input still works, so… looks good to me!

emilk:
I just noticed a bug caused by this PR: resizing the browser window will no longer immediately resize the egui canvas, making it feel VERY unresponsive.

PR #1308, opened 2022-02-25
Add method to set a Plot's margin

Titaniumtown:
This PR Implements the `set_margin_fraction` method which allows the `margin_fraction` private field in the Plot struct to be set. This is useful as it then allows for the margins to be changed easily.


egui/src/widgets/plot/mod.rs
```diff
@@ -287,6 +287,12 @@ impl Plot {
         self
     }
 
+    /// Sets the `margin_fraction` field used to determine the margin of the Plot
+    pub fn set_margin_fraction(mut self, margin_fraction: Vec2) -> Self {
```
emilk, line 291: Naming what private field is set is not a very useful docstring. Maybe:

```suggestion
    /// Set the side margin as a fraction of the plot size.
    ///
    /// For instance, a value of `0.1` will add 10% space on both sides.
    pub fn set_margin_fraction(mut self, margin_fraction: Vec2) -> Self {
```

emilk (changes requested):

egui/src/widgets/plot/mod.rs
Titaniumtown: fixed!

PR #1329, opened 2022-03-05
Remove warning about cpal drag and drop

DusterTheFirst:
Given that the issue https://github.com/rust-windowing/winit/issues/1255 was closed by https://github.com/rust-windowing/winit/pull/1524, it would make sense to remove the warning about the issue from NativeOptions.

I also propose changing the default for `NativeOptions::drag_and_drop_support` to `true`, since the issue with winit has been resolved. Because this may be a breaking change, I have separated it into its own commit so it can be reverted if the default should stay the same.


emilk:
thanks!

PR #1338, opened 2022-03-07
Fix multiple partial updates of the same texture

### LINKED ISSUE #1337: Calling `set_partial` on the same texture rapidly causes some changes to get lost
**Describe the bug**
When [`set_partial`](https://github.com/emilk/egui/blob/37c9f116bfb717243e70ecbc48f2d2280106a51d/epaint/src/texture_handle.rs#L75) is called on a [`TextureHandle`](https://github.com/emilk/egui/blob/37c9f116bfb717243e70ecbc48f2d2280106a51d/epaint/src/texture_handle.rs#L19), it calls [`set`](https://github.com/emilk/egui/blob/master/epaint/src/textures.rs#L47) on its associated [`TextureManager`](https://github.com/emilk/egui/blob/37c9f116bfb717243e70ecbc48f2d2280106a51d/epaint/src/textures.rs#L10). [`set` internally inserts the TextureDelta](https://github.com/emilk/egui/blob/37c9f116bfb717243e70ecbc48f2d2280106a51d/epaint/src/textures.rs#L53) in a [`AHashMap`](https://github.com/emilk/egui/blob/37c9f116bfb717243e70ecbc48f2d2280106a51d/epaint/src/textures.rs#L153), replacing a possibly unprocessed image delta.

This manifests in the form that only the last `set_partial` call in a frame actually has an effect on the texture, the previous calls are silently discarded. We have to patch multiple small areas in a big texture every frame. Therefore it would be very nice if the `ImageDelta`s would actually be enqueued, so that no deltas are lost. 

**To Reproduce**
+ Create a texture
+ use `set_partial` to patch multiple areas of the texture in a single frame
+ observe how only the last patch is actually applied 

**Expected behavior**
+ All patches are applied

 - OS: NixOS 21.09

wucke13:
introduces a queue for texture changes

Closes <https://github.com/emilk/egui/issues/1337>.

### DIFF
--- a/epaint/src/textures.rs
+++ b/epaint/src/textures.rs
@@ -39,7 +39,7 @@ impl TextureManager {
             filter,
         });
 
-        self.delta.set.insert(id, ImageDelta::full(image, filter));
+        self.delta.set.push((id, ImageDelta::full(image, filter)));
         id
     }
 
@@ -57,9 +57,10 @@ impl TextureManager {
                 // whole update
                 meta.size = delta.image.size();
                 meta.bytes_per_pixel = delta.image.bytes_per_pixel();
+                // since we update the whole image, we can discard all old enqueued deltas
+                self.delta.set.retain(|(x, _)| x != id);
             }
-
-            self.delta.set.insert(id, delta);
+            self.delta.set.push((id, delta));
         } else {
             crate::epaint_assert!(false, "Tried setting texture {id:?} which is not allocated");
         }
@@ -175,7 +176,7 @@ impl TextureMeta {
 #[must_use = "The painter must take care of this"]
 pub struct TexturesDelta {
     /// New or changed textures. Apply before painting.
-    pub set: AHashMap<TextureId, ImageDelta>,
+    pub set: Vec<(TextureId, ImageDelta)>,
 
     /// Textures to free after painting.
     pub free: Vec<TextureId>,
--- a/epaint/src/textures.rs
+++ b/epaint/src/textures.rs
@@ -58,7 +58,7 @@ impl TextureManager {
                 meta.size = delta.image.size();
                 meta.bytes_per_pixel = delta.image.bytes_per_pixel();
                 // since we update the whole image, we can discard all old enqueued deltas
-                self.delta.set.retain(|(x, _)| x != id);
+                self.delta.set.retain(|(x, _)| x != &id);
             }
             self.delta.set.push((id, delta));
         } else {


epaint/src/textures.rs
```diff
@@ -150,7 +153,7 @@ impl TextureMeta {
 #[must_use = "The painter must take care of this"]
 pub struct TexturesDelta {
     /// New or changed textures. Apply before painting.
-    pub set: AHashMap<TextureId, ImageDelta>,
+    pub set: AHashMap<TextureId, Vec<ImageDelta>>,
```
emilk, line 156: Maybe it would be even simpler to just make this a `Vec<(TextureId, ImageDelta)>`? I don't really see the point of having it be a hashmap anymore.

wucke13: Excellent suggestion. That avoids the double allocations. However, we would then always push to the end, and pop the front, like a FIFO, right? If so, maybe using a [`VecDequeue`]([VecDeque](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#)) would be clever? On the other hand, we always consume the full `Vec` at once when we consume it, so using `into_iter()` will probably be even more efficient than the VecDequeue, then again. Does `into_iter()` on a normal vector cause any de-allocations?

emilk: A `Vec` is the best fit since we always push to the back and consume the whole thing with either `std::mem::take` or `drain`

emilk:
I think we should make it a `Vec<(TextureId, ImageDelta)>` for simplicity sake. Are you still planning on working on this, or should I take it over?

epaint/src/textures.rs
```diff
@@ -49,8 +49,9 @@ impl TextureManager {
             if delta.is_whole() {
                 meta.size = delta.image.size();
                 meta.bytes_per_pixel = delta.image.bytes_per_pixel();
+                // since we update the whole image, we can discard all old enqueued deltas
```
emilk, line 52: seems like an odd comment to add considering the new code keeps the old deltas :)

wucke13: Good catch! I altered the code a bit to comply with the comment. My reasoning here would be: removing a couple of elements from the vector of texture deltas is likely less computational intensive than altering a texture a bunch of times only the to replace it completely.

wucke13:
Oof. `Vec::retain_mut` seems to be a rather new thing. @emilk what do you propose?

emilk:
I suggest:

``` rust
self.delta.set.retain(|(x, _)| x != id); 
self.delta.set.push((id, delta));
```

wucke13:
@emilk Please take a look again :)

emilk (approved):

epaint/src/textures.rs
```diff
@@ -57,9 +57,10 @@ impl TextureManager {
                 // whole update
                 meta.size = delta.image.size();
                 meta.bytes_per_pixel = delta.image.bytes_per_pixel();
+                // since we update the whole image, we can discard all old enqueued deltas
+                self.delta.set.retain(|(x, _)| x != id);
```
emilk, line 61:
```suggestion
                self.delta.set.retain(|(x, _)| x != &id);
```

Issue #1339, opened 2022-03-08
A way to access the current cursor/next-widget position


ascclemens:
**Is your feature request related to a problem? Please describe.**
Coming from Dear ImGui, I am very used to being able to access the current cursor position. It seems as if egui has made a conscious design choice to not expose this information. However, it is very useful for laying out elements.

For example, I have a program that is rendering markdown from a YAML file into egu in chunksi. I would like to centre certain chunks of the rendered markdown in the window. Normally, this could be done by using a group and scrolling the group to the centre, but since the markdown is being processed via a pull parser, I can't easily do that. A very easy solution would be to record the cursor position before rendering, then after, then compute the centre and scroll to that position.

egui does not expose the necessary information to do this, though, so I'm at an impasse.

**Describe the solution you'd like**
Make public at least `Ui::next_widget_position(&self) -> Pos2`. `Ui::cursor(&self) -> Rect` becoming public would also be nice.

PR #1342, opened 2022-03-08
Expose cursor and next_widget_position

ascclemens:
Change `Ui::cursor` and `Ui::next_widget_position` from crate-visible to public. These are useful for layout code and scrolling.

Closes <https://github.com/emilk/egui/issues/1339>.

### DIFF
--- a/egui/src/ui.rs
+++ b/egui/src/ui.rs
@@ -783,6 +783,16 @@ impl Ui {
         &self.placer
     }
 
+    /// Where the next widget will be put.
+    ///
+    /// One side of this will always be infinite: the direction in which new widgets will be added.
+    /// The opposing side is what is incremented.
+    /// The crossing sides are initialized to `max_rect`.
+    ///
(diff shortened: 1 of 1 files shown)


egui/src/ui.rs
```diff
@@ -783,7 +783,7 @@ impl Ui {
         &self.placer
     }
 
-    pub(crate) fn cursor(&self) -> Rect {
+    pub fn cursor(&self) -> Rect {
```
Mingun, line 786: If you make this public, you should add a documentation for it (i.e. or figure out its meaning itself, or ask @emilk)

emilk: I agree. You can find the documentation for it in `Region::cursor`. Note that `cursor` is a complicated thing, and also that it may change in future egui versions if/when the layout engine is rewritten (again). This is why it hasn't been exposed (yet).

ascclemens: I copied the docs from `Region::cursor`.

emilk (approved):

PR #1346, opened 2022-03-09
Fix typo in syntax_highlighting demo

mbillingr:
I had this *HUH?* moment when attempting a text search for the definition of `Highlighter`, which was actually spelled `Highligher`. 
Nothing in the sources indicated this was intentional, so I thought I'd go ahead and fix it. 

This seems like a trivial enough change so I didn't bother creating an issue first. Hopefully this won't create any inconveniences.


emilk (approved):

emilk:
Good catch!

PR #1349, opened 2022-03-09
Add input_filter method to TextEdit & example

### LINKED ISSUE #1348: Allow user to impl its own type checking directly in TextEdit
## Problem

User input in a `TextEdit` has to be manually validated each time because it allows basically any characters. This adds a lot of additional unnecessary validation steps when you expect anything more restricting than a `String`, for instance a `u64`.

## Solution

I would like something akin to [imgui::InputInt](https://docs.rs/imgui/0.8.0/imgui/struct.InputInt.html) which is like a more restrictive version of a `TextEdit` but that only allow numeric characters and prevents the user from entering an invalid integer. I think ideally however, you would want something more generic so the user can implement its own type checking etc.

jean-airoldie:
Allow input validation directly in the TextEdit using a function pointer. This allow the user to do the input checking in the text box itself and not have to do an extra validation step, making the UI more user friendly. Validation is implemented for some primitive types, and the user can easily extend it to his custom types.

As for the behavior of the `TextEdit` itself, I basically copied `imgui` data type widget behavior. Character inputs that are invalid are simply ignored, character inputs that are valid but would result in a value to large or too small result in the buffer being replaced by the maximum or minimum value. The behavior is the same with pasted text.

## TODO
* [x] Implement input validation for unsigned integers & example
* [x] Implement input validation for signed integers & example
* [x] Allow user to implement his own input validation via a custom type & example

## Open Questions
* ~The `InputValidation` and validator name isn't exact because we are really intercepting input and alter it. Perhaps filter is a better name.~ Renamed to filter.
* Currently we assume that the buffer was not filled with forbidden characters by the user for maximum performance. This is probably a bad assumption to make.
* Implement input validation for floating point numbers? I don't think its a good idea to expose floating point numbers to users directly because they are unintuitive, a decimal type would be better.
* ~Should information on the end of the cursor be provided in the `InputData`?~ Added `#[non_exhaustive]` flag to allow future extension.

Closes #1348

edit: Remove outdated stuff

### DIFF
--- a/egui/src/widgets/text_edit/builder.rs
+++ b/egui/src/widgets/text_edit/builder.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 use epaint::mutex::Arc;
 
 use epaint::text::{cursor::*, Galley, LayoutJob};
@@ -47,7 +49,7 @@ use super::{CCursorRange, CursorRange, TextEditOutput, TextEditState};
 /// ## Advanced usage
 /// See [`TextEdit::show`].
 #[must_use = "You should put this widget in an ui with `ui.add(widget);`"]
-pub struct TextEdit<'t> {
+pub struct TextEdit<'t, 'v> {
     text: &'t mut dyn TextBuffer,
     hint_text: WidgetText,
     id: Option<Id>,
@@ -64,13 +66,14 @@ pub struct TextEdit<'t> {
     desired_height_rows: usize,
     lock_focus: bool,
     cursor_at_end: bool,
+    validator: Option<&'v mut dyn FnMut(InputData<'_>) -> Action>,
 }
 
-impl<'t> WidgetWithState for TextEdit<'t> {
+impl<'t, 'v> WidgetWithState for TextEdit<'t, 'v> {
     type State = TextEditState;
 }
 
-impl<'t> TextEdit<'t> {
+impl<'t, 'v> TextEdit<'t, 'v> {
     pub fn load_state(ctx: &Context, id: Id) -> Option<TextEditState> {
         TextEditState::load(ctx, id)
     }
@@ -80,7 +83,7 @@ impl<'t> TextEdit<'t> {
     }
 }
 
-impl<'t> TextEdit<'t> {
+impl<'t, 'v> TextEdit<'t, 'v> {
     /// No newlines (`\n`) allowed. Pressing enter key will result in the `TextEdit` losing focus (`response.lost_focus`).
     pub fn singleline(text: &'t mut dyn TextBuffer) -> Self {
         Self {
@@ -109,6 +112,7 @@ impl<'t> TextEdit<'t> {
             desired_height_rows: 4,
             lock_focus: false,
             cursor_at_end: true,
+            validator: None,
         }
     }
 
@@ -248,15 +252,53 @@ impl<'t> TextEdit<'t> {
     }
 }
 
+/// Instructions on how the user input should be processed.
+pub enum Action {
+    /// Insert the string at the cursor.
+    ///
+    /// Note that inserting an empty string does nothing.
+    Insert(String),
+
+    /// Overwrite the entire buffer with the given string.
+    ///
+    /// This can be used for instance, to automatically fill in the
+    /// minimum or maximum value possible, if it is exceeded.
+    Overwrite(Cow<'static, str>),
+}
+
+/// The information required to validate text input.
+pub struct InputData<'a> {
+    /// The text that was input by the user.
+    pub input: String,
+
+    /// The current text in the text box buffer.
+    pub buffer: &'a str,
+
+    /// The current position of the start of the cursor.
+    ///
+    /// This represents where the data will be inserted.
+    pub cursor_start: usize,
+}
+
+impl<'t, 'v> TextEdit<'t, 'v> {
+    /// Set the function to use for input validation.
+    ///
+    /// Defaults to no input validation.
+    pub fn validator(mut self, validator: &'v mut dyn FnMut(InputData<'_>) -> Action) -> Self {
+        self.validator.replace(validator);
+        self
+    }
+}
+
 // ----------------------------------------------------------------------------
 
-impl<'t> Widget for TextEdit<'t> {
+impl<'t, 'v> Widget for TextEdit<'t, 'v> {
     fn ui(self, ui: &mut Ui) -> Response {
         self.show(ui).response
     }
 }
 
-impl<'t> TextEdit<'t> {
+impl<'t, 'v> TextEdit<'t, 'v> {
     /// Show the [`TextEdit`], returning a rich [`TextEditOutput`].
     ///
     /// ```
@@ -348,6 +390,7 @@ impl<'t> TextEdit<'t> {
             desired_height_rows,
             lock_focus,
             cursor_at_end,
+            validator,
         } = self;
 
         let text_color = text_color
@@ -496,6 +539,11 @@ impl<'t> TextEdit<'t> {
                 CursorRange::default()
             };
 
+            let mut default_validator = |data: InputData<'_>| Action::Insert(data.input);
+            let validator = match validator {
+                Some(validator) => validator,
+                None => &mut default_validator,
+            };
             let (changed, new_cursor_range) = events(
                 ui,
                 &mut state,
@@ -507,6 +555,7 @@ impl<'t> TextEdit<'t> {
                 multiline,
                 password,
                 default_cursor_range,
+                validator,
             );
 
             if changed {
@@ -662,6 +711,7 @@ fn events(
     multiline: bool,
     password: bool,
     default_cursor_range: CursorRange,
+    validator: &mut dyn FnMut(InputData<'_>) -> Action,
 ) -> (bool, CursorRange) {
     let mut cursor_range = state.cursor_range(&*galley).unwrap_or(default_cursor_range);
 
@@ -681,7 +731,7 @@ fn events(
     let mut any_change = false;
 
     let events = ui.input().events.clone(); // avoid dead-lock by cloning. TODO: optimize
-    for event in &events {
+    for event in events {
         let did_mutate_text = match event {
             Event::Copy => {
                 if cursor_range.is_empty() {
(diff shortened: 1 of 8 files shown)


egui/src/widgets/text_edit/builder.rs
```diff
@@ -81,17 +86,15 @@ impl<'t> TextEdit<'t> {
...
+    pub fn new(text: &'t mut dyn TextBuffer) -> TextEdit<'t, String> {
+        Self::with_validator(text)
     }
+}
 
-    /// A `TextEdit` for multiple lines. Pressing enter key will create a new line.
-    pub fn multiline(text: &'t mut dyn TextBuffer) -> Self {
+impl<'t, V> TextEdit<'t, V> {
+    /// Set the input validator type.
+    pub fn with_validator(text: &'t mut dyn TextBuffer) -> TextEdit<'t, V> {
```
emilk, line 97: Did you consider the alternative of passing a function pointer instead? i.e. `validator: fn(data: InputData<'_>) -> Action` or `validator: &dyn Fn(data: InputData<'_>) -> Action`. This makes it more flexible, allowing dynamic filters, and allows us to pass the validator as another builder step: `TextEdit::singleline(text).with_validator(&my_validator)`

egui/src/widgets/text_edit/validation.rs
```diff
@@ -0,0 +1,204 @@
...
+    ///
+    /// Note that inserting an empty string does nothing.
+    Insert(String),
+
+    /// Overwrite the entire buffer with the given string.
+    ///
+    /// This can be used for instance, to automatically fill in the
+    /// minimum or maximum value possible, if it is exceeded.
+    Overwrite(Cow<'static, str>),
+}
```
emilk, line 15: I think this is a very good design!

egui/src/widgets/text_edit/validation.rs
```diff
@@ -0,0 +1,204 @@
...
+                            Action::Insert(filter)
+                        }
+                    }
+                }
+            }
+        }
+    };
+}
+
+impl_validate_input_signed!(i8, "127", "-128");
```
emilk, line 201: nitpicky, but `min` should come before `max`

egui/src/widgets/text_edit/builder.rs
jean-airoldie: Interesting idea, I think its better design indeed. I'm not to used to the immediate semantics of basically creating temporary structures each frame, so I didn't even think about it.

jean-airoldie: Also solves the naming problem.

jean-airoldie: Yep its much nicer.

jean-airoldie:
Did a force push to remove all the pollution I previously did with that unnecessary breaking change.

jean-airoldie:
I am unsure if this is a breaking change since this adds an extra lifetime to the `TextEdit`, however, this lifetime isn't relevent until you actually set a validator. So I'm guessing its a breaking change.

egui/src/widgets/text_edit/validation.rs
jean-airoldie: thanks

egui/src/widgets/text_edit/validation.rs
jean-airoldie: agreed

jean-airoldie:
> Implement input validation for floating point numbers? I don't think its a good idea to expose floating point numbers to users directly because they are unintuitive, a decimal type would be better.

Regarding this question, is it undefined behavior if a user modify the content of the `&mut dyn TextBuffer` directly without going through the UI? If so then it should be documented and the current validator implementation would be sane.

emilk:
> > Currently we assume that the buffer was not filled with forbidden characters by the user for maximum performance. This is probably a bad assumption to make.
> 
> @emilk Regarding this point, do you consider it undefined behavior if a user modifies the content of the `&mut dyn TextBuffer` directly without going through the UI? If so then it should be documented and the current validator implementation would be sane.

Assuming that by "user" here you mean "the programmer using egui" (and not the user typing in things in the UI):

I think if makes sense to assume that if the programmer sets the buffer to contain 'invalid' text, they did so for a valid reason, and we shouldn't try to protect the programmer from themselves. Only user input should be filtered, methinks.

egui_demo_lib/src/apps/demo/input_validation.rs
```diff
@@ -0,0 +1,122 @@
...
+                .validator(&mut i64::validate_input)
+                .hint_text("a valid `i64`")
+                .show(ui);
+        });
+
+        ui.horizontal(|ui| {
+            ui.label("custom validator:");
+            TextEdit::singleline(&mut self.a_buffer)
+                .validator(&mut custom_validator)
+                .hint_text("can only be filled with `a` characters")
```
emilk, line 118: Perhaps do a slightly more useful demo, e.g. only allow numbers

egui_demo_lib/src/apps/demo/input_validation.rs
```diff
@@ -0,0 +1,122 @@
...
+                .show(ui);
+        });
+
+        ui.horizontal(|ui| {
+            ui.label("i64:");
+            TextEdit::singleline(&mut self.i64_buffer)
+                .validator(&mut i64::validate_input)
+                .hint_text("a valid `i64`")
+                .show(ui);
+        });
```
emilk, line 112: Perhaps one or two of these is enough :)

emilk (commented):
Starting to look good!

egui/src/widgets/text_edit/builder.rs
```diff
@@ -47,7 +49,7 @@ use super::{CCursorRange, CursorRange, TextEditOutput, TextEditState};
 /// ## Advanced usage
 /// See [`TextEdit::show`].
 #[must_use = "You should put this widget in an ui with `ui.add(widget);`"]
-pub struct TextEdit<'t> {
+pub struct TextEdit<'t, 'v> {
```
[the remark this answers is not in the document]
jean-airoldie: Yeah, I'm trying to think of a use case where the two lifetimes are useful, but I can't think of anything.

egui_demo_lib/src/apps/demo/input_validation.rs
jean-airoldie: Yeah, I used those to test if the behavior is as expected, since there's no real other way to tests UI features.

egui/src/widgets/text_edit/builder.rs
```diff
@@ -248,15 +252,53 @@ impl<'t> TextEdit<'t> {
...
+    ///
+    /// This represents where the data will be inserted.
+    pub cursor_start: usize,
+}
+
+impl<'t, 'v> TextEdit<'t, 'v> {
+    /// Set the function to use for input validation.
+    ///
+    /// Defaults to no input validation.
+    pub fn validator(mut self, validator: &'v mut dyn FnMut(InputData<'_>) -> Action) -> Self {
```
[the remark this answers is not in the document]
jean-airoldie: Ouais filter is a better name. I'll add a minimal doctest example.

egui_demo_lib/src/apps/demo/input_validation.rs
jean-airoldie: I'll do something more substantial.

jean-airoldie:
I took the liberty of adding `non_exhaustive` flags to `InputData` as well as `Action` for some future-proofness.

jean-airoldie:
> I think if makes sense to assume that if the programmer sets the buffer to contain 'invalid' text, they did so for a valid reason, and we shouldn't try to protect the programmer from themselves. Only keyboard input should be filtered, methinks.

Alright. So in my filter implementation of integer primitives, I don't parse integers because its really slow (cuz radix 10 is slow) and instead I do character lexicographic comparison which only makes sense if the buffer exclusively contains numeric characters. Do you think its fine? Maybe I should document this behavior somewhere.

egui/src/widgets/text_edit/builder.rs
```diff
@@ -248,6 +252,84 @@ impl<'t> TextEdit<'t> {
...
+/// The information required to validate text input.
+#[non_exhaustive]
+pub struct InputData<'a> {
+    /// The text that was input by the user.
+    pub input: String,
+
+    /// The current text in the text box buffer.
+    pub buffer: &'a str,
+
+    /// The current position of the start of the cursor.
```
emilk, line 279:
```suggestion
    /// The current position of the start of the cursor, in characters (not bytes!).
```

egui/src/widgets/text_edit/builder.rs
```diff
@@ -248,6 +252,84 @@ impl<'t> TextEdit<'t> {
...
+    ///
+    /// This represents where the data will be inserted.
+    pub cursor_start: usize,
+}
+
+impl<'t> TextEdit<'t> {
+    /// Set the function to use for input filtering.
+    ///
+    /// An input filter can be used to force the user to enter a valid
+    /// value in the `TextEdit`. This can be used to do type checking
```
emilk, line 289:
```suggestion
    /// value in the [`TextEdit`]. This can be used to do type checking
```

egui/src/widgets/text_edit/builder.rs
```diff
@@ -496,6 +579,11 @@ impl<'t> TextEdit<'t> {
                 CursorRange::default()
             };
 
+            let mut default_filter = |data: InputData<'_>| Action::Insert(data.input);
+            let filter = match filter {
+                Some(filter) => filter,
+                None => &mut default_filter,
+            };
```
emilk, line 586:
```suggestion
            let mut no_filter = |data: InputData<'_>| Action::Insert(data.input);
            let filter = filter.unwrap_or(&mut no_filter);
```

egui/src/widgets/text_edit/builder.rs
```diff
@@ -248,6 +252,84 @@ impl<'t> TextEdit<'t> {
...
+    /// The text that was input by the user.
+    pub input: String,
+
+    /// The current text in the text box buffer.
+    pub buffer: &'a str,
+
+    /// The current position of the start of the cursor.
+    ///
+    /// This represents where the data will be inserted.
+    pub cursor_start: usize,
```
emilk, line 282: If a range of characters are selected than this is misleading. Perhaps remove it, or replace with `CursorRange` or `CCursorRange`.

egui/src/widgets/text_edit/filter.rs
```diff
@@ -0,0 +1,178 @@
...
+                let new_size = buffer.len() + filter.len();
+
+                match new_size.cmp(&max_str_len) {
+                    // Guaranteed to be below the limit so we insert.
+                    cmp::Ordering::Less => Action::Insert(filter),
+                    // Guaranteed to exceed the limit so we overwrite.
+                    cmp::Ordering::Greater => Action::Overwrite(Cow::Borrowed(max_str)),
+                    // Slow case where we need to actually compare with the limit.
+                    cmp::Ordering::Equal => {
+                        // Note that we don't parse the strings because its too slow.
```
emilk, line 47: This is not a great motivation IMHO. We only need to run this code when new text is entered, so it is not even every frame.

egui/src/widgets/text_edit/filter.rs
```diff
@@ -0,0 +1,178 @@
...
+                        }
+
+                        let back = max_str[buffer.len()..].chars();
+                        for (current, max) in iter.zip(back) {
+                            if current > max {
+                                return Action::Overwrite(Cow::Borrowed(max_str));
+                            }
+                        }
+
+                        Action::Insert(filter)
```
emilk, line 63: afaict this compares the existing buffer, but does not take into account the new input, nor where it is to be inserted.

I think we should consider an API where the user gets handed the suggested `new_buffer`, i.e. what would be produced without any filtering. The user could then very easily check that with a e.g. a regex, or lexicographically against a maximum value. In such an API we could get away with a single input (the suggested new `String`) and an `enum Action { Revert, Replace(String) }`, where `Revert` would revert to the previous value, forbidding the new input (e.g. because it doesn't fit the regex).

Presumably filters will be used for short `TextEdit`, and since the code is only run on new input we don't need to worry much about performance and should focus on usability.

emilk (changes requested):
Reading your filtering code a bit closer got me thinking we should have a much simpler API - see the relevant comment.

egui/src/widgets/text_edit/filter.rs
jean-airoldie: > afaict this compares the existing buffer, but does not take into account the new input, nor where it is to be inserted.

This basically always assumes the worst case scenario that the user has single cursor (and thus nothing selected). So this won't work properly when strings are selected indeed.

> I think we should consider an API where the user gets handed the suggested new_buffer, i.e. what would be produced without any filtering.

Yeah that sounds like a more user friendly API.

jean-airoldie:
I'll be abandoning this PR unfortunately, I ran out of free time for now and I found an alternative solution for my problem.

I'm sorry for wasting your time.

emilk:
I'm sorry to hear that - it was a great effort, and maybe me or someone else will resume it!

jean-airoldie:
Yeah, at least the next person to try will have a good starting point if they look at this PR.

TimTheBig:
So sad

Issue #1198, opened 2022-02-02
Expose glow::Context via `epi::Frame`


emilk:
Since we are converging on using `glow` as the rendering backend for `eframe` on both native and web, it would be quite useful to expose the glow GL context to the user, in particular in order to do paint-to-texture from within `eframe`.

This would need to be opt-in thing though (optional dependency of `epi`) and be fallible (e.g. `frame.glow_context()` should return `None` if we aren't using a glow renderer).

This would obviously be for advanced users.

I'm not sure if there is interest in this though. My thinking is that if one wants to do 3D stuff with egui, one should probably use `bevy_egui`, `egui-macrquad`, or similar. Still, `eframe` has many features these libraries lack (reactive mode, better i/o on web).

coderedart:
there could be some considerations for the still imaginary multi-window feature. gl contexts are usually *per* window.
I don't see any negatives here. as long as it is marked unsafe, those who want to experiment can do so. some more questions:
does the user get to request a specific version of opengl (4.6?) at the start of program?
what state is okay to change (since its all global) eg: drawing into texture means changing framebuffers/attachments and maybe more exotic stuff like polygon mode /depth test / stencil test  which egui doesn't set every frame right now. 

there could be a slight chance that people will be hitting discord with a lot of opengl questions and we might need to pin renderdoc download link at the top

emilk:
I'm now looking into adding a custom `Shape` with a callback to which the glow context will be passed. That would allow users to embed 3D stuff and other custom painting directly into an egui Ui. This will be similar to how Dear ImGui has `AddCallback/UserCallback`

PR #1351, opened 2022-03-10
Add Shape::Callback to do custom rendering inside of an egui UI

emilk:
This adds a new shape `Shape::Callback` which can be used to inline custom rendering code inside an egui UI. The callback will be given the underlying rendering context, and will therefore be integration specific. When using `eframe`, it will be `glow::Context` (unless one opts in to the glium or WebGL backends).

This means you can paint anything anywhere, including behind or on top of other egui windows and with transparency. Previously this was only possible by first rendering to a texture and then displaying that texture using `egui::TextureId`.

The design is very similar to Dear ImGui's `AddCallback/UserCallback`.

Mar-10-2022 12-50-57

Closes https://github.com/emilk/egui/issues/1198

## Unresolved issues
One unresolved issue is where to free any resources (shaders, buffer, textures, …) allocated by the custom painting callback. Perhaps we could pass a `Rc<Cell<glow::Context>>` and let the user clone it and destroy the resources on `Drop`?


coderedart:
glow::Context doesn't need to be mutable. so, a simple `Rc<glow::Context>` should be good. window must be alive for `Rc<glow::Context>` to work in their User struct Drop impl (otherwise, segfaults). .

one question i have is whether the user will be able to access the textures loaded into egui already? I would assume they don't want to duplicate textures in their app used both by egui::Image as well as their own custom rendering code, so reusing the textures from egui's tex manager would be cool.

emilk:
> one question i have is whether the user will be able to access the textures loaded into egui already? I would assume they don't want to duplicate textures in their app used both by egui::Image as well as their own custom rendering code, so reusing the textures from egui's tex manager would be cool.

Great question!

My initial thinking was that a user would either want to either manage their textures manually OR ask egui to do so, but not both. But I guess we could pass the full `egui_glow::Painter` along with `glow::Context` so the users could ask for any textures they desire.

t18b219k:
if we able to redirecting to Texture would save energy.

emilk:
There - now the callback code has access to `egui_glow::Painter` which contains the `glow::Context` and all textures.

emilk:
The top-level types need some documentation. Running `cargo doc --no-deps -p egui_extras --open` gives this:

Screen Shot 2022-03-12 at 08 15 59

(no summary of what these are)

emilk:
I think this is a decent start.

Future work: Make `glow` a first-class part of the `epi/eframe` interface so that users can run `glow`-related setup- and teardown code. Coming in:
* https://github.com/emilk/egui/pull/1363

PR #1359, opened 2022-03-12
WIP: use fontdue instead of ab_glyph for text rendering

manuel-woelker:
**Work in Progress**

A quick proof of concept to use [fontdue](https://github.com/mooman219/fontdue) instead of ab_glyph. Just wanted to see how hard this would be and if there is any interest in this. If you have any thoughts, please let me know!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1207,6 +1207,7 @@ dependencies = [
  "cint",
  "criterion",
  "emath",
+ "fontdue",
  "nohash-hasher",
  "parking_lot 0.12.0",
  "serde",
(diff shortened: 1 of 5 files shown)


emilk:
What would be the main motivation for this change?

There is slightly less dependencies, which is nice. It would be nice to see before/after benchmarks to see if there is any improvement.

The main thing I want to address with text is to use a text layout library supporting things like right-to-left, which fontdue doesn't have yet.

adals:
I don't know but maybe one of them could help and support RTL(right to left)

[https://github.com/RazrFalcon/rustybuzz](https://github.com/RazrFalcon/rustybuzz)
[https://github.com/dfrg/swash](https://github.com/dfrg/swash)
[https://github.com/yeslogic/allsorts](https://github.com/yeslogic/allsorts)

manuel-woelker:
Yeah, RTL and advanced features like shaping currently don't seem to be supported in fontdue, so it make sense to revisit this at a later date.

Performance/better feature support might be reasons for a switch, although a deeper analysis should be done to inform any decisions.

As I said, I mainly wanted to see how easy it would be to swap the font rasterization engine.

PR #1363, opened 2022-03-14
eframe app creation refactor

emilk:
This refactors how eframe creates an app, removing the need for `App::setup`.

The purpose of this is to be able to access the following state before/during creation of ones app:
* `egui::Context`: alllow users to set fonts, style and maybe create textures.
* `epi::Storage`: for reading app state from storage.
* `epi::Frame`: can be cloned and stored so we can run [`request_repaint`](https://docs.rs/epi/latest/epi/struct.Frame.html#method.request_repaint).
* `glow::Context`: for creating rendering resources.

Previously eframe was designed so that these were passed into `App::setup`, but that meant one would have to create an app before running the setup of it, making `App::setup` a weird "second constructor". This PR solves that problem.

`App::name` is now also removed, and is instead passed in by argument to `eframe::run_native`.

Additionally, `eframe` now re-exports everything in `epi`, so `eframe` users can ignore `epi` completely.

---

This now changes the `eframe` use to look like so:

``` rust
use eframe::egui;

fn main() {
    eframe::run_native("MyApp", native_options, |cc| Box::new(MyEguiApp::new(cc)));
}

#[wasm_bindgen]
pub fn start(canvas_id: &str) -> Result<(), eframe::wasm_bindgen::JsValue> {
    eframe::start_web(canvas_id, |cc| Box::new(MyApp::new(cc)))
}

#[derive(Default)]
struct MyEguiApp {}

impl MyEguiApp {
        Self::default()
    }
}

   fn update(&mut self, ctx: &egui::Context, frame: &eframe::Frame) {
           ui.heading("Hello World!");
       });
   }
}
```

### DIFF
--- a/eframe/examples/confirm_exit.rs
+++ b/eframe/examples/confirm_exit.rs
@@ -9,10 +9,6 @@ struct MyApp {
 }
 
 impl epi::App for MyApp {
-    fn name(&self) -> &str {
-        "Confirm exit"
-    }
-
     fn on_exit_event(&mut self) -> bool {
         self.is_exiting = true;
         self.can_exit
@@ -45,5 +41,7 @@ impl epi::App for MyApp {
 
 fn main() {
     let options = eframe::NativeOptions::default();
-    eframe::run_native(Box::new(MyApp::default()), options);
+    eframe::run_native("Confirm exit", options, |_, _, _, _| {
+        Box::new(MyApp::default())
+    });
 }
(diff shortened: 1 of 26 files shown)


AngelOnFira:
Sorry for the comment noise, just wanted to mention that it took me a while to find this while trying to understand why setup + name existed on `ewebsocket`'s example app. This PR seems to give a good overview of how an app should be created; would it be worth specifying in some of `eframe`/`epi`'s docs about how code *used* to look, so that people know how to migrate from setup if they need? If so, I can try upstreaming some docs.

I'm working on updating `ewebsock`'s example app, so hopefully confusion shouldn't spawn from there in the future :)

Issue #1279, opened 2022-02-21
Add texture/image painting transformations


quietvoid:
**Is your feature request related to a problem? Please describe.**
There doesn't seem to be a way to make modifications to an existing texture or when painting an image.
My current situation is that I have to store an image, and process it on demand (for example, zooming or translating), before recreating the final image to load into a texture.
This seems very inefficient, as I'm constantly copying the image to perform resize/crop operations.

Coming from `piston_window`, there exists a public API to setup a transform for use when drawing an image.

I am using `eframe`, and only tested in native.

**Describe the solution you'd like**
It would be nice to be able to transform textures directly.
Or maybe through `ui.image()`?

**Describe alternatives you've considered**
Currently, I'm processing intermediate images using `image` and `fast_image_resize` separately.
So the texture is reloaded every modification required.

emilk:
`egui::Image` is very simple at the moment. If you want to do more complex things (scaling, rotating, …) you will have to use the `Painter` API directly.

quietvoid:
Thanks. I'll have a look at the Painter API.

songlinshu:
I like egui very much. As the functional requirements of widgets increase, egui will be added_ Extras is absolutely right. By the way, do you plan to merge dynamic grid? thank you!

PR #1371, opened 2022-03-17
Add Image::rotate and Mesh::rotate

Hunter522:
Related to https://github.com/emilk/egui/issues/1279

This adds rotation capabilities to `Image` and `Mesh` (since `Image` uses it). You can specify an origin to rotate about too.

Updated the `eframe/examples/image.rs` example:

image

### DIFF
--- a/egui/src/widgets/image.rs
+++ b/egui/src/widgets/image.rs
@@ -1,4 +1,5 @@
 use crate::*;
+use emath::Rot2;
 
 /// An widget to show an image of a given size.
 ///
@@ -36,6 +37,7 @@ pub struct Image {
     bg_fill: Color32,
     tint: Color32,
     sense: Sense,
+    rotation: Option<(Rot2, Vec2)>,
 }
 
 impl Image {
@@ -47,6 +49,7 @@ impl Image {
             bg_fill: Default::default(),
             tint: Color32::WHITE,
             sense: Sense::hover(),
+            rotation: None,
         }
     }
 
@@ -75,6 +78,15 @@ impl Image {
         self.sense = sense;
         self
     }
+
+    /// Rotate the image about an origin by some angle
+    ///
+    /// Positive angle is clockwise.
+    /// Origin is a vector relative to the top-left corner of the image.
+    pub fn rotate(mut self, angle: f32, origin: Vec2) -> Self {
+        self.rotation = Some((Rot2::from_angle(angle), origin));
+        self
+    }
 }
 
 impl Image {
@@ -92,6 +104,7 @@ impl Image {
                 bg_fill,
                 tint,
                 sense: _,
+                rotation,
             } = self;
 
             if *bg_fill != Default::default() {
@@ -104,6 +117,10 @@ impl Image {
                 // TODO: builder pattern for Mesh
                 let mut mesh = Mesh::with_texture(*texture_id);
                 mesh.add_rect_with_uv(rect, *uv, *tint);
+                if let Some((rot, origin)) = rotation {
+                    let origin_abs = rect.min.to_vec2() + *origin;
+                    mesh.rotate(*rot, origin_abs);
+                }
                 ui.painter().add(Shape::mesh(mesh));
             }
         }
--- a/egui/src/widgets/image.rs
+++ b/egui/src/widgets/image.rs
@@ -82,9 +82,11 @@ impl Image {
     /// Rotate the image about an origin by some angle
     ///
     /// Positive angle is clockwise.
-    /// Origin is a vector relative to the top-left corner of the image.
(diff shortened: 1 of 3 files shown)


epaint/src/mesh.rs
```diff
@@ -242,6 +242,17 @@ impl Mesh {
...
+
+    /// Rotate by some angle about an origin, in-place.
+    ///
+    /// Origin is relative to the top left corner of the screen.
+    pub fn rotate(&mut self, rot: Rot2, origin: Vec2) {
+        self.translate(-origin);
+        for v in &mut self.vertices {
+            v.pos = (rot * v.pos.to_vec2()).to_pos2();
+        }
+        self.translate(origin);
```
emilk, line 254: This iterates three times instead of one - just to the translation back and forth inside the loop instead for better performance

emilk (changes requested):
Nice addition!

epaint/src/mesh.rs
Hunter522: Good thinking, I reduced it to one line now.

egui/src/widgets/image.rs
```diff
@@ -75,6 +78,15 @@ impl Image {
         self.sense = sense;
         self
     }
+
+    /// Rotate the image about an origin by some angle
+    ///
+    /// Positive angle is clockwise.
+    /// Origin is a vector relative to the top-left corner of the image.
```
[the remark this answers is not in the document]
Hunter522: I think I agree. Using UV space here would be more ergonomic. Updated in latest and documented the coordinate spaces in both new added rotate methods.

Hunter522:
Rebased onto master to fix merge conflict.

epaint/src/mesh.rs
```diff
@@ -242,6 +242,15 @@ impl Mesh {
             v.pos += delta;
         }
     }
+
+    /// Rotate by some angle about an origin, in-place.
+    ///
+    /// Origin is a vector in screen space.
+    pub fn rotate(&mut self, rot: Rot2, origin: Vec2) {
+        for v in &mut self.vertices {
+            v.pos = (rot * (v.pos - origin).to_vec2()).to_pos2() + origin;
```
emilk, line 251:
```suggestion
    /// Origin is a position in screen space.
    pub fn rotate(&mut self, rot: Rot2, origin: Pos2) {
        for v in &mut self.vertices {
            v.pos = origin + rot * (v.pos - origin);
```

egui/src/widgets/image.rs
```diff
@@ -104,6 +119,10 @@ impl Image {
                 // TODO: builder pattern for Mesh
                 let mut mesh = Mesh::with_texture(*texture_id);
                 mesh.add_rect_with_uv(rect, *uv, *tint);
+                if let Some((rot, origin)) = rotation {
+                    let origin_abs = rect.min.to_vec2() + (*origin * *size);
+                    mesh.rotate(*rot, origin_abs);
```
emilk, line 124:
```suggestion
                    let origin_abs = rect.min + *origin * *size;
                    mesh.rotate(*rot, origin_abs);
```

egui/src/widgets/image.rs
```diff
@@ -75,6 +78,17 @@ impl Image {
...
     }
+
+    /// Rotate the image about an origin by some angle
+    ///
+    /// Positive angle is clockwise.
+    /// Origin is a vector in UV space (normalized space by default unless set by [`Image::uv`])
+    ///
+    /// To rotate about the center you can pass `Vec2::splat(0.5)` as the origin.
+    pub fn rotate(mut self, angle: f32, origin: Vec2) -> Self {
+        self.rotation = Some((Rot2::from_angle(angle), origin / self.uv.size()));
```
emilk, line 89: dividing by uv.size doesn't make much sense to me – you need to take the whole uv rect into account.

emilk (changes requested):

Hunter522:
@emilk thanks for your suggestions. Take a look at the latest commit. 

Here's a screenshot of how it looks when I set a different uv range (bottom left quadrant of image):

image

```rust
ui.add(
    egui::Image::new(self.image.texture_id(ctx), self.image.size_vec2())
        .uv(egui::Rect::from_min_size(
            egui::pos2(0.0, 0.5),
            egui::Vec2::splat(0.5),
        ))
        .rotate(self.angle.to_radians(), egui::Vec2::splat(0.5)),
);
```

You can still pass `Vec2::splat(0.5)` to rotate about the center of the *resulting* image with the new UV range. I think this keeps rotating easy to use here even if a new UV range is set. Let me know what you think.

emilk (approved):
Looks good!

PR #1380, opened 2022-03-19
Fix code that could lead to a possible deadlock.

DusterTheFirst:
This pull request aims to address some accidental double-read locks on `RwLock`s made by egui. When acquiring 2 read locks, if a thread in parallel happened to attempt to acquire a write lock between the two read locks, the second read lock will deadlock. This deadlock occurs due to `RwLock`s refusing any new read locks on a `RwLock` that has a pending Write lock which in most cases can lead to better fairness.

Each commit has a detailed description of the deadlock and how the changes prevent it.

Rather than taking one lock and reusing it, it could be possible to use something like `parking_lot`'s [`read_recursive`](https://docs.rs/lock_api/0.4.6/lock_api/struct.RwLock.html#method.read_recursive) which would bypass the aforementioned second read lock from blocking if a write lock is queued, but in my opinion it would be less clear and maintainers would need to understand the difference between `read()` and `read_recursive()`. Using the already commonly understood scoping/drop system of Rust seems like the better choice.

**Note:** Each of these deadlocks have been found by accident, I did not do an extensive comb through the egui code for these situations so others are bound to exist in the codebase that I have not encountered. These deadlocks were just encountered in my code where a second thread was calling `ctx.request_repaint` very frequently.

**Note 2:** There is no way that I know of to prove that the deadlocks are gone, but the deadlocks have gone away in my application after these changes.

### DIFF
--- a/egui/src/menu.rs
+++ b/egui/src/menu.rs
@@ -277,10 +277,10 @@ impl MenuRoot {
         root: &mut MenuRootManager,
         id: Id,
     ) -> MenuResponse {
-        let pointer = &response.ctx.input().pointer;
-        if (response.clicked() && root.is_menu_open(id))
-            || response.ctx.input().key_pressed(Key::Escape)
-        {
+        // Lock the input once for the whole function call
+        let input = response.ctx.input();
+
+        if (response.clicked() && root.is_menu_open(id)) || input.key_pressed(Key::Escape) {
             // menu open and button clicked or esc pressed
             return MenuResponse::Close;
         } else if (response.clicked() && !root.is_menu_open(id))
@@ -290,8 +290,8 @@ impl MenuRoot {
             // or button hovered while other menu is open
             let pos = response.rect.left_bottom();
             return MenuResponse::Create(pos, id);
-        } else if pointer.any_pressed() && pointer.primary_down() {
-            if let Some(pos) = pointer.interact_pos() {
+        } else if input.pointer.any_pressed() && input.pointer.primary_down() {
+            if let Some(pos) = input.pointer.interact_pos() {
                 if let Some(root) = root.inner.as_mut() {
                     if root.id == id {
                         // pressed somewhere while this menu is open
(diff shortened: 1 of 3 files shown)


emilk (approved):
Good catch!

egui/src/menu.rs
```diff
@@ -277,10 +277,10 @@ impl MenuRoot {
         root: &mut MenuRootManager,
         id: Id,
     ) -> MenuResponse {
-        let pointer = &response.ctx.input().pointer;
-        if (response.clicked() && root.is_menu_open(id))
-            || response.ctx.input().key_pressed(Key::Escape)
-        {
+        // Lock the input once for the whole function call (see #1380)
```
emilk, line 280:
```suggestion
        // Lock the input once for the whole function call (see https://github.com/emilk/egui/pull/1380).
```

egui/src/widgets/slider.rs
```diff
@@ -455,10 +455,14 @@ impl<'a> Slider<'a> {
 
     fn value_ui(&mut self, ui: &mut Ui, position_range: RangeInclusive<f32>) -> Response {
         // If `DragValue` is controlled from the keyboard and `step` is defined, set speed to `step`
-        let change = ui.input().num_presses(Key::ArrowUp) as i32
-            + ui.input().num_presses(Key::ArrowRight) as i32
-            - ui.input().num_presses(Key::ArrowDown) as i32
-            - ui.input().num_presses(Key::ArrowLeft) as i32;
+        let change = {
+            // Hold one lock rather than 4 (see #1380)
```
emilk, line 459:
```suggestion
            // Hold one lock rather than 4 (see https://github.com/emilk/egui/pull/1380).
```

emilk (approved):

emilk:
Another solution is to use a standard `Mutex` instead of a `RwLock`.

emilk:
This problem would be mitigated by https://github.com/emilk/egui/pull/1389

PR #1382, opened 2022-03-20
Add Plot::allow_scroll

rukai:
I needed to disable scrolling as it conflicts with the scroll action of a ScrollArea but it is currently tied to the zoom functionality which I want to keep.

### DIFF
--- a/egui_demo_lib/src/apps/demo/context_menu.rs
+++ b/egui_demo_lib/src/apps/demo/context_menu.rs
@@ -21,6 +21,7 @@ pub struct ContextMenus {
     show_axes: [bool; 2],
     allow_drag: bool,
     allow_zoom: bool,
+    allow_scroll: bool,
     center_x_axis: bool,
     center_y_axis: bool,
     width: f32,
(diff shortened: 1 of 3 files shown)


rukai:
@emilk I left this PR as a draft as requested by the `pull_request_template.md`, waiting for green CI before marking as ready for review.
However that requires you to approve CI for this PR first.
And I suspect you arent looking at this PR yet because its still a draft.

So I'll mark this as ready to review and maybe the template should be changed to prevent newbie PRs getting stuck like this in the future.

emilk:
Good call @rukai  - PR template updated in https://github.com/emilk/egui/commit/aa6a2bb73fc4e96064bdc17b9acf80767811a64b

emilk (approved):

egui_demo_lib/src/apps/demo/context_menu.rs
```diff
@@ -99,6 +101,7 @@ impl super::View for ContextMenus {
                     ui.end_row();
                     if ui.checkbox(&mut self.allow_drag, "Drag").changed()
                         || ui.checkbox(&mut self.allow_zoom, "Zoom").changed()
+                        || ui.checkbox(&mut self.allow_scroll, "Scroll").changed()
```
emilk, line 104: This should really be part of the plot demo instead, but that's for another PR

PR #1384, opened 2022-03-20
Expose more tessellator method

### LINKED ISSUE #1383: [epaint] Expose more shape tessellation methods
Currently, to use a `Tessellator`, a user has to use the `tessellate_shape` method, meaning he has to use the `Shape` enum.

This can be a problem because this method takes ownership of the `Shape` which might carry a non-copy type (such as a Mesh), which might force the user to clone unnecessarily. This can also be a problem for a user that doesn't need to use a `Shape` altogether.

A solution to this problem would be to expose more of the `Tessellator` internal methods (like the already public tessellate_text) and have `tessellate_mesh` take a reference instead.

jean-airoldie:
* Make public the Tessellator methods to tessellate a circle, a
    mesh, a rectangle, a line, a path, a quadratic and cubic
    bezier curve.
* Add doc to tessellate_text & make it take a reference.
* Add Mesh::append_ref method.

## Notes
* I'm not too happy with the `Mesh::append_ref` method name, but I couldn't think of anything else.
* I'm not exactly sure what should be the behavior of `Mesh::append_ref`
* This is a breaking change because tessellate_text now takes a reference. A non-breaking alternative would be to make it take a `AsRef<TextShape>`.

Closes #1383

### DIFF
--- a/egui_demo_lib/benches/benchmark.rs
+++ b/egui_demo_lib/benches/benchmark.rs
@@ -129,7 +129,7 @@ pub fn criterion_benchmark(c: &mut Criterion) {
         let font_image_size = fonts.font_image_size();
         c.bench_function("tessellate_text", |b| {
             b.iter(|| {
-                tessellator.tessellate_text(font_image_size, text_shape.clone(), &mut mesh);
+                tessellator.tessellate_text(font_image_size, &text_shape, &mut mesh);
                 mesh.clear();
             })
         });
--- a/epaint/src/mesh.rs
+++ b/epaint/src/mesh.rs
@@ -91,16 +91,28 @@ impl Mesh {
         if self.is_empty() {
             *self = other;
         } else {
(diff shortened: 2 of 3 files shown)


jean-airoldie:
A side effect of not having to clone is that it improves the `egui_demo_lib::tessellate_text benchmark`.
## before
```
tessellate_text         time:   [68.282 ns 69.360 ns 70.408 ns]
                        change: [+83.723% +86.104% +88.444%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

## after
```
tessellate_text         time:   [36.897 ns 37.308 ns 37.777 ns]
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) low mild
  4 (4.00%) high mild
  4 (4.00%) high severe
```

emilk (approved):
Nice!

Issue #1379, opened 2022-03-19
Changes to the repaint signal causes problems with multi-threaded code


DusterTheFirst:
Issue created from comment: https://github.com/emilk/egui/commit/c768d1d48e4c8b7620391f4b732ff836bc17adad#r68942300
@quietvoid

---

Removing `frame.request_repaint` causes problems with calling `ctx.request_repaint` on another thread that did not exist when calling on `frame`.

`Frame` always uses a `std::sync::Mutex` to lock its state which makes this code fine to call from another thread as it will just block if the mutex is locked. `Context` on the other hand switches between using an `AtomicRefCell` and a `parking_lot::Mutex` depending on if `epaint/multi_threaded` is enabled (which it normally is not when using eframe. So when `ctx.request_repaint` is called from another thread and the ui thread has a lock on the `Context` the application just panics with the following message:

```
The application panicked (crashed).
Message:  already mutably borrowed
Location: C:\Users\ME\.cargo\registry\src\github.com-1ecc6299db9ec823\atomic_refcell-0.1.8\src\lib.rs:126
```

The solution to this would be to just have a way to enable or enable by default the `multi_threaded` feature on epaint. But it would leave this discrepency between different parts of code using `std::sync::Mutex`, `parking_lot::Mutex` and `AtomicRefCell` for depending on features or just what part of the code they are in. Standardizing all of them to use `parking_lot`'s synchronization would be my suggestion, but again the `Cargo.toml` for `epaint` does note a minor performance impact when not using `AtomicRefCell` in a single threaded context: 

https://github.com/emilk/egui/blob/c8f6cae3624e2c1303f7d7ca0158ae3b07b8ac15/epaint/Cargo.toml#L52-L54

quietvoid:
I modified the `eframe` `download_image` example to reproduce this:
```diff
diff --git a/eframe/examples/download_image.rs b/eframe/examples/download_image.rs
index 9cbb51d1..c6471014 100644
--- a/eframe/examples/download_image.rs
+++ b/eframe/examples/download_image.rs
@@ -36,17 +36,39 @@ impl eframe::App for MyApp {
             promise
         });
 
-        egui::CentralPanel::default().show(ctx, |ui| match promise.ready() {
-            None => {
-                ui.add(egui::Spinner::new()); // still loading
-            }
-            Some(Err(err)) => {
-                ui.colored_label(egui::Color32::RED, err); // something went wrong
+        let mut new_promise = None;
+
[...]

PR #1389, opened 2022-03-20
Put repaint callback and tex manager behind separate mutexes

emilk:
Closes https://github.com/emilk/egui/issues/1379

Please take a look @DusterTheFirst 

This fattens up `Context` so that the repaint callback and texture manager can be used from any thread, even when you are _not_ compiling with the `multi_threaded` feature flag:

``` rust
pub struct Context {
    ctx: Arc<RwLock<ContextImpl>>,
    repaint_info: Arc<RepaintInfo>,
    tex_manager: WrappedTextureManager,
}
```

This is one solution, but I'm not sure it is the best one.

The solution in this PR has the added advantage of mitigating https://github.com/emilk/egui/pull/1380

---
Performance: the e2e tests shows very little change, but some specific tests (`Painter::rect` and `label format!`) shows 10% slowdowns.

---

I think the simpler solution is to remove the `single_threaded/multi_threaded` features and to always use proper mutexes, as proposed in:

* https://github.com/emilk/egui/pull/1390


DusterTheFirst:
Do you have an idea of how these changes caused such an observable slow down in those specific test cases?

Issue #1271, opened 2022-02-19
Detecting click without sensing dragged first


shouya:
Hi, I'm working on a custom widget that senses both click and drag. I want to be able to clearly distinguish the two actions, however I find it somewhat difficult to do.

The `response.clicked` is always a few frames later than `response.drag_started`. I think I understand why this happens - dragging detection doesn't require mouse button up event, and is thus detected much earlier. The problem is that this makes the "drag_started" check unreliable - it can either mean a real "drag_started", or an unfinished "click". There is no way to tell the two apart when `response.drag_started == true`.

What I think may be ideal is to delay the report for `drag_started` until it is impossible a click can be interpreted.

I kludged some dirty code as follows, and it works for my intention fairly well:

```rust
enum Interaction {
  Clicked,
  // origin: Pos2
  DragStarted(egui::Pos2),
  DragReleased,
  Dragged,
}

// https://docs.rs/egui/latest/src/egui/input_state.rs.html#11-15
const MAX_CLICK_DIST: f32 = 6.0;
const MAX_CLICK_DURATION: f64 = 0.6;

fn detect_interaction(response: &Response) -> Option<Interaction> {
  use Interaction::*;
  #[derive(Clone)]
  struct ClickDragDetectionFinished(bool);

  let mut memory = response.ctx.memory();
  let click_detection_finished: &mut bool = &mut memory
    .data
    .get_temp_mut_or(response.id, ClickDragDetectionFinished(false))
    .0;

  let pointer = &response.ctx.input().pointer;
  if !pointer.any_down() {
    *click_detection_finished = false;
  }

  if !*click_detection_finished && response.clicked() {
    *click_detection_finished = true;
    return Some(Clicked);
  }

  if response.drag_released() {
    return Some(DragReleased);
  }

  if !*click_detection_finished && response.dragged() {
    let origin = pointer.press_origin().unwrap();
    if let Some(pos) = pointer.hover_pos() {
      let dx = (pos - origin).length_sq();
      if dx > MAX_CLICK_DIST * MAX_CLICK_DIST {
        *click_detection_finished = true;
        return Some(DragStarted(origin));
      }
    }

    let dt = response.ctx.input().time - pointer.press_start_time().unwrap();
    if dt > MAX_CLICK_DURATION {
      *click_detection_finished = true;
      return Some(DragStarted(origin));
    }

    return None;
  }

  if *click_detection_finished && response.dragged() {
    return Some(Dragged);
  }

  None
}
```

[...]

PR #1396, opened 2022-03-21
Add example of custom window frame for native window using eframe

emilk:
Often requested. There are many ways to accomplish it. This is the one that gives the most control, by custom-painting everything.

Looks like this:

Screen Shot 2022-03-21 at 22 05 19

Screen Shot 2022-03-21 at 22 05 16

### DIFF
--- a/eframe/examples/custom_window_frame.rs
+++ b/eframe/examples/custom_window_frame.rs
@@ -1,0 +1,117 @@
+//! Show a custom window frame instead of the default OS window chrome decorations.
+
+#![cfg_attr(not(debug_assertions), windows_subsystem = "windows")] // hide console window on Windows in release
+
+use eframe::egui;
+
+fn main() {
+    let options = eframe::NativeOptions {
+        // Hide the OS-specific "chrome" around the window:
+        decorated: false,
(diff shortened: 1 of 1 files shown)


zohnannor:
I works very inconsistently for some reason. When I touch an X button or move the window, theme switching buttons stop working. X button doesn't work either.
Sorry, should I open a new issue referencing this PR instead?

Animation1

emilk:
@zohnannor sounds like it could be a problem with `winit::Window::drag_window`. Have you tried running https://github.com/rust-windowing/winit/blob/master/examples/drag_window.rs ?

zohnannor:
@emilk sorry for late response. I tried it just now at it works as expected, I can drag windows by clicking anywhere in the window and choose drag target by pressing "x" on keyboard.

> sounds like it could be a problem with `winit::Window::drag_window`

Is this a problem really? In the egui case, I can't even close the window, clicking "X" just drags the whole window.

emilk:
For me, clicking X closes the window, but dragging the X drags the window. Not the best behavior perhaps, but consistent with how button in `egui::Window` work. Related: https://github.com/emilk/egui/issues/1271

PR #1397, opened 2022-03-22
eframe vs glow refactor

emilk:
The purpose of this is to allow accessing the glow painter (via a new `trait Renderer`) from the `App::update` function.


emilk:
Closed in favor of https://github.com/emilk/egui/pull/1546

PR #1398, opened 2022-03-22
Add example using `three-d`

emilk:
https://github.com/asny/three-d recently merged a PR adding `glow` support: https://github.com/asny/three-d/pull/210.
This means it is a prime candidate for embedding 3D painting inside an eframe app.

Screen Shot 2022-03-22 at 10 27 24

There are currently a few kinks that need to be figured out:

* [x] sudden egui paint failure
* [x] viewport
* [ ] anti-aliasing
* [ ] `Send+Sync`

## Sudden egui paint failure
When reusing the same `three_d` context over time (as one should), we get two(!) good frames of `egui+three_d`, but frame three and onwards the `egui` painting breaks (nothing is painted), but the `three_d` still works. Very weird!

**EDIT**: fixed by https://github.com/asny/three-d/commit/07bbf3b6e3336b758d7dbdfcc14c72f95f361376

## Viewport
The `thee-d` camera requires a viewport (in pixels). This need to be passed in to the `PaintCallback` function, or there needs to be some way of telling `three-d` to use the currently set viewport (less likely).

**EDIT**: fixed by https://github.com/emilk/egui/pull/1398/commits/9181dc7a0e52c48fbdc2648d4b328d95dcaac051

## Anti-aliasing
It would be nice to enable MSAA to smooth out the edges.

## `Shape: Send + Sync`
`Shape` is `Send + Sync` and `three_d::Context` is not. This means we cannot store a three_d context and send it to the `Shape::Callback`.

So we either need to recreate the three_d context each frame (obviously a bad idea), or access it through a `thread_local` hack. This PR adds both as examples, with a checkbox to switch.

We could consider making `Shape: !Send + !Sync`, but that has a lot of downstream implications, as discussed in https://github.com/emilk/egui/issues/1399

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,6 +82,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "approx"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3f2a05fd1bd10b2527e20a2cd32d8873d115b8b39fe219ee25f42a8aca6ba278"
+dependencies = [
+ "num-traits",
+]
+
+[[package]]
 name = "arrayref"
 version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


emilk:
Rebased and split into https://github.com/emilk/egui/pull/1406 and https://github.com/emilk/egui/pull/1407

PR #1405, opened 2022-03-22
Add `ComboBox::icon()`

4JX:
Inspired by and based on the commit for `CollapsingHeader::icon()`. Decided to take the whole `&Ui` rather than `&Painter` for flexibility.

### DIFF
--- a/egui/src/containers/combo_box.rs
+++ b/egui/src/containers/combo_box.rs
@@ -1,6 +1,9 @@
 use crate::{style::WidgetVisuals, *};
 use epaint::Shape;
 
+/// A function that paints the `ComboBox` icon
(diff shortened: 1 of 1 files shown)


emilk (changes requested):
Cool! Would it perhaps make sense to also give a `opened: bool` argument, so users can paint different icons when it is opened and closed?

4JX:
> Cool! Would it perhaps make sense to also give a `opened: bool` argument, so users can paint different icons when it is opened and closed?

Seems like a good idea, added in the latest commit

emilk (approved):

emilk:
Thanks!

PR #1412, opened 2022-03-23
Improve text contrast in bright mode

### LINKED ISSUE #1410: Bright text is too faint/weak
Compare white text on black with black text on white:

Screen Shot 2022-03-23 at 13 40 08 Screen Shot 2022-03-23 at 13 40 13

Even though the text colors are `#fffff` and `#00000` they do not appear as such. `ab_glyph` computes per-pixel coverage values, and no pixels in the text are fully covered by the glyph. This makes the text looks a bit too faint in both cases.

This problem is much more pronounced on the bright text. The effect of this is that the bright text looks fainter, thinner and weaker than the black text. This just makes bright mode worse overall in egui.

It is possible this could be solved by either by switching to gamma-blending, or by tweaking how we interpret/store the coverage values we get from ab_glyph.

emilk:
Closes #1410

This PR changes how the text coverage values (from `ab_glyph`) are interpreted and translated to `Color32` values in the font texture. The effect is that dark text on bright background appear darker, and therefore thicker.

The change in dark mode is minimal, but in bright mode it makes a big difference:

## Before (bright):
before-bright

## After (bright):
after-bright

## Before (dark):
before-dark

## After (dark):
after-dark

### DIFF
--- a/epaint/src/image.rs
+++ b/epaint/src/image.rs
@@ -6,7 +6,7 @@ use crate::Color32;
 ///
 /// In order to paint the image on screen, you first need to convert it to
 ///
-/// See also: [`ColorImage`], [`AlphaImage`].
+/// See also: [`ColorImage`], [`FontImage`].
 #[derive(Clone, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub enum ImageData {


felixfaire:
I appreciate that legible text in light mode is definitely a priority, though i find the solution of hacking the blending to make the thin font work (and look worse for all other fonts in the process) a bit extreme.. I think proper solutions to this might include.

1. Using an appropriately weighted font for the default font size (Ubuntu regular for example).
2. Using a darker colour for text in light mode (it's currently set at a relatively light grey).
3. Making the default font size bigger.

For me (mostly on a macbook) the text is perfectly legible on this PR in light mode (even more-so with black as the font color vs grey). Though i appreciate this isn't for everyone. Perhaps the defaults of egui should change size / color to be more accessible to more users who have an issue with correct blending, then there is always the option to make the font smaller / lighter for users who need the space / density.

I don't consider this issue resolved / PR closed.. So please reconsider looking over this as a serious issue (with solutions outlined above).

Issue #1377, opened 2022-03-19
Use full browser width by default


rukai:
At first I assumed that egui was limited to a hardcoded width in browsers (why else wouldnt it use all the available space?)
After noticing the max width slider in the demo I followed the code to find where this was set: https://github.com/emilk/egui/blob/c69f39e8697c56d50b1f53199abea9122573353b/epi/src/lib.rs#L184
There I found that the reason for setting the max width is to work around this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1010527#c0 
The bug was closed 5 months ago so does that mean egui should be changed to use the full browser width by default?

Edit: this should probably be a feature request...

emilk:
Oh wow, you are right - it has been fixed! The perf in firefox is now great on my Mac 😃 

I'll change this ASAP!

PR #1413, opened 2022-03-23
Remove FAQs for resolved issues

DusterTheFirst:
This PR removes the [**Why is `egui_web` using so much CPU in Firefox?**](https://github.com/emilk/egui#why-is-egui_web-using-so-much-cpu-in-firefox) and [**Why does my web app not fill the full width of the screen?**](https://github.com/emilk/egui#why-does-my-web-app-not-fill-the-full-width-of-the-screen) FAQs from the readme as it was noted that the problems with Firefox on Linux/MacOS have been resolved in #1377 and the limitation on the canvas size was lifted in commit 465c96122cf6aea72e7fbd8d6e7c2c048e88424f


Titaniumtown (approved):
LGTM

emilk (approved):
Good catch!

PR #1419, opened 2022-03-25
Add more clickable struct names in comments

4JX:
Replaced all instances found through the following regex:
```regex
([^[])(`(?!Arc)(?!Rc)(?!PartialEq)(?!Clone)(?!Some)(?!None)(?!MacOS)(?!EasyMark)(?!Err)(?!Home)(?!End)(?!Trackpad)[A-Z][a-z]{1}\w*`)
```

Tried to keep it conservative as to not make false positives, haven't found any wrong replacements myself.

Example of a clickable struct name:


emilk:
Thank you for this!

Unfortunately it is not enough to just enclose in `[]`, they also need to be in scope, or have a qualified path to them.

You can see the problem by running

``` sh
cargo doc -p emath -p epaint -p egui -p eframe -p epi -p egui_web -p egui-winit -p egui_glium -p egui_glow --lib --no-deps --all-features
…
warning: unresolved link to `PathShapes`
  --> epaint/src/bezier.rs:57:56
   |
57 |     /// Convert the cubic Bézier curve to one or two [`PathShapes`].
   |                                                        ^^^^^^^^^^ no item named `PathShapes` in scope
   |
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
   = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
```

(unfortunately I haven't figured out how to get `cargo doc` to fail on warnings to that the CI can catch this)

4JX:
> Thank you for this!
> 
> Unfortunately it is not enough to just enclose in `[]`, they also need to be in scope, or have a qualified path to them.
> 
> You can see the problem by running
> 
> ```shell
> cargo doc -p emath -p epaint -p egui -p eframe -p epi -p egui_web -p egui-winit -p egui_glium -p egui_glow --lib --no-deps --all-features
> …
> warning: unresolved link to `PathShapes`
>   --> epaint/src/bezier.rs:57:56
>    |
> 57 |     /// Convert the cubic Bézier curve to one or two [`PathShapes`].
>    |                                                        ^^^^^^^^^^ no item named `PathShapes` in scope
>    |
>    = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
>    = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
> ```
> 
> (unfortunately I haven't figured out how to get `cargo doc` to fail on warnings to that the CI can catch this)

Whoops, I assumed they were working as VSCode did correctly place links on them. Will fix it.

Mingun:
> (unfortunately I haven't figured out how to get `cargo doc` to fail on warnings to that the CI can catch this)

Maybe adding `#![deny(rustdoc::broken_intra_doc_links)]` at the crate root solve this?

egui/src/data/input.rs
```diff
@@ -192,7 +192,7 @@ pub enum Event {
     ///
     /// This means there is no longer a cursor on the screen for hovering etc.
     ///
-    /// On touch-up first send `PointerButton{pressed: false, …}` followed by `PointerLeft`.
+    /// On touch-up first send `PointerButton{pressed: false, …}` followed by [`PointerLeft`].
```
4JX, line 195: What does `PointerLeft` refer to? I can't seem to find it anywhere.

epaint/src/color.rs
```diff
@@ -828,7 +828,7 @@ impl From<Hsva> for HsvaGamma {
 // ----------------------------------------------------------------------------
 
 /// Cheap and ugly.
-/// Made for graying out disabled `Ui`:s.
+/// Made for graying out disabled [`Ui`]:s.
```
4JX, line 831: How should I qualify this one? Since the `egui` crate itself is not imported here I seem to be unable to please cargo without importing it.

emilk: I don't think `epaint` can point to `egui`, since it is not a dependency except by linking explicitly to https://docs.rs/egui/latest/egui/struct.Ui.html

4JX:
One fight with github's merging later this should be ready for merging with no more broken links.

4JX:
Spoke too soon, that one does not show on my machine but it does on CI. Maybe caused by #1125 stopping the build from completing.

emilk:
Wow, you are amazing for doing all this work! There are still 12 broken links with `cargo doc --document-private-items --no-deps --all-features` though:

```
warning: unresolved link to `Ui`
  --> egui/src/layout.rs:10:43
   |
10 |     /// This is the minimal size of the [`Ui`].
   |                                           ^^ no item named `Ui` in scope
   |
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
   = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
…
```

emilk:
Thanks ❤️

PR #1425, opened 2022-03-27
Add glow::Context to epi::Frame

emilk:
This can be used, for instance, to:

* Render things to offscreen buffers.
* Read the pixel buffer from the previous frame (`glow::Context::read_pixels`).
* Render things behind the egui windows.

Note that all egui painting is deferred to after the call to `App::update` (`egui` only collects `egui::Shape`s and then eframe paints them all in one go later on).


BernardBorio:
Hi, I'm trying to export the painter_layer content as an image, and I've seen you made this function to read the pixels but I have no idea how to use it, can you explain me a bit? It could be great 🙏🏼

Issue #1391, opened 2022-03-21
ScrollArea::show_rows() does not scroll smoothly with very large total_rows


parasyte:
**Describe the bug**

I have a UI with about 16 million rows (but this is on the low end of what I want to support). I'm using `ScrollArea::show_rows()` to keep a sane viewport size.

Up to about 2 million rows, scrolling with the trackpad or by dragging the background with a mouse is relatively smooth. Above 2 million total rows, scrolling has some obvious hitching. Above 100 million total rows, the scrolling gets very broken.

**To Reproduce**
1. Patch the demo lib to increase the `num_rows` to `16_000_000`: https://github.com/emilk/egui/blob/ccbddcfe951e01c55efd0ed19f2f2ab5edfad5d9/egui_demo_lib/src/apps/demo/scrolling.rs#L85
2. Run the demo and go to the "Scolling" demo and then the "Scroll a lot of lines" tab.
3. Scroll with the trackpad or by dragging the background with the mouse.
4. You'll see it does not scroll smoothly.
5. Increase `num_rows` to `160_000_000` and try the same test.
6. Scrolling is almost completely unusable with this many rows.

**Expected behavior**

Always scroll smoothly regardless of the total number of rows.

**Desktop (please complete the following information):**
 - OS: Windows 11, macOS 12.2.1

emilk:
Yup, `f32` will cause trouble here. Seems like we need to switch to `f64` in `ScrollArea::show_rows`, which should be pretty straight-forward.

PR #1444, opened 2022-04-03
egui_extras: enable virtual scroll for heterogenous rows

waynr:
Introduce `TableBody.heterogenous_rows` and `TableRowBuilder`, the former of which takes as an argument the latter. `TableRowBuilder` provides two methods that enable virtual scrolling for rows with non-uniform heights. Those methods are:

* `TableRowBuilder.row_heights` which returns an iterator over `f32` to allow incremental virtual scroll buffer calculation.
* `TableRowBuilder.populate_row` which `TableBody.heterogenous_rows` uses to allow `TableRowBuilder` implementations to, you guessed it, populate rows that are visible.

One thought that occurs to me while writing this description is that `TableBody.heterogenous_rows` could look more like the following:

```
    pub fn heterogenous_rows(
        mut self,
        row_heights: impl Iterator<Item = f32> + '_,
        mut row: impl FnMut(usize, TableRow<'_, '_>),
    )
```

This could potentially be easier to use, considering all the trouble I had coming up with and implementing the trait. Happy to make this change if the maintainers prefer.

### DIFF
--- a/egui_demo_lib/src/apps/demo/table_demo.rs
+++ b/egui_demo_lib/src/apps/demo/table_demo.rs
@@ -30,9 +30,15 @@ impl super::Demo for TableDemo {
 
 impl super::View for TableDemo {
     fn ui(&mut self, ui: &mut egui::Ui) {
+        let mut settings_height = 44.0;
+        if self.virtual_scroll {
+            settings_height = 66.0;
+        } else {
+            self.heterogeneous_rows = false
+        }
         // Leave room for the source code link after the table demo:
         StripBuilder::new(ui)
-            .size(Size::exact(66.0)) // for the settings
+            .size(Size::exact(settings_height)) // for the settings
             .size(Size::remainder()) // for the table
             .size(Size::exact(10.0)) // for the source code link
             .vertical(|mut strip| {
@@ -42,16 +48,20 @@ impl super::View for TableDemo {
                         .size(Size::remainder())
                         .horizontal(|mut strip| {
                             strip.cell(|ui| {
-                                ui.checkbox(&mut self.heterogeneous_rows, "Heterogeneous rows");
                                 ui.checkbox(&mut self.virtual_scroll, "Virtual Scroll");
+                                if self.virtual_scroll {
+                                    ui.checkbox(&mut self.heterogeneous_rows, "Heterogeneous rows");
+                                }
                                 ui.checkbox(&mut self.resizable, "Resizable columns");
                             });
-                            strip.cell(|ui| {
-                                ui.add(
-                                    egui::Slider::new(&mut self.num_rows, 0..=300_000)
-                                        .text("Num rows"),
-                                );
-                            });
+                            if self.virtual_scroll {
+                                strip.cell(|ui| {
+                                    ui.add(
+                                        egui::Slider::new(&mut self.num_rows, 0..=300_000)
+                                            .text("Num rows"),
+                                    );
+                                });
+                            }
                         });
                 });
                 strip.cell(|ui| {
@@ -94,23 +104,26 @@ impl TableDemo {
                 });
             })
             .body(|mut body| {
-                if !self.heterogeneous_rows {
-                    body.rows(text_height, self.num_rows, |row_index, mut row| {
-                        row.col(|ui| {
-                            ui.label(row_index.to_string());
+                if self.virtual_scroll {
+                    if !self.heterogeneous_rows {
+                        body.rows(text_height, self.num_rows, |row_index, mut row| {
+                            row.col(|ui| {
+                                ui.label(row_index.to_string());
+                            });
+                            row.col(|ui| {
+                                ui.label(clock_emoji(row_index));
+                            });
+                            row.col(|ui| {
+                                ui.add(
+                                    egui::Label::new("Thousands of rows of even height")
+                                        .wrap(false),
+                                );
+                            });
                         });
-                        row.col(|ui| {
-                            ui.label(clock_emoji(row_index));
-                        });
-                        row.col(|ui| {
-                            ui.add(
-                                egui::Label::new("Thousands of rows of even height").wrap(false),
-                            );
-                        });
-                    });
-                } else if self.virtual_scroll {
-                    let rows = DemoRows::new(self.num_rows);
-                    body.heterogeneous_rows(rows, DemoRows::populate_row);
+                    } else {
+                        let rows = DemoRows::new(self.num_rows);
+                        body.heterogeneous_rows(rows, DemoRows::populate_row);
+                    }
                 } else {
                     for row_index in 0..20 {
                         let thick = row_index % 6 == 0;
--- a/egui_demo_lib/src/apps/demo/table_demo.rs
+++ b/egui_demo_lib/src/apps/demo/table_demo.rs
@@ -5,8 +5,9 @@ use egui_extras::{Size, StripBuilder, TableBuilder, TableRow, TableRowBuilder};
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 #[derive(Default)]
 pub struct TableDemo {
-    virtual_scroll: bool,
+    heterogeneous_rows: bool,
     resizable: bool,
+    num_rows: usize,
 }
 
 impl super::Demo for TableDemo {
@@ -28,14 +29,29 @@ impl super::Demo for TableDemo {
 
 impl super::View for TableDemo {
     fn ui(&mut self, ui: &mut egui::Ui) {
-        ui.checkbox(&mut self.virtual_scroll, "Virtual scroll");
-        ui.checkbox(&mut self.resizable, "Resizable columns");
-
         // Leave room for the source code link after the table demo:
         StripBuilder::new(ui)
+            .size(Size::exact(50.0)) // for the settings
             .size(Size::remainder()) // for the table
             .size(Size::exact(10.0)) // for the source code link
             .vertical(|mut strip| {
+                strip.cell(|ui| {
+                    StripBuilder::new(ui)
+                        .size(Size::exact(150.0))
+                        .size(Size::remainder())
+                        .horizontal(|mut strip| {
+                            strip.cell(|ui| {
+                                ui.checkbox(&mut self.heterogeneous_rows, "Heterogeneous rows");
+                                ui.checkbox(&mut self.resizable, "Resizable columns");
+                            });
+                            strip.cell(|ui| {
+                                ui.add(
+                                    egui::Slider::new(&mut self.num_rows, 0..=300_000)
+                                        .text("Num rows"),
+                                );
+                            });
+                        });
+                });
                 strip.cell(|ui| {
                     self.table_ui(ui);
                 });
@@ -76,8 +92,8 @@ impl TableDemo {
                 });
             })
             .body(|body| {
-                if self.virtual_scroll {
-                    body.rows(text_height, 100_000, |row_index, mut row| {
+                if !self.heterogeneous_rows {
+                    body.rows(text_height, self.num_rows, |row_index, mut row| {
                         row.col(|ui| {
                             ui.label(row_index.to_string());
                         });
@@ -92,9 +108,9 @@ impl TableDemo {
                     });
                 } else {
                     let row_builder = DemoRowBuilder {
-                        row_count: 100000,
+                        row_count: self.num_rows,
                     };
-                    body.heterogenous_rows(row_builder);
+                    body.heterogeneous_rows(row_builder);
                 }
             });
     }
@@ -125,7 +141,7 @@ impl Iterator for DemoRows {
 
 impl TableRowBuilder for DemoRowBuilder {
     fn row_heights(&self, _: &Vec<f32>) -> Box<dyn Iterator<Item = f32>> {
-        Box::new(DemoRows{
+        Box::new(DemoRows {
             row_count: self.row_count,
             current_row: 0,
         })
--- a/egui_demo_lib/src/apps/demo/table_demo.rs
+++ b/egui_demo_lib/src/apps/demo/table_demo.rs
@@ -1,5 +1,5 @@
 use egui::TextStyle;
-use egui_extras::{Size, StripBuilder, TableBuilder, TableRow, TableRowBuilder};
+use egui_extras::{Size, StripBuilder, TableBuilder, TableRow};
 
 /// Shows off a table with dynamic layout
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
@@ -91,7 +91,7 @@ impl TableDemo {
                     });
                 });
             })
-            .body(|body| {
+            .body(|mut body| {
                 if !self.heterogeneous_rows {
                     body.rows(text_height, self.num_rows, |row_index, mut row| {
                         row.col(|ui| {
@@ -107,17 +107,11 @@ impl TableDemo {
                         });
                     });
                 } else {
-                    let row_builder = DemoRowBuilder {
-                        row_count: self.num_rows,
-                    };
-                    body.heterogeneous_rows(row_builder);
+                    let rows = DemoRows::new(self.num_rows);
+                    body.heterogeneous_rows(rows, DemoRows::populate_row);
                 }
             });
     }
-}
-
-struct DemoRowBuilder {
-    row_count: usize,
 }
 
 struct DemoRows {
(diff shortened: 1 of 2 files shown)


egui_demo_lib/src/apps/demo/table_demo.rs
```diff
@@ -75,7 +75,7 @@ impl TableDemo {
                     });
                 });
             })
-            .body(|mut body| {
+            .body(|body| {
                 if self.virtual_scroll {
```
waynr, line 79: note: technically both branches here use virtual scroll, so it might be helpful to come up with a different distinction here. while doing that it could also be helpful from a development standpoint to make the number of rows (in either case) configurable to make it easier to test edge cases

egui_extras/src/table.rs
```diff
@@ -351,7 +351,65 @@ pub struct TableBody<'a> {
     end_y: f32,
 }
 
+pub trait TableRowBuilder {
+    fn row_heights(&self, widths: &Vec<f32>) -> Box<dyn Iterator<Item = f32> + '_>;
```
waynr, line 355: note for reviewers: widths isn't used in the demo app, but I included it here because my main use case for the `Table` type is to implement a chat box where text will be wrapped based on width and so height will be tied to cell width.

actually, come to think of it this was the main reason i thought it would be good to use a trait rather than simply taking an iterator over `f32` without any input from `heterogenous_rows`.

waynr: if there were a `widths(&self) -> Vec<32>` method on `TableBody` then coming up with an iterator of floating point height values wouldn't explicitly require it as an input

egui_extras/src/table.rs
```diff
@@ -351,7 +351,65 @@ pub struct TableBody<'a> {
...
+        // cumulative height of all rows above those being displayed
+        let mut height_above_visible = 0.0;
+        // cumulative height of all rows below those being displayed
+        let mut height_below_visible = 0.0;
+
+        let mut row_index = 0;
+
+        let mut above_buffer_set = false;
+        let mut current_height = 0.0; // used to track height of visible rows
+        while let Some(height) = heights.next() {
```
waynr, line 375: note: just occurred to me that breaking this into three separate loops might make it a bit easier to reason about. can use breaks at the boundaries between non-visible and visible areas.

waynr:
For what it's worth, when I ran this in the demo app using 100_000 rows in the heterogeneous rows (same as in the homogeneous rows example) I noticed that CPU time / frame seemed to average around 9 or 10 ms on my computer vs the 4 or 5 of the homogeneous rows example. Of course this is many orders of magnitude better than the heterogeneous example at 100_000 rows prior to this PR...

egui_extras/src/table.rs
```diff
@@ -351,7 +351,65 @@ pub struct TableBody<'a> {
...
+            // when delta is greater than height above 0, we need to increment the row index and
+            // update the height above visble with the current height then continue
+            if height_above_visible < delta {
+                row_index += 1;
+                height_above_visible += height;
+                continue;
+            }
+
+            // if height above visible is > 0 here then we need to add a buffer to allow the table
+            // to calculate the "virtual" scrollbar position, reset height_above_visisble to 0 to
```
waynr, line 385: TODO: fix out of date comment

egui_demo_lib/src/apps/demo/table_demo.rs
emilk: I think it would be good to demo all three version - the simple version for short tables, the version with homogeneous row heights, and the new heterogeneous version

egui_extras/src/table.rs
```diff
@@ -351,7 +351,65 @@ pub struct TableBody<'a> {
...
+        let mut row_index = 0;
+
+        let mut above_buffer_set = false;
+        let mut current_height = 0.0; // used to track height of visible rows
+        while let Some(height) = heights.next() {
+            // when delta is greater than height above 0, we need to increment the row index and
+            // update the height above visble with the current height then continue
+            if height_above_visible < delta {
+                row_index += 1;
+                height_above_visible += height;
```
emilk, line 380: Accumulating `f32` will lead to precision issues really quickly (even worse than those [already present](https://github.com/emilk/egui/issues/1391)). Better to use `f64` for `height_above_visible`,  `height_below_visible` and `current_height`.

egui_extras/src/table.rs
```diff
@@ -412,6 +457,16 @@ impl<'a> TableBody<'a> {
 
         self.row_nr += 1;
     }
+
+    pub fn buffer(&mut self, height: f32) {
```
emilk, line 461: needs a docstring or a better name (`advance_height`?)

egui_extras/src/table.rs
```diff
@@ -351,7 +351,65 @@ pub struct TableBody<'a> {
     end_y: f32,
 }
 
+pub trait TableRowBuilder {
+    fn row_heights(&self, widths: &Vec<f32>) -> Box<dyn Iterator<Item = f32> + '_>;
+    fn populate_row(&self, index: usize, row: TableRow<'_, '_>);
+}
+
 impl<'a> TableBody<'a> {
+    pub fn heterogenous_rows(mut self, builder: impl TableRowBuilder) {
```
emilk, line 360: Needs a docstring with an example

egui_extras/src/table.rs
emilk: I agree that using a trait for this is quite cumbersome. It would be much nicer to use with just closures.

Adding `fn width(&self) -> &[f32]` to `TableBody` is easy enough.

emilk (changes requested):
This is a really good addition! But let's see if we can do it without the `trait`!

egui_extras/src/table.rs
waynr: Should the `height` values be given in `f64` then also? I don't know much about precision issues between different bit widths. But I did go ahead and make this change.

egui_extras/src/table.rs
waynr: I added a docstring and removed the `pub`, but `buffer` still makes more sense to me semantically -- but since you're the maintainer if you have a strong preference I'll go with whatever your suggestion is.

egui_extras/src/table.rs
emilk: No, `height` can be `f32`. It is the accumulator that needs to be wide.

egui_extras/src/table.rs
emilk: buffer is so many different things (including a verb). `add_buffer` or `add_space`?

egui_demo_lib/src/apps/demo/table_demo.rs
```diff
@@ -28,14 +30,30 @@ impl super::Demo for TableDemo {
 
 impl super::View for TableDemo {
     fn ui(&mut self, ui: &mut egui::Ui) {
-        ui.checkbox(&mut self.virtual_scroll, "Virtual scroll");
-        ui.checkbox(&mut self.resizable, "Resizable columns");
-
         // Leave room for the source code link after the table demo:
         StripBuilder::new(ui)
+            .size(Size::exact(66.0)) // for the settings
```
emilk, line 35: You should be able to just put the settings before the `StripBuilder`. We only need `StripBuilder` here because of the table uses `remainder`.

egui_demo_lib/src/apps/demo/table_demo.rs
```diff
@@ -28,14 +30,30 @@ impl super::Demo for TableDemo {
...
             .size(Size::exact(10.0)) // for the source code link
             .vertical(|mut strip| {
+                strip.cell(|ui| {
+                    StripBuilder::new(ui)
+                        .size(Size::exact(150.0))
+                        .size(Size::remainder())
+                        .horizontal(|mut strip| {
+                            strip.cell(|ui| {
+                                ui.checkbox(&mut self.heterogeneous_rows, "Heterogeneous rows");
+                                ui.checkbox(&mut self.virtual_scroll, "Virtual Scroll");
```
emilk, line 46: These are 2*2=4 options for three different demos. Perhaps use an enum and radio buttons, or only show `heterogeneous_rows` in case `virtual_scroll==true`

egui_extras/src/table.rs
```diff
@@ -352,24 +352,145 @@ pub struct TableBody<'a> {
 }
 
 impl<'a> TableBody<'a> {
+    fn delta(&self) -> f32 {
+        self.start_y - self.layout.current_y()
+    }
```
emilk, line 357: I think this is a weird name and a weird behavior - the _negative_ amount of y progress?

I think this would be better as

```suggestion
    fn y_progress(&self) -> f32 {
        self.layout.current_y() - self.start_y
    }
```

so that the returned value is positive (or am I misunderstanding it?)

egui_extras/src/table.rs
```diff
@@ -352,24 +352,145 @@ pub struct TableBody<'a> {
...
+        // cumulative height of all rows below those being displayed
+        let mut height_below_visible: f64 = 0.0;
+
+        let mut row_index = 0;
+
+        // calculate height above visible table range
+        while let Some((_, height)) = striped_heights.next() {
+            // when delta is greater than height above 0, we need to increment the row index and
+            // update the height above visble with the current height then continue
+            if height_above_visible >= delta as f64 {
```
emilk, line 440: I don't quite follow the logic of everything, but it seems like this line should be taking `height` into account. If the new row height is so wide that it should be visible, then we should `break` and paint it. I think if you do that, then we can remove `VIRTUAL_EXTENSION` completely.

egui_extras/src/table.rs
```diff
@@ -352,24 +352,145 @@ pub struct TableBody<'a> {
...
+            row_index += 1;
+            current_height += height as f64;
+        }
+
+        // calculate height below the visible table range
+        while let Some((_, height)) = striped_heights.next() {
+            height_below_visible += height as f64
+        }
+
+        // if height bloew visible is > 0 here then we need to add a buffer to allow the table to
```
emilk, line 476:
```suggestion
        // if height below visible is > 0 here then we need to add a buffer to allow the table to
```

emilk (changes requested):

egui_demo_lib/src/apps/demo/table_demo.rs
waynr: I don't know any other way to get a horizontal layout (checkbox options on the left, slider on the right). If I move the setting strip builder with the horizontal layout outside of the top-level strip builder than the whole table demo window automatically expands in height well below the bottom edge of the application window.

Anyway, if this is a demo what's the problem with demonstrating how to embed a horizontally-oriented strip inside the cell of a vertically-oriented strip?

egui_extras/src/table.rs
```diff
@@ -352,24 +352,145 @@ pub struct TableBody<'a> {
...
+    fn delta(&self) -> f32 {
+        self.start_y - self.layout.current_y()
+    }
+
+    /// Return a vector containing all column widths for this table body.
+    ///
+    /// This is primarily meant for use with [`TableBody::heterogeneous_rows`] in cases where row
+    /// heights are expected to according to the width of one or more cells -- for example, if text
+    /// is wrapped rather than clippped within the cell.
+    pub fn widths(&self) -> &Vec<f32> {
```
parasyte, line 364: This can return a slice.

```suggestion
    pub fn widths(&self) -> &[f32] {
```

waynr: I'm still fairly new to rust so I don't fully understand the signficance -- is there a reason to prefer one over the other?

egui_extras/src/table.rs
waynr: I could also be misunderstanding, but if you look at the way [delta was calculated prior to this PRd in `TableBody::rows`](https://github.com/emilk/egui/blob/5dff1e42c6ec8a3740c9cf0b18400743d73a0266/egui_extras/src/table.rs#L359-L363) the result was expected to be negative if any buffer spaces was needed above the visible range. My intent with this `delta` helper method was to clean that up a little and to reuse the calculation in `TableBody::heterogeneous_rows`.

So I agree with the intent of your feedback here (ie that it's easier to reason about a positive result), just not with the suggested change since to my understanding what I've written should always result in >= 0 value for delta. But my understanding is definitely limited since I don't have a good conceptual understanding of the difference between `self.start_y` and `self.layout.current_y()` nor even where the origin of the coordinate system in egui is. My guess is that `self.start_y` is the coordinate of the top of the visible rectangle and `self.layout.current_y()` is the top of the "virtual" rectangle pushed upward by the `TableRow`. Or something like that. All I know is that if I flip the subtraction operands around then I have to either negate the result or flip a bunch of comparison operands.

I'll definitely make the suggested name change though.

egui_extras/src/table.rs
waynr: While I think this suggestion is ultimately correct -- it solve the problem in the demo app where row index 0 isn't considered an "extra thick row" despite `0 % 6 == 0` (the logic that determines which rows should be thicker) -- it doesn't solve the same problem that I intended to solve with `VIRTUAL_EXTENSION`, which is that when scrolling up rows abruptly disappear rather than slide "under" the lip of the visible rectangle area. In theory even this `VIRTUAL_EXTENSION` constant may not work for sufficiently tall rows...

egui_demo_lib/src/apps/demo/table_demo.rs
waynr: I took the approach of only showing the `heterogeneous_rows` setting if `virtual_scroll==true`.

egui_extras/src/table.rs
parasyte: There a few reasons to return a slice. One is that it provides an abstraction that hides implementation details. For instance the list of widths could internally be stored in any collection which can Deref to slice and the type could be changed freely without changing the API.

Also if you look through the docs for `Vec` you'll see that it only implements very few methods that take `&self`, and the useful ones are also implemented on slice.

Lastly, it's just more idiomatic. In the same way you would pass a slice as an argument instead of a borrowed vector. A slice works with multiple types of collections.

waynr: That's super helpful to know, thanks for the explanation! I did make the suggested change fyi.

parasyte: Glad to help!

egui_extras/src/table.rs
```diff
@@ -352,24 +352,144 @@ pub struct TableBody<'a> {
...
+        // cumulative height of all rows above those being displayed
+        let mut height_above_visible: f64 = 0.0;
+        // cumulative height of all rows below those being displayed
+        let mut height_below_visible: f64 = 0.0;
+
+        // calculate height above visible table range
+        while let Some((row_index, striped, height)) = striped_heights.next() {
+            // when y_progress is greater than height above 0, we need to increment the row index
+            // and update the height above visble with the current height then continue
+            if height_above_visible >= y_progress as f64 {
```
emilk, line 438: I believe `VIRTUAL_EXTENSION` could be removed if we changed this logic to take `height` into account (`if height_above_visible + height as f64 >= …`).

(PS: I thought I already made this comment, but now I can't find it on GitHub…)

emilk:
You don't need to wait for CI - just run `./sh/check.sh` (https://github.com/emilk/egui/blob/master/CONTRIBUTING.md#making-a-pr)

waynr:
@emilk
```
<wayne@thing3:~/projects/egui>
zsh/3 4901 [101]  (git)-[enable-virtual-scroll-for-heterogeneous-rows]-% ./sh/check.sh
./sh/check.sh: 2: Bad substitution
+ cargo check --workspace --all-targets
error: could not find `Cargo.toml` in `/home/wayne/projects` or any parent directory
```

emilk:
> @emilk
> 
> ```
> <wayne@thing3:~/projects/egui>
> zsh/3 4901 [101]  (git)-[enable-virtual-scroll-for-heterogeneous-rows]-% ./sh/check.sh
> ./sh/check.sh: 2: Bad substitution
> + cargo check --workspace --all-targets
> error: could not find `Cargo.toml` in `/home/wayne/projects` or any parent directory
> ```

Oh, I guess bash scripts aren't very portable 😭

waynr:
> Oh, I guess bash scripts aren't very portable 

Technically these scripts aren't using bash unless the user has `sh` linked to `/bin/bash`: https://github.com/emilk/egui/blob/68d5806b41725fb8f8a5e62d7acd94b5ac9140f0/sh/check.sh#L1

emilk (approved):

emilk:
Thanks!

PR #1471, opened 2022-04-09
Add to_hex_str convenience methods for Rgba and Color32

grantshandy:
This is really a pretty simple convenience method for `Rgba` and `Color32`, it creates a standard lowercase formatted hex color.

```rust
pub fn to_hex_str(&self) -> String {
    format!(
        "#{:02x}{:02x}{:02x}",
        self.0[0] as i32, self.0[1] as i32, self.0[2] as i32
    )
}
```

Here's what it would look like in action:
```rust
assert_eq!(
    Rgba::from_rgb(92.0, 117.0, 51.0).to_hex_str(),
    "#5c7533".to_string()
);

assert_eq!(
    Color32::from_rgb(174, 81, 164).to_hex_str(),
    "#ae51a4".to_string()
);
```

I don't know if this is out of scope for `Rgba` and `Color32`, but it seems pretty useful to me :)

### DIFF
--- a/epaint/src/color.rs
+++ b/epaint/src/color.rs
@@ -154,6 +154,15 @@ impl Color32 {
         Rgba::from(self).to_opaque().into()
     }
 
+    // Returns a hex string representing the rgb value
+    #[inline(always)]
+    pub fn to_hex_str(&self) -> String {
+        format!(
+            "#{:02x}{:02x}{:02x}",
+            self.0[0] as i32, self.0[1] as i32, self.0[2] as i32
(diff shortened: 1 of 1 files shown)


emilk:
This only works for opaque colors, which leaves room for way too many bugs. Taking alpha into account will be difficult, since egui uses premultiplied alpha.

Perhaps you should take a look at https://crates.io/crates/csscolorparser instead?

grantshandy:
Thanks 👍🏽

PR #1498, opened 2022-04-15
Add line breaking rules for Japanese text

### LINKED ISSUE #1497: Japanese word wrapping is weird
**Describe the bug**
Japanese text wraps at weird points of sentences.

image

**To Reproduce**
Display the following text in egui:

> ゲームを下記より選択してください。
>
> リストが空の場合は、「roms」ディレクトリにROMファイルを、「saves」ディレクトリにセーブファイルを配置してください。

**Expected behavior**
Wrapping that looks more like this:

> ゲームを下記より選択してください。
>
> リストが空の場合は、「roms」ディレクトリにROMファイル
> を、「saves」ディレクトリにセーブファイルを配置してくだ
> さい。

**Screenshots**
See above.

 - OS: macOS
 - Browser N/A
 - Version 0.17.0

bigfarts:
This allows line breaks after any kana, unless they are immediately followed by a gyōtō kinsoku character, in which case the line may not break at that point.

Also pedantically renamed `is_chinese` to `is_cjk_ideograph` as `is_chinese` will also cause line breaks on Japanese kanji.

Fixes #1497.

### DIFF
--- a/epaint/src/text/text_layout.rs
+++ b/epaint/src/text/text_layout.rs
@@ -194,8 +194,8 @@ fn line_break(
     let mut row_start_idx = 0;
     let mut non_empty_rows = 0;
 
-    for (i, glyph) in paragraph.glyphs.iter().enumerate() {
-        let potential_row_width = glyph.max_x() - row_start_x;
+    for i in 0..paragraph.glyphs.len() {
+        let potential_row_width = paragraph.glyphs[i].max_x() - row_start_x;
 
         if job.wrap.max_rows > 0 && non_empty_rows >= job.wrap.max_rows {
             break;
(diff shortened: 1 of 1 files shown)


epaint/src/text/text_layout.rs
```diff
@@ -194,8 +194,8 @@ fn line_break(
     let mut row_start_idx = 0;
     let mut non_empty_rows = 0;
 
-    for (i, glyph) in paragraph.glyphs.iter().enumerate() {
-        let potential_row_width = glyph.max_x() - row_start_x;
+    for (i, glyphs) in (0..paragraph.glyphs.len()).map(|i| &paragraph.glyphs[i..]).enumerate() {
+        let potential_row_width = glyphs[0].max_x() - row_start_x;
```
emilk, line 198: This is a bit of a mouthful. I would suggest something simpler:

```suggestion
    for i in ..paragraph.glyphs.len() {
        let potential_row_width = paragraph.glyphs[i].max_x() - row_start_x;
```

And then use `row_break_candidates.add(i, &paragraph.glyphs[i..]);` later

bigfarts: Done, thanks!

emilk:
Thanks!

Issue #1500, opened 2022-04-15
Unresponsive browser resizing


emilk:
Resizing the browser window (Brave or Firefox, Mac) will not resize the egui canvas immediately, making resizing feel VERY laggy.

`git bisect` found out it was caused by https://github.com/emilk/egui/pull/1306/

DusterTheFirst:
It is very curious that the changes would affect performance, especially so much. The main change is the addition of an atomic Boolean which is slower than non-atomic operations, but I wouldn't see that causing a 10x slow down.

emilk:
No, it must be something far more serious, like the event not being registered or the `needs_repaint` flag not being set or read successfully.

DusterTheFirst:
I think I found the culprit, It looks like the event handlers meant for the html element do not get registered. It can be seen in the screenshots that the canvas and the input elements have their event handlers registered, while the html element has no event handlers. 

Before #1306
![image](https://user-images.githubusercontent.com/14093962/163632516-e21bd8fc-5c0b-4636-b2f3-3d177e18cc4d.png)

After #1306
![image](https://user-images.githubusercontent.com/14093962/163632531-985608ab-5aab-4b6f-a8b7-ccacc7d2f4d9.png)

kud1ing:
Moving the windows in the web demo has always been fast. To my surprise now on macOS it is very slow. It's not good in Firefox and unbearable in Safari. Can this be related?

emilk:
That sounds unrelated @kud1ing - please file an issue if you are having performance issues in a browser. Please note the difference between `egui::Window`, the browser window and the native `winit::Window`.

kud1ing:
This may be only on my system. It's better after a reboot. Not sure why it's shows up now and only in the web demo.
I've been toying with it for a long time. Sorry for the probably false alarm.

PR #1503, opened 2022-04-15
[egui_web] Fix misplaced resize and other event handlers

DusterTheFirst:
Fixes copy-paste error introduced in #1306 by putting events meant for the window back on the window instead of on the document.

Closes <https://github.com/emilk/egui/issues/1500>.


emilk:
Nicely done; thanks!

Issue #408, opened 2021-05-17
Setting native window position


TimUntersberger:
**Is your feature request related to a problem? Please describe.**

I need to be able to set the window position of the window.

**Describe the solution you'd like**

Adding a `window_position` field of type `Vec2` to the `NativeOptions` struct should be enough.

emilk:
Sounds good! PR welcome.

(It should be `Pos2` rather than `Vec2` though)

emilk:
`NativeOptions` has an `initial_window_pos`, but I'll keep this issue open for adding a `epi::Frame::set_window_pos` too (for changing window position at runtime).

PR #1505, opened 2022-04-16
added `epi::Frame::set_window_pos`

bobyclaws:
this allows setting the position of window at runtime

Closes <https://github.com/emilk/egui/issues/408>.


emilk (approved):
Great!

PR #1512, opened 2022-04-18
Add triple-click support

trevyn:
Straightforward extension of double-click.

Line / paragraph selection in TextEdit fields is the same as word selection logic with a different character filter. There's quite a bit of copy/paste there, let me know if you want a deeper refactoring.

Thanks for egui, it's awesome!!

### DIFF
--- a/egui/src/widgets/text_edit/builder.rs
+++ b/egui/src/widgets/text_edit/builder.rs
@@ -1231,15 +1231,15 @@ fn select_line_at(text: &str, ccursor: CCursor) -> CCursorRange {
         let mut it = it.skip(ccursor.index - 1);
         if let Some(char_before_cursor) = it.next() {
             if let Some(char_after_cursor) = it.next() {
-                if is_line_char(char_before_cursor) && is_line_char(char_after_cursor) {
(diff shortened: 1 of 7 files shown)


egui/src/widgets/text_edit/builder.rs
```diff
@@ -1250,10 +1308,32 @@ fn next_word_boundary_char_index(it: impl Iterator<Item = char>, mut index: usiz
...
+        }
+    }
+    index
+}
+
 fn is_word_char(c: char) -> bool {
     c.is_ascii_alphanumeric() || c == '_'
 }
 
+fn is_line_char(c: char) -> bool {
```
emilk, line 1333: nitpick: "line character" is not a very established expression afaik. When I read `is_line_char(x)` I didn't understand it. I think it would make the code more readable to replace `is_line_char(c)` with `!is_linebreak(c)`

trevyn: Done

emilk (approved):

emilk:
Good addition; thanks!

PR #1520, opened 2022-04-20
FileStorage: Ensure we gracefully flush when we terminate

dbuch:
Also try hard not to race with auto-save

### DIFF
--- a/epi/src/file_storage.rs
+++ b/epi/src/file_storage.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::HashMap,
     path::{Path, PathBuf},
+    sync::mpsc,
 };
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


epi/src/file_storage.rs
```diff
@@ -1,26 +1,59 @@
...
 }
 
 impl FileStorage {
     /// Store the state in this .ron file.
     pub fn from_ron_filepath(ron_filepath: impl Into<PathBuf>) -> Self {
         let ron_filepath: PathBuf = ron_filepath.into();
+        let cloned_filepath = ron_filepath.clone();
+        let (tx, rx): (mpsc::Sender<Message>, mpsc::Receiver<Message>) = mpsc::channel();
+        let flush_handle = Some(std::thread::spawn(move || loop {
+            let message = rx.recv().unwrap();
```
emilk, line 39:
```suggestion
        let flush_handle = Some(std::thread::spawn(move ||
            while let Ok(message) = rx.recv() {
```

less unwraps, and closes the thread when the sender drops. This also means we don't need `Message::Terminate`

epi/src/file_storage.rs
```diff
@@ -61,15 +94,8 @@ impl crate::Storage for FileStorage {
...
-            let file_path = self.ron_filepath.clone();
             let kv = self.kv.clone();
-
-            std::thread::spawn(move || {
-                let file = std::fs::File::create(&file_path).unwrap();
-                let config = Default::default();
-                ron::ser::to_writer_pretty(file, &kv, config).unwrap();
-                tracing::trace!("Persisted to {:?}", file_path);
-            });
+            self.sender.send(Message::DoFlush(kv)).unwrap();
```
emilk, line 98: The problem with this is that there is no backpressure in `std::sync::mpsc::channel`, because it is an infinite buffer. This means that if serialization is slow, we will just create an ever-growing queue of `Message::DoFlush` until we run out of memory.

I think the simpler solution is to just store a join handle from `std::thread::spawn` and join any old thread before spawning a new one. Considering the default auto-save policy is 30s, the overhead of spawning a thread is negligible in comparison.

emilk (changes requested):
This has a slow-consumer problem (see comments). I think we should use a simpler solution.

Issue #1000, opened 2021-12-26
[Request] Consistent `TextEdit`


alexmozaidze:
**Is your feature request related to a problem? Please describe.**
I don't like how immutable `TextEdit` looks different than the regular one AKA `egui::TextEdit::new(&mut some_string)` looks different than `egui::TextEdit::new(&mut some_string.as_str())`. Immutable `TextEdit` looks like it's just `Label` except you can select it, and when you give it background it changes background behind the text, but the box that the text is inside of stays the same, while mutable `TextEdit` looks sexy af but is editable.

**Describe the solution you'd like**
Make both "versions" of `TextEdit` look the same in order to be less confusing.

**Describe alternatives you've considered**
I guess it's possible to regenerate `some_string` every time user changes it, maybe? Or have 2  versions of `some_string` -  one for display and other one for program to use, but it's more of a hack than an actual  solution.

**Additional context**
I am making a map editor that returns a copy-able output in form of TOML with all entities' info (position, name, e.t.c.)

LordMZTE:
I don't quite understand how they look different. After all, the `TextEdit` doesn't even know if it's editable (if you make it immutable just by passing a mutable reference to some data that won't be preserved on the next frame). That said, I'm not even sure if this is the right approach to make an immutable `TextEdit`. I've looked at this before and it seems to be the only way sadly.

emilk:
We currently have three (relevant) categories of text in egui:

* Text that can be selected and edited (mutable + selectable)
* Text that can be selected but not edited (imutable + selectable)
* Text that can neither be selected nor edited (immutable and unselectable)

I think it is very important that the first two categories have different looks (by default). A user must be able to look at a egui widget and understand if it can be interacted with, and if so how.

Currently the second two categories (immutable text) also look the same, which is not ideal. Ideally we should have three different looks, not two.

As far as I know there is no consistent UI language for "this text can be selected, but not edited". On the web, most text is selectable, except for buttons. In native apps, most text is _not_ selectable, with rare exceptions. In both cases selectable text looks the same as text which cannot be selected, which is also the current egui behavior.

---

**In summary:** changing immutable text to look like all other `TextEdit`:s is in my opinion a bad idea, which will cause a lot of confusion, and would make egui apps wildly different from any app I know of.

I think we have a choice of two paths forwards (not mutually exclusive)

* Make selectable text look different from both editable text and unselectable text
[...]

PR #1523, opened 2022-04-21
Don't paint insertion point if text is immutable

trevyn:
To match the general behavior/look of selectable-but-immutable text in other UIs, including web.

See also https://github.com/emilk/egui/issues/1000

https://user-images.githubusercontent.com/230691/164389025-0c51b3f6-4ab4-47c9-89ee-830edd68829f.mov


emilk (approved):
Makes sense!

Issue #1036, opened 2022-01-04
Generalize `CollapsingHeader` to show any widget as a header


Mingun:
**Is your feature request related to a problem? Please describe.**
In order to implement a tree table widget it is needed to make a row of cells as header of collapsible block. Unfortunately, right now this is impossible out-of-box as `CollapsingHeader` can only use a text as a header.

**Describe the solution you'd like**
Allow to accept a closure for a header, maybe something like:
```rust
let response = CollapsingHeader::new(...)
    .header(ui, |ui| {...})
    .show(ui, |ui| {...});
```

PR #1538, opened 2022-04-27
Collapsing header with custom header

emilk:
This makes it possible to construct collapsing headers with custom header contents.

Closes https://github.com/emilk/egui/pull/754
Closes https://github.com/emilk/egui/issues/1036

```rust
let id = ui.make_persistent_id("my_collapsing_header");
egui::collapsing_header::CollapsingState::load_with_default_open(ui.ctx(), id, false)
    .show_header(ui, |ui| {
        ui.label("Header"); // you can put checkboxes or whatever here
    })
    .body(|ui| ui.label("Body"));
```

I think this is a better approach for making header selectable, which means we could/should deprecate `CollapsingHeader::selectable`, which is buggy anyway (i.e. clicking both selects and toggles, which is very annoying).

This introduces a separate `CollapsingState` which is perhaps a bit confusing, but `CollapsingHeader` is a bit overloaded with features and I didn't want to pile onto it. `CollapsingState` can also be used as a more low-level building block (and is indeed used by `CollapsingHeader` and by `Window`).

# TODO
* [x] Fix window collapse animation
* [x] Improve UI: break up `show_custom_header` into two calls

### DIFF
--- a/egui/src/containers/collapsing_header.rs
+++ b/egui/src/containers/collapsing_header.rs
@@ -3,6 +3,9 @@ use std::hash::Hash;
 use crate::*;
 use epaint::Shape;
 
+/// This is a a building block for building collapsing regions.
+///
+/// It is used by [`CollapsingHeader`] and [`Window`], but can also be used on its own.
 #[derive(Clone, Copy, Debug)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub struct CollapsingState {
@@ -36,13 +39,12 @@ impl CollapsingState {
         })
     }
 
-    /// `None` if this is is a new [`CollapsingState`].
-    pub fn is_open(ctx: &Context, id: Id) -> Option<bool> {
-        if ctx.memory().everything_is_visible() {
-            Some(true)
-        } else {
-            CollapsingState::load(ctx, id).map(|state| state.open)
-        }
+    pub fn is_open(&self) -> bool {
+        self.open
+    }
+
+    pub fn set_open(&mut self, open: bool) {
+        self.open = open;
     }
 
     pub fn toggle(&mut self, ui: &Ui) {
--- a/egui/src/containers/collapsing_header.rs
+++ b/egui/src/containers/collapsing_header.rs
@@ -16,6 +16,8 @@ pub(crate) struct InnerState {
 /// This is a a building block for building collapsing regions.
 ///
 /// It is used by [`CollapsingHeader`] and [`Window`], but can also be used on its own.
+///
+/// See [`CollapsingState::show_custom_header`] for how to show a collapsing header with a custom header.
 #[derive(Clone, Debug)]
 pub struct CollapsingState {
     id: Id,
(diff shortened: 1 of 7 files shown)


Mingun:
I think, this will help with #1036 as well?

gents83:
I really like the idea of having header separate from the body itself and It looks nicer.
I just didn't get the example that show the selected state, that seems now just a tree (Sorry I'm looking at it from mobile and it's not really helping 😅)
Is It left to the user instead of being managed automatically?

emilk:
@gents83 yeah, that's the point. Instead of having "selectable" as part of `CollapsingHeader` you can built it yourself, with all the details you want:


```rust
let id = ui.make_persistent_id("my_collapsing_header");
egui::collapsing_header::CollapsingState::load_with_default_open(ui.ctx(), id, true)
    .show_custom_header(
        ui,
        |ui| {
            ui.toggle_value(&mut self.selected, "Click to select/unselect header");
        },
        |ui| {
            ui.label("Body");
        },
    );
```

emilk:
I think I can improve the UI of this feature a bit though

egui/src/containers/collapsing_header.rs
```diff
@@ -3,102 +3,256 @@ use std::hash::Hash;
...
+    /// # egui::__run_test_ui(|ui| {
+    /// let id = ui.make_persistent_id("my_collapsing_header");
+    /// egui::collapsing_header::CollapsingState::load_with_default_open(ui.ctx(), id, false)
+    ///     .show_custom_header(ui, |ui| {
+    ///         ui.label("Header"); // you can put checkboxes or whatever here
+    ///     })
+    ///     .body(|ui| ui.label("Body"));
+    /// # });
+    /// ```
+    pub fn show_custom_header<HeaderRet>(
```
Mingun, line 131: Why not `header` or `show_header`? It's looks that header always "custom", so not need to explicitly define that?

emilk: `egui` uses the `show` prefix for all container widgets that take an `impl FnOnce(&mut Ui)` (`Window::show`, `Frame::show`, …).

…but I agree with `show_header` being shorter and sweeter

lumlune:
Is it possible to maintain "click header to expand tree" behavior with these semantics? I would like to use this to make a normal tree but with a context menu on each header, for example.

emilk:
@lumlune you should be able to do that like so:

```rust
let response = CollapsingHeader::new("header").show(ui, |ui| ui.label("body"));
response.header_response.context_menu(|ui| ui.label("Shown on right-clicks"));
```

it should work already on egui 0.17.0 IIRC

KokeBlacksmith:
With this implementation is there a way to handle the collapsing (open, close) within the show_header method?

`let id = ui.make_persistent_id("my_collapsing_header");
egui::collapsing_header::CollapsingState::load_with_default_open(ui.ctx(), id, true)
    .show_header(ui, |ui| {
        // access CollapsingState to handle open or close
    }).body(|ui| {
        ui.label("Body");
    });`

PR #1540, opened 2022-04-28
Move code from `egui_demo_lib` to `egui_demo_app`

emilk:
This moves all eframe specific code out of `egui_demo_lib` so that it can more easily be used by 3rd party integrations for tests and demos (as done in e.g. https://github.com/not-fl3/egui-miniquad/blob/master/examples/demo.rs).

This also cleans up the feature names and dependencies.


nagisakuya:
It seems link to source cord in egui demo is broken.
https://www.egui.rs/index.html

Issue #1553, opened 2022-05-01
Add sense param for tables


aleokdev:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to interact with a table when I click on one of its rows, e.g. 
```rust
let do_thing = row.col(|ui| { ui.label(layer_name); }).clicked();
```

**Describe the solution you'd like**
I'd like to pass in a `sense` parameter to the table builder.

**Describe alternatives you've considered**
Adding a `sense` parameter to the `rows` function.

Issue #1516, opened 2022-04-19
Z-ordering of shapes


emilk:
egui and epaint are currently based on "painters algorithm", i.e. painting things from back-to-front in order to get some shapes to cover others. This means when painting widgets you need to do so back-to-front. Sometimes this is difficult. Sometimes you want to paint things that go in the background last, or things that go in the foreground first.

This is currently cumbersome. For instance, when painting the `egui::Frame` we need to wait until the contents has been added before we know how big to paint the frame. So the code adds a dummy `Shape::Noop` and gets a `ShapeIdx` where the background it later put.

One way to make this less cumbersome is to introduce a `Z` value for each `Shape`. This could be an integer or float, and shapes with a lower Z will always be behind those with a higher Z. Shapes with the same Z are painted in the order they are added.

This could be implemented by habing `PaintList` store `Vec<(ZIndex, ClippedShape)>`, where `ZIndex` is a a typedef for e.g. `f64`. Each `PaintList` would do a stable sort on the Z value before draining in `GraphicLayers::drain`.

`egui::Painter` could have a setting for which `Z` value it uses so that one can set it for a `Painter` and then pass the `Painter` to a function that does the painting, with that function being agnostic to what Z it is painting to (just as it is agnostic to what paint layer it is painting to).

## Questions
What type for Z index: `i64`, `f32`, `f64`?

Should we have standard indexes for foreground and background?

What are the performance implications of this?

Should we still have separate `PaintList`s in `GraphicLayers`, or just put the `LayerId` together with `ZIndex` and `ClippedShape` and stable-sort on that too while we're at it?

setzer22:
Hi! I'd like to tackle this. I think this would help simplify things a lot. As you say, widget call order is coupling together interaction and draw order in a way that makes some things more cumbersome than they should.

The plan you outline above sounds good. But I have a few questions / comments.

First, how is this going to interact with the `order` in the `Areas` memory object? Right now, `GraphicLayers::drain` is special cased for areas and those are ordered using their `LayerId`s. 

> What type for Z index: i32, i64, f32, f64?
Float sounds like the most flexible, because you can always sneak in a shape between every pair of shapes (as long as you're not within precision limits, but it's still better than an integer). As for 32 or 64 bits, I'm ambivalent. Egui seems to be using f32 everywhere else (Pos2, Rect, ...) so it sounds like `f32` might be a good option here too.

[...]

PR #1554, opened 2022-05-01
Add `StripBuilder::sense` & `TableBuilder::sense`

aleokdev:
Closes <https://github.com/emilk/egui/issues/1553>.


emilk:
Adding a sense is good, but I think we should add a hover effect to it as well. Everything clickable in egui changes appearance when you hover it. This is what let's you know you can click it. You can follow this pattern:

``` rust
if sense.click || sense.drag {
    let visuals = ui.style().interact(&response);
    ui.painter().rect(
        rect.expand(visuals.expansion),
        visuals.rounding,
        visuals.bg_fill,
        visuals.bg_stroke,
    );
}
```

aleokdev:
Not too sure about how to do that. :sweat_smile: I've tried to put that in `StripLayout::add` after the rect allocation, using the allocated rect as the rect to use in the painter, but I just get entire rectangles:
Screenshot_20220504_122107

I'm not too sure how strips work internally either. Will need to dive into this a bit further.

emilk:
Perhaps don't draw the `fill` part, and only the stroke part. Ideally it should be drawn behind the row, but that's harder because https://github.com/emilk/egui/issues/1516 isn't done yet

emilk:
* Feature implemented in https://github.com/emilk/egui/pull/3347

Issue #1555, opened 2022-05-02
Consider making `epi` functionality implementation-agnostic


EmbersArc:
**Is your feature request related to a problem? Please describe.**
So far I've been able to reuse most of `epi`'s functionality in my own WGPU integration, strongly reducing the need for code duplication as it also uses winit. However, with the 0.18 release, `epi` was moved into `eframe` and made dependent on a GL context.

**Describe the solution you'd like**
If possible, epi should be separate and implementation-agnostic. This could significantly reduce the need for code duplication and would make egui's behavior between integrations more consistent.

**Describe alternatives you've considered**
Copying the epi code, removing all GL-related fields and functionality and keeping it in sync with future changes to `eframe::epi`.

I see that this request is a bit inconvenient right now as this change has just been made and there might be good reasons for the tighter GL dependency. But I still wanted to bring it up for discussion as `epi` by itself has been very useful without being dependent on GL.

emilk:
I had no idea anyone was using `epi` outside of `eframe`. This was indeed bad timing 😓

PR #1561, opened 2022-05-03
Official WGPU integration

s-nie:
Based on the discussion in #1555, it looks like providing an official wgpu backend might have its benefits.

This builds on Nils Hasenbanck's [egui_wgpu_backend](https://github.com/hasenbanck/egui_wgpu_backend) but removes the need for [egui_winit_platform](https://github.com/hasenbanck/egui_winit_platform) and some boilerplate code and instead uses egui's own `egui-winit` and `eframe`. This approach worked well in egui 0.17 using the separate `epi` crate. 0.18 made this approach invalid as `epi` was moved into `eframe`.

@hasenbanck Since this is a heavily modified fork from your repository from around egui 0.16, you are still listed as the primary author. I also left the original license files in here, at least for now. Let me know if this is okay for you.

This does not compile right now. But as outlined in #1555, there is some potential for sharing code between the glow and wgpu backend which we can experiment with in this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -133,6 +133,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
+name = "ash"
+version = "0.34.0+1.2.203"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


hasenbanck:
@s-nie From a license and copyright standpoint this likes fine.

emilk:
Great! I continued work on this in https://github.com/emilk/egui/pull/1564 (because I don't know how to commit to this branch), so I'm closing this now.

PR #1569, opened 2022-05-04
CollapsingState Improvements

4JX:
As per https://github.com/emilk/egui/discussions/1566

Concerns are the `FnOnce` > `Fn` change, which shouldn't realistically impact any icon drawing functions and the ergonomics of a "no collapser" mode, since something like https://github.com/4JX/egui/commit/f3234a1f0dd400d33bedfe46c4f3ddff6cbff423 would need to be done.

### DIFF
--- a/egui/src/containers/collapsing_header.rs
+++ b/egui/src/containers/collapsing_header.rs
@@ -106,7 +106,7 @@ impl CollapsingState {
     ///
     /// let id = ui.make_persistent_id("my_collapsing_header");
     /// egui::collapsing_header::CollapsingState::load_with_default_open(ui.ctx(), id, false)
-    ///     .icon(circle_icon);
+    ///     .icon(circle_icon)
     ///     .show_header(ui, |ui| {
     ///         ui.label("Header"); // you can put checkboxes or whatever here
     ///     })
--- a/egui/src/containers/collapsing_header.rs
+++ b/egui/src/containers/collapsing_header.rs
@@ -25,11 +25,11 @@ pub enum CollapserPosition {
 /// It is used by [`CollapsingHeader`] and [`Window`], but can also be used on its own.
 ///
 /// See [`CollapsingState::show_header`] for how to show a collapsing header with a custom header.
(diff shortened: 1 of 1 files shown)


4JX:
Would like to add the commit mentioned above as well if it isn't deemed that much of an issue

egui/src/containers/collapsing_header.rs
```diff
@@ -134,10 +189,22 @@ impl CollapsingState {
...
+                    ui.spacing_mut().item_spacing.x = 0.0; // the toggler button uses the full indent width
+                    let collapser = self.show_default_button_indented(ui);
+                    ui.spacing_mut().item_spacing.x = ui.spacing_mut().icon_spacing; // Restore spacing
+                    let add_header_ret = add_header(ui);
+                    (collapser, add_header_ret)
+                }
+                CollapserPosition::After => {
+                    let add_header_ret = add_header(ui);
+                    ui.spacing_mut().item_spacing.x = 0.0; // the toggler button uses the full indent width
+                    let collapser = self.show_default_button_indented(ui);
```
emilk, line 203: what does it mean to match the indentation in this case?

emilk:
My intent for `CollapsingState` was to make it more of a building block, and not something you customize. I would rather see a solution that looks something like this (from a users perspective):

``` rust
let mut collapsing_state = …;
let header_response = ui.horizontal(|ui| {
    ui.label("Header");
    collapsing_state.show_toggle_button(ui, my_icon_painter);
});
collapsing_state.show_body(&header_response.response, ui, |ui| {
    ui.label("Body");
});
``` 

This keeps `CollapsingState` simple, and makes for a much more powerful and flexible solution.

4JX:
> This keeps `CollapsingState` simple, and makes for a much more powerful and flexible solution.
Oh, I was unaware a state could be constructed with an arbitrary response, hence the approach of this PR. I'll close it and draft a new one with the approach you described, as it seems a lot more useful.

PR #1575, opened 2022-05-04
Fix CircleShape::visual_bounding_rect()

carrotflakes:
I fixed a bug that `CircleShape::visual_bounding_rect()` returns the wrong value.
The Rect returned was the half size of the circle.


emilk (approved):
Good find, thanks!

Issue #1580, opened 2022-05-05
Add `primary_clicked` & `secondary_clicked` for `PointerState`.


aleokdev:
**Is your feature request related to a problem? Please describe.**
I'd like to detect if the user clicked something with the primary button somewhere in order to cancel their selection.
While we have `.any_click()` and `.primary_down()`, we don't have `.primary_clicked()`!

**Describe the solution you'd like**
I'd like a way to know if one button was clicked on this frame.

**Describe alternatives you've considered**
Using `.any_click() && .primary_down()`, but this is faulty: The user could be holding down the primary button and then have clicked with another button.

emilk:
Sounds good, please make a PR adding the missing methods!

PR #1582, opened 2022-05-05
Add *_released & *_clicked PointerState methods

aleokdev:
Closes <https://github.com/emilk/egui/issues/1580>.


aleokdev:
Added release methods too because they were also missing and it was trivial to add them.

emilk (approved):
thanks!

PR #1591, opened 2022-05-06
eframe::App::post_rendering

elwerene:
First try in having a post rendering function as discussed at https://github.com/emilk/egui/discussions/1578#discussioncomment-2692266

I'm not sure if the full `PaintCallbackInfo` would be needed. Either way I don't know how to fill half its fields.

`Frame` is needed to get access to the `gl` handle.

Should I wrap the screen size and the `gl` handle into an own Struct? Or is it ok to provide a readonly reference to `Frame`?


emilk:
Looks good!

It would be great with an example app (that also serves as a test) which paints a GUI and then can take a screenshot of it using this feature!

eframe/src/native/epi_integration.rs
```diff
@@ -264,6 +264,14 @@ impl EpiIntegration {
         full_output
     }
 
+    pub fn post_rendering(&mut self, app: &mut dyn epi::App) {
+        let pixels_per_point = self.frame.info().native_pixels_per_point.unwrap_or(1.0);
+        let screen_size = self.egui_ctx.input().screen_rect.size() * pixels_per_point;
+        let screen_size_px = [screen_size.x as u32, screen_size.y as u32];
```
emilk, line 270: There's a subtle bug here - you are using the native_pixels_per_point, but  `InputState::screen_rect` is based on the current `pixels_per_point` which is customizable. To get the pixel size of the display it is better to ask the `winit::Window` directly with https://docs.rs/winit/latest/winit/window/struct.Window.html#method.inner_size

examples/screenshot/src/main.rs
```diff
@@ -0,0 +1,81 @@
...
+                ui.spinner();
+            }
+
+            ctx.request_repaint();
+        });
+    }
+
+    #[allow(unsafe_code)]
+    fn post_rendering(&mut self, screen_size_px: [u32; 2], frame: &eframe::Frame) {
+        if let Some(gl) = frame.gl() {
```
emilk, line 51: Taking a screenshot each frame is very slow, especialy in debug builds. I think it would be better to only do it when clicking a button.

examples/screenshot/src/main.rs
```diff
@@ -0,0 +1,81 @@
...
+            if ui
+                .add(egui::Label::new("Screenshot (hover me!):").sense(egui::Sense::hover()))
+                .hovered()
+            {
+                ctx.set_visuals(egui::Visuals::dark());
+            } else {
+                ctx.set_visuals(egui::Visuals::light());
+            };
+            if let Some(texture) = self.texture.as_ref() {
+                ui.image(texture, texture.size_vec2());
```
emilk, line 40: nit: This is what it looks like on a high-dpi screen. I think it would be less confusing to instead show the screenshot scaled down to fit inside of `ui.available_size()`

Screen Shot 2022-05-13 at 09 25 00

emilk (changes requested):

examples/screenshot/src/main.rs
elwerene: I like the continuous mode, so I made it toggable :)

PR #1592, opened 2022-05-06
Allow manually painting the CollapsingState icon

4JX:
As described in https://github.com/emilk/egui/pull/1569#issuecomment-1117784113

Perhaps it would be a good idea to also have a (or expose the existing) "paint the default icon" function? Or wrap the icon field in an `Option`?

### DIFF
--- a/egui/src/containers/collapsing_header.rs
+++ b/egui/src/containers/collapsing_header.rs
@@ -92,7 +92,8 @@ impl CollapsingState {
         self.paint_button_indented(ui, paint_default_icon)
     }
 
-    fn paint_button_indented(
+    /// Will toggle when clicked, etc.
+    fn show_button_indented(
         &mut self,
         ui: &mut Ui,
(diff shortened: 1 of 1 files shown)


egui/src/containers/collapsing_header.rs
```diff
@@ -89,6 +89,14 @@ impl CollapsingState {
 
     /// Will toggle when clicked, etc.
     fn show_default_button_indented(&mut self, ui: &mut Ui) -> Response {
+        self.paint_button_indented(ui, paint_default_icon)
+    }
+
+    fn paint_button_indented(
```
emilk, line 95:
```suggestion

    /// Will toggle when clicked, etc.
    fn show_button_indented(
```

(it does more than just paint)

emilk (changes requested):
Looks good!

emilk (approved):

PR #1595, opened 2022-05-07
Fix image button padding on hover

newcomb-luke:
This PR addresses one issue that has been encountered by multiple people where image buttons don't display correctly when hovering. This updates 2 lines of code that fix the problem as shown in the image below.

There are no clippy warnings.

./sh/check.sh was not able to be completed for some unrelated error below in speech-dispatcher:

```
error[E0308]: mismatched types
  --> /home/luke/.cargo/registry/src/github.com-1ecc6299db9ec823/speech-dispatcher-0.12.0/src/lib.rs:35:13
   |
35 |     Male1 = SPDVoiceType::SPD_MALE1,
   |             ^^^^^^^^^^^^^^^^^^^^^^^ expected `u32`, found `i32`
```

The fixed image buttons shown:

fixed_egui

Closes <https://github.com/emilk/egui/issues/1528>.

Issue #1492, opened 2022-04-13
Color32 from hex value


tami5:
**Is your feature request related to a problem? Please describe.**

Most of the time when implementing a UI design, colors comes in hex values. Having to manually convert hex to RGBA is kind tedious and time consuming.

**Describe the solution you'd like**

new method in Color32

```rust
Color32::from_hex("#0d1117")
```

emilk:
This would be nice - especially if it could be a `const fn` or macro, so any typos could be caught at compile time.

I think what we would want to support is only the `#rrggbb` and `#rrggbbaa` formats, not the other weird CSS formats

PR #1596, opened 2022-05-07
Add ability to convert from a hex string to Color32 using macros

newcomb-luke:
This PR closes <https://github.com/emilk/egui/issues/1492>.

Again currently `./sh/check.sh` is currently failing for me, see my other [PR](https://github.com/emilk/egui/pull/1595).
Cargo clippy, cargo test, and cargo fmt have all been run.
The CHANGELOG.md has been updated under "Unreleased"

This change allows you to create Color32's from hex strings:

```rust
let color = epaint::hex_rgb!("#4c4c4c");
```

And it does so at compile-time.


emilk:
Very nice!

But I would suggest we simplify it. As far as I can recall I've never seen premultiplied alpha expressed in hex form. So I suggest we boil it down to a single macro that accepts both RGB and unmultiplied RGBA (`epaint::color_hex!("#55eedd") == epaint::color_hex!("#55eeddff")`).

For additive, users can just do `epaint::color_hex!("#ffeedd").additive()` (we can make `Color32::additive` `const fn`)

newcomb-luke:
Upon further testing, previously the from_rgb, from_rgb_additive, from_rgba_premultiplied were `const`, however from_rgba_unmultiplied due to the floating-point conversions that have to take place, is unable to be `const` [yet](https://github.com/rust-lang/rust/issues/57241). So only the conversion from the string literal to the array of values can be const. So the macro invocation itself cannot yet be `const`, but soon it could be. This isn't much slower than using the regular from_rgba_unmultiplied though.

epaint/src/color.rs
```diff
@@ -186,6 +186,36 @@ impl Color32 {
     }
 }
 
+#[macro_export]
```
emilk, line 189:
```suggestion
/// Construct a [`Color32`] from a hex RGB or RGBA string.
///
/// ```
/// assert_eq!(hex_color!("#202122"), Color32::from_rgb(0x20, 0x21, 0x22));
/// assert_eq!(hex_color!("#abcdef12"), Color32::from_rgba_unmultiplied(0xab, 0xcd, 0xef, 0x12));
/// ```
#[macro_export]
```

emilk (changes requested):
It's a bit of a shame that it is not `const` but yeah, not much to do about that right now. I think this looks good, but the macro needs a docstring!

epaint/src/color.rs
newcomb-luke: Done!

coderedart:
is hex string assumed to be in linear space? or sRGB?
docs could mention that so its not confusing. 
`Color32` mention that `Internally this uses 0-255 gamma space sRGBA color with premultiplied alpha. Alpha channel is in linear space.`.

PR #1599, opened 2022-05-07
Fix Plot auto_bounds when LinkedAxisGroup one axis

enomado:
Without this change there was a bug with `center_y_axis = false` but it was centered anyways.

### DIFF
--- a/egui/src/widgets/plot/mod.rs
+++ b/egui/src/widgets/plot/mod.rs
@@ -69,7 +69,7 @@ const MIN_LINE_SPACING_IN_POINTS: f64 = 6.0; // TODO: large enough for a wide la
 
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 #[derive(Clone)]
-pub struct AutoBounds {
+struct AutoBounds {
     x: bool,
     y: bool,
 }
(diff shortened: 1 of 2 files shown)


egui/src/widgets/plot/mod.rs
```diff
@@ -67,11 +67,34 @@ impl Default for CoordinatesFormatter {
 
 const MIN_LINE_SPACING_IN_POINTS: f64 = 6.0; // TODO: large enough for a wide label
 
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+#[derive(Clone)]
+pub struct AutoBounds {
```
emilk, line 72:
```suggestion
struct AutoBounds {
```

I don't think it needs to be `pub`?

emilk (changes requested):

enomado:
Sure. Thank you.

egui/src/widgets/plot/transform.rs
```diff
@@ -109,6 +129,16 @@ impl PlotBounds {
...
+        let width = self.width().max(0.0);
+        self.expand_x(margin_fraction.x as f64 * width);
+    }
+
+    pub(crate) fn add_relative_margin_y(&mut self, margin_fraction: Vec2) {
+        let height = self.height().max(0.0);
+        self.expand_y(margin_fraction.y as f64 * height);
+    }
+
     pub(crate) fn add_relative_margin(&mut self, margin_fraction: Vec2) {
```
emilk, line 142: let's remove this unused function to appease clippy

enomado: done

PR #1603, opened 2022-05-08
Added Extra1 & Extra2 pointer buttons

sy1ntexx:
This PR adds support for 2 buttons (Extra1 & Extra2) which are present on most modern mice.


eframe/src/web/input.rs
```diff
@@ -14,6 +14,8 @@ pub fn button_from_mouse_event(event: &web_sys::MouseEvent) -> Option<egui::Poin
         0 => Some(egui::PointerButton::Primary),
         1 => Some(egui::PointerButton::Middle),
         2 => Some(egui::PointerButton::Secondary),
+        3 => Some(egui::PointerButton::Extra1),
+        4 => Some(egui::PointerButton::Extra2),
```
emilk, line 18: Seems a bit weird to begin indexing at `1`, but perhaps these are standard names?

emilk (approved):
Looks good, assuming it has been tested.

eframe/src/web/input.rs
sy1ntexx: Sorry I didn't come up with better names. I can rename them to Forward & Backward but I don't know if it's meaningful enough because they go by these names mostly in browsers.

emilk: Let's keep it as is

Issue #1602, opened 2022-05-08
Use cursor grab / mouse lock api


enomado:
https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_cursor_grab
https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API

For example it could be awesome to lock pointer when setting DragValue for example in cases when UX breaks when widget is near screen edge.

maxjvh:
one caveat is that at least Chromium constantly shows a warning that the cursor is hidden which becomes annoying really quick

emilk:
Closing: cursor locking is available via `ViewportCommand::CursorGrab(CursorGrab::Locked)` (#3172), and raw mouse motion via `Event::MouseMoved` (#4063).

PR #1605, opened 2022-05-08
Feature Mouse Lock API

enomado:
Trying to investigate on #1602 
Seems awesome.

https://user-images.githubusercontent.com/707007/167320729-9f56e8a9-47bc-4a0f-942b-2af1bd6ce01c.mp4

There is a demo you can try your own

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2209,6 +2209,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "mouse_lock"
+version = "0.1.0"
+dependencies = [
+ "eframe",
+]
+
+[[package]]
 name = "naga"
 version = "0.8.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ members = [
     "examples/hello_world",
     "examples/puffin_profiler",
     "examples/retained_image",
+    "examples/mouse_lock",
     "examples/svg",
 ]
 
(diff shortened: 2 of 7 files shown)


enomado:
Some todos: don't change position on emit release event.
For some reason if we lose focus while dragging, we would not get release event and cant unlock cursor

enomado:
Just realized I can use https://docs.rs/winit/latest/winit/event/enum.DeviceEvent.html#variant.MouseMotion for delta to modify DragValue just by delta and thats all.

emilk:
converted to draft since you still have some `todos`

enomado:
Just want you to check out and write what do you think. Is it worth feature of it costs? 

Btw first todo was fixed, and second todo - looks like a bug in winit. browser behave the same way. and looks like we can turn off locking in each component by asking `clicked_anywhere`. 

I think we can add this as experimental feature anyway, but I already happy with my deltas, so I can enjoy DragValue. I just worried about excess complexity. If you are okay with that I'll polish this.

Thank you for reply

enomado:
This could be merged now I guess

emilk:
is it really worth the added complexity?

enomado:
> is it really worth the added complexity?

If we dont merge it, anybody who want this must do it from scratch.
If we merge this - this add just few odd, experimental methods that could motivate someone to play with this.

I really dont know. perhaps there should be some voting process about it. As for me, the solution is not very kludgy.

bglw:
Not sure how hard this would be to get across the line, but it is something I'm hoping egui could provide an api for. Very useful for things like draggable values — anyone using Figma will have likely appreciated it when dragging a value and the cursor wraps around the screen, rather than hitting the right edge of the screen and stopping.

emilk:
Closing: this landed in a different form. Cursor locking is available via `ViewportCommand::CursorGrab(CursorGrab::Locked)` (#3172), and raw mouse motion via `Event::MouseMoved` (#4063). Thanks for the PR!

Issue #1612, opened 2022-05-09
Clipboard support is broken on Wayland


Detegr:
**Describe the bug**

The clipboard is not working on Wayland. `arboard` supports only Wayland terminal applications, not graphical applications.

**To Reproduce**
```
use eframe::egui;

fn main() {
    let options = eframe::NativeOptions::default();
    eframe::run_native(
        "My egui App",
        options,
        Box::new(|_cc| Box::new(MyApp::default())),
    );
}

struct MyApp {
    s: String,
}

impl Default for MyApp {
    fn default() -> Self {
        Self { s: String::new() }
    }
}

impl eframe::App for MyApp {
    fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
        egui::CentralPanel::default().show(ctx, |ui| {
            if ui.button("copy test string to clipboard").clicked() {
                ctx.output().copied_text = "hello, world!".into();
            }
            ui.text_edit_multiline(&mut self.s);
        });
    }
}
```
1. Run the demo on Wayland.
2. Try clicking the button, or copying from the text box.
3. Nothing happens.

**Expected behavior**
Clipboard should work as expected.

**Desktop (please complete the following information):**
 - OS: Linux, Wayland

quietvoid:
The current clipboard was super inconsistent or not working at all for me in Wayland, using `eframe`.
Fix in #1613 works consistently now.

Thanks!

Issue #1474, opened 2022-04-10
Clipboard dependency unmaintained


Shatur:
This crate depends on https://github.com/alacritty/copypasta which depends on unmaintained https://github.com/quininer/x11-clipboard/issues/19. Which also depends on xkb which is [unsound and also unmaintained](https://github.com/RustSec/advisory-db/issues/653) :rofl:
I would suggest to switch to https://github.com/1Password/arboard

emilk:
I didn't know about `arboard` - good find!

PR #1613, opened 2022-05-09
Fix clipboard on Wayland

Detegr:
arboard advertises that it works with Wayland, but in reality it only works with Wayland terminal applications. To make the clipboard work with applications that draw Wayland surfaces, arboard isn't going to work.

Copypasta does support Wayland's graphical clipboard, but the usage isn't documented. However, for the reasons mentioned in #1474 the move from Copypasta to arboard makes sense.

To resolve the issue, this commit brings in an optional dependency smithay-clipboard, that is a crate that correctly handles the Wayland clipboard in graphical applications. It is used by default if a Wayland window handle is found. If for some reason the handle to the Wayland window handle cannot be fetched, arboard is used as a backup.

It feels a bit wrong adding the `wayland_display` argument to `from_pixels_per_point`, but I couldn't figure out any better alternatives. Maybe pass the whole window to it?

Fixes https://github.com/emilk/egui/issues/1612

### DIFF
--- a/egui-winit/Cargo.toml
+++ b/egui-winit/Cargo.toml
@@ -25,7 +25,7 @@ bytemuck = ["egui/bytemuck"]
 
 # enable cut/copy/paste to OS clipboard.
 # if disabled a clipboard will be simulated so you can still copy/paste within the egui app.
-clipboard = ["arboard"]
+clipboard = ["arboard", "smithay-clipboard"]
 
 # enable opening links in a browser when an egui hyperlink is clicked.
 links = ["webbrowser"]
@@ -55,3 +55,6 @@ webbrowser = { version = "0.7", optional = true }
 
 # feature screen_reader
 tts = { version = "0.20", optional = true } # stuck on old version due to compilation problems on linux
+
+[target.'cfg(any(target_os="linux", target_os="dragonfly", target_os="freebsd", target_os="netbsd", target_os="openbsd"))'.dependencies]
+smithay-clipboard = { version = "0.6.3", optional = true }
--- a/egui-winit/src/lib.rs
+++ b/egui-winit/src/lib.rs
@@ -5,6 +5,8 @@
 
 #![allow(clippy::manual_range_contains)]
 
+use std::os::raw::c_void;
+
 pub use egui;
 pub use winit;
 
@@ -13,6 +15,15 @@ pub mod screen_reader;
 mod window_settings;
 
 pub use window_settings::WindowSettings;
+
+#[cfg(any(
+    target_os = "linux",
+    target_os = "dragonfly",
+    target_os = "freebsd",
+    target_os = "netbsd",
+    target_os = "openbsd"
+))]
+use winit::platform::unix::WindowExtUnix;
 
 pub fn native_pixels_per_point(window: &winit::window::Window) -> f32 {
     window.scale_factor() as f32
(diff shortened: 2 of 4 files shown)


emilk:
wouldn't it make more sense to make this PR to `arboard`?

Detegr:
Maybe. I'll try preparing and submitting a PR to them. Anyway, `egui` needs a modification to pass the wayland display handle to the clipboard crate, whatever it might be, so I'd leave this open for now.

egui-winit/Cargo.toml
```diff
@@ -49,6 +49,7 @@ tracing = "0.1"
 winit = "0.26.1"
 
 arboard = { version = "2.1", optional = true, default-features = false }
+smithay-clipboard = { version = "0.6.3", optional = true }
```
emilk, line 52: We should put this under `[target.'cfg(target_os = "linux")'.dependencies]`

and then add a`#[cfg(target_os = "linux")]` to all uses of it, or
`#[cfg(all(target_os = "linux", feature = "smithay-clipboard"))]`

emilk (changes requested):

emilk:
> Maybe. I'll try preparing and submitting a PR to them. Anyway, `egui` needs a modification to pass the wayland display handle to the clipboard crate, whatever it might be, so I'd leave this open for now.

Let's merge this fix into `egui-winit` now and then, later, we can hopefully remove it when this gets upstreamed to `arboard`.

egui-winit/Cargo.toml
Detegr: I went for `cfg(unix)` everywhere instead as at least FreeBSD has [some kind of support](https://docs.freebsd.org/en/books/handbook/x11/#x-wayland) for Wayland.

egui-winit/src/lib.rs
```diff
@@ -53,13 +58,26 @@ impl State {
     /// * `max_texture_side`: e.g. `GL_MAX_TEXTURE_SIZE`
     /// * the native `pixels_per_point` (dpi scaling).
     pub fn new(max_texture_side: usize, window: &winit::window::Window) -> Self {
-        Self::from_pixels_per_point(max_texture_side, native_pixels_per_point(window))
+        #[cfg(unix)]
+        let wayland_display = window.wayland_display();
+        #[cfg(not(unix))]
+        let wayland_display = None;
```
Detegr, line 64: I think this needs to be wrapped in conditional compilation too, as Windows won't have the `wayland_display()` function that comes from a platform dependent trait.

egui-winit/Cargo.toml
Detegr: However it seems `winit` is using:
```
#![cfg(any(
    target_os = "linux",
    target_os = "dragonfly",
    target_os = "freebsd",
    target_os = "netbsd",
    target_os = "openbsd"
))]
```
instead for the unix platform module. I guess that means that compiling this as-is will fail on some esoteric operating system that is an unix but isn't any of those in the list.

egui-winit/src/lib.rs
emilk: I added Windows and Max to the CI matrix (in https://github.com/emilk/egui/pull/1631), so if you merge in `master` to this PR branch the CI will test for you!

Detegr: Nice! A welcome change. Rebased `master`. I guess I need a maintainer approval to get it the CI running as I'm a first-time contributor.

emilk:
So it seems the new CI checks works! The problem with `#[cfg(not(unix))]` is that it includes macos!

enomado:
cargo check --target wasm32-unknown-unknown

emits

    69 |         Ok(Clipboard { platform: PlatformClipboard::new()? })
       |                                  ^^^^^^^^^^^^^^^^^ use of undeclared type `PlatformClipboard`

Detegr:
I see now why `#[cfg(unix)]` is problematic. Let's use the same OS listing as `winit` does for the conditional compilation, that should work I guess.

Detegr:
> cargo check --target wasm32-unknown-unknown
> 
> emits
> 
> ```
> 69 |         Ok(Clipboard { platform: PlatformClipboard::new()? })
>    |                                  ^^^^^^^^^^^^^^^^^ use of undeclared type `PlatformClipboard`
> ```

I think this issue is unrelated to this PR. The same happens with the `master` branch too.

emilk:
I would prefer if you made new commits instead of squashing. New commits means I can see the changes since the previous one! I always squash on merge anyway, so don't worry about making a pretty history.

Detegr:
You can use the `Compare` button to see changes. But if you're squashing them on merge anyway, I'll do separate commits in the future :)

emilk:
> You can use the `Compare` button to see changes

Today I learned! thanks :)

Issue #1611, opened 2022-05-09
use DeviceEvent::MouseMotion to get mouse delta when mouse go out of screen.


enomado:
This could enhance DragValue UX when mouse stuck in screen rect. 

For example this http://idflood.github.io/draggable-number.js/ can increment number when mouse is out of screen, So I guesst browsers delta works the same way.

eframe/src/native/run.rs:206

PR #1614, opened 2022-05-10
Mouse delta from DeviceEvent

enomado:
This can enhance usability of DragValue much. #1611

https://user-images.githubusercontent.com/707007/167677844-db6a0514-e5c1-427e-b563-b18e00fea289.mp4

### DIFF
--- a/egui-winit/src/lib.rs
+++ b/egui-winit/src/lib.rs
@@ -6,6 +6,7 @@
 #![allow(clippy::manual_range_contains)]
 
 pub use egui;
+use egui::Vec2;
 pub use winit;
 
 pub mod clipboard;
@@ -119,6 +120,13 @@ impl State {
         self.egui_input.take()
     }
 
+    pub fn on_mouse_delta(&mut self, delta: (f64, f64)) {
+        self.egui_input.events.push(egui::Event::MouseDelta(Vec2 {
+            x: delta.0 as f32,
+            y: delta.1 as f32,
+        }))
+    }
+
     /// Call this when there is a new event.
     ///
     /// The result can be found in [`Self::egui_input`] and be extracted with [`Self::take_egui_input`].
@@ -152,6 +160,7 @@ impl State {
             }
             WindowEvent::CursorMoved { position, .. } => {
                 self.on_cursor_moved(*position);
+
                 egui_ctx.is_using_pointer()
             }
             WindowEvent::CursorLeft { .. } => {
@@ -226,10 +235,8 @@ impl State {
                 };
                 false
             }
-            _ => {
-                // dbg!(event);
-                false
-            }
+
+            _ => false,
         }
     }
 
--- a/egui/src/data/input.rs
+++ b/egui/src/data/input.rs
@@ -177,6 +177,11 @@ pub enum Event {
     /// The mouse or touch moved to a new place.
     PointerMoved(Pos2),
 
+    /// The mouse delta from winit::DeviceEvent. MouseDelta and PointerMoved can be emmited simultaneously.
+    /// If MouseDelta accumulated value is zero per frame, then `old_pos-new_pos` will be used as fallback.
+    /// For now it used in situations when we need delta but mouse pointer is on screen edge.
+    MouseDelta(Vec2),
+
     /// A mouse button was pressed or released (or a touch started or stopped).
     PointerButton {
         /// Where is the pointer?
--- a/egui/src/data/input.rs
+++ b/egui/src/data/input.rs
@@ -177,9 +177,9 @@ pub enum Event {
     /// The mouse or touch moved to a new place.
     PointerMoved(Pos2),
 
-    /// The mouse delta from winit::DeviceEvent. MouseDelta and PointerMoved can be emmited simultaneously.
-    /// If MouseDelta accumulated value is zero per frame, then `old_pos-new_pos` will be used as fallback.
-    /// For now it used in situations when we need delta but mouse pointer is on screen edge.
+    /// Can be used as mouse delta when mouse pointer is on screen edge, like in some CAD systems.
+    //  Currently uses only `winit::DeviceEvent`
+    /// `MouseDelta` and `PointerMoved` can be emmited simultaneously.
     MouseDelta(Vec2),
 
     /// A mouse button was pressed or released (or a touch started or stopped).
--- a/egui/src/input_state.rs
+++ b/egui/src/input_state.rs
@@ -450,6 +450,11 @@ pub struct PointerState {
     /// How much the pointer moved compared to last frame, in points.
     delta: Vec2,
 
+    /// How much the pointer moved despite on cursor location, window, in unspecified units.
+    /// Platform specific.
+    /// See `winit::event::DeviceEvent::MouseMotion`
+    motion: Vec2,
+
     /// Current velocity of pointer.
     velocity: Vec2,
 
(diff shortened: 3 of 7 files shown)


egui-winit/src/lib.rs
```diff
@@ -680,3 +687,4 @@ macro_rules! profile_scope {
 }
 #[allow(unused_imports)]
 pub(crate) use profile_scope;
+use winit::event::DeviceEvent;
```
emilk, line 690: This is a very weird place for a `use` statement. Is it even used?

egui/src/data/input.rs
```diff
@@ -177,6 +177,8 @@ pub enum Event {
     /// The mouse or touch moved to a new place.
     PointerMoved(Pos2),
 
+    MouseDelta(Vec2),
```
emilk, line 180: This need documentation!

For instance: should it always be emitted alongside `PointerMoved`? Is it emitted on touch screens?

emilk (changes requested):
Thanks! This needs a line in the `egui` changelog so other integrations knows to use the new event.

Also: should this be implemented for `eframe` web?

egui/src/data/input.rs
enomado: Well, the whole solution is workaround . It `needs` to be exact `one` event with position and delta both. But as we can see, winit has different event for window-mouse-position change, and DeviceEvent.

For now I use MouseDelta when it emmited, if not - there is fallback to position, so is would not affect touch screens.

I'll add this note is comment

enomado:
> Also: should this be implemented for `eframe` web?

it could be, i think. web-event has delta already. anyway, as i said it should not break existing behavior, if there was no MouseDelta emmited

enomado:
Bad news. web just cant delta. I thought [this numeric input](http://idflood.github.io/draggable-number.js/) works, I thought this works https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/movementX

But seems not. So in wasm it works the old way.

egui/src/data/input.rs
```diff
@@ -177,6 +177,11 @@ pub enum Event {
     /// The mouse or touch moved to a new place.
     PointerMoved(Pos2),
 
+    /// The mouse delta from winit::DeviceEvent. MouseDelta and PointerMoved can be emmited simultaneously.
+    /// If MouseDelta accumulated value is zero per frame, then `old_pos-new_pos` will be used as fallback.
+    /// For now it used in situations when we need delta but mouse pointer is on screen edge.
+    MouseDelta(Vec2),
```
emilk, line 183:
```suggestion
    /// Mouse movement in unspecified units.
    ///
    /// Both [`Self::MouseDelta`] and [`Self::PointerMoved`] can both be emitted for the same movement,
    /// but when a pointer is grabbed or at the edge of the screen, only [`Self::MouseDelta`] is emitted.
    ///
    /// Some integrations do not emit any [`Self::MouseDelta`]. In these cases the change in [`Self::PointerMoved`]  will be used for [`PointerState::delta`].
    MouseDelta(Vec2),
```

egui/src/input_state.rs
```diff
@@ -473,6 +473,8 @@ impl PointerState {
         let old_pos = self.latest_pos;
         self.interact_pos = self.latest_pos;
 
+        let mut delta_acc = Vec2::ZERO;
```
emilk, line 476: this code can be simplified by removing `delta_acc` and just writintg to `self.delta` directly.

emilk (changes requested):
There is a problem with this approach: the units emitted by `MouseMotion` is in unspecified units: https://docs.rs/winit/latest/winit/event/enum.DeviceEvent.html#variant.MouseMotion

For instance, these units may be ignoring mouse sensitivity or acceleration. Therefor it is not equivalent to comparing changes in pointer position. I think if you really want to use this in a UI we need to assign it to something else than `PointerState::delta`, and keep that to mean "mosue movements in points", and perhaps add a separate `motion: Vec2` for these unspecified units (which can still fall back to changes in pointer position for integrations that don't emit `MosueDelta` events.

egui/src/input_state.rs
enomado: Its my habit to leave variable local as long as I can, and do `global` mutations just after all actions. But yeah, I'll change that.

enomado:
> perhaps add a separate `motion: Vec2` for these unspecified units
Great idea. I'll do it. And change DragValue to use them.

enomado:
Squashed and rebased

egui/src/data/input.rs
emilk: "The mouse delta from winit::DeviceEvent." is true for one of many backends, but not for all.

These docstrings have two targets: egui users, and people writing egui integrations.

egui/src/response.rs
```diff
@@ -277,6 +277,24 @@ impl Response {
         self.drag_released
     }
 
+    /// Mouse dragged, nevertheless it's stuck on screen's edge or not.
+    pub fn motion_delta(&self) -> Vec2 {
```
emilk, line 281: This is very confusing. What does it return on touch screens? Does it matter if `self.dragged` is true or false? Why does it act differently when compiled for wasm?

NOTE: wasm != web. There are those who run egui inside of wasmtime on desktop.

emilk (changes requested):

egui/src/response.rs
enomado: I've changed the docs

enomado:
Sorry. Is any changes still needed? I was trying to document my best, but there is long story behind this stuff

egui/src/response.rs
```diff
@@ -277,6 +277,27 @@ impl Response {
...
+            let res = self.ctx.input().pointer.motion();
+
+            if res == Vec2::ZERO {
+                // in a case of touchscreen or whatever
+                self.ctx.input().pointer.delta()
+            } else {
+                res
+            }
+        }
+    }
```
emilk, line 299: It's not `egui`s responsibility to have platform-specific code like this. Instead it is up to the integrations to feed `egui` with the correct input.

In this case the winit backend should feed different data than the web backend. [Other integrations](https://github.com/emilk/egui#integrations) needs to do whats right for them.

 So if this is called `mtion_delta` then it should probably use `self.ctx.input().pointer.motion()` regardless of what the compilation target is.

Also: 'wasm" doesn't mean "web". There are egui users who run egui in wasmtime on desktop.

egui/src/response.rs
```diff
@@ -277,6 +277,27 @@ impl Response {
...
+        #[cfg(target_family = "wasm")]
+        return self.ctx.input().pointer.delta();
+
+        #[cfg(not(target_family = "wasm"))]
+        {
+            let res = self.ctx.input().pointer.motion();
+
+            if res == Vec2::ZERO {
+                // in a case of touchscreen or whatever
+                self.ctx.input().pointer.delta()
```
emilk, line 294: maybe touch screens should set `motion` too?

Better yet: `RawInput` could have `motion: Option<Vec2>` and then `InputState::begin_frame` would use it if set, or else fall back to the pointer delta. That way it will work correctly for all existing backends.

emilk (changes requested):

emilk:
Closing: raw mouse motion landed in #4063 as `Event::MouseMoved` and `Response::drag_motion`. Thanks for the PR!

PR #1617, opened 2022-05-10
eframe: read native window position and size

TicClick:
the window's position and dimensions are now available via `eframe::Frame::info().window_info` on desktop (and probably mobile platforms as well, as I've patched the native backend). no issue to close, as this is mostly my own need.

this is my first PR in Rust, so feel free to point out anything you deem poor.

### DIFF
--- a/eframe/src/epi.rs
+++ b/eframe/src/epi.rs
@@ -341,6 +341,17 @@ pub struct WebInfo {
     pub location: Location,
 }
 
+/// Information about the application's main window, if available.
+#[derive(Clone, Debug)]
+pub struct WindowInfo {
+    /// Coordinates of the window's top left corner, relative to the top left corner
+    /// of a user's first monitor.
+    pub position: egui::Vec2,
+
+    /// Window dimensions, in pixels.
+    pub size: egui::Vec2,
+}
+
 /// Information about the URL.
 ///
 /// Everything has been percent decoded (`%20` -> ` ` etc).
@@ -411,6 +422,9 @@ pub struct IntegrationInfo {
 
     /// The OS native pixels-per-point
     pub native_pixels_per_point: Option<f32>,
+
+    /// Window-specific geometry information, if provided by the platform.
+    pub window_info: Option<WindowInfo>,
 }
 
 // ----------------------------------------------------------------------------
--- a/eframe/src/epi.rs
+++ b/eframe/src/epi.rs
@@ -344,7 +344,7 @@ pub struct WebInfo {
 /// Information about the application's main window, if available.
 #[derive(Clone, Debug)]
 pub struct WindowInfo {
-    /// Coordinates of the window's outer top left corner, relative to the top left corner.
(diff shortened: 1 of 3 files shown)


eframe/src/native/epi_integration.rs
```diff
@@ -8,6 +8,25 @@ pub fn points_to_size(points: egui::Vec2) -> winit::dpi::LogicalSize<f64> {
...
+        Ok(pos) => {
+            let size = window.inner_size();
+            Some(WindowInfo {
+                position: egui::Vec2 {
+                    x: pos.x as f32,
+                    y: pos.y as f32,
+                },
+                size: egui::Vec2 {
+                    x: size.width as f32,
+                    y: size.height as f32,
```
emilk, line 22: `outer_position()` returns physical pixels, but egui uses logical pixels (called "points") almost everywhere. Most importantly, eframe uses points for `Frame::set_window_size` and `Frame::set_window_pos`, so we should be consistent with that methinks.

You can use `window.outer_position().to_logical::<f32>(egui_ctx.pixels_per_point())` to get the size in logical points (egui coordinates).

eframe/src/epi.rs
```diff
@@ -341,6 +341,17 @@ pub struct WebInfo {
     pub location: Location,
 }
 
+/// Information about the application's main window, if available.
+#[derive(Clone, Debug)]
+pub struct WindowInfo {
+    /// Coordinates of the window's top left corner, relative to the top left corner
+    /// of a user's first monitor.
+    pub position: egui::Vec2,
```
emilk, line 349:
```suggestion
    pub position: egui::Pos2,
```

eframe/src/epi.rs
```diff
@@ -341,6 +341,17 @@ pub struct WebInfo {
     pub location: Location,
 }
 
+/// Information about the application's main window, if available.
+#[derive(Clone, Debug)]
+pub struct WindowInfo {
+    /// Coordinates of the window's top left corner, relative to the top left corner
```
emilk, line 347:
```suggestion
    /// Coordinates of the window's outer top left corner, relative to the top left corner.
    /// Unit: egui points.
```

eframe/src/epi.rs
```diff
@@ -341,6 +341,17 @@ pub struct WebInfo {
...
+    /// of a user's first monitor.
+    pub position: egui::Vec2,
+
+    /// Window dimensions, in pixels.
```
emilk, line 351:
```suggestion
    /// Window inner size in egui points.
    ///
    /// Equivalent to `egui_ctx.input().screen_rect().size()`.
```

emilk (changes requested):

eframe/src/native/epi_integration.rs
TicClick: thanks for catching that, fixed.

emilk (approved):

Issue #1633, opened 2022-05-17
Android Support


rib:
It's currently tricky to run egui natively on Android due to some general assumptions that it's possible to create a winit Window and immediately use that to create a graphics context and surface for rendering.

Unlike other winit platforms the native windows on Android are given to the application via lifecycle events and the application will be notified of a new window for each Resumed event and the native window will be invalidated by the next Paused event. (The winit backend on Android will panic if the app does anything that depends on dereferencing the `.native_window()` for a winit window while the app is paused).

The changes required to support this aren't too complicated but it generally means:

- Splitting up surface creation/configuration from graphics context setup
- Deferring graphics context / render state setup until we have our first (native) window (since device selection is technically dependent on being compatible with a given surface)
- On Android: the graphics context should only be lazily initialized when the application is first Resumed.
- On Android: for each Resumed event a new surface should be created, and then dropped on the next Paused event.

Problem areas currently are with the `::new()` constructors for apis like `egui_winit::State` `egui_wgpu::winit::Painter`, `epi::EpiIntegration` and `EguiGlium` which currently all tend to assume that they have a valid native window and can immediately initialize all required render state.

### proposal

It will require some small API changes to the above constructors but it shouldn't be too tricky to remove this assumption and instead allow for deferred initialization on graphics state on platforms where that's needed (i.e. Android).

For example `egui_wgpu::Painter::new()` might only take arguments for a power preference, present mode and msaa sample count and could have a `set_window()` method for asynchronously associating (or disassociating) the painter with a native window. For platforms other than Android this could be called immediately if a window has already been created, but on Android it would be called when handling Resumed and Paused events. Internally the `set_window()` method would handle deferring the creation of all render state until the first time `set_window()` is called, and thereafter would handle just creating a new surface (since the render state should remain valid/reusable).

In the case of `egui_winit::State::new()` it wouldn't be assumed that `max_texture_side` or `pixels_per_point` are known up front and instead there would be setters for both of these that would allow for the `max_texture_side` to be initialized whenever the gpu context gets (lazily) initialized and `pixels_per_point` can be explicitly set each time a new native window is created.

The epi and glium apis would then just need updating in line with those lower-level changes.

For reference I have a branch with the above proposed  `egui_wgpu::Painter` and `egui_winit::State` changes which I've tested on Android (and also on Windows) which I can use to open a draft PR. So far I haven't looked at the epi/glium changes required to really support Android, I just made minimal changes to work with the updated Painter/State APIs.

[...]

PR #1634, opened 2022-05-17
egui_winit/wgpu: enable Android support

rib:
I've recently been looking at using egui to create an Android app that I can use to test a Bluetooth library that I'm building but I found that it would panic due to some assumptions that it's possible to access a native window for initializing render + surface state while the application is paused.

This implements the proposal in #1633 which enables graphics state to be lazily initialized on Android when the application is Resumed (at which point it has a valid native window) as well as support dropping/recreating surface state each time the application is paused/resumed on Android.

Since I'm currently using `egui_winit` and `egui_wgpu` directly this PR only focuses on enabling Android support via these APIs. The corresponding epi/glium changes adapt to the egui_wini/wgpu API changes but have some remaining assumptions about when a window is valid which means they would require some further work to add Android support.

Closes https://github.com/emilk/egui/issues/1633.

### DIFF
--- a/egui-wgpu/CHANGELOG.md
+++ b/egui-wgpu/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to the `egui-wgpu` integration will be noted in this file.
 
 
 ## Unreleased
-
+Enables deferred render + surface state initialization for Android ([#1634](https://github.com/emilk/egui/pull/1634))
 
 ## 0.18.0 - 2022-05-15
 First published version since moving the code into the `egui` repository from <https://github.com/LU15W1R7H/eww>.
--- a/egui-wgpu/src/winit.rs
+++ b/egui-wgpu/src/winit.rs
@@ -1,72 +1,203 @@
+use tracing::error;
+use wgpu::{Adapter, Instance, Surface, TextureFormat};
+
 use crate::renderer;
+
+struct RenderState {
+    device: wgpu::Device,
+    queue: wgpu::Queue,
+    target_format: TextureFormat,
+    egui_rpass: renderer::RenderPass,
+}
+
+struct SurfaceState {
+    surface: Surface,
+    width: u32,
+    height: u32,
+}
 
 /// Everything you need to paint egui with [`wgpu`] on [`winit`].
 ///
 /// Alternatively you can use [`crate::renderer`] directly.
-pub struct Painter {
-    device: wgpu::Device,
-    queue: wgpu::Queue,
-    surface_config: wgpu::SurfaceConfiguration,
-    surface: wgpu::Surface,
-    egui_rpass: renderer::RenderPass,
+pub struct Painter<'a> {
+    power_preference: wgpu::PowerPreference,
+    device_descriptor: wgpu::DeviceDescriptor<'a>,
+    present_mode: wgpu::PresentMode,
+    msaa_samples: u32,
+
+    instance: Instance,
+    adapter: Option<Adapter>,
+    render_state: Option<RenderState>,
+    surface_state: Option<SurfaceState>,
 }
 
-impl Painter {
-    /// Creates a [`wgpu`] surface for the given window, and things required to render egui onto it.
+impl<'a> Painter<'a> {
+    /// Manages [`wgpu`] state, including surface state, required to render egui.
     ///
-    /// # Safety
-    /// The given `window` must outlive the returned [`Painter`].
-    pub unsafe fn new(window: &winit::window::Window, msaa_samples: u32) -> Self {
-        let instance = wgpu::Instance::new(wgpu::Backends::PRIMARY | wgpu::Backends::GL);
-        let surface = instance.create_surface(&window);
-
-        let adapter = pollster::block_on(instance.request_adapter(&wgpu::RequestAdapterOptions {
-            power_preference: wgpu::PowerPreference::HighPerformance,
-            compatible_surface: Some(&surface),
-            force_fallback_adapter: false,
-        }))
-        .unwrap();
-
-        let (device, queue) = pollster::block_on(adapter.request_device(
-            &wgpu::DeviceDescriptor {
-                features: wgpu::Features::default(),
-                limits: wgpu::Limits::default(),
-                label: None,
-            },
-            None,
-        ))
-        .unwrap();
-
-        let size = window.inner_size();
-        let surface_format = surface.get_preferred_format(&adapter).unwrap();
-        let surface_config = wgpu::SurfaceConfiguration {
-            usage: wgpu::TextureUsages::RENDER_ATTACHMENT,
-            format: surface_format,
-            width: size.width as u32,
-            height: size.height as u32,
-            present_mode: wgpu::PresentMode::Fifo, // TODO(emilk): make vsync configurable
-        };
-        surface.configure(&device, &surface_config);
-
-        let egui_rpass = renderer::RenderPass::new(&device, surface_format, msaa_samples);
+    /// Only the [`wgpu::Instance`] is initialized here. Device selection and the initialization
+    /// of render + surface state is deferred until the painter is given its first window target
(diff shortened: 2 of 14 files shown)


lunixbochs:
If we're changing the `wgpu` constructor, can we maybe go a step further and allow the user to fully set `wgpu::Backends`, `wgpu::RequestAdapterOptions` and `wgpu::DeviceDescriptor`?

rib:
yeah, sounds good, it was fairly arbitrary that I just exposed the power preference and present mode

egui-wgpu/CHANGELOG.md
```diff
@@ -1,9 +1,9 @@
-# Changelog for egui_glow
-All notable changes to the `egui-glow` integration will be noted in this file.
+# Changelog for egui_wgpu
+All notable changes to the `egui-wgpu` integration will be noted in this file.
 
 
 ## Unreleased
-
+Enables deferred render + surface state initialization for Android ([#1633]()https://github.com/emilk/egui/issues/1633)
```
emilk, line 6:
```suggestion
Enables deferred render + surface state initialization for Android ([#1633](https://github.com/emilk/egui/pull/1634)).
```

egui-winit/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ All notable changes to the `egui-winit` integration will be noted in this file.
 
 
 ## Unreleased
+Allow deferred render + surface state initialization for Android ([#1633]()https://github.com/emilk/egui/issues/1633)
```
emilk, line 6:
```suggestion
Allow deferred render + surface state initialization for Android ([#1633](https://github.com/emilk/egui/pull/1634)).
```

emilk (changes requested):
Thanks for working on Android support!

I assume the FIXME:s in the code need fixin' before mergin'?

rib:
> Thanks for working on Android support!
> 
> I assume the FIXME:s in the code need fixin' before mergin'?

In this case I added the FiXMEs to code where I was maintaining status quo (where it wont currently work on Android) but I had to make integration changes to use the updated egui_wgpu and egui_winit API.

My intent was to just focus first on supporting Android better via egui_winit + egui_wgpu directly (i'm not cuurently using eframe) but I figured it was worth highlighting where the current integration details for epi / eframe aren't portable to Android.

I can try to take another pass later to also support Android via epi / eframe by addressing the FIXMEs, but would it be ok perhaps to leave them for this PR? Or perhaps just remove for cleaner aesthetics?

egui-wgpu/src/winit.rs
```diff
@@ -37,36 +75,131 @@ impl Painter {
...
+            let adapter =
+                pollster::block_on(self.instance.request_adapter(&wgpu::RequestAdapterOptions {
+                    power_preference: self.power_preference,
+                    compatible_surface: Some(surface),
+                    force_fallback_adapter: false,
+                }))
+                .unwrap();
+
+            self.adapter = Some(adapter);
+        }
```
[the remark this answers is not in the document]
rib: I just tried this out but it's doesn't seem to work out well in terms of borrow checking...

`get_or_insert_with` returns a mutable reference which we don't really want here and that ends up indirectly holding a mutable reference to `self` that then means the adapter initialization needs to be wrapped in a redundant block to make sure the borrow checker will let us ensure the render_state afterwards. Even trying to downgrade the adapter reference to immutable afterwards doesn't work since it's still based on the mutable reference to self.

I guess an API conceptually like `.as_ref_or_insert_with(|| ..)` for setting up the adapter would have been convenient (which wouldn't take a mutable reference), and then the adapter reference could have been used to set up the render_state too.

rib:
> If we're changing the `wgpu` constructor, can we maybe go a step further and allow the user to fully set `wgpu::Backends`, `wgpu::RequestAdapterOptions` and `wgpu::DeviceDescriptor`?

With the updated patches I've just pushed the `Painter::new()` now takes a `wgpu::Backends` mask and `wgpu::DeviceDescriptor` struct and but not `wgpu::RequestAdapterOption`. The adapter options struct has a lifetime that's normally associated with the surface that can be given, which is not expected to be available when calling `Painter::new()`. It still accepts the `wgpu::PowerPreference` so it's currently just the `force_fallback_adapter` adapter option that's not exposed.

egui-wgpu/src/winit.rs
```diff
@@ -1,72 +1,202 @@
...
```
emilk, line 101: What I meant is this:

```suggestion
        self.adapter.get_or_insert_with(|| {
            pollster::block_on(self.instance.request_adapter(&wgpu::RequestAdapterOptions {
                power_preference: self.power_preference,
                compatible_surface: Some(surface),
                force_fallback_adapter: false,
            }))
            .unwrap()
        });
```

we don't care about what `get_or_insert_with` returns, so I don't see the borrow checker issue – but I might me missing something

rib: Yep, sorry I understood what you were suggesting, and it's ok to change the adapter initialization as suggesting but it doesn't work out to also change the render_state to be initialized in the same way which you had also suggested.

Actually I was mistaken in blaming the issue with the `render_state` initialization on the preceding adapter initialization. The problem with the render_state init is that the closure passed to `get_or_insert_with` also needs to borrow self when it calls `self.init_render_state`. (I was thinking the compiler was upset about the borrow returned from the preceding adapter initialization - sorry for the confusion)

So we could apply the change for the first block but then it ends up being inconsistent and then I personally didn't find that the code scanned better. Happy to make the suggested change still for initializing the adapter if you don't mind the inconsistency?

So just to be extra clear. If the code is changed to this:
```rust
    fn ensure_render_state_for_surface(&mut self, surface: &Surface) {
        self.adapter.get_or_insert_with(|| {
            pollster::block_on(self.instance.request_adapter(&wgpu::RequestAdapterOptions {
                power_preference: self.power_preference,
                compatible_surface: Some(surface),
                force_fallback_adapter: false,
            }))
            .unwrap()
        });

        self.render_state.get_or_insert_with(|| {
            let adapter = self.adapter.as_ref().unwrap();
            let swapchain_format = surface.get_preferred_format(adapter).unwrap();

            pollster::block_on(self.init_render_state(adapter, swapchain_format))
        });
    }
```
The compiler will complain like this:
```rust
error[E0502]: cannot borrow `self.render_state` as mutable because it is also borrowed as immutable
   --> C:\Users\Robert\src\egui\egui-wgpu\src\winit.rs:102:9
    |
102 |           self.render_state.get_or_insert_with(|| {
    |           ^                 ------------------ -- immutable borrow occurs here
    |           |                 |
    |  _________|                 immutable borrow later used by call
    | |
103 | |             let adapter = self.adapter.as_ref().unwrap();
104 | |             let swapchain_format = surface.get_preferred_format(adapter).unwrap();
105 | |
106 | |             let rs = pollster::block_on(self.init_render_state(adapter, swapchain_format));
    | |                                         ---- first borrow occurs due to use of `*self` in closure
107 | |             rs
108 | |         });
    | |__________^ mutable borrow occurs here
```

This works ok, even though it's inconsistent:
```rust
    fn ensure_render_state_for_surface(&mut self, surface: &Surface) {
        self.adapter.get_or_insert_with(|| {
            pollster::block_on(self.instance.request_adapter(&wgpu::RequestAdapterOptions {
                power_preference: self.power_preference,
                compatible_surface: Some(surface),
                force_fallback_adapter: false,
            }))
            .unwrap()
        });

        if self.render_state.is_none() {
            let adapter = self.adapter.as_ref().unwrap();
            let swapchain_format = surface.get_preferred_format(adapter).unwrap();

            let rs = pollster::block_on(self.init_render_state(adapter, swapchain_format));
            self.render_state = Some(rs);
        }
    }
```

emilk: ah right, thanks for explaining!

emilk (approved):
Looks good, thanks! Only a CI doc failure that need to be fixed

Issue #186, opened 2021-02-20
Improve documentation


emilk:
Tracking issue for improving the egui docs. Other suggestions for improvements welcome!

All doc-examples need to be part of the doctests (i.e. they should never be marked `ignore`, though `no_run` can maybe sometimes be motivated).

# egui
The crate-level docs for `egui` (at https://docs.rs/egui generated from [`egui/src/lib.rs`](https://github.com/emilk/egui/blob/master/egui/src/lib.rs)) should be improved with a tutorial. In addition, all major parts of egui should have a quick example in the crate-level docs

* Crate-level:
   * [ ] intro tutorial
   * [ ] FAQ (see https://github.com/emilk/egui/issues and https://github.com/emilk/egui/discussions/categories/q-a for what the common questions are)
   * [ ] examples for common widgets, layouts etc (see below)
* Crate-level AND item-level examples for:
  * [ ] `Label`
  * [ ] `Hyperlink`
  * [ ] `TextEdit`
  * [ ] `Button`
  * [ ] `CheckBox`
  * [ ] `RadioButton`
  * [ ] `SelectableLabel`
  * [ ] `combo_box`
  * [ ] `Slider`
  * [ ] `DragValue`
  * [ ] `Separator`
  * [ ] `CollapsingHeader`
  * [ ] `Plot`
  * [ ] horizontal and vertcial layouts
  * [ ] centered vertical layouts
  * [ ] `Grid`
* [ ] Add a few gifs to the docs (not too many though, as they are difficult to maintain)
* Links from the web demo to the docs
  * [x] Link from the widget gallery to the docs
  * [ ] Link from the slider demo to the docs for the slider
  * [ ] …
* Links from the docs to the web demo
  * [x] Link from the crate-level docs to the web demo
  * [ ] Link from the `Slider`-docs to the slider demo window
* [x] Links from the web demo to its own source code

# other
* [ ] Meta: make the CI run all the doc tests
* [ ] Fix the `eframe` doc build issues: https://github.com/emilk/egui/issues/146

olivereanderson:
Thank you for creating this issue Emil :)

Tastaturtaste:
Don't know if you are already aware of it, but [this](https://docs.rs/egui/latest/src/egui/widgets/plot/mod.rs.html#145) Doc example does not compile.
This snippet:
```
use egui::plot::{Line, Plot, Value, Values};

fn main() {

    let sin = (0..1000).map(|i| {
[...]

PR #1635, opened 2022-05-17
Added examples for ComboBox, SelectableLabel and Layouts

Maldus512:
I added a couple of examples (`selection` and `layout`) that cover 3 or 4 of the remaining points from #186 .
The examples are quite simple and contain what I would have needed to get started faster with egui when I first took it up.


emilk:
The `examples` are mainly used for `eframe`-related examples, i.e. things that cannot be demonstrated by `egui_demo_lib` / <www.egui.rs>.

In fact, a lot of the things demoed in this PR is already present at www.egui.rs. Selection is pretty much covered by the widget gallery. There is also the `Layout Test` which could use some love though.

Could you consider improving the existing `egui_demo_lib` instead of adding new demo apps?

PR #1636, opened 2022-05-17
Add the Ability to Specify Egui Texture Filters

zicklag:
Very quick draft, but it works for the `egui_glow` backend.

Not sure that it's a good API or not, and it needs more thought, but pushing this up because I needed it in my project, and other people might want to look at it.

### DIFF
--- a/egui_extras/src/image.rs
+++ b/egui_extras/src/image.rs
@@ -1,3 +1,4 @@
+use egui::epaint::textures::TextureFilter;
 use egui::mutex::Mutex;
 
 /// An image to be shown in egui.
@@ -12,8 +13,8 @@ pub struct RetainedImage {
     image: Mutex<egui::ColorImage>,
     /// Lazily loaded when we have an egui context.
     texture: Mutex<Option<egui::TextureHandle>>,
+    filter: TextureFilter,
 }
-
 impl RetainedImage {
     pub fn from_color_image(debug_name: impl Into<String>, image: ColorImage) -> Self {
         Self {
@@ -21,7 +22,37 @@ impl RetainedImage {
             size: image.size,
             image: Mutex::new(image),
             texture: Default::default(),
+            filter: Default::default(),
         }
+    }
+
+    /// Set the texture filter to use for the image.
+    ///
+    /// **Note:** If the texture has already been uploaded to the GPU, this will require
+    /// re-uploading the texture with the updated filter.
+    ///
+    /// # Example
+    /// ```rust
+    /// # use egui_extras::RetainedImage;
+    /// # use egui::{Color32, epaint::{ColorImage, textures::TextureFilter}};
+    /// # let pixels = vec![Color32::BLACK];
+    /// # let color_image = ColorImage {
+    /// #   size: [1, 1],
+    /// #   pixels,
+    /// # };
+    /// #
+    /// // Upload a pixel art image without it getting blurry when resized
+    /// let image = RetainedImage::from_color_image("my_image", color_image)
+    ///     .with_texture_filter(TextureFilter::Nearest);
+    /// ```
+    pub fn with_texture_filter(mut self, filter: TextureFilter) -> Self {
+        self.filter = filter;
+
+        // If the texture has already been uploaded, this will force it to be re-uploaded with the
+        // updated filter.
+        *self.texture.lock() = None;
+
+        self
     }
 
     /// Load a (non-svg) image.
@@ -86,7 +117,7 @@ impl RetainedImage {
             .get_or_insert_with(|| {
                 let image: &mut ColorImage = &mut self.image.lock();
                 let image = std::mem::take(image);
-                ctx.load_texture(&self.debug_name, image)
+                ctx.load_texture(&self.debug_name, image, self.filter)
             })
             .id()
     }
--- a/egui_glow/src/painter.rs
+++ b/egui_glow/src/painter.rs
@@ -20,20 +20,13 @@ pub use glow::Context;
 const VERT_SRC: &str = include_str!("shader/vertex.glsl");
 const FRAG_SRC: &str = include_str!("shader/fragment.glsl");
 
-#[derive(Copy, Clone)]
-pub enum TextureFilter {
-    Linear,
-    Nearest,
-}
+pub type TextureFilter = egui::epaint::textures::TextureFilter;
 
-impl Default for TextureFilter {
-    fn default() -> Self {
-        TextureFilter::Linear
-    }
+trait TextureFilterExt {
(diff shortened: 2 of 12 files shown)


epaint/src/textures.rs
```diff
@@ -36,9 +36,10 @@ impl TextureManager {
             size: image.size(),
             bytes_per_pixel: image.bytes_per_pixel(),
             retain_count: 1,
+            filter: TextureFilter::Nearest,
```
emilk, line 39:
```suggestion
            filter,
```

epaint/src/texture_atlas.rs
```diff
@@ -176,12 +176,12 @@ impl TextureAtlas {
...
             None
         } else if dirty == Rectu::EVERYTHING {
-            Some(ImageDelta::full(self.image.clone()))
+            Some(ImageDelta::full(self.image.clone(), TextureFilter::Linear))
         } else {
             let pos = [dirty.min_x, dirty.min_y];
             let size = [dirty.max_x - dirty.min_x, dirty.max_y - dirty.min_y];
             let region = self.image.region(pos, size);
-            Some(ImageDelta::partial(pos, region))
+            Some(ImageDelta::partial(pos, region, TextureFilter::default()))
```
emilk, line 184:
```suggestion
            Some(ImageDelta::partial(pos, region, TextureFilter::Linear))
```

better be explicit

egui_glow/src/painter.rs
```diff
@@ -32,6 +32,15 @@ impl Default for TextureFilter {
     }
 }
 
+impl From<egui::epaint::textures::TextureFilter> for TextureFilter {
```
emilk, line 35:
```suggestion
impl From<egui::epaint::textures::TextureFilter> for glow::TextureFilter {
```

for extra clarity

egui_glow/src/painter.rs
```diff
@@ -481,6 +490,7 @@ impl Painter {
 
                 let data: &[u8] = bytemuck::cast_slice(image.pixels.as_ref());
 
+                self.set_texture_filter(delta.filter.into());
```
emilk, line 493: I would prefer the stateful `set_texture_filter` was removed and that the filter was passed by argument to `upload_texture_srgb` - but I can also do that at a later time.

egui_extras/src/image.rs
```diff
@@ -59,8 +73,12 @@ impl RetainedImage {
     /// # Errors
     /// On invalid image
     #[cfg(feature = "svg")]
-    pub fn from_svg_str(debug_name: impl Into<String>, svg_str: &str) -> Result<Self, String> {
-        Self::from_svg_bytes(debug_name, svg_str.as_bytes())
+    pub fn from_svg_str(
+        debug_name: impl Into<String>,
+        svg_str: &str,
+        filter: TextureFilter,
```
emilk, line 79: Instead of passing it in by argument to all constructors I suggest we add a `set_texture_filter` member instead. Most users will just want to use a safe default (linear).

eframe/src/epi.rs
```diff
@@ -213,6 +213,8 @@ pub struct NativeOptions {
 
     /// What rendering backend to use.
     pub renderer: Renderer,
+
+
```
emilk, line 217:
```suggestion
```

emilk (commented):
I think this looks good overall!

I am keen to also add mipmapping as a filtering option, but let's save that for a later PR to keep down the complexity.

egui_extras/src/image.rs
zicklag: Good point, I tried to do a chain-able `with_texture_filter` method, because it will need to be called at the creation of the texture, or more accurately, before the texture is uploaded to the GPU, so I think the chaining is convenient.

egui_glow/src/painter.rs
zicklag: Good point! I'll update that real quick.

egui_glow/src/painter.rs
zicklag: This would be either `crate::painter::TextureFilter` or `self::TextureFilter`, because the struct is defined above.

Do you want it to be either of those?

zicklag (commented):
Awesome!

I addressed most of your points and pushed an update, with just a couple of questions.

**Edit:** I've still got some build problems to fix, I'll get to that as soon as I can.

egui_glow/src/painter.rs
emilk: Oh, I forgot about `egui_glow::TextureFilter`. Let's just remove that and replace it with `pub type TextureFilter = gui::epaint::textures::TextureFilter;`

egui_extras/src/image.rs
```diff
@@ -21,6 +35,36 @@ impl RetainedImage {
...
+    /// # use egui::{Color32, epaint::{ColorImage, textures::TextureFilter}};
+    /// # let pixels = vec![Color32::BLACK];
+    /// # let color_image = ColorImage {
+    /// #   size: [1, 1],
+    /// #   pixels,
+    /// # };
+    /// #
+    /// let image = RetainedImage::from_color_image("my_image", color_image)
+    ///     .with_texture_filter(TextureFilter::Nearest)
+    ///     .unwrap();
```
emilk, line 60: I don't like how it forces people to write `.unwrap()` in correct code (that cannot fail).

It would be possible to implement a `set_texture_filter` which actually changed texture filter. The simple way to implement it is by setting `*self.texture.lock() = None;` which would force the texture to be re-uploaded with the new filter. It is certainly a bit inefficient, but changing filters should be rare and we can optimize it later (by adding `TextureDelta::TextureFilterChanged` or something), and right now it is much nicer than using `Result` and `unwrap` imho.

zicklag: I didn't think about setting `texture = None`, that's a great idea, and much better! :+1:

Pushed an update.

emilk (approved):

egui_glow/src/painter.rs
```diff
@@ -500,12 +484,18 @@ impl Painter {
                     .flat_map(|a| a.to_array())
                     .collect();
 
-                self.upload_texture_srgb(delta.pos, image.size, &data);
+                self.upload_texture_srgb(delta.pos, image.size, delta.filter.into(), &data);
```
emilk, line 487: Is the `.into()` necessary? What does `cargo clippy` think?

zicklag: Whoops, my bad. Fixed it.

zicklag:
Make another push, hopefully that fixes CI, sorry for the re-runs.

emilk:
thanks!

PR #1640, opened 2022-05-18
Take Glow context using Arc.

### LINKED ISSUE #1637: egui_glow::Painter should take Arc<Context> and not Rc<Context>
`egui_glow::Painter::new` takes the Glow GL context using `Rc<Context>`. This forces me to put the context into an `Rc<_>` although I require using an `Arc<_>` to share it between threads. Furthermore, this makes `egui_glow::Painter` not `Send` and `Sync` which makes it use more complicated in a multi-threaded program.

Change `egui_glow::Painter::new` to take `Arc<Context>`.

Or, alternatively, do not store a reference to the context inside it at all and require passing in a `&glow::Context` for each method call.

We require accessing the Glow context from multiple threads due to the way our application is architected. Glow itself supports this, i.e. `Context` is `Send` and `Sync`. I am aware that an OpenGL context can only be current on a single thread at a time. It is however legal to move the OpenGL context between threads.

surban:
This allows usage with a Glow context that is passed between threads.

This is WIP because it first requires merging of https://github.com/asny/three-d/pull/248.

Closes #1637.

### DIFF
--- a/examples/custom_3d_three-d/Cargo.toml
+++ b/examples/custom_3d_three-d/Cargo.toml
@@ -12,4 +12,4 @@ publish = false
 eframe = { path = "../../eframe", features = ["glow"] }
 egui_glow = { path = "../../egui_glow" }
 glow = "0.11"
-three-d = { version = "0.11", default-features = false }
+three-d = { version = "0.12", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,3 +37,6 @@ opt-level = 2 # fast and small wasm, basically same as `opt-level = 's'`
 # opt-level = 3 # unecessarily large wasm for no performance gain
 
 # debug = true # include debug symbols, useful when profiling wasm
+
+[patch.crates-io]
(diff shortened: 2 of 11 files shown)


surban:
https://github.com/asny/three-d/pull/248 has been merged, so this is ready.

Only thing remaining is to remove the override, once three-d version 0.12 is published.

examples/custom_3d_three-d/Cargo.toml
```diff
@@ -12,4 +12,4 @@ publish = false
 eframe = { path = "../../eframe", features = ["glow"] }
 egui_glow = { path = "../../egui_glow" }
 glow = "0.11"
-three-d = { version = "0.11", default-features = false }
+three-d = { version = "0.12", default-features = false }
```
emilk, line 15: I think we should use `git = "https://github.com/asny/three-d.git", rev = "7ac4f3e1e14335290e505a5799a0b88717474678"` here to make it easier for others to copy-paste this example without getting surprise errors!

emilk (commented):
Look good - thanks!

PR #1650, opened 2022-05-20
Graceful exit from web 

### LINKED ISSUE #1642: Work in html container and graceful-stop
I've started migration of my web-app on egui and want to show and stop canvas, so I started some changes.
- resize on 100% of parent container, not screen
- handle with destroy method, and runner-singleton
- is_destroyed flag, unsubscribing on events.

PR is coming

enomado:
- holds all event-handles and gracefully unsubscribes them.
- destroys GL painter.
- tries to resize canvas to previous container's size, not screen size (it should be 100% height thought)

With this PR egui could be running as a part of web-app and stop and start again without resource leaks(I hope)

Fixes #1642

# ⚠️ IMPORTANT!
You need to update your CSS with `html, body: { height: 100%; width: 100%; }`!

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -33,6 +33,8 @@
             overflow: hidden;
             margin: 0 !important;
             padding: 0 !important;
+            height: 100%;
+            width: 100%;
         }
 
         /* Position canvas in center-top: */
--- a/docs/multiple_apps.html
+++ b/docs/multiple_apps.html
@@ -17,6 +17,12 @@
             /* Light mode background color for what is not covered by the egui canvas,
             or where the egui canvas is translucent. */
             background: #909090;
+            display:flex;
+        }
+
+        .canvas_wrap{
+            /* height: 200px; */
+            width: 400px;
         }
 
         @media (prefers-color-scheme: dark) {
@@ -39,13 +45,15 @@
 
         /* Position canvas in center-top: */
         canvas {
-            margin-right: auto;
-            margin-left: auto;
-            display: block;
+            /* margin-right: auto;
+            margin-left: auto; */
+            /* display: block;
             position: absolute;
             top: 0%;
             left: 50%;
-            transform: translate(-50%, 0%);
+            transform: translate(-50%, 0%); */
+            width:90%;
+            height:90%;
         }
 
         .centered {
@@ -96,7 +104,23 @@
 
 <body>
     <!-- The WASM code will resize the canvas dynamically -->
-    <canvas id="the_canvas_id"></canvas>
+
+    <div>controls</div>
+
+    <button class="stop_one">
+    stop    
+    </button>
+
+    <div class="canvas_wrap one">
+        <canvas id="the_canvas_id_one"></canvas>
+    </div>
+
+    <div  class="canvas_wrap two">
+        <canvas id="the_canvas_id_two"></canvas>
+    </div>
+
+
+
     <div class="centered" id="center_text">
         <p style="font-size:16px">
             Loading…
@@ -133,11 +157,23 @@
             console.debug("wasm loaded. starting app…");
 
             // This call installs a bunch of callbacks and then returns:
-            const handle = wasm_bindgen.start("the_canvas_id");
+
+            wasm_bindgen.init_wasm_hooks()
+
+            const handle_one = wasm_bindgen.start_separate("the_canvas_id_one");
+            const handle_two = wasm_bindgen.start_separate("the_canvas_id_two");
+
+            const button = document.getElementsByClassName("stop_one")[0]
+
+            button.addEventListener("click", ()=>{
+                handle_one.stop_web()
+                handle_one.free()
+            });
+    
 
             // call `handle.stop_web()` to stop
             // uncomment to quick result
-            // setTimeout(() => {handle.stop_web(); handle.free())}, 2000)
+            // setTimeout(() => {handle.stop_web()}, 2000)
 
             console.debug("app started.");
             document.getElementById("center_text").remove();
--- a/eframe/src/lib.rs
+++ b/eframe/src/lib.rs
@@ -103,7 +103,7 @@ pub use web_sys;
 /// /// This is the entry-point for all the web-assembly.
 /// /// This is called from the HTML.
 /// /// It loads the app, installs some callbacks, then returns.
-/// /// It creates singleton app-handle that could be stopped calling `stop_web`
+/// /// It returns a handle to the running app that can be stopped calling `AppRunner::stop_web`.
 /// /// You can add more callbacks like this if you want to call in to your code.
 /// #[cfg(target_arch = "wasm32")]
 /// #[wasm_bindgen]
--- a/eframe/src/lib.rs
+++ b/eframe/src/lib.rs
@@ -77,11 +77,19 @@ pub use epi::*;
 // When compiling for web
 
 #[cfg(target_arch = "wasm32")]
-mod web;
+use egui::mutex::Mutex;
+
+#[cfg(target_arch = "wasm32")]
+use std::sync::Arc;
+
+#[cfg(target_arch = "wasm32")]
+pub mod web;
 
 #[cfg(target_arch = "wasm32")]
 pub use wasm_bindgen;
 
+#[cfg(target_arch = "wasm32")]
+use web::AppRunner;
 #[cfg(target_arch = "wasm32")]
 pub use web_sys;
 
@@ -93,12 +101,13 @@ pub use web_sys;
 /// use wasm_bindgen::prelude::*;
 ///
 /// /// This is the entry-point for all the web-assembly.
-/// /// This is called once from the HTML.
+/// /// This is called from the HTML.
 /// /// It loads the app, installs some callbacks, then returns.
+/// /// It creates singleton app-handle that could be stopped calling `stop_web`
 /// /// You can add more callbacks like this if you want to call in to your code.
 /// #[cfg(target_arch = "wasm32")]
 /// #[wasm_bindgen]
-/// pub fn start(canvas_id: &str) -> Result<(), eframe::wasm_bindgen::JsValue> {
+/// pub fn start(canvas_id: &str) -> Result<Arc<Mutex<AppRunner>>, eframe::wasm_bindgen::JsValue> {
 ///     let web_options = eframe::WebOptions::default();
 ///     eframe::start_web(canvas_id, web_options, Box::new(|cc| Box::new(MyEguiApp::new(cc))))
 /// }
@@ -108,9 +117,10 @@ pub fn start_web(
     canvas_id: &str,
     web_options: WebOptions,
     app_creator: AppCreator,
-) -> Result<(), wasm_bindgen::JsValue> {
-    web::start(canvas_id, web_options, app_creator)?;
-    Ok(())
+) -> Result<Arc<Mutex<AppRunner>>, wasm_bindgen::JsValue> {
+    let handle = web::start(canvas_id, web_options, app_creator)?;
+
+    Ok(handle)
 }
 
 // ----------------------------------------------------------------------------
--- a/eframe/src/web/backend.rs
+++ b/eframe/src/web/backend.rs
@@ -2,9 +2,10 @@ use super::{glow_wrapping::WrappedGlowPainter, *};
 
 use crate::epi;
 
-use egui::mutex::{Mutex, MutexGuard};
-use egui::TexturesDelta;
-
+use egui::{
+    mutex::{Mutex, MutexGuard},
+    TexturesDelta,
+};
 pub use egui::{pos2, Color32};
 
 // ----------------------------------------------------------------------------
@@ -64,6 +65,24 @@ impl NeedRepaint {
 
     pub fn repaint_asap(&self) {
         *self.0.lock() = f64::NEG_INFINITY;
+    }
+}
+
+pub struct IsDestroyed(std::sync::atomic::AtomicBool);
+
+impl Default for IsDestroyed {
+    fn default() -> Self {
+        Self(false.into())
+    }
+}
+
+impl IsDestroyed {
+    pub fn fetch(&self) -> bool {
+        self.0.load(SeqCst)
+    }
+
+    pub fn set_true(&self) {
+        self.0.store(true, SeqCst);
     }
 }
 
@@ -147,11 +166,19 @@ pub struct AppRunner {
     pub(crate) input: WebInput,
     app: Box<dyn epi::App>,
     pub(crate) needs_repaint: std::sync::Arc<NeedRepaint>,
+    pub(crate) is_destroyed: std::sync::Arc<IsDestroyed>,
     last_save_time: f64,
     screen_reader: super::screen_reader::ScreenReader,
     pub(crate) text_cursor_pos: Option<egui::Pos2>,
     pub(crate) mutable_text_under_cursor: bool,
     textures_delta: TexturesDelta,
+    pub events_to_unsubscribe: Vec<EventToUnsubscribe>,
+}
+
+impl Drop for AppRunner {
+    fn drop(&mut self) {
+        tracing::debug!("AppRunner has fully dropped");
+    }
 }
 
 impl AppRunner {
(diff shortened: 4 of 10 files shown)


eframe/src/lib.rs
```diff
@@ -95,9 +100,44 @@ pub use web_sys;
...
+
+#[cfg(target_arch = "wasm32")]
+unsafe impl Send for AppRunner {}
+
+#[cfg(target_arch = "wasm32")]
+unsafe impl Sync for AppRunner {}
+
+#[cfg(target_arch = "wasm32")]
+static GLOBAL_HANDLE: ArcSwapOption<Mutex<AppRunner>> = ArcSwapOption::const_empty();
+
```
emilk, line 114: You moved the docstring for `start_web` to now document `use arc_swap::ArcSwapOption;`.

eframe/src/lib.rs
```diff
@@ -95,9 +100,44 @@ pub use web_sys;
...
+#[cfg(target_arch = "wasm32")]
+pub fn stop_web() -> Result<(), wasm_bindgen::JsValue> {
+    let u = GLOBAL_HANDLE.load_full();
+
+    if let Some(uu) = u {
+        {
+            let mut oo = uu.lock();
+            oo.destroy();
+        }
```
emilk, line 132:
```suggestion
        uu.lock().destroy();
```

eframe/src/lib.rs
```diff
@@ -95,9 +100,44 @@ pub use web_sys;
...
+    GLOBAL_HANDLE.swap(Some(handle));
+
+    Ok(())
+}
+
+#[cfg(target_arch = "wasm32")]
+pub fn stop_web() -> Result<(), wasm_bindgen::JsValue> {
+    let u = GLOBAL_HANDLE.load_full();
+
+    if let Some(uu) = u {
```
emilk, line 128: Please use names, not letters

eframe/src/web/backend.rs
```diff
@@ -372,14 +453,22 @@ impl AppRunnerContainer {
...
+            target: target.clone(),
+            event_name: event_name.to_string(),
+            closure,
+        };
+
+        self.events.push(EventToUnsubscribe::TargetEvent(handle));
+
         // Bypass closure drop so that event handler can call the closure
-        closure.forget();
+        // closure.forget();
```
emilk, line 471: mistake?

eframe/src/web/backend.rs
```diff
@@ -398,24 +487,30 @@ pub fn start(canvas_id: &str, app_creator: epi::AppCreator) -> Result<AppRunnerR
...
     super::events::paint_and_schedule(&runner_container.runner, runner_container.panicked.clone())?;
 
     // Disable all event handlers on panic
     let previous_hook = std::panic::take_hook();
-    let panicked = runner_container.panicked;
+
+    {
+        let mut runner_locked = runner_container.runner.lock();
+        runner_locked.events_to_unsubscribe = runner_container.events;
+    }
```
emilk, line 509:
```suggestion
    runner_container.runner.lock().events_to_unsubscribe = runner_container.events;
```

eframe/src/web/mod.rs
```diff
@@ -38,6 +39,7 @@ pub fn now_sec() -> f64 {
         / 1000.0
 }
 
+#[allow(dead_code)]
```
emilk, line 42: remove dead code instead

eframe/src/web/backend.rs
```diff
@@ -340,18 +380,59 @@ impl AppRunner {
...
+
+        match self {
+            EventToUnsubscribe::TargetEvent(handle) => {
+                handle
+                    .target
+                    .remove_event_listener_with_callback(
+                        handle.event_name.as_str(),
+                        handle.closure.as_ref().unchecked_ref(),
+                    )
+                    .unwrap();
```
emilk, line 411: Perhaps safer with an `.ok()` here - do we really want to panic if this fails?

eframe/src/lib.rs
```diff
@@ -95,9 +100,44 @@ pub use web_sys;
...
+use arc_swap::ArcSwapOption;
+
+#[cfg(target_arch = "wasm32")]
+unsafe impl Send for AppRunner {}
+
+#[cfg(target_arch = "wasm32")]
+unsafe impl Sync for AppRunner {}
+
+#[cfg(target_arch = "wasm32")]
+static GLOBAL_HANDLE: ArcSwapOption<Mutex<AppRunner>> = ArcSwapOption::const_empty();
```
emilk, line 113: Adding a dependency on `ArcSwapOption` seems a bit unnecessary to me. Perhaps `Mutex<Option<AppRunnerRef>>` could do?

eframe/src/lib.rs
```diff
@@ -95,9 +100,44 @@ pub use web_sys;
...
+
 #[cfg(target_arch = "wasm32")]
 pub fn start_web(canvas_id: &str, app_creator: AppCreator) -> Result<(), wasm_bindgen::JsValue> {
-    web::start(canvas_id, app_creator)?;
+    let handle = web::start(canvas_id, app_creator)?;
+
+    GLOBAL_HANDLE.swap(Some(handle));
```
emilk, line 119: what should be the semantics of calling `start` many times? Should we warn about it? Destroy the previous instances first?

eframe/src/lib.rs
emilk, line 128:
```suggestion
    if let Some(app_runner) = GLOBAL_HANDLE.load_full() {
```

emilk (changes requested):
Did a quick read-through. Thanks for working on this!

eframe/src/web/mod.rs
enomado: I thought it could be useful in future. but okay

eframe/src/lib.rs
enomado: I dont know how to do it. I have `Arc<Mutex<AppRunnerRef>>` how I can  place it in static variable?

enomado: Good option could be return app handle to js, but singleton is good too, for me

emilk: Returning it to JS is much nicer actually, since we can then also support multiple egui:s from the same WASM.

emilk:
Did you try returning the handle from `start_web`? I think it would be a much nicer design (since it can support multiple apps in the same WASM), but I can understand if you feel it is outside the scope of this PR.

enomado:
That was planned at first, but I was not sure I understand how it will be managed by garbage collector, but looks like its fine. I'll do it, but cant find time for now.

enomado:
I've made this dont use global variable

eframe/src/web/backend.rs
```diff
@@ -367,18 +408,59 @@ impl AppRunner {
...
+
+pub struct RepetitativeHandle {
+    pub handle: i32,
+    pub closure: Closure<dyn FnMut()>,
+}
+
+pub enum EventToUnsubscribe {
+    TargetEvent(TargetEvent),
+    #[allow(dead_code)]
+    RepititativeHandle(RepetitativeHandle),
```
emilk, line 425: remove dead code?

eframe/src/web/mod.rs
```diff
@@ -96,13 +98,21 @@ pub fn canvas_size_in_points(canvas_id: &str) -> egui::Vec2 {
...
     let canvas = canvas_element(canvas_id)?;
+    let parent = canvas.parent_element()?;
+
+    let width = parent.scroll_width();
+    let height = parent.scroll_height();
+
+    let canvas_real_size = Vec2 {
+        x: width as f32,
+        y: height as f32,
+    };
```
emilk, line 109: On my Firefox and Chrominum on Mac, this gives a size of `2001 x 0` which makes nothing visible at all.

This should be reverted to the old code (calling `screen_size_in_native_points`), especially as I don't see how this is related to a graceful exit!

egui_demo_app/src/lib.rs
```diff
@@ -19,23 +28,48 @@ pub(crate) fn seconds_since_midnight() -> f64 {
 #[cfg(target_arch = "wasm32")]
 use eframe::wasm_bindgen::{self, prelude::*};
 
+#[cfg(target_arch = "wasm32")]
+#[wasm_bindgen]
+pub struct WebHandle {
+    handle: Arc<Mutex<AppRunner>>,
+}
```
emilk, line 35: This `WenHandle` wrapper should be moved into `eframe` instead so that all `eframe` users can stop their instances!

emilk (changes requested):

emilk:
> I've made this dont use global variable

Thanks ❤️

eframe/src/web/backend.rs
```diff
@@ -367,18 +408,59 @@ impl AppRunner {
 
 pub type AppRunnerRef = Arc<Mutex<AppRunner>>;
 
+pub struct TargetEvent {
+    target: EventTarget,
+    event_name: String,
+    closure: Closure<dyn FnMut(web_sys::Event)>,
+}
+
+pub struct RepetitativeHandle {
```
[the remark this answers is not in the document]
enomado: Its a feature. It can be used in future. For possible events in future. It was used for that that rerendering each each second event, which was removed for some reason.

eframe/src/web/mod.rs
enomado: > On my Firefox and Chrominum on Mac, this gives a size of `2001 x 0` which makes nothing visible at all.

The idea was to control width of canvas by wrapping in by div, and set width of that div manually.



> This should be reverted to the old code (calling `screen_size_in_native_points`), especially as I don't see how this is related to a graceful exit!

Its all about having control of egui from web. To embed egui in div, not full-frame document.

egui_demo_app/src/lib.rs
enomado: Again, I tried to make egui embeddable, not just run in whole document. WebHandle is extensible. You have handle in js and you can call some methods to interact with egui app

eframe/src/web/backend.rs
```diff
@@ -267,6 +296,18 @@ impl AppRunner {
         Ok(())
     }
 
+    pub fn destroy(&mut self) -> Result<(), JsValue> {
+        tracing::debug!("destroying...");
```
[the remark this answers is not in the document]
enomado: Makes sense, but I'm not sure how it should be. panic? return "false" ? Perhaps, it could be added later

eframe/src/web/events.rs
```diff
@@ -5,9 +5,11 @@ pub fn paint_and_schedule(
     runner_ref: &AppRunnerRef,
     panicked: Arc<AtomicBool>,
 ) -> Result<(), JsValue> {
-    fn paint_if_needed(runner_ref: &AppRunnerRef) -> Result<(), JsValue> {
+    fn paint_if_needed(runner_ref: &AppRunnerRef) -> Result<bool, JsValue> {
```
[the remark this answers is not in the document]
enomado: `struct IsDestroyed(pub bool);
Result<IsDestroyed, JsValue>
`
?

eframe/src/web/backend.rs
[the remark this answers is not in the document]
emilk: The safest is to log a warning and return. No need to cause a panic

eframe/src/web/mod.rs
emilk: That may be, but the new code breaks the the web demo.

eframe/src/web/events.rs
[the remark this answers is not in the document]
emilk: Yeah, that's an improvement imho!

eframe/src/web/mod.rs
enomado: That' is why i made this https://github.com/emilk/eframe_template/pull/60

Anyway, may be I add fallback to screen_size_in_native_points

enomado:
Btw, what do you think on adding fn as_any_mut(&mut self) -> &mut dyn Any; in App trait?

I need it to get original App from WebHandle

eframe/src/web/events.rs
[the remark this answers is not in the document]
Mingun: Type alias also could be used, similar to [yew](https://yew.rs) approach in the past (now [removed](https://github.com/yewstack/yew/pull/2011)), but I think, that just documenting should be enough

eframe/src/web/mod.rs
enomado: Oh. Okay. I'll try to change it too

enomado:
Fixed Web, IntervalHandle, looks like most of issues. Quick checked and seems everything is fine

enomado:
squashed and rebased

eframe/src/web/mod.rs
emilk: It works now ❤️ thanks!

docs/index.html
```diff
@@ -33,6 +33,8 @@
             overflow: hidden;
             margin: 0 !important;
             padding: 0 !important;
+            height: 100%;
+            width: 100%;
```
emilk, line 37: This is an important (breaking!) change that should be documented in `eframe/CHANGELOG.md`

eframe/src/lib.rs
```diff
@@ -93,12 +101,13 @@ pub use web_sys;
 /// use wasm_bindgen::prelude::*;
 ///
 /// /// This is the entry-point for all the web-assembly.
-/// /// This is called once from the HTML.
+/// /// This is called from the HTML.
 /// /// It loads the app, installs some callbacks, then returns.
+/// /// It creates singleton app-handle that could be stopped calling `stop_web`
```
emilk, line 106:
```suggestion
/// /// It returns a handle to the running app that can be stopped calling `AppRunner::stop_web`.
```

I think it is also worth documenting whether or not it is safe to ignore the returned handle. As it is an `Arc`, I as a user would be afraid that dropping it would free the last handle, thus stopping the egui app.

egui_demo_app/src/lib.rs
emilk: Sure, but what is the point of defining `WebHandle` in the example app? I don't see `stop_web()` being used anywhere. Is it dead code? Should `docs/index.html` store the handle as a demonstration?

emilk (changes requested):
Looks good! Still needs a line in `eframe/CHANGELOG.md` noting the breaking changes (in particular the need to add `body { height: 100%; width: 100% }` in the CSS).

All lines in the changelog should link to the relevant PR, and so it would be great if the PR description could be updated with:

A) breaking changes (things all users must do when updating `eframe`)
B) a short summary or example of what the new features enable (having two egui instances in the same web page), and how to achieve it (does the user need need to call `stop_web` form JS when deleting a `canvas` with eframe running in it?)

emilk:
> squashed and rebased

thanks for the rebase, but please don't squash - it is easier to follow changes via commit history. I always squash when merging anyway, so the history will still end up clean!

egui_demo_app/src/lib.rs
enomado: Well. Ok. I'll create another wrapper in eframe

enomado:
> thanks for the rebase, but please don't squash - it is easier to follow changes via commit history. I always squash when merging anyway, so the history will still end up clean!

got it

enomado:
>I don't see stop_web() being used anywhere. Is it dead code? Should docs/index.html store the handle as a demonstration?

it works in scenario where egui app is just a canvas in normal web page. I'll create another page

emilk:
> it works in scenario where egui app is just a canvas in normal web page. I'll create another page

Well, that's exactly what the web-demo in `index.html` is too, just that the canvas fills the screen.

enomado:
> Well, that's exactly what the web-demo in `index.html` is too, just that the canvas fills the screen.

I mean we need html with some button to stop app

enomado:
Added page with stop button on http://127.0.0.1:8888/multiple_contexts.html

docs/multiple_apps.html
```diff
@@ -0,0 +1,200 @@
...
+                <p style="font-size:14px">
+                    Make sure you use a modern browser with WebGL and WASM enabled.
+                </p>`;
+        }
+    </script>
+</body>
+
+</html>
+
+<!-- Powered by egui: https://github.com/emilk/egui/ -->
```
emilk, line 200: The example is not very pretty, but it works for now :)

Screen Shot 2022-08-02 at 17 31 04

emilk (approved):
Thanks for working on this!

PR #1653, opened 2022-05-20
Make Popups Dodge Window Edges

zicklag:
I was using a popup to show a fixed-pallet color picker, using `popup_below_widget`, but it was getting cut off because it was close to the right side of the window and it was a wide popup:

image

By changing this line to use `.default_pos` instead of `.fixed_pos`, it fixes my issue, and seems like it would be a good improvement, but I'm not sure if there are side-effects I'm unaware of.

https://github.com/emilk/egui/blob/3c685d7bf6a36bda9ffbee36ee822b4205b29b36/egui/src/containers/popup.rs#L300

### DIFF
--- a/egui/src/containers/popup.rs
+++ b/egui/src/containers/popup.rs
@@ -297,7 +297,7 @@ pub fn popup_below_widget<R>(
     if ui.memory().is_popup_open(popup_id) {
         let inner = Area::new(popup_id)
             .order(Order::Foreground)
-            .fixed_pos(widget_response.rect.left_bottom())
+            .default_pos(widget_response.rect.left_bottom())
             .show(ui.ctx(), |ui| {
                 // Note: we use a separate clip-rect for this area, so the popup can be outside the parent.
                 // See https://github.com/emilk/egui/issues/825


emilk:
There are some downsides yes:

May-21-2022 13-08-15

zicklag:
Ah, OK, then that doesn't make sense for for the default setting. I wonder if we could set it up so it uses the logic to doge the window border like when we set `default_pos`, but it doesn't allow you to move the popup, and if  you move the widget and re-open the dropdown or something, it makes sure to recalculate the position relative to the widget. If this ends up bugging me, too much I'll try to look into doing that.

Anyway, I'll close this for now.

zicklag:
I was able to get something working in my project with this code:

```rust
use egui::{Align, Area, Frame, Id, Key, Layout, NumExt, Order, Rect, Response, Ui, Vec2};

#[derive(Clone, Default)]
struct State {
    size: Vec2,
}

/// Like `egui::popup_under_widget, but pops up to the left, so that the popup doesn't go off the screen
pub(crate) fn popup_under_widget<R>(
    ui: &Ui,
    popup_id: Id,
    widget_response: &Response,
    add_contents: impl FnOnce(&mut Ui) -> R,
) -> Option<R> {
    if ui.memory().is_popup_open(popup_id) {
        let state: Option<State> = ui.data().get_temp(popup_id);

        // If this is the first draw, we don't know the popup size yet, so we don't know how to
        // position the popup
        if state.is_none() {
            ui.ctx().request_repaint();
        }

        let mut state = state.unwrap_or_default();

        let rect = Rect {
            min: widget_response.rect.left_bottom(),
            max: widget_response.rect.left_bottom() + state.size,
        };
        let inner = Area::new(popup_id)
            .order(Order::Foreground)
            .fixed_pos(constrain_window_rect_to_area(ui.ctx(), rect, None).min)
            .movable(true)
            .show(ui.ctx(), |ui| {
                // Note: we use a separate clip-rect for this area, so the popup can be outside the parent.
                // See https://github.com/emilk/egui/issues/825
                let frame = Frame::popup(ui.style());
                let frame_margin = frame.inner_margin + frame.outer_margin;
                let result = frame
                    .show(ui, |ui| {
                        ui.with_layout(Layout::top_down_justified(Align::LEFT), |ui| {
                            ui.set_width(widget_response.rect.width() - frame_margin.sum().x);
                            add_contents(ui)
                        })
                        .inner
                    })
                    .inner;

                state.size = ui.min_rect().size();

                result
            })
            .inner;

        *ui.data().get_temp_mut_or_default(popup_id) = state;

        if ui.input().key_pressed(Key::Escape) || widget_response.clicked_elsewhere() {
            ui.memory().close_popup();
        }
        Some(inner)
    } else {
        None
    }
}

/// I just had to copy this function below from egui because it was a private function on `egui::Context`

/// Constrain the position of a window/area so it fits within the provided boundary.
///
/// If area is `None`, will constrain to [`ctx::available_rect`].
pub(crate) fn constrain_window_rect_to_area(
    ctx: &egui::Context,
    window: Rect,
    area: Option<Rect>,
) -> Rect {
    let mut area = area.unwrap_or_else(|| ctx.available_rect());

    if window.width() > area.width() {
        // Allow overlapping side bars.
        // This is important for small screens, e.g. mobiles running the web demo.
        area.max.x = ctx.input().screen_rect().max.x;
        area.min.x = ctx.input().screen_rect().min.x;
    }
    if window.height() > area.height() {
        // Allow overlapping top/bottom bars:
        area.max.y = ctx.input().screen_rect().max.y;
        area.min.y = ctx.input().screen_rect().min.y;
    }

    let mut pos = window.min;

    // Constrain to screen, unless window is too large to fit:
    let margin_x = (window.width() - area.width()).at_least(0.0);
    let margin_y = (window.height() - area.height()).at_least(0.0);

    pos.x = pos.x.at_most(area.right() + margin_x - window.width()); // move left if needed
    pos.x = pos.x.at_least(area.left() - margin_x); // move right if needed
    pos.y = pos.y.at_most(area.bottom() + margin_y - window.height()); // move right if needed
    pos.y = pos.y.at_least(area.top() - margin_y); // move down if needed

    // pos = ctx.round_pos_to_pixels(pos);

    Rect::from_min_size(pos, window.size())
}
```

It copies the pattern that the Area uses to position itself within the window borders. Do you think that's a reasonable approach to build into the popup, or is there maybe a better way to do it?

PR #1665, opened 2022-05-22
Added support for 20 fn keys

sy1ntexx:
This PR adds support for 20 fn keys. I chose 20 keys because that's how many keys are defined in Mozilla's web documentation. Winit supports all 20 keys and 4 more, that are not included in this pr.

### DIFF
--- a/egui/src/data/input.rs
+++ b/egui/src/data/input.rs
@@ -498,6 +498,26 @@ pub enum Key {
     X,
     Y,
     Z, // Used for cmd+Z (undo)
+    F1,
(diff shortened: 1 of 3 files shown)


egui/src/data/input.rs
```diff
@@ -498,6 +498,28 @@ pub enum Key {
     X,
     Y,
     Z, // Used for cmd+Z (undo)
+    
+    // The function keys:
```
emilk, line 502:
```suggestion

    // The function keys:
```

PR #1671, opened 2022-05-23
Add some padding to the egui-wgpu uniform buffer for WebGL

expenses:
I'm using egui-wgpu with the wgpu's WebGL backend. At the moment I get the following error on Chromium:

```
[.WebGL-0x38140334e800] GL_INVALID_OPERATION: It is undefined behaviour to use a uniform buffer that is too small.
```

with nothing rendering. I've encountered this issue before and have added some padding to the `UniformBuffer` struct. There is now no warning and it renders normally.

### DIFF
--- a/egui-wgpu/src/renderer.rs
+++ b/egui-wgpu/src/renderer.rs
@@ -554,7 +554,7 @@ impl RenderPass {
             0,
             bytemuck::cast_slice(&[UniformBuffer {
                 screen_size_in_points,
-                _padding: Default::default()
+                _padding: Default::default(),
             }]),
         );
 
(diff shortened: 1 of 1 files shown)


egui-wgpu/src/renderer.rs
```diff
@@ -38,6 +38,7 @@ impl ScreenDescriptor {
 #[repr(C)]
 struct UniformBuffer {
     screen_size_in_points: [f32; 2],
+    _padding: [u32; 2],
```
Mingun, line 41: You can write your findings, that you describe in ticket, here as doc comment for further readers, because otherwise intentions of this field a very unclear!

emilk: 100%

```suggestion
    _padding: [u32; 2], // Required on Chromium, see https://github.com/emilk/egui/pull/1671
```

emilk (approved):

emilk:
`cargo fmt` is unhappy

PR #1680, opened 2022-05-24
Fix uneven table striping

### LINKED ISSUE #1679: Striped table has uneven striping
**Describe the bug**

When using `TableBuilder` from `egui_extras` with `.striped(true)`, the stripes are drawn unevenly making the rows look like they are different heights:
image

Here's an edited image showing the difference by overlaying a striped row with a non-striped row:
image

**To Reproduce**
1. Open the [egui demo](egui.rs)
2. Open the Table demo
3. Select "Thousands of rows of same height"
4. Compare the height of dark vs light rows

**Expected behavior**

All rows should be equal height.

**Screenshots**

 - OS: Arch Linux 5.17.4-arch1-1
 - Firefox 100.0b8
 - Master branch

auderer:
Closes #1679 

The problem was caused by the painted stripes not being expanded to include vertical spacing. The rows are the correct size, just painted with too small of stripes.

Before:
image

After:
image

Edited to show overlay:
image

### DIFF
--- a/egui_extras/src/layout.rs
+++ b/egui_extras/src/layout.rs
@@ -110,7 +110,7 @@ impl<'l> StripLayout<'l> {
         let rect = self.cell_rect(&width, &height);
 
         // Make sure we don't have a gap in the stripe background:
-        let rect = rect.expand2(egui::vec2(0.5 * self.ui.spacing().item_spacing.x, 0.0));
+        let rect = rect.expand2(egui::vec2(0.5 * self.ui.spacing().item_spacing.x, 0.5 * self.ui.spacing().item_spacing.y));
 
         self.ui
             .painter()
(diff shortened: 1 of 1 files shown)


egui_extras/src/layout.rs
```diff
@@ -110,7 +110,7 @@ impl<'l> StripLayout<'l> {
         let rect = self.cell_rect(&width, &height);
 
         // Make sure we don't have a gap in the stripe background:
-        let rect = rect.expand2(egui::vec2(0.5 * self.ui.spacing().item_spacing.x, 0.0));
+        let rect = rect.expand2(egui::vec2(0.5 * self.ui.spacing().item_spacing.x, 0.5 * self.ui.spacing().item_spacing.y));
```
emilk, line 113:
```suggestion
        let rect = rect.expand2(0.5 * self.ui.spacing().item_spacing);
```

emilk (changes requested):

Issue #1095, opened 2022-01-11
Allow opting out of the dedicated GPU on macOS


terhechte:
**Is your feature request related to a problem? Please describe.**

There's an issue on [Postsack](https://github.com/terhechte/postsack/issues/32) about high battery consumption of the app when run on MacBook Pros that contain a dedicated GPU. Basically by default any `egui` app switches macOS to the more powerful dedicated GPU. In most cases, though, the internal GPU (of the Intel chipset) would be enough for rendering the app.

Glium supports this via the `with_hardware_acceleration` call. However, `egui_glium` [currently doesn't expose this](https://github.com/emilk/egui/blob/master/egui_glium/src/epi_backend.rs#L17).

**Describe the solution you'd like**

My proposal would be to add an additional `requires_dedicated_gpu` field to [`epi::NativeOptions`](https://docs.rs/epi/0.16.0/epi/struct.NativeOptions.html) and then use that in the `epi_backend`. The default value could be `true` but it would be nice to be able to opt out. The whole thing could also be limited to macOS via `target_os = "macos"`. I'd be willing to implement this in a PR if this solution is approved.

**Describe alternatives you've considered**

Not implementing it would make any egui app on most of the older MacBook Pros out there consume their battery faster (also, heat up and turn on the fans I guess). 

**Additional context**

https://github.com/rust-windowing/glutin/issues/980

emilk:
The new default native backend for `eframe` is [`glow`](https://github.com/grovesNL/glow), so we need to look at how that could be accomplished there.

hacknus:
is there an update to this? I am also interested in a solution for that, but I have no idea where the problematic line is in the glow backend...

coderedart:
it seems glow has nothing to do with this. this seems to be a opengl context creation / loading thing. glow just wraps around the already loaded functions with more rusty types. 

glium uses glutin. egui also uses glutin https://github.com/emilk/egui/blob/master/eframe/src/native/run.rs#L19 . 
just need to add the option and pass it to the glutin builder ^^ and this should work.

terhechte:
Thanks @coderedart I hadn't seen this earlier, I've created a PR to add the functionality to run.rs :)

https://github.com/emilk/egui/pull/1681

Issue #1358, opened 2022-03-12
eframe X11 crash after using thread spawn


griffi-gh:
**Describe the bug**
Random crash 1 out of 4 times when spawning a thread with `thread::spawn` on X11
```
[xcb] Unknown sequence number while processing queue
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
rustyboi: ../../src/xcb_io.c:269: poll_for_event: Assertion `!xcb_xlib_threads_sequence_lost' failed.
```

**To Reproduce**
- Create a simple `eframe` app
- Build it, it shoul start just fine
- Add `thread::spawn` with `loop{}`
- Build it again, it should start randomly crashing on start

**Expected behavior**
N/A

**Screenshots**
![изображение](https://user-images.githubusercontent.com/45996170/158017169-a3eaea24-500d-45cb-a95b-4956489bba5c.png)

**Desktop (please complete the following information):**
 - OS: Pop OS + KDE Plasma X11 (gdm3)
 - Browser: N/A

Vaimer9:
Same here 
![image](https://user-images.githubusercontent.com/81609819/158060928-123c5eab-604f-4953-ac31-0fd57d709cf6.png)
I am only using the code in [the eframe example](https://github.com/emilk/egui/blob/master/eframe/examples/hello_world.rs)

emilk:
I think this is a problem that should be reported to `winit`

uniconductive:
Have same on Arch Linux x64 (Manjaro), X11, Intel HD 3000, mesa 21.3.7-2. 
This error is there 8 months as minimum.

emilk:
Anyone wants to dig into this? Likely culprits include `glutin`, `glow`, `winit`, or a combination of the three.

uniconductive:
I have this in Cargo.toml:
eframe = { version = "0.17.0", default-features = false, features = ["default_fonts", "egui_glow"] }
I see no way to set glutin...

Ar37-rs:
same issue. i suspect there's an issue somewhere inside glutin, i tried (ubuntu/latest mesa driver [from obiaf repo]) using both fltk and sdl with glow... even winit with egui wgpu backend, everything's working fine.

ambihelical:
Just as a data point I just upgraded to a much faster machine (from circa 2009 4 core i7 to 16 core ryzen), and my egui app does this _every_ time.  I have yet to see it run :-(

[...]

PR #1681, opened 2022-05-25
Add flag to disable hardware acceleration

terhechte:
Closes <https://github.com/emilk/egui/issues/1095>

This is a fix for the behaviour on macOS platforms where any egui app would use the dedicated GPU and consume more power than needed. Not all apps might have dedicated GPU requirements.

The change here is to allow the user to opt in / out of hardware acceleration via the `glutin::ContextBuilder` [with_hardware_acceleration](https://docs.rs/glutin/latest/glutin/struct.ContextBuilder.html#method.with_hardware_acceleration) method.

The setting has been added to `NativeOptions` ans is used in the `glutin` setup.

### DIFF
--- a/eframe/src/epi.rs
+++ b/eframe/src/epi.rs
@@ -211,6 +211,11 @@ pub struct NativeOptions {
     /// `egui` doesn't need the stencil buffer, so the default value is 0.
     pub stencil_buffer: u8,
 
+    /// Use hardware acceleration if available. On macOS, this will possibly
+    /// use a dedicated GPU which will lead to higher power consumption.
+    /// The default value is `Some(true)`
+    pub hardware_acceleration: Option<bool>,
+
     /// What rendering backend to use.
     pub renderer: Renderer,
 }
@@ -233,6 +238,7 @@ impl Default for NativeOptions {
             multisampling: 0,
             depth_buffer: 0,
             stencil_buffer: 0,
+            hardware_acceleration: Some(true),
             renderer: Renderer::default(),
         }
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui-w
 
 ## Unreleased
 ### Added ⭐
+* Added `NativeOptions::hardware_acceleration` to allow opting out of hardware acceleration for less power consumption ([#1681](https://github.com/emilk/egui/pull/1681)).
 * Added `*_released` & `*_clicked` methods for `PointerState` ([#1582](https://github.com/emilk/egui/pull/1582)).
 * Added `egui::hex_color!` to create `Color32`'s from hex strings under the `color-hex` feature ([#1596](https://github.com/emilk/egui/pull/1596)).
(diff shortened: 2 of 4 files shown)


eframe/src/epi.rs
```diff
@@ -211,6 +211,11 @@ pub struct NativeOptions {
     /// `egui` doesn't need the stencil buffer, so the default value is 0.
     pub stencil_buffer: u8,
 
+    /// Use hardware acceleration if available. On macOS, this will possibly
+    /// use a dedicated GPU which will lead to higher power consumption.
+    /// The default value is `Some(true)`
+    pub hardware_acceleration: Option<bool>,
```
emilk, line 217: Do you know what `None` means? It is not documented at https://docs.rs/glutin/latest/glutin/struct.ContextBuilder.html#method.with_hardware_acceleration

coderedart: on mac, glutin only sets the relevant attribute (mentioned below) if it is some, otherwise it just skips setting that attribute.

from https://developer.apple.com/documentation/appkit/1436213-opengl_pixel_format_attributes/nsopenglpfaaccelerated 

A Boolean attribute. If present, this attribute indicates that only hardware-accelerated renderers are considered. If not present, accelerated renderers are still preferred.

so, hw acceleration preferred by default (None). if `Some(true)`, only hw acceleration contexts are considered. but ofcourse, someone with mac needs to test it out just to be sure.

emilk: Interesting - so the current default (before and after this PR) means that we ONLY consider hardware-accelerated renderers, which could explain https://github.com/emilk/egui/issues/1358#issuecomment-1091893038

Perhaps the default should be `None` to allow software rendered backends too.

emilk (approved):

hacknus:
Can this also be implemented in the glow (new standard) setup? In my setup, it does not change anything..  
`eframe = {version = "0.19.0", features = ["persistence"]}
egui_extras = { version = "0.19.0", features = ["image"] }`  

`let options = eframe::NativeOptions {
        drag_and_drop_support: true,
        initial_window_size: Option::from(vec2(gui_settings.x, gui_settings.y)),
        hardware_acceleration : HardwareAcceleration::Off,
        ..Default::default()
    };`

emilk:
> Can this also be implemented in the glow (new standard) setup?

This PR was for `glow`

hacknus:
that's weird, did I configure something wrong? because for me the flag for the dedicated GPU is still set..

hacknus:
I noticed that this does not disable the GPU flag on macOS 12.4, so the dedicated GPU is still enabled when the application is launched. But it causes the application to panic on Windows 10 (native and virtual): `NoAvailablePixelFormat`

hacknus:
in the latest egui version (0.21.0) I get the following error under windows, when I set 
```
        drag_and_drop_support: true,
        initial_window_size: Option::from(vec2(gui_settings.x, gui_settings.y)),
        hardware_acceleration : HardwareAcceleration::Off,
        ..Default::default()
    };
```
I get the following error:
```
thread 'main' panicked at 'some OpenGL error occurred egui_glow requires opengl 2.0+. 
', C:\Users\ls21y493\.cargo\registry\src\github.com-1ecc6299db9ec823\eframe-0.21.0\src\native\run.rs:667:45
```
but when I leave out the `hardware_acceleration` flag, it runs fine.

Issue #1652, opened 2022-05-20
Feature Request: table columns added with variable widths, a one liner


ray-5050:
I'm using a table with about 9 columns and each has its own relative width so that seems to leave out using the columns method.  Currently I do the following
.column(Size::relative(self.col_widths[0]))
.column(Size::relative(self.col_widths[1]))
....
.column(Size::relative(self.col_widths[8]))
As you can see, this is a bit tedious.

Proposal

I added a method in the TableBuilder of the following

/// Allocate space for several columns at once with varying sizes.
    pub fn columns_with_sizes(mut self, size: Vec<Size>, count: usize) -> Self {
        if size.len() != count {
            panic!("The size vec must equal the count")
        }
        for i in 0..count {
            self.sizing.add(size[i]);
        }
        self
    }

This seems to work well and fills my needs, I just pass a vec containing the sizes and the Size enum I want
i.e. 
col_widths = vec![Size::relative(0.05), Size::relative(0.2), Size::relative(0.2), .....]
then
.columns_with_sizes(self.col_widths.clone(), 9)

Thanks for your consideration.

emilk:
I don't see the need for the `count: usize` parameter, but sure, add such a method if you like

PR #1682, opened 2022-05-25
#1652 Feature - Added columns_with_sizes method for TableBuilder

ray-5050:
#1652 Feature request
To reduce the lines of code when constructing table columns with varying sizes, a new fn was created by passing a &Vec<Size> parameter to new fn columns_with sizes.  The function is very similar to columns() by simply iterating over the vec. 

Closes <https://github.com/emilk/egui/issues/#1652>.

### DIFF
--- a/egui_extras/src/table.rs
+++ b/egui_extras/src/table.rs
@@ -126,8 +126,8 @@ impl<'a> TableBuilder<'a> {
     }
 
     /// Allocate space for several columns at once with varying sizes.
-    /// Example: let vec = &vec![Size::relative(0.5), Size::relative(0.3)];
-    pub fn columns_with_sizes(mut self, sizes: &Vec<Size>) -> Self {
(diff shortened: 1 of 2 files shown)


egui_extras/src/table.rs
```diff
@@ -125,6 +125,15 @@ impl<'a> TableBuilder<'a> {
         self
     }
 
+    /// Allocate space for several columns at once with varying sizes.
+    /// Vector example, let vec = &vec!`[Size::relative(0.5), Size::relative(0.3)]`
```
emilk, line 129:
```suggestion
    ///
    /// Example: `.columns_with_sizes(&[Size::relative(0.5), Size::relative(0.3)])`
```

emilk (changes requested):

emilk (approved):

PR #1670, opened 2022-05-23
Update wgpu to 0.13

expenses:
**This should only be merged when a new wgpu release is made**

I'm using the git version of [naga](https://github.com/gfx-rs/naga/) and currently egui-wgpu panics during shader processing with that because it expects a different syntax for decorations, so I needed to make these changes to get it to work.

PR #1684, opened 2022-05-26
Re-implement PaintCallbacks With Support for WGPU

zicklag:
This makes breaking changes to the PaintCallback system, but makes it
flexible enough to support both the WGPU and glow backends with custom
rendering.

Also adds a WGPU equivalent to the glow demo for custom painting.

Closes #1661.

----

This includes breaking changes to the way that `PaintCallbacks` are used. Instead of the paint callback downcasting a `context` argument to a specific backend context, the backend will downcast the `callback` to a backend-specific callback function.

This is required to allow greater divergence from the backend APIs. For example, the WGPU backend needs _two_ callbacks with different arguments for each, and it also has arguments that are not `'static`, meaning that having a `dyn Any` argument doesn't work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ members = [
 
     "examples/confirm_exit",
     "examples/custom_3d_glow",
-    "examples/custom_3d_three-d",
+    # "examples/custom_3d_three-d",
     "examples/custom_font",
     "examples/custom_font_style",
     "examples/custom_window_frame",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ members = [
 
     "examples/confirm_exit",
     "examples/custom_3d_glow",
-    # "examples/custom_3d_three-d",
+    "examples/custom_3d_three-d",
     "examples/custom_font",
     "examples/custom_font_style",
     "examples/custom_window_frame",
--- a/eframe/src/native/run.rs
+++ b/eframe/src/native/run.rs
@@ -62,6 +62,8 @@ pub fn run_glow(
         gl_window.window(),
         storage,
         Some(gl.clone()),
+        #[cfg(feature = "wgpu")]
+        None,
     );
 
     {
@@ -76,6 +78,8 @@ pub fn run_glow(
         integration_info: integration.frame.info(),
         storage: integration.frame.storage(),
         gl: Some(gl.clone()),
+        #[cfg(feature = "wgpu")]
+        render_state: None,
     });
 
     if app.warm_up_enabled() {
@@ -230,6 +234,8 @@ pub fn run_wgpu(
         painter
     };
 
+    let render_state = painter.get_render_state().expect("Uninitialized");
+
     let mut integration = epi_integration::EpiIntegration::new(
         &event_loop,
         painter.max_texture_side().unwrap_or(2048),
@@ -237,6 +243,7 @@ pub fn run_wgpu(
         storage,
         #[cfg(feature = "glow")]
         None,
+        Some(render_state.clone()),
     );
 
     {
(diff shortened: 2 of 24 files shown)


egui_demo_app/src/apps/custom3d_wgpu_shader.wgsl
```diff
@@ -0,0 +1,39 @@
...
+    out.position.x = out.position.x * cos(uniforms.angle);
+    out.color = v_colors[v_idx];
+
+    return out;
+}
+
+[[stage(fragment)]]
+fn fs_main(in: VertexOut) -> [[location(0)]] vec4<f32> {
+    return in.color;
+}
```
emilk, line 39: See also https://github.com/emilk/egui/pull/1670

egui_glow/src/painter.rs
```diff
@@ -68,6 +68,18 @@ pub struct Painter {
     destroyed: bool,
 }
 
+// TODO: Documentation
```
emilk, line 71: do it!

egui_glow/src/painter.rs
```diff
@@ -381,7 +393,11 @@ impl Painter {
                             screen_size_px,
                         };
 
-                        callback.call(&info, self);
+                        if let Some(callback) = callback.callback.downcast_ref::<CallbackFn>() {
+                            (callback.f)(info, self);
+                        } else {
+                            tracing::warn!("Warning: Unsupported render callback");
```
emilk, line 399:
```suggestion
                            tracing::warn!("Warning: Unsupported render callback. Expected egui_glow::CallbackFn");
```

Cargo.toml
```diff
@@ -14,7 +14,7 @@ members = [
 
     "examples/confirm_exit",
     "examples/custom_3d_glow",
-    "examples/custom_3d_three-d",
+    # "examples/custom_3d_three-d",
```
emilk, line 17:
```suggestion
    "examples/custom_3d_three-d",
```

emilk (changes requested):
Amazing job! Thank you ❤️

zicklag:
Great review points, I addressed them all, I believe, and CI is happy, too!

emilk (approved):

eframe/src/native/run.rs
```diff
@@ -230,13 +234,16 @@ pub fn run_wgpu(
         painter
     };
 
+    let render_state = painter.get_render_state().expect("Uninitialized");
```
Zoxc, line 237: This always panics on Android, so this breaks the support added in https://github.com/emilk/egui/pull/1634.

zicklag: Well that's not cool! I don't have an Android to test on unfortunately.

Still, I did try to build it for Android real quick. A quick look over and, it appears that there is multiple things breaking the ability to build `eframe` for android right now, that isn't related to this PR. It looks like #1634 might not have actually gotten `eframe` building for Android, just `egui_wgpu` and `egui_winit`.

Some things I noticed need to be fixed for `eframe` will build for Android:

- The inclusion of the `arboard` crate for clipboard support needs to be made optional, because it doesn't build for android. 
- Same fix for the `glutin` dependency
- It appears that the `winit::event::Event::Paused` event doesn't exist, but it's used in the android event loop code.

For what I can tell, it looks like the `eframe` crate hasn't ever actually worked for Android yet, but it might not be a big deal to make work. I'm not sure if I'll be able to get the time to try to get it working yet, though.

zicklag: Cool! We shouldn't actually need glutin anyway if we are using the WGPU backend. I'm not sure why text wouldn't be working.

emilk: Nice to hear some progress on this! Feel free to open a PR to add `cargo check --target aarch64-linux-android` to CI, and any other needed changes

Zoxc:
This also apparently broke DX12's CPU renderer as `egui_demo_app` just gives a black window after this PR.

zicklag:
That's another platform I can't test for unfortunately. :confused: 

I just have a Linux machine with Vulkan and OpenGL.

PR #1687, opened 2022-05-27
feat: add feature for load system fonts according user input

lj94093:
Solved the issure mentioned in <https://github.com/emilk/egui/discussions/1344>.

For now,this PR adds a `system_fonts` feature which depend the font-kits and skia-safe(`gl` feature) and it makes the binary size very big.So I add it as a optional feature in `epaint` ,`egui` and `eframe` for some people who need it.

Maybe I should develop a library to choose the font for characters.But it sounds like there are lots of work to do.If i finished it or find a proper library,i will make a PR again add as the default feature of `eframe`.So for now,Let's just make it as a experimental feature.

图片

### DIFF
--- a/eframe/Cargo.toml
+++ b/eframe/Cargo.toml
@@ -24,7 +24,7 @@ default = ["default_fonts", "glow"]
 
 # detect dark mode system preference
 dark-light = ["dep:dark-light"]
-
+system_fonts = ["egui/system_fonts"]
 # If set, egui will use `include_bytes!` to bundle some fonts.
 # If you plan on specifying your own fonts you may disable this feature.
 default_fonts = ["egui/default_fonts"]
--- a/egui/Cargo.toml
+++ b/egui/Cargo.toml
@@ -36,7 +36,7 @@ deadlock_detection = ["epaint/deadlock_detection"]
 # If set, egui will use `include_bytes!` to bundle some fonts.
 # If you plan on specifying your own fonts you may disable this feature.
 default_fonts = ["epaint/default_fonts"]
-
+system_fonts = ["epaint/system_fonts"]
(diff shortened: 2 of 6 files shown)


emilk:
Hi, and thanks for the PR!

Can you describe what this is supposed to do? The new code is very sparse on documentation.

I ran it on my mac and pasted some Japanese characters into a text box. The app froze for a second, and then the familiar `◻◻◻` replacement characters showed up. Maybe it doesn't work on Mac?

`epaint` and `egui` are designed to be completely disconnected from the platform they are running on, and this PR breaks this tradition. I think we should think about whether we can design this differently. For instance, this could be a plugin system where you can tell epaint "use this callback to look for fonts for missing characters". That way we could use the same plugin system to also implement this on web (some sunny day).

lj94093:
Sorry for didn't test it on mac os.I tested it on my Windows.

The correct way for use this feature is add `system_fonts` in project's Cargo.toml.

The reason for it didn't load correct fonts on mac is that `font-kits`'s family name didn't match with the name in skia.

Luckily,I find that skia-safe already load the font data from disk.So we don't need to load fonts by `font-kits` .So we can decrease the font loading time and remove the `font-kits` dependency.

I fix it in later 2 commits.

Reason for didn't implement it with a plugin system:
The skia-safe is too big for egui.If we implement a plugin system for load system fonts with this feature opened,the binary size of egui will be very large.It is not suitable for such a lightweight GUI framework.

If I find a better way to to choose the font for characters and make the binary size small enough.Then we can implement such a plugin system.

emilk:
Closing: system font loading landed in #8512 via the new `egui_system_fonts` crate and the `system_fonts` feature in eframe. Thanks for the PR!

Issue #1691, opened 2022-05-28
provide manual timeout to egui in reactive mode to force refresh even if there's no new input


coderedart:
**Is your feature request related to a problem? Please describe.**
egui's reactive mode is really great for applications, but the UI not updating AT ALL is a pain point for many newbies. at present, they are usually told to call `request_repaint` from another thread. not really user friendly imo. 

**Describe the solution you'd like**
just let users set a optional timeout, and when there's no input for the duration of the timeout, egui does the request repaint thing itself.

**Describe alternatives you've considered**
I don't see any alternatives.

**Additional context**
all big three windowing libraries:  winit, GLFW and SDL2, have support for wait_event timeout function, where the event loop will yield control back to us when it fails to get any events for the specified timeout. so, it should be pretty clean to implement.

emilk:
`eframe`:s web backend already repaints every second by default, and I agree this should be controllable and also implemented on native. Is it something you'd like to work on?

PR #1694, opened 2022-05-28
Add Context::request_repaint_after

coderedart:
Closes <https://github.com/emilk/egui/issues/1691>.

kinda my first PR. 

I am wondering whether i should store this idle timeout interval in `platform output` our `full output`. I did not implement this for wgpu yet, but should be pretty similar. 

by storing it in platform output, it affects the `take` function and the `Default` derive to be manually implemented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui-w
 
 ### Changed
 * `PaintCallback` shapes now require the whole callback to be put in an `Arc<dyn Any>` with the value being a backend-specific callback type. ([#1684](https://github.com/emilk/egui/pull/1684))
-* Replaced `needs_repaint` in `FullOutput` with `repaint_after`. used to force repaint after the set duration in reactive mode.([#1694](https://github.com/emilk/egui/pull/1694))
+* Replaced `needs_repaint` in `FullOutput` with `repaint_after`. Used to force repaint after the set duration in reactive mode.([#1694](https://github.com/emilk/egui/pull/1694)).
 
 ### Fixed 🐛
 * Fixed `ImageButton`'s changing background padding on hover ([#1595](https://github.com/emilk/egui/pull/1595)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui-w
 
 ### Changed
 * `PaintCallback` shapes now require the whole callback to be put in an `Arc<dyn Any>` with the value being a backend-specific callback type. ([#1684](https://github.com/emilk/egui/pull/1684))
+* Replaced `needs_repaint` in `FullOutput` with `repaint_after`. used to force repaint after the set duration in reactive mode.([#1694](https://github.com/emilk/egui/pull/1694))
 
 ### Fixed 🐛
 * Fixed `ImageButton`'s changing background padding on hover ([#1595](https://github.com/emilk/egui/pull/1595)).
--- a/eframe/src/native/run.rs
+++ b/eframe/src/native/run.rs
@@ -111,7 +111,7 @@ pub fn run_glow(
                 textures_delta,
                 shapes,
             } = integration.update(app.as_mut(), window);
-
+            let idle_duration = platform_output.idle_timeout_interval;
             integration.handle_platform_output(window, platform_output);
 
             let clipped_primitives = {
@@ -136,6 +136,10 @@ pub fn run_glow(
             } else if needs_repaint {
                 window.request_redraw();
                 winit::event_loop::ControlFlow::Poll
+            } else if let Some(idle_duration) = idle_duration {
+                let idle_duration = std::time::Instant::now()
+                    + idle_duration;
+                winit::event_loop::ControlFlow::WaitUntil(idle_duration)
             } else {
                 winit::event_loop::ControlFlow::Wait
             };
@@ -152,7 +156,6 @@ pub fn run_glow(
                 std::thread::sleep(std::time::Duration::from_millis(10));
             }
         };
-
         match event {
             // Platform-dependent event handlers to workaround a winit bug
             // See: https://github.com/rust-windowing/winit/issues/987
@@ -194,7 +197,12 @@ pub fn run_glow(
                 painter.destroy();
             }
             winit::event::Event::UserEvent(RequestRepaintEvent) => window.request_redraw(),
-            _ => (),
+            winit::event::Event::NewEvents(winit::event::StartCause::ResumeTimeReached {
+                ..
+            }) => {
+                window.request_redraw();
+            }
+            _ => {}
         }
     });
 }
--- a/eframe/src/native/run.rs
+++ b/eframe/src/native/run.rs
@@ -399,7 +399,7 @@ pub fn run_wgpu(
             }
             winit::event::Event::UserEvent(RequestRepaintEvent) => window.request_redraw(),
             winit::event::Event::NewEvents(winit::event::StartCause::ResumeTimeReached {
-            ..
+                ..
             }) => {
                 window.request_redraw();
             }
--- a/eframe/src/native/run.rs
+++ b/eframe/src/native/run.rs
@@ -398,6 +398,11 @@ pub fn run_wgpu(
                 painter.destroy();
             }
             winit::event::Event::UserEvent(RequestRepaintEvent) => window.request_redraw(),
+            winit::event::Event::NewEvents(winit::event::StartCause::ResumeTimeReached {
+            ..
+            }) => {
+                window.request_redraw();
+            }
             _ => (),
         }
     });
--- a/eframe/src/native/run.rs
+++ b/eframe/src/native/run.rs
@@ -107,11 +107,11 @@ pub fn run_glow(
 
             let egui::FullOutput {
                 platform_output,
-                needs_repaint,
+                repaint_after,
                 textures_delta,
                 shapes,
             } = integration.update(app.as_mut(), window);
-            let idle_duration = platform_output.idle_timeout_interval;
+
             integration.handle_platform_output(window, platform_output);
 
             let clipped_primitives = {
@@ -133,13 +133,18 @@ pub fn run_glow(
 
             *control_flow = if integration.should_quit() {
                 winit::event_loop::ControlFlow::Exit
-            } else if needs_repaint {
+            } else if repaint_after.is_zero() {
                 window.request_redraw();
                 winit::event_loop::ControlFlow::Poll
-            } else if let Some(idle_duration) = idle_duration {
-                let idle_duration = std::time::Instant::now()
-                    + idle_duration;
-                winit::event_loop::ControlFlow::WaitUntil(idle_duration)
+            } else if let Some(repaint_after_instant) =
+                std::time::Instant::now().checked_add(repaint_after)
+            {
+                // if repaint_after is something huge and can't be added to Instant,
+                // we will use `ControlFlow::Wait` instead.
+                // technically, this might lead to some weird corner cases where the user *WANTS*
+                // winit to use `WaitUntil(MAX_INSTANT)` explicitly. they can roll their own
+                // egui backend impl i guess.
+                winit::event_loop::ControlFlow::WaitUntil(repaint_after_instant)
             } else {
                 winit::event_loop::ControlFlow::Wait
             };
@@ -287,7 +292,7 @@ pub fn run_wgpu(
 
             let egui::FullOutput {
                 platform_output,
-                needs_repaint,
+                repaint_after,
                 textures_delta,
                 shapes,
             } = integration.update(app.as_mut(), window);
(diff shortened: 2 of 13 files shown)


emilk:
Thanks for starting to work on this!

I think it is worth thinking through what the most common use cases for this feature would be.

If this is a setting ("minimal refresh interval") I think it would make most sense to have it as a member of [`eframe::App`](https://docs.rs/eframe/latest/eframe/trait.App.html) (similar to `auto_save_interval`, for instance).

However, there could be cases were an egui widget want to schedule a deferred repaint. For instance, a clock widget may want an update every second to show move the second hand. In this case it makes more sense to have it as a member of `egui::Context` (like is currently the case of this PR). In that case, we also want to make sure that if several widgets calls the function then we do a repaint after the _shortest_ requested time. So if one widget needs updating every two seconds, and one every 500ms, the latter time is used no matter what order the widgets are called. Let's call this something like `request_repaint_in` with a `Duration` member. In this case we probably also want `request_repaint` and `request_repaint_in(Duration::ZERO)` to mean the exact same thing.

Thoughts?

coderedart:
I thought it would be best to implement this in eframe too, but then i realized other non-eframe implementations might want this feature too and decided to use `Context` instead. 

The primary purpose of this issue was to just have some sort of timeout interval for the *next* repaint callback. 

I had two use cases in mind:
1. a simple stopwatch or timer app. if it is only showing seconds, then they only need to refresh every second. but `request_repaint` doesn't allow them to delay the repaint signal. and every time the app repaints, they can set the interval for the next frame by checking the current time and calculating how long before the seconds text displayed in gui needs to change. then, set that as the delay for the next frame. here, you want to make sure that the gui refreshes at this particular instant in future, regardless of whether it had input or not. if there's input, causing an earlier refresh, app will recalculate the timeout and set it again this frame. 
2.  a normal application like weather app or something, where you only want the gui to not be completely stale / dead due to idling in the background. here, you don't care about *how long till a particular moment in time*, but rather just a simple, maximum timeout like 5 seconds or something from the frame where it received the last input event. you can just provide the same duration every frame with no calculation.

I did not think about widgets explicitly wanting to set a timeout. but, I don't think that changes anything here. 

my plan was to only just store the shortest duration too. as long as the repaint happens at or before the Instant desired by widget / user, they can recalculate their next duration according to their repaint needs just like in stop watch usecase. 

I guess, that settles most questions. 
`request_repaint_in` will take a duration, and will keep track of the smallest duration inside context. will provide that duration to backend in `FullOutput` along with the `needs_repaint` bool. winit/glfw/sdl will need to make sure to not idle longer than that duration waiting for events. 

the name `request_repaint_in` is slightly misleading, but it should be fine. duration starts at the next frame. lets say for example that its a very inefficient app and takes 500 milliseconds per frame at 2 fps. the widget / user might want a repaint in next 500 milliseconds. now, app takes 1000 ms per frame (1 fps) because the backend event timeout takes 500 milli seconds AFTER the vsync swap buffer. 

so, its not that we are requesting repaint within X duration. we are rather timing out during app idle time without any input events. 
just need to mention this in the docs, so that they understand what is happening in the background.

emilk:
Good points! Perhaps `request_repaint_after` is a better name to communicate that the time starts ticking _after_ the current frame finishes?

You can also use `Duration::MAX` to avoid using an `Option` (will make `.min(…)` easier to call)

coderedart:
should i just replace `needs_repaint` bool with the `repaint_after` duration? then, checking for `needs_repaint` is simply `repaint_after == Duration::ZERO`

it will be a breaking change

egui/src/context.rs
```diff
@@ -574,6 +575,39 @@ impl Context {
         }
     }
 
+    /// request repaint after the specified duration elapses in the case of no new input
```
coderedart, line 578: I guess i like writing documentation xD

coderedart:
I have no idea how to deal with web, so i left it alone except changing the type signature in the logic function. 
todo:
- [ ]  add to changelog about breaking change

eframe/src/web/backend.rs
```diff
@@ -248,7 +248,7 @@ impl AppRunner {
     /// Returns `true` if egui requests a repaint.
```
emilk, line 248:
```suggestion
    /// Returns how long to wait until the next repaint.
```

egui/src/context.rs
```diff
@@ -813,19 +839,18 @@ impl Context {
...
-            false
-        };
-        self.write().requested_repaint_last_frame = needs_repaint;
-
+        let repaint_after = self.read().repaint_after;
+        self.write().requested_repaint_last_frame = repaint_after.is_zero();
+        // make sure we reset the repaint_after duration.
+        // otherwise, if repaint_after is low, then any widget setting repaint_after next frame,
+        // will fail to overwrite the previous lower value. and thus, repaints will never
+        // go back to higher values.
```
emilk, line 850: Is the `repaint_requests` logic gone, or just moved?

The current behavior is that calling `request_repaint` actually does _two_ repaints. That's because often you call it on an event, and then there are some thing in egui that has one frame of delay (e.g. window sizes are calculated one frame and then used to position the window the next frame). I'd like to keep that behavior! But only `request_repaint` (and maybe `request_repaint_after(Duration::ZERO)`) should do a double-repaint. `request_repaint_after(0.1s)` has no real reason to do so imho.

coderedart: hmm, its there in my local copy. let me check if i messed up with the commit smh

coderedart: its right there on line 842-244. I just removed the assignment of the `needs_repaint` bool. 

but i understand the issue now. I will instead set `repaint_after` to `Duration::ZERO` when the `repaint_requests` is greater than zero.

coderedart:
Unfortunately libspeechd is failing to link on my arch. So, I could not check for failures offline with --all-features. Had to commit and wait for ci workflow to run and show me errors to fix.

I hope this works now.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ NOTE: [`epaint`](epaint/CHANGELOG.md), [`eframe`](eframe/CHANGELOG.md), [`egui-w
 
 ### Changed
 * `PaintCallback` shapes now require the whole callback to be put in an `Arc<dyn Any>` with the value being a backend-specific callback type. ([#1684](https://github.com/emilk/egui/pull/1684))
+* Replaced `needs_repaint` in `FullOutput` with `repaint_after`. used to force repaint after the set duration in reactive mode.([#1694](https://github.com/emilk/egui/pull/1694))
```
emilk, line 20:
```suggestion
* Replaced `needs_repaint` in `FullOutput` with `repaint_after`. Used to force repaint after the set duration in reactive mode ([#1694](https://github.com/emilk/egui/pull/1694)).
```

eframe/src/native/run.rs
```diff
@@ -194,7 +202,12 @@ pub fn run_glow(
                 painter.destroy();
             }
             winit::event::Event::UserEvent(RequestRepaintEvent) => window.request_redraw(),
-            _ => (),
+            winit::event::Event::NewEvents(winit::event::StartCause::ResumeTimeReached {
+                ..
+            }) => {
+                window.request_redraw();
+            }
+            _ => {}
```
emilk, line 210: This needs to be put in `run_wgpu` also

eframe/src/web/events.rs
```diff
@@ -9,9 +9,9 @@ pub fn paint_and_schedule(
         let mut runner_lock = runner_ref.lock();
         if runner_lock.needs_repaint.fetch_and_clear() {
             runner_lock.clear_color_buffer();
-            let (needs_repaint, clipped_primitives) = runner_lock.logic()?;
+            let (repaint_after, clipped_primitives) = runner_lock.logic()?;
             runner_lock.paint(&clipped_primitives)?;
-            if needs_repaint {
+            if repaint_after.is_zero() {
                 runner_lock.needs_repaint.set_true();
```
emilk, line 15:
```suggestion
            }
            // TODO: schedule a repaint after `repaint_after`
```

egui-winit/src/lib.rs
```diff
@@ -475,7 +475,6 @@ impl State {
             mutable_text_under_cursor: _, // only used in eframe web
             text_cursor_pos,
         } = platform_output;
-
```
emilk, line 478: revert

egui/src/context.rs
```diff
@@ -28,7 +28,6 @@ impl Default for WrappedTextureManager {
 }
 
 // ----------------------------------------------------------------------------
-
```
emilk, line 31: revert

egui/src/context.rs
```diff
@@ -574,6 +575,39 @@ impl Context {
         }
     }
 
+    /// request repaint after the specified duration elapses in the case of no new input
+    /// events being received.
+    ///
+    /// function can be multiple times, but only the *smallest* duration will be considered.
```
emilk, line 581:
```suggestion
    /// The function can be called multiple times, but only the *smallest* duration will be considered.
```

egui/src/context.rs
```diff
@@ -574,6 +575,39 @@ impl Context {
...
+    /// NOTE: only works if called before `Context::end_frame()`. to force egui to update,
+    /// use `Context::request_repaint()` instead.
+    ///
+    /// Quirk:
+    ///     duration begins at the next frame. lets say for example that its a very inefficient app
+    ///     and takes 500 milliseconds per frame at 2 fps. the widget / user might want a repaint in
+    ///     next 500 milliseconds. now, app takes 1000 ms per frame (1 fps) because the backend event
+    ///     timeout takes 500 milli seconds AFTER the vsync swap buffer.
+    ///     so, its not that we are requesting repaint within X duration. we are rather timing out
+    ///     during app idle time where we are not receiving any new input events.
```
emilk, line 604:
```suggestion
    /// ### Quirk:
    /// duration begins at the next frame. lets say for example that its a very inefficient app
    /// and takes 500 milliseconds per frame at 2 fps. the widget / user might want a repaint in
    /// next 500 milliseconds. now, app takes 1000 ms per frame (1 fps) because the backend event
    /// timeout takes 500 milli seconds AFTER the vsync swap buffer.
    /// so, its not that we are requesting repaint within X duration. we are rather timing out
    /// during app idle time where we are not receiving any new input events.
```

Indentation becomes a code block in markdown. See for yourself with `cargo doc -p egui --no-deps --open`

egui_demo_app/src/backend_panel.rs
```diff
@@ -220,6 +243,16 @@ impl BackendPanel {
...
             ui.label("Only running UI code when there are animations or input.");
+            ui.label("but if there's no input for the repaint_after duration, we force an update");
+            ui.label("repaint_after (in seconds)");
+            let mut milli_seconds = self.repaint_after_timeout.as_secs_f32();
+            if egui::DragValue::new(&mut milli_seconds)
+                .clamp_range(0.1..=10.0)
+                .ui(ui)
+                .changed()
+            {
+                self.repaint_after_timeout = std::time::Duration::from_secs_f32(milli_seconds);
```
emilk, line 254:
```suggestion
            let mut seconds = self.repaint_after_timeout.as_secs_f32();
            if egui::DragValue::new(&mut seconds)
                .clamp_range(0.1..=10.0)
                .ui(ui)
                .changed()
            {
                self.repaint_after_timeout = std::time::Duration::from_secs_f32(seconds);
```

egui_glium/examples/native_texture.rs
```diff
@@ -44,9 +44,13 @@ fn main() {
...
             *control_flow = if quit {
                 glutin::event_loop::ControlFlow::Exit
-            } else if needs_repaint {
+            } else if repaint_after.is_zero() {
                 display.gl_window().window().request_redraw();
                 glutin::event_loop::ControlFlow::Poll
+            } else if let Some(repaint_after_instant) =
+                std::time::Instant::now().checked_add(repaint_after)
+            {
+                glutin::event_loop::ControlFlow::WaitUntil(repaint_after_instant)
```
emilk, line 53: You also need to listen for `winit::event::StartCause::ResumeTimeReached`, right?

egui_glium/examples/pure_glium.rs
```diff
@@ -25,9 +25,13 @@ fn main() {
...
             *control_flow = if quit {
                 glutin::event_loop::ControlFlow::Exit
-            } else if needs_repaint {
+            } else if repaint_after.is_zero() {
                 display.gl_window().window().request_redraw();
                 glutin::event_loop::ControlFlow::Poll
+            } else if let Some(repaint_after_instant) =
+                std::time::Instant::now().checked_add(repaint_after)
+            {
+                glutin::event_loop::ControlFlow::WaitUntil(repaint_after_instant)
```
emilk, line 34: Again with the `winit::event::StartCause::ResumeTimeReached`

egui_glow/examples/pure_glow.rs
```diff
@@ -28,9 +28,13 @@ fn main() {
...
             *control_flow = if quit {
                 glutin::event_loop::ControlFlow::Exit
-            } else if needs_repaint {
+            } else if repaint_after.is_zero() {
                 gl_window.window().request_redraw();
                 glutin::event_loop::ControlFlow::Poll
+            } else if let Some(repaint_after_instant) =
+                std::time::Instant::now().checked_add(repaint_after)
+            {
+                glutin::event_loop::ControlFlow::WaitUntil(repaint_after_instant)
```
emilk, line 37: And again with `winit::event::StartCause::ResumeTimeReached`

egui_glow/src/winit.rs
```diff
@@ -48,11 +48,11 @@ impl EguiGlow {
         &mut self,
         window: &winit::window::Window,
         run_ui: impl FnMut(&egui::Context),
-    ) -> bool {
+    ) -> std::time::Duration {
```
emilk, line 51: the docstring of this functions needs updatin'

egui/src/context.rs
emilk: I guess you don't like capitalization :P

emilk (changes requested):
Excellent job - sorry for the slow review, but almost there now!

eframe/src/native/run.rs
```diff
@@ -194,7 +202,12 @@ pub fn run_glow(
                 painter.destroy();
             }
             winit::event::Event::UserEvent(RequestRepaintEvent) => window.request_redraw(),
-            _ => (),
+            winit::event::Event::NewEvents(winit::event::StartCause::ResumeTimeReached {
```
emilk, line 205: You should be able to combine these with `winit::event::Event::UserEvent(RequestRepaintEvent | winit::event::Event::NewEvents…`

coderedart:
apologies for the delay. I had sudden irl stuff to take care of until yesterday and only now got to finally open my PC to work on this.

coderedart:
think all CI should pass now. its been so long since i cared about grammar, that i forgot you capitalize first letter in a sentence :)

emilk:
Great job!

Issue #381, opened 2021-05-12
epi::http not showing up in the docs


emilk:
I just noticed that `epi::http` is not showing up in the docs at <https://docs.rs/epi/0.12.0/epi/>.

How do we configure cargo doc to generate docs with all feature enabled?

memoryruins:
docs.rs will read `[package.metadata.docs.rs]` metadata in `Cargo.toml` https://docs.rs/about/metadata

emilk:
Thanks, I didn't know about that!

PR #1700, opened 2022-05-31
Ensure docs for `winit` feature are generated.

follower:
### Purpose

Without this patch the documentation for [`egui-wgpu::winit`](https://github.com/emilk/egui/blob/e6cfa5028e2dd222ead311cf790d192ff401d0b9/egui-wgpu/src/winit.rs) is not generated which makes Ferris sad.

The configuration was copied from https://github.com/emilk/egui/blob/f27f67b76be7ce614680487e85fb2728d5c53ea8/egui-winit/Cargo.toml#L16-L19 but is currently untested.

### Current state

Currently the only mention of the [`winit` feature](https://github.com/emilk/egui/blob/e6cfa5028e2dd222ead311cf790d192ff401d0b9/egui-wgpu/Cargo.toml#L28-L30) on docs.rs is [here](https://docs.rs/crate/egui-wgpu/0.18.0/features) which (unhelpfully :D ) states "This feature flag does not enable additional features".[0]

### Other context

My particular interest was in `paint_and_update_textures()` which, perhaps ironically, turns out not to be documented (but even documentation of its mere existence would be helpful):
https://github.com/emilk/egui/blob/e6cfa5028e2dd222ead311cf790d192ff401d0b9/egui-wgpu/src/winit.rs#L215-L222

### Other related code

https://github.com/emilk/egui/blob/e6cfa5028e2dd222ead311cf790d192ff401d0b9/egui-wgpu/src/lib.rs#L11-L15

### References

 * https://docs.rs/about/metadata
 * https://github.com/emilk/egui/commit/442b9539643e4ce037da768bc2cd26fe095e0df1
 * https://github.com/emilk/egui/issues/381#issuecomment-839707073

[0] Which according to [this](https://stackoverflow.com/questions/71636468/rust-claims-dashmap-into-par-iter-trait-bounds-not-satisfied#71637846) is intended to be read as "[this] feature does not activate other feature _flags_."

### DIFF
--- a/egui-wgpu/Cargo.toml
+++ b/egui-wgpu/Cargo.toml
@@ -23,6 +23,8 @@ include = [
   "Cargo.toml",
 ]
 
+[package.metadata.docs.rs]
+all-features = true
 
 
 [features]


emilk (approved):
Good catch! We have these lines in all the other `Cargo.toml`:s, but this one was forgotten.

There is no standard way to document features that I know of. I usually look in the `Cargo.toml` of a crate I'm interested in, and that's where the `winit` feature flag has at least on line of comment:

```toml
[features]
# Make it easy to create bindings for winit.
winit = ["dep:pollster", "dep:winit"]
```

Mingun:
Tip: you can use [document-features](https://docs.rs/document-features/0.2.1/document_features/) crate.

emilk:
That's a great tip @Mingun - I'll get right on that!

PR #1703, opened 2022-06-01
Red Green Color Blindness Debug Build Text

thomas992:
caption

### DIFF
--- a/egui/src/lib.rs
+++ b/egui/src/lib.rs
@@ -371,7 +371,7 @@ pub fn warn_if_debug_build(ui: &mut crate::Ui) {
         ui.label(
             RichText::new("‼ Debug build ‼")
                 .small()
-                .color(crate::Color32::RED),
+                .color(crate::Color32::from_rgb(126, 130, 255)),
         )
         .on_hover_text("egui was compiled with debug assertions enabled.");
     }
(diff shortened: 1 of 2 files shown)


thomas992:
I did some math and figured out my color was off
caption
255*3=765 total color value to white/2=382.5 the medium between white and black
-255(from blue)= 127.5 remaining color / 2 = 67 for red and green

thomas992:
caption

used the color dropper to adjust my math

emilk:
The color red signifies warning, which is what I was going for. Sure, we can find a nicer tint of red that works better in dark and light mode, but I still think it can and should be red.

thomas992:
One in twelve men are colorblind, more so in countries like Japan
[statistic](https://www.clintoneye.com/color-blindness.html#:~:text=There%20are%20an%20estimated%20300,are%20color%20blind%20(0.5%25).)
I think of it as a design philosophy if there is only one accent color it should be blue.

emilk:
Being red-green color blind (Deutan) does not mean one cannot see red or green. I should know, I'm one. And using only one color where we could use several does not improve anything for anyone.

thomas992:
the 580 line on the right side of the spectral formation might be better for text

thomas992:
I just made it a hyperlink to Github. I did not really want to add a hyperlink but in retrospect it makes sense for people to pull up the GitHub from debug build.

thomas992:
caption
I changed the hyperlink color on light theme as it was too bright to be clearly visible

emilk:
If you want to be able to change the default warning color, perhaps add `warning_color: Color32` to [`Visuals`](https://docs.rs/egui/latest/egui/style/struct.Visuals.html) ?

PR #1704, opened 2022-06-01
WGPU PaintCallback Fixes

zicklag:
Makes a couple of fixes to the WGPU paint callback implementation:

- Makes the egui textures public. Similar to the glow backend, this lets you access textures allocated by EGUI for custom rendering.
- Fixes a subtle but crucial bug: depending on the relative order of egui meshes and paint callbacks being rendered, some of the last rendered items would just be skipped in apparently random ways.

### DIFF
--- a/egui-wgpu/src/renderer.rs
+++ b/egui-wgpu/src/renderer.rs
@@ -139,7 +139,7 @@ impl RenderPass {
     /// Creates a new render pass to render a egui UI.
     ///
     /// If the format passed is not a *Srgb format, the shader will automatically convert to `sRGB` colors in the shader.
-    pub fn new(
+    pub(crate) fn new(
         device: &wgpu::Device,
         output_format: wgpu::TextureFormat,
         msaa_samples: u32,
@@ -292,7 +292,7 @@ impl RenderPass {
     }
 
     /// Executes the egui render pass.
-    pub fn execute(
+    pub(crate) fn execute(
         &self,
         encoder: &mut wgpu::CommandEncoder,
         color_attachment: &wgpu::TextureView,
@@ -326,7 +326,7 @@ impl RenderPass {
     }
 
     /// Executes the egui render pass onto an existing wgpu renderpass.
-    pub fn execute_with_renderpass<'rpass>(
+    pub(crate) fn execute_with_renderpass<'rpass>(
         &'rpass self,
         rpass: &mut wgpu::RenderPass<'rpass>,
         paint_jobs: &[egui::epaint::ClippedPrimitive],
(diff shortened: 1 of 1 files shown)


egui-wgpu/src/renderer.rs
```diff
@@ -128,7 +128,7 @@ pub struct RenderPass {
     /// Map of egui texture IDs to textures and their associated bindgroups (texture view +
     /// sampler). The texture may be None if the TextureId is just a handle to a user-provided
     /// sampler.
-    textures: HashMap<egui::TextureId, (Option<wgpu::Texture>, wgpu::BindGroup)>,
+    pub textures: HashMap<egui::TextureId, (Option<wgpu::Texture>, wgpu::BindGroup)>,
```
emilk, line 131: should everyone be able to just mutate this as is? Would it be better to use accessor functions instead?

egui-wgpu/src/renderer.rs
```diff
@@ -384,6 +378,9 @@ impl RenderPass {
             match primitive {
                 Primitive::Mesh(mesh) => {
                     if let Some((_texture, bind_group)) = self.textures.get(&mesh.texture_id) {
+                        let index_buffer = index_buffers.next().unwrap();
+                        let vertex_buffer = vertex_buffers.next().unwrap();
```
emilk, line 382: In what circumstance would this `unwrap` crash?

zicklag: It can only crash if we had a bug in `update_buffers()` ( [here](https://github.com/emilk/egui/pull/1704/files/6160a9507f3c4afe18a8b49cc4cb80a1d00bd3aa#diff-9a3a7bbf3e24ed9aa455942f77365cfe970b486b43f9f1d97d00ca56ab7cfba2R669) ). If we failed to push a vertex and index buffer to the list for every `Primitive::Mesh`, then the unwrap would panic.

It can't fail because of any action on the part of the user of egui.

Before this PR, if there was a bug in `update_buffers()` it would have just silently not rendered certain meshes, because `zip()`-ing iterators of different sizes will just stop at the shortest iterator. Now it is actually better, because if we mess that up, it will crash, instead of silently and randomly fail to render certain meshes. That's actually what made this bug difficult to find.

egui-wgpu/src/renderer.rs
zicklag: OK, I made `textures` private again, and I also made all of the functions that users shouldn't need to call private to the crate.

emilk: great!

zicklag:
Ah, it doesn't happen with `cargo check --features wgpu`, but I haven't figured out why it's doing that with just plain `cargo check` yet.

zicklag:
Oh, I figured it out, it's that all those `pub(crate)` functions are only used by the `winit` feature. So we'd have to add a `cfg(feature = "winit")` to all those functions to make them go away. You see any problem with that?

emilk (approved):
awesome, thanks!

PR #1709, opened 2022-06-02
Add mention of screen-13-egui crate

attackgoat:
We now have another integration courtesy of @DoeringChristian to add to the already impressive list


emilk (approved):
Great!

PR #1722, opened 2022-06-07
Allow cursors to be linked for plots

Zoxc:
This adds `set_link_cursor` to `LinkedAxisGroup` which allow the cursor to show up in linked plots.

This only handles the default `PlotItem` case. I think changing `PlotItem::on_hover` to produce a list of cursors would make sense.

The `PartialEq` implementation for `LinkedAxisGroup` also seems a bit questionable given it carries identity.

### DIFF
--- a/crates/egui/src/widgets/plot/items/bar.rs
+++ b/crates/egui/src/widgets/plot/items/bar.rs
@@ -2,7 +2,7 @@ use crate::emath::NumExt;
 use crate::epaint::{Color32, RectShape, Rounding, Shape, Stroke};
 
 use super::{add_rulers_and_text, highlighted_color, Orientation, PlotConfig, RectElement};
-use crate::plot::{BarChart, PlotPoint, ScreenTransform};
+use crate::plot::{BarChart, Cursor, PlotPoint, ScreenTransform};
 
 /// One bar in a [`BarChart`]. Potentially floating, allowing stacked bar charts.
 /// Width can be changed to allow variable-width histograms.
@@ -142,13 +142,14 @@ impl Bar {
         parent: &BarChart,
         plot: &PlotConfig<'_>,
         shapes: &mut Vec<Shape>,
+        cursors: &mut Vec<Cursor>,
     ) {
         let text: Option<String> = parent
             .element_formatter
             .as_ref()
             .map(|fmt| fmt(self, parent));
 
-        add_rulers_and_text(self, plot, text, shapes);
+        add_rulers_and_text(self, plot, text, shapes, cursors);
     }
 }
 
--- a/crates/egui/src/widgets/plot/items/box_elem.rs
+++ b/crates/egui/src/widgets/plot/items/box_elem.rs
@@ -2,7 +2,7 @@ use crate::emath::NumExt;
 use crate::epaint::{Color32, RectShape, Rounding, Shape, Stroke};
 
 use super::{add_rulers_and_text, highlighted_color, Orientation, PlotConfig, RectElement};
-use crate::plot::{BoxPlot, PlotPoint, ScreenTransform};
+use crate::plot::{BoxPlot, Cursor, PlotPoint, ScreenTransform};
 
 /// Contains the values of a single box in a box plot.
 #[derive(Clone, Debug, PartialEq)]
@@ -221,13 +221,14 @@ impl BoxElem {
         parent: &BoxPlot,
         plot: &PlotConfig<'_>,
         shapes: &mut Vec<Shape>,
+        cursors: &mut Vec<Cursor>,
     ) {
         let text: Option<String> = parent
             .element_formatter
             .as_ref()
             .map(|fmt| fmt(self, parent));
 
-        add_rulers_and_text(self, plot, text, shapes);
+        add_rulers_and_text(self, plot, text, shapes, cursors);
     }
 }
 
--- a/crates/egui/src/widgets/plot/items/mod.rs
+++ b/crates/egui/src/widgets/plot/items/mod.rs
@@ -1,6 +1,6 @@
 //! Contains items that can be added to a plot.
 
-use std::ops::RangeInclusive;
+use std::{cell::RefCell, ops::RangeInclusive};
 
 use epaint::util::FloatOrd;
 use epaint::Mesh;
@@ -28,6 +28,7 @@ pub(super) struct PlotConfig<'a> {
     pub transform: &'a ScreenTransform,
     pub show_x: bool,
     pub show_y: bool,
+    pub cursors: &'a RefCell<Vec<(Orientation, f64)>>,
 }
 
 /// Trait shared by things that can be drawn in the plot.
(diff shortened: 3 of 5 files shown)


emilk:
It would be nice of an example of this so we can test it!

enomado:
Wanted this feature

Zoxc:
I did some refactoring and added the ability to link cursors to the linked axis demo.

egui/src/widgets/plot/mod.rs
```diff
@@ -115,22 +121,48 @@ impl PlotMemory {
 
 // ----------------------------------------------------------------------------
 
+#[derive(Copy, Clone, PartialEq)]
+struct Cursor {
+    orientation: Orientation,
+    point: f64,
```
emilk, line 127: I think it is worth documenting the struct as a whole, but also the unit of `point`

emilk: Reading how this is used, I think it would be more clear as:

```rust
enum Cursor {
    Horizontal { y: f64 },
    Vertical { x: f64 },
}
```

egui/src/widgets/plot/mod.rs
```diff
@@ -115,22 +121,48 @@ impl PlotMemory {
...
 /// Defines how multiple plots share the same range for one or both of their axes. Can be added while building
 /// a plot with [`Plot::link_axis`]. Contains an internal state, meaning that this object should be stored by
 /// the user between frames.
 #[derive(Clone, PartialEq)]
 pub struct LinkedAxisGroup {
     pub(crate) link_x: bool,
     pub(crate) link_y: bool,
+    pub(crate) link_cursor: bool,
     pub(crate) bounds: Rc<Cell<Option<PlotBounds>>>,
+    cursors: Rc<RefCell<Vec<(Id, Vec<Cursor>)>>>,
```
emilk, line 155: I agree with clippy

egui/src/widgets/plot/mod.rs
```diff
@@ -115,22 +121,48 @@ impl PlotMemory {
...
 }
```
emilk, line 156: Reusing the same `LinkedAxisGroup` for the cursor feature has a few downsides. In particular, it is (afaict) not possible to JUST link cursors without also linking scrolling. It also makes it more difficult to refactor one feature without disturbing the other. Is there advantage in coupling them that I'm missing?

---

As an alternative, I would propose a separate `CursorGroup`:

```rust
struct CursorGroup {
    /// with a nice dosctring here
    cursors: Vec<(Id, Vec<Cursor>)> 
}
```

```rust
impl Plot {
    pub fn cursor_group(mut self, group: Rc<Cell<CursorGroup>>) -> Self {
        self.cursor_group = Some(group);
        self
    }
}
```

egui/src/widgets/plot/mod.rs
emilk: Also needs documentation explaining it.

egui/src/widgets/plot/mod.rs
```diff
@@ -1254,7 +1352,7 @@ impl PreparedPlot {
         }
     }
 
-    fn hover(&self, ui: &Ui, pointer: Pos2, shapes: &mut Vec<Shape>) {
+    fn hover(&self, ui: &Ui, pointer: Pos2, shapes: &mut Vec<Shape>) -> Vec<Cursor> {
```
emilk, line 1355: it's a bit unintuitive how hovering can return more than one cursor. I guess it is the horizontal + vertical lines that are considered separate cursors? Could perhaps the code be made simpler by using `struct Cursor { point: PlotPoint, horizontal: bool, vertical: bool }` ?

egui/src/widgets/plot/items/mod.rs
```diff
@@ -1575,44 +1594,41 @@ fn add_rulers_and_text(
...
+                Orientation::Horizontal => Cursor::horizontal(argument.y),
+                Orientation::Vertical => Cursor::vertical(argument.x),
             };
-            shapes.push(line);
+            cursors.push(cursor);
         };
 
         for pos in elem.arguments_with_ruler() {
-            push_argument_ruler(pos, shapes);
+            push_argument_ruler(pos, cursors);
```
emilk, line 1616: Not new in this PR, but I really don't see why `push_argument_ruler` exists and is not just inlined in the loop

emilk (changes requested):
Very nice feature, but I think there is an opportunity to simplify the implementation while also making it more powerful!

egui/src/widgets/plot/mod.rs
Zoxc: Things like bar charts have too many lines for that to work.

emilk: Then I don't understand what a `Cursor` is. I thought it was referring to the mouse cursor

Zoxc: It's the lines for objects near the mouse cursor or lines for the cursor itself. See the Plot demo -> Charts -> Box Plot and hover over things.

egui/src/widgets/plot/mod.rs
```diff
@@ -115,6 +121,70 @@ impl PlotMemory {
...
 // ----------------------------------------------------------------------------
 
+/// Indicates a vertical or horizontal cursor line in plot coordinates.
+#[derive(Copy, Clone, PartialEq)]
+enum Cursor {
+    Horizontal { y: f64 },
+    Vertical { x: f64 },
+}
+
+/// Contains the cursors drawn for a specific frame of a plot.
```
emilk, line 131: `frame` as in "frames per second"? Or just is it just "for a specific plot widget"? (each widget has a unique `id`)

egui/src/widgets/plot/mod.rs
```diff
@@ -661,6 +741,29 @@ impl Plot {
...
+
+            // Look for our previous frame
+            let index = frames
+                .iter()
+                .enumerate()
+                .find(|(_, frame)| frame.id == plot_id)
+                .map(|(i, _)| i);
+
+            // Remove our previous frame and all older frames as these are no longer displayed.
+            index.map(|index| frames.drain(0..=index));
```
emilk, line 756: I don't really follow this. What is `frames` sorted on? The docstring doesn't mention it being ordered by anything.

emilk (changes requested):
Looks good, but the use of the word "frame" through me off. In egui it either means "one ui layout time slice" (as in "frames per second"), or it means `egui::Frame` (for drawing a border around widgets).

If _think_ in this PR it just means "plot widget", so it would be better to call it `CursorsForPlot` or similar, with `plot_id: Id` for extra clairy.

egui/src/widgets/plot/mod.rs
Zoxc: It's what was drawn for a specific plot widget in a frame with the "frames per second" meaning.

egui/src/widgets/plot/mod.rs
Zoxc: `frames` is the oldest drawn widget to the newest. Each time a plot widget is drawn, it is appended to `frames`. Previous entries are removed here to avoid unbounded growth.

emilk: Oh, now I see. Thanks for explaining! Could you please add those explanations to the code and then I will merge this!

Zoxc: I added some more explanations.

emilk (approved):
Thanks!

PR #1730, opened 2022-06-11
Refactor common code in `egui-wgpu` vertex shader entry points

mitchmindtree:
Creates `unpack_color` and `position_from_screen` functions to share common logic between the alternative vertex shader entry points. Behaviour remains the same.

I noticed these when checking the difference between `vs_main` and `vs_conv_main` in order to solve some breakage in colour output after updating from 0.15 to 0.18 (I'll open a separate issue/PR about this).


emilk (approved):
Look good! But before I merge: have you tested the code?

mitchmindtree:
Yes, both entry points successfully tested with a downstream project.

PR #1732, opened 2022-06-13
Remove lock from `Context`

### LINKED ISSUE #1399: Consider making egui::Context !Send+!Sync
Prompted by https://github.com/emilk/egui/pull/1398

Currently `Context` and `Shape` are both `Send+Sync`. This means `Shape::Callback` can only capture things that are `Send+Sync`, even though almost all users will be running their UI code on the the paint thread.

We could consider making `Shape: !Send + !Sync`, but that would mean `egui::Context` could not be `Send+Sync` either (because the egui context stores shapes). This could actually be fine. `egui::Context` should only be used from a background thread for calling `request_repaint` and allocating textures. These could be made separate parts of the egui Context, so that one would do:

``` rust
let repaint_signal = egui_ctx.repaint_signal();
let tex_mngr = egui_ctx.tex_mngr();
std::thread::spawn(move || {
    // We can use repaint_signal and tex_mngr here,
    // but NOT `egui_ctx`.
}):
```

Making `Context: !Sync` would also solve deadlocks discussed in #1379 and #1380.

The downside of this is that it would stop users from running their UI code in a separate thread from their main window/paint thread.

Related:
* https://github.com/emilk/egui/issues/1379
* https://github.com/emilk/egui/pull/1380
* https://github.com/emilk/egui/pull/1389

Pjottos:
This PR removes the main lock in `Context` and changes some parts of the API to accommodate this change, mainly necessary to satisfy the borrow checker.

This change brings the following advantages:
- No more deadlocks, code that would create a deadlock now gives a compile time error.
- Since `Context` is no longer required to be `Send + Sync`, callback shapes can capture `!Send + !Sync` types.  For example the `three-d` context.
- Improves performance by a couple percent.
- It is no longer possible to use the same `Context` to draw UI on multiple threads at the same time, which is bound to give incorrect results. 

  Closes #1399

Apologies for the huge PR, but the bulk of the changes is basically search and replace so I hope it's not too difficult to review.


Titaniumtown:
Just merged this into my own fork of egui, so I have a few comments as I just migrated my egui application to base off of this:
- maybe make a note of the requirement of using `memory_mut` instead of simply `memory` to gain mutable access to memory
- make a note of requiring a mutable access to context with things like `on_hover_text`

If this PR is to be accepted, I would recommend something of a migration guide or some deep documentation of the developer-facing changes so projects can more easily migrate.

emilk:
This is a very nice effort, but I am not so happy with the worsening ergonomics. Adding `<'_>` everywhere is very ugly imho, and many methods now require an extra `&mut Context` argument.

Titaniumtown:
Will any work be done on this? This was very promising.

emilk:
I decided to go this route instead:
* https://github.com/emilk/egui/pull/2625

Still, thanks so much for all the work in this PR 🙏

Issue #1737, opened 2022-06-14
eframe: persistence stores "inner_pos"


Barugon:
**Describe the bug**
Enabling persistence will cause the window's "inner_pos" to be stored. When this position is read on the next launch, it appears to be applied to the window's outer position, resulting in the window being moved down 37 pixels each launch.

**To Reproduce**
1. Enable persistence feature
2. Launch app
3. Note window's position
5. Close and launch app again
6. Note window's position

**Expected behavior**
Window is initialized in the same place each launch.

**Desktop (please complete the following information):**
 - OS: Fedora 36 (Gnome)

Barugon:
From the winit documentation:

- **macOS:** The top left corner position of the window content, the window’s “inner” position. The window title bar will be placed above it. The window will be positioned such that it fits on screen, maintaining set `inner_size` if any. If you need to precisely position the top left corner of the whole window you have to use [Window::set_outer_position](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_outer_position) after creating the window.
- **Windows:** The top left corner position of the window title bar, the window’s “outer” position. There may be a small gap between this position and the window due to the specifics of the Window Manager.
- **X11:** The top left corner of the window, the window’s “outer” position.

So the only platform that uses `inner_pos` to set the window position is MacOS, all others use the outer window position.

PR #1745, opened 2022-06-16
Fix persistence of window position

Barugon:
**Issue**
#1737

**Changes**
Store the inner position only for MacOS; other platforms store the outer position.

### DIFF
--- a/egui-winit/src/window_settings.rs
+++ b/egui-winit/src/window_settings.rs
@@ -2,8 +2,9 @@
 #[derive(Clone, Copy, Debug)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub struct WindowSettings {
-    /// Inner position of window in physical pixels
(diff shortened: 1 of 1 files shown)


egui-winit/src/window_settings.rs
```diff
@@ -2,22 +2,32 @@
...
 }
 
 impl WindowSettings {
     pub fn from_display(window: &winit::window::Window) -> Self {
         let inner_size_points = window.inner_size().to_logical::<f32>(window.scale_factor());
-
-        Self {
-            inner_pos: window
+        let position = if cfg!(macos) {
+            // Use the inner position for MacOS
```
Mingun, line 16: I think, it is better to wording like "MacOS uses inner position when positioning windows" or something, added details from your findings also would be helpful.

Right now this comment looks like Captain Obvious statement.

emilk (approved):
Great, thanks!

Barugon:
Thanks.

Issue #1518, opened 2022-04-20
Example custom_window_frame close button doesn't work on gnome and when I click on X button it drags the window


NiiightmareXD:
Video of the bug:

https://user-images.githubusercontent.com/90005793/164396930-0b76827e-5a88-4816-9f0e-d61b510759a9.mp4

**To Reproduce**
1. run cargo run -p custom_window_frame
2. click on X button (close button)

**Expected behavior**
Close the window

**Desktop (please complete the following information):**
 - OS: Fedora Linux
 - Version 36

Titaniumtown:
Could you maybe provide more info? There's a bug template 😕

NiiightmareXD:
> Could you maybe provide more info? There's a bug template confused

Ok edited it, I Also added a vidoe.

emilk:
Anyone feel like digging into this? Perhaps see if https://github.com/rust-windowing/winit/blob/master/examples/drag_window.rs works  for you

PikminGuts92:
> Anyone feel like digging into this? Perhaps see if https://github.com/rust-windowing/winit/blob/master/examples/drag_window.rs works for you

Yes, that example does seem to work fine for me without any noticeable issues.

Additional info is I'm on a 4K display with 150% scaling but switching to 100% didn't change anything.

NiiightmareXD:
> > Anyone feel like digging into this? Perhaps see if https://github.com/rust-windowing/winit/blob/master/examples/drag_window.rs works for you
> 
> Yes, that example does seem to work fine for me without any noticeable issues.
> 
> Additional info is I'm on a 4K display with 150% scaling but switching to 100% didn't change anything.

Are you using windows or linux? What distribution?

PikminGuts92:
> Are you using windows or linux? What distribution?

Windows 10 21H2 19044.1645

It seems that the 'X' button click event is caught somehow and prevents me from using the dark/light theme toggle buttons anymore. Force restarting the application is the only way to restore interaction.

https://user-images.githubusercontent.com/15707039/166829967-49b2a347-7ea6-42b3-a125-655b48b0df0c.mp4

NiiightmareXD:
> > Are you using windows or linux? What distribution?
> 
> Windows 10 21H2 19044.1645
> 
> It seems that the 'X' button click event is caught somehow and prevents me from using the dark/light theme toggle buttons anymore. Force restarting the application is the only way to restore interaction.
> 
>  Code_d4z00bR4B0.mp4

[...]

PR #1750, opened 2022-06-17
Fix bug in custom window example

lorrenbiffin:
Addresses a bug in the drag behavior implementation for the `custom_window_frame` example.

---
#### Existing Behavior:
1. `Sense::drag()` enables "sensing" a mouse drag event on an interactive widget.
2. `title_bar_response.drag_started()` returns true when a drag event is "sensed" on `title_bar_rect`.
3. `frame.drag_window()` sets the `drag_window` property on the `AppOutput` struct.
4. `AppOutput` is passed to `winit` via `handle_app_output(..)` within eframe's epi integration.
5. `handle_app_output(..)` calls winit's `Window::drag_window()` method, which (finally) calls `winuser::PostMessageW(..)` to perform the window move behavior in-sync with the current mouse position.

#### Problem(s):
1. Winit and egui widget's internal drag behavior are at odds with each other when invoked at the same time.

#### Proposed Solution:
1. Allow the window drag handle (in this case, `title_bar_rect`) to sense clicks with `Sense::click()`.
    - Note: We don't need to (and probably shouldn't) sense dragging on the widget itself.
2. Detect when the mouse button is down on the title bar with `title_bar_response.is_pointer_button_down_on()`
3. Call `frame.drag_window()` and allow winit to do its job.
    - When the mouse button is released, `is_pointer_button_down_on()` no longer returns true, which allows `AppOutput` to tell winit that we're finished dragging, which ceases to call `PostMessageW(..)`, allowing mouse capture to return to egui.

---
Closes https://github.com/emilk/egui/issues/1518


emilk:
Thanks for this, but I still don't understand what the bug is (`master` works on my Mac).

https://github.com/emilk/egui/issues/1518#issuecomment-1156314231 hints that it is maybe `dragged` that stops being `true` on non-mac/windows? If so, it would be great to get to the bottom of that problem.

masc-it:
I am experiencing the same behaviour on Windows 10, only with the custom windows frame though. @lorrenbiffin your fix works, thank you 💯

emilk:
Ok, let’s merge this then though I would love a deeper understanding of what goes wrong 🤔

Ethanwolfie:
I also came across this problem on Window 10, with the latest commit [cd0f66b](https://github.com/emilk/egui/commit/cd0f66b9ae661be737ce57f26fb305cf124f6b73). After I swapped the code order as showed in the screeshot, the problem was gone.
codesnap

PR #1765, opened 2022-06-23
Draft: Premultiply font textures for linear blending

felixfaire:
It is my understanding that egui blends colours in linear space (which is arguably more correct than the SRGB blending of yore). However font textures in egui currently use alpha premultiplied pixels values in srgb space. This leads to darkened text edges when using linear blending, this is particularly noticeable with white text on a white background (see below).

image

This PR changes the font pixel premultiplication function to get (afaict) correctly premultiplied values for linear blending.

image
 
This problem and solution is well described in this article: https://hacksoflife.blogspot.com/2022/06/srgb-pre-multiplied-alpha-and.html

Notes: The Ubuntu thin font on white backgrounds can appear thinner due to this change (because it removes the grey border around text). I think the result is still legible and looks better but some other users may disagree.  A future release may wish to use a slightly thicker default font if this comes up in feedback.

Related: 
https://github.com/emilk/egui/issues/1410
https://github.com/emilk/egui/pull/1412
https://github.com/emilk/egui/pull/1411

### DIFF
--- a/egui-wgpu/src/renderer.rs
+++ b/egui-wgpu/src/renderer.rs
@@ -485,7 +485,7 @@ impl RenderPass {
                     image.pixels.len(),
                     "Mismatch between texture size and texel count"
                 );
-                Cow::Owned(image.srgba_pixels(1.0).collect::<Vec<_>>())
+                Cow::Owned(image.linear_premultiplied_srgb_pixels().collect::<Vec<_>>())
             }
         };
         let data_bytes: &[u8] = bytemuck::cast_slice(data_color32.as_slice());
--- a/egui_glium/src/painter.rs
+++ b/egui_glium/src/painter.rs
@@ -235,13 +235,10 @@ impl Painter {
                 );
                 image.pixels.iter().map(|color| color.to_tuple()).collect()
             }
-            egui::ImageData::Font(image) => {
-                let gamma = 1.0;
-                image
-                    .srgba_pixels(gamma)
-                    .map(|color| color.to_tuple())
-                    .collect()
-            }
+            egui::ImageData::Font(image) => image
+                .linear_premultiplied_srgb_pixels()
+                .map(|color| color.to_tuple())
+                .collect(),
         };
         let glium_image = glium::texture::RawImage2d {
             data: std::borrow::Cow::Owned(pixels),
--- a/egui_glow/src/painter.rs
+++ b/egui_glow/src/painter.rs
@@ -505,13 +505,14 @@ impl Painter {
                     "Mismatch between texture size and texel count"
                 );
 
+                // TODO (felix): confirm what this conditional is about
                 let gamma = if self.is_embedded && self.post_process.is_none() {
                     1.0 / 2.2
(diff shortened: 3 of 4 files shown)


emilk:
This effectively reverses https://github.com/emilk/egui/pull/1412

The problem with egui before #1412 was that dark text on white background was too faint. #1412 fixed this, but added the above problem of ugly edges of text.

It of course depends on your use case what problem is the biggest one, but I got many complaints about text being too faint in bright mode, that I believe that problem to be the bigger one.

We should be looking for a solution that fixed both problems at once! Some more clues can be found in https://github.com/emilk/egui/pull/1411

## Current `master`:
master

## This PR:
pr1765

felixfaire:
I appreciate that legible text in light mode is definitely a priority, though i find the solution of hacking the blending to make the thin font work (and look worse for all other fonts in the process) a bit extreme.. I think proper solutions to this might include.

Using an appropriately weighted font for the default font size (Ubuntu regular for example).
Using a darker colour for text in light mode (it's currently set at a relatively light grey).
Making the default font size bigger.
For me (mostly on a macbook) the text is perfectly legible on this PR in light mode (even more-so with black as the font color vs grey). Though i appreciate this isn't for everyone. Perhaps the defaults of egui should change size / color to be more accessible to more users who have an issue with correct blending, then there is always the option to make the font smaller / lighter for users who need the space / density.

I don't consider this issue resolved / PR closed.. So please reconsider looking over this as a serious issue (with solutions outlined above).

felixfaire:
I appreciate that legible text in light mode is definitely a priority, though i find the solution of hacking the blending to make the thin font work (and look worse for all other fonts in the process) a bit extreme.. I think proper solutions to this might include.

1. Using an appropriately weighted font for the default font size (Ubuntu regular for example).
2. Using a darker colour for text in light mode (it's currently set at a relatively light grey).
3. Making the default font size bigger.

For me (mostly on a macbook) the text is perfectly legible on this PR in light mode (even more-so with black as the font color vs grey). Though i appreciate this isn't for everyone. Perhaps the defaults of egui should change size / color to be more accessible to more users who have an issue with correct blending, then there is always the option to make the font smaller / lighter for users who need the space / density.

I don't consider this issue resolved / PR closed.. So please reconsider looking over this as a serious issue (with solutions outlined above).

emilk:
I agree with you @felixfaire - https://github.com/emilk/egui/pull/1412 is a hack in lieu of a proper fix.

On a high DPI display (e.g. a macbook) all this blending problems becomes much less noticeable simply because the pixels are smaller. So for instance, a 12 point font covers more pixels, so there is more "meat" to each letter.

Before this PR the same font, size and color looked substantially different in egui compared to any other place (especially on low resolution displays), so https://github.com/emilk/egui/issues/1410 is a proper bug that absolutely needed fixing, and I don't want to merge any PR that reopens that rather serious bug.

Your suggestions are great, but they also need to be matched with some fixes in the code along the lines of https://github.com/emilk/egui/pull/1411

PR #1774, opened 2022-06-29
Support rich text convert from `Cow<str>`

KmolYuan:
Since `ToString` is not used here, type `Cow<str>` is sometimes used as a string variable, such as [`String::from_utf8_lossy`](https://doc.rust-lang.org/alloc/string/struct.String.html#method.from_utf8_lossy) and [Fluent](https://github.com/projectfluent/fluent-rs) project.


emilk:
Thanks!

PR #1776, opened 2022-06-29
Add doc example for hint_text persistence

zenofsahil:
I had a requirement of persisting the hint text in the TextEdit widget. There currently isn't a straightforward way to achieve this without making changes to the API so I figured this workaround example could be useful for someone with the same use case.


emilk (approved):
Nice!

Issue #1531, opened 2022-04-24
ScrollArea scroll state modified by sibling textedit


sphaerophoria:
**Describe the bug**
If I add a text edit, then a scroll area, typing in the text edit moves the position of the scroll area.
E.g.
![scroll_on_text_entry](https://user-images.githubusercontent.com/1069811/164999923-29f99770-4286-48f8-88c7-c18ca95b54da.gif)

**To Reproduce**
Sample program...
```
use eframe::egui::{self, ScrollArea};

fn main() {
    let native_options = eframe::NativeOptions::default();
    eframe::run_native("My egui App", native_options, Box::new(|cc| Box::new(MyEguiApp::new(cc))));
}

#[derive(Default)]
struct MyEguiApp {
    input: String,
}

impl MyEguiApp {
    fn new(cc: &eframe::CreationContext<'_>) -> Self {
        // Customize egui here with cc.egui_ctx.set_fonts and cc.egui_ctx.set_visuals.
        // Restore app state using cc.storage (requires the "persistence" feature).
        // Use the cc.gl (a glow::Context) to create graphics shaders and buffers that you can use
        // for e.g. egui::PaintCallback.
        Self::default()
    }
}

impl eframe::App for MyEguiApp {
   fn update(&mut self, ctx: &egui::Context, frame: &mut eframe::Frame) {
       egui::CentralPanel::default().show(ctx, |ui| {
           ui.text_edit_singleline(&mut self.input);
           ScrollArea::vertical()
               .show(ui, |ui| {
                   for i in 0..100 {
                       ui.label(format!("test{}", i));
                   }
               });
       });
   }
}
```

**Expected behavior**
Scroll area stays fixed when typing in adjacent boxes

**Desktop (please complete the following information):**
 - OS: Linux, observed on both Ubuntu 20.04 and Arch Linux

**Additional context**
* Typing when the text edit is not selected does not move the scroll area
* I tried a few workarounds thinking that it might be related to internal state from the text edit leaking into the scroll area
    * Adding id sources to the text edit or the scroll area didn't seem to help
    * Wrapping the text edit within a sub-ui with ui.allocate_rect also doesn't resolve the issue

sphaerophoria:
I'm unsure if this is by design or a mistake.

[...]

PR #1779, opened 2022-06-30
Don't call scroll if TextEdit is fully in view

Barugon:
Very narrow fix for #1531

I believe the logic implemented to scroll a `TextEdit` into view is just incorrect (see comments in issue) but this change will at least prevent an adjacent scroll area from scrolling while entering text.

### DIFF
--- a/egui/src/widgets/text_edit/builder.rs
+++ b/egui/src/widgets/text_edit/builder.rs
@@ -590,8 +590,8 @@ impl<'t> TextEdit<'t> {
                             &cursor_range.primary,
                         );
 
-                        if (response.changed || selection_changed)
-                            && !ui.clip_rect().contains_rect(rect)
+                        let is_fully_visible = ui.clip_rect().contains_rect(rect); // TODO: remove this HACK workaround for https://github.com/emilk/egui/issues/1531
+                        if (response.changed || selection_changed) && !is_fully_visible
                         {
                             ui.scroll_to_rect(cursor_pos, None); // keep cursor in view
                         }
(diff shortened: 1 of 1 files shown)


egui/src/widgets/text_edit/builder.rs
```diff
@@ -590,7 +590,9 @@ impl<'t> TextEdit<'t> {
                             &cursor_range.primary,
                         );
 
-                        if response.changed || selection_changed {
+                        if (response.changed || selection_changed)
+                            && !ui.clip_rect().contains_rect(rect)
+                        {
```
emilk, line 595:
```suggestion
                        let is_fully_visible = !ui.clip_rect().contains_rect(rect); // TODO: remove this HACK workaround for https://github.com/emilk/egui/issues/1531
                        if (response.changed || selection_changed) && is_fully_visible
                        {
```

egui/src/widgets/text_edit/builder.rs
```diff
@@ -590,7 +590,9 @@ impl<'t> TextEdit<'t> {
                             &cursor_range.primary,
                         );
 
-                        if response.changed || selection_changed {
+                        let is_fully_visible = ui.clip_rect().contains_rect(rect); // TODO: remove this HACK workaround for https://github.com/emilk/egui/issues/1531
+                        if (response.changed || selection_changed) && !is_fully_visible
+                        {
```
emilk, line 595:
```suggestion
                        if (response.changed || selection_changed) && !is_fully_visible {
```

Issue #1755, opened 2022-06-19
egui-wgpu on web


Titaniumtown:
How could one use the wgpu integration for egui on wasm? I can't find a way to do this. Is this possible?

emilk:
It probably requires some work, but should definitely be doable (and desirable!).

@expenses claims to be using `egui-wgpu` for WebGL in https://github.com/emilk/egui/pull/1671 - maybe they have some input?

expenses:
Yep, but it's currently locked to using webgl and not webgpu: https://github.com/emilk/egui/blob/bd5f553c3a33fd9de0f396634044117bd519342d/egui-wgpu/Cargo.toml#L43

We'd just need to remove this feature to get it work for both backends

Titaniumtown:
I assume this cannot be done as easily as `eframe::start_web`... how did you do it exactly?

felixfaire:
We are using egui-wgpu successfully on wasm (with the webgl2 backend currently, though i don't think it should be different for webgpu). 

We have an existing wgpu context and custom event loop, input events etc (driven by typescript side), i'm not au fait with how eframe does it, but we followed the 'integrations' part of the eframe repo and got it up and running fairly seamlessly. Which bit are you struggling with in particular? @Titaniumtown

emilk:
@geolehmann thanks for that example! I notice that you use winit (and egui-winit) on the web, which is interesting (see https://github.com/emilk/egui/issues/1032).

metacean:
I have done my own barebones integration of egui on top of a 'raw' wgpu app originally based on the 'learn wgpu' tutorials. this runs on native and in browser wth wgpu (no webgl fallback). I also have hooked up basic mouse move and button click events as i required them.

I dont really have a 'releaseable' example app i can publish (and I use camel case which likely makes me an heretic in the wider rust community) but, in case its a helpful to anyone - my renderers state struct has the following added:

guiContext: egui::Context,
guiRenderPass: egui_wgpu::renderer::RenderPass,
guiScreenDescriptor: egui_wgpu::renderer::ScreenDescriptor,
guiInput: egui::RawInput,
guiMousePositionInPoints: egui::Pos2, //cache our mouse position as we need it for events

these are initialised as follows:

let guiContext = egui::Context::default();
let guiRenderPass = egui_wgpu::renderer::RenderPass::new(&device, config.format, 1);
let guiScreenDescriptor = egui_wgpu::renderer::ScreenDescriptor {
[...]

PR #1781, opened 2022-06-30
Don't force the webgl backend in egui-wgpu

expenses:
See https://github.com/emilk/egui/issues/1755#issuecomment-1159812326. At the moment we force the backend of `wgpu` when compiling for `wasm32-unknown-unknown` to be webgl, but we don't need to do this.

This PR should probably have some testing to make sure that egui works with a WebGPU backend. If no tests for `wasm32-unknown-unknown` exist, I'm not sure I have the time to set that them at this time.


emilk (approved):
Makes sense!

Issue #1800, opened 2022-07-05
Way to set eframe window visiblity


NotNite:
It would be nice to have some sort of `Frame::set_visible()` to toggle the window visibility, as I want my app to sit in the background and open from a hotkey.

NotNite:
I've written this functionality in https://github.com/NotNite/egui/commits/eframe-visible, but not sure if I should PR it.

PR #1808, opened 2022-07-07
(eframe) add Frame.set_visible

NotNite:
Closes <https://github.com/emilk/egui/issues/1800>.

You can use `frame.set_visibility(false);` to hide the window - good for background applications that open from a hotkey.

### DIFF
--- a/eframe/src/epi.rs
+++ b/eframe/src/epi.rs
@@ -544,7 +544,7 @@ impl Frame {
     }
 
     /// Set the visibility of the window.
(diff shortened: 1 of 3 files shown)


NotNite:
(Still needs support for web.)

rdrpenguin04:
Is there a reason you chose `set_visibility` instead of `set_visible`?

Other than that, I think it looks good, but I'm not a maintainer.

emilk:
Looks good, but I agree that `set_visible` is a better name.

NotNite:
Renamed it.

PR #1812, opened 2022-07-08
Add support for ctrl+h as backspace

zenofsahil:
Ctrl+h is commonly supported when editing text input. Would be good to have for the egui textedits as well.

### DIFF
--- a/egui/src/widgets/text_edit/builder.rs
+++ b/egui/src/widgets/text_edit/builder.rs
@@ -1090,6 +1090,11 @@ fn on_key_press(
             None
         }
 
+        Key::H if modifiers.ctrl => {
+            let ccursor = delete_previous_char(text, cursor_range.primary.ccursor);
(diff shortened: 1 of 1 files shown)


Barugon:
What OS would this apply to, because it's certainly not a Linux thing and I don't ever recall it being a Windows hotkey for backspace.

zenofsahil:
Ctrl+h deletes the previous character in bash/zsh. Its supported in vim. It's supported within most applications that accept any text input on macs. The GNU readline library supports it AFAIK so pretty sure its there on Linux too.

Barugon:
Well, I didn't know that about bash. Learn something new every day.

zenofsahil:
Bash uses the readline library to accept text input. readline supports emacs/vi keybindings. The keybindings section in `man readline` list out all the supported emacs/vi bindings.

emilk (approved):

Issue #1485, opened 2022-04-12
Enormous cloning, background work?


wbrickner:
Hi!

## Cloning

I'm trying to use the `Plot` and `Line` API.
This requires that I consume my data by value, I'd much rather keep a cached copy & provide a reference.

My plot data is fairly large (1M+ elements, sometimes 50M).

If this is not possible, can I add the API to accept something which can be `.as_ref()` to a `&[(f64, f64)]`?

I suspect the by-value nature of the API has something to do with the immediate mode design of `egui`... however, in cases like this it doesn't seem to actually make sense if it is driven only be design purity, ykwim?  Your thoughts on this / advice would help me.

## Background Work

I am running massive simulations (~hours completion time) & interactively showing the results over time is the goal.  I'd like to be able to do this in the background as mentioned.  It is mentioned a few times in the documentation that long-running tasks should be deferred to background tasks in an async manner.  I agree.  I'm lost on how to accomplish this.  I can think of a few clumsy ways, but I want to know if there is a proper way to do this that makes things easy and performant in `egui`.

Thank you!

wbrickner:
Update: confirmed, displaying a small subset of my data (~2%, not more than a few hundred kb of floats) causes frame rate to become unusable on release + LTO-fat build, like 1-5hz.

`Value` structs created ahead of time and then must be cloned every frame.  I don't know for sure that the cloning is the culprit, but I strongly suspect.

After some small amount of brain-using, I think it is actually exactly because of the immediate mode architecture that there is no reason for this API to take ownership of the data, or even to require `Value` construction.

As for background work, all I've done for now is set an `Arc<Mutex<XYZ>>` field on the `App` struct that can be directly given to spawned threads etc.  Not sure if this is a good idea!

wbrickner:
Ok after reviewing the repo locally, the story is a little more complicated.  

The original thinking was that because everything is immediate mode, there can never be any non-trivial borrowing patterns involved, so it would be safe to throw in lifetime bounds and references inside `Line` etc & it wouldn't ever get complicated later.

I think this may actually be true, but I'm unsure.  This is going to depend on when and how long the ref can possibly be asked to live.  I think because `PreparedPlot::ui` is only called from `Plot::show`, and the `PreparedPlot` is dropped there, that this only  introduces the constraint that the series data referenced exists outside the closure on `Plot::show`, which isn't very burdensome at all, and probably aligns with basically all use cases better.

[...]

PR #1816, opened 2022-07-10
Improve plot item UX

EmbersArc:
Improves the ergonomics of creating plot items. I also tried to address #1485 by allowing the user to pass borrowed data. I however couldn't find a way to do this for non-`static` life times. Maybe someone can find a way to add this as well.

This adds `bytemuck` as a dependency for the `egui` crate but I figured it's okay since all integrations have it as a dependency anyway. I can remove it again otherwise.


egui/src/widgets/plot/items/values.rs
```diff
@@ -7,16 +7,23 @@ use crate::plot::transform::PlotBounds;
...
     /// Goes from left to right.
     pub x: f64,
     /// Goes from bottom to top (inverse of everything else in egui!).
     pub y: f64,
 }
 
-impl Value {
+impl From<[f64; 2]> for PlotPoint {
+    fn from(point: [f64; 2]) -> Self {
+        bytemuck::cast(point)
```
emilk, line 22:
```suggestion
    #[inline]
    fn from(point: [f64; 2]) -> Self {
        Self { x: point[0], y: point[1] }
```

egui/src/widgets/plot/items/values.rs
```diff
@@ -140,22 +147,45 @@ impl Default for Orientation {
 
 // ----------------------------------------------------------------------------
 
-#[derive(Default)]
-pub struct Values {
-    pub(super) values: Vec<Value>,
-    generator: Option<ExplicitGenerator>,
+pub enum PlotPoints<'v> {
```
emilk, line 150: Good improvement, but I think it would be nice to also add a docstring

```suggestion
/// Represents many [`PlotPoint`]s.
///
/// These can be a borrowed slice, an owned `Vec`, or generated with a function.
pub enum PlotPoints<'v> {
```

egui/src/widgets/plot/items/values.rs
```diff
@@ -140,22 +147,45 @@ impl Default for Orientation {
...
+}
+
+impl<'v> PlotPoints<'v> {
+    pub fn new(points: Vec<[f64; 2]>) -> Self {
+        Self::Owned(bytemuck::cast_vec(points))
+    }
+
+    pub fn from_slice(points: &'v [[f64; 2]]) -> Self {
+        Self::Borrowed(bytemuck::cast_slice(points))
+    }
```
emilk, line 181: `bytemuck` is nice, but it also makes compiling egui from scratch take ~twice as long (because of the `syn` crate). I'd rather keep the number of required dependencies low, so let's make `bytemuck` opt-in.

I think it's good to prefix these with `#[cfg(feature = "bytemuck")]`

egui_demo_app/Cargo.toml
```diff
@@ -45,7 +45,7 @@ egui_demo_lib = { version = "0.18.0", path = "../egui_demo_lib", features = ["ch
 
 # Optional dependencies:
 
-bytemuck = { version = "1.9.1", optional = true }
+bytemuck = { version = "1.7.1", optional = true }
```
emilk, line 48: why roll back to an earlier version?

egui_demo_lib/src/demo/context_menu.rs
```diff
@@ -116,17 +116,21 @@ impl super::View for ContextMenus {
...
+                .map(|i| {
+                    use std::f64::consts::TAU;
+                    let x = egui::remap(i as f64, 0.0..=n as f64, -TAU..=TAU);
+                    match self.plot {
+                        Plot::Sin => [x, x.sin()],
+                        Plot::Bell => [x, 10.0 * gaussian(x)],
+                        Plot::Sigmoid => [x, sigmoid(x)],
+                    }
+                })
+                .collect::<Vec<_>>(),
```
emilk, line 132: Is `.collect::<Vec<_>>()` really more ergonomic than `PlotPoints::from_iter(…)` ?

emilk (changes requested):
Overall, I like the new names (`PlotPoint` is much better than `Value`).

But let's make a `bytemuck` opt-in - sure the official integrations all use it, but not all 3rd party does, and as I noted it makes compiling egui a lot slower.

egui_demo_lib/src/demo/context_menu.rs
EmbersArc: Good idea, I implemented `FromIterator` for `PlotPoints` and used this in the examples.

egui_demo_app/Cargo.toml
EmbersArc: I noticed that the other crates use 1.7, so this should avoid building two different versions. But I can change it back or update all of them if you like.

egui/src/widgets/plot/items/values.rs
EmbersArc: Ah good point. I removed the dependency again. I also removed `PlotPoints::Borrowed` for now because I couldn't find a way to to make it work with arbitrary lifetimes. Once somebody finds a way to get that working, it will make sense to add bytemuck again.

egui/src/widgets/plot/items/mod.rs
```diff
@@ -837,12 +837,13 @@ impl PlotItem for Points {
             stem_stroke.width *= 2.0;
         }
 
-        let y_reference = stems.map(|y| transform.position_from_value(&Value::new(0.0, y)).y);
+        let y_reference =
+            stems.map(|y| transform.position_from_point(&PlotPoint::new(0.0, y)).y as f32);
```
emilk, line 841:
```suggestion
            stems.map(|y| transform.position_from_point(&PlotPoint::new(0.0, y)).y);
```

emilk (approved):
Awesome!

egui/src/widgets/plot/items/mod.rs
```diff
@@ -837,12 +837,13 @@ impl PlotItem for Points {
             stem_stroke.width *= 2.0;
         }
 
-        let y_reference = stems.map(|y| transform.position_from_value(&Value::new(0.0, y)).y);
+        let y_reference =
+            stems.map(|y| transform.position_from_point(&PlotPoint::new(0.0, y)).y);
```
emilk, line 841:
```suggestion
        let y_reference = stems.map(|y| transform.position_from_point(&PlotPoint::new(0.0, y)).y);
```

emilk:
Closes https://github.com/emilk/egui/pull/1843

egui/src/widgets/plot/items/values.rs
```diff
@@ -3,20 +3,30 @@ use std::ops::{Bound, RangeBounds, RangeInclusive};
...
 }
 
-impl Value {
+impl From<[f64; 2]> for PlotPoint {
+    #[inline]
+    fn from(point: [f64; 2]) -> Self {
+        Self {
+            x: point[0],
+            y: point[1],
+        }
```
emilk, line 25:
```suggestion
    fn from([x, y]: [f64; 2]) -> Self {
        Self { x, y }
```

enomado:
> Overall, I like the new names (`PlotPoint` is much better than `Value`).
> 

I like Value. The idea is that you have some metric `space` of functions you draw, it translates into screen-plot-space. but you already name it PlotPoint. 

In other words we have screen-space, plot-space, and value-space. but value-space is plot-space now. 

That could be misunderstood

emilk:
There are only two coordinate systems though, afaict: the normal screen-space point-based coordinate system of egui/epaint, and the coordinate system of the data in the plot.

The reason I like "PlotPoint" is that it is descriptive it is a point in plot space. "Value" is about as vague as "Data" or "Thing". After all, what variable is _not_ a value?

PR #1824, opened 2022-07-13
Remove dependency on AsRef trait for TextBuffer

Kryptos-FR:
**Summary**
Remove the dependency on the `AsRef<str>` trait for `TextBuffer`.

**Rationale**
That dependency is not needed as it is trivial to implement `fn as_str(&self) -> &str` for both `str` and `String` which are the default implementations provided by `egui` and very likely the one that most user will ever need. Then `TextEdit` can just use `as_str()` in places where it was using `as_ref()` before, which in my opinion makes it more consistent with the trait definition.

In one of the project I am working on we did implement a custom `TextBuffer` for our `struct Var`. The forced dependency on `AsRef<str>` caused some pains because:
* our `struct Var` is a kind of container that can have all kind of underlying data type (not only string).
* in other places, it forced us to use the turbofish where it was not needed before:
```rust
// before implementing TextBuffer for Var
let headers: &mut Var = self.some_function();
let mut headers: Table = headers.as_ref().try_into()?; // here it uses the usual as_ref()

// after
let headers: &mut Var = self.some_function();
let mut headers: Table = <&mut Var as AsRef<Var>>::.as_ref(&headers).try_into()?; // here as_ref() was ambiguous
```
We anticipate that other occurrences will appear over time and we would like to avoid them.

**Checklist**
- [x] It is a trivial change.
- [x] I have run `./sh/check.sh`.
  - *There are errors but unrelated to this change.*

### DIFF
--- a/egui/src/widgets/text_edit/builder.rs
+++ b/egui/src/widgets/text_edit/builder.rs
@@ -378,7 +378,7 @@ impl<'t> TextEdit<'t> {
             // .unwrap_or_else(|| ui.style().interact(&response).text_color()); // too bright
             .unwrap_or_else(|| ui.visuals().widgets.inactive.text_color());
 
-        let prev_text = text.as_ref().to_owned();
+        let prev_text = text.as_str().to_owned();
 
         let font_id = font_selection.resolve(ui.style());
         let row_height = ui.fonts().row_height(&font_id);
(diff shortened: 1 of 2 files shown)


emilk (approved):
Sounds good to me!

PR #1826, opened 2022-07-13
Make EasyMark numbered lists allow more than 2 digits

YgorSouza:


emilk (approved):
Shorter and better! 👏

PR #1831, opened 2022-07-15
Add a stick to bottom feature to the `TableBuilder`

Kerollmops:
This PR introduces a new `TableBuilder::stick_to_bottom` method that makes the table follow the content as it comes.
Do you think that I need to open an issue for this small addition? If so, I will.

I ran the `./sh.check.sh` script but it returned a warning (converted into an error), not sure this PR touches anything related to that:

```
error: associated function is never used: `post_rendering`
   --> eframe/src/native/epi_integration.rs:290:12
    |
290 |     pub fn post_rendering(&mut self, app: &mut dyn epi::App, window: &winit::window::Window) {
    |            ^^^^^^^^^^^^^^
    |
    = note: `-D dead-code` implied by `-D warnings`
```


emilk (approved):

emilk:
Ooh, there were two PR:s at the same time! This has already been added in https://github.com/emilk/egui/pull/1849

Issue #1836, opened 2022-07-17
First element of the list of elements, takes up entire window height, when using RTL or LTR layout methods


gouravkhator:
## Issue

I faced one layout issue while rendering the list of elements. 

### Issue Description with Code Snippet

I have created a dummy app [here](https://github.com/gouravkhator/rusty-users/blob/main/src/render_ui.rs), in which the hyperlink needs to be rendered with `right_to_left` layout. 

If I don't use the `with_cross_align()` method, then as checked in below screenshot, the first element in the list takes up entire window size.

**Code Snippet:**

```rust
ui.with_layout(
    Layout::right_to_left(),
    |ui| {
        ui.add(egui::Hyperlink::from_label_and_url(
            "See User's Profile",
            format!("http://example.com/u/{}", user.id),
        ));
    },
);
```

**Issue Screenshot**

![Issue_screenshot](https://user-images.githubusercontent.com/36273363/179375146-9c04f369-1667-418a-877d-0c9bd8157732.png)

### Workaround to be performed by every developers

We can fix the above using the `with_cross_align` method and passing the `Align::LEFT` in the params. But this is just a workaround, and not the actual fix.

The code snippet for that dummy app, along with the fix is also mentioned [here](https://github.com/gouravkhator/rusty-users/blob/1802ba52b724095edf4f82afd58630602d99ff51/src/render_ui.rs#L41).

**Code Snippet**:

```rust
ui.with_layout(
    Layout::right_to_left().with_cross_align(Align::LEFT),
    |ui| {
        ui.add(egui::Hyperlink::from_label_and_url(
            "See User's Profile",
            format!("http://example.com/u/{}", user.id),
        ));
    },
);
```

**Screenshot**

![Fixed_Issue_Screenshot](https://user-images.githubusercontent.com/36273363/179375154-3b796ac8-f006-407e-855c-711226bdbf61.png)

> Note: I have submitted PR [#1837](https://github.com/emilk/egui/pull/1837) for the actual fix. I would like the creator of this awesome library, to look into that and help merge the fix asap.

emilk:
The reason for this behavior is twofold:

* [`with_layout` uses up all available space](https://docs.rs/egui/latest/egui/struct.Ui.html#method.with_layout)
* horizontal layouts centers vertically (as you noted)

[...]

PR #1837, opened 2022-07-17
Modified the cross_align attribute of the layout, to fix the first element in the list, taking up entire window height

gouravkhator:
## Fixing Issue #1836 

- In the `eframe::egui::Layout` struct, in the functions `right_to_left` and `left_to_right`, I have modified the default value of `cross_align` attribute to `Align::LEFT`.
- To keep the examples intact, where cross_align is not explicitly set using `with_cross_align`, I have added an explicit call to `with_cross_align()` method with `Align::Center` as the parameter value.

### DIFF
--- a/egui/src/layout.rs
+++ b/egui/src/layout.rs
@@ -158,7 +158,7 @@ impl Layout {
             main_wrap: false,
             main_align: Align::Center, // looks best to e.g. center text within a button
             main_justify: false,
-            cross_align: Align::Center,
+            cross_align: Align::LEFT,
             cross_justify: false,
         }
     }
@@ -170,7 +170,7 @@ impl Layout {
             main_wrap: false,
             main_align: Align::Center, // looks best to e.g. center text within a button
             main_justify: false,
-            cross_align: Align::Center,
+            cross_align: Align::LEFT,
             cross_justify: false,
         }
     }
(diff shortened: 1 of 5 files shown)


egui/src/layout.rs
```diff
@@ -158,7 +158,7 @@ impl Layout {
             main_wrap: false,
             main_align: Align::Center, // looks best to e.g. center text within a button
             main_justify: false,
-            cross_align: Align::Center,
+            cross_align: Align::LEFT,
```
emilk, line 161: This is confusing - the _cross_ axis is the vertical axis in this case, so `Align::TOP` would be a better description here.

emilk:
This is a scary change, as it is a breaking change.

The current value is intentional. Say you allocate a menu bar (maybe 24 pixels high) and you use `right_to_left` in it. In this case you likely want the widgets you add to it to be centered. For instance, adding a label and a button of different height it looks best if they are vertically centered. This is also the behavior of `ui.horizontal`.

Perhaps this can be solved by adding better helpers to the `Ui` interface instead? There is currently `ui.horizontal` and `ui.horizontal_top` - perhaps adding `ui.right_to_left` and `ui.right_to_left_top` would scratch your itch without adding a breaking change?

PR #1854, opened 2022-07-25
Make sliders correctly generate events on change.

ndarilek:
It doesn't appear as if change events are correctly generated when sliders update. This at least generates the events, though I'm a bit unclear on how keyboard-handling is intended to work with these.

### DIFF
--- a/egui/src/widgets/slider.rs
+++ b/egui/src/widgets/slider.rs
@@ -327,9 +327,6 @@ impl<'a> Slider<'a> {
             self.set_value(new_value);
         }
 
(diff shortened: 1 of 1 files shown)


ndarilek:
Not sure this failure is my fault--looks like wasm-bindgen may need to be pinned in the job definition.

emilk (approved):

emilk:
> Not sure this failure is my fault--looks like wasm-bindgen may need to be pinned in the job definition.

Correct you are! Fixing in https://github.com/emilk/egui/pull/1855

PR #1856, opened 2022-07-26
Add dry run feature for anchor calculation. (#1)

zapp88:
This PR resolves issue: emilk#1852
We introduce dry_run flag which makes component invisible until we do second pass of rendering - which allows us to properly calculate position for anchor. (This removes rapid flicker when new window is drawn for the first time).

### DIFF
--- a/egui/src/containers/area.rs
+++ b/egui/src/containers/area.rs
@@ -172,7 +172,12 @@ pub(crate) struct Prepared {
     pub(crate) movable: bool,
     enabled: bool,
     drag_bounds: Option<Rect>,
-    dry_run: bool,
+    /// Set the first frame of new windows with anchors.
+    ///
+    /// This is so that we use the first frame to calculate the window size,
+    /// and then can correctly position the window the next frame,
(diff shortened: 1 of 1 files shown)


egui/src/containers/area.rs
```diff
@@ -172,6 +172,7 @@ pub(crate) struct Prepared {
     pub(crate) movable: bool,
     enabled: bool,
     drag_bounds: Option<Rect>,
+    dry_run: bool,
```
emilk, line 175:
```suggestion
    /// Set the first frame of new windows with anchors.
    ///
    /// This is so that we use the first frame to calculate the window size,
    /// and then can correctly position the window the next frame,
    /// without having one frame where the window is positioned in the wrong place.
    temporarily_invisible: bool,
```

emilk (changes requested):
Interesting idea! Needs some docs / better naming though

egui/src/containers/area.rs
zapp88: The change has been applied.

emilk (approved):

PR #1858, opened 2022-07-26
Reset the scissor rect after rendering onto a renderpass

cwfitzgerald:
This prevents a easy-to-hit stumbling block when using egui-wgpu.

This actively resets the wgpu scissor rect back to being the whole output texture. For people who don't use scissor rects - this is the behavior that they expect - for people who do use them the value of the scissor rect doesn't matter because they will need to manually set it afterwards.


emilk (approved):
Good catch

Issue #1845, opened 2022-07-23
Global focus tracking


mwcampbell:
egui needs to know whether the UI as a whole (e.g. host window or web page) has keyboard focus. This is important for accessibility, but also for changing the way the UI is rendered when the window is visible but not focused. I have a solution for this that works with winit.

PR #1859, opened 2022-07-27
Track the global focus state of the UI

mwcampbell:
Closes <https://github.com/emilk/egui/issues/1845>.

### DIFF
--- a/egui/src/response.rs
+++ b/egui/src/response.rs
@@ -213,7 +213,9 @@ impl Response {
 
     /// This widget has the keyboard focus (i.e. is receiving key presses).
     pub fn has_focus(&self) -> bool {
-        self.ctx.memory().has_focus(self.id)
+        // Access input and memory in separate statements to prevent deadlock.
+        let has_global_focus = self.ctx.input().raw.has_focus;
+        has_global_focus && self.ctx.memory().has_focus(self.id)
     }
 
     /// True if this widget has keyboard focus this frame, but didn't last frame.
--- a/egui/src/response.rs
+++ b/egui/src/response.rs
@@ -212,6 +212,10 @@ impl Response {
     }
 
     /// This widget has the keyboard focus (i.e. is receiving key presses).
+    ///
+    /// This function only returns true if the UI as a whole (e.g. window)
+    /// also has the keyboard focus. That makes this function suitable
+    /// for style choices, e.g. a thicker border around focused widgets.
     pub fn has_focus(&self) -> bool {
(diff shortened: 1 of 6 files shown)


egui/src/response.rs
```diff
@@ -213,7 +213,9 @@ impl Response {
 
     /// This widget has the keyboard focus (i.e. is receiving key presses).
     pub fn has_focus(&self) -> bool {
-        self.ctx.memory().has_focus(self.id)
+        // Access input and memory in separate statements to prevent deadlock.
+        let has_global_focus = self.ctx.input().raw.has_focus;
+        has_global_focus && self.ctx.memory().has_focus(self.id)
```
emilk, line 218: This makes `Response::has_focus` and `Memory::has_focus` act differently, which seems like a potential footgun.

I'm unsure how to resolve it in a nice way. Perhaps it is good enough to document the difference, i.e. add a docstring to both `Response::has_focus` and `Memory::has_focus` indicating wether or not they are heeding `RawInput::has_focus`.

emilk (changes requested):
Very nice, I like it! But please add some docstrings to the two `fn has_focus` to explain which one heeds `RawInput::has_focus`, and why

mwcampbell:
Another option would be to make the `Focus::id` field `pub(crate)` (I have to do that anyway on my AccessKit branch), then replace `Memory::has_focus` with a new `Context::has_focus` function that checks both the memory and raw input. Then `Response::has_focus` can simply call `Context::has_focus`. Of course, I'd have to fix the widgets that directly call `Memory::has_focus`, but it would be more consistent that way.

emilk:
@mwcampbell I like that approach!

mwcampbell:
@emilk On closer investigation, I don't think my suggested solution will work. The most obvious reason is that other functions in `Memory` call `Memory::has_focus`. So I would at least have to keep that function and make it internal or `pub(crate)`. But also, in some cases where `Memory::has_focus` is used, we don't want the outcome to depend on whether the window has focus. For example, if you put a `DragValue` into keyboard editing mode, then move focus away from the window, then come back, you probably don't want the `DragValue` to switch to a button in the meantime. So I think I just have to document the difference between `Response::has_focus` and `Memory::has_focus`, or even add a second function in `Response` that takes global focus into account. Sometimes widgets really should check the global focus, e.g. when determining whether to render a border around a text edit.

emilk:
Let's just go with added docstrings for now!

emilk (approved):

Issue #1558, opened 2022-05-02
Unexpected Tab Behavior in hello_world Example


tataDan:
**Expected behavior**
When I press the `<Tab>` key, I would expect the focus to go to the `text_edit_singleline` for the name field. However, the focus instead goes to the slider below the `text_edit_singleline`.  If I then press `<Shift> + <Tab>` the focus does go to the  `text_edit_singleline`. Alternatively, I can click in the `text_edit_singleline` when the application starts and this works.

**Desktop (please complete the following information):**
 - OS: Linux 
 - Version: Debian 11 (Gnome)

**Additional context**

PR #1861, opened 2022-07-27
Fix focus behavior when pressing Tab with no focus

mwcampbell:
Closes <https://github.com/emilk/egui/issues/1558>.


emilk:
Thanks!

PR #1869, opened 2022-07-29
Improve documentation about panel order

xNxExOx:
Closes nothing.
I did not create an issue, because it seems unnecessary to create an issue, because documentation actually contains the information, I just think I can make it more "visible".


emilk (approved):
This is great, thanks!

PR #1872, opened 2022-07-30
better scroll inertia

enomado:
It is more natural, I guess :+1: 

btw my try to make standalone inertia, to use it in plot (but only in one direction) https://gist.github.com/enomado/b6995c7a17b2d0abf0a211242e2fc257

### DIFF
--- a/egui/src/containers/scroll_area.rs
+++ b/egui/src/containers/scroll_area.rs
@@ -627,14 +627,14 @@ impl Prepared {
                 }
             } else {
                 let stop_speed = 20.0; // Pixels per second.
-                let friction_coeff = 1000.0; // Pixels per second squared.
+                let friction_coeff = 68.0;
                 let dt = ui.input().unstable_dt;
 
                 let friction = friction_coeff * dt;
                 if friction > state.vel.length() || state.vel.length() < stop_speed {
                     state.vel = Vec2::ZERO;
                 } else {
-                    state.vel -= friction * state.vel.normalized();
+                    state.vel = state.vel / friction;
                     // Offset has an inverted coordinate system compared to
                     // the velocity, so we subtract it instead of adding it
                     state.offset -= state.vel * dt;


egui/src/containers/scroll_area.rs
```diff
@@ -627,14 +627,14 @@ impl Prepared {
...
-                let friction_coeff = 1000.0; // Pixels per second squared.
+                let friction_coeff = 68.0;
                 let dt = ui.input().unstable_dt;
 
                 let friction = friction_coeff * dt;
                 if friction > state.vel.length() || state.vel.length() < stop_speed {
                     state.vel = Vec2::ZERO;
                 } else {
-                    state.vel -= friction * state.vel.normalized();
+                    state.vel = state.vel / friction;
```
emilk, line 637: With a 120Hz display, `dt = 1/120`, so `friction = 68.0 / 120 = 0.5666…`, which means each frame the velocity will increase, which is a very serious bug to introduce.

emilk:
I agree it could be nice to separate out the inertia into a separate `struct`, but I'd rather not mess with code that works without taking a lot of care to avoid serious regressions like these.

enomado:
Current behavior is wrong(possible). It was frustrating for me, because I used to [this implementation](https://github.com/chrisbateman/impetus/blob/master/src/Impetus.js#L404)

I start googling and almost [everything](https://stackoverflow.com/a/33861363/720545) says that `vy *= friction` is more natural.

egui/src/containers/scroll_area.rs
enomado: The idea of doing `x * dt` is no maintain same travel vs various frame rates. if it behaves opposite, then we should divide instead of multiplicate. I'm agree with that I will test it.

enomado:
@emilk please, just try this version and feel how better it is.

emilk:
I tried `master` and this PR side by side, and honestly, on my Mac at 60 Hz, they feel mostly the same to me.

emilk:
> I start googling and almost [everything](https://stackoverflow.com/a/33861363/720545) says that `vy *= friction` is more natural.

The link is about how to slow down when you reach to top/bottom of the scrollable area ("rubber-banding") which egui doesn't implement yet.

--

egui currently use sliding friction, which is how something slows down when sliding on top of a surface (e.g. how you slow down when sliding on a frozen lake).

Multiplication is a different friction model that looses more velocity when moving fast than moving slow. This means if you scroll quickly, it will quickly slow down to a low speed, but keep that low speed for longer. I have no idea if this feels more or less natural, but I would love some references to some sources that have investigated this.

In any case, if you want to experiment with multiplicative friction I suggest you make use of the new `emath::exponential_smooth_factor` to get framerate independent behavior.

PR #1878, opened 2022-08-02
Add depth buffer in native 3D example

asny:
The `custom_3d_three-d` example does not enable a depth buffer since it is only rendering a triangle. However, if it is used as a starting point for other projects, it is highly likely that a depth buffer is actually needed, therefore I propose to enable it by default.
Also see [this](https://github.com/asny/three-d/issues/268) issue for full context.


emilk:
Good catch! And the example code already clears the non-existing depth-buffer 🤦 😄

asny:
Thanks 💪 And yeah a bit silly. I remember that I thought it was weird that the depth buffer only needed to be cleared on web, but I didn't connect the dots 🙂

PR #1881, opened 2022-08-02
Add `egui/default_fonts` feature to `pure_glow` example

asny:
It seems to me like the `pure_glow` example was broken sometime in april because of changes to feature flags. The text simply didn't show up which is due to missing fonts unless you figured out that you needed the `egui/default_fonts` feature flag. This change enforces the use of the `egui/default_fonts` feature flag in this example.


emilk (approved):

emilk:
thanks!

PR #1886, opened 2022-08-03
eframe web: access app from WebHandle

enomado:
This is a way to communicate with ConcreteApp from JavaScript.

There are few ways how to do it - 
1. place ConcreteApp in a Arc<Mutex> and save it somewhere before it goes to AppRunner, so it must be not Box.
2. make shared data in app and do nothing
3. This way. Its actually elegant and **not** quite hacky. 

I tried to cast `Box<dyn App>` to `&dyn Any` other ways, and dont find a solution. 

I stuck with `static lifetime, or [this](https://doc.rust-lang.org/std/any/index.html#smart-pointers-and-dyn-any)

here is the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=caccd8fa96005d23492194d3fb2e81da)


enomado:
I'm not understand what lint wants from me

>error: missing documentation for an associated function

Is there is another place where it should be documented?

eframe/src/epi.rs
```diff
@@ -49,6 +52,15 @@ pub trait App {
...
     fn update(&mut self, ctx: &egui::Context, frame: &mut Frame);
 
+    // Handle to the app.
+    //
+    // Can be used from web to interact or other external context
+    // Implementation is needed, because downcasting Box<dyn App> -> Box<dyn Any> to get &ConcreteApp is not simple in current rust.
+    //
+    // Just return &mut *self
+    #[cfg(target_arch = "wasm32")]
+    fn as_any_mut(&mut self) -> &mut dyn Any;
```
emilk, line 62: You accidentally used `//` instead of `///` - only the latter generated documentation!

egui_demo_app/src/lib.rs
```diff
@@ -43,6 +43,25 @@ impl WebHandle {
...
+        F: Fn(&mut WrapApp) -> (),
+    {
+        let mut runner_ref = self.handle.lock();
+        let app_ref = runner_ref.get_app_mut();
+        let app = app_ref.downcast_mut::<WrapApp>().unwrap();
+        func(app);
+    }
+
+    #[wasm_bindgen]
+    #[cfg(target_arch = "wasm32")]
```
emilk, line 59:
```suggestion
```

already covered by `WebHandle`

emilk:
This is a really great feature to implement, but I wonder if we can make it more ergonomic.

----

Let's investigate the current approach first (3.).

Every user must add

```rust
    fn as_any_mut(&mut self) -> &mut dyn Any {
        &mut *self
    }
```

I guess this is fine, and not much to do about it afacit.

However, I think we can improve this part:

```rust
    fn with_app<F>(&mut self, func: F) -> ()
    where
        F: Fn(&mut ConcreteApp) -> (),
    {
        let mut runner_ref = self.handle.lock();
        let app_ref = runner_ref.get_app_mut();
        let app = app_ref.downcast_mut::<ConcreteApp>().unwrap();
        func(app);
    }

   #[wasm_bindgen]	
    pub fn set_some_content_from_javasript(&mut self, _some_data: &str) {
        self.with_app(|app| {
            app.data = some_data;
        });
    }
```

With a helper function, this could become:


```rust
   #[wasm_bindgen]	
    pub fn set_some_content_from_javasript(&mut self, _some_data: &str) {
        let app = self.handle.lock().app_mut::<ConcreteApp>();
        app.data = some_data;
    }
```

much nicer!

----

About option 1 (`Arc`). We would need something like `Arc<Mutex<Box<dyn App>>>` on one end and `Arc<Mutex<Box<ConcreteApp>>>` on the other end. Is that even possible? If it was it would save users from having to implement `as_any_mut`, which is pretty nice.

emilk:
I don't understand GitHub. I pushed one commit, and it shows up here. I push another, and it doesn't show up.

```diff
diff --git a/eframe/src/web/backend.rs b/eframe/src/web/backend.rs
index 7849ea72..f231172a 100644
--- a/eframe/src/web/backend.rs
+++ b/eframe/src/web/backend.rs
@@ -35,7 +35,7 @@ impl WebInput {
 
 // ----------------------------------------------------------------------------
 
-use std::{any::Any, sync::atomic::Ordering::SeqCst};
+use std::sync::atomic::Ordering::SeqCst;
 
 /// Stores when to do the next repaint.
 pub struct NeedRepaint(Mutex<f64>);
@@ -265,8 +265,9 @@ impl AppRunner {
         &self.egui_ctx
     }
 
-    pub fn get_app_mut(&mut self) -> &mut dyn Any {
-        self.app.as_any_mut()
+    /// Get mutable access to the concrete [`App`] we enclose.
+    pub fn app_mut<ConreteApp: 'static + crate::App>(&mut self) -> &mut ConreteApp {
+        self.app.as_any_mut().downcast_mut::<ConreteApp>().unwrap()
     }
 
     pub fn auto_save(&mut self) {
diff --git a/egui_demo_app/src/lib.rs b/egui_demo_app/src/lib.rs
index b00cad8e..aafd53ae 100644
--- a/egui_demo_app/src/lib.rs
+++ b/egui_demo_app/src/lib.rs
@@ -32,7 +32,6 @@ pub struct WebHandle {
 #[wasm_bindgen]
 impl WebHandle {
     #[wasm_bindgen]
-    #[cfg(target_arch = "wasm32")]
     pub fn stop_web(&self) -> Result<(), wasm_bindgen::JsValue> {
         let mut app = self.handle.lock();
         let res = app.destroy();
@@ -44,22 +43,10 @@ impl WebHandle {
         res
     }
 
-    // helper for mutating original app from javascript
-    fn with_app<F>(&mut self, func: F) -> ()
-    where
-        F: Fn(&mut WrapApp) -> (),
-    {
-        let mut runner_ref = self.handle.lock();
-        let app_ref = runner_ref.get_app_mut();
-        let app = app_ref.downcast_mut::<WrapApp>().unwrap();
-        func(app);
-    }
-
     #[wasm_bindgen]
     pub fn set_some_content_from_javasript(&mut self, _some_data: &str) {
-        self.with_app(|_app| {
-            // app.data = some_data;
-        });
+        let _app = self.handle.lock().app_mut::<WrapApp>();
+        // _app.data = some_data;
     }
 }
```

enomado:
Can we merge this now?

enomado:
Hmm I dont get where is your modifications. May be we can merge this as is, since I'm only person who need this for now?

I'm so tired to manage all my outdated branches...

emilk:
Can you please apply my diff first?

enomado:
done

emilk:
Thanks!

PR #1897, opened 2022-08-10
Changed datepicker widget from egui_extras to use DateTime

terafo:
Chrono crate advises not to use **Date** type, and, to discourage to use it, that type is non serializable with serde. Since persistence feature of eframe requires serde, is is impossible to make datepicker state persistent without conversions. Switching **Date** type to **Datetime** fixes that issue.

### DIFF
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -232,7 +232,7 @@ pub enum HardwareAcceleration {
 
     /// Do NOT use graphics acceleration.
     ///
-    /// On some platforms (MacOS) this is ignored and treated the same as [`Self::Preferred`].
+    /// On some platforms (macOS) this is ignored and treated the same as [`Self::Preferred`].
     Off,
 }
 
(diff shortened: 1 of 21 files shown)


egui_extras/src/datepicker/button.rs
```diff
@@ -18,7 +18,7 @@ pub struct DatePickerButton<'a> {
 }
 
 impl<'a> DatePickerButton<'a> {
-    pub fn new(selection: &'a mut Date<Utc>) -> Self {
+    pub fn new(selection: &'a mut DateTime<Utc>) -> Self {
```
emilk, line 21: I'm worried that this is a confusing API. Why would an API that is about picking a date (without time) take an argument for selecting the data _and time_?

I think perhaps there is a chance here for a better fix: to stop using `chrono::Date` in favor of our own date type, `pub struct Date { pub year: u16, pub month: u8, pub day: u8 }` or whatever. That way we don't force the use of `chrono` (a non-maintained crate with several open RUSTSEC:s) onto our users.

dvg-vitol:
Whilst I agree that attaching time to explicit date functionality is not perfect, I'd strongly shy away from any custom Date type.
The cleanest choice might be to extend the widget to include time (and default to zero time as in the pull), and adopt the ecosystem time standard. I found this pull looking for a date time picker widget!

JoshLambda:
Chrono API documentation [recommends](https://docs.rs/chrono/latest/chrono/struct.Date.html) using `NaiveDate` instead of `Date`, which would make sense here since the datepicker widget is time zone agnostic.

PR #1900, opened 2022-08-10
Add CI for android

Zoxc:
This just runs `cargo check` on `eframe` for an Android target. It doesn't currently pass due to [`arboard` intentionally not building on unsupported platforms](https://github.com/1Password/arboard/pull/50). Not quite sure where that should be resolved.


emilk:
Let's add `[target.'cfg(not(target_os = "android"))'.dependencies]` before the `arboard` dependency in `egui-winit/Cargo.toml`

Zoxc:
I did the `arboard` fix and added some fixes to `eframe` too.

emilk:
This is great; thank you!

PR #1917, opened 2022-08-15
add rows function for Ui , split an Ui into several rows.

shuibing811:


emilk:
I know there is a `ui.columns` and this may seem natural, but have you tried out `egui_extras::StripBuilder`? It is a more powerful API, and I am considering deprecating `ui.columns` in favor of it.

Issue #1918, opened 2022-08-15
eframe: create multiple windows in sequence


emilk:
Since https://github.com/emilk/egui/pull/1889 we can now open an `eframe` window, close it, and keep running the application.

What we want to do is to then be able to open another window and run that for a while, etc.

Thanks to https://github.com/emilk/egui/pull/1919 this now works on Windows, but this does not yet work on Mac, due to https://github.com/rust-windowing/winit/issues/2431. Linux is unknown.

emilk:
Blocked on https://github.com/rust-windowing/winit/issues/2431

winne42:
>  Linux is unknown.
Just ran `examples/serial_windows` on Ubuntu 22.10/Gnome. The second window opens up and then the program crashes immediately. That sounds like the behavior on Mac as described in rust-windowing/winit#2431.

emilk:
https://github.com/emilk/egui/pull/3649 does not help on Mac :(

strike-digital:
Hi, I'm having some issues with this:

https://github.com/emilk/egui/assets/59890307/d7ea2448-8eb1-4bd0-9b11-9f15b2906757

I've seen that closing the window while continuing program execution is supported, but for me, the window will freeze until either a new window is closed or the program exits.
This is with `eframe = "0.26.2"`, and using `run_and_return: true` in the native options, on Windows 11.

Any workarounds, or fixes? Thanks!

strike-digital:
Hi, has there been an update on this? I see that this has been marked as completed, but I've updated my projects to 0.29.1 and the behaviour I showed above is still happening.

Even when trying the official `serial_windows` example, this problem still occurs for me.

I assume this isn't the intended behaviour, but I don't know if this is experienced by others, I'm on Windows 11 if that helps.

Thanks for your time!

lucasmerlin:
Just tested this on MacOS and Windows and can confirm that this doesn't work as expected (instead of disappearing, the window hangs until the next window is opened)

qrnch-jan:
What's the status on this? Is this supposed to be fixed in 0.31.0?  I've run the serial_windows example on macos, and it kind-of, sort-of seems to work?

Basically it does what it's advertised to do, but I can provoke it to misbehave (it doesn't open up the next window) if I try to interact with the window[*] after pressing the `Close` button.  But if I press the `Close` button, then move the mouse pointer away from the window and let it timeout in peace, then it closes the window and opens the next one.


[...]

PR #1919, opened 2022-08-15
eframe: several windows in series

emilk:
Part of https://github.com/emilk/egui/issues/1918

This attempts to allow opening several eframe windows after each other (one at the time).

Try it out with `cargo run -p serial_windows`.

This works on Windows (thanks to @zu1k), but on Mac I get `winit::event::Event::LoopDestroyed` right away for the second window, and then the even loops closes.

### DIFF
--- a/eframe/src/native/run.rs
+++ b/eframe/src/native/run.rs
@@ -152,6 +152,13 @@ fn run_and_return(event_loop: &mut EventLoop<RequestRepaintEvent>, mut winit_app
     tracing::debug!("eframe window closed");
 
     winit_app.save_and_destroy();
+
+    drop(winit_app);
+
+    // Needed to clean the event_loop:
+    event_loop.run_return(|_, _, control_flow| {
+        control_flow.set_exit();
+    });
 }
 
 fn run_and_exit(
--- a/eframe/src/native/run.rs
+++ b/eframe/src/native/run.rs
@@ -113,6 +113,14 @@ fn run_and_return(event_loop: &mut EventLoop<RequestRepaintEvent>, mut winit_app
                 ..
             }) => EventResult::RepaintAsap,
 
+            winit::event::Event::WindowEvent { window_id, .. }
+                if window_id != winit_app.window().id() =>
+            {
+                // This can happen if we close a window, and then reopen a new one,
+                // or if we have multiple windows open.
+                EventResult::Wait
+            }
+
             event => winit_app.on_event(event),
         };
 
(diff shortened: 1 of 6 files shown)


zu1k:
I add the following code to the start of GlowWinitApp.on_event() func, and found the second and third window receive some events which are not belong to them.

```rust
 if window_id != self.window().id() {
    println!("{event:?}");
    println!("window_id mismatch !");
    return EventResult::Wait;
}
```

```
# Running `target\debug\serial_windows.exe`
ModifiersChanged((empty))
window_id mismatch !
Focused(false)
window_id mismatch !
Destroyed
window_id mismatch !
ModifiersChanged((empty))
window_id mismatch !
Focused(false)
window_id mismatch !
```

eframe/src/native/run.rs
```diff
@@ -424,12 +437,15 @@ mod glow_integration {
...
-        let event_loop = glutin::event_loop::EventLoopBuilder::with_user_event().build();
-        let glow_eframe = GlowWinitApp::new(&event_loop, app_name, native_options, app_creator);
-
         if native_options.run_and_return {
-            run_and_return(event_loop, glow_eframe);
+            with_event_loop(|event_loop| {
+                let glow_eframe =
+                    GlowWinitApp::new(event_loop, app_name, native_options, app_creator);
+                run_and_return(event_loop, glow_eframe);
+            });
```
zu1k, line 445:
```suggestion
            with_event_loop(|event_loop| {
                {
                    let glow_eframe =
                        GlowWinitApp::new(event_loop, app_name, native_options, app_creator);
                    run_and_return(event_loop, glow_eframe);
                }

                // rerun to clean the event_loop
                use winit::platform::run_return::EventLoopExtRunReturn as _;
                event_loop.run_return(|_, _, control_flow| {
                    control_flow.set_exit();
                });
            });
```

zu1k: We should rerun to clean the event_loop after GlowWinitApp dropped.

zu1k: This works on my Windows 10 PC.

emilk: I pushed it, but it doesn't help on my Mac :(

emilk:
Ignoring events to the wrong window seems like a good idea, so I added that. The extra call to `run_return` does nothing on my Mac.

Does it now work as expected on Windows?

zu1k:
> Does it now work as expected on Windows?

Yes, it works.

https://user-images.githubusercontent.com/42370281/184650345-16c9b79b-dd99-4363-b2ff-352f4ea35bff.mp4

emilk:
Well, that's nice! On Mac I get `winit::event::Event::LoopDestroyed` right away, and then the event loop closes. Well, perhaps I can merge this as at least a step in the right direction.

Issue #1744, opened 2022-06-16
Artifacts rendering on the web using three-d


Yeicor:
**Describe the bug**
Using a custom glow context for web is buggy (see demo gif). This happens for both desktop and mobile browsers (see versions below), so I think it is a web backend + custom glow code bug.

![demo](https://user-images.githubusercontent.com/4929005/174131468-2bee6136-a3ae-4703-8d4f-fbb07f0f57d1.gif)

**To Reproduce**
1. Download the latest source code (7eeb292adfacd9311a420ac3ea225e2261a8f8d3).
2. Apply this patch to be able to run the example for web: [custom_3d_three_d_web.diff.txt](https://github.com/emilk/egui/files/8921112/custom_3d_three_d_web.diff.txt)
3. From `examples/custom_3d_three-d/`, execute `wasm-pack build --target web --out-dir ./target/pkg` and `cp index.html target/pkg` to prepare the static site.
4. Serve the files at `target/pkg`, for example with `python -m http.server`, and open the directory in any browser

**Expected behavior**
I expected the triangle to render exactly like the desktop version on the same machine:

![demo_14](https://user-images.githubusercontent.com/4929005/174131797-5ce97733-1699-4cca-8faa-21d6ca39fc74.gif)

**Desktop**
 - OS: Manjaro Linux
 - Browser: tested on Google Chrome 101.0.4951.54 (Official Build) (64-bit) and Firefox 101.0.1 (64-bit) with the same results.

**Mobile**
 - Device: Poco F3
 - OS: Android 12
 - Browser: tested on Google Chrome 102.0.5005.78 and Firefox 101.2.0 with the same results.

Yeicor:
FWIW, custom glow code works on the web if instead of rendering to a widget I render directly to the initial glow context. [Demo](https://yeicor.github.io/sdf-viewer/) with [these](https://github.com/Yeicor/sdf-viewer/blob/v0.0.15/src/app/mod.rs#L61) [sources](https://github.com/Yeicor/sdf-viewer/blob/v0.0.15/src/app/scene/mod.rs#L136).

Yeicor:
This bug seems partially related to rendering the background of Frame::canvas for web, as setting `Frame::canvas(ui.style()).fill(Color32::from_rgba_unmultiplied(0, 0, 0, 0))` almost solves it.

[...]

PR #1923, opened 2022-08-16
Improve custom_3d_three-d example

asny:
Improve the `custom_3d_three-d` example by:

- Render the `three-d` stuff into the correct frame buffer (see #1744 and a827c3e0333438c1f6afa8f709c66bbdb79b1c65 )
- Update to latest version of `three-d`
- Made it easy to build the example to web
- Only create the data (camera and model) once at startup and not every frame which is what you usually want
- Restructured the example into structs to separate the functionality (`egui`, `three-d` and the interface between them)


emilk (approved):
Awesome, thanks!

PR #1938, opened 2022-08-19
Use Trunk to build and deploy demo web app

### LINKED ISSUE #1937: Use Trunk to build and deploy demo web app
@coderedart did an excellent job converting `eframe_template` to using [Trunk](https://trunkrs.dev/) in https://github.com/emilk/eframe_template/pull/69/files and it would be great to use Trunk for the egui demo app (egui.rs) too. That way we can deploy a new web app on each commit to `master`. We can also get rid of a lot of annoying scripts.

emilk:
Closes https://github.com/emilk/egui/issues/1937

Copied a lot from what @coderedart  did in https://github.com/emilk/eframe_template/pull/69

One remaining issue is that trunk generates the code that starts egui. This means that `assets/multiple_apps.html` no longer works. I believe this is blocked on https://github.com/thedodd/trunk/issues/298


emilk:
Closing this stale draft. Still tracked in #1937.

PR #1940, opened 2022-08-20
Move all crates into a `crates` directory

emilk:
This is just a lot cleaner and easier to navigate.


iitalics:
FYI this breaks a bunch of links in the README afaict

emilk:
Thanks @iitalics I will fix asap!

PR #1936, opened 2022-08-19
egui_extras: Add Table::vertical_scroll_offset

njust:
This PR adds the ability to scroll the table to a specific vertical scroll offset. Useful if items are not yet rendered, and therefore it cannot be scrolled to them via Response::scroll_to_me.

PR #1946, opened 2022-08-20
egui_extras: Add Table::vertical_scroll_offset

njust:
Sorry, I messed up PR #1936, so I created a new one. This PR adds the ability to scroll the table to a specific vertical scroll offset. Useful if items are not yet rendered, and therefore it cannot be scrolled to them via Response::scroll_to_me.

### DIFF
--- a/crates/egui_extras/CHANGELOG.md
+++ b/crates/egui_extras/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to the `egui_extras` integration will be noted in this file.
 
 
 ## Unreleased
-* Added `TableBuilder::vertical_scroll_offset`: method to set vertical scroll offset position for a table.
+* Added `TableBuilder::vertical_scroll_offset`: method to set vertical scroll offset position for a table ([#1946](https://github.com/emilk/egui/pull/1946)).
 
 ## 0.19.0 - 2022-08-20
 * MSRV (Minimum Supported Rust Version) is now `1.61.0` ([#1846](https://github.com/emilk/egui/pull/1846)).
--- a/crates/egui_extras/CHANGELOG.md
+++ b/crates/egui_extras/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to the `egui_extras` integration will be noted in this file.
 
 
 ## Unreleased
-
+* Added `TableBuilder::vertical_scroll_offset`: method to set vertical scroll offset position for a table.
 
 ## 0.19.0 - 2022-08-20
 * MSRV (Minimum Supported Rust Version) is now `1.61.0` ([#1846](https://github.com/emilk/egui/pull/1846)).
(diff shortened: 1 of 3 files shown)


RobWalt:
Unknowingly I just worked on the same feature 😅 

I wrote a little example [here](https://github.com/RobWalt/egui/blob/master/examples/table/src/main.rs). As I saw your PR, I wanted to delete my changes, but I can't get it to work like my code with your changes. Can you maybe explain how to use your code with an example?

I want to open the app and immediately scroll to a position in the table with one jump.

crates/egui_extras/src/table.rs
```diff
@@ -288,17 +301,22 @@ impl<'a> Table<'a> {
...
 
         let mut new_widths = widths.clone();
 
-        let scroll_area = egui::ScrollArea::new([false, scroll])
+        let mut scroll_area = egui::ScrollArea::new([false, scroll])
             .auto_shrink([true; 2])
             .stick_to_bottom(stick_to_bottom);
 
+        if let Some(scroll_offset_y) = scroll_offset_y {
+            scroll_area = scroll_area.vertical_scroll_offset(scroll_offset_y);
```
RobWalt, line 317: Just an idea: We could use `scroll_offset` using a `Vec2` and set the `x` component to `0` to be more future proof (if horizontal scroll will be available some day).

njust:
> Unknowingly I just worked on the same feature sweat_smile
> 
> I wrote a little example [here](https://github.com/RobWalt/egui/blob/master/examples/table/src/main.rs). As I saw your PR, I wanted to delete my changes, but I can't get it to work like my code with your changes. Can you maybe explain how to use your code with an example?
> 
> I want to open the app and immediately scroll to a position in the table with one jump.

I extended the table demo with many homogenous rows, to be able to scroll to a specific row. It also works with an initial scroll offset.

crates/egui_extras/src/table.rs
njust: I wouldn't. None means don't change the scroll position, whereas 0 means scroll to 0.

crates/egui_extras/CHANGELOG.md
```diff
@@ -3,7 +3,7 @@ All notable changes to the `egui_extras` integration will be noted in this file.
 
 
 ## Unreleased
-
+* Added `TableBuilder::vertical_scroll_offset`: method to set vertical scroll offset position for a table.
```
emilk, line 6:
```suggestion
* Added `TableBuilder::vertical_scroll_offset`: method to set vertical scroll offset position for a table ([#1946](https://github.com/emilk/egui/pull/1946)).
```

emilk (approved):
thanks!

PerryCameron:
Any plans to make the column headers sortable in the same manner that TableView in JavaFX does? For example, if you click on a column head it sorts the whole table to that column and has an arrow that points up. If you click on the column head again it sorts the entire table in reverse order of that column and then the arrow points down. To me this is the most important feature of a table view style widget.

247gzs:
@PerryCameron 
Hi, how did you deal with this problem?

Issue #1949, opened 2022-08-21
Panic when running native custom_3d_three-d


dbuch:
Egui crash when running example with three-d

**Describe the bug**
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: NoAvailablePixelFormat', crates/eframe/src/native/run.rs:45:14

**To Reproduce**
1. RUST_BACKTRACE=1 cargo run --release -p custom_3d_three-d

**Expected behavior**
The app runs without crash also on semi-old integrate gpu

**Stack backtrace**
```
   0: rust_begin_unwind
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/panicking.rs:142:14
   2: core::result::unwrap_failed
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/result.rs:1805:5
   3: eframe::native::run::create_display
   4: eframe::native::run::glow_integration::GlowWinitApp::new
   5: std::thread::local::LocalKey<T>::with
   6: eframe::native::run::glow_integration::run_glow
   7: eframe::run_native
   8: custom_3d_three_d::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

**Desktop (please complete the following information):**
 - OS: Linux

**Additional context**
I believe it's caused by a 32-bit depth buffer setting. On my machine it works if i drop to 24-bit.

PR #1950, opened 2022-08-21
Example: three-d downgrade depth_buffer to 24 bit

dbuch:
Closes: #1949 

This works on my pc. I'm not sure if there's another and better fix.


emilk (approved):
thanks. yeah, 24 is a much better value

PR #1952, opened 2022-08-21
Android support for EFrame

### LINKED ISSUE #1951: Android support for EFrame, including support for suspend + resume events
**Describe the bug**
As a continuation of the work started with https://github.com/emilk/egui/pull/1634 it would also be good to update EFrame to be able to defer the initialization of render state until the app is first resumed (and be able to re-create windows/surfaces for further Suspend/Resumed events) so that EFrame can be used to build Android applications.

Right now if you try and run an EFrame based application on Android you will simply get a black screen.

**To Reproduce**

The Android example here can potentially be used to reproduce this: https://github.com/rib/android-activity/tree/eframe/examples/agdk-eframe - though there's a chicken / egg situation since that test has actually been updated to work with [a branch](https://github.com/rib/egui/commits/eframe-android) I will submit as a PR soon.

**Expected behavior**

It should be possible to run an eframe application on Android with input handling and correct rendering, as well as be able to suspend and resume when switching between applications.

rib:
This continues the work I started in https://github.com/emilk/egui/pull/1634 by updating eframe to also defer how it initializes graphics state, which ensures that on Android we wait until the application has an associated `SurfaceView`.

The first commit (which affects the `NativeOption` API) is needed for me to use the [android-activity](https://github.com/rib/android-activity) Android glue crate with Egui which was a pre-requisite for being able to create an Android test/example using eframe. Notably; the aim here was to create a general mechanism that allows customizing the event loop so it's not necessary for Egui/EFrame to have a direct dependency on the `android-activity` glue crate (since Winit will generally handle the Android platform integration).

The minimal example [here](https://github.com/rib/android-activity/tree/eframe/examples/agdk-eframe) can be used to test this - and that will hopefully be updated soon to use demo lib.

The main issue hit while making this change was realizing that it was only really possible to support Android fully with Wgpu currently, not with Glow/Glutin.

The difficulty with Glutin is that it currently tightly couples a GL context with a window which means you can't destroy and recreate just the window surface when an Android app is suspended - you're forced to also destroy the full GL context, at which point you're basically forced to tear everything down. There is work in progress to improve the Glutin API so suspend and resume will be possible to support in the future, but I'm not sure what the eta is currently for [this work](https://github.com/rust-windowing/glutin/pull/1435).

For now though I have still updated the Glow integration in-line with the Wgpu changes - so it would be possible to launch an Glow based application on Android and it should work until the application is suspended - which is at least an incremental improvement and hopefully it will just be another incremental change to support suspend and resume once the `Surface` API lands for Glutin.

Closes #1951

### DIFF
--- a/crates/eframe/CHANGELOG.md
+++ b/crates/eframe/CHANGELOG.md
@@ -5,7 +5,7 @@ NOTE: [`egui-winit`](../egui-winit/CHANGELOG.md), [`egui_glium`](../egui_glium/C
 
 
 ## Unreleased
-
+* Added `NativeOptions::event_loop_builder` hook for apps to change platform specific event loop options ([#1952](https://github.com/emilk/egui/pull/1952)).
 
 ## 0.19.0 - 2022-08-20
 * MSRV (Minimum Supported Rust Version) is now `1.61.0` ([#1846](https://github.com/emilk/egui/pull/1846)).
--- a/crates/eframe/CHANGELOG.md
+++ b/crates/eframe/CHANGELOG.md
@@ -6,6 +6,7 @@ NOTE: [`egui-winit`](../egui-winit/CHANGELOG.md), [`egui_glium`](../egui_glium/C
 
 ## Unreleased
 * Added `NativeOptions::event_loop_builder` hook for apps to change platform specific event loop options ([#1952](https://github.com/emilk/egui/pull/1952)).
+* Enabled deferred render state initialization to support Android ([#1952](https://github.com/emilk/egui/pull/1952)).
 
 ## 0.19.0 - 2022-08-20
 * MSRV (Minimum Supported Rust Version) is now `1.61.0` ([#1846](https://github.com/emilk/egui/pull/1846)).
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -5,52 +5,15 @@ use std::time::Duration;
 use std::time::Instant;
 
 use egui_winit::winit;
-use winit::event_loop::{ControlFlow, EventLoop, EventLoopBuilder};
+use winit::event_loop::{
+    ControlFlow, EventLoop, EventLoopBuilder, EventLoopProxy, EventLoopWindowTarget,
+};
 
 use super::epi_integration::{self, EpiIntegration};
 use crate::epi;
 
 #[derive(Debug)]
 pub struct RequestRepaintEvent;
-
-#[cfg(feature = "glow")]
-#[allow(unsafe_code)]
-fn create_display(
-    native_options: &NativeOptions,
-    window_builder: winit::window::WindowBuilder,
-    event_loop: &EventLoop<RequestRepaintEvent>,
-) -> (
-    glutin::WindowedContext<glutin::PossiblyCurrent>,
-    glow::Context,
-) {
-    crate::profile_function!();
-
-    use crate::HardwareAcceleration;
-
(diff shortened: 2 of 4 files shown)


crates/eframe/CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ NOTE: [`egui-winit`](../egui-winit/CHANGELOG.md), [`egui_glium`](../egui_glium/C
 
 
 ## Unreleased
-
+* Added `NativeOptions::event_loop_builder` hook for apps to change platform specific event loop options.
+* Enabled deferred render state initialization to support Android.
```
emilk, line 9:
```suggestion
* Added `NativeOptions::event_loop_builder` hook for apps to change platform specific event loop options ([#1952](https://github.com/emilk/egui/pull/1952)).
* Enabled deferred render state initialization to support Android ([#1952](https://github.com/emilk/egui/pull/1952)).
```

crates/eframe/src/native/run.rs
```diff
@@ -65,23 +28,53 @@ enum EventResult {
...
+        // do that as part of the lazy thread local storage initialization and so we instead
+        // create the event loop lazily here
+
+        let mut event_loop = event_loop.borrow_mut();
+        if event_loop.is_none() {
+            event_loop.replace(create_event_loop_builder(&mut native_options).build());
+        }
+        if let Some(event_loop) = event_loop.as_mut() {
+            f(event_loop, native_options);
+        }
```
emilk, line 77: This can be simplified by using [`Option::get_or_insert_with`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.get_or_insert_with).

emilk (changes requested):
It complicates things a bit, but not more then necessary afaict. Thanks for working on this!

crates/eframe/src/native/run.rs
rib: :) I guess I have a blind spot for some combinator APIs - a vaguely recall you suggested something similar with my previous PR

rib:
> It complicates things a bit, but not more then necessary afaict. Thanks for working on this!

No worries, thanks for taking a look.  Yeah, it's a bit more involved, though hopefully not too unwieldy. There's perhaps also a chance that having a clearer line between app state, render state and window state could help with other things, besides Android, later too - maybe a stepping stone for supporting multiple windows with desktop window systems.

rib:
I just pushed a change to remove a redundant `#[allow(unused)]` above `create_event_loop_builder` (the original iteration of that code had some `#[cfg(target_os = "android")]` code that could lead to "unused" warnings, before adding the hook mechanism).

Zoxc:
I noticed that the destructor of the app is not called when the app window is closed on Android with this branch.

Zoxc:
I also noticed that this currently continuously redraws, which is not the best behavior on a mobile device.

rib:
Okey, thanks for experimenting with the branch @Zoxc. At least for the latter I'd figure it might be reasonable to tackle that as an orthogonal follow up issue (unless its affecting other platforms), and for the first one we should at least double check things are cleaning up correctly e.g for Windows - to check this doesnt regress somthing.

I have seen issues with wgpu resource cleanup with egui before on Windows (non-eframe) so would be curious if thats related to this deferred initialization pattern somehow. 

I might not get around to testing for some days I guess so if you manage to figue out any more details it would be great to hear what you find from experimenting @Zoxc

emilk (approved):

PR #1956, opened 2022-08-22
Only require something GLES 3 compatible from wgpu

Zoxc:
This allows the wgpu backend to be used on older devices and is required by the Android emulator. There should probably be a way to request higher limits for applications doing their own rendering however.


emilk:
tnx

Issue #1961, opened 2022-08-24
`egui_glium` doesn't use the `TextureFilter`


TomasKralCZ:
**Describe the bug**
`egui_glium` doesn't respect the `TextureFilter` and draws every texture with linear filtering.

**To Reproduce**
1. Create a texture with nearest filtering
2. Draw an image with the respective texture

**Expected behavior**
The texture should be sampled with nearest filtering.

PR #1962, opened 2022-08-24
`egui_glium` correct texture filtering

TomasKralCZ:
Closes <https://github.com/emilk/egui/issues/1961>.

I'm not sure if the name `UserTexture` is good... feel free to edit.


emilk (commented):
Looks good, but `UserTexture` is a bit confusing considering it is also used for the egui font texture. Perhaps `EguiTexture`?

Issue #1953, opened 2022-08-21
Number base specifiers for `DragValue` and `Slider`


Adanos020:
**Is your feature request related to a problem? Please describe.**

The newly added `custom_formatter` allows for, among other things, displaying numbers in `DragValue` and `Slider` in number systems other than decimal. What I didn't consider while adding this feature is that if you type a number into those widgets on keyboard, the input is still always parsed as a decimal number.

**Describe the solution you'd like**

First of all, get rid of `custom_formatter` since having separate methods for the input and output allows to specify different number bases on both ends, which can be confusing and difficult to debug.

Add an enum called `NumberBase`.
```rust
#[derive(Copy, Clone, Debug)]
pub enum NumberBase {
    Binary,
    Octal,
    Decimal,
    Hexadecimal,
}
```

Add new fields and methods for the `DragValue` and `Slider` structs.
```rust
/// Sets the number base in which the numeric input is parsed from and displayed to the widget.
pub fn number_base(mut self, base: NumberBase) -> Self {
    self.number_base = base;
    self
}

/// Sets the minimum number of digits displayed. If the number is shorter than this, it is prefixed with a required number of zeroes.
pub fn number_width(mut self, zeroes: u32) -> Self {
    self.number_width = width;
    self
}
```

**Describe alternatives you've considered**

1. Instead of an enum, the number base could be specified as a `u8`, allowing for more number systems. However, Rust's standard library doesn't provide formatting numbers with an arbitrary radix, so we would need either our own implementation for that or an additional dependency such as the [radix_fmt](https://crates.io/crates/radix_fmt) crate.
2. The deletion of `custom_formatter` comes with two drawbacks: the API breaks and the number of ways a number can be formatted becomes limited. Instead of completely replacing it with a whole new API, we could only have an additional method specifying the number base for the input. However, as I mentioned earlier, that would allow for having the inconsistency where the e.g. number would be displayed in binary but parsed in hexadecimal on input.

**Additional context**

--

emilk:
I wouldn't like to get rid of `custom_formatter` since it supports many interesting scenarios such as displaying time in `HH::MM::SS` format, etc.

Instead it can be coupled with a `custom_parser`. A `NumberBase` API could then be built on top of those two features.

Adanos020:
I see. This sounds like a good solution 👍

PR #1851, opened 2022-07-24
Add methods for custom number formatting in `DragValue` and `Slider`

Adanos020:
Closes <https://github.com/emilk/egui/issues/1221>.

As the title says, this PR adds a `custom_formatter` method for both `DragValue` and `Slider`. A custom formatter takes a `f64` for the numeric value and a `RangeInclusive<usize>` representing the decimal range.

The following code:

```rs
ui.add(egui::Slider::new(&mut self.age, 0..=120)
    .prefix("0x")
    .custom_formatter(|n, _| format!("{:X}", n as i64)));
ui.add(egui::DragValue::new(&mut self.age)
    .custom_formatter(|n, _| format!("{:03}", n as i64)));
```

produces the following result:

image

PR #1967, opened 2022-08-26
Add custom parser to `DragValue` and `Slider`, with additional support for binary, octal, and hexadecimal numbers.

Adanos020:
Closes <https://github.com/emilk/egui/issues/1953>.

Adds a `custom_parser` method to `DragValue` and `Slider`, along with extra `binary_u64`, `octal_u64`, and `hexadecimal_u64` which overwrite the `custom_formatter` and `custom_parser`.

The purpose of `custom_parser` is to convert the text input from both widgets to a `f64`. A parser function returns an `Option<f64>` set to `Some` when parsing succeeds or `None` when it fails.

Additional methods for different number systems all take in a `min_width` parameter that specifies the minimum number of displayed digits. If the number is shorter than `min_width`, it will be prefixed with additional 0s to match it. `hexadecimal_u64` also takes in a `bool` which specifies if in the output text alphabetic digits should be uppercase or not.

Demo (apologies for quality):
untitled

Code for the demo:
```rust
ui.add(
    egui::Slider::new(&mut self.age, 0..=1023)
        .text("bin")
        .binary_u64(10),
);
ui.add(
    egui::Slider::new(&mut self.age, 0..=1023)
        .text("oct")
        .octal_u64(4),
);
ui.add(
    egui::Slider::new(&mut self.age, 0..=1023)
        .text("hex")
        .hexadecimal_u64(3, false),
);
ui.add(
    egui::Slider::new(&mut self.age, 0..=1023)
        .text("HEX")
        .hexadecimal_u64(3, true),
);
ui.add(
    egui::Slider::new(&mut self.age, 0..=((60 * 60 * 24) - 1))
        .text("time")
        .custom_formatter(|n, _| {
            let n = n as i32;
            let hours = n / (60 * 60);
            let mins = (n / 60) % 60;
            let secs = n % 60;
            format!("{hours:02}:{mins:02}:{secs:02}")
        })
        .custom_parser(|s| {
            let parts: Vec<&str> = s.split(':').collect();
            if parts.len() == 3 {
                parts[0]
                    .parse::<i32>()
                    .and_then(|h| {
                        parts[1].parse::<i32>().and_then(|m| {
                            parts[2]
                                .parse::<i32>()
                                .map(|s| ((h * 60 * 60) + (m * 60) + s) as f64)
                        })
                    })
                    .ok()
            } else {
                None
            }
        }),
);
```

### DIFF
--- a/crates/egui/src/widgets/drag_value.rs
+++ b/crates/egui/src/widgets/drag_value.rs
@@ -240,11 +240,16 @@ impl<'a> DragValue<'a> {
         self
     }
 
-    /// Set `custom_formatter` and `custom_parser` to display and parse numbers in binary.
-    /// Signed integers and floating point numbers are not supported.
+    /// Set `custom_formatter` and `custom_parser` to display and parse numbers as binary signed integers. Floating
+    /// point numbers are *not* supported.
     ///
     /// `min_width` specifies the minimum number of displayed digits; if the number is shorter than this, it will be
     /// prefixed with additional 0s to match `min_width`.
+    ///
+    /// If `twos_complement` is true, negative values will be displayed as the 2's complement representation. Otherwise
+    /// they will be prefixed with a '-' sign.
+    ///
+    /// See also: [`binary_uint`]
     ///
     /// # Panics
     ///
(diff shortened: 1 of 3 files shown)


Mingun:
May be return a `Result<T, String>` where the `Err` variant will represent a text that can be shown to the user why his input is incorrect? For now egui can just use `.ok()` and not shown anything, but in the future it could.

Adanos020:
> May be return a `Result<T, String>` where the `Err` variant will represent a text that can be shown to the user why his input is incorrect? For now egui can just use `.ok()` and not shown anything, but in the future it could.

If we do want this, I think `Result<T, Option<String>>` would be better as it'd give you the option not to have an error message.

crates/egui/src/widgets/drag_value.rs
```diff
@@ -170,6 +198,170 @@ impl<'a> DragValue<'a> {
...
+                    return None;
+                }
+                let num_digits = s.len().min(64);
+                let mut n = 0;
+                for (i, digit) in s.chars().rev().take(num_digits).enumerate() {
+                    if digit != '0' && digit != '1' {
+                        return None;
+                    }
+                    n |= (digit as u64 - '0' as u64) << (i as u64);
+                }
```
emilk, line 276: Use https://doc.rust-lang.org/stable/std/primitive.i64.html#method.from_str_radix here and below - less code, and automatic support for signed integers!

emilk (changes requested):
nice!

emilk:
> May be return a `Result<T, String>` where the `Err` variant will represent a text that can be shown to the user why his input is incorrect? For now egui can just use `.ok()` and not shown anything, but in the future it could.

Let's wait with that until we actually implement showing the errors (if ever)

crates/egui/src/widgets/drag_value.rs
Adanos020: I was thinking about using that but wasn't sure if we wanted another dependency. I'll add it now!

Adanos020: Oh cool, I don't know how I missed this!

Adanos020: @emilk According to [`Binary`]() documentation:

> For primitive signed integers ([i8](https://doc.rust-lang.org/stable/std/primitive.i8.html) to [i128](https://doc.rust-lang.org/stable/std/primitive.i128.html), and [isize](https://doc.rust-lang.org/stable/std/primitive.isize.html)), negative values are formatted as the two’s complement representation.

Are we happy with this?

Adanos020: I added a bool that lets you choose if you want the number to be displayed with a sign or in 2's complement.

Adanos020:
Applied suggested changes, also added a `twos_complement` flag to give the choice whether you want to display negative numbers with a `-` sign or in the 2's complement representation.

emilk (approved):

enomado:
Would be awesome to add thousands separator,  e.g. `1,000,000.0`

Adanos020:
> Would be awesome to add thousands separator, e.g. `1,000,000.0`

@enomado Right now you can do it with `custom_formatter` I added in #1851, it's available in v0.19 :smile:
```rust
use num_format::{Locale, ToFormattedString};
ui.add(DragValue::new(&mut n).custom_formatter(|n, _| n.to_formatted_string(&Locale::en)));
```

PR #1968, opened 2022-08-27
Fix bit more code that could lead to a possible deadlock.

MaximOsipenko:
Continuation of #1380 story. Fixed bit more code that can possibly lead to double-read caused deadlocks.
 
Some of this deadlocks were found with test panels and gdb, others are found by reading during more global solution research on how to avoid same deadlocks in the future.

First deadlock was found in attempt to integrate egui with last amethyst version, when tried to draw panels in parallel from ecs systems.


emilk (approved):

emilk:
thanks!

PR #1971, opened 2022-08-27
egui_winit: support winit with default features disabled

koute:
Currently if you include `egui_winit` in your crate graph it forces all of the default `winit` features. This PR makes it so that if you depend on `egui_winit` with `default-features = false` then it will also propagate that to the `winit` crate, letting you chose the features you want yourself.


emilk:
The winit features: https://github.com/rust-windowing/winit/blob/master/Cargo.toml#L36-L43
Some docs for them: https://github.com/rust-windowing/winit#cargo-features

So it seems like the primary use for the winit features is to select between x11 or wayland, which indicates to me that we should have both a `wayland` and a `x11` feature in `egui-winit` too. Interestingly enough, winit has both x11 and wayland enabled by default.

koute:
What about other features of `winit`? For example, in my case I want only the following features enabled from `winit`: `x11`, `wayland`, `wayland-dlopen`. So with the current state of the PR I'll have to manually add `winit` to my deps and enable those features. If I do what you propose and also add a `x11` feature to `egui_winit` then I'll still have to include `winit` myself. 

Or in other words, there are three possibilities here:

1) We only add the bare minimum of feature flags necessary to `egui_winit` (only `wayland`). You have to manually add `winit` to your deps if you want to enable other `winit` flags.
2) We only add *some* of the feature flags from `winit` to `egui_winit` (only `x11` and `wayland`). You have to manually add `winit` to your deps if you want to enable other `winit` flags.
3) We add *all* of the feature flags from `winit` to `egui_winit`. Now you don't have to add `winit` at all to your deps, but `egui_winit` needs to keep parity with all of `winit`'s feature flags.

I think that if someone is particular enough to actually disable the default features there's probably a high chance that they do want to enable other flags, in which case either (1) or (3) makes the most sense to me.

But I really don't care which one it is, so please decide and I'll update the PR accordingly. (:

emilk:
Thanks for elaborating on the design space here @koute!

Here's the option that now seems the most attractive to me:
* `--no-default-features` to `egui-winit` provides no features to `winit` (users will have to opt-in to winit features)
* `default` features of `egui-winit` provides the default `winit` features
* There are no added features for `wayland`, `x11` etc added to `egui-winit`

This seems the most parsimonious approach, and shows no favoring of one specific winit feature (wayland) over any other. In other words - exactly what this PR does, but remove the `wayland` feature from `egui-winit` (and let users add that themselves directly to `winit`).

What do you think?

koute:
> This seems the most parsimonious approach, and shows no favoring of one specific winit feature (wayland) over any other. In other words - exactly what this PR does, but remove the `wayland` feature from `egui-winit` (and let users add that themselves directly to `winit`).
> 
> What do you think?

I would actually have done that originally, but the problem with it is that we have this piece of code in `egui-winit`:

```rust
/// Returns a Wayland display handle if the target is running Wayland
fn wayland_display<T>(_event_loop: &EventLoopWindowTarget<T>) -> Option<*mut c_void> {
    #[cfg(any(
        target_os = "linux",
        target_os = "dragonfly",
        target_os = "freebsd",
        target_os = "netbsd",
        target_os = "openbsd"
    ))]
    {
        return _event_loop.wayland_display();
    }

    #[allow(unreachable_code)]
    {
        let _ = _event_loop;
        None
    }
}
```

And the `wayland_display` in `winit` is defined like this:

```rust
    #[cfg(feature = "wayland")]
    fn wayland_display(&self) -> Option<*mut raw::c_void>;
```

So we can only access the `wayland_display` method if `winit` has the `wayland` feature enabled, so we *need* the `wayland` feature in `egui-winit`.

I guess the alternative here could be to make a PR to `winit` so that the `wayland_display` method is always defined but just returns `None` if the `wayland` feature is not enabled? Not sure if they'd accept it though.

emilk (approved):
Thanks for this, and for explaining it to me :)

PR #1986, opened 2022-08-30
Improve mixed CJK/Latin linebreaking.

bigfarts:
This avoids prioritizing kana above spaces and we will break at the first possible break location, rather than via an implicit order of breaking.

Before:
image
image

After:
image
image

### DIFF
--- a/crates/epaint/src/text/text_layout.rs
+++ b/crates/epaint/src/text/text_layout.rs
@@ -719,11 +719,11 @@ struct RowBreakCandidates {
     /// is always the primary candidate.
     space: Option<usize>,
 
-    /// Logograms (single character representing a whole word) are good candidates for line break.
-    logogram: Option<usize>,
+    /// Logograms (single character representing a whole word) or kana (Japanese hiragana and katakana) are good candidates for line break.
+    cjk: Option<usize>,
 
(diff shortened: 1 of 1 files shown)


emilk:
I have no knowledge of kana, but in English text I would say it is preferable to break at a space rather than at punctuation or dashes. Take for instance: `Temperature: 3.2 Kelvin". We do not want to break this as:

```
Temperature: 3.
2 Kelvin
```

So the current ordering is very deliberate when it comes to spaces, dashes and punctuation, and I don't want to break that.

If the problem is that spaces are prioritized over kana, let's just focus on that.

Perhaps something like:

``` rust
let best = self.space.or(self.logogram).or(self.dash).or(self.punctuation);
let pos = match (best, kana) {
    (None, None) => None,
    (None, Some(pos)) => Some(pos),
    (Some(pos), None) => Some(pos),
    (Some(best), Some(kana)) => Some(best.max(kana)),
};
pos.or(self.any)
```

Or we special-case it based on whether or not there is kana:

``` rust
if let Some(kana) = self.kana {
    // Whatever logic makes sense for kana
} else {
    self.space.or(self.logogram).or(self.dash).or(self.punctuation).or(self.any)
}
```


It would also be great if you added a test for this so we don't break it in the future!

emilk:
Great!

PR #2002, opened 2022-09-04
Add depth buffer support for egui-wgpu's render pass

amPerl:
`depth_buffer` in `NativeOptions` is currently completely unused when using eframe with the wgpu backend, making it a bit tedious to use render callbacks. This PR uses a 32-bit float depth buffer when `depth_buffer` is non-zero. It also conditionally raises the 2d texture size limit, which very quickly became an issue due to the low webgl2 compatibility defaults.

### DIFF
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -259,6 +259,10 @@ pub struct NativeOptions {
     /// Sets the number of bits in the depth buffer.
     ///
     /// `egui` doesn't need the depth buffer, so the default value is 0.
+    ///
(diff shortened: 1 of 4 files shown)


emilk (changes requested):
Looks good, but it's a bit unfortunate that `NativeOptions::depth_bits` will be interpreted differently on `wgpu` (though still better than not being used at all). We could have a separate wgpu_options later perhaps, but for now please just add a line to the docstring of `NativeOptions::depth_bits` explaining how it works on `wgpu` backends!

PR #2021, opened 2022-09-07
egui-wgpu renderer renaming

LU15W1R7H:
Adjust naming in egui-wgpu.
We want to avoid confusion with names like `RenderPass::execute_with_renderpass`.

- `RenderPass` -> `Renderer`
- `RenderPass::execute` -> `Renderer::render`
- `RenderPass::execute_with_renderpass` -> `Renderer::render_onto_renderpass`
- reexport `Renderer` in `lib.rs`


LU15W1R7H:
The CI failure wasn't caused by me?

crates/egui-wgpu/src/renderer.rs
```diff
@@ -323,8 +323,8 @@ impl RenderPass {
         self.depth_texture = Some((texture, view));
     }
 
-    /// Executes the egui render pass.
-    pub fn execute(
+    /// Executes the renderer on it's own render pass.
```
emilk, line 326:
```suggestion
    /// Executes the renderer on its own render pass.
```

PR #2028, opened 2022-09-08
Allow box zoom in any direction for plots

Zoxc:
Previously only the top-left to bottom-right direction worked.

### DIFF
--- a/crates/egui/src/widgets/plot/mod.rs
+++ b/crates/egui/src/widgets/plot/mod.rs
@@ -876,14 +876,18 @@ impl Plot {
                     let box_start_pos = transform.value_from_position(box_start_pos);
                     let box_end_pos = transform.value_from_position(box_end_pos);
                     let new_bounds = PlotBounds {
-                        min: [box_start_pos.x, box_end_pos.y],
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/plot/mod.rs
```diff
@@ -876,14 +876,18 @@ impl Plot {
...
+                        max: [
+                            box_start_pos.x.max(box_end_pos.x),
+                            box_start_pos.y.max(box_end_pos.y),
+                        ],
                     };
                     if new_bounds.is_valid() {
                         transform.set_bounds(new_bounds);
                         auto_bounds = false.into();
-                    } else {
-                        auto_bounds = true.into();
```
emilk, line 886: why this change?

emilk: If you don't know why it was there, maybe it is not a good idea to remove it? :)

Zoxc: I don't want box zoom to ever trigger auto bounds, we already have double clicking to reset the view.

PR #2029, opened 2022-09-08
Plot: Add `auto_bounds_x`, `auto_bounds_y` and `reset`

Zoxc:
This is based on https://github.com/emilk/egui/pull/2028.

This adds new `auto_bounds_x` and `auto_bounds_y` methods which allows auto bounds to be used in combination with `include_x` and `include_y`. The plot code is refactored to track when bounds are modified and it will use auto bounds and the values from `include_x` and `include_y` until then.

Double clicking is changed to reset the bounds to their initial value.

A new `reset` method is added which resets the plot state which is useful if you're loading a new dataset.

### DIFF
--- a/crates/egui/src/widgets/plot/mod.rs
+++ b/crates/egui/src/widgets/plot/mod.rs
@@ -73,7 +73,7 @@ impl Default for CoordinatesFormatter {
 const MIN_LINE_SPACING_IN_POINTS: f64 = 6.0; // TODO(emilk): large enough for a wide label
 
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
-#[derive(Clone)]
+#[derive(Copy, Clone)]
 struct AutoBounds {
     x: bool,
     y: bool,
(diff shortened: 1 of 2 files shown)


crates/egui/src/widgets/plot/mod.rs
```diff
@@ -73,36 +73,25 @@ impl Default for CoordinatesFormatter {
...
+        AxisBools { x: val, y: val }
     }
 }
 
 /// Information about the plot that has to persist between frames.
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 #[derive(Clone)]
 struct PlotMemory {
-    auto_bounds: AutoBounds,
+    modified: AxisBools,
```
emilk, line 92: Has _what_ been modified? Bounds? If so `bounds_modified` is a better name.

Zoxc: Yeah. Changed the name and added some documentation to the field.

emilk (approved):
thanks!

PR #2035, opened 2022-09-10
eframe: Add `center` to `NativeOptions` and `monitor_size` to `WindowInfo`

ar37-rs:
- Make window always on top and center positioning to be configured both at runtime and initialization.
-  Centering the window position at initialization it will fix the window being shown at different positions twice as if centering the window at runtime (which is doesn't seem right).

### DIFF
--- a/crates/eframe/CHANGELOG.md
+++ b/crates/eframe/CHANGELOG.md
@@ -10,7 +10,7 @@ NOTE: [`egui-winit`](../egui-winit/CHANGELOG.md), [`egui_glium`](../egui_glium/C
 * Allow empty textures with the glow renderer.
 * Added `shader_version` to `NativeOptions` for cross compiling support on different target OpenGL | ES versions (on native `glow` renderer only) ([#1993](https://github.com/emilk/egui/pull/1993)).
 * Fix: app state is now saved when user presses Cmd-Q on Mac ([#2013](https://github.com/emilk/egui/pull/2013)).
-* Added `center` to `NativeOptions`and `monitor_size` to `WindowInfo` on Desktop ([#2035](https://github.com/emilk/egui/pull/2035))
+* Added `center` to `NativeOptions` and `monitor_size` to `WindowInfo` on desktop ([#2035](https://github.com/emilk/egui/pull/2035)).
 
 ## 0.19.0 - 2022-08-20
 * MSRV (Minimum Supported Rust Version) is now `1.61.0` ([#1846](https://github.com/emilk/egui/pull/1846)).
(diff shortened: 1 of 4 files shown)


crates/eframe/src/epi.rs
```diff
@@ -698,6 +706,31 @@ impl Frame {
...
+    /// (Wayland desktop currently not supported)
+    #[cfg(not(target_arch = "wasm32"))]
+    pub fn set_centered(&mut self) {
+        if let Some(monitor_size) = self.info.window_info.monitor_size {
+            let inner_size = self.info.window_info.size;
+            let logical_one = 1.0;
+            if monitor_size.x > logical_one && monitor_size.y > logical_one {
+                let logical_two = 2.0;
+                let x = (monitor_size.x - inner_size.x) / logical_two;
+                let y = (monitor_size.y - inner_size.y) / logical_two;
```
emilk, line 728:
```suggestion
            if monitor_size.x > 1.0 && monitor_size.y > 1.0 {
                let x = (monitor_size.x - inner_size.x) / 2.0;
                let y = (monitor_size.y - inner_size.y) / 2.0
```

crates/eframe/src/native/run.rs
```diff
@@ -223,6 +223,27 @@ fn run_and_exit(
...
+fn centere_window_pos(
+    monitor: Option<winit::monitor::MonitorHandle>,
+    native_options: &mut epi::NativeOptions,
+) {
+    // Get the current_monitor.
+    if let Some(monitor) = monitor {
+        let monitor_size = monitor.size();
+        let inner_size = native_options
+            .initial_window_size
+            .unwrap_or(egui::Vec2 { x: 800.0, y: 600.0 });
```
emilk, line 235: This seems like a rather random assumption.

Perhaps `NativeOption::initial_window_size` shouldn't be an `Option`, and instead be a `Vec2` with a default of 800x600?

emilk (changes requested):

crates/eframe/src/native/run.rs
ar37-rs: @emilk yup, i also think so.

ar37-rs: @emilk  not sure about that though, it turns out breaking change on a lot of  `egui/examples` and there's also `min_window_size` /`max_window_size` on `NativeOptions` which is `Option<Vec2>`

crates/eframe/CHANGELOG.md
```diff
@@ -9,8 +9,8 @@ NOTE: [`egui-winit`](../egui-winit/CHANGELOG.md), [`egui_glium`](../egui_glium/C
 * Enabled deferred render state initialization to support Android ([#1952](https://github.com/emilk/egui/pull/1952)).
 * Allow empty textures with the glow renderer.
 * Added `shader_version` to `NativeOptions` for cross compiling support on different target OpenGL | ES versions (on native `glow` renderer only) ([#1993](https://github.com/emilk/egui/pull/1993)).
-* Fix: app state is now saved when user presses Cmd-Q on Mac ([#1993](https://github.com/emilk/egui/pull/1993)).
-
+* Fix: app state is now saved when user presses Cmd-Q on Mac ([#2013](https://github.com/emilk/egui/pull/2013)).
+* Added `center` to `NativeOptions`and `monitor_size` to `WindowInfo` on Desktop ([#2035](https://github.com/emilk/egui/pull/2035))
```
emilk, line 13:
```suggestion
* Added `center` to `NativeOptions` and `monitor_size` to `WindowInfo` on desktop ([#2035](https://github.com/emilk/egui/pull/2035)).
```

PR #2039, opened 2022-09-12
egui-wgpu: Fix crash on zero-sized scissor rects

### LINKED ISSUE #2038: wgpu crash if scissor rectangle is outside of the window
It looks like https://github.com/emilk/egui/commit/09d636b08943b14f182c25d65d25085959e7fc68 introduced the possibility of having a scissor rectangle that is outside of the screen. This leads to a wgpu panic:

```
ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
    In a RenderPass
      note: encoder = `encoder`
    In a set_scissor_rect command
    Scissor Rect { x: 1280, y: 720, w: 1, h: 1 } is not contained in the render target Extent3d { width: 1280, height: 720, depth_or_array_layers: 1 }
```

One way to produce this was to increase the `set_pixels_per_point` and then on the same frame placing an image in the bottom-right corner. But it also seems to happen in other scenarios.

emilk:
Closes #2038

Please confirm @s-nie

### DIFF
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -74,7 +74,13 @@ pub trait App {
     /// Can be used from web to interact or other external context
     /// Implementation is needed, because downcasting Box<dyn App> -> Box<dyn Any> to get &ConcreteApp is not simple in current rust.
     ///
-    /// Just return &mut *self
+    /// Just copy-paste this as your implementation:
+    /// ```ignore
+    /// #[cfg(target_arch = "wasm32")]
+    /// fn as_any_mut(&mut self) -> &mut dyn std::any::Any {
+    ///     &mut *self
+    /// }
+    /// ```
     #[cfg(target_arch = "wasm32")]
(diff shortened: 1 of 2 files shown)


crates/eframe/src/epi.rs
```diff
@@ -74,7 +74,13 @@ pub trait App {
...
     /// Implementation is needed, because downcasting Box<dyn App> -> Box<dyn Any> to get &ConcreteApp is not simple in current rust.
     ///
-    /// Just return &mut *self
+    /// Just copy-paste this as your implementation:
+    /// ```ignore
+    /// #[cfg(target_arch = "wasm32")]
+    /// fn as_any_mut(&mut self) -> &mut dyn std::any::Any {
+    ///     &mut *self
+    /// }
+    /// ```
```
Mingun, line 83: Maybe default implementation can be provided? Or it will return not right `Any`?

s-nie (approved):
That fixed it, thanks so much!

PR #2051, opened 2022-09-17
Change several methods from pub(crate) to pub

setzer22:
I'm working on custom widgets, and I realized most of the core widgets use the "auto_id" mechanism. I think having that available would let custom widget authors provide the same level of ergonomics as every other core widget that makes use of this feature.

The functionality is marked as `pub(crate)`, but I couldn't find any good reason for that, so this PR makes that be `pub` and available to everyone.

I found the same issue with `Button`'s `min_size`. I'm trying to implement my own `DragValue`, so I find myself in the need of using the same workaround that was used there and call the `min_size` method of the underlying button. There, too, I could not find any rationale for `pub(crate)`.

### DIFF
--- a/crates/egui/src/ui.rs
+++ b/crates/egui/src/ui.rs
@@ -586,11 +586,11 @@ impl Ui {
         self.id.with(&id_source)
     }
 
-    pub(crate) fn next_auto_id(&self) -> Id {
+    pub fn next_auto_id(&self) -> Id {
         Id::new(self.next_auto_id_source)
     }
 
-    pub(crate) fn auto_id_with<IdSource>(&self, id_source: IdSource) -> Id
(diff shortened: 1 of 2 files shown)


crates/egui/src/ui.rs
```diff
@@ -586,11 +586,11 @@ impl Ui {
         self.id.with(&id_source)
     }
 
-    pub(crate) fn next_auto_id(&self) -> Id {
+    pub fn next_auto_id(&self) -> Id {
```
emilk, line 589:
```suggestion
    /// This is the `Id` that will be assigned to the next widget added to this `Ui`.
    pub fn next_auto_id(&self) -> Id {
```

setzer22:
Thanks! :+1:

PR #2072, opened 2022-09-22
Add disable_srgb to NativeOptions.

ar37-rs:
VirtualBox_Ubuntu22 04_22_09_2022_16_42_45
Hi @emilk eframe broken again on virtualbox gl es 2.0 since #2070 (probably because it has GL_EXT_texture_sRGB_decode, but doesn't do anything).
adding `disable_srgb` will avoid srgb textures support misdetection.


emilk:
I want to merge https://github.com/emilk/egui/pull/2071 before merging this, but a few quick comments:

* Don't use negations in names. Call it `enable_srgb_textures: bool` instead
* I don't see a need for `Option<bool>`. Just use `bool`.

Is it possible to detect virtualbox some other way? i.e. via `gl.get_parameter_string(glow::SHADING_LANGUAGE_VERSION)` or similar?

ar37-rs:
@emilk exactly, agree on that. since OpenGL ES 2.0 driver usually has no `SRGB8_ALPHA8` support, what if we just disable srgb by default if the shader version detected is `egui_glow::ShaderVersion::Es100 `? that would be much more simpler.

ar37-rs:
Like so:
```rust
        let is_webgl_1 = shader_version == ShaderVersion::Es100;
        //...
        let srgb_textures = if is_webgl_1 { false } else { true };
        tracing::debug!("SRGB Support: {:?}", srgb_textures);
```

ar37-rs:
@emilk current master woking fine on virtualbox, thank you.

PR #2082, opened 2022-09-25
Add support to enable sRGB pixel format on NativeOptions.

ar37-rs:
fixes: #2085


emilk:
Does it perhaps work to just remove the `with_srgb(…)` line altogether instead?

ar37-rs:
@emilk i don't know, but the sRGB issue which causes crashing  reported by @enomado, for me it's working fine though on windows with intel graphics and vmsvga driver on virtualbox as well, with or without sRGB being enabled

PR #2087, opened 2022-09-26
fix: default to changing y axis when changing data aspect

### LINKED ISSUE #2077: Plot always zooms out when changing data aspect
This is a regression, I mentioned the original bug in this issue: #1174
and it was fixed in the linked PR: #1184

Seems like this PR broke it #1865 in `plot/mod.rs` line `723-728`.
If no axis is linked it will always expand.

https://github.com/emilk/egui/blob/29fa63317ea4f950a098be6240efc65c14b8c7a5/crates/egui/src/widgets/plot/mod.rs#L835
In the else case `change_x` should probably be defaulted to `false` as before.

Saecki:
Closes #2077.

@cessen @DusterTheFirst Can you guys check if I broke anything in this PR?

### DIFF
--- a/crates/egui/src/widgets/plot/mod.rs
+++ b/crates/egui/src/widgets/plot/mod.rs
@@ -79,6 +79,13 @@ struct AxisBools {
     y: bool,
 }
 
+impl AxisBools {
+    #[inline]
+    fn any(&self) -> bool {
+        self.x || self.y
+    }
+}
+
 impl From<bool> for AxisBools {
     fn from(val: bool) -> Self {
         AxisBools { x: val, y: val }
@@ -843,8 +850,10 @@ impl Plot {
             if let Some(linked_axes) = &linked_axes {
                 let change_x = linked_axes.link_y && !linked_axes.link_x;
                 transform.set_aspect_by_changing_axis(data_aspect as f64, change_x);
-            } else {
+            } else if auto_bounds.any() {
                 transform.set_aspect_by_expanding(data_aspect as f64);
+            } else {
+                transform.set_aspect_by_changing_axis(data_aspect as f64, false);
             }
         }
 


Saecki:
Ok so that definitely broke something

Saecki:
This feels a little bit inconsistent, but I've found it to actually work quite well, though other suggestions are welcome.
If `auto_bounds` is active, changing the `data_aspect` works as before.
If `auto_bounds` is inactive, changing the `data_aspect` only changes the y-axis.

[Screencast from 2022-09-26 17:59:43.webm](https://user-images.githubusercontent.com/43008152/192326519-8ca4cd43-5a6e-4961-8a91-90bb227d21d7.webm)

emilk:
There are now merge conflicts with https://github.com/emilk/egui/pull/2029, but it should be pretty trivial to fix if you merge in `master` and apply:

```diff
index a517db1e..86faad9a 100644
--- a/crates/egui/src/widgets/plot/mod.rs
+++ b/crates/egui/src/widgets/plot/mod.rs
@@ -79,6 +79,13 @@ struct AxisBools {
     y: bool,
 }
 
+impl AxisBools {
+    #[inline]
+    pub fn any(&self) -> bool {
+        self.x || self.y
+    }
+}
+
 impl From<bool> for AxisBools {
     fn from(val: bool) -> Self {
         AxisBools { x: val, y: val }
```

Saecki:
Rebased onto master and applied the patch

emilk (approved):

emilk:
thanks!

Zoxc:
Note that the `auto_bounds` variable is something entirely different after https://github.com/emilk/egui/pull/2029 and it doesn't seem to make sense with the use in this PR.

PR #2089, opened 2022-09-28
Bug fix: Allocate Frame outer margins

paulrouget:
Without that, adding a frame with an outer margin would not push the sibling content right or bottom.


emilk (approved):
Good catch!

PR #2102, opened 2022-10-02
Make private types related to ScrollArea public

Demiu:
The ScrollAreaOutput that show()ing a ScrollArea returns is a private. It cannot be used as a parameter, a return type or a type hint. It doesn't even have documentation page. Same with the State struct of the scroll_area module (Prepared remains private).


Demiu:
Already fixed

Issue #2000, opened 2022-09-04
Table row selection


DanielSWolf:
**Is your feature request related to a problem? Please describe.**

I'd like to write an egui application with a typical master-detail interface: On the left-hand side, there is a table of things; on the right-hand side, the thing(s) currently selected in the table are displayed with more detail. Think your typical e-mail client.

It is my understanding that the table component currently doesn't support selecting rows.

**Describe the solution you'd like**

Ideally, I'd like the following features:

* Choice between single-row selection (at most one row can be selected at any time) and multi-row selection (any number of rows can be selected at a time)
* Clicking on any row selects this row, unselecting any other rows
* Pressing the up or down cursor keys selects the single row above or below the current selection
* In multi-row mode, the following additional interactions are possible:
  * Ctrl-clicking allows adding individual rows to the selection
  * Shift-clicking allows selecting a range of rows
  * Ctrl+A selects all rows

terhechte:
In case you need it right now. There's a table Widget that supports selection in my Postsack app:
https://github.com/terhechte/postsack/blob/main/ps-gui/src/widgets/table.rs

I forked it from somewhere else and extended it. Maybe it suits your needs until the official table supports selection.

DanielSWolf:
@terhechte Thanks for the code! I'm currently writing the backend logic, so it'll be some time before I actually need the table. But it's good to know there's an option!

samitbasu:
I put a sketch of an implementation on PR for a selectable table.  It's probably (definitely) in need of work.  But I'm happy to move it forward, since I need this functionality also.

apoorv569:
Hi, I am also looking for a similar thing. I have a table from `egui_extras` and I want the ability to hover/select a row and potentially get the state that row is clicked and/or data from columns when clicked if needed.

I tried the link that @terhechte shared above, it uses very out of date version of `egui` so the code doesn't work.

PR #2105, opened 2022-10-03
Draft of an update to the Table widget to allow for selection.

samitbasu:
Hello!  I needed a selectable table for a project, and saw this [issue](https://github.com/emilk/egui/issues/2000).  The code same referenced in that issue no longer works with the latest versions of `egui`, so I modified the `Table` instead.  It definitely needs work!  But before going too far down the path, I thought I would at least float the basics since I'm still getting my feet wet in `egui`.

There are definitely some oddities and it feels unnatural.  For example, in the calling code, you need to know if the row is selected or not when adding the widget:

```rust
           .body(|mut body| {
                for row_index in 0..self.things.len() {
                    let selected_row = if let Some(index) = self.selected {
                        index == row_index
                    } else {
                        false
                    };
                    body.row_with_selection(text_height, selected_row, |mut row| {
                        let selected = row.selected();
                        if row
                            .col(|ui| {
                                if selected {
                                    ui.strong(row_index.to_string());
                                } else {
                                    ui.label(row_index.to_string());
                                }
                            })
                            .clicked()
                        {
                            self.selected = Some(row_index);
                        };
                        if row
                            .col(|ui| {
                                if selected {
                                    ui.strong(self.things.len().to_string());
                                } else {
                                    ui.label(self.things.len().to_string());
                                }
                            })
                            .clicked()
                        {
                            self.selected = Some(row_index);
                        };
                        if row
                            .col(|ui| {
                                if selected {
                                    ui.strong(&self.things[row_index]);
                                } else {
                                    ui.label(&self.things[row_index]);
                                }
                            })
                            .clicked()
                        {
                            self.selected = Some(row_index);
                        }
                    });
                }
            })

```

This seems both repetitive and verbose.  I guess it allows you to decide which part of the row you are using to determine clicks?  Not sure what a cleaner way to do this would be, but I'm happy to keep working on it if you think it's worth pursuing.

Closes <https://github.com/emilk/egui/issues/2000>.

[demo.webm](https://user-images.githubusercontent.com/5354222/193632435-6b4d93fb-73a0-4658-9fec-687d13fc84ee.webm)

### DIFF
--- a/crates/eframe/README.md
+++ b/crates/eframe/README.md
@@ -2,7 +2,6 @@
 
 [![Latest version](https://img.shields.io/crates/v/eframe.svg)](https://crates.io/crates/eframe)
 [![Documentation](https://docs.rs/eframe/badge.svg)](https://docs.rs/eframe)
-[![unsafe forbidden](https://img.shields.io/badge/unsafe-forbidden-success.svg)](https://github.com/rust-secure-code/safety-dance/)
 ![MIT](https://img.shields.io/badge/license-MIT-blue.svg)
 ![Apache](https://img.shields.io/badge/license-Apache-blue.svg)
 
--- a/crates/egui-winit/README.md
+++ b/crates/egui-winit/README.md
@@ -2,7 +2,6 @@
 
 [![Latest version](https://img.shields.io/crates/v/egui-winit.svg)](https://crates.io/crates/egui-winit)
 [![Documentation](https://docs.rs/egui-winit/badge.svg)](https://docs.rs/egui-winit)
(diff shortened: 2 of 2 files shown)


samitbasu:
One consideration is that "selected" in `egui` currently seems to imply a text field.  So it make make sense to specialize `Table` into e.g., `SelectableTable` with an additional requirement that the entries in the table `impl Into<WidgetText>` or whatever.

deft01:
Hi do you know if this is still working ? Thanks !

deft01:
Im really interested in this feature

PR #2115, opened 2022-10-06
Add ability to control double click reset in plot widget

Raymi306:
While building a tool that uses the plot widget, I wanted to be able to disable the double click reset.
This PR allows for that.

### DIFF
--- a/crates/egui/src/widgets/plot/mod.rs
+++ b/crates/egui/src/widgets/plot/mod.rs
@@ -816,7 +816,7 @@ impl Plot {
         };
 
         // Allow double clicking to reset to the initial bounds?
-        if response.double_clicked_by(PointerButton::Primary) && allow_double_click_reset {
+        if allow_double_click_reset && response.double_clicked_by(PointerButton::Primary) {
             bounds_modified = false.into();
         }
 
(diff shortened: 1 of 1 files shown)


Raymi306 (commented):
I did some small housekeeping and tried to organize the allow_ struct fields in a consistent way.
I added a new struct field allow_double_click_reset.
The important bit is on line 818 where I added to the if guard using the new struct field.

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -406,6 +408,12 @@ impl Plot {
         self
     }
 
+    /// Whether to allow double clicking to reset the
```
emilk, line 411:
```suggestion
    /// Whether to allow double clicking to reset the view.
    /// Default: `true`.
```

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -805,8 +814,8 @@ impl Plot {
             }
         };
 
-        // Allow double clicking to reset to the initial bounds.
-        if response.double_clicked_by(PointerButton::Primary) {
+        // Allow double clicking to reset to the initial bounds?
+        if response.double_clicked_by(PointerButton::Primary) && allow_double_click_reset {
```
emilk, line 818: small optimization

```suggestion
        if allow_double_click_reset && response.double_clicked_by(PointerButton::Primary) {
```

emilk (changes requested):

PR #2126, opened 2022-10-10
Only call the request_repaint_callback at most once per frame

emilk:
egui-winit adds new repaint events to the event loop on each call, and on some platforms this becomes very expensive, and can even lead to deadlocks (reported by @teh-cmc) when called thousands of times in quick succession.

@teh-cmc please try this and see if it fixes your deadlock issues.

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -46,17 +46,25 @@ struct ContextImpl {
     output: PlatformOutput,
 
     paint_stats: PaintStats,
+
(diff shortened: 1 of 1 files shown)


teh-cmc:
Surprisingly/interestingly, this doesn't seem to help. It does give me a bunch of new avenues to try and continue my investigations later on though.

Aside from that, I guess it's a worthwhile addition anyhow?

emilk:
Oh 😞 

Yeah, I don't think it can hurt, so might as well merge it.

PR #2143, opened 2022-10-12
Use WidgetText for the slider's text

Kryptos-FR:
**Summary**
Allow to use a custom `RichText` as the label of a slider.

**Rationale**
Currently, the label of a `Slider` is constructed from text only (`impl ToString`). It cannot be customized/styled except for its color (through the `text_color()` function). Sometimes, users might want to use `italics`, `underline`, `strikethrough` or change the size.

By taking a `Into<WidgetText>` instead, the previous behavior is maintained, but users can also provide their own custom `RichText`:

```rust
let text = RichText::new(&self.text).italics();
let slider = Slider::new(value, min..=max).text(text);
```

Checklist
- [x] `cargo fmt`
- [x] `cargo clippy`
- [x] `./sh/check.sh`
   all green until `+ ./sh/wasm_bindgen_check.sh`
   >    Compiling wasm-bindgen-cli v0.2.83
   > error[E0425]: cannot find function `match_assets` in crate `rouille`

### DIFF
--- a/crates/egui/src/widgets/slider.rs
+++ b/crates/egui/src/widgets/slider.rs
@@ -76,8 +76,7 @@ pub struct Slider<'a> {
     orientation: SliderOrientation,
     prefix: String,
     suffix: String,
-    text: String,
-    text_color: Option<Color32>,
+    text: WidgetText,
     /// Sets the minimal step of the widget value
     step: Option<f64>,
     min_decimals: usize,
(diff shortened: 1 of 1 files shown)


emilk (approved):
Great!

Kryptos-FR:
Should I add samples using this feature? Maybe somewhere in the `demo_app`?

emilk:
Nah, this is trivial enough imho :) thanks for the PR!

PR #2163, opened 2022-10-18
Add option to disable mouse dragging on scroll area.

alphaqu:
Adds an option in scrollarea to disable dragging with the mouse seperatley from the option to disable scrolling entirely.
There is not much more to this commit and i am unsure if it is a trivial change.

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -90,6 +90,8 @@ pub struct ScrollArea {
     offset_y: Option<f32>,
     /// If false, we ignore scroll events.
     scrolling_enabled: bool,
+    /// If false we ignore mouse drag events.
+    mouse_dragging_enabled: bool,
 
     /// If true for vertical or horizontal the scroll wheel will stick to the
     /// end position until user manually changes position. It will become true
@@ -132,6 +134,7 @@ impl ScrollArea {
             offset_x: None,
             offset_y: None,
             scrolling_enabled: true,
+            mouse_dragging_enabled: true,
             stick_to_end: [false; 2],
         }
     }
@@ -251,6 +254,19 @@ impl ScrollArea {
         self
     }
 
+    /// Control the dragging behavior
+    /// If `true` (default), the scroll area will respond to mouse dragging.
+    /// If `false`, the scroll area will not respond to mouse dragging
+    ///
+    /// This can be used, for example for code editors where
+    /// selecting and scrolling with a mouse is required.
+    ///
+    /// This controls both scrolling directions.
+    pub fn enable_mouse_dragging(mut self, enable: bool) -> Self {
+        self.mouse_dragging_enabled = enable;
+        self
+    }
+
     /// For each axis, should the containing area shrink if the content is small?
     ///
     /// If true, egui will add blank space outside the scroll area.
@@ -305,6 +321,7 @@ struct Prepared {
     /// `viewport.min == ZERO` means we scrolled to the top.
     viewport: Rect,
     scrolling_enabled: bool,
+    mouse_dragging_enabled: bool,
     stick_to_end: [bool; 2],
 }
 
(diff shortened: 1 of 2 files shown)


crates/egui/src/containers/scroll_area.rs
```diff
@@ -251,6 +254,19 @@ impl ScrollArea {
         self
     }
 
+    /// Control the dragging behavior
+    /// If `true` (default), the scroll area will respond to mouse dragging.
```
emilk, line 258:
```suggestion
    /// Control the dragging behavior.
    ///
    /// If `true` (default), the scroll area will respond to mouse dragging.
```

crates/epaint/src/text/text_layout.rs
```diff
@@ -122,14 +122,13 @@ fn layout_section(
 
             paragraph.glyphs.push(Glyph {
                 chr,
-                pos: pos2(paragraph.cursor_x, f32::NAN),
+                pos: pos2(font.round_to_pixel(paragraph.cursor_x), f32::NAN),
```
emilk, line 125: wrong PR!

crates/egui/src/containers/scroll_area.rs
```diff
@@ -251,6 +254,19 @@ impl ScrollArea {
...
+    /// If `false`, the scroll area will not respond to mouse dragging
+    ///
+    /// This can be used, for example for code editors where
+    /// selecting and scrolling with a mouse is required.
+    ///
+    /// This controls both scrolling directions.
+    pub fn enable_mouse_dragging(mut self, enable: bool) -> Self {
```
emilk, line 265: I think we should communicate that this is in particular for "drag to scroll". Perhaps it should be called `drag_to_scroll`?

emilk (changes requested):
thanks!

crates/epaint/src/text/text_layout.rs
alphaqu: I swear i double checked lmao

crates/egui/src/containers/scroll_area.rs
```diff
@@ -608,7 +629,7 @@ impl Prepared {
 
         if content_is_too_large[0] || content_is_too_large[1] {
             // Drag contents to scroll (for touch screens mostly):
-            let sense = if self.scrolling_enabled {
+            let sense = if self.drag_to_scroll_enabled {
```
mkrueger, line 632: Shouldn't it check both things ?
let sense = if self.scrolling_enabled && self.drag_to_scroll_enabled {

mkrueger:
I can really, really need that. Thanks.

crates/egui/src/containers/scroll_area.rs
```diff
@@ -90,6 +90,8 @@ pub struct ScrollArea {
     offset_y: Option<f32>,
     /// If false, we ignore scroll events.
     scrolling_enabled: bool,
+    /// If false we ignore mouse drag events.
+    drag_to_scroll_enabled: bool,
```
emilk, line 94: When would you want to turn this off for just one ScrollArea an not another?
Wouldn't it make more sense to put this as global setting in `Style` instead?

crates/egui/src/containers/scroll_area.rs
```diff
@@ -521,6 +541,7 @@ impl Prepared {
             content_ui,
             viewport: _,
             scrolling_enabled,
+            drag_to_scroll_enabled: mouse_dragging_enabled,
```
emilk, line 544:
```suggestion
            drag_to_scroll_enabled,
```

crates/egui/src/containers/scroll_area.rs
emilk: Yes, it should

emilk (changes requested):
EDIT: wrong comment

PR #2164, opened 2022-10-18
Fix some monospace fonts spacing being non-monospace

alphaqu:
On fonts like Jetbrains Mono the rounding to the closest pixel after every glyph made the spacing between the characters non-monospace. This fixes that issue by rounding the glyph location instead of the cursor x position which affects the future characters in a non-expected way.

### DIFF
--- a/crates/epaint/src/text/text_layout.rs
+++ b/crates/epaint/src/text/text_layout.rs
@@ -122,14 +122,13 @@ fn layout_section(
 
             paragraph.glyphs.push(Glyph {
                 chr,
-                pos: pos2(paragraph.cursor_x, f32::NAN),
+                pos: pos2(font.round_to_pixel(paragraph.cursor_x), f32::NAN),
                 size: vec2(glyph_info.advance_width, font_height),
                 uv_rect: glyph_info.uv_rect,
                 section_index,
             });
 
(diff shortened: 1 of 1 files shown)


emilk:
I tried this out with the default font with `pixels_per_point = 1.15`.

## Before:
old-rounding


## After:
new-rounding


The new code produces some bad kerning in many places:
Screen Shot 2022-10-18 at 17 17 37


Screen Shot 2022-10-18 at 17 18 44

Screen Shot 2022-10-18 at 17 18 53

alphaqu:
A possible fix might be to only apply this fix on monospace fonts?

alphaqu:
Another solution to this problem is to provide the glyph width correctly in the first place. Rendering rounds the glyph_width which makes it be different than the application (for example a cursor drawer). Just rounding the glyph_width on the getter makes the behavior consistent across the codebase.

parasyte:
@alphaqu could you try the patch in #2490? I don't have an app using Jetbrains Mono, and I'm not exactly sure what I should be looking for to compare between actual and expected rasterization because there are no screenshots here.

The linked patch may address the issue you had if your `pixels_per_point` is greater than 1.0. (I.e. on a high DPI display.)

emilk:
I think this is now fixed on main

PR #2167, opened 2022-10-19
Allows creating a `ColorImage` struct without an alpha channel

paulhazen:
Allows creating a `ColorImage` struct without an alpha channel.

- [X] ran `cargo test`
- [X] ran `cargo fmt` 
- [X] ran `cargo clippy`
- [X] ran `./sh/check.sh`

**Background:**
I'm a student working on an image processing library in rust. As a base we are required to implement our own IO / image manipulation centered around the [PPM image format](https://netpbm.sourceforge.net/doc/ppm.html). The PPM image format does not use an alpha channel. 

In order to get `epaint` to play nice with this concept, I have to design my data structures with an extra `0xff` alpha channel. With this change, folk can create a `ColorImage` struct without an alpha channel, which lets me save like, a few bytes in my data structure. But... it feels like the *right* way to do it so 🤷

### DIFF
--- a/crates/epaint/src/image.rs
+++ b/crates/epaint/src/image.rs
@@ -107,7 +107,7 @@ impl ColorImage {
     /// you are ignoring the alpha channel - considering it to always be 0xff)
     ///
     /// Panics if `size[0] * size[1] * 3 != rgba.len()`.
-    pub fn from_rgb_unmultiplied(size: [usize; 2], rgb: &[u8]) -> Self {
+    pub fn from_rgb(size: [usize; 2], rgb: &[u8]) -> Self {
         assert_eq!(size[0] * size[1] * 3, rgb.len());
         let pixels = rgb
             .chunks_exact(3)
--- a/crates/epaint/src/image.rs
+++ b/crates/epaint/src/image.rs
@@ -101,6 +101,21 @@ impl ColorImage {
         Self { size, pixels }
     }
 
+    /// Create a [`ColorImage`] from flat un-multiplied RGB data.
+    ///
+    /// This is what you want to use after having loaded an image file (and if
+    /// you are ignoring the alpha channel - considering it to always be 0xff)
+    ///
+    /// Panics if `size[0] * size[1] * 3 != rgba.len()`.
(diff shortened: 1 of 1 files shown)


crates/epaint/src/image.rs
```diff
@@ -101,6 +101,21 @@ impl ColorImage {
         Self { size, pixels }
     }
 
+    /// Create a [`ColorImage`] from flat un-multiplied RGB data.
+    ///
+    /// This is what you want to use after having loaded an image file (and if
+    /// you are ignoring the alpha channel - considering it to always be 0xff)
+    ///
+    /// Panics if `size[0] * size[1] * 3 != rgba.len()`.
+    pub fn from_rgb_unmultiplied(size: [usize; 2], rgb: &[u8]) -> Self {
```
emilk, line 110:
```suggestion
    pub fn from_rgb(size: [usize; 2], rgb: &[u8]) -> Self {
```

if you only have rgb, there is no alpha for it to have been multiplied with

crates/epaint/src/image.rs
```diff
@@ -101,6 +101,21 @@ impl ColorImage {
         Self { size, pixels }
     }
 
+    /// Create a [`ColorImage`] from flat un-multiplied RGB data.
```
emilk, line 104:
```suggestion
    /// Create a [`ColorImage`] from flat RGB data.
```

emilk (approved):

crates/epaint/src/image.rs
```diff
@@ -101,6 +101,21 @@ impl ColorImage {
         Self { size, pixels }
     }
 
+    /// Create a [`ColorImage`] from flat RGB data.
+    ///
+    /// This is what you want to use after having loaded an image file (and if
+    /// you are ignoring the alpha channel - considering it to always be 0xff)
+    ///
+    /// Panics if `size[0] * size[1] * 3 != rgba.len()`.
```
Mingun, line 109: Misprint or copy-paste error:
```suggestion
    /// Panics if `size[0] * size[1] * 3 != rgb.len()`.
```

Issue #2018, opened 2022-09-07
Can not run official example with error NoAvailablePixelFormat...


rikonaka:
**Describe the bug**

As my title say...

**To Reproduce**
1. Clone the repo to local
2. Into the repo dir and run command `cargo run -p hello_world`

**Expected behavior**
Run and show a gui.

**Screenshots**
![image](https://user-images.githubusercontent.com/13602602/188772086-5a3aecc8-5f0a-4547-b8c4-da240a409500.png)

**Desktop (please complete the following information):**
 - OS:  Ubuntu 22.04 with Gnome and Wayland.
 - Browser  Both Firefox and Chrome.
 - Version  Default version..

**Additional context**

No.

ar37-rs:
glxinfo, which opengl version? @rikonaka

tyrumus:
Not OP, but I have the exact same error. Here's the output that was asked for.

```
     Running `/home/naddan/git/egui/target/debug/hello_world`
2022-09-14T00:33:43.096531Z DEBUG eframe: Using the glow renderer
2022-09-14T00:33:43.099789Z DEBUG eframe::native::run: event_loop.run_return
2022-09-14T00:33:43.105911Z DEBUG crossfont::ft: Loaded Face Face { ft_face: Font Face: Regular, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }    
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: NoAvailablePixelFormat', crates/eframe/src/native/run.rs:324:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Configuration details:
Desktop, Arch Linux
Kernel: 5.15.67-1-lts
Graphics Platform: Wayland (sway/wlroots)
GPU: NVIDIA GTX 1070 (driver version 515.65.01)
```
$ lspci | grep VGA
65:00.0 VGA compatible controller: NVIDIA Corporation GP104 [GeForce GTX 1070] (rev a1)
```

When I run `hello_world` on my laptop with the same configuration, except for the GPU being different, it behaves normally.
```
$ lspci | grep VGA
00:02.0 VGA compatible controller: Intel Corporation TigerLake-LP GT2 [Iris Xe Graphics] (rev 01)
```

unbelievable as it may be, methinks this is possibly an NVIDIA/Wayland problem.

samitbasu:
Also not the OP, but I ran into this exact same problem.  Config:
```
[...]

PR #2187, opened 2022-10-27
Glutin Upgrade 

coderedart:
updating glutin to `0.30`. verified on latest arch plasma/gnome with both x11 and wayland. 

for some reason, text becomes more blurry on x11. but it doesn't happen using x11 backend on wayland. so, needs more debugging before merge. 

Closes https://github.com/emilk/egui/issues/2018


coderedart:
this glutin upgrade can actually simplify eframe a lot by reducing the glow / wgpu split as glutin no longer holds the winit window.

Although, this might also break existing programs as glutin just finished this new way of doing things.

coderedart:
tested on 
Windows:
Nvidia PC

Arch Linux:
AMD igpu laptop
Nvidia PC

TODO:
- [ ] need testers on windows /linux with intel GPUs. as well as Mac. 
- [ ] remove the dbg! macro / lint after testing
- [ ] maybe update the egui_glow's glutin example?


I kept this change as minimal as possible so that we can revert this easily in case of too many issues. But if this works fine for most users, we can remove a lot of boilerplate code between wgpu/glutin split in a future larger patch. and maybe get android glow working :)

crates/eframe/src/native/run.rs
```diff
@@ -276,7 +276,41 @@ mod glow_integration {
...
+            self.gl_surface.resize(
+                &self.gl_context,
+                physical_size
+                    .width
+                    .try_into()
+                    .expect("physical size must not be zero"),
+                physical_size
+                    .height
+                    .try_into()
+                    .expect("physical size must not be zero"),
```
emilk, line 303: I seem to recall some platforms reporting 0x0 when window is minimized, so I'd rather have a safe fallback here (e.g. falling back to `1`).

emilk:
Works great on my M1 Mac! 🙏

crates/eframe/src/native/run.rs
coderedart: eframe already checks if size is `> 0` and only then calls the resize fn with arguments, so it should be fine i think. 

> same file, line 708
```rust
       winit::event::WindowEvent::Resized(physical_size) => {
                                // Resize with 0 width and height is used by winit to signal a minimize event on Windows.
                                // See: https://github.com/rust-windowing/winit/issues/208
                                // This solves an issue where the app would panic when minimizing on Windows.
                                if physical_size.width > 0 && physical_size.height > 0 {
                                    running.gl_window.resize(*physical_size);
                                }
```

kchibisov:
You might be interested in using https://crates.io/crates/glutin-winit if you're plan on using glutin just with winit. There's also a glutin 0.30.1 which fixes issues on Windows wrt `swap_buffers` performance.

coderedart:
the glutin part is really tiny atm, but maybe `glutin-winit` can ease some future maintenance as its an official crate too? it also explicitly supports an API for multi-window which is something we will need in the future. 

the decision is upto emlik. i will clean up the code and finalize the PR tomorrow as we need some time to fix bugs that might arise from shifting to the new glutin version in master branch.

kchibisov:
>the glutin part is really tiny atm, but maybe glutin-winit can ease some future maintenance as its an official crate from glutin repo? it also explicitly supports an API for multi-window which is something we will need in the future.

It basically hides stuff like `x11_visual` and `DisplayApiPreference`, so you don't have platform specific code, but it's up to you.

fgimian:
Would it be possible to also update the example in the egui-glow crate? https://github.com/emilk/egui/tree/master/crates/egui_glow/examples

This would be nice as the example is easier to read and understand when attempting to integrate egui directly with glow in our applications 😄 

Cheers
Fotis

crates/eframe/src/native/run.rs
```diff
@@ -364,25 +395,135 @@ mod glow_integration {
...
+                .with_transparent(native_options.transparent)
                 // Keep hidden until we've painted something. See https://github.com/emilk/egui/pull/2279
                 // We must also keep the window hidden until AccessKit is initialized.
                 .with_visible(false);
-
+            let winit_window = window_builder
+                .build(event_loop)
+                .expect("failed to create winit window");
+            // a lot of the code below has been lifted from glutin example in their repo.
             let gl_window = unsafe {
```
emilk, line 406: it would be nice if everything within this `unsafe` block (returning `gl_window`) would be extracted to its own function!

crates/eframe/src/native/run.rs
```diff
@@ -364,25 +395,135 @@ mod glow_integration {
...
+                        &config.num_samples(),
+                        &config.alpha_size(),
+                        &config.api(),
+                        &config.srgb_capable(),
+                        config.depth_size(),
+                        config.config_surface_types(),
+                        config.color_buffer_type()
+                    );
+                }
+
```
emilk, line 496: and yes, remove this :)

emilk (approved):
I think we should try to clean this up and merge this. That's the only way we will get more people to test it :P

coderedart:
I think its ready now. we can eventually add `glutin-winit` to cleanup some platform gated code and support android.

emilk:
thanks!

Issue #2176, opened 2022-10-21
Open a menu upwards instead of downwards


jsguay-haven:
This may be similar to #1176 but only in a general sense that if I use the egui::TopBottomPanel::bottom() and put a menu button in it, the menu still opens downwards so it's outside the screen.

**Is your feature request related to a problem? Please describe.**

I'd like to be able to have menu buttons on the egui::TopBottomPanel::bottom() panel, and have the menus open upwards.

**Describe the solution you'd like**

Ideally I think if a popup is going to open outside the screen it should be automatically repositioned to be inside the screen, as mentioned in the above issue.

**Describe alternatives you've considered**

I tried to write a version of popup_below_widget (called popup_above_widget) but I don't know how to get the size of the menu that would be created by add_contents(ui) in order to offset the popup (in .fixed_pos(...)). This is down to my lack of knowledge, so if this is something that's easy to answer, it may be sufficient to do what I want.

**Additional context**

Nothing to add really.

Thanks in advance!

PR #2191, opened 2022-10-28
Fixed menu popups going outside of the screen

ItsEthra:
Closes <https://github.com/emilk/egui/issues/2176>.

### DIFF
--- a/crates/egui/src/menu.rs
+++ b/crates/egui/src/menu.rs
@@ -296,8 +296,7 @@ impl MenuRoot {
                 let screen_rect = response.ctx.input().screen_rect;
 
                 if pos.y + menu_rect.height() > screen_rect.max.y {
-                    pos.y = screen_rect.max.y - menu_rect.height();
-                    pos.x += response.rect.width();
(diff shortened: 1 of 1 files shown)


crates/egui/src/menu.rs
```diff
@@ -288,7 +288,23 @@ impl MenuRoot {
...
+            drop(input);
+
+            let mut pos = response.rect.left_bottom();
+            if let Some(root) = root.inner.as_mut() {
+                let menu_rect = root.menu_state.read().rect;
+                let screen_rect = response.ctx.input().screen_rect;
+
+                if pos.y + menu_rect.height() > screen_rect.max.y {
+                    pos.y = screen_rect.max.y - menu_rect.height();
+                    pos.x += response.rect.width();
```
emilk, line 300: I guess this is so that the menu doesn't cover the button, but wouldn't it make more sense to move the menu up by another `response.rect.height()`, so the menu is now above the button? That way a menu-button at the bottom of the screen will have its menu open above it, which makes sense to me.

ItsEthra: It does make sense, I fixed it.

emilk (approved):

PR #2195, opened 2022-10-29
fix bug on examples/custom_3d_three-d/lib.rs

kotlin2018:


emilk:
I don't think there is a bug here

PR #2202, opened 2022-10-30
Keyboard shortcut helpers

emilk:
This adds `egui::KeyboardShortcut` which is `Modifiers` + `Key`.

# TODO:
* [ ] PR description
* [ ] Changelog
* [ ] Menu button support

### DIFF
--- a/crates/egui/src/data/input.rs
+++ b/crates/egui/src/data/input.rs
@@ -727,6 +727,34 @@ impl Key {
 
 // ----------------------------------------------------------------------------
 
+/// A keyboard shortcut, e.g. `Ctrl+Alt+W`
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct KeyboardShortcut {
+    pub modifiers: Modifiers,
+    pub key: Key,
+}
+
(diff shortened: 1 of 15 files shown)


crates/egui/src/data/input.rs
```diff
@@ -563,6 +635,129 @@ pub enum Key {
...
 
+impl Key {
+    /// Emoji or name representing the key
+    pub fn symbol_or_name(self) -> &'static str {
+        match self {
+            Key::ArrowDown => "⏷",
+            Key::ArrowLeft => "⏴",
+            Key::ArrowRight => "⏵",
+            Key::ArrowUp => "⏶",
+            _ => self.name(),
```
winne42, line 646: https://wincent.com/wiki/Unicode_representations_of_modifier_keys has even more Unicode symbols for special keys

emilk: Nice link I think I'll add it as a TODO, because we need to check what characters are supported by a font before we use them (the egui default font supports those arrow triangles, but no others in that link).

teh-cmc (approved):

Issue #2203, opened 2022-10-31
`DragValue::clamp_range(0.0..=max)` allows `-0.0`


SludgePhD:
**Describe the bug**

Using `DragValue::clamp_range(0.0..=max)` to force a `DragValue` to be positive does not fully work: It allows the value to be dragged to `-0.0`, which can then cause other methods to fail (eg. [`Duration::from_secs_f32`](https://doc.rust-lang.org/stable/std/time/struct.Duration.html#method.from_secs_f32)).

While `0.0` and `-0.0` compare equal, `is_sign_negative` will still return `true`, since the sign bit is set in the value.

**To Reproduce**
1. Create a `egui::DragValue::new(&mut value).speed(0.01).clamp_range(0.0..=f32::MAX)`
2. Drag it to `0.01`
3. Drag it downwards
4. (might require some fiddling, it doesn't always happen)

**Expected behavior**

I would not expect to see any values with the sign bit set if `.clamp_range(0.0..=f32::MAX)` is used.

**Desktop (please complete the following information):**
 - OS: Linux (Wayland) with winit/bevy, but this does not look like platform-specific behavior

ItsEthra:
This happens because `clamp`, `min` and `max` functions used by `clamp_to_range` in file `drag_value.rs:510` are not using `total_cmp`. One easy fix is to implement it manual clamping using [total_cmp](https://doc.rust-lang.org/std/primitive.f64.html#method.total_cmp) but this would require bumping minimum supported rust version from 1.61 to 1.62.

PR #2213, opened 2022-10-31
Use total_cmp for clamping drag value in order to avoid floating point ambiguities

ItsEthra:
This pr increases MSRV because total_cmp was stabilized only in 1.62

Closes <https://github.com/emilk/egui/issues/2203>.


crates/egui/src/widgets/drag_value.rs
```diff
@@ -508,8 +508,32 @@ impl<'a> Widget for DragValue<'a> {
...
-        range.start().min(*range.end()),
-        range.start().max(*range.end()),
-    )
+    match x.total_cmp(range.start()) {
+        Ordering::Less | Ordering::Equal => *range.start(),
+        Ordering::Greater => match x.total_cmp(range.end()) {
+            Ordering::Greater | Ordering::Equal => *range.end(),
+            Ordering::Less => x,
+        },
+    }
```
emilk, line 517: Would this simpler solution also work:

``` rust
let min = range.start().min(*range.end());
let max = range.start().max(*range.end());
if x <= min {
    min
} else if x <= max {
    x
} else {
    max
}
```

I think that should also cover the `-0.0` case

crates/egui/src/widgets/drag_value.rs
```diff
@@ -508,8 +508,32 @@ impl<'a> Widget for DragValue<'a> {
 }
 
 fn clamp_to_range(x: f64, range: RangeInclusive<f64>) -> f64 {
-    x.clamp(
-        range.start().min(*range.end()),
-        range.start().max(*range.end()),
-    )
+    match x.total_cmp(range.start()) {
```
emilk, line 511: The previous implementation handles the case of reversed range, i.e. `clamp_to_range(5.0, 10.0..=0.0) == 5.0`. The new versions fails here.

crates/egui/src/widgets/drag_value.rs
ItsEthra: Documentation states that `PartialEq` and `PartialOrd` implementation of `f32` and `f64` differs from ordering of [total_cmp](https://doc.rust-lang.org/nightly/std/primitive.f64.html#method.total_cmp). So I believe it won't be as simple as just using operators. 

> The ordering established by this function does not always agree with the [PartialOrd](https://doc.rust-lang.org/nightly/std/cmp/trait.PartialOrd.html) and [PartialEq](https://doc.rust-lang.org/nightly/std/cmp/trait.PartialEq.html) implementations of f64. For example, they consider negative and positive zero equal, while total_cmp doesn’t.

crates/egui/src/widgets/drag_value.rs
ItsEthra: Fixed with new commit

crates/egui/src/widgets/drag_value.rs
emilk: Yeah, but I'm pretty sure I set up the code above to work correctly for the `-0` vs `+0` cases (and chose to ignore NaN, because whatever we do there is bad)

ItsEthra: Yeah you're right, but it won't work in the case of `clamp_to_range(0.0, (-1.0)..=(-0.0))`, although I am not sure if there is any use-case for this kind of input.

emilk: ah, right. perhaps your original code was better. Best to cover it all with tests though.

ItsEthra: I think I covered all cases in the new commit

PR #2230, opened 2022-11-03
wgpu: add support for user-level command buffers + viewport clarification

teh-cmc:
This PR adds support for returning user-defined command buffers from the `prepare` callback when using the wgpu backend.
The main command buffer as well as all user-defined ones are submitted together in a single call.

In particular, this gives the user an opportunity to parallelize command registration and prevents a faulty callback from poisoning the main wgpu pipeline.

Also adds some comments regarding viewports, and how the user can override them.

cc @Wumpf

### DIFF
--- a/crates/egui-wgpu/src/renderer.rs
+++ b/crates/egui-wgpu/src/renderer.rs
@@ -14,17 +14,12 @@ use wgpu::util::DeviceExt as _;
 /// A callback function that can be used to compose an [`egui::PaintCallback`] for custom WGPU
 /// rendering.
 ///
-/// The callback is composed of two functions: `prepare` and `paint`.
-///
-/// `prepare` is called every frame before `paint`, and can use the passed-in [`wgpu::Device`] and
-/// [`wgpu::Buffer`] to allocate or modify GPU resources such as buffers.
-/// Additionally, a [`wgpu::CommandEncoder`] is provided in order to allow creation of
-/// custom [`wgpu::RenderPass`]/[`wgpu::ComputePass`] or perform buffer/texture copies
-/// which may serve as preparation to the final `paint`.
-/// (This allows reusing the same [`wgpu::CommandEncoder`] for all callbacks and egui rendering itself)
-///
-/// `paint` is called after `prepare` and is given access to the [`wgpu::RenderPass`] so that it
-/// can issue draw commands into the same [`wgpu::RenderPass`] that is used for all other egui elements.
+/// The callback is composed of two functions: `prepare` and `paint`:
+/// - `prepare` is called every frame before `paint`, and can use the passed-in
+///   [`wgpu::Device`] and [`wgpu::Buffer`] to allocate or modify GPU resources such as buffers.
+/// - `paint` is called after `prepare` and is given access to the [`wgpu::RenderPass`] so
+///   that it can issue draw commands into the same [`wgpu::RenderPass`] that is used for
+///   all other egui elements.
 ///
 /// The final argument of both the `prepare` and `paint` callbacks is a the
 /// [`paint_callback_resources`][crate::renderer::Renderer::paint_callback_resources].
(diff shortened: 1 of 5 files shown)


crates/egui-wgpu/src/renderer.rs
```diff
@@ -60,10 +66,26 @@ impl CallbackFn {
...
+    /// The passed-in `CommandEncoder` is egui's and can be used directly to register
+    /// wgpu commands for simple use cases.
+    ///
+    /// For more complicated use cases, one can also return a list of arbitrary
+    /// `CommandBuffer`s and have complete control over they get created and fed.
+    /// In particular, this gives an opportunity to parallelize command registration and
+    /// prevents a faulty callback from poisoning the main wgpu pipeline.
+    ///
+    /// The main egui command buffer, as well as all user-defined command buffers returned
+    /// by this function, are guaranteed to all be submitted at once in a single call.
```
Wumpf, line 80: I really like your description here, but wonder if the doc block at the beginning should be shortened accordingly

Wumpf: one detail though:
```
    /// The main egui command buffer, as well as all user-defined command buffers returned
    /// by this function, are guaranteed to all be submitted at once in a single call.
```

If someone uses the `Renderer` without eframe/egui_wgpu::winit.rs, it is their responsibility to do so. So it's less a guarantee. So maybe
```
    /// by this function, can then be submitted at once in a single draw call (as done by eframe).
```

Wumpf (approved):
love it

crates/egui_demo_app/src/apps/custom3d_wgpu.rs
```diff
@@ -135,12 +135,18 @@ impl Custom3d {
         // Device and Queue, which can be used, for instance, to update buffers and uniforms before
         // rendering.
         //
+        // You can use the main `CommandEncoder` that is passed-in, return an arbitrary number
+        // of user-defined `CommandBuffer`s, or both.
```
emilk, line 139: Why the two ways of doing it? Are the `CommandEncoder` expensive to create, so better to reuse the same one for the common case?

Wumpf: correct. Although there is some lack of clarity how expensive exactly

Wumpf: the only thing that is very clear is that submit calls are very expensive - that's even in the official wgpu advice
https://github.com/gfx-rs/wgpu/wiki/Do's-and-Dont's#dont-submit-many-times-per-frame

emilk (approved):

crates/egui_demo_app/src/apps/custom3d_wgpu.rs
Wumpf: Actually https://developer.nvidia.com/blog/vulkan-dos-donts/ is talking about 5-15 command buffers per frame. So maybe we should not pass the command encoder after all. Doing so is a very thing I added and it wasn't connected to a user ask

Wumpf: I'll ask in wgpu matrix chat what folks recommend there :)

Wumpf: there is this bit in favor of encouraging the user to pile in more stuff:
```
Don’t submit small command buffers. If a submission is processed on the GPU faster than new ones can be submitted on the CPU, it will result in wasted / idle GPU cycles.
```

Wumpf: for reference, the same in DX12 flavor https://developer.nvidia.com/dx12-dos-and-donts#worksubmit
What I'm reading in all of this is that CommandBuffers should be [lagom](https://en.wikipedia.org/wiki/Lagom) ;)

Wumpf: for posterity, comment from @cwfitzgerald (one of the primary maintainers of wgpu) on the topic:

> This is a bit of an unknown question for me. I did some primative testing a long long time ago on the submission performance relative to command buffer count and found it was nearly quadratic, but that was well before I was at in touch with the stack so I don't put too much value in that number
Note that wgpu splits each renderpass into its own command buffer and the cost of this is still up in the air (and part of why this question is important)

PR #2232, opened 2022-11-03
add screenshots to examples README.md

winne42:


emilk:
This is great, but it also adds some maintenance cost, as changing the egui style should probably update the screenshots.

It would be amazing if these could be automatically generated somehow. Perhaps an environment variable caught by `eframe` (`EFRAME_SCREENSHOT_TARGET` or something), that would just run the app for just one frame, and then output the screenshot as png to the target filepath. Then all is needed is a script to run all the examples…

emilk:
Thanks for your PR - it made me spend the time to auto-generate these :)

PR #2233, opened 2022-11-03
fix comment for WebOptions follow_system_theme

msheldyakov:
it took me a while to figure out why the theme doesn't change :)
```
impl Default for WebOptions {
    fn default() -> Self {
        Self {
            follow_system_theme: true,
```


emilk (approved):

emilk:
oops!

Issue #980, opened 2021-12-21
Overlaping widgets


AlphaSteam:
Hello,

I'm using egui for a game engine project for card roguelikes and I wanted to use egui for the cards as ui elements.
The problem is that no matter what I do, I can't put text inside an image. As I read on another issue, widgets are expected not to overlap inside a window.

Could it be possible to allow widgets to overlap each other if explicitly told to do so?
This could help do composed widgets like you do on HTML.

I've tried using Areas, Windows, and Layouts. I tried putting the widget with put and using rects. I tried changing the order of the Widgets.

I also tried creating two areas at the same place, but they don't move at the same time. Perhaps there's a way to link two areas so they move at the same time, but I couldn't figure it out.

![card](https://user-images.githubusercontent.com/30818940/147002605-77d11e6d-36a4-4e9f-824c-b8ad61612f32.png)
I would like to put the text above and below, in front of the card.

If this is already possible, any help to achieve it will be appreciated.

Thanks for everyone's work on this project, it's an awesome library.

BctfN0HUK7Yg:
I think the better way is to create custom widget for card.
https://github.com/emilk/egui/blob/master/egui_demo_lib/src/apps/demo/toggle_switch.rs

parasyte:
For a little more context and a theoretical basis, thinking of a "widget" as its individual components (image, text, shapes, etc.) is probably not the right level of abstraction. A widget is a rectangular area that is allocated from the layout, but what exists inside that rectangle is up to the widget itself; it can render any shapes it wants inside that space.

This is how, e.g., the [`plot`](https://github.com/emilk/egui/tree/master/egui/src/widgets/plot) widget works. The paths and labels inside the `plot` are children of the parent rectangle. In short, widgets are a hierarchy, and "overlapping" can be naturally defined by the parent-child relationship.

emilk:
Overlapping widgets is something I also miss from time to time, e.g. to put controls on top of a canvas that can also be interacted with.

Right now the only way to have overlapping widgets is to use `Area`:s (e.g. `Windows`). egui remember the position and sizes of each `Area` and can therefor on each frame know which one is on top for any single position, and therefore which one should receive a click. The problem is: if you have multiple overlapping widgets in the same `Area`, which should receive the click?

Some idea for solutions:

* The first widget you add gets the interaction. Requires rendering the widgets front-to-back, which requires adding some better layering (sub-layers per Area) or z-order on the `Shapes`. This would be a nice feature in any case.
[...]

PR #2244, opened 2022-11-06
Allow overlapping interactive widgets

emilk:
Closes https://github.com/emilk/egui/issues/980

By interactive widget, I mean any widget with `Sense::click` or `Sense::drag` (or both).

This design is based on the idea that the latest added (which usually mean the top-most) widget gets the interaction.

For instance: say you have a big interactive canvas and on top of that put some widgets. If you hover the widgets, they get all the input. If you hover the canvas outside of the widgets, the canvas gets the input.

This is done by storing where all interactive widgets are, frame-to-frame. When a new interactive widget is added, we check if there is anything interactive above it (=added later in the previous frame), and if there is, the first widget gets `hovered=false`.

This change required some changes to `ScrollArea`, `Area` and `Window` to make sure they were properly layering interaction back-to-front.

There may be other looming bugs I haven't found yet.

Most user code should be unaffected, unless the user has added layers of interaction on top of each other.

You can visualize this using `style.debug.show_blocking_widget` and `style.debug.show_interactive_widgets`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 
 ## Unreleased
 * ⚠️ BREAKING: egui now expects integrations to do all color blending in gamma space ([#2071](https://github.com/emilk/egui/pull/2071)).
+* ⚠️ BREAKING: if you have overlapping interactive widgets, only the top widget (last added) will be interactive ([#2244](https://github.com/emilk/egui/pull/2244)).
 
 ### Added ⭐
 * Added helper functions for animating panels that collapse/expand ([#2190](https://github.com/emilk/egui/pull/2190)).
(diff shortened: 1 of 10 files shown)


trivigy:
@emilk This PR introduces a usability issue for `ScrollArea`.

use-case: If a scroll area has interactive widgets such as buttons which occupy majority of available space then, because those end up being places after the scrolling interactivity (in content function) and therefore steal interaction, it is impossible to actually scroll.

I am currently trying to figure out a solution/workaround and the only possible solution I imagine is to use `ctx.input` directly inside of the scroll area rather than `ui.interact()` in order to prevent the scroll area interactivity to use the `memory.interaction` entirely altogether.

Kryptos-FR:
> Most user code should be unaffected, unless the user has added layers of interaction on top of each other.

I am exactly in this situation. What is the workaround to still allow that?

Basically my scenario is: I have a some kind of graph with nodes represented as boxes which can have `DragValue` inside (to input some integer). Before `0.20.0`, I could implemented dragging the whole boxes by just allocating a UI of the same size on top of the rest, sensing `Sense::click_and_drag()`, getting the drag_delta and apply that to some value that will then be used the next time to draw the shape. Now because this takes the whole area, I can move the shape but the underlying `DragValue` can't get the mouse focus to input value anymore. This is a big regression.

What was the rationale behind that change? I can't see the benefit of not allowing it any more.

PR #2249, opened 2022-11-06
`Painter` extend accepts `IntoIter`

maxwase:
Non-breaking interface improvement for `Painter` and `PaintList` `extend` methods.


emilk (approved):
Nice!

PR #2252, opened 2022-11-07
Add a way to provide custom scaling to SVG rasterization

bluskript:
This provides a way to select a custom size at which SVGs get rasterized.

This should make it easier for users to render SVGs at the proper resolution without aliasing, and possibly provide a way to rasterize dynamically in the future (although I would like to first explore enabling mipmapping for RetainedImages because it would be a more efficient solution)

### DIFF
--- a/crates/egui_extras/src/image.rs
+++ b/crates/egui_extras/src/image.rs
@@ -1,5 +1,8 @@
 use egui::{mutex::Mutex, TextureFilter, TextureOptions};
 
+#[cfg(feature = "svg")]
+pub use usvg::FitTo;
+
 /// An image to be shown in egui.
 ///
 /// Load once, and save somewhere in your app state.
@@ -52,10 +55,7 @@ impl RetainedImage {
     /// On invalid image
     #[cfg(feature = "svg")]
     pub fn from_svg_bytes(debug_name: impl Into<String>, svg_bytes: &[u8]) -> Result<Self, String> {
-        Ok(Self::from_color_image(
-            debug_name,
-            load_svg_bytes(svg_bytes)?,
-        ))
+        Self::from_svg_bytes_with_size(debug_name, svg_bytes, FitTo::Original)
     }
 
     /// Pass in the str of an SVG that you've loaded.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 
 ### Added ⭐
 * Added helper functions for animating panels that collapse/expand ([#2190](https://github.com/emilk/egui/pull/2190)).
+* Added `RetainedImage::from_svg_bytes_with_size` to be able to specify a size for SVGs to be rasterized at.
```
emilk, line 12: Please move to `crates/egui_extras/CHANGELOG.md`

Cargo.lock
```diff
@@ -4557,9 +4563,9 @@ checksum = "d2d7d3948613f75c98fd9328cfdcc45acc4d360655289d0a7d4ec931392200a3"
 
 [[package]]
 name = "xmlparser"
-version = "0.13.3"
+version = "0.13.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "114ba2b24d2167ef6d67d7d04c8cc86522b87f490025f39f0303b7db5bf5e3d8"
```
emilk, line 4562: Please revert changes to `Cargo.lock` that aren't related to this PR

crates/egui_extras/src/image.rs
```diff
@@ -200,29 +217,54 @@ pub fn load_image_bytes(image_bytes: &[u8]) -> Result<egui::ColorImage, String>
...
+        },
+        match fit_to {
+            FitTo::Original => pixmap_size.height(),
+            FitTo::Size(_, h) => h,
+            FitTo::Height(h) => h,
+            FitTo::Width(w) => {
+                (pixmap_size.height() as f32 * (w as f32 / pixmap_size.width() as f32)) as u32
+            }
+            FitTo::Zoom(z) => (pixmap_size.height() as f32 * z) as u32,
+        },
```
emilk, line 258: I think this would be easier to read as one `match` statement

emilk (changes requested):
thanks

emilk (approved):

Issue #2093, opened 2022-09-30
[regression] eframe persistence doesn't work anymore for some UI states


quietvoid:
**Describe the bug**
In my app I have a `egui::Window` and some `egui::CollapsingHeader`, which stopped properly restoring from the persisted state.
The window is placed seemingly randomly, while the collapsing headers are always closed on app launch.

**To Reproduce**
1. Create a `CollapsingHeader`, in a native app project using persistence.
2. Toggle the header.
3. Quit the app.
4. Reopen app, the header isn't open anymore.

**Expected behavior**
The header should be open at app launch, as the state is persisted.

**Desktop (please complete the following information):**
 - OS: Arch Linux, Wayland (Sway)

**Additional context**
I have bisected the issue and it's caused by 5514a8afda53c282d2be670bef657d5c7812d467.
The last previous working commit is eeeb4b7de2d08b7923a0e19cf1a711b7219afa1f.

Not working 5514a8afda53c282d2be670bef657d5c7812d467 persisted file: [not-working.txt](https://github.com/emilk/egui/files/9679825/not-working.txt)
Working eeeb4b7de2d08b7923a0e19cf1a711b7219afa1f persisted file: [working.txt](https://github.com/emilk/egui/files/9679826/working.txt)

quietvoid:
Seems to be related strictly to the `ahash` related changes.

PR #2254, opened 2022-11-07
use `RandomState::with_seeds` replace `AHasher::default`

axxop:
Fix: #2093

AHash 0.8 related: https://github.com/tkaitchuck/aHash/issues/132

### DIFF
--- a/crates/egui/src/id.rs
+++ b/crates/egui/src/id.rs
@@ -45,16 +45,16 @@ impl Id {
 
     /// Generate a new [`Id`] by hashing some source (e.g. a string or integer).
     pub fn new(source: impl std::hash::Hash) -> Id {
-        use std::hash::Hasher;
-        let mut hasher = epaint::ahash::AHasher::default();
+        use std::hash::{BuildHasher, Hasher};
+        let mut hasher = epaint::ahash::RandomState::with_seeds(15, 26, 37, 48).build_hasher();
         source.hash(&mut hasher);
         Id(hasher.finish())
     }
 
     /// Generate a new [`Id`] by hashing the parent [`Id`] and the given argument.
     pub fn with(self, child: impl std::hash::Hash) -> Id {
-        use std::hash::Hasher;
-        let mut hasher = epaint::ahash::AHasher::default();
+        use std::hash::{BuildHasher, Hasher};
+        let mut hasher = epaint::ahash::RandomState::with_seeds(15, 26, 37, 48).build_hasher();
         hasher.write_u64(self.0);
         child.hash(&mut hasher);
         Id(hasher.finish())
--- a/crates/egui/src/id.rs
+++ b/crates/egui/src/id.rs
@@ -46,7 +46,7 @@ impl Id {
     /// Generate a new [`Id`] by hashing some source (e.g. a string or integer).
     pub fn new(source: impl std::hash::Hash) -> Id {
         use std::hash::{BuildHasher, Hasher};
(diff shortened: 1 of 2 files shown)


crates/egui/src/id.rs
```diff
@@ -45,16 +45,16 @@ impl Id {
 
     /// Generate a new [`Id`] by hashing some source (e.g. a string or integer).
     pub fn new(source: impl std::hash::Hash) -> Id {
-        use std::hash::Hasher;
-        let mut hasher = epaint::ahash::AHasher::default();
+        use std::hash::{BuildHasher, Hasher};
+        let mut hasher = epaint::ahash::RandomState::with_seeds(15, 26, 37, 48).build_hasher();
```
Mingun, line 49: Why exactly that seed is used? It is important? It definitely needs a comment. Also it is repeated three times, it is easy to change some of the places accidentally, which can break something, that is non-obvious from the code. I think, a function that will return `RandomState` is needed

emilk: Let's use a more clear [nothing-up-my-sleeve number](https://en.wikipedia.org/wiki/Nothing-up-my-sleeve_number) like `1, 2, 3, 4`

emilk, line 49:
```suggestion
        use std::hash::{BuildHasher as _, Hasher as _};
        let mut hasher = epaint::ahash::RandomState::with_seeds(1, 2, 3, 4).build_hasher();
```

crates/egui/src/id.rs
```diff
@@ -45,16 +45,16 @@ impl Id {
...
         source.hash(&mut hasher);
         Id(hasher.finish())
     }
 
     /// Generate a new [`Id`] by hashing the parent [`Id`] and the given argument.
     pub fn with(self, child: impl std::hash::Hash) -> Id {
-        use std::hash::Hasher;
-        let mut hasher = epaint::ahash::AHasher::default();
+        use std::hash::{BuildHasher, Hasher};
+        let mut hasher = epaint::ahash::RandomState::with_seeds(15, 26, 37, 48).build_hasher();
```
emilk, line 57:
```suggestion
        use std::hash::{BuildHasher as _, Hasher as _};
        let mut hasher = epaint::ahash::RandomState::with_seeds(1, 2, 3, 4).build_hasher();
```

crates/epaint/src/util/mod.rs
```diff
@@ -5,10 +5,8 @@ pub use ordered_float::*;
 /// Hash the given value with a predictable hasher.
 #[inline]
 pub fn hash(value: impl std::hash::Hash) -> u64 {
-    use std::hash::Hasher as _;
-    let mut hasher = ahash::AHasher::default();
-    value.hash(&mut hasher);
-    hasher.finish()
+    use ahash::RandomState;
+    RandomState::with_seeds(15, 26, 37, 48).hash_one(value)
```
emilk, line 9:
```suggestion
    use ahash::RandomState as _;
    RandomState::with_seeds(1, 2, 3, 4).hash_one(value)
```

emilk (changes requested):

emilk:
Good catch!

axxop:
updated, please check

emilk (approved):
Thanks!

crates/egui/src/id.rs
Mingun: I think, it is also need to mention in a comment, that the selected numbers are just random. Should they need to be the same in all three places? Or it is just a [random values](https://xkcd.com/221/), not connected to each other? What's happened if in the future someone change the one of them to another values?

PR #2258, opened 2022-11-07
Fix ScrollArea::show_rows

qthree:
I've encountered incorrect behavior of `ScrollArea::show_rows` when `min_row` becomes larger than `max_row`.
The way to fix it is to change `min_row` and `max_row` by the same amount when `max_row` larger then `total_rows`.
I can't provide test case atm.


emilk (approved):
Nice fix!

PR #2261, opened 2022-11-07
Panel visual improvements

emilk:
## Before
panel-framing-before

## After
panel-framing-after


trivigy:
@emilk This commit creates a regression bug. Because of this (https://github.com/emilk/egui/pull/2261/files#diff-1b1b7f67a664568e1201e4336ca926fa687c506ffaaf2f237a21a1fa4619fab7L715) change, 'TopBottomPanel' forces every usecase to always have a seperator hline irrelevant with regards to '.resizable(false)' or '.frame(Frame { stroke: Stroke::none(), ..Frame::none() })'. Those are simply ignored.

emilk:
@trivigy good catch. Fixed in https://github.com/emilk/egui/commit/df01db2df1d201c6d79139bb9dfb82fa85ffc8c9

Issue #2253, opened 2022-11-07
Auto-generated IDs of elements inside of `child_ui_with_id_source` don't use the given `id_source`


ervanalb:
Currently, it is not possible to give a `Slider` an `id` that is independent of the global draw order, even inside of a `child_ui_with_id_source`.

**Context**

I am writing an app that draws a series of "tiles" on a freeform canvas. The tile draw order is dictated by visual superposition and changes frame-to-frame. Each tile has a `child_ui_with_id_source` where the explicit `id_source` is used to track the tile's identity as the draw order changes.

**Bug**

![image](https://user-images.githubusercontent.com/4504093/200226326-057223f5-f8f2-4803-9113-d5060d04c375.png)

The problem is that the elements rendered inside of the child UI (such as `Slider`) get auto-generated IDs based on the global draw order. Therefore, when dragging a slider, which slider is being dragged bounces around as the draw order changes.

The expected behavior is that elements get auto-generated IDs based on their draw order within their parent `child_ui`, seeded by the parent UI's `id_source`. This means that as long as the tile's child UI has a consistent ID, the slider ID will be independent of tile draw order.

**Fix**

I was not able to find a workaround, and so I have created a [fork](https://github.com/emilk/egui/compare/0.19.0...ervanalb:egui:0.19.0-radiance-patches?expand=1) that fixes the problem.

I can open a PR with this change if, architecturally, this is the direction that egui wants to take.

enomado:
just use 
`    ui.push_id("inner", |ui| {...})`

ervanalb:
The buggy behavior is still present with `push_id`. It looks like internally, `push_id` just uses [child_ui_with_id_source](https://github.com/emilk/egui/blob/5515d2db7738b3b108c51ba668ad177ecd30a9e3/crates/egui/src/ui.rs#L1721). This is why I decided to make my patch to `child_ui_with_id_source`.

emilk:
I think you are probably are right that `child_ui_with_id_source` should generate a new `next_auto_id_source` based on the parent `id`, parent `next_auto_id_source`, and the given `id_source`.

ervanalb:
I have updated my app to use the latest version of egui and this problem has gone away. Thanks for all the updates to ID handling!

PR #2262, opened 2022-11-08
Auto-ids are seeded based on Ui::id

emilk:
Closes https://github.com/emilk/egui/issues/2253

I would appreciate help testing this on different projects.

### DIFF
--- a/crates/egui/src/ui.rs
+++ b/crates/egui/src/ui.rs
@@ -98,9 +98,7 @@ impl Ui {
     ) -> Self {
         crate::egui_assert!(!max_rect.any_nan());
         let child_id = self.id.with(id_source);
-        let child_next_auto_id_source = child_id.with(self.next_auto_id_source).value();
-
-        self.next_auto_id_source = self.next_auto_id_source.wrapping_add(1);
+        let child_next_auto_id_source = child_id.with("auto").value();
 
         Ui {
             id: child_id,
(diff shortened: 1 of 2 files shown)


crates/egui/src/ui.rs
```diff
@@ -97,17 +97,19 @@ impl Ui {
         id_source: impl Hash,
     ) -> Self {
         crate::egui_assert!(!max_rect.any_nan());
-        let next_auto_id_source = Id::new(self.next_auto_id_source).with("child").value();
+        let child_id = self.id.with(id_source);
+        let child_next_auto_id_source = child_id.with(self.next_auto_id_source).value();
+
         self.next_auto_id_source = self.next_auto_id_source.wrapping_add(1);
```
ervanalb, line 103: As-is, this PR does not fix the problem in my application (#2253.) I think this is because the child UI's `next_auto_id_source` is still dependent on how much stuff was rendered before the child_ui. Something like this works for me:

```suggestion
        let child_next_auto_id_source = child_id.with("auto").value();
```

emilk: The problem is `id_source` is not always unique in this instance. A UI can have many child-ui:s, all created with `child_ui` which use the same id_source (`"child"`), so we need to seed the auto-ids differently for all these children.

Applying this diff gets me:

Screen Shot 2022-12-02 at 10 08 12

emilk:
* Closed in favor of https://github.com/emilk/egui/pull/4840

PR #2270, opened 2022-11-09
Fixed color edit popup going outside the screen

ItsEthra:


crates/egui/src/widgets/color_picker.rs
```diff
@@ -350,13 +350,29 @@ pub fn color_edit_button_hsva(ui: &mut Ui, hsva: &mut Hsva, alpha: Alpha) -> Res
...
+    if anchor_pos.y + COLOR_SLIDER_WIDTH * WIDTH_TO_HIGHT > screen_max.y {
+        anchor_pos.y = screen_max.y - COLOR_SLIDER_WIDTH * WIDTH_TO_HIGHT;
+    }
+
     // TODO(emilk): make it easier to show a temporary popup that closes when you click outside it
     if ui.memory().is_popup_open(popup_id) {
         let area_response = Area::new(popup_id)
             .order(Order::Foreground)
-            .fixed_pos(button_response.rect.max)
+            .fixed_pos(anchor_pos)
```
emilk, line 373: It would be rather easy to add something like `.contrain(true)` here, and then have `Area::begin` do this:

```rust
if contraint {
    state.pos = ctx.constrain_window_rect_to_area(state.rect(), drag_bounds).min;
}
```

That would be a smaller change, a more robust change, and something more useful over all!

ItsEthra: That's neat because It also fixes ComboBox(or any popup) going outside the screen, thanks!

emilk (approved):
Thanks for taking your time to do this!

Please add a line in `CHANGELOG.md` about the new `Area::constrain` feature (and maybe add that feature to `Window` as well?) and I'll merge this!

ItsEthra:
Hm, Window is already constrained to the screen bounds and I don't see a reason why anyone would not want that.

emilk:
@ItsEthra yup, maybe `constrain` should be on by default.

PR #2274, opened 2022-11-09
bump msrv 1.64 and tts + bindgen dep

coderedart:
- bumps msrv to 1.64 in toolchain file and workflows
- manually bump `bindgen` dep using `cargo update -p bindgen` 
- bump tts version to latest in `egui_winit` and `eframe` crates
- change runner image from `ubuntu-latest` to `ubuntu-22.04`

This is specifically to fix the `speech-dispatcher` crate compilation issues on platforms like `Arch Linux`.

We ended up at this PR due to the following series of events:
1.  The main issue is that workflow uses `ubuntu-20.04` which has an old version of library in its repository. but other distros like Arch have the latest version which will fail to link due to breaking ABI changes. 
2. To update `speech-dispatcher` crate, we need to update `tts` crate version. 
3. latest version of `speech-dispatcher` crate has some issues with `bindgen` version `0.60`, so we need to manually bump that to `0.61` (latest).
4. `bindgen` latest version requires minimum version of rust to be `1.64` as it uses the `std::ffi::*` types . 

Finally,
* `speech-dispatcher` in `ubuntu-latest` (20.04 LTS) is version [`0.9`](https://packages.ubuntu.com/focal/speech-dispatcher). our changes won't compile because its too old and that library had breaking changes since then.
* `speech-dispatcher` in `ubuntu-22.04` (will be default for `ubuntu-latest` by this month end) is version [`0.11`](https://packages.ubuntu.com/jammy/speech-dispatcher) and will compile successfully.

if we want to stick with `ubuntu-latest`, then this PR will have to wait for December i guess. 

https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/
> Ubuntu-22.04 is ready to be the default version for the ubuntu-latest label in GitHub Actions workflows. This change will be rolled out over a period of 8 weeks beginning on October 1, 2022.

Closes https://github.com/emilk/egui/issues/1125

clippy got a few new lints. for now, i added them to allow section of cranky.toml so that i can get this to compile for now. 
I could fix those lints, but i am afraid it will touch a lot of files. If its okay to fix those lints, i can create a follow up PR to this. 

I wanted to keep the noise low for this PR, so kept the changes as minimal as i could

### DIFF
--- a/crates/eframe/Cargo.toml
+++ b/crates/eframe/Cargo.toml
@@ -145,5 +145,5 @@ web-sys = { version = "0.3.58", features = [
 
 # optional web:
 egui-wgpu = { version = "0.19.0", path = "../egui-wgpu", optional = true } # if wgpu is used, use it without (!) winit
-tts = { version = "0.20", optional = true }                                # Can't use 0.21-0.24 due to compilation problems on linux
+tts = { version = "0.24", optional = true }                                # Can't use 0.21-0.24 due to compilation problems on linux
 wgpu = { version = "0.14", optional = true, features = ["webgl"] }
--- a/crates/eframe/Cargo.toml
+++ b/crates/eframe/Cargo.toml
@@ -145,5 +145,5 @@ web-sys = { version = "0.3.58", features = [
 
 # optional web:
 egui-wgpu = { version = "0.19.0", path = "../egui-wgpu", optional = true } # if wgpu is used, use it without (!) winit
-tts = { version = "0.24", optional = true }                                # Can't use 0.21-0.24 due to compilation problems on linux
+tts = { version = "0.24", optional = true }
 wgpu = { version = "0.14", optional = true, features = ["webgl"] }
--- a/crates/egui-winit/Cargo.toml
+++ b/crates/egui-winit/Cargo.toml
@@ -61,7 +61,7 @@ puffin = { version = "0.14", optional = true }
 serde = { version = "1.0", optional = true, features = ["derive"] }
 
 # feature screen_reader
-tts = { version = "0.24", optional = true } # Can't use 0.21-0.24 due to compilation problems on linux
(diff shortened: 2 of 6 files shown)


crates/egui-winit/Cargo.toml
```diff
@@ -61,7 +61,7 @@ puffin = { version = "0.14", optional = true }
 serde = { version = "1.0", optional = true, features = ["derive"] }
 
 # feature screen_reader
-tts = { version = "0.20", optional = true } # Can't use 0.21-0.24 due to compilation problems on linux
+tts = { version = "0.24", optional = true } # Can't use 0.21-0.24 due to compilation problems on linux
```
emilk, line 64:
```suggestion
tts = { version = "0.24", optional = true }
```

crates/eframe/Cargo.toml
```diff
@@ -145,5 +145,5 @@ web-sys = { version = "0.3.58", features = [
 
 # optional web:
 egui-wgpu = { version = "0.19.0", path = "../egui-wgpu", optional = true } # if wgpu is used, use it without (!) winit
-tts = { version = "0.20", optional = true }                                # Can't use 0.21-0.24 due to compilation problems on linux
+tts = { version = "0.24", optional = true }                                # Can't use 0.21-0.24 due to compilation problems on linux
```
emilk, line 148:
```suggestion
tts = { version = "0.24", optional = true }
```

emilk (commented):
This is great!

coderedart:
ooh.. i can just press the commit suggestion buttom and it will add your changes without me doing anything. never knew i could do that. nice.

emilk:
Let's do the clippy lints in a separate PR - I think that was a good choice

Issue #1802, opened 2022-07-05
eframe: window flashes white on launch on Windows


HideakiAtsuyo:
eframe = { git = "https://github.com/emilk/egui", branch = "master" }

**Describe the bug**
Wheb the window show it's kinda buggy it's light then dark

**To Reproduce**
A bit modified the example code with name and age
![](https://i.imgur.com/ohRGWRa.gif)

**Expected behavior**
Appear dark

**Screenshots**
See to reproduce for GIF

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser I don't use a browser
 - Version I'm still using the libs directly 

**Additional context**
None

emilk:
How annoying! On Mac it works well. Is it the same using `egui-wgpu` as the backend?

Perhaps add a few breakpoints and try to find out what goes wrong!

HideakiAtsuyo:
> How annoying! On Mac it works well. Is it the same using `egui-wgpu` as the backend?
> 
> Perhaps add a few breakpoints and try to find out what goes wrong!

Hello, it's not very annoying but I was thiking it would be good if it's possible to fix it ! And yeah even with using egui-wgpu I get the same result !

HideakiAtsuyo:
"Bump"

emilk:
@HideakiAtsuyo since this affects your computer and you are a programmer, maybe you can investigate?

HideakiAtsuyo:
> @HideakiAtsuyo since this affects your computer and you are a programmer, maybe you can investigate?

The only solution that I have for this situation that would not necessarely require to modify the crate is to hide the windows until it's fully loaded

HideakiAtsuyo:
+ I'm not a maintainer of egui 🤔

emilk:
You are not the maintainer, but you can become a contributor.

I have a thousand things to do that has a higher priority, so if you want this to be fixed, you better roll up your sleeves.

HideakiAtsuyo:
> You are not the maintainer, but you can become a contributor.
> 
> I have a thousand things to do that has a higher priority, so if you want this to be fixed, you better roll up your sleeves.

It's not a priority I never said that, I'm using Rust very rarely sadly it was just for 1 project nothing more

rlidwka:
when you create `glutin` context on windows, it creates white window, which stays white until content gets swapped into it

when you create `glutin` context on linux, it creates a hidden window (icon in taskbar, but no ability to click into it)

[...]

PR #2279, opened 2022-11-11
eframe: Don't show window until after initialization

LoganDark:
Shortens #1802, but does not completely solve it. This looks to be a step in the right direction, though, and eframe should have been doing this anyway.

Marked as draft because:

- I have not run cranky since I don't have it installed
- This could be done better, for example by only showing the window after the first frame has already been painted
- Ultimately I just wanted to submit this as a first step, even if it's not yet perfect.

### DIFF
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -146,7 +146,7 @@ pub fn handle_app_output(
         fullscreen,
         drag_window,
         window_pos,
-        visible: _,
+        visible: _, // handled in post_present
         always_on_top,
     } = app_output;
 
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -324,10 +324,8 @@ impl EpiIntegration {
             if app_output.close {
                 self.close = app.on_close_event();
             }
-            let visible = app_output.visible;
+            self.frame.output.visible = app_output.visible; // this is handled by post_present
             handle_app_output(window, self.egui_ctx.pixels_per_point(), app_output);
-            // TODO(logandark): find a nicer way to do this
-            self.frame.output.visible = visible;
         }
 
         let frame_time = (std::time::Instant::now() - frame_start).as_secs_f64() as f32;
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -146,7 +146,7 @@ pub fn handle_app_output(
         fullscreen,
         drag_window,
         window_pos,
-        visible,
+        visible: _,
         always_on_top,
     } = app_output;
 
@@ -181,10 +181,6 @@ pub fn handle_app_output(
 
     if drag_window {
         let _ = window.drag_window();
-    }
-
-    if let Some(visible) = visible {
-        window.set_visible(visible);
     }
 
     if let Some(always_on_top) = always_on_top {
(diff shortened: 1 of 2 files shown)


emilk:
> I have not run cranky since I don't have it installed

`cargo install cranky`

emilk:
Thanks for working on this - this seems like a promising avenue to explore!

LoganDark:
> > I have not run cranky since I don't have it installed
> 
> `cargo install cranky`

I think it's actually `cargo install cargo-cranky`.

Alright, it passes, and I also ran `cargo fmt` (which I previously forgot).

> Thanks for working on this - this seems like a promising avenue to explore!

I'm going to see if I can delay this until the first frame is already painted. That could eliminate the white flash altogether. A lot of the overhead seems to be glutin context creation which takes 500ms at best (iGPU) or 779ms at worst (dGPU).

LoganDark:
The window being shown is now properly delayed until the first frame is *naturally* rendered. The rendering is not forced / done prematurely, either.

This also makes it possible to do something that you could not do before: keep the window completely hidden by calling `Frame.set_visible(false)` on the first update, in which case it will not be shown at any point.

There is still a very very brief (1 frame?) white flash. Maybe we could show the app on the second frame instead :^).

Marking as ready for review.

crates/eframe/src/native/epi_integration.rs
```diff
@@ -341,6 +343,12 @@ impl EpiIntegration {
         app.post_rendering(window_size_px, &self.frame);
     }
 
+    pub fn post_present(&mut self, window: &winit::window::Window) {
+        if let Some(visible) = self.frame.output.visible.take() {
+           window.set_visible(visible);
```
emilk, line 348:
```suggestion
        if let Some(visible) = self.frame.output.visible.take() {
            window.set_visible(visible);
```

crates/eframe/src/native/run.rs
```diff
@@ -330,8 +330,9 @@ mod glow_integration {
             };
             let window_settings = epi_integration::load_window_settings(storage);
 
-            let window_builder =
-                epi_integration::window_builder(native_options, &window_settings).with_title(title);
+            let window_builder = epi_integration::window_builder(native_options, &window_settings)
+                .with_title(title)
+                .with_visible(false);
```
emilk, line 335:
```suggestion
                .with_visible(false); // Keep hidden until we painted something. See https://github.com/emilk/egui/pull/2279
```

crates/eframe/src/native/run.rs
```diff
@@ -692,6 +695,7 @@ mod wgpu_integration {
             let window_settings = epi_integration::load_window_settings(storage);
             epi_integration::window_builder(native_options, &window_settings)
                 .with_title(title)
+                .with_visible(false)
```
emilk, line 698:
```suggestion
                .with_visible(false) // Keep hidden until we painted something. See https://github.com/emilk/egui/pull/2279
```

crates/eframe/src/native/epi_integration.rs
```diff
@@ -146,7 +146,7 @@ pub fn handle_app_output(
         fullscreen,
         drag_window,
         window_pos,
-        visible,
+        visible: _,
```
emilk, line 149:
```suggestion
        visible: _, // handled in post_present
```

crates/eframe/src/native/epi_integration.rs
```diff
@@ -325,7 +324,10 @@ impl EpiIntegration {
             if app_output.close {
                 self.close = app.on_close_event();
             }
+            let visible = app_output.visible;
             handle_app_output(window, self.egui_ctx.pixels_per_point(), app_output);
+            // TODO(logandark): find a nicer way to do this
+            self.frame.output.visible = visible;
```
emilk, line 330:
```suggestion
            self.frame.output.visible = app_output.visible; // this is handled by post_present
            handle_app_output(window, self.egui_ctx.pixels_per_point(), app_output);
```

emilk (approved):

PR #2280, opened 2022-11-11
eframe: Repaint immediately on RepaintAsap, fixes #903

### LINKED ISSUE #903: UI jittering/flickering on fast window resize
**Describe the bug**
When the window is resized at high speed, the UI starts to flicker/jitter.

**To Reproduce**
1. Run the eframe example.
2. Resize the window fast.

**Expected behavior**
For the elements of the UI to stay put during resizing

**Screenshots**
Here's a quick screen recording displaying the issue:

https://user-images.githubusercontent.com/25753364/142637840-cae07e3d-e756-4e67-a513-324244b4de72.mov

 - OS: MacOS Montery 12.0.1
 - M1 Max SoC - if that's applicable at all

LoganDark:
This completely eliminates the white flickering seen on Windows when rapidly resizing a window on the glow backend. The reason that happens is because DWM only waits for the resize event to be delivered before displaying the window at its new size. You must repaint synchronously inside that iteration of the event loop or else you get flickering.

Draft because there may be another reason why it was done this way; there was no comment explaining it so I am unsure.

### DIFF
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -26,12 +26,12 @@ enum EventResult {
     /// be used in special situations if the window must be repainted while
     /// handling a specific event. This occurs on Windows when handling resizes.
     ///
-    /// `RepaintAsap` introduces a one-frame delay, and should therefore only be
+    /// `RepaintNow` introduces a one-frame delay, and should therefore only be
     /// used for extremely urgent repaints.
-    RepaintAsap,
+    RepaintNow,
     /// Queues a repaint for once the event loop handles its next redraw. Exists
     /// so that multiple input events can be handled in one frame. Does not
-    /// cause any delay like `RepaintAsap`.
+    /// cause any delay like `RepaintNow`.
     RepaintNext,
     RepaintAt(Instant),
     Exit,
@@ -129,7 +129,7 @@ fn run_and_return(event_loop: &mut EventLoop<RequestRepaintEvent>, mut winit_app
 
         match event_result {
             EventResult::Wait => {}
-            EventResult::RepaintAsap => {
+            EventResult::RepaintNow => {
                 tracing::trace!("Repaint caused by winit::Event: {:?}", event);
                 next_repaint_time = Instant::now() + Duration::from_secs(1_000_000_000);
                 winit_app.paint();
@@ -210,7 +210,7 @@ fn run_and_exit(
 
         match event_result {
             EventResult::Wait => {}
-            EventResult::RepaintAsap => {
+            EventResult::RepaintNow => {
                 next_repaint_time = Instant::now() + Duration::from_secs(1_000_000_000);
                 winit_app.paint();
             }
@@ -557,7 +557,7 @@ mod glow_integration {
                     if self.running.is_none() {
                         self.init_run_state(event_loop);
                     }
-                    EventResult::RepaintAsap
+                    EventResult::RepaintNow
                 }
                 winit::event::Event::Suspended => {
                     #[cfg(target_os = "android")]
@@ -630,7 +630,7 @@ mod glow_integration {
                             EventResult::Exit
                         } else if event_response.repaint {
                             if repaint_asap {
-                                EventResult::RepaintAsap
+                                EventResult::RepaintNow
                             } else {
                                 EventResult::RepaintNext
                             }
@@ -954,7 +954,7 @@ mod wgpu_integration {
                         );
                         self.init_run_state(event_loop, storage, window);
                     }
-                    EventResult::RepaintAsap
+                    EventResult::RepaintNow
                 }
                 winit::event::Event::Suspended => {
                     #[cfg(target_os = "android")]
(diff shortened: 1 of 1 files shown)


crates/eframe/src/native/run.rs
```diff
@@ -23,6 +23,7 @@ pub use epi::NativeOptions;
 enum EventResult {
     Wait,
     RepaintAsap,
+    RepaintNext,
```
Mingun, line 26: > there was no comment explaining it so I am unsure.

...and you continue this bad practice by does not including comment that describe the difference from `RepaintAsap` :smile:

LoganDark: ...whoops. :D

I shall write some doc comments about it, commit incoming soon.

crates/eframe/src/native/run.rs
```diff
@@ -22,7 +22,17 @@ pub use epi::NativeOptions;
 #[derive(Debug)]
 enum EventResult {
     Wait,
+    /// Causes a synchronous repaint inside the event handler. This should only
+    /// be used in special situations if the window must be repainted while
+    /// handling a specific event. This occurs on Windows when handling resizes.
+    ///
+    /// `RepaintAsap` introduces a one-frame delay, and should therefore only be
+    /// used for extremely urgent repaints.
     RepaintAsap,
```
emilk, line 31: with this change, maybe `RepaintNow` is a more apt name?

emilk (commented):
Interesting!

On my Mac I notice no difference. In particular, https://github.com/emilk/egui/issues/903 is not solved.

But if this solves the white flickering on Windows, that's a big win!

crates/eframe/src/native/run.rs
Mingun: According to the doc, wouldn't be `RepaintNextFrame` the best?

crates/eframe/src/native/run.rs
```diff
@@ -22,7 +22,17 @@ pub use epi::NativeOptions;
 #[derive(Debug)]
 enum EventResult {
     Wait,
+    /// Causes a synchronous repaint inside the event handler. This should only
+    /// be used in special situations if the window must be repainted while
+    /// handling a specific event. This occurs on Windows when handling resizes.
+    ///
+    /// `RepaintAsap` introduces a one-frame delay, and should therefore only be
```
Mingun, line 29: Isn't that sentence contradicts the intention of this variant? If you paint immediately then where one-frame delay can get from? From my understanding of the situation, the delay would be in `RequestNext`, no?

crates/eframe/src/native/run.rs
Mingun: Or `RepaintThisFrame`, if that is what it's doing

LoganDark: @Mingun This isn't "repaint this frame", that's what RepaintNext does (multiple events are handled before the next frame). This is "repaint **now**". Synchronously, immediately, before the event loop continues. It holds up all subsequent events in the loop, so in effect, it produces a new frame.

crates/eframe/src/native/run.rs
LoganDark: By one-frame delay I mean `RepaintNow` (I renamed it) blocks the event loop for a frame. It delays further processing of events until the frame has been synchronously produced. This is normal and expected when it comes to resize events on Windows, since the compositor uses them to tell when the application has handled the resize. But as far as I can tell, no other event acts this way. (I used `RepaintNow` for `Resumed` events as well, but in hindsight that is probably not necessary.)

LoganDark:
> On my Mac I notice no difference. In particular, #903 is not solved.

Interesting. I also happen to have a Mac - would you like me to look into the stretching effect there?

Just beware that *that* is likely a `glutin` / `glow` issue. For example, even with this PR, if you boot up the `wgpu` backend on Windows, it will still stretch and wobble horribly. This isn't due to `eframe` not handling events correctly, this is due to the graphics context itself.

> But if this solves the white flickering on Windows, that's a big win!

It does, completely! To the point where resizing is perfectly smooth, take a look at this:

https://user-images.githubusercontent.com/4723091/201410347-43b20f1c-d63e-43f6-a9f3-262b32f5490a.mp4

That is *perfection*. I did not know it was even *possible* to do that. But, somehow, eframe does it now, and it's *amazing*.

LoganDark:
It looks like this patch works as intended (glow is now perfect, even though wgpu still looks like crap; at least it didn't get worse); marking as ready for review.

LoganDark:
oh wait, I didn't realize that's what was going on. *derp.*

GitHub is so unintuitive sometimes.

Really sorry for the spam

crates/eframe/src/native/run.rs
Mingun: Ah.. Then maybe rewrite this sentence, because it is a very unclear, that the time of current frame is just increased (so it takes a time that normally two frames take)

LoganDark: > it is a very unclear, that the time of current frame is just increased (so it takes a time that normally two frames take)

This doesn't increase the time of the current frame. This creates a brand new frame into the current iteration of the event loop.

emilk (approved):
Awesome - thanks!

Issue #889, opened 2021-11-08
Axis legends for Plots


jossriLDR:
I think there is no 'native' way to add x and y axis legends.
This is one of the most basic features of plotting and should be added (and also be demonstrated in the web demo).

Any suggestions how to achieve this (maybe even workarounds)?

Other than this, I can't figure out how to rotate labels. #428 refers to this but the mentioned commit 
a0f5d7e is in another fork of this repo.

JohannesProgrammiert:
## Question to the maintainers @emilk @nknotts 

I am currently working on the plot library. Labels and ticks generally work as expected. One remaining challenge is calculating the y-axis label width for the widget layout (marked blue in the picture):

![egui_legend](https://user-images.githubusercontent.com/80203331/201482904-af617cef-5a51-4e5a-adf8-71cfa0f420d2.png)

Part of the solution is that the x-axis label height is constant

### Idea

1. Layout the widget in y-direction
2. Make the plot grid in y-direction (needs screen-to-plot and plot-to-screen transformation in 1D)
3. Initialize the galleys for y-ticks and calculate maximum width
4. Do the remaining layout

However, I noticed transformation between plot space and screen space are rather limited and non-standard/'manual' in the current state. I know that [`euclid`](https://docs.rs/euclid/latest/euclid/) makes such transformations easy and has an intuitive API. On a quick glance, it even might be suitable to replace the `emath` library completely.

Implementing basic linear algebra myself is cumbersome and and is not the task I want to focus on now.

### My questions:

* why don't you use that or another widely known crate that provides the required functionality?
* may I add `euclid` as dependency to `egui` and refactor the plot-to-screen transformations in `eplot`?

nknotts:
@JohannesProgrammiert I am not a maintainer, but this is exactly what I am looking for! Great work.

Pardoxa:
I would really love to have this functionality as labeling your axis is kind of essential. So I'd like to +1 this feature request

jakkosdev:
@JohannesProgrammiert that looks great!

I'd love to have this functionality, so +1 too

wolro:
Thank you for your efforts! I am currently slowly introducing Rust / egui for easy-to-deploy data acquisition and monitoring tools at my workplace, and the lack of axis labels and (to a lesser degree) axis ticks outside the plot are giving me a hard time convincing my fellow Pythonistas :-)

Otherwise, egui + plot is just wonderful.

JohannesProgrammiert:
[...]

Issue #2245, opened 2022-11-06
Heatmap Widget for Plots


JohannesProgrammiert:
I did not find a feature request about heat maps. Heat map plotting is available in the [ImPlot](https://traineq.org/implot_demo/src/implot_demo.html) library. The related demo widget looks like this:

![implot_heatmap](https://user-images.githubusercontent.com/80203331/200167209-daab31da-a948-4747-863b-f0da62d50a47.png)

I started implementing a similar feature for the egui plot widget which currently looks like this:

[simplescreenrecorder.webm](https://user-images.githubusercontent.com/80203331/200167865-b68dddf9-a7cf-4ed3-acac-b7ffc593b96d.webm)

And the API do create such a thing is:

```rust
let mut vals = Vec::new();
const ROWS: usize = 16; // tiles y
const COLS: usize = ROWS; // tiles x
// populate values
for x in 0..COLS {
    for y in 0..ROWS {
        let r = f32::sqrt((x.pow(2) + y.pow(2)) as f32);
        vals.push(f64::sin(r as f64 + self.t as f64 / 100.0));
    }
}

const RESOLUTION: usize = 256; // number of color steps in interpolated color palette
let heatmap = egui::plot::Heatmap::<RESOLUTION>::new(vals, COLS)
    .unwrap()
    .show_labels(self.show_labels) // bool
    .palette(self.color_gradient.clone()) // Vec<Color32>
    .name("something");

Plot::new("plot")
    .legend(Legend::default())
    .allow_zoom(false)
    .allow_scroll(false)
    .allow_boxed_zoom(false)
    .allow_drag(false)
    .set_margin_fraction(Vec2::new(0.0, 0.0))
    .width(500.0)
    .view_aspect(1.0)
    .show(ui, |plot_ui| {
        plot_ui.heatmap(heatmap);
    });
```

## TODO

* axis legends (related to/depends on https://github.com/emilk/egui/issues/889). I also think axis labels and ticks should be **outside** of the plotting window, similar to ImPlot. ImPlot also has an option to move the legend out of the plotting window which I find convenient.
* Value to color mapping is sometimes weird, debugging needed
* Legend (similar to ImPlot, an extra window that shows value-to-color-mapping).

PR #2284, opened 2022-11-11
Draw axis labels and ticks outside of plotting window

JohannesProgrammiert:
* ⚠️ BREAKING: `egui::widgets::plot::Plot::show_axes` renamed to `egui::widgets::plot::Plot::show_grid`. The origin `show_axes` API now refers to the axis labels
* Added plot axis labels and the required API to customize them
* Added support for multiple x-axis- and y-axis-labels and -ticks per plot

---

All better known plotting libraries like gnuplot, matlab, and also Implot draw the ticks outside of the plotting window:

matlab_plot

imgui_plot

egui paints the ticks inside the plotting window, which is unusual and, in my opinion, inconvenient.
In the same region, axis labels should be drawn.

- [x] add a screenshot or gif.
- [x] add a line to the relevant `CHANGELOG.md` under "Unreleased".
- [x] consider adding a demo for it to `egui_demo_lib`.
- [x] run `cargo fmt` and `cargo clippy`.
- [x] run `./scripts/check.sh`.

* Blocks https://github.com/emilk/egui/issues/2245
* Closes https://github.com/emilk/egui/issues/889


JohannesProgrammiert:
Support for labels is done. I changed the API for `Plot` but further information will follow once I managed to move tick labels to the outside.

[egui_labels.webm](https://user-images.githubusercontent.com/80203331/201413956-496f8a1e-ee10-405f-8c40-03abcfa0f98d.webm)

JohannesProgrammiert:
multi_axis

nknotts:
I am testing out this branch. Would you like feedback or wait until you finish developement?

JohannesProgrammiert:
I would love some feedback if you have time.

However, somebody will have to review this again in 1-2 weeks because I plan to make the axis-API a bit more convenient.

nknotts:
Screen Shot 2022-11-15 at 2 43 34 PM

This has allowed me to make our engineering dashboard with good looking plots!

JohannesProgrammiert:
`./sh/check.sh` doesn't work for me.

```sh
❯ ./sh/check.sh
+ cargo install cargo-cranky
    Updating crates.io index
     Ignored package `cargo-cranky v0.3.0` is already installed, use --force to override
+ export 'RUSTFLAGS=--cfg=web_sys_unstable_apis -D warnings'
+ RUSTFLAGS='--cfg=web_sys_unstable_apis -D warnings'
+ export 'RUSTDOCFLAGS=-D warnings'
+ RUSTDOCFLAGS='-D warnings'
+ cargo check --all-targets
    Finished dev [optimized + debuginfo] target(s) in 0.30s
+ cargo check --all-targets --all-features
    Checking serde v1.0.143
    Checking tracing v0.1.36
    Checking zerocopy v0.6.1
   Compiling speech-dispatcher-sys v0.7.0
    Checking wgpu-hal v0.14.0
   Compiling zvariant_derive v3.6.0
   Compiling zbus_macros v2.3.2
    Checking resvg v0.23.0
    Checking smithay-clipboard v0.6.6
    Checking web-sys v0.3.60
    Checking sctk-adwaita v0.4.2
    Checking winit v0.27.2
    Checking half v1.8.2
error: failed to run custom build command for `speech-dispatcher-sys v0.7.0`

Caused by:
  process didn't exit successfully: `egui/target/debug/build/speech-dispatcher-sys-511605648ffbec9f/build-script-build` (exit status: 101)
  --- stdout
  cargo:rustc-link-lib=speechd

  --- stderr
  wrapper.h:1:10: fatal error: 'speech-dispatcher/libspeechd.h' file not found
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ClangDiagnostic("wrapper.h:1:10: fatal error: 'speech-dispatcher/libspeechd.h' file not found\n")', $CARGO_HOME/registry/src/github.com-1ecc6299db9ec823/speech-dispatcher-sys-0.7.0/build.rs:22:10
warning: build failed, waiting for other jobs to finish...  
```

Everything else seems ok.

emilk:
I'm (finally) planning on taking a proper look at this today!

I do like this feature, but one immediate problem I see is how much whitespace it uses up, especially on the X axis. Take a look at these before-and-after pictures from the demo:

**Before:**
Screen Shot 2022-12-04 at 10 21 53

**After:**
Screen Shot 2022-12-04 at 10 21 19

JohannesProgrammiert:
Thanks for your feedback.

> a thinner default width

Can be changed in the `AxisHints::Default` trait implementation.

> Is it configurable?

The width is configurable by `AxisHints::max_digits(usize)` for each `AxisHints` that is passed to `Plot::custom_y_axes()`. This is, however, more of a 'proof-of-concept'-feature because it does not yet account for different font sizes. It always assumes 12pt per digit.

I didn't figure out how to determine the size of one digit without drawing it first...

> one immediate problem I see is how much whitespace it uses up, especially on the X axis

In issue https://github.com/emilk/egui/issues/889 I discussed this and made a proposal for automatically calculated y-axis-width. I am not sure it'd pay off because this feature might introduce some weird edge case behavior and makes the widget layout computation unnecessarily complex.

> Perhaps we need to tweak the default. I see no need for the names here ("x" and "y") when they were not explicitly given.

I think plots *without* axis labels should be the exception, not the other way round. If someone really wants a plot without axis labels it should be specified explicitly. But if you are sure about removing labels by default I'll just do it.

emilk:
> I think plots without axis labels should be the exception, not the other way round. If someone really wants a plot without axis labels it should be specified explicitly. But if you are sure about removing labels by default I'll just do it.

I think this is a user problem. That is: users should name their axes, and we should encourage that (e.g. by always doing so in all examples), but "x" and "y" aren't _names_. If the user is plotting cpu usage over time, calling the axes "x" and "y" is probably worse than having no names at all.

nknotts:
I agree with @emilk. Axis labels should be optional, but should be promoted in all examples.

JohannesProgrammiert:
I had some time and rebased on current master and removed axis labels by default as suggested. I also added some more API improvements, e.g. an easy way to set the y-label width.

The demo app is updated accordingly. As you can see, the spacing is reduced:

linked_axes

This is again ready for review.

Malven31:
Hi there =)
I'm quite new to open source merge requests and all that, and quite noob about rust, so my question is for sure going to sound silly : is there any chance this MR gets merged ? I'm trying to compile this by my own (to use eframe with this fork of egui), and struggling like hell...
Sorry to bother !

jakkosdev:
@Malven31 It may be worth looking into https://github.com/Gip-Gip/egui-plotter while we wait for this to be merged. I haven't tried it myself yet, but it looks promising.

CHANGELOG.md
```diff
@@ -5,7 +5,14 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
...
 
 ## Unreleased
+* ⚠️ BREAKING: `egui::widgets::plot::Plot::show_axes` renamed to `egui::widgets::plot::Plot::show_grid`. The origin `show_axes` API now refers to the axis labels. ([#2284](https://github.com/emilk/egui/pull/2284))
 
+### Added
+* Added plot axis labels and the required API to customize them. ([#2284](https://github.com/emilk/egui/pull/2284))
+* Added support for multiple x-axis- and y-axis-labels and -ticks per plot.  ([#2284](https://github.com/emilk/egui/pull/2284))
+
+### Fixed
+* Draw plot ticks outside of plotting window. ([#2284](https://github.com/emilk/egui/pull/2284))
```
emilk, line 15: I've switch to generating the changelog on each release using `scripts/generate_changelog.py`.

Please add these details to the PR description instead!

```suggestion
```

crates/egui/src/widgets/plot/axis.rs
```diff
@@ -0,0 +1,297 @@
...
+            transform: None,
+            steps: Vec::new(),
+        }
+    }
+}
+
+impl<const AXIS: usize> Widget for AxisWidget<AXIS> {
+    fn ui(self, ui: &mut Ui) -> Response {
+        // --- add label ---
+        let response = ui.allocate_rect(self.rect, Sense::click_and_drag());
```
emilk, line 165: does the axis ever response to clicking and dragging, or should this be changed to `Sense::hover()` instead?

crates/egui/src/widgets/plot/axis.rs
```diff
@@ -0,0 +1,297 @@
...
+        if ui.is_rect_visible(response.rect) {
+            let visuals = ui.style().visuals.clone();
+            let text = self.hints.label;
+            let galley = text.into_galley(ui, Some(false), f32::INFINITY, TextStyle::Body);
+            let text_color = visuals
+                .override_text_color
+                .unwrap_or_else(|| ui.visuals().text_color());
+            let angle: f32 = match AXIS {
+                X_AXIS => 0.0,
+                Y_AXIS => -std::f32::consts::PI * 0.5,
```
emilk, line 175:
```suggestion
                Y_AXIS => -std::f32::consts::TAU * 0.25,
```

("a quarter turn" is easier to think about than "half of a half-turn")

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -1526,36 +1763,6 @@ impl PreparedPlot {
...
-                    text_pos[1 - axis] = text_pos[1 - axis]
-                        .at_most(transform.frame().max[1 - axis] - galley.size()[1 - axis] - 2.0)
-                        .at_least(transform.frame().min[1 - axis] + 1.0);
-
-                    shapes.push((Shape::galley(text_pos, galley), text_strength));
-                }
-            }
         }
 
         fn color_from_contrast(ui: &Ui, contrast: f32) -> Color32 {
```
emilk, line 1768: Is `color_from_contrast` now duplicated?

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -583,6 +565,107 @@ impl Plot {
         self
     }
 
+    /// Set the x axis label of the main x-axis
+    ///
+    /// If no x-axis has been specified so far a new one will be created.
+    pub fn x_axis_label(mut self, label: impl Into<WidgetText>) -> Self {
```
emilk, line 571: It would be nice to add an example of using these new functions to the doctest of `Plot` (perhaps as a new example)!

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -651,10 +737,148 @@ impl Plot {
...
+        // y|x|x|    Plot Window     |x|x|
+        //  |i|i|                    |i|i|
+        //  |s|s|                    |s|s|
+        //  |1|0|                    |2|3|
+        //  +-+-+--------------------+-+-+
+        //      |      x-Axis 0      |   |
+        //      +--------------------+   a
+        //      |      x-Axis 1      |   |
+        //  +   +--------------------+-d-+
+        //
```
emilk, line 765: Very nice ASCII art!

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -651,10 +737,148 @@ impl Plot {
...
+
+        let mut plot_rect: Rect = {
+            // find dimensions of axis labels
+            // for a, b, c, d meanings see picture
+            let mut a = 0.0;
+            let mut b = 0.0;
+            let mut c = 0.0;
+            let mut d = 0.0;
+            if show_axes.x {
```
emilk, line 774: isn't it a bit superfluous with the `show_axes` setting? Why would the user add axis if they don't want to show them?

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -651,10 +737,148 @@ impl Plot {
...
+        //  +   +--------------------+-d-+
+        //
+
+        let mut plot_rect: Rect = {
+            // find dimensions of axis labels
+            // for a, b, c, d meanings see picture
+            let mut a = 0.0;
+            let mut b = 0.0;
+            let mut c = 0.0;
+            let mut d = 0.0;
```
emilk, line 773: These names are not very descriptive. `bottom`, `left`, `top`, and `right` would be much better names. You can use `egui::Margin` for this: https://docs.rs/egui/latest/egui/style/struct.Margin.html

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -651,10 +737,148 @@ impl Plot {
...
+            Rect {
+                min: complete_rect.min + Vec2::new(b, c),
+                max: complete_rect.max - Vec2::new(d, a),
+            }
+        };
+        let mut x_axis_widgets = Vec::<XAxisWidget>::new();
+        let mut y_axis_widgets = Vec::<YAxisWidget>::new();
+        {
+            // determine absolute rectangle for each axis label widget
+            // widget cnt per border of plot in order left, top, right, bottom
```
emilk, line 809: "cnt"?

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -651,10 +737,148 @@ impl Plot {
...
+            struct WidgetCnt {
```
emilk, line 810:
```suggestion
            struct NumWidgets {
```

is a name more consistent with the rest of the code base

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -651,10 +737,148 @@ impl Plot {
...
+                            Rect {
+                                min: plot_rect.right_top() + size_x * off,
+                                max: plot_rect.right_bottom() + size_x * (off + 1.0),
+                            }
+                        }
+                    };
+                    y_axis_widgets.push(YAxisWidget::new(cfg.clone(), rect));
+                }
+            }
+        }
```
emilk, line 870: The `show_dyn` functions is already way too long. Can you try to move out some of this code into helper function so as not to increase the complexity of `show_dyn` further?

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -651,10 +737,148 @@ impl Plot {
...
 
-        // Allocate the space.
-        let (rect, response) = ui.allocate_exact_size(size, Sense::drag());
+        // if to little space, remove axis widgets
```
emilk, line 872:
```suggestion
        // If too little space, remove axis widgets
```

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -947,6 +1171,39 @@ impl Plot {
...
+            let input = GridInput {
+                bounds: (bounds.min[Y_AXIS], bounds.max[Y_AXIS]),
+                base_step_size: transform.dvalue_dpos()[Y_AXIS] * MIN_LINE_SPACING_IN_POINTS * 2.0,
+            };
+            (grid_spacers[Y_AXIS])(input)
+        };
+        for mut widget in x_axis_widgets {
+            widget.range = x_axis_range.clone();
+            widget.transform = Some(transform);
+            widget.steps = x_steps.clone();
```
emilk, line 1197: This clones a `Vec<GridMark>` for each widget, which is a bit costly. If you wrap it in an `Arc` we can save some CPU time here!

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -1335,16 +1593,11 @@ impl PreparedPlot {
...
-                self.paint_axis(
-                    ui,
-                    d,
-                    self.show_axes[1 - d],
-                    &mut axes_shapes,
-                    self.sharp_grid_lines,
-                );
-            }
+        if self.show_grid.x {
+            self.paint_grid::<{ X_AXIS }>(ui, &mut axes_shapes);
```
emilk, line 1597: What weird syntax is this? O.o

Will this work: ?
```suggestion
            self.paint_grid::<X_AXIS>(ui, &mut axes_shapes);
```

crates/egui/src/widgets/plot/axis.rs
```diff
@@ -0,0 +1,297 @@
...
+pub enum Placement {
+    Default,
+    Opposite,
+}
+
+// shorthand types for AxisHints, public API
+/// Configuration for x-axis
+pub type XAxisHints = AxisHints<X_AXIS>;
+/// Configuration for y-axis
+pub type YAxisHints = AxisHints<Y_AXIS>;
```
emilk, line 33: Using const generics like for axis is a bit weird, and will cause code bloat due to monomorpization (basically duplicating all the generated code, increasing e.g. `.wasm` sizes). It would be nice if it could be avoided by passing in the `axis` as an argument/member instead.

emilk (changes requested):
It looks and works great! I left some comments on the code though, but nothing major

wlwatkins:
This is great work, but i'm confused. Is it possible to keep the axis in old fashion way? it's quite useful for highly compact multiplots.

Issue #2286, opened 2022-11-11
egui-glow fails cargo check --all-features 


coderedart:
When i tried to run `./sh/check.sh`, it fails at the step `cargo check --all-features` for the egui_glow crate. 

apparently, winit requires one of  x11 or wayland features to be enabled. because the `default-features` is false for the egui-winit dep in egui_glow for a long long time. so, its either a recent winit change or nobody ever checked `egui_glow` crate on linux until now. 

**To Reproduce**
1. Go to `crates/egui_glow`
2. run `cargo check --all-features`

**Output** 
```
Checking winit v0.27.2
error: Please select a feature to build for unix: `x11`, `wayland`
  --> /home/red/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.27.2/src/platform_impl/linux/mod.rs:10:1
   |
10 | compile_error!("Please select a feature to build for unix: `x11`, `wayland`");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

   Compiling wayland-scanner v0.29.4
error[E0392]: parameter `T` is never used
  --> /home/red/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.27.2/src/event_loop.rs:35:22
   |
35 | pub struct EventLoop<T: 'static> {
   |                      ^ unused parameter
   |
   = help: consider removing `T`, referring to it in a field, or using a marker such as `PhantomData`

error[E0392]: parameter `T` is never used
```

**Desktop (please complete the following information):**
 - OS: Arch Linux

PR #2287, opened 2022-11-11
enable winit default features so that egui_glow can pass cargo check

coderedart:
we can either fix this by exposing the x11/winit features to users of `egui_glow` or just enable the default features of `winit`. I went with the second option for now. 

Closes https://github.com/emilk/egui/issues/2286


emilk:
Hum. In https://github.com/emilk/egui/pull/1971 we disabled the default features to not force any default features onto users.

coderedart:
should i instead expose x11/wayland features in egui_glow -> egui-winit?
or just remove the `cargo check --all-features` for egui_glow crate?

sorry for late reply.

PR #2292, opened 2022-11-12
eframe: add set_minimized and set_maximized

SunDoge:
This PR adds some window actions to enable the possibility to create client-side decorations in egui applications.

window_actions


examples/custom_window_frame/src/main.rs
```diff
@@ -104,6 +98,38 @@ fn custom_window_frame(
...
+            if close_response.clicked() {
+                frame.close();
+            }
+
+            let minimized_response = ui.put(
+                Rect::from_min_size(
+                    rect.left_top() + vec2((height - 4.0) * 1.0, 0.0),
+                    Vec2::splat(height),
+                ),
+                Button::new(RichText::new("--").size(height - 4.0)).frame(false),
```
emilk, line 115: The default egui fonts have icons for minimize and maximize:

```suggestion
                Button::new(RichText::new("🗕").size(height - 4.0)).frame(false),
```

examples/custom_window_frame/src/main.rs
```diff
@@ -104,6 +98,38 @@ fn custom_window_frame(
...
+            if minimized_response.clicked() {
+                frame.set_minimized(true);
+            }
+
+            let maximized_response = ui.put(
+                Rect::from_min_size(
+                    rect.left_top() + vec2((height - 4.0) * 2.0, 0.0),
+                    Vec2::splat(height),
+                ),
+                Button::new(RichText::new("O").size(height - 4.0)).frame(false),
```
emilk, line 126:
```suggestion
                Button::new(RichText::new("🗖").size(height - 4.0)).frame(false),
```

examples/custom_window_frame/src/main.rs
```diff
@@ -104,6 +98,38 @@ fn custom_window_frame(
...
+            );
+            if maximized_response.clicked() {
+                app.maximized = !app.maximized;
```
emilk, line 129: it would be even better if we could add `maximized` to `eframe::WindowInfo`. You can fill it in with https://docs.rs/winit/0.27.5/winit/window/struct.Window.html#method.is_maximized

emilk (commented):
This is really cool, thanks for working on this!

SunDoge:
The cranky step should pass now.

emilk (commented):
On my Mac, this PR behaves very badly. The titlebar is back and I cannot move the window 😅 

Screen Shot 2022-11-16 at 11 28 05

SunDoge:
Is this problem caused by this PR?

emilk:
Yeah, I looked a bit closer. Apearently asking `window.is_maximized()` is what is causing this. 🤦

And there is no winit event for maximize state changing.

yunfengwangluo:
When learning EGUI, I found that when I click the maximize button, the window in Windows will ignore the position of the task bar and directly enter the full screen state In ubuntu, the taskbar can be explicitly displayed normally

Under normal circumstances, clicking Maximize should only maximize the window, and should not enter the full screen state!

SunDoge:
Hi @emilk, could you please check if this can be fixed by maintaining the window maximized state in the app?

SunDoge:
Hi @yunfengwangluo, could you show the command you run? I haven't witnessed this behavior.

emilk:
It seems `winit` is quite buggy on Mac. Minimize work, but maximize makes the app hang on the next call to `window.fullscreen()` (i.e. asking if it is fullscreen)

emilk:
So the main winit problem was fixed by https://github.com/rust-windowing/winit/pull/2636, but we need to wait for a new winit release, but that's coming pretty soon:
* https://github.com/rust-windowing/winit/pull/2634

emilk:
It now works well on my mac (with the winit 0.28 update).

Update the `min_window_size` and `initial_window_size` in `examples/custom_window_frame/src/main.rs` to avoid this:

image

Issue #167, opened 2021-02-08
Accessibility (A11y)


nvzqz:
**Is your feature request related to a problem? Please describe.**
Not all GUI app users are sighted or can see well. This project does not indicate any effort to make apps accessible to the visually impaired.

**Describe the solution you'd like**
For this project to make a conscious and focused effort to support apps that can be used by the visually impaired.

**Describe alternatives you've considered**
None.

**Additional context**
I (and many others) will not use a GUI library in production unless it is designed with accessibility in mind.

emilk:
From the top of my head, here are some tasks to improving accessibility:

1) A high-contrast, large text visual theme
2) Output data necessary for screen readers, braille displays and similar tools
3) Hook up such data in `egui_web` (and maybe `egui_glium`)

The accessibility data needs structure ("there is a window here with a scroll area") and semantics ("this is a heading, this is a checkbox"). What are the good, open standards for such accessibility data?

Is there an accessibility API for the web that doesn't require `equi_web` to create a fake DOM that mirrors the GUI?

I'd appreciate some help with this as I don't have any experience here!

follower:
@emilk There's some potentially helpful content in this Bevy issue:

 * Starting from ~ https://github.com/bevyengine/bevy/issues/254#issuecomment-678642730
 * But particularly: https://github.com/bevyengine/bevy/issues/254#issuecomment-679359772 (from developer of https://github.com/lightsoutgames/godot-accessibility).

emilk:
Thanks for those links @follower !

follower:
I'd hoped I might've been able to find some Free/Open Source Software developer specific accessibility tools/guidelines[0] but didn't manage to find much from a brief search.

> [0] I wonder if there's one or more companies that might want to support such an effort? \*cough\* :D

There's a reasonable amount of coverage of WWW/HTML/DOM aspects but less so for Canvas/desktop applications. (There's some older somewhat "enterprisey" content but it's a little difficult to separate the wheat from the seo content marketing...)

A couple of links that might still have some applicability:

 * https://wiki.gnome.org/Accessibility

 * https://webaim.org/standards/wcag/checklist

### Additional thoughts

There definitely seems to be *opportunity* for egui to have accessibility as a strong part of its story which has the potential to be compelling for both legally mandated & philosophical reasons--and seems consistent with the values of the Rust ecosystem itself.

Of course, if it were easy everyone would be doing it, I guess...

[...]

PR #2294, opened 2022-11-12
Implement accessibility APIs via AccessKit

mwcampbell:
This PR makes egui accessible with assistive technologies such as screen readers. In contrast with egui's existing built-in screen reader, this PR uses AccessKit to implement platform accessibility APIs that are used by existing screen readers and other assistive technologies.

AccessKit is only implemented for Windows so far. I will focus on a Mac implementation over the next week or two.

We should discuss what level of completeness this branch needs to reach before being merged upstream. I haven't yet implemented support for all egui widgets. In general, this branch supports the widgets that were already accessible via `WidgetInfo`.

Closes <https://github.com/emilk/egui/issues/167>.

### DIFF
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -272,8 +272,17 @@ impl EpiIntegration {
         window: &winit::window::Window,
         event_loop_proxy: winit::event_loop::EventLoopProxy<E>,
     ) {
+        let egui_ctx = self.egui_ctx.clone();
         self.egui_winit
-            .init_accesskit(window, event_loop_proxy, self.egui_ctx.clone());
+            .init_accesskit(window, event_loop_proxy, move || {
+                // This function is called when an accessibility client
+                // (e.g. screen reader) makes its first request. If we got here,
+                // we know that an accessibility tree is actually wanted.
+                egui_ctx.enable_accesskit();
+                // Enqueue a repaint so we'll receive a full tree update soon.
+                egui_ctx.request_repaint();
+                egui::accesskit_placeholder_tree_update()
+            });
     }
 
     pub fn warm_up(&mut self, app: &mut dyn epi::App, window: &winit::window::Window) {
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -3,6 +3,10 @@ use winit::event_loop::EventLoopWindowTarget;
 #[cfg(target_os = "macos")]
 use winit::platform::macos::WindowBuilderExtMacOS as _;
 
+#[cfg(feature = "accesskit")]
+use egui::accesskit;
+#[cfg(feature = "accesskit")]
+use egui_winit::accesskit_winit;
 use egui_winit::{native_pixels_per_point, EventResponse, WindowSettings};
 
 use crate::{epi, Theme, WindowInfo};
@@ -262,6 +266,15 @@ impl EpiIntegration {
         }
     }
 
+    #[cfg(feature = "accesskit")]
+    pub fn init_accesskit<E: From<accesskit_winit::ActionRequestEvent> + Send>(
+        &mut self,
+        window: &winit::window::Window,
+        event_loop_proxy: winit::event_loop::EventLoopProxy<E>,
+    ) {
+        self.egui_winit.init_accesskit(window, event_loop_proxy);
+    }
+
     pub fn warm_up(&mut self, app: &mut dyn epi::App, window: &winit::window::Window) {
         crate::profile_function!();
         let saved_memory: egui::Memory = self.egui_ctx.memory().clone();
@@ -299,6 +312,11 @@ impl EpiIntegration {
         }
 
         self.egui_winit.on_event(&self.egui_ctx, event)
+    }
+
+    #[cfg(feature = "accesskit")]
+    pub fn on_accesskit_action_request(&mut self, request: accesskit::ActionRequest) {
+        self.egui_winit.on_accesskit_action_request(request);
     }
 
     pub fn update(
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -272,7 +272,8 @@ impl EpiIntegration {
         window: &winit::window::Window,
         event_loop_proxy: winit::event_loop::EventLoopProxy<E>,
     ) {
-        self.egui_winit.init_accesskit(window, event_loop_proxy);
+        self.egui_winit
+            .init_accesskit(window, event_loop_proxy, self.egui_ctx.clone());
     }
 
     pub fn warm_up(&mut self, app: &mut dyn epi::App, window: &winit::window::Window) {
--- a/crates/eframe/src/web/backend.rs
+++ b/crates/eframe/src/web/backend.rs
@@ -400,6 +400,8 @@ impl AppRunner {
             events: _, // already handled
             mutable_text_under_cursor,
             text_cursor_pos,
+            #[cfg(feature = "accesskit")]
+                accesskit_update: _, // not currently implemented
         } = platform_output;
 
         set_cursor_icon(cursor_icon);
--- a/crates/egui/Cargo.toml
+++ b/crates/egui/Cargo.toml
@@ -57,7 +57,6 @@ serde = ["dep:serde", "epaint/serde", "accesskit?/serde"]
 [dependencies]
 epaint = { version = "0.19.0", path = "../epaint", default-features = false }
 
-accesskit = { version = "0.8.1", optional = true }
 ahash = { version = "0.8.1", default-features = false, features = [
     "no-rng", # we don't need DOS-protection, so we let users opt-in to it instead
     "std",
@@ -65,6 +64,10 @@ ahash = { version = "0.8.1", default-features = false, features = [
 nohash-hasher = "0.2"
 
 #! ### Optional dependencies
+## Exposes detailed accessibility implementation required by platform
+## accessibility APIs. Also requires support in the egui integration.
+accesskit = { version = "0.8.1", optional = true }
+
 ## Enable this when generating docs.
 document-features = { version = "0.2", optional = true }
 
--- a/crates/egui/Cargo.toml
+++ b/crates/egui/Cargo.toml
@@ -52,11 +52,12 @@ mint = ["epaint/mint"]
 persistence = ["serde", "epaint/serde", "ron"]
 
 ## Allow serialization using [`serde`](https://docs.rs/serde).
-serde = ["dep:serde", "epaint/serde"]
+serde = ["dep:serde", "epaint/serde", "accesskit?/serde"]
 
 [dependencies]
 epaint = { version = "0.19.0", path = "../epaint", default-features = false }
 
+accesskit = { version = "0.8.0", optional = true }
 ahash = { version = "0.8.1", default-features = false, features = [
     "no-rng", # we don't need DOS-protection, so we let users opt-in to it instead
     "std",
--- a/crates/egui/Cargo.toml
+++ b/crates/egui/Cargo.toml
@@ -57,7 +57,7 @@ serde = ["dep:serde", "epaint/serde", "accesskit?/serde"]
 [dependencies]
 epaint = { version = "0.19.0", path = "../epaint", default-features = false }
 
-accesskit = { version = "0.8.0", optional = true }
+accesskit = { version = "0.8.1", optional = true }
 ahash = { version = "0.8.1", default-features = false, features = [
     "no-rng", # we don't need DOS-protection, so we let users opt-in to it instead
     "std",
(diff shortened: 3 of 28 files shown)


crates/eframe/src/native/epi_integration.rs
```diff
@@ -75,7 +77,9 @@ pub fn window_builder(
         .with_maximized(*maximized)
         .with_resizable(*resizable)
         .with_transparent(*transparent)
-        .with_window_icon(window_icon);
+        .with_window_icon(window_icon)
+        // We must not show the window until AccessKit is initialized.
+        .with_visible(false);
```
emilk, line 82: Already done in https://github.com/emilk/egui/pull/2279, but it would be good with this comment with the added motivation

crates/egui/Cargo.toml
```diff
@@ -52,11 +52,12 @@ mint = ["epaint/mint"]
 persistence = ["serde", "epaint/serde", "ron"]
 
 ## Allow serialization using [`serde`](https://docs.rs/serde).
-serde = ["dep:serde", "epaint/serde"]
+serde = ["dep:serde", "epaint/serde", "accesskit/serde"]
```
emilk, line 55:
```suggestion
serde = ["dep:serde", "epaint/serde", "accesskit?/serde"]
```

mwcampbell:
@emilk Are there any particular milestones you want this work to reach before I take the PR out of draft status?

mwcampbell:
I wonder if I should make AccessKit an optional feature in egui-winit as it is in egui. To do that, I may need to refactor my changes to egui-winit, to initialize the AccessKit winit adapter separately. But by doing this, only eframe would be opinionated about whether to use AccessKit, and I suppose I'd be able to revert my modifications to egui_glium and egui_glow.

emilk:
@mwcampbell I'll take a look as soon as I find time. I have a very busy week, but I'll do my best to squeeze this in!

And thanks so much for working on this 🙏

mwcampbell:
I just modified accesskit_winit so it can be used without requiring an `EventLoopProxy` and a specific user event type. So I should be able to refactor my changes to eframe to get back to requiring only `RequestRepaintEvent`. The downside of such a change is that I'll have to put either `egui_winit::State` or its `egui_input` field behind a mutex, since an AccessKit `ActionHandler` implementation may be called from any thread. That might still be preferrable to the changes that I currently have to make to plumb the user event type that accesskit_winit used to require.

emilk:
> I just modified accesskit_winit so it can be used without requiring an EventLoopProxy and a specific user event type. So I should be able to refactor my changes to eframe to get back to requiring only RequestRepaintEvent

Even if you do, I think a `pub enum UserEvent {  RequestRepaint }` still makes sense.

mwcampbell:
I decided that putting the egui_winit state, or even just the `egui_input` field of that state, behind a mutex was just as invasive a change as the previous change I made to the user event type, so I decided not to bother with that. However, I did make AccessKit an optional feature in egui-winit, which allowed me to revert my previous changes to egui_glium and egui_glow. So now, only eframe is opinionated about using AccessKit.

I could make AccessKit an optional feature even in eframe, though enabled by default. On the one hand, I'm apprehensive about making it too easy to disable accessibility; I wouldn't want to make it easy for a developer to disable accessibility just because they're idly reducing the size of their executable or trying to improve hypothetical performance, leading to a situation where a disabled person can't use an application they need for their job. On the other hand, I don't want to be so dogmatic about forcing accessibility onto everyone that it leads to a backlash.

crates/egui/src/context.rs
```diff
@@ -436,16 +473,23 @@ impl Context {
...
+                response.fill_accesskit_node_common(&mut node);
+            }
+        }
+
         let clicked_elsewhere = response.clicked_elsewhere();
         let ctx_impl = &mut *self.write();
         let memory = &mut ctx_impl.memory;
         let input = &mut ctx_impl.input;
 
-        // We only want to focus labels if the screen reader is on.
```
emilk, line 444: I think we should keep this behavior. Tab to select the next button is what I ususally want from a GUI. Selecting labels is only interesting if the screen reader is on

emilk (commented):
This is looking really nice!

I would still like for `accesskit` to be an optional (but perhaps default) feature of `eframe`. Seems like it should be fairly easy.

I made [another PR](https://github.com/emilk/egui/pull/2311), introducing `UserEvent`, to reduce the diff of this PR and to lessen the chance for merge conflicts for others working in `eframe` in parallel.

crates/egui/src/context.rs
mwcampbell: Please see the corresponding change I made in the label widget. Short version: the label widget now adjusts its sense setting based on whether the built-in screen reader is on.

mwcampbell:
@emilk I rebased on your recent PR, and now AccessKit is an optional feature even in eframe. You seem unsure about whether it should be enabled by default in eframe. What are your concerns about enabling it by default?

To explain why I think accessibility should be the default at least in the highest-level framework, I'll share an anecdote: My best friend, who is totally blind, bought a piece of hardware that had a companion app to set it up. He couldn't independently use that app, because it was Qt-based, and it was shipped without Qt's optional accessibility support. So apparently the first person to find out that accessibility was missing was the end-user, and you can imagine the layers such a bug report would have to go through before the app developer would even look at it, let alone ship a fix. Now imagine that he had to use the app to do his job. That's why I think that requiring application developers to enable accessibility is a mistake. I can appreciate the reasons for disabling it by default in egui and even egui-winit; properly supporting accessibility in a custom egui integration requires some extra work. But in eframe, it should just work by default.

emilk:
@mwcampbell my only concern is build times and other potential build problems from extra dependencies, especially since AccessKit is currently only working on Windows, yet is pulled in on all platforms.

But, you make a very good point about it being enabled by default, so let's keep it like that!

mwcampbell:
I'm eager to minimize all possible barriers to implementing accessibility, so I'm willing to do work to minimize AccessKit's impact on build time, but not necessarily immediately. For now, I need to focus on getting it working at all, across platforms. BTW, I'm currently working on a macOS implementation, but don't yet have anything usable to show.

mwcampbell:
@emilk I need to write changelog entries for this PR. To acknowledge the funding that Google has provided for both this integration work and the development of AccessKit itself, I propose including something like the following in the entry in the main egui changelog:

Special thanks to the Google Fonts team for funding both this integration and the development of AccessKit itself.

Is this OK? I already cleared it with my main contact at Google.

mwcampbell:
I'm reconsidering my previous changes to `DragValue`. I now believe that automatically switching into edit mode when the widget receives keyboard focus is the correct behavior. But it means that because the button itself can never receive keyboard focus, the current code for incrementing and decrementing the value with the keyboard doesn't work. I'm going to address that in a separate issue and PR, then rebase this branch.

mwcampbell:
This is now blocked on #2342.

mwcampbell:
I've decided it's time to take this PR out of draft status. There are, of course, more widget types to support, and more things to polish. But I believe the foundation is now solid, and for the sake of keeping the scope of a single PR manageable, I won't do any more work on this one, except for review feedback.

I've done some more refactoring. Most notably, AccessKit support is now lazy. That is, an egui context now only starts generating AccessKit tress once the integration enables AccessKit support in that context. This way, the integration can defer AccessKit initialization until there is actually a request for accessibility info, e.g. from a screen reader. So, even when the AccessKit feature is enabled, the performance impact for users that don't need accessibility is negligible. The downside, at least for egui-winit and eframe, is that the integration has to return a placeholder tree at first, then push a real one on the next repaint. The relevant code in eframe uses `egui::Context::request_repaint` to request a repaint. In theory, a screen reader could decide how to read a window that has just appeared or received focus based on the (lack of) content in the placeholder tree. But so far, this isn't happening on either Windows or macOS. And there's precedent; Chromium's accessibility implementation has to return a placeholder tree for the web content area at first, then get a real one from the renderer process. I think lazy initialization is critical for wide acceptance of AccessKit, so I'm going to go with this compromise.

crates/egui/Cargo.toml
```diff
@@ -52,11 +52,12 @@ mint = ["epaint/mint"]
 persistence = ["serde", "epaint/serde", "ron"]
 
 ## Allow serialization using [`serde`](https://docs.rs/serde).
-serde = ["dep:serde", "epaint/serde"]
+serde = ["dep:serde", "epaint/serde", "accesskit?/serde"]
 
 [dependencies]
 epaint = { version = "0.19.0", path = "../epaint", default-features = false }
 
+accesskit = { version = "0.8.0", optional = true }
```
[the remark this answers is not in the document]
emilk: You can see the feature flags documented if you run `cargo doc --open --no-deps -p egui --all-features` and scroll down to "Feature flags"

crates/egui/src/context.rs
```diff
@@ -132,6 +154,24 @@ impl ContextImpl {
...
+        // insertion path also modifies another entry, to establish
+        // the parent/child relationship. Using `HashMap::entry` here
+        // would require two mutable borrows at once.
+        #[allow(clippy::map_entry)]
+        if !nodes.contains_key(&id) {
+            nodes.insert(id, Default::default());
+            let parent_id = state.parent_stack.last().unwrap();
+            let parent = nodes.get_mut(parent_id).unwrap();
+            parent.children.push(id.accesskit_id());
+        }
```
emilk, line 172:
```suggestion
        if let std::collections::hash_map::Entry::Vacant(entry) = nodes.entry(id) {
            entry.insert(Default::default());
            let parent_id = state.parent_stack.last().unwrap();
            let parent = nodes.get_mut(parent_id).unwrap();
            parent.children.push(id.accesskit_id());
        }
```

works for me!

crates/egui/src/context.rs
emilk: Oh I see. Why did you opt to handle that behavior in `Label` rather than here?

crates/egui/src/widgets/drag_value.rs
```diff
@@ -425,17 +449,18 @@ impl<'a> Widget for DragValue<'a> {
             }
         };
 
+        #[allow(clippy::redundant_clone)] // some clones below are dundant if AccessKit is disabled
```
emilk, line 452:
```suggestion
        #[allow(clippy::redundant_clone)] // some clones below are redundant if AccessKit is disabled
```

crates/egui/src/response.rs
```diff
@@ -526,8 +526,95 @@ impl Response {
...
+            node.checked_state = Some(if selected {
+                CheckedState::True
+            } else {
+                CheckedState::False
+            });
+        }
+    }
+
+    /// Associate a label with a control for accessibility.
+    pub fn labelled_by(self, id: Id) -> Self {
```
emilk, line 607: I think deserves a doc-example to show how it is supposed to be used

examples/hello_world/src/main.rs
```diff
@@ -33,10 +33,15 @@ impl eframe::App for MyApp {
...
-                ui.label("Your name: ");
-                ui.text_edit_singleline(&mut self.name);
+                let name_label = ui.label("Your name: ");
+                ui.text_edit_singleline(&mut self.name)
+                    .labelled_by(name_label.id);
             });
-            ui.add(egui::Slider::new(&mut self.age, 0..=120).text("age"));
+            ui.add(
+                egui::Slider::new(&mut self.age, 0..=120)
+                    .step_by(1.0)
```
emilk, line 42: hum… this should already be the default for all integer sliders, but I see that it isn't! The fix should be this:

```diff
--- a/crates/egui/src/widgets/slider.rs
+++ b/crates/egui/src/widgets/slider.rs
@@ -450,7 +450,7 @@ impl<'a> Slider<'a> {
     /// If you use one of the integer constructors (e.g. `Slider::i32`) this is called for you,
     /// but if you want to have a slider for picking integer values in an `Slider::f64`, use this.
     pub fn integer(self) -> Self {
-        self.fixed_decimals(0).smallest_positive(1.0)
+        self.fixed_decimals(0).smallest_positive(1.0).step_by(1.0)
     }
 
     fn get_value(&mut self) -> f64 {
```

emilk (changes requested):
This looks amazing - I only have a few small nits - fix those and I'll merge!

crates/egui/src/context.rs
mwcampbell: Since the label widget is the one special case, I thought it made sense to encapsulate the special logic in that widget. And this way, the AccessKit integration code can just look at `sense.focus` without having to know about the label special case. Does that make sense?

crates/eframe/src/web/backend.rs
```diff
@@ -400,6 +400,8 @@ impl AppRunner {
             events: _, // already handled
             mutable_text_under_cursor,
             text_cursor_pos,
+            #[cfg(feature = "accesskit")]
+                accesskit_update: _, // not currently implemented
```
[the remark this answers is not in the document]
mwcampbell: The current spacing is what rustfmt wants. If I change it, rustfmt just puts it back.

mwcampbell:
@emilk I think that covers everything. Thanks for taking time to review. Looking forward to getting this merged. Also, do you think it would be a good idea to publish egui 0.20 when this is merged?

emilk:
@mwcampbell merged! 🍾 🥳 

Amazing work!

Yes, I'm try to find time to publish `egui` 0.20 this week!

PR #2304, opened 2022-11-13
Improve mouse selection accuracy 

### LINKED ISSUE #1536: Click interactions aren't as a accurate as they can be
Referencing [this code](https://github.com/emilk/egui/commit/b8a4c9c8ce3228fd48d255d29be094b3edb4442f), it would be great if we could reduce the size of the gap and make the interact rect bounds more accurate. I was having issues were my clicks weren't being triggered on the edges and decided to reduce the 0.5 values to 0.01. Though I believe there probably should be a middle ground to that range.

arshiyasolei:
Closes #1536.


crates/egui/src/context.rs
```diff
@@ -328,11 +328,11 @@ impl Context {
         sense: Sense,
         enabled: bool,
     ) -> Response {
-        let gap = 0.5; // Just to make sure we don't accidentally hover two things at once (a small eps should be sufficient).
+        let gap = 0.1; // Just to make sure we don't accidentally hover two things at once (a small eps should be sufficient).
 
         // Make it easier to click things:
         let interact_rect = rect.expand2(
-            (0.5 * item_spacing - Vec2::splat(gap))
+            (0.1 * item_spacing - Vec2::splat(gap))
```
emilk, line 335: The point of expanding by half of the item spacing is to remove the dead space between items. The default `item_spacing.x` is 8, so we have 8 points of space between buttons. But it is nice to be able to click the four pixels to the left or right of a button and still hit the button.

In other words: your change here will make it much _harder_ to hit a button.

```suggestion
            (0.5 * item_spacing - Vec2::splat(gap))
```

emilk (changes requested):

crates/egui/src/context.rs
arshiyasolei: Ah I see 👍

PR #2316, opened 2022-11-17
[wgpu] Add depth stencil initialization to `Painter`

lsh:
I was getting panics in the `egui_demo_app` when using a depth buffer. To add the depth buffer I had modified:

```rust
// egui_demo_app/src/main.rs
    drag_and_drop_support: true,

    initial_window_size: Some([1280.0, 1024.0].into()),

    #[cfg(feature = "wgpu")]
    depth_buffer: 1, // <-- enable depth buffer

    ..Default::default()
};
```
and
```rust
// egui_demo_app/src/apps/custom3d_wgpu.rs
// in create_render_pipeline
depth_stencil: Some(wgpu::DepthStencilState {
    format: wgpu::TextureFormat::Depth32Float,
    depth_write_enabled: true,
    depth_compare: wgpu::CompareFunction::Less,
    stencil: wgpu::StencilState::default(),
    bias: wgpu::DepthBiasState::default(),
}),
```

I realized that because the first frame can render before the window is set, it is possible for `depth_buffer` to exist, while `depth_texture_view` to be `None`. This check stopped the crash for me, but if there is an issue with this solution I would be happy to refactor.

### DIFF
--- a/crates/egui-wgpu/CHANGELOG.md
+++ b/crates/egui-wgpu/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to the `egui-wgpu` integration will be noted in this file.
 * Only a single vertex & index buffer is now created and resized when necessary (previously, vertex/index buffers were allocated for every mesh) ([#2148](https://github.com/emilk/egui/pull/2148)).
 * `Renderer::update_texture` no longer creates a new `wgpu::Sampler` with every new texture ([#2198](https://github.com/emilk/egui/pull/2198))
 * `Painter`'s instance/device/adapter/surface creation is now configurable via `WgpuConfiguration` ([#2207](https://github.com/emilk/egui/pull/2207))
+* Fix panic on using a depth buffer ([#2316](https://github.com/emilk/egui/pull/2316))
 
 ## 0.19.0 - 2022-08-20
 * Enables deferred render + surface state initialization for Android ([#1634](https://github.com/emilk/egui/pull/1634)).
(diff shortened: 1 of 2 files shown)


Wumpf (changes requested):
putting this into `paint_and_update_textures` doesn't really make sense as we don't check for a configured surface every frame either. Also don't quite understand why you're doing `map().iter().for_each` on a single element - you can just map to the correct thing right away :).

Anyways, it would be best if `set_window` would do the same on `Some(window)` as `on_window_resized` does on `self.surface_state.is_some()`.

I.e. best to put 
```rust
            self.configure_surface(width_in_pixels, height_in_pixels);
            let device = &self.render_state.as_ref().unwrap().device;
            self.depth_texture_view = self.depth_format.map(|depth_format| {
                device
                    .create_texture(&wgpu::TextureDescriptor {
                        label: Some("egui_depth_texture"),
                        size: wgpu::Extent3d {
                            width: width_in_pixels,
                            height: height_in_pixels,
                            depth_or_array_layers: 1,
                        },
                        mip_level_count: 1,
                        sample_count: 1,
                        dimension: wgpu::TextureDimension::D2,
                        format: depth_format,
                        usage: wgpu::TextureUsages::RENDER_ATTACHMENT
                            | wgpu::TextureUsages::TEXTURE_BINDING,
                    })
                    .create_view(&wgpu::TextureViewDescriptor::default())
            });
```
into a function and call it at both locations. If that's unclear I can help out!

lsh:
@Wumpf thanks for explaining! Also feel free to suggest a better name for the texture view generator than `resize_and_generate_depth_texture_view`.

Wumpf (approved):
looks great and thanks for contributing!

I added a change-log entry myself since I forgot to mention that earlier (:

PR #2320, opened 2022-11-18
add set_plot_bounds method, giving users the ability to set the plot bounds themselves.

flxzt:
This PR adds the method `set_plot_bounds()` to give users more control over the bounds of the plot. Some requested this in discussion #1328 . To use this, the user needs to be able to create `PlotBounds` structs so I added a `PlotBounds::new()` method as well.  
To give an example for usecases: implementing a digital oscilloscope or a serial plotter where you have fixed ranges controlled from other parts in or outside of the app.  
This is sort of continuing #2145, which added the ability to translate the plot bounds by a delta.

I am not sure about adding this to the changelog? Let me know if I should add it, or if it is trivial and not worth mentioning.

Also: thank you for creating egui!

### DIFF
--- a/crates/egui/src/widgets/plot/transform.rs
+++ b/crates/egui/src/widgets/plot/transform.rs
@@ -18,6 +18,10 @@ impl PlotBounds {
         max: [-f64::INFINITY; 2],
     };
 
+    pub fn new(min: [f64; 2], max: [f64; 2]) -> Self {
+        Self { min, max }
+    }
+
     pub fn min(&self) -> [f64; 2] {
         self.min
     }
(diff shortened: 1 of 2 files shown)


crates/egui/src/widgets/plot/transform.rs
```diff
@@ -18,6 +18,10 @@ impl PlotBounds {
         max: [-f64::INFINITY; 2],
     };
 
+    pub fn new(min: [f64; 2], max: [f64; 2]) -> Self {
```
emilk, line 21:
```suggestion
    pub fn from_min_max(min: [f64; 2], max: [f64; 2]) -> Self {
```

emilk (approved):

PR #2322, opened 2022-11-19
Add an example project demonstrating animation

fgimian:
Hey there, I was struggling to find examples of doing animation with egui, especially when animating in two directions and using different times.

Thus I created a little project demonstrating the use of `animate_value_with_time` and `clear_animations` in case it's helpful to others getting started with egui.  This project also provides a little exponential curve function I found on Stack Exchange which applies a curve to the animation to make it more natural.

I didn't file an issue on this as it's just an example project and I had already finished it 😄 

Feel free to decline this PR if you don't wish to include the example.

Cheers
Fotis

### DIFF
--- a/examples/animation/README.md
+++ b/examples/animation/README.md
@@ -1,3 +1,9 @@
+This demo shows how to perform animation using `animate_value_with_time` to adjust the opacity of a rectangle over time.
+
+Also demonstrated here is how to use `clear_animations` and reset the animated value which is required to ensure smooth transitions if animation times differ for fade in and fade out.
+
+Click the **Animate** button to toggle the opacity of the rectangle with animation.
+
 ```sh
 cargo run -p animation
 ```
--- a/examples/animation/README.md
+++ b/examples/animation/README.md
@@ -1,0 +1,3 @@
+```sh
+cargo run -p animation
+```
--- a/examples/animation/src/main.rs
+++ b/examples/animation/src/main.rs
@@ -1,58 +1,34 @@
 #![cfg_attr(not(debug_assertions), windows_subsystem = "windows")] // hide console window on Windows in release
 
+mod scaled_color;
+
 use eframe::egui;
-use egui::{color, Color32, Id, Rect, Rounding};
-use std::ops::Range;
+use egui::{Color32, Id, Rect, Rounding};
+use scaled_color::ScaledColor32;
 
 fn main() {
     let options = eframe::NativeOptions::default();
     eframe::run_native(
         "My egui App",
         options,
-        Box::new(|cc| {
-            let app = MyApp::default();
-            // Set the initial opacity of the rectangle to write the current value to memory in
-            // the animation manager.
-            cc.egui_ctx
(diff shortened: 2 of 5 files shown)


examples/animation/README.md
```diff
@@ -0,0 +1,3 @@
+```sh
+cargo run -p animation
+```
```
Mingun, line 3: I think, that readme should contain a little more information about what that example is doing and what problem it solves

fgimian:
Just want to open one more point up for discussion.  The only reason I needed the curve is to counterart the effect of `linear_f32_from_gamma_u8` which causes the result not to be a smooth linear fade.

In my own little app, I used a helper like this instead which does a completely linear transform of RGB based on alpha:

```rust
fn apply_alpha(color: Color32, alpha: f32) -> Color32 {
    let r = color::linear_f32_from_linear_u8(color.r()) * alpha;
    let r = color::linear_u8_from_linear_f32(r);

    let g = color::linear_f32_from_linear_u8(color.g()) * alpha;
    let g = color::linear_u8_from_linear_f32(g);

    let b = color::linear_f32_from_linear_u8(color.b()) * alpha;
    let b = color::linear_u8_from_linear_f32(b);

    let a = color::linear_f32_from_linear_u8(color.a()) * alpha;
    let a = color::linear_u8_from_linear_f32(a);

    Color32::from_rgba_premultiplied(r, g, b, a)
}
```

I'm happy to take suggestions for how to better accomplish this in the example and remove the use of `fn curve`.

Cheers
Fotis

fgimian:
Rebased against `master` just now.  Would love your feedback on the way alpha is applied as per my comment above and I can then adjust this PR. 😄

examples/animation/src/main.rs
```diff
@@ -0,0 +1,126 @@
...
+    FadeOut,
+}
+
+struct MyApp {
+    rectangle_id: Id,
+    opacity_range: Range<f32>,
+    animate_direction: AnimateDirection,
+    fade_in_time: f32,
+    fade_out_time: f32,
+    current_opacity: f32,
```
emilk, line 35: The whole point of the egui animation system is so you don't need to keep this much state around. Really, for you example, you should only need a single `bool`.

The `toggle_switch.rs` example demonstrates this pretty well: https://github.com/emilk/egui/blob/master/crates/egui_demo_lib/src/demo/toggle_switch.rs#L49

fgimian:
@emilk Really appreciate your feedback as it helped me understand the animation system further in egui and simplify the code heavily.

I've pushed a new commit now which only leaves 4 fields in the app:

- `rectangle_id`: The id of the rectangle.  If there is a neater way to compute this, please let me know.
- `animate_direction`: The current animation direction.
- `fade_in_time` and `fade_out_time`: Independent fade in and fade out times which may be configured.  This is one of the main points I wanted to demonstrate as I didn't see it shown anywhere when I was learning egui.

Since you do already have examples of animation, feel free to close this without merging if you feel they are adequate.

Cheers
Fotis

PR #2334, opened 2022-11-22
Fix key pressed event

palako:
A key can be pressed or not pressed. This is the status of the key, not an event. The `InputState::keys_down` attribute contains a collection of any key being held down at any given frame.  `InputState::key_down(Key)` returns true if a key is being held down. So far so good.

A key being pressed or released is an event, i.e. the change from being pressed to not being pressed, or from not being pressed to being pressed. While` InputState::key_released(Key`) returns true correctly only when the key was released on a given frame,` InputState::key_pressed(Key)` doesn't behave that way, and instead returns true as long as the key is held down.

Besides not allowing this method to be used to detect the actual key press, another unintended consequence is that on the frame where the key is released, sometimes both events will be there, and thus both _key_pressed_ and _key_released_ will return true.

The example in the second commit of this PR helps seeing this. Before applying the patch to _InputState_, press and hold the letter A, then release. First you'll see that during the key being pressed, not only you get true from `key_down()` but also from `key_press()`. Also if you try to do this a few times, you will see that sometimes the message "pressed" is printed AFTER the message "released" (note there's no release messages in the following screenshot, they key is being held down)

Screenshot 2022-11-22 at 15 20 13

The root of the problem is the difference in meaning between pressed/released events in egui/eframe and egui-winit. For egui-winit, keyboard events mean something different, as implemented on State::on_keyboard_input. Keys pressed are added as Event::Key with pressed true (and this applies while the key is held down).

The patch on the first commit is very simple, and it just removes from the raw input events the key press if this key was already in the set of keys pressed from the previous frame. It of course leaves the keys_pressed set unaltered, so that the key can still be queried for being held down (there are only "held" messages in between a "pressed" and a "released" message in the following screenshot)

Screenshot 2022-11-22 at 15 13 43

The API to the user doesn't change, but with this fix key_pressed will change behaviour. If someone was using key_pressed to detect if a key was being held down, this will stop working. The fix would be trivial though, just changing the API call from key_pressed() to key_down(); the parameters are all the same.


crates/egui/src/input_state.rs
```diff
@@ -160,24 +164,26 @@ impl InputState {
...
-                }
-                Event::Scroll(delta) => {
-                    scroll_delta += *delta;
-                }
-                Event::Zoom(factor) => {
-                    zoom_factor_delta *= *factor;
+        new.events.retain(|event| match event {
+            Event::Key { key, pressed, .. } => {
+                if *pressed {
+                    //return false (thus filtering out the Event) if the key already existed in the set of pressed keys
```
emilk, line 170: rm dead code / confusing comment

emilk: Maybe
```suggestion
                    // We only retain presses that are novel (i.e. the first Press event, not those generated by key-repeat)
```

emilk (changes requested):
I tried to push an improvement to this PR but failed - perhaps you can `git cherrypick` it? 

https://github.com/emilk/egui/commit/a646c1259d8fcced4a281f42ca5de5ecfe0e56d1

palako:
I've pushed a commit with all the requested improvements. 
Not sure what the formatting problem was. When I tried to cargo format as you explain in the doc, I got cargo suggesting hundreds of format changes all over the place in files I was not touching, so I thought better to leave it alone. Let me know if you want me to adjust something.
Cheers!

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 * MSRV (Minimum Supported Rust Version) is now `1.65.0` ([#2314](https://github.com/emilk/egui/pull/2314)).
 * ⚠️ BREAKING: egui now expects integrations to do all color blending in gamma space ([#2071](https://github.com/emilk/egui/pull/2071)).
 * ⚠️ BREAKING: if you have overlapping interactive widgets, only the top widget (last added) will be interactive ([#2244](https://github.com/emilk/egui/pull/2244)).
+* Keyboard press events are only pressent at the frame when the key was pressed, consistent with how key releases work.
```
emilk, line 11:
```suggestion
* Keyboard press events are only present at the frame when the key was pressed, consistent with how key releases work ([#2334](https://github.com/emilk/egui/pull/2334)).
```

emilk:
Thanks!

PR #2335, opened 2022-11-23
[WGPU] Allow for depth buffer in web target

lsh:
Currently depth textures are not supported on web. This PR takes steps to allow a depth stencil. I'm expecting that I'll have to rename some parts and I might have to shift some things around. For example, I'm not sure if the depth texture option should be in `WebOptions`, as it is currently in the PR, or in `WgpuConfiguration`. Another consideration is if I should be moving this logic into some higher abstraction layer or not. Either way, would love feedback! cc @Wumpf

### DIFF
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -472,6 +472,10 @@ pub struct WebOptions {
     /// Configures wgpu instance/device/adapter/surface creation and renderloop.
     #[cfg(feature = "wgpu")]
     pub wgpu_options: egui_wgpu::WgpuConfiguration,
+
+    /// Configures if the render pass should use a depth buffer.
+    #[cfg(feature = "wgpu")]
+    pub depth_texture: bool,
 }
 
 #[cfg(target_arch = "wasm32")]
@@ -484,6 +488,8 @@ impl Default for WebOptions {
             #[cfg(feature = "glow")]
             webgl_context_option: WebGlContextOption::BestFirst,
 
+            #[cfg(feature = "wgpu")]
+            depth_texture: false,
             #[cfg(feature = "wgpu")]
             wgpu_options: egui_wgpu::WgpuConfiguration {
                 // WebGPU is not stable enough yet, use WebGL emulation
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -472,10 +472,6 @@ pub struct WebOptions {
     /// Configures wgpu instance/device/adapter/surface creation and renderloop.
     #[cfg(feature = "wgpu")]
     pub wgpu_options: egui_wgpu::WgpuConfiguration,
-
-    /// Configures if the render pass should use a depth buffer.
-    #[cfg(feature = "wgpu")]
-    pub depth_texture: bool,
 }
 
 #[cfg(target_arch = "wasm32")]
@@ -488,8 +484,6 @@ impl Default for WebOptions {
             #[cfg(feature = "glow")]
             webgl_context_option: WebGlContextOption::BestFirst,
 
-            #[cfg(feature = "wgpu")]
-            depth_texture: false,
             #[cfg(feature = "wgpu")]
             wgpu_options: egui_wgpu::WgpuConfiguration {
                 // WebGPU is not stable enough yet, use WebGL emulation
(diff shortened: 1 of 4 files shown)


crates/eframe/src/epi.rs
```diff
@@ -472,6 +472,10 @@ pub struct WebOptions {
     /// Configures wgpu instance/device/adapter/surface creation and renderloop.
     #[cfg(feature = "wgpu")]
     pub wgpu_options: egui_wgpu::WgpuConfiguration,
+
+    /// Configures if the render pass should use a depth buffer.
+    #[cfg(feature = "wgpu")]
+    pub depth_texture: bool,
```
emilk, line 478: maybe this should be in `WgpuConfiguration` since it only affects wgpu?

Wumpf: it would actually make much more sense to have a `Option<wgpu::TextureFormat>` passed in which then internally is constrained to be a depth format. That's a short comin of the `winit.rs` frontend of egui-wgpu right now.
That said, there is the slightly bigger task here to move the multisampling and depth_buffer options from native options over to something that is used both for web and native.

crates/eframe/src/web/web_painter_wgpu.rs
```diff
@@ -26,6 +28,30 @@ impl WebPainterWgpu {
...
+                            width: width_in_pixels,
+                            height: height_in_pixels,
+                            depth_or_array_layers: 1,
+                        },
+                        mip_level_count: 1,
+                        sample_count: 1,
+                        dimension: wgpu::TextureDimension::D2,
+                        format: depth_format,
+                        usage: wgpu::TextureUsages::RENDER_ATTACHMENT
+                            | wgpu::TextureUsages::TEXTURE_BINDING,
```
Wumpf, line 48: why is it a texture binding, render attachment only should be enough?

crates/eframe/src/web/web_painter_wgpu.rs
```diff
@@ -103,6 +132,7 @@ impl WebPainter for WebPainterWgpu {
         textures_delta: &egui::TexturesDelta,
     ) -> Result<(), JsValue> {
         let size_in_pixels = [self.canvas.width(), self.canvas.height()];
+        self.generate_depth_texture_view(size_in_pixels[0], size_in_pixels[1]);
```
Wumpf, line 135: doing this every frame is extremely wasteful. Creating a texture has major costs!
This should only be done on startup & resize. You actually can piggy-back on the 
```rust
            if size_in_pixels[0] != self.surface_configuration.width
                || size_in_pixels[1] != self.surface_configuration.height
                || size_in_pixels[1] != self.surface_configuration.height
            {
```
check further down which resizes the backbuffer

crates/eframe/src/web/web_painter_wgpu.rs
```diff
@@ -193,7 +223,16 @@ impl WebPainter for WebPainterWgpu {
                             store: true,
                         },
                     })],
-                    depth_stencil_attachment: None,
+                    depth_stencil_attachment: self.depth_texture_view.as_ref().map(|view| {
+                        wgpu::RenderPassDepthStencilAttachment {
+                            view,
+                            depth_ops: Some(wgpu::Operations {
+                                load: wgpu::LoadOp::Clear(1.0),
+                                store: true,
```
Wumpf, line 231: Quite common misconception :). You want to set `store` to false here since the texture is not read later anymore. This doesn't mean it is not written to as in no depth writes are happening, but rather that its contents can be discarded at the end of the renderpass. This means that if the gpu renders in a tile based fashion (the rule in mobile gpus, desktops are picking up as well!) it can optimize away writing to main gpu memory and keep the (potentially tiled) depth buffer only in on-chip memory!

```suggestion
                                store: false,
```

Wumpf (changes requested):

crates/eframe/src/epi.rs
lsh: Done in the latest commit. A minor papercut I ran into that's worth considering is `WgpuConfiguration::default()`. With the current setup, the following code will result in a mostly opaque error (due to using the wrong backend):
```rust
let options = WebOptions {
  wgpu_options: {
    depth_format: Some(wgpu::TextureFormat::Depth32Float),
    ..Default::default()
  },
  ..Default::default()
};
```

It might be worth having a separate `#[cfg(target = "wasm32")] impl Default for WgpuConfig`.

lsh:
@Wumpf thanks for mentoring on this PR!!! :)

Issue #2341, opened 2022-11-24
`DragValue` keyboard operation is broken


mwcampbell:
I observe these three related issues which I want to fix in one PR:

1. `DragValue` can't be incremented and decremented with the keyboard. In theory the code does this when the button form of the widget has focus, but the button never has focus.
2. When focus is moved to `DragValue` by tabbing, the widget is rendered in button mode for one frame, because egui doesn't know that the widget is interested in receiving focus until it is rendered as a button. This instability isn't good for screen readers.
3. If the `DragValue` widget is focused, then the user moves to an associated slider with Shift+Tab, adjusts the slider, then returns to the `DragValue`, the previous edit string is still there, and it overrides the change the user made to the slider.

PR #2342, opened 2022-11-24
Fix keyboard support in `DragValue`

mwcampbell:
Closes <https://github.com/emilk/egui/issues/2341>.

### DIFF
--- a/crates/egui/src/memory.rs
+++ b/crates/egui/src/memory.rs
@@ -403,7 +403,7 @@ impl Memory {
     /// This is normally done automatically when handling interactions,
     /// but it is sometimes useful to pre-register interest in focus,
     /// e.g. before deciding which type of underlying widget to use,
-    /// as in the `DragValue` widget, so a widget can be focused
+    /// as in the [`crate::DragValue`] widget, so a widget can be focused
     /// and rendered correctly in a single frame.
     #[inline(always)]
     pub fn interested_in_focus(&mut self, id: Id) {
--- a/crates/egui/src/memory.rs
+++ b/crates/egui/src/memory.rs
@@ -400,8 +400,13 @@ impl Memory {
 
     /// Register this widget as being interested in getting keyboard focus.
     /// This will allow the user to select it with tab and shift-tab.
+    /// This is normally done automatically when handling interactions,
+    /// but it is sometimes useful to pre-register interest in focus,
+    /// e.g. before deciding which type of underlying widget to use,
+    /// as in the `DragValue` widget, so a widget can be focused
+    /// and rendered correctly in a single frame.
     #[inline(always)]
-    pub(crate) fn interested_in_focus(&mut self, id: Id) {
+    pub fn interested_in_focus(&mut self, id: Id) {
         self.interaction.focus.interested_in_focus(id);
     }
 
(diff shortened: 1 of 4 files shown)


crates/egui/src/memory.rs
```diff
@@ -400,8 +400,13 @@ impl Memory {
 
     /// Register this widget as being interested in getting keyboard focus.
     /// This will allow the user to select it with tab and shift-tab.
+    /// This is normally done automatically when handling interactions,
+    /// but it is sometimes useful to pre-register interest in focus,
+    /// e.g. before deciding which type of underlying widget to use,
+    /// as in the `DragValue` widget, so a widget can be focused
```
emilk, line 406:
```suggestion
    /// as in the [`crate::DragValue`] widget, so a widget can be focused
```

crates/egui/src/widgets/drag_value.rs
```diff
@@ -371,18 +371,40 @@ impl<'a> Widget for DragValue<'a> {
...
         let auto_decimals = (aim_rad / speed.abs()).log10().ceil().clamp(0.0, 15.0) as usize;
         let auto_decimals = auto_decimals + is_slow_speed as usize;
-
         let max_decimals = max_decimals.unwrap_or(auto_decimals + 2);
         let auto_decimals = auto_decimals.clamp(min_decimals, max_decimals);
+
+        if is_kb_editing {
+            let mut input = ui.input_mut();
+            let change = input.count_and_consume_key(Modifiers::NONE, Key::ArrowUp) as f64
+                - input.count_and_consume_key(Modifiers::NONE, Key::ArrowDown) as f64;
```
emilk, line 394: The previous code also listened for `ArrowRight` and `ArrowLeft`, which matches the direction of most sliders, and is also a valid drag-axis.

emilk (changes requested):
Looks good, but I would like to keep the behavior of pressing right and left!

crates/egui/src/widgets/drag_value.rs
mwcampbell: I thought I replied to this one a few days ago, but apparently I didn't actually post it.

The problem is that the left and right arrows can be used to move the caret when in keyboard editing mode. So we can't consume those keystrokes in `DragValue` itself.

To look at an existing accessible stepper/spinner implementation, I created a little web page with an `<input type="number">` element. In Chromium's implementation of that input type, at least on Windows, the up and down arrows adjust the number, but the left and right arrows move the caret.

Before this PR, `DragValue` was supposed to be able to listen for keypresses while in button mode, but that never actually happened, because as soon as the widget receives keyboard focus, it switches to keyboard editing mode. In an earlier iteration of my AccessKit branch, I modified `DragValue` to stay in button mode when it received the keyboard focus, unless the user pressed Enter, at which point it would switch to editing mode. This allows all arrows to be used to adjust the number. The problem with this is that the JAWS screen reader for Windows assumes that a spinner (the Windows name for this type of widget) is always in edit mode when it has the focus. (NVDA and Narrator seem to have no problem with a non-editable spinner, though.) And that previous change was also more drastic than what I'm doing now. So I figured that, all things considered, it's best to keep the current behavior of immediately going into edit mode on focus, and not respond to the left and right arrows. I'm open to other thoughts though.

emilk (approved):

emilk:
Thanks!

PR #2343, opened 2022-11-25
Do not emit changed responses on unchanged values in selectable_value/radio_value

xTibor:
The `selectable_value` and `radio_value` methods did not take into account that you can click on the very same widget multiple times in a row. They emitted `changed` responses on every click, even when the underlying value did not change.


emilk (approved):
good catch!

Issue #2333, opened 2022-11-22
Flash of white when running the egui_glow / pure_glow example on Windows


fgimian:
**Describe the bug**

I'm running the pure_glow example as follows on Windows 11 Pro with dark mode enabled:

```powershell
cd crates\egui_glow\examples
cargo run --features="winit egui/default_fonts" --example pure_glow
```

Upon starting up, there's a brief flash of white before the UI shows up.  I'm experiencing this with all egui related projects too, but can narrow it down to something to do with the rendering with glow.

**To Reproduce**
Steps to reproduce the behavior: See above

**Expected behavior**
The window should not show up until it's ready and thus no bursts of white should show if the background is dark.

**Screenshots**
![White Flash egui](https://user-images.githubusercontent.com/1811813/203248084-053239f5-707d-463f-b7f8-5c0d7b0e7c88.gif)

**Desktop (please complete the following information):**
 - OS: Windows 11 Pro
 - Browser N/A
 - Version N/A

**Smartphone (please complete the following information):**: N/A

**Additional context**

If I disable `egui_glow.paint(gl_window.window());` and ensure that `visible` is only set to `true` after the context is created, I see no flash of white:

e.g.

```rust
...
    let window_builder = glutin::window::WindowBuilder::new()
        .with_resizable(true)
        .with_inner_size(glutin::dpi::LogicalSize {
            width: 800.0,
            height: 600.0,
        })
        .with_title("egui_glow example")
        // FG: Ensure that the window is not visible.
        .with_visible(false);

    let gl_window = unsafe {
        glutin::ContextBuilder::new()
            .with_depth_buffer(0)
            .with_stencil_buffer(0)
            .with_vsync(true)
            .build_windowed(window_builder, event_loop)
            .unwrap()
            .make_current()
            .unwrap()
    };

    // FG: Only make the window visible after the context is created to avoid the flash of white.
    gl_window.window().set_visible(true);
...
```

You can see the results below:

![Fixed Flash](https://user-images.githubusercontent.com/1811813/203249212-96701605-8a35-4d50-bec4-c085770c0d6f.gif)

That tweak gets me half way there, but the moment I paint with egui by re-enabling the following line:

```rust
egui_glow.paint(gl_window.window());
```

I see the flash of white again.

Cheers
Fotis

LoganDark:
Is this related to the eframe fix #2279 (which is my PR)?

emilk:
Is this still a problem on latest `master`?

[...]

PR #2345, opened 2022-11-26
Update the glow example to avoid a white flash when the app starts.

fgimian:
Closes <https://github.com/emilk/egui/issues/2333>.

### DIFF
--- a/crates/egui_glow/examples/pure_glow.rs
+++ b/crates/egui_glow/examples/pure_glow.rs
@@ -53,6 +53,7 @@ fn main() {
                 // draw things on top of egui here
 
                 gl_window.swap_buffers().unwrap();
+                gl_window.window().set_visible(true);
             }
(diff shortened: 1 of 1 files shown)


fgimian:
I will admit that I'm sure if running `set_visible` on each renedred frame as is also done in eframe is a wise decision  Doing this actually does quite a bit of work on Windows for example:

```rust
    pub fn set_visible(&self, visible: bool) {
        let window = self.window.clone();
        let window_state = Arc::clone(&self.window_state);
        self.thread_executor.execute_in_thread(move || {
            let _ = &window;
            WindowState::set_window_flags(window_state.lock(), window.0, |f| {
                f.set(WindowFlags::VISIBLE, visible)
            });
        });
    }
```

However, I've aligned my solution with that of eframe.

Cheers
Fotis

Issue #2354, opened 2022-11-28
Subpixel antialiasing using freetype


lulitao1997:
Adding subpixel font antialising will greatly improve user experience on non-retina displays, I implemented this using rust's freetype, please see this preliminary PR.

Some screenshots--

rendered using freetype --
![image](https://user-images.githubusercontent.com/20862509/204182472-889a5e45-4660-4a29-840a-6ff53427ced6.png)

vs rendered using ab_glyph
![image](https://user-images.githubusercontent.com/20862509/204182544-46183035-0ebb-4c40-b51e-35862baa3635.png)

related issue ##1259

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #2356, opened 2022-11-28
WIP use freetype for epaint text rendering

lulitao1997:
implemented subpixel AA using freetype, this PR is working but still need to make freetype a optional dependency, maybe controled by a feature tag?
related issue #2354

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,22 +3,6 @@
 version = 3
 
 [[package]]
-name = "ab_glyph"
-version = "0.2.16"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "846ffacb9d0c8b879ef9e565b59e18fb76d6a61013e5bd24ecc659864e6b1a1f"
-dependencies = [
- "ab_glyph_rasterizer",
- "owned_ttf_parser",
-]
-
-[[package]]
-name = "ab_glyph_rasterizer"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a13739d7177fbd22bb0ed28badfff9f372f8bef46c863db4e1c6248f6b223b6e"
-
-[[package]]
 name = "addr2line"
 version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -873,6 +857,7 @@ name = "custom_font"
 version = "0.1.0"
 dependencies = [
  "eframe",
+ "rfd",
 ]
 
 [[package]]
@@ -1364,7 +1349,6 @@ dependencies = [
 name = "epaint"
 version = "0.19.0"
 dependencies = [
- "ab_glyph",
  "ahash 0.8.1",
  "atomic_refcell",
  "backtrace",
@@ -1374,6 +1358,7 @@ dependencies = [
  "criterion",
  "document-features",
  "emath",
+ "freetype-rs",
  "nohash-hasher",
  "parking_lot",
  "serde",
@@ -2573,15 +2558,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "88cfece6e95d2e717e0872a7f53a8684712ad13822a7979bc760b9c77ec0013b"
 dependencies = [
  "shared_library",
-]
-
-[[package]]
-name = "owned_ttf_parser"
-version = "0.15.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "07ef1a404ae479dd6906f4fa2c88b3c94028f1284beb42a47c183a7c27ee9a3e"
-dependencies = [
- "ttf-parser",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ opt-level = 2 # fast and small wasm, basically same as `opt-level = 's'`
 # debug = true # include debug symbols, useful when profiling wasm
 
 
-[profile.dev]
+[profile.dev.'cfg(!windows)']
 split-debuginfo = "unpacked" # faster debug builds on mac
 # opt-level = 1                # Make debug builds run faster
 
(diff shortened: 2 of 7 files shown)


emilk:
I would like to try this, but on my Mac I get:

```
error[E0308]: mismatched types
   --> crates/epaint/src/text/font.rs:96:63
    |
96  |                     freetype::ffi::FT_Get_Next_Char(raw_font, charcode, &mut self.gindex)
    |                     -------------------------------           ^^^^^^^^ expected `u64`, found `u32`
    |                     |
    |                     arguments to this function are incorrect
    |
note: function defined here
   --> /Users/emilk/.cargo/registry/src/github.com-1ecc6299db9ec823/freetype-sys-0.13.1/src/lib.rs:953:12
    |
953 |     pub fn FT_Get_Next_Char(face: FT_Face, char_code: FT_ULong, agindex: *mut FT_UInt) -> FT_ULong;
    |            ^^^^^^^^^^^^^^^^
help: you can convert a `u32` to a `u64`
    |
96  |                     freetype::ffi::FT_Get_Next_Char(raw_font, charcode.into(), &mut self.gindex)
    |                                                                       +++++++
```

I'd be far more interested in trying out https://github.com/pop-os/cosmic-text for the layouting and font rendering in egui.

LoganDark:
Subpixel antialiasing can be done for the entire canvas and without freetype, PoC : (click)

image

This is also higher quality than doing it only for text in isolation

lulitao1997:
> It is possible to perform subpixel antialiasing on the entire app and without freetype:
> 
> Demos
> It works for all graphics, including widgets and vector drawings and even a 3D view or images:
> 
> 3D painting
> EasyMark Editor
> Color test

LoganDark:
> how do u do that?

After 14 hours of searching I can't seem to find the article that I read that explains subpixel antialiasing. But basically, you can apply it to an entire application in the exact same way that you would apply it to text.

The basic idea is to render to a buffer that is three times as wide as the actual window, and then downsample it using a subpixel antialiasing filter. This can be done efficiently with a fragment shader on the GPU, and doesn't require switching the entire text rasterization backend (esp. when FreeType would be one more non-Rust dependency to build).

Let me explain in more detail, since that blog post is unfortunately lost to my browsers' lack of good history search.

Let's say you are trying to subpixel-antialias a lowercase `e`. Here is a close-up:

expe1

The point of subpixel antialiasing is to approximate the shape better, like this:

expe3

We want to find a way to use the full precision of the LCD's subpixels to achieve this, since they can theoretically achieve a horizontal resolution of three times what grayscale AA can.

First, though, we need to create a rendering where what will become each subpixel is represented individually. This can be done by rendering it at 3x width:

expe2

Each three columns represents what will become only one column of output. The not-quite-right approach is to try to combine them as-is - that is, for each set of three columns, take the red channel from the left column, the green channel from the center column, and the blue channel from the right column. (And vice versa for BGR panels)

This results in:

expe4

But this is obviously too intense - for example, there is bright yellow to the left side. We're only missing one step, though.

The trick is to average all pixels with their horizontal neighbors, so that the color fringing is less pronounced. That results in this rendering:

expe5

and this subpixel-antialiased result:

expe6

This approach is very easily generalizable - it can be applied to much more than just text. Here it is applied to this GitHub issue:

image

Notice how not only is the text subpixel-antialiased, but everything is - border radii, emoji, your profile picture, icons and solid shapes.

However, this reveals an issue with my implementation - solid pixels bordered by other solid pixels bleed into each other by one subpixel. I'm sure this can be fixed somehow - and I've tried not taking into account neighboring full-pixels - but I just can't seem to get a result as good as what I've shown so far.

As always, more research is needed.

(It took me four hours to compose this comment lol)

lulitao1997:
> ### The subpixel AA in this post, and above, accidentally assume a BGR panel instead of RGB, oops. It is trivially reversible, and the same concepts still apply
> > how do u do that?
> 
> After 14 hours of searching I can't seem to find the article that I read that explains subpixel antialiasing. But basically, you can apply it to an entire application in the exact same way that you would apply it to text.
> 
> The basic idea is to render to a buffer that is three times as wide as the actual window, and then downsample it using a subpixel antialiasing filter. This can be done efficiently with a fragment shader on the GPU, and doesn't require switching the entire text rasterization backend (esp. when FreeType would be one more non-Rust dependency to build).
> 
> Let me explain in more detail, since that blog post is unfortunately lost to my browsers' lack of good history search.
> 
> Let's say you are trying to subpixel-antialias a lowercase `e`. Here is a close-up:
> 
> expe1
> 
> The point of subpixel antialiasing is to approximate the shape better, like this:
> 
> expe3
> 
> We want to find a way to use the full precision of the LCD's subpixels to achieve this, since they can theoretically achieve a horizontal resolution of three times what grayscale AA can.
> 
> First, though, we need to create a rendering where what will become each subpixel is represented individually. This can be done by rendering it at 3x width:
> 
> expe2
> 
> Each three columns represents what will become only one column of output. The not-quite-right approach is to try to combine them as-is - that is, for each set of three columns, take the red channel from the left column, the green channel from the center column, and the blue channel from the right column. (And vice versa for BGR panels)
> 
> This results in:
> 
> expe4
> 
> But this is obviously too intense - for example, there is bright yellow to the left side. We're only missing one step, though.
> 
> The trick is to average all pixels with their horizontal neighbors, so that the color fringing is less pronounced. That results in this rendering:
> 
> expe5
> 
> and this subpixel-antialiased result:
> 
> expe6
> 
> This approach is very easily generalizable - it can be applied to much more than just text. Here it is applied to this GitHub issue:
> 
> image
> 
> Notice how not only is the text subpixel-antialiased, but everything is - border radii, emoji, your profile picture, icons and solid shapes.
> 
> However, this reveals an issue with my implementation - solid pixels bordered by other solid pixels bleed into each other by one subpixel. I'm sure this can be fixed somehow - and I've tried not taking into account neighboring full-pixels - but I just can't seem to get a result as good as what I've shown so far.
> 
> As always, more research is needed.
> 
> (It took me four hours to compose this comment lol)

thanks so much for the detailed reply!

emilk:
Closing: the text rasterizer was replaced by skrifa + vello_cpu in #7694, with hinting, so this no longer applies. Thanks for the PR!

PR #2389, opened 2022-12-05
Added KeyRepeat event.

mkrueger:
KeyPress no longer repeats.


emilk:
Ah nice, that's a pretty elegant solution!

emilk:
…but it needs `cargo fmt`

mkrueger:
Did run cargo fmt & force pushed.

btw. nice didn't know about cargo fmt. Will use that for my projects as well.

emilk:
you seem to have forgotten to push

mkrueger:
Thanks

PR #2397, opened 2022-12-06
Chrono update

elwerene:
As asked for by @emilk: https://github.com/emilk/egui/pull/963#issuecomment-1337102242

* update chrono to 0.4.23
* switch to NaiveDate
* remove use of deprecated functions


emilk (approved):
Thanks!

PR #2399, opened 2022-12-06
Split out ecolor crate

Wumpf:
Makes color its own crate and splits it in turn into lot of smaller modules.
Due to the generous imports of color structs this should not be a (hugely) breaking change!


emilk:
Thanks! Please silence clippy with `#![allow(clippy::manual_range_contains)]`. I'm not a big fan of that warning. Perhaps we can add it to `allow`-section in `Cranky.toml`?

crates/ecolor/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Changelog for ecolor
+All notable changes to the `ecolor` integration will be noted in this file.
```
emilk, line 2:
```suggestion
All notable changes to the `ecolor` crate will be noted in this file.
```

emilk (approved):

PR #2400, opened 2022-12-06
fix wrong variable name

coderedart:
I used rust analyzer's rename functionality and it seems the feature gated code is not affected by that.

I am surprised that the CI hasn't caught this one.


emilk (approved):
We don't have a CI job for Windows… that's why.

PR #2410, opened 2022-12-08
Improved plot groups and bounds handling

### LINKED ISSUE #2409: Improve implementation of linked plot bounds and cursors
The linked bounds I originally introduced and the cursor groups that have since been implemented in a similar fashion have some drawbacks. Users have to create and store them between frames, which goes against egui's approach to store data between frames in its own memory. The stored `Rc` in these groups also puts some avoidable limitations on user code, e.g. parallelism.
I've also noticed a regression in the plot bound handling. Changing the plot bounds manually when linked axes are active makes the plot non-interactive.

I have a PR ready to addresses both of those issues.

s-nie:
Closes #2409


crates/egui/src/widgets/plot/mod.rs
```diff
@@ -35,9 +32,14 @@ type AxisFormatter = Option<Box<AxisFormatterFn>>;
 type GridSpacerFn = dyn Fn(GridInput) -> Vec<GridMark>;
 type GridSpacer = Box<GridSpacerFn>;
 
+type CoordinatesFormatterFn = dyn Fn(&PlotPoint, &PlotBounds) -> String;
+
+const LINKED_PLOT_AXES_KEY: &str = "egui_linked_plot_axes";
+const LINKED_PLOT_CURSORS_KEY: &str = "egui_linked_plot_cursors";
```
s-nie, line 38: Not sure what the best approach is to add something to the egui memory.

CHANGELOG.md
```diff
@@ -7,11 +7,12 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 ## Unreleased
 ### Added ⭐
 * `Event::Key` now has a `repeat` field that is set to `true` if the event was the result of a key-repeat ([#2435](https://github.com/emilk/egui/pull/2435)).
+
+### Changed 🔧
+* ⚠️ BREAKING: `Plot::link_axis` and `Plot::link_cursor` now take the name of the group ([#2410](https://github.com/emilk/egui/pull/2410)).
```
emilk, line 12: Bad merge? The following entries are addition, while the new entry is a Change

crates/egui/src/widgets/plot/mod.rs
emilk: If data is to only be identified by type, use `Id::null`: https://docs.rs/egui/latest/egui/util/id_type_map/struct.IdTypeMap.html

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -617,17 +510,29 @@ impl Plot {
         self
     }
 
-    /// Add a [`LinkedAxisGroup`] so that this plot will share the bounds with other plots that have this
-    /// group assigned. A plot cannot belong to more than one group.
-    pub fn link_axis(mut self, group: LinkedAxisGroup) -> Self {
-        self.linked_axes = Some(group);
+    /// Add this plot to an axis link group so that this plot will share the bounds with other plots in the
+    /// same group. A plot cannot belong to more than one axis group.
+    pub fn link_axis(mut self, group_name: impl Into<Id>, link_x: bool, link_y: bool) -> Self {
```
emilk, line 515: this should be `group_id` - "name" makes it sounds like it is human readable

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -799,23 +708,26 @@ impl Plot {
         let mut bounds = *last_screen_transform.bounds();
 
         // Find the cursors from other plots we need to draw
-        let draw_cursors: Vec<Cursor> = if let Some(group) = linked_cursors.as_ref() {
-            let mut frames = group.frames.borrow_mut();
+        let draw_cursors: Vec<Cursor> = if let Some((id, _)) = linked_cursors.as_ref() {
+            let data = &mut ui.memory().data;
+            let frames: &mut HashMap<Id, Vec<PlotFrameCursors>> =
```
emilk, line 713: `HashMap<Id, Vec<PlotFrameCursors>>` is a pretty complex type that needs to match exactly when reading and writing data to `memory().data`. It would be a good idea to use a type alias or newtype for it.

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -824,27 +736,43 @@ impl Plot {
...
         // Transfer the bounds from a link group.
-        if let Some(axes) = linked_axes.as_ref() {
-            if let Some(linked_bounds) = axes.get() {
-                if axes.link_x {
-                    bounds.set_x(&linked_bounds);
-                    // Mark the axis as modified to prevent it from being changed.
-                    bounds_modified.x = true;
+        if let Some((id, axes)) = linked_axes.as_ref() {
+            let data = &mut ui.memory().data;
+            let link_groups: &mut HashMap<Id, PlotBounds> =
```
emilk, line 741: Same here

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -824,27 +736,43 @@ impl Plot {
...
+        // Allow double clicking to reset to the initial bounds.
+        if allow_double_click_reset && response.double_clicked() {
             bounds_modified = false.into();
         }
 
-        // Reset bounds to initial bounds if we haven't been modified.
+        // Apply bounds modifications.
+        bounds_modifications
+            .into_iter()
+            .for_each(|modification| match modification {
```
emilk, line 761:
```suggestion
        for modification in bounds_modifications {
            match modification {
```

crates/egui_demo_lib/src/demo/plot_demo.rs
```diff
@@ -520,37 +516,35 @@ impl LinkedAxisDemo {
...
-        self.group.set_link_y(self.link_y);
         ui.horizontal(|ui| {
             ui.label("Linked cursors:");
             ui.checkbox(&mut self.link_cursor_x, "X");
             ui.checkbox(&mut self.link_cursor_y, "Y");
         });
-        self.cursor_group.set_link_x(self.link_cursor_x);
-        self.cursor_group.set_link_y(self.link_cursor_y);
+
+        const LINK_GROUP_NAME: &str = "linked_demo";
```
emilk, line 525: Better to use a scoped `Id` than a global one:

```suggestion
        let link_group_id = ui.id().with("linked_demo");
```

emilk (approved):
Very good!

s-nie:
Thanks for taking the time @emilk. I addressed the comments. It looks like CI is failing, but for a reason unrelated to this PR.

emilk:
I think this broke something:

bad-groups

s-nie:
@emilk Good catch, https://github.com/emilk/egui/pull/2410/commits/d8efd9a07c82cb391cf8263e14aa34f9bf20740c should fix it.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 * Add `Slider::trailing_fill` for trailing color behind the circle like a `ProgressBar` ([#2660](https://github.com/emilk/egui/pull/2660)).
 
 ### Changed 🔧
+* ⚠️ BREAKING: `Plot::link_axis` and `Plot::link_cursor` now take the name of the group ([#2410](https://github.com/emilk/egui/pull/2410)).
```
emilk, line 35: Move to `## Unreleased`

s-nie: Moved in https://github.com/emilk/egui/pull/2410/commits/ba13680a9d3747fcc2a057bb73e180322b2c9eec

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -1038,14 +980,20 @@ impl Plot {
...
+
 /// Provides methods to interact with a plot while building it. It is the single argument of the closure
 /// provided to [`Plot::show`]. See [`Plot`] for an example of how to use it.
 pub struct PlotUi {
     items: Vec<Box<dyn PlotItem>>,
     next_auto_color_idx: usize,
     last_screen_transform: ScreenTransform,
-    bounds_modified: AxisBools,
     response: Response,
+    bounds_modifications: Vec<BoundsModification>,
```
emilk, line 996: why are you accumulating modifications instead of applying them right away?

s-nie: So that they can be applied together at the right time, there are some other transforms that should be done between lines 711 and 763 before the modifications.

I'm sure there are other ways to do this as well, but the bound handling is pretty complex and I didn't want to mess with it too much.

OmegaJak:
I made some improvements to the interaction with auto_bounds in a PR on the source repo: https://github.com/s-nie/egui/pull/1

We're in fork-of-a-fork territory here to if there's a better way to handle it than making a PR on the fork's repo, then please let me know!

s-nie:
Thanks for the improvement @OmegaJak! I merged it into this branch.

crates/egui/src/widgets/plot/mod.rs
s-nie: Added a comment in https://github.com/emilk/egui/pull/2410/commits/eaec082d8ddf56450e63a12916e70be8c6cd72b2.

emilk (approved):

Issue #2415, opened 2022-12-08
Key repeat with backspace doesn't work on TextEdit


ItsEthra:
**Describe the bug**
With newly added `Event::KeyRepeat` backspace stopped properly working with TextEdit. When being held it only removes one character.

**To Reproduce**
```rust
use eframe::{run_native, App, NativeOptions};
use egui::CentralPanel;

#[derive(Default)]
struct ExampleApp {
    buf: String,
}

impl App for ExampleApp {
    fn update(&mut self, ctx: &egui::Context, _: &mut eframe::Frame) {
        CentralPanel::default().show(ctx, |ui| {
            ui.text_edit_singleline(&mut self.buf);
        });
    }
}

fn main() {
    run_native(
        "Example",
        NativeOptions::default(),
        Box::new(|_| Box::<ExampleApp>::default()),
    );
}

```

**Expected behavior**
When holding backspace multiple character are expected to be removed.

PR #2416, opened 2022-12-08
Fixed backspace repeat with TextEdit

ItsEthra:
Closes <https://github.com/emilk/egui/issues/2415>.


emilk:
Good catch, that's quite irritating. I think the arrow keys should probably also heed repeats. I'll merge this and fix that soon after

PR #2419, opened 2022-12-09
fix all clippy lints and remove them from allow list in cranky

coderedart:
As a new version has been published, its better to clean up clippy lints before any other major pull requests.


coderedart:
It seems the majority of the lints were already fixed in https://github.com/emilk/egui/commit/5093b67e9b095072033e54557825f3bfc9d39c49

I will leave type complexity lint alone. It requires naming a bunch of closures types in plot module which i have no idea about.

Wumpf (approved):
neat, thanks for tidying up!

Issue #2418, opened 2022-12-09
DragValue not support empty content in text editing mode


tkkcc:
### Discussed in https://github.com/emilk/egui/discussions/2376

<div type='discussions-op-text'>

<sup>Originally posted by **tkkcc** December  4, 2022</sup>
A normal way to edit `2` into `0` is `backspace` + `0`. Current DragValue does not support this.</div>

tkkcc:
One DragValue works well. Two or more DragValues boom.

The problem is `ui.memory().drag_value.edit_string` is shared by all DragValues. If any DragValue is not in keyboard mode(this is always true), it's `update` will reset `edit_string` to None. The current editing DragValue has to reconstruct `edit_string` from it's `value`, so boom.

same issue: https://github.com/emilk/egui/issues/2370

Issue #2370, opened 2022-11-30
DragValue max_decimals not working on master


paul-hansen:
**Describe the bug**
Max decimals for drag value seems to not be working on master? Maxes out at 2-4 seemingly random.
![drag_value_max_decimals](https://user-images.githubusercontent.com/7019130/204860454-a26050e4-0389-4b00-91c5-84d3e9e5015d.gif)

**To Reproduce**
1. Change the hello_world example to use:
    ``` rust
    ui.add(
        egui::DragValue::new(&mut self.height)
            .min_decimals(1)
            .max_decimals(10),
    );
    ```
2. Add a height f32 to the `MyApp` struct.
3. Try to enter a value with 10 decimals
4. Observe that you cannot get past 2-4 decimal places.

**Expected behavior**
It should allow you to enter more decimal places like it did in 0.19.0

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: 11 22H2 22621.819

**Additional context**
Might be related to the recent change to dragvalue in 0336816faf9f361e59ca77f99c19fcfa2bf7c993 which will be greatly appreciated and is the reason I tried out master, but for my application I'll need to be able to enter higher precision values sometimes.

PR #2421, opened 2022-12-09
Fix drag_value.edit_string unexpected reset

tkkcc:
Make drag_value.edit_string None only during current focused widget update

Only two cases `edit_string` will be set to None (and reconstructed from `value`):

1. current widget gains focus.
2. current widget consumes up down key to change value.

Editing, Tabing and UpDown tested.

Close https://github.com/emilk/egui/issues/2418
Close https://github.com/emilk/egui/issues/2370

### DIFF
--- a/crates/egui/src/widgets/drag_value.rs
+++ b/crates/egui/src/widgets/drag_value.rs
@@ -499,6 +499,7 @@ impl<'a> Widget for DragValue<'a> {
             }
 
             if response.clicked() {
+                ui.memory().drag_value.edit_string = None;
                 ui.memory().request_focus(id);
             } else if response.dragged() {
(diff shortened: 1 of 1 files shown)


tkkcc:
Still problematic. Editing DragValue then click another DragValue, the edit_string does not reset.

tkkcc:
Fixed, add another reset on `request_focus`. Seems `gained_focus` is not satisfied this case.

emilk:
@mwcampbell worked on this code recently - maybe you wanna take a look at this so it doesn't break AccessKit?

mwcampbell:
I reviewed the code and tested it using only the keyboard and a screen reader, and it looks good to me.

emilk:
Thanks for taking a look @mwcampbell! And thanks for the PR @tkkcc!

PR #2422, opened 2022-12-09
Fixed docs comment in web::start

tomoyanonymous:
The current example code in `web::start` returns `Result<AppRunnerRef, eframe::wasm_bindgen::JsValue>` but `#[wasm_bindgen]` macro returns an error for this.
In the demo app, this was fixed in #1650 and #1886 to wrap `AppRunnerRef` with a single struct, `WebHandle`.

(btw, `WebHandle` can not be defined on eframe side as commented in <https://github.com/emilk/egui/pull/1650#discussion_r933031615> ,right? )


emilk:
Thanks!

Issue #2423, opened 2022-12-09
Text editor no longer repeats caret keys/backspace etc.


mkrueger:
The text editor needs to use the KeyRepeat event, otherwise only the first key is taken.

However IDK which other controls are affected by this issue. I've not thought about the other controls when I introduced the KeyRepeat event. Maybe it's better to revert that event and the commit that introduced the other key down behavior. 

If an application doesn't want additional key down events it's very easy to check. As alternative the key repeat event needs to be taken into account everywhere key input is needed.

See the PR.

mkrueger:
ok already fixed in the editor - only saw it in the egui demo that it doesn't work  - my fault.

mkrueger:
Only suggestion would be to use that instead of the 2  branches:

`            Event::KeyRepeat { key, modifiers }
            | Event::Key {
                key,
                pressed: true,
                modifiers,
            } => on_key_press(&mut cursor_range, text, galley, *key, modifiers),
`

Patten matching in rust is awesome :).

PR #2424, opened 2022-12-09
Fixed text edit widget.

mkrueger:
Closes https://github.com/emilk/egui/issues/2423


mkrueger:
I don't have an overview of all widgets that could need key repeat support.

Issue #2427, opened 2022-12-10
egui-wgpu panics if device is not found


lunixbochs:
If the device_descriptor doesn't match a device, egui-wgpu will panic here:
https://github.com/emilk/egui/blob/f4d8ab9/crates/egui-wgpu/src/winit.rs#L72-L73

This means there's no good way for the caller to fall back to a different egui backend, their app just panics.

In Parallels, on an M1 Mac, running Ubuntu 22.04 in the VM, the GPU driver is limited enough that a default wgpu instantiation by egui fails to match a driver:

```rust
egui_wgpu::winit::Painter::new(Default::default(), 1, 0);
```

I worked around this by relaxing the device target, but I know there are also machines with no OpenGL at all, and I'd like to be able to provide a software fallback for those. It might be good to make these code paths return a Result somehow.

emilk:
yup, we should indeed be avoiding any panics and `unwrap`:s, so returning a `Result` makes sense. Do you want to make a PR?

PR #2428, opened 2022-12-10
egui-wgpu: don't panic if we can't find a device (#2427)

lunixbochs:
`(egui_wgpu::Painter).set_window()` now returns a `Result<(), wgpu::RequestDeviceError>`

I'm not really sure you want to handle this kind of error in eframe, so for now I just added an unwrap there, which is at least the same behavior as before. In my case, I'm not using eframe, so I am able to handle this Result.

Closes <https://github.com/emilk/egui/issues/2427>.

### DIFF
--- a/crates/egui-wgpu/CHANGELOG.md
+++ b/crates/egui-wgpu/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to the `egui-wgpu` integration will be noted in this file.
 
 
 ## Unreleased
+* Fix panic if we can't find a device ([#2428](https://github.com/emilk/egui/pull/2428))
 
 
 ## 0.20.0 - 2022-12-08 - web support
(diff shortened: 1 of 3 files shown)


crates/egui-wgpu/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ All notable changes to the `egui-wgpu` integration will be noted in this file.
 * `Renderer::update_texture` no longer creates a new `wgpu::Sampler` with every new texture ([#2198](https://github.com/emilk/egui/pull/2198))
 * `Painter`'s instance/device/adapter/surface creation is now configurable via `WgpuConfiguration` ([#2207](https://github.com/emilk/egui/pull/2207))
 * Fix panic on using a depth buffer ([#2316](https://github.com/emilk/egui/pull/2316))
+* Fix panic if we can't find a device ([#2428](https://github.com/emilk/egui/pull/2428))
```
emilk, line 21: Move to the `Unreleased` section please!

emilk (commented):
Looks good! It would be even better if we kept bubbling up the errors through eframe too, but we can also do that later

crates/egui-wgpu/CHANGELOG.md
lunixbochs: Fixed

PR #2436, opened 2022-12-12
eframe: App::on_exit is no longer available for wasm32

emilk:
It was never called in wasm32, so no need to have it there


emilk:
…but it is also annoying to have to add `#[cfg(not(target = "wasm32-unknown-unknown"))]` to everything, any maybe we will add teardown one day so… nevermind

PR #2451, opened 2022-12-14
fix: Compute mesh u16 split size corretly

pum-purum-pum-pum:
The u16 max value is `1<<16 - 1` not `1<<16` (https://doc.rust-lang.org/std/u16/constant.MAX.html)


emilk:
This is more a case of a bad name. If you make this change you also need to change all `< MAX_SIZE` to `<= MAX_SIZE`

emilk:
Ah yes, it is the check lower down that is wrong! I'll clean it up slightly

emilk:
I added some clarifications in https://github.com/emilk/egui/pull/2459

PR #2452, opened 2022-12-14
Don't render the \r (Carriage Return) character, because it sucks

emilk:
It is the textual equivalent of ~coriander~ a stubbed toe


Wumpf (changes requested):
code looks great but can't let that PR description stand this way

Issue #2458, opened 2022-12-15
split_to_u16 produces mesh that doesn't pass validity check


pum-purum-pum-pum:
This [line](https://github.com/emilk/egui/blob/37fd141dd189196f673a4281114409f95b13b643/crates/epaint/src/mesh.rs#L289) will fail for a `MAX_SIZE` batch. 
It is failing for me in the `egui-miniquad` crate after splitting a large correct batch(with `split_to_u16`) -- it produces a batch that fails the check. At least it happens on the old crate version, but this code hasn't changed since.

PR #2459, opened 2022-12-15
Fix bug in `Mesh::split_to_u16`

emilk:
Builds on https://github.com/emilk/egui/pull/2451 (I failed to push more commits onto that branch)
Closes https://github.com/emilk/egui/issues/2458

Please try it out @pum-purum-pum-pum


pum-purum-pum-pum:
Yeah it's correct, thanks for clarifying! 
Patched the old version I'm on -- it works

PR #2476, opened 2022-12-18
Impls From<Pos2> for Vec2, From<Vec2> for Pos2

fundon:
Impls `From<Pos2>` for `Vec2`, `From<Vec2>` for `Pos2`.

### DIFF
--- a/crates/emath/src/pos2.rs
+++ b/crates/emath/src/pos2.rs
@@ -26,6 +26,15 @@ pub const fn pos2(x: f32, y: f32) -> Pos2 {
     Pos2 { x, y }
 }
 
+/// Treat this vector as a position.
+/// `v.into()` is equivalent to `Pos2::default() + v`.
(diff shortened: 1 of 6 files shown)


emilk:
There are a lot of interfaces that take `impl Into<Vec2>` as argument. With this PR, it would be very easy to accidentally pass a position in these places.

fundon:
https://github.com/search?q=repo%3Aemilk%2Fegui%20Into%3CPos2%3E&type=code

Before I was trying to modify the windows position, `default_pos` and `current_pos` accept the `Into<Pos2>` parameter, which would be more convenient if `Vec2` implemented From trait.

emilk:
@fundon why would it be more convenient?  You can use `pos2(x,y)` or `[x,y]` and it will work fine.

fundon:
Ok, we can add it if needed in the future. :)

PR #2482, opened 2022-12-19
Z-ordering of shapes

mxsdev:
Closes <https://github.com/emilk/egui/issues/1516>.

This PR implements the feature in the exact way described by @emilk. Library consumers are expected to set z-index using `ui.with_z(...)`.

I've added an example to `examples` demonstrating the feature.

screenshot


emilk:
Thanks for working on this! 🙏 

I'm heading off to a long vacation, so it will be at least a month until I will get to reviewing this. Just a heads-up!

crates/egui/src/context.rs
```diff
@@ -411,16 +411,18 @@ impl Context {
...
+                .push((id, layer.z, interact_rect));
 
             if hovered {
                 let pointer_pos = slf.input.pointer.interact_pos();
                 if let Some(pointer_pos) = pointer_pos {
-                    if let Some(rects) = slf.layer_rects_prev_frame.get(&layer_id) {
-                        for &(prev_id, prev_rect) in rects.iter().rev() {
-                            if prev_id == id {
+                    if let Some(rects) = slf.layer_rects_prev_frame.get_mut(&layer.area_layer) {
+                        rects.sort_by_key(|(_id, z, ..)| *z);
```
emilk, line 422: re-sorting this on each interaction is unnecessary. Better to sort once at the end of each frame instead.

crates/egui/src/layers.rs
```diff
@@ -108,36 +117,187 @@ impl LayerId {
...
+
+impl Default for ZOrder {
+    fn default() -> Self {
+        Self::BASE
+    }
+}
+
+/// An identifier for a paint layer which supports Z-indexing
+///
+/// This says: draw on `area_layer` with index z. This only affects the display
```
emilk, line 178:
```suggestion
/// This says: draw on [`AreaLayerId`] with index z. This only affects the display
```

crates/egui/src/ui.rs
```diff
@@ -310,8 +316,13 @@ impl Ui {
 
     /// Use this to paint stuff within this [`Ui`].
     #[inline]
-    pub fn layer_id(&self) -> LayerId {
-        self.painter().layer_id()
+    pub fn layer_id(&self) -> AreaLayerId {
```
emilk, line 319: let's call this `area_layer` (but also keep `layer_id` as a deprecated alias)

crates/egui/src/ui.rs
```diff
@@ -310,8 +316,13 @@ impl Ui {
...
+        self.painter().layer()
+    }
+
+    #[inline]
+    pub fn layer(&self) -> ZLayer {
```
emilk, line 324: let's call this `z_layer`

crates/egui/src/ui.rs
```diff
@@ -310,8 +316,13 @@ impl Ui {
...
+        self.painter().zlayer()
```
emilk, line 325:
```suggestion
        self.painter().z_layer()
```

or I will pronounce it as "slayer" :)

crates/egui/src/ui.rs
```diff
@@ -1737,6 +1748,33 @@ impl Ui {
         })
     }
 
+    /// Set z-index and layer id at the same time
+    pub fn with_layer<R>(
```
emilk, line 1752:
```suggestion
    pub fn with_z_layer<R>(
```

examples/z_order/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+                        ui.with_z(order2, |ui| {
+                            let response = ui.allocate_rect(
+                                Rect::from_min_size(pos + DELTA + shift, SIZE),
+                                Sense::click(),
+                            );
+                            let painter = ui.painter_at(response.rect);
+                            painter.rect_filled(response.rect, Rounding::none(), Color32::BLUE);
+
+                            if response.clicked() {
+                                log::info!("Clicked blue");
```
emilk, line 69: really nice that you test interaction too ⭐

crates/egui/src/layers.rs
```diff
@@ -108,36 +117,187 @@ impl LayerId {
...
+    #[inline(always)]
+    pub fn above_all() -> Self {
+        Self::ABOVE_ALL
+    }
+
+    /// Minimal z-order - nothing on the same area layer can render below this
+    #[inline(always)]
+    pub fn below_all() -> Self {
+        Self::BELOW_ALL
+    }
```
emilk, line 147:
```suggestion
    /// Base z-order, corresponding to a level of 0.
    pub const BASE: ZOrder = ZOrder(0);

    /// Maximal z-order - nothing on the same area layer can render above this.    
    pub const ABOVE_ALL: ZOrder = ZOrder(i32::MAX);
    
    /// Minimal z-order - nothing on the same area layer can render below this.
    pub const BELOW_ALL: ZOrder = ZOrder(i32::MIN);
```

emilk (changes requested):
Really nice - thanks for working on this!

Most of my comments is about naming, but there is one real blocker which is the sorting of the layers. It should be done once, instead of on each interaction.

emilk:
Continued in https://github.com/emilk/egui/pull/2633

Issue #382, opened 2021-05-12
Uneven text for non-integral DPI scaling


emilk:
with 1.25 `pixels_per_point` text can become uneven:

![egui_0 12](https://user-images.githubusercontent.com/1148717/118013180-20911000-b352-11eb-8ea1-978f4ab16faa.png)

The reason it gets like that is because I render each character as a textured quad. The font kerning requires a non-integral number of pixels of offset between characters. If I do that I get either blurry fonts (bilinear sampling) or weird offsets (round to nearest pixel).

Rounding to nearest pixel is done in `Font::layout_single_row_fragment` and in `Tessellator::tessellate_text`.

If we don't round to closest pixel, the text becomes even, but blurry:

![no_rounding](https://user-images.githubusercontent.com/1148717/118013218-2dadff00-b352-11eb-9267-8470735eca5c.png)

emilk:
An alternative solution is to support sub-pixel rendering of glyphs.

For instance, rendering each glyph with an x-offset of 0.0, 0.25, 0.5, 0.75 pixels, and then picking the closest one. That allows us to have non-integer letter spacing with both sharp glyphs and even letter spacing. But it also means each glyph gets added four times to the font atlas. For latin characters that's fine, but we probably want to disable that for logographic scripts.

PR #2490, opened 2022-12-20
Fix rounding errors with font scaling

parasyte:
I am working on something like a terminal emulator. It needs to have a fixed width font that renders identically on all platforms. One of the problems I am facing right now is that with the `pixels_per_point` value at 2.0 on macOS and 1.0 on Windows, scaling the fonts ends up rendering slightly differently in these environments.

Here are some screenshots with a grid of characters rendered on Windows and macOS. I initially developed this on macOS, so it is showing what I "expected" to see on Windows:

macOS:

prosaic-macos

Windows:

prosaic-windows

They both have exactly 800x600 `LogicalSize` inner window size, e.g. it's 800x600 physical pixels on Windows, and 1600x1200 physical pixels on macOS. But clearly the font size is ever so slightly different enough that the cursor isn't clipped in the corner of the window on Windows, as it is on macOS.

This PR attempts to address the bug.

<details><summary>Reveal outdated patch description</summary>

First, we notice that `Font::round_to_pixel()` is called with a point size that is already scaled by `pixels_per_point`, and it internally resizes by dividing after rounding. This leads to rounding errors. For example, suppose the point size is 16.27 and `pixels_per_point` is 2.0:

```rust
(16.27 * 2.0).round() == 33.0

33.0 / 2.0 == 16.5
```

Compare this when `pixels_per_point` is 1.0:

```rust
(16.27 * 1.0).round() == 16.0

16.0 / 1.0 == 16.0
```

If instead we `floor` the value, then we get the same result, regardless of the value of `pixels_per_point`:

```rust
(16.27 * 2.0).floor() == 32.0

32.0 / 2.0 == 16.0
```

Secondly, the same issue occurs with the horizontal advance, so we change that `round` to `floor` also.
</details>

----

Description of the new patch:

We defer the scaling of the font to as late as possible (e.g. when drawing glyphs). The fonts themselves now only store sizes in points (unscaled). Rounding of positions is still done, but it uses the font _points_ coordinate space instead of _physical pixels_. This accounts for the kerning differences on high DPI displays as shown in screenshots below.

The result is that the rasterization looks identical on Windows where `pixels_per_point == 1.0`, but the rasterization on macOS (`pixels_per_point == 2.0`) now more closely resembles what is seen on Windows:

prosaic-macos-fixed

----

I'll leave this as a draft PR to open discussion. I could be wrong, but I believe this is the correct way to address the problem. I'll leave a list of items that I know need to be done before this can be accepted.

TODO:

- [x] Add changelog entry
- [x] ~~Update comments and function names to reference flooring instead of rounding~~

### DIFF
--- a/crates/epaint/src/text/text_layout.rs
+++ b/crates/epaint/src/text/text_layout.rs
@@ -129,7 +129,26 @@ fn layout_section(
             });
 
             paragraph.cursor_x += glyph_info.advance_width;
-            paragraph.cursor_x = font.round_to_pixel(paragraph.cursor_x);
+
+            // This opt-in feature flag makes fixed width fonts scale proportionally to their
+            // surroundings regardless of the `pixels_per_point` setting. I.e., the text bounding
+            // box will be proportionally identical even when the window is moved between displays
+            // with varying DPIs.
+            //
+            // By default this feature is off, and text layout will nudge glyph positions to align
+            // with the pixel grid to retain the appearance of decent spacing between characters.
+            //
+            // See https://github.com/emilk/egui/pull/2490 for additional details.
+            #[cfg(feature = "proportional_font_scaling")]
+            {
+                paragraph.cursor_x = paragraph.cursor_x.round();
+            }
+
+            #[cfg(not(feature = "proportional_font_scaling"))]
+            {
+                paragraph.cursor_x = font.round_to_pixel(paragraph.cursor_x);
+            }
+
             last_glyph_id = Some(glyph_info.id);
         }
     }
@@ -817,6 +836,36 @@ fn test_zero_max_width() {
     layout_job.wrap.max_width = 0.0;
     let galley = super::layout(&mut fonts, layout_job.into());
     assert_eq!(galley.rows.len(), 1);
+}
+
+#[test]
+fn test_font_scaling() {
+    fn measure_max_glyph_size(pixels_per_point: f32) -> Vec2 {
+        let mut fonts = FontsImpl::new(pixels_per_point, 1024, super::FontDefinitions::default());
+        let layout_job = LayoutJob::single_section(
+            "Hello, world!".into(),
+            super::TextFormat {
+                font_id: super::FontId::monospace(14.0),
+                ..Default::default()
+            },
+        );
+        let galley = super::layout(&mut fonts, layout_job.into());
+
+        galley
+            .rows
+            .iter()
+            .flat_map(|row| row.glyphs.iter().map(|g| g.size))
+            .fold(Vec2::ZERO, |acc, size| acc.max(size))
+    }
+
+    let expected_size = Vec2::new(8.275167, 16.0);
+
+    assert_eq!(measure_max_glyph_size(1.0), expected_size);
+    assert_eq!(measure_max_glyph_size(1.15), expected_size);
+    assert_eq!(measure_max_glyph_size(1.5), expected_size);
+    assert_eq!(measure_max_glyph_size(2.0), expected_size);
+    assert_eq!(measure_max_glyph_size(3.0), expected_size);
+    assert_eq!(measure_max_glyph_size(3.333), expected_size);
 }
 
 #[test]
--- a/crates/epaint/CHANGELOG.md
+++ b/crates/epaint/CHANGELOG.md
@@ -7,6 +7,8 @@ All notable changes to the epaint crate will be documented in this file.
 * Don't render `\r` (Carriage Return) ([#2452](https://github.com/emilk/egui/pull/2452)).
 * Fix bug in `Mesh::split_to_u16` ([#2459](https://github.com/emilk/egui/pull/2459)).
 * Improve rendering of very thin rectangles.
+* Fix rounding errors with font scaling ([#2490](https://github.com/emilk/egui/pull/2490)):
+  * Added opt-in feature `proportional_font_scaling` to maintain text bounding box sizes when display DPI is changed.
 
 
 ## 0.20.0 - 2022-12-08
--- a/crates/epaint/Cargo.toml
+++ b/crates/epaint/Cargo.toml
@@ -49,6 +49,9 @@ deadlock_detection = ["dep:backtrace"]
 ## If you plan on specifying your own fonts you may disable this feature.
 default_fonts = []
 
+## Enabling this feature will allow fixed-width fonts to scale correctly for any display DPI.
+proportional_font_scaling = []
+
 ## Enable additional checks if debug assertions are enabled (debug builds).
 extra_debug_asserts = [
   "emath/extra_debug_asserts",
--- a/crates/epaint/src/text/font.rs
+++ b/crates/epaint/src/text/font.rs
@@ -61,7 +61,7 @@ pub struct FontImpl {
     name: String,
     ab_glyph_font: ab_glyph::FontArc,
     /// Maximum character height
-    scale_in_pixels: u32,
+    scale_in_points: f32,
     height_in_points: f32,
     // move each character by this much (hack)
     y_offset: f32,
@@ -76,13 +76,13 @@ impl FontImpl {
         pixels_per_point: f32,
         name: String,
         ab_glyph_font: ab_glyph::FontArc,
-        scale_in_pixels: u32,
+        scale_in_points: f32,
         y_offset_points: f32,
     ) -> FontImpl {
-        assert!(scale_in_pixels > 0);
+        assert!(scale_in_points > 0.0);
         assert!(pixels_per_point > 0.0);
 
-        let height_in_points = scale_in_pixels as f32 / pixels_per_point;
+        let height_in_points = scale_in_points;
 
         // TODO(emilk): use these font metrics?
         // use ab_glyph::ScaleFont as _;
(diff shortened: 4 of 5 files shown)


parasyte:
I don't know for sure, but this might also fix #2164 ?

parasyte:
This patch introduces some issues with font rendering, even with `pixels_per_point == 1.0`.

Before:

before

After:

after

Kerning is badly impacted, much like it is in #2164.

Perhaps it is worth attempting to address the problem by performing the scaling only once, as late as possible?

parasyte:
Ok, the latest commit does a much better job at fixing the rounding errors when `pixels_per_point != 1.0`. The before and after screenshots with `pixels_per_point == 1.0` look pixel-perfect identical, AFAICT.

Kerning is _absolutely_ different with this PR on high DPI displays with `pixels_per_point > 1.0`. Overall difference is minor (1 physical pixel here and there) but it's worth pointing out. For instance, here are some comparison shots on macOS with `pixels_per_point == 2.0`:

Before:

macOS-before

After:

macOS-after

And here are some detail shots that show the kerning differences. The changes may be better, but I don't necessarily think it is any worse than before. Open these each in a new tab and switch between them. Or put each into a separate layer in photoshop and toggle layer visibility.

# Detail 1

Before:

detail-1-before

After:

detail-1-after

# Detail 2

Before:

detail-2-before

After:

detail-2-after

crates/epaint/src/text/text_layout.rs
```diff
@@ -129,7 +129,7 @@ fn layout_section(
             });
 
             paragraph.cursor_x += glyph_info.advance_width;
-            paragraph.cursor_x = font.round_to_pixel(paragraph.cursor_x);
+            paragraph.cursor_x = paragraph.cursor_x.round();
```
emilk, line 132: This rounds each glyph to point coordinates. The previous code rounded it to pixel coordinates.

The reason we want to round it to pixel coordinates is that it creates more crisp text. Each glyph is rasterized to a little bitmap (part of the font-atlas), and rounding to pixel coordinates ensures that we don't get any billinear filtering-blur when reading them.

In other words: we have a trade-off to make. Do we want crisp text, or well-kerned text? We can't really have both with the current font rendering technique.

Rounding to points coordiantes makes no sense at all, and will just create _worse_ kerning whenever pixels_per_point is non-integral.

For instance, here is the demo app at 0.95 scale, before:

font-0 95-before

And this PR:
font-0 95-after

You can notice that a lot of the text is blurrier (especially visible in the monospace code) and that the kerning is worse in many places.

emilk (changes requested):
The current code is designed to produce crisp text and ok kerning for any `pixels_per_point` scale. The new code fails pretty badly on non-integral `pixels_per_point` (which are common).

There may be some kerning improvements to be made, but if it comes at the cost of crisp text it should be made optional imho.

parasyte:
It is not the keening that I am trying to address. That was just a side effect. The situation is that the code which rounds to pixels is causing the fonts to appear at the wrong scale when pixels_per_point is not 1.0. See the new tests in the PR.

emilk:
I appreciate that, but the point is that your PR breaks something else: crisp text.

Getting both crisp text and kerning right is difficult, and I suspect the current code may be doing the best job that could be done, but I'd be happy to be proven wrong.

parasyte:
I guess I'll have to look more closely, but on initial inspection I do not see any additional blurriness in the text when rendered with non-integer scale factors. What I do see, and what this PR wants to achieve, is correctly scaling the text proportionally to the scale factor.

The following screens show the difference. (The window size was not changed, only the scale value.)

# Before

With `pixels_per_point == 1.0`:

before-1.0

With `pixels_per_point == 0.9`:

before-0.9

With `pixels_per_point == 0.75`:

before-0.75

Notice that the words "feature" and "syntect" are laid out on different lines depending on the value of `pixels_per_point`. This is because the character size is _literally different proportionally to the UI_ when the scale factor is changed. And this is unfortunately caused by rounding in the pixel coordinate system before the point coordinate system has been fully "materialized".

# After

With `pixels_per_point == 1.0`:

after-1.0

With `pixels_per_point == 0.9`:

after-0.9

With `pixels_per_point == 0.75`:

after-0.75

Now notice that the text layout looks proportionately identical across the board.

----

I still can't make out the blurriness you mentioned, even at 0.9. Here is a comparison from both the before and after magnified by 10x:

# Before

before-0.9-detail

# After

after-0.9-detail

Neither looks blurrier to me. The only difference is in kerning (as mentioned in the OP) and even then, I don't think either is objectively worse. For instance, in the "before" there is 1px of space between the "u" and "r" that doesn't exist in the "after". But also in the "after" there is 1px of space between "{" and "s" that doesn't exist in the "before". So, it appears to be a net neutral change. No better, no worse, just different.

Ok, so that's the proportional font, what about the monospace font?

# Before

before-0.9-detail-2

# After

after-0.9-detail-2

Now it is obvious that the font sizes are literally different. In the "after" image, the "H" and the "d" are more clearly defined as a result of it just being larger. The double quotes and "w" are less clearly defined, also as a result of the larger font. And kerning may be worse. (Particularly the extra space between the two "l"s and the extra space between the "o" and "r".)

On the font size with the monospace font, the text is more accurately proportional to the width of the text input widget (i.e., the raison d'être for this PR):

# Before

| Scale | Widget width | Width of "println" | line_width / widget_width |
|-------|--------------|--------------------|---------------------------|
| 1.0   | 342 px       | 178 px             | 0.52046783625730994152046 |
| 0.9   | 307 px       | 145 px             | 0.47231270358306188925081 |
| 0.75  | 256 px       | 143 px             | 0.55859375                |

# After

| Scale | Widget width | Width of "println" | line_width / widget_width |
|-------|--------------|--------------------|---------------------------|
| 1.0   | 342 px       | 174 px             | 0.50877192982456140350877 |
| 0.9   | 307 px       | 159 px             | 0.51791530944625407166123 |
| 0.75  | 257 px       | 135 px             | 0.52529182879377431906614 |

The proportionality of the monospace font in this test varies by about 8% before this PR, but only about 2% after. It can never be perfect because it rounds each character to the pixel grid (for #382), but I'll take 2% over 8% any day!

----

I'm happy to address any feedback that can demonstrate blurring, but I honestly don't see it, here.

For the kerning issues in the monospace font ... Yeah, that's not ideal. But it accomplishes the goal of retaining proportionality (and character legibility), which is far more important in a monospace font than perfect kerning on a pixel grid. (See OP for rationale.)

> Getting both crisp text and kerning right is difficult

Ain't it the truth! I think I've shown that the crispness has not been affected at all by this PR. Kerning is still problematic, but also, I believe the real solution to that is probably subpixel rendering or another antialiasing technology (hinting, et al).

crates/epaint/src/text/text_layout.rs
parasyte: My understanding is that this specific line only changes the kerning on glyphs. Rasterization is rounded to the pixel grid in `tessellate_glyphs()`, which is not changed by this PR. It isn't clear to me that rounding here is useful at all, but I kept it (in point space) because it does improve the kerning at some scales.

But you can just as well `floor`, `ceil` or even remove the rounding entirely on this line and it will not affect glyph clarity or the pixel grid in the tessellator.

emilk:
I was wrong, the text seems crisp, sorry about that.

But the kerning for non-integer scales is pretty bad.

Try 1.1 pixels_per_point and then enter `iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii` in a text field and you will see a very uneven lettering:

Screen Shot 2023-01-26 at 11 40 18

compared to `master`:

Screen Shot 2023-01-26 at 11 42 35

parasyte:
Yes, that is intentional, unfortunately. It is the only way to both maintain the correct text bounding box regardless of scale, and also align each glyph to the pixel grid independently.

In one opinion it is unacceptable to have uneven kerning, and in another opinion, it is unacceptable to change the size of text bounding boxes when moving the window between different displays.

Since we are at an impasse, I propose a compromise with a new opt-in feature flag in `epaint` to disable the per-pixel rounding during glyph layout. That way, you get today's kerning behavior by default at the expense of fixed width fonts not actually having a fixed width. Or you can opt-in to better fixed width font behavior at the expense of poor kerning.

The real fix is of course better antialiasing (subpixel rendering, hinting, super sampling, etc. that does not rely on rounding positions to the pixel grid) for which I will create a separate ticket for tracking. Meanwhile this PR can potentially go forward to fix rounding errors introduced by forcing glyph kerning to the pixel grid.

Does this sound reasonable to you? I have tried the proposed feature flag locally and it is doing what I need it to.

emilk:
Yes, having flags for all these options would be _great_!

I thin we can give full control if we have flags for:

* Rounding font sizes to even pixels on/off (`FontImplCache::font_impl`)
* Rounding glyphs to even pixels on/off (`crates/epaint/src/text/text_layout.rs`)

`true, true`: stepwise scaling, crisp text, correct kerning
`false, true`: fluid scaling, crisp text, bad kerning
`false, false`: fluid scaling, blurry text, correct kerning
`true, false`: stepwise scaling, blurry text, correct kerning  (no real sense in this I think)

It would be extra great if these were runtime flags. For instance, for high-resolution displays it could make sense to turn off both types of rounding to get fluid scaling and perfect kerning at the cost of slightly blurry text (which makes much less difference on high resolution displays)

parasyte:
In terms of design, should the new flag stater be attached to the `egui::Context` and automatically inherited by fonts as they are added? I'll see what it takes to make it a runtime config.

emilk:
Closing: the ab_glyph-based font code this patched was replaced by skrifa + vello_cpu in #7694, so this no longer applies. Thanks for the PR!

PR #2491, opened 2022-12-20
`Plot::items::highlight` set  highlight in plot and in legend plot

haricot:
While working on a proposal for feature additions for the Plot widget, I realized that the Plot: items were not reacting with the parameter highlight function.
 I take this opportunity to propose `BoolState` which will make it a little clearer than the double boolean necessary for its operation, it seems to me. ~~Also the same case repeats itself if we want to add this functionality with allow_hover~~ , which I would like to  propose soon among others.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 * Improved the algorithm for picking the number of decimals to show when hovering values in the `Plot`.
 
 ### Fixed 🐛
+* `Plot::items`: fix bool setting of `highlight` function. 
 * Expose `TextEdit`'s multiline flag to AccessKit ([#2448](https://github.com/emilk/egui/pull/2448)).
 * Don't render `\r` (Carriage Return) ([#2452](https://github.com/emilk/egui/pull/2452)).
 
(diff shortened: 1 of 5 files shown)


haricot:
otherwise I propose a [branch](https://github.com/haricot/egui/pull/new/plot_item_highlight_bool) with double booleans:
   If the first boolean is true, it highlights
   and the second boolean is its changing state.

emilk:
Can you expand a bit on the problem? I would expect `.highlight(true)` to highlight the given item this frame. What is the difference between `init` and `state`?

haricot:
In fact, the highlight function doesn't seem to work without the two booleans.
in Plot::Items:
```
    pub fn highlight(mut self, highlight: bool) -> Self {
        self.highlight = highlight;
        self
    }
```
And personally, I don't see any point in disabling it (since PR #2558 allow_hover), so another possibility is to remove those chunks if you don't want to give the option to enable or disable highlighting in setting.

haricot:
> What is the difference between init and state?

if init is `false` and highlight becomes true (state = true) then it will not be highlighted.
if init is `true` and highlight becomes true (state = true) then it will be highlighted.

haricot:
I understood my mistake in what I expected from the highlight function.
what I was proposing here is to be able to authorize or not highlight in the plot and when hovering over the legend of the plot.
For example if a plot item has `allow_hover = false` we could expect it to also be in the plot legend.

haricot:
I close this PR because it seems more necessary to me and should be named Default_highlight if necessary because it allows elements not to be highlighted when hovering in the legend of the plot.

PR #2493, opened 2022-12-21
New feature to support unity vertex layout requirement

lazytiger:
Unity needs the vertex layout to be in order like "position, color, uv", but the current layout of epaint is "position, uv, color". So I  add a new feature to support this and fulfill the back-compatibility.


emilk:
Cannot Unity can be programmed to accept different vertex orderings?

It seems a bit weird to me to add several layouts of basic data because some backends cannot re-arrange data.

lazytiger:
As far as I know, there is no way to do that. Unity will do an import action to the external resources, but the runtime mesh must obey the rule.

crates/egui/Cargo.toml
```diff
@@ -54,6 +54,9 @@ persistence = ["serde", "epaint/serde", "ron"]
 ## Allow serialization using [`serde`](https://docs.rs/serde).
 serde = ["dep:serde", "epaint/serde", "accesskit?/serde"]
 
+unity = ["epaint/unity"]
```
emilk, line 57:
```suggestion
## Change Vertex layout to be compatible with unity
unity = ["epaint/unity"]
```

emilk (approved):

PR #2506, opened 2022-12-26
fix bezier curve assertion fail

xyning:
bezier curve can cause assertion fail and panic in some edge cases (typically when there are overlapping points).
<details>
<summary>example</summary>

image
</details>

this patch adds a check to ensure that there are enough points in the curve flatten result.


emilk (approved):
Thanks!

PR #2508, opened 2022-12-26
Checkbox without a text label

SDesya74:
Allows checkbox without a text label.

TODO:
 - [x] Unless this is a trivial change, add a line to the relevant CHANGELOG.md under "Unreleased".
 - [x] run cargo fmt
 - [x] run cargo clippy
 - [x] self-review
 - [x]  run ./sh/check.sh

### DIFF
--- a/crates/egui/src/widgets/button.rs
+++ b/crates/egui/src/widgets/button.rs
@@ -263,9 +263,7 @@ impl<'a> Checkbox<'a> {
     }
 
     pub fn without_text(checked: &'a mut bool) -> Self {
-        Checkbox {
-            checked,
-            text: None,
+        Self::new(checked, WidgetText::default())
         }
     }
 }
--- a/crates/egui/src/widgets/button.rs
+++ b/crates/egui/src/widgets/button.rs
@@ -234,7 +234,6 @@ impl Widget for Button {
 
 // ----------------------------------------------------------------------------
 
-// TODO(emilk): allow checkbox without a text label
 /// Boolean on/off control with text label.
 ///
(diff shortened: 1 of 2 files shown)


crates/egui/src/widgets/button.rs
```diff
@@ -246,18 +245,27 @@ impl Widget for Button {
...
             checked,
-            text: text.into(),
+            text: Some(text.into()),
+        }
+    }
+
+    pub fn without_text(checked: &'a mut bool) -> Self {
+        Checkbox {
+            checked,
+            text: None,
```
emilk, line 268:
```suggestion
    pub fn without_text(checked: &'a mut bool) -> Self {
        Self::new(checked, WidgetText::default())
```

and then you won't need any other change

emilk (changes requested):

crates/egui/src/widgets/button.rs
```diff
@@ -249,20 +248,26 @@ impl Widget for Button {
...
 /// ui.add(egui::Checkbox::new(&mut my_bool, "Checked"));
 /// # });
+/// // You can also create a checkbox without a text label
+/// ui.add(egui::Checkbox::without_text(&mut my_bool));
 /// ```
 #[must_use = "You should put this widget in an ui with `ui.add(widget);`"]
 pub struct Checkbox<'a> {
     checked: &'a mut bool,
-    text: WidgetText,
+    text: Option<WidgetText>,
```
emilk, line 257: No need for this to be an `Option` - the empty text label (`WidgetText::default`) already means "no label"

Issue #2509, opened 2022-12-26
Margin for Image Buttons


RobWalt:
As a user of `egui` I want to have more control of how the button is styled. Currently, if we include an image on a button, the image is positioned on the far left of the button right on the edge as we can see in the following picture.

![image](https://user-images.githubusercontent.com/26892280/209557785-15b55032-954a-4def-bffe-4b940a9823aa.png)

If we want to get some space between the button and the edge there are only two options, both of which are not very optimal:

1. We can add a frame with a margin around the button which pushes it to the side, but the image is still on the edge of the button. Here is a picture of this alternative applied to the example above
![image](https://user-images.githubusercontent.com/26892280/209558087-00711e0a-e79f-4d02-8d7a-9f97e2e81a3c.png)
2. We can edit the image and add margin there, which might be undesireable aswell, as we just increase the image size without adding real content

So it would be nice to have some method on the button to configure it. I think an interface like `image_margin(margin: Vec2)` would already do the trick.

yunfengwangluo:
I can get the required modification by modifying button.rs

RobWalt:
I already filed a PR for this issue !2510

PR #2510, opened 2022-12-26
Fix: `button_padding` when using image+text buttons

RobWalt:
Changes of the PR:

- fix `button_padding` functionality for buttons with images

Only the last comment of this PR is important. All other comments aren't relevant after my self review anymore.

----

### TODO:

- [x] Unless this is a trivial change, add a line to the relevant `CHANGELOG.md` under "Unreleased".
- [x] run `cargo fmt`
- [x] run `cargo clippy`
- [x] self-review
- [x] run `./sh/check.sh`

Closes <https://github.com/emilk/egui/issues/2509>.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 * Add `Plot::clamp_grid` to only show grid where there is data ([#2480](https://github.com/emilk/egui/pull/2480)).
 * Add `ScrollArea::drag_to_scroll` if you want to turn off that feature.
 * Add `Response::on_hover_and_drag_cursor`.
+* Add `image_margin` field and method on `Button` widget to configure its layout ([#2510](https://github.com/emilk/egui/pull/2510)).
 
 ### Changed 🔧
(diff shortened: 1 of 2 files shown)


RobWalt:
Question @reviewer: Should we maybe extend the margin feature to respect the text aswell, i.e. the `right` field of the `Margin` creates a margin on the right side of the text in the button instead?

RobWalt:
> Question @Reviewer: Should we maybe extend the margin feature to respect the text aswell, i.e. the `right` field of the `Margin` creates a margin on the right side of the text in the button instead?

It seems like this would be the better solution as we already have the `icon_spacing` to configure the space between the image and the text. I implemented a PoC of the changes above and the example with `image_margin(Margin::same(20.0))` and `icon_spacing = 30.0;` would look as follows:

image

RobWalt:
I just noticed that the style option `button_padding` does the trick. It just wasn't applied to the left of the button when there was an image. Setting the `button_padding` to `Vec2::splat(20.0)` in the example above looks like this

before

image


after

image

Note that this makes much more sense, since I also explicitly set `icon_spacing = 0.0`, which isn't the case in the "before" picture.

emilk (approved):
Good catch

PR #2511, opened 2022-12-27
Improvement to auto plot grid colors

apessino:
Currently, the automatic coloring of plot grid lines/labels produces harsh (and inconsistent) results with color schemes other than very light or very dark, for example:

https://user-images.githubusercontent.com/8810462/209590843-ae6462de-a0d4-40ec-a2c6-5c6c07395b33.mp4

Ideally, the user should have control over all aspects of the coloring, but because the contrasted color calculation is based on the "extreme" background color blended with the foreground stroke color of the widget, and these same colors are used for other aspects of the plot, it is not possible to control the result externally. By simply clamping the contrast factor to a value higher than 0.0 (I used 0.2 and it seems to work well) the behavior is much improved, and the light and dark themes still look good. Here it is in use:

https://user-images.githubusercontent.com/8810462/209590971-41edef74-bc9a-4267-a5d5-752e619f99ac.mp4


crates/egui/src/widgets/plot/mod.rs
```diff
@@ -1488,7 +1488,7 @@ impl PreparedPlot {
                 let line_strength = remap_clamp(
                     spacing_in_points,
                     MIN_LINE_SPACING_IN_POINTS as f32..=300.0,
-                    0.0..=1.0,
+                    0.2..=1.0,
```
emilk, line 1491: On `master`, when zooming in, new grid level smoothly fade in. With this change, they now suddenly pop onto screen in a very jarring manner.

You may want to tweak the gamma curve or something else, but we should definitely fade from zero here and below

emilk (changes requested):

PR #2523, opened 2022-12-28
Add feature to snap slider to user-specified values

ac-freeman:
Screenshot from 2022-12-28 12-54-54

Allows user to set "snap" points with `Slider::smart_aim_values`. When the slider marker is within a user-specified range of one of these points, it will jump to that point. If `smart_aim` is disabled, then these values don't do anything. If `smart_aim` is enabled, then regular smart aim is used as normal whenever the marker is outside the user-specified range of these snap points.

I added it to the demo app for seeing how it works. I'm open to feedback to make sure it fits in with the structural and stylistic requirements of the project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 * Add `Plot::clamp_grid` to only show grid where there is data ([#2480](https://github.com/emilk/egui/pull/2480)).
 * Add `ScrollArea::drag_to_scroll` if you want to turn off that feature.
 * Add `Response::on_hover_and_drag_cursor`.
+* Add `Slider::smart_aim_values`, which lets the user define certain points to snap the slider to when clicking or dragging.
 
 ### Changed 🔧
 * Improved plot grid appearance ([#2412](https://github.com/emilk/egui/pull/2412)).
--- a/crates/egui/src/widgets/slider.rs
+++ b/crates/egui/src/widgets/slider.rs
@@ -1,5 +1,6 @@
 #![allow(clippy::needless_pass_by_value)] // False positives with `impl ToString`
 
+use std::cmp::Ordering;
 use std::ops::RangeInclusive;
 
 use crate::*;
@@ -72,6 +73,8 @@ pub struct Slider<'a> {
     spec: SliderSpec,
     clamp_to_range: bool,
     smart_aim: bool,
+    smart_aim_values: Vec<f64>,
+    snap_radius: f32,
     show_value: bool,
     orientation: SliderOrientation,
     prefix: String,
@@ -118,6 +121,8 @@ impl<'a> Slider<'a> {
             },
             clamp_to_range: true,
             smart_aim: true,
+            smart_aim_values: vec![],
+            snap_radius: 0.1,
             show_value: true,
             orientation: SliderOrientation::Horizontal,
             prefix: Default::default(),
@@ -210,6 +215,25 @@ impl<'a> Slider<'a> {
     /// There is almost no point in turning this off.
     pub fn smart_aim(mut self, smart_aim: bool) -> Self {
         self.smart_aim = smart_aim;
+        self
+    }
+
+    /// Set specific points for the slider to snap to. These points are visualized with small dots.
+    /// This only has an effect when [`smart_aim`] is ON. `snap_radius` defines the range around
+    /// the mouse where the slider will snap to the given values. To be useful, this should be
+    /// somewhat large (and definitely larger than [`InputState::aim_radius`]). Default is 0.1.
+    pub fn smart_aim_values<Num: emath::Numeric>(mut self, smart_aim_values: Vec<Num>, snap_radius: f32) -> Self {
+        self.snap_radius = snap_radius.clamp(0.0, 1.0);
+        self.smart_aim_values = smart_aim_values.into_iter()
+            .map(|n| n.to_f64().clamp(*self.range.start(), *self.range.end()))
+            .collect();
+        self.smart_aim_values
+            .sort_by(|a, b| match a.partial_cmp(b) {
+                Some(ordering) => ordering,
+
+                // NaNs are handled silently
+                None => Ordering::Greater,
+            });
         self
     }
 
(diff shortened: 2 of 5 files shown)


crates/egui/src/widgets/slider.rs
```diff
@@ -72,6 +73,8 @@ pub struct Slider<'a> {
     spec: SliderSpec,
     clamp_to_range: bool,
     smart_aim: bool,
+    smart_aim_values: Vec<f64>,
```
emilk, line 76:
```suggestion
    snap_values: Vec<f64>,
```

crates/egui/src/widgets/slider.rs
```diff
@@ -72,6 +73,8 @@ pub struct Slider<'a> {
...
+    snap_radius: f32,
```
emilk, line 77: I think it's better to just use the normal `aim_radius`

```suggestion
```

crates/egui/src/widgets/slider.rs
```diff
@@ -213,6 +218,30 @@ impl<'a> Slider<'a> {
...
+        self.smart_aim_values = smart_aim_values
+            .into_iter()
+            .map(|n| n.to_f64().clamp(*self.range.start(), *self.range.end()))
+            .collect();
+        self.smart_aim_values
+            .sort_by(|a, b| match a.partial_cmp(b) {
+                Some(ordering) => ordering,
+
+                // NaNs are handled silently
+                None => Ordering::Greater,
```
emilk, line 240: this is unsound. This can make `a < b` and `b < a` for some values of a,b, which could theoretically make the sort hang.

crates/egui/src/widgets/slider.rs
```diff
@@ -213,6 +218,30 @@ impl<'a> Slider<'a> {
...
+            });
```
emilk, line 241:
```suggestion
            .sort_by(|a, b| a.total_cmp(b));
```

emilk (changes requested):
Good start, but I actually think smart-aim and snap-values are orthogonal. That is: I see no reason why you shouldn't be able to turn either feature on/off independently.

If smart-aim is enabled, perform that. If there are snap-values, snap to the closest one (if within aim_radius). Done!

crates/egui/src/widgets/slider.rs
ac-freeman: If only using aim_radius, the snapping will behave something like this:


But do we not also want the flexibility to support this kind of snapping? And to be able to decide for an individual slider what kind of granularity between snap values is supported? That's what I was thinking with having a separate `snap_radius` for each slider.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 * Add `Plot::clamp_grid` to only show grid where there is data ([#2480](https://github.com/emilk/egui/pull/2480)).
 * Add `ScrollArea::drag_to_scroll` if you want to turn off that feature.
 * Add `Response::on_hover_and_drag_cursor`.
+* Add `Slider::smart_aim_values`, which lets the user define certain points to snap the slider to when clicking or dragging.
```
emilk, line 15:
```suggestion
* Add `Slider::smart_aim_values`, which lets the user define certain points to snap the slider to when clicking or dragging ([#2523](https://github.com/emilk/egui/pull/2523)).
```

crates/egui/src/widgets/slider.rs
emilk: Ah, good point. I think these should be two separate modes. One is "snap if close" the other is "only use snap-values".

ac-freeman: Okay, I added that with a `snap_values_only` bool in `Slider`

crates/egui/src/widgets/slider.rs
```diff
@@ -539,15 +563,37 @@ impl<'a> Slider<'a> {
...
+                        closest_distance = distance;
+                    } else {
+                        break;
+                    }
+                }
+
+                // Divide aim_radius by the granularity we want
+                if self.snap_values_only || closest_distance < ui.input().aim_radius() as f64 / 15.0 {
+                    new_value = closest_snap;
+                }
```
emilk, line 594: The docs aren't very good around `position_from_value` etc, but you should measure the distances in ui space, NOT in the space of the slider values.

So do this:

```suggestion
                // We use ui points as the unit for the measurements:
                let mut closest_snap = 0.0;
                let new_value_pos = self.position_from_value(new_value, position_range.clone());
                let mut closest_distance = f64::INFINITY;
                for snap_value in self.snap_values.iter().copied() {
                    let snap_pos = self.position_from_valuee(snap_value, self.range.clone());
                    let distance = (new_value_pos - snap_pos).abs();
                    if distance < closest_distance {
                        closest_snap = snap;
                        closest_distance = distance;
                    } else {
                        break;
                    }
                }

                // Divide aim_radius by the granularity we want
                if self.snap_values_only || closest_distance < ui.input().aim_radius() {
                    new_value = closest_snap;
                }
```

crates/egui_demo_lib/src/demo/sliders.rs
```diff
@@ -135,6 +146,15 @@ impl super::View for Sliders {
                 .text("right"),
         );
 
+        if *snap_aim {
+            let range = *max - *min;
+
+            // Set the smart aim values to 1/3 and 2/3 of the range, for demo-ing
```
emilk, line 152:
```suggestion
            // Set the snap values to 1/3 and 2/3 of the range, for demo-ing
```

emilk (changes requested):

crates/egui/src/widgets/slider.rs
```diff
@@ -578,23 +578,23 @@ impl<'a> Slider<'a> {
...
                         closest_distance = distance;
                     } else {
                         break;
                     }
                 }
 
-                // Divide aim_radius by the granularity we want
-                if self.snap_values_only || closest_distance < ui.input().aim_radius() as f64 / 15.0 {
+                // Multiply aim_radius by the granularity we want
+                if self.snap_values_only || closest_distance < ui.input(|i| i.aim_radius()) * 5.0 {
```
ac-freeman, line 597: I think some kind of multiplier is necessary here, but I'm not sure how we should determine what it is. 

With no multiplier, the aim radius is too small for it to actually snap. The precision is too tight for using a mouse.
2c9d976390

By multiplying the aim radius by 5 as I have here, the snap works more like I expect.
c0c8b43de6

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 * Add `Plot::clamp_grid` to only show grid where there is data ([#2480](https://github.com/emilk/egui/pull/2480)).
 * Add `ScrollArea::drag_to_scroll` if you want to turn off that feature.
 * Add `Response::on_hover_and_drag_cursor`.
+* Add `Slider::snap_values`, which lets the user define certain points to snap the slider to when clicking or dragging ([#2523](https://github.com/emilk/egui/pull/2523)).
```
emilk, line 24: This needs to be moved up to the `Unreleased` section.

Next time, open your PR from another branch then `master`, then I can help out by adding my own commits!

Anyways, after this I will merge it (finally!).

Thank you so much for your work an patience!

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 
 
 ## Unreleased
-
+* Add `Slider::snap_values`, which lets the user define certain points to snap the slider to when clicking or dragging ([#2523](https://github.com/emilk/egui/pull/2523)).
```
emilk, line 8:
```suggestion
```

emilk: we autogenerate the changelog now

Issue #2520, opened 2022-12-28
failed to create glutin window surface


qiuzhiqian:
**Describe the bug**
when i run the egui example hello_world,but it run failed.

```
$ cd examples
$ cargo run -p hello_world
warning: /home/xml/egui-0.20.1/crates/eframe/Cargo.toml: unused manifest key: target.cfg(not(target_arch = "wasm32")).dependencies.glutin.es
    Finished dev [unoptimized + debuginfo] target(s) in 0.35s
     Running `/home/xml/egui-0.20.1/target/debug/hello_world`
2022-12-28T10:02:53.267553Z  INFO winit::platform_impl::platform::x11::window: Guessed window scale factor: 1.25    
libEGL warning: DRI2: failed to authenticate
thread 'main' panicked at 'failed to create glutin window surface: Error { raw_code: Some(12299), raw_os_message: None, kind: BadNativeWindow }', crates/eframe/src/native/run.rs:409:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

when i run it with `RUST_BACKTRACE=1`,i got that:
```
$ RUST_BACKTRACE=1 cargo run -p hello_world
warning: /home/xml/egui-0.20.1/crates/eframe/Cargo.toml: unused manifest key: target.cfg(not(target_arch = "wasm32")).dependencies.glutin.es
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
     Running `/home/xml/egui-0.20.1/target/debug/hello_world`
2022-12-28T10:12:18.211554Z  INFO winit::platform_impl::platform::x11::window: Guessed window scale factor: 1.25    
libEGL warning: DRI2: failed to authenticate
thread 'main' panicked at 'failed to create glutin window surface: Error { raw_code: Some(12299), raw_os_message: None, kind: BadNativeWindow }', crates/eframe/src/native/run.rs:409:18
stack backtrace:
   0: rust_begin_unwind
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/panicking.rs:142:14
   2: core::result::unwrap_failed
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/result.rs:1785:5
   3: core::result::Result<T,E>::expect
[...]

PR #2526, opened 2022-12-29
polish glutin upgrade with glutin-winit crate

coderedart:
1. Cleaned up all the platform specific code by replacing it with `glutin-winit` helper crate.
2. refactored the glutin backend to be more android friendly.

TODO: 
1. once  winit 0.28 is released, we can add an example for eframe + android using `android-activity = "0.4"` and maybe some instructions for eframe-template on how to build


coderedart:
I think this is ready for review. 


the PR is just mostly polishing the glutin upgrade by using glutin_winit helper crate and adding some logging to help with debugging glutin issues. In particular, egui glow using shader lang version constant will cause a seg fault if context comes from Microsoft S/W renderer with gl 1.1 context. this happens if user doesn't have proper drivers installed (apparently the case in virtual machines and remote desktop) for their GPU. 

there could be some improvement using https://github.com/emilk/egui/pull/2541 in future, where we just run the whole glutin window context creation function with `preferEgl` instead of `FallbackEgl`. I want to delay this as it might cause some weird bugs due to interaction of both egl/native displays, as glutin doesn't support destroying/cleaning up these resources yet.

emilk:
What is the motivation behind this PR? It adds a lot of code (and a dependency), so what problems is it solving? What features is it adding?

coderedart:
1. push the context creation responsibility to `glutin-winit`, which needed some ugly platform gated code before. the crate is also maintained within the glutin repo. was recommended by one of the glutin devs during the first PR https://github.com/emilk/egui/pull/2187#issuecomment-1312422231
> It basically hides stuff like x11_visual and DisplayApiPreference, so you don't have platform specific code

2. Some users have problems if we  use Egl API first. so, changed the preference from `PreferEgl` to `FallbackEgl`. according to glutin devs, this was the behavior in older versions and the more reliable one. https://github.com/emilk/egui/pull/2541#issuecomment-1370767582

3. refactor the code to match wgpu backend and be android friendly. mainly, separating context creation (once at startup) from window/surface creation (multiple times over the lifecycle of app). makes it seem like a lot of code, but its just existing code moved around for this. android support meant not always having a winit `Window` and i didn't want to deal with that complexity in the first PR, only focusing on desktop.

4. add a bunch of logging because debugging opengl issues is hard. we had 2 issues about problems on virtual machines, where eframe might segfault for querying `SHADER_LANGUAGE_VERSION` (doesn't exist in opengl 1.1). logs will now immediately make this obvious.  https://github.com/emilk/egui/issues/2545 https://github.com/emilk/egui/issues/2520

coderedart:
Is there still some problem with merging this? The merge conflicts seem to be growing and dealing with git scares me.

If you don't want to merge this, then this can be closed in favor of https://github.com/emilk/egui/pull/2541 .

crates/egui_glow/src/painter.rs
```diff
@@ -106,6 +106,23 @@ impl Painter {
         crate::profile_function!();
         crate::check_for_gl_error_even_in_release!(&gl, "before Painter::new");
 
+        // some useful debug info. all three of them are present in gl 1.1.
+        unsafe {
+            let version = gl.get_parameter_string(glow::VERSION);
+            let renderer = gl.get_parameter_string(glow::RENDERER);
+            let vendor = gl.get_parameter_string(glow::VENDOR);
+            tracing::info!(
```
emilk, line 114:
```suggestion
            tracing::debug!(
```

crates/egui_glow/src/painter.rs
```diff
@@ -106,6 +106,23 @@ impl Painter {
...
+                "\nopengl version: {version}\nopengl renderer: {renderer}\nopengl vendor: {vendor}"
+            );
+        }
+
+        #[cfg(not(target_arch = "wasm32"))]
+        if gl.version().major < 2 {
+            // this checks on desktop that we are not using opengl 1.1 microsoft sw rendering context.
+            // ShaderVersion::get fn will segfault due to SHADING_LANGUAGE_VERSION (added in gl2.0)
+            return Err("egui-glow requires opengl 2.0+. ".to_owned());
```
emilk, line 123:
```suggestion
            return Err("egui_glow requires opengl 2.0+. ".to_owned());
```

crates/egui_glow/examples/pure_glow.rs
```diff
@@ -17,66 +17,91 @@ impl GlutinWindowContext {
...
         let context_attributes =
-            glutin::context::ContextAttributesBuilder::new().build(Some(raw_window_handle));
-        // for surface creation.
-        let (width, height): (u32, u32) = winit_window.inner_size().into();
+            glutin::context::ContextAttributesBuilder::new().build(raw_window_handle);
+        // by default, glutin will try to create a core opengl context. but, if it is not available, try to create a gl-es context using this fallback attributes
+        let fallback_context_attributes = glutin::context::ContextAttributesBuilder::new()
+            .with_context_api(glutin::context::ContextApi::Gles(None))
+            .build(raw_window_handle);
+        let not_current_gl_context = unsafe {
```
emilk, line 68: This is weirdly named. Usually you don't name things after what they are _not_.

crates/eframe/src/native/run.rs
```diff
@@ -342,132 +348,258 @@ mod glow_integration {
         painter: egui_glow::Painter,
         integration: epi_integration::EpiIntegration,
         app: Box<dyn epi::App>,
-
         // Conceptually this will be split out eventually so that the rest of the state
         // can be persistent.
         gl_window: GlutinWindowContext,
     }
+    /// This struct will contain both persistent and temporary glutin state.
```
emilk, line 355:
```suggestion
    }
    
    /// This struct will contain both persistent and temporary glutin state.
```

crates/eframe/src/native/run.rs
```diff
@@ -342,132 +348,258 @@ mod glow_integration {
...
+    ///
+    /// Platform Quirks:
+    /// Microsoft Windows: requires that we create a window before opengl context.
+    /// Android: window and surface should be destroyed when we receive a suspend event. recreate on resume event.
```
emilk, line 359:
```suggestion
    /// * Microsoft Windows: requires that we create a window before opengl context.
    /// * Android: window and surface should be destroyed when we receive a suspend event. recreate on resume event.
```

(markdown treats a single newline as no newline)

crates/eframe/src/native/run.rs
```diff
@@ -342,132 +348,258 @@ mod glow_integration {
...
+            let fallback_context_attributes = glutin::context::ContextAttributesBuilder::new()
+                .with_context_api(glutin::context::ContextApi::Gles(None))
+                .build(raw_window_handle);
+            let gl_context = match gl_config
+                .display()
+                .create_context(&gl_config, &context_attributes)
+            {
+                Ok(it) => it,
+                Err(err) => {
+                    tracing::error!("failed to create context using default context attributes {context_attributes:?} due to error: {err}");
```
emilk, line 474:
```suggestion
                    tracing::warn!("failed to create context using default context attributes {context_attributes:?} due to error: {err}");
```

crates/eframe/src/native/run.rs
```diff
@@ -749,26 +882,25 @@ mod glow_integration {
...
-                        //
-                        // When the application resumes then it will get a new `SurfaceView` but
-                        // we have no practical way currently of creating a new EGL surface
-                        // via the Glutin API while keeping the GL context and the rest of
-                        // our app state. This will likely result in a black screen or
-                        // frozen screen.
-                        //
-                        //self.running = None;
-                    }
+                    tracing::warn!("suspending app");
```
emilk, line 901: why is this a warning?

emilk (approved):
Ok, looks good!

Sorry for the slowness

crates/eframe/src/native/run.rs
coderedart: I haven't verified android support, so eframe might just crash after suspend event when using glutin. thought it would be better to have it clearly visible in logs that the crash is due to suspend event and not something else.

crates/egui_glow/examples/pure_glow.rs
coderedart: comes directly from glutin https://docs.rs/glutin/latest/glutin/context/enum.NotCurrentContext.html

coderedart:
In epi integration, `window_builder` fn was changed to `build_window`. but glutin requires window_builder (because it decides when to actually build the window). so, i split build_window fn into `window_builder` and `build_window` functions.

https://github.com/coderedart/egui/blob/glutin_winit/crates/eframe/src/native/epi_integration.rs#L163

crates/eframe/src/native/epi_integration.rs
```diff
@@ -64,13 +64,12 @@ pub fn read_window_info(
...
-pub fn build_window<E>(
+pub fn window_builder<E>(
     event_loop: &EventLoopWindowTarget<E>,
     title: &str,
     native_options: &epi::NativeOptions,
     window_settings: Option<WindowSettings>,
-) -> Result<winit::window::Window, winit::error::OsError> {
+) -> winit::window::WindowBuilder {
     let epi::NativeOptions {
         always_on_top,
```
emilk, line 74:
```suggestion
```

emilk: build fix

coderedart:
nice. this is done. is it okay to bump glow to 0.12 in a new pr?

emilk:
yes, go for it!

emilk:
Actually, I went ahead myself: https://github.com/emilk/egui/pull/2695

Issue #1834, opened 2022-07-16
Close button Not Clickable, when added inside the Custom Window frame #eframe_library_example_not_working


gouravkhator:
## Issue Description

Even the eframe's close button example [here](https://github.com/emilk/egui/blob/898f4804b7b998ffeb1ff9f457b935e1364d6827/examples/custom_window_frame/src/main.rs#L91) seems to not work on my machine.

When I click on the close button in the custom window frame, it does not let me to click, it thinks that I have to drag the frame.

## Expected Behavior

When I click on the close button, it should close the frame.

## Machine tested on:

```
System:
  Kernel: 5.15.53-1-MANJARO arch: x86_64 bits: 64 compiler: gcc v: 12.1.0
  Desktop: GNOME v: 42.3.1 tk: GTK v: 3.24.34 wm: gnome-shell dm: GDM
    v: 42.0 Distro: Manjaro Linux base: Arch Linux
```

gouravkhator:
#1770 Not sure if these two issues are related. But in issue #1770 also, the `.clicked()` method is not working for `RetainedImage`.

Please help with these two bugs, and confirm if they are related to the same root problem or not.

PR #2533, opened 2023-01-02
Fix close button not working

RadonCoding:
By adding the close button after the title bar drag listener the close button will sense clicks.

Closes <https://github.com/emilk/egui/issues/1834>.


emilk (approved):

emilk:
Nice catch - thanks!

PR #2539, opened 2023-01-03
window starting collapsed state

joffreybesos:
A small PR to added a default collapsed state to the Window widget.\
Currently it always starts as expanded.

So now you can do:
`egui::Window::new("window name").default_open(true)`


Wumpf (approved):
nice addition, thank you!

PR #2552, opened 2023-01-07
Point to caller's location when using log_or_panic

kornelski:
Incorrect use of methods like table's `col()` can panic, but the panic points to egui's guts instead of the call site that user would recognize. This adds `#[track_caller]` to make the panic location point to user's code.


Wumpf (approved):
neat! Thank you

Issue #2395, opened 2022-12-05
egui::widgets::plot: Allow reading/writing of plot items hidden state


Boscop:
I have a lot of items in my plot:

![image](https://user-images.githubusercontent.com/535593/205733081-9eca03eb-ca06-45fc-a1c6-3dbc80177762.png)

Whenever my app starts, the first thing I as a user have to do is disable most of the plot items. Every time the app restarts.
It would be nicer if there was a way to set items to hidden through code instead of only via the mouse at runtime.
And being able to read hidden states of items, for persistence/restoring.

Is this possible somehow, or planned? :)

PR #2557, opened 2023-01-08
`Plot::Items::default_visible` allow hidden on init

haricot:
added feature described in this issue  [#2395](https://github.com/emilk/egui/issues/2395)

### DIFF
--- a/crates/egui/src/widgets/plot/items/mod.rs
+++ b/crates/egui/src/widgets/plot/items/mod.rs
@@ -45,6 +45,8 @@ pub(super) trait PlotItem {
 
     fn highlighted(&self) -> bool;
 
+    fn visible(&self) -> bool;
+
     fn geometry(&self) -> PlotGeometry<'_>;
 
     fn bounds(&self) -> PlotBounds;
@@ -119,6 +121,7 @@ pub struct HLine {
     pub(super) stroke: Stroke,
     pub(super) name: String,
     pub(super) highlight: bool,
+    pub(super) visible: bool,
     pub(super) style: LineStyle,
 }
 
@@ -129,6 +132,7 @@ impl HLine {
             stroke: Stroke::new(1.0, Color32::TRANSPARENT),
             name: String::default(),
             highlight: false,
+            visible: true,
             style: LineStyle::Solid,
         }
     }
@@ -136,6 +140,12 @@ impl HLine {
     /// Highlight this line in the plot by scaling up the line.
     pub fn highlight(mut self, highlight: bool) -> Self {
         self.highlight = highlight;
+        self
+    }
+
+    /// Hidden on initialization. Default is true.
+    pub fn visible(mut self, visible: bool) -> Self {
+        self.visible = visible;
         self
     }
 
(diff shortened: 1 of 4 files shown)


crates/egui/src/widgets/plot/items/mod.rs
```diff
@@ -139,6 +143,12 @@ impl HLine {
         self
     }
 
+    /// Hidden on initialization. Default is true.
+    pub fn visible(mut self, visible: bool) -> Self {
```
emilk, line 147:
```suggestion
    /// Is the item visible on initialization? Default is `true`.
    #[inline]
    pub fn visible(mut self, visible: bool) -> Self {
```

emilk: and everywhere else

crates/egui/src/widgets/plot/legend.rs
```diff
@@ -166,29 +169,30 @@ impl LegendWidget {
...
-                        let checked = !hidden_items.contains(item.name());
-                        LegendEntry::new(color, checked)
-                    });
+                entries.entry(item.name().to_owned()).or_insert_with(|| {
+                    let color = item.color();
+                    let checked = !hidden_items.contains(item.name());
+                    let visible = item.visible();
+                    let hiddened= !(checked && visible || !checked && !visible);
+
+                    if hiddened {
```
emilk, line 188:
```suggestion
                    if checked != visible {
```

emilk: …though I struggle to follow he logic

crates/egui/src/widgets/plot/items/mod.rs
emilk: Also, I think this is better named `default_visible`, since it shouldn't control the current visibility, just the visibility on the first time we show the plot

emilk (changes requested):

crates/egui/src/widgets/plot/items/mod.rs
haricot: Thank ! I did not suspect that it was necessary to add [#inline]  and renamed to default_visible which indicates well that it is an initialization parameter.

crates/egui/src/widgets/plot/legend.rs
haricot: another possible logic without default_visible but with visible checking if it's in hidden_items.
(_requires a PR to access hidden items, which seems necessary to know if the item is hidden in a future case.
and surely a specific logic for the visible._)
```
let visibility_points = !hidden_items.contains("Points");
plot_ui.points(points.name("Points").visible(visibility_points);
```


although with default_visible we set a state from the start, we don't have access to its modified state.
while with visible it will allow to have its state hidden or not dynamically.

I don't think they are the same thing all the same.
with default_visible we define a state from the start which is clear to me.
When to know if it is in hidden_items it is another problem.
I think I have a preference for default_visible to solve this feature.
But maybe an alternative and necessary PR to compare.

emilk:

See also:

PR #2569, opened 2023-01-10
Add `Plot::allow_auto_bounds`

haricot:
Whether to allow auto bounds. Default: `true`.
If false, when the bounds is modified it will not be recalculated.


crates/egui/src/widgets/plot/mod.rs
```diff
@@ -437,6 +439,14 @@ impl Plot {
         self
     }
 
+    /// Whether to allow auto bounds. Default: `true`.
+    /// `false` keep bounds, it gives the ability with a pointer down to the drawn outside the plot area 
+    /// without ask auto bounds.
```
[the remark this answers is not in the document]
haricot: This represents an example of use and does not describe the functionality.
I meant " If `false`, when the bounds is modified it will not be recalculated." but that doesn't seem necessary.

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -437,6 +439,12 @@ impl Plot {
         self
     }
 
+    /// Whether to allow auto bounds. Default: `true`.
```
emilk, line 442: I think you should try to explain what it means for the user to set this to `true` or `false`.

Does "bounds" here mean the view of the plot (zoom/pan)?
How does this relate to `PlotUi::set_plot_bounds`?
What bounds will be used when this is set to `false`?

haricot: I suggest this:
> If `false`, it set `bounds_modified` to true to prevent auto adjusting bounds,
> also it check if the user double clicked to reset bounds. If so, it set `bounds_modified` to false.
> 
> It's needed when a new plot is started, such as when resetting egui memory and pan/zoom that
> have not been used before, as the limits can be saved as persistent data.
> 
> Currently, if `allow_auto_bounds(false)` and `plot_ui.set_plot_bounds(PlotBounds)` are used in
> same time `set_plot_bounds` will be ignored.
>
> TODO: Refactor `Plot::allow_auto_bounds` and `Plot::set_plot_bounds` feature to work together.


Possible solution for `allow_auto_bounds` and `set_plot_bounds` by passing something like this when initializing plot:
`Plot::new("myplot").allow_auto_bounds(false).preset_bounds(plotbounds)` or something a enum.

haricot: or something like:
```
enum LogicBound{
  SkipRecalculated,
  Bounds(PlotBounds)
}

Plot::set_plot_bounds( LogicBounds::SkipRecalculated )
```

I will try to test this.

haricot: or rather just this:

widgets/plot/mod.rs 
```
        .unwrap_or_else(|| 
            PlotMemory {
            bounds_modified: if let Some(custom_bounds) = custom_bounds {
                true.into()
                } else { 
                false.into()
                }  ,
            hovered_entry: None,
            hidden_items: Default::default(),
            last_plot_transform: PlotTransform::new(
                rect,
                if let Some(custom_bounds) = custom_bounds {
                    custom_bounds
                    } else { 
                        min_auto_bounds
                    },
                center_axis.x,
                center_axis.y,
            ),
            last_click_pos_for_zoom: None,
        });
```

haricot: I opened  issue  #3371 related this pull request,  how i think it resolve this indirect approach.


Also i realize i think we can add  set_bounds and dynamic  zoom and translate under  process default bounds (currently enforce aspect ratio)  like vec ![bounds_modifications] but  in setting by using factor_delta_changed slider.

```
let data_zoom_aspect = scalar_aspect_zoom.current ;
let last_data_aspect_zoom  = scalar_aspect_zoom.last_changed;

let data_zoom_aspect  = if last_data_aspect_zoom > data_zoom_aspect {
    1.0 * (data_zoom_aspect / last_data_aspect_zoom)
} else {
    (data_zoom_aspect / last_data_aspect_zoom) / 1.0
};

scalar_aspect_zoom.last_changed  =  scalar_aspect_zoom.current ;

scalar_aspect_zoom.factor_delta_changed  =  data_zoom_aspect;

```

emilk:
I think this has already been added by a different PR:

https://docs.rs/egui_plot/latest/egui_plot/struct.PlotUi.html#method.set_auto_bounds

PR #2570, opened 2023-01-10
Add `Plot::is_hovered_entry` Returns `true` if the plot area is hovered entry.

haricot:


haricot:
Closed in favor of a new PR `Plot::is_hover_rect_legend` Returns `true` if the plot pointer is hovering over the rectangle containing the plot legend.

PR #2571, opened 2023-01-10
Add `Plot::plot_interact` to check if the plot area has interaction pointer.

haricot:


emilk (commented):
Wouldn't `pointer_coordinate().is_some()` suffice for most users?

haricot:
In the case where we interact with the pointer in the plot window to draw, if we are outside the plot window with the pointer up when the pointer is down, it will draw in the plot window while we don't want it.

_in fact, I would like to submit a demo for egui that I call "Plotboard" (in preparation) which illustrates among other things this functionality without external dependency.  The objective being to be able to use these functionalities necessary in a future project on this subject._

With this PR: we can allow drawing outside the plot window when the pointer down comes out of the plot window and not allow when we are outside the plot window the pointer up and it is becomes down.

emilk:
Superseded by https://github.com/emilk/egui/pull/3223

PR #2580, opened 2023-01-13
Add Chinese font: 得意黑 Smiley Sans

xuxiaowei-com-cn:
https://github.com/atelier-anchor/smiley-sans
Use version v1.1.1


emilk:
Sadly, we cannot add fonts for all languages in the world. Instead, we make it easy for users to add their own fonts.

PR #2621, opened 2023-01-23
Add ability to hide button backgrounds

emilk:
With this PR you can turn off background for buttons.

This required a bit of work. We want to be able to turn off background colors for buttons, comboboxes, dragvalues, etc, but keep them for checkbox-boxes, radiobutton-circles, slider-backgrounds, etc. Previously these were all controlled by `WidgetVisuals::bg_fill`. Now a new ~`WidgetVisuals::optional_bg_fill`~ `WidgetVisuals::weak_bg_fill` has been added, which is used by buttons, comboboxes, dragvalues etc. `weak_bg_fill` can be safely set to `Color32::TRANSPARENT` without messing up checkboxes, radiobuttons, sliders, etc. ~I don't really like the name `optional_bg_fill`, and suggestions are welcome.~ I kept the old `bg_fill` named as-is, to avoid breaking to much 3rd party code.

Turning off the background on ComboBoxes requires making them tighter, to keep the arrow close to the text, so `Spacing::combo_width` was added.

En example of this style:

image

### DIFF
--- a/crates/egui/src/containers/combo_box.rs
+++ b/crates/egui/src/containers/combo_box.rs
@@ -268,7 +268,7 @@ fn combo_box_dyn<'c, R>(
             // Currently selected value's text will be wrapped if needed, so occupy the available width.
             ui.available_width()
         } else {
-            // Occupy at least the minimum width assigned to Slider and ComboBox.
+            // Occupy at least the minimum width assigned to ComboBox.
             let width = width.unwrap_or_else(|| ui.spacing().combo_width);
             width - 2.0 * margin.x
         };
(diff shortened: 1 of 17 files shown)


crates/egui/src/containers/combo_box.rs
```diff
@@ -263,18 +262,20 @@ fn combo_box_dyn<'c, R>(
 
     let margin = ui.spacing().button_padding;
     let button_response = button_frame(ui, button_id, is_popup_open, Sense::click(), |ui| {
+        let icon_spacing = ui.spacing().icon_spacing;
         // We don't want to change width when user selects something new
         let full_minimum_width = if wrap_enabled {
             // Currently selected value's text will be wrapped if needed, so occupy the available width.
             ui.available_width()
         } else {
             // Occupy at least the minimum width assigned to Slider and ComboBox.
```
Wumpf, line 271:
```suggestion
            // Occupy at least the minimum width assigned to ComboBox.
```

Wumpf (approved):
`weak_bg_fill`? then one could distinguish between elements having regular vs weak backgrounds. Essentially it comes down to having different kinds of backgrounds that can be styled individually, less so much about them being allowed to be transparent, no?

PR #2558, opened 2023-01-08
`Plot::Item::allow_hover` give possibility to masked the interaction on hovered item

haricot:
This is particularly interesting if you want to authorize a single hover tooltip on an item in the event of an interaction.

PR #2628, opened 2023-01-25
Add `Plot::hover_indexes` & `Plot::items::group`  to retrieve indexes of the source

haricot:
In the plot when the hover function is used, I notice that we can retrieve the indexes from the source without additional cost (without iterating twice).

Once we have the indexes, given that the items of the plot are presented in the order in which they were pushed, if one of them is hidden in the plot via the legend then the order of the items shifts and it is therefore difficult to know what source items it is. (or else we rely on hidden_item to update the indexes but if they don't have a name we can't know, they won't appear in hidden_items).
Hence the addition of groups which allows this to be overloaded via `SourceIndex`.

This allows to find the indexes and the source of the item in the following cases:
     if the group has not been defined, we can still recover but is retain items.
     if the group is defined and sub_index is true  and it is a line for example.
     if group is defined and sub_index is false and it is points in a vector that nests others.
     if multiple groups are defined and they represent a graph.

This can allow to select the element with which we can interact and it would not necessarily need the PR allow_hover https://github.com/emilk/egui/pull/2558.

With groups this leaves enough flexibility to be able to present the data as desired.

Alternative would be another way to generate the hidden_items so that one is not lag if one becomes hidden.
it wouldn't need retain items, but such a visible filter.
This would involve keeping track of the index when initializing the plot and then we would only need the optional sub_index but it seems to me that we would not have the point indices indexes embedded in several vectors for example .


haricot:
I need to go back and use `#[allow(clippy::needless_pass_by_value)]` instead Option  to make it typed better.

haricot:
and i need to refactor for use Vec<usize> for retrieve all nested shape.

icanwalkonwater:
Hi, is there any news on the status of this PR ?

There is still interest in it as evidenced by myself and several others (#1778, #3028, #3784) asking for a similar thing.

haricot:
I think I have something more concise but I've gotten a little confused with other features that seem to me to take precedence.  I think the principle is there.
I've had a lot of setbacks lately, I can't guarantee anything about the progress of this pull request yet.  In any case, I would like to get back to it.

emilk:

See also:

PR #2633, opened 2023-01-26
zlayer (continued)

emilk:
Continuation of https://github.com/emilk/egui/pull/2482


mxsdev:
Hey, sorry for not getting to this when you got back! I got a little busy this week so it was pretty unfortunate timing. Thanks for taking care of it - feel free to lmk if there's anything more you'd like from me.

emilk:
Oh no worry at all, I was just curious to try this out a bit :)

I tried yesterday putting panel frames on a negative Z layer so that widgets could use a smaller negative Z to paint behind other widgets, but it didn't work on my first go, and then I got distracted by something else… I would like to get something like that tried and working before merging.

mxsdev:
@emilk 

Gotcha! If I recall correctly, with this implementation ordering happens only within the same layer id - that is, Z-ordering determines the ordering within a specific layer. So, it's different than CSS z-order for example. So ordering something like a widget behind a panel probably wouldn't work, since they're on separate layers.

I originally tried to implement it like in CSS, but it becomes pretty complicated. For example, if a widget has z-order 1, but all windows have z-order 0 (and are on the same `Order`), then the widget will appear over all other windows. If you're making a node editor and want edges to appear beneath nodes, for example, then this behavior isn't desirable.

I think to solve this problem you'd need to kinds of Z-ordering. One "inner" z ordering for the order of elements on a layer, and one "outer" z ordering for the order of layers among other layers. One potential direction here would be to expand the `Order` struct to be user definable.

If you think having two kinds of Z-ordering is a good solution I can merge your changes into my PR and implement it. It's a little cumbersome, though, so maybe I'm missing a more elegant solution.

mxsdev:
To expand on the two-z system: The z in `ZLayer` would now determine the z order-ing between different layers. The z in `PaintList` would determine the z order-ing between elements on the same layer. `with_zlayer` is how you use the former, and `with_z` controls the z ordering on the same layer.

One other challenge here is that two `ZLayer`s could have the same Id but different Z levels. I'm assuming we would tackle this by hashing `ZLayer` with both it's Id and it's `ZOrder`, but it might be confusing to the user if two layers with the same Id are considered "different" internally.

polwel:
This'd be a great feature.

I've been browsing the PR by @mxsdev (also with the intent that perhaps I can make myself useful 😉) , and it seems it stalled over the unresolved question of what the draw order should be between shapes on different `AreaLayers`. I'd like to better understand what the use case is here.

(Asking my question here rather than in the PR, as it's bound to get lost otherwise.)

As far as I can tell, `AreaLayer` serves two purposes: assigning draw commands a coarse `Order` (presumably so that e.g. modal windows are always in the foreground?), and a view transform. IMHO, these are unrelated concepts: in a scenario like `egui_plot`, it's reasonable that all the elements in the scene share a common transform, but also that the grid lines are always drawn behind the traces, labels, etc. That should already be possible to accomplish with the new z-ordering.

Maybe I am asking why we still need `AreaLayerId::Order` at all.

Could we replace it with the concept of a 'z-offset' that is applied when sorting the shape list? Then `Order::Background` and friends would just be some value, say, -1000.0, by convention. If we nest layers (for instance for a pannable scene in a sub-window), we could just increment the z-offset for the new layer by some amount, and a widget that paints itself at z=0 would by default still be drawn on top.

Something else: I see that since this PR was created, the concept of `WidgetRect` was introduced, which _also_ uses `LayerId`. How exactly does this play into things? AFAICT this is yet _another_ kind of layering, which UI receives _input events_. Having this separate from the UI hierarchy makes sense to me! But I am still confused...

PR #2637, opened 2023-01-27
add `Plot::items::Text::font_styles` and `Plot::items::Text::font_size`

haricot:
In plot items text:
Set `font_style` of the text. Default is `TextStyle::Small`.
or
Set `font_size`  of the text.
If you want to use a custom font size, you can use `font_style` instead.
This will override the font size of the `font_style`.
The font key name is defined with prefixed by "Plot_" the "name" of `FondID` spaced
by a underscore and the "size" to prevent redundant `text_styles` in ui.

### DIFF
--- a/crates/egui/src/widgets/plot/items/mod.rs
+++ b/crates/egui/src/widgets/plot/items/mod.rs
@@ -695,14 +695,13 @@ impl Text {
     /// Set Font size of the text.
     /// If you want to use a custom font size, you can use `font_style` instead.
     /// This will override the font size of the `font_style`.
-    /// How are defined name font: prefixed by "Plot_" the "name" of FondID spaced by a underscore and "size"
-    /// to prevent redundant text_styles.
+    /// The font key name is defined with prefixed by "Plot_" the "name" of `FondID` spaced
+    /// by a underscore and the "size" to prevent redundant `text styles` in ui.
     pub fn font_size(mut self, font_size: f32) -> Self {
         self.font_size = Some(font_size);
         self
     }
-    
-    
+
     /// Name of this text.
     ///
     /// This name will show up in the plot legend, if legends are turned on.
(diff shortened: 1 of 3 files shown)


crates/egui/src/widgets/plot/items/mod.rs
```diff
@@ -682,6 +686,22 @@ impl Text {
...
+
+    /// Set Font size of the text.
+    /// If you want to use a custom font size, you can use `font_style` instead.
+    /// This will override the font size of the `font_style`.
+    /// The font key name is defined with prefixed by "Plot_" the "name" of `FondID` spaced
+    /// by a underscore and the "size" to prevent redundant `text styles` in ui.
+    pub fn font_size(mut self, font_size: f32) -> Self {
+        self.font_size = Some(font_size);
+        self
+    }
```
emilk, line 703: A more general approach is to be able to set `FontId`. Users can still use `TextStyle` to get a `FontId` via `TextStyle::resolve`

crates/egui/src/widgets/plot/items/mod.rs
```diff
@@ -703,10 +723,31 @@ impl PlotItem for Text {
...
+            {
+                ui.style_mut().text_styles.insert(
+                    TextStyle::Name(name_font.clone().into()),
+                    FontId {
+                        size: self.font_size.unwrap(),
+                        family: Default::default(),
+                    },
+                );
+            }
+            TextStyle::Name(name_font.into())
```
emilk, line 742: this is a bad idea. There is no need to modify the ui style. Just `FontId` instead!

emilk (changes requested):

crates/egui/src/widgets/plot/items/mod.rs
haricot: Thank !  I need to integrate these concepts particularly FontId and AB_glyph. Especially since what I really wanted was for the text to appear frozen in the plot, ie when zooming in the text should appear to grow.

crates/egui/src/widgets/plot/items/mod.rs
haricot: Thank you for these clarifications !

haricot:
Closed in favor better approach.

PR #2649, opened 2023-01-30
make dragvalue textedit style consistent with button

xyning:
* use the same alignment as button in textedit
* expand width to fit texts instead of clipping


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -67,6 +67,8 @@ pub struct TextEdit<'t> {
     desired_height_rows: usize,
     lock_focus: bool,
     cursor_at_end: bool,
+    align: Align,
+    use_content_width: bool,
```
emilk, line 71: please add docstrings!

crates/egui/src/widgets/drag_value.rs
```diff
@@ -466,7 +466,7 @@ impl<'a> Widget for DragValue<'a> {
                 .memory_mut(|mem| mem.drag_value.edit_string.take())
                 .unwrap_or_else(|| value_text.clone());
             let response = ui.add(
-                TextEdit::singleline(&mut value_text)
+                TextEdit::singleline_button(&mut value_text, ui.layout().horizontal_align())
```
emilk, line 469: Better to add `align` and `use_content_width` helpers instead of `singleline_button`

```suggestion
                TextEdit::singleline(&mut value_text)
                    .align(ui.layout().horizontal_align())
                    .use_content_width(true)
```

crates/egui/src/widgets/text_edit/builder.rs
emilk, line 71: Can we find a better name for `use_content_width`, like `follow_text_width`? Or maybe the opposite makes more sense `clip_text: bool` ?

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -547,7 +562,11 @@ impl<'t> TextEdit<'t> {
             cursor_range = Some(new_cursor_range);
         }
 
-        let mut text_draw_pos = response.rect.min;
+        let mut text_draw_pos = match align {
+            Align::Min => response.rect.min,
+            Align::Center => response.rect.center_top() - Vec2::new(galley.size().x, 0.0) / 2.0,
+            Align::Max => response.rect.right_top() - Vec2::new(galley.size().x, 0.0),
+        };
```
emilk, line 569: If you make `align` an `Align2`, then you can use `Align2::align_size_within_rect`

crates/egui/src/widgets/text_edit/builder.rs
xyning: added to the helper functions

emilk (approved):
Looks great; works great!

PR #2650, opened 2023-01-30
select all text when drag value gets focus

xyning:
currently drag value textboxes reuse the cursor state from the previous focus, which doesnt really make sense imo.

this patch ensures all texts are selected when a drag value is clicked, so user can type directly to overwrite the value.
seems that other common ui toolkits also behave like this.


emilk (approved):
Great catch - thanks!

Issue #2652, opened 2023-01-30
`PlotUi::set_plot_bounds` do not change Plot bounds properly


lvaroqui:
Bug still present in latest master :white_check_mark: 

**Describe the bug**

`PlotUi::set_plot_bounds` and `PlotUi::translate_bounds` do not change Plot bounds properly.

**To Reproduce**
1. Run the following code without persistence (which tends to hide the problem):
```rust
let sin: PlotPoints = (0..1000)
                .map(|i| {
                    let x = i as f64 * 0.01;
                    [x, x.sin()]
                })
                .collect();
            let line = Line::new(sin);
            Plot::new("my_plot")
                .view_aspect(2.0)
                .allow_boxed_zoom(false)
                .allow_double_click_reset(false)
                .allow_drag(false)
                .allow_scroll(false)
                .allow_zoom(false)
                .show(ui, |plot_ui| {
                    plot_ui.set_plot_bounds(PlotBounds::from_min_max([-1., -1.], [1., 1.]));
                    plot_ui.line(line)
                });
```

**Expected behavior**
I'd expect the plot to have `[-1., -1.], [1., 1.]` bounds, however it defaults to showing the whole line as is the default behavior. 

**Screenshots**

**Desktop:**
 - OS: Ubuntu
 - Version: 22.04

**Additional context**
After some digging, I found out that the `PlotUi::set_plot_bounds` and `PlotUi::translate_bounds` do not enable the `bounds_modified` variable, which then makes `PlotUi` fall back to `min_auto_bounds`. ([see here](https://github.com/lvaroqui/egui/blob/8ce0e1c5206780e76234842b94ceb0edf5bb8b75/crates/egui/src/widgets/plot/mod.rs#L847-L853))

I have a fix that I will share with a PR.

PR #2653, opened 2023-01-30
Fix set_plot_bounds

lvaroqui:
Closes https://github.com/emilk/egui/issues/2652.


s-nie:
Can you check if https://github.com/emilk/egui/pull/2410 would fix it as well?

lvaroqui:
Just tested, it seems like so yes!

PR #2666, opened 2023-02-03
Clear color values are now explicitely sent to the rendering backend as-is.

Wumpf:
Previously, converting from Color32 to Rgba caused an srgb->linear conversion. This conversion is incorrect if the backbuffer doesn't perform automatic conversion from linear->srgb (lack of this conversion is generally what egui assumes!).

Manually tested on Mac native & web both with glow & wgpu.
(although I only checked on the exact clear color with wgpu in native and web, but this should be pretty universal)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 
 
 ## Unreleased
+* ⚠️ BREAKING: `App::clear_color` now expects you to return a raw float array ([#TODO](https://github.com/emilk/egui/pull/TODO)):
 * ⚠️ BREAKING: `egui::Context` now use closures for locking ([#2625](https://github.com/emilk/egui/pull/2625)):
   * `ctx.input().key_pressed(Key::A)` -> `ctx.input(|i| i.key_pressed(Key::A))`
   * `ui.memory().toggle_popup(popup_id)` -> `ui.memory_mut(|mem| mem.toggle_popup(popup_id))`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 
 
 ## Unreleased
-* ⚠️ BREAKING: `App::clear_color` now expects you to return a raw float array ([#2666](https://github.com/emilk/egui/pull/2666)):
 * ⚠️ BREAKING: `egui::Context` now use closures for locking ([#2625](https://github.com/emilk/egui/pull/2625)):
   * `ctx.input().key_pressed(Key::A)` -> `ctx.input(|i| i.key_pressed(Key::A))`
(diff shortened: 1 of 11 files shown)


Wumpf:
converted to draft, found an issue on web @ wgpu with this after all

Wumpf:
Was false alarm - the app ;) I tested with has a different eframe::App impl for web and native. Web had a similar background color but not the same.
So works as expected on web & native

crates/ecolor/src/color32.rs
```diff
@@ -198,4 +198,19 @@ impl Color32 {
         // we need a somewhat expensive conversion to linear space and back.
         Rgba::from(self).multiply(factor).into()
     }
+
+    /// Converts to floating point values in the range 0-1 without any color space conversion.
+    ///
+    /// Use this with great care! In almost all cases, you want to convert to [`crate::Rgba`] instead
+    /// in order to obtain linear space color values.
+    #[inline]
+    pub fn to_srgba_f32(self) -> [f32; 4] {
```
emilk, line 207:
```suggestion
    /// Converts to floating point values in the range 0-1 without any gamma space conversion.
    ///
    /// Use this with great care! In almost all cases, you want to convert to [`crate::Rgba`] instead
    /// in order to obtain linear space color values.
    #[inline]
    pub fn to_normalized_gamma_f32(self) -> [f32; 4] {
```

I got schooled on this a couple of years ago: sRGB is the name of color space. It can be encoded in many ways, most commonly in gamma-space and in linear space. But it's still the same color space.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 
 
 ## Unreleased
+* ⚠️ BREAKING: `App::clear_color` now expects you to return a raw float array ([#2666](https://github.com/emilk/egui/pull/2666)):
```
emilk, line 8: -> `crates/eframe/CHANGELOG.md`

emilk (changes requested):
good find!

Wumpf:
Thanks for fixing the comments and explaining srgb-gamma again. We talked about this not too long ago 🤔

crates/eframe/src/epi.rs
```diff
@@ -149,13 +149,21 @@ pub trait App {
...
+    /// Background color values for the app, e.g. what is sent to `gl.clearColor`.
+    ///
     /// This is the background of your windows if you don't set a central panel.
-    fn clear_color(&self, _visuals: &egui::Visuals) -> egui::Rgba {
+    ///
+    /// ATTENTION:
+    /// Since these float values go to the render as-is, any color space conversion as done
+    /// e.g. by converting from [`egui::Color32`] to [`egui::Rgba`] may cause incorrect results.
+    /// egui recommends that rendering backends use a normal "gamma-space" (non-sRGB-aware) blending,
+    ///  which means the values you return here should also be in sRGB gamma-space in the 0-1 range.
```
emilk, line 160:
```suggestion
    ///  which means the values you return here should also be in `sRGB` gamma-space in the 0-1 range.
```

Issue #2642, opened 2023-01-28
Capturing egui's frames


amfaber:
Hi, loving the crate!
Question / feature request: Is it possible to capture the frame rendered by egui?

Context:
I've built a simple UI to visualize some image processing for microscopy, and everything is working flawlessly!
I am using a custom wgpu shader to map the intensities of the data to an image with a colormap, and then adding some circles and lines on top with epaint. However, at this point it would be awesome if I could export the visuals as a video, for embedding in presentations and the like. I had initially hoped that I could simply get a handle to the surface texture that egui is rendering to (I'm using the wgpu backend). From that I would be able to get the final image back from the GPU to the CPU and do anything with it I like from there (save to an image, concatenate multiple images for a video, cut out parts of the frame that is UI and not video etc). However I have not been able to find a way to access the surface texture from egui / egui-wgpu / eframe / egui-winit.

Is it possible? If not, is it a feature that is plausible to add?
Is there some other, better way to accomplish what I want to do?

Thanks for the awesome work :)

PR #2676, opened 2023-02-05
eframe: capture a screenshot using `Frame::request_screenshot`

amfaber:
Hi!

As per https://github.com/emilk/egui/issues/2642, I would like to capture the frames produced by egui.
I decided to take a stab at implementing the feature myself, and hopefully contributing back to this awesome repo. The implementation concerns two distinct areas.

1. Exposing the possibility for the app the communicate to the backend that it would like to capture the current frame, as well as allowing the backend to hand the data back to the app. This is all done in eframe.
2. Implementing the actual screen capture for the wgpu backend.

As I am only familiar with wgpu, I have not attempted a glow implementation.

### Description of the changes
1. is fairly straightforward. It adds the boolean field "pixels_requested" to `AppOutput`, allowing the app to communicate to the backend that it wants a screencapture, as well as the field "pixel_data" to `Frame`, in order to be able to pass the data back to the app during `App::post_rendering`.

2. is a little bit heavier, but localized to the wgpu backend. The fundamental strategy is that when a screencapture is requested, instead of rendering to the surface texture, we do all the rendering of the frame to a texture that is `wgpu::TextureUsages::COPY_SRC`, which allows us to copy the frame's data to both the surface texture as well as a `wgpu::Buffer` which is `wgpu::BufferUsages::MAP_READ`, from which the data can be loaded back onto the cpu. This new pair of  texture and buffer need to have the same dimensions as the surface texture, and so they are updated if they need to be used for a frame capture but their dimensions no longer match up with the surface. They are never initialized if a screen capture is never requested.
As the intention is to deliver back a Vec<u8> of rgba values for the screen, the bytes need to be reshuffled in the case that the surface texture is `wgpu::TextureFormat::Bgra8Unorm`. Currently, screen capturing is only supported if the surface is either `wgpu::TextureFormat::Rgba8Unorm` or `wgpu::TextureFormat::Bgra8Unorm`, panic'ing if the screen is another format AND a capture is requested.

### Concerns
I have tried to make 1. as non-intrusive as possible, still only exposing a `&Frame` to the app's `post_rendering`, while allowing the app to take the generated screenshot from `Frame` via compile-time interior mutability with `std::cell::Cell`.
One thing that bugs me is that there is now another flag on `AppOutput` that isn't handled by either `handle_app_output` or `handle_platform_output`, but since the nature of the feature is so tightly coupled with the rendering backend, I don't see a way around that.

As for 2., it ended up being very coupled to `Painter::paint_and_update_textures`. Again I don't see a way around this, as during testing on my Macbook, I found that `wgpu::TextureUsages::COPY_SRC` is not an allowed usage for the surface texture, necessitating that we instead make the surface `wgpu::TextureUsages::COPY_DST` and render to a separate texture which is `wgpu::TextureUsages::COPY_SRC`. The fact that we need to render to a completely different texture when we intend to take a screencapture means that we will have to interact closely with `Painter::paint_and_update_textures`. My current solution is by adding a boolean argument for whether we are doing a screen capture, and having the function return a Some(Vec<u8>) only if a capture was requested and it succeded, and otherwise None.

For now I have handled all the fallible operations (except if the surface format isn't recognized) through Options, but I think would probably make sense to make an Error type for this that can be exposed to the user in `Frame::frame_pixels`. Failure cases include:
Trying to get the frame pixels without having requested them first
Attempting to use this feature with a backend that doesn't support it yet.
Using it with a surface format that isn't covered by out cases of how to convert to rgba.
A failure during the callback of `wgpu::BufferSlice::map_async`.


crates/egui-wgpu/src/winit.rs
```diff
@@ -13,6 +13,66 @@ struct SurfaceState {
     height: u32,
 }
 
+// A texture and a buffer for reading the rendered frame back to the cpu.
+// The texture is required since wgpu::TextureUsages::COPY_DST is not an allowed
+// flag for the surface texture on all platforms. This means that anytime we want to
+// capture the frame, we first render it to this texture, and then we can copy it to
+// both the surface texture and the buffer, from where we can pull it back to the cpu.
+struct CaptureState {
```
emilk, line 21:
```suggestion
/// A texture and a buffer for reading the rendered frame back to the cpu.
/// The texture is required since wgpu::TextureUsages::COPY_DST is not an allowed
/// flag for the surface texture on all platforms. This means that anytime we want to
/// capture the frame, we first render it to this texture, and then we can copy it to
/// both the surface texture and the buffer, from where we can pull it back to the cpu.
struct CaptureState {
```

crates/egui-wgpu/src/winit.rs
```diff
@@ -274,22 +359,94 @@ impl Painter {
...
+        buffer_slice.map_async(wgpu::MapMode::Read, move |v| {
+            drop(sender.send(v));
+        });
+        device.poll(wgpu::Maintain::WaitForSubmissionIndex(id));
+        receiver.recv().ok()?.ok()?;
+
+        let to_rgba = match tex.format() {
+            wgpu::TextureFormat::Rgba8Unorm => [0, 1, 2, 3],
+            wgpu::TextureFormat::Bgra8Unorm => [2, 1, 0, 3],
+            _ => panic!("Video capture not supported for the used surface format"),
```
emilk, line 411: let's use `tracing::error` instead, and also include the unsupported format in the error message

crates/eframe/src/epi.rs
```diff
@@ -672,6 +674,28 @@ impl Frame {
         self.storage.as_deref()
     }
 
+    /// Request the current frame's pixel data. Needs to be retrieved by calling [`eframe::Frame::frame_pixels`]
+    /// during [`eframe::App::post_rendering`].
+    pub fn request_pixels(&mut self) {
```
emilk, line 679: Maybe this should be called something like `take_screenshot` instead?

crates/eframe/src/epi.rs
```diff
@@ -672,6 +674,28 @@ impl Frame {
         self.storage.as_deref()
     }
 
+    /// Request the current frame's pixel data. Needs to be retrieved by calling [`eframe::Frame::frame_pixels`]
+    /// during [`eframe::App::post_rendering`].
```
emilk, line 678: This feature would be a lot more useful if the user could call `frame_pixels` at any time. Basically `request_pixels` would set some values, and they would remain until the user retrieves them

crates/eframe/src/epi.rs
```diff
@@ -672,6 +674,28 @@ impl Frame {
...
+    pub fn cancel_request_pixels(&mut self) {
+        self.output.pixels_requested = false;
+    }
+
+    /// During [`eframe::App::post_rendering`], use this to retrieve the pixel data that was requested during
+    /// [`eframe::App::update`] via [`eframe::Frame::request_pixels`]. Currently only implemented with wgpu backend.
+    /// Returns None if
+    ///     Called in [`eframe::App::update`]
+    ///     [`eframe::Frame::request_pixels`] wasn't called on this frame during [`eframe::App::update`]
+    ///     The rendering backend doesn't support this feature (yet). Currently only implemented for the wgpu backend.
```
emilk, line 693: Implementing it for `glow` shouldn't be much work at all. You can read from the framebuffer in OpenGL. See `examples/screenshot/src/main.rs` for an example!

crates/eframe/src/epi.rs
```diff
@@ -672,6 +674,28 @@ impl Frame {
...
+    ///     Retrieving the data was unsuccesful in some way.
+    pub fn frame_pixels(&self) -> Option<Vec<u8>> {
```
emilk, line 695: This needs to document the format (premultiplied RGBA), or switch it to using `egui::Color32`.

Does the pixel data start from the top of the screen, or the bottom?

crates/eframe/src/epi.rs
```diff
@@ -672,6 +674,28 @@ impl Frame {
...
+    /// Cancel a request made with [`eframe::Frame::request_pixels`].
+    pub fn cancel_request_pixels(&mut self) {
+        self.output.pixels_requested = false;
+    }
+
+    /// During [`eframe::App::post_rendering`], use this to retrieve the pixel data that was requested during
+    /// [`eframe::App::update`] via [`eframe::Frame::request_pixels`]. Currently only implemented with wgpu backend.
+    /// Returns None if
+    ///     Called in [`eframe::App::update`]
+    ///     [`eframe::Frame::request_pixels`] wasn't called on this frame during [`eframe::App::update`]
```
emilk, line 692: You should use `*` when making lists. You can check how the docs look with `cargo doc -p eframe --no-deps --open`

emilk (commented):
This is a very nice feature to have!

I suggest looking at `examples/screenshot/src/main.rs` for how to implement it for `glow`, and then you could use the new feature in that same example, to make it a lot simpler!

crates/eframe/src/epi.rs
amfaber: With this setup, that behaviour can still be achieved by simply retrieving the frame's data and manually saving it to the apps state in whatever way the user sees fit. I think that by clearing the data out after each frame there we achieve a small amount of clarity as to where a frame originated from, as it can only be retrieved from the post_rendering stage that created it. That being said I'm not against just letting stick around on the eframe::Frame if you think its more ergonomic.

crates/eframe/src/epi.rs
amfaber: Great points! I ended up refactoring to emit an egui::ColorImage from both backends to avoid ambiguity

amfaber:
From the example I figured out how to implement it for glow, and refactored example to use the new feature. I ended up returning an egui::ColorImage from the method as well instead of Vec<u8>, as that seems a lot more ergonomic to me. Because of that I also added some utility functions to egui::ColorImage.
ColorImage::region takes an egui::Rect out of an existing image, which covers my use-case. The rest of the new methods make it easier to access the image's data in raw form in the interest of interoperating with other crates such as image, or doing additional rendering on the gpu.

I know that ColorImage::into_raw is a little spicy, since it's using unsafe to return the data as a Vec<u8> without a copy.  In my mind it should be safe since the underlying data is a Vec<Color32> which is essentially just a Vec<[u8; 4]>, meaning we don't have any issues with endianness since the order of RGBA is always well-defined. Admittedly I have only tested it on my own setup.
To me, there is a case for avoiding the copy as a lot of data is potentially passed through this pipeline when recording.
Let me know if you think it should be included and if it should be annotated as unsafe if it is. For now I've left it as safe as I believe it to be so.

crates/epaint/src/image.rs
```diff
@@ -101,6 +101,65 @@ impl ColorImage {
...
+
+    /// Reinterpret the underlying data as `Vec<u8>` instead of `Vec<egui::Color32>`. Useful for interoperating with other crates such as [`image`](crates.io/crates/image). This
+    /// is a no-copy, but uses unsafe code internally. For a safer alternative, you can use [`Self::as_raw`] and manually copy into a new collection.
+    pub fn into_raw(mut self) -> Vec<u8> {
+        let ratio = std::mem::size_of::<Color32>() / std::mem::size_of::<u8>();
+        let length = self.pixels.len() * ratio;
+        let capacity = self.pixels.capacity() * ratio;
+        let ptr = self.pixels.as_mut_ptr() as *mut u8;
+        std::mem::forget(self.pixels);
+        unsafe { Vec::from_raw_parts(ptr, length, capacity) }
```
emilk, line 131: I believe this is unsound. From reading the `bytemuck` docs, you can only do this if the alignment of the source and target type is the same, which it isn't

I suggest removing this method. Users can to it themselves if they wish to.

amfaber: Done. I assume the cast_slice's are still fine, as Color32 is annotated with bytemuck::Pod?

amfaber:
Polite bump 😃

CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 * Add `Response.highlight` to highlight a widget ([#2632](https://github.com/emilk/egui/pull/2632)).
 * Add `Separator::grow` and `Separator::shrink` ([#2665](https://github.com/emilk/egui/pull/2665)).
 * Add `Slider::trailing_fill` for trailing color behind the circle like a `ProgressBar` ([#2660](https://github.com/emilk/egui/pull/2660)).
+* Add `Frame::request_screenshot` and `Frame::screenshot` to communicate to the backend that a screenshot of the current frame should be exposed by `Frame` during `App::post_rendering`
+* Add `read_screan_rgba` to the egui-wgpu `Painter`, to allow for capturing the current frame when using wgpu. Used in conjuction with `Frame::request_screenshot`
```
emilk, line 34: these should go into the `eframe` and `egui-wgpu` changelogs, and include al ink to this PR

Cargo.lock
```diff
@@ -4367,24 +4368,32 @@ checksum = "8c9864e83243fdec7fc9c5444389dcbbfd258f745e7853198f365e3c4968a608"
 
 [[package]]
 name = "windows_aarch64_msvc"
+=======
+>>>>>>> 83b5b81227b4496b5f25a458bfb9ea7aa891149a
```
emilk, line 4372: Broken `Cargo.lock` makes it difficult to test this PR

crates/eframe/src/epi.rs
```diff
@@ -651,6 +651,9 @@ pub struct Frame {
     /// Can be used to manage GPU resources for custom rendering with WGPU using [`egui::PaintCallback`]s.
     #[cfg(feature = "wgpu")]
     pub(crate) wgpu_render_state: Option<egui_wgpu::RenderState>,
+
+    #[cfg(not(target_arch = "wasm32"))]
+    pub(crate) screenshot: std::cell::Cell<Option<egui::ColorImage>>,
```
emilk, line 656: docstring

crates/eframe/src/epi.rs
```diff
@@ -668,10 +671,69 @@ impl Frame {
     }
 
     /// A place where you can store custom data in a way that persists when you restart the app.
+    #[cfg(not(target_arch = "wasm32"))]
```
emilk, line 674: We have storage on web (local storage)

crates/eframe/src/epi.rs
```diff
@@ -668,10 +671,69 @@ impl Frame {
...
+    ///             image::save_buffer(
+    ///                 "top_left.png",
+    ///                 top_left_corner.as_raw(),
+    ///                 top_left_corner.width() as u32,
+    ///                 top_left_corner.height() as u32,
+    ///                 image::ColorType::Rgba8,
+    ///             ).unwrap();
+    ///         }
+    ///     }
+    /// }
```
emilk, line 730:
```suggestion
    /// impl eframe::App for MyApp {
    ///     fn update(ctx: &egui::Context, frame: &mut eframe::Frame) {
    ///         // In real code the app would render something here
    ///         frame.request_screenshot()
    ///         // Things that are added to the frame after the call to
    ///         // request_screenshot() will still be included.
    ///     }
    ///
    ///     fn post_rendering(&mut self, _window_size: [u32; 2], frame: &eframe::Frame) {
    ///         if let Some(screenshot) = frame.screenshot() {
    ///             let pixels_per_point = frame.info().native_pixels_per_point;
    ///             let region = egui::Rect::from_two_pos(
    ///                 egui::Pos2::ZERO,
    ///                 egui::Pos2{ x: 100., y: 100. },
    ///             );
    ///             let top_left_corner = screenshot.region(&region, pixels_per_point);
    ///             image::save_buffer(
    ///                 "top_left.png",
    ///                 top_left_corner.as_raw(),
    ///                 top_left_corner.width() as u32,
    ///                 top_left_corner.height() as u32,
    ///                 image::ColorType::Rgba8,
    ///             ).unwrap();
    ///         }
    ///     }
    /// }
```

crates/eframe/src/epi.rs
```diff
@@ -668,10 +671,69 @@ impl Frame {
...
 
     /// A place where you can store custom data in a way that persists when you restart the app.
+    #[cfg(not(target_arch = "wasm32"))]
     pub fn storage(&self) -> Option<&dyn Storage> {
         self.storage.as_deref()
     }
 
+    /// Request the current frame's pixel data. Needs to be retrieved by calling [`Frame::screenshot`]
+    /// during [`App::post_rendering`].
+    #[cfg(not(target_arch = "wasm32"))]
```
emilk, line 681: What is the limitation for implementing this on web?

crates/eframe/src/native/run.rs
```diff
@@ -794,6 +794,14 @@ mod glow_integration {
                     &textures_delta,
                 );
 
+                let screenshot_requested = &mut integration.frame.output.screenshot_requested;
+
+                if *screenshot_requested == true {
```
emilk, line 799:
```suggestion
                if *screenshot_requested {
```

examples/screenshot/src/main.rs
```diff
@@ -40,6 +40,12 @@ impl eframe::App for MyApp {
                     "continuously take screenshots",
                 );
 
+                if self.continuously_take_screenshots {
+                    if ui.button("save to 'top_left.png'").clicked() {
```
emilk, line 44: It's a bit confusing that this button is only shown in continuous mode (and it is hard to find, because in continuous mode, the screen is… weird)

emilk (commented):
You accidentally checked in `top_left.png`, which is probably an indication that it should be added to an `.gitignore`.

@Wumpf could you pleas take a quick look at the wgpu parts of this? 🙏

examples/screenshot/src/main.rs
amfaber: I agree. Made it constantly visible + clicking the button will request a screenshot.

crates/eframe/src/epi.rs
amfaber: Mostly my inexperience with wasm and any kind of web-dev. I took a stab at trying to port the screenshot example to wasm with the eframe-template though, and it seems like App::post_rendering is never called on the web?
From the backend perspectiv, 3 things are required;

1. A flag needs to be set during App::update.
2. The flag needs to be read during the backend's paint call, such that the data can be pulled from the GPU
3. The data needs to be delivered back to the app during App::post_rendering

I chose to use the Frame struct for delivering the flag and image data between the app and the backend, which was easy on native since everything could be handled inside the WinitApp::paint function. I didn't look too closely at the architecture of the webpainter yet, but as App::post_rendering is never called, there would be no way to deliver the image data back to the app.

If this is solved, I believe it would be possible to more or less copy-paste the actual screen capturing logic for wgpu and glow.

Cargo.lock
amfaber: I had some merge issues 😅 It should be fixed now!

amfaber:
I've incorporated the latest set of feedback and took a look at what it would take to implement the feature for wasm. The main limitation right now is that App::post_rendering is never called outside native/run.rs

Also, sorry about the failing CI tests, this is my first time contributing to a larger project. This stuff is hard

amfaber:
I finally figured out how to run the CI for my own fork so that I could fix the failing tests.
However, due to https://github.com/ebarnard/rust-plist/commit/aa3821390fe804ad73bec28180a07c7ac31e0dc8 updating their dependency on base64 before bumping their version number, we are now using two versions of base64 - 0.13.1 and 0.21.0. For now I've added an ignore flag for base64 to deny.toml, basically ignoring the problem. Would love some thoughts on how to resolve an issue like this.

Wumpf (approved):
checked over the wgpu code and looks pretty straight forward! :)
Was hoping for something that is fast enough for streaming out videos, but that's for another time then (see (unactionable) comments)

emilk:
> I finally figured out how to run the CI for my own fork so that I could fix the failing tests. However, due to [ebarnard/rust-plist@aa38213](https://github.com/ebarnard/rust-plist/commit/aa3821390fe804ad73bec28180a07c7ac31e0dc8) updating their dependency on base64 before bumping their version number, we are now using two versions of base64 - 0.13.1 and 0.21.0. For now I've added an ignore flag for base64 to deny.toml, basically ignoring the problem. Would love some thoughts on how to resolve an issue like this.

Why update `Cargo.lock` at all in this PR?

```sh
git checkout main Cargo.lock deny.toml
cargo check
git add -u
git commit -m 'Revert changes to Cargo.lock and deny.toml`
```

amfaber:
> ```shell
> git checkout main Cargo.lock deny.toml
> cargo check
> git add -u
> git commit -m 'Revert changes to Cargo.lock and deny.toml`
> ```
Thanks! That was just what I needed :)

> Why update `Cargo.lock` at all in this PR?

It wasn't intentional. I feel like something broke when integrating a bunch of new commits from the master to my fork a few weeks ago and I ended up with some merge conflicts in the Cargo.toml somehow. At the time I think I tried to revert it, but it didn't work out for whatever reason. Then I naively thought "its autogenerated from the Cargo.toml anyways, so I could just delete it and let cargo handle it for me". I now see the error of those ways. Simply reverting just worked now, so I must have messed something up when I tried it initially

hacknus:
any updates on when this can be merged?

crates/eframe/src/epi.rs
```diff
@@ -691,10 +696,71 @@ impl Frame {
     }
 
     /// A place where you can store custom data in a way that persists when you restart the app.
+    #[cfg(not(target_arch = "wasm32"))]
```
emilk, line 699:
```suggestion
```

crates/egui-wgpu/CHANGELOG.md
```diff
@@ -11,7 +11,7 @@ All notable changes to the `egui-wgpu` integration will be noted in this file.
 * `winit::Painter::set_window` is now `async` ([#2434](https://github.com/emilk/egui/pull/2434)).
 * `egui-wgpu` now only depends on `epaint` instead of the entire `egui` ([#2438](https://github.com/emilk/egui/pull/2438)).
 * `winit::Painter` now supports transparent backbuffer ([#2684](https://github.com/emilk/egui/pull/2684)).
-
+* Add `read_screan_rgba` to the egui-wgpu `Painter`, to allow for capturing the current frame when using wgpu. Used in conjuction with `Frame::request_screenshot`. ([#2676](https://github.com/emilk/egui/pull/2676))
```
emilk, line 14: need to move up to the `# Unreleased` section

crates/egui-wgpu/src/winit.rs
```diff
@@ -274,22 +335,123 @@ impl Painter {
...
-    ) {
+        capture: bool,
+    ) -> Option<epaint::ColorImage> {
         crate::profile_function!();
 
         let render_state = match self.render_state.as_mut() {
             Some(rs) => rs,
-            None => return,
+            None => return None,
         };
```
emilk, line 451: `let render_state = self.render_state.as_mut()?;` here and below

amfaber:
The changelog changes have been moved to unreleased, and the ? is now employed, as it should it be

emilk (approved):
Thanks!

amfaber:
Thanks for the help with my first open source contribution, I appreciate it :)

Issue #2677, opened 2023-02-05
AccessKit breaking change drastically reduces memory usage


mwcampbell:
I just released a major breaking change in AccessKit that drastically reduces the memory usage of each AccessKit node, from more than 1 KB per node to as little as 32 bytes per node (depending on which properties are set). I'm about to open a PR to update egui.

PR #2678, opened 2023-02-05
Update for AccessKit refactor that drastically reduces memory usage

mwcampbell:
Closes <https://github.com/emilk/egui/issues/2677>.


emilk (approved):
Super!

Issue #2687, opened 2023-02-07
During keyboard input, only update a DragValue value when focus is lost.


harrisonmg:
I'm filing this issue to mention in an upcoming PR.

Currently, when focusing and editing a DragValue via keyboard input, the value updates for every character you type (I think it actually updates every single frame).

As per a discussion on the egui Discord server, this behavior is not desirable. It should only update the value when enter is pressed and / or focus is lost on the widget.

PR #2688, opened 2023-02-07
DragValue: when keyboard editing, only update the value on focus lost

harrisonmg:
Closes <https://github.com/emilk/egui/issues/2687>

Since the underlying `TextEdit` is `singleline`, pressing enter also results in lost focus.

I demonstrate two instances of this below. I press enter after the first edit, then click away after the second.

demo

I also shuffled some of the accesskit stuff. The up and down arrow keys probably shouldn't change the value during keyboard editing either. That little change allowed me to consolidate the accesskit code a bit better. I'm not sure how to test it.

As far as testing goes, `./sh/check.sh` failed in my environment (WSL 2) for some winit / platform reason. Clippy has no complaints about my changes.

### DIFF
--- a/crates/egui/src/widgets/drag_value.rs
+++ b/crates/egui/src/widgets/drag_value.rs
@@ -475,6 +475,8 @@ impl<'a> Widget for DragValue<'a> {
                     .desired_width(ui.spacing().interact_size.x)
                     .font(text_style),
             );
+            // Only update the value when the user presses enter, or clicks elsewhere. NOT every frame.
+            // See https://github.com/emilk/egui/issues/2687            
             if response.lost_focus() {
                 let parsed_value = match custom_parser {
                     Some(parser) => parser(&value_text),
--- a/crates/egui/src/widgets/drag_value.rs
+++ b/crates/egui/src/widgets/drag_value.rs
@@ -395,22 +395,21 @@ impl<'a> Widget for DragValue<'a> {
         let max_decimals = max_decimals.unwrap_or(auto_decimals + 2);
         let auto_decimals = auto_decimals.clamp(min_decimals, max_decimals);
 
-        let change = ui.input_mut(|input| {
-            let mut change = 0.0;
-
-            #[cfg(feature = "accesskit")]
-            if !is_kb_editing {
-                use accesskit::Action;
-                change += input.num_accesskit_action_requests(id, Action::Increment) as f64
-                    - input.num_accesskit_action_requests(id, Action::Decrement) as f64;
-            }
-
-            change
-        });
-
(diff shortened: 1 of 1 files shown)


emilk:
>  The up and down arrow keys probably shouldn't change the value during keyboard editing either.

Actually, it is very convenient to be able to move keyboard focus to DragValue using Tab, and then and press up/down to change its value!

emilk (changes requested):
I like the addition of `if response.lost_focus()` - it is really a much better behavior - but please revert the change where you can no longer change the value with up/down

crates/egui/src/widgets/drag_value.rs
```diff
@@ -475,13 +475,15 @@ impl<'a> Widget for DragValue<'a> {
...
                     .font(text_style),
             );
-            let parsed_value = match custom_parser {
-                Some(parser) => parser(&value_text),
-                None => value_text.parse().ok(),
-            };
-            if let Some(parsed_value) = parsed_value {
-                let parsed_value = clamp_to_range(parsed_value, clamp_range.clone());
-                set(&mut get_set_value, parsed_value);
+            if response.lost_focus() {
```
Mingun, line 478: It would be nice if you also added a comment justifying why this check should be done. The fact that it was absent before allows us to conclude that it is not self evident. Description of the reasons why this check was added and what it affects will make it easier to resolve problems if this change leads to them.

A good tone will also leave a link to the GitHub Issue, where you can get more information.

emilk, line 478:
```suggestion
            // Only update the value when the user presses enter, or clicks elsewhere. NOT every frame.
            // See https://github.com/emilk/egui/issues/2687            
            if response.lost_focus() {
```

crates/egui/src/widgets/drag_value.rs
```diff
@@ -475,13 +475,17 @@ impl<'a> Widget for DragValue<'a> {
...
             );
-            let parsed_value = match custom_parser {
-                Some(parser) => parser(&value_text),
-                None => value_text.parse().ok(),
-            };
-            if let Some(parsed_value) = parsed_value {
-                let parsed_value = clamp_to_range(parsed_value, clamp_range.clone());
-                set(&mut get_set_value, parsed_value);
+            // Only update the value when the user presses enter, or clicks elsewhere. NOT every frame.
+            // See https://github.com/emilk/egui/issues/2687
```
emilk, line 479:
```suggestion
            // Only update the value when the user presses enter, or clicks elsewhere. NOT every frame.
            // See https://github.com/emilk/egui/issues/2687
```

PR #2695, opened 2023-02-08
Update `glow` to 0.12

emilk:


coderedart:
will we get a new version of egui-glow on crates.io soon?
three-d's [glow upgrade](https://github.com/asny/three-d/issues/321) is blocked on egui-glow.

PR #2701, opened 2023-02-09
Re-enable egui-winit for egui-glow on web

Talon1024:
I tried this myself, and I do not think there is any reason to block the winit feature from egui-glow on wasm platforms.


t18b219k:
winit don't support IME on wasm32 platform so we have to use own eframe's implementation.
If we disable IME we can't type Chinese Japanese Korean languages.

emilk:
Closing: this landed in #4421, which enabled the `winit` feature of `egui_glow` on wasm. Thanks for the PR!

PR #2713, opened 2023-02-11
Size `Slider` based off `Spacing` instead of `TextStyle::Body`

hinto-janai:
Most widgets are statically sized based off a value in `Spacing`, but `Slider` is not, instead its size is determined by [the height of `TextStyle::Body`](https://github.com/emilk/egui/blob/master/crates/egui/src/widgets/slider.rs#L784-L786). This leads to awkward scaling.

This PR adds `slider_thickness` to `Spacing` that all sliders will use instead.

The default size is `12.5`, which is the same default size as right now (since `TextStyle::Body` is also `12.5`).

It also adds overrides per slider:
```rust
Slider::new(&mut s, 0.0..1.0).thickness(my_f32);
```
This may be a breaking change as some users may be relying on `Slider` scailing with their `TextStyle::Body` text size.

Before:
before

After:
after

Also: Sliders that are even slightly bigger than the default have a rectangle that peeks out of the circle so the default circle-to-rect ratio was changed slightly (circle is a bit bigger):

Before:
before2
After:
after2

Very large sliders will still have peeking rectangles so an override was added to this as well:
```rust
Slider::new(&mut s, 0.0..1.0).circle_size(my_f32);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 
 ## Unreleased
 
+### Added ⭐
+* Add `Spacing::slider_thickness` to globally set thickness (not width/length!) of `Slider`.
+* Add `Slider::thickness` to set thickness per `Slider`.
+* Add `Slider::circle_size` to set the circle size of a `Slider`.
 
 ## 0.21.0 - 2023-02-08 - Deadlock fix and style customizability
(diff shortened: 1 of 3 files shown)


emilk (commented):
I find it a big confusing if `thickness` refers to the visual thickness of the line, or the height of the entire slider.

And considering how it behaves, I think the code is a bit confused too:

slider-thickness

emilk:
I am curious about how one now adjust the thickness of the "stem" of the slider

emilk:
Closed by https://github.com/emilk/egui/pull/4092

Issue #2700, opened 2023-02-08
Custom fonts have vertical offset with unicode symbols


woelper:
Egui default font with `0.21`:

![image](https://user-images.githubusercontent.com/3613791/217669500-baa5e755-65ab-4836-bd0b-6679905a45cf.png)

Noto sans with `0.19`:
![image](https://user-images.githubusercontent.com/3613791/217669720-bb3e6962-f501-4662-94fa-3a20402bc3b8.png)

Noto sans with `0.21`:
![image](https://user-images.githubusercontent.com/3613791/217669933-c791da0d-a4a9-4e6a-a287-e8dbad22eee5.png)

Although a bit hard to see, all symbols are offset in the y direction. I tried different fonts such as IBM plex, and they show this as well.

**Describe the bug**
Symbols are not vertically aligned with text if using a custom font.

**To Reproduce**
```rust
pub fn new(cc: &eframe::CreationContext<'_>) -> Self {
    let mut fonts = FontDefinitions::default();
    fonts.font_data.insert(
        "my_font".to_owned(),
        FontData::from_static(include_bytes!("../res/fonts/NotoSans-Regular.ttf")),
        // FontData::from_static(include_bytes!("../res/fonts/IBMPlexSans-Regular.ttf")),
    );
    fonts
        .families
        .get_mut(&FontFamily::Proportional)
        .unwrap()
        .insert(0, "my_font".to_owned());
    cc.egui_ctx.set_fonts(fonts);
    Default::default()
}
```

woelper:
Actually after reading [your post](https://twitter.com/ernerfeldt/status/1621216119955619841) and trying `Inter` (which worked and is vertically aligned with symbols) I am not so sure any more if I just picked the wrong fonts. Noto and Plex are pretty common though and I wonder what they do differently.

emilk:
There may be a "vertical offset" font attribute that egui ignores, and shouldn't. In any case, you can tweak the y-offset when you load your fonts using https://docs.rs/egui/latest/egui/struct.FontData.html#method.tweak

aevyrie:
I've run into this using `Roboto`, and had to add a manual y offset in the font tweaks. Issue found when upgrading to `v0.20`.

PR #2724, opened 2023-02-12
improve fallback fonts alignment

xyning:
try to proper align glyph from different fallback fonts on the baseline without some manually tricky tweaks

i *guess* in most cases (where no font fallback happens) it should look same as before
`FontTweak::scale` now should only affect glyphs themselves, not full text layouts (row heights, alignments etc). imo for the latter case is better to change the actual font size instead of using this hacky scale...

----
test using `Cantarell` as primary font and `Noto Sans CJK` as cjk fallback
***all font tweak values are default***
| before | after |
|----|----|
|image|image|

----
note the row height is now calculated using the font metrics from the tallest glyph
example: the second row is *slightly* taller than the first because of the noto cjk glyphs
| before | after |
|----|----|
|image|image|

----
might also resolve https://github.com/emilk/egui/issues/2700
| before | after |
|----|----
|image|image|

----

<details>
<summary>why theres still some tweaks for the <code>emoji-icon-font</code></summary>
  
  
| emoji-icon-font | notoemoji-regular |
|----|----|
|image|image|

note that <code>emoji-icon-font</code> has a significantly lower baseline than other common fonts. 
this will make the *correct* rendering look a bit off center: 
|image|
|:--:|
| image taken from libreoffice |

...which is not so good

</details>


crates/epaint/src/text/font.rs
```diff
@@ -42,6 +43,10 @@ pub struct GlyphInfo {
     /// Unit: points.
     pub advance_width: f32,
 
+    pub ascent: f32,
+
+    pub row_height: f32,
```
emilk, line 48: These needs docstrings

emilk (commented):
This looks extremely promising - thanks for working on this!

emilk:
Thank you!

Issue #2725, opened 2023-02-12
Make egui-wgpu Painter independent from winit


senneh:
The egui-wgpu crate provides a Painter struct. Right now this can only be used with winit, but it would be useful for integrations with other windowing libraries (like glazier). 
There is no reason for it to depend on winit when it only needs a raw window handle and the physical size of the client area.

In short, change:
```rust
    pub async unsafe fn set_window(
        &mut self,
        window: Option<&winit::window::Window>,
    ) -> Result<(), crate::WgpuError>
```
to:
```rust
    pub async unsafe fn set_window<W>(
        &mut self,
        window: Option<&W>,
        width_in_pixels: u32,
        height_in_pixels: u32,
    ) -> Result<(), crate::WgpuError>
    where
        W: raw_window_handle::HasRawWindowHandle + raw_window_handle::HasRawDisplayHandle,
```

PR #2726, opened 2023-02-12
Make egui-wgpu Painter independent from winit

senneh:
Implements <https://github.com/emilk/egui/issues/2725>.

The painter module is an exact copy of the winit module, but with the `set_window` method changed. It now depends on raw-window-handle.
The winit module is still there but marked as deprecated. This way it doesn't require a semver bump.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1271,6 +1271,7 @@ dependencies = [
  "document-features",
  "epaint",
  "puffin",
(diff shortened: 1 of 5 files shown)


emilk:
Instead of making an exact copy, can you just add a feature-flag that changes how `set_window` works?

Also: we don't care about breaking changes here. We break compatibility on each new release :)

Issue #2608, opened 2023-01-19
Option to hide scrollbar


senchpimy:
### Option to hide scrollbar

something like the function **[always_show_scroll](https://docs.rs/egui/latest/egui/containers/scroll_area/struct.ScrollArea.html#method.always_show_scroll)** but hiding the scroll all the time, just enable scrolling

IVAN-MK7:
I'll be working on it

PR #2729, opened 2023-02-12
Scroll bar visibility options

IVAN-MK7:
Enables the user to set whether the horizontal and vertical scroll bars must be always visible, hidden or visible when needed, in the context of a ScrollArea.

`egui::ScrollArea::vertical().scroll_bar_visibility(ScrollBarVisibility::AlwaysVisible).show(ui, |ui| {});`

To consider adding a method for each option in order to shorter and simplify the code, e.g.:
`egui::ScrollArea::vertical().scroll_bar_always_visible().show(ui, |ui| {});`

Closes <https://github.com/emilk/egui/issues/2608>.

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -74,6 +74,14 @@ pub struct ScrollAreaOutput<R> {
     pub inner_rect: Rect,
 }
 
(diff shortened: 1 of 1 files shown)


emilk (approved):
Very nice!

Issue #2716, opened 2023-02-11
RGBA values input fields for color_picker


IVAN-MK7:
The color_picker widget does not currently allow the user to manually set the colors' RGBA values.

I've added some DragValue widgets to my egui fork to achieve just that (RGBA values with 0-255 range):

![image](https://user-images.githubusercontent.com/68190772/218269349-bb44a345-9c80-4290-9a5b-050d4c036602.png)

Would be great to have this feature added to the master branch.

PR #2734, opened 2023-02-13
Add `DragValue`s for RGB(A) in the color picker

IVAN-MK7:
Added some DragValue widgets in the color_picker widget as input fields for managing the RGBA values.
In case the provided values result in a not valid premultiplied alpha RGBA color, a button will appear next to the input fields, to be used to multiply the values with the alpha channel.

image

Closes <https://github.com/emilk/egui/issues/2716>.


crates/egui/src/widgets/color_picker.rs
```diff
@@ -311,6 +314,51 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsva: &mut HsvaGamma, alpha: Alpha) {
...
+            rgb_changed = true;
+        }
+        if ui.add(DragValue::new(&mut g).speed(0.5)).changed() {
+            rgb_changed = true;
+        }
+        if ui.add(DragValue::new(&mut b).speed(0.5)).changed() {
+            rgb_changed = true;
+        }
+        if ui.add(DragValue::new(&mut a).speed(0.5)).changed() {
+            a_changed = true;
```
emilk, line 336:
```suggestion
        if ui.add(DragValue::new(&mut r).speed(0.5).prefix("R: ")).changed() {
            rgb_changed = true;
        }
        if ui.add(DragValue::new(&mut g).speed(0.5).prefix("G: ")).changed() {
            rgb_changed = true;
        }
        if ui.add(DragValue::new(&mut b).speed(0.5).prefix("B: ")).changed() {
            rgb_changed = true;
        }
        if ui.add(DragValue::new(&mut a).speed(0.5).prefix("A: ")).changed() {
            a_changed = true;
```

crates/egui/src/widgets/color_picker.rs
```diff
@@ -311,6 +314,51 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsva: &mut HsvaGamma, alpha: Alpha) {
     color_slider_2d(ui, v, s, |v, s| HsvaGamma { s, v, ..opaque }.into());
 }
 
+fn srgba_edit_ui(ui: &mut Ui, hsvag: HsvaGamma) -> Option<HsvaGamma> {
+    let [mut r, mut g, mut b, mut a] = Color32::from(hsvag).to_array();
```
emilk, line 318:
```suggestion
fn srgba_edit_ui(ui: &mut Ui, rgba: Color32) -> Option<Color32> {
    let [mut r, mut g, mut b, mut a] = rgba.to_array();
```

crates/egui/src/widgets/color_picker.rs
```diff
@@ -311,6 +314,51 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsva: &mut HsvaGamma, alpha: Alpha) {
     color_slider_2d(ui, v, s, |v, s| HsvaGamma { s, v, ..opaque }.into());
 }
 
+fn srgba_edit_ui(ui: &mut Ui, hsvag: HsvaGamma) -> Option<HsvaGamma> {
```
emilk, line 317: This function should take `alpha: Alpha` as argument and not show an alpha-slider for `Alpha::Opaque`, for instance.

emilk (changes requested):
It is quite confusing how this works with pre-multiplied alpha.

I think we should keep premultiplied alpha out of the UI.

Instead, unmultiply the alpha before showing the buttons, and then special-case alpha=0 (additive).

Otherwise it is very easy to end up with things like this:

image

IVAN-MK7:
I've applied the suggestions, had to work around the case when Additive blending is signaled with a negative alpha on Alpha::BlendOrAdditive.
Now the DragValues take unmultiplied RGBA values and the Alpha DragValue is hidden in case the provided Alpha is Alpha::Opaque (Alpha::Opaque is automatically sent to the fn in case of a negative alpha on Alpha::BlendOrAdditive).

color_picker_srgba_unmultiplied_alpha_input_values

Given the negative alpha thing, users still can't switch to "Additive" blending when alpha is 0, even after this PR change.

crates/egui/src/widgets/color_picker.rs
IVAN-MK7: Change applied.

crates/egui/src/widgets/color_picker.rs
IVAN-MK7: Added alpha: Alpha parameter, alpha-slider won't be shown when Alpha::Opaque

crates/egui/src/widgets/color_picker.rs
```diff
@@ -311,6 +340,49 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsva: &mut HsvaGamma, alpha: Alpha) {
     color_slider_2d(ui, v, s, |v, s| HsvaGamma { s, v, ..opaque }.into());
 }
 
+/// Shows 4 `DragValue` widgets to be used to edit the Color32's RGBA values.
+/// Alpha's `DragValue` is hidden when `Alpha::Opaque`.
+///
+/// Returns `Some(Color32)` on change.
+fn srgba_edit_ui(ui: &mut Ui, rgba: Color32, alpha: Alpha) -> Option<Color32> {
```
emilk, line 347: Using premultiplied alpha in a UI is just confusing, especially since the rest of the color picker separates out the alpha.

IVAN-MK7: Are you referring to the DragValues? The values they display and let the user change are not premultiplied with the alpha channel, the premultiplied alpha values are only shown in the color_text_ui below the DragValues.

emilk: The `DragValues` below takes the alpha from a `Color32`, which per definition is always pre-multiplied.

…but I just now realize the code on `master` is already displaying pre-multiplied values, which is my bad. We should never use pre-multiplied alpha in the UI imho.

crates/egui/src/widgets/color_picker.rs
```diff
@@ -244,15 +244,43 @@ fn color_text_ui(ui: &mut Ui, color: impl Into<Color32>, alpha: Alpha) {
...
-fn color_picker_hsvag_2d(ui: &mut Ui, hsva: &mut HsvaGamma, alpha: Alpha) {
+fn color_picker_hsvag_2d(ui: &mut Ui, hsvag: &mut HsvaGamma, alpha: Alpha) {
+    let srgba_unmultiplied = (std::convert::Into::<Hsva>::into(*hsvag)).to_srgba_unmultiplied();
+    // Send an Opaque Alpha also when Alpha is BlendOrAdditive with negative alpha value (signals Additive blending),
+    // so to hide the alpha's DragValue in both cases.
+    let alpha_control = if alpha == Alpha::Opaque || hsvag.a < 0.0 {
+        Alpha::Opaque
+    } else {
+        alpha
+    };
```
tosti007, line 255: This would be equivalent right?
```suggestion
    let alpha_control = if hsvag.a < 0.0 {
        Alpha::Opaque
    } else {
        alpha
    };
```

crates/egui/src/widgets/color_picker.rs
```diff
@@ -311,6 +339,49 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsva: &mut HsvaGamma, alpha: Alpha) {
     color_slider_2d(ui, v, s, |v, s| HsvaGamma { s, v, ..opaque }.into());
 }
 
+/// Shows 4 `DragValue` widgets to be used to edit the RGBA values.
+/// Alpha's `DragValue` is hidden when `Alpha::Opaque`.
+///
+/// Returns `Some([u8;4])` on change, with unmultiplied alpha values.
+fn srgba_edit_ui(ui: &mut Ui, rgba: [u8; 4], alpha: Alpha) -> Option<[u8; 4]> {
```
tosti007, line 346: Nit: Would it be possible for this to take a `&mut [u8; 4]` and return a `bool` instead?

tosti007 (commented):
Awesome, I'd really like this addition! To maybe help this further I gave it a review.

crates/egui/src/widgets/color_picker.rs
IVAN-MK7: This suggestion makes the code slimmer and reduces the amount of variables to allocate. Committing.

crates/egui/src/widgets/color_picker.rs
```diff
@@ -311,6 +335,32 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsva: &mut HsvaGamma, alpha: Alpha) {
...
+///
+/// Returns `true` on change.
+fn srgba_edit_ui(ui: &mut Ui, rgba: &mut [u8; 4], alpha: Alpha) -> bool {
+    let [r, g, b, a] = rgba;
+
+    let mut edited = false;
+    ui.horizontal(|ui| {
+        if ui.add(DragValue::new(r).speed(0.5).prefix("R: ")).changed() {
+            edited = true;
+        }
```
tosti007, line 349: Just realized that a `|` is also possible here. Also maybe using the `Widget`'s `ui` function might result in a nice `rustfmt` (since it's all function chains then)? So something like (but still needs a `fmt`): 

```suggestion
        edited |= DragValue::new(r).speed(0.5).prefix("R: ").ui(ui).changed();
```

tosti007 (commented):
Added another small nit i just realised. Also: Maybe it'd be good to support `f32` drag values, since that would allow for HDR color pickers (which is my use case)?

IVAN-MK7:
> Added another small nit i just realised. Also: Maybe it'd be good to support `f32` drag values, since that would allow for HDR color pickers (which is my use case)?



>

crates/egui/src/widgets/color_picker.rs
IVAN-MK7: committed

imsi32:
Hello, while I was looking demos, I faced hard time to get the correct colors, so I find this pull request. Because I don't know egui yet, I couldn't help with my suggestion.

I think buttons on the top doesn't look good. Instead of putting them there replacing with rgba(1,1,1,1) text will be better idea like this image:
egui-color-picker-1

Secondly, even though I don't understand the additive page. When I open the additive pages, alpha area disables and all page change place. So putting alpha area in end of the page will make page less shaking. Unfortunately,  color changing area needs to put upper of the alpha area to eliminate color mixing between color choosing area and alpha area.
egui-color-picker-2

IVAN-MK7:
I've added the
```rust
pub color_picker_input_values_type: ColorPickerInputType,
```
field to style's Visuals.
Also rearranged color_picker's color sliders, and made color_text_ui display unmultiplied rgba values, this is the final result:

colorpicker_u8_f32_rearrange_2

The resulting appearance definitely needs some polish, it would be better to put buttons and drag values into a StripBuilder, but it must be imported from the egui_extras crate. Help and advices are welcome.

crates/egui/src/widgets/color_picker.rs
```diff
@@ -227,38 +227,146 @@ pub enum Alpha {
...
 }
 
-fn color_text_ui(ui: &mut Ui, color: impl Into<Color32>, alpha: Alpha) {
-    let color = color.into();
+fn color_text_ui(ui: &mut Ui, hsvag: &HsvaGamma, alpha: Alpha) {
+    use crate::style::ColorPickerInputType;
     ui.horizontal(|ui| {
-        let [r, g, b, a] = color.to_array();
+        let [r, g, b, a] = match ui.style().visuals.color_picker_input_values_type {
+            ColorPickerInputType::U8 => std::convert::Into::<Hsva>::into(*hsvag)
```
emilk, line 234:
```suggestion
            ColorPickerInputType::U8 => Hsva::from(*hsvag)
```

emilk: here and eerywhere

crates/egui/src/widgets/color_picker.rs
```diff
@@ -227,38 +227,146 @@ pub enum Alpha {
...
+                .to_srgba_unmultiplied()
+                .map(|c| c.to_string()),
+            ColorPickerInputType::F32 => {
+                let mut formatted_colors = (std::convert::Into::<Hsva>::into(*hsvag))
+                    .to_rgba_unmultiplied()
+                    .map(|c| {
+                        emath::round_to_decimals(c.into(), 3)
+                            .to_string()
+                            .char_range(0..5)
```
emilk, line 243: why the `char_range`?

crates/egui/src/widgets/color_picker.rs
```diff
@@ -227,38 +227,146 @@ pub enum Alpha {
...
+                            .to_owned()
+                    });
+                for color in &mut formatted_colors {
+                    if color.contains('.') {
+                        *color = format!("{color:0<5}");
+                    } else if *color == "1" || *color == "0" {
+                        *color = format!("{color}.000");
+                    }
```
emilk, line 251: this looks like a very complicated way of doing `format!("{c:.03}`) ?

crates/egui/src/widgets/color_picker.rs
```diff
@@ -227,38 +227,146 @@ pub enum Alpha {
...
+                let mut visuals = ui.style().visuals.clone();
+                match ui.style().visuals.color_picker_input_values_type {
+                    ColorPickerInputType::U8 => {
+                        visuals.color_picker_input_values_type = ColorPickerInputType::F32;
+                    }
+                    ColorPickerInputType::F32 => {
+                        visuals.color_picker_input_values_type = ColorPickerInputType::U8;
+                    }
+                }
+                ui.ctx().set_visuals(visuals);
```
emilk, line 276: this reads the visuals from the current ui, but sets it globally for the `ctx`.

I think it makes more sense to just do `ctx.style_mut(|s| s.visuals.color_picker_input_values_type = …)` here

crates/egui/src/widgets/color_picker.rs
```diff
@@ -227,38 +227,146 @@ pub enum Alpha {
...
             }
-        }
+            if ui.button("📋").on_hover_text("Click to copy").clicked() {
+                if alpha == Alpha::Opaque {
+                    ui.ctx().copy_text(format!("{r}, {g}, {b}"));
+                } else {
+                    ui.ctx().copy_text(format!("{r}, {g}, {b}, {a}"));
+                }
+            }
```
emilk, line 284: I think we should move this copy-button before the new RGB-buttons. Right now we show the RGB values twice - first as buttons, and then as labels. That is redundant and redundant.

emilk (changes requested):

crates/egui/src/widgets/color_picker.rs
IVAN-MK7: Applied

crates/egui/src/widgets/color_picker.rs
IVAN-MK7: You're right, thanks for the suggestion.

crates/egui/src/widgets/color_picker.rs
IVAN-MK7: Must have forgot to remove it.

crates/egui/src/widgets/color_picker.rs
IVAN-MK7: Applied.

crates/egui/src/widgets/color_picker.rs
IVAN-MK7: Rearranged the buttons layout, removed the label.

IVAN-MK7:
This is the resulting UI following the recent changes:

u8f32rearranged



Needs a finishing touch for the layout, I tried to make so that the change of RGBA values and input type (U8/F32) made the least amount of popping, but could't go very far without a StripBuilder.

Ved-s:
Would be nice seeing this implemented in. Also I'd suggest adding hex color field so it's easier to copy/paste

emilk:
I have trouble pushing commits to this PR, for whatever reason. Or rather, I push my commits successfully to `git@github.com:IVAN-MK7/egui` but they don't show up in the PR :/

There is a lot of `a < 0.0` everywhere, so I tried cleaning that up with this patch:

```
diff --git a/crates/egui/src/widgets/color_picker.rs b/crates/egui/src/widgets/color_picker.rs
index 138b9724..a293b813 100644
--- a/crates/egui/src/widgets/color_picker.rs
+++ b/crates/egui/src/widgets/color_picker.rs
@@ -216,6 +216,13 @@ fn color_slider_2d(
     response
 }
 
+/// We use a negative alpha for additive colors within this file (a bit ironic).
+///
+/// We use alpha=0 to mean "transparent".
+fn is_additive_alpha(a: f32) -> bool {
+    a < 0.0
+}
+
 /// What options to show for alpha
 #[derive(Clone, Copy, PartialEq, Eq)]
 pub enum Alpha {
@@ -230,9 +237,11 @@ pub enum Alpha {
 fn color_picker_hsvag_2d(ui: &mut Ui, hsvag: &mut HsvaGamma, alpha: Alpha) {
     use crate::style::ColorPickerInputType;
 
-    // Send an Opaque Alpha also when Alpha is BlendOrAdditive with negative alpha value (signals Additive blending),
-    // so to hide the alpha's DragValue in both cases.
-    let alpha_control = if hsvag.a < 0.0 { Alpha::Opaque } else { alpha };
+    let alpha_control = if is_additive_alpha(hsvag.a) {
+        Alpha::Opaque // no alpha control for additive colors
+    } else {
+        alpha
+    };
 
     match ui.style().visuals.color_picker_input_values_type {
         ColorPickerInputType::U8 => {
@@ -240,8 +249,7 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsvag: &mut HsvaGamma, alpha: Alpha) {
             // Update hsvag only if the converted srgba is changed, this is because hsvag is made of f32,
             // and the conversion between u8 and f32 loses a bit of the color precision, causing little flickering on hsvag based ui widgets.
             if srgba_edit_ui(ui, &mut srgba_unmultiplied, alpha_control) {
-                // Additive blending, signaled by the negative Alpha.
-                if hsvag.a < 0.0 {
+                if is_additive_alpha(hsvag.a) {
                     let stored_a = hsvag.a;
                     // Alpha to 0 instead of negative, so it won't pop back to Normal blending when RGB are modified.
                     srgba_unmultiplied[3] = 0;
@@ -250,9 +258,8 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsvag: &mut HsvaGamma, alpha: Alpha) {
                     // stored_a keeps the Alpha value that was set during Normal blending so that in case we alter RGB in Additive blending (negative Alpha)
                     // and then switch back to Normal blending it gets that Alpha value back.
                     hsvag.a = stored_a;
-                }
-                // Normal blending.
-                else {
+                } else {
+                    // Normal blending.
                     *hsvag = HsvaGamma::from(Hsva::from_srgba_unmultiplied(srgba_unmultiplied));
                 }
             }
@@ -262,8 +269,7 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsvag: &mut HsvaGamma, alpha: Alpha) {
             // Update hsvag only if the converted srgba is changed, this is because hsvag is made of f32,
             // and the conversion between u8 and f32 loses a bit of the color precision, causing little flickering on hsvag based ui widgets.
             if rgba_edit_ui(ui, &mut rgba_unmultiplied, alpha_control) {
-                // Additive blending, signaled by the negative Alpha.
-                if hsvag.a < 0.0 {
+                if is_additive_alpha(hsvag.a) {
                     let stored_a = hsvag.a;
                     // Alpha to 0 instead of negative, so it won't pop back to Normal blending when RGB are modified.
                     rgba_unmultiplied[3] = 0.0;
@@ -276,9 +282,8 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsvag: &mut HsvaGamma, alpha: Alpha) {
                     // stored_a keeps the Alpha value that was set during Normal blending so that in case we alter RGB in Additive blending (negative Alpha)
                     // and then switch back to Normal blending it gets that Alpha value back.
                     hsvag.a = stored_a;
-                }
-                // Normal blending.
-                else {
+                } else {
+                    // Normal blending.
                     *hsvag = HsvaGamma::from(Hsva::from_rgba_unmultiplied(
                         rgba_unmultiplied[0],
                         rgba_unmultiplied[1],
@@ -294,9 +299,8 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsvag: &mut HsvaGamma, alpha: Alpha) {
     show_color(ui, *hsvag, current_color_size).on_hover_text("Selected color");
 
     if alpha == Alpha::BlendOrAdditive {
-        // We signal additive blending by storing a negative alpha (a bit ironic).
         let a = &mut hsvag.a;
-        let mut additive = *a < 0.0;
+        let mut additive = is_additive_alpha(*a);
         ui.horizontal(|ui| {
             if ui
                 .button(
@@ -355,7 +359,7 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsvag: &mut HsvaGamma, alpha: Alpha) {
     })
     .on_hover_text("Hue");
 
-    let additive = hsvag.a < 0.0;
+    let additive = is_additive_alpha(hsvag.a);
 
     if alpha == Alpha::Opaque {
         hsvag.a = 1.0;
@@ -363,7 +367,7 @@ fn color_picker_hsvag_2d(ui: &mut Ui, hsvag: &mut HsvaGamma, alpha: Alpha) {
         let a = &mut hsvag.a;
 
         if alpha == Alpha::OnlyBlend {
-            if *a < 0.0 {
+            if is_additive_alpha(*a) {
                 *a = 0.5; // was additive, but isn't allowed to be
             }
             color_slider_1d(ui, a, |a| HsvaGamma { a, ..opaque }.into()).on_hover_text("Alpha");
```

emilk:
I think the `U8/F32` switch belong on the same row as the drag values

emilk (approved):

Issue #2739, opened 2023-02-14
Demo keyboard input "p" not working


jiricodes:
# Short Description
In the deployed [egui demo](https://www.egui.rs/) the character `p` is being ignored in some TextEdit fields.

# Expected behaviour
Press `P` on keyboard and a `p` character appears in selected editable field.
Press `Shift + P` on keyboard and a `P` character appears in selected editable field.

# Observed behaviour
The keyboard input `P` is ignored in following fields:

- Demos
    - [Code Editor](https://github.com/emilk/egui/blob/38849fe3816d24538e200c4dc62854e38cb47a09/crates/egui_demo_lib/src/demo/code_editor.rs#L87)
    - [Code Example](https://github.com/emilk/egui/blob/38849fe3816d24538e200c4dc62854e38cb47a09/crates/egui_demo_lib/src/demo/code_example.rs#L34)
    - [Font Book](https://github.com/emilk/egui/blob/38849fe3816d24538e200c4dc62854e38cb47a09/crates/egui_demo_lib/src/demo/font_book.rs#L62)
    -  Misc Demos - [password](https://github.com/emilk/egui/blob/38849fe3816d24538e200c4dc62854e38cb47a09/crates/egui_demo_lib/src/demo/password.rs#L41)
        - [ Desktop ] works as expected until visibility is toggled on or off
            - gets fixed by selecting all the text and deleting it
            - one by one character deletion doesn't fix
    - [TextEdit](https://github.com/emilk/egui/blob/38849fe3816d24538e200c4dc62854e38cb47a09/crates/egui_demo_lib/src/demo/text_edit.rs#L44)
        - [ Desktop ] - buttons [*start*](https://github.com/emilk/egui/blob/38849fe3816d24538e200c4dc62854e38cb47a09/crates/egui_demo_lib/src/demo/text_edit.rs#L87) and [*end*](https://github.com/emilk/egui/blob/38849fe3816d24538e200c4dc62854e38cb47a09/crates/egui_demo_lib/src/demo/text_edit.rs#L97) seems to fix things, making the 'P' key being registered as expected. Until the cursor is moved manually. (maybe a hint towards the root cause?)
    - [Window Options](https://github.com/emilk/egui/blob/38849fe3816d24538e200c4dc62854e38cb47a09/crates/egui_demo_lib/src/demo/window_options.rs#L92)
[...]

PR #2740, opened 2023-02-15
eframe: Fix inputting of the letter P on web

emilk:
Closes https://github.com/emilk/egui/issues/2739


emilk:
eframe 0.21.3 released with fix

PR #2742, opened 2023-02-15
Fix UI `data()` read mutability

### LINKED ISSUE #2741: The `ui.data()` requires mutable borrow to `get_temp()`?
**Describe the bug**

It seems that inside the `ui.data()` lambda I can't `get_temp()` anything from the `IdTypeMap` because the reference to it passed to the lambda  apparently needs to be mutable. Why?

**To Reproduce**
1. Create an `eframe` project and plop down some widget
2. Inside the lambda that accepts `&mut egui::Ui` do
```rust
let x_id = ui.id().with("x");
let x = ui.data(|d| d.get_temp::<u32>(x_id).unwrap_or_default());
```
3.
```
error[E0596]: cannot borrow `*d` as mutable, as it is behind a `&` reference
  --> [DATA EXPUNGED].rs:41:33
   |
41 |             let x = ui.data(|d| d.get_temp::<u32>(x_id).unwrap_or_default());
   |                              -  ^^^^^^^^^^^^^^^^^^^^^^^ `d` is a `&` reference, so the data it refers to cannot be borrowed as mutable
   |                              |
   |                              consider changing this binding's type to be: `&mut IdTypeMap`

```

**Expected behavior**

Since `get_temp()` clones the value, the `IdTypeMap` is not required to be mutable. Applies to all "read" methods.

**Screenshots**

 - OS: `Linux 5.18.0-14parrot1-amd64 #1 SMP PREEMPT_DYNAMIC Debian 5.18.14-1parrot1 (2022-08-07) x86_64 GNU/Linux`
 - Browser 
 - Version `master` (7215fdfb7c7407b8085d53052582dac10124bdfc)

**Additional context**

Am I missing something? This seems very weird.

Possibly related:
- #2625

IS2511:
Closes #2741

### DIFF
--- a/crates/egui/src/util/id_type_map.rs
+++ b/crates/egui/src/util/id_type_map.rs
@@ -366,10 +366,7 @@ impl IdTypeMap {
     #[inline]
     pub fn get_temp<T: 'static + Clone>(&self, id: Id) -> Option<T> {
         let hash = hash(TypeId::of::<T>(), id);
-        self.0
-            .get(&hash)
-            .and_then(|x| x.get_temp())
-            .cloned()
+        self.0.get(&hash).and_then(|x| x.get_temp()).cloned()
     }
 
     /// Read a value, optionally deserializing it if available.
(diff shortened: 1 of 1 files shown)


emilk (approved):

emilk:
needs `cargo fmt`

IS2511:
Should be all fixed now.
My bad, didn't notice the suggested change. Kinda gave up trying to run `sh/check.sh` with it complaining about me missing `gdk-3.0`.

PR #2750, opened 2023-02-18
eframe: Automatically change theme when system dark/light mode changes

bash:
I noticed that `eframe` does not change the theme when the system dark/light mode is changed while the app is running.

For `winit` reacting to the change is fairly easy: There's a [`ThemeChanged`] event. It's not supported everywhere, but I guess it's better than nothing :).

For web, there's a [`change`](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/change_event) event on `MediaQueryList`.

https://user-images.githubusercontent.com/4602612/219874514-6bdedbad-14b8-4069-bd7d-fdf03a8c2a67.mp4

[`ThemeChanged`]: https://docs.rs/winit/latest/winit/event/enum.WindowEvent.html#variant.ThemeChanged
[`change`]: https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/change_event

### DIFF
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -308,6 +308,7 @@ pub struct EpiIntegration {
     close: bool,
     can_drag_window: bool,
     window_state: WindowState,
+    follow_system_theme: bool,
 }
 
 impl EpiIntegration {
@@ -316,6 +317,7 @@ impl EpiIntegration {
         max_texture_side: usize,
         window: &winit::window::Window,
         system_theme: Option<Theme>,
+        follow_system_theme: bool,
         storage: Option<Box<dyn epi::Storage>>,
         #[cfg(feature = "glow")] gl: Option<std::sync::Arc<glow::Context>>,
         #[cfg(feature = "wgpu")] wgpu_render_state: Option<egui_wgpu::RenderState>,
@@ -363,6 +365,7 @@ impl EpiIntegration {
             close: false,
             can_drag_window: false,
             window_state,
+            follow_system_theme,
         }
     }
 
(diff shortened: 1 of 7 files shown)


crates/eframe/src/native/epi_integration.rs
```diff
@@ -426,6 +432,12 @@ impl EpiIntegration {
             WindowEvent::ScaleFactorChanged { scale_factor, .. } => {
                 self.frame.info.native_pixels_per_point = Some(*scale_factor as _);
             }
+            #[cfg(feature = "dark-light")]
+            WindowEvent::ThemeChanged(winit_theme) if self.follow_system_theme => {
```
emilk, line 436: I don't see any reason to limit this on the `dark-light` feature.

`dark-light` pulls in a specific dependency for detecting the system theme on start-up. Maybe we can do that via `winit` instead?

emilk (commented):
Really cool, thank you so much for working on this!

Since `winit` now has `WindowEvent::ThemeChanged` I wonder if we even need the `dark-light` feature. Can we detect the initial system theme with just winit? There is [`Window::theme`](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.theme) but it seems to be Mac only (Windows not mentioned) and it's unclear if it reads it from system settings or just the previous call to `window.set_theme`.

In either case, I think we should listen for `ThemeChanged` regardless of whether or not the `dark-light` feature is enabled.

crates/eframe/src/native/epi_integration.rs
bash: I have removed `#[cfg(feature = "dark-light")]` from all places where it's not necessary.

Detecting the theme via winit's `theme` function works for Windows and macOS—I have tested the latest change on both my Windows 11 and my macOS devices.

The `theme` function always returns `None` on Linux though (for both [X11](https://github.com/rust-windowing/winit/blob/08bdca19b14fae099076e51816c799d4902fdcbb/src/platform_impl/linux/x11/window.rs#L1620) and [Wayland](https://github.com/rust-windowing/winit/blob/08bdca19b14fae099076e51816c799d4902fdcbb/src/platform_impl/linux/wayland/window/mod.rs#L689)), so there we still have to rely on `dark-light`.

The `NativeOptions::system_theme` function has become a bit of a liar, since it no longer corresponds with the effectively used system theme on Windows and macOS.

emilk (approved):

PR #2751, opened 2023-02-18
Enabled wayland feature for winit when running native

ItsEthra:
This PR fixes clipboard not working on wayland.

### DIFF
--- a/crates/eframe/Cargo.toml
+++ b/crates/eframe/Cargo.toml
@@ -38,6 +38,9 @@ default_fonts = ["egui/default_fonts"]
 ## Use [`glow`](https://github.com/grovesNL/glow) for painting, via [`egui_glow`](https://github.com/emilk/egui/tree/master/crates/egui_glow).
 glow = ["dep:glow", "dep:egui_glow", "dep:glutin", "dep:glutin-winit"]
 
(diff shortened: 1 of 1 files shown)


crates/eframe/Cargo.toml
```diff
@@ -90,6 +90,7 @@ serde = { version = "1", optional = true, features = ["derive"] }
 egui-winit = { version = "0.21.1", path = "../egui-winit", default-features = false, features = [
   "clipboard",
   "links",
+  "wayland",
```
emilk, line 93: This probably should be an opt-in feature, instead of forcing it onto all `eframe` users.

In other words: add a `wayland` feature-flag to `eframe` with `wayland = ["egui-winit/wayland"]`

emilk:
Have you tried https://github.com/emilk/egui/pull/2765?

ItsEthra:
Yes I just tried their fork. When attempting to paste it shows `arboard paste error: The clipboard contents were not available in the requested format or the clipboard is empty.`

crates/eframe/Cargo.toml
ItsEthra: Makes sense, fixed

emilk (approved):

Issue #2752, opened 2023-02-18
show_blocking_widget debug option makes the application hang


YgorSouza:
**Describe the bug**

Enabling the `show_blocking_widget` option causes the application to hang. Observed both locally and on https://www.egui.rs/

**To Reproduce**
1.  Go to <https://www.egui.rs/>
2. Open Settings windows and expand the Debug collapsing header
3. Check the last option "Show what widget blocks the interaction of another widget"
4. The UI stops responding, without using any CPU

**Expected behavior**

The debug option draws the overlapping rectangles and doesn't freeze the UI.

**Screenshots**

**Desktop (please complete the following information):**
 - OS:  Fedora 37
 - Browser  Firefox
 - Version 110.0

**Additional context**

This is the code that runs when this option is enabled:
https://github.com/emilk/egui/blob/7215fdfb7c7407b8085d53052582dac10124bdfc/crates/egui/src/context.rs#L587-L598

The problem seems to be that this code being called inside a `Context::write()` call, whereas both acquiring and using the layer painter requires some `Context::read()` and `Context::write()` calls, so it causes a deadlock.

crumblingstatue:
I can reproduce it in the web version on chromium 110, Linux

PR #2753, opened 2023-02-18
Fix deadlock when using show_blocking_widget

YgorSouza:
Calling the layer painter from inside a write() call causes a deadlock on the Context. This change stores the necessary data (the two overlapping Rects) in the write() call but uses them outside.

Closes <https://github.com/emilk/egui/issues/2752>.


emilk (approved):
Thanks!

PR #2756, opened 2023-02-20
Update serial window example

c-git:
Hi,

I'm new to egui and my main experience so far has been to go through the examples and I found that for the `serial_windows` example I wasn't sure what to expect. I've updated it to what I think would have helped me set better expectations like the other examples. I wasn't quite sure what to do with the screenshot so I left it still to get feedback on the rest first.

I tried my best not to make the code more complicated but felt it was necessary to add a bit more to the UI to help the viewer understand what is happening. I was on the fence about including that the delay is 2 seconds. I think it would help but it would either need to be passed in to keep it in sync which makes the example more complicated or it would be in two places and then might go out of sync.


emilk (approved):
nice!

c-git:
Added the updated screen shot based on the feedback and took the PR out of draft. Let me know if there is anything else I need to do.

emilk (approved):

PR #2763, opened 2023-02-25
fix window position display error (physical pixels converted to logic…

get200:


crates/eframe/src/native/epi_integration.rs
```diff
@@ -236,13 +237,10 @@ pub fn handle_app_output(
...
-            winit::dpi::PhysicalSize {
-                width: (current_pixels_per_point * window_size.x).round(),
-                height: (current_pixels_per_point * window_size.y).round(),
-            }
-            .to_logical::<f32>(native_pixels_per_point(window) as f64),
-        );
+        window.set_inner_size(winit::dpi::LogicalSize {
+            width: window_size.x.round(),
+            height: window_size.y.round(),
+        });
```
emilk, line 243: I actually think the old code is more correct.

The new code is subtlety wrong when the dpi scaling of egui (`current_pixels_per_point`) differs from the native scaling known to winit. `window_size` is the size in _egui points_, which the old code would first convert to physical pixels using `current_pixels_per_point` before handing it over to winit.

emilk (approved):
Great fixes, thanks! Just one thing that _I believe_ is now slightly wrong

crates/eframe/src/native/epi_integration.rs
get200: I think the content zoom should not affect the set window size, so I modified this code, I don't know if this is appropriate, I can change it back, what do you think？

emilk: I think the most common reason for an eframe user to ask for a specific window size is to make their UI elements fit nicely. Therefore it makes sense that the window size is measures in the same units as the UI elements.

get200: 👌

emilk (approved):
Thanks!

emilk:
You got a `cargo cranky` issue to fix!

PR #2765, opened 2023-02-26
egui-winit: fix unsafe API of Clipboard::new

dhardy:
So, `window_clipboard` is not (very actively) maintained and I had a look around...

- Winit doesn't [*yet*](https://github.com/rust-windowing/winit/pull/2156) have a Clipboard API
- `copypasta` is maintained, supports the "primary" clipboard, but makes no attempt to provide a Wayland-X11 abstraction
- `arboard` is maintained, supports HTML and uncompressed RGBA images, but makes no attempt to provide a Wayland-X11 abstraction
- `egui-winit` does have this abstraction, but with an unsafe API

[smithay_clipboard::Clipboard::new](https://docs.rs/smithay-clipboard/latest/smithay_clipboard/struct.Clipboard.html#method.new) notes:

> `display` must be a valid `*mut wl_display` pointer, and it must remain valid for as long as Clipboard object is alive.

This change addresses the first point (prevents calling `Clipboard::new` with an arbitrary pointer).

Addressing the second point with Rust's lifetimes would *probably* be very frustrating. I haven't tried. Addressing it by marking `Clipboard::new` and `State::new` as `unsafe` would be another approach, but would result in all code within `State::new` being compiled in `unsafe` mode, would violate your `#[forbid(unsafe)]` policy and is probably needlessly onerous (given that observing broken behaviour is unlikely), so I opted merely to add documentation here.

---

I ran `sh/check` but `egui-glow` fails `cargo check --all-features` (on X11). Nothing to do with me.

### DIFF
--- a/crates/egui-winit/src/clipboard.rs
+++ b/crates/egui-winit/src/clipboard.rs
@@ -1,4 +1,4 @@
-use std::os::raw::c_void;
+use winit::event_loop::EventLoopWindowTarget;
 
 /// Handles interfacing with the OS clipboard.
 ///
@@ -25,8 +25,13 @@ pub struct Clipboard {
 }
 
 impl Clipboard {
+    /// Construct a new instance
+    ///
+    /// # Safety
+    ///
+    /// The returned `Clipboard` must not outive the input `_event_loop`.
     #[allow(unused_variables)]
-    pub fn new(#[allow(unused_variables)] wayland_display: Option<*mut c_void>) -> Self {
+    pub fn new<T>(_event_loop: &EventLoopWindowTarget<T>) -> Self {
         Self {
             #[cfg(all(feature = "arboard", not(target_os = "android")))]
             arboard: init_arboard(),
@@ -41,7 +46,7 @@ impl Clipboard {
                 ),
                 feature = "smithay-clipboard"
             ))]
-            smithay: init_smithay_clipboard(wayland_display),
+            smithay: init_smithay_clipboard(_event_loop),
 
             clipboard: Default::default(),
         }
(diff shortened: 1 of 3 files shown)


crates/egui-winit/src/lib.rs
```diff
@@ -85,11 +83,12 @@ pub struct State {
 }
 
 impl State {
+    /// Construct a new instance
+    ///
+    /// # Safety
+    ///
+    /// The returned `State` must not outive the input `_event_loop`.
```
emilk, line 90:
```suggestion
    /// The returned `State` must not oultive the input `_event_loop`.
```

crates/egui-winit/src/clipboard.rs
```diff
@@ -25,8 +25,13 @@ pub struct Clipboard {
 }
 
 impl Clipboard {
+    /// Construct a new instance
+    ///
+    /// # Safety
+    ///
+    /// The returned `Clipboard` must not outive the input `_event_loop`.
     #[allow(unused_variables)]
```
emilk, line 33:
```suggestion
```

crates/egui-winit/src/lib.rs
dhardy: The typo fix has a typo!

crates/egui-winit/src/clipboard.rs
```diff
@@ -132,15 +136,20 @@ fn init_arboard() -> Option<arboard::Clipboard> {
...
+    {
+        use winit::platform::wayland::EventLoopWindowTargetExtWayland as _;
+        if let Some(display) = _event_loop.wayland_display() {
+            tracing::debug!("Initializing smithay clipboard…");
+            #[allow(unsafe_code)]
+            return Some(unsafe { smithay_clipboard::Clipboard::new(display) });
+        }
     }
+
+    tracing::debug!("Cannot initialize smithay clipboard without a display handle");
```
emilk, line 153: maybe this makes more sense?

```suggestion
        } else {
            tracing::debug!("Cannot initialize smithay clipboard without a display handle");
            return;
        }
    }
```

emilk (commented):
Looks good!

I think it would be good to document in `crates/egui-winit/Cargo.toml` that the `wayland` feature is required to get the clipboard working on wayland.

dhardy:
Without the `wayland` feature Winit won't support Wayland at all, so I don't think there's any point.

emilk:
needs rustfmt

emilk:
`cargo check --locked --all-features --all-targets` fails

Issue #2470, opened 2022-12-17
Save plot to image file


cdhf:
**Is your feature request related to a problem? Please describe.**
I am using the plot widget to visualize log data for failure analysis of hardware systems. I often need to save the plots to image files to include them in written reports. Often I need the info shown for the current mouse position as well, so I need to make a screenshot of the whole desktop, paste that into a separate program and then clip the image to the desired region.

**Describe the solution you'd like**
It would be great if there was builtin functionality to save a plot to an image file with user defined size. It would be even better, if there was a way generic way to save some part of the window to an image.

**Describe alternatives you've considered**
I've looked at the screenshot example, but it seems that I can only get a screenshot of the whole window, and that the quality of the image is limited.

saturn77:
I am having the same issue, in terms of figuring out a way to save the plot image only to an image. 

Since I have the plot points available in a buffer in egui, one options is to simply make a button such as "save plot image" which would then use the current PlotPoints and write them to a png file. 

Since matplotlib is kind of a standard, one consideration is to use one of the the Rust ports of matplotlib to create a PNG file from the data.

hacknus:
I would also really be interested in a egui built-in solution!

dashdeckers:
I would love this as well! Would be an amazing feature for those using egui to visualize data to be able to save those visualizations.

PR #2769, opened 2023-03-01
Added an example to save plot to image

hacknus:
Added an example to save a plot to an image, based on the [screenshot example](https://github.com/emilk/egui/tree/master/examples/screenshot) - uses unsafe code as well.
Closes <https://github.com/emilk/egui/issues/2470>.


emilk:
This will be easier when https://github.com/emilk/egui/pull/2676 is merged, with no need for unsafe code!

hacknus:
Ok, I guess I'll update this example once 2676 is merged.

hacknus:
adapted to #2676. no unsafe code is required now. Happy to get inputs about this.

examples/save_plot/src/main.rs
```diff
@@ -0,0 +1,148 @@
...
+                });
+            });
+
+            // add some whitespace in y direction
+            ui.add_space(border_y);
+        });
+
+        match &self.screenshot {
+            None => {}
+            Some(screenshot) => {
```
emilk, line 103:
```suggestion
        if let Some(screenshot) = self.screenshot.take() {
```

examples/save_plot/src/main.rs
```diff
@@ -0,0 +1,148 @@
...
+    eframe::run_native(
+        "My egui App with a plot",
+        options,
+        Box::new(|_cc| Box::new(MyApp::default())),
+    )
+}
+
+struct MyApp {
+    picked_path_plot: PathBuf,
+    plot_location: [f32; 4],
```
emilk, line 26:
```suggestion
    /// Unit: fraction of full screen
    plot_location: egui::Rect,
```

examples/save_plot/src/main.rs
```diff
@@ -0,0 +1,148 @@
...
+            ui.add_space(border_y);
+
+            if ui.button("Save Plot").clicked() {
+                if let Some(mut path) = rfd::FileDialog::new().save_file() {
+                    path.set_extension("png");
+                    frame.request_screenshot();
+                    self.save_plot = true;
+                    self.picked_path_plot = path;
+                }
+            }
```
emilk, line 65: You can simplify the state here - no need for `save_plot` nor `picked_path_plot` if you instead use the logic:


```rs
if ui.button("Save Plot").clicked() {
    frame.request_screenshot();
}

```

and then late

```rs
if let Some(screenshot) = self.screenshot.take() {
    if let Some(mut path) = rfd::FileDialog::new().save_file() {
        // save to path
    }
}
```

emilk (changes requested):

hacknus:
Thanks! Implemented your comments - I still have to do the transformation of the plot location with relative values, but it looks cleaner with `egui::Rect`

examples/save_plot/src/main.rs
hacknus: done

hacknus:
Oops, ok I removed the unused import that made the CI workflow fail.

examples/save_plot/src/main.rs
```diff
@@ -0,0 +1,144 @@
...
+                    },
+                );
+                let my_plot = Plot::new("My Plot")
+                    .height(height)
+                    .width(width)
+                    .legend(Legend::default());
+
+                // let's create a dummy line in the plot
+                let graph: Vec<[f64; 2]> = vec![[0.0, 1.0], [2.0, 3.0], [3.0, 2.0]];
+                my_plot.show(ui, |plot_ui| {
```
emilk, line 89: `.show` returns an [`InnerResponse`](https://docs.rs/egui/latest/egui/struct.InnerResponse.html) containing a [`Response`](https://docs.rs/egui/latest/egui/struct.Response.html) which has the `Rect` that the plot covered. You should be able to use that for initializing your `plot_location`

emilk: it also seems easiest to store the location in ui points (as returned by egui) and then pass in `pixels_per_point` to `ColorImage::region` so you don't need to do your own normalization of coordinates

hacknus: Thanks!!! That's what I was missing, now it's much cleaner!

hacknus:
I do not get how to mitigate the failed check concerning the `Cargo.lock` file.

examples/save_plot/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+rust-version = "1.65"
+publish = false
+
+[dependencies]
+eframe = { path = "../../crates/eframe", features = [
+    "__screenshot", # __screenshot is so we can dump a screenshot using EFRAME_SCREENSHOT_TO
+] }
+image = { version = "0.24", default-features = false, features = ["png"] }
+rfd = "0.11.0"
+tracing-subscriber = "0.3"
```
emilk, line 16: Please use `env_logger` instead of `tracing-subscriber` (see other examples). I think this changed since you opened the PR (you may also want to re-base on latest `master` 😬)

examples/save_plot/src/main.rs
```diff
@@ -0,0 +1,102 @@
...
+                    .height(height)
+                    .width(width)
+                    .legend(Legend::default());
+
+                // let's create a dummy line in the plot
+                let graph: Vec<[f64; 2]> = vec![[0.0, 1.0], [2.0, 3.0], [3.0, 2.0]];
+                let inner = my_plot.show(ui, |plot_ui| {
+                    plot_ui.line(Line::new(PlotPoints::from(graph)).name("curve"));
+                });
+                // pass along the position of the plot
```
emilk, line 64:
```suggestion
                // Remember the position of the plot
```

examples/save_plot/src/main.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        if let Some(screenshot) = self.screenshot.take() {
+            if let Some(mut path) = rfd::FileDialog::new().save_file() {
+                path.set_extension("png");
+
+                // for a full size application, we should put this in a different thread,
+                // so that the GUI doesn't lag during saving
+
+                let pixels_per_point = frame.info().native_pixels_per_point;
+                if let Some(plot_location) = plot_rect {
```
emilk, line 80: This can be combined with the earlier if-let:

```rust
if let (Some(screenshot), Some(plot_rect)) = (self.screenshot.take(), plot_rect) {
```

emilk (approved):
Looks good! Just a few small nits

emilk (approved):

examples/save_plot/src/main.rs
```diff
@@ -0,0 +1,99 @@
...
+                // Remember the position of the plot
+                plot_rect = Some(inner.response.rect);
+            });
+
+            // add some whitespace in y direction
+            ui.add_space(border_y);
+        });
+
+        if let (Some(screenshot), Some(plot_location)) = (self.screenshot.take(), plot_rect) {
+            if let Some(mut path) = rfd::FileDialog::new().save_file() {
```
bryevdv, line 72: Currently always defaults to "Untitled" (.png) but we could set a a different default filename if desired

PR #2771, opened 2023-03-01
Wip

amfaber:
Cranky and deny have been massaged to get checks to pass.

Because of https://github.com/ebarnard/rust-plist/commit/aa3821390fe804ad73bec28180a07c7ac31e0dc8 upgrading its dependency on base64 before bumping its version, base64 has been added to ban skips of cargo deny.


amfaber:
Accidentally created a PR upstream instead of to the main of my own fork. Disregard.

PR #2772, opened 2023-03-02
Fix: return correct index for out of range cursor

tomcur:
For out of range character indices, `byte_index_from_char_index(s, char_index)`  should return the last byte index + 1, not the last char index + 1.

### DIFF
--- a/crates/egui/src/widgets/text_edit/text_buffer.rs
+++ b/crates/egui/src/widgets/text_edit/text_buffer.rs
@@ -123,5 +123,5 @@ fn byte_index_from_char_index(s: &str, char_index: usize) -> usize {
             return bi;
         }
     }
-    s.len()
+    s.as_bytes().len()
 }


crates/egui/src/widgets/text_edit/text_buffer.rs
```diff
@@ -123,5 +123,5 @@ fn byte_index_from_char_index(s: &str, char_index: usize) -> usize {
             return bi;
         }
     }
-    s.len()
+    s.as_bytes().len()
```
Mingun, line 126: This change does nothing because `as_bytes()` for string just reinterprets type `&str` as a type `&[u8]`. This is the same thing

tomcur: > This change does nothing because `as_bytes()` for string just reinterprets type `&str` as a type `&[u8]`. This is the same thing

Whoops, thanks, you're right. I had been working under the assumption `&str::len` to be the number of characters, but it is indeed the byte length.

Issue #2723, opened 2023-02-12
Missing forbid unsafe_code attributes


YgorSouza:
Some crates have the [safety dance](https://github.com/rust-secure-code/safety-dance/) badge in their individual README files, but none of them have the required `#![forbid(unsafe_code)]` attribute. Indeed, some of them display the badge even though they actually have unsafe code, namely eframe and egui-winit, which is misleading. The badge should be removed from crates that have unsafe code, and those that don't should have the attribute so they are parsed correctly by cargo-geiger.

emilk:
Good catch! I've removed the misleading badges.

As for parsing correctly with `cargo-grieger` - is that something you'd be willing to contribute?

PR #2773, opened 2023-03-02
Add #![forbid(unsafe_code)] where unsafe isn't needed

YgorSouza:
To be consistent with the use of the safety badge in the README.

Closes <https://github.com/emilk/egui/issues/2723>.


emilk (approved):
Thanks!

PR #2775, opened 2023-03-02
Implement toggle for `egui::Window` to only accept move events from the title bar.

### LINKED ISSUE #2774: Allow setting window to only support dragging from title bar
I want custom dragging behaviour inside my `egui::Window` (controlling a 3D camera), currently this is not possible because dragging the mouse inside the window moves the window itself.

A simple toggle on the window builder would be nice, `window.movable_from_content(false)` or something.

A custom widget that implements this behaviour could be possible, however this is a lot of extra work either on the user, or would share much duplicate code with `egui::Window`.

NotAPenguin0:
Closes #2774. For a more detailed description of the problem, refer to that issue.

[demo gif](https://gifyu.com/image/S7ukY).

List of changes
- Added `egui::Window::movable_from_contents()` to toggle the new behaviour on.
- In `window::window_interaction()`, split resize and move behaviour, since they might now use two different rects.

My apologies that there are so many commits, it's the unfortunate result of some dependencies I have to manage while developing. Just squashing on merge should get rid of the problem.

### DIFF
--- a/crates/egui/src/containers/window.rs
+++ b/crates/egui/src/containers/window.rs
@@ -32,6 +32,7 @@ pub struct Window<'open> {
     collapsible: bool,
     default_open: bool,
     with_title_bar: bool,
+    move_title_bar_only: bool,
 }
 
 impl<'open> Window<'open> {
@@ -54,6 +55,7 @@ impl<'open> Window<'open> {
             collapsible: true,
             default_open: true,
             with_title_bar: true,
+            move_title_bar_only: false,
         }
     }
 
@@ -88,6 +90,12 @@ impl<'open> Window<'open> {
     /// If `false` the window will be immovable.
     pub fn movable(mut self, movable: bool) -> Self {
         self.area = self.area.movable(movable);
+        self
+    }
+
+    /// Only allow moving by dragging from the title bar.
+    pub fn movable_from_contents(mut self, movable: bool) -> Self {
+        self.move_title_bar_only = !movable;
         self
     }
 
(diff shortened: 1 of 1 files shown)


crates/egui/src/containers/window.rs
```diff
@@ -91,6 +93,12 @@ impl<'open> Window<'open> {
         self
     }
 
+    /// Only allow moving by dragging from the title bar.
+    pub fn movable_from_contents(mut self, movable: bool) -> Self {
```
emilk, line 97:
```suggestion
    /// Allow moving the window by dragging its content?
    ///
    /// Default: `true`.
    /// Regardless of this value, dragging the title-bar will move the window (if moveable).
    pub fn movable_from_contents(mut self, movable: bool) -> Self {
```

crates/egui/src/containers/window.rs
```diff
@@ -91,6 +93,12 @@ impl<'open> Window<'open> {
...
+        self.move_title_bar_only = !movable;
```
emilk, line 98: avoid negation in names

```suggestion
        self.movable_from_contents = movable;
```

emilk (changes requested):

emilk (changes requested):
This seems like a lot of new code and complexity for a small change.

We should be able to just pipe in a smaller rect into `move_and_resize_window`?

NotAPenguin0:
That is what I did initially, the problem is that this also moves the resize handle to a very weird and awkward position. I've also been able to implement this exact behaviour without touching egui internals at all, by simply doing
```rs
let (response, painter) = ui.allocate_painter(remaining_size, Sense::drag());
```
and handling the `drag` value from this `Response` myself, so I don't think this change is needed anymore

emilk:
Closing: this landed in #8183 as `Window::drag(WindowDrag::TitleBar)`. Thanks for the PR!

PR #2778, opened 2023-03-04
[WIP] Fix submenu flickering

### LINKED ISSUE #2706: Context submenu flickers when menu contacts edge of window
New in 0.21.0.

https://user-images.githubusercontent.com/3969183/217979574-71cd83f6-aef6-454d-8aa7-d71325a08b9e.mp4

juancampa:
## Problem
Submenus flicker and are generally unusable, as shown in #2706.

There are two separate causes for this:

 1. Menu Areas were being created with both `constrain` _and_ `fixed_pos`. These two properties fight with each other as they are both trying to position the menu in a different position.
 2. `MenuState::area_contains` uses it's own logic to determine if the pointer is inside of the menu but this logic is not the same as `Context::layer_id_at`.
 
 (1) causes flickering when the submenu is constrained and ends up on top of the parent menu
 (2) causes flickering when the pointer is 5px away from the submenu

## Solution

(1) ~~seems fully solved, although there might be a better solution~~ see first comment below

(2) I hard-coded a value for `resize_grab_radius_side` for now, which fixes the issue unless someone tweaks this value. Feedback on how to properly solve this would be appreciated. Left some comments explaining the issue

Closes #2706

### DIFF
--- a/crates/egui/src/containers/area.rs
+++ b/crates/egui/src/containers/area.rs
@@ -142,7 +142,7 @@ impl Area {
 
     /// Positions the window and prevents it from being moved
     pub fn fixed_pos(mut self, fixed_pos: impl Into<Pos2>) -> Self {
-        self.new_pos = Some(fixed_pos.into());
+        self.default_pos = Some(fixed_pos.into());
         self.movable = false;
         self
     }
--- a/crates/egui/src/containers/area.rs
+++ b/crates/egui/src/containers/area.rs
@@ -142,7 +142,7 @@ impl Area {
 
     /// Positions the window and prevents it from being moved
     pub fn fixed_pos(mut self, fixed_pos: impl Into<Pos2>) -> Self {
-        self.default_pos = Some(fixed_pos.into());
+        self.new_pos = Some(fixed_pos.into());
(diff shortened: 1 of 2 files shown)


juancampa:
I found an issue with (1). If the menu is constrained once, but the window is resized to be bigger, the menu stays in the constrained position when it's opened again.

A solution would be clear the memory for the menu once it closes or two somehow make `constrain` and `fixed_pos` get along.

emilk:
Thanks for the fix!

I think we should consider re-defining what keeps a sub-menu open.

The current behavior is, as far as I recall, "keep sub-menu open while hovering it or any child". This leads to sub-menus closing very easily when the mouse strays outside their active area.

I think we should change this to "remain open until another sub-menu-button is hovered, unless is is a descendant UI". That's how most menu systems work.

This is of course a bigger change, and this PR is a good band-aid in the mean-time.

emilk:
OK; I'll convert this to a draft for now then!

juancampa:
Closing since it doesn't seem to be happening anymore: https://github.com/emilk/egui/issues/2706#issuecomment-1615800751

PR #2780, opened 2023-03-04
Fix resizable columns option in the table demo

Bobo1239:
The "Resizable columns" check box wasn't doing anything in the table demo before.


emilk (approved):
Good catch

PR #2781, opened 2023-03-04
demo app: use `enum` instead of strings for demo-selector anchor

XyLyXyRR:


crates/egui_demo_app/src/wrap_app.rs
```diff
@@ -75,6 +76,37 @@ impl eframe::App for ColorTestApp {
...
+    #[cfg(feature = "http")]
+    Http,
+    Clock,
+    #[cfg(any(feature = "glow", feature = "wgpu"))]
+    Custom3d,
+    Colors,
+}
+
+impl Display for Anchor {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
```
emilk, line 93: no need to pollute the file namespace

```suggestion
impl std::fmt::Display for Anchor {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```

emilk (changes requested):
Nice!

Issue #2718, opened 2023-02-11
Expose raw mouse wheel delta to allow alternative input


YgorSouza:
**Is your feature request related to a problem? Please describe.**

egui converts `MouseWheel` events to `Scroll` or `Zoom` events based on which modifier keys are pressed ([here](https://github.com/emilk/egui/blob/530e9f667c0969b09edb68f7701f20c52a273d3a/crates/egui-winit/src/lib.rs#L551-L575) and [here](https://github.com/emilk/egui/blob/530e9f667c0969b09edb68f7701f20c52a273d3a/crates/eframe/src/web/events.rs#L396-L439)). The raw MouseWheel delta does not seem to be available anywhere. This makes it hard to implement alternative behaviors involving the mouse wheel, such as the one requested in #1164.

**Describe the solution you'd like**

Make the raw `MouseWheel` delta available somewhere in the [InputState](https://github.com/emilk/egui/blob/530e9f667c0969b09edb68f7701f20c52a273d3a/crates/egui/src/input_state.rs#L25) struct.

**Describe alternatives you've considered**

I have implemented an example of alternative plot zoom/scroll controls in #2714, showing how this can be done currently. You basically have to reverse what egui does internally to deduce the mouse wheel delta from the `Zoom` and `Scroll` events, which is cumbersome and forces the developer to know internal details of the library. Moreover, it seems impossible to distinguish a mouse wheel event when Ctrl is down from other input events that translate to `Zoom`, so this workaround does not cover all cases.

**Additional context**

I am willing to implement this myself, but I am not sure of how to reconcile it with the existing logic. The current `Scroll` and `Zoom` events should of course stay as they are to avoid breaking the entire library, but should the raw `MouseWheelDelta` be an additional duplicate event? Or an `Option` field that is reset at the start of every frame and then assigned if there is a MouseWheel event in the backend? Neither option seems very good, so I would like to brainstorm some alternatives.

YgorSouza:
Looking at how the keyboard events are handled, egui translates them to `Copy` and `Paste` events where appropriate, but it also sends the key event itself in this case, so there can be two events for the same key press, which allows users of the library to use Ctrl+C, Ctrl+V etc for other purposes. So it seems that it would be consistent to add a `MouseWheel` event that is sent in addition to the `Scroll` or `Zoom` event. However, there are a few things to consider:

[...]

PR #2782, opened 2023-03-04
Add raw mouse wheel event

YgorSouza:
The event is sent as it comes from the backend, so it will follow different conventions depending on the target, and it is up to the user code to deal with that. The goal is to allow advanced users to implement alternative UI controls, e.g., using Ctrl to scroll the plot horizontally instead of zooming, or use Shift to scroll faster instead of changing direction.

Closes <https://github.com/emilk/egui/issues/2718>.

### DIFF
--- a/crates/eframe/src/web/events.rs
+++ b/crates/eframe/src/web/events.rs
@@ -409,7 +409,8 @@ pub fn install_canvas_events(runner_container: &mut AppRunnerContainer) -> Resul
                     web_sys::WheelEvent::DOM_DELTA_PAGE => egui::MouseWheelUnit::Page,
                     _ => return,
                 };
-                let delta = egui::vec2(event.delta_x() as f32, event.delta_y() as f32);
+                // delta sign is flipped to match native convention
+                let delta = -egui::vec2(event.delta_x() as f32, event.delta_y() as f32);
                 let modifiers = runner_lock.input.raw.modifiers;
 
                 runner_lock.input.raw.events.push(egui::Event::MouseWheel {
--- a/crates/eframe/src/web/events.rs
+++ b/crates/eframe/src/web/events.rs
@@ -402,7 +402,7 @@ pub fn install_canvas_events(runner_container: &mut AppRunnerContainer) -> Resul
         &canvas,
         "wheel",
         |event: web_sys::WheelEvent, mut runner_lock| {
-            let mut push_raw_event = || {
+            {
                 let unit = match event.delta_mode() {
                     web_sys::WheelEvent::DOM_DELTA_PIXEL => egui::MouseWheelUnit::Point,
                     web_sys::WheelEvent::DOM_DELTA_LINE => egui::MouseWheelUnit::Line,
@@ -417,8 +417,7 @@ pub fn install_canvas_events(runner_container: &mut AppRunnerContainer) -> Resul
                     delta,
                     modifiers,
                 });
-            };
-            push_raw_event();
+            }
             let scroll_multiplier = match event.delta_mode() {
                 web_sys::WheelEvent::DOM_DELTA_PAGE => {
                     canvas_size_in_points(runner_lock.canvas_id()).y
--- a/crates/eframe/src/web/events.rs
+++ b/crates/eframe/src/web/events.rs
@@ -404,7 +404,7 @@ pub fn install_canvas_events(runner_container: &mut AppRunnerContainer) -> Resul
         |event: web_sys::WheelEvent, mut runner_lock| {
             let mut push_raw_event = || {
                 let unit = match event.delta_mode() {
-                    web_sys::WheelEvent::DOM_DELTA_PIXEL => egui::MouseWheelUnit::Pixel,
+                    web_sys::WheelEvent::DOM_DELTA_PIXEL => egui::MouseWheelUnit::Point,
                     web_sys::WheelEvent::DOM_DELTA_LINE => egui::MouseWheelUnit::Line,
                     web_sys::WheelEvent::DOM_DELTA_PAGE => egui::MouseWheelUnit::Page,
                     3u32..=u32::MAX => return,
--- a/crates/eframe/src/web/events.rs
+++ b/crates/eframe/src/web/events.rs
@@ -407,7 +407,7 @@ pub fn install_canvas_events(runner_container: &mut AppRunnerContainer) -> Resul
                     web_sys::WheelEvent::DOM_DELTA_PIXEL => egui::MouseWheelUnit::Point,
                     web_sys::WheelEvent::DOM_DELTA_LINE => egui::MouseWheelUnit::Line,
                     web_sys::WheelEvent::DOM_DELTA_PAGE => egui::MouseWheelUnit::Page,
-                    3u32..=u32::MAX => return,
+                    _ => return,
                 };
                 let delta = egui::vec2(event.delta_x() as f32, event.delta_y() as f32);
                 let modifiers = runner_lock.input.raw.modifiers;
--- a/crates/eframe/src/web/events.rs
+++ b/crates/eframe/src/web/events.rs
@@ -402,37 +402,33 @@ pub fn install_canvas_events(runner_container: &mut AppRunnerContainer) -> Resul
         &canvas,
         "wheel",
         |event: web_sys::WheelEvent, mut runner_lock| {
-            {
-                let unit = match event.delta_mode() {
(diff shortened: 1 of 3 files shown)


YgorSouza:
This implementation provides maximum information and flexibility to the user of egui, but it also gives them the burden of having to know the specifics of the backend/target they're building for. For example, in winit, scrolling down one step produces a delta of (0.0, -1.0) lines, whereas in web the delta is (0.0, 6.0) lines. Also, the x and y coordinates are swapped when holding down shift on macOS (both winit and web), but not on other targets.

An alternative would be to create a universal convention, similar to how the `Scroll` event is handled currently, while still providing a specific `MouseWheel` event in addition to the opinionated conversion to `Zoom`/`Scroll`, so the user can handle the wheel event as they prefer but still get consistent behavior across all platforms without having to know the specifics of each one.

But this convention would end up being very similar to the `Scroll` event itself, so maybe an easier solution would be to just not convert the mouse wheel to `Zoom` automatically, or allow the user to opt out of this conversion and always get a Scroll event even when Ctrl is down.

crates/egui-winit/src/lib.rs
```diff
@@ -549,6 +549,24 @@ impl State {
...
+                }) => (egui::MouseWheelUnit::Pixel, egui::vec2(x as f32, y as f32)),
+            };
+            let modifiers = self.egui_input.modifiers;
+            self.egui_input.events.push(egui::Event::MouseWheel {
+                unit,
+                delta,
+                modifiers,
+            });
+        };
+        push_raw_event();
```
emilk, line 569: no need for a lambda here afaict

crates/eframe/src/web/events.rs
```diff
@@ -402,6 +402,23 @@ pub fn install_canvas_events(runner_container: &mut AppRunnerContainer) -> Resul
         &canvas,
         "wheel",
         |event: web_sys::WheelEvent, mut runner_lock| {
+            let mut push_raw_event = || {
+                let unit = match event.delta_mode() {
+                    web_sys::WheelEvent::DOM_DELTA_PIXEL => egui::MouseWheelUnit::Pixel,
```
emilk, line 407: What is called a "pixel" on the web, we call "point" in egui

```suggestion
                    web_sys::WheelEvent::DOM_DELTA_PIXEL => egui::MouseWheelUnit::Point,
```

crates/egui-winit/src/lib.rs
```diff
@@ -549,6 +549,24 @@ impl State {
...
     fn on_mouse_wheel(&mut self, delta: winit::event::MouseScrollDelta) {
+        let mut push_raw_event = || {
+            let (unit, delta) = match delta {
+                winit::event::MouseScrollDelta::LineDelta(x, y) => {
+                    (egui::MouseWheelUnit::Line, egui::vec2(x, y))
+                }
+                winit::event::MouseScrollDelta::PixelDelta(winit::dpi::PhysicalPosition {
+                    x,
+                    y,
+                }) => (egui::MouseWheelUnit::Pixel, egui::vec2(x as f32, y as f32)),
```
emilk, line 560: Let's use points for this instead to be consistent across different DPI:s:
```suggestion
                }) => (egui::MouseWheelUnit::Pixel, egui::vec2(x as f32, y as f32) / self.pixels_per_point()),
```

crates/egui/src/data/input.rs
```diff
@@ -278,6 +278,22 @@ pub enum Event {
         force: f32,
     },
 
+    /// A raw mouse wheel event as sent by the backend (minus the z coordinate),
+    /// for implementing alternative custom controls.
+    /// If this event is handled, Zoom and Scroll should be ignored, as they might be duplicates
+    /// coming from the same input device.
```
emilk, line 284:
```suggestion
    /// Note that the same event can also trigger [`Self::Zoom`] and [`Self::Scroll`],
    /// so you probably want to handle only one of them.
```
```

File: crates/egui/src/data/input.rs [hunk 5]
```diff
@@ -278,6 +278,22 @@ pub enum Event {
...
+    MouseWheel {
+        /// The unit of scrolling: pixels, lines etc
```
emilk, line 286:
```suggestion
        /// The unit of scrolling: points, lines, or pages.
```

crates/egui/src/data/input.rs
```diff
@@ -891,6 +907,20 @@ pub enum TouchPhase {
     Cancel,
 }
 
+/// The unit associated with the numeric value of a mouse wheel event
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub enum MouseWheelUnit {
+    /// Number of pixels
+    Pixel,
```
emilk, line 915:
```suggestion
    /// Number of ui points (logical pixels)
    Points,
```

emilk (changes requested):
Good idea, but let's use the same unit (points) for both web and native

crates/eframe/src/web/events.rs
```diff
@@ -402,6 +402,22 @@ pub fn install_canvas_events(runner_container: &mut AppRunnerContainer) -> Resul
...
+                let delta = egui::vec2(event.delta_x() as f32, event.delta_y() as f32);
+                let modifiers = runner_lock.input.raw.modifiers;
+
+                runner_lock.input.raw.events.push(egui::Event::MouseWheel {
+                    unit,
+                    delta,
+                    modifiers,
+                });
+            }
             let scroll_multiplier = match event.delta_mode() {
```
emilk, line 421: you can exhaustively match on `unit` here instead now
```suggestion
            let scroll_multiplier = match unit {
```

crates/eframe/src/web/events.rs
```diff
@@ -402,6 +402,22 @@ pub fn install_canvas_events(runner_container: &mut AppRunnerContainer) -> Resul
...
         "wheel",
         |event: web_sys::WheelEvent, mut runner_lock| {
+            {
+                let unit = match event.delta_mode() {
+                    web_sys::WheelEvent::DOM_DELTA_PIXEL => egui::MouseWheelUnit::Point,
+                    web_sys::WheelEvent::DOM_DELTA_LINE => egui::MouseWheelUnit::Line,
+                    web_sys::WheelEvent::DOM_DELTA_PAGE => egui::MouseWheelUnit::Page,
+                    3u32..=u32::MAX => return,
+                };
+                let delta = egui::vec2(event.delta_x() as f32, event.delta_y() as f32);
```
emilk, line 412: further down in this code we flip the sign of the delta to match that of winit - don't you think we should do that here too?

Please test your code on both native and web to make sure the signs are the same!

YgorSouza: That's actually what I was asking about in the first comment in this PR. There are multiple differences between targets, and if egui changes only some of them, users still have to be aware of the others, which might make it even more confusing. And if we change them all, then it just becomes a duplicate of the `Scroll` event (but without accounting for the modifier keys). That could be a choice, but I wasn't sure of where to draw the line. Do I just flip the sign of the delta, or do I handle the other differences as well?

crates/eframe/src/web/events.rs
YgorSouza: I tried to keep the raw event part self-contained to avoid introducing bugs, and to make it easy to remove or comment out if needed. But depending on the answer to the comment below about the target differences, it might make more sense to merge the two parts into one, indeed.

crates/eframe/src/web/events.rs
emilk: Without the signs being correct it is almost useless imho.

A common use-case for this is to catch a single step of the scroll-wheel. The magnitude of this is different on different platforms, but as long as the sign is the same a user can still do `let change = delta.y.sign();` to catch the discreet scroll-tick events.

YgorSouza: I see your point. Although I guess you would have to do `let change = (delta.x + delta.y).sign();`, because of the macOS thing as mentioned down below in the code. But in any case, this seems like a good compromise. I changed it.

I also noticed that the touchpad gesture is also returned as a wheel event by the backend, but it uses the unit `Point` on native and `Line` on web, both with about the same delta for the same speed. I don't know if that is worth dealing with as well. At least the sign is the same, though, and this already affects the `Scroll` event anyway. Scrolling with the touchpad on web goes much faster than on native.

emilk (approved):

PR #2790, opened 2023-03-07
📜 Replace ComboBox::show_index String with Into<TextWidget>

tosti007:
The `ComboBox::show_index` function currently takes a `impl Fn(usize) -> String` and passes that `String` to functions taking a `impl Into<WidgetText>`. We can make the `show_index` function more generic by using `Into<WidgetText>` as well.


emilk (approved):
Makes sense!

Issue #2578, opened 2023-01-12
Text wrapping too late


mikedilger:
**Describe the bug**

Under certain conditions, text wrapping leaves one line of text too wide compared to the container, in my case pushing the container width (or it's ancestors) beyond the bounds of the window.

**To Reproduce**

1. Within an area that is not as wide as the window...
2. Horizontally wrapped...
3. Render a short link, and then a very long string of characters with no space to natrually break them on.
4. The first line of that very long string of characters will be too long, but subsequent lines won't.

**Expected behavior**

Wrapping that doesn't expand the width of the container.

**Screenshots**

https://user-images.githubusercontent.com/1653275/212086814-c878a200-f7ab-4a84-a5f0-a0b89ec740a5.png

Normally subsequent posts (below the horizontal lines) line up on the left edge. Here, they are pushed left beyond the edge of the window.

**Desktop (please complete the following information):**

 - OS: ArchLinux (X11/winit)
 - Browser: n/a
 - Version: eframe 0.20.1

emilk:
A simple reproduce would help a lot!

mikedilger:
Ok I did some digging and this reproduces it

main.rs:
````
use eframe::egui;

fn main() {
    let native_options = eframe::NativeOptions::default();
    eframe::run_native("My egui App", native_options, Box::new(|cc| Box::new(MyEguiApp::new(cc))));
}

#[derive(Default)]
struct MyEguiApp {}

impl MyEguiApp {
    fn new(cc: &eframe::CreationContext<'_>) -> Self {
        Self::default()
    }
}

impl eframe::App for MyEguiApp {
   fn update(&mut self, ctx: &egui::Context, frame: &mut eframe::Frame) {
       egui::CentralPanel::default().show(ctx, |ui| {
           ui.horizontal_wrapped(|ui| {
               ui.hyperlink_to("@npub1vdaeclr2mnntmyw...", "whocares");
[...]

PR #2791, opened 2023-03-08
Fix issue #2578

bu5hm4nn:
- `row_start_x` tracks a virtual position in the source paragraph (the one that is too long) for which length has already been processed. When creating an empty row, this position should not be updated as no glyphs were consumed from the source paragraph.
- added example that would demonstrate the problem if the line was included, and that is fixed with this commit


crates/epaint/src/text/text_layout.rs
```diff
@@ -198,28 +201,17 @@ fn line_break(
...
 
         if job.wrap.max_rows > 0 && non_empty_rows >= job.wrap.max_rows {
             break;
         }
 
-        if potential_row_width > job.wrap.max_width {
-            if first_row_indentation > 0.0
-                && !row_break_candidates.has_good_candidate(job.wrap.break_anywhere)
-            {
-                // Allow the first row to be completely empty, because we know there will be more space on the next row:
```
emilk, line 211: If `first_row_indentation == 0.0` then we never want an empty first row. We would rather overflow `max_width`, because otherwise we will just run into the same problem on the next row. For instance if `max_width == 0.0` we want a layout where we have exactly one glyph per row.

If `first_row_indentation > 0.0` however we are fine with having an empty row, because there will be more space available on the next row, and that might be enough to fit the first word or whatever.

bu5hm4nn: I just ran this in examples/wrapped_layout and it seems to work as you describe (one glyph per row).

```
let mut chui = ui.child_ui(
    Rect::from_two_pos(ui.cursor().min, ui.cursor().min),
    Layout::left_to_right(Align::Center),
);
chui.horizontal_wrapped(|ui| {
    ui.label("Frick");
});
```

emilk:
Is there more work to be done here, or is it ready for review?

bu5hm4nn:
Since this bug in `ui.horizontal_wrapped()` persists in the latest release, I have updated this branch by cherry picking the fixes ontop of a fresh checkout from master. Therefore the force push.

jb55:
since I'm hitting this issue on https://github.com/damus-io/notedeck I will try this and report back!

jb55:
tried @bu5hm4nn 's fork for 0.28.1 at f40370c48ae2e07d2bc1d7ec33d094d29dc34e70 with no luck. whatever it fixed must have been specific to gossip? I'm still seeing wrapping issues in [notedeck] columns:

Screenshot 2024-11-26 at 11 49 03


[notedeck]: https://github.com/damus-io/notedeck/pull/482

PR #2792, opened 2023-03-08
Upgrade AccessKit-related dependencies.

ndarilek:


ndarilek:
I can't read the CI output for `cargo deny` with my screen readerto figure out what the issue is. Sorry.

Techie-Pi:
The errors from `cargo deny` are the following:
- error, duplicate: found 2 duplicate entries for crate 'base64'
- error, duplicate: found 2 duplicate entries for crate 'jni'
- error, duplicate: found 2 duplicate entries for crate 'memoffset'

Looks like `accesskit_winit` and `accesskit` should be on the same version, but that's only my guess.
Hope this helps! 😄

ndarilek:
Thanks. They're not version-locked--I don't necessarily know which 
versions are explicitly compatible but the latest versions should work fine.

emilk:
The updated `accesskit` pulls in newer dependencies that cause duplicates of some dependencies. `base64` and `memoffset` are so minor that we can ignore them in `deny.toml`, but `jni` sounds a bit worse.

I'll take a look once this is merged:
* https://github.com/emilk/egui/pull/2852

PR #2804, opened 2023-03-13
Prefer the client width/height for the canvas parent

samitbasu:
Normally, `eframe` takes over the entire webpage, but sometimes, you
would like it to "play nice" and keep its canvas to a finite area.  This mostly
works as long as the containing `div` has the right style properties. 
However, the use of `scrollheight` and `scrollwidth` breaks the responsive case in which the page is resized.  The canvas will grow, but it will not shrink.

This PR resolves https://github.com/emilk/egui/issues/2805


samitbasu:
Yeah - I'm not sure what to do about the `cargo deny` warnings.

samitbasu:
I resync-ed with `master`.  It looks like the `cargo deny` issues were fixed there.

PR #2813, opened 2023-03-15
Apply 'with_active' option to winit

Dixeran:
There is a option in winit's WindowBuilder called `with_active` which allow to avoid granting focus when window initially opened. This PR is tend to add this option into `NativeOptions` of eframe.


emilk (approved):

emilk (commented):
Looks good, but needs build fix

PR #2816, opened 2023-03-16
Add character limit to TextEdit

wzid:
Added `char_limit` to TextEdit to provide a way to limit the amount of characters.

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -926,12 +926,13 @@ fn events(
                 if !text_to_insert.is_empty() {
                     let mut ccursor = delete_selected(text, &cursor_range);
 
-                    if !multiline && char_limit.is_some() {
-                        let mut new_string = text_to_insert.clone();
-                        new_string.truncate(char_limit.unwrap() - text.as_str().len());
-                        insert_text(&mut ccursor, text, &new_string);
-                    } else {
-                        insert_text(&mut ccursor, text, &text_to_insert);
+                    match char_limit {
+                        Some(limit) if !multiline => {
+                            let mut new_string = text_to_insert.clone();
+                            new_string.truncate(limit - text.as_str().len());
+                            insert_text(&mut ccursor, text, &new_string);
+                        }
+                        _ => insert_text(&mut ccursor, text, text_to_insert),
                     }
 
                     Some(CCursorRange::one(ccursor))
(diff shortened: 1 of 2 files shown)


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -922,7 +944,16 @@ fn events(
                 // Newlines are handled by `Key::Enter`.
                 if !text_to_insert.is_empty() && text_to_insert != "\n" && text_to_insert != "\r" {
                     let mut ccursor = delete_selected(text, &cursor_range);
-                    insert_text(&mut ccursor, text, text_to_insert);
+
+                    match char_limit {
+                        Some(limit) if !multiline => {
+                            let mut new_string = text_to_insert.clone();
+                            new_string.truncate(limit - text.as_str().len());
```
xTibor, line 951: This chops UTF-8 sequences at the middle, causing panics.

```
thread 'main' panicked at 'assertion failed: self.is_char_boundary(new_len)', /rustc/897e37553bba8b42751c67658967889d11ecd120/library/alloc/src/string.rs:1280:13
stack backtrace:
   ...
   4: egui::widgets::text_edit::builder::events
             at ./crates/egui/src/widgets/text_edit/builder.rs:951:29
   ...
```

wzid: Thank you for bringing this up, I have just fixed it.
😃

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -922,7 +951,23 @@ fn events(
...
+                            let cutoff = limit.saturating_sub(text.as_str().len());
+
+                            new_string = match new_string.char_indices().nth(cutoff) {
+                                None => new_string,
+                                Some((idx, _)) => new_string[..idx].to_owned(),
+                            };
+
+                            insert_text(&mut ccursor, text, &new_string);
+                        }
+                        _ => insert_text(&mut ccursor, text, text_to_insert),
```
emilk, line 968: inline this code into `insert_text` instead of duplicating the code

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -70,6 +70,7 @@ pub struct TextEdit<'t> {
     min_size: Vec2,
     align: Align2,
     clip_text: bool,
+    char_limit: Option<usize>,
```
emilk, line 73: This could be `char_limit: usize` with a default of `usize::MAX` for simpler code

emilk (changes requested):

crates/egui/src/widgets/text_edit/builder.rs
wzid: Done

emilk (approved):

Issue #2826, opened 2023-03-20
scroll bar visibility default changed


lunixbochs:
https://github.com/emilk/egui/commit/5910144112c69426b45c4750e6dfed35a7e6c75c#diff-14b897b3019738ddcdb1d3bb8cad1346486ea4aea57bde3299c774925323a84dR149

This patch changed the default scroll bar visibility from VisibleWhenNeeded to AlwaysHidden. This means scroll bars don't show up on HEAD.

Barugon:
Scroll bars are now always hidden for `ComboBox`.

PR #2827, opened 2023-03-20
fix scroll bars not appearing (#2826)

lunixbochs:
This fixes a scroll bar visibility default regression introduced by https://github.com/emilk/egui/pull/2729.

Closes <https://github.com/emilk/egui/issues/2826>.


emilk (approved):
Good catch 😬

PR #2828, opened 2023-03-20
Allow temporarily preventing focusable widgets

jabuwu:
Currently there is no way to prevent focusing of widgets without locking focus to a particular widget. There are cases where it makes sense to disable focus for entire sections of widgets, particularly for modals. An outstanding issue in all of the modal implementations I've seen is that the tab key allows interacting with widgets behind the modal.

By calling `prevent_focus` before rendering all non-modal widgets, then `allow_focus` before a modal, it's possible to allow tab focus only within the modal's widgets. It's also necessary to reset focus when the modal opens by focusing some widget in the modal.

I don't think this is the ideal solution, since it might not work in a generic context like that of `egui-modal`.

I'm curious if there's other thoughts about how this can be achieved.

Related: https://github.com/n00kii/egui-modal/issues/4

Todo:
- [ ] Add to CHANGELOG.md

### DIFF
--- a/crates/egui/src/memory.rs
+++ b/crates/egui/src/memory.rs
@@ -199,6 +199,9 @@ pub(crate) struct Focus {
 
     /// Set at the beginning of the frame, set to `false` when "used".
     pressed_shift_tab: bool,
+
+    /// If `true`, calls to [`Memory::interested_in_focus`] will be ignored.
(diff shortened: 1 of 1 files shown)


jabuwu:
We don't need this anymore after egui got modals.

PR #2832, opened 2023-03-23
Fix the OS check for windows

jleibs:


emilk:
Please write a unit-test before submitting a PR

troll-face

Issue #2818, opened 2023-03-17
`DragValue` edit is discarded when clicked off


Barugon:
**Describe the bug**
If I click a `DragValue` and type in a new value then that new value is discarded if I click on another `DragValue` or `TextEdit`.

Video: https://youtube.com/shorts/_wxqB0XjkIY

paul-hansen:
This is really annoying when you want to edit a list of values with the keyboard by tabbing through them. The only way to make the changes save is to hit the return key, which resets the tabbing state so you have to tab back to where you were from the start.

For example, I have this Ui:
![image](https://user-images.githubusercontent.com/7019130/227364697-ad4cbb3a-8919-4289-9884-a7317cbe03a4.png)

I would like to be able to select the first value, translation.x, type the new value, hit tab, type the next value, hit tab, etc. Currently that's not possible as it doesn't save the values you typed when you hit tab.

Barugon:
I'm using the return key to mean "Ok" when displaying a dialog (standard OS behavior) so using it to enter a single value is not good.

paul-hansen:
Looks like it was introduced in this commit: https://github.com/emilk/egui/commit/b40dba18c642fa959c4bad59fea99416a31f0825

Used git bisect and tested with the hello_world example to find the first commit where it wasn't working. The drag value created by the Slider in the example is affected so it made a good test since DragValue isn't used in the examples directly.

Barugon:
The problem here seems to be that `Memory` doesn't see that the `TextEdit` has lost focus.

Barugon:
Okay, I found the culprit. If you click on another `DragValue` or `TextEdit` then `Memory::stop_text_input` gets called, which sets the focus ID to `None`. This happens before [this check](https://github.com/emilk/egui/blob/eb0812a9530f77712bf77aad0956f932a3a8f23d/crates/egui/src/widgets/drag_value.rs#L382), so `is_kb_editing` is false, which essentially kills the edit session.

Clicking on other widgets (e.g. `Label`) does not call `TextEdit` then `Memory::stop_text_input`, which allows `Memory::surrender_focus` to be executed correctly, resulting in the correct behavior.

Simply commenting out [this line](https://github.com/emilk/egui/blob/eb0812a9530f77712bf77aad0956f932a3a8f23d/crates/egui/src/memory.rs#L456) fixes the issue but I don't know if there would be any other issues because of it.
@emilk

Barugon:
[...]

PR #2838, opened 2023-03-23
Fix DragValue not saving in all cases where focus can be lost

paul-hansen:
Closes https://github.com/emilk/egui/issues/2818

This fixes a bug that was introduced in https://github.com/emilk/egui/commit/b40dba18c642fa959c4bad59fea99416a31f0825 . That commit added checking focus_lost before applying changes to the DragValue to prevent modifying the value every frame or keypress. The problem was it only checked this inside of a condition that was only true when it still had focus so it was never run unless the focus changed while inside the code block.

Note that I'm a Windows user so I haven't been able to run `/sh/check.sh` in it's entirety. I looked through it and ran the checks that looked most relevant. Any errors I found seemed not related to my code and already in master.

### DIFF
--- a/crates/egui/src/widgets/drag_value.rs
+++ b/crates/egui/src/widgets/drag_value.rs
@@ -475,18 +475,6 @@ impl<'a> Widget for DragValue<'a> {
                     .desired_width(ui.spacing().interact_size.x)
                     .font(text_style),
             );
-            // Only update the value when the user presses enter, or clicks elsewhere. NOT every frame.
-            // See https://github.com/emilk/egui/issues/2687
-            if response.lost_focus() {
-                let parsed_value = match custom_parser {
-                    Some(parser) => parser(&value_text),
-                    None => value_text.parse().ok(),
-                };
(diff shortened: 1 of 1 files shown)


paul-hansen:
Moving this to a draft, still found one case where it isn't saving.

If you have a list of drag values, clicking from one drag value to a drag value that comes after it, does not save.

Barugon:
Any progress on this?

paul-hansen:
None on my end unfortunately. If someone wants to pick this up that would be appreciated!

I thought I had figured out what it was so I made this PR. It has it working for tabbing, but I discovered after creating this PR that it still wasn't working if you click on another drag value that comes after the drag value you are editing.

If a solution isn't found, perhaps it would be worth it to just revert https://github.com/emilk/egui/commit/b40dba18c642fa959c4bad59fea99416a31f0825 so we have something that functions again.

Just for clarity, I'm not a normal egui maintainer or anything, just a user also ran into this and thought I could help. I don't hold any authority here.

Issue #2811, opened 2023-03-13
`ScrollArea` ignores `Spacing::scroll_bar_inner_margin` for vertical scrollbar


Barugon:
**Describe the bug**
I show a `ScrollArea` after setting `Spacing::scroll_bar_inner_margin` to 0.0 and here's the result that I get:

![Screenshot from 2023-03-13 13-10-49](https://user-images.githubusercontent.com/16503728/224822445-aa544619-246d-4d67-869d-49cef7926964.png)

Note that the horizontal scrollbar is correct while the vertical scrollbar still has the default margin.

This behavior started with v0.21.

Barugon:
I don't know what the intent of that code is, so I can't make a fix for it.

PR #2839, opened 2023-03-25
Remove problematic clip code from `ScrollArea`

Barugon:
This code causes incorrect clipping. It also uses incorrect indexing (1 - d instead of just d). With this code removed, I have not been able to find a case where the scroll area contents invade the scroll bar area.

Closes <https://github.com/emilk/egui/issues/2811>.

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -439,20 +439,12 @@ impl ScrollArea {
         {
             // Clip the content, but only when we really need to:
             let clip_rect_margin = ui.visuals().clip_rect_margin;
-            let scroll_bar_inner_margin = ui.spacing().scroll_bar_inner_margin;
             let mut content_clip_rect = ui.clip_rect();
             for d in 0..2 {
                 if has_bar[d] {
                     if state.content_is_too_large[d] {
                         content_clip_rect.min[d] = inner_rect.min[d] - clip_rect_margin;
                         content_clip_rect.max[d] = inner_rect.max[d] + clip_rect_margin;
-                    }
-
-                    if state.show_scroll[d] {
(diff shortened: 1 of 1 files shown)


crates/egui/src/containers/scroll_area.rs
```diff
@@ -439,21 +439,13 @@ impl ScrollArea {
...
                         content_clip_rect.min[d] = inner_rect.min[d] - clip_rect_margin;
                         content_clip_rect.max[d] = inner_rect.max[d] + clip_rect_margin;
                     }
-
-                    if state.show_scroll[d] {
-                        // Make sure content doesn't cover scroll bars
-                        let tiny_gap = 1.0;
-                        content_clip_rect.max[1 - d] =
-                            inner_rect.max[1 - d] + scroll_bar_inner_margin - tiny_gap;
-                    }
```
emilk, line 456: The `1-d` is not incorrect. When we have a horizontal scroll-bar (`d = 0`), the scrollbar itself is at he bottom of the vertical dimension (1-d), and the other way around.

I believe this code was added to handle nested scroll-areas. Have you tried that?

Barugon: I have not. I'll tinker with it and see what I can find.

Barugon:
Closing this for now.

crates/egui/src/containers/scroll_area.rs
Barugon: > I believe this code was added to handle nested scroll-areas. Have you tried that?

Do you have some code to specifically test this?

Barugon: What I mean is something that specifically shows what the change was trying to correct.

PR #2860, opened 2023-03-31
Remove problematic clip code from `ScrollArea`

Barugon:
Closes <https://github.com/emilk/egui/issues/2811>.

This code causes incorrect clipping. It also uses incorrect indexing (1 - d instead of just d). With this code removed, I have not been able to find a case where the scroll area contents invade the scroll bar area.

Because this was asked about in the original PR (#2839), here is a simple project with embedded `ScrollArea`s. I cannot see any difference between this branch and  master.
[scroll_test.zip](https://github.com/emilk/egui/files/11125180/scroll_test.zip)

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -439,20 +439,12 @@ impl ScrollArea {
         {
             // Clip the content, but only when we really need to:
             let clip_rect_margin = ui.visuals().clip_rect_margin;
-            let scroll_bar_inner_margin = ui.spacing().scroll_bar_inner_margin;
             let mut content_clip_rect = ui.clip_rect();
             for d in 0..2 {
                 if has_bar[d] {
                     if state.content_is_too_large[d] {
                         content_clip_rect.min[d] = inner_rect.min[d] - clip_rect_margin;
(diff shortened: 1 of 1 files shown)


Barugon:
Here's a screenshot of a `ComboBox` showing a date and time where I use `UI::set_min_width` in order to keep it from wrapping when the scroll bar appears.

Using the master branch:
Screenshot from 2023-03-31 12-19-10

Using this branch:
Screenshot from 2023-03-31 12-26-10

juancampa:
This code also causes an issue when a ScrollArea is rendered inside a menu.

The content is clipped horizontally to the hard-coded menu width when the scrollbar appears.

Removing the code seems to fix it as well.

https://github.com/emilk/egui/assets/1410520/569ab260-80e1-4a90-80e2-c308d3010978

Barugon:
There are a number of issues that it causes but I haven't discovered what it's supposed to fix.

emilk:
The code was added here: https://github.com/emilk/egui/pull/2665/files

I wish I had added an example or test to justify it… I'm a bit scared of removing it now, but let's try it and see what pops up :)

emilk (approved):

Issue #2829, opened 2023-03-20
Egui almost certainly shouldn't be depending on android-activity


rib:
This dependency on the "native-activity" feature for `android-activity` means that applications can't use Egui with the game-activity backend and it also locks Egui applications to a specific version of android-activity which is probably not necessary.

https://github.com/emilk/egui/blob/59c2c4c70d9e23b673b4cf6fcdfc030a91074d7e/crates/egui-winit/Cargo.toml#L76

Unlike most crates it's strictly never possible to have more than one version of android-activity linked into an application and so ideally very few crates should be directly depending on a specific version (increasing the chance of conflicts).

The android-activity crate is re-exported from Winit so that most crates can avoid needing to specify a direct dependency.

E.g. see:
https://github.com/rust-mobile/rust-android-examples/blob/41e72ca9de45e2add4776ec004f916ad979cf781/agdk-egui/src/lib.rs#L4

I can't see anything depending on the android_activity API though, so hopefully this can be removed.

ardocrat:
I made PR to use android-activity as optional dependency cause their CI is requiring it: https://github.com/emilk/egui/pull/2834

Issue #2720, opened 2023-02-12
Build fails when depending on `winit` with `android-game-activity` feature enabled


metent:
**Describe the bug**
[This example](https://github.com/metent/rust-android-examples/tree/0beec1d0b8c2f3311852b08318d3a7d31a404c84/agdk-eframe) fails to build with latest egui.

**To Reproduce**
1. Clone [this repository](https://github.com/metent/rust-android-examples).
2. Navigate to `agdk-eframe` directory.
3. Follow the instructions in `README.md` to build the example.
4. Build fails with the following error.

```
[2023-02-12T06:45:41Z INFO  cargo_ndk::cli] Using NDK at path: /home/rishabh/Android/Sdk/ndk/25.2.9519653 (ANDROID_NDK_HOME)
[2023-02-12T06:45:41Z INFO  cargo_ndk::cli] NDK API level: 21
[2023-02-12T06:45:41Z INFO  cargo_ndk::cli] Building targets: arm64-v8a
[2023-02-12T06:45:41Z INFO  cargo_ndk::cli] Building arm64-v8a (aarch64-linux-android)
warning: file `/home/rishabh/software/myprogs/rust-android-examples/agdk-eframe/src/lib.rs` found to be present in multiple build targets:
  * `lib` target `main`
  * `bin` target `agdk-eframe`
    Updating git repository `https://github.com/emilk/egui`
   Compiling ecolor v0.21.0 (https://github.com/emilk/egui?branch=master#530e9f66)
   Compiling emath v0.21.0 (https://github.com/emilk/egui?branch=master#530e9f66)
   Compiling android-activity v0.4.0
error[E0252]: the name `activity_impl` is defined multiple times
   --> /home/rishabh/.cargo/registry/src/github.com-1ecc6299db9ec823/android-activity-0.4.0/src/lib.rs:107:5
    |
102 | use native_activity as activity_impl;
    |     -------------------------------- previous import of the module `activity_impl` here
...
107 | use game_activity as activity_impl;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `activity_impl` reimported here
    |
    = note: `activity_impl` must be defined only once in the type namespace of this module
help: you can use `as` to change the binding name of the import
    |
107 | use game_activity as other_activity_impl;
    |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

error: The "game-activity" and "native-activity" features cannot be enabled at the same time
[...]

PR #2863, opened 2023-04-01
egui-winit: Remove `android-activity` dependency + add `Activity` backend features

rib:
Instead of depending on `android-activity` directly, this exposes the `android-native-activity` and `android-game-activity` features from Winit (via `egui-winit` and `eframe`).

The overall issue here is that Android applications will very likely need to implement a subclass of `Activity` in Java or Kotlin and when using `android-activity` that means they will have to choose to subclass `NativeActivity` or `GameActivity`. Since there are significant pros and cons to each one it isn't really appropriate for Egui to force a choice here.

Conceptually the `android-activity` crate is the lowest-level, runtime crate for a native Android application written in Rust and then middleware like Winit and Egui sit on top of this. With the way Cargo features work then if we want make the low-level crate configurable for applications then we don't really have any choice but to expose equivalent features via middleware frameworks (such as Egui) that applications depend on.

This ensures that applications can choose what `android-activity` backend they use while also relying on Winit to decide what version of `android-activity` to use - without increasing the risk of a version conflict by having a direct dependency.

_(NB: Egui doesn't currently use the android-activity API itself)_

Since `android-activity` provides the `android_main()` entry point for Android applications it's not possible to link in multiple versions of the android-activity crate and so it's particularly important to avoid unnecessary direct dependencies that could cause a version conflict in the future.

To help avoid the need for applications to directly depend on `android-activity` the Winit crate re-exports the `android-activity` API and exposes features to configure the backend so that application crates can instead rely on Winit to pull in a compatible version of `android-activity`. (This way version bumps for `android-activity` only need to be synchronized with the Winit crate).

CI now enables the `android-native-activity` feature for testing.

Fixes: #2829
Fixes: #2720
Closes: #2834

Cc: @ardocrat

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -199,7 +199,7 @@ jobs:
       - name: Set up cargo cache
         uses: Swatinem/rust-cache@v2
 
-      - run: cargo check --features wgpu --target aarch64-linux-android
+      - run: cargo check --features wgpu,android-native-activity --target aarch64-linux-android
         working-directory: crates/eframe
 
   # ---------------------------------------------------------------------------
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1272,7 +1272,6 @@ name = "egui-winit"
 version = "0.21.1"
 dependencies = [
  "accesskit_winit",
- "android-activity",
  "arboard",
  "document-features",
  "egui",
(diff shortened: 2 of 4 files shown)


emilk (approved):
Thank you!

PR #2865, opened 2023-04-01
eframe: Fix building for iOS with wgpu

modmuss50:
This PR fixes a compile error when building an eframe app (using wgpu) for iOS due to winit's `run_return` function not being present.

As noted here [/winit/blob/master/src/platform/mod.rs#L12-L14](https://github.com/rust-windowing/winit/blob/master/src/platform/mod.rs#L12-L14)  in winit `run_return` is only available on specific platforms.

This PR contains the minimal changes required to build for IOS, I can envision a few other solutions such as:

- opt-ing in to using `run_and_return` only on the supported platforms
- Removing `native_options.run_and_return` on unsupported platforms
- Applying the same change to `run_glow` even though that appears to be unsupported (glutin fails to compile) on iOS.

I am just getting into rust so very much look forward to your feedback. Thanks in advance 👍 

See the following screenshot of the unmodifed (with wgpu eframe feature) hello world example running on ios:
Screenshot 2023-04-01 at 22 13 14

### DIFF
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -91,6 +91,7 @@ fn create_event_loop_builder(
 ///
 /// We reuse the event-loop so we can support closing and opening an eframe window
 /// multiple times. This is just a limitation of winit.
+#[cfg(not(target_os = "ios"))]
 fn with_event_loop<R>(
     mut native_options: epi::NativeOptions,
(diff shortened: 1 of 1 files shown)


emilk:
Cool to see egui on iOS! What happened with the PR? Did you find some other way forward?

PR #2866, opened 2023-04-02
Fix typos

fezjo:


emilk (approved):
Thank you!

Did you use some automated tool for this? Maybe some we could add to the CI?

Before I saw this PR, I started adding a CI step with a another spell checker, but it didn't find everything you found: https://github.com/emilk/egui/pull/2918

fezjo:
Hi, I used [cspell](https://cspell.org/) (for no particular reason). It had around 300 unique false positives (which is many times more than real positives) when I ran it, so maybe it's not ideal for CI. I could give you the whitelist that I used, but some other tool may be better.

PR #735, opened 2021-09-22
Split out new crate egui-winit from egui_glium

emilk:
This extracts a lot of code from egui_glium into a reusable crate ~`egui_for_winit`~ `egui-winit`. This can be reused for `egui_glow` (https://github.com/emilk/egui/pull/685) and many other integrations.

This replaces the 3rd party [`egui_winit_platform`](https://github.com/hasenbanck/egui_winit_platform) which is used by a few integrations: <https://crates.io/crates/egui_winit_platform/reverse_dependencies>. `egui_for_winit` handles a lot more things, like touch input, screen reader, etc.

~The name `egui_for_winit` is not great, but `egui_winit` [was already taken](https://crates.io/crates/egui-winit). I would welcome a better name suggestion!~

# API

egui_for_winit_docs

PR #2872, opened 2023-04-03
🏓 `egui-winit`: Document `State::on_event` from unconditionally consuming Tab keypresses 

tosti007:
In PR #735, the initial `egui-winit` version was added. That PR contains a check whether the `Tab` key was pressed within the `State::on_event` and returns `consumed = true`, regardless of `egui`'s `wants_keyboard_input`.

Since I could not find any reference or further logic that this is intended behavior and I'd like to handle the `Tab` key, similar to any other key, this PR removes the check.

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -307,6 +307,7 @@ impl State {
             }
             WindowEvent::KeyboardInput { input, .. } => {
                 self.on_keyboard_input(input);
+                // When pressing the Tab key, egui focuses the first focusable element, hence Tab always consumes.
                 let consumed = egui_ctx.wants_keyboard_input()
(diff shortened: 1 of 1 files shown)


tosti007:
I just figured out why this is the case; when no element is focused, pressing `Tab` focuses the first element. Since `egui` has `ctx.wants_keyboard_input() = false` at that stage, the additional check is added. I think this line needs to be better documented to ensure the same mistake isn't made in the future, I'll change this PR to reflect this.

Nonetheless, maybe it'd be a good idea to add a toggle to allow/disallow the initial-tab behavior? Another solution would be to have `egui` return some kind of `true` when this is the case, but I don't think this is possible with the current input handling implementation.

emilk:
I agree that improving the comment here is the better fix, for now

tosti007:
This slipped my mind, but I have now updated the PR to only add a comment.

PR #2879, opened 2023-04-06
Revert "DragValue: when keyboard editing, only update the value on focus lost"

Barugon:
Reverts emilk/egui#2688, which causes  a lot of problems with `DragValue` editing (see #2818).


Barugon:
Actually, a better solution would be to use `Response::changed` so that the value isn't being parsed every frame. I'm closing this PR in order to create a new PR with that change.

Issue #2877, opened 2023-04-06
Keyboard navigation: focus change loses new value in DragValue


tajtiattila:
In the Widget Gallery of Egui Demo, focus the DragValue and type in a new number. Tab or Shift+Tab moves the focus to the next widget, but the new value is lost. The value is kept only when Enter is pressed, but then the focus is lost.

https://user-images.githubusercontent.com/247562/230308477-13e59cd0-e612-49f8-acaf-a7b36c848b23.mp4

This somewhat defeats keyboard navigation with DragValues because one can either navigate or change values, but not both.

**Expected behavior**
DragValue should keep the value entered when Tab is pressed.

Tested on the web demo at egui.rs and eframe on Windows using egui version 0.21.3-28-geb0812a9 (2023-03-31).

There seems to be a related issue: the internal state of the TextEdit/DragValue (text cursor position, selection) seems to be remembered when navigating away with Tab, and is reset only when focusing the widget with the mouse (which selects the whole number). Perhaps focusing the DragValue with Tab should work like focusing with the mouse.

harrisonmg:
relevant: #2880

YgorSouza:
The code that decides whether the value should be updated is this:

https://github.com/emilk/egui/blob/d77c44657206d6a5a9eda0d3b10fac6e189acc27/crates/egui/src/widgets/drag_value.rs#L494-L500

The code is conditional on the `is_kb_editing` variable defined here:

https://github.com/emilk/egui/blob/d77c44657206d6a5a9eda0d3b10fac6e189acc27/crates/egui/src/widgets/drag_value.rs#L389-L400

It seems that the problem is that on the frame where the TextEdit would lose focus after pressing tab, `is_kb_editing` is already set to false, so we never get the `lost_focus` condition to update the value. We get it with Enter because in that case the TextEdit itself calls surrender_focus on the same frame:

https://github.com/emilk/egui/blob/d77c44657206d6a5a9eda0d3b10fac6e189acc27/crates/egui/src/widgets/text_edit/builder.rs#L967-L981

Given that, I wonder what would be the best way to fix it. Maybe we could change the logic for `is_kb_editing` to render an extra frame after focus is lost, but that would mess with accessibility, according to the comment. Another option would be to add a `pub(crate)` method to `TextEdit` that would make it capture the Tab key and surrender focus manually like it does with Enter, but that seems a little convoluted.

Issue #2142, opened 2022-10-12
lost_focus on a single_text_edit widget does not get called in several cases


Yvo68:
Has been tested in master branch.

**Describe the bug**
lost_focus gets correctly called when clicking in an area without widget or when pressing ENTER but

lost_focus does not get called in the following cases:
1. when focus is given to another widget (e.g.  clicking on another single_text_edit)
2. when tab is pressed (focus is given to the first widget of the panel while you would expect the next widget to get focus and the current widget to lose focus 

Case 1:
Click on a single_text_edit to get focus
Click on a second single_text_edit --> you get focus on the second widget without lost_focus getting called on the first one

Case 2:
Click on a single_text_edit to get focus
Press tab --> the first widget gets focus without lost_focus being called on our focused widget
Press tab until you reach the correct widget --> the lost_focus gets correctly called when leaving the considered widget

**Expected behavior**
a) lost_focus should be called when focus is given to another widget
b) Pressing TAB for the first time should move to the widget just after the currently widget under focus (and call lost_focus)

**Desktop **
 - OS: MACOS
 - Browser Chrome
 - Version 105.0.5195.125

Yvo68:
Sorry (my bad), seems the case 2 ("TAB")  has been fixed between 0.19.0 and master branch.
The bug on case 1 is still present in master branch.

phire:
I've run into this myself (case 1), and it's still present on the latest master.

Digging a bit deeper, the problem is actually dependant on the ordering. If you click from one TextEdit field to another created later, then the `lost_focus()` event is missed. But if you click to one created previously, then lost_focus()` works as expected.

willbicks:
Can confirm that the case 1 of clicking another widget not firing `lost_focus` is still an issue in v0.30.0. 

I ended up working around it by checking `if widget.lost_focus() || widget.clicked_elsewhere()`.

tectin0:
Can confirm that the case 1 of clicking another widget not firing `lost_focus` is still an issue in v0.33.3

The "workaround" of calling `clicked_elsewhere()` in no way actually fixes the issue though

The `lost_focus` event should specifically be called when a widget had focus and then lost it. With this "workaround" it will just fire anytime you click anywhere in the GUI (except on the widget) which is just entirely different behavior and possibly harmful depending on what your code does when the widget loses focus

[...]

PR #2880, opened 2023-04-06
Fix `DragValue` editing

### LINKED ISSUE #2818: `DragValue` edit is discarded when clicked off

Barugon:
Closes #2818.

Use `Response::changed` instead of `Response::lost_focus` (which is problematic).

### DIFF
--- a/crates/egui/src/widgets/drag_value.rs
+++ b/crates/egui/src/widgets/drag_value.rs
@@ -475,7 +475,7 @@ impl<'a> Widget for DragValue<'a> {
                     .desired_width(ui.spacing().interact_size.x)
                     .font(text_style),
             );
-            // Only update the value when the user presses enter, or clicks elsewhere. NOT every frame.
+            // Only update when the edit content has changed. NOT every frame.
             // See https://github.com/emilk/egui/issues/2687
             if response.changed() {
                 let parsed_value = match custom_parser {
--- a/crates/egui/src/widgets/drag_value.rs
+++ b/crates/egui/src/widgets/drag_value.rs
@@ -477,7 +477,7 @@ impl<'a> Widget for DragValue<'a> {
             );
             // Only update the value when the user presses enter, or clicks elsewhere. NOT every frame.
             // See https://github.com/emilk/egui/issues/2687
-            if response.lost_focus() {
+            if response.changed() {
                 let parsed_value = match custom_parser {
                     Some(parser) => parser(&value_text),
                     None => value_text.parse().ok(),


paul-hansen (approved):
This does mean that it will change the value as you are typing now, which essentially reverts what #2687 wanted. Pinging @harrisonmg to give them a chance to comment. 

I would prefer the way this PR has it anyway. Updating the value on every keypress allows you to see the change happening in real time as you type for quickly previewing small adjustments, otherwise you have to press enter and re-focus it every time you want to preview the change. Think about positioning objects precisely in a 3d scene. So I think this is better anyway.

You can always make your own temporary variable and only update it when focus is lost to recreate the functionality #2687 wanted:
```rust
// see the hello_world example for the where self.age would come from
let mut temp_value =
    ui.data_mut(|data| *data.get_temp_mut_or(Id::new("my_temp"), self.age));
let response = ui.add(egui::DragValue::new(&mut temp_value));
if response.changed() {
    ui.data_mut(|data| data.insert_temp(Id::new("my_temp"), temp_value));
}
if response.lost_focus() {
    self.age = temp_value;
}
```
Obviously would want to wrap that in a function or your own widget if you use it a lot.

tldr; I approve of this PR's change, but it should be noted that it is changing the functionality back closer to what it was, but I think it's for the better anyway.

Barugon:
> This does mean that it will change the value as you are typing now

Yes, it's kind of a middle ground. Previously, it would parse the text every frame. With this change, it would only do that work when the `TextEdit` changes. After looking extensively at the focus logic, it's not possible to use `lost_focus` as there are too many issues with it. The focus logic needs a lot of work to make it reliable enough before using `lost_focus`.

Barugon:
I prefer it updating while editing the value also.

harrisonmg:
In my case, I'm updating motor controller tunings, so this behavior presents a bit of an issue for that.

Seems like an option might be nice in the future.

paul-hansen:
It sounds like the problem you want to solve is that updating your motor controller tunings too frequently causes problems with the motor. Personally I wouldn't see that as a problem that a UI library should be making changes to try to solve. Instead, you should use something like a debounce function when setting the motor's values from anywhere to prevent it from being updated more often than it can support. This would be more robust and would prevent things like users still being able to focus and unfocus inputs more quickly than is supported.

That said, if there's a way to add an option to update the value only on focused lost I wouldn't be opposed to it. It needs to work correctly when tabbing and unfocusing by clicking on other dragvalues/widgets, which is hard to do because of how DragValue changes it's widget based on if it's focused or not, which causes checks using lost_focus in DragValue's internal logic to be inconsistent.

Barugon:
> In my case, I'm updating motor controller tunings, so this behavior presents a bit of an issue for that.
> 
> Seems like an option might be nice in the future.

Is it possible, in your code, to update on `lost_focus`?

harrisonmg:
> > In my case, I'm updating motor controller tunings, so this behavior presents a bit of an issue for that.
> > 
> > Seems like an option might be nice in the future.
> 
> Is it possible, in your code, to update on `lost_focus`?

Maybe, I'll probably stick to an older egui version for now.

Specifically, my issue is that typing "100" results in the gain being set to 1, then 10, then 100. It's the order of magnitude changes, not the frequency.

Anyway, I do support this change especially if it improves consistency.

emilk:
I think we should add an option for this, so users can control what behavior they want. I definitely see the use in both

harrisonmg:
> it's not possible to use lost_focus as there are too many issues with it

@Barugon could you give a few examples of when you found that this fails? I think it would be helpful to drive future improvements to the focus logic if need be.

Barugon:
> > it's not possible to use lost_focus as there are too many issues with it
> 
> @Barugon could you give a few examples of when you found that this fails? I think it would be helpful to drive future improvements to the focus logic if need be.

Edit the value and then click on another `DragValue` or press tab and your edits are lost.

Barugon:
There's a video in the issue linked above that show the issue.

harrisonmg:
Ok, so nothing beyond what's described in #2142?

If so, seems like this option + fixing 2142 would be perfect for everyone

Barugon:
> Ok, so nothing beyond what's described in #2142?
> 
> If so, seems like this option + fixing 2142 would be perfect for everyone

Loosing your edit due to tabbing or clicking on another `DragValue` (or `TextEdit`) are actually two separate issues that cause the same problem.

Barugon:
In #2818 I found the culprit for clicking off but it doesn't fix tabbing.

tajtiattila:
@Barugon This seems to fix the main issue in #2877, thank you!

Barugon:
> @Barugon This seems to fix the main issue in #2877, thank you!

Unfortunately hasn't been merged though.

Trivaxy:
Would be nice to see this merged. Some of my users complained that the way it is now is sort of unintuitive and annoying, especially since we're doing rapid data entry on many different `DragValue`s where it's a cycle of "type into box -> click on the next one -> repeat"

Barugon:
@emilk 
Are there more changes you require?

crates/egui/src/widgets/drag_value.rs
```diff
@@ -352,6 +354,12 @@ impl<'a> DragValue<'a> {
         }
         .custom_parser(|s| i64::from_str_radix(s, 16).map(|n| n as f64).ok())
     }
+
+    /// Update the value when there are changes while editing. (Default: true)
```
emilk, line 358:
```suggestion
    /// Update the value on each key press when text-editing the value.
    ///
    /// Default: `true`.
    /// If `false`, the value will only be updated when user presses enter or deselects the value.
```

emilk (approved):

emilk:
Thanks!

PR #2881, opened 2023-04-06
Add option to customize progress bar rounding

YgorSouza:
The minimum filled width is adapted to the radius and the animation is disabled so it renders correctly.

### DIFF
--- a/crates/egui/src/widgets/progress_bar.rs
+++ b/crates/egui/src/widgets/progress_bar.rs
@@ -16,6 +16,7 @@ pub struct ProgressBar {
     text: Option<ProgressBarText>,
     fill: Option<Color32>,
     animate: bool,
+    rounding: Option<Rounding>,
 }
 
 impl ProgressBar {
@@ -28,6 +29,7 @@ impl ProgressBar {
             text: None,
             fill: None,
             animate: false,
+            rounding: None,
         }
     }
 
@@ -74,6 +76,14 @@ impl ProgressBar {
         self.animate = animate;
         self
     }
+
+    /// Set the rounding of the progress bar.
+    /// If this option is set, the [`Self::animate`] option is ignored, since the loading animation
+    /// requires a perfect circle to render correctly.
+    pub fn rounding(mut self, rounding: impl Into<Rounding>) -> Self {
+        self.rounding = Some(rounding.into());
+        self
+    }
 }
 
 impl Widget for ProgressBar {
@@ -85,6 +95,7 @@ impl Widget for ProgressBar {
             text,
             fill,
             animate,
+            rounding,
         } = self;
 
         let animate = animate && progress < 1.0;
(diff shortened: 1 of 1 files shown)


YgorSouza:
To test, add e.g., `.rounding(2.0)` here:

https://github.com/emilk/egui/blob/eb0812a9530f77712bf77aad0956f932a3a8f23d/crates/egui_demo_lib/src/demo/widget_gallery.rs#L196-L198

Should I add this somewhere in the demo?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
 * Add `char_limit` to `TextEdit` singleline mode to limit the amount of characters
 * ⚠️ BREAKING: `Plot::link_axis` and `Plot::link_cursor` now take the name of the group ([#2410](https://github.com/emilk/egui/pull/2410)).
 * Update `Plot::allow_zoom` and `Plot::allow_drag` to allow setting those values for X and Y axes independently ([#2901](https://github.com/emilk/egui/pull/2901)).
+* Add `ProgressBar::rounding` to enable progress bars with sharp corners ([#2881](https://github.com/emilk/egui/pull/2881)).
```
emilk, line 11:
```suggestion
```

I'll generate a changelog on release instead with `scripts/generate_changelog.py`

YgorSouza:
I'm thinking that instead of setting an explicit rounding value, it could instead be a boolean that makes it use the same rounding as the other widgets, like buttons and textboxes, so it is easier to keep the UI consistent. Does that make sense? I also have to fix some visual bugs at low fill percentages before this is ready to merge.

emilk (commented):
It would be nice with a image or gif for this!

YgorSouza:
> It would be nice with a image or gif for this!

[Screencast from 2023-11-10 23-25-49.webm](https://github.com/emilk/egui/assets/43298013/16ff990d-f3a4-44a9-992d-fc7596ec1406)

crates/egui/src/widgets/progress_bar.rs
```diff
@@ -95,16 +112,15 @@ impl Widget for ProgressBar {
...
             ui.painter()
                 .rect(outer_rect, rounding, visuals.extreme_bg_color, Stroke::NONE);
-            let inner_rect = Rect::from_min_size(
-                outer_rect.min,
-                vec2(
-                    (outer_rect.width() * progress).at_least(outer_rect.height()),
-                    outer_rect.height(),
-                ),
-            );
+            let min_width = 2.0 * rounding.sw.at_least(rounding.nw).at_most(corner_radius);
```
emilk, line 120: why only look at `sw` and `nw` here? Wouldn't it make sense to use something like `rounding.max()` instead?

crates/egui/src/widgets/progress_bar.rs
```diff
@@ -64,10 +66,24 @@ impl ProgressBar {
...
         self.animate = animate;
         self
     }
+
+    /// Set the rounding of the progress bar.
+    ///
+    /// If [`Self::rounding`] and [`Self::animate`] are used simultaneously, the animation is not
+    /// rendered, since it requires a perfect circle to render correctly. However, the UI is still
+    /// redrawn.
+    pub fn rounding(mut self, rounding: impl Into<Rounding>) -> Self {
```
emilk, line 83:
```suggestion
    #[inline]
    pub fn rounding(mut self, rounding: impl Into<Rounding>) -> Self {
```

crates/egui/src/widgets/progress_bar.rs
```diff
@@ -64,10 +66,24 @@ impl ProgressBar {
     /// Whether to display a loading animation when progress `< 1`.
     /// Note that this will cause the UI to be redrawn.
     /// Defaults to `false`.
+    ///
+    /// If [`Self::rounding`] and [`Self::animate`] are used simultaneously, the animation is not
+    /// rendered, since it requires a perfect circle to render correctly. However, the UI is still
+    /// redrawn.
     pub fn animate(mut self, animate: bool) -> Self {
```
emilk, line 73:
```suggestion
    #[inline]
    pub fn animate(mut self, animate: bool) -> Self {
```

emilk (approved):

crates/egui/src/widgets/progress_bar.rs
YgorSouza: Since this `min_width` is meant to make the rounding look right when the progress bar is empty, I figured only the left side mattered. But maybe it would be better to just use `max()`. I'll do a quick test and commit it.

YgorSouza: So there is no method `Rounding::max` currently. Should I add one?

It doesn't seem to be necessary, though. The problem this min_width avoids is that the filled rect is flattened when it is too small, and then it ends up appearing outside the background rect, like this:

image

Actually, if there was a way to avoid this overflow without setting a minimum width, I think a lot of people would prefer that (see for example [#3999](https://github.com/emilk/egui/discussions/3399)). But I don't know if it's possible.

PR #2884, opened 2023-04-07
wgpu 0.16

niklaskorz:
wgpu 0.16 has been released today: https://github.com/gfx-rs/wgpu/releases/tag/v0.16.0

Not much to do actually, just the type changes of `NonZeroU32` to `u32` and the replacement of `TextureFormat::describe` with `is_srgb`.


cwfitzgerald:
I don't think the right action here is to get us to downgrade, rather to get ash to upgrade. It was reverted due to Firefox not bring able to use windows-rs

emilk:
`libloading` seems like a pretty small dependency, so we can whitelist it in `deny.toml`.

I wonder how it wasn't caught by the `cargo deny` check in the `wgpu` CI though 🤔

Wumpf (approved):
since emil is good with libloading dupl we're good to go, changes look all straight forward. Tested web and after the fix I pushed yesterday stuff still works.
Just waiting for ci to finish, there was some hickup

deny.toml
```diff
@@ -46,6 +46,7 @@ skip = [
     { name = "windows_x86_64_msvc" }, # old version via glutin
     { name = "windows-sys" },         # old version via glutin
     { name = "windows" },             # old version via accesskit
+    { name = "libloading" },          # wgpu-hal itself depends on 0.8 while some of its dependencies, like ash and d3d12, depend on 0.7
```
emilk, line 49: please sort 🙏

emilk:
🥳

PR #2886, opened 2023-04-08
Menu State rect now uses menu frame rect instead of contents rect.

### LINKED ISSUE #2862: Menu buttons do not work near the edge of screen
When a context menu is in contact with the edge of the screen, the inner buttons won't work and it will act like if the user clicked outside.

This was reproduced with 0.21.0 and master.

https://user-images.githubusercontent.com/124705253/229283905-02b4243b-f60a-4bfa-ac87-9e0e5e1d358d.mp4

Tested on Arch Linux under both Wayland(KDE, sway) and X11(dwm).

For context I use a 3700x and an RTX3080.

hats-np:
When a menu is opened near the edge of screen, the rect of the `add_contents` response is always `> cursor_pos` as expected during regular function, but not when the area moves itself to accommodate the screen edges.

As a solution I made the menu_state rect be the same as the menu frame, which fixes the issue. I am unsure if there is a better solution, such as making the `add_contents` aware that their parent has been moved.

I also removed a redundant call that set the menu_state rect a second time unnecessarily as it had already been set during  `MenuState::show`.

Closes #2862 .

### DIFF
--- a/crates/egui/src/menu.rs
+++ b/crates/egui/src/menu.rs
@@ -147,21 +147,22 @@ pub(crate) fn menu_ui<'c, R>(
         .fixed_pos(pos)
         .interactable(true)
         .drag_bounds(ctx.screen_rect());
-    let inner_response = area.show(ctx, |ui| {
+
+    area.show(ctx, |ui| {
         set_menu_style(ui.style_mut());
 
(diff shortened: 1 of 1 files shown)


emilk (approved):
Thanks!

PR #2891, opened 2023-04-10
egui_extras: add `load_texture()` for `RetainedImage`

hinto-janai:
Since `RetainedImage`'s are lazily loaded into textures, the GUI thread ends up being the one doing the work upon `.show()`.

With many images or high resolution ones (or both...!), it freezes the GUI for a few seconds.

This function allows some other thread to load them into textures.

The code is just `RetainedImage::texture_id()` but without the return value.


hinto-janai:
This ends up locking `Context` (freezes GUI) in high-contention situations anyway. I'll submit a PR with a better way to load textures in bulk.

Issue #1164, opened 2022-01-26
Custom zoom configuration


SwayStar123:
**Is your feature request related to a problem? Please describe.**

Currently, the ctrl + scroll zooms in on both x and y axis at the same time, when trying to see granular candlestick data this is not ideal.

**Describe the solution you'd like**

It would be nice if I could add my own custom zoom logic, like regular scroll to zoom on the x axis, ctrl scroll to zoom on the y axis, etc.

**Describe alternatives you've considered**

Ive tried setting the bar width at a higher number, so that zooming would be unnecessary but its impossible to find a zoom level that works for every need

**Additional context**

![image](https://user-images.githubusercontent.com/46050679/151140924-b7ec1ebc-906d-41ca-a3ba-9590391d9ca8.png)
Pretty much impossible to make any good use out of this chart with the current y axis zoom, only option is to zoom in on both axes and only view a small segment of the chart.

SwayStar123:
its not as important as a seperate x & y zoom but it would be nice if I could add a maximum and a minimum zoom level aswell

calbaker:
Yep, I'd benefit hugely from being able to make check boxes for `lock x zoom` and/or `lock y zoom` so that I could zoom in one axis only.  It'd also be nice to be able to specify ranges of x- and y-data to plot.

g-tejas:
This would be of great help to me too! Any ideas anyone?

IFcoltransG:
For my use case, I'd simply like to change zoom so it happens on regular scrolling, rather than ctrl + scrolling. A convenience method to zoom by a certain amount per each direction (around some centre point) would help me immensely, and would allow everyone to hook up whatever zooming behaviour they need. Sort of like the `translate_bounds` but for scale. Granted, they'd still need process the scroll deltas and mouse position themselves, but it would still be a lot friendlier than messing with `set_plot_bounds`.
A possible API could be `pub fn scale_bounds(&mut self, scale: Vec2, centre: Pos2)` where the components of the `scale` vector represent the amounts to scale `x` and `y` by. Or alternatively two separate functions `scale_x/y_bounds(&mut self, scale: f64, centre: f64)`.

YgorSouza:
[...]

PR #2901, opened 2023-04-14
Plot widget - allow disabling zoom and drag for x and y separately

OmegaJak:
Currently, `allow_zoom` and `allow_drag` apply to both axes at the same time. This PR makes it possible to specify whether zooming and dragging is allowed per axis. This helps in a scenario where the range of one axis is fixed for the whole range-- like a sinusoid. In that case, it's nice if zooming and dragging is only possible on the X axis.

This partially addresses <https://github.com/emilk/egui/issues/1164>, but isn't exactly what was asked for there.

[Egui Plot zoom drag demo.webm](https://user-images.githubusercontent.com/4239887/231917584-fca1017f-1006-4040-ac1b-20ec3ddda01f.webm)


crates/egui/src/widgets/plot/mod.rs
```diff
@@ -402,9 +402,19 @@ impl Plot {
         self
     }
 
-    /// Whether to allow zooming in the plot. Default: `true`.
-    pub fn allow_zoom(mut self, on: bool) -> Self {
```
OmegaJak, line 406: It would be very easy to maintain backwards compatibility here by keeping these methods and having them set both x and y at the same time.

Question to the reviewers: what's the preference - keep backwards compatibility here or go with a more minimal API?

OmegaJak:
The referenced issue #1164 wasn't _exactly_ asking for this - should I open a separate issue asking for this exactly?

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -894,9 +910,16 @@ impl Plot {
         }
 
         // Dragging
-        if allow_drag && response.dragged_by(PointerButton::Primary) {
+        if (allow_drag.x || allow_drag.y) && response.dragged_by(PointerButton::Primary) {
```
emilk, line 913:
```suggestion
        if allow_drag.any() && response.dragged_by(PointerButton::Primary) {
```

crates/egui/src/widgets/plot/mod.rs
```diff
@@ -953,12 +976,18 @@ impl Plot {
         }
 
         if let Some(hover_pos) = response.hover_pos() {
-            if allow_zoom {
-                let zoom_factor = if data_aspect.is_some() {
+            if allow_zoom.x || allow_zoom.y {
```
emilk, line 979:
```suggestion
            if allow_zoom.any() {
```

crates/egui/src/widgets/plot/mod.rs
emilk: I vote keeping the old one. Not just for backwards compatibility, but also ergonomics: in most cases, you want to control both the same

emilk (changes requested):
Looks good, but please reintroduce the old functions as a simplified interface

crates/egui/src/widgets/plot/mod.rs
OmegaJak: Oh nice, I hadn't noticed that :+1:

crates/egui/src/widgets/plot/mod.rs
OmegaJak: Yeah, good point. It occurs to me now that having it be generic in `T` `where T : Into<AxisBools>` would produce the best of both worlds. That way there's only the one method needed, and it maintains the current ergonomics for setting both at once, while allowing the new flexibility.

The main cost I see there is that it requires exposing AxisBools, which has been private until now. I guess it will also be a little more verbose to disable just one of the axes, as you'll have to do something like `.allow_zoom(AxisBools::new(false, true))` (unless we keep the `allow_zoom_x/y` methods, which doesn't seem worth it). I'll go ahead and update with the new signature, and probably drop the x/y-specific methods.

emilk (approved):

PR #2905, opened 2023-04-16
Allow for requesting the user's attention to the window

TicClick:
- [ ] should be rebased onto `master once/if #2900 is merged

calls [`request_user_attention()`](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.request_user_attention) with different platform-dependent effects, as seen in [`UserAttentionType`'s documentation](https://docs.rs/winit/latest/winit/window/enum.UserAttentionType.html). the attention type enum mirrors that of winit's but I'm open to other options if this is not how it should be done.

<details>
<summary>example action (windows)</summary>

source: https://github.com/TicClick/egui-attention-tester

https://user-images.githubusercontent.com/4686101/232264061-15ca06e0-b566-4ce0-bbc9-78bf709e6448.mp4

</details>


emilk:
https://github.com/emilk/egui/pull/2900 has been merged - please rebase on it 🙏

TicClick:
all done, thanks for the quick response

crates/eframe/src/epi.rs
```diff
@@ -1145,6 +1159,10 @@ pub(crate) mod backend {
         #[cfg(not(target_arch = "wasm32"))]
         pub focus: Option<bool>,
 
+        /// Set to request a user's attention to the native window. Setting this to `None` resets the attention request.
```
emilk, line 1162: This docstring is wrong I think. `None` means "keep the current attention", afaict

```suggestion
        /// Set to request a user's attention to the native window.
```

emilk (commented):
Looks good - it would be nice with a test of this in `egui_demo_app`:s `backend_panel.rs` - e.g. a button to request attention after a two-second delay (so the user can alt-tab away and test it properly)

TicClick:
> Looks good - it would be nice with a test of this in `egui_demo_app`:s `backend_panel.rs` - e.g. a button to request attention after a two-second delay (so the user can alt-tab away and test it properly)

can do that in the present PR, or file a new one -- what do you prefer? I have a [tester app](https://github.com/TicClick/egui-attention-tester) already so it won't be too much work

emilk:
A separate app to test this would make sense! Please add it to `examples` 🙏

examples/user_attention/src/main.rs
```diff
@@ -0,0 +1,128 @@
+use eframe::egui::{Button, CentralPanel, Context, UserAttentionType};
+use eframe::{CreationContext, NativeOptions};
+
+use chrono::{DateTime, Duration, Utc};
```
emilk, line 4: using `chrono` is way overkill - you can just use `std::time::SystemTime::now()` and `std::time::Duration`.

(`chrono` has a dependency on an old, broken, version of the `time` crate, so it is more than just extra compile time)

examples/user_attention/src/main.rs
```diff
@@ -0,0 +1,128 @@
...
+                                Self::attention_request_timeout().num_seconds()
+                            ),
+                            false => match self.reset_at {
+                                None => "Unfocus the window, fast!".to_owned(),
+                                Some(t) => format!(
+                                    "Resetting attention in {} s...",
+                                    (t - Utc::now()).num_seconds()
+                                ),
+                            },
+                        }),
```
emilk, line 93: `match` on bool is more usually (and more tersely) done with `if bool { } else { }`

examples/user_attention/src/main.rs
```diff
@@ -0,0 +1,128 @@
...
+    let native_options = NativeOptions {
+        initial_window_size: Some(eframe::egui::vec2(400., 200.)),
+        ..Default::default()
+    };
+    eframe::run_native(
+        "User attention test",
+        native_options,
+        Box::new(|cc| Box::new(Application::new(cc))),
+    )
+    .unwrap();
```
emilk, line 127:
```suggestion
fn main() -> eframe::Result<()> {
    let native_options = NativeOptions {
        initial_window_size: Some(eframe::egui::vec2(400., 200.)),
        ..Default::default()
    };
    eframe::run_native(
        "User attention test",
        native_options,
        Box::new(|cc| Box::new(Application::new(cc))),
    )
```

examples/user_attention/src/main.rs
emilk: it would also enhance readability if the match was assigned to a temporary `let button_text = if enabled { … } else { … };`

emilk (approved):
Good example, and works like a charm, but please get rid of the `chrono` dependency 🙏

emilk:
Thank you!

Issue #2912, opened 2023-04-17
egui_winit::State::new should not depend on &EventLoopWindowTarget just for wayland clipboard init


lunixbochs:
I cannot construct a `egui_winit::State` at all on egui HEAD, as of #2765.

I use egui in an actor model system I built that allows me to treat winit windows as actors. This works is great, it allows me to create and use multiple windows, and multiple instances of a single window type. I can also control these windows from any thread in my app. To make this work, I can't pass the `EventLoopWindowTarget` into an actor's window constructor, because my winit user_event references the message types for all of the actor types. No actor's constructor actually knows the concrete type of the `EventLoopWindowTarget` UserEvent, because I don't have a circular type dependency from each actor impl -> all actor definitions (I don't know if that's even possible).

I'm currently constructing `egui_winit::State` myself, without access to an `EventLoopWindowTarget`. I use my own `ActorWindowTarget` type, which implements `HasRawDisplayHandle`.

I also don't use egui_winit's clipboard support at all (I have my own clipboard backend that is constructed and managed entirely outside of the egui state).

I think a good solution is to switch from requiring an `EventLoopWindowTarget` handle, to just `impl HasRawDisplayHandle` (which is implemented by `EventLoopWindowTarget`), if all you're looking for is a wayland display handle.

bu5hm4nn:
I believe this can be closed with the merge of #2914

PR #2914, opened 2023-04-17
use RawDisplayHandle for smithay clipboard init

lunixbochs:
I believe this addresses #2912

### DIFF
--- a/crates/egui-winit/Cargo.toml
+++ b/crates/egui-winit/Cargo.toml
@@ -74,6 +74,7 @@ webbrowser = { version = "0.8.3", optional = true }
 
 [target.'cfg(not(target_arch="wasm32"))'.dependencies]
 instant = { version = "0.1" }
+raw-window-handle = "0.5.0"
 
 [target.'cfg(target_arch="wasm32")'.dependencies]
 instant = { version = "0.1", features = [
--- a/crates/egui-winit/src/clipboard.rs
+++ b/crates/egui-winit/src/clipboard.rs
@@ -1,4 +1,4 @@
-use winit::event_loop::EventLoopWindowTarget;
+use raw_window_handle::HasRawDisplayHandle;
 
 /// Handles interfacing with the OS clipboard.
 ///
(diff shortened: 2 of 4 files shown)


crates/egui-winit/src/clipboard.rs
```diff
@@ -30,7 +30,7 @@ impl Clipboard {
     /// # Safety
     ///
     /// The returned `Clipboard` must not outlive the input `_event_loop`.
-    pub fn new<T>(_event_loop: &EventLoopWindowTarget<T>) -> Self {
+    pub fn new(_display_target: impl HasRawDisplayHandle) -> Self {
```
Mingun, line 33: Doc comment should be updated to refer to the new parameter name

crates/egui-winit/src/lib.rs
```diff
@@ -88,7 +88,7 @@ impl State {
     /// # Safety
     ///
     /// The returned `State` must not outlive the input `_event_loop`.
-    pub fn new<T>(event_loop: &EventLoopWindowTarget<T>) -> Self {
+    pub fn new(display_target: impl HasRawDisplayHandle) -> Self {
```
Mingun, line 91: And here

crates/egui-winit/src/clipboard.rs
```diff
@@ -29,8 +29,8 @@ impl Clipboard {
     ///
     /// # Safety
     ///
-    /// The returned `Clipboard` must not outlive the input `_event_loop`.
-    pub fn new<T>(_event_loop: &EventLoopWindowTarget<T>) -> Self {
+    /// The returned `Clipboard` must not outlive the input `_display_target`.
+    pub fn new(_display_target: impl HasRawDisplayHandle) -> Self {
```
emilk, line 33: could use `&dyn HasRawDisplayHandle` here and elsewhere to avoid monomorphization

crates/egui-winit/Cargo.toml
```diff
@@ -61,6 +61,7 @@ puffin = { version = "0.14", optional = true }
 serde = { version = "1.0", optional = true, features = ["derive"] }
 
 webbrowser = { version = "0.8.3", optional = true }
+raw-window-handle = "0.5.0"
```
emilk, line 64: Move under the native dependency list below (`[target.'cfg(not(target_arch="wasm32"))'.dependencies]`)

emilk (changes requested):
Looks good, but need `cargo clippy`

lunixbochs:
addressed the inline comments

emilk:
tip: configure your editor to run `cargo fmt` on save

emilk (approved):

emilk:
thanks!

Issue #2674, opened 2023-02-04
Context::request_repaint_after doesn't repaint unless the mouse is moving


ShayBox:
I need to display constantly changing information, I tried `Context::request_repaint_after` but the labels only update when the mouse is moving, inside or outside the window, focused or unfocused.

emilk:
Using `eframe`? On what OS?

ShayBox:
I'm using eframe, the eframe template, on the desktop (windows)

ShayBox:
Here's a video of the problem, this is a big problem for my program, it displays constantly changing information but if the screen doesn't change when the mouse isn't moving or the mouse isn't over the window when it's unfocused, the program will be useless.

https://user-images.githubusercontent.com/9505196/221392892-33febdfc-6253-40bb-ae61-4e77d94d6600.mp4

PR #2939, opened 2023-04-20
Fix: `request_repaint_after` works even when called from background thread

emilk:
Closes https://github.com/emilk/egui/issues/2674

The callback installed with `Context::set_request_repaint_callback` is now given a `RequestRepaintInfo` which contains an `after: std::time::Duration` specifying how long to wait until repainting. It is up to integration to heed this.

With `eframe`, both web and native now supports requesting repaints, with delay, from background threads and tasks.


dnsco:
@emilk is there any way you could cut a release with this fix please? We're having issues with our build infra + git dependencies so it would be very helpful if we could just point at a released version.

Thanks for all your work on this awesome library!

PR #2945, opened 2023-04-21
Make egui work on WebGPU out of the box.

### LINKED ISSUE #2137: Eframe : Support for experimental webGPU backend
Hi,
Recently quite a lot of development has been made on the webGPU side of things on egui (yay!), but right now, only the webgl2 backend seems supported on the web. I know this is probably very early, but I would be interested in getting compute shaders in the pipeline, which can only be done with webGPU. There is a great introduction to webGPU compute shaders [here](https://web.dev/gpu-compute/).
I spent some time trying to get a working webGPU backend (diregarding any compute pipeline for now) with Eframe but couldn't manage to get it to work. 

First of all, is this something that would be relevant for Egui ? And has anybody tried something similar with any success? (Maybe @Wumpf which worked on #2107 ?)
Best,
Philippe

Wumpf:
**BREAKING:** Makes WebGL for wgpu opt-in. By default wasm builds for egui will now use WebGPU. To use WebGL, you need to add `wgpu = { version = "0.16.0", features = ["webgl"] }` now to your `Cargo.toml`.

Works around a surface creation issue which made any subsequent call to any method on the surface fail (in a rather arcane manner) - see https://github.com/gfx-rs/wgpu/issues/3710

Regarding the breaking change: Wgpu doesn't allow both WebGPU and a "direct" backend, like WebGL, to live side by side (i.e. wgpu is either in a "wrap WebGPU" or "implement WebGPU" mode). Therefore, egui now inherits the webgl-opt-in behavior. Since Web is memory constraint and availability of WebGPU can be queried easily, two separate builds for WebGPU and WebGL seem to be desirable regardless, so I don't think this is a big issue.

Fixes #2137

### DIFF
--- a/crates/eframe/CHANGELOG.md
+++ b/crates/eframe/CHANGELOG.md
@@ -12,6 +12,7 @@ NOTE: [`egui-winit`](../egui-winit/CHANGELOG.md), [`egui_glium`](../egui_glium/C
 
 #### Web:
 * Bug fix: modifiers keys getting stuck on alt-tab ([#2857](https://github.com/emilk/egui/pull/2857)).
+* ⚠️ BREAKING: Makes WebGL for wgpu opt-in. By default wasm builds for egui will now use WebGPU. To use WebGL, you need to add `wgpu = { version = "0.16.0", features = ["webgl"] }` now to your `Cargo.toml`. ([#2945](https://github.com/emilk/egui/pull/2945))
 
 ## 0.21.3 - 2023-02-15
 * Fix typing the letter 'P' on web ([#2740](https://github.com/emilk/egui/pull/2740)).
--- a/crates/eframe/CHANGELOG.md
+++ b/crates/eframe/CHANGELOG.md
@@ -12,7 +12,7 @@ NOTE: [`egui-winit`](../egui-winit/CHANGELOG.md), [`egui_glium`](../egui_glium/C
 
 #### Web:
 * Bug fix: modifiers keys getting stuck on alt-tab ([#2857](https://github.com/emilk/egui/pull/2857)).
-* ⚠️ BREAKING: Makes WebGL for wgpu opt-in. By default wasm builds for egui will now use WebGPU. To use WebGL, you need to add `wgpu = { version = "0.16.0", features = ["webgl"] }` now to your `Cargo.toml`. ([#2945](https://github.com/emilk/egui/pull/2945))
+* ⚠️ BREAKING: WebGPU is now the default web renderer when using the `wgpu` feature of `eframe`. To use WebGL with `wgpu`, you need to add `wgpu = { version = "0.16.0", features = ["webgl"] }` to your own `Cargo.toml`. ([#2945](https://github.com/emilk/egui/pull/2945))
 
 ## 0.21.3 - 2023-02-15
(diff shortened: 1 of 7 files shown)


Wumpf:
`./scripts/build_demo_web.sh --release` with feature `glow` changed to `wgpu`

`egui_demo_app_bg.wasm` filesize:

before: 6.62mb
after: 4.96mb

not too bad (:

emilk:
So WebGPU must be configured at compile-time to user either WebGL or WebGPU, but you cannot ever configure it so that you pick at runtime?

So in effect we'll need to have a `app_gl.wasm` and `app_wgpu.wasm` and pick the correct one using JavaScript?

Wumpf:
yes, that's my understanding of the state of things!

crates/eframe/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ NOTE: [`egui-winit`](../egui-winit/CHANGELOG.md), [`egui_glium`](../egui_glium/C
 
 #### Web:
 * Bug fix: modifiers keys getting stuck on alt-tab ([#2857](https://github.com/emilk/egui/pull/2857)).
+* ⚠️ BREAKING: Makes WebGL for wgpu opt-in. By default wasm builds for egui will now use WebGPU. To use WebGL, you need to add `wgpu = { version = "0.16.0", features = ["webgl"] }` now to your `Cargo.toml`. ([#2945](https://github.com/emilk/egui/pull/2945))
```
emilk, line 15: We should make clear this is only if you are using the `wgpu` flag for `eframe`

crates/eframe/src/web/web_painter_wgpu.rs
```diff
@@ -59,14 +77,20 @@ impl WebPainterWgpu {
     pub async fn new(canvas_id: &str, options: &WebOptions) -> Result<Self, String> {
         log::debug!("Creating wgpu painter");
 
+        // Workaround for https://github.com/gfx-rs/wgpu/issues/3710:
+        // Don't use `create_surface_from_canvas`, but `create_surface` instead!
```
emilk, line 81: You could keep the old code behind `if false { … }` so we can more easily switch back when the bug is fixed

Wumpf:
confirmed that now: https://github.com/gfx-rs/wgpu/discussions/3119#discussioncomment-3890627
Concluded that more from the code I debugged, but didn't go into the details. Personally I only ever looked at the backends that go through `direct.rs` (all backends except WebGPU, *including* webgl) and never looked at the mechanism that dispatches to `web.rs` instead.

emilk (approved):
Very cool!

It would be nice with a `--webgpu` flag to `scripts/build_demo_web.sh` so this can be easily tested

crates/eframe/CHANGELOG.md
emilk, line 15:
```suggestion
* ⚠️ BREAKING: WebGPU is now the default web renderer when using the `wgpu` feature of `eframe`. To use WebGL with `wgpu`, you need to add `wgpu = { version = "0.16.0", features = ["webgl"] }` to your own `Cargo.toml`. ([#2945](https://github.com/emilk/egui/pull/2945))
```

Wumpf:
confirmed non-workaround path working with pending fix on wgpu

PR #2951, opened 2023-04-24
[egui-wgpu] Device configuration is now dependent on adapter

Wumpf:
Additionally, renamed `backends` into `supported_backends` and improved & unified wgpu config defaults.

### DIFF
--- a/crates/eframe/src/web/web_painter_wgpu.rs
+++ b/crates/eframe/src/web/web_painter_wgpu.rs
@@ -159,9 +159,9 @@ impl WebPainter for WebPainterWgpu {
     }
 
     fn max_texture_side(&self) -> usize {
-        self.render_state
-            .as_ref()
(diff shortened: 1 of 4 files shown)


emilk (approved):

crates/eframe/src/web/web_painter_wgpu.rs
```diff
@@ -161,7 +159,10 @@ impl WebPainter for WebPainterWgpu {
     }
 
     fn max_texture_side(&self) -> usize {
-        self.limits.max_texture_dimension_2d as _
+        self.render_state
+            .as_ref()
+            .map(|state| state.device.limits().max_texture_dimension_2d as _)
+            .unwrap_or(0)
```
emilk, line 165:
```suggestion
            .map_or(0, |state| state.device.limits().max_texture_dimension_2d as _)
```

PR #2954, opened 2023-04-25
[wgpu] Expose wgpu::Adapter via RenderState

Wumpf:
Also improved  errors, docs, more code sharing and use stencil bits (no other backend seems to use those? Oo)

### DIFF
--- a/crates/egui-wgpu/src/lib.rs
+++ b/crates/egui-wgpu/src/lib.rs
@@ -56,7 +56,7 @@ impl RenderState {
     /// Creates a new `RenderState`, containing everything needed for drawing egui with wgpu.
     ///
     /// # Errors
-    ///     Wgpu initialization may fail due to incompatible hardware or driver for a given config.
+    /// Wgpu initialization may fail due to incompatible hardware or driver for a given config.
     pub async fn create(
         config: &WgpuConfiguration,
         instance: &wgpu::Instance,
--- a/crates/egui-wgpu/src/lib.rs
+++ b/crates/egui-wgpu/src/lib.rs
@@ -21,13 +21,75 @@ use std::sync::Arc;
 
 use epaint::mutex::RwLock;
 
+#[derive(thiserror::Error, Debug)]
+pub enum WgpuError {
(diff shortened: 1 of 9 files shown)


crates/egui-wgpu/src/winit.rs
```diff
@@ -1,8 +1,8 @@
 use std::sync::Arc;
 
-use epaint::{self, mutex::RwLock};
+use epaint::{self};
```
emilk, line 4:
```suggestion
```

crates/egui-wgpu/src/winit.rs
```diff
@@ -235,20 +203,20 @@ impl Painter {
...
         &mut self,
         window: Option<&winit::window::Window>,
     ) -> Result<(), crate::WgpuError> {
         match window {
             Some(window) => {
-                let surface = self.instance.create_surface(&window)?;
+                let surface = unsafe { self.instance.create_surface(&window)? };
 
                 self.ensure_render_state_for_surface(&surface).await?;
+                let render_state = self.render_state.as_ref().unwrap();
```
emilk, line 215: `unwrap` 😱  - please return an error instead, or restructure the code so it is impossible to hit

emilk (dismissed):
LGTM, except for the `unwrap`!

crates/egui-wgpu/src/lib.rs
```diff
@@ -116,39 +175,20 @@ pub fn preferred_framebuffer_format(formats: &[wgpu::TextureFormat]) -> wgpu::Te
     }
     formats[0] // take the first
 }
```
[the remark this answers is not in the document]
Wumpf: good catch, didn't notice. Let's add another thing to the now established error thiserror enum!

crates/egui-wgpu/src/winit.rs
Wumpf: Hmm I tried to return a reference from the line above, but can't do that because it takes ownership of `self`. Probably that method can just be inlined here and done

PR #2959, opened 2023-04-26
[DragValue] Use rounded values for checking if value needs setting

kamirr:
If `old_value` is not rounded, the modified if-expression would always trigger clearing the edit memory, and hence making editing by keyboard impossible -- the edit buffer would only last a single frame.

Alternatively, `old_value` could abe always rounded if we rounded the result of parsing the keyboard inpu and accesskit input, but this prevents the user from inputting values more precise than allowed by default, which is a behavior I've personally found useful. Regardless of that, this solution is more likely to prevent this bug from being reintroduced if new input methods are added and don't respect rounding.

### DIFF
--- a/crates/egui/src/widgets/drag_value.rs
+++ b/crates/egui/src/widgets/drag_value.rs
@@ -440,7 +440,10 @@ impl<'a> Widget for DragValue<'a> {
         }
 
         value = clamp_to_range(value, clamp_range.clone());
-        if old_value != value {
+
+        let old_rounded = emath::round_to_decimals(old_value, auto_decimals);
+        let new_rounded = emath::round_to_decimals(value, auto_decimals);
+        if old_rounded != new_rounded {
             set(&mut get_set_value, value);
             ui.memory_mut(|mem| mem.drag_value.edit_string = None);
         }


kamirr:
`./scripts/check.sh` fails due to the following:
```
+ cd crates/egui_glow
+ cargo check --all-features
   Compiling smithay-client-toolkit v0.16.0
   Compiling wayland-protocols v0.29.5
   Compiling winit v0.28.1
    Checking epaint v0.21.0 (/home/kamil/workspace/egui/crates/epaint)
    Checking egui v0.21.0 (/home/kamil/workspace/egui/crates/egui)
error: The platform you're compiling for is not supported by winit
  --> /home/kamil/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.28.1/src/platform_impl/mod.rs:67:1
   |
67 | compile_error!("The platform you're compiling for is not supported by winit");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I don't think it's relevant to this patch.

PR #2961, opened 2023-04-27
Log warning instead of error when failing to decode RON in storage

emilk:
I'm not sure why, but using `console.error` causes a hard crash.

image

But this PR fixes it. I don't understand why.


teh-cmc (approved):
Can confirm this works.

To reproduce/test, manually introduce a typo in one of the keys stored in local storage.

Firefox Linux, using egui.rs:
image

Firefox Linux: using this branch:
image

Issue #1600, opened 2022-05-07
Support setting application ID for Wayland


Detegr:
**Is your feature request related to a problem? Please describe.**
Applications using `egui` on Wayland have their [app_id](https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_app_id) property set as `null`. As far as I can tell there is currently no way to set it.

**Describe the solution you'd like**
I'd like to have the window title to be used as `app_id` for Wayland. This looks to be as simple as calling the WindowBuilder's [with_app_id](https://docs.rs/winit/latest/winit/platform/unix/trait.WindowBuilderExtUnix.html#tymethod.with_app_id) function with the title string as the value.

**Additional context**
Some window managers, `sway` to name one, can make use of `app_id` value to detect windows of certain applications, and handle those windows in some different manner than others.

I'm willing to submit a pull request implementing this if this is something you agree would be an useful feature to have.

dimtpap:
This would also allow some DEs/WMs to show an icon for the application in various places such as the title bar, if the application has provided desktop configuration files and icons

Detegr:
> I assume the hard part is to configure this source to only run on Wayland (I'm not sure if this is exposed to the eframe crate as configuration yet).

`eframe` has a `wayland` feature so you should be able to use `#[cfg(feature = "wayland")]`.

> I am also not sure if it is wise to re-use the application title here (which is the same as the application name currently, but will this always be the case?), or if a more specific option e.g. in `eframe::NativeOptions` would better.

I'm not sure about this either.

dimtpap:
> I am also not sure if it is wise to re-use the application title here (which is the same as the application name currently, but will this always be the case?), or if a more specific option e.g. in `eframe::NativeOptions` would better.

App IDs are supposed to be in reverse domain format like `org.example.Application` but the title would usually be just `Application` so a more specific option would be better. [Relevant documentation](https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_app_id)

thomaskrause:
I added a pull request (https://github.com/emilk/egui/pull/3007) for this, but decided against using the title as default. This would alter the behavior of existing programs, and I am uncertain if there are any limitations to application IDs that would conflict with more "weird" window titles. I used the `NativeOptions` instead, since there is already a macOS specific configuration value and so it makes sense to add a Wayland specific one there.

PR #3007, opened 2023-05-21
Add an optional app_id field to eframe's NativeOptions for Wayland

thomaskrause:
Closes <https://github.com/emilk/egui/issues/1600>.

The new `app_id` is used in the window builder to set the Wayland application ID using the [winit API](https://docs.rs/winit/latest/winit/platform/wayland/trait.WindowBuilderExtWayland.html#tymethod.with_name), which is e.g. important for a proper configuration in `.desktop` files.
There should not be any changes in the behavior for existing applications.

The new field is only available when the `target_os` is set to "linux"  and the "wayland" feature is enabled  (similar to the `fullsize_content` field in macOS). The limitation to `target_os = "linux"` is needed because the winit crate does not export the needed trait on Windows and other Non-Unix systems (causing compilation issues otherwise). 
Because this makes the configuration a little bit more difficult to use, I added an example in its documentation.

### DIFF
--- a/crates/eframe/src/epi/mod.rs
+++ b/crates/eframe/src/epi/mod.rs
@@ -383,6 +383,21 @@ pub struct NativeOptions {
     /// Configures wgpu instance/device/adapter/surface creation and renderloop.
     #[cfg(feature = "wgpu")]
     pub wgpu_options: egui_wgpu::WgpuConfiguration,
+
+    /// On Wayland: Application ID for the window.
+    ///
+    /// The application ID is used in several places of the compositor, e.g. for
+    /// grouping windows of the same application. It is also important for
+    /// connecting the configuration of a `.desktop` file with the window, by
+    /// using the application ID as file name. This allows e.g. a proper icon
+    /// handling under Wayland.
+    ///
+    /// See [Waylands XDG shell documentation][xdg-shell] for more information
+    /// on this Wayland-specific option.
+    ///
+    /// [xdg-shell]: https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_app_id
+    #[cfg(feature = "wayland")]
+    pub app_id: Option<String>,
 }
 
 #[cfg(not(target_arch = "wasm32"))]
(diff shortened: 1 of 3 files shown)


crates/eframe/src/epi/mod.rs
```diff
@@ -383,6 +383,40 @@ pub struct NativeOptions {
...
+    /// The application ID is used in several places of the compositor, e.g. for
+    /// grouping windows of the same application. It is also important for
+    /// connecting the configuration of a `.desktop` file with the window, by
+    /// using the application ID as file name. This allows e.g. a proper icon
+    /// handling under Wayland.
+    ///
+    /// See [Waylands XDG shell documentation][xdg-shell] for more information
+    /// on this Wayland-specific option.
+    ///
+    /// [xdg-shell]: https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_app_id
```
emilk, line 398: Is there a restriction of what the app id can be? If it can be anything, why not use the app title passed to `run_simple_native`?

Mingun: I think, that application title can be a) dynamic and b) localized which makes it bad choice for such purpose.

emilk: I agree - I think we should remove `#[cfg(all(feature = "wayland", target_os = "linux"))]` and then also use `app_id` when calling `epi_integration::create_storage`. But I guess I can do that in a follow-up PR

crates/eframe/src/native/epi_integration.rs
```diff
@@ -118,6 +121,11 @@ pub fn window_builder<E>(
             .with_fullsize_content_view(true);
     }
 
+    #[cfg(all(feature = "wayland", target_os = "linux"))]
+    if let Some(app_id) = native_options.app_id.as_deref() {
```
emilk, line 125: little shorter 🤷 

```suggestion
    if let Some(app_id) = &native_options.app_id {
```

emilk (approved):

crates/eframe/src/epi/mod.rs
thomaskrause: Yes, removing the app_id for the storage would probably also be a good fit for this field. Especially, if the window title might become dynamic at some point (e.g. for indicating non-saved files and this kind of things), the `app_id`  could still be static and only set at startup.

thomaskrause:
Thanks for the review! I agree with all the proposed (it is already committed), and would also like the idea of using the `app_id` for storage (but in a separate PR).

PR #3021, opened 2023-05-23
Add hifitime datepicker & make datepicker generic

### LINKED ISSUE #2932: Remove implicit dependency on `time 0.1`
`chrono`, by default, depends on an old version of `time`, which has a RUSTSEC: https://github.com/emilk/egui/security/dependabot/7

`egui` doesn't use `chrono`, but `egui_extras` does as does `egui_demo_lib`. It would be nice to replace it with pure `time 0.3` use, or use a`chrono` without the `oldtime` feature it has.

It's worth noting that next release of `chrono` seem to have removed the dependency on `time` completely, so maybe this will sort itself out in time

Massou31:
I've copy/pasted my changes to the datepicker that uses https://github.com/nyx-space/hifitime

I've tested succinctly and it works but there may be some edge cases not handled yet.

scripts/check.sh do not work yet because the features are not mutually exclusive (after a quick check I found this: https://github.com/rust-lang/cargo/issues/2980 but it seem that some mitigations are possible). 

I'll dig further on the features later this week or next week as well as more testing.

Closes <https://github.com/emilk/egui/issues/2932>.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1265,6 +1265,7 @@ dependencies = [
  "chrono",
  "document-features",
  "egui",
+ "hifitime",
  "image",
  "log",
  "resvg",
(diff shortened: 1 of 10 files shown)


Massou31:
I've just changed the implementation in order to make an abstraction on the datepicker, this way any date/time library could be used as long as the required trait is implemented.

There is 1 known bug regarding how I compute the week number with hifitime (I plan to fix it sometime in the next 2 weeks, I'll remove the draft status once it's fixed)

@emilk I think that this pull request could have an initial review.

emilk:
Closing: the datepicker moved from chrono to jiff in #8008, so this no longer applies. Thanks for the PR!

PR #3022, opened 2023-05-23
Add an optional skip_taskbar field to eframe's NativeOptions for hide taskbar icon under windows

hoothin:
Add support for `skip_taskbar` of winit to hide taskbar icon under windows.


hoothin (commented):
use with_skip_taskbar only under windows

PR #3026, opened 2023-05-25
Fix a document: `eframe::start_web`

kerryeon:
In `0.22.0`, `eframe::start_web` has been replaced with `eframe::WebRunner`.

[But, there is a missed document in `eframe` `/lib.rs`.](https://github.com/emilk/egui/blob/307565efa55158cfa6b82d2e8fdc4c4914b954ed/crates/eframe/src/lib.rs#L10)

https://github.com/emilk/egui/blob/307565efa55158cfa6b82d2e8fdc4c4914b954ed/crates/eframe/src/lib.rs#L10

[I think it should be like:](https://github.com/ulagbulag/egui/blob/d9cf9e6f3a4c85b7029fb36abf3e85ac7afcf9df/crates/eframe/src/lib.rs#L10)

```rust
//! call [`crate::run_native`] from your `main.rs`, and/or use `eframe::WebRunner` from your `lib.rs`.
```

And thanks for driving the artistic Rust GUI framework!

### DIFF
--- a/crates/eframe/src/lib.rs
+++ b/crates/eframe/src/lib.rs
@@ -7,7 +7,7 @@
 //! To learn how to set up `eframe` for web and native, go to <https://github.com/emilk/eframe_template/> and follow the instructions there!
 //!
 //! In short, you implement [`App`] (especially [`App::update`]) and then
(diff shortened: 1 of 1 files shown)


emilk (approved):
Thanks!

PR #3031, opened 2023-05-27
Derive Default for Rect

e00E:
Makes it slightly nicer to work with and I don't see a reason against the derive. `Pos2` already is Default.


emilk:
There is a reason: what should the default be? You have now defined it to be the rectangle containing the origin (0,0). When does this default make sense? I don't know. I do know that this default is a foot gun when you want to use `Rect` as a bounding box - then `Rect::NOTHING` is a better default. Yes, I have seen that exact problem cause bugs :)

Issue #576, opened 2021-07-23
Hover not recognized right next to Area


mankinskin:
![Peek 2021-07-23 19-34](https://user-images.githubusercontent.com/20745737/126819972-3eb5949f-3168-4445-aca3-9cd1894db8e6.gif)

This is what the container looks like:
```rust
    let area = Area::new(menu_id)
        .order(Order::Foreground)
        .fixed_pos(pos)
        .drag_bounds(Rect::EVERYTHING);
    let mut frame = Frame::menu(&style);
    frame.shadow = Default::default();
    area.show(ctx, |ui| {
        frame
            .show(ui, |ui| {
                //style.visuals.widgets.active.bg_fill = Color32::TRANSPARENT;
                style.visuals.widgets.active.bg_stroke = Stroke::none();
                //style.visuals.widgets.hovered.bg_fill = Color32::TRANSPARENT;
                style.visuals.widgets.hovered.bg_stroke = Stroke::none();
                style.visuals.widgets.inactive.bg_fill = Color32::TRANSPARENT;
                style.visuals.widgets.inactive.bg_stroke = Stroke::none();
                ui.set_style(style);
                ui.with_layout(Layout::top_down_justified(Align::LEFT), add_contents)
                    .inner
            })
    })
```
The same thing happens on master with the `menu`:
![Peek 2021-07-23 19-38](https://user-images.githubusercontent.com/20745737/126820464-bbabf410-7d00-4e8c-9acb-885a2d21709c.gif)

debug on hover also shows that in that area none of the two overlapping widgets are hovered.

I suspect it is something with [`Context::interact`](https://github.com/emilk/egui/blob/master/egui/src/context.rs#L151), but I didn't figure it out yet. Any help would be appreciated 👍🏻 

required for #543

mankinskin:
seems to be caused by `Context::layer_id_at` https://github.com/emilk/egui/blob/master/egui/src/context.rs#L730

mankinskin:
So the issue is `context.options.style.interaction.resize_grab_radius_side` which [catches the hover](https://github.com/emilk/egui/blob/master/egui/src/memory.rs#L479) in an area designated for dragging the sides of resizable windows. This is used for all Areas though, while it is only necessary for resizable areas.

I am now looking into disabling this extra radius for areas which are not resizable.

mankinskin:
[...]

PR #3039, opened 2023-05-31
Have non-resizable Areas ignore mouse input just beyond their bounds

fleabitdev:
As described in #576, the method `Memory::layer_id_at` expands the hit-testing region for every `Area` slightly. This is necessary so that, when the user clicks to resize a `Window`, the mouse input isn't routed to a different `Window`.

For non-resizable `Area`s (such as dropdown menus, context menus, date pickers, and any non-resizable `Window`), this causes them to be surrounded by a "dead zone" where any underlying widgets can't be hovered or clicked. The effect is particularly noticeable in menu bars.

This commit adds a persisted `edges_padded_for_resize` property to `Area`, which is `true` when the `Area` belongs to a resizable `Window` and `false` for all other `Area`s. The hit-testing region is only expanded when this property is `true`.


fleabitdev:
I'm a little uneasy about my approach here, because it forces `Window` to manually keep `self.area.edges_padded_for_resize` in sync with `self.resize.resizable`. It would be better to have a single source of truth, but that would probably require `Area` to be made responsible for resizing, which seems like a much larger change. Any thoughts?

emilk (approved):
It's a bit hacky I agree, but I fail to see any obviously better way

emilk:

PR #3042, opened 2023-06-01
Improve rendering of color buttons

fleabitdev:
This commit introduces several small improvements in the appearance of color buttons:
- Checks now properly line up with one another in the X axis.
- Checks are now anti-aliased.
- There must be an even number of checks in the horizontal direction. (This produces a more aesthetically pleasing result when the checkerboard is split into left and right halves.)

The last change incidentally solves an unreported bug where color buttons, at their default size, would display a different check pattern when hovered.


emilk (commented):
A before/after image would be nice 🙏

emilk:
converted to draft until there is progress

emilk:
Closing: the checkerboard is now a repeating texture since #8439, so this no longer applies. Thanks for the PR!

Issue #3046, opened 2023-06-01
Menus cast a shadow over the widget which opened them


fleabitdev:
<img src="https://github.com/emilk/egui/assets/65809749/e507aeb7-f40c-4c3a-b639-e8de406423e2" width=300>
<img src="https://github.com/emilk/egui/assets/65809749/d8b32d88-55fe-45db-b878-a1448b05130e" width=300>

GTK 4 and Qt 6 ameliorate this by having a relatively small drop shadow on menus, and by shifting that drop shadow downwards. GTK shifts the whole shadow down by about half of its radius, while Qt moves the shadow's top edge down by its entire radius.

<img src="https://github.com/emilk/egui/assets/65809749/25009124-85e9-43a1-b697-b878827e61cc" width=150>
<img src="https://github.com/emilk/egui/assets/65809749/7d3317cd-9b51-4765-8f2e-a790b4b23690" width=300>

In both cases, the same vertical shadow offset is applied to submenus. This means that submenus are allowed to shadow the right edge of their parent menu.

PR #3047, opened 2023-06-01
Adjust menu shadows

fleabitdev:
This commit makes menu and tooltip shadows slightly fainter and smaller.

It also introduces the option for shadows' top edges to be slightly inset relative to their parent window. This causes menus and tooltips to cast less of a shadow over the widget which opened them.

Closes <https://github.com/emilk/egui/issues/3046>.

### DIFF
--- a/crates/epaint/src/shadow.rs
+++ b/crates/epaint/src/shadow.rs
@@ -11,27 +11,34 @@ pub struct Shadow {
 
     /// Color of the opaque center of the shadow.
     pub color: Color32,
+
+    /// Y-offset for the top edge of the rectangle. Positive values make the shadow
+    /// smaller, negative values make the shadow larger. The bottom edge is unchanged.
+    pub top_offset: f32,
 }
 
 impl Shadow {
     pub const NONE: Self = Self {
         extrusion: 0.0,
         color: Color32::TRANSPARENT,
+        top_offset: 0.0,
     };
 
     /// Tooltips, menus, …, for dark mode.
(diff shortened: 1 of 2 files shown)


fleabitdev:
Before/after comparison (animated):

before-after

crates/epaint/src/shadow.rs
```diff
@@ -11,27 +11,34 @@ pub struct Shadow {
 
     /// Color of the opaque center of the shadow.
     pub color: Color32,
+
+    /// Y-offset for the top edge of the rectangle. Positive values make the shadow
+    /// smaller, negative values make the shadow larger. The bottom edge is unchanged.
+    pub top_offset: f32,
```
emilk, line 17: This will leave users wanting to do the same for the other three sides. You could use `Margin` here instead of `f32`, allowing users to tweak all four sides. You could use `Margin::expand_rect` to easily tweak the resulting rectangle.

I think it makes more sense that large values means "expand", so that to offset the whole shadow down and to the right (make it more like a drop-shadow) you could use `Margin { left: -8., right: 8., top: -8., bottom: 8. }`, for instance

emilk, line 17:
```suggestion
    pub offset: Margin,
```

emilk (commented):
Results looks good for the given examples, though it's worth noting that when you move a menu button to the bottom of the screen, the resulting menu is opened above the button.

emilk:
bump

PR #1889, opened 2022-08-03
Continue execution after closing native eframe window

emilk:
Closes https://github.com/emilk/egui/issues/1223

This adds `NativeOptions::run_and_return` with default value `true`.

If `true`, execution will continue after the eframe window is closed. This is a new behavior introduced in this PR.
If `false`, the app will close once the eframe window is closed. The is the old behavior.

This is `true` by default, and the `false` option is only there so we can revert if we find any bugs.

When `true`, [`winit::platform::run_return::EventLoopExtRunReturn::run_return`](https://docs.rs/winit/latest/winit/platform/run_return/trait.EventLoopExtRunReturn.html#tymethod.run_return) is used. The winit docs warns of its usage, recommending `EventLoop::run`, but 🤷 
When `false`, [`winit::event_loop::EventLoop::run`](https://docs.rs/winit/latest/winit/event_loop/struct.EventLoop.html#method.run) is used.

This is a really useful feature. You can now use `eframe` to quickly open up a window and show some data or options, and then continue your program after the `eframe` window is closed

My previous attempt at this caused some problems, but my new attempt seems to be working much better, at least on my Mac.

PR #3053, opened 2023-06-04
eframe: Only run_return twice on Windows

pan93412:
The approach of #1889 may remove the observer of a view twice, which produces the Obj-C Exception:

    Cannot remove an observer <...> for the key path "nextResponder" from <WinitView ...> because it is not registered as an observer.

The above message can only be seen when attaching the application to debugger. Users normally see:

    [1]    *** trace trap  cargo run

This commit fixes it by only running event_loop twice on Windows. Besides:

* We have set ControlFlow::Exit on 'Event::LoopDestoryed', 'EventResult::Exit' and on error; therefore, it is safe to not calling `set_exit()`.
* This commit also fix the persistence function in macOS. It can't store the content in Memory due to this exception.

Fixed: #2768 (eframe: "App quit unexpectedly" on macOS)


emilk (approved):
Makes sense - thanks!

Issue #2853, opened 2023-03-29
Sub-menus are too eager to close


emilk:
I think we should consider re-defining what keeps a sub-menu open.

The current behavior is "keep sub-menu open while hovering it or any child". This leads to sub-menus closing very easily when the mouse strays outside their active area.

I think we should change this to "remain open until another sub-menu-button is hovered, unless is is a descendant UI". That's how most menu systems work.

Se the problem here:

![menus-close-too-quickly](https://user-images.githubusercontent.com/1148717/228555066-8a8fab04-c602-4b5f-9d15-1c0764da83dc.gif)

PR #3055, opened 2023-06-06
Changed logic for submenus closing

DannyStoll1:
This tiny change removes the trigger for a submenu to close if the mouse is no longer hovering over it or one of its children. Now, the menu stays open until a different menu item is selected. This aims to address <https://github.com/emilk/egui/issues/2853>.

I suspected that a new trigger would need to be implemented so that multiple menu items don't remain open simultaneously, but after some testing, this does not appear to occur.

This change leaves `MenuState::hovering_current_submenu` and `MenuState::close_submenu` as dead code in case they might be useful later.


emilk (approved):
Nice, thanks for the fix! I think we should remove the dead code though. We can always revert it if we want it again.

PR #3058, opened 2023-06-06
Remove panic on plot grid

nrot:
Close [issue](https://github.com/emilk/egui/pull/1180#issuecomment-1570874066).

### DIFF
--- a/crates/egui/src/containers/area.rs
+++ b/crates/egui/src/containers/area.rs
@@ -52,7 +52,7 @@ impl State {
 ///
 /// ```
 /// # egui::__run_test_ctx(|ctx| {
-/// egui::Area::new("my_area")
+/// egui::Area::new(egui::Id::new("my_area"))
 ///     .fixed_pos(egui::pos2(32.0, 32.0))
 ///     .show(ctx, |ui| {
 ///         ui.label("Floating text!");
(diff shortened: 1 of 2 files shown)


crates/egui/src/widgets/plot/mod.rs
```diff
@@ -1650,6 +1650,10 @@ fn generate_marks(step_sizes: [f64; 3], bounds: (f64, f64)) -> Vec<GridMark> {
 /// Fill in all values between [min, max] which are a multiple of `step_size`
 fn fill_marks_between(out: &mut Vec<GridMark>, step_size: f64, (min, max): (f64, f64)) {
     assert!(max > min);
+    if step_size.abs() == 0.0 {
+        trace!("To small step_size");
```
emilk, line 1654:
```suggestion
    if step_size == 0.0 {
        log::trace!("Ignoring step_size of zero");
```

… though can we still get a panic when `step_size` is very small, or `NaN`? What was the actual panic? Can we fix it at the source instead of adding a guard like this?

nrot:
Panic this [extend](https://github.com/emilk/egui/blob/master/crates/egui_plot/src/lib.rs#L1900). When step_size to small last - first = very big. And then allocated more memory than have system and this is panic in extend.

PR #3073, opened 2023-06-11
Expose Raw Window and Display Handles

### LINKED ISSUE #2757: Keep filedialog in front of main window with eframe - Add the `HasRawWindowHandle` to `eframe::Frame` maybe?
When using `rfd` to open a file fialog window, It easly goes behind the main window, freezing the latter and confusing the user. 
I know this is a solved issue with egui-winit, since it is possible to give to rfd::FileDialog().set_parent(var [of type winit::window::Window])

However, when using eframe, i cannot seem to find the window that was generated when the frameware ran `winit::windw::WindowBuilder`. I don't think it is public. 
In the main update loop, i have the `ctx` and the `frame`, and it would be good if `frame` had the  `HasRawWindowHandle`  `rfd` requires to be set as parent.

I tried directly using the egui-winit example, but I have mean drawbacks specifically in terms of quality of the rendering. eframe has been the best performing for me

**Additional context**
Some of the docs are below

https://docs.rs/winit/0.28.1/winit/window/struct.Window.html#method.raw_window_handle
https://docs.rs/raw-window-handle/0.5.0/raw_window_handle/trait.HasRawWindowHandle.html
https://docs.rs/rfd/latest/rfd/struct.FileDialog.html#method.set_parent

bash:
I went ahead and made this PR based on @mvirkkunen's [patch](https://github.com/emilk/egui/issues/2757#issuecomment-1455865907).

Closes #2757.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1164,6 +1164,7 @@ dependencies = [
  "raw-window-handle",
  "ron",
  "serde",
+ "static_assertions",
  "thiserror",
  "tts",
  "wasm-bindgen",
--- a/crates/eframe/CHANGELOG.md
+++ b/crates/eframe/CHANGELOG.md
@@ -5,7 +5,7 @@ NOTE: [`egui-winit`](../egui-winit/CHANGELOG.md), [`egui_glium`](../egui_glium/C
 
 
 ## Unreleased
-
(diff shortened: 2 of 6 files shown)


emilk (approved):
Nice, thank you!

Issue #3068, opened 2023-06-10
egui/eframe using wgpu/GL panics at startup after upgrade to v0.22


NoisyCoil:
**Describe the bug**
On an Nvidia graphics card (NVIDIA Corporation GK107M [GeForce GT 750M Mac Edition]), every egui/eframe app panics at startup with an `'Error in Surface::configure: Validation Error`, caused by `Requested usage is not supported`, when using wgpu. This didn't happen before the upgrade to v0.22.

**To Reproduce**
1. Enable the "wgpu" feature in eframe
2. Add `Renderer::Wgpu` to eframe's `NativeOptions`
3. Compile

**Expected behavior**
The application should not panic.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ubuntu 23.04 x86_64 (both Wayland and X11)
 - Version 0.22 and current master

**Additional context**
The error and backtrace I see e.g. with the hello_world example on Wayland is

```
   Compiling hello_world v0.1.0 (/home/user/egui/examples/hello_world)
    Finished dev [unoptimized + debuginfo] target(s) in 12.07s
     Running `/home/user/egui/target/debug/hello_world`
thread 'main' panicked at 'Error in Surface::configure: Validation Error

Caused by:
    Requested usage is not supported
', /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.16.0/src/backend/direct.rs:734:18
stack backtrace:
   0: rust_begin_unwind
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/panicking.rs:142:14
   2: wgpu::backend::direct::Context::handle_error_fatal
             at /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.16.0/src/backend/direct.rs:308:9
   3: <wgpu::backend::direct::Context as wgpu::context::Context>::surface_configure
             at /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.16.0/src/backend/direct.rs:734:13
   4: <T as wgpu::context::DynContext>::surface_configure
             at /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.16.0/src/context.rs:2056:9
   5: wgpu::Surface::configure
[...]

PR #3078, opened 2023-06-14
Fix panic on wgpu GL backend due to new screenshot capability

amfaber:
This is a first-pass triage for https://github.com/emilk/egui/issues/3068. The fix involves simply disabling the screen shot feature when using the GL backend with wgpu. If a screenshot is requested on this backend, None will be provided and an error is logged out. The docs have been updated with an small comment on how the feature isn't available with the GL backend.

The reason that the feature is incompatible with the GL backend is that this backend doesn't allow the surface to have the TextureUsages::COPY_DST flag, and so the approach of rendering to a texture and subsequently copying the result to the surface and a buffer for extracting to the CPU is fundamentally incompatible.

Perhaps @Wumpf can help with ideas on how to circumvent this?

### DIFF
--- a/crates/eframe/src/epi/mod.rs
+++ b/crates/eframe/src/epi/mod.rs
@@ -798,7 +798,8 @@ impl Frame {
     /// Returns None if:
     /// * Called in [`App::update`]
     /// * [`Frame::request_screenshot`] wasn't called on this frame during [`App::update`]
-    /// * The rendering backend doesn't support this feature (yet). Currently implemented for wgpu and glow, but not with wasm as target or gl as backend in the case of wgpu.
+    /// * The rendering backend doesn't support this feature (yet). Currently implemented for wgpu and glow, but not with wasm as target.
+    /// * Wgpu's GL target is active (not yet supported)
     /// * Retrieving the data was unsuccessful in some way.
     ///
     /// See also [`egui::ColorImage::region`]
--- a/crates/eframe/src/epi/mod.rs
+++ b/crates/eframe/src/epi/mod.rs
@@ -798,7 +798,7 @@ impl Frame {
     /// Returns None if:
     /// * Called in [`App::update`]
     /// * [`Frame::request_screenshot`] wasn't called on this frame during [`App::update`]
-    /// * The rendering backend doesn't support this feature (yet). Currently implemented for wgpu and glow, but not with wasm as target.
+    /// * The rendering backend doesn't support this feature (yet). Currently implemented for wgpu and glow, but not with wasm as target or gl as backend in the case of wgpu.
     /// * Retrieving the data was unsuccessful in some way.
     ///
     /// See also [`egui::ColorImage::region`]
--- a/crates/egui-wgpu/src/winit.rs
+++ b/crates/egui-wgpu/src/winit.rs
@@ -581,7 +581,7 @@ impl Painter {
                 Self::read_screen_rgba(screen_capture_state, render_state, &output_frame)
             }
             (true, false) => {
-                log::error!("Screenshots are not supported on backend that don't allow the render surface to have the TextureUsages::COPY_DST flag");
+                log::error!("The active render surface doesn't support taking screenshots.");
                 None
             }
             (false, _) => None,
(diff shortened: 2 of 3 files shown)


crates/eframe/src/epi/mod.rs
```diff
@@ -798,7 +798,7 @@ impl Frame {
     /// Returns None if:
     /// * Called in [`App::update`]
     /// * [`Frame::request_screenshot`] wasn't called on this frame during [`App::update`]
-    /// * The rendering backend doesn't support this feature (yet). Currently implemented for wgpu and glow, but not with wasm as target.
+    /// * The rendering backend doesn't support this feature (yet). Currently implemented for wgpu and glow, but not with wasm as target or gl as backend in the case of wgpu.
```
Wumpf, line 801:
```suggestion
    /// * The rendering backend doesn't support this feature (yet). Currently implemented for wgpu and glow, but not with wasm as target.
    /// * Wgpu's GL target is active (not yet supported)
```
it's a bit too long otherwise

crates/egui-wgpu/src/winit.rs
```diff
@@ -560,13 +569,24 @@ impl Painter {
...
+                .texture
+                .usage()
+                .contains(wgpu::TextureUsages::COPY_DST),
+        ) {
+            (true, true) => {
+                let screen_capture_state = self.screen_capture_state.as_ref()?;
+                Self::read_screen_rgba(screen_capture_state, render_state, &output_frame)
+            }
+            (true, false) => {
+                log::error!("Screenshots are not supported on backend that don't allow the render surface to have the TextureUsages::COPY_DST flag");
```
Wumpf, line 584:
```suggestion
                log::error!("The active render surface doesn't support taking screenshots.");
```

crates/egui-wgpu/src/winit.rs
```diff
@@ -487,7 +491,12 @@ impl Painter {
 
         {
             let renderer = render_state.renderer.read();
-            let frame_view = if capture {
+            let frame_view = if capture
+                & output_frame
+                    .texture
+                    .usage()
+                    .contains(wgpu::TextureUsages::COPY_DST)
```
Wumpf, line 498: the repeated check on the texture usage is a bit tedious and arcane if one doesn't know about this. How about putting a `supports_screenshots` bool on `surface_state` which we then use to set `capture` always to false + log a warning (I'm on the fence if that's worth a `log::error`, in many cases this isn't critical enough)

Wumpf (changes requested):
Great catch, makes sense!

But the culprit isn't really GL backend not supporting a thing, but that that we blindly assume the COPY_DST usage is supported. This can happen on any backend in theory, so therefore we should check the availability of that usage flag and then draw the right conclusion.... only that I noticed now that this is not possible right now, I opened an issue here https://github.com/gfx-rs/wgpu/issues/3869

_Even_ better would be ofc to just not use texture copies here, but instead do screen triangle pass to render the final texture, but let's not get into additional complexities right now :)

So given the situation I think we can move on with this, but I'd like some small improvements in here, see comments :)

crates/egui-wgpu/src/winit.rs
amfaber: Great call! On the error / warning debate, I originally wrote this for my application in which I want to be able to export the shown video as an mp4. In this case if the user requests the video to be recorded but it can't be, I would consider it an error - just my 2 cents.

Wumpf (approved):
ci is unhappy, otherwise good to go imho

amfaber:
Thanks for the input! As I am still getting into graphics programming, I am especially interested in the benefits of doing a triangle screen (re) render rather than a copy: Are there any benefits other than not requiring COPY_DST on the surface (though this benefit is clearly sufficient on its own)? In my mind I would have thought that a straight copy would be one of the most efficient operations that we could perform. Really looking to learn ^^

Wumpf:
It should generally be way better to do the copy via the copy texture operation, but with these things it's very much dependent on the driver and what other loads are in flight. But yes the only actual reason here would be compatibility

Issue #2037, opened 2022-09-12
Pretty way for leak private Placer for calc(100% - 96px)


enomado:
Hi. I have `ui.group()` in which I have `ui.vertical()` and Table. I want something like

![Screenshot_2022-09-12_06-18-55](https://user-images.githubusercontent.com/707007/189567757-b1757916-d9a8-46e6-9a17-405daa35b296.png)

So, group-frame is responsive, Table is responsive, buttons-row is static.  So I want something like height: `calc(100% - 71px)`

I figured out this would work for me.

```
    {
        let pp = ui.placer_mut();
        pp.region.max_rect.max.y -= 71.;
    }
    {   render table   } 


```

Or is there better way to do that?

enomado:
I came up with this.


            let mut rect = ui.available_rect_before_wrap();
            rect.max.y = rect.max.y - 32.;

            let p = ui.allocate_ui_at_rect(rect, |ui| {table...})
            
            last_row()

PR #3089, opened 2023-06-17
make layout, region public

enomado:
Hi. It is stuff from my fork to do some tricky layouts like https://github.com/emilk/egui/issues/2037, table-row hovering and custom styling, etc.

And also there is my custom context menu with some additional control.

Seems I cant do this things different way.

### DIFF
--- a/crates/egui/src/placer.rs
+++ b/crates/egui/src/placer.rs
@@ -1,10 +1,10 @@
 use crate::*;
 
-pub(crate) struct Placer {
+pub struct Placer {
     /// If set this will take precedence over [`layout`].
     grid: Option<grid::GridLayout>,
     layout: Layout,
-    region: Region,
+    pub region: Region,
 }
 
 impl Placer {
--- a/crates/egui/src/ui.rs
+++ b/crates/egui/src/ui.rs
@@ -815,6 +815,10 @@ impl Ui {
         &self.placer
     }
 
+    pub fn placer_mut(&mut self) -> &mut Placer {
+        &mut self.placer
+    }
+
     /// Where the next widget will be put.
     ///
     /// One side of this will always be infinite: the direction in which new widgets will be added.
--- a/crates/egui_extras/src/table.rs
+++ b/crates/egui_extras/src/table.rs
@@ -1010,14 +1010,14 @@ impl<'a> Drop for TableBody<'a> {
 /// The row of a table.
 /// Is created by [`TableRow`] for each created [`TableBody::row`] or each visible row in rows created by calling [`TableBody::rows`].
 pub struct TableRow<'a, 'b> {
-    layout: &'b mut StripLayout<'a>,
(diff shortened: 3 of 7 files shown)


crates/egui/src/ui.rs
```diff
@@ -815,6 +815,10 @@ impl Ui {
         &self.placer
     }
 
+    pub fn placer_mut(&mut self) -> &mut Placer {
```
emilk, line 818: You should probably make `player()` public too then?

crates/egui/src/placer.rs
```diff
@@ -1,10 +1,10 @@
 use crate::*;
 
-pub(crate) struct Placer {
+pub struct Placer {
```
emilk, line 3: Now that this is public, it deserves a docstring explaining it and what it does

crates/egui/src/ui.rs
emilk: Making this public and mutable means the user can change layout in the middle of a UI, which is something new, and might break the expectations of other widgets and containers. I suggest we should add a strong "here be dragons" warning here

crates/egui_extras/src/table.rs
```diff
@@ -1010,14 +1010,14 @@ impl<'a> Drop for TableBody<'a> {
...
-    layout: &'b mut StripLayout<'a>,
+    pub layout: &'b mut StripLayout<'a>,
     columns: &'b [Column],
     widths: &'b [f32],
     /// grows during building with the maximum widths
     max_used_widths: &'b mut [f32],
     col_index: usize,
     striped: bool,
-    height: f32,
+    pub height: f32,
```
emilk, line 1020: Do we support changing the height in the middle of a row? What does that do?

crates/egui_extras/src/table.rs
```diff
@@ -1010,14 +1010,14 @@ impl<'a> Drop for TableBody<'a> {
 /// The row of a table.
 /// Is created by [`TableRow`] for each created [`TableBody::row`] or each visible row in rows created by calling [`TableBody::rows`].
 pub struct TableRow<'a, 'b> {
-    layout: &'b mut StripLayout<'a>,
+    pub layout: &'b mut StripLayout<'a>,
```
emilk, line 1013: What would it mean to change layout in the middle of a row? Should we really support that? Wouldn't a non-mutable accessor be better?

emilk (commented):
I agree that there are things we can make more public, and many of the things in this PR fall under that category, but making things public brings a bigger responsibility when it comes to maintaining invariances. If a user can change the internals, what does that do with the state of the thing? Is that something that we now officially support, or something only power users should do? It at least requires a lot more documentation.

A user will look at the `pub`lic API in the docs and should reasonably expect that any use of the public API will leave the object in a good state. If it is public, it should be fine to mutate. Have you really thought that through in all the places in this PR?

PR #3100, opened 2023-06-21
Add `TableBuilder::drag_to_scroll`

### LINKED ISSUE #3097: Add `drag_to_scroll` to scroll options of the table

I would like to control the `drag_to_scroll` behavior of the table contents.  Currently, the scroll options assembled by the `TableBuilder` and passed to the `ScrollArea` do not include this option.  That means the `ScrollArea` gets the default behavior, which is `drag_to_scroll = true`, but I may not want that behavior.

I would add the `drag_to_scroll` flag to the table builder and allow the builder to populate this flag.

The only work around at the moment is to disable the scrolling in the table itself, and put the table into a scroll area, where you can control the option.  However, this is not the same, as the header of the table will scroll off the area.

**Additional context**

I can put this on PR assuming it's a reasonable feature to have.

KYovchevski:
Closes #3097

### DIFF
--- a/crates/egui_extras/src/table.rs
+++ b/crates/egui_extras/src/table.rs
@@ -157,6 +157,7 @@ fn to_sizing(columns: &[Column]) -> crate::sizing::Sizing {
 
 struct TableScrollOptions {
     vscroll: bool,
+    drag_to_scroll: bool,
     stick_to_bottom: bool,
     scroll_to_row: Option<(usize, Option<Align>)>,
     scroll_offset_y: Option<f32>,
@@ -169,6 +170,7 @@ impl Default for TableScrollOptions {
     fn default() -> Self {
         Self {
             vscroll: true,
+            drag_to_scroll: true,
             stick_to_bottom: false,
             scroll_to_row: None,
             scroll_offset_y: None,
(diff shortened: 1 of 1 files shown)


vvv:
@KYovchevski If I understand #3097 correctly, you want to be able to set `drag_to_scroll = false`.

> `ScrollArea` gets the default behavior, which is `drag_to_scroll = true`, but I may not want that behavior.

I gave `TableBuilder::drag_to_scroll` a try, but failed to notice any difference from the default behaviour.

I ran 3 versions of the demo app (`cargo run -p egui_demo_app --release`):
1. `master` (9478e50d012c5138551c38cbee16b07bc1fcf283)
2. your branch (e8865d13ad4fcb173423e04e430746adccce970e)
3. your branch with the patch below

```diff
diff --git a/crates/egui_demo_lib/src/demo/table_demo.rs b/crates/egui_demo_lib/src/demo/table_demo.rs
index 8fdea420..ceeadf4a 100644
--- a/crates/egui_demo_lib/src/demo/table_demo.rs
+++ b/crates/egui_demo_lib/src/demo/table_demo.rs
@@ -127,6 +127,7 @@ impl TableDemo {
         let mut table = TableBuilder::new(ui)
             .striped(self.striped)
             .resizable(self.resizable)
+            .drag_to_scroll(false)
             .cell_layout(egui::Layout::left_to_right(egui::Align::Center))
             .column(Column::auto())
             .column(Column::initial(100.0).range(40.0..=300.0))

```

I failed to notice any difference between the cases — I was always able to scroll the table by dragging its body:

https://github.com/emilk/egui/assets/17462/5eaa9a55-1581-4abf-9cbe-354a97335452

@KYovchevski What is the desired effect?

vvv:
@KYovchevski Ah, I understand why it didn't work for me! I used “two fingers scrolling” gesture on MacBook's trackpad. That is equivalent to scrolling with mouse's wheel.

When I use a mouse or click-and-drag properly with a trackpad, **the PR works as intended**.

Sorry for the noise.

emilk (approved):

emilk:

Issue #3106, opened 2023-06-24
title not being used when app_id is not set in NativeOptions


jacekpoz:
**Describe the bug**

according to [the docs](https://docs.rs/eframe/0.22.0/eframe/struct.NativeOptions.html#structfield.app_id), when you don't set an app_id in the NativeOptions it automatically uses the title, in reality it doesn't, I'm making a pull request for it right now

**To Reproduce**
no steps, just code:
```rust
use eframe::egui;

fn main() -> Result<(), eframe::Error> {
    let options = eframe::NativeOptions {
        initial_window_size: Some(egui::vec2(800., 600.)),
        ..Default::default()
    };

    eframe::run_native(
        "this should be app_id but isn't",
        options,
        Box::new(|_cc| Box::<Test>::default()),
    )
}

#[derive(Default)]
struct Test {}

impl eframe::App for Test {
    fn update(&mut self, _ctx: &egui::Context, _frame: &mut eframe::Frame) {}
}
```

**Expected behavior**

the app_id should be equal to the title

**Screenshots**

https://files.catbox.moe/tt0n2k.png

**Desktop (please complete the following information):**
 - OS:  Arch Linux
 - Browser 
 - Version 

**Smartphone (please complete the following information):**
 - Device: 
 - OS: 
 - Browser 
 - Version 

**Additional context**

PR #3107, opened 2023-06-24
eframe: use window title as fallback when app_id is not set

jacekpoz:
Closes <https://github.com/emilk/egui/issues/3106>.


emilk (approved):
Looks good!

jacekpoz:
needs a label

PR #3115, opened 2023-06-27
Make `Table` min/max_scroll_height Optional by default

### LINKED ISSUE #3114: egui_extras Table has default `max_scroll_height` of 800 and must be overridden to fit in taller windows
**Description**

egui_extras::Table has a default `max_scroll_height` of 800.0 which means that Tables placed in any container that is taller than 800px will have their scroll area cut off well before the actual end of the table.

**To Reproduce**
1. Create a CentralPanel within a resizable window (or a window with a default height > 800px)
2. Create a Table, using TableBuilder, fill with data.
3. Observe that the Table scroll area cuts off before the bottom of the Window.

**Additional Context**
See discussion and gifs on #2430.

LeonardMH:
Changes the `TableScrollOptions` to use Optional values for `min_scroll_height` and `max_scroll_height`, requiring the user to call the respective builder functions if they desire those behaviors.

This addresses the surprising effect that by default Tables limit themselves to a maximum height of 800px (#3114).

Closes <https://github.com/emilk/egui/issues/3114>.

### DIFF
--- a/crates/egui_extras/src/table.rs
+++ b/crates/egui_extras/src/table.rs
@@ -160,8 +160,8 @@ struct TableScrollOptions {
     stick_to_bottom: bool,
     scroll_to_row: Option<(usize, Option<Align>)>,
     scroll_offset_y: Option<f32>,
-    min_scrolled_height: f32,
-    max_scroll_height: f32,
+    min_scrolled_height: Option<f32>,
+    max_scroll_height: Option<f32>,
     auto_shrink: [bool; 2],
 }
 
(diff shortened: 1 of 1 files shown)


LeonardMH:
I saw the contributing guideline to not open the PR from my `master` branch and assumed I would be able to edit it after opening the PR but I guess I can't. I have created a separate branch `bugfix/tables-have-bounded-height-by-default` in my fork if you'd rather work from that.

LeonardMH:
Need to update the relevant documentation / comments.

emilk:
This will match how `ScrollArea` works, so I guess it makes sense. But no need for `Option`, just use `f32::INFINITY` instead, just like `ScrollArea::max_height` does

emilk:
Check the docs for `ScrollArea` to see a bit more about why we have these settings.

emilk:
Converted to draft awaiting progress

emilk:
Closing: the default `max_scroll_height` was changed to `f32::INFINITY` in #4817, so tables are no longer capped at 800 points by default. Thanks for the PR!

PR #3118, opened 2023-06-28
Add `Window::drag-to-scroll`

KYovchevski:
Similar to #3100, exposes the underlying `Window`'s underlying `ScrollArea`'s `drag_to_scroll` field.


emilk (approved):
Looks good!

PR #3119, opened 2023-06-28
Return `TableOutput` struct from `TableBuilder::body()`

KYovchevski:
Currently it appears to be impossible, or very difficult, to receive information about the table that will be drawn to the window. This PR adds a barebones solution to this by returning a `TableOutput` struct from `fn body()` which describes the basic details about the header and the content body of the table.

I'm ok with spending the extra time to expand this further to cover more of the data about the table, such as being able to return values from the header and body functions, or being able to specify the sense on them. However, I want to verify this is the preferred way of doing this beforehand.

### DIFF
--- a/crates/egui_extras/src/table.rs
+++ b/crates/egui_extras/src/table.rs
@@ -539,6 +539,7 @@ pub struct Table<'a> {
     cell_layout: egui::Layout,
 
     scroll_options: TableScrollOptions,
+
     /// The `header_response` is passed down from [`TableBuilder::header()`] so that it can be returned to the user after the body is built.
     /// If no header is specified, it is `None`.
     header_response: Option<Response>,
--- a/crates/egui_extras/src/table.rs
+++ b/crates/egui_extras/src/table.rs
@@ -3,7 +3,9 @@
 //! | fixed size | all available space/minimum | 30% of available width | fixed size |
 //! Takes all available height, so if you want something below the table, put it in a strip.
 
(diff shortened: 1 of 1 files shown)


emilk (approved):
Yes, this is a good start!

emilk:

KYovchevski:
> I'm not sure why the CI isn't kicking in. Perhaps try merging in latest `master` and see if it helps?

I just finished a rebase on top of master, and it seems to require approval from a maintainer to run the CI.

emilk (approved):

crates/egui_extras/src/table.rs
```diff
@@ -519,6 +525,18 @@ pub struct Table<'a> {
     cell_layout: egui::Layout,
 
     scroll_options: TableScrollOptions,
+    /// The `header_response` is passed down from [`TableBuilder::header()`] so that it can be returned to the user after the body is built.
```
emilk, line 528:
```suggestion

    /// The `header_response` is passed down from [`TableBuilder::header()`] so that it can be returned to the user after the body is built.
```

emilk:
Looks good, except for the missing newline

emilk:
bump! I cannot commit my suggested change

emilk:
also needs merge of `master`

emilk:
Closing: `Table::body` returns `ScrollAreaOutput` since #4829. If you still need the header response, please create a new PR for just that. Thanks for the PR, and sorry for the slow response

Issue #3112, opened 2023-06-26
How do i roate a plot image


ThundR67:
I want to rotate a plot image but can't find an option to do so. It only works on normal image.

Issue #1646, opened 2022-05-19
Rotate and Flip PlotImage.


kkngsm:
Hi, I would like the ability to rotate and flip `PlotImage`.

Usually using `Vec<_>.push()`, if I want to append to the image data, it will be  append to bottom.
However, in some cases (such as drawing a spectrogram) I may want to append to the left or right.
Therefore, we need the ability to rotate or flip the image to make it appear as if it is being added to the left or right.

The function is like `widget::Image`
```rust
PlotImage::new(/*inline*/)
    .rotate(angle, origin)
    .flip([true, false])
```

With the current functionality, the only way to do this is to `Vec<Vec<u8>>`, but this is inefficient.

ThundR67:
How do i rotate plot images for the time being?

PR #3121, opened 2023-06-28
Add rotation feature to plot images

ThundR67:
Add roation to plot images
Closes https://github.com/emilk/egui/issues/3112 
Closes https://github.com/emilk/egui/issues/1646


emilk (approved):

emilk:

PR #3122, opened 2023-06-28
Fix the "ever-growing height" problem of Strip and Table demos

### LINKED ISSUE #3029: Table Demo and Strip Demo window height grow when mouse cursor moves
Even if I refresh my browser, the egui.rs demo page windows for the Table Demo and Strip Demo grow on their own when the mouse cursor moves. Using Latest Firefox 113.0.2 64-bit on an M1 running MacOS Ventura 13.4.

vvv:
Problem
-------

The height of "Table Demo" or "Strip Demo" floating window grows
when the demo app is interacted with. If 'Continuous' mode is enabled
in 'Backend' settings, the window grows irrespectively of user interaction.

Observations
------------

I noticed that [`area_content_ui.min_rect().max.y`][1] is increasing
monotonically with speed 0.5 px/frame.

I also noticed that commenting out `ui.add(crate::egui_github_link_file!());`
[statement][2] in `table_demo.rs` makes the problem disappear.

The "Fix"
---------

I added 0.5 to the height of the row with GitHub link.
This solved the problem.

Before:

https://github.com/emilk/egui/assets/17462/1cb100b8-5fc8-4233-87eb-685d8a78650b

After:

https://github.com/emilk/egui/assets/17462/bd373d30-4775-4fa3-b8f7-425e13c6993c

Closes #3029.

Warning
-------

I failed to find the root cause of the problem.
I don't understand why this change makes the problem disappear.

[1]: https://github.com/emilk/egui/blob/9478e50d012c5138551c38cbee16b07bc1fcf283/crates/egui/src/containers/window.rs#L403
[2]: https://github.com/emilk/egui/blob/9478e50d012c5138551c38cbee16b07bc1fcf283/crates/egui_demo_lib/src/demo/table_demo.rs#L114

### DIFF
--- a/crates/egui/src/containers/area.rs
+++ b/crates/egui/src/containers/area.rs
@@ -426,7 +426,7 @@ impl Prepared {
             temporarily_invisible: _,
         } = self;
 
-        state.size = content_ui.min_rect().size();
+        state.size = content_ui.min_size();
 
         ctx.memory_mut(|m| m.areas.set_state(layer_id, state));
 
--- a/crates/egui_demo_lib/src/demo/strip_demo.rs
+++ b/crates/egui_demo_lib/src/demo/strip_demo.rs
@@ -37,7 +37,7 @@ impl super::View for StripDemo {
             .size(Size::exact(50.0))
             .size(Size::remainder())
             .size(Size::relative(0.5).at_least(60.0))
-            .size(Size::exact(10.0))
+            .size(Size::exact(10.5))
             .vertical(|mut strip| {
                 strip.cell(|ui| {
                     ui.painter().rect_filled(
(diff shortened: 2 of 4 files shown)


vvv:
@rustrust [wrote](https://github.com/emilk/egui/issues/3029#issuecomment-1616009939):

> I am unclear from the description, was this an issue only with the demo itself or was there an underlying problem with the egui library itself? I see that area.rs and rect.rs were modified.

The problem is with the demo app, not the library.

This PR comprises
- the fix (https://github.com/emilk/egui/pull/3122/commits/85c269852e3eb4d5f26f823b49845b603ac70d15) and 
- decorative changes (https://github.com/emilk/egui/pull/3122/commits/de5fe83f2a4c9ce3cf0a3ffd6dee41036564bca8) that are not related to #3122.

I don't mind dropping the “decorative changes” commit. Should I?

emilk (approved):
Thanks for the fix. It would be nicer with a root-cause fix, but this is better than nothing :)

emilk:

vvv:
Rebased onto the latest master.

vvv:
> I'm not sure why the CI isn't kicking in. Perhaps try merging in latest `master` and see if it helps?

@emilk If I understand [this workflow file](https://github.com/emilk/egui/blob/af6419343d8e864e0c9afaa559363315abb55424/.github/workflows/labels.yml) correctly, at least one label is required for CI to kick in.

PR #3138, opened 2023-07-07
egui: Plot Arrows add method to specify tip_size

nagua:
Hello everyone,

For my application I want the plot arrows to always have the same size and currently the tip size
depends on the length of the arrow. So this patch adds a method to explicitly set the arrow tip size.

If this needs more discussion, I can additionally open an issue.

Have a nice day everyone!

### DIFF
--- a/crates/egui/src/widgets/plot/items/mod.rs
+++ b/crates/egui/src/widgets/plot/items/mod.rs
@@ -997,6 +997,7 @@ impl PlotItem for Points {
 pub struct Arrows {
     pub(super) origins: PlotPoints,
     pub(super) tips: PlotPoints,
+    pub(super) tip_length: Option<f32>,
     pub(super) color: Color32,
     pub(super) name: String,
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/plot/items/mod.rs
```diff
@@ -1019,6 +1021,12 @@ impl Arrows {
         self
     }
 
+    /// Set the size of the arrow tips
+    pub fn tip_size(mut self, tip_size: f32) -> Self {
+        self.tip_size = Some(tip_size);
```
emilk, line 1026: "size" is not very clear. Is it with width or the length of the tip? Calling it `tip_length` would make more sense imho

nagua: Ok, I'm fine with "length". I already changed the name.

emilk (approved):

PR #3161, opened 2023-07-14
🧢 Expose `area::State`'s `Rect` in Memory

tosti007:
Currently, there is no way of obtaining any information on windows from `Memory`, even though this is in fact stored in there. This PR exposes the `area::State` struct as public and adds a function to obtain said `State` for a window.

### DIFF
--- a/crates/egui/src/containers/window.rs
+++ b/crates/egui/src/containers/window.rs
@@ -21,6 +21,9 @@ use super::*;
 ///    ui.label("Hello World!");
 /// });
 /// # });
+/// ```
+///
+/// The previous rectangle used by this window can be obtained through [`crate::Memory::area_rect()`].
 #[must_use = "You should call .show()"]
 pub struct Window<'open> {
     title: WidgetText,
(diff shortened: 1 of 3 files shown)


crates/egui/src/memory.rs
```diff
@@ -487,6 +487,11 @@ impl Memory {
     pub fn reset_areas(&mut self) {
         self.areas = Default::default();
     }
+
+    /// Obtain the previous area rect of a window.
+    pub fn window_rect(&self, id: impl Into<Id>) -> Option<Rect> {
```
emilk, line 492: It would be great to link to this function from the docs of `Window` and `Area`

emilk (approved):

crates/egui/src/memory.rs
tosti007: Done!

tosti007:
I added the doc-links on `Window` and `Area`. I renamed `window_rect` to `area_rect` as it's actually generic for all areas.

crates/egui/src/containers/window.rs
```diff
@@ -21,6 +21,9 @@ use super::*;
 ///    ui.label("Hello World!");
 /// });
 /// # });
+/// ```
```
tosti007, line 24: Also did a drive-by fix here.

emilk (approved):

PR #3166, opened 2023-07-17
epaint: Add `ColorImage::from_gray`

thomaseliot:
This PR adds a new function `from_gray_unmultiplied()` . This allows programs working with grayscale source images to load the images directly as grayscale. 

Previously, an application working with grayscale images would have load the image as rgba. By providing a way to load images in grayscale, the size of the copy is reduced by 4x and the caller no longer needs to perform a conversion from grayscale to rgba. This has resulted in significant performance improvements in my application. It seems likely that anyone working with grayscale source imagery in the future would find this functionality useful.

### DIFF
--- a/crates/epaint/src/image.rs
+++ b/crates/epaint/src/image.rs
@@ -110,10 +110,10 @@ impl ColorImage {
         Self { size, pixels }
     }
 
-    /// Create a [`ColorImage`] from flat un-multiplied gray data.
+    /// Create a [`ColorImage`] from flat opaque gray data.
     ///
     /// Panics if `size[0] * size[1] != gray.len()`.
-    pub fn from_gray_unmultiplied(size: [usize; 2], gray: &[u8]) -> Self {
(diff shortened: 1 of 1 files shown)


crates/epaint/src/image.rs
```diff
@@ -110,6 +110,15 @@ impl ColorImage {
         Self { size, pixels }
     }
 
+    /// Create a [`ColorImage`] from flat un-multiplied gray data.
+    ///
+    /// Panics if `size[0] * size[1] != gray.len()`.
+    pub fn from_gray_unmultiplied(size: [usize; 2], gray: &[u8]) -> Self {
```
emilk, line 116: Since there is no alpha-channel here, the whole `unmultiplied` thing can be skipped:

```suggestion
    /// Create a [`ColorImage`] from flat opaque gray data.
    ///
    /// Panics if `size[0] * size[1] != gray.len()`.
    pub fn from_gray(size: [usize; 2], gray: &[u8]) -> Self {
```

emilk (approved):

PR #3167, opened 2023-07-18
Fix panic with persistence without window

sagebind:
A window may not always be available and may have already been closed by the time an eframe app is closing. An example of this is Android, where the main activity window may have been stopped or discarded because the app is no longer in the foreground, and then the user decides to close your app without resuming it using the multitasking view.

In this case, skip the window persistence step if it does not exist anymore by the time we are saving the persistence data. Currently eframe will panic with `winit window doesn't exist` when using the glow backend.

This also fixes an inconsistency between the wgpu and glow backends, where the wgpu backend would skip saving data in the persistence store entirely if a window does not exist, rather than just persisting everything but the window information.


emilk (approved):
Thanks!

PR #3171, opened 2023-07-22
Fix auto_bounds when only one axis has restricted navigation

KoffeinFlummi:
This fixes an issue with plots where the auto-bounds feature for the y-axis would not work if navigation in the y-axis was restricted, but moving the x-axis was allowed. Any dragging or scrolling would lead to the auto-bounding being disabled, meaning the y-axis was not rescaled when new data came into view.

If that explanation makes no sense:

before: https://i.imgur.com/2bwRFNr.mp4
after: https://i.imgur.com/5TtD9fp.mp4

The solution is to only consider an axis to be modified if modifying that axis is actually allowed.

### DIFF
--- a/crates/egui/src/widgets/plot/mod.rs
+++ b/crates/egui/src/widgets/plot/mod.rs
@@ -864,7 +864,7 @@ impl Plot {
                 delta.y = 0.0;
             }
             transform.translate_bounds(delta);
-            bounds_modified = true.into();
+            bounds_modified = allow_drag;
         }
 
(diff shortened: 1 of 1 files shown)


emilk (approved):

KoffeinFlummi:
Thanks for merging.

> Though your links are broken. You can just drag-drop mp4 videos into a GitHub PR description next time

Yeah, looks like those links expired. I'll keep that in mind for next time.

PR #3172, opened 2023-07-23
Multiple viewports/windows

### LINKED ISSUE #1044: Multiple native windows
We should add an example `egui_glow/examples/multiple_windows.rs` that opens multiple native `winit` windows and runs one egui instance is each.

Probably requires some improvements to the `egui_glow` API.

konkitoman:
* Closes #1044

---
(new PR description written by @emilk)

## Overview
This PR introduces the concept of `Viewports`, which on the native eframe backend corresponds to native OS windows.

You can spawn a new viewport using `Context::show_viewport` and `Cotext::show_viewport_immediate`.
These needs to be called every frame the viewport should be visible.

This is implemented by the native `eframe` backend, but not the web one.

## Viewport classes
The viewports form a tree of parent-child relationships.

There are different classes of viewports.

### Root vieport
The root viewport is the original viewport, and cannot be closed without closing the application.

### Deferred viewports
These are created with `Context::show_viewport`.
Deferred viewports take a closure that is called by the integration at a later time, perhaps multiple times.
Deferred viewports are repainted independenantly of the parent viewport.
This means communication with them need to done via channels, or `Arc/Mutex`.

This is the most performant type of child viewport, though a bit more cumbersome to work with compared to immediate viewports.

### Immediate viewports
These are created with `Context::show_viewport_immediate`.
Immediate viewports take a `FnOnce` closure, similar to other egui functions, and is called immediately. This makes communication with them much simpler than with deferred viewports, but this simplicity comes at a cost: whenever tha parent viewports needs to be repainted, so will the child viewport, and vice versa. This means that if you have `N` viewports you are poentially doing `N` times as much CPU work. However, if all your viewports are showing animations, and thus are repainting constantly anyway, this doesn't matter.

In short: immediate viewports are simpler to use, but can waste a lot of CPU time.

### Embedded viewports
These are not real, independenant viewports, but is a fallback mode for when the integration does not support real viewports. In your callback is called with `ViewportClass::Embedded` it means you need to create an `egui::Window` to wrap your ui in, which will then be embedded in the parent viewport, unable to escape it.

## Using the viewports
Only one viewport is active at any one time, identified wth `Context::viewport_id`.
You can send commands to other viewports using `Context::send_viewport_command_to`.

There is an example in <https://github.com/emilk/egui/tree/master/examples/multiple_viewports/src/main.rs>.

## For integrations
There are several changes relevant to integrations.

* There is a [`crate::RawInput::viewport`] with information about the current viewport.
* The repaint callback set by `Context::set_request_repaint_callback` now points to which viewport should be repainted.
* `Context::run` now returns a list of viewports in `FullOutput` which should result in their own independant windows
* There is a new `Context::set_immediate_viewport_renderer` for setting up the immediate viewport integration
* If you support viewports, you need to call `Context::set_embed_viewports(false)`, or all new viewports will be embedded (the default behavior).

## Future work
* Make it easy to wrap child viewports in the same chrome as `egui::Window`
* Automatically show embedded viewports using `egui::Window`
* Use the new `ViewportBuilder` in `eframe::NativeOptions`
* Automatically position new viewport windows (they currently cover each other)
* Add a `Context` method for listing all existing viewports

Find more at https://github.com/emilk/egui/issues/3556

---

<details>
<summary>
Outdated PR description by @konkitoman
</summary>

## Inspiration
- Godot because the app always work desktop or single_window because of embedding
- Dear ImGui viewport system

## What is a Viewport

A Viewport is a egui isolated component!
Can be used by the egui integration to create native windows!

When you create a Viewport is possible that the backend do not supports that!
So you need to check if the Viewport was created or you are in the normal egui context!
This is how you can do that:
```rust
if ctx.viewport_id() != ctx.parent_viewport_id() {
    // In here you add the code for the viewport context, like
    egui::CentralPanel::default().show(ctx, |ui|{
        ui.label("This is in a native window!");
    });
}else{
    // In here you add the code for when viewport cannot be created!
   // You cannot use CentralPanel in here because you will override the app CentralPanel
   egui::Window::new("Virtual Viewport").show(ctx, |ui|{
       ui.label("This is without a native window!\nThis is in a embedded viewport");
   });
}
```

This PR do not support for drag and drop between Viewports!

After this PR is accepted i will begin work to intregrate the Viewport system in `egui::Window`!
The `egui::Window` i want to behave the same on desktop and web
The `egui::Window` will be like Godot Window

## Changes and new

These are only public structs and functions!

<details>
<summary>

## New
</summary>

- `egui::ViewportId`
- `egui::ViewportBuilder`
This is like winit WindowBuilder

- `egui::ViewportCommand`
With this you can set any winit property on a viewport, when is a native window!

- `egui::Context::new`
- `egui::Context::create_viewport`
- `egui::Context::create_viewport_sync`
- `egui::Context::viewport_id`
- `egui::Context::parent_viewport_id`
- `egui::Context::viewport_id_pair`
- `egui::Context::set_render_sync_callback`
- `egui::Context::is_desktop`
- `egui::Context::force_embedding`
- `egui::Context::set_force_embedding`
- `egui::Context::viewport_command`
- `egui::Context::send_viewport_command_to`
- `egui::Context::input_for`
- `egui::Context::input_mut_for`
- `egui::Context::frame_nr_for`
- `egui::Context::request_repaint_for`
- `egui::Context::request_repaint_after_for`
- `egui::Context::requested_repaint_last_frame`
- `egui::Context::requested_repaint_last_frame_for`
- `egui::Context::requested_repaint`
- `egui::Context::requested_repaint_for`
- `egui::Context::inner_rect`
- `egui::Context::outer_rect`

- `egui::InputState::inner_rect`
- `egui::InputState::outer_rect`

- `egui::WindowEvent`

</details>

<details>
<summary>

## Changes
</summary>

- `egui::Context::run`
Now needs the viewport that we want to render!

- `egui::Context::begin_frame`
Now needs the viewport that we want to render!

- `egui::Context::tessellate`
Now needs the viewport that we want to render!

- `egui::FullOutput`
```diff
- repaint_after
+ viewports
+ viewport_commands
```

- `egui::RawInput`
```diff
+ inner_rect
+ outer_rect
```

- `egui::Event`
```diff
+ WindowEvent
```
</details>

### Async Viewport

Async means that is independent from other viewports!

Is created by `egui::Context::create_viewport`

To be used you will need to wrap your state in `Arc<RwLock<T>>`
Look at viewports example to understand how to use it!

### Sync Viewport

Sync means that is dependent on his parent!

Is created by `egui::Context::create_viewport_sync`

This will pause the parent then render itself the resumes his parent!

#### :warning: Attention

You will need to do this when you render your content
```rust
ctx.create_viewport(ViewportBuilder::new("Simple Viewport"), | ctx | {
    let content = |ui: &mut egui::Ui|{
        ui.label("Content");
    };

    // This will make the content a popup if cannot create a native window
    if ctx.viewport_id() != ctx.parent_viewport_id() {
        egui::CentralPanel::default().show(ctx, content);
    } else {
        egui::Area::new("Simple Viewport").show(ctx, |ui| {
            egui::Frame::popup(ui.style()).show(ui, content);
        });
    };
});
````

### If you are using eframe

You don't need to change anything!

### If you have a manual implementation

Now `egui::run` or `egui::begin` and `egui::tessellate` will need the current viewport id!
You cannot create a `ViewportId` only `ViewportId::MAIN`

If you make a single window app you will set the viewport id to be `egui::ViewportId::MAIN` or see the `examples/pure_glow`
If you want to have multiples window support look at `crates/eframe` glow or wgpu implementations!

## If you want to try this

- cargo run -p viewports

## This before was wanted to change

This will probably be in feature PR's

### egui::Window

To create a native window when embedded was set to false
You can try that in viewports example before: [78a0ae8](https://github.com/emilk/egui/pull/3172/commits/78a0ae879e68621543b9ca700ece3d9456eee8a4)

### egui popups, context_menu, tooltip

To be a native window

</details>

### DIFF
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -1,4 +1,4 @@
-use std::{sync::Arc, time::Instant};
+use std::time::Instant;
 
 use winit::event_loop::EventLoopWindowTarget;
 
@@ -92,16 +92,18 @@ pub fn window_builder<E>(
         .with_maximized(*maximized)
         .with_resizable(*resizable)
         .with_transparent(*transparent)
-        .with_window_icon(icon_data.clone().map(|d| {
-            Arc::new(egui::ColorImage::from_rgba_premultiplied(
-                [d.width as usize, d.height as usize],
-                &d.rgba,
-            ))
-        }))
         .with_active(*active)
         // Keep hidden until we've painted something. See https://github.com/emilk/egui/pull/2279
         // We must also keep the window hidden until AccessKit is initialized.
         .with_visible(false);
+
+    if let Some(icon_data) = icon_data {
+        viewport_builder =
+            viewport_builder.with_window_icon(egui::ColorImage::from_rgba_premultiplied(
+                [icon_data.width as usize, icon_data.height as usize],
+                &icon_data.rgba,
+            ));
+    }
 
     #[cfg(target_os = "macos")]
     if *fullsize_content {
@@ -120,10 +122,10 @@ pub fn window_builder<E>(
     }
 
     if let Some(min_size) = *min_window_size {
-        viewport_builder = viewport_builder.with_min_inner_size(Some(min_size));
+        viewport_builder = viewport_builder.with_min_inner_size(min_size);
     }
     if let Some(max_size) = *max_window_size {
-        viewport_builder = viewport_builder.with_max_inner_size(Some(max_size));
+        viewport_builder = viewport_builder.with_max_inner_size(max_size);
     }
 
     viewport_builder = viewport_builder.with_drag_and_drop(*drag_and_drop_support);
@@ -141,13 +143,13 @@ pub fn window_builder<E>(
         window_settings.inner_size_points()
     } else {
         if let Some(pos) = *initial_window_pos {
-            viewport_builder = viewport_builder.with_position(Some(pos));
+            viewport_builder = viewport_builder.with_position(pos);
         }
 
         if let Some(initial_window_size) = *initial_window_size {
             let initial_window_size =
                 initial_window_size.at_most(largest_monitor_point_size(event_loop));
-            viewport_builder = viewport_builder.with_inner_size(Some(initial_window_size));
+            viewport_builder = viewport_builder.with_inner_size(initial_window_size);
         }
 
         *initial_window_size
@@ -161,7 +163,7 @@ pub fn window_builder<E>(
             if monitor_size.width > 0.0 && monitor_size.height > 0.0 {
                 let x = (monitor_size.width - inner_size.x) / 2.0;
                 let y = (monitor_size.height - inner_size.y) / 2.0;
-                viewport_builder = viewport_builder.with_position(Some(egui::Pos2::new(x, y)));
+                viewport_builder = viewport_builder.with_position([x, y]);
             }
         }
     }
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -4,7 +4,7 @@ use winit::event_loop::EventLoopWindowTarget;
 
 use raw_window_handle::{HasRawDisplayHandle as _, HasRawWindowHandle as _};
 
-use egui::{NumExt as _, ViewportBuilder, ViewportId, ViewportIdPair, ViewportUiCallback};
+use egui::{DeferredViewportUiCallback, NumExt as _, ViewportBuilder, ViewportId, ViewportIdPair};
 use egui_winit::{native_pixels_per_point, EventResponse, WindowSettings};
 
 use crate::{epi, Theme, WindowInfo};
@@ -501,7 +501,7 @@ impl EpiIntegration {
     pub fn update(
         &mut self,
         app: &mut dyn epi::App,
-        viewport_ui_cb: Option<&ViewportUiCallback>,
+        viewport_ui_cb: Option<&DeferredViewportUiCallback>,
         mut raw_input: egui::RawInput,
     ) -> egui::FullOutput {
         raw_input.time = Some(self.beginning.elapsed().as_secs_f64());
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -85,7 +85,7 @@ pub fn window_builder<E>(
         ..
     } = native_options;
 
-    let mut viewport_builder = egui::ViewportBuilder::ROOT
+    let mut viewport_builder = egui::ViewportBuilder::DEFAULTS
         .with_title(title)
         .with_decorations(*decorated)
         .with_fullscreen(*fullscreen)
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -85,8 +85,7 @@ pub fn window_builder<E>(
         ..
     } = native_options;
 
-    let mut viewport_builder = egui::ViewportBuilder::ROOT;
-    viewport_builder
+    let mut viewport_builder = egui::ViewportBuilder::ROOT
         .with_title(title)
         .with_decorations(*decorated)
         .with_fullscreen(*fullscreen)
@@ -106,7 +105,7 @@ pub fn window_builder<E>(
 
     #[cfg(target_os = "macos")]
     if *fullsize_content {
-        viewport_builder
+        viewport_builder = viewport_builder
             .with_title_hidden(true)
             .with_titlebar_transparent(true)
             .with_fullsize_content_view(true);
@@ -114,20 +113,20 @@ pub fn window_builder<E>(
 
     #[cfg(all(feature = "wayland", target_os = "linux"))]
     {
-        match &native_options.app_id {
+        viewport_builder = match &native_options.app_id {
             Some(app_id) => viewport_builder.with_name(app_id, ""),
             None => viewport_builder.with_name(title, ""),
         };
     }
 
     if let Some(min_size) = *min_window_size {
-        viewport_builder.with_min_inner_size(Some(min_size));
+        viewport_builder = viewport_builder.with_min_inner_size(Some(min_size));
     }
     if let Some(max_size) = *max_window_size {
-        viewport_builder.with_max_inner_size(Some(max_size));
+        viewport_builder = viewport_builder.with_max_inner_size(Some(max_size));
     }
 
-    viewport_builder.with_drag_and_drop(*drag_and_drop_support);
+    viewport_builder = viewport_builder.with_drag_and_drop(*drag_and_drop_support);
 
     // Always use the default window size / position on iOS. Trying to restore the previous position
     // causes the window to be shown too small.
@@ -138,17 +137,17 @@ pub fn window_builder<E>(
         window_settings.clamp_size_to_sane_values(largest_monitor_point_size(event_loop));
         window_settings.clamp_position_to_monitors(event_loop);
 
-        window_settings.initialize_viewport_builder(&mut viewport_builder);
+        viewport_builder = window_settings.initialize_viewport_builder(viewport_builder);
         window_settings.inner_size_points()
     } else {
         if let Some(pos) = *initial_window_pos {
-            viewport_builder.with_position(Some(pos));
+            viewport_builder = viewport_builder.with_position(Some(pos));
         }
 
         if let Some(initial_window_size) = *initial_window_size {
             let initial_window_size =
                 initial_window_size.at_most(largest_monitor_point_size(event_loop));
-            viewport_builder.with_inner_size(Some(initial_window_size));
+            viewport_builder = viewport_builder.with_inner_size(Some(initial_window_size));
         }
 
         *initial_window_size
@@ -162,7 +161,7 @@ pub fn window_builder<E>(
             if monitor_size.width > 0.0 && monitor_size.height > 0.0 {
                 let x = (monitor_size.width - inner_size.x) / 2.0;
                 let y = (monitor_size.height - inner_size.y) / 2.0;
-                viewport_builder.with_position(Some(egui::Pos2::new(x, y)));
+                viewport_builder = viewport_builder.with_position(Some(egui::Pos2::new(x, y)));
             }
         }
     }
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -85,10 +85,9 @@ pub fn window_builder<E>(
         ..
     } = native_options;
 
-    let mut viewport_builder = egui::ViewportBuilder::new(ViewportId::ROOT);
+    let mut viewport_builder = egui::ViewportBuilder::ROOT;
     viewport_builder
         .with_title(title)
-        .with_close_button(true) // The default for all other viewports is `false`!
         .with_decorations(*decorated)
         .with_fullscreen(*fullscreen)
         .with_maximized(*maximized)
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -495,7 +495,7 @@ impl EpiIntegration {
 
     /// Run user code - this can create immediate viewports, so hold no locks over this!
     ///
-    /// If `viewport_ui_cb` is None, we are in the root viewport and will call [`App::update`].
+    /// If `viewport_ui_cb` is None, we are in the root viewport and will call [`crate::App::update`].
     pub fn update(
         &mut self,
         app: &mut dyn epi::App,
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -407,6 +407,8 @@ impl EpiIntegration {
         window: &winit::window::Window,
         event_loop_proxy: winit::event_loop::EventLoopProxy<E>,
     ) {
+        crate::profile_function!();
+
         let egui_ctx = self.egui_ctx.clone();
         egui_winit.init_accesskit(window, event_loop_proxy, move || {
             // This function is called when an accessibility client
@@ -549,14 +551,15 @@ impl EpiIntegration {
     }
 
     pub fn post_rendering(&mut self, app: &mut dyn epi::App, window: &winit::window::Window) {
+        crate::profile_function!();
         let inner_size = window.inner_size();
         let window_size_px = [inner_size.width, inner_size.height];
-
         app.post_rendering(window_size_px, &self.frame);
     }
 
     pub fn post_present(&mut self, window: &winit::window::Window) {
         if let Some(visible) = self.frame.output.visible.take() {
+            crate::profile_scope!("window.set_visible");
             window.set_visible(visible);
         }
     }
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -85,7 +85,7 @@ pub fn window_builder<E>(
         ..
     } = native_options;
 
-    let mut viewport_builder = egui::ViewportBuilder::DEFAULTS
+    let mut viewport_builder = egui::ViewportBuilder::default()
         .with_title(title)
         .with_decorations(*decorated)
         .with_fullscreen(*fullscreen)
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -507,13 +507,14 @@ mod glow_integration {
             crate::profile_scope!("frame");
 
             {
-                // This will only happen if the viewport is sync
-                // That means that the viewport cannot be rendered by itself and needs his parent to be rendered
                 let glutin = self.glutin.borrow();
                 let viewport = &glutin.viewports[&viewport_id];
-                if viewport.viewport_ui_cb.is_none() && viewport_id != ViewportId::ROOT {
+                let is_immediate = viewport.viewport_ui_cb.is_none();
+                if is_immediate && viewport_id != ViewportId::ROOT {
                     if let Some(parent_viewport) = glutin.viewports.get(&viewport.id_pair.parent) {
                         if let Some(window) = parent_viewport.window.as_ref() {
+                            // This will only happen if this is an immediate viewport.
+                            // That means that the viewport cannot be rendered by itself and needs his parent to be rendered.
                             return EventResult::RepaintNext(window.id());
                         }
                     }
@@ -701,6 +702,7 @@ mod glow_integration {
         id_pair: ViewportIdPair,
 
         /// The user-callback that shows the ui.
+        /// None for immediate viewports.
         viewport_ui_cb: Option<Arc<ViewportUiCallback>>,
 
         egui_winit: Option<egui_winit::State>,
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -457,8 +457,8 @@ mod glow_integration {
     };
 
     use egui::{
-        epaint::ahash::HashMap, ImmediateViewport, NumExt as _, ViewportIdMap, ViewportIdPair,
-        ViewportIdSet, ViewportOutput, ViewportUiCallback,
+        epaint::ahash::HashMap, DeferredViewportUiCallback, ImmediateViewport, NumExt as _,
+        ViewportClass, ViewportIdMap, ViewportIdPair, ViewportIdSet, ViewportOutput,
     };
     use egui_winit::{create_winit_window_builder, process_viewport_commands, EventResponse};
 
@@ -781,11 +781,12 @@ mod glow_integration {
 
     struct Viewport {
         ids: ViewportIdPair,
+        class: ViewportClass,
         builder: ViewportBuilder,
 
         /// The user-callback that shows the ui.
         /// None for immediate viewports.
-        viewport_ui_cb: Option<Arc<ViewportUiCallback>>,
+        viewport_ui_cb: Option<Arc<DeferredViewportUiCallback>>,
 
         gl_surface: Option<glutin::surface::Surface<glutin::surface::WindowSurface>>,
         window: Option<Rc<Window>>,
@@ -944,6 +945,7 @@ mod glow_integration {
                 ViewportId::ROOT,
                 Viewport {
                     ids: ViewportIdPair::ROOT,
+                    class: ViewportClass::Root,
                     builder: viewport_builder,
                     viewport_ui_cb: None,
                     gl_surface: None,
@@ -1166,6 +1168,7 @@ mod glow_integration {
                 viewport_id,
                 ViewportOutput {
                     parent,
+                    class,
                     builder,
                     viewport_ui_cb,
                     commands,
@@ -1178,6 +1181,7 @@ mod glow_integration {
                 initialize_or_update_viewport(
                     &mut self.viewports,
                     ids,
+                    class,
                     builder,
                     viewport_ui_cb,
                     focused_viewport,
@@ -1208,6 +1212,7 @@ mod glow_integration {
     fn initialize_or_update_viewport(
         viewports: &mut ViewportIdMap<Viewport>,
         ids: ViewportIdPair,
+        class: ViewportClass,
         mut builder: ViewportBuilder,
         viewport_ui_cb: Option<Arc<dyn Fn(&egui::Context) + Send + Sync>>,
         focused_viewport: Option<ViewportId>,
@@ -1227,6 +1232,7 @@ mod glow_integration {
                 log::debug!("Creating new viewport {:?} ({:?})", ids.this, builder.title);
                 entry.insert(Viewport {
                     ids,
+                    class,
                     builder,
                     viewport_ui_cb,
                     window: None,
@@ -1240,6 +1246,7 @@ mod glow_integration {
                 let viewport = entry.get_mut();
 
                 viewport.ids.parent = ids.parent;
+                viewport.class = class;
                 viewport.viewport_ui_cb = viewport_ui_cb;
 
                 let (delta_commands, recreate) = viewport.builder.patch(&builder);
@@ -1519,8 +1526,14 @@ mod glow_integration {
         {
             let mut glutin = glutin.borrow_mut();
 
-            let viewport =
-                initialize_or_update_viewport(&mut glutin.viewports, ids, builder, None, None);
+            let viewport = initialize_or_update_viewport(
+                &mut glutin.viewports,
+                ids,
+                ViewportClass::Immediate,
+                builder,
+                None,
+                None,
+            );
 
             if viewport.gl_surface.is_none() {
                 glutin.init_viewport(ids.this, event_loop).expect(
@@ -1799,8 +1812,8 @@ mod wgpu_integration {
     use parking_lot::Mutex;
 
     use egui::{
-        FullOutput, ImmediateViewport, ViewportIdMap, ViewportIdPair, ViewportIdSet,
-        ViewportOutput, ViewportUiCallback,
+        DeferredViewportUiCallback, FullOutput, ImmediateViewport, ViewportClass, ViewportIdMap,
+        ViewportIdPair, ViewportIdSet, ViewportOutput,
     };
     use egui_winit::{create_winit_window_builder, process_viewport_commands};
 
@@ -1809,10 +1822,12 @@ mod wgpu_integration {
     pub struct Viewport {
         ids: ViewportIdPair,
 
+        class: ViewportClass,
+
         builder: ViewportBuilder,
 
         /// `None` for sync viewports.
-        viewport_ui_cb: Option<Arc<ViewportUiCallback>>,
+        viewport_ui_cb: Option<Arc<DeferredViewportUiCallback>>,
 
         /// Window surface state that's initialized when the app starts running via a Resumed event
         /// and on Android will also be destroyed if the application is paused.
@@ -2049,6 +2064,7 @@ mod wgpu_integration {
                 ViewportId::ROOT,
                 Viewport {
                     ids: ViewportIdPair::ROOT,
+                    class: ViewportClass::Root,
                     builder,
                     viewport_ui_cb: None,
                     window: Some(Rc::new(window)),
@@ -2139,7 +2155,14 @@ mod wgpu_integration {
                 painter,
                 viewport_from_window,
             } = &mut *shared.borrow_mut();
-            let viewport = initialize_or_update_viewport(viewports, ids, builder, None, None);
+            let viewport = initialize_or_update_viewport(
+                viewports,
+                ids,
+                ViewportClass::Immediate,
+                builder,
+                None,
+                None,
+            );
 
             if viewport.window.is_none() {
                 viewport.init_window(viewport_from_window, painter, event_loop);
@@ -2676,6 +2699,7 @@ mod wgpu_integration {
             viewport_id,
             ViewportOutput {
                 parent,
+                class,
                 builder,
                 viewport_ui_cb,
                 commands,
@@ -2688,6 +2712,7 @@ mod wgpu_integration {
             initialize_or_update_viewport(
                 viewports,
                 ids,
+                class,
                 builder,
                 viewport_ui_cb,
                 focused_viewport,
@@ -2706,6 +2731,7 @@ mod wgpu_integration {
     fn initialize_or_update_viewport(
         viewports: &mut Viewports,
         ids: ViewportIdPair,
+        class: ViewportClass,
         mut builder: ViewportBuilder,
         viewport_ui_cb: Option<Arc<dyn Fn(&egui::Context) + Send + Sync>>,
         focused_viewport: Option<ViewportId>,
@@ -2723,6 +2749,7 @@ mod wgpu_integration {
                 log::debug!("Creating new viewport {:?} ({:?})", ids.this, builder.title);
                 entry.insert(Viewport {
                     ids,
+                    class,
                     builder,
                     viewport_ui_cb,
                     window: None,
@@ -2734,6 +2761,7 @@ mod wgpu_integration {
                 // Patch an existing viewport:
                 let viewport = entry.get_mut();
 
+                viewport.class = class;
                 viewport.ids.parent = ids.parent;
                 viewport.viewport_ui_cb = viewport_ui_cb;
 
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -766,6 +766,7 @@ mod glow_integration {
         }
     }
 
+    #[cfg(feature = "__screenshot")]
     fn save_screeshot_and_exit(
         path: &str,
         painter: &egui_glow::Painter,
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -1169,6 +1169,7 @@ mod glow_integration {
                     builder,
                     viewport_ui_cb,
                     commands,
+                    repaint_delay: _, // ignored - we listend to the repaint callback instead
                 },
             ) in viewport_output
             {
@@ -1386,7 +1387,7 @@ mod glow_integration {
                     .egui_ctx
                     .set_request_repaint_callback(move |info| {
                         log::trace!("request_repaint_callback: {info:?}");
-                        let when = Instant::now() + info.after;
+                        let when = Instant::now() + info.delay;
                         let frame_nr = info.current_frame_nr;
                         event_loop_proxy
                             .lock()
@@ -1989,7 +1990,7 @@ mod wgpu_integration {
                     .egui_ctx
                     .set_request_repaint_callback(move |info| {
                         log::trace!("request_repaint_callback: {info:?}");
-                        let when = Instant::now() + info.after;
+                        let when = Instant::now() + info.delay;
                         let frame_nr = info.current_frame_nr;
 
                         event_loop_proxy
@@ -2676,6 +2677,7 @@ mod wgpu_integration {
                 builder,
                 viewport_ui_cb,
                 commands,
+                repaint_delay: _, // ignored - we listend to the repaint callback instead
             },
         ) in viewport_output
         {
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -1892,6 +1892,8 @@ mod wgpu_integration {
     struct WgpuWinitRunning {
         integration: epi_integration::EpiIntegration,
         app: Box<dyn epi::App>,
+
+        /// Wrapped in an `Rc<RefCell<…>>` so it can be re-entrantly shared via a weak-pointer.
         shared: Rc<RefCell<SharedState>>,
     }
 
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -1872,6 +1872,18 @@ mod wgpu_integration {
     }
 
     impl WgpuWinitRunning {
+        fn set_window(&self, id: ViewportId) -> Result<(), egui_wgpu::WgpuError> {
+            crate::profile_function!();
+            let mut shared = self.shared.borrow_mut();
+            let SharedState {
+                viewports, painter, ..
+            } = &mut *shared;
+            if let Some(Viewport { window, .. }) = viewports.get(&id) {
+                return pollster::block_on(painter.set_window(id, window.as_deref()));
+            }
+            Ok(())
+        }
+
         fn on_window_event(
             &mut self,
             window_id: WindowId,
@@ -2032,20 +2044,6 @@ mod wgpu_integration {
             }
         }
 
-        fn set_window(&mut self, id: ViewportId) -> Result<(), egui_wgpu::WgpuError> {
-            if let Some(running) = &mut self.running {
-                crate::profile_function!();
-                let mut shared = running.shared.borrow_mut();
-                let SharedState {
-                    viewports, painter, ..
-                } = &mut *shared;
-                if let Some(Viewport { window, .. }) = viewports.get(&id) {
-                    return pollster::block_on(painter.set_window(id, window.as_deref()));
-                }
-            }
-            Ok(())
-        }
-
         #[cfg(target_os = "android")]
         fn drop_window(&mut self) -> Result<(), egui_wgpu::WgpuError> {
             if let Some(running) = &mut self.running {
@@ -2062,7 +2060,7 @@ mod wgpu_integration {
             storage: Option<Box<dyn epi::Storage>>,
             window: Window,
             builder: ViewportBuilder,
-        ) -> Result<(), egui_wgpu::WgpuError> {
+        ) -> Result<&mut WgpuWinitRunning, egui_wgpu::WgpuError> {
             crate::profile_function!();
 
             #[allow(unsafe_code, unused_mut, unused_unsafe)]
@@ -2203,13 +2201,11 @@ mod wgpu_integration {
                 );
             }
 
-            self.running = Some(WgpuWinitRunning {
+            Ok(self.running.insert(WgpuWinitRunning {
                 integration,
                 app,
                 shared,
-            });
-
-            Ok(())
+            }))
         }
     }
 
@@ -2703,7 +2699,7 @@ mod wgpu_integration {
 
             Ok(match event {
                 winit::event::Event::Resumed => {
-                    if let Some(running) = &self.running {
+                    let running = if let Some(running) = &self.running {
                         if !running
                             .shared
                             .borrow()
@@ -2716,8 +2712,9 @@ mod wgpu_integration {
                                 &self.app_name,
                                 &mut self.native_options,
                             )?;
-                            self.set_window(ViewportId::ROOT)?;
+                            running.set_window(ViewportId::ROOT)?;
                         }
+                        running
                     } else {
                         let storage = epi_integration::create_storage(
                             self.native_options
@@ -2731,9 +2728,8 @@ mod wgpu_integration {
                             &self.app_name,
                             &mut self.native_options,
                         )?;
-                        self.init_run_state(event_loop, storage, window, builder)?;
-                    }
-                    let running = self.running.as_ref().unwrap(); // Can't fail - we just initialized it
+                        self.init_run_state(event_loop, storage, window, builder)?
+                    };
                     EventResult::RepaintNow(
                         running.shared.borrow().viewports[&ViewportId::ROOT]
                             .window
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -566,6 +566,7 @@ mod glow_integration {
                 current_gl_context,
                 ..
             } = &mut *glutin;
+
             let viewport = viewports.get_mut(&viewport_id).unwrap();
             let window = viewport.window.as_ref().unwrap();
             let gl_surface = viewport.gl_surface.as_ref().unwrap();
@@ -699,6 +700,7 @@ mod glow_integration {
                         })
                         .flatten();
                 }
+
                 winit::event::WindowEvent::Resized(physical_size) => {
                     repaint_asap = true;
 
@@ -712,6 +714,7 @@ mod glow_integration {
                         }
                     }
                 }
+
                 winit::event::WindowEvent::ScaleFactorChanged { new_inner_size, .. } => {
                     let glutin = &mut *self.glutin.borrow_mut();
                     repaint_asap = true;
@@ -719,17 +722,18 @@ mod glow_integration {
                         glutin.resize(*id, **new_inner_size);
                     }
                 }
-                winit::event::WindowEvent::CloseRequested
-                    if self
+
+                winit::event::WindowEvent::CloseRequested => {
+                    let is_root = self
                         .glutin
                         .borrow()
                         .viewport_from_window
                         .get(&window_id)
-                        .map_or(false, |id| *id == ViewportId::ROOT)
-                        && self.integration.should_close() =>
-                {
-                    log::debug!("Received WindowEvent::CloseRequested");
-                    return EventResult::Exit;
+                        .map_or(false, |id| *id == ViewportId::ROOT);
+                    if is_root && self.integration.should_close() {
+                        log::debug!("Received WindowEvent::CloseRequested");
+                        return EventResult::Exit;
+                    }
                 }
                 _ => {}
             }
@@ -830,7 +834,7 @@ mod glow_integration {
 
         viewports: ViewportIdMap<Viewport>,
         viewport_from_window: HashMap<WindowId, ViewportId>,
-        window_maps: ViewportIdMap<WindowId>,
+        window_from_viewport: ViewportIdMap<WindowId>,
     }
 
     impl GlutinWindowContext {
@@ -937,8 +941,8 @@ mod glow_integration {
             let gl_context = match gl_context_result {
                 Ok(it) => it,
                 Err(err) => {
-                    log::warn!("failed to create context using default context attributes {context_attributes:?} due to error: {err}");
-                    log::debug!("retrying with fallback context attributes: {fallback_context_attributes:?}");
+                    log::warn!("Failed to create context using default context attributes {context_attributes:?} due to error: {err}");
+                    log::debug!("Retrying with fallback context attributes: {fallback_context_attributes:?}");
                     gl_config
                         .display()
                         .create_context(&gl_config, &fallback_context_attributes)?
@@ -947,10 +951,10 @@ mod glow_integration {
             let not_current_gl_context = Some(gl_context);
 
             let mut viewport_from_window = HashMap::default();
-            let mut window_maps = ViewportIdMap::default();
+            let mut window_from_viewport = ViewportIdMap::default();
             if let Some(window) = &window {
                 viewport_from_window.insert(window.id(), ViewportId::ROOT);
-                window_maps.insert(ViewportId::ROOT, window.id());
+                window_from_viewport.insert(ViewportId::ROOT, window.id());
             }
 
             let mut viewports = ViewportIdMap::default();
@@ -979,7 +983,7 @@ mod glow_integration {
                 viewports,
                 viewport_from_window,
                 max_texture_side: None,
-                window_maps,
+                window_from_viewport,
             };
 
             slf.on_resume(event_loop)?;
@@ -1024,9 +1028,8 @@ mod glow_integration {
                 .get_mut(&viewport_id)
                 .expect("viewport doesn't exist");
 
-            // make sure we have a window or create one.
-            let window = viewport.window.take().unwrap_or_else(|| {
-                log::debug!("window doesn't exist yet. creating one now with finalize_window");
+            let window = viewport.window.get_or_insert_with(|| {
+                log::trace!("Window doesn't exist yet. Creating one now with finalize_window");
                 Rc::new(
                     glutin_winit::finalize_window(
                         event_loop,
@@ -1036,6 +1039,7 @@ mod glow_integration {
                     .expect("failed to finalize glutin window"),
                 )
             });
+
             {
                 // surface attributes
                 let (width_px, height_px): (u32, u32) = window.inner_size().into();
@@ -1045,15 +1049,16 @@ mod glow_integration {
                     glutin::surface::WindowSurface,
                 >::new()
                 .build(window.raw_window_handle(), width_px, height_px);
-                log::debug!("creating surface with attributes: {surface_attributes:?}");
-                // create surface
+
+                log::trace!("creating surface with attributes: {surface_attributes:?}");
                 let gl_surface = unsafe {
                     self.gl_config
                         .display()
                         .create_window_surface(&self.gl_config, &surface_attributes)?
                 };
-                log::debug!("surface created successfully: {gl_surface:?}. making context current");
-                // make surface and context current.
+
+                log::trace!("surface created successfully: {gl_surface:?}. making context current");
+
                 let not_current_gl_context =
                     if let Some(not_current_context) = self.not_current_gl_context.take() {
                         not_current_context
@@ -1065,13 +1070,15 @@ mod glow_integration {
                             .unwrap()
                     };
                 let current_gl_context = not_current_gl_context.make_current(&gl_surface)?;
+
                 // try setting swap interval. but its not absolutely necessary, so don't panic on failure.
-                log::debug!("made context current. setting swap interval for surface");
+                log::trace!("made context current. setting swap interval for surface");
                 if let Err(err) =
                     gl_surface.set_swap_interval(&current_gl_context, self.swap_interval)
                 {
-                    log::error!("failed to set swap interval due to error: {err}");
+                    log::warn!("Failed to set swap interval due to error: {err}");
                 }
+
                 // we will reach this point only once in most platforms except android.
                 // create window/surface/make context current once and just use them forever.
 
@@ -1087,9 +1094,10 @@ mod glow_integration {
                 self.current_gl_context = Some(current_gl_context);
                 self.viewport_from_window
                     .insert(window.id(), viewport.ids.this);
-                self.window_maps.insert(viewport.ids.this, window.id());
+                self.window_from_viewport
+                    .insert(viewport.ids.this, window.id());
             }
-            viewport.window = Some(window);
+
             Ok(())
         }
 
@@ -1167,6 +1175,8 @@ mod glow_integration {
             viewport_output: Vec<ViewportOutput>,
             focused_viewport: Option<ViewportId>,
         ) {
+            crate::profile_function!();
+
             let mut active_viewports_ids = ViewportIdSet::default();
             active_viewports_ids.insert(ViewportId::ROOT);
 
@@ -1192,7 +1202,7 @@ mod glow_integration {
                 .retain(|id, _| active_viewports_ids.contains(id));
             self.viewport_from_window
                 .retain(|_, id| active_viewports_ids.contains(id));
-            self.window_maps
+            self.window_from_viewport
                 .retain(|id, _| active_viewports_ids.contains(id));
         }
     }
@@ -1204,6 +1214,8 @@ mod glow_integration {
         viewport_ui_cb: Option<Arc<dyn Fn(&egui::Context) + Send + Sync>>,
         focused_viewport: Option<ViewportId>,
     ) -> &mut Viewport {
+        crate::profile_function!();
+
         if builder.icon.is_none() {
             // Inherit icon from parent
             builder.icon = viewports
@@ -1232,7 +1244,7 @@ mod glow_integration {
                 viewport.ids.parent = ids.parent;
                 viewport.viewport_ui_cb = viewport_ui_cb;
 
-                let (commands, recreate) = viewport.builder.patch(&builder);
+                let (delta_commands, recreate) = viewport.builder.patch(&builder);
 
                 if recreate {
                     log::debug!(
@@ -1244,7 +1256,7 @@ mod glow_integration {
                     viewport.egui_winit = None;
                 } else if let Some(window) = &viewport.window {
                     let is_viewport_focused = focused_viewport == Some(ids.this);
-                    process_viewport_commands(commands, window, is_viewport_focused);
+                    process_viewport_commands(delta_commands, window, is_viewport_focused);
                 }
 
                 entry.into_mut()
@@ -1331,7 +1343,10 @@ mod glow_integration {
             Ok((glutin_window_context, painter))
         }
 
-        fn init_run_state(&mut self, event_loop: &EventLoopWindowTarget<UserEvent>) -> Result<()> {
+        fn init_run_state(
+            &mut self,
+            event_loop: &EventLoopWindowTarget<UserEvent>,
+        ) -> Result<&mut GlowWinitRunning> {
             crate::profile_function!();
 
             let storage = epi_integration::create_storage(
@@ -1358,6 +1373,7 @@ mod glow_integration {
             }
 
             let system_theme = system_theme(&glutin.window(ViewportId::ROOT), &self.native_options);
+
             let mut integration = epi_integration::EpiIntegration::new(
                 &glutin.window(ViewportId::ROOT),
                 system_theme,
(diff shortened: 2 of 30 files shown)

### CODE REVIEW

**Comment by emilk** [comment]
Whoa, this looks very promising! I tried running `cargo run -p viewports` on my Mac and after resizing the window and clicking "Show Sync Viewport" I do indeed get two windows 🥳 

My biggest feedback so far is to add docstrings to the code. I'm trying to follow along in `examples/viewports/src/main.rs` but it is quite opaque what `default_embedded`, `show_async`, `get_viewport_id`, `set_render_sync_callback` etc does.

Second piece of feedback: wrap the naked `u64` viewport id in a newtype, e.g. `struct ViewportId(u64)`. That way we get type safety and better readability of the code (I can look at the docstring of `ViewportId` to get an explanation of what a viewport id _is_).

Lastly: run [`cargo cranky`](https://github.com/ericseppanen/cargo-cranky) on the code ;)

**Comment by konkitoman** [comment]
Thank you for more motivation and feedback! :heart:

`examples/viewports/src/main.rs` is currently my playground, and is really messy!
All current errors are in `crates/glow` because is a backend and currently i am only working for eframe glow and winit!
`crates/glow` should have support for multiples windows?
 
Today i will add `ViewportId` and some documentation!

**Comment by konkitoman** [comment]
This is the end of the day for me!
Now every where viewport id is needed, now is used `ViewportId`
`ViewportId` cannot be created, only we can get the main viewport id with `ViewportId::MAIN`
In the feature will be able to get the `ViewportId` of a viewport if we know his name!

I updated the viewports example and i added some documentation!

Tomorrow i will make every Popup to use a `Context::create_viewport_sync` to be rendered!

**Comment by konkitoman** [comment]
This is the end of the day for me!

I made `crates/egui/src/containers/popup.rs` `show_tooltip_area_dyn` to create a sync viewport!
Now every tool-tip will open a native window if he can, but is really broken because the window has a fixed size!

I changed ViewportBuilder, everything in it to be optional. this will allow to create a viewport and run stuff in it without changing the window attributes, this will be used for `was_tooltip_open_last_frame` or to change a window attributes with commands manually.
I implemented all winit window attributes for `ViewportCommand`

And move code from eframe to egui-winit to be more generic.

**Comment by konkitoman** [comment]
@emilk
The last commit [4bbdab1](https://github.com/emilk/egui/pull/3172/commits/4bbdab178826666d7810855748a66d4fe89d4c78)
The `screen_rect` should not work like that?

Is was not storing the window position!
I was always `Pos2::ZERO` as the position.

If did i understated `screen_rect` purpose wrong, how the window position should be stored?

**Comment by emilk** [comment]
All coordinates in egui, including `ctx.screen_rect`, is relative to the native window containing egui. `egui` has no notion of anything outside the main window (at least until this PR!).

How do you plan on handling the coordinate system of multiple windows?

**Comment by konkitoman** [comment]
Currently in `ctx.screen_rect.min` is the outside position of the native window and when we cannot get the outside position will be `Pos2::ZERO`
Currently in `ctx.screen_rect.max` is the exact size of the window size!

I can make it relative so the `ctx.screen_rect.size()` to work again!
Should i do that?

You should run viewports example to see how is currently working!
Currently the window is not redrawn if is moved!

File: crates/eframe/src/native/epi_integration.rs [hunk 1]
```diff
@@ -339,10 +313,9 @@ pub fn create_storage(_app_name: &str) -> Option<Box<dyn epi::Storage>> {
 pub struct EpiIntegration {
     pub frame: epi::Frame,
     last_auto_save: std::time::Instant,
+    pub beagining: Instant,
```
stefnotch, line 316: Typo, should say "beginning"

stefnotch:
> If you see this
>
>    Please give feedback!

If I want to try this out, is there any sort of feedback you'd be looking for in particular? Or should I simply tell you if I notice anything odd.

konkitoman:
@stefnotch
I am looking for feedback to find out what is needed to build a docking system!

stefnotch:
Is it intentional that the windows apparently don't update unless they're being resized? I tried out the viewport demo, and this is what I got.

[Code_2023-08-23-0321.webm](https://github.com/emilk/egui/assets/10220080/4d5b702b-5c89-4dc3-b595-ce10d5942dd6)

konkitoman:
> Is it intentional that the windows apparently don't update unless they're being resized? I tried out the viewport demo, and this is what I got. I cannot press any of the buttons.
> 
>  Code_2023-08-23-0321.webm 
> Operating system: Windows 10

That is not normal!
This should fix the issue [3b41253](https://github.com/emilk/egui/pull/3172/commits/3b412534420569b672b6594a995baee0d3e54952)
Please tell me if that fixed the issue!

stefnotch:
@konkitoman Lovely, that does fix the issue.
I did find another odd bug, where the application will hang while the console spams
```
Set size!
Set size!
Set size!
Set size!
Set size!
```

Operating system: Windows 10

Steps to reproduce:
Start the viewport demo. Press "Show Async Window". (Or press all of them.)
Then quickly toggle the "Force embedding". After a few tries, the application lands in that bugged state.

[firefox_2023-08-23-0324.webm](https://github.com/emilk/egui/assets/10220080/ed5d6a0e-87b7-4333-bc12-82411022ae56)

konkitoman:
> @konkitoman Lovely, that does fix the issue. I did find another odd bug, where the application will hang while the console spams
> 
But you can resize the `egui::Window`?
I'am referring to any thing that has "Window" as a name in viewports example.

The spamming of `Set Size!` is for testing, but should not hang the application!
> ```
> Set size!
> Set size!
> Set size!
> Set size!
> Set size!
> ```
> 
> Operating system: Windows 10
> 
> Steps to reproduce: Start the viewport demo. Press "Show Async Window". (Or press all of them.) Then quickly toggle the "Force embedding". After a few tries, the application lands in that bugged state.
> 
>  firefox_2023-08-23-0324.webm

You are sure that is a bugged state or is crashing, look in the console, but that thing is not happening to me!

konkitoman:
This is how is working for me!
And how should work!

OS: Arch Linux
Desktop Env: KDE Plasma X11

https://github.com/emilk/egui/assets/51934166/0935e031-7293-43fb-ad4f-2eecc027692d

The video is on speed up some times to 130% or 200%, because i was needed to be less than 10MB

The starching is because of X11!

When appear two viewports from nowhere, is because "Sync Viewport" and "Async Viewport" has the same state for showing the two other viewports! 

I unchecked every window/viewport from the main viewport at the end to show how the child viewport of the child viewport is closing!

stefnotch:
It might be a Windows specific bug. It only seems to happen if I repeatedly, and quickly, toggle "Force embedding".

stefnotch:
@konkitoman I tested it on Windows 10.
"Show Async Window" and "Show Sync Window" leads to windows that
- can be moved
- cannot be resized
- will glitch out when pressing the triangle button twice (close and reopen)

konkitoman:
Thanks @stefnotch

> - cannot be resized

The problem with resizing is fixed in [winit 0.29.1-beta](https://github.com/rust-windowing/winit/releases/tag/v0.29.1-beta)
I'am waiting for winit to get out of beta before updating!

> - will glitch out when pressing the triangle button twice (close and reopen)

If you are referring to this image is not really a bug is implemented badly, because if the window can be resize is only annoying!

Now i see that when you said:
> Then quickly toggle the "Force embedding". After a few tries, the application lands in that bugged state.

The window is getting smaller and smaller, and the bugged state is happening when the window has width=1, at the beginning the blue bar i was thinking that was a encoder artifact!
Then the:
> - will glitch out when pressing the triangle button twice (close and reopen)

You are not referring to this, as glitch out?
image

konkitoman:
I wanted to see the auto merge conflict and i pressed "Ready for review" by mistake!

konkitoman:
I see this as finished!

The only problem that i see that, for every sync viewport the fps is cut in half, because if a sync viewport needs to be redraw his parent needs to, the only fixes that i see is to set vsync to off or shadow update the parent!

I should marked this as "Ready for review"?

emilk:
It sounds like "async" viewports are always better.

> A sync Viewport should only be used if we don't or can use a Async Viewport

when can't we create an async viewport?

crates/egui/src/input_state.rs
```diff
@@ -54,6 +54,22 @@ pub struct InputState {
...
+    /// If is not working as expected is a winit bug!
+    pub inner_pos: Pos2,
+
+    /// Viewport outer position, drowable area + decorations
+    /// If is not working as expected is a winit bug!
+    pub outer_pos: Pos2,
+
+    /// Viewport inner size, only drowable area
+    /// If is not working as expected is a winit bug!
+    pub inner_size: Pos2,
```
emilk, line 67: Instead of `inner_pos` and `inner_size`, use a `inner_rect: Rect` and `outer_rect: Rect`.

emilk (commented):
This is amazing, but a lot to review - I only had a few minutes today. I'll try to find more time soon!

crates/egui/src/input_state.rs
konkitoman: The `Rect` i should create like:
```rust
Rect::from_min_size(viewport_pos, viewport_size);
```
Or
```rust
Rect::from_min_max(viewport_pos, viewport_size);
```

From what i see in the code base, i should use the first variant!

crates/egui/src/input_state.rs
```diff
@@ -54,6 +54,22 @@ pub struct InputState {
...
+
+    /// Viewport outer size, drowable area + decorations
+    /// If is not working as expected is a winit bug!
+    pub outer_size: Pos2,
```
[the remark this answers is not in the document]
konkitoman: The unit is physical pixels directly from `winit::window::Window::outer_size` but converted to f32!

konkitoman:
> It sounds like "async" viewports are always better.
> 
> > A sync Viewport should only be used if we don't or can use a Async Viewport
> 
> when can't we create an async viewport?

You always can use a async viewport if you don't want the result!
But for example if you have a popup is easier to use a sync viewport, because returns the result like `egui::Window`

For a async viewport you will always need to have the state in a `Arc<Mutex<T>>`, and cannot access current scope!

crates/egui/src/context.rs
```diff
@@ -171,33 +232,81 @@ struct ContextImpl {
...
-            let ratio = self.input.pixels_per_point / new_pixels_per_point;
-            let mut rect = self.input.screen_rect;
-            rect.min = (ratio * rect.min.to_vec2()).to_pos2();
-            rect.max = (ratio * rect.max.to_vec2()).to_pos2();
-            new_raw_input.screen_rect = Some(rect);
+    fn begin_frame_mut(
+        &mut self,
+        mut new_raw_input: RawInput,
+        viewport_id: ViewportId,
+        parent_viewport_id: ViewportId,
```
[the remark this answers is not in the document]
konkitoman: The idea is that `viewport_id` is the current viewport and `parent_viewport_id` is his parent

crates/eframe/src/native/run.rs
```diff
@@ -414,21 +525,28 @@ mod glow_integration {
...
-        window: Option<winit::window::Window>,
+
+        viewports: HashMap<ViewportId, Arc<RwLock<Window>>>,
+        viewports_maps: HashMap<winit::window::WindowId, ViewportId>,
+        builders: HashMap<ViewportId, ViewportBuilder>,
     }
 
+    #[allow(unsafe_code)]
+    unsafe impl Sync for GlutinWindowContext {}
+    #[allow(unsafe_code)]
```
emilk, line 541: Please explain in a comment why this is safe!

crates/eframe/src/native/run.rs
```diff
@@ -565,63 +703,111 @@ mod glow_integration {
...
+            let values = self
+                .viewports
+                .values()
+                .cloned()
+                .collect::<Vec<Arc<RwLock<Window>>>>();
+            for win in values {
+                if win.read().gl_surface.is_some() {
+                    continue;
+                }
+                self.init_window(&win, event_loop)?;
```
emilk, line 718:
```suggestion
            let windows = self
                .viewports
                .values()
                .cloned()
                .collect::<Vec<Arc<RwLock<Window>>>>();
            for window in windows {
                if window.read().gl_surface.is_some() {
                    continue;
                }
                self.init_window(&window, event_loop)?;
```

docs/egui_demo_app.js
```diff
@@ -7,13 +7,40 @@ let wasm_bindgen;
     }
```
emilk, line 7: please revert `.js` and `.wasm` - see https://github.com/emilk/egui/blob/master/CONTRIBUTING.md#making-a-pr

crates/egui_glow/src/winit.rs
```diff
@@ -53,14 +54,15 @@ impl EguiGlow {
...
+            ..
+        } = self.egui_ctx.run(raw_input, ViewportIdPair::MAIN, run_ui);
 
         self.egui_winit
             .handle_platform_output(window, &self.egui_ctx, platform_output);
 
         self.shapes = shapes;
         self.textures_delta.append(textures_delta);
-        repaint_after
+        repaint_after.last().map(|(_, t)| *t).unwrap_or_default()
```
emilk, line 65: why only look at the last here?

crates/egui/src/viewport.rs
```diff
@@ -0,0 +1,348 @@
...
+        self
+    }
+
+    pub fn with_transparent(mut self, transparent: bool) -> Self {
+        self.transparent = Some(transparent);
+        self
+    }
+
+    /// The icon needs to be wrapped in Arc because will be copied every frame
+    pub fn with_window_icon(mut self, icon: Option<Arc<(u32, u32, Vec<u8>)>>) -> Self {
```
emilk, line 188:
```suggestion
    pub fn with_window_icon(mut self, icon: Option<Arc<ColorImage>>) -> Self {
```

crates/egui/src/viewport.rs
```diff
@@ -0,0 +1,348 @@
...
+/// The filds in this struct should not be change directly, but is not problem tho!
+/// Every thing is wrapped in Option<> indicates that thing should not be changed!
+#[derive(PartialEq, Eq, Clone)]
+#[allow(clippy::option_option)]
+pub struct ViewportBuilder {
+    pub id: Id,
+    pub title: String,
+    pub name: Option<(String, String)>,
+    pub position: Option<Option<Pos2>>,
+    pub inner_size: Option<Option<Pos2>>,
```
emilk, line 67: What is the difference between `None` and `Some(None)` ? Can we avoid using an Option of an Option?

crates/egui/src/viewport.rs
```diff
@@ -0,0 +1,348 @@
...
+        self.drag_and_drop = Some(value);
+        self
+    }
+
+    pub fn with_position(mut self, value: Option<Pos2>) -> Self {
+        self.position = Some(value);
+        self
+    }
+
+    pub fn with_name(mut self, id: impl Into<String>, instance: impl Into<String>) -> Self {
```
emilk, line 262: All these functions should have docstrings - can we copy the ones from winit maybe?

crates/egui/src/input_state.rs
```diff
@@ -54,6 +54,14 @@ pub struct InputState {
     /// Position and size of the egui area.
     pub screen_rect: Rect,
 
+    /// Viewport inner position and size, only the drowable area
+    /// unit = physical pixels
+    pub inner_rect: Rect,
+
+    /// Viewport outer position and size, drowable area + decorations
+    /// unit = physical pixels
```
emilk, line 62: Why use physical pixels for these? We use logical points for everything else in egui!

Have you tested this PR with `pixels_per_point != 1`? You can use `egui::gui_zoom::zoom_with_keyboard_shortcuts` in an application to zoom in/out with cmd +/-

crates/egui/src/data/output.rs
```diff
@@ -18,7 +19,7 @@ pub struct FullOutput {
     /// duration elapses. when in reactive mode, egui spends forever waiting for input and only then,
     /// will it repaint itself. this can be used to make sure that backend will only wait for a
     /// specified amount of time, and repaint egui without any new input.
-    pub repaint_after: std::time::Duration,
+    pub repaint_after: Vec<(ViewportId, std::time::Duration)>,
```
emilk, line 22: I think it makes sense to have at most one duration per viewport

```suggestion
    pub repaint_after: BTreeMap<ViewportId, std::time::Duration>,
```

crates/eframe/src/web/app_runner.rs
```diff
@@ -199,6 +203,7 @@ impl AppRunner {
 
         self.frame.info.cpu_usage = Some((now_sec() - frame_start) as f32);
 
+        let (_, repaint_after) = repaint_after[0];
```
emilk, line 206: why only look at one element?

crates/eframe/src/native/run.rs
```diff
@@ -791,190 +1016,569 @@ mod glow_integration {
...
+                                                    "failed to get current context to swap buffers",
+                                                ),
+                                            );
+                                        winit_state.handle_platform_output(
+                                            &win,
+                                            egui_ctx,
+                                            output.platform_output,
+                                        );
+                                    } else {
+                                        break 'try_render;
```
emilk, line 1163: isn't this just a `return`?

crates/eframe/src/native/run.rs
```diff
@@ -791,190 +1016,569 @@ mod glow_integration {
...
+                                    }
+                                } else {
+                                    break 'try_render;
+                                }
+                            }
+                        }
+                    }
+                },
```
emilk, line 1171: This is a lot of complex code - can you please move it out into own functions

crates/eframe/src/native/run.rs
```diff
@@ -160,94 +190,122 @@ fn run_and_return(
...
-            winit::event::Event::RedrawRequested(_) => {
-                next_repaint_time = extremely_far_future();
-                winit_app.run_ui_and_paint()
+            // #[cfg(target_os = "windows")]
+            // winit::event::Event::RedrawEventsCleared => {
+            // windows_next_repaint_times.clear();
+            // winit_app.run_ui_and_paint(None)
+            // vec![EventResult::Wait]
+            // }
+            // #[cfg(not(target_os = "windows"))]
```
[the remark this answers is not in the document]
konkitoman: The problem is that for the windows workaround to work on RedrawEventsCleared i need to render all the viewports and in here i don't know all the window_ids of all the viewports.
I can do that if you want!

But other problem will be the performance, because if we interact with any viewport all viewports will be rendered!

On wine works!
But only if the winit control flow is set to poll when a redraw is requested, but when a redraw is requested from egui will always be set to pull at line 292

From what i seen from @stefnotch is currently working but i will test and on my friend computer!
I will add a new comment when i test this on Windows 10

crates/eframe/src/native/run.rs
```diff
@@ -565,63 +703,111 @@ mod glow_integration {
...
+                    log::error!("failed to set swap interval due to error: {e:?}");
+                }
+                // we will reach this point only once in most platforms except android.
+                // create window/surface/make context current once and just use them forever.
+
+                let native_pixels_per_point = window.scale_factor() as f32;
+
+                if win.egui_winit.is_none() {
+                    let mut egui_winit = egui_winit::State::new(event_loop);
+                    // egui_winit.set_max_texture_side(max_texture_side);
```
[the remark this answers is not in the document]
konkitoman: But why is called side and not size is not `GL_MAX_TEXTURE_SIZE`?

crates/egui/src/input_state.rs
konkitoman: Because this is for the native window!
Is not breaking any thing when you zoom, because is not used anywhere!

But i can make it to be logical pixels, but only for `InputState`!
In `RawInput` will be physical always!

crates/egui/src/data/output.rs
konkitoman: I used a `HashMap` because i'am not interested in the order and because in `Repaint` was stored as a `HashMap` now i only call clone!

crates/eframe/src/web/app_runner.rs
konkitoman: This is because in a browser we cannot create multiples viewports!
So we only interested in the main viewport!

Is a way to create a new browser windows with shared state?

crates/eframe/src/native/run.rs
konkitoman: I moved in `render_sync_viewport`

crates/eframe/src/native/run.rs
konkitoman: That was part of the past implementations!
I got under the radar for me, thx for pointing that out!

crates/eframe/src/native/run.rs
```diff
@@ -160,94 +190,122 @@ fn run_and_return(
...
-            EventResult::RepaintNow => {
-                log::trace!("Repaint caused by winit::Event: {:?}", event);
-                if cfg!(target_os = "windows") {
-                    // Fix flickering on Windows, see https://github.com/emilk/egui/pull/2280
-                    next_repaint_time = extremely_far_future();
-                    winit_app.run_ui_and_paint();
-                } else {
-                    // Fix for https://github.com/emilk/egui/issues/2425
-                    next_repaint_time = Instant::now();
+        for event in events {
```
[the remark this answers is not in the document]
konkitoman: If we have multiples viewports like 3 async viewports, we can receive 3 repaint requests at the same time, this is why that is needed!
And for example to make the old window workaround to work needs to render all the viewports and that will return a lot of events!

crates/egui/src/viewport.rs
konkitoman: First when you call `Context::create_viewport` the builder will patch the existing builder of that viewport if exists, and will make `ViewportCommand` for the changes to be applied if a native window already exists!

For example like in android the app suspends and resumes or the compositor crashes on linux
on suspend the windows is destroyed or is already destroyed
on resume will build a new window and the window will be build with the ViewportBuilder that was patched
this make possible for a window to be restored and the and the ui is not be affected!

The difference between None and Some(None)
- None indicate that not change is needed
- Some(None) on inner_size will set the default native window size to be chosen by the OS when is rebuild!

This system can be removed if you want!

crates/egui/src/viewport.rs
konkitoman: I will copy and adapt winit documentations for every function in ViewportBuilder!

crates/egui_glow/src/winit.rs
konkitoman: This is because glow implementation do not have support for multiples viewports!

Glow implementation should have support for multiples windows?

docs/egui_demo_app.js
konkitoman: I tried to do but some how i commited with that!
I don't know the commit i did that!

I run `git restore docs --source upstream/master` now is ok!

crates/eframe/src/native/run.rs
konkitoman: That is not safe because the gl context is not thread safe and neither clipboard!
That was a workaround to work on sync viewports when i didn't think sync viewport was possible! 

I did this commit to resolve that [449f38a](https://github.com/emilk/egui/pull/3172/commits/449f38adf2837c0e3663914f65aeaf595faa3139)

But this makes it impossible to have two different `egui::Context` on the same thread!
There is a way to store `std::thread::LocalKey` on a struct?

crates/egui/src/input_state.rs
konkitoman: But when you set the size or position of the window with a `ViewportCommand`
Should not be physical pixels?

ajfrantz:
I've spent an hour or two playing with this changeset with an internal tool and so far it's a considerable improvement on some of the hacks we used previously. As a bonus, it allows us to revert back to using `eframe`, which should cut down on our poor half-baked reimplementations of some features. Thanks for all the work!

> > It sounds like "async" viewports are always better.
> > > A sync Viewport should only be used if we don't or can use a Async Viewport
> > 
> > 
> > when can't we create an async viewport?
> 
> You always can use a async viewport if you don't want the result! But for example if you have a popup is easier to use a sync viewport, because returns the result like `egui::Window`
> 
> For a async viewport you will always need to have the state in a `Arc<Mutex<T>>`, and cannot access current scope!

I'm in full support of having a sync viewport option; we have a few places where we want to drive a 'child' window from a 'parent' window, but the 'child' window isn't really autonomous. Using a sync window prevents us from needing to wrap a bunch of internal state in `Arc<Mutex<>>` when we'll never really be able to do anything sensibly 'in parallel' anyway.

workingj:
hi there, thanks for the great work on egui, looking forward to use it in an project.
any progress here in sight?

konkitoman:
@workingj

This PR from my point of view is ready!

But is needed to be tested on Windows!

workingj:
> @workingj
> 
> This PR from my point of view is ready!
> 
> But is needed to be tested on Windows!

ok thanks, maybe i can look in to it if it does not take too much time i think i can trie it.
i am still learning may be you @konkitoman can give me an advide how to start testing your code on windows.

konkitoman:
## Today i tested on:

- Arch Linux, Plasma 5.27.8, X11, Nvidia 535.104.05, Glow and WGPU
- Windows 10 21H2 in QEMU/KVM with mesa-dist-win for OpenGL, Glow
- NixOS, Plasma 5.27.7, X11 and Wayland, Intel UHD Graphics G4, Glow and WGPU

The fixed problems are excluded!

## Problem 1:

On the Windows 10 VM, when a viewport was resized the content was flickering,
but is the same on the master so is probably because of the VM!

## Problem 2:

In `egui_demo_app` when scale/pixels per point are changed is taking a long time to refresh
I didn't look to find the cause!
But is probably because now any repaint is rendered only once!

## What you can do @workingj

You can chose both options!

### Option 1

You can run `cargo run -p viewports`, that is the current test suit and example!
The example should work as in [this post](https://github.com/emilk/egui/pull/3172#issuecomment-1690131096)
The example was a little improved and the `egui::Window` was removed from it!
You can send a video in that you check all the `Show * Viewport` check boxes and interact with everyone, resize them and
drag and drop, drag the "Count" or press "Add"!
And check and unchecked the "Force Embedding"
Try to spam any checkbox!
And if you want you can run `cargo run -p viewports --features wgpu` and do the same!

### Option 2

You can run `cargo run -p egui_demo_app` on `konkitoman/multiples_viewports` and `emilk/master`

They resize the same on Windows?

The `konkitoman/multiples_viewports` is more or the same as responsive as the `emilk/master`?

workingj:
Hi, 
thanks for the instruction!
tested it on: Win 10 Home 22H2

- `cargo run -p viewports`
every thing worked well and performant,
spaming triggers worked fine,
resizing worked well too, no stuttering or glitching
actually i am impressed by the performance and what i have seen so far in comparison what i tried till now... great work!

- `cargo run -p viewports --features wgpu`
resizing worked but there was some stuttering or glitching of the fonts and somtimes black or white areas that appeard while fastly resizing the windows
spaming triggers worked fine,
but suddenly there occured a panic which i could not reproduce while wildy clicking an draging the windows:
```shell
thread 'main' panicked at 'Error in Surface::configure: Validation Error

Caused by:
    Both `Surface` width and height must be non-zero. Wait to recreate the `Surface` until the window has non-zero area.
', C:\Users\worki\.cargo\registry\src\index.crates.io-6f17d22bba15001f\wgpu-0.17.0\src\backend\direct.rs:771:18
error: process didn't exit successfully: `C:\Users\...
```

on issue i found in both versions:
setting the parent position to 0:0 did not work correctly in that it left a gap of 7px left edge of the screen.
after moving it to the edge it showed Outer Rect Pos: `-7.0 0.0`

having no screen recoder on windows at hand and not the time messing with stuff from the internet i tried one recoder and it did not work, so i cant provide a video as of now, many some one can hint a simple os solution.

hope this helps option 2 i try tomorrow...

konkitoman:
# Update

The initial implementation is completed!

One minor bug remains:
If we have a sync viewport in a sync viewport in a async viewport when changing the pixels_per_point depending from where is changed the fonts will look weird for less then 1ms from my testing, but doesn't happen every time!

Other problems are that `InputState::outer_rect`/`ViewportCommand::SetPosition`/`ViewportBuilder::with_position` will not work as expected because wayland do not supports to get the window position and other problems with winit

Drag and Drop between native windows is not implemented, and i don't know how!
But for example in viewports example if you drag some thing and drop it in a invalid space then go to other viewport and click on drop_target will be there, this happens because of a bug in example!

crates/egui/src/input_state.rs
konkitoman: `inner_rect` and `outer_rect` was made to position a new viewport, but winit doesn't work well with that!
So can be removed if you want!

emilk:
I'm finally back for more reviews, and this time I'm determined to get this merged, hopefully some time next week!

There is a big problem though - I'm trying this on my Mac and running _any_ eframe application results in a panic on shutdown (e..g when closing the native window, or hitting Cmd+Q):

```
> RUST_BACKTRACE=1 RUST_LOG=trace cargo r -p hello_world
…

thread '<unnamed>' panicked at 'cannot access a Thread Local Storage value during or after destruction: AccessError', /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/thread/local.rs:228:26
stack backtrace:
   0: rust_begin_unwind
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:578:5
   1: core::panicking::panic_fmt
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/panicking.rs:67:14
   2: core::result::unwrap_failed
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/result.rs:1687:5
   3: core::result::Result<T,E>::expect
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/result.rs:1046:23
   4: std::thread::local::LocalKey<T>::with
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/thread/local.rs:228:9
   5: objc2::rc::autorelease::AutoreleasePool::new
             at /Users/emilk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/objc2-0.3.0-beta.3.patch-leaks.3/src/rc/autorelease.rs:53:9
   6: objc2::rc::autorelease::autoreleasepool
             at /Users/emilk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/objc2-0.3.0-beta.3.patch-leaks.3/src/rc/autorelease.rs:312:25
   7: winit::platform_impl::platform::util::async::close_sync::{{closure}}
             at /Users/emilk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.28.6/src/platform_impl/macos/util/async.rs:199:9
   8: winit::platform_impl::platform::util::async::run_on_main
             at /Users/emilk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.28.6/src/platform_impl/macos/util/async.rs:33:9
   9: winit::platform_impl::platform::util::async::close_sync
             at /Users/emilk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.28.6/src/platform_impl/macos/util/async.rs:198:5
  10: core::ptr::drop_in_place<winit::platform_impl::platform::Window>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  11: core::ptr::drop_in_place<winit::window::Window>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  12: core::ptr::drop_in_place<core::cell::UnsafeCell<winit::window::Window>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  13: core::ptr::drop_in_place<lock_api::rwlock::RwLock<parking_lot::raw_rwlock::RawRwLock,winit::window::Window>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  14: core::ptr::drop_in_place<epaint::mutex::rw_lock_impl::RwLock<winit::window::Window>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  15: alloc::sync::Arc<T>::drop_slow
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/alloc/src/sync.rs:1259:18
  16: <alloc::sync::Arc<T> as core::ops::drop::Drop>::drop
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/alloc/src/sync.rs:1886:13
  17: core::ptr::drop_in_place<alloc::sync::Arc<epaint::mutex::rw_lock_impl::RwLock<winit::window::Window>>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  18: core::ptr::drop_in_place<core::option::Option<alloc::sync::Arc<epaint::mutex::rw_lock_impl::RwLock<winit::window::Window>>>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  19: core::ptr::drop_in_place<eframe::native::run::glow_integration::Window>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  20: core::ptr::drop_in_place<core::cell::UnsafeCell<eframe::native::run::glow_integration::Window>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  21: core::ptr::drop_in_place<lock_api::rwlock::RwLock<parking_lot::raw_rwlock::RawRwLock,eframe::native::run::glow_integration::Window>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  22: core::ptr::drop_in_place<epaint::mutex::rw_lock_impl::RwLock<eframe::native::run::glow_integration::Window>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  23: alloc::sync::Arc<T>::drop_slow
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/alloc/src/sync.rs:1259:18
  24: <alloc::sync::Arc<T> as core::ops::drop::Drop>::drop
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/alloc/src/sync.rs:1886:13
  25: core::ptr::drop_in_place<alloc::sync::Arc<epaint::mutex::rw_lock_impl::RwLock<eframe::native::run::glow_integration::Window>>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  26: core::ptr::drop_in_place<(egui::viewport::ViewportId,alloc::sync::Arc<epaint::mutex::rw_lock_impl::RwLock<eframe::native::run::glow_integration::Window>>)>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  27: core::ptr::mut_ptr::<impl *mut T>::drop_in_place
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mut_ptr.rs:1431:18
  28: hashbrown::raw::Bucket<T>::drop
             at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hashbrown-0.12.3/src/raw/mod.rs:335:23
  29: hashbrown::raw::RawTable<T,A>::drop_elements
             at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hashbrown-0.12.3/src/raw/mod.rs:598:17
  30: <hashbrown::raw::RawTable<T,A> as core::ops::drop::Drop>::drop
             at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hashbrown-0.12.3/src/raw/mod.rs:1806:17
  31: core::ptr::drop_in_place<hashbrown::raw::RawTable<(egui::viewport::ViewportId,alloc::sync::Arc<epaint::mutex::rw_lock_impl::RwLock<eframe::native::run::glow_integration::Window>>)>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  32: core::ptr::drop_in_place<hashbrown::map::HashMap<egui::viewport::ViewportId,alloc::sync::Arc<epaint::mutex::rw_lock_impl::RwLock<eframe::native::run::glow_integration::Window>>,ahash::random_state::RandomState>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  33: core::ptr::drop_in_place<std::collections::hash::map::HashMap<egui::viewport::ViewportId,alloc::sync::Arc<epaint::mutex::rw_lock_impl::RwLock<eframe::native::run::glow_integration::Window>>,ahash::random_state::RandomState>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  34: core::ptr::drop_in_place<eframe::native::run::glow_integration::GlutinWindowContext>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  35: core::ptr::drop_in_place<core::cell::UnsafeCell<eframe::native::run::glow_integration::GlutinWindowContext>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  36: core::ptr::drop_in_place<lock_api::rwlock::RwLock<parking_lot::raw_rwlock::RawRwLock,eframe::native::run::glow_integration::GlutinWindowContext>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  37: core::ptr::drop_in_place<epaint::mutex::rw_lock_impl::RwLock<eframe::native::run::glow_integration::GlutinWindowContext>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  38: alloc::sync::Arc<T>::drop_slow
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/alloc/src/sync.rs:1259:18
  39: <alloc::sync::Arc<T> as core::ops::drop::Drop>::drop
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/alloc/src/sync.rs:1886:13
  40: core::ptr::drop_in_place<alloc::sync::Arc<epaint::mutex::rw_lock_impl::RwLock<eframe::native::run::glow_integration::GlutinWindowContext>>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  41: core::ptr::drop_in_place<eframe::native::run::glow_integration::GlowWinitApp::init_run_state::{{closure}}>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  42: core::ptr::drop_in_place<alloc::boxed::Box<dyn core::ops::function::Fn<(&egui::context::Context,egui::viewport::ViewportBuilder,egui::viewport::ViewportIdPair,alloc::boxed::Box<dyn core::ops::function::FnOnce<(&egui::context::Context,)>+Output = ()>)>+Output = ()>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  43: core::ptr::drop_in_place<core::option::Option<alloc::boxed::Box<dyn core::ops::function::Fn<(&egui::context::Context,egui::viewport::ViewportBuilder,egui::viewport::ViewportIdPair,alloc::boxed::Box<dyn core::ops::function::FnOnce<(&egui::context::Context,)>+Output = ()>)>+Output = ()>>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  44: core::ptr::drop_in_place<core::cell::UnsafeCell<core::option::Option<alloc::boxed::Box<dyn core::ops::function::Fn<(&egui::context::Context,egui::viewport::ViewportBuilder,egui::viewport::ViewportIdPair,alloc::boxed::Box<dyn core::ops::function::FnOnce<(&egui::context::Context,)>+Output = ()>)>+Output = ()>>>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  45: core::ptr::drop_in_place<lock_api::rwlock::RwLock<parking_lot::raw_rwlock::RawRwLock,core::option::Option<alloc::boxed::Box<dyn core::ops::function::Fn<(&egui::context::Context,egui::viewport::ViewportBuilder,egui::viewport::ViewportIdPair,alloc::boxed::Box<dyn core::ops::function::FnOnce<(&egui::context::Context,)>+Output = ()>)>+Output = ()>>>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  46: core::ptr::drop_in_place<epaint::mutex::rw_lock_impl::RwLock<core::option::Option<alloc::boxed::Box<dyn core::ops::function::Fn<(&egui::context::Context,egui::viewport::ViewportBuilder,egui::viewport::ViewportIdPair,alloc::boxed::Box<dyn core::ops::function::FnOnce<(&egui::context::Context,)>+Output = ()>)>+Output = ()>>>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  47: core::ptr::drop_in_place<core::option::Option<epaint::mutex::rw_lock_impl::RwLock<core::option::Option<alloc::boxed::Box<dyn core::ops::function::Fn<(&egui::context::Context,egui::viewport::ViewportBuilder,egui::viewport::ViewportIdPair,alloc::boxed::Box<dyn core::ops::function::FnOnce<(&egui::context::Context,)>+Output = ()>)>+Output = ()>>>>>
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ptr/mod.rs:490:1
  48: core::mem::drop
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/mem/mod.rs:979:24
  49: std::sys::common::thread_local::fast_local::fast::destroy_value::{{closure}}
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/sys/common/thread_local/fast_local.rs:249:13
  50: core::ops::function::FnOnce::call_once
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ops/function.rs:250:5
  51: <core::panic::unwind_safe::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/panic/unwind_safe.rs:271:9
  52: std::panicking::try::do_call
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:485:40
  53: ___rust_try
  54: std::panicking::try
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:449:19
  55: std::panic::catch_unwind
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panic.rs:140:14
  56: std::sys::common::thread_local::fast_local::fast::destroy_value
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/sys/common/thread_local/fast_local.rs:246:25
  57: std::sys::unix::thread_local_dtor::register_dtor::run_dtors
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/sys/unix/thread_local_dtor.rs:79:17
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
fatal runtime error: thread local panicked on drop
zsh: abort      RUST_BACKTRACE=1 RUST_LOG=trace cargo r -p hello_world
```

I'm investigating…

emilk:
Actually, maybe the solution is to make `EGUI_RENDER_SYNC` and `WINIT_EVENT_LOOP` proper globals, i.e. not `thread_local` ones. I'll try that tomorrow.

emilk:
I fixed the crash! `set_render_sync_callback` was holding onto strong `Rc/Arc` pointers, which meant a lot of stuff was kept alive for too long, like `glow::Context` and `GlutinWindowContext`, and then freed when the main thread died. Switching to weak pointers solved the problem.

crates/egui/src/viewport.rs
```diff
@@ -0,0 +1,493 @@
...
+    pub decorations: Option<bool>,
+    pub icon: Option<Option<Arc<ColorImage>>>,
+    pub active: Option<bool>,
+    pub visible: Option<bool>,
+    pub title_hidden: Option<bool>,
+    pub titlebar_transparent: Option<bool>,
+    pub fullsize_content_view: Option<bool>,
+    pub min_inner_size: Option<Option<Vec2>>,
+    pub max_inner_size: Option<Option<Vec2>>,
+    pub drag_and_drop: Option<bool>,
```
emilk, line 125: Why are some of these options of options?

crates/egui/src/context.rs
```diff
@@ -146,24 +154,35 @@ struct ContextImpl {
...
 
+    viewports: ViewportMap<Viewport>,
+    viewport_commands: Vec<(ViewportId, ViewportCommand)>,
+
+    is_desktop: bool,
+    force_embedding: bool,
+
     /// Written to during the frame.
     layer_rects_this_frame: ahash::HashMap<LayerId, Vec<(Id, Rect)>>,
+    layer_rects_this_viewports: ViewportMap<HashMap<LayerId, Vec<(Id, Rect)>>>,
```
emilk, line 181: Why do we need to duplicate these and others? Isn't it enough to use `layer_rects_this_viewports[current_viewport_id]` everywhere? It would mean less bookkeeping at the start and end of each frame

crates/egui/src/viewport.rs
konkitoman: That is because the `ViewportBuilder` is like a patch that is applied over the existing ViewportBuilder!
The `egui_winit::changes_between_builders` will make ViewportCommands and say if the window needs recreation!

So for `min_inner_size`
If first Option is None that means that will remain the same!
If first Option is Some and then second one is None that means that can have any size!
If first Option is Some and then second one is Some(10) that means that the Viewport will have the min size of 10

Why:
This was implemented when i was experimenting with making any popup a native window, but if forgot for what was used!

But if was not like that, only the first ViewportBuilder was applied and others only if the viewport was destroyed by the os!

Now that i rethink this: i will say that the `egui_winit::changes_between_builders` needs to be moved in egui because do not uses the winit!

crates/egui/src/context.rs
konkitoman: We can!
Was like that from the stage when i was trying to make stuff to work!

I will do that now!

konkitoman:
> I wonder if we can come up with better names for `sync` and `async` viewports to not confuse it with Rust `async` feature.
> 
> Maybe something like `immediate` vs `deferred`?

immediate and deferred are pretty long but in docs are nice!
So i will say that the deferred viewport should be called "viewport" and immediate "sync viewport" or "immediate viewport"
So the creation will be like `create_viewport` for deferred and `create_viewport_immediate` for immediate

This is because the deferred viewport is recommended!

crates/eframe/src/native/run.rs
```diff
@@ -789,190 +1005,666 @@ mod glow_integration {
...
+                    let mut glutin = glutin_ctx.borrow_mut();
+                    let last_builder = glutin.builders.entry(*id).or_insert(builder.clone());
+                    let (commands, recreate) = changes_between_builders(builder, last_builder);
+                    drop(glutin);
+                    if let Some(viewport) = glutin_ctx.borrow().viewports.get(id) {
+                        let mut viewport = viewport.borrow_mut();
+                        if recreate {
+                            viewport.window = None;
+                            viewport.gl_surface = None;
+                            viewport.id_pair.parent = *id;
```
emilk, line 1279: Why is parent changed to point to itself here?

konkitoman: That was a mistake!

konkitoman: Fixed!

crates/egui/src/memory.rs
```diff
@@ -68,23 +71,37 @@ pub struct Memory {
...
+    pub(crate) window_interactions: ViewportIdMap<window::WindowInteraction>,
+
     #[cfg_attr(feature = "persistence", serde(skip))]
     pub(crate) drag_value: crate::widgets::drag_value::MonoState,
 
+    #[cfg_attr(feature = "persistence", serde(skip))]
     pub(crate) areas: Areas,
 
+    #[cfg_attr(feature = "persistence", serde(skip))]
+    pub(crate) viewports_areas: ViewportIdMap<Areas>,
```
emilk, line 103: Same here - I don't think we need to duplicate `Interaction`, `WindowInteraction` and `Areas`. We can just use `interactions.entry(viewport_id)` instead of `interaction`, etc.

konkitoman: I tried to do that!
But requires to many changes!
And will be worst performance because needs a lookup every time!

I am thinking to move all of that to a `ViewportMemory` or something.

What is your opinion?

emilk: I'll fix

crates/egui/src/context.rs
```diff
@@ -50,92 +54,96 @@ impl Default for WrappedTextureManager {
...
     /// The current frame number.
     ///
     /// Incremented at the end of each frame.
-    frame_nr: u64,
-
-    /// The duration backend will poll for new events, before forcing another egui update
-    /// even if there's no new events.
-    ///
-    /// Also used to suppress multiple calls to the repaint callback during the same frame.
-    repaint_after: std::time::Duration,
```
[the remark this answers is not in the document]
konkitoman: Because now only the repaint callback is used!

emilk (commented):
The `wgpu` backend always starts with `pixels_per_point = 1`, ignoring the native gui scaling

**EDIT:** fixed

crates/egui/src/input_state.rs
```diff
@@ -54,6 +54,14 @@ pub struct InputState {
...
+    pub outer_rect: Rect,
```
emilk, line 63: I suggest we combine these two into a `ViewportInfo` or similar, e.g.:

```struct
ViewportInfo {
    pub inner_rect: Option<Rect>,
    pub outer_rect: Option<Rect>,
}
```

they should be `Option`s because we don't always know where the viewport is (e.g. on web).

This would also be the natural place to add more info later, such as wether or not the viewport is maximied, focused, etc. Basically this could replace `eframe::WindowInfo`, though we don't need to do so in this PR.

crates/egui/src/lib.rs
```diff
@@ -125,7 +125,7 @@
 //! loop {
 //!     let raw_input: egui::RawInput = gather_input();
 //!
-//!     let full_output = ctx.run(raw_input, |ctx| {
+//!     let full_output = ctx.run(raw_input, egui::ViewportIdPair::ROOT, |ctx| {
```
emilk, line 128: What do you think about adding the `egui::ViewportIdPair::ROOT` to the `RawInput` instead, perhaps as part of the `struct ViewportInfo` I suggested earlier?

That way a `RawInput` is self-descriptive, saying "This is the input for this viewport"

crates/egui/src/input_state.rs
konkitoman: That will be nice!

So we can move the `WindowEvent::CloseRequested` to `ViewportInfo::close_requested`

crates/egui/src/lib.rs
konkitoman: So do you mean like:
```rust
struct ViewportInfo{
    id_pair: ViewportIdPair,
    inner_rect: Option<Rect>,
    outer_rect: Option<Rect>,
}

struct RawInput{
    viewport: ViewportInfo
    ..
}
```

And removing the `ViewportIdPair` from `Context::{begin_frame,  run}`
That will be nice, and no more changes are needed from the user!
But that means the `egui_winit::State::take_egui_input` needs the `ViewportIdPair`

pablosichert:
Heyo, great to see work on this feature happening - thank you for your efforts!

Just tried this out locally and chiming in with a few comments. I ran the code on macOS 13.0.1.
- When I set `NativeOptions::decorated` to `false` the main window never launches / when calling `.with_decorations(false)` on `ViewportBuilder` the process hangs and the viewport doesn't launch.
- Would it be possible that `Context::show_viewport` accepts a `FnMut` instead of `Fn` - similar to `App::update` taking a `&mut self`? That way it would be possible to mutate local state for the viewport without interior mutability patterns.
- How would the following code look like for viewports, since the `show_viewport` callback does not provide any `eframe::Frame`:

```Rust
if ui.interact(ui.max_rect(), Id::new("window-drag"), Sense::drag()).dragged() {
    frame.drag_window();
}
```

konkitoman:
> Heyo, great to see work on this feature happening - thank you for your efforts!
> 
> Just tried this out locally and chiming in with a few comments. I ran the code on macOS 13.0.1.
> 
> * When I set `NativeOptions::decorated` to `false` the main window never launches / when calling `.with_decorations(false)` on `ViewportBuilder` the process hangs and the viewport doesn't launch.

On Windows and Linux works as expected!
I cannot test on Mac OS
@emilk You can test this?
> * Would it be possible that `Context::show_viewport` accepts a `FnMut` instead of `Fn` - similar to `App::update` taking a `&mut self`? That way it would be possible to mutate local state for the viewport without interior mutability patterns.

Give me an example in what case that will help.
> * How would the following code look like for viewports, since the `show_viewport` callback does not provide any `eframe::Frame`:
> 
> ```rust
> if ui.interact(ui.max_rect(), Id::new("window-drag"), Sense::drag()).dragged() {
>     frame.drag_window();
> }
> ```

That is like this:

```rust
 if ui.interact(ui.max_rect(), Id::new("window-drag"), Sense::drag()).dragged() { 
   ui.ctx().viewport_command(egui::ViewportCommand::Drag);
}
```

pablosichert:
> On Windows and Linux works as expected!

Ok, good to know. I will investigate myself, but in case you have any idea where the issue might be, happy to take your guesses.

> Give me an example in what case that will help.

Sure, a minimal example would be something like this:

```rust
...
struct App {
    value: bool,
}

impl eframe::App for App {
        ...
        ctx.show_viewport(
            ...,
            |ctx| {
                    ui.add(egui::Slider::new(&mut self.value, 0..=100));
                });
            },
        );
    }
}
...
```

But I see that you restrict the callback in `show_viewport` to be `'static` because you store it in a `Box` later. By looking at the examples in `test_viewport` I noticed that you build state using `egui::Window`/`egui::CentralPanel` and `FnOnce`, makes sense and works fine.

Thank you for the `viewport_command(egui::ViewportCommand::Drag)` snippet, works great!

konkitoman:
> Sure, a minimal example would be something like this:
> 
> ```rust
> ...
> struct App {
>     value: bool,
> }
> 
> impl eframe::App for App {
>     fn update(&mut self, ctx: &egui::Context, frame: &mut eframe::Frame) {
>         ...
>         ctx.show_viewport(
>             ...,
>             |ctx| {
>                 egui::CentralPanel::default().show(ctx, |ui| {
>                     ui.add(egui::Slider::new(&mut self.value, 0..=100));
>                 });
>             },
>         );
>     }
> }
> ...
> ```

That is not possible, read the docs of `Context::show_viewport`
In short the `Context::show_viewport` is like creating a thread!
You can use `Context::show_viewport_immediate` for that to work, but cannot be rendered by itself!
I only recommend ` Context::show_viewport_immediate` for dialog's or experimenting because has a performance impact!

crates/egui-winit/src/lib.rs
```diff
@@ -163,22 +177,81 @@ impl State {
...
+    ) -> egui::RawInput {
         let pixels_per_point = self.pixels_per_point();
 
         self.egui_input.time = Some(self.start_time.elapsed().as_secs_f64());
 
+        // TODO remove this in winit 0.29
+        // This hack make the window outer_position and size to be valid, X11 Only
+        // That was happending because winit get the window state before the compositor adds decorations!
+        #[cfg(feature = "x11")]
+        window.set_maximized(window.is_maximized());
```
pablosichert, line 193: This caused an infinite loop on macOS. `x11` appears to be in the default features. Maybe we can change this to:

```suggestion
        // TODO remove this in winit 0.29
        // This hack make the window outer_position and size to be valid, X11 Only
        // That was happending because winit get the window state before the compositor adds decorations!
        #[cfg(all(feature = "x11", target_os = "linux"))]
        window.set_maximized(window.is_maximized());
```

crates/egui/src/viewport.rs
```diff
@@ -0,0 +1,680 @@
...
+
+    pub close_button: Option<bool>,
+    pub minimize_button: Option<bool>,
+    pub maximize_button: Option<bool>,
+
+    pub hittest: Option<bool>,
+}
```
pablosichert, line 137: There seem to be some options missing in `ViewportBuilder` that are present in `eframe::NativeOptions`, or them having different names, e.g.:

| NativeOptions | ViewportBuilder |
| - | - |
| always_on_top | - |
| centered | - |
| decorat**ed** | decorat**ions** |
| drag_and_drop<strong>_support</strong> | drag_and_drop |
| fullsize_content | fullsize_content<strong>_view</strong> |
| ... | ... |

Are the missing options left out on purpose?

If not, would it be a good idea to share them via a struct `NativeWindowOptions` (this could be a bit challenging since for `NativeOptions` all of them are `T` and for `ViewportBuilder` they are `Option<T>`) or a macro? That way it would also be easier to synchronize and maintain the options in the future.

emilk: I'd like to remove all the window/viewport-related options from `NativeOptions` and use a `ViewportBuilder` there, but this could perhaps wait for a later PR

crates/egui-winit/Cargo.toml
```diff
@@ -49,10 +49,10 @@ puffin = ["dep:puffin", "egui/puffin"]
 serde = ["egui/serde", "dep:serde"]
 
 ## Enables Wayland support.
-wayland = ["winit/wayland"]
+wayland = ["winit/wayland", "bytemuck"]
 
 ## Enables compiling for x11.
-x11 = ["winit/x11"]
+x11 = ["winit/x11", "bytemuck"]
```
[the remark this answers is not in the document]
konkitoman: Because of `ColorImage` for the viewport icon uses `ColorImage::as_raw()`

emilk:
I have a feeling `run.rs` has way too many `Rc` and `RefCell`. Do we really need this much sharing?

pablosichert:
Here's a patch to add `always_on_top` for viewports:

```
diff --git a/crates/egui-winit/src/lib.rs b/crates/egui-winit/src/lib.rs
index 38003b7b..adc6df76 100644
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -1138,6 +1138,12 @@ pub fn create_winit_window_builder(builder: &ViewportBuilder) -> winit::window::
     crate::profile_function!();
 
     let mut window_builder = winit::window::WindowBuilder::new()
+        .with_window_level(
+            builder
+                .always_on_top
+                .map(|_| WindowLevel::AlwaysOnTop)
+                .unwrap_or_default(),
+        )
         .with_title(
             builder
                 .title
diff --git a/crates/egui/src/viewport.rs b/crates/egui/src/viewport.rs
index c073d940..322f3058 100644
--- a/crates/egui/src/viewport.rs
+++ b/crates/egui/src/viewport.rs
@@ -112,6 +112,7 @@ pub struct ViewportBuilder {
     /// This is wayland only. See [`Self::with_name`].
     pub name: Option<(String, String)>,
 
+    pub always_on_top: Option<bool>,
     pub position: Option<Option<Pos2>>,
     pub inner_size: Option<Option<Vec2>>,
     pub fullscreen: Option<bool>,
@@ -145,6 +146,7 @@ impl ViewportBuilder {
             id,
             title: None,
             name: None,
+            always_on_top: None,
             position: None,
             inner_size: Some(Some(Vec2::new(300.0, 200.0))),
             fullscreen: None,
@@ -179,6 +181,7 @@ impl ViewportBuilder {
             id,
             title: None,
             name: None,
+            always_on_top: None,
             position: None,
             inner_size: None,
             fullscreen: None,
@@ -210,6 +213,12 @@ impl ViewportBuilder {
         self
     }
 
+    /// Sets the window always on top.
+    pub fn with_always_on_top(&mut self, always_on_top: bool) -> &mut Self {
+        self.always_on_top = Some(always_on_top);
+        self
+    }
+
     /// Sets whether the window should have a border, a title bar, etc.
     ///
     /// The default is `true`.
@@ -415,6 +424,17 @@ impl ViewportBuilder {
             }
         }
 
+        if let Some(new_always_on_top) = new.always_on_top {
+            if Some(new_always_on_top) != self.always_on_top {
+                self.always_on_top = Some(new_always_on_top);
+                commands.push(ViewportCommand::WindowLevel(if new_always_on_top {
+                    2
+                } else {
+                    0
+                }));
+            }
+        }
+
         if let Some(new_position) = new.position {
             if Some(new_position) != self.position {
                 self.position = Some(new_position);
```

pablosichert:
```rust
ctx.show_viewport(
    &egui::ViewportBuilder::new(egui::ViewportId::from_hash_of("thread_ctx")),
    |ctx| {
            let ctx = ctx.clone();
            println!("viewport id: {:?}", ctx.viewport_id());
            std::thread::spawn(move || {
                println!("thread viewport id: {:?}", ctx.viewport_id());
            });
        });
    },
);
```

prints

```
viewport id: "B17D"
thread viewport id: "0000"
```

Is this intended behavior? I would have expected `thread viewport id == viewport id`.

(For context: In my code I'm handing a `ctx` reference to a thread that repaints a texture at a later point and then calls `ctx.request_paint()`. Nothing happens because here only the parent window (`"0000"`) is repainted.)

emilk:
I meant the other way around @konkitoman - keep a single lock around `GlowWinitRunning` and remove all the smaller locks inside of it.

To avoid deadlock we need to make sure nothing is locked during the call to `egui_ctx.run` inside of `render_immediate_viewport`. This should be easy - we lock the single `GlowWinitRunning` during setup, then unlock it to call `egui_ctx.run`, and then lock it again to process the output. Much nicer than having so many small locks!

konkitoman:
I tried that but then i will not be able to return references like `&EpiIntegration`

I'am trying to remove the Rc<RefCell<T>> from `WinitApp`

crates/eframe/src/native/run.rs
```diff
@@ -157,22 +179,24 @@ fn run_and_return(
                 return;
             }
 
-            // Platform-dependent event handlers to workaround a winit bug
-            // See: https://github.com/rust-windowing/winit/issues/987
-            // See: https://github.com/rust-windowing/winit/issues/1619
```
emilk, line 162: Is this workaround no longer needed? Have you tested this on Windows?

emilk: oh, this is now handled later instead?

crates/eframe/src/native/run.rs
```diff
@@ -233,19 +264,36 @@ fn run_and_return(
...
+                        next_repaint_time.map_or(*repaint_time, |last| last.min(*repaint_time)),
+                    );
+                }
             }
-            next_repaint_time = extremely_far_future();
-            ControlFlow::Poll
-        } else {
+        }
+
+        if let Some(next_repaint_time) = next_repaint_time {
```
emilk, line 291: Something is wrong with this logic: `next_repaint_time` is only ever lowered (`last.min(*repaint_time)`) and never reset. What is it used for? Why do we need it?

Can't we just take the minimum of all `windows_next_repaint_times` instead? That's _almost_ what the code does, but `next_repaint_time` should be decalred  inside the event loop, instead of outside of it

crates/eframe/src/native/run.rs
```diff
@@ -340,19 +397,47 @@ fn run_and_exit(event_loop: EventLoop<UserEvent>, mut winit_app: impl WinitApp +
...
+                    control_flow.set_poll();
+                } else {
+                    next_repaint_time = Some(
+                        next_repaint_time.map_or(*repaint_time, |last| last.min(*repaint_time)),
+                    );
+                }
+            }
+        }
+
+        if let Some(next_repaint_time) = next_repaint_time {
```
emilk, line 424: Same comment here about `next_repaint_time`

crates/eframe/src/native/run.rs
```diff
@@ -823,153 +1226,399 @@ mod glow_integration {
...
+                )
+            };
 
-            egui_glow::painter::clear(
-                gl,
-                screen_size_in_pixels,
-                app.clear_color(&integration.egui_ctx.style().visuals),
-            );
+            // This will only happen if the viewport is sync
+            // That means that the viewport cannot be rendered by itself and needs his parent to be rendered
```
emilk, line 1448: What will only happen?

crates/eframe/src/native/run.rs
konkitoman: I tested on Windows VM and works fine.
And from what i heard from @stefnotch and @workingj works good!

Setting the control flow to Pull when we request a repaint, seams to fix it!

crates/eframe/src/native/run.rs
konkitoman: > What is it used for? Why do we need it?
This is an assumption, Is used to make the event_loop to wait until is needed and then request redraws!
I was there before, so i adapt it to work with multiples viewports.

> Can't we just take the minimum of all windows_next_repaint_times instead?
Yes, that was what that was supposed to do!

crates/eframe/src/native/run.rs
konkitoman: The closure underneath that!

That request a repaint for the parent of that viewport!

emilk:
ab67a310c275d355519632a4ca686adcdd261526 broke clicking things 😬 
I'm investigating

emilk:
I think this is ready to be merged (finally!)

Turun:
Amazing! 

I was very confused why the examples in this Repo did not compile for me, despite being on the latest release of egui/eframe. Then I noticed that this was only merged (at the time of surprise) 4h ago!

In case other people stumble here, you need to switch your `Cargo.toml` to:
```toml
eframe = { git = "https://github.com/emilk/egui/", rev = "83aa3109d31eb7ab09c40530ef4b7d5f3e370fd4" , features = ["persistence"]}
# other egui crates can be used like this too:
egui_plot = { git = "https://github.com/emilk/egui/", rev = "83aa3109d31eb7ab09c40530ef4b7d5f3e370fd4"}
```
(or any commit after this one, so that the multiple windows code is included)

PR #3173, opened 2023-07-23
Fix TextEdit's character limit

Serverator:
# Objective
Fixes a issue with `TextEdit`'s character limit. Currently, the character limit is using bytes instead of chars to count the existing length of the text, which causes incorrect behavior with non-ASCII text. This small PR fixes the issue.

## Issue example
If I set a character limit to 8, `TextEdit` will not allow me to write anything after

- "Language" (8 bytes, 8 chars)
- "Морж" (8 bytes, 4 chars)
- "冰淇淋" (9 bytes, 3 chars)

While still not being at the character limit.


Wumpf (approved):
makes sense! Wondering if there's more of these hidden

PR #3182, opened 2023-07-27
Plot items: Image rotation and size in plot coordinates, polygon fill color 

### LINKED ISSUE #3112: How do i roate a plot image

s-nie:
* `PlotImage` dimensions are now in plot coordinates.
* Plot `Polygon`s can now have a custom fill color, different from the outline color.

Closes https://github.com/emilk/egui/issues/3276


CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`eframe`](crates/eframe/CHANGELOG
...
 ## Unreleased
 
+### ⭐ Added
+
+- Plot images can now be rotated [#3182](https://github.com/emilk/egui/pull/3182)
+
+### 🔧 Changed
+
+- The scale of plot images is now in plot coordinates instead of screen coordinates [#3182](https://github.com/emilk/egui/pull/3182)
+- Plot polygons can now have a custom fill color, replaced `Polygon::fill_alpha` and `Polygon::color` with `Polygon::fill_color` [#3182](https://github.com/emilk/egui/pull/3182)
```
emilk, line 16:
```suggestion
```

I've switched to generating the changelog at release-time using `scripts/generate_changelog.py`.

Please add the above details to the PR description instead!

crates/egui/src/widgets/plot/items/mod.rs
```diff
@@ -1164,6 +1167,12 @@ impl PlotImage {
         self
     }
 
+    /// Rotate the image counter-clockwise around its center.
```
emilk, line 1170: egui uses clock-wise as the standard rotation direction. See for instance `Rot2::from_angle`, `Vec2::angled`, `TextShape::angle`, …

emilk: Also, it is worth noting that this is in radians.

emilk (changes requested):
This has significant overlap with the recently merged PR:
* https://github.com/emilk/egui/pull/3121

crates/egui/src/widgets/plot/items/mod.rs
s-nie: Indeed, but the plot coordinates are already different from the screen coordinates. As the y-axis points up in those coordinates, it would make more sense for the rotation to be counter-clockwise according the the right-hand rule.

emilk: Good point, I didn't think of that!

s-nie:
The highlight outline is now also rotated and I removed the rotation center parameter as it is not very practical the way it it currently implemented. Setting it to anything other than the image's center almost never makes sense in my opinion. What do you think @ThundR67?

ThundR67:
> The image highlight outline is now also rotated and I removed the rotation center parameter as it is not very practical the way it it currently implemented. Setting it to anything other than the image's center almost never makes sense in my opinion. What do you think @ThundR67?

Yup it doesn't. I just added it because egui's normal image does take the center as parameter.

tgross35:
Does this also change stroke and text to use plot coordinates? As in https://github.com/emilk/egui/discussions/3277

s-nie:
>Does the scaling change affect other plot items to use plot coordinates, such as stroke and text? As in https://github.com/emilk/egui/discussions/3277

No this only affects the plot image itself.

PR #3186, opened 2023-07-28
Set the correct unicode character for "ctrl" shortcuts

abey79:
This PR change the character used to display "ctrl" shortcuts from `^` to `⌃`, which is the proper unicode char for that.

Before:

image

After:

image

Fixes rerun-io/rerun#2862


Wumpf (approved):
subtle! :)

martenbjork:
This PR is my happy place ❤️

PR #3219, opened 2023-08-09
Add raw_window_handle to WindowInfo

### LINKED ISSUE #3050: eframe: acrylic/mica/blur background vibrancy

Is there a method to apply window vibrancy effects to an eframe window? [Tauri](https://github.com/tauri-apps/window-vibrancy/blob/dev/src/windows.rs#L64) has an external implementation using the windows-sys crate and by setting the hwnd of the window you want to apply the effect on. 

Has this been done before, and is there then a way to get the eframe hwnd at runtime?

OmegaJak:
All this does is expose the winit window raw_window_handle and put it into the WindowInfo so it can be accessed through the Frame's or CreationContexts's IntegrationInfo.

I'm a bit uncertain about the safety of this since most of `raw_window_handle`'s structs are wrapping `*mut c_void`, but I think it's fine? Any input there would be appreciated.

Closes #3050

### DIFF
--- a/crates/eframe/src/epi/mod.rs
+++ b/crates/eframe/src/epi/mod.rs
@@ -985,6 +985,9 @@ pub struct WindowInfo {
 
     /// Current monitor size in egui points (logical pixels)
     pub monitor_size: Option<egui::Vec2>,
+
+    /// The raw window handle for the main window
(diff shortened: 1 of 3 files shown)


OmegaJak:
I had missed (due to a branching mistake) that this was already implemented in #3073 and is available on current master, closing in favor of that.

Issue #3102, opened 2023-06-21
On wasm32, egui::mutex::Mutex<T> requires T: Sync to be Sync


NickHu:
This is inherited from the underlying implementation for wasm32, from `atomic_refcell`, and means that you can't do the standard pattern of sharing `Send` values between threads by wrapping them in an `Arc<Mutex<_>>`.

Issue #1401, opened 2022-03-22
Firefox panic with parking_lot (Parking not supported on this platform)


Titaniumtown:
**Describe the bug**
With the latest egui commit, egui randomly fails with the error "Parking not supported on this platform" in relation to `parking_lot`.

**To Reproduce**
1. Browse to egui site using latest commit of egui with Firefox browser
2. Interact with site for a bit
3. Observe panic in browser console

**Expected behavior**
The site should run fine on Firefox

**Screenshots**
![image](https://user-images.githubusercontent.com/11786225/159482979-0b7c315a-c243-4869-bf71-f8f379b9ce91.png)

**Desktop (please complete the following information):**
 - OS: `Arch Linux`
 - Browser: Firefox
 - Version: 100 (nightly)

Titaniumtown:
This is on firefox though.

emilk:
It's a bit dissapointing (and surprising) that you don't get a proper Rust callstack in the console - or did you accidentally crop that from the image?

emilk:
Please try out https://github.com/emilk/egui/pull/1404 @Titaniumtown

imxood:
Using latest code, and this is Ok,
OS: Deepin Linux
Browser: Firefox
![image](https://user-images.githubusercontent.com/14890680/159506190-10f7c142-af06-4b7a-a5d2-a0a753b6a123.png)

PR #3236, opened 2023-08-11
Support multi-threaded Wasm

emilk:
Replace `atomic_refcell` with `parking_lot` on wasm32.

`parking_lot` has had problems running on wasm32 before (https://github.com/emilk/egui/issues/1401) but it works these days.
If we have problems again we can always switch to `std::sync::Mutex`.

Closes https://github.com/emilk/egui/issues/3102

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -327,12 +327,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atomic_refcell"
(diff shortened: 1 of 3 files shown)


juancampa:
Nevermind!

This panic happened when the code was about to deadlock. Maybe this is how deadlock detection manifests in WASM right now?

PR #3238, opened 2023-08-11
Refactor: Clarify os detection

emilk:
We had a bunch of `cfg!(windows)` which should have been `cfg!(target_os = "windows")`.

I wonder what the effects of this PR will be for Windows 😬

**EDIT**: Apparently I was confused. `cfg!(windows)` acts exactly like `cfg!(target_os = "windows")`.

However, `cfg!(macos)` is always `false` while `cfg!(target_os = "macos")` works :man-facepalming: 

https://twitter.com/ernerfeldt/status/1690010817415041024?s=20

I'll merge this anyway so I won't get confused again

### DIFF
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -145,12 +145,12 @@ fn run_and_return(
             // Platform-dependent event handlers to workaround a winit bug
             // See: https://github.com/rust-windowing/winit/issues/987
             // See: https://github.com/rust-windowing/winit/issues/1619
-            #[cfg(windows)]
+            #[cfg(target_os = "windows")]
             winit::event::Event::RedrawEventsCleared => {
                 next_repaint_time = extremely_far_future();
                 winit_app.run_ui_and_paint()
             }
-            #[cfg(not(windows))]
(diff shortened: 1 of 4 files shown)


crates/eframe/src/native/run.rs
```diff
@@ -270,11 +270,11 @@ fn run_and_exit(event_loop: EventLoop<UserEvent>, mut winit_app: impl WinitApp +
             // Platform-dependent event handlers to workaround a winit bug
             // See: https://github.com/rust-windowing/winit/issues/987
             // See: https://github.com/rust-windowing/winit/issues/1619
-            winit::event::Event::RedrawEventsCleared if cfg!(windows) => {
+            winit::event::Event::RedrawEventsCleared if cfg!(target_os = "windows") => {
```
emilk, line 273: I suspect we could remove this special handling of `RedrawEventsCleared` here and elsewhere, since it has effectively been disabled for a long time due to the `if cfg!(windows)` mistake. Maybe someone on Windows can test and verify?

emilk: Let's remove it when we update winit - the event loop is completely rewritten in the next winit release

PR #3241, opened 2023-08-11
Fix iOS support in `eframe`

lucasmerlin:
This allows egui and eframe to work correctly on iOS. 

Without these changes it wouldn't compile because winit doesn't support run_and_return on iOS. 
Also request_redraw was not working, seems like WaitUntil was not working correctly, I'm not sure why though. My fix is to repaint immideately but that might need to be investigated some more.

Then, after changing screen orientation, the screen wouldn't return to full size, because egui was trying to restore the previous window size. I've disabled this behaviour for iOS since it doesn't seem like it would make sense there anyways.

### DIFF
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -1057,6 +1057,7 @@ mod glow_integration {
         mut native_options: epi::NativeOptions,
         app_creator: epi::AppCreator,
     ) -> Result<()> {
+        #[cfg(not(target_os = "ios"))]
         if native_options.run_and_return {
             with_event_loop(native_options, |event_loop, native_options| {
                 let glow_eframe =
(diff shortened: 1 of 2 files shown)


crates/eframe/src/native/run.rs
```diff
@@ -280,6 +281,10 @@ fn run_and_exit(event_loop: EventLoop<UserEvent>, mut winit_app: impl WinitApp +
             }
 
             winit::event::Event::UserEvent(UserEvent::RequestRepaint { when, frame_nr }) => {
+                // WaitUntil seems to not work on iOS
+                #[cfg(target_os = "ios")]
+                winit_app.window().unwrap().request_redraw();
```
emilk, line 286: I think it makes more sense to move this to line 340 (covers more cases)

emilk (approved):
Thank you!

Issue #3129, opened 2023-07-01
Web Only: add option to open hyperlink in a new browser tab when using via eframe


FreddyFunk:
It is currently impossible to open a hyperlink without leaving the current app when running a egui/eframe web app. This is because text is not selectable and right clicks on hyperlinks with an option to open in a new tab is currently not supported.

I would prefer the following two-part solution:

1. When right clicking any hyperlink it shows the option to be opened in a new browser tab.
2. `egui:ui:Ui:hyperlink` and `egui:ui:Ui:hyperlink_to` get an optional boolean parameter called `open_in_new_tab`. When this boolean is set to `true`, a normal left click opens the hyperlink in a new tab as well. This bool can be set to `false` by default to do not break any backwards compatibility. This boolen will also not affect egui when running on native OS, since hyperlinks are opened as a new tab anyway in this case.

PR #3242, opened 2023-08-11
Add option to always open hyperlink in a new browser tab

FreddyFunk:
Partially Closes <https://github.com/emilk/egui/issues/3129>.


emilk (approved):
Nice!

ctaggart:
Thanks for this! I was looking for how to use it and figured it out:

``` rust
ui.hyperlink_to(label, url); // opens in same tab
ui.add(Hyperlink::from_label_and_url(label, url).open_in_new_tab(true)); // opens in new tab
```

Yes, it would be cool if the default was a global option.

Issue #1727, opened 2022-06-10
Option to truncate text in Label


AlanRace:
Label currently ignores the max size when adding via `add_sized`. This only seems to be taken into account when `wrap` is set to `true`. I would ideally like to set a maximum size for the label and then simply have it not display any text that doesn't fit in the specified width (possibly replacing it with `...` to indicate the full label is not being displayed). 

```rust
ui.horizontal(|ui| {
    ui.add_sized([100.0, 10.0], egui::Label::new("My very long description"));
    ui.button("Remove");
}
```

Regardless of the specified width in `add_sized`, the label is displayed in full.

As a proposed solution, perhaps an API similar to the below:

```rust
/// Describes how to handle exceeding allocated size for the widget (label)
enum ExceedDimension {
     /// Force extend the dimension (e.g. width/height) of the label so that all text is displayed
     Extend,
     /// Wrap the text so that it fits within the specified width, but 
     /// potentially extends the height past the requested size
     Wrap,
     /// Truncate text so that it fits within the specified bounds
     Truncate,
}

// ...

ui.horizontal(|ui| {
    ui.add_sized(
        [100.0, 10.0], 
        egui::Label::new("My very long description").on_exceed_width(ExceedDimension::Truncate)
    );
    ui.button("Remove");
}
```

Alternatively, maybe there is a different way to achieve what I am trying to achieve - essentially as in the example above, I would like to always have the button visible at the cost of the full description. This appears in a `SidePanel` so I want the width to be tied to the current width of the `SidePanel` (minus the width of the button) so that the button always appears at the very right of the screen/panel.

emilk:
Truncating with `…` would be a nice option to have.

PR #3244, opened 2023-08-12
Add option to truncate text at wrap width

emilk:
* Closes https://github.com/emilk/egui/issues/1727

You can now easily elide text that doesn't fit a certain max width.

This is done using `Label::new(long_text).truncate(true)`. The text will be truncated to fit the max width of the parent `Ui`, and the [`…` ellipsis](https://www.compart.com/en/unicode/U+2026) will replace the omitted text. The text may get truncated in the middle of a word. When hovering a label with elided text, the full non-elided text will be shown in a tool-tip.

You can use the options in `TextWrapping` for finer control.

⚠️ BREAKING: `TextWrapping::max_rows == 0` will now output zero rows. Use `usize::MAX` to set no limit.
⚠️ BREAKING: `TextWrapping::max_rows` will now count total rows (not rows per paragraph, as previous).
ADDED: `Galley::elided`, to see if the text layout elided the input text

Most of the implementation had already be done by @awaken1ng in https://github.com/emilk/egui/pull/1291

### Future work
There are normally three modes of elision that can be desired:

* Start:  `… jumps over the lazy dog`
* Middle: `The quick … the lazy dog`
* End: `The quick brown fox jumps… `

This PR only adds support for the latter kind.

It would also be good to have an option to only omit whole words, never breaking them.

Finally, being able to set the truncate character to a string (e.g. ` …` with a leading space) would be nice.

### DIFF
--- a/crates/egui/src/widget_text.rs
+++ b/crates/egui/src/widget_text.rs
@@ -676,7 +676,7 @@ impl WidgetTextGalley {
         self.galley.size()
     }
 
-    /// Size of the laid out text.
+    /// The full, non-elided text of the input job.
     #[inline]
     pub fn text(&self) -> &str {
         self.galley.text()
--- a/crates/egui/src/widgets/label.rs
+++ b/crates/egui/src/widgets/label.rs
@@ -194,8 +194,12 @@ impl Label {
 
 impl Widget for Label {
     fn ui(self, ui: &mut Ui) -> Response {
-        let (pos, text_galley, response) = self.layout_in_ui(ui);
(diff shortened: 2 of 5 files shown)


abey79 (approved):
Great feature! Nothing strikes me in particular with the implementation. I just have a weak preference for `truncate`instead of `elide` for the API until we have various forms of elision.

paul-hansen:
Just a heads up, this breaking change didn't seem to make it into the changelog for 0.22.0 in epaint or egui. Took a bit to figure out why text wasn't showing, but not the end of the world. Might want check to make sure whatever processes for updating the changelog is working for future breaking changes though.

PR #3245, opened 2023-08-12
Add tooltip_delay option

### LINKED ISSUE #3232: Allow showing tooltips with a slight time delay
The current way to show a tooltip, as far as I can tell, is to use [on_hover_text](https://docs.rs/egui/latest/egui/struct.Response.html#method.on_hover_text) and friends.

These methods show a custom tooltip almost immediately when an item is hovered. This might be desired in some cases, but it can also be distracting because the tooltip is shown even when the mouse just quickly crosses an item. That's why tooltips are usually shown after a small time delay so that they only show up when actually needed.

I have no idea about the implementation, but I think it would be nice to have some methods like `on_hover_text_delayed` or similar that show a tooltip only after a certain time of continuous hovering. The time delay could be part of the style.

YgorSouza:
Similar to the show_tooltips_only_when_still option, but allows the user to configure a specific delay in seconds, and also makes the tooltip disappear if the mouse is moved again.

Closes #3232

### DIFF
--- a/crates/egui/src/response.rs
+++ b/crates/egui/src/response.rs
@@ -435,14 +435,11 @@ impl Response {
             }
         }
 
-        if self.ctx.style().interaction.tooltip_delay > 0.0
-            && self.ctx.style().interaction.tooltip_delay
-                > self.ctx.input(|i| i.pointer.time_since_last_movement())
+        if !self.is_tooltip_open()
+            && self.ctx.input(|i| i.pointer.time_since_last_movement())
+                < self.ctx.style().interaction.tooltip_delay
         {
-            // If a non-zero delay was specified, we only show the tooltip if the mouse has been
-            // still for the specified amount of time. If the mouse is moved, the tooltip
-            // disappears.
-
+            // Keep waiting until the mouse has been still for a while
             self.ctx.request_repaint();
             return false;
         }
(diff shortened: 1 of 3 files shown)


crates/egui/src/response.rs
```diff
@@ -435,6 +435,18 @@ impl Response {
...
         }
 
+        if self.ctx.style().interaction.tooltip_delay > 0.0
+            && self.ctx.style().interaction.tooltip_delay
+                > self.ctx.input(|i| i.pointer.time_since_last_movement())
+        {
+            // If a non-zero delay was specified, we only show the tooltip if the mouse has been
+            // still for the specified amount of time. If the mouse is moved, the tooltip
+            // disappears.
+
```
emilk, line 445: As it stands, `tooltip_delay` has double duty:

* Controls delay
* Controls wether or not to close a tooltip by just moving the mouse (when > 0)

Closing tooltips on mouse movement is quite uncommon, and I suggest we never do it (or enable it via its own `bool`).


```suggestion
        if !self.is_tooltip_open()
            && self.ctx.input(|i| i.pointer.time_since_last_movement())
                < self.ctx.style().interaction.tooltip_delay
        {
            // Keep waiting until the mouse has been still for a while
```

emilk (changes requested):
Thanks for working on this!

emilk (approved):

PR #3247, opened 2023-08-13
Fix consistency of lost_focus and gained_focus

### LINKED ISSUE #2142: lost_focus on a single_text_edit widget does not get called in several cases

phire:
This PR aims to make `focus.id` stay consistent throughout the entire frame.

Otherwise `gained_focus` and `lost_focus` can sometimes produce inconsistent results as new widgets are added.

In my testing, this fixes #2142, where `lost_focus` on TextEdit widgets wouldn't fire if you clicked on a TextEdit widget that was added to the ui later than the currently focused widget. I suspect it fixes other weird bugs too.

Closes <https://github.com/emilk/egui/issues/2142>.

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -1250,7 +1250,7 @@ impl Context {
                     nodes,
                     tree: Some(accesskit::Tree::new(root_id)),
                     focus: has_focus.then(|| {
-                        let focus_id = self.memory(|mem| mem.interaction.focus.id);
+                        let focus_id = self.memory(|mem| mem.interaction.focus.focused());
                         focus_id.map_or(root_id, |id| id.accesskit_id())
                     }),
                 });
--- a/crates/egui/src/memory.rs
+++ b/crates/egui/src/memory.rs
@@ -185,7 +185,10 @@ pub(crate) struct Focus {
     id_previous_frame: Option<Id>,
 
     /// Give focus to this widget next frame
-    id_next_frame: Option<Id>,
+    /// We want `id` to stay consistent for the entire frame, so we only update it
+    /// at the start of the frame. Otherwise `gained_focus` and `lost_focus` will
+    /// produce inconsistent results.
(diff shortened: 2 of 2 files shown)


phire:
This change probably causes some regressions, but I don't know the codebase enough to know where to look or how to test.

But one I've noticed: Previously you could both focus and highlight a range in a TextEdit with a single click-and-drag. Now clicking and dragging prevents the TextEdit from gaining focus, you need to focus then highlight.

Because the code that calls `stop_text_input` is relying on focus.id to be updated immediately in order to avoid matching on the currently selected element.

phire:
What's the desired way to deal with this kind ofAPI breakage? I see two potential options:

1. Fix all supplied widgets and accept that some external code will break.
2. Explicitly define `Memory.focus()` and `Memory.has_focus()` as keeping their current broken behaviour, allowing widgets that requests focus then check focus on the same frame to continue working. 
We could also define new methods that always report which widget that had focus at the beginning of the frame.

PR #3253, opened 2023-08-14
Improved wgpu callbacks

### LINKED ISSUE #3163: Painter's `CallbackFn` redesign
It is a little hard to describe but I'll try my best. It is based on wgpu backend but glow could has similar points.

First of all example for using glow/wgpu directly is way too simple. It covers at max one control and stores all data in `paint_callback_resources` which is not always possible when using multiple controls. For example it is not possible to have custom UVs per-control because there is only a single vertex buffer stored `globally` in context.

Currently it is somewhat hard to implement custom controls that require its own retainable data (for ex. uniform/vertex buffers). We can register our 'global' context in the `paint_callback_resources` during App initialization and use it as storage for uniforms/vertex/shaders/etc.

Ideally we could have a big buffer for all controls and just update it in `CallbackFn::prepare` (while storing offset) and than just bind it in `CallbackFn::paint` with stored offset. Unfortunately unlike in egui itself we don't know how big the buffer must because we don't know how many controls we have. This can be worked around by using per-control uniform/vertex buffer instead of single huge one.

So possible solution is to:
- register global object in `paint_callback_resources`
- write all the data to global object's intermediate Vec<u8> in `CallbackFn::prepare`, and store offsets in per-control shared data
- between `CallbackFn::prepare` and `CallbackFn::paint` reallocate uniform/vertes buffers (if needed) and upload data (similar to what egui does)
- simply schedule draw calls in `CallbackFn::paint` with all of the offsets above

There are few problems:
1. we don't have a callback between of `prepare`/`paint` to update all buffers. we can instead use a boolean flag which will be reset in `CallbackFn::prepare` and perform buffer updates in the first call to `CallbackFn::paint` but `paint_callback_resources` are readonly there.
2. the only way to pass per-control data from `CallbackFn::prepare` to `CallbackFn::paint` is to use shared arc-ed struct. so all offsets need to be stored atomically.
3. if we had a single trait instead of two callbacks, we could store said offsets in the object implementing it. but because  `CallbackFn` itself is stored in an Arc, mutable access to this object is not possible (so again... atomics).

Ideally:
1. store `CallbackFn` in `PaintCallback` as Box<dyn Any> instead of Arc<dyn Any>
2. combine `CallbackFn::prepare` and `CallbackFn::paint` into a single trait (also reduces number of allocations)
3. add additional `&mut self` callback between `CallbackFn::prepare` and `CallbackFn::paint` to know when we can upload intermediate buffers to final buffers.

Realistically:
I guess 1 is very painfull to implement. so we can at the very try to least implement 2 & 3. those are much less painfull:
```rust
// old
#[derive(Clone)]
pub struct CallbackFn {
    prepare: Box<PrepareCallback>,
    paint: Box<PaintCallback>,
}

// new
// ideally wouldn't need it but `dyn Any` can't be case to `dyn Trait` in Rust, `CallbackFn` is used just as a wrapper:
pub struct CallbackFn {
    implementation: Box<dyn CallbackFnTrait>
}

pub trait CallbackFnTrait : Send + Sync {
    fn prepare(
        &self,
        device: &wgpu::Device,
        queue: &wgpu::Queue,
        encoder: &mut wgpu::CommandEncoder,
        resources: &mut TypeMap,
    ) -> Vec<wgpu::CommandBuffer>;

    fn finish_prepare(
        &self,
        device: &wgpu::Device,
        queue: &wgpu::Queue,
        encoder: &mut wgpu::CommandEncoder,
        resources: &mut TypeMap,
    ) -> Vec<wgpu::CommandBuffer>;

    fn paint<'a>(
        &self,
        info: PaintCallbackInfo,
        render_pass: &mut wgpu::RenderPass<'a>,
        resources: &'a TypeMap,
    );
}
```

Drawbacks:
- a breaking change
- requires GATs, so MSRV must be increased to 1.65

Just allocate shared state and callbacks separately, per-control unified/vertex buffers.

**Additional context**
This is only a potential improvement that fits my use case, it might not be worth overall but lets see.

Wumpf:
⚠️ BREAKING: Callbacks for egui_wgpu have been reworked.
* Registering a callback now requires implementing `egui_wgpu::CallbackTrait`
* new `CallbackTrait::finish_prepare` called after all prepare callbacks have been called
* resources shared between callbacks:
  - are now stored in `egui_wgpu::Renderer::shared_paint_callback_resources`
  - are not required to be `Send`/`Sync` on wasm

---------

* Fixes #3163

I followed @MatrixDev's proposal fairly closely. Key differences:

* only a single `finish_prepare_callback` that is not part of the callback trait. For convenience it still gets references to all callback objects. This technically makes the regular `prepare` callbacks redundant, but I figured it's convenient to have both around. I think @MatrixDev correctly describes the typical scenario as each individual `prepare` pushing to some form of system/manager and then processing the overall results in a `finish_prepare
* kicked out `TypeMap` and replaced with integer hash map since:
   * no longer need to push objects in a singleton-style way. The new map is slightly harder to use but a lot more flexible
   * needed to do something to support non Sync/Send going forward anyways since on wasm wgpu resources aren't Send/Sync by default (since they contain references to JS objects)
   * one dependency less!

I ventured a little bit into trying to change egui's callback system on a more deeper level as hinted in the ticket, but I concluded the only way forward there would be to define _separate_ callback data on the `Shape` (always sync+send, immutable, comparable etc.) and separately callback "systems". I think it should be easier to transition there with this PR in the future but this would be a much more far-reaching change.

### DIFF
--- a/crates/epaint/src/shape.rs
+++ b/crates/epaint/src/shape.rs
@@ -817,7 +817,8 @@ pub struct PaintCallback {
     ///
     /// The rendering backend is also responsible for restoring any state, such as the bound shader
     /// program, vertex array, etc.
-    pub callback: Arc<dyn Any + Sync + Send>,
+    /// TODO: update doc
+    pub callback: Arc<dyn Any + Send + Sync>,
 }
 
 impl std::fmt::Debug for PaintCallback {
--- a/crates/epaint/src/shape.rs
+++ b/crates/epaint/src/shape.rs
@@ -817,6 +817,8 @@ pub struct PaintCallback {
     ///
     /// The rendering backend is also responsible for restoring any state, such as the bound shader
     /// program, vertex array, etc.
+    ///
+    /// Shape has to be clone, therefore this has to be an `Arc` instead of a `Box`.
     pub callback: Arc<dyn Any + Send + Sync>,
 }
 
--- a/crates/epaint/src/shape.rs
+++ b/crates/epaint/src/shape.rs
@@ -807,7 +807,7 @@ pub struct PaintCallback {
     ///
     /// The concrete value of `callback` depends on the rendering backend used. For instance, the
     /// `glow` backend requires that callback be an `egui_glow::CallbackFn` while the `wgpu`
-    /// backend requires a `egui_wgpu::CallbackFn`.
+    /// backend requires a `egui_wgpu::Callback`.
     ///
     /// If the type cannot be downcast to the type expected by the current backend the callback
     /// will not be drawn.
@@ -817,7 +817,6 @@ pub struct PaintCallback {
     ///
     /// The rendering backend is also responsible for restoring any state, such as the bound shader
     /// program, vertex array, etc.
-    /// TODO: update doc
     pub callback: Arc<dyn Any + Send + Sync>,
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1193,7 +1193,6 @@ dependencies = [
  "log",
  "puffin",
  "thiserror",
- "type-map",
  "wgpu",
  "winit",
 ]
@@ -3707,15 +3706,6 @@ dependencies = [
  "wasm-bindgen",
  "web-sys",
  "windows 0.48.0",
-]
-
-[[package]]
-name = "type-map"
-version = "0.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "deb68604048ff8fa93347f02441e4487594adc20bb8a084f9e564d2b827a0a9f"
-dependencies = [
- "rustc-hash",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1193,6 +1193,7 @@ dependencies = [
  "log",
  "puffin",
  "thiserror",
+ "type-map",
  "wgpu",
  "winit",
 ]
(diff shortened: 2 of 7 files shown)


crates/epaint/src/shape.rs
```diff
@@ -817,7 +817,7 @@ pub struct PaintCallback {
     ///
     /// The rendering backend is also responsible for restoring any state, such as the bound shader
     /// program, vertex array, etc.
-    pub callback: Arc<dyn Any + Sync + Send>,
+    pub callback: Arc<dyn Any + Send + Sync>,
```
Wumpf, line 820: accidental reordering of traits due to intra-pr-historical reasons ;)

MatrixDev:
> @MatrixDev, want to have a look if this fixes the issue for you? Doesn't address glow backend, but we can take that separately

Wow, this is some great work and it will fully cover my usecase.

The only thing I'm not sure about is `Renderer::finish_prepare_callback`. Don't get me wrong, it fits me 100% and is probably the most performant way to implement it. But having a single callback for all controls reduces ease of use for 3rd-party library controls (when they inevitably appear in the future). Instead of just calling extension function on `UI` users would also be required to provide manual implementation of `Renderer::finish_prepare_callback` which will group callbacks by controls and than call `finish_prepare_callback` for a specific control from a 3rd-party library.

emilk (approved):

Issue #721, opened 2021-09-14
Make rounded corners clip the image in ImageButton


Kesanov:
Currently, the image remains in its original shape - usually rectangular.

Furthermore, the border is painted under the image.
If it had been painted on top, it would appear to clip the image beneath.

Kesanov:
Additionally, the border seems to be few pixels wider than it should:
![image](https://user-images.githubusercontent.com/22682767/133285357-f4dcaf22-1fea-430f-85f9-685a23019b00.png)

PR #3257, opened 2023-08-14
Support images with rounded corners

emilk:
* Closes https://github.com/emilk/egui/issues/721

You can now control the rounding of images using `Image::rounding`. This is implemented by augmenting `RectShape` with texture support. The corners are anti-aliased.

Example output:
image


crates/epaint/src/shape.rs
```diff
@@ -476,17 +480,45 @@ impl From<PathShape> for Shape {
...
+    /// If the rect should be filled with a texture, which one?
+    ///
+    /// The texture is multiplied with [`Self::fill`].
+    pub fill_texture_id: TextureId,
+
+    /// What UV coordinates to use for the texture?
+    ///
+    /// To display a texture, set [`Self::fill_texture_id`],
+    /// and set this to `Rect::from_min_max(pos2(0.0, 0.0), pos2(1.0, 1.0))`.
+    pub uv: Rect,
```
abey79, line 501: I assume this is to spare the `Option` on fill_texture_id, and the assorted 1 byte + padding impact on the struct size?

abey79 (approved):
So timely :) Looks great 👍🏻

crates/epaint/src/shape.rs
emilk: That, and for symmetry with the backend where the top-left texel of the default texture is always white

PR #3267, opened 2023-08-20
Added Context::is_context_menu_open()

dmlary:
I encountered a case where I needed to know if an egui context menu was open, even if the mouse cursor was not in an egui area (full details in discord: https://discord.com/channels/900275882684477440/1142659953102966884).

I found that every resource I needed to detect that the context menu was open was either private, or pub(crate).  Also, the `Response` returned by `Response::context_menu()` contains no way to detect that the context menu was opened or closed, so there's nothing to hook there.

While it is possible to wrap the `Response::context_menu()` in code to do state-tracking, it becomes necessary to duplicate that code in every place you call `context_menu()`.

In this commit, I add `Context::is_context_menu_open()`.  Named similarly to `Context::is_pointer_over_area()`, this method will return true if any context menu is open.  This is possible because the context menu uses a temp var with a fixed Id to store its state.  This method just fetches the state, and interrogates it to see if there is a menu present.

One `pub(crate)` helper method, `BarState::has_root()`, was added as all the fields needed to perform the check were private to the `menu` module.

I've also updated the Context Menu demo to show the result of `is_context_menu_open()` to verify the code functions as expected.

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -1389,6 +1389,14 @@ impl Context {
     pub fn highlight_widget(&self, id: Id) {
         self.frame_state_mut(|fs| fs.highlight_next_frame.insert(id));
     }
+
+    /// Is an egui context menu open?
+    pub fn is_context_menu_open(&self) -> bool {
+        self.data(|d| {
+            d.get_temp::<crate::menu::BarState>(menu::CONTEXT_MENU_ID_STR.into())
+                .map_or(false, |state| state.has_root())
+        })
+    }
 }
 
 // Ergonomic methods to forward some calls often used in 'if let' without holding the borrow
--- a/crates/egui/src/menu.rs
+++ b/crates/egui/src/menu.rs
@@ -48,6 +48,10 @@ impl BarState {
         MenuRoot::stationary_click_interaction(response, &mut self.open_menu, response.id);
         self.open_menu.show(response, add_contents)
     }
+
+    pub(crate) fn has_root(&self) -> bool {
+        self.open_menu.inner.is_some()
+    }
 }
 
 impl std::ops::Deref for BarState {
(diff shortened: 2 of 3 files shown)


dmlary:
Video of updated demo; github wouldn't inline it in the description.
https://github.com/emilk/egui/assets/857742/9233bf6e-8ccd-41b8-86ae-fba06e2fa698

dmlary:
And here's the equivalent code needed to track the context_menu state from outside of egui:

```rust
        // create an area covering the entire screen that we use to detect
        // right-click to open the context menu.
        egui::Area::new("context_menu_area")
            // set the area as background so 'is_pointer_over_area()' ignores
            // this area when checking the pointer location.
            .order(egui::Order::Background)
            // set the position to (0, 0) to override padding/spacing and cover
            // the entire screen
            .fixed_pos([0.0, 0.0])
            .show(ctx, |ui| {
                // We need to know when the context menu is open so we don't
                // move the tile-selection cursor.  There is no mechanism in
                // egui to do this, so instead we need to manually track it
                // ourselves.
                //
                // Assume the menu is always closed; update it to open when it
                // is drawn.  If the menu state doesn't match the UiState value
                // at the end, update the UiState.
                let mut menu_is_open = false;

                // create a response covering the entire screen to detect click
                // for the context menu
                ui.allocate_response(ui.available_size(), egui::Sense::click())
                    .context_menu(|ui| {
                        let mut close_menu = false;
                        if ui.button("Link to new Chunk").clicked() {
                            close_menu = true;
                        }
                        if ui.button("Link to Existing Chunk").clicked() {
                            close_menu = true;
                        }
                        if ui.button("Link to Existing Tile").clicked() {
                            close_menu = true;
                        }

                        if close_menu {
                            ui.close_menu();
                        } else {
                            menu_is_open = true;
                        }
                    });

                // if the menu state changed, upsdate the ui state
                if menu_is_open != ui_state.context_menu {
                    debug!("context menu state change: {:?}", menu_is_open);
                    world.resource_mut::<UiState>().context_menu = menu_is_open;
                }
            });
```

emilk (approved):

PR #3272, opened 2023-08-23
Change focused widget with arrow keys

### LINKED ISSUE #1250: Gampad and arrow-keys navigation support

I'd like to be able to navigate on the gui elments using a gamepad

TimonPost:
We have the use case of gamepad control for a user-facing game menu. This PR makes it possible to change widget focus with the keyboard arrows. It allows integrations to convert joysticks to arrow keys in order to toggle between menu items. It does not add support for gamepad events in egui as this seems like a bigger topic that I desire not to touch. 

Initially, I examine the currently focused widget to determine the one with the highest dot product when compared to the desired focus direction. Subsequently, I identify the widget that is nearest to the current one. This process enables the selection of the widget in the desired direction which not only possesses the smallest angle relative to the current widget but also maintains the closest proximity.

The widget selection code can be improved over time ofcourse, there are some potential issues right now:
- This solution has no sense of the window it exists in, so if a focusable widget in another window is a better candidate it will pick that one.  
- If I click in an input box the `Focus::interested_in_focus` is not called and thus I do not know what the interactable widgets and their rectangles are. I only know this after using a key like arrow or tab (+shift). 

Closes #1250

### DIFF
--- a/crates/egui/src/memory.rs
+++ b/crates/egui/src/memory.rs
@@ -428,8 +428,8 @@ impl Focus {
             return None;
         };
 
-        // Check if the widget has the right dot product and length.
-        let focus_direction = match self.focus_direction {
+        // In what direction we are looking for the next widget.
+        let search_direction = match self.focus_direction {
             FocusDirection::Up => Vec2::UP,
             FocusDirection::Right => Vec2::RIGHT,
             FocusDirection::Down => Vec2::DOWN,
@@ -459,17 +459,22 @@ impl Focus {
                 continue;
             }
 
+            // There is a lot of room for improvement here.
             let to_candidate = vec2(
                 range_diff(candidate_rect.x_range(), current_rect.x_range()),
                 range_diff(candidate_rect.y_range(), current_rect.y_range()),
             );
 
-            let acos_angle = to_candidate.normalized().dot(focus_direction);
+            let acos_angle = to_candidate.normalized().dot(search_direction);
 
-            // Only interested in widgets that fall in a 90° cone (+/- 45°)
-            if 0.5_f32.sqrt() <= acos_angle {
+            // Only interested in widgets that fall in a 90° cone (±45°)
+            // of the search direction.
+            let is_in_search_cone = 0.5_f32.sqrt() <= acos_angle;
+            if is_in_search_cone {
                 let distance = to_candidate.length();
-                let score = distance / acos_angle;
+
+                // There is a lot of room for improvement here.
+                let score = distance / (acos_angle * acos_angle);
 
                 if score < best_score {
                     best_score = score;
--- a/crates/egui/src/memory.rs
+++ b/crates/egui/src/memory.rs
@@ -440,14 +440,14 @@ impl Focus {
                 continue;
             }
 
-            let x_overlaps = -current_focus_widget_rect
+            let x_similarity = -current_focus_widget_rect
                 .x_range()
-                .overlaps(widget_rect.x_range());
-            let y_overlaps = -current_focus_widget_rect
+                .similarity(widget_rect.x_range());
+            let y_similarity = -current_focus_widget_rect
                 .y_range()
-                .overlaps(widget_rect.y_range());
+                .similarity(widget_rect.y_range());
 
-            let current_to_candidate = vec2(x_overlaps, y_overlaps);
+            let current_to_candidate = vec2(x_similarity, y_similarity);
 
             let dot_current_candidate = current_to_candidate.normalized().dot(focus_direction);
             let distance_current_candidate = current_to_candidate.length();
--- a/crates/egui/src/memory.rs
+++ b/crates/egui/src/memory.rs
@@ -93,22 +93,22 @@ pub struct Memory {
 
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 pub enum FocusDirection {
-    // Select the widget closest above the current focused widget.
+    /// Select the widget closest above the current focused widget.
     Up,
 
-    // Select the widget to the right of the current focused widget.
+    /// Select the widget to the right of the current focused widget.
     Right,
 
-    // Select the widget below the current focused widget.
+    /// Select the widget below the current focused widget.
     Down,
 
-    // Select the widget to the left of the the current focused widget.
+    /// Select the widget to the left of the the current focused widget.
     Left,
 
-    // Select the previous widget that had focus.
+    /// Select the previous widget that had focus.
     Previous,
 
-    // Select the next widget that wants focus.
+    /// Select the next widget that wants focus.
     Next,
 
     /// Don't change focus.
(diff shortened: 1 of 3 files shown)


crates/egui/src/memory.rs
```diff
@@ -322,31 +355,121 @@ impl Focus {
...
+            Vec::with_capacity(self.focus_widgets_cache.len());
+
+        for (widget_id, widget_rect) in &mut self.focus_widgets_cache {
+            if Some(*widget_id) == self.id {
+                continue;
+            }
+
+            let candidate_widget_pos = widget_rect.left_center();
+
+            let current_to_candidate = candidate_widget_pos - current_widget_pos;
```
emilk, line 423: This could be improved. Instead of taking an arbitrary point (`left_center`) in the rect, we should consider the full rects.

For instance, these right-aligned rects should be considered to be perfectly above each other:

```
[ first widget ]
     [ another ]
        [ last ]
```

We can accomplish this with a function that checks the x-ranges and y-ranges and if they overlap, returns 0 for that axis.

```rs
/// * negative if `a` is strictly left of `b`
/// * positive if `a` is strictly right of `b`
/// * zero if the ranges overlap
fn range_diff(a: Rangef, b: Rangef) -> f32 {
    if a.max < b.min {
        a.max - b.min
    } else if b.max < a.min {
        a.min - b.max
    } else {
        0.0 // overlaps
    }
}

…

let current_to_candate = vec2(
    range_diff(selected_rect.x_range(), candidate_rect.x_range(),
    range_diff(selected_rect.y_range(), candidate_rect.y_range(),
);
```

I'm not a 100% this is an improvement, but I believe so.

examples/hello_world/src/main.rs
```diff
@@ -17,14 +17,18 @@ fn main() -> Result<(), eframe::Error> {
 
 struct MyApp {
     name: String,
+    name2: String,
     age: u32,
+    age2: u32,
```
emilk, line 22: please revert the changes in this file. Use `egui_demo_app` (`cargo r`) for a better test of your code!

emilk (changes requested):
Exciting!

The navigation works okish, but I left some ideas on how to improve it

crates/egui/src/memory.rs
```diff
@@ -322,31 +355,121 @@ impl Focus {
...
+            if let Some(new_rect) = new_rects.get(id) {
+                *old_rect = *new_rect;
+                true // Keep the item
+            } else {
+                false // Remove the item
+            }
+        });
+
+        let current_focus_widget_rect = self.focus_widgets_cache.get(&focus_id).unwrap();
+        let current_widget_pos = current_focus_widget_rect.left_center();
```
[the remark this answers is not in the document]
TimonPost: I am not entirely sure what the best metrics is. I think going from center, left, right, top down all have their consequences. 

Left The center might change based on the contents, the left center will not change based on content. I had the issue were i wanted to go to the widget to the right but the widget to the right top was closer center wise so I went up right instead.

crates/egui/src/memory.rs
```diff
@@ -322,31 +355,121 @@ impl Focus {
...
+            // Check if the widget has the right dot product and length.
+            let focus_direction = match self.focus_direction {
+                FocusDirection::Up => Vec2::UP,
+                FocusDirection::Right => Vec2::RIGHT,
+                FocusDirection::Down => Vec2::DOWN,
+                FocusDirection::Left => Vec2::LEFT,
+                _ => {
+                    continue;
+                }
+            };
```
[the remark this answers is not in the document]
TimonPost: If `current_to_candidate` is perpendicular to `focus_direction` it is 0. If `current_to_candidate` is parallel to `focus_direction` it is 0. `aligns_with_focus_direction` exists to prevent picking the perpendicular case. I do not see how this change solves this problem. Can you elaborate?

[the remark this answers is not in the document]
TimonPost: Yes your right!

TimonPost:
I updated the selection metric but it is still not perfect.
```
focus_candidates.sort_by(|(_, dot1, len1), (_, dot2, len2)| {
    (len1 / dot1).partial_cmp(&(len2 / dot2)).unwrap()
});
focus_candidates.sort_by(|(_, dot1, _), (_, dot2, _)| dot2.partial_cmp(dot1).unwrap());
```

Not sorting by dot product afterwards results in close by widgets to be picked evenwhile they have a big angle. I am not sure how perfect this feature has to be. 

An other issue is that sometimes it's hard to tell where focus went to as it might be a window behind the current window, or some widget in a different panel. 

The rectangle thing does seem to improve the overall behavior.

crates/emath/src/range.rs
```diff
@@ -49,12 +49,48 @@ impl Rangef {
...
+        self.min + (self.span() / 2.0)
+    }
+
+    /// Returns the similarity between `self` and `other`.
+    ///
+    ///  * negative if `a` is left of `b`
+    /// * positive if `a` is right of `b`
+    /// * zero if the ranges overlap significantly
+    #[inline]
+    pub fn similarity(self, other: Rangef) -> f32 {
```
emilk, line 64: This is a heuristic specific to the keyboard problem and doesn't belong in this file

crates/emath/src/range.rs
```diff
@@ -49,12 +49,48 @@ impl Rangef {
...
+
     #[inline]
     #[must_use]
     pub fn contains(self, x: f32) -> bool {
         self.min <= x && x <= self.max
     }
 
+    /// Returns if `self` contains `other` for at least 50%.
+    #[inline]
+    pub fn contains_range(&self, other: Rangef) -> bool {
```
emilk, line 80: This is a very surprising definition of `contains_range`. Either change it to `self.contains(other.min) && self.contains(other.max)`, or move it to where it is being used.

TimonPost: Was trying to get the overlap of more than 50 percent to work. So that self looks if it overlaps at least 50% with other. Changing it to self.contains(other.min) && self.contains(other.max) would restrict it to a full overlap where other is fully within self?

emilk: the 50% overlap may be a good idea for what you are trying to do, but it is not a good idea for a `pub` function called `contains_range`. Call it `has_significant_overlap` or something and move it to where it is used (`crates/egui/src/memory.rs`)

emilk: the 50% overlap may be a good idea for what you are trying to do, but it is not a good idea for a `pub` function called `contains_range`. Call it `has_significant_overlap` or something and move it to where it is used (`crates/egui/src/memory.rs`)

emilk (approved):

emilk:
There is still plenty of room for improvement, but this is a good start!

Jakkestt:
>It allows integrations to convert joysticks to arrow keys in order to toggle between menu items.
How do you do that?

TimonPost:
Gotta make some custom logic to achieve that. E.g detect the cardinal directions of the joystick and map them to key combo's

PR #3275, opened 2023-08-23
Inadequate fixes for 3109 (request_repaint_after)

### LINKED ISSUE #3109: Duration passed to `request_repaint_after` is ignored if called each frame
**Describe the bug**
Using egui 0.22.0, calling `request_repaint_after()` causes the application to redraw the UI with a high refresh rate, regardless of the duration passed into `request_repaint_after()`. It behaves (mostly) as if calling `request_repaint()`.
The bug was not there in version 0.21.0.

**Steps to reproduce the behavior:**
The following code was adapted from the hello world example. I replaced the controls with a fps (frames per second) and counter label. Rust file:
```rs
use eframe::egui;
use std::time::{Duration, Instant};

        initial_window_size: Some(egui::vec2(320.0, 240.0)),
        ..Default::default()
    };
    eframe::run_native(
        "My egui App",
        options,
        Box::new(|_cc| Box::<MyApp>::default()),
    )
}

struct MyApp {
    previous_instant: Instant,
    previous_count: u64,
}

impl Default for MyApp {
    fn default() -> Self {
        Self {
            previous_instant: Instant::now(),
            previous_count: 0,
        }
    }
}

impl eframe::App for MyApp {
            let fps = 1.0 / self.previous_instant.elapsed().as_secs_f64();
            self.previous_instant = Instant::now();
            self.previous_count += 1;
            ui.label(format!("fps: {:.2}, counter: {}", fps, self.previous_count));
        });
        ctx.request_repaint_after(Duration::from_secs(1));
    }
}
```
toml:
```toml
[package]
name = "egui_request_repaint_bug"
version = "0.1.0"
edition = "2021"

[dependencies]
egui = "0.22.0"
eframe = { version = "0.22.0", default-features = false, features = [
    "default_fonts", # Embed the default egui fonts.
    "glow",          # Use the glow rendering backend. Alternative: "wgpu".
] }

[patch.crates-io]
egui = { git = "https://github.com/emilk/egui", branch = "master" }
eframe = { git = "https://github.com/emilk/egui", branch = "master" }
```

Build and run on Windows 10.

When the application has focus, the "fps" labels shows approximately a value of 144 on my monitor (which has a refresh rate of 144 Hz), regardless of where the mouse is. When the application does not have focus, the "fps" label continues updating with a high frequency and shows approximately 90. (I have no idea where the 90 is coming from.)

If I remove the call to `request_repaint_after()`, the application no longer updates with a high rate. 
Keeping the `request_repaint_after()` call and going back to egui 0.21.0, I get the expected behavior. Considering the [release notes](https://github.com/emilk/egui/blob/master/crates/eframe/CHANGELOG.md#0220---2023-05-23), I guess this is related to #2939.

**Expected behavior**
When the application does NOT have focus, the "fps" label should show a value of approximately 1. In other words, the GUI should update only once a second because I passed 1s as duration to `request_repaint_after()`.
When the application does have focus, the "fps" label should show a value of 144 only while moving around the mouse.

**Desktop:**
 - OS:  Windows 10
 - Browser: Not applicable; running on native
 - Version of egui: 0.22.0 (version 0.21.0 is not affected)

OmegaJak:
Closes #3109.

### DIFF
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -143,26 +143,22 @@ fn run_and_return(
                 return;
             }
 
-            // Platform-dependent event handlers to workaround a winit bug
-            // See: https://github.com/rust-windowing/winit/issues/987
-            // See: https://github.com/rust-windowing/winit/issues/1619
-            #[cfg(target_os = "windows")]
-            winit::event::Event::RedrawEventsCleared => {
-                next_repaint_time = extremely_far_future();
-                winit_app.run_ui_and_paint()
-            }
-            #[cfg(not(target_os = "windows"))]
             winit::event::Event::RedrawRequested(_) => {
+                log::trace!("Repaint caused by RedrawRequested");
                 next_repaint_time = extremely_far_future();
                 winit_app.run_ui_and_paint()
             }
 
             winit::event::Event::UserEvent(UserEvent::RequestRepaint { when, frame_nr }) => {
-                if winit_app.frame_nr() == *frame_nr {
+                if winit_app.frame_nr().abs_diff(*frame_nr) <= 1 {
                     log::trace!("UserEvent::RequestRepaint scheduling repaint at {when:?}");
                     EventResult::RepaintAt(*when)
                 } else {
-                    log::trace!("Got outdated UserEvent::RequestRepaint");
+                    log::trace!(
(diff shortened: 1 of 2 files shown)


emilk:
Closing: #3109 was fixed in #3978. Thanks for the PR!

PR #3288, opened 2023-08-29
Changing `menu_image_button` to use `ImageButton` builder

v-kat:
I'm changing `menu_image_button` to allow passing in an ImageButton so that I can modify the frame of the passed in ImageButton. It's a breaking API change but I kept the shorthand `ui` version the same.


emilk (approved):
Makes sense!

PR #3294, opened 2023-09-01
Added focus & highlight to the topmost window.

GuillaumeSchmid:
Added a function to memory to retrieve the topmost panel LayerId on a given Order.
Added a color for focused window header.
The tomost window header appears with a specific color (this effect shoudl be enhanced by completely filling the header).
This effect is visible in the demo.
image

### DIFF
--- a/crates/egui/src/containers/frame.rs
+++ b/crates/egui/src/containers/frame.rs
@@ -29,6 +29,8 @@ pub struct Frame {
 
     pub fill: Color32,
 
+    pub selected_header_color: Color32,
+
     pub stroke: Stroke,
 }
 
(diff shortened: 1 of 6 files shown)


GuillaumeSchmid:
This solution is not perfect, but it is simple.
There is a small problem indeed: if you close a focused window, no other window gets the focus as the memory keeps even the closed window in the ordered list.
To solve that, I could implement a simple window manager struct in the context that would keep track of the focused window.
Each window would register itself and update its status.
Would such an approach be acceptable?
The idea behind this feature is the possibility to have multiple images or documents open and a side panel that updates itself according to the focused one.

GuillaumeSchmid:
Ok, I did not know if the closed windows on the order list were left on purpose.
I will try your suggestion.

GuillaumeSchmid:
I implemented closed window culling. It seems to work!

GuillaumeSchmid:
Removed the has_focus on window as it is not necessary.
the test can be performed from he UI function by comparing the ui.layer_id with the topmost layer_id provided by the context.
Updated the demo.

GuillaumeSchmid:
I fixed the aspect of the selected window. Now, the whole header area of the widget is painted.
I also fixed the problem with the collapsible icon not showing.

Issue #3291, opened 2023-09-01
Flexible and convenient texture loading


emilk:
## Intro
Showing images in egui is currently very cumbersome. There is `egui_extras::RetainedImage`, but it requires you to store the `RetainedImage` image somewhere and is obviously not very immediate mode.

Ideally we want users to be able to write something like:

```rs
ui.image("file://image.png");
ui.image("https://www.example.com/imag.png");
```

We also want 3rd party crates like [`egui_commonmark`](https://github.com/lampsitter/egui_commonmark) to be able to use the same system, without having to implement their own image loading.

## Desired features
egui is designed to have minimal dependencies and never doing any system calls, and I'd like to keep it that way. Therefor the solution needs to be some sort of plugin system with callbacks. This is also the best choice for flexibility.

There is three steps to loading an image:
* Getting the image bytes
    * file://
    * https://
    * From a static list of `include_bytes!("my_icon.png")`
    * …
* Decoding the image
    * png, jpg, …
    * svg rasterization
    * …
* Uploading it to a texture

In most cases we want the loaded image to be passed to `egui::Context::load_texture`, which wil hand the image off to whatever rendering backend egui is hooked up to. This will allow the image loaded to work with any egui integration.

We should also allow users to have more exotic ways of loading images. At Rerun we have a data store where we store images, and it would be great if we could just reference images in that store with e.g. `ui.image(rerun://data/store/path);`.

In some cases however, the user may want to refer to textures that they themselves have uploaded to the GPU, i.e. return a `TextureId::User`. We do this at Rerun.

## Proposal
I propose we introduce three new traits in `egui`:
* `BytesLoader`
* `ImageLoader`
* `TextureLoaader`

The user can then mix-and-match these as they will.

We will supply good default implementations for these traits in `egui_extras`. Most users will never need to look at the details of these.

All these traits are designed to be _immediate_, i.e. callable each frame. It is therefor up to the implementation to cache any results.

They are designed to be able to suppor background loading (e.g. slow downloading of images).
For this they return `Pending` when something is being loaded. When the loading is done, they are responsible for calling `ctx.request_repaint` so that the now-loaded image will be shown.

They can also return an error.

`Pending` will be shown in egui using `ui.spinner`, and errors with red text.

### Common code
``` rs
enum LoadError {
    /// This loader does not support this protocol or image format.
    ///
    /// Try the next loader instead!
    NotSupported,

    /// A custom error string (e.g. "File not found: foo.png")
    Custom(String),
}

[...]

PR #3297, opened 2023-09-04
Managed texture loading

jprochazk:
Part of https://github.com/emilk/egui/issues/3291.

- [x] Add types from the proposal
- [x] Add loader implementations from proposal to `egui_extras`
- [x] Update usage in examples
- [x] Documentation


crates/epaint/src/image.rs
```diff
@@ -11,7 +12,7 @@ use crate::{textures::TextureOptions, Color32};
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub enum ImageData {
     /// RGBA image.
-    Color(ColorImage),
+    Color(Arc<ColorImage>),
```
emilk, line 15: Or add the `Arc` to inside `ColorImage` 🤷

crates/epaint/src/image.rs
```diff
@@ -226,6 +227,12 @@ impl std::ops::IndexMut<(usize, usize)> for ColorImage {
 impl From<ColorImage> for ImageData {
     #[inline(always)]
     fn from(image: ColorImage) -> Self {
+        Self::Color(Arc::new(image))
+    }
+}
+
+impl From<Arc<ColorImage>> for ImageData {
+    fn from(image: Arc<ColorImage>) -> Self {
```
emilk, line 235:
```suggestion
    #[inline]
    fn from(image: Arc<ColorImage>) -> Self {
```

emilk (changes requested):
Really awesome!

jprochazk:
> Image loading can be slow.

@emilk is it slow eonugh that we should also make `ImageCrateLoader` load images on a separate thread?

emilk:
> > Image loading can be slow.
> 
> @emilk is it slow eonugh that we should also make `ImageCrateLoader` load images on a separate thread?

The preference here actually depend on the application.

If you only have a few images, you may prefer a slow first frame than the flicker that comes from loading things in the background (also goes for loading the bytes from disk). But if you have hundreds of images, then you definitely want to load them in background threads. But spawning one thread for each image (or two - one for the loading from disk, one for the parsing of the image file) is not ideal, so we should also consider adding `rayon` as an opt-in dependency for that.

I'm not sure how a user would want to configure this. Probably with a flag in `egui::Options` that's then passed in to the loaders. But let's punt on that for now - let's start with always using background threads, either with `std::thread` or `rayon`, if that feature is enabled. That will yield the most responsive applications, and is pretty standard these days (that's what people is used to on the web, and in electron, etc)

crates/egui_extras/src/loaders.rs
```diff
@@ -0,0 +1,51 @@
...
+/// Install the default set of loaders:
+/// - `file` loader on non-Wasm targets
+/// - `http` loader with the `ehttp` feature
+/// - `image` loader with the `image` feature
+///   - the supported set of image formats is configured by enabling `image` crate features.
+/// - `svg` loader with the `svg` feature
+///
+/// The `file` and `http` loaders are bytes loaders, they do not know how to turn those
+/// bytes into an image. `svg` is for loading `.svg` files, and `image` can load any
+/// other image format enabled on the `image` crate.
```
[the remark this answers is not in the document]
jprochazk: Documentation is tricky! Hopefully I made it clearer here

crates/egui_extras/src/loaders/ehttp_loader.rs
```diff
@@ -0,0 +1,102 @@
+use egui::{
+    ahash::HashMap,
+    load::{BytesLoadResult, BytesLoader, BytesPoll, LoadError},
+    mutex::Mutex,
+};
+use std::{sync::Arc, task::Poll};
+
+#[derive(Default)]
+pub struct EhttpLoader {
+    cache: Arc<Mutex<HashMap<String, Poll<Result<Arc<[u8]>, String>>>>>,
```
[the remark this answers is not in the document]
jprochazk: I'm not sure if those two are the same. `BytesLoadResult` contains `BytesPoll` in the `Ok` variant, so this change would make it something like `Poll<Result<Poll>>`, which I believe is not what we want here.

jprochazk:
> let's start with always using background threads, either with std::thread or rayon, if that feature is enabled

Didn't think this through properly, but spawning a background thread for anything other than `FileLoader` would not work for Wasm. I say let's punt on it completely for now and figure out later how we'll make it non-blocking on Wasm, or if we do that at all.

jprochazk:
One thing that isn't done here in any way is constructing an `Image2` from non-static bytes, e.g. `egui::ColorImage::example()`. Not exactly sure how that should work. Should the `load_include_bytes` function and `IncludeBytesLoader` just allow non-`'static` lifetimes?

emilk:
> One thing that isn't done here in any way is constructing an `Image2` from non-static bytes, e.g. `egui::ColorImage::example()`. Not exactly sure how that should work. Should loading from bytes just allow non-`'static` lifetimes?

I think so - the `IncludeBytesLoader` still copies the bytes to an `Arc<[u8]>` anyways, so why not.

A potential optimization is to replace `Arc<[u8]>` with `enum Bytes { Static(&'static [u8]), Shared(Arc<[u8]>) }` to avoid both the copies and the extra memory use.

emilk (approved):
Exciting!

crates/egui/src/load.rs
```diff
@@ -274,17 +314,27 @@ pub trait TextureLoader {
...
-        self.cache.lock().entry(uri).or_insert_with(|| bytes.into());
+impl DefaultBytesLoader {
+    pub(crate) fn insert_static(&self, uri: &'static str, bytes: &'static [u8]) {
+        self.cache
+            .lock()
+            .entry(uri)
+            .or_insert_with(|| Bytes::Static(bytes));
+    }
+
+    pub(crate) fn insert_shared(&self, uri: &'static str, bytes: impl Into<Arc<[u8]>>) {
```
emilk, line 329:
```suggestion
    pub(crate) fn insert_static(&self, uri: &'static str, bytes: impl Into<Bytes>) {
        self.cache
            .lock()
            .entry(uri)
            .or_insert_with(|| bytes.into());
    }
```

jprochazk:
I'm not totally sure about the API I came up with for `Bytes`, but we can always improve it later.

crates/egui/src/load.rs
jprochazk: There's a tiny optimization here: `Bytes` requires `Arc<[u8]>` to already be constructed, `impl Into<Arc<[u8]>>` doesn't. So here the user would have to copy their bytes into a `Arc<[u8]>` first if the data is not `'static`, where as with this split, we only copy them if there isn't already an entry for the given `uri`.

Is it worth keeping?

crates/epaint/src/texture_handle.rs
```diff
@@ -95,6 +95,11 @@ impl TextureHandle {
         crate::Vec2::new(w as f32, h as f32)
     }
 
+    /// `width x height x bytes_per_pixel`
+    pub fn byte_size(&self) -> usize {
+        self.tex_mngr.read().meta(self.id).unwrap().bytes_used()
```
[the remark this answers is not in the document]
jprochazk: The `size` method also unwraps, so I just copied that. I'll change both to not unwrap.

PR #3300, opened 2023-09-04
Add `scroll_area::State::velocity`

Barugon:
Add a method to `State` in order to check if a scroll area has momentum.

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -55,8 +55,8 @@ impl State {
         ctx.data_mut(|d| d.insert_persisted(id, self));
     }
 
-    /// Returns true if the scroll area has momentum.
-    pub fn has_momentum(&self) -> bool {
+    /// Returns true if the scroll area currently has a velocity due to momentum.
+    pub fn has_velocity(&self) -> bool {
         self.vel.x != 0.0 || self.vel.y != 0.0
     }
 }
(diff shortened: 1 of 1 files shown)


crates/egui/src/containers/scroll_area.rs
```diff
@@ -54,6 +54,11 @@ impl State {
     pub fn store(self, ctx: &Context, id: Id) {
         ctx.data_mut(|d| d.insert_persisted(id, self));
     }
+
+    /// Returns true if the scroll area has momentum.
+    pub fn has_momentum(&self) -> bool {
```
emilk, line 59: This can be misinterpreted as to mean "will have momentum if given the chance"

```suggestion
    /// Returns true if the scroll area currently has a velocity due to momentum.
    pub fn has_velocity(&self) -> bool {
```

emilk (approved):

crates/egui/src/containers/scroll_area.rs
```diff
@@ -54,6 +54,11 @@ impl State {
     pub fn store(self, ctx: &Context, id: Id) {
         ctx.data_mut(|d| d.insert_persisted(id, self));
     }
+
+    /// Returns true if the scroll area currently has a velocity due to momentum.
+    pub fn has_velocity(&self) -> bool {
+        self.vel.x != 0.0 || self.vel.y != 0.0
```
emilk, line 60: We could also just return the current velocity, which gives the user a lot more info

emilk (approved):

PR #3301, opened 2023-09-04
`eframe::Frame::info` returns a reference

Barugon:
Add a way to get `IntegrationInfo` as a reference in order to avoid a deep copy.

### DIFF
--- a/crates/eframe/src/epi/mod.rs
+++ b/crates/eframe/src/epi/mod.rs
@@ -774,6 +774,10 @@ impl Frame {
         self.info.clone()
     }
 
+    pub fn info_ref(&self) -> &IntegrationInfo {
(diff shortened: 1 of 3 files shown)


crates/eframe/src/epi/mod.rs
```diff
@@ -774,6 +774,11 @@ impl Frame {
         self.info.clone()
     }
 
+    /// Get `IntegrationInfo` as a reference.
+    pub fn info_ref(&self) -> &IntegrationInfo {
+        &self.info
+    }
```
emilk, line 780: Let's just change the old function to return a reference instead. Users can just call `.clone()` themselves.

emilk (approved):

Issue #2738, opened 2023-02-13
Add support for custom `line_height` and `letter_spacing`


mikedilger:
I'm considering working on a PR and want feedback first.

I'd like to be able to follow the "visual design rules" on this site https://anthonyhobday.com/sideprojects/saferules/
specifically this PR would allow users to "lower letter spacing and line height with larger text. Raise them with smaller text" as shown in this image https://anthonyhobday.com/sideprojects/saferules/images/6.png

For this I'm suggesting `row_height` and `character_spacing` fields within epaint::TextFormat.

`row_height` would be a multiplier on `glyph.size.y` within `text_layout.rs` and `character_spacing` a multiplier on `advance_width`.

Users would have to manually set these; I'm not suggesting we automatically do anything differently from now.

abey79:
Piggy-backing on this existing issue, I suggest following as closely CSS naming/behaviour as possible.

- `line_height`: https://developer.mozilla.org/en-US/docs/Web/CSS/line-height
- `letter_spacing`: https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing

cc @martenbjork

PR #3302, opened 2023-09-04
Add control of line height and letter spacing

emilk:
* Closes https://github.com/emilk/egui/issues/2738

Line height can now be explicitly overridden, and the spacing between letter adjusted.

They can be set in the low-level `epaint::TextFormat` or the high-level `egui::RichText`.

Both are set in points, but for even kerning you should round to pixels.

You can see both in action in the new "Text Layout" demo (using integer pixels rather than points for the ui):

line-height
letter-spacing


crates/egui/src/widget_text.rs
```diff
@@ -100,6 +101,28 @@ impl RichText {
         self
     }
 
+    /// Extra spacing between letters, in points.
+    ///
+    /// Default: 0.0. Round to whole _pixels_ for crisp text.
```
abey79, line 106: Do you have a section on pixel vs. point in the docs you could link to from here?

abey79 (approved):
Nice!

crates/egui/src/widget_text.rs
emilk: I'll add something!

Issue #3311, opened 2023-09-05
Should `egui::Button`s change the cursor to PointingHand on hover?


zkldi:
At the moment, mousing over an egui::Button does nothing, but I think it should change to `PointingHand` to be consistent with most other UIs.

For example, mousing over any button on this page, or in vscode, etc. will turn the cursor into `PointingHand`.

This could be achieved with `ctx.output_mut(|out| { out.cursor_icon = CursorIcon::PointingHand })` inside the `.hovered()` call for a button.

It would be painful to implement this everywhere yourself, so maybe it should just be part of the component itself?

LoganDark:
this is a web browser thing that absolutely does not belong on native controls! leave this up to the user please and do NOT use the pointing hand cursor by default

PR #3312, opened 2023-09-05
feat: change to PointingHand when hovering over a button

zkldi:
When hovering over an `egui::Button`, the cursor should change to a PointingHand. This is consistent with the buttons in most other UIs (the browser, etc.)

Closes <https://github.com/emilk/egui/issues/3311>.

### DIFF
--- a/crates/egui/src/widgets/button.rs
+++ b/crates/egui/src/widgets/button.rs
@@ -238,7 +238,7 @@ impl Widget for Button {
             }
         }
 
-        if response.hovered && ui.visuals().pointinghand_if_interactable {
+        if response.hovered && ui.visuals().pointinghand_if_interactive {
             ui.ctx().set_cursor_icon(CursorIcon::PointingHand);
         }
 
--- a/crates/egui/src/widgets/button.rs
+++ b/crates/egui/src/widgets/button.rs
@@ -238,6 +238,10 @@ impl Widget for Button {
             }
         }
 
+        if response.hovered {
+            ui.ctx().set_cursor_icon(CursorIcon::PointingHand);
+        }
+
         response
     }
 }
--- a/crates/egui/src/widgets/button.rs
+++ b/crates/egui/src/widgets/button.rs
@@ -238,8 +238,11 @@ impl Widget for Button {
             }
         }
 
-        if response.hovered && ui.visuals().pointinghand_if_interactive {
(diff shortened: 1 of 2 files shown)


emilk:
This should be opt-in, set as an option in `Visuals`.

I personally prefer pointing-hand to mean "clicking here will bring you someplace else", which not all buttons do, but of course this is a matter of preference.

You should also consider if this should apply to other widgets too. Maybe you want anything that is clickable to have the pointing hand?

zkldi:
Makes sense to me. I'll look into making this a visual option and take a look at other clickable widgets in egui. Should be committed by the end of the day.

Thanks for such a fast review!

LoganDark:
https://github.com/emilk/egui/issues/3311#issuecomment-1709445746

> this is a web browser thing that absolutely does not belong on native controls! leave this up to the user please and do NOT use the pointing hand cursor by default. maybe a setting would be reasonable but not a default

zkldi:
Sounds good to me! I'm putting this into an option now.

zkldi:
This is now an option that is off by default. From what I can tell browsers don't invoke this behaviour when hovering over a checkbox or a radio button. It's just for normal buttons.

There's an argument to be made that things like `collapsing_header` should also inherit this behaviour, since they're button-ish (vscode would agree with me here, but things like `<summary>` disagree.)

*****

`pointinghand_if_interactive` is quite frankly a terrible name but I couldn't think of a better one. Totally open for any suggestions that suck less.

LoganDark:
> It's just for normal buttons.

Nope - it's only for hyperlinks. Sites override this for buttons, but for desktop UI elements this isn't common.

zkldi:
> Nope - it's only for hyperlinks. Sites override this for buttons, but for desktop UI elements this isn't common.

Oh wow, yea. Everything I checked this on overrides that functionality. I guess I've just subconsciously gotten used to it and thought it was the default.

> `interact_cursor` (or something like) and have it be an Option that can contain any cursor one would like?

Yeah this is a huge improvement. One sec.

crates/egui/src/widgets/button.rs
```diff
@@ -238,6 +238,13 @@ impl Widget for Button {
             }
         }
 
+        // n.b. no let chains here...
```
emilk, line 241:
```suggestion
```

emilk (approved):

Issue #3057, opened 2023-06-06
Phosh on-screen keyboard pops up when app starts


Barugon:
**Describe the bug**
When running an eframe app on Phosh, the on-screen keyboard pops up when the app starts. This is because `Window::set_ime_allowed` is being called with a parameter of `true` (#2046) .

**To Reproduce**
Build and run any eframe app on Phosh (Librem 5, PinePhone).

Barugon:
Perhaps calling `Window::set_ime_allowed` with `true` when a `TextEdit` gains focus and then calling it with `false` when focus is lost would actually bring up the on-screen keyboard at the right time?

liquidhelium:
Related to #2614.

PR #3313, opened 2023-09-06
Implement `Frame::allow_ime` method

Barugon:
Blindly enabling IME is problematic for Phosh (and probably other Linux desktops that use touch) as it brings up the on-screen keyboard when the app starts. A better solution would be to enable IME when a text edit gains focus and disable IME when it looses focus (which actually shows and hides the on-screen keyboard nicely in Phosh). In order to facilitate this, I have added an `allow_ime` method to `Frame`.

I also added an `allow_ime` member to `NativeOptions` in order to enable IME at startup.

### DIFF
--- a/crates/eframe/src/epi/mod.rs
+++ b/crates/eframe/src/epi/mod.rs
@@ -455,6 +455,9 @@ pub struct NativeOptions {
     /// }
     /// ```
     pub app_id: Option<String>,
+
+    /// Enable/disable IME at startup.
+    pub allow_ime: bool,
 }
 
 #[cfg(not(target_arch = "wasm32"))]
@@ -529,6 +532,7 @@ impl Default for NativeOptions {
             wgpu_options: egui_wgpu::WgpuConfiguration::default(),
 
             app_id: None,
+            allow_ime: false,
         }
     }
 }
--- a/crates/eframe/src/epi/mod.rs
+++ b/crates/eframe/src/epi/mod.rs
@@ -730,6 +730,9 @@ pub struct Frame {
     /// Raw platform display handle for window
     #[cfg(not(target_arch = "wasm32"))]
     pub(crate) raw_display_handle: RawDisplayHandle,
+
+    /// Option for enabling/disabling IME.
+    pub(crate) allow_ime: Option<bool>,
 }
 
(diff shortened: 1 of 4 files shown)


crates/eframe/src/epi/mod.rs
```diff
@@ -455,6 +455,9 @@ pub struct NativeOptions {
     /// }
     /// ```
     pub app_id: Option<String>,
+
+    /// Enable/disable IME at startup.
+    pub allow_ime: bool,
```
emilk, line 460: Please describe this a bit more.

If `true`, will an on-screen-keyboard usually be shown for touch-screens on start? Will it hide once the user stops editing text?

If `false`, will it still show up when you start editing text? Will it hide once you stop editing the text?

Barugon: > If `true`, will an on-screen-keyboard usually be shown for touch-screens on start? Will it hide once the user stops editing text?

You have to manually hide the keyboard on Phosh.

> If `false`, will it still show up when you start editing text? Will it hide once you stop editing the text?

It will not automatically show the keyboard when you enter a text edit. You have to manually show it.

Adding this to my code will automatically show/hide the keyboard on Phosh:
```Rust
          let response = ui.text_edit_singleline(&mut self.text);
          if response.gained_focus() {
            frame.allow_ime(true);
          } else if response.lost_focus() {
            frame.allow_ime(false);
          }
```

I added `allow_ime` to `NativeOptions` in case someone wants to keep the previous behavior.

Barugon: I may be that enabling and disabling IME shows and hides the keyboard for other OS' but I really don't have a way to test that.

emilk: It seems to me the desirable behavior is to have one flag that controls wether or not the on-screen-keyboard shows on startup (with `false` being a reasonable default).

And then, regardless of that flag, always show the on-screen-keyboard when the user clicks a text field

Barugon: That's my ultimate goal but egui is rather disconnected from the windowing system, so this was the best that I could come up with.

emilk:
* Btw, this attempts to close https://github.com/emilk/egui/issues/3057, correct?

Barugon:
Closing in favor of #3362.

PR #3319, opened 2023-09-07
Add method to RichText making it easier to construct layout jobs

### LINKED ISSUE #3318: Difficult to create an Into<WidgetText> that looks like a composite of RichTexts
**The Problem**
I'm looking to using text with mixed formatting as the text of a `CollapsingHeader`, whose parameter is an `Into<WidgetText>`. In my case, I am trying to get ordinary body text, with one word in the middle bolded (like this: "Using: **Chrome** (started using 2 minutes ago)"). As far as I can tell, this currently requires the use of a LayoutJob, whose API, while allowing a great deal of control, is not very ergonomic and takes some work to make it visually match a `RichText`.

**Example**
The best way I know how to achieve this currently looks something like this:
```rust
let layout_job = LayoutJob::default();
let font_id = ui
    .style()
    .text_styles
    .get(&egui::TextStyle::Body)
    .unwrap()
    .clone();
layout_job.append(
    "Using: ",
    0.0,
    TextFormat {
        font_id: font_id.clone(),
        color: ui.style().visuals.text_color(),
        ..Default::default()
    },
);
layout_job.append(
    &format!("{} ", self.current_window.app_name),
    0.0,
    TextFormat {
        font_id: font_id.clone(),
        color: ui.style().visuals.strong_text_color(),
        ..Default::default()
    },
);
layout_job.append(
    &format!("(started using {})", time_using_current),
    0.0,
    TextFormat {
        font_id: font_id.clone(),
        color: ui.style().visuals.text_color(),
        ..Default::default()
    },
);
```

I could reduce a little duplication by pulling the 'normal text' `TextFormat` into a variable, but this is still a lot of code just to get one part of the text bolded. Manipulating `TextFormat` directly is much more difficult than using a `RichText`.

**My Dream**
I would love if there were an easy way to make something implementing `Into<WidgetText>` from multiple `RichText`s. The `RichText` api is very ergonomic, and this would ensure visual consistency across the application.

I will be raising a PR shortly that takes a small step towards this, but I'm not sure if it's the best solution.

OmegaJak:
Closes #3318

This refactors the existing code that was used inside `RichText.into_text_job()` to generate a `LayoutJob` to also expose a method that appends to a `LayoutJob` passed in from the outside. This makes it possible to construct a `LayoutJob` using only `RichText`s, without having to manually touch `TextFormat`.

In my code I added an extension method to UI (which didn't seem fitting for core egui) that looks like this:
```rust
fn create_default_layout_job(&self, rich_texts: Vec<RichText>) -> LayoutJob {
    let mut layout_job = LayoutJob::default();
    for rich_text in rich_texts.into_iter() {
        rich_text.append_to(
            &mut layout_job,
            self.style(),
            egui::FontSelection::Default,
            egui::Align::Center,
        )
    }

    layout_job
}
```

which can be used like so:
```rust
let layout_job = ui.create_default_layout_job(vec![
    RichText::new("Using: ").color(ui.style().visuals.text_color()),
    RichText::new(format!("{} ", self.current_window.app_name)).strong(),
    RichText::new(format!("(started using {})", time_using_current))
        .color(ui.style().visuals.text_color()),
]);
```

This, in my opinion, is a much more ergonomic API that was available previously. Compare the above code to the code inside the linked #3318.

Note that the color of text must be specified in the RichText since `WidgetText.into_text_job()` always sets `WidgetTextJob::job_has_color` to true when converting from a LayoutJob, or the text will show up as green.

I'm not sure if this change is appropriate for core `egui`, but since it helped me, I thought I'd share and see.

### DIFF
--- a/crates/egui/src/widget_text.rs
+++ b/crates/egui/src/widget_text.rs
@@ -269,6 +269,35 @@ impl RichText {
         fonts.row_height(&font_id)
     }
 
+    /// Append to an existing [`LayoutJob`]
+    ///
+    /// Note that the color of the [`RichText`] must be set, or may default to an undesirable color.
+    ///
+    /// ### Example
+    /// ```
+    /// use egui::{Style, RichText, text::LayoutJob, Color32, FontSelection, Align};
+    ///
+    /// let style = Style::default();
+    /// let mut layout_job = LayoutJob::default();
+    /// RichText::new("Normal")
+    ///     .color(style.visuals.text_color())
+    ///     .append_to(
+    ///         &mut layout_job,
+    ///         &style,
+    ///         FontSelection::Default,
+    ///         Align::Center,
+    ///     );
+    /// RichText::new("Large and underlined")
+    ///     .color(style.visuals.text_color())
+    ///     .size(20.0)
+    ///     .underline()
+    ///     .append_to(
+    ///         &mut layout_job,
+    ///         &style,
+    ///         FontSelection::Default,
+    ///         Align::Center,
+    ///     );
+    /// ```
     pub fn append_to(
         self,
         layout_job: &mut LayoutJob,
--- a/crates/egui/src/widget_text.rs
+++ b/crates/egui/src/widget_text.rs
@@ -269,12 +269,34 @@ impl RichText {
         fonts.row_height(&font_id)
     }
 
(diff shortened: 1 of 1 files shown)


emilk (approved):
Nice!

PR #3323, opened 2023-09-08
Support remote rendering

linguini11:
Support the ability to render egui remotely using eframe, inspired by these imgui projects:
- https://github.com/ocornut/imgui/issues/75
- https://github.com/sammyfreg/netImgui
- https://github.com/ggerganov/imgui-ws

Addresses the feature request discussed here: https://github.com/emilk/egui/discussions/3292

An example is included in examples/remote_rendering of how to set this up. This example uses the tungstenite crate to create websockets for the client and server, but other client/server set ups should work too.
As a basic overview, the server sends the serialised FullOutput to the client to render it, and the client sends RawInput back to the server.

serde-diff is being included as a feature flag (though this won't currently run, I need to fix some errors). Performance is greatly improved when using it, but it requires a couple of pull requests to be merged in order for it to work (https://github.com/amethyst/serde-diff/pull/36 and https://github.com/amethyst/serde-diff/pull/45), and currently the crate is unmaintained. While a way forward is being looked into for that, the remote_rendering works without but is a bit slow.

This is quite a rough draft to get some feedback on the approach I've taken. The most recent working commit is b1ebaf14a428b2ef8559ddf0fa060ead3c476dc8.


examples/remote_rendering/src/server.rs
```diff
@@ -0,0 +1,59 @@
+use serde_diff::Diff;
```
emilk, line 1: Better to put `server.rd` and `client.rs` into a `bin` folder, then users can run `cargo r -p remote_rendering --bin client`

examples/remote_rendering/src/client.rs
```diff
@@ -0,0 +1,118 @@
...
+            .expect("Error reading message from server");
+        // Filter out ping messages
+        if let tungstenite::Message::Ping(_) = received_message {
+            return (self.full_output.clone(), self.pixels_per_point);
+        }
+
+        let full_output_diff = received_message.into_text();
+
+        if let Ok(diff) = full_output_diff {
+            let deserialized: egui::FullOutput = serde_json::from_str(&diff).unwrap();
```
emilk, line 112: Using json for this will be very slow and wasteful - I suggest using `bincode` or `MsgPack` instead

examples/remote_rendering/src/server.rs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        let ctx = context.clone();
+        ctx.set_pixels_per_point(current_pixels_per_point);
+        ctx.begin_frame(latest_raw_input.clone());
+        egui::Window::new("Hello world!")
+            .default_pos(egui::pos2(100.0, 0.0))
+            .show(&ctx, |ui| {
+                ui.label(format!("Count: {:?}", count));
+            });
```
emilk, line 46: You can use `egui_demo_lib` instead - that would give a more convincing example, and a better benchmark for performance!

crates/epaint/Cargo.toml
```diff
@@ -66,6 +66,14 @@ mint = ["emath/mint"]
 ## Allow serialization using [`serde`](https://docs.rs/serde).
 serde = ["dep:serde", "ahash/serde", "emath/serde", "ecolor/serde"]
 
+## Allow diffing serialization using [`serde`](https://docs.rs/serde).
```
emilk, line 69:
```suggestion
## Allow diffing serialization using [`serde-diff`](https://docs.rs/serde-diff).
```

crates/emath/src/pos2.rs
```diff
@@ -11,16 +11,34 @@ use crate::*;
...
     pub y: f32,
     // implicit w = 1
 }
 
+#[cfg(feature = "serde")]
+fn serialize_f32_custom<S>(x: &f32, s: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    // Workaround for the fact that INFINITY and NEG_INFINITY serialize as null and then fail to deserialize
```
emilk, line 32: Only in JSON. Most other formats support INF and NAN.

emilk (commented):
Very interesting! Using `serde-diff` is a nice idea that i missed in [`eterm`](https://github.com/emilk/eterm).

I only took a brief look: are you serializing `Shape`s directly then? Does that work when adding new glyphs, e.g. by pasting in some emojis in a `TextEdit`? I had problems with that in `eterm`.

Is it running synchronously, waiting for the next frame before returning?

examples/remote_rendering/src/client.rs
linguini11: In my earlier tests I experimented with those formats, and found no visible difference in performance. I ended up sticking with json because it was (surprisingly) the quickest at serializing, and server-side speed was my priority. This may have been specific to my use case though. Given that json also causes the INF and NAN issues it may be best to change it regardless.

examples/remote_rendering/src/server.rs
linguini11: Perfect, thanks

linguini11:
It serializes FullOutput directly by implementing Serialize and Deserialize for FullOutput and all of its fields. Shapes are done in this way, with the exception that Callback does not work (serde(skip) is used to ignore it).
I had not tried with Glyphs. Quickly testing it out now, the thumbs up emoji has got a blank box beside it when rendered. Is that similar to the problems you were getting?

image

The examples are running synchronously yes, as this was sufficient for my purposes. I don't think switching to asynchronous would be too much of a problem if that is preferable.

lucasmerlin:
This could be interesting for #5418

PR #3328, opened 2023-09-12
Final polish for new image loading

jprochazk:
Part of https://github.com/emilk/egui/issues/3291

- [x] Add egui logo to demo widget gallery
- [x] Improve error message when no image loaders are installed
  image
- [x] Support `Cow<'static, str>` instead of only `&'static str` for `ImageSource::Bytes` URIs
- [x] Remove any remaining usage of `RetainedImage`
  - There is one `#![allow(deprecated)]` remaining in `egui_extras::image` itself.
- [x] Add `#[doc(hidden)]` to `egui_extras::image`
- [x] Replace `RawImage` in `Button` with just `Image`
- [x] Remove `RawImage`
- [x] Add a `Style` option for wether or not an image should show a spinner while loading (see https://github.com/rerun-io/rerun/issues/3244)


crates/egui_demo_lib/src/demo/widget_gallery.rs
```diff
@@ -209,7 +209,10 @@ impl WidgetGallery {
         let img_size = 16.0 * texture.size_vec2() / texture.size_vec2().y;
 
         ui.add(doc_link_label("Image", "Image"));
-        ui.raw_image((texture.id(), img_size));
+        ui.allocate_ui(img_size * 2.0 + egui::Vec2::new(8.0, 0.0), |ui| {
+            ui.image((texture.id(), img_size));
+            ui.image(egui::include_image!("../../assets/icon.png"));
+        });
```
emilk, line 215: Remove the old `texture` and just use the new API

```suggestion
        let egui_icon = egui::include_image!("../../assets/icon.png");
        ui.add(egui::Image(egui_icon).max_size(egui::Vec2::splat(16.0));
```

emilk: Also, if the icon is only being used here, we can downsize the .png quite a bit, e.g. to 32x32. It will improve its visual quality too since we don't do mip-mapping on the renderer yet.

crates/egui_plot/src/items/mod.rs
```diff
@@ -1234,12 +1234,13 @@ impl PlotItem for PlotImage {
...
-            .uv(*uv)
-            .rotate(screen_rotation, Vec2::splat(0.5))
-            .paint_at(ui, image_screen_rect);
+        ui.add(
+            Image::from_texture((*texture_id, image_screen_rect.size()))
+                .bg_fill(*bg_fill)
+                .tint(*tint)
+                .uv(*uv)
+                .rotate(screen_rotation, Vec2::splat(0.5)),
+        );
```
emilk, line 1243: no, calling `ui.add` here won't work well at all. That will put the image just anywhere. Check out the `Plot` demo in `egui_demo_app`: 
image

You should still use `Image::paint_at` here.

Btw, that plot demo could also use the egui icon image instead of `ColorImage::example`.

emilk (changes requested):

crates/egui/src/widgets/button.rs
```diff
@@ -235,11 +238,23 @@ impl Widget for Button {
...
+                    Ok(TexturePoll::Ready { texture }) => {
+                        image.paint_at(ui, image_rect, &texture);
+                    }
+                    Ok(TexturePoll::Pending { .. }) => {
+                        ui.add(Spinner::new().size(image_rect.size().min_elem()))
+                            .on_hover_text(format!("Loading {:?}…", image.uri()));
+                    }
+                    Err(err) => {
+                        ui.colored_label(ui.visuals().error_fg_color, "⚠")
+                            .on_hover_text(err.to_string());
```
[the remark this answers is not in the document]
jprochazk: How do I paint text into a specific `image_rect`? `Label` doesn't seem to have anything like a `paint_at`.

PR #3105, opened 2023-06-22
Implement table row framing and selection

vvv:
- Add framing effect to table rows
- Select table row by clicking
- Configure sensitivity of table cells

Demo
----

https://github.com/emilk/egui/assets/17462/5335a191-1bee-496e-a0c8-ad8d1c552afe

Closes #1519, #3069

PR #3347, opened 2023-09-16
Implement table row selection and hover highlighting

### LINKED ISSUE #1553: Add sense param for tables

laurooyen:
* Based on #3105 by @vvv.

## Additions and Changes

- Add `TableBuilder::sense()` and `StripBuilder::sense()` to enable detecting clicks or drags on table and strip cells.
- Add `TableRow::select()` which takes a boolean that sets the highlight state for all cells added after a call to it. This allows highlighting an entire row or specific cells.
- Add `TableRow::response()` which returns the union of the `Response` of all cells added to the row up to that point. This makes it easy to detect interactions with an entire row. See below for an alternative design.
- Add `TableRow::index()` and `TableRow::col_index()` helpers.
- Remove explicit `row_index` from callback passed to `TableBody::rows()` and `TableBody::heterogeneous_rows()`, possible due to the above. This is a breaking change but makes the callback compatible with `TableBody::row()`.
- Update Table example to demonstrate all of the above.

## Design Decisions

An alternative design to `TableRow::response()` would be to return the row response from `TableBody`s `row()`, `rows()` and `heterogeneous_rows()` functions. `row()` could just return the response. `rows()` and `heterogeneous_rows()` could return a tuple of the hovered row index and that rows response. I feel like this might be the cleaner soluction if only returning the hovered rows response isn't too limiting.

I didn't implement `TableBuilder::select_rows()` as described [here](https://github.com/emilk/egui/pull/3105#issuecomment-1618062533) because it requires an immutable borrow of the selection state for the lifetime of the `TableBuilder`. This makes updating the selection state from within the body unnecessarily complicated. Additionally the current design allows for selecting specific cells, though that could be possible by modifying `TableBuilder::select_rows()` to provide row and column indices like below.

```rust
pub fn select_cells(is_selected: impl Fn(usize, usize) -> bool) -> Self
```

## Hover Highlighting

EDIT: Thanks to @samitbasu we now have hover highlighting too.

~This is not implemented yet. Ideally we'd have an api that allows to choose between highlighting the hovered cell, column or row. Should cells containing interactive widgets, be highlighted when hovering over the widget or only when hovering over the cell itself? I'd like to implement that before this gets merged though.~

Feedback is more than welcome. I'd be happy to make any changes necessary to get this merged.

* Closes #1519
* Closes #1553
* Closes #3069


jazzay:
@emilk I am wondering if we could have this merged sooner than later without waiting for the hover. There have been multiple attempts at table row selection over the past year or so, and I am also now looking at such a capability for my own app. Rather than me implementing yet another solution, I could use this. 

I get the sense that hover is trickier to sort out, so perhaps that can come in later. What blocks application development is lack of row selection interactions. Hover while nice, is not critical in my opinion.

In the meantime @laurooyen would you be able to update your branch to include the latest 0.23 release?

laurooyen:
I'll update my branch in the comming days.

I agree that actually being able to select rows is more important than the hovering effect. If this gets merged without it, I'd be willing to do a follow up pull-request for it. In fact I had it working previously, the difficult part is deciding on the api, or none at all and enabling it automatically when sense is interactive.

jazzay:
Thanks @laurooyen. No rush, I have since merged your changes across to my own branch w 0.23 as base.

laurooyen:
One final problem I noticed is that a row is not hover highlighted when hovering an interactive widget within that row. This is visible in the [video](https://github.com/emilk/egui/pull/3347#issuecomment-1836707710) above when the mouse hovers the *Click me* checkbox.

The interactive widgets response intercepts the hover from the rows response. A solution could be to check if the row is hovered by checking if it contains the cursor position. However that causes rows to be hover highlighted when another window is on top of it as well. Not sure what the solution would be.

samitbasu:
@laurooyen  - I didn't notice that before.  However, now that I see it, it  does make sense that it behaves the way it does (to me anyway).  When you hover on the clickable widget, the row is not highlighted, but the row is also not selected when you click the widget.  So if highlighting the row means the row will become selected when clicked, it behaves intuitively.

laurooyen:
@samitbasu - I see where you're comming from, however I do believe the behaviour I described is more common. For example the outliner in both Unreal Engine and Blender behave this way. Since this isn't a deal breaker though I'll leave it as is and mark this ready for review.

samitbasu:
@laurooyen - do you want to add @vvv  and myself as reviewers?  We can review the PR.  Also, should we consider just breaking the table widget out into it's own crate?  Or is it better to leave it here?

jazzay:
Would be great to see this merged in soon. Thanks for all the efforts @laurooyen @samitbasu !
My opinion is it should be fine to leave the table in the already separate egui_extras crate. Table is a key widget for many apps. But up to @emilk.

crates/egui_extras/src/table.rs
```diff
@@ -836,7 +875,7 @@ impl<'a> TableBody<'a> {
     ///     .body(|mut body| {
     ///         let row_height = 18.0;
     ///         let num_rows = 10_000;
-    ///         body.rows(row_height, num_rows, |row_index, mut row| {
+    ///         body.rows(row_height, num_rows, |mut row| {
     ///             row.col(|ui| {
     ///                 ui.label("First column");
```
emilk, line 880: It would be good to demonstrate here how to get the row index

crates/egui_extras/src/table.rs
```diff
@@ -802,16 +835,22 @@ impl<'a> TableBody<'a> {
...
     /// If you have many thousands of row it can be more performant to instead use [`Self::rows`] or [`Self::heterogeneous_rows`].
     pub fn row(&mut self, height: f32, add_row_content: impl FnOnce(TableRow<'a, '_>)) {
+        let mut response: Option<Response> = None;
         let top_y = self.layout.cursor.y;
         add_row_content(TableRow {
             layout: &mut self.layout,
             columns: self.columns,
             widths: self.widths,
             max_used_widths: self.max_used_widths,
+            row_index: self.row_nr,
```
emilk: we should be consistent with the naming here

crates/egui_extras/src/table.rs
```diff
@@ -768,6 +798,9 @@ pub struct TableBody<'a> {
     /// If we find the correct row to scroll to,
     /// this is set to the y-range of the row.
     scroll_to_y_range: &'a mut Option<Rangef>,
+
+    hovered_row: Option<usize>,
+    hovered_row_id: egui::Id,
```
emilk, line 803: this is a bit confusing name - it sounds like the id of the row that is being hover, rather than the id of the storage of the hovered row index.

crates/egui_extras/src/layout.rs
```diff
@@ -94,34 +110,53 @@ impl<'l> StripLayout<'l> {
...
         add_cell_contents: impl FnOnce(&mut Ui),
     ) -> (Rect, Response) {
         let max_rect = self.cell_rect(&width, &height);
 
-        if striped {
-            // Make sure we don't have a gap in the stripe background:
-            let stripe_rect = max_rect.expand2(0.5 * self.ui.spacing().item_spacing);
+        // Make sure we don't have a gap in the stripe/frame/selection background:
+        let item_spacing = self.ui.spacing().item_spacing;
+        let gapless_rect = || max_rect.expand2(0.5 * item_spacing);
```
emilk, line 122: unnecessary use of a closure

emilk (approved):
Beautiful! ⭐⭐⭐

I just have a few nits that I'll push fixes for

enomado:
So, how exactly I can configure row hovering effect, leaving ability to interact with row contents?

enomado:
As for me, there must be only API for row styling and pointer accessing in table.

Event processing, list of selected rows should be in user level

PR #3353, opened 2023-09-18
file_storage: Wrap `File` in `BufWriter`

sebbert:
When exiting my application, it would freeze for around a minute waiting to finish writing ca. 40MB to file storage. Some quick profiling revealed that it was spending all that time in `File::write`, presumably making a large number of system calls writing tiny amounts of data at a time. We can avoid this by buffering writes using a `BufWriter`, just like we already do with `BufReader`.

With this change, my application takes around 1-2 seconds to exit, with the majority of that time spent serializing `ron`. I'm sure there are further potential performance improvements there, but this is already an order of magnitude or two better.

I also fixed the call to `log::warn!`, which was using incorrect formatting syntax and not including the error message.

I've run `scripts/check.sh`, all checks passed.

~~Closes <https://github.com/emilk/egui/issues/THE_RELEVANT_ISSUE>.~~

I didn't raise an issue first, since this is such a small change. Let me know if you want me to create one.

### DIFF
--- a/crates/eframe/src/native/file_storage.rs
+++ b/crates/eframe/src/native/file_storage.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::HashMap,
+    io::Write,
     path::{Path, PathBuf},
 };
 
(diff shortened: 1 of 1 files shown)


emilk (approved):
Wow, great find! Thanks

PR #3362, opened 2023-09-19
Only show on-screen-keyboard and IME when editing text

### LINKED ISSUE #2614: Windows shouldn't allow IME events when not editing text (On Desktop)
**Describe the bug**
We can always enable or disable IME because IME is always allowed. In a typical application, this attribute should be determined by whether a text is being edited.
https://github.com/emilk/egui/blob/0eabd894bd0e4466d53e3098e8cce1f10183f3b0/crates/eframe/src/native/run.rs#L563-L572
https://github.com/emilk/egui/blob/0eabd894bd0e4466d53e3098e8cce1f10183f3b0/crates/eframe/src/native/run.rs#L1011-L1024

**To Reproduce**
1. You'll need to have a input method editor.
2. Start a *native* egui app (on web it works fine) and you'll find that you can call up the IME when not editing (even there's no `textedit`).

**Expected behavior**
When editing, IME is allowed; Disallowed otherwise.

**Screenshots**
IME is on it's initial position.
(IME is on it's initial position.)

 - OS: Windows 22H2 (Windows 11)

**Additional context**
There's a field `mutable_text_under_cursor` in [`egui::output::PlatformOutput`](https://docs.rs/egui/latest/egui/output/struct.PlatformOutput.html).

currently it's used by `eframe` web, but it should also applicable to `egui_winit` or other crates.

### LINKED ISSUE #3057: Phosh on-screen keyboard pops up when app starts

Barugon:
* Closes #2614
* Closes #3057.

Allow IME if `PlatformOutput::text_cursor_pos` is `Some`.

* Hopefully this is a better approach than #3313.

### DIFF
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -777,7 +777,6 @@ mod glow_integration {
             let theme = system_theme.unwrap_or(self.native_options.default_theme);
             integration.egui_ctx.set_visuals(theme.egui_visuals());
 
-            gl_window.window().set_ime_allowed(true);
             if self.native_options.mouse_passthrough {
                 gl_window.window().set_cursor_hittest(false).unwrap();
             }
@@ -1268,8 +1267,6 @@ mod wgpu_integration {
             }
             let theme = system_theme.unwrap_or(self.native_options.default_theme);
             integration.egui_ctx.set_visuals(theme.egui_visuals());
-
-            window.set_ime_allowed(true);
 
             {
                 let event_loop_proxy = self.repaint_proxy.clone();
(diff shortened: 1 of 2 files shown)


emilk (approved):
This looks much better! How have you tested it?

Barugon:
I've tested it on a PinePhone Pro. 

It seems to be a bit expensive to call `Window::set_ime_allowed`, so I added some code to only call it when the state of `text_cursor_pos` changes.

Barugon:
On Phosh, the keyboard pops up when a text widget is focused and it's hidden when the text widget looses focus.

emilk (approved):

Issue #3372, opened 2023-09-23
Need more key mappings


mkrueger:
For my application I need some keys mapped:
'[', ']' and '\' - potentially ','

emilk:
Closed in favor of https://github.com/emilk/egui/issues/3653

PR #3373, opened 2023-09-23
Added key mappings for `[`, `]`, `,` and `\`.

mkrueger:
Added key mappings for '[', ']', ',' and '\'.

* Closes https://github.com/emilk/egui/issues/3372

## Added keys

* egui::Key::Backslash
* egui::Key::OpenBracket
* egui::Key::CloseBracket
* egui::Key::Comma

### DIFF
--- a/crates/egui/src/data/input.rs
+++ b/crates/egui/src/data/input.rs
@@ -962,6 +962,18 @@ pub enum Key {
     /// `,`
     Comma,
 
+    /// '/'
+    Backslash,
+
+    // '['
+    OpenBracket,
+
+    // ']'
+    CloseBracket,
+
+    /// '`'
+    Backtick,
+
     /// `-`
     Minus,
 
@@ -1094,6 +1106,10 @@ impl Key {
         Self::Plus,
         Self::Equals,
         Self::Semicolon,
+        Self::OpenBracket,
+        Self::CloseBracket,
+        Self::Backtick,
+        Self::Backslash,
         // Digits:
         Self::Num0,
         Self::Num1,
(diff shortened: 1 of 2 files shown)


crates/egui/src/data/input.rs
```diff
@@ -694,6 +694,12 @@ pub enum Key {
     PageUp,
     PageDown,
 
+    Backslash,
+    OpenBracket,
+    CloseBracket,
+    Comma,
+    Backquote,
```
emilk, line 701: Please add docstrings for these!

emilk (approved):

emilk:
Have you tested this on web?

mkrueger:
works on web for me.

crates/egui/src/data/input.rs
mkrueger: done.

mkrueger:
rebased the PR on current HEAD - no conflicts.

emilk:
Thanks for the PR! Unfortunately this requires another rebase 😬

mkrueger:
Too much changes - I'll look at it when I continue on my terminal program.

mkrueger:
there were many changes in that area :/

crates/egui/src/data/input.rs
```diff
@@ -962,6 +962,18 @@ pub enum Key {
...
+    Backslash,
+
+    // '['
+    OpenBracket,
+
+    // ']'
+    CloseBracket,
+
+    /// '`'
+    Backquote,
```
emilk, line 975: This has many different names, but I suggest we follow wikipedia (https://en.wikipedia.org/wiki/Backtick) and call it `Backtick`, but add "backquote" and "grave" as pseudonyms in the docstring.

mkrueger: fixed - I don't mind how it's called doesn't matter if it's grave, backtick or backquote - I just need the key :).

emilk (approved):

emilk:
` cargo check --locked --all-features --all-targets` fails

mkrueger:
ok fixed - some more renamings. Assumed that a VirtualKeyCode -> KeyCode but they renamed some keys.

LBracket -> BracketLeft
RBracket -> BracketRight
Grave -> Backquote :)

PR #3382, opened 2023-09-24
eframe: fix unexpected change with winit 0.28

icedrocket:
As of winit 0.28, with the addition of the `WindowsButtons` API, `resizable` no longer controls whether the maximize button is enabled or not. So in this pull request, I've made changes here to match the behavior of older versions of winit until we change our API.


emilk (commented):
So `resizable` used to also affect wether not the maximize button was shown, then winit fixed that rather weird behavior?

I think the preferable way to fix this is to introduce a way to control the maximize button in `NativeOptions` so that eframe users can toggle the two independently of each other.

There is some work towards that in https://github.com/emilk/egui/pull/3172

Issue #2747, opened 2023-02-18
Add hook to tap into window creation similar to EventLoopBuilder hook in eframe


twop:
### Problem
I would love to customize appearance beyond what is available in `eframe::NativeOptions` (example below is on macOS).

Specifically I would love to achieve this look (top window, code snippet is taken from `winit`):
<img width="703" alt="image" src="https://user-images.githubusercontent.com/6633216/219827978-6fcd0bce-7721-41c1-a242-fc18ec56560f.png">

It is a combination of these `winit` settings:
```rs
    let event_loop = EventLoop::new();

    let window_1 = WindowBuilder::new()
        .with_titlebar_buttons_hidden(true)
        .with_title_hidden(true)
        .with_titlebar_transparent(true)
        .build(&event_loop);
```

Note that these methods are specific to macOS (defined in `WindowBuilderExtMacOS`).
The closest that `eframe` provides today is 
```rs
 eframe::NativeOptions {
        fullsize_content: true,
        decorated: false
}
```
But it removes the ability to resize and "roundness" of the window.

### Desired solution
I would love to to tap into `WindowBuilder` with a hook similar to `EventLoopBuilder`.
Something like that
```rs
let options = eframe::NativeOptions {
    // proposed API for window builder
    window_builder: Some(Box::new(|builder| {
        #[cfg(target_os = "macos")]
        {
            use winit::platform::macos::WindowBuilderExtMacOS;
            builder
                .with_titlebar_buttons_hidden(true)
                .with_title_hidden(true)
                .with_titlebar_transparent(true)
        }
    })),

    // existing hook for event loop
    event_loop_builder: Some(Box::new(|builder| {
        #[cfg(target_os = "macos")]
        {
            use winit::platform::macos::{ActivationPolicy, EventLoopBuilderExtMacOS};
            builder.with_activation_policy(ActivationPolicy::Accessory);
        }
    })),

    ..Default::default()
};
```

### Alternative
Mimic `winit` window builder API in `eframe::NativeOptions`. It feels less desirable to me because
1. There is already a precedent for `EventLoop` hook
2. It seems that `eframe::NativeOptions` is mostly 1:1 to `winit` options, thus it feels repetitive.
3. With little code in `eframe` this provides access to a lot of features, AND it is future proof, e.g. when `winit` adds features `eframe` gets these features for "free".

---
Happy to contribute the code

twop:
[...]

PR #3390, opened 2023-09-26
Add `NativeOptions::window_builder` for more customization

twop:
Closes https://github.com/emilk/egui/issues/2747

- this PR is my attempt to tap into winit options to style the window
- I'm building an app that behaves akin to macOS spotlight, e.g. no standard controls to minimize or close the window

here is the snippet from my app code base
```rs
        resizable: true,
        always_on_top: true,
        run_and_return: true,
        window_builder: Some(Box::new(|builder| {
            #[cfg(target_os = "macos")]
            use winit::platform::macos::WindowBuilderExtMacOS;
            // code below allows for richer customization
            #[cfg(target_os = "macos")]
            return builder
                .with_fullsize_content_view(true)
                .with_titlebar_buttons_hidden(true)
                .with_title_hidden(true)
                .with_titlebar_transparent(true);

            builder
        })),
        ..Default::default()
    };
```

So that makes it possible to create completely custom look and feel
image

Happy to rework if needed, maintaining a fork is a pain ^_^

### DIFF
--- a/crates/eframe/src/epi/mod.rs
+++ b/crates/eframe/src/epi/mod.rs
@@ -41,7 +41,7 @@ pub type EventLoopBuilderHook = Box<dyn FnOnce(&mut EventLoopBuilder<UserEvent>)
 /// Hook into the building of a the native window.
 ///
 /// You can configure any platform specific details required on top of the default configuration
-/// done by `EFrame`.
+/// done by `eframe`.
 #[cfg(not(target_arch = "wasm32"))]
 #[cfg(any(feature = "glow", feature = "wgpu"))]
 pub type WindowBuilderHook = Box<dyn FnOnce(WindowBuilder) -> WindowBuilder>;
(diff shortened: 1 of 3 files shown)


crates/eframe/src/epi/mod.rs
```diff
@@ -38,6 +38,14 @@ pub use winit::event_loop::EventLoopBuilder;
 #[cfg(any(feature = "glow", feature = "wgpu"))]
 pub type EventLoopBuilderHook = Box<dyn FnOnce(&mut EventLoopBuilder<UserEvent>)>;
 
+/// Hook into the building of a the native window.
+///
+/// You can configure any platform specific details required on top of the default configuration
+/// done by `EFrame`.
```
emilk, line 44:
```suggestion
/// done by `eframe`.
```

crates/eframe/src/native/epi_integration.rs
```diff
@@ -75,7 +75,7 @@ pub fn read_window_info(
 pub fn window_builder<E>(
     event_loop: &EventLoopWindowTarget<E>,
     title: &str,
-    native_options: &epi::NativeOptions,
+    native_options: &mut epi::NativeOptions,
```
emilk, line 78: Adding `mut` everywhere is quite ugly.

This can be fixed by using `Rc<dyn FnOnce(…)>` as the callback type, instead of just `Box`. That will also allow cloning.

emilk (approved):
Very good idea

Issue #3419, opened 2023-09-30
Some minor rough edges around Painter and Vec2 API


Phen-Ro:
**Describe the bug**
While using egui, I ran into some (very) minor rough edges in some of its APIs.
* All of the Painter functions that supports Stroke use Into<Stroke> as a parameter, all except arrow().
* Vec2 supports `*=` but not `/=`.
* Vec2::angled() could be optimized to use sin_cos().
* Then while testing my patch for this, my system deadlocked on the lock_reentry_single_thread mutext test.

**To Reproduce**
```rust
painter.line_segment([egui::pos2(2.0, 2.0), egui::pos2(3.0, 3.0)], (1.0, egui::Color32::GREEN)); // OK
painter.arrow(egui::pos2(2.0, 2.0), egui::vec2(0.0, 1.0), (1.0, egui::Color32::RED)); // "arguments to this method are incorrect"
```

```rust
let mut my_vec2 = egui::vec2(1.0, 1.0);
my_vec2 *= 3.0; // OK
my_vec2 /= 2.0; // "binary assignment operation `/=` cannot be applied to type `emath::Vec2`"
```

**Expected behavior**
* arrow() should accept Into<Stroke> like all the other painter methods.
* Vec2 should implement DivAssign<f32>.
* Tests should pass.

**Additional context**
Of course all API issues these are unimportant, but if I ran into them maybe other people will too.  I'll submit a PR in a moment to address those.  There a few other places that take a parameter of Stroke instead of Into<Stroke>.  I haven't run into those personally, but I'll update them anyway for consistency.

Additionally, while running tests, I came across a problem with the lock_reentry_single_thread mutext test.  The [documentation for lock()](https://doc.rust-lang.org/stable/std/sync/struct.Mutex.html#method.lock) reads, "*The exact behavior on locking a mutex in the thread which already holds the lock is left unspecified. However, this function will not return on the second call (it might panic or deadlock, for example).*"  Nevertheless, the test asserts that taking a lock twice on the same thread will panic.  On my system, rust 1.70.0 does not panic but instead deadlocks.  I've added an `#[ignore]` on the test, but you may want to handle the situation differently.  It may be preferable to remove the test altogether.

PR #3420, opened 2023-09-30
Use `impl Into<Stroke>` as argument in a few more places

Phen-Ro:
* Functions that take Stroke were updated to take Into<Stroke> to make them consistent with other Into<Stroke> parameters.
* Vec2 implements DivAssign<f32>, to make it consistent with already implementing MulAssign<f32> and Div<f32>.
* Vec2::angled() uses sin_cos() rather than an individual sin() and cos() call for an immeasurable but hypothetical performance improvement.
* Disable the lock_reentry_single_thread() mutex test.  Lock()ing twice on the same thread is not guaranteed to panic.

* Closes <https://github.com/emilk/egui/issues/3419>.


Phen-Ro:
I was unable to complete scripts/check.sh because the winit crate failed with "The platform you're compiling for is not supported by winit".  I hope you don't mind that I'll move this PR to ready anyway.

Phen-Ro:
In particular, these were the lines that failed complaining about winit platform support:
`(cd crates/eframe && cargo check --no-default-features --features "wgpu")`
`(cd crates/egui_demo_app && cargo check --no-default-features --features "wgpu")`
`(cd crates/egui_glow && cargo check --all-features)`

Everything else passed.

crates/epaint/src/mutex.rs
```diff
@@ -389,6 +389,7 @@ mod tests {
 
     #[test]
     #[should_panic]
+    #[ignore = "Documentation does not gaurantee panic. \"The exact behavior on locking a mutex in the thread which already holds the lock is left unspecified. However, this function will not return on the second call (it might panic or deadlock, for example).\""]
```
emilk, line 392: Interesting! Then let's remove the test altogether instead

emilk (approved):

Phen-Ro:
Merge conflict resolved and mutex test removed.

PR #3429, opened 2023-10-02
Allow changing handle shape of a slider

### LINKED ISSUE #1974: Change slider indicator to a rounded rect
Although this is an opinionated topic, I think for the sake of themeability/skinning, users should be able to modify the look of sliders indicators and choose forms other than `epaint::CircleShape`. For a lack of a better word, the indicator is the small bar (or currently circle) in a slider:
 ```
======0======
       ^
```

A solution could be to use a rectangle and set rounding to `ui.visuals().widgets.inactive.rounding`, although this would change the overall look of egui. Another more backwards compatible solution would be to set the rounding to a default value which produces a circle, and add a method like `rounding()` which would override the rounding behavior.

I have considered building my own widget for this, but there are so many great attributes and methods in the current `Slider` implementation that this would feel overkill to me.

### PR DESCRIPTION
**Pull request by YgorSouza**
* Closes #1974

### DIFF
--- a/crates/egui/src/style.rs
+++ b/crates/egui/src/style.rs
@@ -815,6 +815,11 @@ pub struct Visuals {
     /// Enabling this will affect ALL sliders, and can be enabled/disabled per slider with [`Slider::trailing_fill`].
     pub slider_trailing_fill: bool,
 
+    /// Shape of the handle for sliders and similar widgets.
+    ///
+    /// Changing this will affect ALL sliders, and can be enabled/disabled per slider with [`Slider::handle_shape`].
+    pub handle_shape: HandleShape,
+
     /// Should the cursor change when the user hovers over an interactive/clickable item?
     ///
     /// This is consistent with a lot of browser-based applications (vscode, github
@@ -962,9 +967,6 @@ pub struct WidgetVisuals {
 
     /// Make the frame this much larger.
     pub expansion: f32,
-
-    /// Shape of the handle for sliders and similar widgets.
-    pub handle_shape: HandleShape,
 }
 
 impl WidgetVisuals {
@@ -1142,6 +1144,7 @@ impl Visuals {
             striped: false,
 
             slider_trailing_fill: false,
+            handle_shape: HandleShape::Circle,
 
             interact_cursor: None,
 
@@ -1213,7 +1216,6 @@ impl Widgets {
                 fg_stroke: Stroke::new(1.0, Color32::from_gray(140)), // normal text color
                 rounding: Rounding::same(2.0),
                 expansion: 0.0,
-                handle_shape: HandleShape::Circle,
             },
             inactive: WidgetVisuals {
                 weak_bg_fill: Color32::from_gray(60), // button background
@@ -1222,7 +1224,6 @@ impl Widgets {
                 fg_stroke: Stroke::new(1.0, Color32::from_gray(180)), // button text
                 rounding: Rounding::same(2.0),
                 expansion: 0.0,
-                handle_shape: HandleShape::Circle,
             },
             hovered: WidgetVisuals {
                 weak_bg_fill: Color32::from_gray(70),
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: crates/egui/src/widgets/slider.rs [hunk 1]
```diff
@@ -676,12 +676,36 @@ impl<'a> Slider<'a> {
...
+                        stroke: visuals.fg_stroke,
+                    });
+                }
+                style::HandleShape::Rect { aspect_ratio } => {
+                    let v = match self.orientation {
+                        SliderOrientation::Horizontal => Vec2::new(radius, radius / aspect_ratio),
+                        SliderOrientation::Vertical => Vec2::new(radius / aspect_ratio, radius),
+                    };
+                    let v = v + Vec2::splat(visuals.expansion);
+                    let rect = Rect {
```
emilk, line 695: You could use `Rect::from_center_size` here

crates/egui/src/widgets/slider.rs
```diff
@@ -720,12 +744,16 @@ impl<'a> Slider<'a> {
...
         let limit = match self.orientation {
             SliderOrientation::Horizontal => rect.height(),
             SliderOrientation::Vertical => rect.width(),
         };
-        limit / 2.5
+        let aspect_ratio = match handle_shape {
+            style::HandleShape::Circle => 1.0,
+            style::HandleShape::Rect { aspect_ratio } => *aspect_ratio,
+        };
+        limit / (2.5 / aspect_ratio)
```
emilk, line 756: If we define the radius to be on the cross axis (i.e. the Y axis on a horizontal slider), then we don't need to use `aspect_ratio` here

emilk (commented):
Looks good, but adding the settings to every `WidgetVisuals` seems overkill - I think the belongs in `Visuals` instead

YgorSouza:
> Looks good, but adding the settings to every `WidgetVisuals` seems overkill - I think the belongs in `Visuals` instead

I thought about that, but I figured maybe someone could want to change the aspect ratio or shape when the slider is hovered or clicked, as an alternative to the `expansion` setting, and that would be very hard to do without the setting on `WidgetVisuals`. On the other hand, it does mean that everyone who doesn't want that is going to have to change this setting 3 times instead of 1, so maybe it's not such a great tradeoff. I'll change it.

crates/egui/src/style.rs
```diff
@@ -1877,3 +1900,22 @@ fn rounding_ui(ui: &mut Ui, rounding: &mut Rounding) {
...
     });
 }
+
+fn handle_shape_ui(ui: &mut Ui, handle_shape: &mut HandleShape) {
+    ui.label("Widget handle shape");
+    ui.horizontal(|ui| {
+        ui.radio_value(handle_shape, HandleShape::Circle, "Circle");
+        if ui
+            .radio(
+                matches!(handle_shape, HandleShape::Rect { aspect_ratio: _ }),
```
emilk, line 1910: nit

```suggestion
                matches!(handle_shape, HandleShape::Rect { .. }),
```

crates/egui/src/widgets/slider.rs
```diff
@@ -284,6 +286,15 @@ impl<'a> Slider<'a> {
         self
     }
 
+    /// Change the shape of the slider handle
+    ///
+    /// This setting can be enabled globally for all sliders with [`Visuals::handle_shape`].
+    /// Changing it here will override the above setting ONLY for this individual slider.
+    pub fn handle_shape(mut self, handle_shape: HandleShape) -> Self {
```
emilk, line 293:
```suggestion
    #[inline]
    pub fn handle_shape(mut self, handle_shape: HandleShape) -> Self {
```

crates/egui_demo_lib/src/demo/sliders.rs
```diff
@@ -132,23 +137,43 @@ impl super::View for Sliders {
...
+                    matches!(handle_shape, HandleShape::Rect { aspect_ratio: _ }),
+                    "Rectangle",
+                )
+                .clicked()
+            {
+                *handle_shape = HandleShape::Rect { aspect_ratio: 0.5 };
+            }
+            if let HandleShape::Rect { aspect_ratio } = handle_shape {
+                ui.add(Slider::new(aspect_ratio, 0.1..=3.0).text("Aspect ratio"));
+            }
```
emilk, line 173: Duplicated code - I suggest you create an `impl HandleSape { fn ui(&mut self, ui: &mut Ui) { … } }` for this in `style.rs`

emilk (approved):

PR #3434, opened 2023-10-03
egui: fix scroll not sticking when scrollbar is hidden

LoganDark:
Before:

https://github.com/emilk/egui/assets/4723091/4b42842c-4f34-45d1-9538-29cd0bfd614e

https://github.com/emilk/egui/assets/4723091/7544e300-5b9a-4b2b-aaa4-9acade69fa75

After:

https://github.com/emilk/egui/assets/4723091/d720286f-a166-4891-874a-350b15683587


LoganDark:
actually, I just realized this fix is incomplete. This keeps the scroll stuck at the end, but it does not actually allow it to stick and unstick depending on scroll.

I have absolutely no clue how a billion checks could be failing just because I moved a block of code like 10 lines up, either. But oh well. I'll see if I can make a more complete fix

LoganDark:
on second thought and some testing, this is fine actually. This fix should be OK as implemented

emilk (approved):

LoganDark:
Do you need the merge conflict resolved?

LoganDark:
There, fixed that up for you.

Issue #3316, opened 2023-09-07
Support floating scrollbars


martenbjork:
When a view starts to overflow, scrollbars appear. These scrollbars take up physical space in the view, and pushes the content to the left. This causes a visual jump by a few pixels.

IMO, it would be better if the scrollbars floated on top of the content, like the default scrollbars on MacOS and iOS (and Window now as well, I believe?). That way, you avoid the jumpiness altogether.

https://github.com/emilk/egui/assets/875708/e0e54563-646a-4a5e-8a64-d6bb73096d20

emilk:
The default scrollbars on MacOS and iOS are invisible until you start scrolling, which can be very confusing since it doesn't tell the user there is anything to scroll. Do you have a suggestion on how to resolve that?

If they are always visible, then they will always cover the content, which is not great. Of course, we can reserve some space for them, but then we are back to them not being floating anymore.

YgorSouza:
In the Gnome desktop environment, the scrollbar disappears completely when the mouse is not moving, appears very thin when the mouse is moving far from it, and becomes wide when the mouse is close to it. 

[Screencast from 2023-09-18 20-53-18.webm](https://github.com/emilk/egui/assets/43298013/d7c9cd30-014d-4ffe-824a-e6287176e97c)

It can still block part of the UI if you don't put any margin, but I guess that's up to the UI designer to decide if that's a problem or not for their particular app.

![image](https://github.com/emilk/egui/assets/43298013/12941c40-b713-4172-995f-73c798b4e5ff)

emilk:
Thats pretty neat - thanks @YgorSouza!

It doesn't quite solve it for touch screens though :/  Perhaps always showing the thin scroll bar makes sense for touch screens.

martenbjork:
which can be very confusing since it doesn't tell the user there is anything to scroll. Do you have a suggestion on how to resolve that?

martenbjork:
(accidental PR close — never mind)

> ...which can be very confusing since it doesn't tell the user there is anything to scroll. Do you have a suggestion on how to resolve that?

(Not trying to be salty! 👉 ) I'd argue that it's not the role of individual apps to correct the operating system's default affordances. IMO, if Apple and Microsoft thinks invisible scrollbars is the best solution for scrollable areas, I think we should follow their lead.

I think there are some exceptions:

- The scrollable content is unusual, so users might not recognize it as scrollable.
[...]

PR #3435, opened 2023-10-03
feat: provide option to hide scroll bar unless interacting with the scroll area

zkldi:
The opacity of the scroll bar may optionally be tied
to the interaction the user is doing with the scroll area.

This has prior art in things like VSCode, which don't show scroll bars
on containers unless you're hovered over them or interacting
with the scroll bar.

This effect is disabled by default.

Peek 2023-10-03 17-27

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -107,6 +107,12 @@ pub struct ScrollArea {
     max_size: Vec2,
     min_scrolled_size: Vec2,
     scroll_bar_visibility: ScrollBarVisibility,
+
+    /// Should the scroll bar only be visible when this area is hovered over?
+    ///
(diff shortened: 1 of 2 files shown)


YgorSouza:
Discussed in #3316

zkldi:
Ah. This is almost what's described in that PR but with two caveats:
- it doesn't fade out if you're idle over the scrollview (this is consistent with vscode, inconsistent with gtk)
- it doesn't do anything about the scroll bar reserving area (which is a separate issue imo)

zkldi:
Update: having a click_n_drag handler on the outer rect completely breaks being able to interact with the actual contents of the scroll container.

That makes a lot of sense obviously. My bad.

emilk:
Closed by https://github.com/emilk/egui/pull/3539

PR #3438, opened 2023-10-06
Fix inconsistent naming of constrain_to versus constraint_to

rinde:
The changelog for 0.23.0 mentioned:

> * Add  `Area::constrain_to` and `Window::constrain_to` [#3396](https://github.com/emilk/egui/pull/3396)

However, the implementation for `Window` was called `constraint_to` (notice the extra 't'). I noticed this because the `deprecated` message points to the non-existing `constrain_to`.

### DIFF
--- a/crates/egui/src/containers/area.rs
+++ b/crates/egui/src/containers/area.rs
@@ -155,7 +155,7 @@ impl Area {
         self
     }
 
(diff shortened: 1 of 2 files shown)


emilk (approved):
Thanks!

Issue #3439, opened 2023-10-06
Broken Links in Demo


c-git:
**Describe the bug**

Some of the links in the [Widget Gallary](https://www.egui.rs/#demo) in the demo are broken. See screenshot for which are broken. Below is a list of the links you are taken to that return no search results.
- <https://docs.rs/egui/latest/egui/?search=label%2Cheading>
- <https://docs.rs/egui/latest/egui/?search=selectable_value%2CSelectableLabel>
- <https://docs.rs/egui/latest/egui/?search=DatePickerButton>
- <https://docs.rs/egui/latest/egui/?search=plot>

**To Reproduce**
1.  Go to demo and navigate to the [Widget Gallery](https://www.egui.rs/#demo)
2.  Click on one of the links indicated in the screenshot.
3.  See that you get no search results

**Expected behavior**

Link expected to take you to the documentation for the widget.

**Screenshots**

![image](https://github.com/emilk/egui/assets/43485962/6d5cdfa7-2897-4d79-b360-aa965deaaf07)

**Desktop (please complete the following information):**
 - OS:  22.04.1-Ubuntu
 - Browser  Firefox
 - Version  118.0.1

**Additional context**

Some of them appear to be just to update the search key, while others I think may have been moved into separate crates so maybe a link like what is done for custom widget at the bottom.

PR #3441, opened 2023-10-07
Fix broken doc links in the demo app widget gallery

YgorSouza:
* Closes <https://github.com/emilk/egui/issues/3439>


emilk (approved):
Nice one!

Issue #3076, opened 2023-06-12
egui_extras: simultaneous use of TableBody::row and TableBody::rows breaks the striping pattern


vvv:
### Describe the bug

Table rows may be striped incorrectly if the closure passed to `Table::body` uses `TableBody::row` together with either `TableBody::rows` or `TableBody::heterogeneous_rows`.

Note that first two rows on this screenshot have identical (lighter) background:

<img width="371" alt="image" src="https://github.com/emilk/egui/assets/17462/19621afa-d576-49ac-ae39-edd645aaaeb7">

### To reproduce

In the closure passed to `Table::body`:
- add one row with `TableBody::row`
- then add a few more rows with `TableBody::rows`

PoC code:

```rust
#![cfg_attr(not(debug_assertions), windows_subsystem = "windows")] // hide console window on Windows in release

use eframe::egui;

#[derive(Debug)]
pub struct HelloApp;

impl eframe::App for HelloApp {
    fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
        egui::CentralPanel::default().show(ctx, |ui| {
            egui::ScrollArea::horizontal().show(ui, |ui| {
                self.table_ui(ui);
            });
        });
    }
}

impl HelloApp {
    fn table_ui(&mut self, ui: &mut egui::Ui) {
        use egui_extras::{Column, TableBuilder};

        let row_height = ui.text_style_height(&egui::TextStyle::Body);
        TableBuilder::new(ui)
            .striped(true)
            .resizable(true)
            .column(Column::auto())
            .column(Column::initial(100.0).at_least(40.0).clip(true))
            .column(Column::remainder())
            .header(20.0, |mut header| {
                header.col(|ui| {
                    ui.strong("row_index");
                });
                header.col(|ui| {
                    ui.strong("Row created with");
                });
            })
            .body(|mut body| {
                body.row(row_height, |mut row| {
                    row.col(|ui| {
                        ui.label("0");
                    });
                    row.col(|ui| {
                        ui.label("TableBody::row");
                    });
                });
                body.rows(row_height, 3, |row_idx, mut row| {
                    row.col(|ui| {
                        ui.label(row_idx.to_string());
                    });
                    row.col(|ui| {
                        ui.label(format!("TableBody::rows"));
                    });
                });
            });
    }
}

fn main() -> eframe::Result<()> {
    eframe::run_native(
        "Wrong Stripes",
        eframe::NativeOptions::default(),
[...]

PR #3442, opened 2023-10-07
Fix Table stripe pattern when combining row() and rows()

YgorSouza:
* Closes <https://github.com/emilk/egui/issues/3076>


emilk (approved):
Sweet!

PR #3451, opened 2023-10-08
Update DragValue when focus is lost

YgorSouza:
Closes <https://github.com/emilk/egui/issues/2877>.

### DIFF
--- a/crates/egui/src/widgets/drag_value.rs
+++ b/crates/egui/src/widgets/drag_value.rs
@@ -473,13 +473,21 @@ impl<'a> Widget for DragValue<'a> {
 
         let text_style = ui.style().drag_value_text_style.clone();
 
+        // We draw both the TextEdit and the Button in all cases, making one of them invisible as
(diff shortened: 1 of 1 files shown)


YgorSouza:
I came up with this not so elegant solution which consists of always drawing both the `TextEdit` and the `Button`, and making one of them invisible. This is both to ensure that the Ids don't change when we toggle between states, and to give the `TextEdit` a chance to update the value after it has lost focus.

As far as I can tell, there is only one case where it still doesn't work: when we click away from the `DragValue` directly into a `TextEdit` that was painted after it. This is due to #2142.

In other cases (pressing Tab or Shift + Tab, clicking on empty space or another type of widget, or clicking on a `TextEdit` painted before the current one) it updates correctly. However, this adds a penalty in performance, as there are always two widgets being added now. And I'm not sure how this affects accessibility, as I have no experience with that. I'm just leaving this PR for consideration in case no one comes up with a better solution.

emilk:
I agree that the solution isn't so elegant :) 

I came up with something else in https://github.com/emilk/egui/pull/3776

PR #3461, opened 2023-10-11
Fix panic when window is missing in GlowWinitApp::window

### LINKED ISSUE #3460: Panic occurs when GlowWinitApp is missing window on android resume
Swapping in an out of an app and requesting repaints in the background causes a panic due to missing window when calling GlowWinitApp::window

Steps to reproduce the behavior on Android:
1. Swap out of a window
2. Request repaint in another thread
3. Swap back to the window

**Expected behavior**
No panic

devjobe:
Removing expect check from GlowWinitApp::window resolved the panic issue.

Closes #3460

### DIFF
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -629,8 +629,8 @@ mod glow_integration {
             Ok(())
         }
 
-        fn window(&self) -> &winit::window::Window {
-            self.window.as_ref().expect("winit window doesn't exist")
+        fn window(&self) -> Option<&winit::window::Window> {
+            self.window.as_ref()
         }
 
         fn resize(&self, physical_size: winit::dpi::PhysicalSize<u32>) {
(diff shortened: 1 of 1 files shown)


crates/eframe/src/native/run.rs
```diff
@@ -629,8 +629,8 @@ mod glow_integration {
             Ok(())
         }
 
-        fn window(&self) -> &winit::window::Window {
-            self.window.as_ref().expect("winit window doesn't exist")
+        fn try_window(&self) -> Option<&winit::window::Window> {
```
emilk, line 632: No need for the extra `try_` prefix methinks

```suggestion
        fn window(&self) -> Option<&winit::window::Window> {
```

devjobe: Fixed. Also removed the unnecessary error log as it's created for the first expect and checked for none before the second expect.

emilk:
This may have been fixed by https://github.com/emilk/egui/pull/3172 - please test!

PR #3475, opened 2023-10-16
Update accesskit and accesskit_winit.

ndarilek:

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -96,7 +96,7 @@ jobs:
           toolchain: 1.70.0
           targets: wasm32-unknown-unknown
 
-      - run: sudo apt-get update && sudo apt-get install libgtk-3-dev
+      - run: sudo apt-get update && sudo apt-get install libgtk-3-dev libatk1.0-dev libgtk-3-dev
 
       - name: Set up cargo cache
         uses: Swatinem/rust-cache@v2
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -96,7 +96,7 @@ jobs:
           toolchain: 1.70.0
           targets: wasm32-unknown-unknown
 
-      - run: sudo apt-get update && sudo apt-get install libgtk-3-dev libatk1.0-dev libgtk-3-dev
+      - run: sudo apt-get update && sudo apt-get install libgtk-3-dev libatk1.0-dev
 
       - name: Set up cargo cache
         uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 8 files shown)


ndarilek:
I'm not clear if *I* need to add labels to make this pass or not, but I don't appear able to.

In any case, it'd be nice if this could land in a patch release somehow--not sure if a dependency upgrade like this qualifies for that even if there aren't user-facing API changes. Bevy 0.12 releasing soon ships AccessKit 0.12 as well, and I'm hoping to release a bevy_egui integration that gets all of this working in Bevy games. But getting everything on the same version of AccessKit is a bit of a challenge.

Thanks!

DataTriny:
Hi @ndarilek 

Please note that it is mandatory to call [`accesskit_winit::Adapter::on_event`](https://docs.rs/accesskit_winit/latest/accesskit_winit/struct.Adapter.html#method.on_event) whenever a winit event is received by the app. If `false` is returned then the event should be discarded before it could be consumed by the app.

Actually this was the case since Unix support was introduced to AccessKit, but it is now a requirement on macOS as well. Not doing so breaks focus tracking on both platforms, as well as hit-testing on Unix.

ndarilek:
Thanks, going to work this in later today. Is there anything else needed for the upgrade? Asking because I'm trying to get Bevy 0.12 upgraded in time for its imminent release and want to make sure there's nothing else that's needed.`

DataTriny:
You should also populate the new `Tree` fields:

- app_name: do egui already have access to that?
- toolkit_name: egui,
- toolkit_version: you should be able to fetch it from a Cargo environment variable

These properties are exposed on Windows and Unix to help assistive technologies apply patches to specific toolkits and/or apps.

mwcampbell:
@ndarilek Do you recall why you modified .github/workflows/rust.yml to install libatk1.0-dev? I don't believe that's actually necessary.

ndarilek:
I thought there may have been a related build failure when I did this. If it builds without it, though, then feel free to remove it.

emilk (approved):

.github/workflows/rust.yml
```diff
@@ -96,7 +96,7 @@ jobs:
           toolchain: 1.70.0
           targets: wasm32-unknown-unknown
 
-      - run: sudo apt-get update && sudo apt-get install libgtk-3-dev
+      - run: sudo apt-get update && sudo apt-get install libgtk-3-dev libatk1.0-dev libgtk-3-dev
```
emilk, line 99:
```suggestion
      - run: sudo apt-get update && sudo apt-get install libgtk-3-dev libatk1.0-dev
```

PR #3478, opened 2023-10-17
egui: add redo support to Undoer

### LINKED ISSUE #3447: Add redo to Undoer

The `Undoer` does not have a redo functionality, and thus neither does the `TextEdit`. most text boxes that have undo also have redo, and there is a TODO for it in the source code.

https://github.com/emilk/egui/blob/38b4234c3282a7c044c18b77234ee8c204efe171/crates/egui/src/widgets/text_edit/builder.rs#L987-L988

The `Undoer` should push undone states to a redo stack, and have a `redo()` method that can be called to revert the `undo()` call, as well as a `has_redo()` method to help set the enabled/disable state for the corresponding UI button.

It might be possible to check the state of `has_undo()` before and after calling `feed_state()` or `add_undo()`, and push a state to a redo stack stored separately, but that seems complicated and error-prone, and this feature is needed for the `TextEdit` anyway, so it may as well be provided by the `Undoer`.

**Additional context**

I am willing to try to implement this feature myself, but I would like to know if there are any pitfalls I should consider, that might have prevented it from being implemented until now.

LoganDark:
* Closes #3447
* Closes #3448

Better implementation than #3448. (by accident since I did not see that PR)

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -978,10 +978,14 @@ fn events(
             Event::Key {
                 key: Key::Z,
                 pressed: true,
-                modifiers,
+                modifiers:
+                    Modifiers {
+                        command: true,
+                        shift: false,
+                        ..
+                    },
                 ..
-            } if modifiers.command && !modifiers.shift => {
-                // TODO(emilk): redo
+            } => {
                 if let Some((undo_ccursor_range, undo_txt)) = state
                     .undoer
                     .lock()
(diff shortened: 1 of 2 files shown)


LoganDark:
Honestly no idea why the labels check is failing. Either that or it means actual github labels which I don't have permission to apply.

LoganDark:
Hey @emilk what would it take to get this merged?

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -978,10 +978,14 @@ fn events(
...
+                modifiers:
+                    Modifiers {
+                        command: true,
+                        shift: false,
+                        ..
+                    },
                 ..
-            } if modifiers.command && !modifiers.shift => {
-                // TODO(emilk): redo
+            } => {
```
emilk, line 988: This is a bad way to match against Modifiers, as this will not catch `Ctrl+Z` on Windows and Linux, because that also sets the `ctrl` fields.

The old way was not great either, because it would ignore the `alt` key.

The preferred way of test `Modifiers` is to use `Modifiers::matches`:

It is better to use `if modifiers.matches(Modifiers::COMMAND)`

emilk:
> Better implementation than https://github.com/emilk/egui/pull/3448

Better how?

crates/egui/src/widgets/text_edit/builder.rs
LoganDark: My apologies, that method was not being used here so I wasn't aware of its existence. I'll fix this up

LoganDark:
> > Better implementation than #3448
> 
> Better how?

AFAIK this implementation just seemed more straightforward to me. I've since forgot the exact reason, though.

LoganDark:
Sigh. I literally ran `cargo cranky`. Not very cranky if it doesn't check formatting.

LoganDark:
In my own application, I have a variation of change history that allows you to have multiple redo stacks and switch between them, rather than them being cleared whenever you perform an action. However, that usually requires supporting UI, and isn't always intuitive to people, which is why I don't think it's right for `egui` to implement that itself. However, `TextEdit` makes it difficult to replace the default undo and redo actions with it. Not sure if there is an easy solution.

crates/egui/src/widgets/text_edit/builder.rs
emilk: No need to apologize - I didn't mean to come off as harsh. This is not easy to get right unless you already know it :)

emilk (approved):
Looks good!

PR #3479, opened 2023-10-17
add methods to manipulate TextEditState undoer

### LINKED ISSUE #3436: Request for Clearing Functionality in TextEdit Undoer for Enhanced Data Security
Dear emilk,

We need a feature to clear all content from the undoer in TextEdit.

For instance:

1. When editing a new file, where we need to get rid of the previously saved Undo function.
2. When entering an ID or password and we don't want the Undo function to work.

Using the id_source() function to disable the Undo feature is possible, but for data security, it would be better to have the ability to erase the content in the undoer.

Thanks, emilk.

LoganDark:
This feels awful, @emilk why does this have to be an Arc?

Should the Arc be replaced with a new one when set_undoer is called, or should it just replace the undoer inside the Arc?

Closes #3436

### DIFF
--- a/crates/egui/src/widgets/text_edit/state.rs
+++ b/crates/egui/src/widgets/text_edit/state.rs
@@ -6,7 +6,7 @@ use crate::*;
 
 use super::{CCursorRange, CursorRange};
 
-type Undoer = crate::util::undoer::Undoer<(CCursorRange, String)>;
+pub type TextEditUndoer = crate::util::undoer::Undoer<(CCursorRange, String)>;
 
 /// The text edit state stored between frames.
 ///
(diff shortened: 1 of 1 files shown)


LoganDark:
@emilk what are your thoughts on this? It's marked as draft just because I'm not sure what the final API design for this should be.

emilk:
Looks good to me. Why is it an `Arc`? The docstring explains it:

> Wrapped in Arc for cheaper clones.

LoganDark:
Alright. I primarily wanted to use this to save and restore the undoer state in response to user input, so cloning the inner part of the `Arc` seems most correct to me.

emilk (approved):

PR #3497, opened 2023-10-23
fix unusable rgba constructor

Norlock:
The constructor was not usable, so this small change will make it usable. I receive a private fields error when pub(crate) is used inside a tuple.


emilk:
There are many constructors:

https://docs.rs/ecolor/latest/ecolor/struct.Rgba.html#method.from_rgba_unmultiplied

PR #3498, opened 2023-10-23
Fix Shift+Tab behavior when no widget is focused

DataTriny:
If no widget is focused (such as when an application just started or after Escape was pressed), pressing Shift+Tab does not set the focus to the last widget. I think this PR fixes that.


emilk (approved):

DataTriny:
@emilk I've rebased this as an attempt to please cargo-deny, but I can't do much about the label job. Let me know if you need something more from me.

PR #3505, opened 2023-10-26
Updated to latest wgpu (0.18.0)

Wumpf:
Tested on M1 Mac:
* native
* webgl, firefox
* webgpu, chrome

all looking normal

Updated minor ahash version because 0.8.1 got yanked. Added some deny exceptions for now - we'll have to update winit soon to resolve glow related cargo deny errors (not a big issue though since we don't expect wgpu and glow backends to be used at the same time)

### DIFF
--- a/crates/egui-wgpu/src/winit.rs
+++ b/crates/egui-wgpu/src/winit.rs
@@ -107,7 +107,7 @@ impl Painter {
     ) -> Self {
         let instance = wgpu::Instance::new(wgpu::InstanceDescriptor {
             backends: configuration.supported_backends,
-            dx12_shader_compiler: Default::default(),
+            ..Default::default()
         });
 
         Self {
@@ -528,6 +528,7 @@ impl Painter {
                 });
 
             let mut render_pass = encoder.begin_render_pass(&wgpu::RenderPassDescriptor {
+                label: Some("egui_render"),
                 color_attachments: &[Some(wgpu::RenderPassColorAttachment {
                     view,
                     resolve_target,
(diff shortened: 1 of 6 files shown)


crates/egui-wgpu/src/winit.rs
```diff
@@ -538,20 +539,21 @@ impl Painter {
...
+                        store: wgpu::StoreOp::Store,
                     },
                 })],
                 depth_stencil_attachment: self.depth_texture_view.as_ref().map(|view| {
                     wgpu::RenderPassDepthStencilAttachment {
                         view,
                         depth_ops: Some(wgpu::Operations {
                             load: wgpu::LoadOp::Clear(1.0),
-                            store: true,
+                            store: wgpu::StoreOp::Discard,
```
torokati44, line 550: In all other places, `true` got replaced with `Store`, and `false` with `Discard`. Is this different intentionally?

Wumpf: well observed and yes it is! I should leave an explaining comment.
Typically the depth buffer used during rendering the main egui screen isn't needed afterwards. There might be a usecase for it but I have a pretty hard time thinking of one. In absence of that this unnecessarily costs bandwidth

torokati44: So it being `true` until now has been unnecessary then, right?

Wumpf: yep, I think so!

torokati44: Cool, thanks!

emilk:
There's quite a few added dependencies in `Cargo.lock` :(

emilk (approved):
Sad to see so many added/duplicated dependencies, but hopefully we can fix that once we update winit and glow

torokati44:
May I ask, which release is this going to be the part of, and when can that be expected? :eyes:

torokati44:
The reason I asked, is that it would have been nice (for Ruffle at least) to get an egui version that updated wgpu and naga, but not yet winit and rwh.
It's not super important though, we can wait for 0.24 just fine.

PR #3509, opened 2023-10-27
Fix: allow using the full Private Use Area for custom fonts

varphone:
The ignored characters are used in some custom fonts.

for example: the \u{F0FF} is used as `cleaning_services` in MaterialIcons-Regular.ttf

### DIFF
--- a/crates/epaint/src/text/font.rs
+++ b/crates/epaint/src/text/font.rs
@@ -139,6 +139,12 @@ impl FontImpl {
     ///
     /// See also [`invisible_char`].
     fn ignore_character(&self, chr: char) -> bool {
+        use crate::text::FontDefinitions;
+
+        if !FontDefinitions::builtin_font_names().contains(&self.name.as_str()) {
+            return false;
+        }
+
         if self.name == "emoji-icon-font" {
             // HACK: https://github.com/emilk/egui/issues/1284 https://github.com/jslegers/emoji-icon-font/issues/18
             // Don't show the wrong fullwidth capital letters:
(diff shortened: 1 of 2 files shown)


emilk (changes requested):
There is a better way to do this - only ignore `'\u{E0FF}' | '\u{EFFD}' | '\u{F0FF}' | '\u{F200}'` for `Ubuntu-Light`

crates/epaint/src/text/font.rs
```diff
@@ -139,6 +139,12 @@ impl FontImpl {
     ///
     /// See also [`invisible_char`].
     fn ignore_character(&self, chr: char) -> bool {
+        use crate::text::FontDefinitions;
+
+        if !FontDefinitions::builtin_font_names().contains(&self.name.as_str()) {
```
emilk, line 144:
```suggestion
        if !FontDefinitions::builtin_font_names().contains(&self.name) {
```

emilk (approved):

crates/epaint/src/text/font.rs
```diff
@@ -139,6 +139,12 @@ impl FontImpl {
     ///
     /// See also [`invisible_char`].
     fn ignore_character(&self, chr: char) -> bool {
+        use crate::text::FontDefinitions;
+
+        if !FontDefinitions::builtin_font_names().contains(&self.name) {
```
emilk, line 144:
```suggestion
        if !FontDefinitions::builtin_font_names().contains(&self.name.as_str()) {
```

emilk (approved):

PR #3513, opened 2023-10-30
Fix copy and cut on Safari

### LINKED ISSUE #3480: Copy doesn't work on Safari
**Describe the bug**

Copying text from https://egui.rs (e.g. from the text edit widget) or indeed Rerun web doesn't work on Safari. This is particularly annoying for Rerun's new Quick Start guide's code block, which are meant to be copied by the user.

**To Reproduce**

1. Go to https://egui.rs -> Widget Gallery
2. Enter text in the TextEdit widget
3. Select and copy that text

=> beeps and generate the following error in the console:

```
ERROR: [eframe::web] eframe/src/web/mod.rs:199: Copy/cut action failed: JsValue(NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.
writeText@[native code]
@https://www.egui.rs/egui_demo_app.js:941:46
<?>.wasm-function[644]@[wasm code]
<?>.wasm-function[516]@[wasm code]
<?>.wasm-function[2521]@[wasm code]
__wbg_adapter_30@https://www.egui.rs/egui_demo_app.js:226:136
real@https://www.egui.rs/egui_demo_app.js:209:21)
```

 - OS: macOS Sonoma 14.0
 - Browser: Safari 17.0

Related to:
- https://github.com/emilk/egui/issues/501

lunixbochs:
* Closes <https://github.com/emilk/egui/issues/3480>

I've tested this on Safari and Chrome on macOS Sonoma 14.0.

Could be improved to only call `event.preventDefault()` if `runner.logic()` actually performed a copy, but I don't see a way to get that information out with the current API.

### DIFF
--- a/crates/eframe/src/web/events.rs
+++ b/crates/eframe/src/web/events.rs
@@ -175,16 +175,30 @@ pub(crate) fn install_document_events(runner_ref: &WebRunner) -> Result<(), JsVa
     )?;
 
     #[cfg(web_sys_unstable_apis)]
-    runner_ref.add_event_listener(&document, "cut", |_: web_sys::ClipboardEvent, runner| {
-        runner.input.raw.events.push(egui::Event::Cut);
-        runner.needs_repaint.repaint_asap();
-    })?;
+    runner_ref.add_event_listener(
+        &document,
+        "cut",
+        |event: web_sys::ClipboardEvent, runner| {
+            runner.input.raw.events.push(egui::Event::Cut);
+            runner.logic();
+            runner.needs_repaint.repaint_asap();
+            event.stop_propagation();
+            event.prevent_default();
+        },
+    )?;
 
     #[cfg(web_sys_unstable_apis)]
-    runner_ref.add_event_listener(&document, "copy", |_: web_sys::ClipboardEvent, runner| {
-        runner.input.raw.events.push(egui::Event::Copy);
-        runner.needs_repaint.repaint_asap();
-    })?;
+    runner_ref.add_event_listener(
+        &document,
+        "copy",
+        |event: web_sys::ClipboardEvent, runner| {
+            runner.input.raw.events.push(egui::Event::Copy);
+            runner.logic();
+            runner.needs_repaint.repaint_asap();
+            event.stop_propagation();
+            event.prevent_default();
+        },
+    )?;
 
     Ok(())
 }


lunixbochs:
I also noticed middle clicking a "source code" URL in egui.rs triggers Safari's popup blocker. This is likely the same issue - eframe should run logic headlessly from inside the click handler so it inherits the "user interaction" permissions when processing the click.

crates/eframe/src/web/events.rs
```diff
@@ -175,16 +175,30 @@ pub(crate) fn install_document_events(runner_ref: &WebRunner) -> Result<(), JsVa
...
-    runner_ref.add_event_listener(&document, "cut", |_: web_sys::ClipboardEvent, runner| {
-        runner.input.raw.events.push(egui::Event::Cut);
-        runner.needs_repaint.repaint_asap();
-    })?;
+    runner_ref.add_event_listener(
+        &document,
+        "cut",
+        |event: web_sys::ClipboardEvent, runner| {
+            runner.input.raw.events.push(egui::Event::Cut);
+            runner.logic();
```
emilk, line 183:
```suggestion
            // In Safari we are only allowed to write to the clipboard during the
            // event callback, which is why we run the app logic here and now.
            runner.logic();
```

crates/eframe/src/web/events.rs
```diff
@@ -175,16 +175,30 @@ pub(crate) fn install_document_events(runner_ref: &WebRunner) -> Result<(), JsVa
...
     #[cfg(web_sys_unstable_apis)]
-    runner_ref.add_event_listener(&document, "copy", |_: web_sys::ClipboardEvent, runner| {
-        runner.input.raw.events.push(egui::Event::Copy);
-        runner.needs_repaint.repaint_asap();
-    })?;
+    runner_ref.add_event_listener(
+        &document,
+        "copy",
+        |event: web_sys::ClipboardEvent, runner| {
+            runner.input.raw.events.push(egui::Event::Copy);
```
emilk, line 195:
```suggestion
            runner.input.raw.events.push(egui::Event::Copy);
            // In Safari we are only allowed to write to the clipboard during the
            // event callback, which is why we run the app logic here and now.
```

emilk (approved):
Excellent, thank you!

PR #3515, opened 2023-10-30
Implemented window header color change when focused (on top).

GuillaumeSchmid:
This PR replaces an old one with many problems (no collapse icon, the header background was not funny colored, ugly...). It fixes those problems.

It implements the highlight of the header to the focused window. And allows the rest of the application to know the selected window. It allows, for instance, to display multiple windows with images and some additional meta information about those images on the side panel of the main window. The side panel updates itself according to the selected image window.

* Added a theme color for the selected window header.
* Added a function to retrieve the LayerId of the focused window.
* Implemented a simple demo of this function in the demo app where a message states if the Option window is focused (at the bottom of the Options window).

Capture d’écran du 2023-10-30 10-30-39

A technical point: 
The header color is applied with a transparency of 125 so the collapsible button becomes visible.
The reason is that the collapsible button is rendered before the rest of the header and before the header size is known.
We cannot draw the background before knowing this value, so rendering with transparency is a solution to see the collapsible button through the header background.

This PR has been sponsored by my company, which left me to do it during my work time.
This is part of an evil plan to convince them to switch to rust for new projects :)


emilk:
> This PR has been sponsored by my company, which left me to do it during my work time.
> This is part of an evil plan to convince them to switch to rust for new projects :)

I fully support your plan :)

emilk:
> A technical point:
> The header color is applied with a transparency of 125 so the collapsible button becomes visible.
> The reason is that the collapsible button is rendered before the rest of the header and before the header size is known.
> We cannot draw the background before knowing this value, so rendering with transparency is a solution to see the collapsible button through the header background.

There is a way to work around this! If you use this pattern:

```rust
let where_to_put_header_background = ui.painter().add(Shape::Noop);

// paint the collapsible button and figure out size of window…

ui.painter().set(
    where_to_put_header_background,
    epaint::RectShape::filled(…),
);
```

emilk:
We should pick a nomenclature and stick with it.

In the demo you use "focus", which is already used in the context of "keyboard focus" (press tab to move keyboard focus, which _doesn't_ change what window is on top).

I think "top" is the best word to use, as it is the most accurate.

crates/egui/src/memory.rs
```diff
@@ -819,6 +819,14 @@ impl Areas {
         }
     }
 
+    pub fn get_top_layer_id(&self, order: Order) -> Option<LayerId> {
```
emilk, line 822:
```suggestion
    pub fn top_layer_id(&self, order: Order) -> Option<LayerId> {
```

crates/egui_demo_lib/src/demo/window_options.rs
```diff
@@ -6,7 +6,6 @@ pub struct WindowOptions {
     closable: bool,
     collapsible: bool,
     resizable: bool,
-    constrain: bool,
```
emilk, line 9: why was the constrain demo removed?

crates/egui/src/style.rs
```diff
@@ -885,6 +893,7 @@ impl Visuals {
 
             window_shadow: Shadow::big_light(),
             window_fill: Color32::from_gray(248),
+            window_selected_header_color: Color32::from_rgb(32, 64, 255),
```
emilk, line 896: The dark mode color is nice a subtle, but this is too distinct imho.

emilk (commented):
Instead of adding yet another color, could we reuse an existing one, e.g. `visuals.widgets.hovered.bg_fill`?

The visual style is usually more relaxing when there is a limited pallett of colors

crates/egui_demo_lib/src/demo/window_options.rs
GuillaumeSchmid: I probably made a mistake using git, it is unintentional, I restored it.

crates/egui/src/containers/frame.rs
```diff
@@ -29,6 +29,8 @@ pub struct Frame {
 
     pub fill: Color32,
 
+    pub selected_header_color: Color32,
```
emilk, line 32: Unused?

GuillaumeSchmid: Yes unused. Sorry, it was part of my early experiments. I removed it.

crates/egui/src/context.rs
```diff
@@ -1905,6 +1905,11 @@ impl Context {
         self.memory_mut(|mem| mem.areas_mut().move_to_top(layer_id));
     }
 
+    /// Retrieve the `LayerId` of the top level windows.
+    pub fn get_top_layer_id(&self) -> Option<LayerId> {
```
emilk, line 1909:
```suggestion
    /// Retrieve the [`LayerId`] of the top level windows.
    pub fn top_layer_id(&self) -> Option<LayerId> {
```

https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

emilk (approved):

tmtbe:
There is a bug. When switching between dark and light tones, the setting of window_highlight_topmost will be set to true again.

margual56:
This PR had the following result when updating to the latest version of `egui`:
image

Is this a bug or am I missing something?

crates/egui/src/containers/window.rs
```diff
@@ -405,10 +405,19 @@ impl<'open> Window<'open> {
...
 
+        let on_top = Some(area_layer_id) == ctx.top_layer_id();
         let mut area = area.begin(ctx);
 
         let title_content_spacing = 2.0 * ctx.style().spacing.item_spacing.y;
 
+        // Calculate roughly how much larger the window size is compared to the inner rect
+        let title_bar_height = if with_title_bar {
+            let style = ctx.style();
+            ctx.fonts(|f| title.font_height(f, &style)) + title_content_spacing * 2.0
```
margual56, line 416: This (the `* 2.0`) is the culprit. When set to `1.0`, the result is correct:


image

emilk: Good catch - I'll revert it

emilk: I'm curious why only your window is broken, and no other one on egui.rs

Maybe you can make a PR to fix it?

Issue #358, opened 2021-05-08
Add methods to collapse a window


emilk:
We should add two methods to `egui::Window`

* `default_collapsed`: set whether the initial state of the window is collapsed
* `collapsed`: set whether the current state of the window is collapsed

nside:
i'll take it

nside:
i've got this https://github.com/nside/egui/commit/9285214cc0495e741b0029cbdf8a36f40accb722 but it's overriding the state in the collapsing header. somehow the ctx is needed to retrieve the header state. how would you handle this?

nside:
why not keep the collapsing header state in the window struct?

emilk:
`struct Window` is a builder - it stores no state at all between frames. The position and size of a window is stored in `area::State`, the scroll offset in `scroll_area::State`, the collapsing state in `collapsing_header::State`.

nside:
I meant adding an `Option<collapsing_header::State>` field to the `struct Window` - similar to how area and scroll are handled. Otherwise the collapsed bool would need to be handled similarly to the window `open` state?

leo848:
Any updates on this? Is there any way to have a window collapsed by default at this point?

lordbenedikt:
I also would like this feature. I'm creating an editor app with lots of functionalities that shouldn't be visible right at startup.

konradmoesch:
I'd also like this feature. Is someone working on this? I see #3517 marked as draft with no progress for four months.
Are you still working on it, @Norlock? 
Or would you be open to me taking over the PR or making a whole new attempt at this feature?

Norlock:
You can also continue, I have been busy with other projects so haven't really looked into it. But let me try if I can make an improvement (today/tomorrow). I will have to look at the code again.

EriKWDev:
I created #5661 which unlike #3517 just implements the suggested functions here. Though the event API also looks interesting, I think this should be a simpler addition

PR #3517, opened 2023-11-01
Manually open / closing windows

Norlock:
This is an API redesign for events. You can open/close/hide/show the window programmatically keeping the borrow checker unlocked. I updated the demo application with two new buttons to showcase. Everything is still backwards compatible. I do favor to deprecate the .open() function.

Closes [358](https://github.com/emilk/egui/issues/358)


crates/egui/src/containers/collapsing_header.rs
```diff
@@ -62,11 +61,23 @@ impl CollapsingState {
...
+        self.state.hidden
+    }
+
+    pub fn toggle_hidden(&mut self) {
+        self.state.hidden = !self.state.hidden;
+    }
+
+    pub fn set_hidden(&mut self, hidden: bool) {
+        self.state.hidden = hidden;
+    }
```
emilk, line 79: What does it mean for a `CollapsingHeader` to be hidden?

Norlock: Well windows can become hidden, collapsing headers won't do anything but I will update the PR, so this code won't be accessible when not using a window. I also have made the same event option for normal widgets to expand/collapse which I will include in this PR as well.

crates/egui/src/containers/collapsing_header.rs
```diff
@@ -6,8 +6,10 @@ use epaint::Shape;
 #[derive(Clone, Copy, Debug)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub(crate) struct InnerState {
-    open: bool,   // Expand / collapse
-    hidden: bool, // Show / Hide
+    /// Expand / collapse
+    open: bool,
+    /// Show / Hide (only on windows)
```
Mingun, line 11: I think, that sentence "only on windows" is misleading -- you may think, that you mean Windows OS

crates/egui/src/containers/collapsing_header.rs
Norlock: I think I will let hidden in the struct for now, otherwise it requires:

A:  Use of generics which I think is an absolute overkill in this instance and ugly.
B: Big restructure of code which I'm willing to do but I will have to take my time for that PR.

Norlock: I build a split between the two states on this branch: https://github.com/Norlock/egui/commit/40b1fb773bd3c327affb49614a505116d45be74b#diff-817170c50650af7962ada22bb6459b20eeb54ce921043e4db8522368ccf8933e. I don't think its a real improvement but you can check it out

PR #3521, opened 2023-11-02
WIP. draft. make PlotPoints accept borrowed data

valyagolev:
> When you have something that works, open a draft PR.

This is a draft PR. It's imperfect, not tested, and has some random edits. I have a question about one of them.

I'll make a better demo if this doesn't turn out to be such a bad idea (and after I've tried it in my projects)

### DIFF
--- a/crates/egui_plot/src/lib.rs
+++ b/crates/egui_plot/src/lib.rs
@@ -689,15 +689,18 @@ impl Plot {
     }
 
     /// Interact with and add items to the plot and finally draw it.
-    pub fn show<R>(self, ui: &mut Ui, build_fn: impl FnOnce(&mut PlotUi) -> R) -> PlotResponse<R> {
-        self.show_dyn(ui, Box::new(build_fn))
-    }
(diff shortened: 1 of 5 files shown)


crates/egui_plot/src/lib.rs
```diff
@@ -689,15 +689,18 @@ impl Plot {
...
+        FN: FnOnce(&mut PlotUi<'data>) -> R,
+    {
+        //     self.show_dyn(ui, Box::new(build_fn))
+        // }
+
+        // fn show_dyn<'a, R>(
+        //     self,
+        //     ui: &mut Ui,
+        //     build_fn: Box<dyn FnOnce(&mut PlotUi) -> R + 'a>,
+        // ) -> PlotResponse<R> {
```
valyagolev, line 703: I wonder about this one. why do we need to box the one-time callback?

s-nie: It's to reduce the amount of generated code by making sure that the plotting function can be reused.

valyagolev: understood, will undo the change

PR #3531, opened 2023-11-05
Fix rounding of `ImageButton`

### LINKED ISSUE #3486: ImageButton ignoring input Image rounding
```Rust
ui.add(ImageButton::new(
    egui::Image::new(SizedTexture::new(
        *a_texture_id,
        egui::Vec2::new(1.0, 1.0)
    )).rounding(50.0)
));
```
https://github.com/emilk/egui/blob/master/crates/egui/src/widgets/button.rs#L620 is the problem line I believe

**Expected behavior**
The image button should respect the input image's rounding. There should also maybe be an exposed `.rounding` method on the `ImageButton`. I might try to fix this later myself....

chriscate:
Updated `ImageButton` to use rounding from the corresponding image via `self.image.image_options().rounding` and added a `rounding()` method to `ImageButton`.

Closes <https://github.com/emilk/egui/issues/3486> <s>if we agree the addition of a `rounding()` method on the `ImageButton` itself, which the issue also mentions, can be discussed and handled separately.</s>

This now fully closes #3486

### DIFF
--- a/crates/egui/src/widgets/button.rs
+++ b/crates/egui/src/widgets/button.rs
@@ -621,7 +621,7 @@ impl<'a> Widget for ImageButton<'a> {
                 let selection = ui.visuals().selection;
                 (
                     Vec2::ZERO,
-                    Rounding::ZERO,
+                    self.image.image_options().rounding,
                     selection.bg_fill,
                     selection.stroke,
                 )
@@ -630,7 +630,7 @@ impl<'a> Widget for ImageButton<'a> {
                 let expansion = Vec2::splat(visuals.expansion);
                 (
                     expansion,
-                    visuals.rounding,
+                    self.image.image_options().rounding,
                     visuals.weak_bg_fill,
                     visuals.bg_stroke,
                 )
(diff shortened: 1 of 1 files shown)


chriscate:
One thing I noticed with the current approach in this PR is that when using e.g., `rounding(50.0)`, the border bumps up against the available space and the edge of the circle flattens out ever so slightly (see bottom circle in below screenshot). Is there a way to auto-scale this to account for that? Maybe that is one reason `rounding` was called out separately to begin with?

image

v-kat:
I'm fine with this although I'd also like to see the `rounding()` method. I have no clue why the original code separated out rounding....

chriscate:
I actually agree we need the `rounding()` method at the same time or else there's no quick way to override if the user chooses. I went ahead and added it in this same PR.

So, now this fully closes #3486 

@v-kat let me know what you think and then I'll open for review.

v-kat:
I haven't tried it but it looks okay

crates/egui/src/widgets/button.rs
```diff
@@ -594,6 +594,15 @@ impl<'a> ImageButton<'a> {
         self.sense = sense;
         self
     }
+
+    /// Set rounding for the `ImageButton`.
+    /// If the underlying image already has rounding, this
+    /// will override that value.
+    pub fn rounding(mut self, rounding: impl Into<Rounding>) -> Self {
+        let image = self.image.rounding(rounding.into());
+        self.image = image;
```
emilk, line 603:
```suggestion
        self.image = self.image.rounding(rounding.into());
```

emilk (approved):
Nice one, thanks!

chriscate:
You're welcome! Happy to contribute.

PR #3536, opened 2023-11-08
Add `ColorImage::from_gray_iter`

wangxiaochuTHU:
Add an alternative method for creating a [`ColorImage`] that accepts `Iterator` as the argument. It can be useful when `&[u8]` is not available but the iterator is.

### DIFF
--- a/crates/epaint/src/image.rs
+++ b/crates/epaint/src/image.rs
@@ -120,6 +120,16 @@ impl ColorImage {
         Self { size, pixels }
     }
 
+    /// Alternative method to `from_gray`.
+    /// Create a [`ColorImage`] from iterator over flat opaque gray data.
+    /// 
+    /// Panics if `size[0] * size[1] != gray_iter.len()`.
+    pub fn from_gray_iter<'_>(size: [usize; 2], gray_iter: impl Iterator<Item = &'_ u8> + ExactSizeIterator) -> Self {
+        assert_eq!(size[0] * size[1], gray_iter.len());
+        let pixels = gray_iter.map(|p| Color32::from_gray(*p)).collect();
+        Self { size, pixels }
+    }
+
     /// A view of the underlying data as `&[u8]`
     #[cfg(feature = "bytemuck")]
     pub fn as_raw(&self) -> &[u8] {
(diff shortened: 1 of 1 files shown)


crates/epaint/src/image.rs
```diff
@@ -120,6 +120,16 @@ impl ColorImage {
         Self { size, pixels }
     }
 
+    /// Alternative method to `from_gray`.
+    /// Create a [`ColorImage`] from iterator over flat opaque gray data.
+    /// 
+    /// Panics if `size[0] * size[1] != gray_iter.len()`.
+    pub fn from_gray_iter<'_>(size: [usize; 2], gray_iter: impl Iterator<Item = &'_ u8> + ExactSizeIterator) -> Self {
+        assert_eq!(size[0] * size[1], gray_iter.len());
+        let pixels = gray_iter.map(|p| Color32::from_gray(*p)).collect();
```
emilk, line 129:
```suggestion
    pub fn from_gray_iter(size: [usize; 2], gray_iter: impl Iterator<Item = u8>) -> Self {
        let pixels = gray_iter.map(Color32::from_gray).collect();
        assert_eq!(size[0] * size[1], pixels.len());
```

emilk: This will make it more general

emilk (approved):

crates/epaint/src/image.rs
wangxiaochuTHU: Absolutely agree. But the pull request is still here, maybe it was caused by my wrong click on the button "Commit Suggestion".

crates/epaint/src/image.rs
```diff
@@ -120,6 +120,16 @@ impl ColorImage {
         Self { size, pixels }
     }
 
+    /// Alternative method to `from_gray`.
+    /// Create a [`ColorImage`] from iterator over flat opaque gray data.
+    /// 
+    /// Panics if `size[0] * size[1] != gray_iter.len()`.
+    pub fn from_gray_iter(size: [usize; 2], gray_iter: impl Iterator<Item = u8>) -> Self {
+        let pixels = gray_iter.map(Color32::from_gray).collect();
```
emilk, line 128:
```suggestion
        let pixels: Vec<_> = gray_iter.map(Color32::from_gray).collect();
```

emilk (approved):

emilk:
please fix the compilation if you want this merged

PR #3537, opened 2023-11-08
Max window size & other size helpers

arduano:
Was a bit confused why the max_size API isn't exposed on `Window`, when it's perfectly functional in `Resize`. Anyway here's the main thing that it affects:

```rs
let screen = ctx.available_rect();
let size = screen.size();

egui::Window::new(self.name())
    .resizable(true)
    .resize(|resize| resize.max_size(size)) // Before
    .max_size(size)                         // After
    .show(ctx, |ui| todo!());
```

I also added some other relevant helpers for consistency.

This PR doesn't change any logic, only forwards along some helper functions that are already public for consistency.

### DIFF
--- a/crates/egui/src/containers/resize.rs
+++ b/crates/egui/src/containers/resize.rs
@@ -123,6 +123,18 @@ impl Resize {
         self
     }
 
+    /// Won't expand to larger than this
+    pub fn max_width(mut self, max_width: f32) -> Self {
+        self.max_size.x = max_width;
+        self
+    }
+
+    /// Won't expand to larger than this
+    pub fn max_height(mut self, max_height: f32) -> Self {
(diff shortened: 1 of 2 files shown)


Lord-Hellgrim (approved):
This looks very reasonable and I don't see any breaking bugs.

emilk (approved):
Looks good, but please add `#[inline]` to the functions

arduano:
@emilk I'm sorry, which functions? just the ones I added, or all size related ones, or all of them in the window struct?

emilk:
Oh, they don't? My bad. All trivial builder functions should have `#[inline]` on them, but I guess we haven't been great at enforcing that. I can take a pass on fixing that after this PR is merged

PR #3548, opened 2023-11-13
Add override to egui_plot

StratusFearMe21:
This PR simply allows you to override the opacity of a `Galley` when you draw it on screen. For my use case ([grezi-next](https://github.com/StratusFearMe21/grezi-next), a slideshow program), I am caching the text on slides as galleys, and drawing them on screen directly using the `ui.painter()`. My previous solution to controlling text opacity was to draw a background colored box in front of the text with the desired opacity. This caused a lot of jank for me, especially with a moving image as the background. Now I can fade in and out text without the jank!

https://github.com/emilk/egui/assets/57533634/6bee622e-8c17-4454-b442-9147f1b3bab0

Closes discussion <https://github.com/emilk/egui/discussions/3416>.

### DIFF
--- a/crates/egui/src/painter.rs
+++ b/crates/egui/src/painter.rs
@@ -443,6 +443,18 @@ impl Painter {
             self.add(Shape::galley_with_color(pos, galley, text_color));
         }
     }
+
+    /// Paint text that has already been laid out in a [`Galley`].
+    ///
+    /// You can create the [`Galley`] with [`Self::layout`].
+    ///
+    /// The text color in the [`Galley`] will be gamma multiplied by the given factor.
+    #[inline(always)]
+    pub fn galley_with_gamma(&self, pos: Pos2, galley: Arc<Galley>, gamma_multiply: f32) {
+        if !galley.is_empty() && gamma_multiply != 0.0 {
+            self.add(Shape::galley_with_gamma(pos, galley, gamma_multiply));
+        }
+    }
 }
 
 fn tint_shape_towards(shape: &mut Shape, target: Color32) {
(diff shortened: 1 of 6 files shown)


StratusFearMe21:
P.S: Just to clarify, the reason I can't use `ui.painter().galley_with_color()` is because, as is shown in the video, syntax highlighted code could be present on slides, and would be un-colored if I used that.

crates/epaint/src/shape.rs
```diff
@@ -685,6 +695,10 @@ pub struct TextShape {
     /// This will NOT replace background color nor strikethrough/underline color.
     pub override_text_color: Option<Color32>,
 
+    /// If set, the text color will be gamma multiplied by the given factor
+    /// This will NOT replace background color nor strikethrough/underline color.
+    pub gamma_multiply: Option<f32>,
```
emilk, line 700: Is there any difference between `None` and `Some(1.0)`? If not, I suggest just using a `f32` here defaulting to `1.0`, and testing against in the tessallalator for early-out-optimization

crates/egui/src/painter.rs
```diff
@@ -443,6 +443,18 @@ impl Painter {
...
         }
     }
+
+    /// Paint text that has already been laid out in a [`Galley`].
+    ///
+    /// You can create the [`Galley`] with [`Self::layout`].
+    ///
+    /// The text color in the [`Galley`] will be gamma multiplied by the given factor.
+    #[inline(always)]
+    pub fn galley_with_gamma(&self, pos: Pos2, galley: Arc<Galley>, gamma_multiply: f32) {
```
emilk, line 453: I think we should just refer to this as `opacity_factor` and then document that it is applied in gamma-space.

PR #3554, opened 2023-11-15
egui-extras: add ReqwestLoader as alternative to EhttpLoader

PingPongun:
This PR adds new http loader (`ReqwestLoader`) to `egui-extras`, as an alternative to `EhttpLoader`. 

If the rest of the project uses `reqwest`+`tokio`, using this new loader (that uses `reqwest` instead of `ehttp`) reduces project dependencies (in my case this was c. -35 crates & -0.3MB binary size in release).

Considering `reqwest` popularity (according to `crates.io` `reqwest` has more daily downloads than `ehttp` of all time, I don't know how this looks like for `egui` users only, but I guess it's still popular), I think this loader would better fit integrated into official `egui-extras` than as separate crate (but if You would find this fits better outside `egui`, I would probably release this as separate crate).

There is a small catch, about using `ReqwestLoader`, but I described it in feature description (even if sb misses it, it should error in a descriptive way).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -288,7 +288,7 @@ dependencies = [
  "polling",
  "rustix 0.37.25",
  "slab",
- "socket2",
+ "socket2 0.4.9",
  "waker-fn",
 ]
 
@@ -1351,10 +1351,12 @@ dependencies = [
  "log",
  "mime_guess2",
  "puffin",
+ "reqwest",
  "resvg",
  "serde",
  "syntect",
  "tiny-skia",
+ "tokio",
  "usvg",
 ]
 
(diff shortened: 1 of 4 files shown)


emilk:
This fits better as a 3rd-party crate. The point of the design of the byte-loaders is that you should easily be able to pick-and-chose which ones to use. In this case, disable the `http` feature of `egui_extras` and then just add the reqwest-based loader from a 3rd party crate (`egui_reqwest_loader` or whatever).

I appreciate that you want to minimize the number of dependencies. That's exactly why I chose `ehttp` (which uses the popular `ureq` on native) as the default. Here is a comparison of the dependency tree of `ureq` vs `reqwest`:

ureq

reqwest

Issue #3556, opened 2023-11-16
Improvement to multi-viewports


emilk:
Multi-viewports were implemented in https://github.com/emilk/egui/pull/3172, but there are several features missing that would make it a lot nicer, including:

* Make it easy to wrap child viewports in the same chrome as `crate::Window`
* Automatically show embedded viewports using `crate::Window`
* Use the new `ViewportBuilder` in `eframe::NativeOptions`
* Automatically position new viewport windows (they currently cover each other)
* Add a `Context` method for listing all existing viewports
* …

juancampa:
Not sure if out of scope for this issue but one that I'm particularly interested in:

 - Implement viewports for the web backed (should the app provide a `create_canvas_for_viewport` callback?)

My use case for this, btw, is improving performance of the [Membrane](https://www.membrane.io) extension by splitting the app into canvases that can be re-rendered independently.

emilk:
That defiantly belongs here! We can break out smaller issues from this one as people start working on them

juancampa:
I spent a few hours this weekend investigating how viewports could work on eframe web. For our particular use-case I was hoping we could run a viewport on a separate iframe using `OffscreenCanvas`.

The plan was this:
 - Run the main application on one iframe (application A)
 - Run an "empty shell" eframe application (application B)
 - Send canvas from iframe B to iframe A using [`transferControlToOffscreen`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/transferControlToOffscreen)
 - A can then use the received `OffscreenCanvas` to paint viewport contents.
 - B forwards all events to A

I'm not sure this approach would work though because some event handlers need to run the `logic` synchronously, which AFAIK cannot be done, unfortunately.

GamingLiamStudios:
I might be mistaken on how to properly use `RawInput` here, but shouldn't there also be a distinction on the `RawInput::events` per viewport? For example, you might alt-tab from one viewport to another. Or if you're using Wayland and get mouse events in surface-local coordinates.

PR #3564, opened 2023-11-17
Replace `eframe::Frame` commands and `WindowInfo` with egui

emilk:
* Part of https://github.com/emilk/egui/issues/3556

## In short
You now almost never need to use `eframe::Frame` - instead use `ui.input(|i| i.viewport())` for information about the current viewport (native window), and use `ctx.send_viewport_cmd` to modify it.

## In detail

This PR removes most commands from `eframe::Frame`, and replaces them with `ViewportCommand`.
So `frame.close()` becomes `ctx.send_viewport_cmd(ViewportCommand::Close)`, etc.

`frame.info().window_info` is now also gone, replaced with `ui.input(|i| i.viewport())`.

`frame.info().native_pixels_per_point` is replaced with `ui.input(|i| i.raw.native_pixels_per_point)`.

`RawInput` now contains one `ViewportInfo` for each viewport.

Screenshots are taken with `ctx.send_viewport_cmd(ViewportCommand::Screenshots)` and are returned in `egui::Event` which you can check with:

``` ust
ui.input(|i| {
    for event in &i.raw.events {
        if let egui::Event::Screenshot { viewport_id, image } = event {
            // handle it here
        }
    }
});
```

### Motivation
You no longer need to pass around the `&eframe::Frame` everywhere.
This also opens the door for other integrations to use the same API of `ViewportCommand`s.

### DIFF
--- a/crates/eframe/src/epi/mod.rs
+++ b/crates/eframe/src/epi/mod.rs
@@ -892,15 +892,6 @@ impl Frame {
         self.wgpu_render_state.as_ref()
     }
 
-    /// Bring the window into focus (native only). Has no effect on Wayland, or if the window is minimized or invisible.
-    ///
-    /// This method puts the window on top of other applications and takes input focus away from them,
-    /// which, if unexpected, will disturb the user.
-    #[cfg(not(target_arch = "wasm32"))]
-    pub fn focus(&mut self) {
-        self.output.focus = Some(true);
-    }
-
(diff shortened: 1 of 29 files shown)


chrisheib:
Previous to this I used to do `frame.info().window_info().position` to get the window coordinates, which is no longer possible. The new viewport doesn't seem to have a position field, and the outer_rect stays `None` for my App.

Issue #3597, opened 2023-11-21
Datepicker year range manual selection


vaqxai:
**Is your feature request related to a problem? Please describe.**
I can't select years less than 5 before current year, this is bad for example when I want to specify a birthdate.

**Describe the solution you'd like**
I'd like to be able to manually define the year range for the year picker. There is already a scrollbar. There should be no such restriction. At least change the hardcoding to allow 100 years into the past.

**Describe alternatives you've considered**
No good alternative datepickers for egui

PR #3599, opened 2023-11-21
egui_extras: Add more years for datepicker

vaqxai:
Birthdates really needed this

Kinda closes [<https://github.com/emilk/egui/issues/3597>](https://github.com/emilk/egui/issues/3597).


Mingun:
It seems it is better to add a setting to the widget rather than trying to guess best value for everyone...

vaqxai:
The change proposed in this PR is just a temporary fix, I'll be adding a full range option soon, too

emilk (approved):

PR #3603, opened 2023-11-21
Keep unsafe code forbidden when puffin is disabled

YgorSouza:
This helps document what unsafe is being used for, and prevent other uses from going unnoticed.


emilk (approved):
Good call

PR #3604, opened 2023-11-21
PaintCallbackInfo::viewport_in_pixels now guarantees to stay in bounds

Wumpf:
* Fixes https://github.com/rerun-io/rerun/issues/4297
* tested against a very hasty and incomplete port of egui/trunk, found [here](https://github.com/rerun-io/rerun/tree/andreas/experimental-egui-trunk)

In rare cases in can happen that the viewport returned by `PaintCallbackInfo` is outside the bounds of the screen.
for at least [wgpu/webgpu in particular](https://www.w3.org/TR/webgpu/#dom-gpurenderpassencoder-setviewport) this is invalid usage, other backends might be affected as well.
Since this happened due to a float rounding error (in one repro case I had I got (width==1126.5625) + (offset=715.4376) = 1842.0001 for a resolution of 1842) I decided to do away with fractional values on the viewport alltogether. They _technically_ make sense since a viewport is only specifying the NDC to pixel coordinate conversion, but practically this may lead to surprising sub-sampling issues.

### DIFF
--- a/crates/egui-wgpu/src/renderer.rs
+++ b/crates/egui-wgpu/src/renderer.rs
@@ -475,10 +475,10 @@ impl Renderer {
                             let viewport_px = info.viewport_in_pixels();
 
                             render_pass.set_viewport(
-                                viewport_px.left_px,
-                                viewport_px.top_px,
-                                viewport_px.width_px,
-                                viewport_px.height_px,
+                                viewport_px.left_px as f32,
+                                viewport_px.top_px as f32,
+                                viewport_px.width_px as f32,
+                                viewport_px.height_px as f32,
                                 0.0,
                                 1.0,
                             );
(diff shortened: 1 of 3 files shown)


emilk (changes requested):
Keeping in bounds makes sense, and rounding to integers makes sense too.
I’m not sure it should be `epaint`s job, but since this will likely be a footgun for many backend, maybe it should.

There is a small detail of _how_ to round.
We want two adjacent Rect (sharing an edge) to have non-overlapping viewports with no gap or overdraw. For that I think we need to round the edges (right/bottom) rather than the sizes (width/height).

Also, if `epaint` is always rounding, then `ViewportInPixels` should use `i32` instead of `f32`

emilk (approved):

PR #3605, opened 2023-11-21
Add indeterminate state to checkbox

YgorSouza:

### DIFF
--- a/crates/egui/src/widgets/button.rs
+++ b/crates/egui/src/widgets/button.rs
@@ -418,11 +418,18 @@ impl<'a> Widget for Checkbox<'a> {
             response.mark_changed();
         }
         response.widget_info(|| {
-            WidgetInfo::selected(
-                WidgetType::Checkbox,
-                *checked,
-                galley.as_ref().map_or("", |x| x.text()),
-            )
+            if indeterminate {
+                WidgetInfo::labeled(
+                    WidgetType::Checkbox,
+                    galley.as_ref().map_or("", |x| x.text()),
+                )
+            } else {
+                WidgetInfo::selected(
+                    WidgetType::Checkbox,
+                    *checked,
+                    galley.as_ref().map_or("", |x| x.text()),
+                )
+            }
         });
 
         if ui.is_rect_visible(rect) {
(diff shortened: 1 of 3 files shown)


YgorSouza:
[Screencast from 2023-11-21 21-38-32.webm](https://github.com/emilk/egui/assets/43298013/eb5ef1ac-d3f4-4f6d-a6a1-ea9d99eb0250)

DataTriny:
Hello,

In case of an indeterminate checked state, `Checked::Mixed` must be passed to [`NodeBuilder::set_checked`](https://docs.rs/accesskit/latest/accesskit/struct.NodeBuilder.html#method.set_checked) when an accessible node is built [here](https://github.com/emilk/egui/blob/9253cafedda22a719ccfc7f898f56e8da631b3b0/crates/egui/src/response.rs#L618).

DataTriny:
From an accessibility standpoint this is now OK, the state of the check box in the demo app is correctly picked up by ATs. Thanks!

crates/egui/src/widgets/button.rs
```diff
@@ -355,24 +355,40 @@ impl Widget for Button<'_> {
...
             text: text.into(),
+            indeterminate: false,
         }
     }
 
     pub fn without_text(checked: &'a mut bool) -> Self {
         Self::new(checked, WidgetText::default())
     }
+
+    /// Display an indeterminate state (neither checked not unchecked)
```
emilk, line 374:
```suggestion
    /// Display an indeterminate state (neither checked nor unchecked)
```

crates/egui/src/widgets/button.rs
```diff
@@ -419,7 +439,16 @@ impl<'a> Widget for Checkbox<'a> {
...
             ));
 
-            if *checked {
+            if indeterminate {
+                // Horizontal line:
+                ui.painter().add(Shape::line(
+                    vec![
+                        pos2(small_icon_rect.left(), small_icon_rect.center().y),
+                        pos2(small_icon_rect.right(), small_icon_rect.center().y),
+                    ],
```
emilk, line 448: nit: use `Shape::hline` for shorter code

emilk (approved):
Nice!

Issue #3602, opened 2023-11-21
Introduce a persistent `zoom_factor`


emilk:
Currently we have `pixels_per_point` which the user can change with `ctx.set_pixels_per_point`. However, it is confusing how that works when e.g. dragging an eframe window from one monitor to another with a different pixel density.

I suggest we introducer a `zoom_factor: f32` (default 1.0) which would be a multiplier on top of the native pixels_per_point. This means dragging a window from one monitor to another would keep the same `zoom_factor` (e.g. 120%), even though the monitors might have different pixel densities.

I'd also like to build-in the cmd+/- gui scaling to egui, which would then operator on `zoom_factor`.

`zoom_factor` would thus replace a settable `pixels_per_point`. `Context` would instead computer `pixels_per_point` as `zoom_factor * native_pixels_per_point`, where `zoom_factor` is read/write, and `pixels_per_point` is read-only.

We would then persist `zoom_factor` in `egui::Memory`.

This is essentially what we do manually at Rerun: https://github.com/rerun-io/rerun/pull/1448/files

----

One wrinkle is: what unit to use for input/output? Currently we use UI "points" everywhere, e.g. when describing mouse coordinates and output triangle vertices.

This means integrations would need to ask for `Context::pixels_per_point()` to calculate it.
We could also change the input/output interface to use native pixels for everything, but that's a pretty big change, and can lead to many subtle bugs.

Resonanz:
Suggestion: allow setting limits on the zoom factor. 

Useful, for example, for limiting the end user's ability to change scaling outside of the limiting bounds of say 1.0  and 1.5.

Resonanz:
Pressing Ctrl + plus or minus changes the zoom factor.

However, in the code below, pressing and holding this key combination results in the scaling jumping back and forth.

Why would the scaling jump around if the zoom factor is applied at the beginning of each redraw cycle?  Doesn't this code lock the zoom factor to 1.0?

```
impl eframe::App for TemplateApp {
    fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
        ctx.set_zoom_factor(1.0);
```

PR #3608, opened 2023-11-22
Introduce global `zoom_factor`

emilk:
* Closes https://github.com/emilk/egui/issues/3602

You can now zoom any egui app by pressing Cmd+Plus, Cmd+Minus or Cmd+0, just like in a browser. This will change the current `zoom_factor` (default 1.0) which is persisted in the egui memory, and is the same for all viewports.
You can turn off the keyboard shortcuts with `ctx.options_mut(|o| o.zoom_with_keyboard = false);`

`zoom_factor` can also be explicitly read/written with `ctx.zoom_factor()` and `ctx.set_zoom_factor()`.

This redefines `pixels_per_point` as `zoom_factor * native_pixels_per_point`, where `native_pixels_per_point` is whatever is the native scale factor for the monitor that the current viewport is in.

This adds some complexity to the interaction with winit, since we need to know the current `zoom_factor` in a lot of places, because all egui IO is done in ui points. I'm pretty sure this PR fixes a bunch of subtle bugs though that used to be in this code.

`egui::gui_zoom::zoom_with_keyboard_shortcuts` is now gone, and is no longer needed, as this is now the default behavior.

`Context::set_pixels_per_point` is still there, but it is recommended you use `Context::set_zoom_factor` instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1317,6 +1317,8 @@ dependencies = [
  "image",
  "log",
  "poll-promise",
+ "puffin",
+ "puffin_http",
  "rfd",
  "serde",
  "wasm-bindgen",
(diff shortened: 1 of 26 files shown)


Barugon:
I really feel that this should not be enabled by default.

PR #3622, opened 2023-11-24
Add a tutorials folder with simple tutorials

Agent-E11:
`tutorials.md` is the "home page"
The other `md` files are different sections (layout, technical, miscellaneous, etc.)

I have found it somewhat difficult getting started with `eframe` and I think it would be immensely helpful to have a collection of how-to's for doing small tasks (e.g. making a `ui.group()` the full available height).

This is (very) obviously not comprehensive, but I think it would be a good starting point. I am also not entirely sure about the layout, I think it's ok, but probably not perfect.

---

Addresses <https://github.com/emilk/egui/issues/186>.

### DIFF
--- a/crates/eframe/tutorials/layout.md
+++ b/crates/eframe/tutorials/layout.md
@@ -1,0 +1,10 @@
+# Layout tutorials
+
+## Make group fill full height of its parent
+
+```rust
+ui.group(|ui| {
+    // Stuff blah blah blah
+    ui.set_height(ui.available_height());
+});
+```
--- a/crates/eframe/tutorials/miscellaneous.md
+++ b/crates/eframe/tutorials/miscellaneous.md
@@ -1,0 +1,1 @@
+# Miscellaneous tutorials
--- a/crates/eframe/tutorials/technical.md
+++ b/crates/eframe/tutorials/technical.md
@@ -1,0 +1,20 @@
+# Technical tutorials
+
+## Press a button when a key is pressed
+
+If you want to have something that happens when either:
+
+- a button is clicked, or
+- a key is pressed
+
+This is how you can do that:
+
+```rust
+let my_button = ui.button("Press me");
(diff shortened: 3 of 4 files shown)


emilk:
Thanks for starting work on this!

The problem with this approach is code-rot: there is nothing checking that the code in the `.md` files are correct, and keep being correct as breaking changes land in `egui`.

I think a better approach is to collect these tricks into one tutorial application that users can compile and run.

If we embed it in the `egui_demo_lib` the tutorial can become a new section on egui.rs, which would link to its own source code, like most of the examples there already does.

Agent-E11:
@emilk
Ok.
1. Would the "one tutorial application" go in `egui/crates/egui_demo_lib`?

2. I think it could be hard to organize the small snippets if they are all in one application. (Maybe split it into the different categories like I did, but with separate apps instead of separate folders?)

3. And, if they were in an application, would it be possible to search for specific tutorials?

4. Looking through the examples in `egui_demo_lib`, I am having trouble figuring out how it works (it isn't laid out like a simple, single window app). Would it be possible to make the tutorial app simpler than the egui.rs site? (Single page, like the [http demo](https://www.egui.rs/#Http)) Possibly have it as another tab in egui.rs?

5. Would it be possible to have the tutorial app's code in a separate folder? So that it is easy to browse the code?:

```
egui_demo_lib/src/tutorial
| - layout
|   | - bla-bla-bla
|
| - technical
|   | - bla-bla-bla
|
| - miscellaneous
|   | - bla-bla-bla
```

6. I am realizing that this is getting really long, should I start a discussion?

YgorSouza:
You can make `cargo test` test .md files by including them in a crate or module's documentation with e.g. `#![doc = include_str!("../../../README.md")]`. They will be tested even if the module is private, so you could add hidden modules to some crate just for that.

Issue #3625, opened 2023-11-24
Window flashes white on launch on Windows


zeozeozeo:
**Describe the bug**
This was previously fixed in https://github.com/emilk/egui/issues/1802, but was reintroduced in commit https://github.com/emilk/egui/commit/1571027556b57cef5908df094038795c09b9090a.

**To Reproduce**
1. Start app in debug or release mode
2. Window flashes white for ~0.5 seconds

**Expected behavior**
No window flash on startup.

**Screenshots**
![window-flash](https://github.com/emilk/egui/assets/108888572/31a675af-8c30-46b1-a1af-653491cf67da)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version 0.24.0

**Additional context**
Perhaps the `is_first_frame` check is broken, or the window is being created with `set_visible(true)` in the first place, when it should start invisible until all assets are loaded.

7sDream:
Can reproduce this after upgrade to 0.24.0. 

OS: Windows 11 Preview Build 25997

zeozeozeo:
Fixed in https://github.com/emilk/egui/pull/3631.

sourcebox:
I also had this issue on macOS, did not do extensive testing though.

MrWheatley:
The white flashing is still happening to me with 0.25.0 and on master as well.

![gif](https://github.com/emilk/egui/assets/28412095/363ae6d9-4ace-4206-ad1a-3acbfa83101a)

emilk:
Is both the `glow` and `wgpu` backends affected?

MrWheatley:
Trying to use the `wgpu` backend I get this error:
```sh
error: failed to select a version for `web-sys`.
    ... required by package `wgpu v0.18.0`
    ... which satisfies dependency `wgpu = "^0.18.0"` of package `eframe v0.25.0 (https://github.com/emilk/egui.git?branch=master#7733d1d8)`
    ... which satisfies git dependency `eframe` of package `temp v0.1.0 (C:\Users\mrmon\PycharmProjects\temp)`
versions that meet the requirements `^0.3.64` (locked to 0.3.67) are: 0.3.67

the package `wgpu` depends on `web-sys`, with features: `GpuComputePassTimestampWrite` but `web-sys` does not have these features.


failed to select a version for `web-sys` which could resolve this conflict
```

torokati44:
[...]

PR #3631, opened 2023-11-25
eframe: Fix window flashing white on launch

zeozeozeo:
This is a bug that can occur on Windows, previously it was fixed by only showing the window after the first frame was rendered, but the bug appeared again with egui 0.24.0. This commit fixes this bug by making the window invisible on startup.

## Before

window-flash

## After

no-window-flash

Closes https://github.com/emilk/egui/issues/3625.

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -154,7 +154,8 @@ impl GlowWinitApp {
             event_loop,
             native_options,
             window_settings,
-        );
+        )
+        .with_visible(false); // Start hidden until we render the first frame
 
         let mut glutin_window_context = unsafe {
             GlutinWindowContext::new(egui_ctx, winit_window_builder, native_options, event_loop)?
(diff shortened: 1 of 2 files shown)


crates/eframe/src/native/glow_integration.rs
```diff
@@ -154,7 +154,8 @@ impl GlowWinitApp {
             event_loop,
             native_options,
             window_settings,
-        );
+        )
+        .with_visible(false); // Start hidden until we render the first frame
```
emilk, line 158:
```suggestion
        .with_visible(false); // Start hidden until we render the first frame to fix white flash on startup (https://github.com/emilk/egui/pull/3631)
```

crates/eframe/src/native/wgpu_integration.rs
```diff
@@ -837,7 +837,9 @@ fn create_window(
         event_loop,
         native_options,
         window_settings,
-    );
+    )
+    .with_visible(false); // Start hidden until we render the first frame
```
emilk, line 841:
```suggestion
    .with_visible(false); // Start hidden until we render the first frame to fix white flash on startup (https://github.com/emilk/egui/pull/3631)
```

emilk (approved):
Thanks!

Issue #3628, opened 2023-11-24
eframe window changes size constantly


RegenJacob:
**Describe the bug**
I'm on Hyprland and when I open an eframe app this happens:

https://github.com/emilk/egui/assets/53652728/8fd27371-8783-48ba-9382-e2e2ac2eb4c4

It worked in 0.23 without those issues. Maybe related to #3625 ?

**To Reproduce**
1. Be on Hyprland (maybe other Wayland Compositors too?)
2. Run eframe app in debug mode (didn't test release yet)
3. The window resizes like shown in the video

**Expected behavior**
Window shouldn't change the size  

**Desktop (please complete the following information):**
 - OS: Linux arch 6.6.2-zen1-1-zen (Hyprland on wayland) 
 - Version: eframe 0.24

zeozeozeo:
Probably related to https://github.com/emilk/egui/issues/3620

RegenJacob:
Also doesn't work on wgpu feature enabled

emilk:
Does `RUST_LOG=trace` give you any clues?

ItsEthra:
This only happens when `x11` feature is enabled.

emilk:
Nice find @ItsEthra - @konkitoman wrote that `set_maximized` - Do we really need it @konkitoman ? It seems like a bad idea. Maybe it should be moved to window initialization instead of running it each frame?

emilk:
Potential fix: https://github.com/emilk/egui/pull/3643

konkitoman:
> Nice find @ItsEthra - @konkitoman wrote that `set_maximized` - Do we really need it @konkitoman ? It seems like a bad idea. Maybe it should be moved to window initialization instead of running it each frame?

Is not really needed, that can be removed.

But the position will be off by 2..12 pixels on X11 Plasma

Issue #3620, opened 2023-11-24
App window is brought to foreground whenever `Context::request_repaint` is called


Barugon:
egui version 0.24.0

**Describe the bug**
In addition to what the title says, even without calling `Context::request_repaint`, the window remains on top when I switch to another app.

**To Reproduce**
1. Build and runt the attached project. It will issue a `Context::request_repaint` every second.
2. Switch to another app and note that this app's window will be moved to the foreground.

Even if `Context::request_repaint` is not deliberately called, moving your mouse over any exposed part of the window will bring it to the front.

**Expected behavior**
`Context::request_repaint` should not effect the desktop window order.

**Desktop**
- Memory:                        32.0 GiB
- Processor:                     AMD Ryzen™ 9 3900X × 24
- Graphics:                       AMD Radeon™ RX 6700 XT
- OS Name:                      Fedora Linux 39 (Workstation Edition)
- OS Type:                        64-bit
- GNOME Version:           45.1
- Windowing System:       Wayland
- Kernel Version:              Linux 6.5.12-300.fc39.x86_64

[testy.zip](https://github.com/emilk/egui/files/13456054/testy.zip)

emilk:
Thanks for the report!

Is this a new bug for 0.24?

I cannot reproduce on Mac, so I suspect this is a bug specific to Linux/Wayland.

sourcebox:
I can confirm that this bug is new in 0.24 and happens also on Linux without using Wayland. I use  `Context::request_repaint_after` in my application.

Barugon:
> Is this a new bug for 0.24?

Yes.

> Does it still happen if you use the `"wgpu"` feature of `eframe`?

If I try to build with the `"wgpu"` feature then I get this error:
```
the package `wgpu` depends on `web-sys`, with features: `GpuComputePassTimestampWrite` but `web-sys` does not have these features.
```

RegenJacob:
I tested it with the wgpu feature enabled and the issue still exist

emilk:
> > Does it still happen if you use the `"wgpu"` feature of `eframe`?
> 
> If I try to build with the `"wgpu"` feature then I get this error:
> 
> ```
> the package `wgpu` depends on `web-sys`, with features: `GpuComputePassTimestampWrite` but `web-sys` does not have these features.
> ```
> 
> [edit] I don't know why `web-sys` needs to be pulled in for a non-wasm project.

[...]

PR #3643, opened 2023-11-27
Fix windowing problems when using the `x11` feature on Linux

emilk:
This was caused by a hack added by @konkitoman in https://github.com/emilk/egui/pull/3172

The solution was reported by @ItsEthra in https://github.com/emilk/egui/issues/3628#issuecomment-1827943346

* Closes https://github.com/emilk/egui/issues/3620
* Closes https://github.com/emilk/egui/issues/3628


konkitoman:
That can be removed!
But on Plasma 5.27.9 X11 the viewport position will be out of sync!

But is a Plasma specific bug, that happens in multiples apps!

emilk:
Ok, then I'll remove the hack completely - thanks for the clarification @konkitoman !

Issue #3542, opened 2023-11-11
Update to winit 0.29


cybersoulK:
It brings a lot of fixes, specially to the wasm backend.
So it's a priority bump for web apps that use egui.

A few that i am aware are:
resize fixes
firefox pointers/mouse scroll issues
fixes to window animation loop

dbalsom:
Winit 0.29 also fixes some major issues with international keyboards. There are entire characters you can't send to egui like ':' on Italian layouts.

Issue #3303, opened 2023-09-05
Add non-portable keys to egui::Key


timstr:
Currently, the [documentation for `egui::Key`](https://docs.rs/egui/latest/egui/enum.Key.html) reads:

> Many keys are omitted because they are not always physical keys (depending on keyboard language), e.g. `;` and `§`, and are therefore unsuitable as keyboard shortcuts if you want your app to be portable.

I think this is an unfortunate restriction for the case of wanting to develop an app for a limited audience with a known keyboard. I want to be able to assign features to e.g. the `/` key on my North American keyboard layout to focus to a search bar. I'm unable to, it seems to be implied, because not _all_ keyboard layouts have a slash key. I really appreciate placing portability first but I think the available keys on my actual keyboard should at least come second.

To this end, would it be acceptable to have an API to detect whether any key is physically present? Non-portable keys could be found and queried this way, but non-portable keys should also be included in e.g. the `events` field of [`egui::RawInput`](https://docs.rs/egui/latest/egui/struct.RawInput.html) so that they can be acted on in the same way as existing keys. Appropriate naming and requiring at least some extra typing on the programmer's part would make them think about why they're using non-portable keys and what it means for their audience.

If this approach sounds reasonable, I could try and put in some time to prototype these changes.

Related: https://github.com/emilk/egui/issues/3266

Thanks by the way for this wonderful project! I've really been enjoying it

ahicks92:
I was actually coming here to open the same issue in order to frontload a few project ideas I've had, since I need accessibility and Egui is one of the only crates with that.  My two use cases:

- It would be nice to use egui as the "top level" library for games, since I'm blind, I'm writing for other blind people, and getting it nested in anything else accessibly is a whole project.
  - Such games usually use *all* the keys.  Not figuratively. Literally.
- I'm considering writing a MIDI sequencer.  In that use case, . (dotted note) and / (divide the note) etc. are useful.  You can think of this as a "text" control except it's not receiving text in the sense of normal text controls, but mnemonics for other things.

Looking at this, it seems we're down at least 9 common keys, namely all punctuation.  And though it doesn't matter to me being as it isn't portable even in English layouts, we're also probably down the numpad too.

emilk:
Will be closed by https://github.com/emilk/egui/pull/3649

Issue #2977, opened 2023-05-07
Forward `scancodes` via egui-winit


zkldi:
**Is your feature request related to a problem? Please describe.**

I'm currently working on reimplementing an existing codebase using `egui` and `eframe`. For part of this application, the keybindings SHIFT+1, SHIFT+2, SHIFT+3, and so on... are used *frequently*. Changing these keybindings isn't really an option as all existing users of this software practically have it burned into their memory.

The problem here is that these inputs (shift+1, etc.) are instead passed as `egui::Event::Text` instead of the raw scancode information. Since Shift+2 might output `Text('@')` or `Text('"')`, it's not possible to hack around this by reading the characters sent to the device.

Furthermore, things like `consume_shortcut` don't work with shortcuts like this.

**Describe the solution you'd like**

I'd like for the `virtual_keycode` part of `winit::event::KeyboardInput` to be accessible somehow from the eframe input handler. That way, I can handle this logic myself. At the moment, it is converted to a key with `translate_virtual_key_code` and that results in the loss of a lot of information.

**Describe alternatives you've considered**

I've tried checking if `Num1` is pressed and `modifiers.shift` is pressed, but that doesn't work as the input is coerced into `"!"`.

It's possible that I'm missing some way already to access the raw `winit::` events that already exists?

**Additional context**

I don't know much about the `egui` codebase or where this could nicely fit in as a feature. Strapping it onto `egui::Event::Key` could work, but seems unorthodox. Some sort of alternative place to access "raw" winit input would be potentially interesting too.

I'd be willing to work on this if there were some pointers on how this could fit in.

edisno:
I would also like this.  To deploy egui apps on Android TV for example, I have been forced to vendor egui-winit to forward the scancodes to get all the special remote control buttons.  Also if I remember correctly, winit's key code list is more complete than egui's.  It would be nice to have the scancode, or a more complete keycode list in egui (copy winit's list for example).  Of course the translation scancode -> key code should probably really happen in winit, but the key code list in egui must be extended or we must have access to the raw scan code I think.

zkldi:
I'd be very interested in that vendoring of egui-winit if you could make that public. Perhaps that could be shaped up into a PR for egui?

sdasda7777:
Completely agree there should be a way to access the raw key information. There are many layouts that don't have keys A-Z, or 0-9. The current approach just assumes too much about user's machine and a more general one is clearly needed.

emilk:
[...]

Issue #3501, opened 2023-10-24
Multiple svg images render blurry


thequver:
**Describe the bug**
When adding multiple svg images from the same source but with different size, the latter image renders blurry as if it takes the texture from the first added (smaller) image.

**To Reproduce**
Code to reproduce:
```
                const IMG_SRC: egui::ImageSource = egui::include_image!("./img.svg");
                ui.add_sized(egui::Vec2{x: 20.0, y: 20.0}, egui::Image::new(IMG_SRC));
                ui.add_sized(egui::Vec2{x: 200.0, y: 200.0}, egui::Image::new(IMG_SRC));
```

Image i used to test:
![img](https://github.com/emilk/egui/assets/70153252/01435573-f557-4ecd-a7e9-723b1185b268)

**Expected behavior**
Both vector images must look sharp and render at respective for its size resolution

**Screenshots**
![Screenshot](https://github.com/emilk/egui/assets/70153252/69c54394-a539-41ce-ac19-23bc8fa2af3b)

**Desktop (please complete the following information):**
 - OS: Win11

**Additional context**
Reproduced on win11, android 10, eframe, egui-winit + egui-wgpu implementation, egui 0.23.0, master branch (as the time of writing)

chriscate:
This appears to be due to the the caching done by the ImageLoader. If you call `ctx.forget_image(IMG_SRC.uri().unwrap());` between your two `ui.add_sized()` calls it will work as expected based on my testing. There is also a TODO [here](https://github.com/emilk/egui/blob/f218825d94f8bbaffd6e9bc1119b1073d013f629/crates/egui_extras/src/loaders.rs#L1) for automatic cache eviction.

If anyone has guidance on how to implement that automatic eviction I'd be happy to take a shot at it.

YgorSouza:
I was looking into this, and found that the SVG loader actually includes the size hint as part of its key:

https://github.com/emilk/egui/blob/5cf99c6308e36b5f36861accea6835f190924cce/crates/egui_extras/src/loaders/svg_loader.rs#L12-L15

Which would solve this problem, but also sounds kind of dangerous, because if you have an SVG that fills the available space, you would end up with hundreds of copies at different sizes as the UI is resized.

But in any case, it isn't working as intended, because the image loader is called by the texture loader, which does not take the size into account.

https://github.com/emilk/egui/blob/5cf99c6308e36b5f36861accea6835f190924cce/crates/egui/src/load/texture_loader.rs#L13-L24

[...]

PR #3649, opened 2023-11-27
Update to winit 0.29

fornwall:
* Closes https://github.com/emilk/egui/issues/3542
* Closes https://github.com/emilk/egui/issues/2977
* Closes https://github.com/emilk/egui/issues/3303


fornwall:
Current status: On mac, running `cargo run --release -p egui_demo_app` brings up the window and it can be resized, but not interacted with.

crates/egui-winit/src/lib.rs
```diff
@@ -1182,10 +1177,10 @@ fn process_viewport_command(
             }));
         }
         ViewportCommand::IMEPosition(pos) => {
-            window.set_ime_position(PhysicalPosition::new(
-                pixels_per_point * pos.x,
-                pixels_per_point * pos.y,
-            ));
+            window.set_ime_cursor_area(
+                PhysicalPosition::new(pixels_per_point * pos.x, pixels_per_point * pos.y),
+                PhysicalSize::new(100, 100),
```
fornwall, line 1182: TODO: What size to use here?

crates/eframe/src/native/run.rs
```diff
@@ -39,7 +39,7 @@ fn create_event_loop(native_options: &mut epi::NativeOptions) -> EventLoop<UserE
     let mut builder = create_event_loop_builder(native_options);
 
     crate::profile_scope!("EventLoopBuilder::build");
-    builder.build()
+    builder.build().unwrap()
```
rib, line 42: can we either change the api to return a `Result` or should we update corresponding public api docs to warn about a possible panic here?

crates/egui-winit/src/lib.rs
```diff
@@ -405,14 +387,20 @@ impl State {
...
+                    state.lshift_state() == Pressed || state.rshift_state() == Pressed;
+                self.egui_input.modifiers.mac_cmd = cfg!(target_os = "macos")
+                    && (state.lsuper_state() == Pressed || state.rsuper_state() == Pressed);
+
                 self.egui_input.modifiers.command = if cfg!(target_os = "macos") {
-                    state.logo()
+                    state.lsuper_state() == Pressed || state.rsuper_state() == Pressed
                 } else {
-                    state.ctrl()
+                    state.lcontrol_state() == Pressed || state.rcontrol_state() == Pressed
```
rib, line 403: Maybe this is really what's intended here but it does seem kinda odd to me that egui_input specifically exposes a "mac_cmd" modifier as the super key clamped to false if not macOS, and then again exposes the same modifier via `modifiers.command` except mapped to `ctrl` if not on macOS.

How are other platforms supposed to check the super key modifier?

Maybe `modifiers.mac_cmd` can be removed and there should just be `modifiers.command` that maps to the "super" key?

(not strictly related to this PR though - it just stands out)

crates/egui-winit/src/lib.rs
```diff
@@ -852,138 +860,125 @@ fn open_url_in_browser(_url: &str) {
...
         || (cfg!(target_os = "windows")
             && modifiers.shift
-            && keycode == winit::event::VirtualKeyCode::Delete)
+            && keycode == winit::keyboard::KeyCode::Delete)
 }
 
-fn is_copy_command(modifiers: egui::Modifiers, keycode: winit::event::VirtualKeyCode) -> bool {
-    (modifiers.command && keycode == winit::event::VirtualKeyCode::C)
+fn is_copy_command(modifiers: egui::Modifiers, keycode: winit::keyboard::KeyCode) -> bool {
+    (modifiers.command && keycode == winit::keyboard::KeyCode::KeyC)
```
rib, line 871: I think this should probably be `winit::keyboard::Key::Character("c".into())` instead of being based on keycodes, so that any user keymap is taken into account.

crates/eframe/src/native/run.rs
emilk: I'll change it to return errors

crates/egui-winit/src/lib.rs
emilk: The rational for the unified `command` key is so that you can write something that e.g. selects all with `cmd+A`, and it will use `ctrl+A` on Windows and `⌘+A` on Mac. Changing this behavior is way outside the scope of this PR.

crates/eframe/Cargo.toml
```diff
@@ -131,7 +131,7 @@ image = { version = "0.24", default-features = false, features = [
   "png",
 ] } # Needed for app icon
 raw-window-handle.workspace = true
-winit = { version = "0.28.1", default-features = false }
+winit = { version = "0.29.4", default-features = false, features = ["rwh_05"] }
```
emilk, line 134: why are we using raw-window-handle 0.5 instead of 0.6?

emilk:
`cargo run -p egui_demo_app` on my Mac results in a completely unresponsive application for me. I cannot click anything. The only interaction that is registered is window resize.

crates/egui-winit/src/lib.rs
rib: yeah I can see the utility of that - though it does still feel odd that there isn't a platform-agnostic modifier for the super/meta/windows/cmd key which is a standard modifier key in its own right. But yep, not related to this PR.

crates/egui-winit/src/lib.rs
emilk: Yeah, I think in general we should probably use `logical_key` instead of `physical_key`

emilk: I've created an issue to better track this: https://github.com/emilk/egui/issues/3653

crates/egui-winit/src/lib.rs
emilk: I added a note about it at https://github.com/emilk/egui/issues/3653

crates/eframe/src/native/run.rs
```diff
@@ -211,30 +210,43 @@ fn run_and_return(
     // we only apply this approach on Windows to minimize the affect.
     #[cfg(target_os = "windows")]
     {
-        event_loop.run_return(|_, _, control_flow| {
+        event_loop.run_on_demand(|_, _, control_flow| {
             control_flow.set_exit();
         });
     }
```
emilk, line 216: Can someone one Windows check if this hack is still needed, or if `cargo run -p serial_windows` works without it?

emilk (approved):
This looks good to me, but I'd love to get some additional testing

* [x] Mac
* [ ] X11
* [ ] Wayland
* [ ] Windows
   * Do we need the `target_os = "windows"` specific hacks?
* [ ] IME input

crates/eframe/Cargo.toml
fornwall: It might make sense to upgrade raw-window-handle as a separate PR, to limit scope (and do we need to await glutin updating to rwh 0.6 as well)?

fornwall:
On Mac: In `examples/multiple_viewports` the `Show deferred child viewport` does nothing on this branch (while it works on the master branch).

gerwin3:
Tested on X11 and found:
* in IME, Every keystroke results in two characters.
* Also in IME: Using Ctrl+key or Win+key still renders the character (it doesn't on master). For example Ctrl+C will type a C.

Apart from that everything looks okay (I think). `cargo run -p custom_window_frame` looks/works the same on this PR vs. master.

emilk:
> On Mac: In `examples/multiple_viewports` the `Show deferred child viewport` does nothing on this branch (while it works on the master branch).

Good find - it is now fixed!

emilk:
@gerwin3 thanks for testing IME! Can you please try to dig a little deeper and see if you can figure out the cause of the two issues? `RUST_LOG=winit=info,trace` may help, though it can become quite verbose

gerwin3:
Just checked and having all those issues in text input as well. let me see if I can debug the doubling issue.

gerwin3:
Okay at least the key doubling is easy to fix:

```diff
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -707,7 +707,9 @@ impl State {
             // On Mac we get here when the user presses Cmd-C (copy), ctrl-W, etc.
             // We need to ignore these characters that are side-effects of commands.
             let is_mac_cmd = cfg!(target_os = "macos") && self.egui_input.modifiers.mac_cmd;
-            if !is_mac_cmd {
+            // Also make sure the key is pressed (not released). On Linux, text might
+            // contain some data even when the key is released.
+            if pressed && !is_mac_cmd {
                 self.egui_input
                     .events
```

Not sure if there will be any regression on Windows or mac? Or if this has any bad interaction. Is there a situation where the key event is other than "Pressed" but text is produced?

extrawurst:
Do you guys also test this in wasm? I am asking for a friend aka bevy :)

emilk:
>  Is there a situation where the key event is other than "Pressed" but text is produced?

Maybe IME? 🤷

PingPongun:
I prepared patch that updates `resvg` to v0.36 (issue: #3652):
[update_resvg_0_36.patch](https://github.com/emilk/egui/files/13511578/update_resvg_0_36.patch)

I think it matches in this PR as:

- both `resvg` and `winit` (through `sctk-adwaita`) depend on `tiny-skia`, so updating them separately results in dependency duplication and `cargo-deny` fail.
- there were already some small changes related to svg (commits `b4e6246` & `7c89b67`) and deny.toml (`0dbd597`) in this PR

In example `images` ferris is scaled differently, but I guess that now it scales in expected way (takes all available space; before this patch it takes up to space that, was available at first render- it does not upscale).

<details>
<summary>Digression 1</summary>

This patch is minimal adaptation to new `resvg` api and small related simplification, however it should be considered to update loaders (currently if svg image initially was small and was scaled up it will be blurred, see #3501). As svg image can now scale over render size, problem will be more often seen now.

(currently `SvgLoader` theoretically should rerender for different sizes (but will result in memory leak in that case), but refreshing is stopped earlier in `DefaultTextureLoader`).

I have some ideas on updating loaders (it will fix issue with svg and enable reloading image if file has been changed):

- add two more variants to `ImagePoll`/`BytesPoll` (`Changing(old_data)` & `Changed(new_data)`)
- change order of loader tasks (currently is: 1. check if loader has image cached, 2. if not call lower level loader -> change to -> 1. call lower level loader 2. if lower level is not changed load cached)
- `SvgLoader`:

  - store `resvg::Tree` & `SizeHint` in `svg_loader::Entry`
  - remove `SizeHint` from `SvgLoader` hashmap key
  - one of the following:

    - rerender svg if image size has been changed,
    - render dynamically (first approach will equal second one if there are multiple images from the same file but in different scale),
    - rerender svg if new image size is larger than rendered one (I guess this will be most performant, but there may be some minor down-scalling artifacts)

As these will require changes that are completely unrelated to `winit`, I guess it will better fit into separate PR, after this PR is merged.

</details>

Digression 2: `check.sh` installs `cranky`, but does not install `typos` nor `cargo-deny` that are also needed

emilk:
@PingPongun thanks for the patch, but let's wait with the resvg update for a separate PR - this one is big enough as is 😓

YgorSouza:
I was testing to see if this branch fixes #3673, which it does, but on my machine it makes the main viewport on the demo app stop repainting whenever it is in the foreground and the extra viewport is open. The person who opened the issue did not observe the same behavior on their machine, which has a similar but different setup.

https://github.com/emilk/egui/assets/43298013/6fa47e11-931c-429e-b87b-61fe488049ff

Here is my setup:

```
OS: Fedora Linux 38 (Workstation Edition) x86_64 
Host: Laptop (12th Gen Intel Core) A6 
Kernel: 6.6.2-101.fc38.x86_64 
Resolution: 2256x1504 
DE: GNOME 44.6 
WM: Mutter 
WM Theme: Adwaita 
Theme: Adwaita-dark [GTK2/3] 
Icons: Adwaita [GTK2/3] 
CPU: 12th Gen Intel i7-1260P (16) @ 4.700GHz 
GPU: Intel Alder Lake-P GT2 [Iris Xe Graphics] 
```

YgorSouza:
Another problem I found: the modifier keys are ignored. Both the left and the right ones. They don't appear as down in the inspection window in the demo app, and they don't have any other effect either (aside from making text upper case, in the case of the shift key).

crates/eframe/Cargo.toml
MarijnS95: Glutin hasn't upgraded yet: https://github.com/rust-windowing/glutin/pull/1582

emilk:
According to @Wumpf this seems to work pretty well on Windows now.

emilk:
IME seems to work on my Mac.

Are there any more blockers to merging this?

Vickerinox:
Just tried it out on Linux Mint 21.2, everything seems to be working perfectly aside from window dragging, which causes the window to fly into the top left corner of the screen (see video) I've tested to make sure every letter on my keyboard works + modifiers, as well as the various demos in the demo app. Nothing obvious that's different aside form the previously mentioned bug. 
https://github.com/emilk/egui/assets/51887867/0a46b244-088c-4181-82cb-8c5ec2d593cf

emilk:
Thanks @Vickerinox - I think that is an acceptable level of regression. If anyone wants that feature to work they can make a PR :)

PR #3660, opened 2023-11-29
Fix cursor clipping in `TextEdit` inside a `ScrollArea`

### LINKED ISSUE #1531: ScrollArea scroll state modified by sibling textedit

juancampa:
* Closes #1531

### Before
Notice how the cursor hides after third enter and when the line is long.

https://github.com/user-attachments/assets/8e45736e-d6c7-4dc6-94d0-213188c199ff

### After
Cursor is always visible

https://github.com/user-attachments/assets/43200683-3524-471b-990a-eb7b49385fa9

 - `ScrollArea` now checks if there's a `scroll_target` in `begin`, if there is, it saves it because it's not from its children, then restore it in `end`.
 - `TextEdit` now allocates additional space if its galley grows during the frame. This is needed so that any surrounding `ScrollArea` can bring the cursor to view, otherwise the cursor lays outside the the `ScrollArea`'s `content_ui`.

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -499,6 +499,11 @@ struct Prepared {
 
     scrolling_enabled: bool,
     stick_to_end: Vec2b,
+
+    /// If there was a scroll target before the ScrollArea was added this frame, it's
+    /// not for us to handle so we save it and restore it after this ScrollArea is done.
+    saved_scroll_target: [Option<frame_state::ScrollTarget>; 2],
+
     animated: bool,
 }
 
@@ -693,6 +698,10 @@ impl ScrollArea {
             }
         }
 
+        let saved_scroll_target = content_ui
+            .ctx()
+            .frame_state_mut(|state| std::mem::take(&mut state.scroll_target));
+
         Prepared {
             id,
             state,
@@ -707,6 +716,7 @@ impl ScrollArea {
             viewport,
             scrolling_enabled,
             stick_to_end,
+            saved_scroll_target,
             animated,
         }
     }
@@ -820,6 +830,7 @@ impl Prepared {
             viewport: _,
             scrolling_enabled,
             stick_to_end,
+            saved_scroll_target,
             animated,
         } = self;
 
@@ -853,7 +864,7 @@ impl Prepared {
                     let (start, end) = (range.min, range.max);
                     let clip_start = clip_rect.min[d];
                     let clip_end = clip_rect.max[d];
-                    let mut spacing = ui.spacing().item_spacing[d];
+                    let mut spacing = content_ui.spacing().item_spacing[d];
 
                     let delta_update = if let Some(align) = align {
                         let center_factor = align.to_factor();
@@ -901,6 +912,15 @@ impl Prepared {
                 }
             }
         }
+
+        // Restore scroll target meant for ScrollAreas up the stack (if any)
+        ui.ctx().frame_state_mut(|state| {
+            for d in 0..2 {
+                if saved_scroll_target[d].is_some() {
+                    state.scroll_target[d] = saved_scroll_target[d].clone();
+                };
+            }
+        });
 
         let inner_rect = {
             // At this point this is the available size for the inner rect.
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -502,7 +502,7 @@ struct Prepared {
 
     /// If there was a scroll target before the ScrollArea was added this frame, it's
     /// not for us to handle so we save it and restore it after this ScrollArea is done.
-    saved_scroll_target: [Option<frame_state::ScrollTarget>; 2],
+    saved_scroll_target: [Option<pass_state::ScrollTarget>; 2],
 
     animated: bool,
 }
@@ -700,7 +700,7 @@ impl ScrollArea {
 
         let saved_scroll_target = content_ui
             .ctx()
-            .frame_state_mut(|state| std::mem::take(&mut state.scroll_target));
+            .pass_state_mut(|state| std::mem::take(&mut state.scroll_target));
 
         Prepared {
             id,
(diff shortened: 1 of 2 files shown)


juancampa:
The reason for this fix btw is because the temporary workaround #153 (cc @Barugon) has a problem. Notice that on the fourth `Enter` press, the scrollarea doesn't follow the cursor. That's because the ScrollArea doesn't set up a clip rect (because contents are small enough) and the text area doesn't know it's been clipped.

https://github.com/emilk/egui/assets/1410520/1d38827a-4110-4d3a-ac09-7cc45167af2a

crates/egui/src/containers/scroll_area.rs
```diff
@@ -751,6 +761,13 @@ impl Prepared {
             }
         }
 
+        // Restore scroll target (if any) meant for ScrollAreas up the stack
+        if saved_scroll_target[0].is_some() || saved_scroll_target[1].is_some() {
+            ui.ctx().frame_state_mut(|state| {
+                state.scroll_target = saved_scroll_target;
```
emilk, line 767: If either the horizontal or the vertical part of `saved_scroll_target` is set, overwrite both.
If neither is set, overwrite neither.

Without having thought deeply about this, this seems wrong to me.

Why not always write to `state.scroll_target`?
Or why not do the logic per dimension?

juancampa: I changed it to be per dimension since it's more readable. I was trying to prevent locking twice but probably not worth it

juancampa: We could also _always_ restore the saved scroll target, a few extra cycles but not worth worrying about. Let me know and I can change it.

crates/egui/src/containers/scroll_area.rs
```diff
@@ -752,6 +762,15 @@ impl Prepared {
             }
         }
 
+        // Restore scroll target (if any) meant for ScrollAreas up the stack
+        for d in 0..2 {
+            if saved_scroll_target[d].is_some() {
+                ui.ctx().frame_state_mut(|state| {
```
emilk, line 768: avoid locking twice

```suggestion
        ui.ctx().frame_state_mut(|state| {
            for d in 0..2 {
                if saved_scroll_target[d].is_some() {
```

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -671,8 +671,7 @@ impl<'t> TextEdit<'t> {
                     let primary_cursor_rect =
                         cursor_rect(galley_pos, &galley, &cursor_range.primary, row_height);
 
-                    let is_fully_visible = ui.clip_rect().contains_rect(rect); // TODO: remove this HACK workaround for https://github.com/emilk/egui/issues/1531
-                    if (response.changed || selection_changed) && !is_fully_visible {
+                    if response.changed || selection_changed {
```
emilk, line 674: Nice to get rid of this hack! Can we get rid of it in `crates/egui/src/text_selection/label_text_selection.rs` too?

juancampa:
@emilk I just reimplemented this on top of the latest `master`. Also posted two videos in the description showing the issue.

crates/egui/src/containers/scroll_area.rs
```diff
@@ -499,6 +499,11 @@ struct Prepared {
 
     scrolling_enabled: bool,
     stick_to_end: Vec2b,
+
+    /// If there was a scroll target before the ScrollArea was added this frame, it's
+    /// not for us to handle so we save it and restore it after this ScrollArea is done.
+    saved_scroll_target: [Option<pass_state::ScrollTarget>; 2],
```
lucasmerlin, line 505: There is also `FrameState::scroll_delta`, I think we should do the same with that?

lucasmerlin (approved):
Looks good to me!

crates/egui/src/containers/scroll_area.rs
juancampa: @lucasmerlin There's no `FrameState` anymore so I think this comment is not valid anymore?

juancampa (commented):
@emilk this should be ready to merge unless there's more feedback.

emilk:
thanks!

IaVashik:
This change breaks `TextEdit::singleline` when it's used outside of a `ScrollArea`. The added `ui.allocate_rect` call causes incorrect horizontal expansion because it allocates extra space based on the formatted text size, which is not appropriate for single-line text edits. This logic should be implemented in `ScrollArea` instead of `TextEdit`. `ScrollArea` should handle the cursor visibility, as it already manages its content's position and size. Imo, `TextEdit` shouldn't have to alter its size for `ScrollArea` to work correctly.

IaVashik:
Also, I haven't been able to reproduce this bug in the current version, *but only when I remove the code from lines 726-734 in `builder.rs`*.  Do we actually need this code at all?

PR #3671, opened 2023-12-01
Log all available wgpu adapters during startup 

emilk:
Useful when debugging, e.g. if the default choice is one of many, and perhaps the other ones would be better.

**EDIT**: Oh, `enumerate_adapters` doesn't work on web… will fix.

### DIFF
--- a/crates/egui-wgpu/src/lib.rs
+++ b/crates/egui-wgpu/src/lib.rs
@@ -69,6 +69,7 @@ impl RenderState {
     ) -> Result<Self, WgpuError> {
         crate::profile_scope!("RenderState::create"); // async yield give bad names using `profile_function`
 
+        #[cfg(not(target_arch = "wasm32"))]
         let adaptors: Vec<_> = instance.enumerate_adapters(wgpu::Backends::all()).collect();
 
         let adapter = {
@@ -81,8 +82,7 @@ impl RenderState {
                 })
                 .await
                 .ok_or_else(|| {
-                    let adaptors: Vec<_> =
-                        instance.enumerate_adapters(wgpu::Backends::all()).collect();
+                    #[cfg(not(target_arch = "wasm32"))]
                     if adaptors.is_empty() {
                         log::info!("No wgpu adaptors found");
                     } else if adaptors.len() == 1 {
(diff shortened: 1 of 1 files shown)


crates/egui-wgpu/src/lib.rs
```diff
@@ -69,6 +69,9 @@ impl RenderState {
     ) -> Result<Self, WgpuError> {
         crate::profile_scope!("RenderState::create"); // async yield give bad names using `profile_function`
 
+        #[cfg(not(target_arch = "wasm32"))]
+        let adaptors: Vec<_> = instance.enumerate_adapters(wgpu::Backends::all()).collect();
```
Wumpf, line 73:
```suggestion
        let adapters: Vec<_> = instance.enumerate_adapters(wgpu::Backends::all()).collect();
```

😄

crates/egui-wgpu/src/lib.rs
```diff
@@ -106,6 +151,24 @@ impl RenderState {
     }
 }
 
+#[cfg(not(target_arch = "wasm32"))]
+fn describe_adaptors(adaptors: &[wgpu::Adapter]) -> String {
```
Wumpf, line 155:
```suggestion
fn describe_adapters(adapters: &[wgpu::Adapter]) -> String {
```
I like your consistency though :)

crates/egui-wgpu/src/lib.rs
```diff
@@ -224,6 +294,40 @@ pub fn depth_format_from_bits(depth_buffer: u8, stencil_buffer: u8) -> Option<wg
 
 // ---------------------------------------------------------------------------
 
+/// A human-readable summary about an adapter
+pub fn adapter_info_summary(info: &wgpu::AdapterInfo) -> String {
+    let wgpu::AdapterInfo {
+        name,
+        vendor: _, // skip integer id
+        device: _, // skip integer id
```
Wumpf, line 302: vendor and device id as hex can be super useful though! For instance wgpu uses that internally to enable workarounds.
Bonus points if you use this (pub!!) table https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-hal/src/auxil/mod.rs#L7 to attempt stringification!

Wumpf (approved):

Issue #3674, opened 2023-12-04
eframe's Event::Resumed on android panics since #3172


m-hugo:
Hiding and resuming any eframe app on android since #3172 causes it to crash, logging (on master) `thread panicked at .cargo/git/checkouts/egui-5e4507fa4153be06/84a6d6f/crates/eframe/src/native/wgpu_integration.rs:392:54: no entry found for key`
https://github.com/emilk/egui/blob/84a6d6f2ab03c36e81c0ba5dc20976e7bdf78951/crates/eframe/src/native/wgpu_integration.rs#L392
the offending line `is running.shared.borrow().viewports[&ViewportId::ROOT]`
the commit after #3172 logs the same line but in eframe/src/native/run.rs
the commit before works fine
digging in #3172's commits, it was broken as early as 7caa042 ` .get(&ViewportId::ROOT) `:  `called Option::unwrap() on a None value`

**Expected behavior**
get back to the app normally like before #3172 

**Smartphone**
 - OS: Android 13

m-hugo:
almost the same with glow but with a black screen and while I got a panic `.cargo/git/checkouts/egui-5e4507fa4153be06/7caa042/crates/eframe/src/native/run.rs:2585:30: called `Option::unwrap()` on a `None` value` on a pre-0.24 commit
current master blackscreens without panics

emilk:
Please see if https://github.com/emilk/egui/pull/3676 helps

m-hugo:
follow-up at #3861

PR #3676, opened 2023-12-04
Fix crash on Android

emilk:
* Closes https://github.com/emilk/egui/issues/3674

Please try it @rednaz1337 @m-hugo


m-hugo:
exactly the same .cargo/git/checkouts/egui-5e4507fa4153be06/a474724/crates/eframe/src/native/wgpu_integration.rs:399:69: called `Option::unwrap()` on a `None` value

m-hugo:
if you want to add some logging, just do some `println!`'s

emilk:
Oh right, the main viewport is removed…

I don't own an android device, so it would be better if someone else worked on this.

emilk:
Converting to Draft - If somebody is interested in fixing Android, just fork this and make a new PR

PR #3698, opened 2023-12-11
Invalidate font atlas on any change to `pixels_per_point`, not matter how small

StarStarJ:
Apparently the font implementation uses a distance check to decide if the font(or whatever) need recalculations, after dpi changed:
https://github.com/emilk/egui/blob/8d4de866d4da1835b31e85f9068a5257e6ccbccb/crates/epaint/src/text/fonts.rs#L381-L382

This leads to warnings when the pixel_per_point diff is very low and spams the log. (<- this happens for me if i resize my window on kwin, e.g. maximize it)

(I don't want to debate if the float difference generally makes sense, so if you want to rework that instead just close this pr)

### DIFF
--- a/crates/epaint/src/text/fonts.rs
+++ b/crates/epaint/src/text/fonts.rs
@@ -396,8 +396,7 @@ impl Fonts {
     pub fn begin_frame(&self, pixels_per_point: f32, max_texture_side: usize) {
         let mut fonts_and_cache = self.0.lock();
 
-        let pixels_per_point_changed =
-            (fonts_and_cache.fonts.pixels_per_point - pixels_per_point).abs() > 1e-3;
+        let pixels_per_point_changed = fonts_and_cache.fonts.pixels_per_point != pixels_per_point;
         let max_texture_side_changed = fonts_and_cache.fonts.max_texture_side != max_texture_side;
         let font_atlas_almost_full = fonts_and_cache.fonts.atlas.lock().fill_ratio() > 0.8;
         let needs_recreate =


emilk:
This seems to be fixing a symptom rather than the problem. Have you investigated why the float difference happen at all? I don't hit this assert.

StarStarJ:
I don't know if i want to dig into the details of winit & x11, but this value is reported by winit, and it's definitely changing by this small amount.

But besides that, if for whatever reason the DPI changes by a very small amount, what does egui do to handle it?

my code snipset shows that it ignores small differences. the warning that I i fixed in this pr does not. So these two code segments behave different on the same logic, or not?

emilk:
But where does the change come from? Are you storing a `Galley` between frames (you shouldn't)?

If you create the galley in one frame, and then tesslleate, then I can't see how the `pixels_per_point` could change.

Have you tried latest `master`?

StarStarJ:
> But where does the change come from? Are you storing a `Galley` between frames (you shouldn't)?

Well the `Galley` comes from from the font:
https://github.com/emilk/egui/blob/327f5994079181f8ef2e8918722e6eebcbc676e2/crates/epaint/src/text/fonts.rs#L563-L567

I honestly dunno how exactly to debug if these `Galley` are stored between frames, but a `GalleyCache` sounds like the implementation does that?

I have to say I don't have the insight into egui, the docs for the Galley also only say:
"Needs to be recreated if the underlying font atlas texture changes, which happens under the following conditions: [...]"

(https://docs.rs/egui/latest/egui/struct.Galley.html)

I am not manually recreating my fonts if `pixels_per_point` change, but the doc sounds like it's doing that internally anyway in [crate::text::Fonts::begin_frame](https://docs.rs/egui/latest/egui/text/struct.Fonts.html#method.begin_frame)


Then bcs of:
https://github.com/emilk/egui/blob/8d4de866d4da1835b31e85f9068a5257e6ccbccb/crates/epaint/src/text/fonts.rs#L381-L382

The `pixels_per_point` change is ignored (because of the floating point difference)
=> `galley.pixels_per_point`

but the line in this pr compares it again `self.pixels_per_point` which **IS** changed
(galley.pixels_per_point: `1.16666663` vs. self.pixels_per_point: `1.16666675` (so very very small => galley is not updated in https://github.com/emilk/egui/blob/8d4de866d4da1835b31e85f9068a5257e6ccbccb/crates/epaint/src/text/fonts.rs#L381-L382

but the warning is still displayed
)

I'm indeed not using latest master, if you say `Galley` is never cached between frames, then it's indeed a bug in my application. I am using custom fonts, but i am not sure which font is triggering this issue.

emilk:
Does

```diff
-        let pixels_per_point_changed =
-            (fonts_and_cache.fonts.pixels_per_point - pixels_per_point).abs() > 1e-3;
+        let pixels_per_point_changed = fonts_and_cache.fonts.pixels_per_point != pixels_per_point;
```

solve your problem?

I really don't see the point of the `1e-3` epsilon there

StarStarJ:
Yes that works just as good :+1:

emilk (approved):

Issue #3624, opened 2023-11-24
Allow holding shift on ScrollAreas with horizontal scrolling disabled


doonv:
**Describe the bug**
Pressing/Tapping <kbd>Shift</kbd> after a `ScrollArea::new([false, true])` (Vertical Scrolling Only) was opened causes scrolling via the scroll wheel to no longer work.

**To Reproduce**
1. Create a `ScrollArea` with only vertical scrolling.
2. Scrolling works
3. Press <kbd>Shift</kbd>
4. Scrolling doesn't work

**Expected behavior**
The `ScrollArea` always scrolls vertically regardless of if <kbd>Shift</kbd> was pressed or not.

**Desktop (please complete the following information):**
 - OS: Kubuntu 23.10
 - Browser: Not using a browser
 - Version: `bevy_egui = "0.23.0"`, `egui = { version = "0.23.0", default-features = false, features = ["bytemuck"] }`

**Additional context**
I am using `egui` in `bevy` using `bevy_egui`.

doonv:
This issue is most likely caused by <https://github.com/emilk/egui/pull/663> as that is when horizontal scrolling was added.

emilk:
is this bug about not being able to scrol while _holding down_ shift, or not being able to scroll after tapping shift once and then releasing it?

emilk:
I can't reproduce. Sounds like a problem of sticky modifier keys on your system.

Does egui report the SHIFT modifier to be pressed even after you released it? (`ctx.input(|i| dbg!(i.modifiers));`)

Please take a look at `crates/egui-winit/src/lib.rs` and see how `modifiers.shift` is updated - perhaps `WindowEvent::ModifiersChanged` is not working properly?

doonv:
It might be a `bevy_egui` issue. I'll try out doing it with eframe (or whatever you use to make native egui applications)

doonv:
Yeah I can't reproduce the bug when using `cargo run -p egui_demo_app`.

I also did `ui.input(|i| i.modifiers.shift)` and, as I suspected it continued being `true` after i realised the shift key. So I'm closing this and making a `bevy_egui` issue instead.

doonv:
I'm reopening this for a different reason. Because even if <kbd>Shift</kbd> is being held, it shouldnt affect the scrolling if horizontal scrolling is disabled.

PR #3704, opened 2023-12-14
Scroll area only horizontal true mouse wheel event

12089897411:
Add setting pulley scrolling to ScrollArea. The default scrolling is vertical scrolling, which is set to horizontal scrolling through the on_mouse_wheel_horizontal function.

for example:
 eframe::egui::scroll_area::ScrollArea::new([true, false]).on_mouse_wheel_horizontal(true).show(ui,add_contents(ui));


untbu:
This would also be done by my PR, which detects if a ScrollArea is only scrollable in one direction and allows scrolling in that direction regardless of pressing shift: https://github.com/emilk/egui/pull/3710
This has the advantage that a normal vertical scrollbar can be scrolled even if shift is pressed (issue: https://github.com/emilk/egui/issues/3624).

PR #3710, opened 2023-12-17
[ScrollArea]: Add option to always scroll the only enabled direction

untbu:
Add option to use the scroll input from both directions to scroll the enabled direction if scrolling is enabled for only one direction. This can be used to allow horizontal scrolling without pressing shift if vertical scrolling is disabled.

Closes <https://github.com/emilk/egui/issues/3624>.

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -161,6 +161,9 @@ pub struct ScrollArea {
     scrolling_enabled: bool,
     drag_to_scroll: bool,
 
+    /// If true and scrolling is enabled for only one direction, allow horizontal scrolling without pressing shift
+    always_scroll_the_only_direction: bool,
+
     /// If true for vertical or horizontal the scroll wheel will stick to the
     /// end position until user manually changes position. It will become true
     /// again once scroll handle makes contact with end.
@@ -207,6 +210,7 @@ impl ScrollArea {
             offset_y: None,
             scrolling_enabled: true,
             drag_to_scroll: true,
+            always_scroll_the_only_direction: true,
             stick_to_end: Vec2b::FALSE,
         }
     }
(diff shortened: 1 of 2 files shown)


emilk:
Do you really want to hav this setting for eah `ScrollArea`? Seems like this would make more sense as a global option in `Style`

crates/egui/src/containers/scroll_area.rs
```diff
@@ -779,15 +779,33 @@ impl Prepared {
         if scrolling_enabled && is_hovering_outer_rect {
             for d in 0..2 {
                 if scroll_enabled[d] {
-                    let scroll_delta = ui.ctx().frame_state(|fs| fs.scroll_delta);
+                    let scroll_delta = ui.ctx().frame_state(|fs| {
+                        if ui.style().always_scroll_the_only_direction
+                            && scroll_enabled[0] != scroll_enabled[1]
```
emilk, line 784: `ui.style().always_scroll_the_only_direction && scroll_enabled[0] != scroll_enabled[1]` is repeated down below. It would be nice to put it in a `bool` and reuse it to make the logic more clear; though I don't know what name to give it…

crates/egui/src/style.rs
```diff
@@ -212,6 +212,9 @@ pub struct Style {
     ///
     /// This only affects a few egui widgets.
     pub explanation_tooltips: bool,
+
+    /// If true and scrolling is enabled for only one direction, allow horizontal scrolling without pressing shift
+    pub always_scroll_the_only_direction: bool,
```
emilk, line 217: I'm not sure we we should have this on by default as it will break existing apps in subtle way: a hscroll inside a vscroll (or the other way around) will no longer work as expected (the inner scroll area will steal all input)

emilk (approved):

PR #3714, opened 2023-12-18
egui-wgpu: reduce input latency by defaulting to double-buffering

emilk:
This makes a huge difference in perceived latency on my 60 Hz monitor on macOS on a native build.

There is no difference on web though afaict.

For me this is a much better default. Tearing is mildly annoying in some circumstances, but input latency makes an application feel sluggish all the time.

### DIFF
--- a/crates/egui-wgpu/src/lib.rs
+++ b/crates/egui-wgpu/src/lib.rs
@@ -238,8 +238,8 @@ impl Default for WgpuConfiguration {
                 }
             }),
 
-            present_mode: wgpu::PresentMode::AutoVsync,
-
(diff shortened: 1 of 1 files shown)


emilk:
Sadly this will burn _a lot of cpu_, since we will repaint as fast as we can (e.g. at 500Hz), ignoring the monitor refresh rate.

Wumpf:
* Blocked by https://github.com/gfx-rs/wgpu/pull/4899

torokati44:
With the PR mentioned above now merged, is this unblocked?

emilk:
Yes! I'll make a new PR though since this the API for this changed

PR #3587, opened 2023-11-20
Plot auto-bounds API improvement (part 1/2): clean-up

abey79:
Part 1 of 2 of adding a better API for egui_plot's auto-bounds feature.

In this PR:
- change the `Plot` builder struct field to `default_auto_bounds` (was `auto_bounds`)
- change the  `Plot` state field to `auto_bounds` (was `bounds_modified`)
- minor improvements to `Vec2b`

PR #3586, opened 2023-11-20
Plot auto-bounds API improvement (part 2/2): add API

abey79:
Part 2 of 2 of adding a better API for egui_plot's auto-bounds feature.

In this PR: add `auto_bounds()`/`set_auto_bounds()` APIs to PlotUI

PR #3722, opened 2023-12-19
Fix plot auto-bounds unset by default

### LINKED ISSUE #3712: egui_plot auto-bounds broken
Repro: try the plots demo at https://www.egui.rs/ - you need to double-click most of the plots for them to reset to a nice default.

Reverting https://github.com/emilk/egui/pull/3586 fixes the problem

@abey79

abey79:
These PR recently cleaned up the code around auto-bounds, but introduced an involuntary change whereby auto-bounds would not be enabled by default. All plots would default to being not properly centred as a result.

- #3587
- #3586

This PR changes the default back to enabled. It also deprecates `auto_bounds_x()` and `auto_bounds_y()`, which could only enable auto-bounds (which is not very useful as auto-bounds were, and now are again, enabled by default). A new `auto_bounds()` API can now be sued to disable auto-bounds if needed.

Fixes #3712 
Fixes https://github.com/rerun-io/rerun/issues/4503

### DIFF
--- a/crates/egui_plot/src/lib.rs
+++ b/crates/egui_plot/src/lib.rs
@@ -225,7 +225,7 @@ impl Plot {
             allow_scroll: true,
             allow_double_click_reset: true,
             allow_boxed_zoom: true,
-            default_auto_bounds: false.into(),
+            default_auto_bounds: true.into(),
             min_auto_bounds: PlotBounds::NOTHING,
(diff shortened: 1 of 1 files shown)


Wumpf (approved):
nice! That's surprisingly straight-forward 😄

PR #3723, opened 2023-12-19
Fix: `Grid` now follows `style.visuals.striped` setting if not explicitly overwritten

Wcubed:
The docs of [`Visuals.striped`](https://docs.rs/egui/latest/egui/style/struct.Visuals.html#structfield.striped) mention that they control the default striping of Grid. However, this seems to be broken.

This pr makes the Grid follow the striped setting if it doesn't yet have a row coloring set.


emilk (approved):
Good catch! Apparently it was only used for `egui_extras::Table`

PR #3724, opened 2023-12-20
✋ `egui_winit`: Allow getting the `clipboard` and `allow_ime` state

tosti007:
The `egui_winit::State` contains both the `clipboard` and `allow_ime`, but these are both private fields. In our implementation we use `egui_winit` to do most of the lifting, but it would come really in handy if we could access these values.

Instead of making the fields `pub` we opted for separate get/set functions, that way calling side can't freely replace any of the values fully and it's more in line with the style of the rest of the codebase.


crates/egui-winit/src/lib.rs
```diff
@@ -172,6 +172,16 @@ impl State {
         self.egui_input.max_texture_side = Some(max_texture_side);
     }
 
+    /// Returns the used clipboard.
+    pub fn clipboard(&mut self) -> &mut clipboard::Clipboard {
+        &mut self.clipboard
+    }
+
+    /// Returns whether the window should get IME events.
+    pub fn allow_ime(&self) -> bool {
```
MarijnS95, line 181: In my variant of the PR, this was mut because we need to _update_ this field :wink:

crates/egui-winit/src/lib.rs
```diff
@@ -172,6 +172,21 @@ impl State {
...
+        self.allow_ime
+    }
+
+    /// Returns whether the window should get IME events.
+    pub fn allow_ime_mut(&self) -> &mut bool {
```
MarijnS95, line 186:
```suggestion
    pub fn allow_ime_mut(&mut self) -> &mut bool {
```

crates/egui-winit/src/lib.rs
```diff
@@ -172,6 +172,21 @@ impl State {
         self.egui_input.max_texture_side = Some(max_texture_side);
     }
 
+    /// Returns the used clipboard.
+    pub fn clipboard(&mut self) -> &mut clipboard::Clipboard {
+        &mut self.clipboard
+    }
+
+    /// Returns whether the window should get IME events.
```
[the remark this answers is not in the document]
MarijnS95:
```suggestion
    /// Returns [`false`] or the last value that [`Window::set_ime_allowed()`] was called with, used for debouncing.
```

crates/egui-winit/src/lib.rs
```diff
@@ -172,6 +172,21 @@ impl State {
...
+    pub fn allow_ime(&self) -> bool {
+        self.allow_ime
+    }
+
+    /// Returns whether the window should get IME events.
```
MarijnS95, line 185:
```suggestion
    /// Returns [`false`] or the last value that [`Window::set_ime_allowed()`] was called with, used for debouncing.
```

tedsteen:
>Instead of making the fields pub I opted for getter functions, such that the calling side is discouraged to freely alter the State's internal data.
but the return type is `&mut` allowing for mutation of the internal data?

MarijnS95:
`allow_ime` needs to be `mut` to be able to tell `State` that we've already called `set_ime_enabled()`. This wasn't/isn't conveyed in the original message but is what we need these changes for.

tedsteen:
> `allow_ime` needs to be `mut` to be able to tell `State` that we've already called `set_ime_enabled()`. This wasn't/isn't conveyed in the original message but is what we need these changes for.

It was a comment on the fact that the fields where private instead of public. I don't know what the convention is in this repo but I would personally make the fields public instead of writing methods that return a &mut

MarijnS95:
@tedsteen that comment was written with immutability in mind, but even after correcting that we need _mutable_ access to this field, this matches the `fn egui_input()`/`fn egui_input_mut()` pattern. Those functions are already available on `State`.

emilk:
The difference is that `egui_input_mut` returns a mutable reference to a big struct. For a simple bool, a `set_` function feels a lot more idiomatic imho.

MarijnS95:
> The difference is that `egui_input_mut` returns a mutable reference to a big struct. For a simple bool, a `set_` function feels a lot more idiomatic imho.

Wasn't suggesting that a `_mut()` API is better than `_set()` for `bool`s (though it allows us to get a mutable reference exactly once), just that the OP comment no longer holds. We might also make _both_ (`allow_ime` _and_ `egui_input`) fields `pub` and do away with the API though.

crates/egui-winit/src/lib.rs
```diff
@@ -172,6 +172,21 @@ impl State {
...
+
+    /// Returns [`false`] or the last value that [`Window::set_ime_allowed()`] was called with, used for debouncing.
+    pub fn allow_ime(&self) -> bool {
+        self.allow_ime
+    }
+
+    /// Returns [`false`] or the last value that [`Window::set_ime_allowed()`] was called with, used for debouncing.
+    pub fn allow_ime_mut(&mut self) -> &mut bool {
+        &mut self.allow_ime
+    }
```
MarijnS95, line 188:
```suggestion
    /// Set the last value that [`Window::set_ime_allowed()`] was called with.
    pub fn set_allow_ime(&mut self, allow: bool) {
        self.allow_ime = allow;
    }
```

MarijnS95:
Ouch, looks like this missed out on the `0.25` release

tosti007:
> The difference is that egui_input_mut returns a mutable reference to a big struct. For a simple bool, a set_ function feels a lot more idiomatic imho.

I opted to take the same approach regarding the clipboard in the last commit, by making a `clipboard` and `set_clipboard` function. This way the backing implementation is also free to change if needed in the future.

> Ouch, looks like this missed out on the 0.25 release

Sorry, this had slipped my mind after the Christmas holiday D:

crates/egui-winit/src/lib.rs
```diff
@@ -172,6 +172,26 @@ impl State {
         self.egui_input.max_texture_side = Some(max_texture_side);
     }
 
+    /// Fetches text from the clipboard and returns it.
+    pub fn clipboard_text(&mut self) -> Option<String> {
```
MarijnS95, line 176: Should we document why this is mutable?

MarijnS95 (approved):

MarijnS95:
@emilk @tedsteen any further feedback here?

emilk (approved):

PR #3725, opened 2023-12-20
Add `Key::Cut`, `Key::Copy`, `Key::Paste`

MarijnS95:
The comment added in commit 8a0bc97e ("[egui_glium] Fix paste") seems to assume that `winit` "should have translated" common "paste" keyboard combos to a `Cut`/`Copy`/`Paste` "KeyCode", but completely glossed over the fact that this `KeyCode` (now also `NamedKey`) maps to a special key dedicated to this purpose found on some keyboards and OSes.  Make sure that this key is still handled in addition to the combo that is detected.

---

Note that this PR does not compile as it is (and I have hence not tested this nor even ran into this limitation), just noticed this inconsistency while failing to understand a code comment.  We'd have to decide if the variants should be added to `egui::Key` or if these helper functions need to take `winit` keys (`ScanCode` or `NamedKey`) directly?

I should have an old keyboard with a physical paste key in a drawer somewhere.  And on Android there are special copy/paste events that can be sent by a virtual keyboard or the debug shell, so that this can be properly tested before it is merged.
(Except that the current `clipboard` implementation is not supported on Android)

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -686,8 +686,6 @@ impl State {
 
         if let Some(logical_key) = logical_key {
             if pressed {
-                // KeyCode::Paste etc in winit are broken/untrustworthy,
-                // so we detect these things manually:
                 if is_cut_command(self.egui_input.modifiers, logical_key) {
                     self.egui_input.events.push(egui::Event::Cut);
                     return true;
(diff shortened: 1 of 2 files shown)


emilk:
I think the proposed solution makes sense: we translate the special keys to `egui::Event`, but also pass them on to the user should they want to make use of them

MarijnS95:
Okay, added the necessary `Key` variants and filled them in inside the relevant `match` statements.

emilk (approved):

PR #3727, opened 2023-12-22
Replace a special `Color32::PLACEHOLDER` with widget fallback color

emilk:
This introduces a special `Color32::PLACEHOLDER` which, during text painting, will be replaced with `TextShape::fallback_color`.

The fallback color is mandatory to set in all text painting. Usually this comes from the current visual style.

This lets users color only parts of a `WidgetText` (using e.g. a `LayoutJob` or a `Galley`), where the uncolored parts (using `Color32::PLACEHOLDER`) will be replaced by a default widget color (e.g. blue for a hyperlink).

For instance, you can color the `⚠️`-emoji red in a piece of text red and leave the rest of the text uncolored. The color of the rest of the text will then depend on wether or not you put that text in a label, a button, or a hyperlink.

Overall this simplifies a lot of complexity in the code but comes with a few breaking changes:

* `TextShape::new`, `Shape::galley`, and `Painter::galley` now take a fallback color by argument
* `Shape::galley_with_color` has been deprecated (use `Shape::galley` instead)
* `Painter::galley_with_color` has been deprecated (use `Painter::galley` instead)
* `WidgetTextGalley` is gone (use `Arc<Galley>` instead)
* `WidgetTextJob` is gone (use `LayoutJob` instead)
* `RichText::into_text_job` has been replaced with `RichText::into_layout_job`
* `WidgetText::into_text_job` has been replaced with `WidgetText::into_layout_job`

### DIFF
--- a/crates/ecolor/src/color32.rs
+++ b/crates/ecolor/src/color32.rs
@@ -7,6 +7,8 @@ use crate::{gamma_u8_from_linear_f32, linear_f32_from_gamma_u8, linear_f32_from_
 ///
 /// Internally this uses 0-255 gamma space `sRGBA` color with premultiplied alpha.
 /// Alpha channel is in linear space.
+///
+/// The special value of alpha=0 means the color is to be treated as an additive color.
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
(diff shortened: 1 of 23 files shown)


crates/ecolor/src/color32.rs
```diff
@@ -61,7 +63,16 @@ impl Color32 {
...
-    pub const TEMPORARY_COLOR: Color32 = Color32::from_rgb(64, 254, 0);
+    ///
+    /// This is an invalid color, in that it does not correspond to a valid multiplied color,
+    /// nor to an additive color.
+    ///
+    /// This is used as a special color key,
+    /// i.e. often taken to mean "no color".
+    pub const PLACEHOLDER: Color32 = Color32::from_rgba_premultiplied(64, 254, 0, 128);
+
+    #[deprecated = "Renmaed to PLACEHOLDER"]
```
Mingun, line 74: Misprint: `Renamed`. Strange, why wasn't catched by CI

PR #3733, opened 2023-12-24
Fix: Let `accesskit` process window events

DataTriny:
Since accesskit_winit version 0.15.0, it is necessary to call `Adapter::process_events` to let AccessKit fire up window focus events on Unix and macOS. Furthermore this has always been needed on Unix (X11 only) to let AccessKit know about the window bounds, which is needed to perform hit-testing on nodes.


emilk (approved):

emilk:
Thanks for the bug fix!

Issue #3742, opened 2023-12-28
Help Wanted


emilk:
Are you looking to help out on the egui project? Then this is a god place to start!

There are a few things you can help out with:
* Answer questions on [Discussions](https://github.com/emilk/egui/discussions)
* Contribute code
* Review & discuss [existing issues](https://github.com/emilk/egui/issues)
* Review [Pull Requests](https://github.com/emilk/egui/pulls)

First, make sure you read [`CONTRIBUTING.md`](https://github.com/emilk/egui/blob/master/CONTRIBUTING.md).

## Contributing code
You can browse [existing issues](https://github.com/emilk/egui/issues). A lot of them are unfortunately uncategorized, but I've tried adding "good first issue" and "help wanted" on some of them, but of course, I actually want help with all of them 😆

* [Issues with the "good first issue" label](https://github.com/emilk/egui/issues?q=is:open+is:issue+label:%22good+first+issue%22)
* [Issues with the "help wanted" label](https://github.com/emilk/egui/issues?q=is:open+is:issue+label:%22help+wanted%22)

Misc cherry-picks issues in no particular order:

* https://github.com/emilk/egui/issues/3653
* https://github.com/emilk/egui/issues/3668
* https://github.com/emilk/egui/issues/686
* https://github.com/emilk/egui/issues/3462
* [Windows and Linux specific `eframe/winit` integration issues for showing a custom window frame](https://github.com/emilk/egui/issues?q=is:issue+is:open+custom_window_frame+)

samitbasu:
Hi @emilk - thanks for making UI's fun again!  On the PR reviews, a couple of questions:

- Can anyone review PRs?  Or do they need to belong to the project somehow?
- What happens with PRs that contain API breaking changes (e.g., something that would cause `egui_extras` to break its API for example.

Also, would it help you if more functionality was broken out of the `egui` repo (like `egui_plot`)?  Would that allowed for a more distributed support model?

emilk:
Great questions!

Anyone is welcome to help review PRs, in order to provide feedback and help the PR author. Of course, someone with merge rights will still have the final say (most likely me), but the more eyes on a PR, the better it usually ends up!

Breaking changes are usually fine, but should be documented in the PR description. Of course, if they _can_ be avoided without creating to much hassle, that's often better. Also, using `#[deprecated]` on removed stuff is encouraged.

[...]

PR #3753, opened 2023-12-30
Bump `winit` to `0.29.7`

dbuch:
Compiled changelog below:

 - On X11, fix Xft.dpi reload during runtime.
 - On X11, fix window minimize.

 - On Web, fix context menu not being disabled by with_prevent_default(true).
 - On Wayland, fix WindowEvent::Destroyed not being delivered after destroying window.
 - Fix EventLoopExtRunOnDemand::run_on_demand not working for consequent invocation

 - On macOS, remove spurious error logging when handling Fn.
 - On X11, fix an issue where floating point data from the server is
 - misinterpreted during a drag and drop operation.
 - On X11, fix a bug where focusing the window would panic.
 - On macOS, fix refresh_rate_millihertz.
 - On Wayland, disable Client Side Decorations when wl_subcompositor is not supported.
 - On X11, fix Xft.dpi detection from Xresources.
 - On Windows, fix consecutive calls to window.set_fullscreen(Some(Fullscreen::Borderless(None))) resulting in losing previous window state when eventually exiting fullscreen using window.set_fullscreen(None).
 - On Wayland, fix resize being sent on focus change.
 - On Windows, fix set_ime_cursor_area.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -211,7 +211,7 @@ dependencies = [
  "parking_lot",
  "thiserror",
  "winapi",
- "x11rb",
+ "x11rb 0.12.0",
 ]
 
 [[package]]
@@ -1705,6 +1705,16 @@ checksum = "bb65d4ba3173c56a500b555b532f72c42e8d1fe64962b518897f8959fae2c177"
 dependencies = [
  "libc",
  "winapi",
+]
+
+[[package]]
+name = "gethostname"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0176e0459c2e4a1fe232f984bca6890e681076abb9934f6cea7c326f3fc47818"
+dependencies = [
+ "libc",
+ "windows-targets 0.48.5",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


dbuch:
@emilk I've marked this ready. I might disclose the motivation of this, now that i initially forgot.
I'd like to contribute to some points in https://github.com/emilk/egui/issues/3742 in particular linux issues.

emilk:
I appreciate you wanting to help out with Linux!

I think bumping the winit version for this repository makes sense, i.e. updating `Cargo.lock`, but by also updating the versions in the `Cargo.toml`:s you are effectively forcing downstream users to update to. This means for instance that if a bug is found in winit 0.29.7 then users of `eframe` _cannot_ downgrade to `0.29.6`.

Usually we should only set a minimal patch version in a `Cargo.toml` if we absolutely depend on that version.

Therefore, I suggest reverting the changes to the `Cargo.toml` files and keep just the changes to `Cargo.lock`.

(tip: in case you don't know, you can update a dependency in a `Cargo.lock` file using `cargo update -p winit`)

emilk (approved):

Issue #3761, opened 2024-01-01
Resize window with custom_window_frame


dbuch:
**Is your feature request related to a problem? Please describe.**
There isn't code example how to resize a `custom_window_frame`

**Describe the solution you'd like**
A simple implementation to show how it could be done.

PR #3762, opened 2024-01-01
Implement `custom_window_frame` resize support example

dbuch:
It includes a convenience helper `Into` to map `ResizeDirection` to `CursorIcon` in Egui crate.

This is tested on linux with glow and wgpu.

Glow has a render glich best show by a video.
[Glow Screencast](https://github.com/emilk/egui/assets/1972049/8a0c80f8-8d8f-4fe1-870e-ad07d4d724da)
Where wgpu works fine
[Wgpu Screencast](https://github.com/emilk/egui/assets/1972049/e406c5cd-eb3b-4d9d-bb50-fa0527c0efa8)

Closes https://github.com/emilk/egui/issues/3761

### DIFF
--- a/examples/custom_window_frame/src/main.rs
+++ b/examples/custom_window_frame/src/main.rs
@@ -53,6 +53,8 @@ fn custom_window_frame(ctx: &egui::Context, title: &str, add_contents: impl FnOn
     };
 
     CentralPanel::default().frame(panel_frame).show(ctx, |ui| {
+        handle_resize(ui);
+
         let app_rect = ui.max_rect();
 
         let title_bar_height = 32.0;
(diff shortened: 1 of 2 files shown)


emilk:
On macOS, `custom_window_frame` is already resizable on `master`.

I suspect adding `.with_resizable(true)` to the `ViewportBuilder` is not enough on Linux?

examples/custom_window_frame/src/main.rs
```diff
@@ -75,6 +77,48 @@ fn custom_window_frame(ctx: &egui::Context, title: &str, add_contents: impl FnOn
...
+        (false, false, false, true) => Some(egui::ResizeDirection::South),
+        _ => None,
+    };
+
+    let Some(resize_direction) = possible_resize_direction else {
+        return false;
+    };
+
+    ui.output_mut(|o| o.cursor_icon = resize_direction.into());
+    if ui.input(|i| i.pointer.primary_pressed()) {
```
Chaojimengnan, line 113: We should stop send resize command when window is maximize, otherwise there will be a strange bug

Chaojimengnan, line 113: We should stop send resize command when window is maximize, otherwise there will be a strange bug

Issue #3766, opened 2024-01-04
IME: TextEdit widget cannot support CompositionEnd only input character


FrankLeeC:
**Describe the bug**
Can not input CompositionEnd character directly into TextEditor.
when input no-latin character such as Chinese into TextEditor, i must first input some word to trigger CompositionStart/CompositionUpdate event, then input some mark to trigger CompositionEnd event.
if i just want to type in a single character that trigger CompositionEnd event, TextEditor won't insert this character to text.

![image](https://github.com/emilk/egui/assets/16380650/7a1fb0b3-0721-4be7-a9d7-c8038e88f6d9)

**To Reproduce**
1.Setup a demo app containing a TextEdit
2.Use Chinese PinYin input.
3.Do not type in any other word, just press [shift+1] , [shift+6] , [,], [.], [?] or any other CompositionEnd character.
4.Nothing typed into TextEdit

**Expected behavior**
Expected to type in Chinese character directly such as ！……，。《》、【】without need to trigger CompositionStart/CompositionUpdate event.

**Desktop (please complete the following information):**
 - OS: Windows11 Naive
 - Browser Naive App
 - Version 0.24.1 and master branch

/crates/egui/src/widgets/text_edit/builder.rs L1061.
```if prediction != "\n" && prediction != "\r" && state.has_ime```
this line ensure that i cannot input any CompositionEnd character unless input something to trigger CompositionStart to make ```has_ime``` as true.
tried to modify this line to
```if prediction != "\n" && prediction != "\r"```
if don't ensure ```has_ime``` be true, input CompositionEnd character directly successfully. Also, following some CompositionStart/CompositionUpdate character works too.
![image](https://github.com/emilk/egui/assets/16380650/765bc28a-7e64-4690-9c5b-056cd413c733)

emilk:
Can you check if `window.set_ime_allowed` is called at the correct time in `crates/egui-winit/src/lib.rs`?

FrankLeeC:
> Can you check if `window.set_ime_allowed` is called at the correct time in `crates/egui-winit/src/lib.rs`?

add log after L732 in crates/egui-winit/src/lib.rs
when mouse focus in textedit, `window.set_ime_allowed` is called, `allow_time` is true and `self.allow_time` is false.
when mouse focus outside textedit,
called again, `allow_time` is false and `self.allow_time` is true.

emilk:
Ok, so that at least seems to work.

[...]

PR #3768, opened 2024-01-04
Fix IME input of  `CompositionEnd` without a `CompositionStart`

FrankLeeC:
* Closes https://github.com/emilk/egui/issues/3766

Add support for type in CompositionEnd only characters without trigger CompositionStart first.
This usually works with no-latin character input.

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -1050,7 +1050,8 @@ fn events(
             }
 
             Event::CompositionEnd(prediction) => {
-                if prediction != "\n" && prediction != "\r" && state.has_ime {
+                // CompositionEnd only characters may be typed into TextEdit without trigger CompositionStart first, so do not check ```state.has_ime = true``` in the following statement.
+                if prediction != "\n" && prediction != "\r" {
                     state.has_ime = false;
                     let mut ccursor = delete_selected(text, &cursor_range);
                     if !prediction.is_empty() {
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -1050,7 +1050,7 @@ fn events(
             }
 
             Event::CompositionEnd(prediction) => {
-                // CompositionEnd only characters may be typed into TextEdit without trigger CompositionStart first, so do not check ```state.has_ime = true``` in the following statement.
+                // CompositionEnd only characters may be typed into TextEdit without trigger CompositionStart first, so do not check `state.has_ime = true` in the following statement.
                 if prediction != "\n" && prediction != "\r" {
                     state.has_ime = false;
                     let mut ccursor = delete_selected(text, &cursor_range);


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -1050,7 +1050,8 @@ fn events(
             }
 
             Event::CompositionEnd(prediction) => {
-                if prediction != "\n" && prediction != "\r" && state.has_ime {
+                // CompositionEnd only characters may be typed into TextEdit without trigger CompositionStart first, so do not check ```state.has_ime = true``` in the following statement.
```
emilk, line 1053:
```suggestion
                // CompositionEnd only characters may be typed into TextEdit without trigger CompositionStart first, so do not check `state.has_ime = true` in the following statement.
```

Is it still correct to check it for `Event::CompositionUpdate` above?

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -1050,7 +1050,8 @@ fn events(
             }
 
             Event::CompositionEnd(prediction) => {
-                if prediction != "\n" && prediction != "\r" && state.has_ime {
+                // CompositionEnd only characters may be typed into TextEdit without trigger CompositionStart first, so do not check `state.has_ime = true` in the following statement.
```
emilk, line 1053: Is it still correct to check it for `Event::CompositionUpdate` above?

emilk (approved):

crates/egui/src/widgets/text_edit/builder.rs
FrankLeeC: [This page](https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionstart_event) show that `compositionstart` could be fired after a user starts entering a Chinese character using a [Pinyin](https://en.wikipedia.org/wiki/Pinyin), so when entering some word except `compositionend` character, `componentstart` fires and `state.has_ime` will always be `true`, and `compositionupdate` will work well. so it's still correct to check it for `compositionupdate ` because `state.has_ime` is always `true` when entring word.

I have test the situation that remove all `state.has_ime` in `CompositionStart`, `CompositionUpdate`,`CompositionEnd`, it works well.

Also, `state.has_ime` is a variable that indicating current input state, maybe some other features will use this variable, so i think keep this variable is a good choice, although it seems useless at current time.

crates/egui/src/widgets/text_edit/builder.rs
FrankLeeC: [This page](https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionstart_event) show that `compositionstart` could be fired after a user starts entering a Chinese character using a [Pinyin](https://en.wikipedia.org/wiki/Pinyin), so when entering some word except `compositionend` character, `componentstart` fires and `state.has_ime` will always be `true`, and `compositionupdate` will work well. so it's still correct to check it for `compositionupdate ` because `state.has_ime` is always `true` when entring word.

I have test the situation that remove all `state.has_ime` in `CompositionStart`, `CompositionUpdate`,`CompositionEnd`, it works well.

Also, `state.has_ime` is a variable that indicating current input state, maybe some other features will use this variable, so i think keep this variable is a good choice, although it seems useless at current time.

Issue #3626, opened 2023-11-24
The keyboard shortcut of zoom in does not work on the JIS keyboard


Umio-Yasuno:
**Describe the bug**
The keyboard shortcut of zoom in (`Ctrl+PlusEqual`) does not work on the JIS keyboard layout.  
On the JIS keyboard layout, we need the shift key to type `+ (Shift+;)` or `= (Shift+-)`.

https://en.wikipedia.org/wiki/Japanese_input_method
https://en.wikipedia.org/wiki/Keyboard_layout#Japanese

I have confirmed that this issue will be fixed with the following patch.  

```diff
diff --git a/crates/egui/src/gui_zoom.rs b/crates/egui/src/gui_zoom.rs
index fb4f9bd..9cb8165 100644
--- a/crates/egui/src/gui_zoom.rs
+++ b/crates/egui/src/gui_zoom.rs
@@ -8,6 +8,8 @@ pub mod kb_shortcuts {

     pub const ZOOM_IN: KeyboardShortcut =
         KeyboardShortcut::new(Modifiers::COMMAND, Key::PlusEquals);
+    pub const ZOOM_IN_WITH_SHIFT: KeyboardShortcut =
+        KeyboardShortcut::new(Modifiers::COMMAND.plus(Modifiers::SHIFT), Key::PlusEquals);
     pub const ZOOM_OUT: KeyboardShortcut = KeyboardShortcut::new(Modifiers::COMMAND, Key::Minus);
     pub const ZOOM_RESET: KeyboardShortcut = KeyboardShortcut::new(Modifiers::COMMAND, Key::Num0);
 }
@@ -21,7 +23,10 @@ pub(crate) fn zoom_with_keyboard(ctx: &Context) {
     if ctx.input_mut(|i| i.consume_shortcut(&kb_shortcuts::ZOOM_RESET)) {
         ctx.set_zoom_factor(1.0);
     } else {
-        if ctx.input_mut(|i| i.consume_shortcut(&kb_shortcuts::ZOOM_IN)) {
+        if ctx.input_mut(|i| {
+            i.consume_shortcut(&kb_shortcuts::ZOOM_IN)
+                || i.consume_shortcut(&kb_shortcuts::ZOOM_IN_WITH_SHIFT)
+        }) {
             zoom_in(ctx);
         }
         if ctx.input_mut(|i| i.consume_shortcut(&kb_shortcuts::ZOOM_OUT)) {
```

emilk:
[...]

Issue #3653, opened 2023-11-28
Tracking issue: improve Key input


emilk:
The keyboard input in egui is limited in a couple of ways:

* Lots of keys missing from `Event::Key`
* We don't report the `super/meta` modifier key
* We don't report modifier key presses (just their state)
* Physical keys not reported on the eframe web backend
* `KeyboardShortcut` should ignore `shift` and `alt` keys for logical keys (e.g. `Ctrl Plus` may require pressing shift on some keyboards). See https://github.com/emilk/egui/issues/3626 for more

## How can I help?
* Look at the code for `egui::Key` and follow the instructions there for adding new keys
* Investigate if we can get physical keycode on web

## Requested keys
* Numpad
* Volume keys
* Curly brackets: `{` and `}`
* CapsLock (https://github.com/emilk/egui/issues/2041)
* Modifier keys

## Prioir art
* `winit`: https://docs.rs/winit/latest/winit/event/struct.KeyEvent.html
* https://github.com/pyfisch/keyboard-types

## Relevant issues
* https://github.com/emilk/egui/issues/3636
* https://github.com/emilk/egui/issues/3626
* https://github.com/emilk/egui/issues/3415
* https://github.com/emilk/egui/issues/3372
* https://github.com/emilk/egui/issues/3303
* https://github.com/emilk/egui/issues/3099
* https://github.com/emilk/egui/issues/3094
* https://github.com/emilk/egui/issues/2977
* https://github.com/emilk/egui/issues/2041
* https://github.com/emilk/egui/issues/4081

## Physical vs Logical keys
* Partially fixed by https://github.com/emilk/egui/pull/3649 (needs web support, if even possible?)

Consider someone owning a physical QWERTY keyboard:
<img width="500" alt="qwerty" src="https://github.com/emilk/egui/assets/1148717/d1a14196-7b07-4ac4-bd32-9093052d6401">

However, they have remapped it to use Dvorak (perhaps even repainting the text on their keyboard):
<img width="500" alt="dvorak" src="https://github.com/emilk/egui/assets/1148717/fcdc7e9b-4801-426f-a26c-c9c2028265dd">

If they start hitting Caps Lock and then moving to the left, the keys they will be hitting will be

* physical (qwerty): CapsLock, A, S, D, F, …
* logical (Dvorak): CapsLock, A, O, E, U …

The logical keys makes sense for most keyboard shortcuts - if you say "Press Cmd+S to save" in your UI, the user will be looking for the logical key "S"
[...]

PR #3769, opened 2024-01-04
Ignore extra SHIFT and ALT when matching modifiers

emilk:
* Closes https://github.com/emilk/egui/issues/3626

Basically, egui now ignores extra SHIFT and ALT pressed when matching keyboard shortcuts.
This is because SHIFT and ALT are often requires to produce some logical keys.
For instance, typing `+` on an English keyboard requires pressing `SHIFT =`,
so the keyboard shortcut looking for `CTRL +` should ignore the SHIFT key.

@abey79 You reported problem using `Cmd +` and `Cmd -` to zoom - does this fix it for you?

You can run with `RUST_LOG=egui_winit=trace cargo run` to see a printout of how winit reports the logical and physical keys, and how egui interprets them.

Weirdly, on Mac winit reports `SHIFT =` as `+`, but `CMD SHIFT =` as `=` (on an English keyboard) so things are… difficult.

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -1073,9 +1073,8 @@ fn key_from_key_code(key: winit::keyboard::KeyCode) -> Option<egui::Key> {
 
         KeyCode::Minus | KeyCode::NumpadSubtract => Key::Minus,
 
-        // Using Mac the key with the Plus sign on it is reported as the Equals key
-        // (with both English and Swedish keyboard).
-        KeyCode::Equal | KeyCode::NumpadAdd => Key::PlusEquals,
+        KeyCode::NumpadAdd => Key::Plus,
+        KeyCode::Equal => Key::Equals,
 
         KeyCode::Digit0 | KeyCode::Numpad0 => Key::Num0,
         KeyCode::Digit1 | KeyCode::Numpad1 => Key::Num1,
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -689,6 +689,15 @@ impl State {
 
         let logical_key = key_from_winit_key(logical_key);
 
+        // Helpful logging to enable when adding new key support
+        log::trace!(
+            "logical {:?} -> {:?},  physical {:?} -> {:?}",
+            event.logical_key,
+            logical_key,
+            event.physical_key,
+            physical_key
+        );
+
         if let Some(logical_key) = logical_key {
             if pressed {
                 if is_cut_command(self.egui_input.modifiers, logical_key) {
--- a/crates/egui/src/data/input.rs
+++ b/crates/egui/src/data/input.rs
@@ -968,8 +968,11 @@ pub enum Key {
     /// `.`
     Period,
 
-    /// The for the Plus/Equals key.
-    PlusEquals,
+    /// `+`
+    Plus,
+
+    /// `=`
+    Equals,
 
     /// `;`
     Semicolon,
@@ -1088,7 +1091,8 @@ impl Key {
         Self::Comma,
         Self::Minus,
         Self::Period,
-        Self::PlusEquals,
+        Self::Plus,
+        Self::Equals,
         Self::Semicolon,
         // Digits:
         Self::Num0,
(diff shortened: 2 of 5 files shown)


abey79:
Unfortunately no, this doesn't solve the issue for some reason. Here is the trace when I hit:
- shift-1 
- cmd-shit-1

```
[2024-01-05T07:58:06Z TRACE egui_winit] Unknown key: Shift
[2024-01-05T07:58:06Z TRACE egui_winit] logical Named(Shift) -> None,  physical Code(ShiftLeft) -> None
[2024-01-05T07:58:07Z TRACE egui_winit] logical Character("+") -> Some(Plus),  physical Code(Digit1) -> Some(Num1)
[2024-01-05T07:58:07Z TRACE egui_winit] logical Character("+") -> Some(Plus),  physical Code(Digit1) -> Some(Num1)
[2024-01-05T07:58:08Z TRACE egui_winit] Unknown key: Shift
[2024-01-05T07:58:08Z TRACE egui_winit] logical Named(Shift) -> None,  physical Code(ShiftLeft) -> None
[2024-01-05T07:58:09Z TRACE egui_winit] Unknown key: Shift
[2024-01-05T07:58:09Z TRACE egui_winit] logical Named(Shift) -> None,  physical Code(ShiftLeft) -> None
[2024-01-05T07:58:09Z TRACE egui_winit] Unknown key: Super
[2024-01-05T07:58:09Z TRACE egui_winit] logical Named(Super) -> None,  physical Code(SuperLeft) -> None
[2024-01-05T07:58:09Z TRACE egui_winit] logical Character("1") -> Some(Num1),  physical Code(Digit1) -> Some(Num1)
[2024-01-05T07:58:09Z TRACE egui_winit] logical Character("1") -> Some(Num1),  physical Code(Digit1) -> Some(Num1)
[2024-01-05T07:58:09Z TRACE egui_winit] Unknown key: Super
[2024-01-05T07:58:09Z TRACE egui_winit] logical Named(Super) -> None,  physical Code(SuperLeft) -> None
[2024-01-05T07:58:09Z TRACE egui_winit] Unknown key: Shift
[2024-01-05T07:58:09Z TRACE egui_winit] logical Named(Shift) -> None,  physical Code(ShiftLeft) -> None
```

shift-1 is where the + is in the ch-fr keyboard:
image


(Bonus keyboard layout trivia: the weird glyph layout in the "é/ö" keys and neighbours is due to the fact that there are two sub-layouts: swiss-french and swiss-german. The only difference between the two is "éèà" by default / "öäü" with shift for the swiss-french, and the opposite for the swiss-gemeran.)

emilk:
Ok, so the problem is in winit (or earlier): the holding down `ctrl`, the logical key is not correct, because `shift` seems ignored.

emilk:
Fix for winit in https://github.com/rust-windowing/winit/pull/3361

emilk:
This should do the right think on Windows and Linux (I hope), but for macOS we have to wait for https://github.com/rust-windowing/winit/pull/3361

seanlinsley:
My application was distinguishing between shortcuts with and without Shift, which this breaks. It also causes surprising behavior:

```rs
let change = if i.consume_key(Modifiers::COMMAND | Modifiers::SHIFT, Key::Plus) {
    // correctly captures Command-Shift-Plus, even though it shouldn't anymore?
    10
} else if i.consume_key(Modifiers::COMMAND, Key::Plus) {
    // never triggers
    1
} else if i.consume_key(Modifiers::COMMAND | Modifiers::SHIFT, Key::Minus) {
    // never triggers
    -10
} else if i.consume_key(Modifiers::COMMAND, Key::Minus) {
    // correctly captures Command-Minus
    -1
} else {
    return;
};
```
If not for `consume_key(Modifiers::COMMAND, Key::Plus)` no longer working I would be able to simulate the previous behavior by also checking `i.modifiers.shift`

seanlinsley:
This almost replicates the previous behavior, except that `Command-Shift-Minus` doesn't seem to create a `InputState` event anymore.

```rs
let mut change = None;
i.events.retain(|event| {
    match event {
        Event::Key { physical_key: Some(Key::Equals), modifiers, pressed: true, .. } => {
            change = Some(if modifiers.shift { 10 } else { 1 });
            false
        }
        Event::Key { physical_key: Some(Key::Minus), modifiers, pressed: true, .. } => {
            change = Some(if modifiers.shift { -10 } else { -1 });
            false
        }
        _ => true,
    }
});

emilk:
Pressing shift and minus at the same time will on my keyboard produce the underscore character: `_`, i.e not a `Key::Minus`.

You are trying to check for a physical key, when egui uses logical keys.

See https://docs.rs/egui/latest/egui/enum.Event.html#variant.Key and https://github.com/emilk/egui/issues/3653 for more

PR #3770, opened 2024-01-05
Add missing links to source code

c-git:

### DIFF
--- a/crates/egui_demo_lib/src/demo/about.rs
+++ b/crates/egui_demo_lib/src/demo/about.rs
@@ -51,6 +51,10 @@ impl super::View for About {
             ui.hyperlink_to("Rerun.io", "https://www.rerun.io/");
             ui.label(", a startup building an SDK for visualizing streams of multimodal data");
         });
+
+        ui.vertical_centered(|ui| {
+            ui.add(crate::egui_github_link_file!());
+        });
     }
 }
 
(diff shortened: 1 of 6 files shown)


c-git:
Please let me know if screenshot are needed. Not sure if it's worth the effort.

c-git:
Tried to run the check script but it can't run as is. See error below.

```sh
$ ./scripts/check.sh 
+ cargo install --quiet cargo-cranky
+ cargo install --quiet typos-cli
error: failed to compile `typos-cli v1.17.0`, intermediate artifacts can be found at `/tmp/cargo-installOumziI`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  package `varcon-core v4.0.4` cannot be built because it requires rustc 1.75 or newer, while the currently active rustc version is 1.72.0
  Try re-running cargo install with `--locked`
```

ryo33:
@c-git I've temporary modified the rust-toolchain.toml to use the Rust version 1.75.0, and it works as a workaround of this error in my PR. I don't know should I open an issue or PR for this error.

c-git:
That's fair, I guessed it was something like that but didn't have the time to check it.

c-git:
I feel like since the PR would be small should be easy enough to do and get feedback on.

emilk (approved):

emilk:
Thanks!

PR #3778, opened 2024-01-07
fix egui-wgpu callback viewport size check (#3759)

### LINKED ISSUE #3759: Viewport out-of-range when first constructing certain UIs rendered with egui-wgpu callbacks
**Describe the bug**

Viewport out-of-range, causing a validation error, when first constructing certain UIs rendered with egui-wgpu callbacks.
This only happens on the first frame, the fix would simply be skipping the callback and set viewport command when the viewport rect is outside of the display.

**To Reproduce**
1. Make a Bottom panel with a scroll area inside
2. Add egui-wgpu Callbacks to it

**Expected behavior**
The item to be skipped if the viewport is outside of display.

**Additional context**
https://github.com/automancy/automancy/blob/main/src/bin/automancy/gui/tile_selection.rs

(Remove the following from the code to actually get the panic:
```rs
        if !ui.ctx().screen_rect().contains_rect(rect) {
            continue;
        }
```
)

msparkles:
Closes #3759

### DIFF
--- a/crates/egui-wgpu/src/renderer.rs
+++ b/crates/egui-wgpu/src/renderer.rs
@@ -450,39 +450,35 @@ impl Renderer {
                         continue;
                     };
 
-                    if callback.rect.is_positive() {
+                    let info = PaintCallbackInfo {
+                        viewport: callback.rect,
+                        clip_rect: *clip_rect,
+                        pixels_per_point,
+                        screen_size_px: size_in_pixels,
+                    };
+
+                    let viewport_px = info.viewport_in_pixels();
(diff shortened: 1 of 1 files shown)


crates/egui-wgpu/src/renderer.rs
```diff
@@ -450,39 +450,35 @@ impl Renderer {
                         continue;
                     };
 
-                    if callback.rect.is_positive() {
```
emilk, line 453: Shouldn't we check `callback.rect.is_positive()` too?

emilk: oh right, nevermind

emilk (approved):

PR #3798, opened 2024-01-09
Update wgpu to 0.19

Dinnerbone:
wgpu 0.19 releases on the 17th, opening this in advance so we can hopefully have a smooth update.

`wgpu::Surface` now takes a lifetime. I've kept it `'static` as that's the equivalent to previous code, but in theory a future (breaking?) change could be to keep that lifetime around.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "ab_glyph"
-version = "0.2.21"
+version = "0.2.23"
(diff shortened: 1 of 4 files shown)


Wumpf:
As mentioned on wgpu matrix chat I didn't see this draft in time and started an independent one when working on the single-wasm wgpu feature: https://github.com/emilk/egui/compare/wgpu-update?expand=1
Since we seem to agree on the changes regarding surface handling I suggest continuing work with that one instead :)
(likely only minor changes are needed between here and the final pr)

PR #3803, opened 2024-01-10
Added command for adding Egui to project

Goodnessuc:
Added command for adding Egui to project; may be handy for Rust newbies looking to experiment


emilk (commented):
`cargo add egui` is not enough though, users need to use eframe or some other egui integration

PR #3812, opened 2024-01-12
egui-winit: Don't consume clipboard shortcuts

Dinnerbone:
This breaking change seems to have snuck in during #3649. Previously it would never consume these, today it consumes them unconditionally, and with this change it'll ~~only consume them if egui wants text input~~ never consume them.

This is a blocker for updating our application, sadly :(

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -341,9 +341,10 @@ impl State {
                 }
             }
             WindowEvent::KeyboardInput { event, .. } => {
+                self.on_keyboard_input(event);
+
                 // When pressing the Tab key, egui focuses the first focusable element, hence Tab always consumes.
-                let consumed = self.on_keyboard_input(event)
-                    || self.egui_ctx.wants_keyboard_input()
+                let consumed = self.egui_ctx.wants_keyboard_input()
                     || event.logical_key
                         == winit::keyboard::Key::Named(winit::keyboard::NamedKey::Tab);
                 EventResponse {
@@ -653,7 +654,7 @@ impl State {
         }
     }
 
-    fn on_keyboard_input(&mut self, event: &winit::event::KeyEvent) -> bool {
+    fn on_keyboard_input(&mut self, event: &winit::event::KeyEvent) {
         let winit::event::KeyEvent {
             // Represents the position of a key independent of the currently active layout.
             //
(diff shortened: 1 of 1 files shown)


crates/egui-winit/src/lib.rs
```diff
@@ -342,8 +342,8 @@ impl State {
             }
             WindowEvent::KeyboardInput { event, .. } => {
                 // When pressing the Tab key, egui focuses the first focusable element, hence Tab always consumes.
-                let consumed = self.on_keyboard_input(event)
-                    || self.egui_ctx.wants_keyboard_input()
```
emilk, line 346: Here is the relevant diff of https://github.com/emilk/egui/pull/3649:

image

The change was was that some events were now marked consumed, namely cut/copy/past commands.

Is it these that are troubling you?

You change is a breaking change, not a revert.

If you want to revert to 0.24, you should ignore the result of `on_keyboard_input`, but your new code doesn't even call `on_keyboard_input`, which means the keyboard events won't make it to egui at all, which isn't great (the eframe app may be very interested in the key events, even though there is no TextEdit or similar listening)

emilk (changes requested):

crates/egui-winit/src/lib.rs
Dinnerbone: Sorry I'm a little confused. The new code **does** call `on_keyboard_input`, and I can see the clipboard events working fine when, for example, a textfield is focused. But when egui unconditionally consumes things when it doesn't need them, it's not possible to observe them outside of egui.

I agree that ignoring the result is closer to 0.24 but I figure this change was likely intentional, even if not documented, so I thought to keep it when egui flags that it wants them. Is that not correct?

emilk: I agree with your diagnosis of the symptom, but your fix is wrong.

You code is `self.egui_ctx.wants_keyboard_input() && self.on_keyboard_input(event)` and will only call `on_keyboard_input` if `wants_keyboard_input()` is true, due to boolean short-circuting. This means egui won't be told of the events, which would be a new bug.

The correct fix would be to call `self.on_keyboard_input(event);` and ignore its return value, like we did in 0.24

Dinnerbone: Alright, I'll revert it to how it was before that PR.

rustbasic:
@emilk @Dinnerbone 
My program uses something like below:

```rust
    if ui.input_mut(|i| i.events.last() == Some(&Event::Copy))
    || ui.input_mut(|i| i.events.last() == Some(&Event::Cut))
    {

    }
```

1. I think it would be a good idea to consume the keycode when the key is released.
( or )
2. I think it would be nice to have the ability to turn on/off events or various basic keys (ex: Ctrl+A, Ctrl+C, Ctrl+K Ctrl+Z...) in TextEdit.

crates/egui-winit/src/lib.rs
torokati44: Is this resolved then?

emilk (approved):

emilk:
@rustbasic let's consider what to do there in a separate issue or PR. It feels like a much larger issue.

For now I think it makes sense to revert to the old behavior of only considering Copy/Cut/Paste consumed if egui has `wants_keyboard_input` set, i.e. if a `TextEdit` is selected.

Issue #3675, opened 2023-12-04
WebGPU build broken


emilk:
`./scripts/build_demo_web.sh --webgpu --release`
`./scripts/start_server.sh`

> panicked at 'canvas context is not a GPUCanvasContext: Object { obj: JsValue(GPUCanvasContext) }', /Users/emilk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.18.0/src/backend/web.rs:975:14

Wumpf:
struggling with this one. It seems to fail a cast that should just work. Suspected some kind of web_sys/wasm-bindgen version issue but updating those didn't help so far. Hoping for responses on wgpu matrix chat right now

emilk:
According to @Wumpf the wgpu web build works again with [this wgpu pr](https://github.com/gfx-rs/wgpu/pull/5044)), so this will likely solve itself when we update wgpu

rakurtz:
Hey emilk, i am not sure if the error message i get when i try to compile the eframe template with wgpu instead of glow is related to the above described issue:

```
error: failed to select a version for `web-sys`.
    ... required by package `wgpu v0.18.0`
    ... which satisfies dependency `wgpu = "^0.18.0"` of package `eframe v0.25.0`
    ... which satisfies dependency `eframe = "^0.25.0"` (locked to 0.25.0) of package `msg-extractor v0.1.0 (/Users/xxxxxx/VisualStudioCodeProjects/Rust/HELPER/msg-extractor)`
versions that meet the requirements `^0.3.64` (locked to 0.3.67) are: 0.3.67

the package `wgpu` depends on `web-sys`, with features: `GpuComputePassTimestampWrite` but `web-sys` does not have these features.
```

Maybe that helps ?

emilk:
no, that's a compilation issue. Looks like you're using an old web-sys

emilk:
A fix is coming in https://github.com/emilk/egui/pull/3824

rakurtz:
> no, that's a compilation issue. Looks like you're using an old web-sys

mmh, i am using the current version of web-sys 0.3.67. I just took the `eframe-template` and switched to `wgpu` in `Cargo.toml` though. Shall i file another issue dedicated to the above error messages?

PR #3824, opened 2024-01-15
Update wgpu to 0.19

emilk:
* Closes https://github.com/emilk/egui/issues/3675


Cargo.toml
```diff
@@ -54,8 +54,6 @@ puffin = "0.18"
...
 thiserror = "1.0.37"
 
-# Make the renderer `Sync` even on wasm32, because it makes the code simpler:
-wgpu = { version = "0.18.0", features = ["fragile-send-sync-non-atomic-wasm"] }
-
-# Use this to build wgpu with WebGL support on the Web *instead* of using WebGPU.
-#wgpu = { version = "0.18.0", features = ["webgl"] }
+wgpu = { git = "https://github.com/gfx-rs/wgpu", commit = "7774f31", features = [
+    "fragile-send-sync-non-atomic-wasm",
+] } # Make the renderer `Sync` even on wasm32, because it makes the code simpler:# TODO: Use a stable version of wgpu.
```
emilk, line 59:
```suggestion
    # Make the renderer `Sync` even on wasm32, because it makes the code simpler:
    "fragile-send-sync-non-atomic-wasm",
] } # TODO: Use a stable version of wgpu.
```

egui_demo_app_3.exe
emilk, line 1: no thank you

crates/egui_demo_app/Cargo.toml
```diff
@@ -57,6 +57,8 @@ log = { version = "0.4", features = ["std"] }
 bytemuck = { version = "1.7.1", optional = true }
 puffin = { version = "0.18", optional = true }
 puffin_http = { version = "0.15", optional = true }
+# Enable both webgl & webgpu for targeting the web (these features have no effect when not targeting wasm32)
+wgpu = { workspace = true, features = ["webgpu", "webgl"], optional = true }
```
emilk, line 61: We should document this in `eframe`

crates/eframe/src/native/wgpu_integration.rs
```diff
@@ -544,7 +546,8 @@ impl WgpuWinitRunning {
 
             {
                 crate::profile_scope!("set_window");
-                if let Err(err) = pollster::block_on(painter.set_window(viewport_id, Some(window)))
+                if let Err(err) =
+                    pollster::block_on(painter.set_window(viewport_id, Some(window.clone())))
```
emilk, line 550: nit: it is often nicer to use `Arc::clone` to explicitly show what we're doing

```suggestion
                    pollster::block_on(painter.set_window(viewport_id, Some(Arc::clone(window))))
```

emilk:
We should show what backend (WebGPU or WebGL) is being used in the egui_demo_app GUI

egui_demo_app_3.exe
Wumpf: huh how did that happen

also, you're welcome!

crates/eframe/Cargo.toml
```diff
@@ -92,7 +92,15 @@ web_screen_reader = [
 ]
 
 ## Use [`wgpu`](https://docs.rs/wgpu) for painting (via [`egui-wgpu`](https://github.com/emilk/egui/tree/master/crates/egui-wgpu)).
+##
 ## This overrides the `glow` feature.
+##
+## If you're targetting the web, you also need to turn on the
+## `webgl` and/or `webgpu` features of the `wgpu` crate.
```
Wumpf, line 99: Maybe worth noting that `webgpu` is enabled by default on wgpu.
I just called them out explicitly in the demo's Cargo.toml for visibility.

emilk:
`glutin` is locked at `raw-window-handle 0.5`, and `wgpu` at `raw-window-handle 0.6`. This is hell.

torokati44:
Is it still with `rwh` `0.6`'s anecdotal ability to coexist with `0.5`, and convert between handles of the two versions? :/

emilk:
@Wumpf I only see the WebGL in the list of available adapters (from `instance.enumerate_adapters(wgpu::Backends::all())`) even in my chromium browser supporting WebGPU

Wumpf:
`enumerate_adapters` is too late. WebGPU is unfortunately the anomaly where the _instance_ has to decide whether it picks WebGPU. So now the question ofc why that one is not seeing WebGPU. For that it's important to specify the right parameters for instance creation. I can have a look

Wumpf:
@emilk after I finished writing the last comment I rememberd a weird anomaly I saw when testing this pr originally. Had to first double check again to confirm this. When you open the website on chrome via `http://0.0.0.0:8888/` it will not work - in fact it doesn't even work from the javascript console then, you'll find that `navigator.gpu` is always undefined. If you however go to http://localhost:8888/ it works just fine for me. Can you confirm?

emilk:
Yup, that was it 🤦  so weird

crates/egui-wgpu/src/lib.rs
```diff
@@ -57,6 +57,10 @@ pub struct RenderState {
     pub adapter: Arc<wgpu::Adapter>,
 
     /// All the available adapters.
+    ///
+    /// This is not availabler on web.
```
torokati44, line 61: nit: typo: "availabler"

crates/eframe/src/web/web_painter_wgpu.rs
```diff
@@ -87,6 +87,10 @@ impl WebPainterWgpu {
     pub async fn new(canvas_id: &str, options: &WebOptions) -> Result<Self, String> {
         log::debug!("Creating wgpu painter");
 
+        if crate::web::web_location().hostname == "0.0.0.0" {
+            log::warn!("Due to a chromium bug, WebGPU doesn't work when the hostname is '0.0.0.0'");
```
Dinnerbone, line 91: WebGPU is only available in secure contexts; that's anything HTTPS and a **few** whitelisted hostnames depending on browser, usually just `localhost`.

emilk: Thanks for the clarification!

Wumpf: ahh this makes a lot of sense! thanks!

crates/eframe/src/web/web_painter_wgpu.rs
```diff
@@ -111,6 +112,7 @@ impl WebPainterWgpu {
             present_mode: options.wgpu_options.present_mode,
             alpha_mode: wgpu::CompositeAlphaMode::Auto,
             view_formats: vec![render_state.target_format],
+            desired_maximum_frame_latency: 2, // TODO(emilk): expose to eframe users
```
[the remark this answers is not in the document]
Wumpf: nope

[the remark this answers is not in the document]
Wumpf: very much approving that you keep to the default of 2 because dealing with the fallout of 1 is pretty much an expert feature right now.
It would aactually be even better if we were to use [`..surface.get_default_config`](https://docs.rs/wgpu/latest/wgpu/struct.Surface.html#method.get_default_config) here :)


(oh and by the way I broke DX12 for a random assortment of drivers with the changes that were made to that backend when introducing the feature https://github.com/gfx-rs/wgpu/issues/5088)

crates/eframe/src/web/web_painter_wgpu.rs
```diff
@@ -87,8 +87,13 @@ impl WebPainterWgpu {
     pub async fn new(canvas_id: &str, options: &WebOptions) -> Result<Self, String> {
         log::debug!("Creating wgpu painter");
 
-        if crate::web::web_location().hostname == "0.0.0.0" {
-            log::warn!("Due to a chromium bug, WebGPU doesn't work when the hostname is '0.0.0.0'");
+        {
+            let is_secure_context = web_sys::window().map_or(false, |w| w.is_secure_context());
+            if !is_secure_context {
+                log::info!(
+                    "WebGPU is only available in seucre contexts, i.e. on HTTPS and on localhost"
```
Mingun, line 94:
```suggestion
                    "WebGPU is only available in secure contexts, i.e. on HTTPS and on localhost"
```

Cargo.toml
```diff
@@ -51,11 +51,10 @@ opt-level = 2
...
-raw-window-handle = "0.5.0"
+raw-window-handle = "0.6.0"
 thiserror = "1.0.37"
-
-# Make the renderer `Sync` even on wasm32, because it makes the code simpler:
-wgpu = { version = "0.18.0", features = ["fragile-send-sync-non-atomic-wasm"] }
-
-# Use this to build wgpu with WebGL support on the Web *instead* of using WebGPU.
-#wgpu = { version = "0.18.0", features = ["webgl"] }
+wgpu = { version = "0.19", features = [
```
cwfitzgerald, line 56: I know this has already landed, but this really should be no-default-features at least for egui-wgpu

emilk: Will fix!

emilk: Here we go: https://github.com/emilk/egui/pull/3875

Issue #3825, opened 2024-01-15
SVG rendering panics when size_hint w or h is 0


amPerl:
**Describe the bug**
The initial frame in the `images` example results in a size_hint of `[4984.0, 0.0]` passed to the ferris SVG. Prior to #3719 an invalid size hint for a SVG would error out while creating the pixmap, but now it runs into an unwrap on creating an IntSize, triggering a panic.

**To Reproduce**
1. Checkout `0.25.0` or 963be247f26f2ddd72c03649e9ebc0ec88ac8431
2. Add `cc.egui_ctx.set_pixels_per_point(2.0);` to the app creator in `examples/images/src/main.rs`
3. Run the example with `cargo run -p images`
4. Get ```thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', crates\egui_extras\src\image.rs:248:83``` at `IntSize::from_wh(w, h).unwrap()` because `h` is 0 from `available_size`

**Expected behavior**
There is no panic

**Desktop (please complete the following information):**
 - OS: Windows
 - Version 0.25.0

PR #3826, opened 2024-01-15
Fix unwraps in SVG scaling

amPerl:
Added error messages when scaling to invalid sizes instead of panicking through unwrap, returning to previous behavior.

Closes <https://github.com/emilk/egui/issues/3825>.


emilk (approved):
Very good

Issue #3830, opened 2024-01-17
`stable_dt` is broken


emilk:
It seems the multi-viewport work broke `stable_dt`, making animations in egui bad

PR #3832, opened 2024-01-17
Fix `stable_dt`

emilk:
This broke in https://github.com/emilk/egui/pull/3172

Since 0.24.1, `stable_dt` has been fixed at 1/60s, which is a little bit _too_ stable.

The code issue was the logic for asking "Is this the result of an immeditate repaint?" was completely broken (always returning false).

* Closes https://github.com/emilk/egui/issues/3830

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -68,6 +68,28 @@ impl Default for WrappedTextureManager {
 
 /// Repaint-logic
 impl ContextImpl {
+    /// This is where we update the repaint logic.
+    fn begin_frame_repaint_logic(&mut self, viewport_id: ViewportId) {
+        let viewport = self.viewports.entry(viewport_id).or_default();
+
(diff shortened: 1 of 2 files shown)


crates/egui/src/context.rs
```diff
@@ -79,13 +101,13 @@ impl ContextImpl {
         // This solves some corner-cases of missing repaints on frame-delayed responses.
         viewport.repaint.outstanding = 1;
 
-        if let Some(callback) = &self.request_repaint_callback {
-            // We save some CPU time by only calling the callback if we need to.
-            // If the new delay is greater or equal to the previous lowest,
-            // it means we have already called the callback, and don't need to do it again.
-            if delay < viewport.repaint.repaint_delay {
-                viewport.repaint.repaint_delay = delay;
```
emilk, line 87: This was a separate bug: `repaint_delay` was wrong for integrations that don't install a repaint-callback

Issue #3321, opened 2023-09-08
eframe: Graphical glitches when window has been minimized


rustbasic:
Title: Regarding severe graphical glitches with egui

Dear emilk,

When continuously running a program that uses the egui library on windows, there is a bug causing severe graphic glitches after about 1 to 6 hours.

Before I explain the circumstances leading to this glitch, I want to highlight another bug.

            frame.set_minimized(true);

When using the above function or the native windows minimize button and then restoring the program to its original size, the frame.info().window_info.minimized remains constantly set to true. Therefore, it's impossible to determine whether the program is in a minimized state. In the minimized state, ui.available_width() slightly decreases, and ui.available_height() becomes 0.0. After restoring from the minimized state, while ui.available_width() returns to normal, sometimes ui.available_height() remains at 0.0.

I've tried to determine the exact conditions that lead to the graphic glitches. The frequency of glitches appears to be higher when restoring the program to its original size after minimizing it for over an hour. The graphical glitches also occur every few hours even without minimizing the program. When these glitches happen, it seems ui.available_height() is at 0.0. It's presumed that the glitches might be linked to the instances when, upon restoring from minimized state, ui.available_height() occasionally remains at 0.0.

Please note that this behavior isn't consistent, as it happens only every few hours. However, these are my observations so far while trying to reproduce and prevent this glitch.

Thanks.

---------------------------------------------------------
egui Version
egui 0.22.0
egui 0.21.0

[ BUG Capture ]
https://github.com/rustbasic/rdesk/blob/main/bug-capture-1.jpg?raw=true

[ No Bug Capture ]
https://github.com/rustbasic/rdesk/blob/main/no-bug-capture-1.jpg?raw=true

---------------------------------------------------------------------

Dear emilk,

I've been working for over 6 months to reproduce the issue of graphics breaking every few hours in a Windows program using the egui library. Through these efforts, I've roughly identified the reason and found a temporary way to prevent it.

When the program is minimized, the value of ui.available_height() decreases.

In the minimized state, the value of ui.available_height() seems to range between 0.0 to 20.0, depending on where the program was launched.

I tried using the method if ui.available_height() < 100.0 { return; }, but graphics still broke every few hours when restoring from a minimized state.

So,
```
let mut ui_available_height = ui.available_height();
if ui_available_height < 100.0 {
    ui_available_height = ui_available_height.at_least(100.0);
}
```
[...]

PR #3837, opened 2024-01-18
Fix graphical glitches on Windows

rustbasic:
Fix graphical glitches on Windows

request_repaint()
if is_minimized return

Issues :
#3321

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -1080,6 +1080,7 @@ impl Context {
                                     response.triple_clicked[*button as usize] =
                                         clicked && click.is_triple();
                                 }
+                                response.is_pointer_button_down_on = false;
                             }
                         }
                     }
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -1240,6 +1240,11 @@ impl Context {
     ///
     /// This will repaint the current viewport.
     pub fn request_repaint(&self) {
+        if cfg!(target_os = "windows") {
+            if self.input(|i| i.viewport().minimized.unwrap_or(false)) {
+                return;
+            }
+        }
         self.request_repaint_of(self.viewport_id());
     }
 
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -1240,10 +1240,8 @@ impl Context {
     ///
     /// This will repaint the current viewport.
     pub fn request_repaint(&self) {
-        if cfg!(target_os = "windows") && !cfg!(target_arch = "wasm32") {
-            if self.input(|i| i.viewport().minimized.unwrap_or(false)) {
+        if self.input(|i| i.viewport().minimized.unwrap_or(false)) {
                 return;
-            }
         }
         self.request_repaint_of(self.viewport_id());
     }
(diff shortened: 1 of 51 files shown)


rustbasic:
@emilk 
Dear emilk.

Pull Request.

Based on long-term testing on Windows, ignoring request_repaint() when minimized prevents graphical corruption.

Since `i.viewport().minimized` is `true` only on Windows,
I don't think we need to do `cfg!(target_os = "windows")`

```
if cfg!(target_os = "windows") && self.input(|i| i.viewport().minimized.unwrap_or(false)) {
    return;
}
```

valaphee:
but with cfg, the if gets optimized out as a result

rustbasic:
> but with cfg, the if gets optimized out as a result

Even in non-windows, there is no need to run request_repaint if it is minimized.

crates/egui/src/context.rs
```diff
@@ -1240,6 +1240,9 @@ impl Context {
     ///
     /// This will repaint the current viewport.
     pub fn request_repaint(&self) {
+        if self.input(|i| i.viewport().minimized.unwrap_or(false)) {
+            return;
```
emilk, line 1244: I don't think this is the right solution.

The core problem is that when minimized, `App::update` is being called with a bad viewport size. We should either make sure `App::update` isn't called when minimized, or if called it is done so with the size of the window from when it was large.

This fix will just fix _some_ reasons for why `App::update` might be called, but not all. For instance, calling `request_repaint_after`, and then minimizing, will still call `App::update`.

TL;DR: the fix needs to into `eframe`, not `egui`

emilk (changes requested):

crates/egui/src/context.rs
rustbasic: @emilk
Dear emilk,

I'm not sure if this is the right solution, but I've been able to avoid the problem with this method after testing it for about a year.

I think App::update() should continue to work and should not only output to the screen.

In most cases, most users only call request_repaint(), and there are no cases where request_repaint_after() is called.

Please make a Pull Request for this temporarily until I find a better way. 
If there is a better way, please fix it.

Thank you, emilk.

rustbasic: If you want, let it return only on Windows.

crates/egui/src/text_selection/text_cursor_state.rs
rustbasic, line 1: rev() is panic with UTF8

crates/egui/src/widgets/text_edit/text_buffer.rs
rustbasic, line 1: Not byte_index
It's char_index

#3846

crates/egui/src/text_selection/text_cursor_state.rs
rustbasic, line 1: @emilk 
Dear emilk.

`let width = unicode_width::UnicodeWidthChar::width_cjk(c).unwrap_or(2);`             

Thank you, emilk.

rustbasic:
@emilk 
Dear emilk.

I work finished, Pull Request.
#3846 
#3850

`let width = unicode_width::UnicodeWidthChar::width_cjk(c).unwrap_or(2);`

Thank you, emilk.

crates/egui/src/context.rs
rustbasic: Changed :
crates/eframe/src/native/run.rs 

The test results are normal.

PR #3849, opened 2024-01-20
Plot borrowed data

GunnarMorrigan:
- [x] Keep your PR:s small and focused
Sorry this is a big change for through egui_plot. I can not make it smaller.
- [x] cargo fmt
- [x] cargo cranky
- [x] Open the PR as a draft until you have self-reviewed it and run `./scripts/check.sh`.
The script does not work on 1.72.0

Partially closes https://github.com/emilk/egui/issues/1485.

After much trial and error this PR enables the user to plot borrowed data.

It allows the user to plot `Iterator`s, `slice`s, `PlotPoints`. While there is still optimization possible per my message in #1485. This is a good first step and allows for easy performance improvement on the users side. 
E.g. the user can now easily create a slice `&[PlotPoint]` that represents their point data or go a step further and represent their data using an `iter().filter(...)` approach.

Will this break current builds: **Yes**

- PlotPoint has been renamed to Coordinate
- Points has been made generic
- build_fn in plot.show() has gotten a different signature
- other changes


emilk (changes requested):

crates/egui_plot/src/lib.rs
```diff
@@ -1342,236 +1349,7 @@ enum BoundsModification {
     Set(PlotBounds),
     Translate(Vec2),
     AutoBounds(Vec2b),
-    Zoom(Vec2, PlotPoint),
-}
-
-/// Provides methods to interact with a plot while building it. It is the single argument of the closure
-/// provided to [`Plot::show`]. See [`Plot`] for an example of how to use it.
-pub struct PlotUi {
```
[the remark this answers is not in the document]
GunnarMorrigan: I will have a look but don't want to sink too much work into this PR anymore. It already took me a full day of trying many different approach leading to many commits and an hour to clear up the commits to what it is now.

I might not have the time available atm to rework the commits or this PR into several PRs as I will have to pull apart the commits again.

emilk:
Did you benchmark to see what kind of performance gains this can give?

crates/egui_plot/src/lib.rs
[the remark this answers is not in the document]
GunnarMorrigan: Adjusted to smaller commits.

crates/egui_plot/src/items/values.rs
```diff
@@ -154,9 +158,8 @@ impl Default for Orientation {
 ///
 /// These can be an owned `Vec` or generated with a function.
 pub enum PlotPoints {
-    Owned(Vec<PlotPoint>),
+    Owned(Vec<Coordinate>),
```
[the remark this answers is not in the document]
GunnarMorrigan: ok,

But this brings the code more in line with the current wording of 'CoordinatesFormatter' and [pointer_coordinate -> Option<PlotPoint>](https://docs.rs/egui_plot/latest/egui_plot/struct.PlotUi.html#method.pointer_coordinate)

GunnarMorrigan:
> Did you benchmark to see what kind of performance gains this can give?

I have done profiling to see what was hurting performance. The recreation of all Points was the first obvious problem.

In my case I store roughly 350k PlotPoints to draw, cloning or recalculating them every repaint gets in the way.
If I only need to map my data to PlotPoints once it is much more efficient. 

In the case of a dynamic element (e.g. a filter) data needs to be recomputed every time. There is still some improvement here as now you can use a lazy evaluated filter over your static PlotPoints. Additionally, there is added ergonomics in using an Iterator as you can easily apply `filter / filter_map / skip` etc

examples/save_plot/src/main.rs
```diff
@@ -32,8 +32,12 @@ impl eframe::App for MyApp {
 
             // let's create a dummy line in the plot
             let graph: Vec<[f64; 2]> = vec![[0.0, 1.0], [2.0, 3.0], [3.0, 2.0]];
-            let inner = my_plot.show(ui, |plot_ui| {
+            let inner = my_plot.show(ui, |plot_ui_builder| {
+                let mut plot_ui = plot_ui_builder.into_plot_ui();
+
                 plot_ui.line(Line::new(PlotPoints::from(graph)).name("curve"));
+
+                ((), plot_ui)
```
emilk, line 40: Every example now starts with `let mut plot_ui = plot_ui_builder.into_plot_ui();` and ends with `((), plot_ui)`.
This is a bit ugly.

I assume this is needed in order to get the lifetime of `plot_ui` correct?

Is there no way around it, e.g. assign a lifetime to `Plot` and use that for `plot_ui` too?

GunnarMorrigan: Agreed, I also found this ugly.

Hmm, maybe.

I thought the root cause was the creation of PlotUi in side the show function which leads to lifetime issues with the build_fn.

What I also tried as a solution was have the return is an `impl Into<BuildFnResultThingy>` which is implemented for tuples of (R, PlotUi) and PlotUi etc. to make returning more easy but this also adds indirection as what you can return or not.

emilk (commented):
Nice, this is looking much better!

It would be great if one of the examples in `crates/egui_demo_lib/src/demo/plot_demo.rs` used the new borrowing mechanism

crates/egui_demo_lib/src/demo/plot_demo.rs
```diff
@@ -368,7 +375,7 @@ impl MarkerDemo {
         markers_plot
             .show(ui, |plot_ui| {
                 for marker in self.markers() {
-                    plot_ui.points(marker);
+                    plot_ui.owned_points(marker);
```
emilk, line 378: This is a slightly annoying name change, but maybe fine

crates/egui_plot/src/items/mod.rs
```diff
@@ -792,7 +836,25 @@ pub struct Points {
     pub(super) stems: Option<f32>,
 }
 
-impl Points {
+impl<T> Points<T>
+where
+    T: GenericPlotPoints,
+{
+    pub fn new_generic(t: T) -> Self {
```
emilk, line 843: This should be called something like `from_iter`, and have a docstring

crates/egui_demo_lib/src/demo/plot_demo.rs
```diff
@@ -789,13 +890,15 @@ impl InteractionDemo {
...
+            let return_value = (
                 plot_ui.screen_from_plot(PlotPoint::new(0.0, 0.0)),
                 plot_ui.pointer_coordinate(),
                 plot_ui.pointer_coordinate_drag_delta(),
                 plot_ui.plot_bounds(),
                 plot_ui.response().hovered(),
-            )
+            );
+
+            return_value
```
emilk, line 901: unnecessary `let` binding

crates/egui_plot/src/plot_ui.rs
```diff
@@ -0,0 +1,328 @@
...
+        if points.color == Color32::TRANSPARENT {
+            points.color = self.auto_color();
+        }
+        self.items.push(Box::new(points));
+
+        self
+    }
+
+    /// Add data points.
+    pub fn borrowed_points<'b, T: 'a>(&mut self, mut points: Points<T>) -> &mut Self
```
emilk, line 240: It would be nice if we could combine `owned_points` and `borrowed_points`.

emilk, line 240: These should have docstrings with examples, since there are now multiple layers of abstractions

crates/egui_plot/src/items/mod.rs
```diff
@@ -1020,6 +1120,209 @@ impl PlotItem for Points {
...
+                        let vertical = [tf(0.0, -1.0), tf(0.0, 1.0)];
+                        let diagonal1 = [tf(-frac_sqrt_3_2, 0.5), tf(frac_sqrt_3_2, -0.5)];
+                        let diagonal2 = [tf(-frac_sqrt_3_2, -0.5), tf(frac_sqrt_3_2, 0.5)];
+                        shapes.push(Shape::line_segment(vertical, default_stroke));
+                        shapes.push(Shape::line_segment(diagonal1, default_stroke));
+                        shapes.push(Shape::line_segment(diagonal2, default_stroke));
+                    }
+                }
+            });
+    }
```
emilk, line 1252: Is all this duplicated code? O.o

emilk (commented):
First of all, I appreciate all the effort you've put into this!

However, this is a lot of new complexity, so I want to make sure this is justified. I'm likely the one who is going to be maintaining it for the foreseeable future.

Whenever I've ran into plot performance issues, it's been on the tessellator side, not here. And the solution is usually to aggregate the data, i.e. downsample it before handing it to `egui_plot` (see for instance https://github.com/rerun-io/rerun/pull/4865).

So I'd be very interested in seeing some before/after flamegraph profiler screenshots, or an added benchmark to justify all this.

crates/egui_demo_lib/src/demo/plot_demo.rs
GunnarMorrigan: Will be fixed

crates/egui_plot/src/items/mod.rs
GunnarMorrigan: I can implement FromIterator for Points<T>?

> method `from_iter` can be confused for the standard trait method `std::iter::FromIterator::from_iter`
consider implementing the trait `std::iter::FromIterator` or choosing a less ambiguous method name
for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#should_implement_trait
`-W clippy::should-implement-trait` implied by `-W clippy::all`

crates/egui_plot/src/items/mod.rs
GunnarMorrigan: Yeah kinda

crates/egui_plot/src/plot_ui.rs
GunnarMorrigan: I agree that it would be nicer.

We need to require that everything impls PlotItem and this is either done through a generic type constraint or well defined types that impl PlotItem.

GunnarMorrigan:
> First of all, I appreciate all the effort you've put into this!
> 
> However, this is a lot of new complexity, so I want to make sure this is justified. I'm likely the one who is going to be maintaining it for the foreseeable future.
> 
> Whenever I've ran into plot performance issues, it's been on the tessellator side, not here. And the solution is usually to aggregate the data, i.e. downsample it before handing it to `egui_plot` (see for instance [rerun-io/rerun#4865](https://github.com/rerun-io/rerun/pull/4865)).
> 
> So I'd be very interested in seeing some before/after flamegraph profiler screenshots, or an added benchmark to justify all this.

In my profiling there were some three things in performance
- Tessellation
Vec the growth is amortized but if you know it will grow very large the moving takes a lot of time.
- Shape Vec
Idem
- PlotPoint construction

There is as I see it no easy way to benchmark anything in egui as `egui::__run_test_ui` does not actually do any work. This makes it extremely hard to perform any reliable performance profiling as you can't see when something slows down or speeds up if the benchmark is not the same every time.

I have a flamegraph where egui spends significant time in the shape vec reallocating which should be preventable:
image

In my [optimizeShapeVec](https://github.com/GunnarMorrigan/egui/tree/optimizeShapeVec) branch I made a simple function to count the shapes. Which should already help a bit.
This optimization is also more focused on egui_plot as the shape vec only grows so large in egui_plot due to the number of points that are drawn. 

Tesselation vec (the reserves are nice but it would be nicer if it could be done at once)
image

This is after improving plot points.
I will say I think there is indeed more gain to be had with shapes and tessellation then plot points but it is not insignificant.

This PR is a start to performance improvements. However, right now I feel like there is a requirement to invest a significant amount of additional time to make this or other PRs to increase performance land.
I really like egui and what you have created, but I also have limited free time.

Issue #3850, opened 2024-01-20
TAB key in TextEdit, fill the space with different sizes


rustbasic:
Feature request:
When pressing the TAB key in TextEdit,
fill the space with different sizes of whitespace based
on the position of the text cursor.

I have completed this feature, so I will submit a Pull Request soon.

Issue #3846, opened 2024-01-19
TextEdit : Shift + TAB


rustbasic:
TextEdit : Shift + TAB

Shift + TAB

UTF8 panic
and other BUG

PR #3857, opened 2024-01-22
Pull Request : #3321 #3846 #3850

rustbasic:
@emilk
Dear emilk.

I work finished, Pull Request.

* #3321

Changed :
crates/eframe/src/native/run.rs

The test results are normal.

* #3846

rev() is panic with UTF8

Not byte_index
It's char_index ( for UTF8 )

* #3850

let width = unicode_width::UnicodeWidthChar::width_cjk(c).unwrap_or(2);

Thank you, emilk.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,17 +1077,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "d3d12"
-version = "0.19.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e3d747f100290a1ca24b752186f61f6637e1deffe3bf6320de6fcb29510a307"
-dependencies = [
- "bitflags 2.4.0",
- "libloading 0.8.0",
- "winapi",
-]
-
-[[package]]
 name = "data-url"
 version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2998,12 +2987,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "range-alloc"
-version = "0.1.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9c8a99fddc9f0ba0a85884b8d14e3592853e787d581ca1816c91349b10e4eeab"
-
-[[package]]
 name = "raw-window-handle"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4277,12 +4260,10 @@ dependencies = [
  "android_system_properties",
  "arrayvec",
  "ash",
- "bit-set",
  "bitflags 2.4.0",
  "block",
  "cfg_aliases",
  "core-graphics-types",
- "d3d12",
  "glow",
  "glutin_wgl_sys",
  "gpu-alloc",
@@ -4300,7 +4281,6 @@ dependencies = [
  "once_cell",
  "parking_lot",
  "profiling",
- "range-alloc",
  "raw-window-handle 0.6.0",
  "renderdoc-sys",
  "rustc-hash",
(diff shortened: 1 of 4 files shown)


rustbasic:
@emilk
Dear emilk,

I'm having trouble with `cargo check errors` while making a Pull Request,
and I'm not sure why.
If there's nothing unusual in the source code,
please either merge it as is or feel free to make modifications as you see fit.

I've extensively tested it in the program I created,
so I believe there shouldn't be any bugs.

Thank you, emilk.

emilk:
please create one PR per issue to make it easier to review

crates/egui/src/text_selection/text_cursor_state.rs
```diff
@@ -300,18 +300,97 @@ pub fn find_line_start(text: &str, current_index: CCursor) -> CCursor {
...
+
+pub fn create_char_line_indexes(text: &str) -> Vec<usize> {
+    let mut line_indexes = vec![0];
 
-    match position {
-        Some(pos) => CCursor::new(current_index.index - pos),
-        None => CCursor::new(0),
+    let mut count = 0;
+    for (_i, c) in text.chars().enumerate() {
+        count += 1;
```
YgorSouza, line 314: Why not just use the `i` from the enumerate? Or just remove the `enumerate()` since it is not doing anything. Or do this:

```rs
    let line_indexes: Vec<_> = text
        .lines()
        .map(|line| line.chars().count())
        .scan(0, |total, next| {
            let index = *total;
            *total += next;
            Some(index)
        })
        .collect();
```

Is the vec supposed to always contain the last char index as well? Even if it is not a new line?

crates/egui/src/text_selection/text_cursor_state.rs
```diff
@@ -300,18 +300,97 @@ pub fn find_line_start(text: &str, current_index: CCursor) -> CCursor {
...
+}
+
+pub fn get_line_start_index(line_indexes: &mut [usize], index: usize) -> usize {
+    let (_current_line, line_start) = get_line_and_start_index(line_indexes, index);
+    line_start
+}
+
+pub fn get_line_and_start_index(line_indexes: &mut [usize], index: usize) -> (usize, usize) {
+    let mut current_line = 1;
+    for (i, line_start) in line_indexes.iter().enumerate() {
```
YgorSouza, line 331: This could use `partition_point`.

crates/egui/src/text_selection/text_cursor_state.rs
```diff
@@ -300,18 +300,97 @@ pub fn find_line_start(text: &str, current_index: CCursor) -> CCursor {
...
+    index: usize,
+) -> (usize, usize, usize) {
+    let line_start_index = get_line_start_index(line_indexes, index);
+
+    let total_lines = line_indexes.len() - 1;
+
+    let mut i: usize = line_start_index;
+    let mut current_column: usize = 1;
+    let mut total_columns: usize = 0;
+    while let Some(c) = text.chars().nth(i) {
```
YgorSouza, line 369: This has quadratic complexity, doesn't it? Each call to `nth()` is a for loop from the start of the string. You could do 

```rs
for c in text.chars().skip(line_start_index) {
```

Since you are already breaking on the newline. Otherwise you can get the index of the next line start and use `take()` to get the chars up to that index.

crates/egui/src/text_selection/text_cursor_state.rs
rustbasic: `'\r' and '\n'` should also be counted.
Saving the last index also reduces `panic or bug` occurrence when using arrays.

crates/egui/src/text_selection/text_cursor_state.rs
rustbasic: This is to find a column according to the width of UTF8 characters.

crates/egui/src/text_selection/text_cursor_state.rs
YgorSouza: Wouldn't it introduce bugs, though? You are saving the last index of the last line as if it were the first index of a new line.

Issue #3861, opened 2024-01-22
Glow/Android Only: Going directly from `eframe` to Home breaks the `eframe` on resume 


m-hugo:
now on master everything android work except for this very specific scenario:
- open app
- press Home button
- go back to the app
- it crashes

```rust
#[no_mangle]
fn android_main(app: AndroidApp) {
    let options = NativeOptions {
        event_loop_builder: Some(Box::new(move |builder| {
            builder.with_android_app(app);
        })),
        ..Default::default()
    };
    eframe::run_native(
        "My egui App",
        options,
        anything,
    )
}
```
```toml
eframe = { git = "https://github.com/emilk/egui", default-features = false, features = [
  "glow",
  "default_fonts",
  "android-native-activity"
] }
```
timeline:
- 0.23 both glow and wgpu work
- 0.24 everything is broken
- 0.25 glow works except for this bug, wgpu still broken
- #3847 wgpu works perfectly

PR #3867, opened 2024-01-22
Fix Android crash on resume with Glow backend

Garoven:
Addition for <https://github.com/emilk/egui/pull/3847>
In previous one i only fixed crash occurring with Wgpu backend. This fixes crash with Glow backend as well.
I only tested this change with android so most things i changed are behind ```#[cfg(target_os = "android")]```.

Both fixes are dirty thought. As <https://github.com/emilk/egui/pull/3172> says that "The root viewport is the original viewport, and cannot be closed without closing the application.". So they break rules i guess? But i can't think about better solution for now.

Closes <https://github.com/emilk/egui/issues/3861>.

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -435,7 +435,7 @@ impl WinitApp for GlowWinitApp {
                         #[cfg(target_os = "android")]
                         glutin.on_resume(self.native_options.viewport.clone(), event_loop)?;
                         #[cfg(not(target_os = "android"))]
-                        glutin.initialize_all_windows(event_loop)
+                        glutin.initialize_all_windows(event_loop);
                     }
 
                     running
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -430,13 +430,10 @@ impl WinitApp for GlowWinitApp {
 
                 let running = if let Some(running) = &mut self.running {
                     // Not the first resume event. Create all outstanding windows.
-                    {
-                        let glutin = &mut running.glutin.borrow_mut();
-                        #[cfg(target_os = "android")]
-                        glutin.on_resume(self.native_options.viewport.clone(), event_loop)?;
-                        #[cfg(not(target_os = "android"))]
-                        glutin.initialize_all_windows(event_loop);
-                    }
+                    running
+                        .glutin
+                        .borrow_mut()
+                        .initialize_all_windows(event_loop);
 
                     running
                 } else {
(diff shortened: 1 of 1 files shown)


MarijnS95 (changes requested):
Just like with https://github.com/emilk/egui/pull/3847, please don't put things behind `cfg`s that should function generically.

Garoven:
I actually think that without these cfg's it should work on all platforms but it's need to be tested
1. the on_resume fn only updates viewport if it is present
2. the only way to remove root viewport is to suspend the app, which means on all other platform that app should exit so it isn't a problem

Will change and test it tomorrow. Changing to draft for now

MarijnS95:
Thanks! Can you also test if they can be simplfied/removed again on `wgpu`?

I'm by no means an expert on `egui`, let alone `eframe`. All I can tell you is that `Resumed` will always be called once. On mobile platforms, it might be followed by a `Suspended`, followed by a `Resumed`, etc. You can assume that they will always alternate like that.

Any state that is implicitly/indirectly owned by a surface should be destroyed in `Suspended` and recreated in `Resumed`. If we follow the documentation comments, that should be all state in the `*WinitRunning` structs:

https://github.com/emilk/egui/blob/aa67d3180be2f1a65d71b304e342b23e93448476/crates/eframe/src/native/wgpu_integration.rs#L47-L58
https://github.com/emilk/egui/blob/aa67d3180be2f1a65d71b304e342b23e93448476/crates/eframe/src/native/glow_integration.rs#L69-L95

While at it, can you remove this comment?:

https://github.com/emilk/egui/blob/aa67d3180be2f1a65d71b304e342b23e93448476/crates/eframe/src/native/glow_integration.rs#L42-L52

It is no longer valid.

Finally, what about https://github.com/emilk/egui/pull/3676?

Garoven:
After digging deeper in codebase i have found that there is a way to fix this bug without dropping ROOT viewport. The same can be done with wgpu but don't know if i should do it in this pr.

m-hugo:
works fine

MarijnS95:
> i have found that there is a way to fix this bug without dropping ROOT viewport

I am not familiar enough with `eframe` to say, but viewports sound like a thing that should sustain a drawing surface disappearing (because the screen is temporarily blanked, or another app is in focus). In fact even the `winit::Window` should persist as it is currently not at all tied to the lifetime of an Android surface (and it shouldn't, it should be tied to an `Activity`, but everyone seems to wrongly assume that those are singletons). In `winit` it's just an empty shell that communicates with the fake `Activity` singleton.

Glad to have the code platform-agnostic. Can you retroactively apply the same to the `wgpu` backend?

Garoven:
I have a branch with wgpu changed already. But it requires to make bigger changes. I had to move some code around to make borrow checker work. Now it looks more similar to glow structure so it should be better. But i think it better to make another pr that refreshes wgpu integration(move\clean code + add comments) to catch up with glow then change Suspend/Resume cycle.

MarijnS95:
@Garoven sounds good, looking forward to see that PR and I agree that the `glow` structure looks more like what platform-agnostic `winit` integrations should be.

Indeed such changes are better served in a separate PR to lighten the review/testing burden :+1:

Garoven:
Now I think that everything is good so it's time to re-open this pr.

emilk (approved):
That was a very pleasant diff!

Issue #3856, opened 2024-01-22
Updating `web-time` to `1.0.0`?


torokati44:
A bot tried to bump our `web-time` dependency to `1.0.0`, but ran into this error:

```
cargo update  --package web-time --precise 1.0.0
    Updating crates.io index
error: failed to select a version for the requirement `web-time = "^0.2"`
candidate versions found which didn't match: 1.0.0
location searched: crates.io index
required by package `egui-winit v0.25.0 (https://github.com/emilk/egui?rev=d319489479c371b15e6419d470551dae5d647396#d3194894)`
    ... which satisfies dependency `egui-winit = "^0.25.0"` (locked to 0.25.0) of package `ruffle_desktop v0.1.0 (/home/attila/projects/ruffle/desktop)`
perhaps a crate was updated and forgotten to be re-vendored?
```

See: https://github.com/ruffle-rs/ruffle/pull/14892

The changelog from `0.2` to `1.0` is close to empty, so this looks like just a marker release: https://github.com/daxpedda/web-time/blob/main/CHANGELOG.md

Therefore, it's not critical to update, and won't fix anything for us, it just "would be nice".
This also means that updating to it should be trivial almost everywhere.

Is `egui-winit` going to update sometime?

torokati44:
Feel free to close this issue too, it will resolve itself with some future winit update.

torokati44:
There have been a couple `winit` releases with a `web-time` dependency of 1.0, IIUC.
So, bump on this.

torokati44:
No, no, no, scratch that. It'll only be in `winit` `v0.30`. 😫
Sorry for the spam.

torokati44:
> No, no, no, scratch that. It'll only be in winit v0.30. 😫

... which is now definitely out, for real! :tada:

PR #3872, opened 2024-01-23
chore: Update web-time to 1.0

torokati44:
Closes https://github.com/emilk/egui/issues/3856.


torokati44:
It seems like `web-time` `0.2` is still pulled in by `winit` `0.29.7 ` on web.
Can I make `cargo-deny` ignore that until https://github.com/rust-windowing/winit/pull/3392 gets into a release?

emilk:
I rather wait for the next winit patch release rather than duplicating dependencies

PR #3874, opened 2024-01-23
Add `WgpuConfiguration::desired_maximum_frame_latency`

emilk:
Setting `desired_maximum_frame_latency` to a low value should theoretically lead to lower latency in winit apps using `egui-wgpu` (e.g. in `eframe` with `wgpu` backend).

* Replaces https://github.com/emilk/egui/pull/3714
* See also https://github.com/gfx-rs/wgpu/pull/4899

----

It seems like `desired_maximum_frame_latency` has no effect on my Mac. I lowered my monitor refresh-rate to 30Hz to test, and can see no difference between `desired_maximum_frame_latency` of `0` or `3`.

Before when experimenting with changing the global `DESIRED_NUM_FRAMES` in `wgpu` I saw a huge difference, so I wonder what has changed.

I verified that `set_maximum_drawable_count` is being called with either `1` or `2`, but I perceive no difference between the two.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4219,9 +4219,9 @@ checksum = "14247bb57be4f377dfb94c72830b8ce8fc6beac03cf4bf7b9732eadd414123fc"
 
 [[package]]
 name = "wgpu"
-version = "0.19.0"
+version = "0.19.1"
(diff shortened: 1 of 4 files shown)


Wumpf (approved):
what could possibly go wrong now 😄 🚢

PR #3875, opened 2024-01-23
`egui-wgpu`: turn off the default features of `wgpu`

emilk:
This makes all `wgpu` features opt-in for `egui-wgpu` users.

For `eframe`, I opted to enable some `wgpu` features so users can still use `eframe` without having to also opt-in to extra ewgpu features (eframe is batteries-included).


cwfitzgerald (approved):
Awesome!

Wumpf (approved):
nice

PR #3878, opened 2024-01-24
Fix: panic when pressing Shift + TAB in a TextEdit

rustbasic:
panic when pressing Shift + TAB in a TextEdit #3846 

rev() is panic with UTF8

Not byte_index
It's char_index ( for UTF8 )


crates/egui/src/text_selection/text_cursor_state.rs
```diff
@@ -302,18 +302,42 @@ pub fn find_line_start(text: &str, current_index: CCursor) -> CCursor {
...
 
-    let position = text
-        .chars()
-        .rev()
-        .skip(chars_count - current_index.index)
-        .position(|x| x == '\n');
+    CCursor::new(line_start_index)
+}
+
+pub fn create_char_line_indexes(text: &str) -> Vec<usize> {
```
emilk, line 311: If you use `CCursor` instead of `usize`, the code becomes self-documenting

```suggestion
pub fn create_char_line_indexes(text: &str) -> Vec<CCursor> {
```

crates/egui/src/text_selection/text_cursor_state.rs
```diff
@@ -302,18 +302,42 @@ pub fn find_line_start(text: &str, current_index: CCursor) -> CCursor {
...
     // We need to know the char index to be able to correctly set the cursor
     // later.
-    let chars_count = text.chars().count();
+    let mut char_line_indexes: Vec<usize> = create_char_line_indexes(text);
+    let line_start_index = get_line_start_index(&mut char_line_indexes, current_index.index);
 
-    let position = text
-        .chars()
-        .rev()
-        .skip(chars_count - current_index.index)
```
emilk, line 310: What was wrong with the old code?

Maybe it just needs to handle the `char_count < current_index.index` case?

rustbasic: Since UTF-8 characters use 1 to 4 bytes,
using rev() can cause a panic and terminate the program.
You can test it by copying various UTF-8 characters
and using them in multiline().

emilk: `.rev()` doesn't panic: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1647ea03b9ad85c69d72ee2189b73fc0

crates/egui/src/text_selection/text_cursor_state.rs
rustbasic: If CCursor is used instead of usize,                                                  
it seems handling functions might become difficult,                                   
so I'm not sure if it's a good approach.                                              
I will respect your final opinion.                                                    
                                                                                      
By the way, the create_char_line_indexes() function is also used                      
in the next Pull Request related to TAB.

crates/egui/src/text_selection/text_cursor_state.rs
rustbasic: Mix the UTF8 characters with English, fill the screen ( with them using multiline() ), and pressing Shift + TAB here and there.

rustbasic: > Mix the UTF8 characters with English, fill the screen `( with them using multiline(), egui_demo_app )`, and pressing Shift + TAB here and there.

rustbasic: It's sample TEXT.                                                                     
fill the screen.                                                                      
"    A äö 日本語 A äö 日本語"

rustbasic: Mix the UTF8 characters with English, fill the screen ( with them using multiline() ), and pressing Shift + TAB here and there.

rustbasic: It seems to happen when transitioning from using char index
to using an array of byte indexes.

There are two more instances of `rev()` in `text_cursor_state.rs`,
fortunately, those two do not cause panics
but instead incorrectly select a block where UTF8 characters exist.
I will register this as a separate issue later.

rustbasic: @emilk
Have you noticed the panic occurring?

emilk: I don't have time for this right now. I believe you that here is a panic in egui, but I don't think the panic is in the call to `.rev()`, and so I think a much simpler solution to your problem can be found.

Run in debug mode and RUST_BACKTRACE=1 and paste the stack-trace to the actual panic.

PR #3879, opened 2024-01-24
TAB key in TextEdit, fill the space with different sizes

rustbasic:
https://github.com/emilk/egui/issues/3850

let width = unicode_width::UnicodeWidthChar::width_cjk(c).unwrap_or(2);

### DIFF
--- a/crates/egui/Cargo.toml
+++ b/crates/egui/Cargo.toml
@@ -96,3 +96,4 @@ log = { version = "0.4", optional = true, features = ["std"] }
 puffin = { workspace = true, optional = true }
 ron = { version = "0.8", optional = true }
 serde = { version = "1", optional = true, features = ["derive", "rc"] }
+unicode-width = "0.1.11"
--- a/crates/egui/src/text_selection/text_cursor_state.rs
+++ b/crates/egui/src/text_selection/text_cursor_state.rs
@@ -314,6 +314,93 @@ pub fn find_line_start(text: &str, current_index: CCursor) -> CCursor {
     }
 }
 
+pub fn create_char_line_indexes(text: &str) -> Vec<usize> {
+    let mut line_indexes = vec![0];
+
+    let mut count = 0;
+    for (_i, c) in text.chars().enumerate() {
+        count += 1;
+        if c == '\n' {
+            line_indexes.push(count);
+        }
+    }
+    line_indexes.push(count);
+
+    line_indexes
+}
+
+pub fn get_line_start_index(line_indexes: &mut [usize], index: usize) -> usize {
+    let (_current_line, line_start) = get_line_and_start_index(line_indexes, index);
+    line_start
+}
+
+pub fn get_line_and_start_index(line_indexes: &mut [usize], index: usize) -> (usize, usize) {
+    let mut current_line = 1;
+    for (i, line_start) in line_indexes.iter().enumerate() {
+        current_line = i;
+        if *line_start > index {
+            break;
+        }
+    }
+
+    (current_line, line_indexes[current_line - 1])
+}
+
+pub fn get_current_column(text: &str, line_indexes: &mut [usize], index: usize) -> usize {
+    let (current_column, _total_column) =
+        get_current_column_total_column(text, line_indexes, index);
+    current_column
+}
+
+pub fn get_current_column_total_column(
+    text: &str,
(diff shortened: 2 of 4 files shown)


crates/egui/src/text_selection/text_cursor_state.rs
```diff
@@ -302,18 +302,108 @@ pub fn find_line_start(text: &str, current_index: CCursor) -> CCursor {
...
-    match position {
-        Some(pos) => CCursor::new(current_index.index - pos),
-        None => CCursor::new(0),
+pub fn get_current_column(text: &str, line_indexes: &mut [usize], index: usize) -> usize {
+    let (current_column, _total_column) =
+        get_current_column_total_column(text, line_indexes, index);
+    current_column
+}
+
+pub fn get_current_column_total_column(
```
emilk, line 349: you shouldn't be using `get_` everywhere

https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

crates/egui/src/text_selection/text_cursor_state.rs
```diff
@@ -302,18 +302,108 @@ pub fn find_line_start(text: &str, current_index: CCursor) -> CCursor {
...
+        if c == '\n' {
+            line_indexes.push(count);
+        }
+    }
+    line_indexes.push(count);
+
+    line_indexes
+}
+
+pub fn get_line_start_index(line_indexes: &mut [usize], index: usize) -> usize {
```
emilk, line 326: There is a lot of naked `usize`. If you use `CCursor` you can make it clear wether or not something is a byte offset or a character offset

emilk (changes requested):
Please try to make your code more readable with:

* better names
* documentation if names are not enough
* strong typing when it makes sense

crates/egui/Cargo.toml
```diff
@@ -98,3 +98,4 @@ log = { workspace = true, optional = true }
 puffin = { workspace = true, optional = true }
 ron = { version = "0.8", optional = true }
 serde = { version = "1", optional = true, features = ["derive", "rc"] }
+unicode-width = "0.1.11"
```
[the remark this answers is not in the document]
rustbasic: > Can we avoid adding this dependency? It's not a huge one, but I wonder how much larger the .wasm becomes with these tables: https://github.com/unicode-rs/unicode-width/blob/master/src/tables.rs

I don't think this dependency can be avoided. These are codes so they will not be very large and are essential for countries that use UTF8 characters. All pictograms are also affected by this.

crates/egui/src/text_selection/text_cursor_state.rs
rustbasic: With this, we can easily find out the current Colume and line within TextEdit. I'm using this, and I'm sure others might want to use it too.

crates/egui/src/text_selection/text_cursor_state.rs
```diff
@@ -302,18 +302,108 @@ pub fn find_line_start(text: &str, current_index: CCursor) -> CCursor {
...
 
-    let position = text
-        .chars()
-        .rev()
-        .skip(chars_count - current_index.index)
-        .position(|x| x == '\n');
+    CCursor::new(line_start_index)
+}
+
+pub fn create_char_line_indexes(text: &str) -> Vec<usize> {
```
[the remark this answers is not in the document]
rustbasic: I'm not from an English-speaking country, so it's hard for me to choose even a single word.
I would appreciate it if `emilk` could make a suggestion.

crates/egui/src/text_selection/text_cursor_state.rs
rustbasic: > you shouldn't be using `get_` everywhere
> 
> https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

I'm not from an English-speaking country, so it's hard for me to choose even a single word.
I would appreciate it if `emilk` could make a suggestion.

crates/egui/src/text_selection/text_cursor_state.rs
emilk: does it return a byte index or a char index? The name doesn't tell me, neither does the return type

Issue #3882, opened 2024-01-24
Improvements to make drag-and-drop easier


abey79:
- #3876 added `men.dragged_id()`, but there is no way to get the value of to an `egui::Id` (which could then be converted to an internal Id).
- It would be nice to have `ctx().anything_being_decidedly_dragged()` (equivalent to ` ui.memory(|mem| mem.is_anything_being_dragged()) && ui.input(|i| i.pointer.is_decidedly_dragging())`
- It would be nice to have `ctx().drag_stopped()`

emilk:
We should take a look at the ImGui API for drag-and-drop

crumblingstatue:
I want to mention [egui_dnd](https://docs.rs/egui_dnd/latest/egui_dnd/) just as a baseline to compare an egui-builtin solution against.

It only seems to support fairly simple layouts.

I would like something that supports drag and drop between different uis for maximum flexibility.

An example use case of mine would be to drag tree nodes in a tree.

For example move `a-1` to `b` in this tree ui made out of `CollapsingHeader`s.
![image](https://github.com/emilk/egui/assets/1521976/bd5e3159-e4c3-424d-a14a-6e8749cde22b)

abey79:
@crumblingstatue I think a proper high-level drag-and-drop API in egui is still a little far off due to the difficulty to properly address the general case(s). This issue is more about the little things that would make lower-level implementation easier.

As for the tree drag-and-drop, I just built a demo of that in the rerun repository: https://github.com/rerun-io/rerun/pull/4879. I plan on making a proper egui example out of it.

abey79:
@crumblingstatue https://github.com/emilk/egui/pull/3903

PR #3887, opened 2024-01-25
Add drag-and-drop APIs with payloads storage

emilk:
* Closes https://github.com/emilk/egui/issues/3882

This adds several methods to make drag-and-drop more ergonomic in egui.

In particular, egui can now keep track of _what_ is being dragged for you (the _payload_).

Low-level:
* `egui::DragAndDrop` hold the payload during a drag

Mid-level:
* `Response::dnd_set_drag_payload` sets it for drag-sources
* `Response::dnd_hover_payload` and `Response::dnd_release_payload` reads it for drop-targets

High-level:
* `ui.dnd_drag_source`: make a widget draggable
* `ui.dnd_drop_zone`: a container where things can be dropped

The drag-and-drop demo is now a lot simpler:

https://github.com/emilk/egui/blob/emilk/drag-and-drop/crates/egui_demo_lib/src/demo/drag_and_drop.rs

### DIFF
--- a/crates/egui/src/drag_and_drop.rs
+++ b/crates/egui/src/drag_and_drop.rs
@@ -1,18 +1,17 @@
-//! Support for drag-and-drop in egui.
-//!
-//! This module implements the low-level drag-and-drop API with support for manually saving, retrieving, and clearing
-//! drag-and-drop payloads. This enables the maximum flexibility for complex drag-and-drop interactions.
-//!
-//! For a higher-level API, see:
-//! - [`crate::Ui::dnd_drag_source`],
-//! - [`crate::Ui::dnd_drop_zone`],
-//! - and the drag-related functions in [`crate::Response`].
-
 use std::{any::Any, sync::Arc};
 
 use crate::{Context, CursorIcon, Id};
 
-/// Helpers for drag-and-drop in egui.
+/// Tracking of drag-and-drop payload.
+///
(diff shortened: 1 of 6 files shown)


crates/egui/src/drag_and_drop.rs
```diff
@@ -0,0 +1,126 @@
+//! Support for drag-and-drop in egui.
+//!
+//! This module implements the low-level drag-and-drop API with support for manually saving, retrieving, and clearing
+//! drag-and-drop payloads. This enables the maximum flexibility for complex drag-and-drop interactions.
+//!
+//! For a higher-level API, see:
+//! - [`crate::Ui::dnd_drag_source`],
+//! - [`crate::Ui::dnd_drop_zone`],
+//! - and the drag-related functions in [`crate::Response`].
```
abey79, line 9: Is that the right place for this doc?

abey79 (approved):
I'm only able to use the low-level API in rerun, because of all the geometry testing I must run on my side, but the ability to store and retrieve the dnd payload (along with using its existence as a marker of on on-going drag) makes the code much cleaner and easy to reason with.

emilk:
@abey79 Oh, you can't even use the `Response` API? :(

abey79:
@emilk The response API works well when moving boxes inside boxes, but not so well when moving boxes _in-between_ boxes :)

PR #3892, opened 2024-01-26
Fix menubar macOS

hacknus:
I noticed that on macOS the menubar is not looking right anymore:
Bildschirmfoto 2024-01-26 um 07 43 21
This PR applies some changes, it would further also allow for a "Settings" window to be implemented, if there would be an egui container for that (we would only need to create a new item in `app_icons.rs`).
The menu bar now looks like this:
Bildschirmfoto 2024-01-26 um 09 04 00

TODO:
- I have not yet managed to replicate the separator lines

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ NOTE: [`epaint`](crates/epaint/CHANGELOG.md), [`egui_plot`](crates/egui_plot/CHA
 This file is updated upon each release.
 Changes since the last release can be found at <https://github.com/emilk/egui/compare/latest...HEAD> or by running the `scripts/generate_changelog.py` script.
 
+# Unreleased
+* Fixed menubar in macOS
 
 ## 0.25.0 - 2024-01-08 - Better keyboard input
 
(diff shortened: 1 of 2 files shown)


emilk:
Wait, what was the problem with the menu bar?

hacknus:
It always says NSMenuItem instead of the name of the application

emilk:
Oh, weird! On my macOS Sonoma 14.2.1 it works I get the name of the app:

image

hacknus:
That is indeed very interesting! On Venture it also worked for me, but once I updated to Sonoma (currently 14.1.2) I was noticing this error...) I'll update to 14.2.1 and check again.

hacknus:
I have updated to the latest version of macOS and this has fixed the issue. This is odd indeed...
On the bright side, I now learned how one could implement a "Settings" menu in macOS (cmd + ,) is this something that is planned as an extension for egui? or not really because this would be too platform specific?

emilk:
It's probably too platform specific too be part of eframe, but it would be good if eframe users could still add it themselves if they want too.

emilk:
Closing: eframe moved from `cocoa`/`objc` to `objc2` in #4395, so this no longer applies. A macOS menubar would need a fresh PR on top of objc2. Thanks for the PR!

PR #3894, opened 2024-01-26
ui.label() : `truncate` default value is `true`

rustbasic:
ui.label() :
`truncate` default value is `true` to prevent egui::window size from increasing.


emilk:
No, the default when showing some text should be for egui to show all of the text.

PR #3895, opened 2024-01-26
Re-try adapter creation if creation with WebGPU enabled fails and WebGL was enabled.

Wumpf:
* Filed this in relation to the changes here https://github.com/gfx-rs/wgpu/issues/5142
* Fixes https://github.com/rerun-io/rerun/issues/4915

Draft until fully confirmed this works on Linux Chrome

### DIFF
--- a/crates/eframe/src/web/web_painter_wgpu.rs
+++ b/crates/eframe/src/web/web_painter_wgpu.rs
@@ -157,7 +157,7 @@ impl WebPainterWgpu {
                     instance = wgpu::Instance::new(wgpu::InstanceDescriptor {
                         backends,
                         ..Default::default()
-                    })
+                    });
                 } else {
                     return Err(
                         "Failed to create WebGPU adapter and WebGL was not enabled.".to_owned()
--- a/crates/eframe/src/web/web_painter_wgpu.rs
+++ b/crates/eframe/src/web/web_painter_wgpu.rs
@@ -83,12 +83,12 @@ impl WebPainterWgpu {
         })
     }
 
-    pub async fn init_wgpu_and_create_render_state(
-        mut backends: wgpu::Backends,
-        wgpu_options: &egui_wgpu::WgpuConfiguration,
-        depth_format: Option<wgpu::TextureFormat>,
-        canvas: &HtmlCanvasElement,
-    ) -> Result<(RenderState, wgpu::Surface<'static>), String> {
+    #[allow(unused)] // only used if `wgpu` is the only active feature.
+    pub async fn new(canvas_id: &str, options: &WebOptions) -> Result<Self, String> {
+        log::debug!("Creating wgpu painter");
+
+        let mut backends = options.wgpu_options.supported_backends;
+
         // Don't try WebGPU if we're not in a secure context.
         if backends.contains(wgpu::Backends::BROWSER_WEBGPU) {
             let is_secure_context = web_sys::window().map_or(false, |w| w.is_secure_context());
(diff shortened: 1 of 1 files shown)


crates/eframe/src/web/web_painter_wgpu.rs
```diff
@@ -83,46 +83,95 @@ impl WebPainterWgpu {
...
+                });
+                let surface = instance
+                    .create_surface(wgpu::SurfaceTarget::Canvas(canvas.clone()))
+                    .map_err(|err| format!("failed to create wgpu surface: {err}"))?;
+                let render_state =
+                    RenderState::create(wgpu_options, &instance, &surface, depth_format, 1)
+                        .await
+                        .map_err(|err| err.to_string())?;
+
+                return Ok((render_state, surface));
```
emilk, line 145: isn't it easier to just recurse here?

```suggestion
                return Sellf::init_wgpu_and_create_render_state(
                    wgpu::Backends::GL,
                    wgpu_options,
                    depth_format,
                    canvas,
                );
                    
```

crates/eframe/src/web/web_painter_wgpu.rs
```diff
@@ -83,46 +83,95 @@ impl WebPainterWgpu {
...
+                //
+                // Since WebGPU is handled in a special way in wgpu, we have to recreate the instance
+                // if we instead want to try with WebGL.
+                //
+                // Note that we also might end up here if wgpu didn't detect WebGPU support,
+                // moved on to create a wgpu-core instance and then failed to create a WebGL adapter.
+                // To detect this we'd need to check what kind of instance wgpu created,
+                // but this is currently not easily possible. See https://github.com/gfx-rs/wgpu/issues/5142
+                // In that case we'd be trying the same thing again which isn't ideal, but we're in a
+                // pretty bad situation anyway.
```
emilk, line 132: We should log what is happening

crates/eframe/src/web/web_painter_wgpu.rs
```diff
@@ -83,46 +83,95 @@ impl WebPainterWgpu {
...
-        let canvas = super::canvas_element_or_die(canvas_id);
-
         let surface = instance
             .create_surface(wgpu::SurfaceTarget::Canvas(canvas.clone()))
             .map_err(|err| format!("failed to create wgpu surface: {err}"))?;
+        let render_state_result =
+            RenderState::create(wgpu_options, &instance, &surface, depth_format, 1).await;
+
+        if let Err(egui_wgpu::WgpuError::NoSuitableAdapterFound) = &render_state_result {
+            if backends == wgpu::Backends::BROWSER_WEBGPU | wgpu::Backends::GL {
```
emilk, line 120: nit: I think this logic is slightly clearer

```suggestion
            if backends.contains(Backends::BROWSER_WEBGPU) && backends.contains(Backends::GL) {
```

crates/eframe/src/web/web_painter_wgpu.rs
Wumpf: I wanted to do that originally but async functions are apparently not allowed to recurse. After a bit back and forth I figured it's easiest to just copy & paste those few lines

crates/eframe/src/web/web_painter_wgpu.rs
Wumpf: ... and might actually be necessary if somieone also sets other backends that are not for web!

Wumpf:
confirmed working now by @jleibs

PR #3905, opened 2024-01-27
Refactor response bools to bitflags

### LINKED ISSUE #3862: Performance: replace bools in `Response` with bit-sets
`Response` is 88 bytes large. Around 26 bytes of that is because of as many bools, which could be combines into a few bit-sets instead of 4 bytes.

This also includes `Sense` which has the same problem.

We also have several; `[bool; NUM_POINTER_BUTTONS]` which could be replaced with a bit-set each (`PointerButtons` or something).

Ideally we could solve this in an ergonomic way without any additional additional dependencies.

BSFishy:
This PR focuses primarily on adding the `bitflags` dependency and converting the immediate `bool`s in `Response` into bitflags. I'm mainly looking for style feedback before diving into the `[bool; NUM_POINTER_BUTTONS]` and `Sense` bitflags.

Tested manually with some basic sanity testing as well as `cargo test`.

Partially fixes #3862

### DIFF
--- a/crates/egui/Cargo.toml
+++ b/crates/egui/Cargo.toml
@@ -92,6 +92,7 @@ backtrace = { version = "0.3", optional = true }
 ## Enable this when generating docs.
 document-features = { version = "0.2", optional = true }
 
+bitflags = "2.4"
 log = { version = "0.4", optional = true, features = ["std"] }
 puffin = { workspace = true, optional = true }
 ron = { version = "0.8", optional = true }
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -1065,7 +1065,7 @@ impl Context {
                                     interaction.drag_id = Some(id);
                                     interaction.drag_is_window = false;
                                     memory.set_window_interaction(None); // HACK: stop moving windows (if any)
-                                    response
+                                    res
                                         .state
                                         .insert(ResponseState::IS_POINTER_BUTTON_DOWN_ON);
 
@@ -1082,13 +1082,10 @@ impl Context {
                         }
 
                         PointerEvent::Released { click, button } => {
-                            res.drag_released = res.dragged;
-                            res.dragged = false;
                             res
                                 .state
-                                .set(ResponseState::DRAG_RELEASED, response.dragged());
+                                .set(ResponseState::DRAG_RELEASED, res.dragged());
                             res.state.remove(ResponseState::DRAGGED);
-
 
                             if sense.click && res.hovered() && res.is_pointer_button_down_on() {
                                 if let Some(click) = click {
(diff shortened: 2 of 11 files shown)


crates/egui/src/response.rs
```diff
@@ -6,6 +6,41 @@ use crate::{
 
 // ----------------------------------------------------------------------------
 
+bitflags::bitflags! {
+    /// The state of a widget.
```
BSFishy, line 10: I have no idea what would be good docs for this struct. Feedback would be appreciated :)

crates/egui/src/response.rs
```diff
@@ -6,6 +6,41 @@ use crate::{
 
 // ----------------------------------------------------------------------------
 
+bitflags::bitflags! {
```
BSFishy, line 9: Is this the right place to put this struct?

crates/egui/src/context.rs
```diff
@@ -1066,28 +1065,31 @@ impl Context {
...
+                            res
+                                .state
+                                .set(ResponseState::DRAG_RELEASED, res.dragged());
+                            res.state.remove(ResponseState::DRAGGED);
 
-                            if sense.click && res.hovered && res.is_pointer_button_down_on {
+                            if sense.click && res.hovered() && res.is_pointer_button_down_on() {
                                 if let Some(click) = click {
-                                    let clicked = res.hovered && res.is_pointer_button_down_on;
+                                    let clicked = res.hovered() && res.is_pointer_button_down_on();
```
BSFishy, line 1092: Since clicked now has more complexity associated with it (performing bitwise operations to check if the bitflags contain those 2), should it be moved out of the 2 ifs and just reused? Or is the added cycles negligible to performance?

BSFishy:
Also, should `bitflags` be added to [the Dependencies section of the README](https://github.com/emilk/egui?tab=readme-ov-file#dependencies)?

crates/egui/src/context.rs
emilk: The optimizer should take care of it for you if `fn hovered` is marked `#[inline]`

emilk:
> Also, should `bitflags` be added to [the Dependencies section of the README](https://github.com/emilk/egui?tab=readme-ov-file#dependencies)?

Yes

crates/egui/Cargo.toml
```diff
@@ -92,6 +92,7 @@ backtrace = { version = "0.3", optional = true }
 ## Enable this when generating docs.
 document-features = { version = "0.2", optional = true }
 
+bitflags = "2.4"
```
emilk, line 95: This should move up. It is currently under the `#! ### Optional dependencies` heading

emilk (commented):
looking good at a quick glance!

emilk:
* Closed by https://github.com/emilk/egui/pull/5556

PR #3906, opened 2024-01-27
Add layer transforms, interaction in layer

### LINKED ISSUE #1811: A panning / zooming container
I work on the [egui_node_graph](https://github.com/setzer22/egui_node_graph) library, which can be used to create node-graph based applications in egui. One particular feature of these kind of applications is being able to zoom and pan the graph view, as showcased in the gif below. To this date, I don't think that's possible to implement in egui in a portable way, but I managed to make it work for a specific integration (winit + wgpu) and a bunch of hacks.

I believe this can also be an interesting feature in other kinds of visualization-heavy applications: Right now there are things like the Plot view, which have this behavior available, but being able to use the full power of egui in a Zoomable / Pannable container feels like it would be the cleanest solution in many situations.

It would be great if egui could offer a portable solution to create some sort of container in which you can apply pan and zoom. Tentatively, the API could look something like this?:
```rust
egui::PanZoomContainer::new("graph_area")
    .with_pan(current_pan)
    .with_zoom(current_zoom)
    .show(|ui| {
        // Here, you get a `ui` disconnected from the parent, with an infinite canvas to draw on.
    });
```

The current alternative, which is working for me but is highly non-portable and requires a bunch of patches on top of `egui`. The core idea is that I build two egui contexts, the one with the graph is rendered into a texture, and then that texture is drawn in the "parent" context as an egui::Image. This requires lots of messing around with the `pixels_per_point`, the offsets of the nodes and the quads that are being sent to wgpu (in the integration) in a very precise way to prevent visual glitches. I also intercept `winit` events that go into the child instance to offset mouse positions and such: Not fun at all, as one might imagine :sweat_smile: Nothing would make me happier than being able to throw away all this code.

I'm not sure I know enough about egui internals to see how a *good* implementation for this might look, but I assume it's possible, since egui is doing the rasterization: It would need to make sure it scales and offsets any meshes inside the Pan/Zoom area. It should also make sure a high-resolution texture with the fonts is available, since scaling up the fonts might look blurry otherwise.

**Additional context**
egui_zoom

Tweoss:
⚠️ Removes `Context::translate_layer`, replacing it with a sticky `set_transform_layer`

Adds the capability to scale layers.
Allows interaction with scaled and transformed widgets inside transformed layers.

I've also added a demo of how to have zooming and panning in a window (see the video below).

This probably closes #1811. Having a panning and zooming container would just be creating a new
`Area` with a new id, and applying zooming and panning with `ctx.transform_layer`. 

I've run the github workflow scripts in my repository, so hopefully the formatting and `cargo cranky` is satisfied. 

I'm not sure if all call sites where transforms would be relevant have been handled. This might also be missing are transforming clipping rects, but I'm not sure where / how to accomplish that. In the demo, the clipping rect is transformed to match, which seems to work. 

https://github.com/emilk/egui/assets/70821802/77e7e743-cdfe-402f-86e3-7744b3ee7b0f


crates/egui/src/memory.rs
```diff
@@ -151,6 +155,32 @@ impl FocusDirection {
     }
 }
 
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "persistence", derive(serde::Deserialize, serde::Serialize))]
+#[cfg_attr(feature = "persistence", serde(default))]
+pub struct LayerTransform {
+    pub translation: Vec2,
+    pub scale: f32,
```
emilk, line 163: The struct and fields needs docstrings; especially what order scale and translation are applied in

crates/egui/src/memory.rs
```diff
@@ -151,6 +155,32 @@ impl FocusDirection {
...
+}
+
+impl Default for LayerTransform {
+    fn default() -> Self {
```
emilk, line 167:
```suggestion
    #[inline]
    fn default() -> Self {
```

crates/egui/src/memory.rs
```diff
@@ -151,6 +155,32 @@ impl FocusDirection {
...
+        Self {
+            translation: Vec2::default(),
+            scale: 1.0,
+        }
+    }
+}
+
+impl LayerTransform {
+    pub(crate) fn apply(&self, rect: Rect) -> Rect {
```
emilk, line 176: we should name it for what it transforms, so we can also transform positions

```suggestion
    #[inline]
    pub fn mul_rect(&self, rect: Rect) -> Rect {
```

crates/egui_demo_lib/src/demo/pan_zoom.rs
```diff
@@ -0,0 +1,106 @@
+#[derive(Clone, Default, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub struct PanZoom {
+    pan: egui::Vec2,
+    zoom: f32,
```
emilk, line 5:
```suggestion
    t: LayerTransform,
```

crates/egui_demo_lib/src/demo/pan_zoom.rs
```diff
@@ -0,0 +1,106 @@
...
+
+        // Plot-like reset
+        if response.double_clicked() {
+            self.zoom = 1.0;
+            self.pan = egui::Vec2::ZERO;
+        }
+
+        if let Some(pointer) = ui.ctx().input(|i| i.pointer.hover_pos()) {
+            // Ignore if some other widget is covering this container.
+            if response.rect.contains(pointer) {
```
emilk, line 46:
```suggestion
        if response.hovered() {
```

crates/egui_demo_lib/src/demo/pan_zoom.rs
```diff
@@ -0,0 +1,106 @@
...
+                let original_zoom = self.zoom;
+                self.zoom *= ui.ctx().input(|i| i.zoom_delta());
+                let delta = pointer / self.zoom - pointer / original_zoom;
+                self.pan += delta;
+
+                // Keep mouse centered.
+                self.pan += ui.ctx().input(|i| i.raw_scroll_delta) / self.zoom;
```
emilk, line 53: why not `smooth_scroll_delta`?

crates/egui_demo_lib/src/demo/pan_zoom.rs
```diff
@@ -0,0 +1,106 @@
...
+                current_size.right_bottom() + egui::Vec2::new(-10.0, -10.0),
+                "right bottom :D",
+            ),
+            (
+                current_size.right_top() + egui::Vec2::new(-10.0, 10.0),
+                "right top ):",
+            ),
+        ]
+        .iter()
+        .map(|(pos, msg)| {
```
emilk, line 77: this would be more readable as normal `for`-loop imho

crates/epaint/src/mesh.rs
```diff
@@ -269,9 +269,10 @@ impl Mesh {
     }
 
     /// Translate location by this much, in-place
-    pub fn translate(&mut self, delta: Vec2) {
+    pub fn transform(&mut self, delta: Vec2, scale: f32) {
         for v in &mut self.vertices {
             v.pos += delta;
+            v.pos = (v.pos.to_vec2() * scale).to_pos2();
```
emilk, line 275: This is another reason to put the transform in `emath` instead

```suggestion
    pub fn transform(&mut self, t: &TSTransform) {
        for v in &mut self.vertices {
            v.pos = t * v.pos;
```

crates/egui/src/context.rs
```diff
@@ -2071,9 +2071,14 @@ impl Context {
     /// Move all the graphics at the given layer.
     ///
     /// Can be used to implement drag-and-drop (see relevant demo).
-    pub fn translate_layer(&self, layer_id: LayerId, delta: Vec2) {
+    pub fn transform_layer(&self, layer_id: LayerId, delta: Vec2, scale: f32) {
```
emilk, line 2074: This method used to be only an immediate thing, applying a transform to shapes created this frame.
Now it _also_ stores the transform in memory, applying it to all interaction, but _not_ to future paint shapes. This is not great.

We need to reconsider this API.

I think it would be ideal if we call it `set_layer_transform` instead. It can store the transform in memory, and then we apply the transform to the shapes during `GraphicLayers::drain`.

crates/egui_demo_lib/src/demo/pan_zoom.rs
Tweoss: Ah, I was trying to match the feel of blender's zooming in and out. I guess `smooth_scroll_delta` is better.

Tweoss:
@emilk Thank you for all the detailed comments and advice! I've taken some time and (hopefully) addressed all of them, could you take another look?

crates/epaint/src/shape.rs
```diff
@@ -356,48 +356,56 @@ impl Shape {
...
-                rect_shape.rect = rect_shape.rect.translate(delta);
+                rect_shape.rect = transform * rect_shape.rect;
             }
             Self::Text(text_shape) => {
-                text_shape.pos += delta;
+                text_shape.pos = transform * text_shape.pos;
+                let mut galley = (*text_shape.galley).clone();
+                let scale_transform = TSTransform::from_scaling(transform.scaling);
+                for row in &mut galley.rows {
+                    row.visuals.mesh.transform(scale_transform);
```
emilk, line 389: This can be optimized slightly

```suggestion
                for row in &mut galley.rows {
                    for v in &mut row.visuals.mesh.vertices {
                        v.pos = Pos2::new(transform.scaling * v.pos.x, transform.scaling * v.pos.y);
                    }
```

crates/epaint/src/shape.rs
```diff
@@ -356,48 +356,56 @@ impl Shape {
...
                 }
             }
             Self::Rect(rect_shape) => {
-                rect_shape.rect = rect_shape.rect.translate(delta);
+                rect_shape.rect = transform * rect_shape.rect;
             }
             Self::Text(text_shape) => {
-                text_shape.pos += delta;
+                text_shape.pos = transform * text_shape.pos;
+                let mut galley = (*text_shape.galley).clone();
```
emilk, line 386:
```suggestion
                text_shape.pos = transform * text_shape.pos;
                
                // Scale text:
                let mut galley = (*text_shape.galley).clone();
```

crates/emath/src/ts_transform.rs
```diff
@@ -0,0 +1,91 @@
...
+    fn mul(self, pos: Pos2) -> Pos2 {
+        self.mul_pos(pos)
+    }
+}
+
+/// Transforms the position.
+impl std::ops::Mul<Rect> for TSTransform {
+    type Output = Rect;
+
+    fn mul(self, rect: Rect) -> Rect {
```
emilk, line 88: All functions in this file should be marked `#[inline]`

crates/emath/src/ts_transform.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub fn invert_pos(&self, pos: Pos2) -> Pos2 {
+        // First, reverse scale around 0, 0, then reverse transform.
+        let pos = pos / self.scaling;
+        pos - self.translation
+    }
+
+    /// Reverses the transformation from screen space to layer space.
+    pub fn invert_rect(&self, rect: Rect) -> Rect {
+        Rect::from_min_max(self.invert_pos(rect.min), self.invert_pos(rect.max))
+    }
```
emilk, line 58: I think a better API is to have an `.inverse()` function, so that `t.invert_pos(x) == t.inverse().mul_pos(x)`

crates/emath/src/ts_transform.rs
```diff
@@ -0,0 +1,91 @@
...
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+#[cfg_attr(feature = "bytemuck", derive(bytemuck::Pod, bytemuck::Zeroable))]
+pub struct TSTransform {
+    /// Translation amount.
+    pub translation: Vec2,
+
+    /// Scaling amount after translation, scaled around (0, 0).
+    pub scaling: f32,
+}
```
emilk, line 17: We should switch this. The common convention is `T * S * v`; see for instance https://docs.rs/glam/latest/glam/f32/struct.Affine2.html#method.from_scale_angle_translation

```suggestion
pub struct TSTransform {
    /// Scaling applied first, scaled around (0, 0).
    pub scaling: f32,
    
    /// Translation amount, applied after translation.
    pub translation: Vec2,
}
```

crates/emath/src/ts_transform.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    /// Reverses the transformation from screen space to layer space.
+    pub fn invert_rect(&self, rect: Rect) -> Rect {
+        Rect::from_min_max(self.invert_pos(rect.min), self.invert_pos(rect.max))
+    }
+
+    /// Transforms the given coordinate by translation then scaling.
+    pub fn mul_pos(&self, pos: Pos2) -> Pos2 {
+        let pos = pos + self.translation;
+        pos * self.scaling
```
emilk, line 63:
```suggestion
        self.translation + self.scaling * pos
```

crates/emath/src/ts_transform.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/// Transforms the position.
+impl std::ops::Mul<Rect> for TSTransform {
+    type Output = Rect;
+
+    fn mul(self, rect: Rect) -> Rect {
+        self.mul_rect(rect)
+    }
+}
```
emilk, line 91: This file could use a few tests!

crates/emath/src/ts_transform.rs
Tweoss: I think this would require having the translation applied in the inverse order. If we generalize the transform internally to be a 3x3 matrix (to account for the affine transformation), then it would be possible to invert.

Tweoss:
I added a few doctests, switched the transform order as you suggested, and completed the small refactoring tasks. 

I think all the tests should be passing as well. Let me know if there's any other changes you'd like me to make. (I'm not sure if you meant for me to add doctests or a test module, I went with doctests for now)

Tweoss:
@emilk Could you take a look at the updates when you have a chance? I'd love to get this merged (and avoid merge conflicts 😛 )

crates/egui_demo_lib/src/demo/pan_zoom.rs
```diff
@@ -0,0 +1,102 @@
...
+
+                // Keep mouse centered.
+                let pan_delta = pointer - new_pos;
+
+                // Handle scrolling.
+                let pan_delta =
+                    pan_delta + ui.ctx().input(|i| i.smooth_scroll_delta) / original_zoom;
+
+                self.transform.scaling = new_zoom;
+                self.transform.translation += pan_delta;
```
emilk, line 57: A simpler way of expressing this would be something like:

```
let pointer_in_layer = self.transform.inverse() * pointer;
let zoom_delta = ui.ctx().input(|i| i.zoom_delta());
let pan_delta = ui.ctx().input(|i| i.smooth_scroll_delta);

// Zoom in on pointer:
self.transform =
    self.transform *
    TSTransform::translate(pointer_in_layer) *
    TSTransform::scale(zoom_delta) *
    TSTransform::translate(-pointer_in_layer);

// Pan:
self.transform = TSTransform::translate(pan_delta) * self.transform;
```

There might be a negation sign wrong there, but the gist of it should be correct.

crates/emath/src/ts_transform.rs
```diff
@@ -0,0 +1,121 @@
...
+/// at `0, 0` (the top left corner)
+#[repr(C)]
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+#[cfg_attr(feature = "bytemuck", derive(bytemuck::Pod, bytemuck::Zeroable))]
+pub struct TSTransform {
+    /// Scaling applied first, scaled around (0, 0).
+    pub scaling: f32,
+
+    /// Translation amount, applied after translation.
```
emilk, line 15: The docstrings are contradictory.

crates/emath/src/ts_transform.rs
```diff
@@ -0,0 +1,121 @@
...
+    pub fn from_translation(translation: Vec2) -> Self {
+        Self::new(translation, 1.0)
+    }
+
+    #[inline]
+    pub fn from_scaling(scaling: f32) -> Self {
+        Self::new(Vec2::ZERO, scaling)
+    }
+
+    /// Reverses the transformation from screen space to layer space.
```
emilk, line 53:
```suggestion
    /// Inverts the transform.
```

crates/emath/src/ts_transform.rs
```diff
@@ -0,0 +1,121 @@
...
+    ///
+    /// ```
+    /// # use emath::{pos2, vec2, TSTransform};
+    /// let p1 = pos2(2.0, 3.0);
+    /// let p2 = pos2(12.0, 5.0);
+    /// let ts = TSTransform::new(vec2(2.0, 3.0), 2.0);
+    /// let inv = ts.inverse();
+    /// assert_eq!(inv.mul_pos(p1), pos2(0.0, 0.0));
+    /// assert_eq!(inv.mul_pos(p2), pos2(5.0, 1.0));
```
emilk, line 62: A good test is to see that the inverse of the inverse is the orignal.

`assert_eq!(ts.inverse().inverse(), ts);`.

Care must be taken to pick translation/scale values that don't produce rounding error for floats though, but it looks like you may already have done so 👍

crates/emath/src/ts_transform.rs
```diff
@@ -0,0 +1,121 @@
...
+    /// ```
+    #[inline]
+    pub fn inverse(&self) -> Self {
+        Self::new(-self.translation / self.scaling, 1.0 / self.scaling)
+    }
+
+    /// Transforms the given coordinate by translation then scaling.
```
emilk, line 69: We don't need to repeat the order in all docstrings. The point is more "transforms a position"

crates/emath/src/ts_transform.rs
```diff
@@ -0,0 +1,121 @@
...
+
+impl TSTransform {
+    pub const IDENTITY: Self = Self {
+        translation: Vec2::ZERO,
+        scaling: 1.0,
+    };
+
+    /// The translation is applied first, then scaling around 0, 0.
+    #[inline]
+    pub fn new(translation: Vec2, scaling: f32) -> Self {
```
emilk, line 36: Wrong docstring again.

The point of applying scale first, is that the transform can also be seen as a pose.

So a scale of 3 means "make it 3 times as big" and a translation of [2,3] mean "put the center of the object at position [2, 3] in the other space".

crates/epaint/src/shape.rs
```diff
@@ -356,48 +356,59 @@ impl Shape {
...
+
+                // Scale text:
+                let mut galley = (*text_shape.galley).clone();
+                for row in &mut galley.rows {
+                    for v in &mut row.visuals.mesh.vertices {
+                        v.pos = Pos2::new(transform.scaling * v.pos.x, transform.scaling * v.pos.y);
+                    }
+                }
+
+                text_shape.galley = Arc::new(galley);
```
emilk, line 395: Better perf:

```suggestion
                let mut galley = Arc::make_mut(&mut text_shape.galley);
                for row in &mut galley.rows {
                    for v in &mut row.visuals.mesh.vertices {
                        v.pos = Pos2::new(transform.scaling * v.pos.x, transform.scaling * v.pos.y);
                    }
                }
```

emilk (approved):
Just some docstrings to fix! Good job :)

crates/epaint/src/shape.rs
```diff
@@ -356,48 +356,59 @@ impl Shape {
     }
 
     /// Move the shape by this many points, in-place.
-    pub fn translate(&mut self, delta: Vec2) {
+    pub fn transform(&mut self, transform: TSTransform) {
```
emilk, line 359: This does not change the stroke-width of any of the shapes.

Was this intentional?

In any case: please document this.

crates/egui_demo_lib/src/demo/pan_zoom.rs
Tweoss: Ahh, that's very clever and much more readable!

crates/epaint/src/shape.rs
Tweoss: Lovely! I didn't think of this optimization

crates/epaint/src/shape.rs
Tweoss: This was not intentional, it was more so I hadn't noticed in the demo anything strange. Upon closer inspection, there are some weird visuals when the stroke width is not changed. Here's a before and after. I'll push the commit adding stroke width scaling, but feel free to revert it! 


https://github.com/emilk/egui/assets/70821802/68f94bcb-b6dd-48d7-b4f0-3b09535e2110


https://github.com/emilk/egui/assets/70821802/a7c46e9b-0519-4061-baa5-e05d239de0ef

Tweoss:
@emilk I fixed up those doc comments on transform order (got myself confused last time, oops!), and added the Arc galley optimization and the simplification for pan and zooms (really neat, thank you!). I also added the inverse doc test like you suggested, and scaled the thickness of strokes. 

There're two videos above showing the effects of scaling the thickness, and I noted in a doc comment that `PaintCallback`s just have their boxes scaled.

MeGaGiGaGon:
I've started messing around with this in the node editor I'm making, and first I'd like to say thank you so much @Tweoss, this makes my life so much easier. In my messing around, I have noticed a few things I'd like to comment on.

---

In the new pan_zoom demo, there is this section:

```rs
        // Uncomment to allow dragging the background as well.
        // self.transform.translation += response.drag_delta() / self.transform.scaling;
```

That is how I want my app to behave, however the `self.transform.scaling` desyncs the mouse movement from the world movement. I think it should just be `self.transform.translation += response.drag_delta()`, since that keeps the movement of the `PanZoom` contents moving 1:1 with the dragging of the mouse.

---

At very close zoom levels, the button text will disappear early when going off the top/left sides of the container. On further testing, the more zoomed in you are, the more down left the disappearance point moves. However, at no level of zoom could I get the text to disappear early off the bottom/left sides of the screen. This can also be seen at 6 seconds into the first recording posted.

https://github.com/emilk/egui/assets/107241144/a0fb68f2-1848-4be1-ab07-eb7ca610acfc

---

When zoomed in/out, the movement rate of dragging the buttons desyncs from the movement rate of the mouse. This is probably because the windows that display the buttons use their own drag-handling code, so they don't get their drag deltas transformed. This is probably the most annoying issue here, since I don't know if it is solvable without changing how `Window` or `InputState` work, or using a custom `Window`-like implementation.

---

Thanks again for the amazing work.

Tweoss:
@MeGaGiGaGon Thanks for the feedback! You are correct, the translation should be without the scaling. 

---

I took a look at the disappearing text (and other shapes) and have reproduced it like you say. I can't figure out exactly what clipping / other mechanism is causing this.

--- 

I'm not clear on what behaviour you're describing. I whipped up an example below, and corrected for the pointer position in `crates/egui/src/response.rs` to get the bottom video. Is this what you're looking for?

https://github.com/emilk/egui/assets/70821802/5a279859-3bd6-43d5-a90a-9205b9315827

https://github.com/emilk/egui/assets/70821802/d591652a-b70b-4ccf-af75-9bd37b2a689f

MeGaGiGaGon:
@Tweoss Thanks for the fix, I hadn't noticed that either. What I was referring to is how the `Area` dragging isn't synced, which sadly wasn't fixed, since it looks to access the `InputState` directly instead of through a `Response`. 

```rs
if movable && move_response.dragged() {
    state.pivot_pos += ctx.input(|i| i.pointer.delta());
}
```

https://github.com/emilk/egui/assets/107241144/d564abf7-44db-4256-a0e5-dbdc6681fe97

Hopefully it is fixable inside of `InputState`/`PointerState`, since if not any custom widget that accesses the pointer position through `ui`/`ctx` `.input` would also be broken.

crates/egui/src/response.rs
```diff
@@ -395,7 +402,14 @@ impl Response {
...
         if self.hovered() {
-            self.ctx.input(|i| i.pointer.hover_pos())
+            let mut pos = self.ctx.input(|i| i.pointer.hover_pos());
+            if let Some(transform) = self
+                .ctx
+                .memory(|m| m.layer_transforms.get(&self.layer_id).cloned())
+            {
+                pos = pos.map(|p| transform * p);
+            }
+            pos
```
emilk, line 412:
```suggestion
            let mut pos = self.ctx.input(|i| i.pointer.hover_pos())?;
            if let Some(transform) = self
                .ctx
                .memory(|m| m.layer_transforms.get(&self.layer_id).cloned())
            {
                pos = transform * pos;
            }
            Some(pos)
```

emilk (approved):

emilk:
The text clip bug is quite annoying.

It would also be good if the demo/example has a description in the UI (e.g. a label at the top, that isn't zoomed/panned)

MeGaGiGaGon:
The two other biggest annoyances I've seen are that remain are the internal `Area`s not moving with the outer `Window`'s movement, and the layering issues. I tried looking into both, but got stumped on changing an `Area`'s `pivot_pos` since it is in `State` which is private. I also couldn't find any method of getting the ordering of the window containing the ui, though since `Areas` stores that in `order: Vec<LayerId>` then you'd just have to ensure the windows in the `PanZoom` are directly after the location of the containing window. This seems to more be a symptom of there not being a good interface for setting the ordering of areas/paintings, but that's probably something for a separate issue/pull.

https://github.com/emilk/egui/assets/107241144/21401034-7811-4cca-8381-20de29100293


It would also be nice if there was an example of painting a shape onto the transforming layer, since painting onto the top level `ui` just makes it not zoom/pan, and having to make a static area to paint on/for every paint seems clunky.

Tweoss:
@MeGaGiGaGon Is this the behaviour you're referring to where the inner windows should have their transformation match the outer demo window?

https://github.com/emilk/egui/assets/70821802/06d1d151-36cc-4ee6-a5ae-ee08af8f7cec

---

Unfortunately, I'm also not sure of a good solution to ensuring that the inner windows stay atop the demo window but not other demo windows.

MeGaGiGaGon:
@Tweoss Yep, that's it :)
I think it would be possible, if not for two things:
`Areas.order` is private, otherwise you could edit it to make the order correct using `ui.ctx().memory_mut`
I have no clue how to get the `LayerID` of the window containing the `PanZoom`, especially with just access to the `ui`

Tweoss:
Regarding the text disappearing, I've finally discovered that it is the initial text boxes' bounding box that goes beyond the clip rectangle, causing the whole text to disappear. I'll try to transform that bounding box as well.

MeGaGiGaGon:
Yay!

---

It's probably better to have the smiley be on it's own un-interactable area, since otherwise odd things happen with the widget it's painted on

image
 
It would also be nice to have an example painting that also moves with a draggable area

---

I did even more thinking, and sadly my idea wouldn't work. Getting the `LayerId` of the window isn't too bad if some spaghetti is allowed, here's my flavor: 

```rs
pub struct PanZoom {
    transform: TSTransform,
    drag_value: f32,
    window_layer_id: Option<egui::LayerId>,
}
// ...
// inside impl super::Demo for PanZoom inside fn show
self.window_layer_id = window.show(ctx, |ui| self.ui(ui)).and_then(|x| Some(x.response.layer_id));
// ...
// inside fn ui
if self.window_layer_id.is_none() {
    return;
}
```
However the issue is still the `Area` ordering. As far as I can tell, the only tool that can inspect the `Area.order` is 
```rs
ui.ctx().memory(|mem| mem.areas().top_layer_id(egui::Order::Middle))
````
This is sadly not powerful enough, since the only tool for manipulating the `order` is 
```rs
 ui.ctx().move_to_top(layer_id)
```
Which also affects the top most `Area`. The only two things I can think of that would make it possible are having `Areas.order` be public, or having a `Context::move_to_bottom` method, which would still cause a lot of spaghetti in this code, but wouldn't expose `Areas`'s internals.

---

Something else I just noticed is that things start glitching out if you zoom in far enough. I'd suspect this is due to floating point imprecision, so meh.

https://github.com/emilk/egui/assets/107241144/f29607c0-eb13-4fdc-aa83-f9f911cd5584

---

I also also just noticed that while the translation is saved on close and open, the scale isn't

Tweoss:
Some widgets may need their mouse input coordinates to be reverse transformed, like in the panels.

MeGaGiGaGon:
What I'm mostly thinking about is if this were to be separated into it's own container/widget struct. I imagine most people making a `PanZoom` would just copy all the complex internal logic into their own struct (like I am) and have a few fns for their own specific use case. Granted, I have absolutely no clue how to make a good interface, and it looks like a lot of works that might be better suited for a later PR so that the actual `TSTransform` can make it into egui sooner rather than later.

Also on more thinking, one of these days there should also be a PR for making `Areas.order` more accessible, since that's the main issue that's being run into, and is probably outside the scope of this one. With what I've seen on how `Memory` added `move_to_top`, I might make a stab at it.

emilk:
This seems to be working quite well now.

The layering issue is difficult, and I think should be fixed in a separate PR.

Potential solutions include:
* Add a good API for reordering layers, so we can tell the scroll-pan-layer to always be just on top of the demo window.
* Add sub-layers, so each window can have multiple layers that are always connected to each other. This could also be the solution for https://github.com/emilk/egui/issues/1516

emilk:
Oops - unfortunately this broke the drag-and-drop preview code. I'll see if I can fix it.

Tweoss:
Thank you @emilk for taking the time to review and add changes! 

Also, thank you @MeGaGiGaGon for the suggestions and noticing those bugs.

Dampfwalze:
`set_transform_layer` messes up text selection:


https://github.com/emilk/egui/assets/46897578/4cba62df-9f7c-49ab-99d4-2338118ebef7

Issue #3492, opened 2023-10-21
egui-wgpu docs failed to build on docs.rs


abey79:
See https://docs.rs/crate/egui-wgpu/0.23.0

abey79:
I believe this is related to https://github.com/emilk/egui/pull/3228. I got that error on my CI when updating to egui 0.23 (https://github.com/abey79/vsvg/pull/54). Without the x11 or wayland feature, winit isn't able to detect the platform on linux (https://github.com/rust-windowing/winit/blob/772b21ce095b655f0fe50b89253910cd2ff9fa78/build.rs#L13). My workaround was to run that part on a mac runner, which is probably not an option for docs.rs.

abey79:
winit issue: https://github.com/rust-windowing/winit/issues/3173

tgross35:
This is still an issue at 0.24.0 https://docs.rs/crate/egui-wgpu/0.24.0

tgross35:
But they seem to build fine on my local, weird...

PR #3907, opened 2024-01-28
Enable winit backends when building egui-wgpu on docs.rs to fix failures

Imberflur:
I tested that this command fails:
```
cargo doc -p egui-wgpu --all-features --no-deps 
```
And that this one, which I think will correspond to these changes, succeeds:
```
cargo doc -p egui-wgpu --all-features --features "winit/x11 winit/wayland" --no-deps 
```

https://github.com/emilk/egui/issues/3492 should be fixed by this

Also, I noticed that the PR template and `CONTRIBUTING.md` mention running `./scripts/check.sh` which in a few cases has the same issue of winit failing to build with no backends enabled. I guess this script hasn't been in use?

### DIFF
--- a/crates/egui-wgpu/Cargo.toml
+++ b/crates/egui-wgpu/Cargo.toml
@@ -25,6 +25,7 @@ include = [
 
 [package.metadata.docs.rs]
 all-features = true
+features = ["winit/x11", "winit/wayland"]
 
 
 [features]


YgorSouza:
Maybe egui-wgpu should have the "x11" and "wayland" features like egui-winit and eframe do? Would that make sense?

YgorSouza:
[egui_glow](https://docs.rs/crate/egui_glow/0.25.0) has the same problem, by the way.

YgorSouza:
I opened #3909 to try and solve the issue more broadly, if anyone wants to take a look.

Imberflur:
> Maybe egui-wgpu should have the "x11" and "wayland" features like egui-winit and eframe do? Would that make sense?

I not sure it makes sense if the user will need to separately depend on winit. I assume that this is the case here but I haven't looked very deeply into it.

emilk (approved):

Imberflur:
I believe this is no longer needed since #3909 merged

PR #3909, opened 2024-01-28
Add `x11` and `wayland` features to egui-wgpu and egui_glow

### LINKED ISSUE #2286: egui-glow fails cargo check --all-features

### LINKED ISSUE #3492: egui-wgpu docs failed to build on docs.rs

YgorSouza:
This allows them to build correctly on Linux by passing one or both of the features alongside `winit`.

Closes #3492
Closes #2286

### DIFF
--- a/crates/egui-wgpu/Cargo.toml
+++ b/crates/egui-wgpu/Cargo.toml
@@ -31,8 +31,14 @@ all-features = true
 ## Enable profiling with the [`puffin`](https://docs.rs/puffin) crate.
 puffin = ["dep:puffin"]
 
-## Enable [`winit`](https://docs.rs/winit) integration.
+## Enable [`winit`](https://docs.rs/winit) integration. On Linux, requires either `wayland` or `x11`
 winit = ["dep:winit"]
+
+## Enables Wayland support for winit.
+wayland = ["winit?/wayland"]
+
+## Enables x11 support for winit.
+x11 = ["winit?/x11"]
 
 
 [dependencies]
--- a/crates/eframe/Cargo.toml
+++ b/crates/eframe/Cargo.toml
@@ -88,7 +88,11 @@ puffin = [
 ]
 
 ## Enables wayland support and fixes clipboard issue.
-wayland = ["egui-winit/wayland", "egui-wgpu?/wayland", "egui_glow?/wayland"]
+wayland = [
+  "egui-winit/wayland",
+  "egui-wgpu?/wayland",
+  "egui_glow?/wayland",
+]
 
 ## Enable screen reader support (requires `ctx.options_mut(|o| o.screen_reader = true);`) on web.
 ##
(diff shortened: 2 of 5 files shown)


crates/egui-wgpu/Cargo.toml
```diff
@@ -31,9 +31,15 @@ all-features = true
...
-## Enable [`winit`](https://docs.rs/winit) integration.
+## Enable [`winit`](https://docs.rs/winit) integration. On Linux, requires either `wayland` or `x11`
 winit = ["dep:winit"]
 
+## Enables Wayland support for winit.
+wayland = ["winit?/wayland"]
+
+## Enables x11 support for winit.
+x11 = ["winit?/x11"]
+
```
emilk, line 42: If a user is using `egui-wgpu` with the `winit` feature, they are also depending directly on `winit`, and can set the `x11/wayland` features themselves, no?

YgorSouza: Yes, but then you have to think about that when running `cargo check`, `cargo doc` etc, which is the problem noted in the issues. Technically you don't have to enable these features in a real project, they just help make the repo more ergonomic to work with. I don't know if there is a better solution.

emilk: ah I see

emilk: The problem would still happen with `cargo check` though, just not with `cargo check --all-features` ?

emilk (approved):

PR #3914, opened 2024-01-29
Postpone call to get_current_texture

emilk:
This should help slightly with CPU/GPU parallelism when vsync is on.

I also return the time spent on vsync, which can help users figure out how much CPU wall-time was used on non-vsync stuff

### DIFF
--- a/crates/egui-wgpu/src/winit.rs
+++ b/crates/egui-wgpu/src/winit.rs
@@ -572,7 +572,7 @@ impl Painter {
 
         let output_frame = {
             crate::profile_scope!("get_current_texture");
-            // This is what vsync-waiting happens everywhere except DX12
+            // This is where vsnc happens:
             let start = web_time::Instant::now();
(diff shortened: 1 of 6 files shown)


crates/egui-wgpu/src/winit.rs
```diff
@@ -668,6 +668,7 @@ impl Painter {
 
         {
             crate::profile_scope!("present");
+            // This is where vsync happens in DX12
```
Wumpf, line 671: apologies I got that wrong at first, plz remove this line again :). The wait is always happening `get_current_texture`, but internally (!) what is waited on can be subtly different

See 
* https://github.com/gfx-rs/wgpu/issues/3283
* https://github.com/gfx-rs/wgpu/issues/2869

Wumpf (approved):
nice! I believe this should make the "low latency" rendering mode a lot more viable

PR #3916, opened 2024-01-29
Add opacity factor to `TextShape`

StratusFearMe21:
This PR simply allows you to override the opacity of a Galley when you draw it on screen. Last year I opened #3548 and some changes were requested to the PR, but unfortunately school got really busy and I wasn't able to apply them. This PR supersedes #3548 and applys the changes requested in that PR

Closes #3548

### DIFF
--- a/crates/egui/src/painter.rs
+++ b/crates/egui/src/painter.rs
@@ -469,29 +469,6 @@ impl Painter {
         }
     }
 
-    /// Paint text that has already been laid out in a [`Galley`].
-    ///
-    /// You can create the [`Galley`] with [`Self::layout`].
-    ///
-    /// The entire [`Galley`] will be rendered with the given opacity in gamma space.
-    #[inline]
-    pub fn galley_with_opacity_factor(
-        &self,
-        pos: Pos2,
-        galley: Arc<Galley>,
-        opacity_factor: f32,
-        fallback_color: Color32,
-    ) {
-        if !galley.is_empty() && opacity_factor > 0.0 {
-            self.add(Shape::galley_with_opacity_factor(
-                pos,
-                galley,
-                opacity_factor,
-                fallback_color,
-            ));
-        }
-    }
-
     #[deprecated = "Use `Painter::galley` or `Painter::galley_with_override_text_color` instead"]
     #[inline]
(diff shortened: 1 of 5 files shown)


crates/epaint/src/tessellator.rs
```diff
@@ -1548,6 +1553,10 @@ impl Tessellator {
                             color = *fallback_color;
                         }
 
+                        if *opacity_factor <= 1.0 {
```
emilk, line 1556:
```suggestion
                        if *opacity_factor < 1.0 {
```

StratusFearMe21: Oops, thanks

crates/egui/src/painter.rs
```diff
@@ -469,6 +469,29 @@ impl Painter {
         }
     }
 
+    /// Paint text that has already been laid out in a [`Galley`].
+    ///
+    /// You can create the [`Galley`] with [`Self::layout`].
+    ///
+    /// The entire [`Galley`] will be rendered with the given opacity in gamma space.
+    #[inline]
+    pub fn galley_with_opacity_factor(
```
emilk, line 478: I think these helpers are too much. `painter.add(TextShape::new(pos, galley, color).with_opacity_factor(opacity))` is ergonomic enough for this rare use case imho

crates/epaint/src/shape.rs
```diff
@@ -288,6 +288,19 @@ impl Shape {
             .into()
     }
 
+    /// The entire [`Galley`] will be rendered with the given opacity in gamma space.
+    #[inline]
+    pub fn galley_with_opacity_factor(
```
emilk, line 293: Same here - no need for this Imho. Users can just use the `TextShape::new(…).with_opacity_factor(…)`

StratusFearMe21: Applied in latest commit

crates/egui/src/painter.rs
StratusFearMe21: Applied in latest commit

crates/epaint/src/tessellator.rs
StratusFearMe21: Applied in latest commit

emilk (approved):

PR #3920, opened 2024-01-30
Plot items now have optional id which is returned in the plot's response when hovered

Wumpf:
This allows users to check which item the user interacts with in the plot.

https://github.com/emilk/egui/assets/1220815/1a174b38-8414-49be-a802-d187cd93d154

### DIFF
--- a/crates/egui_demo_lib/src/demo/plot_demo.rs
+++ b/crates/egui_demo_lib/src/demo/plot_demo.rs
@@ -791,8 +791,26 @@ impl InteractionDemo {
         let PlotResponse {
             response,
             inner: (screen_pos, pointer_coordinate, pointer_coordinate_drag_delta, bounds, hovered),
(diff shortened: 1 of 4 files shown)


crates/egui_demo_lib/src/demo/plot_demo.rs
```diff
@@ -823,6 +841,7 @@ impl InteractionDemo {
             pointer_coordinate_drag_delta.x, pointer_coordinate_drag_delta.y
         );
         ui.label(format!("pointer coordinate drag delta: {coordinate_text}"));
+        ui.label(format!("hovered plot item: {hovered_plot_item:?}"));
```
emilk, line 844: Let's go all the way and print if it was "sin" or "cos" being hovered (or neither)

emilk (approved):

emilk (approved):

Issue #3927, opened 2024-01-31
Customize eframe persistence file path


C0D3-M4513R:
**Is your feature request related to a problem? Please describe.**

When making an app, where you want to allow saving to a custom location (e.g. because you want your users to be able to work on multiple projects at once), you basically have to create a custom `eframe::Storage` implementation, which for me boils down to just copying the file that `eframe::native::FileStorage` resides in, removing the profiler macros and making the `eframe::native::FileStorage::from_ron_filepath` public.

 
**Describe the solution you'd like**

just make `eframe::native::FileStorage::from_ron_filepath` public

**Describe alternatives you've considered**

just copy the existing implementation

**Additional context**

emilk:
Sure - you wanna make a PR? 🙏

YgorSouza:
Did #4423 resolve this?

emilk:
I'll mark this as closed by https://github.com/emilk/egui/pull/4423 
If you have further requests, please open a new issue

PR #3939, opened 2024-02-01
feat: make `eframe::native::FileStorage` creatable from other crates

C0D3-M4513R:
This PR makes `eframe::native::FileStorage` creatable from other crates.
To make `eframe::native::FileStorage` nice to use, some api's have been added.
Closes: #3927

### DIFF
--- a/crates/eframe/src/native/file_storage.rs
+++ b/crates/eframe/src/native/file_storage.rs
@@ -25,6 +25,8 @@ pub fn storage_dir(app_id: &str) -> Option<PathBuf> {
 /// Used to restore egui state, glium window position/size and app state.
 pub struct FileStorage {
     ron_filepath: PathBuf,
+    properly_read: bool,
+    last_save_success: Option<bool>,
     kv: HashMap<String, String>,
     dirty: bool,
     last_save_join_handle: Option<std::thread::JoinHandle<()>>,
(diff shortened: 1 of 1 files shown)


C0D3-M4513R:
I'm sorry, but I am unable to test this pr, as:
- `error: could not find system library 'glib-2.0' required by the 'glib-sys' crate`
- `error: could not find system library 'gobject-2.0' required by the 'gobject-sys' crate`
and idk how to resolve this on nixos. I already tried bringing those modules into my current environment.

crates/eframe/src/native/file_storage.rs
```diff
@@ -41,16 +41,16 @@ impl Drop for FileStorage {
 
 impl FileStorage {
     /// Store the state in this .ron file.
-    fn from_ron_filepath(ron_filepath: impl Into<PathBuf>) -> Self {
+    pub fn from_ron_filepath(ron_filepath: impl Into<PathBuf>) -> Option<Self> {
```
emilk, line 44: why return an Option for a function that never returns `None`?

```suggestion
    pub fn from_ron_filepath(ron_filepath: impl Into<PathBuf>) ->Self {
```

emilk (changes requested):

crates/eframe/src/native/file_storage.rs
C0D3-M4513R: Please see https://github.com/emilk/egui/pull/3939/commits/cdcfe7a263473eafabcb130b847bdcb281eef073#diff-99baf00dab62955842a4603f181399c5733a58a9a8201d8c7a09279b7870f0fdR49

It does return None here, through `?`.

C0D3-M4513R: The motivation behind this change is, that I don't want people thinking that this will always return a valid `FileStorage`. (e.g. `read_ron` fails if it cannot access the file or cannot de-serialize the contents properly)

emilk:
Please give this PR a proper title and description

PR #3945, opened 2024-02-02
`eframe`: add `links` feature to opt-out of some deps

jrudolph:
Seems useful to avoid pulling in some dependencies in an eframe app if you don't need links.

### DIFF
--- a/crates/eframe/Cargo.toml
+++ b/crates/eframe/Cargo.toml
@@ -32,6 +32,7 @@ default = [
   "accesskit",
   "default_fonts",
   "glow",
+  "links",
   "wayland",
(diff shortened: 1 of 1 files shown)


crates/eframe/Cargo.toml
```diff
@@ -65,6 +66,8 @@ glow = [
   "winit/rwh_05",
 ]
 
+links = ["egui-winit/links"]
```
emilk, line 69: Needs documentation

emilk:
Closing: eframe got an opt-out `links` feature in #8372. Thanks for the PR!

PR #3946, opened 2024-02-02
more selectively select x11 or wayland from glutin

jrudolph:
In an attempt to strip down features and dependencies needed for development, I tried to only select x11 in eframe but still got a bunch of wayland dependencies.

Therefore, I tried to more selectively choose features from glutin. (I only selected the glx API, not sure what the consequence is of that, the default is ` ["egl", "glx", "x11", "wayland", "wgl"]`. Maybe `egl` and `wgl` should stay as well?

### DIFF
--- a/crates/eframe/Cargo.toml
+++ b/crates/eframe/Cargo.toml
@@ -92,6 +92,8 @@ wayland = [
   "egui-winit/wayland",
   "egui-wgpu?/wayland",
   "egui_glow?/wayland",
+  "glutin-winit?/wayland",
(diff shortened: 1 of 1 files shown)


emilk:
> Maybe `egl` and `wgl` should stay as well?

🤷  it's hard to know when glutin doesn't document the features 🤦

emilk:
Closing: this landed in #5391 / #5490, which made eframe forward its `x11` and `wayland` features to glutin. Thanks for the PR!

PR #3951, opened 2024-02-02
adds a image loader for DynamicImage

JustFrederik:
The macro converts the dynamic_image to rgba8 bytes and the rest is handled by the new image loader(only processes Uris that start with "rgba8://"

```
let img = image::load_from_memory(include_bytes!("../../testimg.png").as_slice()).unwrap();
let v = include_dynamic_image!("testimg", img);
let img = Image::new(v);
```

### DIFF
--- a/crates/egui_extras/src/image.rs
+++ b/crates/egui_extras/src/image.rs
@@ -190,6 +190,8 @@ impl RetainedImage {
 
 // ----------------------------------------------------------------------------
 
+#[cfg(feature = "image")]
+use egui::load::Bytes;
 use egui::ColorImage;
 
 /// Load a (non-svg) image.
@@ -209,6 +211,26 @@ pub fn load_image_bytes(image_bytes: &[u8]) -> Result<egui::ColorImage, String> 
     Ok(egui::ColorImage::from_rgba_unmultiplied(
         size,
         pixels.as_slice(),
+    ))
+}
+#[cfg(feature = "image")]
+pub fn load_rgba(image_bytes: &[u8]) -> Result<egui::ColorImage, String> {
+    crate::profile_function!();
+    let size_of_usize = std::mem::size_of::<usize>();
+    if image_bytes.len() < std::mem::size_of::<usize>() * 2 {
+        return Err("Not enough bytes for image".to_string());
+    }
+
+    let width = usize::from_le_bytes(image_bytes[0..size_of_usize].try_into().unwrap());
+    let height = usize::from_le_bytes(
+        image_bytes[size_of_usize..size_of_usize * 2]
+            .try_into()
+            .unwrap(),
+    );
+    let size = [width, height];
+    Ok(egui::ColorImage::from_rgba_unmultiplied(
+        size,
+        &image_bytes[size_of_usize * 2..],
     ))
 }
 
@@ -278,4 +300,26 @@ pub fn load_svg_bytes_with_size(
     let image = egui::ColorImage::from_rgba_unmultiplied([w as _, h as _], pixmap.data());
 
     Ok(image)
+}
+
+#[cfg(feature = "image")]
+pub fn convert_image_to_bytes(image: image::DynamicImage) -> Bytes {
+    let image_buffer = image.to_rgba8();
+    let pixels = image_buffer.as_flat_samples();
+    let mut result: Vec<u8> = Vec::new();
+    result.extend_from_slice(&(image.width() as usize).to_le_bytes());
+    result.extend_from_slice(&(image.height() as usize).to_le_bytes());
+    result.extend_from_slice(pixels.as_slice());
+    result.into()
+}
+
+#[cfg(feature = "image")]
+#[macro_export]
+macro_rules! include_dynamic_image {
+    ($uri: literal, $image: expr) => {
+        egui::ImageSource::Bytes {
+            uri: ::std::borrow::Cow::Borrowed(concat!("rgba8://", $uri)),
+            bytes: $crate::image::convert_image_to_bytes($image),
+        }
+    };
 }
--- a/crates/egui_extras/src/loaders.rs
+++ b/crates/egui_extras/src/loaders.rs
@@ -78,6 +78,14 @@ pub fn install_image_loaders(ctx: &egui::Context) {
         log::trace!("installed ImageCrateLoader");
     }
 
+    #[cfg(feature = "image")]
+    if !ctx.is_loader_installed(self::raw_image_loader::Rgba8ImageLoader::ID) {
+        ctx.add_image_loader(std::sync::Arc::new(
+            self::raw_image_loader::Rgba8ImageLoader::default(),
+        ));
+        log::trace!("installed ImageCrateLoader");
+    }
+
     #[cfg(feature = "svg")]
     if !ctx.is_loader_installed(self::svg_loader::SvgLoader::ID) {
         ctx.add_image_loader(std::sync::Arc::new(self::svg_loader::SvgLoader::default()));
@@ -103,6 +111,9 @@ mod ehttp_loader;
 
 #[cfg(feature = "image")]
 mod image_loader;
+
+#[cfg(feature = "image")]
+mod raw_image_loader;
 
 #[cfg(feature = "svg")]
 mod svg_loader;
--- a/crates/egui_extras/src/loaders/raw_image_loader.rs
+++ b/crates/egui_extras/src/loaders/raw_image_loader.rs
@@ -1,0 +1,84 @@
+use egui::{
+    ahash::HashMap,
+    load::{BytesPoll, ImageLoadResult, ImageLoader, ImagePoll, LoadError, SizeHint},
+    mutex::Mutex,
+    ColorImage,
+};
+use std::{mem::size_of, sync::Arc};
+
+type Entry = Result<Arc<ColorImage>, String>;
+
+#[derive(Default)]
+pub struct Rgba8ImageLoader {
+    cache: Mutex<HashMap<String, Entry>>,
+}
+
+impl Rgba8ImageLoader {
+    pub const ID: &'static str = egui::generate_loader_id!(DynamicImageCrateLoader);
+}
+
+fn is_supported_uri(uri: &str) -> bool {
+    uri.starts_with("rgba8://")
+}
+
+fn is_unsupported_mime(mime: &str) -> bool {
+    mime.contains("svg")
+}
+
+impl ImageLoader for Rgba8ImageLoader {
+    fn id(&self) -> &str {
+        Self::ID
+    }
+
+    fn load(&self, ctx: &egui::Context, uri: &str, _: SizeHint) -> ImageLoadResult {
+        // three stages of guessing if we support loading the image:
+        // 1. URI extension
+
+        // (1)
+        if !is_supported_uri(uri) {
+            return Err(LoadError::NotSupported);
+        }
+
+        let mut cache = self.cache.lock();
+        if let Some(entry) = cache.get(uri).cloned() {
+            match entry {
+                Ok(image) => Ok(ImagePoll::Ready { image }),
+                Err(err) => Err(LoadError::Loading(err)),
+            }
+        } else {
+            match ctx.try_load_bytes(uri) {
+                Ok(BytesPoll::Ready { bytes, mime, .. }) => {
+                    log::trace!("started loading {uri:?}");
+                    let result = crate::image::load_rgba(&bytes).map(Arc::new);
+                    log::trace!("finished loading {uri:?}");
+                    cache.insert(uri.into(), result.clone());
+                    match result {
+                        Ok(image) => Ok(ImagePoll::Ready { image }),
+                        Err(err) => Err(LoadError::Loading(err)),
+                    }
+                }
+                Ok(BytesPoll::Pending { size }) => Ok(ImagePoll::Pending { size }),
+                Err(err) => Err(err),
+            }
+        }
+    }
+
+    fn forget(&self, uri: &str) {
+        let _ = self.cache.lock().remove(uri);
+    }
+
+    fn forget_all(&self) {
+        self.cache.lock().clear();
+    }
+
+    fn byte_size(&self) -> usize {
+        self.cache
+            .lock()
+            .values()
+            .map(|result| match result {
+                Ok(image) => image.pixels.len() * size_of::<egui::Color32>(),
+                Err(err) => err.len(),
+            })
+            .sum()
+    }
+}


JustFrederik:
the use case I intended for it is to display gifs. an example for that would be:
```
use image::codecs::gif::GifDecoder;
use image::{AnimationDecoder, EncodableLayout, ImageDecoder};
use std::fs::File;

pub fn gif() -> Vec<u8> {
    let file_in = File::open("src/assets/loading.gif").unwrap();
    let decoder = GifDecoder::new(file_in).unwrap();
    let (width, height) = decoder.dimensions();
    for frame in decoder.into_frames() {
        let frame = frame.unwrap();
        let img = frame.buffer();
        let mut bytes = vec![];
        bytes.extend_from_slice(&(width as usize).to_le_bytes());
        bytes.extend_from_slice(&(height as usize).to_le_bytes());
        bytes.extend_from_slice(img.as_bytes());
        let delay = frame.delay();
    }
}
```

I plan to implement a gif component. as of right now it is possible to create a image for each frame

crates/egui_extras/src/image.rs
```diff
@@ -279,3 +301,25 @@ pub fn load_svg_bytes_with_size(
...
+#[cfg(feature = "image")]
+pub fn convert_image_to_bytes(image: &image::DynamicImage) -> Bytes {
+    let image_buffer = image.to_rgba8();
+    let pixels = image_buffer.as_flat_samples();
+    let mut result: Vec<u8> = Vec::new();
+    result.extend_from_slice(&(image.width() as usize).to_le_bytes());
+    result.extend_from_slice(&(image.height() as usize).to_le_bytes());
+    result.extend_from_slice(pixels.as_slice());
+    result.into()
+}
```
emilk, line 314: A simpler solution here is to encode it as a lossless, uncompressed `.tga`. Then `egui_extras::ImageLoader` will load it for you (no need to write an extra loader!).

crates/egui_extras/src/image.rs
```diff
@@ -279,3 +301,25 @@ pub fn load_svg_bytes_with_size(
...
+
+#[cfg(feature = "image")]
+#[macro_export]
+macro_rules! include_dynamic_image {
```
emilk, line 318: I don't see why this needs to be a macro?

crates/egui_extras/src/image.rs
JustFrederik: my problem with that is that I encode something to decode it right after

crates/egui_extras/Cargo.toml
```diff
@@ -79,7 +79,7 @@ chrono = { version = "0.4", optional = true, default-features = false, features
 ## Enable this when generating docs.
 document-features = { workspace = true, optional = true }
 
-image = { version = "0.24", optional = true, default-features = false }
+image = { version = "0.24", optional = true, features = ["gif"], default-features = false }
```
emilk, line 82: gif should be opt-in

crates/egui_extras/src/image.rs
emilk: Ok, but then I suggest you add the decoding function here, next to the encoding function, and that you at least add some magic header to it as a sanity check

crates/egui_extras/src/loaders.rs
```diff
@@ -104,5 +112,8 @@ mod ehttp_loader;
 #[cfg(feature = "image")]
 mod image_loader;
 
+#[cfg(feature = "image")]
```
emilk, line 115: This is not really what the `image` feature means: https://docs.rs/egui_extras/latest/egui_extras/

emilk (changes requested):

crates/egui_extras/src/loaders.rs
JustFrederik: The description is: `Add support for loading images with the [image](https://docs.rs/image) crate`.
The image_loader is loading DynamicImages from the [image](https://docs.rs/image) crate. Im not quite sure as what I should label it. I created it for the use with gifs, but it can be used with images that were already preprocessed

crates/egui/src/widgets/image.rs
```diff
@@ -43,15 +44,93 @@ use epaint::{util::FloatOrd, RectShape};
...
     texture_options: TextureOptions,
     image_options: ImageOptions,
     sense: Sense,
     size: ImageSize,
     pub(crate) show_loading_spinner: Option<bool>,
 }
 
+#[derive(Debug, Clone)]
+pub enum ImageSources<'a> {
+    Multi(Vec<(ImageSource<'a>, Duration)>),
```
emilk, line 58: Needs a docstring

crates/egui/src/widgets/image.rs
```diff
@@ -43,15 +44,93 @@ use epaint::{util::FloatOrd, RectShape};
...
+    Single(ImageSource<'a>),
+}
+
+impl<'a> ImageSources<'a> {
+    pub fn get_source(&self, ctx: &Context, id: Id) -> &'a ImageSource<'_> {
```
emilk, line 63: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter
```suggestion
    pub fn source(&self, ctx: &Context, id: Id) -> &'a ImageSource<'_> {
```

crates/egui/src/widgets/image.rs
```diff
@@ -43,15 +44,93 @@ use epaint::{util::FloatOrd, RectShape};
...
+        match self {
+            ImageSources::Multi(imgs) => {
+                let now = Instant::now();
```
emilk, line 66: This won't work in Wasm. Use `ctx.input(|i| i.time)` instead

crates/egui/src/widgets/image.rs
```diff
@@ -43,15 +44,93 @@ use epaint::{util::FloatOrd, RectShape};
...
+                    let data = data.get_temp_mut_or_insert_with(id, || ImageData {
+                        image: 0,
+                        last_refresh: now,
+                    });
+                    let dur = imgs[data.image].1;
+                    if data.last_refresh + dur < now {
+                        data.image = (data.image + 1) % imgs.len();
+                        data.last_refresh = now;
+                        //TODO:
+                        //ctx.request_repaint_after(dur);
```
emilk, line 77: You should request repaint after waiting for how long is left of the frame

crates/egui/src/widgets/image.rs
```diff
@@ -43,15 +44,93 @@ use epaint::{util::FloatOrd, RectShape};
...
+impl<'a> From<Vec<(ImageSource<'a>, Duration)>> for ImageSources<'a> {
+    fn from(value: Vec<(ImageSource<'a>, Duration)>) -> Self {
+        Self::Multi(value)
+    }
+}
+
 impl<'a> Image<'a> {
+    /// Loads gif from id & `[ImageSources]`
+    /// `egui::Image::new_gif(egui_extras::include_gif!("../../assets/ferris.gif"))`
+    pub fn new_gif((id, sources): (impl ToString, impl Into<ImageSources<'a>>)) -> Self {
```
emilk, line 103:
```suggestion
    pub fn new_aimation((id, sources): (impl ToString, impl Into<ImageSources<'a>>)) -> Self {
```

crates/egui/src/widgets/image.rs
```diff
@@ -43,15 +44,93 @@ use epaint::{util::FloatOrd, RectShape};
 #[must_use = "You should put this widget in an ui with `ui.add(widget);`"]
 #[derive(Debug, Clone)]
 pub struct Image<'a> {
-    source: ImageSource<'a>,
+    id: Id,
```
emilk, line 47: What is this used for?

crates/egui/src/widgets/image.rs
```diff
@@ -43,15 +44,93 @@ use epaint::{util::FloatOrd, RectShape};
...
+        match self {
+            ImageSources::Multi(imgs) => {
+                let now = Instant::now();
+                let index = ctx.data_mut(|data| {
+                    let data = data.get_temp_mut_or_insert_with(id, || ImageData {
+                        image: 0,
+                        last_refresh: now,
+                    });
+                    let dur = imgs[data.image].1;
+                    if data.last_refresh + dur < now {
```
emilk, line 73: We shouldn't need to store any state. Just calculate the sum of the frame time, then modulate the current time with that and calculate the frame index and how long until the next frame. No need for `id`

crates/egui/src/widgets/image.rs
```diff
@@ -275,8 +354,8 @@ impl<'a> Image<'a> {
     }
 
     #[inline]
-    pub fn size(&self) -> Option<Vec2> {
-        match &self.source {
+    pub fn size(&self, ctx: &Context) -> Option<Vec2> {
+        match self.source(ctx) {
```
emilk, line 358: Do we really expect the size to change over time? 😬

crates/egui_extras/src/loaders.rs
emilk: It looks to me like nothing in `raw_image_loader` is using the `image` crate, but maybe I'm missing something

emilk (changes requested):

crates/egui/src/widgets/image.rs
```diff
@@ -43,15 +44,93 @@ use epaint::{util::FloatOrd, RectShape};
...
     pub(crate) show_loading_spinner: Option<bool>,
 }
 
+#[derive(Debug, Clone)]
+/// This type tells the [`Ui`] if the image is an animated Image and when to render the next frame.
+///
+/// This is used by [`Image::new_animation`]
+pub enum ImageSources<'a> {
+    Multi(Vec<(ImageSource<'a>, Duration)>),
+    Single(ImageSource<'a>),
```
emilk, line 61: these names might be more descriptive

```suggestion
    /// A static image.
    Static(ImageSource<'a>),
    
    /// An animation of images, with the length of time to show each frame.
    Animated(Vec<(ImageSource<'a>, Duration)>),
```

crates/egui/src/widgets/image.rs
```diff
@@ -43,15 +44,93 @@ use epaint::{util::FloatOrd, RectShape};
...
+        match self {
+            ImageSources::Multi(imgs) => {
+                let now = (ctx.input(|i| i.time) * 1000.) as u128;
+                let frames: u128 = imgs.iter().map(|v| v.1.as_millis()).sum();
+                let pos = now % frames;
+                let mut cumulative_duration = 0;
+                let mut index = 0;
+                for (i, (_, duration)) in imgs.iter().enumerate() {
+                    cumulative_duration += duration.as_millis();
+                    if cumulative_duration >= pos {
```
emilk, line 75: Doing this math in milliseconds seems rather arbitrary. Why not use `Duration` which already supports `saturating_add`?

crates/egui/src/widgets/image.rs
```diff
@@ -43,15 +44,93 @@ use epaint::{util::FloatOrd, RectShape};
...
+                        index = i;
+                        break;
+                    }
+                }
+                let (img, duration) = &imgs[index];
+                ctx.request_repaint_after(*duration);
+                img
+            }
+            ImageSources::Single(img) => img,
```
emilk, line 84: Put the simple case first please

crates/egui_extras/src/loaders/raw_image_loader.rs
```diff
@@ -0,0 +1,80 @@
...
+impl ImageLoader for Rgba8ImageLoader {
+    fn id(&self) -> &str {
+        Self::ID
+    }
+
+    fn load(&self, ctx: &egui::Context, uri: &str, _: SizeHint) -> ImageLoadResult {
+        // three stages of guessing if we support loading the image:
+        // 1. URI extension
+
+        // (1)
```
emilk, line 33:
```suggestion
```

crates/egui_extras/src/image.rs
```diff
@@ -279,3 +283,98 @@ pub fn load_svg_bytes_with_size(
...
+) -> egui::ImageSource<'static> {
+    egui::ImageSource::Bytes {
+        uri: std::borrow::Cow::Owned(format!("rgba8://{}", uri.to_string())),
+        bytes: convert_image_to_bytes(image),
+    }
+}
+
+#[cfg(feature = "gif")]
+#[macro_export]
+macro_rules! include_gif {
```
emilk, line 353: Needs a docstring

crates/egui/src/widgets/image.rs
```diff
@@ -43,15 +44,93 @@ use epaint::{util::FloatOrd, RectShape};
...
+
+impl<'a> From<Vec<(ImageSource<'a>, Duration)>> for ImageSources<'a> {
+    fn from(value: Vec<(ImageSource<'a>, Duration)>) -> Self {
+        Self::Multi(value)
+    }
+}
+
 impl<'a> Image<'a> {
+    /// Loads gif from id & `[ImageSources]`
+    /// `egui::Image::new_animation(egui_extras::include_gif!("../../assets/ferris.gif"))`
```
emilk, line 103: Won't this load the gif each frame?

It feels like this deserves a proper example in `examples/images/src/main.rs` so we can try this out

crates/egui/src/widgets/image.rs
```diff
@@ -43,15 +44,93 @@ use epaint::{util::FloatOrd, RectShape};
...
+}
+
+impl<'a> From<ImageSource<'a>> for ImageSources<'a> {
+    fn from(value: ImageSource<'a>) -> Self {
+        Self::Single(value)
+    }
+}
+
+impl<'a> From<Vec<(ImageSource<'a>, Duration)>> for ImageSources<'a> {
+    fn from(value: Vec<(ImageSource<'a>, Duration)>) -> Self {
```
emilk, line 96: it could be helpful with a `struct AnimatedImageSource<'a>(Vec<(ImageSource<'a>, Duration)>)` to avoid complex types like this inside `ImageSources`

crates/egui/src/widgets/image.rs
JustFrederik: let pos = now % frames won't work with Durations

JustFrederik:
@emilk is this fine?

examples/images/src/main.rs
```diff
@@ -20,18 +20,27 @@ fn main() -> Result<(), eframe::Error> {
...
+            gif: egui::Image::new_animation(egui_extras::include_gif!("ferris.gif")),
+        }
+    }
+}
 
 impl eframe::App for MyApp {
     fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
         egui::CentralPanel::default().show(ctx, |ui| {
             egui::ScrollArea::both().show(ui, |ui| {
+                ui.add(self.gif.clone());
```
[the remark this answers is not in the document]
JustFrederik: it could be changed to something like this:

```rs
macro_rules! include_gif {
    ($path:expr $(,)?) => {
        $crate::ImageSource::Bytes {
            uri: ::std::borrow::Cow::Borrowed(concat!("gif_bytes://", $path)),
            bytes: $crate::load::Bytes::Static(include_bytes!($path)),
        }
    };
}
```
the new function will check if the uri starts with gif_bytes & look it up/call gif_to_sources

The problem is how to cache the gif.  @emilk do you have an idea how to cache the decoded gif?

Something like this would be needed, but how. static could be used, but it would require unsafe code. `fn get_cached(uri: &str) -> Option<AnimatedImageSource<'a>>;`

Issue #3481, opened 2023-10-17
Eframe option to repeat texture


JumpyLionnn:
**Is your feature request related to a problem? Please describe.**
I wanted to create a big checker board pattern and cover the screen with it, it requires a lot of rectangles. It is slow to generate and draw that many quads every frame(i already use some caches).
In addition I wanted to flip an image visually without messing with the image crate or messing with the image data, I want to be able to negate the uv coordinates and flip the image easily.

**Describe the solution you'd like**
The default texture wrap mode is `GL_CLAMP_TO_EDGE`
I suggest an option in the `NativeOptions` to set the repeat texture property or maybe to set the texture wrap mode:
```rust
enum TextureWrapMode{
    Repeat,
    MirroredRepeat,
    ClampToEdge,
    ClampToBorder(Color32)
}

struct NativeOptions {
    ...
    texture_wrap_mode: TextureWrapMode // defaults to ClampToEdge
    // or
    repeat_textures: bool // defaults to false
    ...
}
```
It can be done with calling some OpenGL functions:
```c
glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_REPEAT);
glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_REPEAT);
```
for the clamp to border, the color can be set like that: 
```c
glTexParameterfv(GL_TEXTURE_2D, GL_TEXTURE_BORDER_COLOR, borderColor);
```

**Describe alternatives you've considered**
I considered manipulating pixels to do what I want but it would be too slow to do that, and I only need it visually, I dont need to use the flipped image for something else.
I also tried generating the checkboard pattern using the mesh API with rectangles. While it works, the performance isn't good enough for me and its not a major part of the thing I am working on.

**Additional context**
This is what happens when I try to flip the texture with uv:
```rust
egui::Rect { min: egui::vec2(0.0, 0.0), max: egui::vec2(-1.0, -1.0) }
```
![image](https://github.com/emilk/egui/assets/79366074/0ffb6191-a03e-442e-9286-de023c38e92d)

Thank you for taking the time to review my feature request :)

emilk:
Feel free to make a PR where you add a wrap mode to `TextureOptions`

PR #3954, opened 2024-02-02
Add `TextureOptions::wrap_mode`

CodedNil:
Exposes support in both glow and wgpu for texture wrap modes

This would be breaking for manual creations of TextureOptions but would work with the current TextureOptions::NEAREST and LINEAR without change, keeping those clamp to edge

I wasn't sure how best to expose the options to the user and added consts for LINEAR_REPEAT LINEAR_MIRRORED_REPEAT NEAREST_REPEAT NEAREST_MIRRORED_REPEAT

This does not include wrap mode clamp to border as it worked fine with glow but with wgpu it panics due to Features Features(ADDRESS_MODE_CLAMP_TO_BORDER) are required but not enabled on the device, and I thought it was probably best not to try to enable that feature, but happy to include that functionality also if that is okay to be toggled

image

### DIFF
--- a/crates/epaint/src/textures.rs
+++ b/crates/epaint/src/textures.rs
@@ -156,6 +156,9 @@ pub struct TextureOptions {
 
     /// How to filter when minifying (when texels are smaller than pixels).
     pub minification: TextureFilter,
+
+    /// How to wrap the texture when the texture coordinates are outside the [0, 1] range.
+    pub wrap_mode: TextureWrapMode,
 }
 
 impl TextureOptions {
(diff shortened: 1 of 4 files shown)


YgorSouza:
Link to #3481 so people know it's being worked on.

crates/epaint/src/textures.rs
```diff
@@ -193,6 +226,20 @@ pub enum TextureFilter {
     Linear,
 }
 
+/// Defines how textures are wrapped around objects when texture coordinates fall outside the [0, 1] range.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub enum TextureWrapMode {
+    /// Stretches the edge pixels to fill beyond the texture's bounds.
+    ClampToEdge,
```
emilk, line 234:
```suggestion
#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, Hash)]
#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
pub enum TextureWrapMode {
    /// Stretches the edge pixels to fill beyond the texture's bounds.
    ///
    /// This is what you want to use for a normal image in a GUI.
    #[default]
    ClampToEdge,
```

emilk (approved):
Great!

Issue #3956, opened 2024-02-03
StripBuilder fails to allocate rect


IVAN-MK7:
**Describe the bug**
Since Commit [5a6d1cb](https://github.com/emilk/egui/commit/5a6d1cb) StripBuilder seems to fail to allocate its used space:

![stripbuilder-broken](https://github.com/emilk/egui/assets/68190772/852b0a5f-c439-4000-9792-5e4583a8145c)

Up until the its previous Commit ([37762f7](https://github.com/emilk/egui/commit/37762f7)), this bug was not present.

![stripbuilder-fixed](https://github.com/emilk/egui/assets/68190772/d33df541-f5d3-47df-ae51-4ea641168748)

**To Reproduce**

1. Clone my forked egui repository, branch: "strip-builder-allocation-rect":
`git clone -b "strip-builder-allocation-rect" --single-branch "https://github.com/IVAN-MK7/egui.git"`
2. Open crates\egui_extras\src\layout.rs
3. Erase the line of code commented with` 37762f7 's working_version.`, de-comment the 2 lines of code commented with `5a6d1cb's breaking_change.`
4. Start the egui demo app with `cargo run --release -p egui_demo_app`, open the Strip Demo from the right-hand panel.

**Expected behavior**
The StripBuilder should allocate its used space, so that the following items won't be displayed upon it, but next to it instead.

The bug was instroduced from 5a6d1cb's crates\egui_extras\src\layout.rs by its allocation strategy:
instead of allocating the allocation_rect:
```
let response = self.ui.allocate_rect(allocation_rect, self.sense);
return (used_rect, response)
```

it allocates the max_rect, for then returning a response with the allocation_rect instead:
```
let response = self.ui.allocate_rect(max_rect, self.sense);
let response = response.with_new_rect(allocation_rect);
return (used_rect, response)
```

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version master, Commit 114f820, February 2nd, 2024

PR #3957, opened 2024-02-03
Fix `StripBuilder` not allocating its used space

IVAN-MK7:
At crates\egui_extras\src\layout.rs :

Allocate allocation_rect instead of max_rect.
Go back from this:
```
let response = self.ui.allocate_rect(max_rect, self.sense);
let response = response.with_new_rect(allocation_rect);
return (used_rect, response)
```

to this:
```
let response = self.ui.allocate_rect(allocation_rect, self.sense);
return (used_rect, response)
```

In order to allocate the

Closes <https://github.com/emilk/egui/issues/3956>.

### DIFF
--- a/crates/egui_extras/src/layout.rs
+++ b/crates/egui_extras/src/layout.rs
@@ -145,7 +145,6 @@ impl<'l> StripLayout<'l> {
             );
         }
 
-        let response = self.ui.allocate_rect(max_rect, self.sense);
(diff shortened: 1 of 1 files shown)


emilk:
The reverted code was introduced in:
* https://github.com/emilk/egui/pull/3347

It seems like table row selection still works after reverting those lines, so 👍

emilk (approved):

PR #3958, opened 2024-02-03
add with_taskbar to viewport builder

AnotherNathan:
Allows removing the taskbar icon of a viewport.

This can be useful when making an overlay type viewport (window is always on top).

### DIFF
--- a/crates/egui/src/viewport.rs
+++ b/crates/egui/src/viewport.rs
@@ -292,7 +292,7 @@ pub struct ViewportBuilder {
 
     // windows:
     pub drag_and_drop: Option<bool>,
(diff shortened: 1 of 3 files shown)


crates/egui/src/viewport.rs
```diff
@@ -442,6 +445,13 @@ impl ViewportBuilder {
         self
     }
 
+    /// windows: Whether show or hide the window icon in the taskbar.
+    #[inline]
+    pub fn with_skip_taskbar(mut self, skip: bool) -> Self {
```
emilk, line 450: Please invert this to `with_taskbar` to avoid double negation

emilk (approved):

PR #3960, opened 2024-02-04
Pass `ScreenDescriptor` to `egui_wgpu::CallbackTrait::prepare`

StratusFearMe21:
`glyphon` requires the screen resolution during the `prepare` stage, and passing that to the callback's `prepare` function seems pretty trivial.


emilk (approved):
Looks good, but one small nit: I think thew screen descriptor should come before the mutable parameters to give the parameters a more logical order

PR #3968, opened 2024-02-04
Override text color with stroke selection color for selected cells

### LINKED ISSUE #3967: Table row selection does not respect selection stroke color
The selected table row always uses the text color:
image

**Expected behavior**
The selection stroke color should be used for selected rows.

njust:
Closes #3967 

image


emilk (approved):
Makes sense.

PR #3977, opened 2024-02-05
Wait with showing tooltip until mouse has been still for 300ms

emilk:
You can change this with `style.interaction.tooltip_delay§

### DIFF
--- a/crates/egui/src/response.rs
+++ b/crates/egui/src/response.rs
@@ -499,13 +499,16 @@ impl Response {
             }
         }
 
-        if !self.is_tooltip_open()
-            && self.ctx.input(|i| i.pointer.time_since_last_movement())
-                < self.ctx.style().interaction.tooltip_delay
-        {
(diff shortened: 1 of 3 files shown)


crates/egui/src/response.rs
```diff
@@ -499,13 +499,16 @@ impl Response {
...
-            self.ctx.request_repaint();
-            return false;
+        if !self.is_tooltip_open() {
+            let time_til_tooltip = self.ctx.style().interaction.tooltip_delay
+                - self.ctx.input(|i| i.pointer.time_since_last_movement());
+
+            if 0.0 < time_til_tooltip {
+                // Wait until the mouse has been still for a while
+                self.ctx
+                    .request_repaint_after(std::time::Duration::from_secs_f32(time_til_tooltip));
```
emilk, line 509: Unfortunately this doesn't work because of https://github.com/emilk/egui/issues/3109

Issue #3980, opened 2024-02-05
Using a custom font causes the table to grow infinitely


njust:
Using a custom font causes the table to grow infinitely.

https://github.com/emilk/egui/assets/2380253/278c64bc-04bb-4316-8156-3bd7c7822c1e

Without the changes of this pull request: https://github.com/emilk/egui/pull/3957 the table does not grow.

To reproduce this issue, you can use 

https://github.com/njust/egui/tree/table-grow-issue

and start the egui_demo_app.

PR #3983, opened 2024-02-05
egui demo: Fix space reserved for source code link

njust:
Clip the cell of the source code link to prevent the strip from growing infinitely if the reserved size for the link is not sufficient.

image

Closes <https://github.com/emilk/egui/issues/3980>.

### DIFF
--- a/crates/egui_demo_lib/src/demo/table_demo.rs
+++ b/crates/egui_demo_lib/src/demo/table_demo.rs
@@ -109,6 +109,7 @@ impl super::View for TableDemo {
         StripBuilder::new(ui)
             .size(Size::remainder().at_least(100.0)) // for the table
             .size(Size::exact(10.5)) // for the source code link
+            .clip(true)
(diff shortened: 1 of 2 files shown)


emilk (approved):

crates/egui_demo_lib/src/demo/table_demo.rs
```diff
@@ -109,6 +109,7 @@ impl super::View for TableDemo {
         StripBuilder::new(ui)
             .size(Size::remainder().at_least(100.0)) // for the table
             .size(Size::exact(10.5)) // for the source code link
```
emilk, line 111: alternatively, we change this to reserve `TextStyle::Body.resolve(ui).height()` instead of the hard-coded 10.5

njust: Done

emilk (approved):

emilk:
Thanks!

PR #3984, opened 2024-02-06
TextEdit: fix crash when hitting SHIFT + TAB around non-ASCII text

### LINKED ISSUE #3846: TextEdit : Shift + TAB

rustbasic:
* Closes #3846 
* Closes #3878

Dear emilk.

Leaving aside other function,
I think this is all you need to fix to patch the panic that occurs when Shift + TAB.

Thank you, emilk.

### DIFF
--- a/crates/egui/src/widgets/text_edit/text_buffer.rs
+++ b/crates/egui/src/widgets/text_edit/text_buffer.rs
@@ -89,10 +89,12 @@ pub trait TextBuffer {
     fn decrease_indentation(&mut self, ccursor: &mut CCursor) {
         let line_start = find_line_start(self.as_str(), *ccursor);
 
-        let remove_len = if self.as_str()[line_start.index..].starts_with('\t') {
+        let remove_len = if self.as_str().chars().nth(line_start.index) == Some('\t') {
             Some(1)
-        } else if self.as_str()[line_start.index..]
+        } else if self
+            .as_str()
             .chars()
+            .skip(line_start.index)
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/text_edit/text_buffer.rs
```diff
@@ -89,10 +89,12 @@ pub trait TextBuffer {
...
         let line_start = find_line_start(self.as_str(), *ccursor);
 
-        let remove_len = if self.as_str()[line_start.index..].starts_with('\t') {
+        let remove_len = if self.as_str().chars().nth(line_start.index) == Some('\t') {
             Some(1)
-        } else if self.as_str()[line_start.index..]
+        } else if self
+            .as_str()
             .chars()
+            .skip(line_start.index)
```
Mingun, line 97: Iteration over chars two times can be time consuming for a long text. It is better to do that only once

rustbasic:
@emilk

PR Merge this for egui v0.26.2, Please.
Let's we to next.

crates/egui/src/widgets/text_edit/text_buffer.rs
rustbasic: May be, you right.
However, it only works when you press SHIFT + TAB, so you don't need to change it.

emilk (approved):

Issue #3982, opened 2024-02-05
Windows: High CPU Usage when minimized


SitiSchu:
**Describe the bug**
When minimized a simple egui application will use up to 17% CPU on my CPU while using 0% when not minimized

**To Reproduce**
1. Take a very simple egui application:
```rust
use eframe::egui;

fn main() -> Result<(), eframe::Error> {
    let options = eframe::NativeOptions {
        viewport: egui::ViewportBuilder::default().with_inner_size([320.0, 240.0]),
        ..Default::default()
    };

    eframe::run_simple_native("My egui App", options, move |ctx, _frame| {
        egui::CentralPanel::default().show(ctx, |ui| {
            ui.heading("My egui Application");
        });
    })
}
```
2. Run it
3. Observe CPU usage while open, if the mouse cursor is not in the window it will drop to 0
4. Minimize window
5. CPU Usage will go up a lot

**Expected behavior**
CPU Usage should stay low or at least close to the same CPU Usage when not minized

**Screenshots**
When idle and not minimized:
![image](https://github.com/emilk/egui/assets/18517178/43841c03-6941-48a3-a2e3-ed486ff4af88)
After being minimized:
![image](https://github.com/emilk/egui/assets/18517178/c161b565-68a7-40a8-b781-bcbc040de621)

**Desktop:**
 - OS: Windows 10 21H2 (19044.3086)
 - Version: latest master (56bf8b7)

**Additional context**
This does not happen with 0.25

rustbasic:
@SitiSchu 
Please refer to patch #3985 to fix it.

SitiSchu:
That fixed the issue, thanks!

yakrider:
This issue seems to still exist when the window is hidden (or set invisible) instead of just minimized .. (in win-11)

PR #3985, opened 2024-02-06
Fix high CPU usage on Windows when app is minimized

rustbasic:
patch #3982

### DIFF
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -192,11 +192,12 @@ fn run_and_return(
 
             if let Some(window) = winit_app.window(*window_id) {
                 log::trace!("request_redraw for {window_id:?}");
-                if !window.is_minimized().unwrap_or(false) {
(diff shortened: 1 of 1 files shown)


crates/eframe/src/native/run.rs
```diff
@@ -194,8 +194,10 @@ fn run_and_return(
                 log::trace!("request_redraw for {window_id:?}");
                 if !window.is_minimized().unwrap_or(false) {
                     window.request_redraw();
+                    true
+                } else {
+                    false
                 }
```
emilk, line 200:
```suggestion
                let is_minimized = window.is_minimized().unwrap_or(false);
                if is_minimized {
                    false
                } else {
                    window.request_redraw();
                    true
                }
```

here and below

emilk (commented):
Remember to give your PR descriptive titles. The PR title is what ends up in the changelog!

emilk (approved):

emilk:
Thanks!

PR #3988, opened 2024-02-06
Prevent egui::Window from becoming larger than Viewport

### LINKED ISSUE #3987: Prevent egui::Window from becoming larger than Viewport
There is a need to prevent `egui::Window` from becoming larger than the Viewport.

rustbasic:
* Closes #3987  

There is a need to prevent egui::Window from becoming larger than the Viewport.

( This is basic. This alone is not enough. )

### DIFF
--- a/crates/egui/src/containers/window.rs
+++ b/crates/egui/src/containers/window.rs
@@ -428,7 +428,7 @@ impl<'open> Window<'open> {
         };
 
         let max_width = area.constrain_rect().unwrap_or(Rect::EVERYTHING).width();
-        let max_height = 
+        let max_height =
             area.constrain_rect().unwrap_or(Rect::EVERYTHING).height() - title_bar_height;
         resize.max_size.x = resize.max_size.x.min(max_width);
         resize.max_size.y = resize.max_size.y.min(max_height);
--- a/crates/egui/src/containers/window.rs
+++ b/crates/egui/src/containers/window.rs
@@ -427,7 +427,7 @@ impl<'open> Window<'open> {
         };
 
         let viewport_rect =
-            ctx.input(|i| i.clone().viewport().inner_rect.unwrap_or(Rect::EVERYTHING));
+            ctx.input(|i| i.viewport().inner_rect.unwrap_or(Rect::EVERYTHING));
         let max_width = viewport_rect.width();
         let max_height = viewport_rect.height() - title_bar_height;
         resize.max_size.x = resize.max_size.x.min(max_width);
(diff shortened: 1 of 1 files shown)


YgorSouza:
@rustbasic if you put "Closes #3987" in the PR description like in the template, GitHub will automatically link the issue to the PR. That makes it easier to see what is being worked on.

rustbasic:
Oh, thank you.

Also, if there are any problems with the source code I wrote, please let me know at any time.

crates/egui/src/containers/window.rs
```diff
@@ -426,6 +426,12 @@ impl<'open> Window<'open> {
             0.0
         };
 
+        let viewport_rect = ctx.input(|i| i.viewport().inner_rect.unwrap_or(Rect::EVERYTHING));
```
emilk, line 429: I think it makes more sense to constrain it to `area.constrain_rect`

emilk (commented):
Should we apply this to all windows or just resizable ones?
Maybe just those with scroll area?
Should it be opt-in or opt-out?
What is the effect of taking e.g. the `WidgetGallery` window in the demo, and putting it in a too small window?

crates/egui/src/containers/window.rs
rustbasic: you're right. May be, It's what I wanted.

rustbasic: It's what I wanted right.

```
        let viewport_rect = ctx.input(|i| i.viewport().inner_rect.unwrap_or(Rect::EVERYTHING));
        let max_width = viewport_rect.width();
        let max_height = viewport_rect.height() - title_bar_height;
```

using `area.constrain_rect` does not limit window from continuing to grow.

rustbasic:
> issue #3987
> 
> There is a need to prevent egui::Window from becoming larger than the Viewport.

crates/egui/src/containers/window.rs
```diff
@@ -429,6 +429,12 @@ impl<'open> Window<'open> {
             (0.0, 0.0)
         };
 
+        let viewport_rect = ctx.input(|i| i.viewport().inner_rect.unwrap_or(Rect::EVERYTHING));
+        let max_width = viewport_rect.width();
+        let max_height = viewport_rect.height() - title_bar_height;
+        resize.max_size.x = resize.max_size.x.min(max_width);
+        resize.max_size.y = resize.max_size.y.min(max_height);
```
emilk, line 436: what I meant was this:

```suggestion
        let max_rect = area.constrain_rect.unwrap_or_else(|| ui.ctx().screen_rect);
        let max_width = max_rect.width();
        let max_height = max_rect.height() - title_bar_height;
        resize.max_size.x = resize.max_size.x.min(max_width);
        resize.max_size.y = resize.max_size.y.min(max_height);
```

PR #4000, opened 2024-02-07
epaint: Break up memozation of text layouts into paragraphs

### LINKED ISSUE #3086: Optimizing TextEdit
# The problem

egui's TextEdit becomes sluggish when it holds large amounts of text; probably, this is due to it rendering all of the text without doing anything clever to cull text which doesn't make it on to the screen (apologies if this is the wrong terminology, I'm not an expert in graphics programming). Even without doing anything clever, there seem to be some optimisation opportunities inside the code, but I wanted to seek some guidance on whether this is worth doing or not before committing to working on this.

At least, with my test program (see end, adapted from #2799), when there are 15 million characters inside the widget, it is noticably laggy to use (>1s per character insert, compiled with `--release`).

I've done some crude `println!` profiling with `std::time::Instant::now()` to find the hot loops, as I will detail below.

Related issues: #1196 #2799 #2906

## Profiling

Here are the measured slow parts of the result of inserting one 'a' into the middle of the buffer (took around ~2s on release mode):

https://github.com/emilk/egui/blob/9478e50d012c5138551c38cbee16b07bc1fcf283/crates/epaint/src/text/text_layout.rs#L110-L136

Took ~0.5s to complete the loop, and reallocated `paragraph.glyphs` 23 times. I think with a call to `Vec::reserve_exact` the allocator pressure can be eliminated entirely, as it should be possible to determine how big each vector is supposed to be. 

https://github.com/emilk/egui/blob/9478e50d012c5138551c38cbee16b07bc1fcf283/crates/epaint/src/text/text_layout.rs#L64

Took ~0.5s to complete.

https://github.com/emilk/egui/blob/9478e50d012c5138551c38cbee16b07bc1fcf283/crates/epaint/src/text/text_layout.rs#L480-L485

Took ~1s to complete.

## Test program

```rust
use eframe::egui::{self, CentralPanel, TextEdit};

        ..Default::default()
    };

    eframe::run_native(
        "editor big file test",
        options,
        Box::new(|_cc| Box::<MyApp>::new(MyApp::new())),
    )
}

struct MyApp {
    text: String,
}

impl MyApp {
    fn new() -> Self {
        let bytes = (0..500000).flat_map(|_| (0u8..10));
        let string: String = bytes.map(|b| format!(" {b:02x}")).collect();
        MyApp { text: string }
    }
}

impl eframe::App for MyApp {
            let code_editor = TextEdit::multiline(&mut self.text)
                .code_editor()
                .desired_width(f32::INFINITY)
                .desired_rows(40);
            let _output = code_editor.show(ui);
        });
    }
}
```

dacid44:
Draft for now, I just want to confirm that I'm on the right track for the general logic, and make sure I know about the edge cases when splitting `LayoutJob`s and merging `Galley`s. It looks like I will need to:
- [ ] Filter which `LayoutSection`s go into each `LayoutJob`, and change their offsets
- [ ] Possibly make sure each `LayoutJob` ends with a newline
- [ ] When merging `Galley`s, merge the `rect`s and `mesh_bounds` of each `Galley` together, and possibly offset them?
- [ ] Make sure the `elided` field makes it onto the full `Galley` if set on any paragraph, and stop adding the following `Galley`s if that field is set
- [ ] Sum `num_vertices` and `num_indices`
- [ ] Anything else?

The code compiles right now, but does not pass `cargo cranky` and panics immediately when that codepath is run (byte index out of bounds, probably due to not filtering and offsetting `LayoutSection`s. Also the `todo!()`.)

* Closes <https://github.com/emilk/egui/issues/3086>

### DIFF
--- a/crates/epaint/src/text/fonts.rs
+++ b/crates/epaint/src/text/fonts.rs
@@ -1,4 +1,4 @@
-use std::{collections::BTreeMap, sync::Arc};
+use std::{collections::BTreeMap, ops::Deref, sync::Arc};
 
 use crate::{
     mutex::{Mutex, MutexGuard},
@@ -703,13 +703,40 @@ impl GalleyCache {
                 cached.galley.clone()
             }
             std::collections::hash_map::Entry::Vacant(entry) => {
-                let galley = super::layout(fonts, job.into());
-                let galley = Arc::new(galley);
-                entry.insert(CachedGalley {
-                    last_used: self.generation,
-                    galley: galley.clone(),
-                });
-                galley
+                // Optimize by splitting the job into paragraphs, if possible.
+                if job.break_on_newline && job.text.contains('\n') {
+                    let mut iter = job.text.split('\n').map(|paragraph| {
+                        // TODO(dacid44): There is almost definitely more logic required to
+                        // properly split up a LayoutJob (i.e., splitting the LayoutSections).
+                        let job = LayoutJob {
+                            text: paragraph.to_owned(),
+                            ..job.clone()
+                        };
+                        self.layout(fonts, job)
+                    });
+                    let first_galley = iter.next()
+                        .expect("if a '\n' is found there should be at least 2 split parts")
+                        .deref()
+                        .clone();
+                    let galley = Arc::new(iter.fold(first_galley, |galley, next_galley| {
+                        // TODO(dacid44): Figure out merging of Galleys.
+                        todo!()
+                    }));
+                    // Can't use the entry, needed to drop it to recursively call self.layout().
+                    self.cache.insert(hash, CachedGalley {
+                        last_used: self.generation,
+                        galley: galley.clone(),
+                    });
+                    galley
+                } else {
+                    let galley = super::layout(fonts, job.into());
+                    let galley = Arc::new(galley);
+                    entry.insert(CachedGalley {
+                        last_used: self.generation,
+                        galley: galley.clone(),
+                    });
+                    galley
+                }
             }
         }
     }
--- a/crates/epaint/src/text/fonts.rs
+++ b/crates/epaint/src/text/fonts.rs
@@ -1,4 +1,4 @@
-use std::{collections::BTreeMap, ops::Deref, sync::Arc};
+use std::{collections::BTreeMap, sync::Arc};
 
 use crate::{
     mutex::{Mutex, MutexGuard},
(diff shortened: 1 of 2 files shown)


dacid44:
This is currently broken. No matter how many things I try to add the y-offset of the previous galleys to, all of the paragraphs render in the top left corner of the text box, on top of each other. I'd appreciate any insights as to why this might be.
image


Also, splitting up the `LayoutJob` (or at least, my current implementation of it) doesn't preserve the indices of the sections. Will I have to offset `Glyph.section_index` to correct this as well?

haricot:
My two cents:
 

The current Paragraph struct is misnamed, as it refers to a block of text that will be split into lines, not a typographic paragraph "/n/n". TextBlock would be a more appropriate name.

When generating the TextBlocks, only the X positioning of glyphs is calculated. The Y positions are ignored at this stage.

Line breaks happen later when converting TextBlocks to Rows, based on the wrap width. A TextBlock can span multiple Rows if it exceeds the wrap width.

To properly position the wrapped Rows vertically, each TextBlock needs a paragraph index that is incremented for each one.

When generating the Rows in line_break(), the paragraph (TextBlock )  index can be used to increment the Y position, so Rows from the same TextBlock are offset vertically.

This prevents the issue of Rows overlapping each other at the same Y position when breaking a TextBlock into multiple Rows.

So in summary, the TextBlocks are intermediate storage of positioned glyphs before line wrapping. The line breaks and Y positioning happen in a later stage when converting TextBlocks to Rows.


Example code:
```
//text_layout.rs

struct TextBlock { // previously Paragraph
  textblock_index: u32, // Added index  
  // other fields..
}

fn layout_section() {

  let mut textblock_index = 0; // Counter

  // In section loop:
  

  let textblock = TextBlock::new(); 
  
  textblock.textblock_index = textblock_index; // Set index
  textblock_index += 1; // Increment counter

  // Rest of section handling...

}

fn line_break() {

  let index = textblock.textblock_index; // Get index

  // In row generation loop:

  let y_offset = index * line_height; // Calc offset
  row.y += y_offset; // Apply offset to Y

  // Rest of row handling...

}
```

dacid44:
Huh, it says your comment was posted on the 10th, but I've been regularly checking this PR and it didn't appear for me until today. Weird.

Anyway, it looks like all use of the `Paragraph` struct seems to happen with the `layout()` function in `text_layout.rs`. I hadn't actually even looked deep enough to see that the `Paragraph` struct existed, as I was mostly treating that function as a black box, and trying to merge `Galley`s together after they are returned from that function. Are you recommending modifying it? The original proposal was to still cache rendering at the `Galley` level, but it could actually be better to memoize further down.

My main question is: what is the frame of reference for the rendering positions inside a `Galley`? If I want to position a `Galley` directly below another one and merge their rows (and thus glyphs) together, what do I need to change to position them properly?

haricot:
Maybe instead of using `galley_from_rows` function here in function layout we could use another function `galley_from_rows_with_previous_offset_y`  or `galley_from_rows_with_full_previous_len_lines_height` ,
so that the paragraphs composed of rows do not overlap (I allow myself to temporarily call them "textBlock" or "textChunk" to avoid any ambiguity).
It may also be necessary to customize ` layout`  functions to receive the previous information.
 
https://github.com/emilk/egui/blob/3672b150ed2c63f87f2a2c8c86ba639f0bbc7d81/crates/epaint/src/text/text_layout.rs#L68-L122

enomado:
I dont get the solution, may be my problem is different than solving here. I want to show BIG 2 megabytes text, so why just dont virtualize lines, and make user pass big text as `Vec<String>`, and virtualize it. 

And make selection api as Range<(row, col)>

emilk:
Closing: per-paragraph layout caching landed in #5411. Thanks for the PR!

PR #4006, opened 2024-02-08
Response Sense Recognition Issue

### LINKED ISSUE #3989: `Response::interact` doesn't always work
The Sense recognition function does not work for `Image` objects, so `response.interact(egui::Sense::click())` is used to check for mouse clicks. 
However, after the recent update, left-clicking and right-clicking on an image are recognized about 1 in 100 times.

```rust
        let response = ui.add(image.show_loading_spinner(false));
        let response = response.interact(egui::Sense::click());
        if response.clicked() {
        }
        if response.secondary_clicked() {
        }
```

egui Version: v0.26.0 & master branch

rustbasic:
Closes #3989

@emilk 
Please review to see if there are any problems.

This Pull Request is Invalid `clip_rect`.
(For example, when the image is enlarged.)

self.ctx.layer_painter(self.layer_id).clip_rect() => Invalid
self.ctx.used_rect() => Invalid
self.ctx.available_rect() => Invalid
self.ctx.screen_rect() => Invalid
self.rect => Invalid

Is there a way to include `ui.clip_rect` in `Response` ?
or
Is there a way to not include `ui.clip_rect` in `Response` ?

I think ui.clip_rect needs to be included in Response.
ui.clip_rect To Response.clip_rect

### DIFF
--- a/crates/egui/src/response.rs
+++ b/crates/egui/src/response.rs
@@ -596,6 +596,15 @@ impl Response {
     /// ```
     #[must_use]
     pub fn interact(&self, sense: Sense) -> Self {
+        self.ctx.interact_with_hovered(
+            self.layer_id,
+            self.id,
+            self.rect,
+            sense,
+            self.enabled,
+            self.hovered,
+        )
+        /*
         self.ctx.interact(
             self.ctx.layer_painter(self.layer_id).clip_rect(),
             self.ctx.style().spacing.item_spacing,
@@ -605,6 +614,7 @@ impl Response {
             sense,
             self.enabled,
         )
+        */
     }
 
     /// Adjust the scroll position until this UI becomes visible.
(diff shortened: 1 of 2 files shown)


rustbasic:
Wait... Problem.

rustbasic:
I think,
This PR is for reference by emilk in fixing bugs.

emilk:
Before we switch away from `interact_with_hovered` I want to understand why it no longer works

rustbasic:
The moment you click, `self.hovered` changes from `true` to `false`, and false is passed to self.ctx.interact_with_hovered().

rustbasic:
Shouldn’t it be this way?

1. First, save all information with click_and_drag().
2. Then, information is removed according to the Sense settings.

emilk:
Closed by #4013

PR #4008, opened 2024-02-08
eframe: Added `App::raw_input_hook` allows for the manipulation or filtering of raw input events

varphone:
# What's New

* eframe: Added `App::raw_input_hook` allows for the manipulation or filtering of raw input events
   A filter applied to raw input before [`Self::update`]
   This allows for the manipulation or filtering of input events before they are processed by egui.
   This can be used to exclude specific keyboard shortcuts, mouse events, etc.
   Additionally, it can be used to add custom keyboard or mouse events generated by a virtual keyboard.
* examples: Added an example to demonstrates how to implement a custom virtual keyboard.

[eframe-custom-keypad.webm](https://github.com/emilk/egui/assets/1274171/a9dc8e34-2c35-4172-b7ef-41010b794fb8)


emilk (approved):
Very cool example!

crates/eframe/src/epi.rs
```diff
@@ -196,6 +196,15 @@ pub trait App {
...
     }
+
+    /// A filter applied to raw input before [`Self::update`]
+    ///
+    /// This allows for the manipulation or filtering of input events before they are processed by egui.
+    ///
+    /// This can be used to exclude specific keyboard shortcuts, mouse events, etc.
+    ///
+    /// Additionally, it can be used to add custom keyboard or mouse events generated by a virtual keyboard.
+    fn filter_raw_input_inplace(&mut self, _ctx: &egui::Context, _raw_input: &mut egui::RawInput) {}
```
emilk, line 207: Can we come up with a better name? It is more than just filtering 🤔

varphone: How abour `on_raw_input` or `raw_input_hijack` ?

rustbasic: on_raw_input()
simulate_input()
send_input()
fake_input()

emilk: I like `raw_input_hook`

varphone: OK, it has been changed to `raw_input_hook`.

PR #4009, opened 2024-02-08
egui: Fix custom `Window` `Frame`s

### LINKED ISSUE #4002: Custom frames for windows not applied correctly
**Describe the bug**
Since version 0.25 there is a visual bug when we set a custom window frame. In 0.24 the code below was making the window all green (as expected) with little margin.

```rust
let frame = egui::containers::Frame::none()
    .fill(egui::Color32::from_rgb(180, 220, 160))
    .inner_margin(egui::Margin::symmetric(20.0, 20.0));
```

In 0.25 and 0.26 the same code makes it only partially green with title bar area still with its original color, which is also not properly adjusted to the title.

image

**To Reproduce**
1. Go to egui_demo_lib -> window_options.rs
2. Find creation of "demo_window_options" window
3. Make it as below:

```rust
use super::View as _;
let frame = egui::containers::Frame::none() // <-- new
	.fill(egui::Color32::from_rgb(180, 220, 160)) // <-- new
	.inner_margin(egui::Margin::symmetric(20.0, 20.0)); // <-- new
let mut window = egui::Window::new(title)
	.id(egui::Id::new("demo_window_options"))
	.frame(frame) // <-- new
	.resizable(resizable)
	.constrain(constrain)
	.collapsible(collapsible)
	.title_bar(title_bar)
	.scroll2(scroll2)
	.enabled(enabled);
```

**Expected behavior**
Just as in 0.24 and earlier, the whole window area should be colored green with no exception for title bar.

varphone:
Fix #4002 

fixed-#4002

### DIFF
--- a/crates/egui/src/containers/window.rs
+++ b/crates/egui/src/containers/window.rs
@@ -390,11 +390,13 @@ impl<'open> Window<'open> {
             with_title_bar,
         } = self;
 
-        let frame = frame.unwrap_or_else(|| Frame::window(&ctx.style()));
+        let header_color =
+            frame.map_or_else(|| ctx.style().visuals.widgets.open.weak_bg_fill, |f| f.fill);
+        let window_frame = frame.unwrap_or_else(|| Frame::window(&ctx.style()));
 
         let is_explicitly_closed = matches!(open, Some(false));
         let is_open = !is_explicitly_closed || ctx.memory(|mem| mem.everything_is_visible());
-        area.show_open_close_animation(ctx, &frame, is_open);
+        area.show_open_close_animation(ctx, &window_frame, is_open);
 
         if !is_open {
             return None;
@@ -419,7 +421,7 @@ impl<'open> Window<'open> {
         // Calculate roughly how much larger the window size is compared to the inner rect
         let (title_bar_height, title_content_spacing) = if with_title_bar {
             let style = ctx.style();
-            let window_margin = style.spacing.window_margin;
+            let window_margin = window_frame.inner_margin;
             let spacing = window_margin.top + window_margin.bottom;
             let height = ctx.fonts(|f| title.font_height(f, &style)) + spacing;
             (height, spacing)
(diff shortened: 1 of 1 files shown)


crates/egui/src/containers/window.rs
```diff
@@ -419,7 +423,11 @@ impl<'open> Window<'open> {
         // Calculate roughly how much larger the window size is compared to the inner rect
         let (title_bar_height, title_content_spacing) = if with_title_bar {
             let style = ctx.style();
-            let window_margin = style.spacing.window_margin;
+            let window_margin = if is_frame_overrode {
+                frame_margin
+            } else {
+                style.spacing.window_margin
```
emilk, line 429: It feels like `Frame::window` should just set this as `inner_margin` instead of having an `if` statement here

crates/egui/src/containers/window.rs
```diff
@@ -518,12 +526,16 @@ impl<'open> Window<'open> {
                             y: title_bar_height,
                         },
                     );
-                    let mut round = area_content_ui.visuals().window_rounding;
+                    let (mut round, header_color) = if is_frame_overrode {
+                        (frame_rounding, frame_fill)
+                    } else {
+                        let visuals = area_content_ui.visuals();
+                        (visuals.window_rounding, visuals.widgets.open.weak_bg_fill)
```
emilk, line 533: Same here, `Frame::window` should just set this values instead imho

crates/egui/src/containers/window.rs
```diff
@@ -390,7 +390,11 @@ impl<'open> Window<'open> {
             with_title_bar,
         } = self;
 
+        let is_frame_overrode = frame.is_some();
         let frame = frame.unwrap_or_else(|| Frame::window(&ctx.style()));
+        let frame_fill = frame.fill;
+        let frame_margin = frame.inner_margin;
+        let frame_rounding = frame.rounding;
```
emilk, line 397: this feels like unnecessary pollution of the namespace. Why not use `frame.fill` later?

crates/egui/src/containers/window.rs
emilk: EDIT: `Frame::window` already has this as the window margin. You can just use `frame.inner_margin` here

crates/egui/src/containers/window.rs
varphone: OK, it has been modified, please check again.

crates/egui/src/containers/window.rs
varphone: OK, it has been modified, please check again.

emilk (approved):
Thanks!

Issue #3997, opened 2024-02-07
`ui.on_disabled_hover_ui()` and friend doesn't respect `tooltip_delay`


abey79:
As the title says ☝🏻 

`on_disabled_hover_ui()` does a simple pointer-in-rect check whereas `on_hover_ui()` has a proper logic that includes testing for `tooltip_delay`:
<img width="707" alt="image" src="https://github.com/emilk/egui/assets/49431240/4db27682-6ee6-4b65-850f-993c45cfaa01">

PR #4012, opened 2024-02-09
Make `on_disabled_hover_ui` respect `tooltip_delay`

YgorSouza:
* Closes <https://github.com/emilk/egui/issues/3997>

### DIFF
--- a/crates/egui/src/response.rs
+++ b/crates/egui/src/response.rs
@@ -509,8 +509,9 @@ impl Response {
 
             if 0.0 < time_til_tooltip {
                 // Wait until the mouse has been still for a while
-                self.ctx
-                    .request_repaint_after(std::time::Duration::from_secs_f32(time_til_tooltip));
+                if let Ok(duration) = std::time::Duration::try_from_secs_f32(time_til_tooltip) {
+                    self.ctx.request_repaint_after(duration);
+                }
                 return false;
             }
         }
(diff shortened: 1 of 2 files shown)


rustbasic:
great.

Future Request:
It would be nice to be able to set the delay time.
If you set it to 0.0, it will have a disable effect.

YgorSouza:
@rustbasic it is already possible to set the delay. Introduced in #3245. If you mean you want to use this value to disable the tooltip altogether, I was thinking you could do that by setting it to infinity, but actually this makes the app panic due to a conversion to `Duration`, so I may as well fix that in this branch.

rustbasic:
I see.
I will look forward to the future.

With the following structure, it seems possible to set delay and duration times as well:

(example)
tooltip_delay
tooltip_duration
tooltip_state
tooltip_start_time
tooltip_end_time

1. Create these five variables to set the delay and duration times.
2. When hovering over the widget to display the tooltip, use the current time along with tooltip_delay and tooltip_duration to input tooltip_start_time and tooltip_end_time.
3. If the current time is between tooltip_show_time and tooltip_hide_time, and the widget is in hover state, display the tooltip and set tooltip_state to true.
4. If the current time is not between tooltip_start_time and tooltip_end_time, modify tooltip_state from true to false, and do not display the tooltip.

Have a good day.

emilk (approved):
Great; thanks you!

abey79:
Thanks @YgorSouza!

Issue #4017, opened 2024-02-10
Take `clip_rect` into account when storing widgets in `interact_with_hovered`


rustbasic:
@emilk

Update Issue : #4013 

After update #4013, you will no longer be able to click any button on an image (in fact any button) that is not visible on screen when the image is enlarged.

If you read this post, please mark it with an eye emoji. 

--------------------------------------------
Reference :

May be,
Should we save the results of all events first?

egui/crates/egui/src/widgets/image.rs : fn ui(self, ui: &mut Ui)
337 line
let (rect, response) = ui.allocate_exact_size(ui_size, self.sense);
to
let (rect, response) = ui.allocate_exact_size(ui_size, egui::Sense::click_and_drag());

it's a good result.
But, One problem found in this case is that the Image in the ScrollArea cannot be dragged with the mouse after being enlarged.
then,

`let (rect, response) = ui.allocate_exact_size(ui_size, egui::Sense::click_and_drag());`
to
`let (rect, response) = ui.allocate_exact_size(ui_size, egui::Sense::click());`
it's a good result.

Or, Should it be this way?

First, save all information with click_and_drag().
Then, information is removed according to the Sense settings.
May be, Remove it from the interact() function
Please refer to the contents so far.

rustbasic:
Should we use widget_rect & clip_rect ?

emilk:
I don't follow this at all. Is there a new bug?

Can you explain with an image or a video? Maybe some code to reproduce it?

And please write good titles. Put yourself in my choses reading this:

![Screenshot 2024-02-10 at 12 34 33](https://github.com/emilk/egui/assets/1148717/52789d59-1fab-431a-8adb-bb5c619107f5)

rustbasic:
When you enlarge the displayed image, only part of the image is visible on the screen, but the image is recognized as displaying the entire screen.
So all buttons outside the image cannot be clicked.

In this case, the button doesn't even display hover.

rustbasic:
![20240210-1](https://github.com/emilk/egui/assets/127506429/47bc5cb9-d28f-4486-8eb1-18e8087c4a72)
![20240210-2](https://github.com/emilk/egui/assets/127506429/d2181f97-3f78-4b0f-adfc-9823eb833feb)

When you enlarge the displayed image, only part of the image is visible on the screen, but the image is recognized as displaying the entire screen.
So all buttons outside the image cannot be clicked.

In this case, the button doesn't even display hover.

emilk:
[...]

PR #4020, opened 2024-02-10
Take clip_rect into account when storing widget rects

emilk:
* Bug introduced in https://github.com/emilk/egui/pull/4013
* Closes https://github.com/emilk/egui/issues/4017

Unfortunately this is a breaking change, since it changes the fields of `Response`, so can't do a patch-release with this.


rustbasic:
It's Good.

Thank you for your hard work today. have a good day.

PR #4021, opened 2024-02-10
Release 0.26.1

emilk:
Some bug fixes!

### DIFF
--- a/examples/hello_world/src/main.rs
+++ b/examples/hello_world/src/main.rs
@@ -49,9 +49,20 @@ impl eframe::App for MyApp {
             }
             ui.label(format!("Hello '{}', age {}", self.name, self.age));
 
-            ui.image(egui::include_image!(
-                "../../../crates/egui/assets/ferris.png"
-            ));
+            // ui.image(egui::include_image!(
+            //     "../../../crates/egui/assets/ferris.png"
+            // ));
+
+            let (response, painter) =
+                ui.allocate_painter(egui::vec2(300.0, 150.0), egui::Sense::click());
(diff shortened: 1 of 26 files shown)


emilk:
@rustbasic can you please test if this branch still fixes your image-interaction bugs?

I had to roll back part of my fix to not break semver.

examples/hello_world/src/main.rs
```diff
@@ -49,9 +49,20 @@ impl eframe::App for MyApp {
             }
             ui.label(format!("Hello '{}', age {}", self.name, self.age));
 
-            ui.image(egui::include_image!(
-                "../../../crates/egui/assets/ferris.png"
-            ));
+            // ui.image(egui::include_image!(
```
YgorSouza, line 52: Is this change to the example intentional? It looks like test code.

emilk: oops! thanks for catching that 😅

rustbasic:
Yes.
Once the v0.26.1 PR is finished, should that test it?

emilk:
I think this is ready to go (I need to generate changelogs etc first ofc), but I would appreciate you testing it @rustbasic since you've been very good at finding bugs

rustbasic:
This is good.
I tested it with the contents of the current master branches, and there is no problem.
  You did a really good job.

rustbasic:
@emilk 

Bug : 
When you open the Viewport, the speed drops to 1 frame per second.

rustbasic:
@emilk 

In egui v0.26.1, many bugs have been improved, and the overall feeling of use is good.
You did a really good job.

PR #4029, opened 2024-02-12
Fix: images with background color now respects rounding

vincent-sparks:
Code:
```rust
use egui::{TextureHandle, Image, Vec2, Layout, Direction};

struct App {
    texture: TextureHandle,
}

const ROW_HEIGHT: f32 = 200.0;

impl App {
    fn new(ctx: &eframe::CreationContext) -> Self {
        Self {
            texture: ctx.egui_ctx.load_texture("sample", egui::ColorImage::example(), egui::TextureOptions::default()),
        }
    }
}

impl eframe::App for App {
        egui::CentralPanel::default().frame(egui::Frame::dark_canvas(&ctx.style())).show(ctx, |ui| {
            ui.with_layout(Layout::centered_and_justified(Direction::TopDown), |ui|{
                ui.add(Image::new(&self.texture)
                       .bg_fill(egui::Color32::WHITE)
                       .rounding(30.0)
                      );
            });
        });
    }
}

    eframe::run_native("egui test", eframe::NativeOptions{
        viewport: egui::ViewportBuilder::default().with_inner_size((256.,200.)),
        ..Default::default()
    }, Box::new(|cc| Box::new(App::new(cc))))
}
```

Before: 
before

After:
after

(please see the comment in the diff -- *definitely* not sure if it's up to code standards.  you might want to change/remove that comment as well as the Mesh codepath if it was not, in fact, done for performance reasons)

### DIFF
--- a/crates/egui/src/widgets/image.rs
+++ b/crates/egui/src/widgets/image.rs
@@ -739,17 +739,7 @@ pub fn paint_texture_at(
     texture: &SizedTexture,
 ) {
     if options.bg_fill != Default::default() {
-        // NOTE(vincent-sparks) originally this was only done with a mesh;
-        // however this created a bug if the image had rounding
-        // I assume the mesh was used in place of the rect for performance reasons
-        // so I have left it in
-        if options.rounding == Rounding::ZERO {
-            let mut mesh = Mesh::default();
-            mesh.add_colored_rect(rect, options.bg_fill);
-            painter.add(Shape::mesh(mesh));
-        } else {
-            painter.add(RectShape::new(rect, options.rounding, options.bg_fill, Stroke::NONE));
-        }
+        painter.add(RectShape::filled(rect, options.rounding, options.bg_fill));
     }
 
     match options.rotation {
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/image.rs
```diff
@@ -739,9 +739,17 @@ pub fn paint_texture_at(
...
+        // however this created a bug if the image had rounding
+        // I assume the mesh was used in place of the rect for performance reasons
+        // so I have left it in
+        if options.rounding == Rounding::ZERO {
+            let mut mesh = Mesh::default();
+            mesh.add_colored_rect(rect, options.bg_fill);
+            painter.add(Shape::mesh(mesh));
+        } else {
+            painter.add(RectShape::new(rect, options.rounding, options.bg_fill, Stroke::NONE));
+        }
```
emilk, line 752: I'm not sure there was a good reason for using the mesh before.
`RectShape` has the advantage of providing anti-aliasing, which would be noticeable if the image is moving smoothly across the screen.
```suggestion
        painter.add(RectShape::filled(rect, options.rounding, options.bg_fill));
```

emilk (approved):

emilk (approved):

PR #4035, opened 2024-02-12
Add `ViewportCommand::RequestCut`, `RequestCopy` and `RequestPaste` to trigger Clipboard actions

bu5hm4nn:
### Motivation
We want to offer our users a context menu with `Cut`, `Copy` and `Paste` actions. `Paste` is not possible out of the box.

### Changes
This PR adds `ViewportCommand::RequestCut`, `ViewportCommand::RequestCopy` and `ViewportCommand::RequestPaste`. They are routed and handled after the example of `ViewportCommand::Screenshot` and result in the same code being executed as when the user uses `CTRL+V` style keyboard commands.

### Reasoning
In our last release we used an instance of `egui_winit::clipboard::Clipboard` in order to get the `Paste` functionality.

However Linux users on Wayland complained about broken clipboard interaction (https://github.com/mikedilger/gossip/issues/617). After a while of digging I could not find the issue although I have found references to problems with multiple clipboards per handle before (https://gitlab.gnome.org/GNOME/mutter/-/issues/1250) but I compared mutter with weston and the problem occured on both.

So to solve this I set out to extend egui to access the clipboard instance already present in egui_winit. Since there was no trivial way to reach the instance of `egui_winit::State` I felt the best approach was to follow the logic of the new `ViewportCommand::Screenshot`.

### Variations
It could make sense to make the introduced `enum ActionRequested` a part of crates/egui/src/viewport.rs and to then wrap them into one single `ViewportCommand::ActionRequest(ActionRequested)`.

### Example
```Rust
let mut text = String::new();
let response = ui.text_edit_singleline(&mut text);
if ui.button("Paste").clicked() {
    response.request_focus();
    ui.ctx().send_viewport_cmd(ViewportCommand::RequestPaste);
}
```


crates/egui/src/viewport.rs
```diff
@@ -965,6 +965,15 @@ pub enum ViewportCommand {
...
     Screenshot,
+
+    /// Request a cut from the clipboard
+    RequestCut,
+
+    /// Request a copy from the clipboard
+    RequestCopy,
+
+    /// Request a paste from the clipboard
+    RequestPaste,
```
emilk, line 976: These needs better docstrings, because I don't get it. Are these effectively injecting events in the next frame?

bu5hm4nn: They are doing the equivalent of pressing `CTRL + C`, `CTRL + X` and `CTRL + V` (or the CMD equivalent on macOS). The reason we had to introduce this is because adding a second clipboard instance broke clipboard interaction for Wayland users. So in order to offer users a context menu with these actions, we needed a way to trigger them. Since we cannot programmatically emit keypresses currently, this seemed like a viable solution.

So maybe the docstring could say `Trigger a CMD + X cut` or something like that

emilk: I see.

Perhaps "Reads text from the system clipboard and pastes it in the next frame"

emilk (commented):
This should be implemented in the web backend too!

bu5hm4nn:
> This should be implemented in the web backend too!

I have never used egui for web and I can't find uses of ViewportCommand outside of winit, I'm afraid I can't help with that.

emilk:
The relevant code is here: 

https://github.com/emilk/egui/blob/ab6c3f91611268e0b86c03d7b7f3a51dec7f7692/crates/eframe/src/web/app_runner.rs#L203-L210

and you can test with the instruction in https://github.com/emilk/egui/blob/master/CONTRIBUTING.md#testing-the-web-viewer

Implementing `RequestPaste` will be a bit of work though, as it requires an async request (a mirror to `set_clipboard_text` in  `crates/eframe/src/web/mod.rs`

bu5hm4nn:
I took a look, but I am sorry, I do not have time to get into that code. Currently web does not handle any of the ViewportCommand's so I personally don't see any hurt in it currently not supporting them.

emilk:
> I took a look, but I am sorry, I do not have time to get into that code. Currently web does not handle any of the ViewportCommand's so I personally don't see any hurt in it currently not supporting them.

Fair point

crates/egui/src/viewport.rs
```diff
@@ -965,6 +965,18 @@ pub enum ViewportCommand {
...
+    /// this is equivalent to the system keyboard shortcut for cut (e.g. CTRL + X)
+    RequestCut,
+
+    /// Request a copy of the current selection
+    /// this is equivalent to the system keyboard shortcut for copy (e.g. CTRL + C)
+    RequestCopy,
+
+    /// Request a paste from the clipboard to the current focused TextEdit if any
+    /// this is equivalent to the system keyboard shortcut for paste (e.g. CTRL + V)
+    RequestPaste,
```
emilk, line 979:
```suggestion

    /// Request cut of the current selection
    ///
    /// This is equivalent to the system keyboard shortcut for cut (e.g. CTRL + X).
    RequestCut,

    /// Request a copy of the current selection.
    ///
    /// This is equivalent to the system keyboard shortcut for copy (e.g. CTRL + C).
    RequestCopy,

    /// Request a paste from the clipboard to the current focused TextEdit if any.
    ///
    /// This is equivalent to the system keyboard shortcut for paste (e.g. CTRL + V).
    RequestPaste,
```

crates/eframe/src/native/glow_integration.rs
```diff
@@ -986,7 +1008,7 @@ impl GlutinWindowContext {
                 class: ViewportClass::Root,
                 builder: viewport_builder,
                 info,
-                screenshot_requested: false,
+                actions_requested: HashSetExt::new(),
```
emilk, line 1011: I'm not sure what `HashSetExt` is, and I don't care to know

```suggestion
                actions_requested: Default::default(),
```

crates/eframe/src/native/wgpu_integration.rs
```diff
@@ -286,7 +287,7 @@ impl WgpuWinitApp {
                     maximized: Some(window.is_maximized()),
                     ..Default::default()
                 },
-                screenshot_requested: false,
+                actions_requested: HashSetExt::new(),
```
emilk, line 290:
```suggestion
                actions_requested: Default::default(),
```

crates/egui_glow/src/winit.rs
```diff
@@ -79,17 +82,17 @@ impl EguiGlow {
             log::warn!("Multiple viewports not yet supported by EguiGlow");
         }
         for (_, ViewportOutput { commands, .. }) in viewport_output {
-            let mut screenshot_requested = false;
+            let mut actions_requested: HashSet<egui_winit::ActionRequested> = HashSetExt::new();
```
emilk, line 85:
```suggestion
            let mut actions_requested: HashSet<egui_winit::ActionRequested> = Default::default();
```

emilk (approved):
Looks good, but needs a merge with `master`

bu5hm4nn:
@emilk Not sure if you get notified when checks pass.

emilk (approved):

PR #4041, opened 2024-02-14
Avoid interacting twice when not required

abey79:
This PR short-circuits `Response::interact()` when the `Response` has already been sufficiently "sensed" already. In some circumstance, this can avoid unnecessarily registering another widget rect that may mask some other widget.

One such instance is Rerun's `ListItem`. Calling `context_menu()` on its response would call `interact` and, in turn, mask its sub-widget (collapsing triangle, show/hide buttons, etc.).


emilk (approved):
Great!

PR #4043, opened 2024-02-14
Immediate Viewport in normal operation

### LINKED ISSUE #3972: When the program is minimized, the viewport stops working.
When a new Immediate Viewport is opened and the program is minimized,
the new Immediate Viewport remains displayed on the screen but becomes unresponsive.
When the program is restored from the minimized state, the Immediate Viewport starts functioning again.

This issue occurs after a recent update.
In previous versions, the new Immediate Viewport was not displayed on the screen when the program was minimized.

OS: Window 10
egui version: master branches

### LINKED ISSUE #4022: Immediate Viewport speed issue
Viewport speed issue

If you run `Immediate Viewport` while the program is at 60 frames (per second),
it will be at 50 frames.

If you have `two monitors` and the Viewport is on a different screen from the program,
the frame rate will be 30.

Is this a normal reaction?

egui version : ALL
OS : Windows 10

rustbasic:
Closes #3972  

in case of program is minimized, Immediate Viewport in normal operation.
Even if the program is minimized, `Immediate Viewport` continues to be displayed and operates normally.

Additional explanation :
In this case,
Create `Deffered Viewport` from `Immediate Viewport`, or
Creating an `Immediate Viewport` from an `Deffered Viewport`, then does not work as desired.
If you create other types of viewports within a viewport, it is a good idea to create them at the ROOT.

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -503,7 +503,7 @@ impl GlowWinitRunning {
             if is_immediate && viewport_id != ViewportId::ROOT {
                 // This will only happen if this is an immediate viewport.
                 // That means that the viewport cannot be rendered by itself and needs his parent to be rendered.
-                if let Some(parent_viewport) = glutin.viewports.get(&viewport.ids.parent) {
+                if let Some(parent_viewport) = glutin.viewports.get(&ViewportId::ROOT) {
                     viewport_id = parent_viewport.ids.this;
                 } else {
                     // Not actually used. Because there is always a `Some()` value.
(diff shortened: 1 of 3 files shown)


rustbasic:
@emilk

Based on `examples/multiple_viewports`, all fixed to work properly.
wgpu only tested 'examples/multiple_viewports', so additional testing on your part is needed.

rustbasic:
@emilk

If https://github.com/emilk/egui/pull/4043 is applied, all viewports will operate normally even if the program is minimized.

However, after applying https://github.com/emilk/egui/pull/4026 yesterday, everything stopped working.

PR #4045, opened 2024-02-14
providing a ui.label_trunc() and ui.label_wrap() and ui.label_rich()

### LINKED ISSUE #3900: Feature Request: providing a ui.label_truncate() function
@emilk
Dear emilk.

How about providing a `ui.label_truncate()` and `ui.label_wrap()` function?

egui/src/ui.rs
```
    #[inline]
    pub fn label_truncate(&mut self, text: impl Into<WidgetText>) -> Response {
        Label::new(text).truncate(true).ui(self)
    }

    #[inline]
    pub fn label_wrap(&mut self, text: impl Into<WidgetText>) -> Response {
        Label::new(text).wrap(true).ui(self)
    }

```

Thank you, emilk.

rustbasic:
Close #3900 

This is an example. 
Please modify and use it as you wish.

### DIFF
--- a/crates/egui/src/ui.rs
+++ b/crates/egui/src/ui.rs
@@ -1251,14 +1251,14 @@ impl Ui {
         Label::new(text).ui(self)
     }
 
-    /// This is like to Label::new(text).truncate(true).ui(self)
+    /// This is like to `Label::new(text).truncate(true).ui(self)`
     /// See also [`Label`].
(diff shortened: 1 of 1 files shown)


emilk:
No, I want to keep https://docs.rs/egui/latest/egui/struct.Ui.html short with only a few convenience helpers for the most used widgets

PR #4054, opened 2024-02-16
Some performance improvements.

### LINKED ISSUE #4022: Immediate Viewport speed issue

### LINKED ISSUE #4053: Unusable When Slowed Down
Unusable When Slowed Down

When the speed slows down, the FPS (frames per second) drops to 0-1, making it almost unusable.

How to slow down the speed :
Opening FontBook (egui_demo_app) after loading a font file with more than 40,000 fonts will slow down the program.

rustbasic:
Close #4022  

Some performance improvements.

### DIFF
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -105,7 +105,7 @@ fn run_and_return(
                 viewport_id,
             }) => {
                 let current_frame_nr = winit_app.frame_nr(*viewport_id);
-                if current_frame_nr == *frame_nr || current_frame_nr == *frame_nr + 1 {
+                if current_frame_nr >= *frame_nr {
                     log::trace!("UserEvent::RequestRepaint scheduling repaint at {when:?}");
                     if let Some(window_id) = winit_app.window_id_from_viewport_id(*viewport_id) {
                         EventResult::RepaintAt(window_id, *when)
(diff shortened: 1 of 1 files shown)


emilk:
Please elaborate on what is happening here

rustbasic:
@emilk 

It's done up to this point.

There are no special bugs.
If you look at the source code, there is no special change from the original intention.

This is what everyone originally intended, and it eliminates commonly used bottlenecks and improves some subtle performance.
It may be difficult to notice performance improvements in simple programs, but performance has improved.

rustbasic:
@emilk

Close #4022 

I think issue #4022 can be closed.
FPS is relatively improved at around 40-50.

Reducing FPS from 60 to 50 when creating a viewport is a separate issue.
I think it would be best to end here for now.

Not directly related to this, but influencing,
#4053 FontBook (bulk widget) issue still remains.

rustbasic:
@emilk 

If #4054 is applied, all viewports will operate normally even if the program is minimized.

However, after applying #4026 yesterday, everything stopped working.

PR #4056, opened 2024-02-16
Add `Response::paint_debug_info()` to make it easy to visualize a widget's id and state

abey79:
This PR adds a new `paint_debug_info()` debug method to `Response` in order to draw the response's `Rect`, it's id, and some of its state (active, hovered) via a color code:

Export-1708083028898


emilk (approved):
Great!

PR #4063, opened 2024-02-17
Add API for raw mouse motion

### LINKED ISSUE #1611: use DeviceEvent::MouseMotion to get mouse delta when mouse go out of screen.

GiantBlargg:
Raw mouse movement is unaccelerated and unclamped by screen boundaries, and does not relate to any position on the screen.
It is useful in certain situations such as draggable values and 3D cameras, where screen position does not matter.

https://github.com/emilk/egui/assets/1700581/1400e6a6-0573-41b9-99a1-a9cd305aa1a3

Added `Event::MouseMoved` for integrations to supply raw mouse movement.
Added `Response:drag_motion` to get the raw mouse movement, but will fall back to delta in case the integration does not supply it.

Nothing should be breaking, but third-party integrations that can send `Event::MouseMoved` should be updated to do so.

Based on #1614 but updated to the current version, and with better fallback behaviour.

* Closes #1611
* Supersedes #1614

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -446,6 +446,33 @@ impl WinitApp for GlowWinitApp {
                 }
             }
 
+            winit::event::Event::DeviceEvent {
+                device_id: _,
+                event: winit::event::DeviceEvent::MouseMotion { delta },
+            } => {
+                if let Some(running) = &mut self.running {
+                    let mut glutin = running.glutin.borrow_mut();
+                    if let Some(viewport) = glutin
+                        .focused_viewport
+                        .and_then(|viewport| glutin.viewports.get_mut(&viewport))
+                    {
+                        if let Some(egui_winit) = viewport.egui_winit.as_mut() {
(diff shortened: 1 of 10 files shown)


emilk (commented):
I like this approach! The only thing that worries me is the "unspecified units" part of this, especially for `DragValue`, who's `speed` is defined as "How much the value changes when dragged one point (logical pixel)", which after this PR is no longer true.

GiantBlargg:
I'm fine with reverting the changes to `DragValue`, my main goal was to add the plumbing to get the raw motion if the app wants it. For now I'm going to keep the changes to `Custom3D` as I feel that it's a good example of where someone might want to use raw motion.

enomado:
Thank you. But main idea was to fix UX of DragValue at corner(lacking of room for the mouse :) )

Issue #4066, opened 2024-02-18
Remove `id_source` parameter for ui.indent


FreePhoenix888:
**Is your feature request related to a problem? Please describe.**

We have to write id_source without a reason when creating an indent. Some of developers have added TODO to code to remove it (with question)
https://github.com/emilk/egui/blob/ca8c8792c2a6a65656105b971998f937edcb836d/crates/egui/src/ui.rs#L1775-L1780

**Describe the solution you'd like**

Remove it. Is there any reason to have it?

**Describe alternatives you've considered**

Do not remove it for compability reasons 🤔 

**Additional context**

emilk:
Yeah, it doesn't really belong there since the ident area doesn't store any state. Feel free to make a PR to remove it. Breaking changes are ok!

PR #4068, opened 2024-02-18
Remove `id_source` parameter from ui.indent

FreePhoenix888:
* Closes <https://github.com/emilk/egui/issues/4066>

I have read all and made all actions you have asked in pull request template but I am unable to run scripts/check.sh cause I am on windows 11 right now. Maybe I will do it later when I boot into ubuntu or someone else can do it later if required

All tests (except 9 ignored) are passing succesfully
image

### DIFF
--- a/crates/egui/src/containers/collapsing_header.rs
+++ b/crates/egui/src/containers/collapsing_header.rs
@@ -175,10 +175,10 @@ impl CollapsingState {
         let id = self.id;
         self.show_body_unindented(ui, |ui| {
             ui.indent(|ui| {
-                            // make as wide as the header:
-                            ui.expand_to_include_x(header_response.rect.right());
-                            add_body(ui)
-                        })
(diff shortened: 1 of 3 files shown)


crates/egui/src/containers/collapsing_header.rs
```diff
@@ -174,7 +174,7 @@ impl CollapsingState {
     ) -> Option<InnerResponse<R>> {
         let id = self.id;
         self.show_body_unindented(ui, |ui| {
-            ui.indent(id, |ui| {
+            ui.indent(|ui| {
```
emilk, line 177: I haven't looked to closely at this yet, but we need to make sure the id of `CollapsingHeader` is not ignored. This is so that two collapsing headers in the same ui won't have an id clash.

Perhaps we need a `indent_with_id_source` just for this use case, or add an extra `ui.push_id` call here

PR #4069, opened 2024-02-18
ImageLoader uses bg thread for processing

### LINKED ISSUE #5375: image loader for `file://` is blocking, while image loader for `http://` runs fine.
**Describe the bug**
When using 
```rust
ui.image("file://image.png");
```
it is blocking until the image is loaded.

But
```rust
ui.image("https://web.site.com/image.png");
```
will run fine and updates the GUI while the image is loading.

**Expected behavior**
All loaders should be non-blocking.

 - OS: macOS 14.6.1
 - Version: egui 0.29

jazzay:
- closes: https://github.com/emilk/egui/issues/5375

Modifies ImageLoader's load function to use background threads for the image decoding work. This avoids blocking the main thread that is especially noticeable when loading many images at once.

This was modelled after the other loader implementations that also use threads.


emilk (changes requested):
Thanks for the PR!

In some cases you want the blocking behavior though (to avoid a frame of showing no image) so it would be better if this was an optional behavior.

Also, this won't work on Wasm

jazzay:
fixed the cranky errors. Will look at making this optional and disabled for Wasm.

emilk:
Closing: background-thread image decoding landed in #5394. Thanks for the PR!

Issue #4033, opened 2024-02-12
Implement `Deref<Target = Context>` for `Ui`


obellish:
**Is your feature request related to a problem? Please describe.**
There are plenty of places where I would like to use a `Context` but have to get a reference to it from the Ui, when it's perfectly feasible to allow `Ui` to deref into a `Context`.

**Describe the solution you'd like**
Implement `Deref` for `Ui` to point to the painter context that the `ctx()` method points to.

**Describe alternatives you've considered**
Implementing an external trait to forward all method calls on `Ui` to `Context`.

PR #4072, opened 2024-02-19
add Deref<Target = Context> for Ui

obellish:
* Closes <https://github.com/emilk/egui/issues/4033>

### DIFF
--- a/crates/egui/src/ui.rs
+++ b/crates/egui/src/ui.rs
@@ -1,7 +1,7 @@
 #![warn(missing_docs)] // Let's keep `Ui` well-documented.
 #![allow(clippy::use_self)]
 
-use std::{any::Any, hash::Hash, sync::Arc};
+use std::{any::Any, hash::Hash, ops::Deref, sync::Arc};
 
 use epaint::mutex::RwLock;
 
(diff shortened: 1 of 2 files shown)


crates/egui/src/ui.rs
```diff
@@ -2320,6 +2320,14 @@ impl Ui {
     }
 }
 
+impl Deref for Ui {
+    type Target = Context;
+
+    fn deref(&self) -> &Self::Target {
```
emilk, line 2326:
```suggestion
    #[inline]
    fn deref(&self) -> &Self::Target {
```

emilk:
Does this mean we can remove `.input()` `.memory` etc from `Ui`?

obellish:
I believe so, as any methods on ctx will be on Ui as well, so duplicate methods can go away

emilk:
We shouldn't implement `DerefMut`, because a `&mut Context` in almost never needed. But we should remove the now redundant `Ui::input` etc

obellish:
Removed all the direct helper methods from `Ui`.

emilk (approved):
Great - thank you!

emilk:
There are some `cargo check` failures

obellish:
Removed the unused deps!

emilk:
a docstring needs updating too

obellish:
Which one?

emilk:
Check the CI output:

Screenshot 2024-03-22 at 12 27 45

emilk:
I tried fixing this myself, but since you opened the PR from your own `master` it failed.

TimTheBig:
Is there any plan to merge this? It would be nice

PR #4085, opened 2024-02-21
Add an API for customizing the return key in TextEdit

lemon-sh:
This PR allows customizing the return key in the TextEdit widget. Right now, it's hard-coded to the Enter key, which is problematic in cases when you want to use the Enter key for something else, and insert the newline in a different way instead.

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -108,7 +108,7 @@ impl<'t> TextEdit<'t> {
         }
     }
 
-    /// A [`TextEdit`] for multiple lines. Pressing enter key will create a new line by default (can be changed with [`return_key`]).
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -348,6 +350,16 @@ impl<'t> TextEdit<'t> {
         self.min_size = min_size;
         self
     }
+
+    /// Set the return key combination.
+    ///
+    /// This combination will cause a newline on multiline,
+    /// whereas on singleline it will cause the widget to lose focus.
+    #[inline]
+    pub fn return_key(mut self, return_key: KeyboardShortcut) -> Self {
```
emilk, line 359: I think this should be called `newline_key` instead

lemon-sh: that was also my first idea, but in the single line mode there is no newline involved at all, so it didn't feel right

emilk:
Thanks!

PR #4088, opened 2024-02-22
Fix panic when dragging window between monitors of different `pixels_per_point`

### LINKED ISSUE #4178: There is on very rare occasions a case where a panic occurred
There is on very rare occasions a case where a panic occurred at line 2,132 of context.rs when a Viewport exit.
( On screens with different pixels_per_point. )

### LINKED ISSUE #4179: There is on very rare occasions a case where a panic occurred at line 152 of layers.rs
There is on very rare occasions a case where a panic occurred at line 152 of layers.rs when a Viewport exit.
( At this time, self.0.len() == 0 )

( On screens with different pixels_per_point. )

thread 'main' panicked at C:\prog\egui\crates\egui\src\layers.rs:152:15:
index out of bounds: the len is 0 but the index is 0

rustbasic:
* Closes #4178 
* Closes #4179 

* Related #4211                                                                       

This is,

1. When moving between screens with different `pixels_per_point`, you can move `quickly` (smooth).
2. Prevents occasional panics (In fact, on very rare occasions a Viewport exit would cause a panic.).


crates/egui/src/layers.rs
```diff
@@ -149,6 +149,10 @@ impl PaintList {
     /// and then later setting it using `paint_list.set(idx, cr, frame);`.
     #[inline(always)]
     pub fn set(&mut self, idx: ShapeIdx, clip_rect: Rect, shape: Shape) {
+        if self.0.len() <= idx.0 {
+            self.add(clip_rect, Shape::Noop);
+        }
```
emilk, line 154: This shouldn't be needed unless the user does something wrong, and even then this only works if it is exactly off-by-one

crates/egui/src/context.rs
```diff
@@ -2123,12 +2123,15 @@ impl Context {
...
-            let texture_atlas = ctx
-                .fonts
-                .get(&pixels_per_point.into())
-                .expect("tessellate called with a different pixels_per_point than the font atlas was created with. \
-                         You should use egui::FullOutput::pixels_per_point when tessellating.")
-                .texture_atlas();
+            let texture_atlas = match ctx.fonts.get(&pixels_per_point.into()) {
+                Some(fonts) => fonts.texture_atlas(),
+                None => {
+                    ctx.fonts.iter().next()
```
emilk, line 2129: This will fall back to a texture atlas for the wrong pixels_per_point, which will likely look bad. I'd rather figure out the root cause of the bug than mask it :/

crates/egui/src/layers.rs
rustbasic: > This shouldn't be needed unless the user does something wrong, and even then this only works if it is exactly off-by-one

I experienced a panic occurring when self.0.len() was 0, and there is no problem if I use this. glow_integration and wgpu_integration must also be changed. I will add that part to this commit as well.

How to reproduce panic:
1. Run examples/multiple_windows.
2. Open all viewports.
3. Close “DD -> Top D -> SS -> Top S”. (order required)

crates/egui/src/context.rs
rustbasic: > This will fall back to a texture atlas for the wrong pixels_per_point, which will likely look bad. The existing check is there to help users that write egui integration to find bugs in their code.
> 
> I'd rather figure out the root cause of the bug than mask it :/

Don't worry, the texture atlas will be read for good pixel_per_point right away.
However, if you don't do this, a panic will occur before the texture atlas for a good pixel_per_point is read.
How to reproduce the panic is the same as the sequence shown above.

This also eliminates the root cause of the bug.
The root cause of another bug was fixed in #4128, so I didn't find it.

crates/egui/src/layers.rs
emilk: I fail to reproduce.

In any case: try to find the root of the problem instead of adding defensive solutions

emilk: And use `RUST_BACKTRACE=1` to get stack traces

emilk:
This is defensive programming that will hide the cause of the bugs instead of actually fixing them. Try to find the root of the problem instead.

PR #4092, opened 2024-02-23
Adjustable Slider rail height

rustbasic:
Adjustable Slider rail height 

explain20240312-2


crates/egui/src/style.rs
```diff
@@ -1224,6 +1227,7 @@ impl Default for Spacing {
             indent: 18.0, // match checkbox/radio-button with `button_padding.x + icon_width + icon_spacing`
             interact_size: vec2(40.0, 18.0),
             slider_width: 100.0,
+            slider_rail_height: 9.6,
```
emilk, line 1230: let's make it round and nice

```suggestion
            slider_rail_height: 8.0,
```

emilk (approved):

PR #4099, opened 2024-02-26
Make ecolor::HexColor Serialize to a Hex Code

Quinntyx:
This PR makes `ecolor::HexColor` Serialize/Deserialize to and from a Hex Code. 

- doesn't break any existing code
- new feature flag in the ecolor crate, without flag enabled will exhibit old serde default behavior
- properly serializes and deserializes the each variant of the enum to the respective correct hexcode length

Example of use (ron + serde), taken from my project [eguitty](https://github.com/quinntyx/eguitty): 

Without:
```rust
(
    profiles: {
        "default": (
            terminal_configuration: (
                bg_color: None,
                fg_color: None,
                cursor_trail: true,
                cursor_trail_color: None,
                font: (
                    size: 12.0,
                    family: Monospace,
                ),
                default_focus_cursor: Block(Hex8(((255, 255, 255, 255)))),
                default_unfocus_cursor: OpenBlock(Hex8(((255, 255, 255, 255)))),
                cursor_stroke: (
                    width: 1.0,
                    color: ((255, 255, 255, 255)),
                ),
            ),
            shell_command: "zsh",
        ),
    },
...
)
```

With:
```rust
(
    profiles: {
        "default": (
            terminal_configuration: (
                bg_color: None,
                fg_color: None,
                cursor_trail: true,
                cursor_trail_color: None,
                font: (
                    size: 12.0,
                    family: Monospace,
                ),
                default_focus_cursor: Block("#ffffffff"),
                default_unfocus_cursor: OpenBlock("#ffffffff"),
                cursor_stroke: (
                    width: 1.0,
                    color: ((255, 255, 255, 255)),
                ),
            ),
            shell_command: "zsh",
        ),
    },
...
)
```

### DIFF
--- a/crates/ecolor/src/hex_color_runtime.rs
+++ b/crates/ecolor/src/hex_color_runtime.rs
@@ -9,7 +9,10 @@ use crate::Color32;
 
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
-#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+#[cfg_attr(
+    all(feature = "serde", not(feature = "serde-hexcolor")),
+    derive(serde::Deserialize, serde::Serialize)
+)]
 /// A wrapper around Color32 that converts to and from a hex-color string
 ///
 /// Implements [`Display`] and [`FromStr`] to convert to and from the hex string.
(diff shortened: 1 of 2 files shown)


Quinntyx:
I actually thought about doing this directly to Color32, which is still also a possibility...
But I concluded HexColor would be better.

However, the downside of this is that many crates use Color32 as a field in, for example, their Style struct (looking at you, [egui_dock](https://github.com/Adanos020/egui_dock)), so they don't benefit at all from this change. 

As you can see, Stroke doesn't benefit either, and it's still quite difficult to tell what's going on in the default Serde behavior here.

Quinntyx:
Here's an example of the change with `serde_json`, the same segment of config

```json
{
  "profiles": {
    "default": {
      "terminal_configuration": {
        "bg_color": null,
        "fg_color": null,
        "cursor_trail": true,
        "cursor_trail_color": null,
        "font": {
          "size": 12.0,
          "family": "Monospace"
        },
        "default_focus_cursor": {
          "Block": {
            "Hex8": [
              255,
              255,
              255,
              255
            ]
          }
        },
        "default_unfocus_cursor": {
          "OpenBlock": {
            "Hex8": [
              255,
              255,
              255,
              255
            ]
          }
        },
        "cursor_stroke": {
          "width": 1.0,
          "color": [
            255,
            255,
            255,
            255
          ]
        }
      },
      "shell_command": "zsh"
    }
  },
  // ...
}
```

vs 
```json
{
  "profiles": {
    "default": {
      "terminal_configuration": {
        "bg_color": null,
        "fg_color": null,
        "cursor_trail": true,
        "cursor_trail_color": null,
        "font": {
          "size": 12.0,
          "family": "Monospace"
        },
        "default_focus_cursor": {
          "Block": "#ffffffff"
        },
        "default_unfocus_cursor": {
          "OpenBlock": "#ffffffff"
        },
        "cursor_stroke": {
          "width": 1.0,
          "color": [
            255,
            255,
            255,
            255
          ]
        }
      },
      "shell_command": "zsh"
    }
  },
  // ...
}
```

Quinntyx:
Okay i'm adding it to Color32 also
it's behind a feature flag so it should be fine

I'll just assume that Color32 is the same as a HexColor::Hex8 since they have the exact same data density

Quinntyx:
I tried to make serializing from `"#00000000"` equally as valid as `"#0000"` and `Color32::TRANSPARENT` (using the enum deserializer)

It should be possible, but I would have to use `deserialize_any`, which is incompatible with some serde backends, so I'm hesitant to take this approach...

For now, I've left myself a TODO, but if anybody knows how to do it without `deserialize_any` be my guest.

Quinntyx:
Should be noted that the feature flag could also be optionally removed if it were possible to do two types without `deserialize_any`, since the legacy format could be preferred for Serialization but the new format could be also valid for deserialization, which would allow non-breaking addition of this feature without a feature flag

emilk:
Color32 uses premultiplied alpha. which normal hex colors don't, so this could be quite confusing.

Quinntyx:
Hmm, that's true. 

I could just leave it as a change on HexColor.

crates/ecolor/src/hex_color_runtime.rs
```diff
@@ -27,6 +30,57 @@ pub enum HexColor {
...
+            ParseHexColorError::InvalidLength => {
+                serde::de::Error::invalid_length(v.len(), &"length 3, 4, 6 or 8 expected")
+            }
+            ParseHexColorError::InvalidInt(i) => serde::de::Error::invalid_type(
+                serde::de::Unexpected::Str(v),
+                &format!("invalid int: {i}").as_str(),
+            ),
+        })
+    }
+}
```
emilk, line 82: a test for this would be nice

PR #4107, opened 2024-02-27
`egui_extras`: Fixed handling of `file://` protocol for images

varphone:
* Remove the leading slash from the path if the target OS is Windows.

  This is because Windows paths are not supposed to start with a slash. For example, `file:///C:/path/to/file` is a valid URI, but `/C:/path/to/file` is not a valid path.

* Use the input URI consistently as the cache key.

  Currently, the cache key is inconsistently set as either the path or the URI, while the forget key is always the URI. This inconsistency should be resolved.

### DIFF
--- a/crates/egui_extras/src/loaders/file_loader.rs
+++ b/crates/egui_extras/src/loaders/file_loader.rs
@@ -25,6 +25,23 @@ impl FileLoader {
 
 const PROTOCOL: &str = "file://";
 
+/// Remove the leading slash from the path if the target OS is Windows.
+///
+/// This is because Windows paths are not supposed to start with a slash.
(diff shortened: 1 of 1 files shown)


crates/egui_extras/src/loaders/file_loader.rs
```diff
@@ -25,19 +25,33 @@ impl FileLoader {
 
 const PROTOCOL: &str = "file://";
 
+/// Remove the leading slash from the path if the target OS is Windows.
+///
+/// This is because Windows paths are not supposed to start with a slash.
+/// For example, `file:///C:/path/to/file` is a valid URI, but `/C:/path/to/file` is not a valid path.
+#[cfg(target_os = "windows")]
```
emilk, line 32:
```suggestion
```

PR #4120, opened 2024-03-02
 Added unhovered style, For TextEdit used bg_stroke in unhovered.

rustbasic:
Added unhovered style, 
For TextEdit used bg_stroke in unhovered.

### DIFF
--- a/crates/egui/src/style.rs
+++ b/crates/egui/src/style.rs
@@ -281,6 +281,9 @@ pub struct Spacing {
     /// Default width of a [`Slider`].
     pub slider_width: f32,
 
+    /// Default rail height of a [`Slider`].
+    pub slider_rail_height: f32,
+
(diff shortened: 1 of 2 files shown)


emilk:
explain why, preferably with an image

rustbasic:
> explain why, preferably with an image

It is good to have a border visible even when the input area is not in focus or unhovered.

explain20240312

crates/egui/src/style.rs
```diff
@@ -1051,6 +1054,9 @@ pub struct Widgets {
     /// The style of an interactive widget, such as a button, at rest.
     pub inactive: WidgetVisuals,
 
+    /// The style of a unhovered widget.
+    pub unhovered: WidgetVisuals,
```
emilk, line 1058: That is what `inactive` means.

emilk:
I think there is a misunderstanding here. `inactive` is already used for things that aren't hovered.

PR #4121, opened 2024-03-02
text cursor blink in TextEdit

rustbasic:
text cursor blink in TextEdit


emilk (changes requested):

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -76,6 +76,7 @@ pub struct TextEdit<'t> {
     desired_height_rows: usize,
     event_filter: EventFilter,
     cursor_at_end: bool,
+    blink: bool,
```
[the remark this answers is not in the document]
rustbasic: @emilk

I am not sure I fully understand your intention, so let me explain this.

TextEdit::blink stores the on/off setting for blinking. Even if there are multiple TextEdit instances in the program, the text cursor only exists in one that has focus, so I think the current configuration is appropriate.

paint_cursor() returns whether the current text cursor has been drawn or not. This is intended to call request_repaint(), but currently, request_repaint() does not work properly when the program does not have focus, so it is for the future and is a common configuration.

Therefore, the following code may or may not exist, and the current result will be the same:
```rust
    if is_drawn {
        ui.ctx().request_repaint();
    }
```

Please refer to this Pull Request and modify and apply it as emilk sees fit.

rustbasic:
@emilk

It's changes were completed.

crates/egui/src/widgets/text_edit/builder.rs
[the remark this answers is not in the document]
emilk: I'm questioning why someone would want a blinking cursor in one TextEdit and not in another. I'd imagine a user would either want a blinking text cursor in all their TextEdits or in none of them, i.e. that the setting to turn on/off a blinking text cursor would belong in `egui::Visuals` instead of as individual options to `TextEdit`

rustbasic:
@emilk

I discovered your post late.

It's changes were completed with ui.visuals()

emilk (changes requested):
Theres a bunch of unrelated changes in this PR, and also a lot of hard-coded constants (5, 8, 300, 700, …) that makes this hard to review

rustbasic:
> Theres a bunch of unrelated changes in this PR, and also a lot of hard-coded constants (5, 8, 300, 700, …) that makes this hard to review

It was mixed with other PRs. Other PR parts have been removed.

crates/egui/src/text_selection/visuals.rs
```diff
@@ -51,7 +51,21 @@ pub fn paint_text_selection(
...
+    cursor_rect: Rect,
+    i_time: f64,
+    blink: bool,
+) -> bool {
+    if blink {
+        // Make i_time_piece between 0 to 9.
+        let i_time_piece = ((i_time % 1.0) * 10.0).trunc() as i64;
+        if i_time_piece >= 5 && i_time_piece <= 8 {
+            return false;
+        }
```
emilk, line 66: why 5 and 8?

If it should be visible half of every second, then do
```
if i_time % 1.0 < 0.5 {
    return false;
}
```

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -680,7 +683,26 @@ impl<'t> TextEdit<'t> {
...
+                            is_blink,
+                        );
+                        if is_blink {
+                            if is_cursor_visible {
+                                ui.ctx()
+                                    .request_repaint_after(std::time::Duration::from_millis(700));
+                            }
+                            if !is_cursor_visible {
+                                ui.ctx()
+                                    .request_repaint_after(std::time::Duration::from_millis(300));
```
emilk, line 703: Make the cursor painter return the number of seconds until the next time the cursor should change from visible to invisible

emilk (changes requested):

crates/egui/src/widgets/text_edit/builder.rs
rustbasic: > Make the cursor painter return the number of seconds until the next time the cursor should change from visible to invisible

Would it be good to be able to set the time the cursor is on and off in `ctx.settings_ui()`?

rustbasic:
@emilk 

It's changes were completed 

I made it possible to set the on/off duration of the text cursor.

crates/egui/src/style.rs
```diff
@@ -728,6 +728,15 @@ pub struct Visuals {
...
     pub text_cursor_preview: bool,
 
+    /// set the text cursor to blink
+    pub text_cursor_blink: bool,
+
+    /// set the text cursor on duration
+    pub text_cursor_on_duration: f64,
+
+    /// set the text cursor off duration
+    pub text_cursor_off_duration: f64,
```
emilk, line 738: I think it's time to group all these in a `struct TextCursorStyle`

crates/egui/src/style.rs
```diff
@@ -1846,6 +1862,15 @@ impl Visuals {
             ui.checkbox(text_cursor_preview, "Preview text cursor on hover");
             ui.add(Slider::new(clip_rect_margin, 0.0..=20.0).text("clip_rect_margin"));
 
+            ui.add(Slider::new(resize_corner_size, 0.0..=20.0).text("resize_corner_size"));
```
emilk, line 1865:
```suggestion
```

crates/egui/src/text_selection/visuals.rs
```diff
@@ -73,3 +73,44 @@ pub fn paint_cursor(painter: &Painter, visuals: &Visuals, cursor_rect: Rect) {
...
+            ui.ctx()
+                .request_repaint_after(std::time::Duration::from_millis(
+                    (on_duration * 1000.0) as u64,
+                ));
+        }
+        if !is_cursor_visible {
+            ui.ctx()
+                .request_repaint_after(std::time::Duration::from_millis(
+                    (off_duration * 1000.0) as u64,
+                ));
```
emilk, line 113:
```suggestion
                .request_repaint_after(std::time::Duration::from_secs_f32(
                    on_duration
                ));
        }
        if !is_cursor_visible {
            ui.ctx()
                .request_repaint_after(std::time::Duration::from_secs_f32(
                    off_duration
                ));
```

crates/egui/src/text_selection/visuals.rs
```diff
@@ -73,3 +73,44 @@ pub fn paint_cursor(painter: &Painter, visuals: &Visuals, cursor_rect: Rect) {
         );
     }
 }
+
+/// Paint text cursor.
+pub fn paint_text_cursor(
```
emilk, line 78: `fn paint_text_cursor` and `fn paint_cursor` have very similar names, but one does blinking, and one does not. We should document that. `fn paint_cursor` should probably be made non-`pub` too

crates/egui/src/text_selection/visuals.rs
```diff
@@ -73,3 +73,44 @@ pub fn paint_cursor(painter: &Painter, visuals: &Visuals, cursor_rect: Rect) {
...
+    ui: &mut Ui,
+    painter: &Painter,
+    primary_cursor_rect: Rect,
+    is_stay_cursor: bool,
+) {
+    let i_time = ui.input(|i| i.time);
```
emilk, line 84: Currently the cursor still blinks while typing, which is quite annoying. If we instead reset a timer each time the user presses a key, it won't blink until `text_cursor_on_duration` after the user last stopped typing.

So, I think you were right to put this timer in `TextEditState`. We could put a `last_edit_time: Option<f64>` there fed by `ui.input(|i| i.time)`, and then `paint_text_cursor` is fed with `time_since_last_edit` which will be used as the basis for the animation

emilk (changes requested):
This is feeling much better now!

emilk:
I'll fix the rest of the things in this PR

rustbasic:
> I'll fix the rest of the things in this PR

Yes, Thank you emilk.

PR #4122, opened 2024-03-02
`epaint`: Add `EllipseShape`

TheTacBanana:
Adds an Ellipse shape draw-able with the Painter, brings egui closer to the SVG Specification.

I've done some optimization towards using less vertices and doing less calculations.

~~Currently the vertices are evenly distributed.
It's possible this could be optimized further taking into account the gradient and increasing the concentration of vertices where the change in gradient is larger.~~

EllipseTall
EllipseWide

### DIFF
--- a/crates/epaint/src/shape.rs
+++ b/crates/epaint/src/shape.rs
@@ -388,7 +388,7 @@ impl Shape {
             }
             Self::Ellipse(ellipse_shape) => {
                 ellipse_shape.center = transform * ellipse_shape.center;
-                ellipse_shape.size *= transform.scaling;
+                ellipse_shape.radius *= transform.scaling;
                 ellipse_shape.stroke.width *= transform.scaling;
             }
             Self::LineSegment { points, stroke } => {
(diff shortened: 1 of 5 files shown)


TheTacBanana:
Points are now distributed based on an easing function
Concentrating points around tighter bends, such as if the width if much greater than the height

crates/epaint/src/shape.rs
```diff
@@ -480,6 +499,61 @@ impl From<CircleShape> for Shape {
...
 // ----------------------------------------------------------------------------
 
+/// How to paint an ellipse.
+#[derive(Copy, Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub struct EllipseShape {
+    pub center: Pos2,
+
+    /// Size is the vector (a, b) where the width of the Ellipse is 2a and the height is 2b
+    pub size: Vec2,
```
emilk, line 509: Sounds like `radius` would be a better name for this

emilk:
I appreciate the PR, but I wonder how far we should be taking `epaint`. The purpose of `epaint` is to support the painting needs for a GUI, not arbitrary _anythings_, and certainly not to be on par with SVG. Still, ellipses isn't adding too much code or complexity, so it's probably fine.

But - what are you planning on using this for?

TheTacBanana:
Was using ellipses to draw nodes of graphs, as it looks much better than using either a large circle with label inside or a small circle with label outside
image

Thought that someone else might find it useful in the future

Have changed to `radius`, definitely makes more sense than `size`

crates/epaint/src/tessellator.rs
```diff
@@ -1315,6 +1318,74 @@ impl Tessellator {
...
+                .expand2(radius + Vec2::splat(stroke.width))
+                .contains(center)
+        {
+            return;
+        }
+
+        // Get the max pixel radius
+        let max_radius = radius.max_elem() * self.pixels_per_point;
+        // Get the power of two below the radius to limit the number of vertices
+        let floored_radius = 2_f32.powf(max_radius.log2().floor()) as u32;
```
emilk, line 1349: could use `next_power_of_two`

emilk: why round to power-of-two though?

TheTacBanana: Not sure why I chose rounding to a power of 2, have changed to just scale with the radius

emilk (approved):

PR #4127, opened 2024-03-03
Progress Circle Widget

bmbenson:
Implement a progress circle widget which can be used for various timer/stopwatch type applications.

Example of the progress moving from 1.0 to 0.0 with some updating text in the center.

ProgCircle

``` rust
// start time declared above.
let total_millis = 60 * 1000;
ui.label("60 second timer");
let now: DateTime<Utc> = Utc::now();
let elapsed = now - start_time.unwrap_or(now);

let remaining_millis = total_millis - elapsed.num_milliseconds();
let rem_percent = 1.0 - (elapsed.num_milliseconds() as f64 / total_millis as f64);
let stroke = Stroke::new(6.0, ui.visuals().selection.bg_fill);
let progress_circle = ProgressCircle::new(rem_percent as f32)
    .text(format!("{:.2}", (remaining_millis as f64  / 1000.0)))
    .clockwise(false)
    .segments(3600)
    .stroke(stroke);
ui.add(progress_circle);
```


DataTriny:
Nice work @bmbenson,

However your new widget is not accessible to assistive technologies and unfortunately there's a missing piece in the puzzle. See #4139. Once it's merged, you can have a look at what changes I made to the progress bar, as it's basically the same thing. I am not completely sure whether you'd want to expose a decreasing value if `clockwise` is set to `false` though. Having something going from 100% to 0% might be confusing to most people who aren't able to see that there's a circle involved.

bmbenson:
Thank you @DataTriny - 
The clockwise option controls the draw direction and is independent of the completion percent, so I believe the response.widget_info block will look very similar to your ProgressBar addition.

emilk:
I appreciate your PR, but I don't think this belong in `egui`. There are a hundred different ways a user may want to show progress, and the goal of egui is not to be a huge toolbox of every single thing a user could want. We already have a progress bar which should be enough for most people, and if they want something different they can just paint it themselves, or use a 3rd party crate.

bmbenson:
Makes sense - thank you for the response(and egui)!

PR #4138, opened 2024-03-05
Hide hover UI when showing the context menu

abey79:
This PR hides the hover UI for a given widget whenever a corresponding context menu is opened.

Fixes:
- https://github.com/rerun-io/rerun/issues/5310


Wumpf (approved):
seems straight forward! 👍

PR #4139, opened 2024-03-05
Add accessibility to `ProgressBar` and `Spinner`

DataTriny:
- Introduces `WidgetType::ProgressIndicator` and maps it to the corresponding AccessKit role.
- A `Spinner` is now exposed as a widget indicating a progress for which a completion state is not known.
- On the other hand, a `ProgressBar` reports a completion state and can possibly be labeled. Note that a label is not used if not explicitly asked by the user, as it would be redundant information. Assistive technologies prefer the numerical value so they can, for instance, emit beeps of which the frequency rise as the completion state increase. I had to call `floor` on the progression as it seems all the ATs I tested would round the value, hence reporting something different than what is displayed on the label.


emilk (approved):
Thanks!

PR #4140, opened 2024-03-06
Change local server IP addr to 127.0.0.1

3119369616:
Change IP addr to 127.0.0.1

### DIFF
--- a/scripts/start_server.sh
+++ b/scripts/start_server.sh
@@ -14,5 +14,5 @@ cargo install basic-http-server
 echo "starting server…"
 echo "serving at http://localhost:${PORT}"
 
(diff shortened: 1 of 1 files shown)


singersbalm:
you can't reach the server from other devices if its 127.0.0.1. for example: `192.168.178.xxx` or `devicename.local`

emilk:
Please explain why

3119369616:
> Please explain why
Because I can't run the local server if the ip addr is 0.0.0.0

OS of my computer is Win10.

singersbalm:
I have used 0.0.0.0 in the past on windows

emilk:
@3119369616 you should go to `127.0.0.1` in your browser, and the HTTP server bound on `0.0.0.0` will answer

PR #4143, opened 2024-03-07
Add optional line to Points PlotItem

jsalort:
This is a proposal to add the option to connect the points in the Points PlotItem, similar to "o-" in Matlab or Python Matplotlib.
I don't know if such addition would be welcome in egui_plot, but I needed this in a project, and initially thought about defining a custom PlotItem. But it is not straightforward to implement custom PlotItem, as some methods are private, so it was simpler to just add the options I needed in the existing Points struct. Besides, other people may want to do this.

It is not the same thing as adding two PlotItem: a Line and a Points, because then clicking on the name in the legend would mask only the Points. There are surely workarounds, but I found simpler to just have the option the connect the points in the Points PlotItem.


crates/egui_plot/src/items/mod.rs
```diff
@@ -972,99 +1031,103 @@ impl PlotItem for Points {
...
 
-        series
+        let values_tf: Vec<_> = series
             .points()
             .iter()
-            .map(|value| transform.position_from_point(value))
-            .for_each(|center| {
-                let tf = |dx: f32, dy: f32| -> Pos2 { center + radius * vec2(dx, dy) };
+            .map(|v| transform.position_from_point(v))
+            .collect();
```
emilk, line 1038: this is paying the price of the `collect` wether or not we have a line

emilk (commented):
This adds a "line" feature to `egui_plot::Points`.
We could instead add a `marker` feature to `egui_plot::Line`

But better yet would probably be to combine `Points` and `Line` into one thing, instead of replicating features between them. What do you think?

jsalort:
> This adds a "line" feature to `egui_plot::Points`. We could instead add a `marker` feature to `egui_plot::Line`
> 
> But better yet would probably be to combine `Points` and `Line` into one thing, instead of replicating features between them. What do you think?

Well yes, I agree completely. In the analogy with other plotting tools like Matplotlib, both kinds of plots are just obtained with `plt.plot`, with option for markers and/or line. Providing both features within the same PlotItem makes sense to me, but deprecating one of the existing things would be a backward-incompatible change (for people using the one that would be removed).
Or maybe it is possible to have just one thing, perhaps something like `egui_plot::Chart`, but keep both previous names, `egui_plot::Points` and `egui_plot::Line` which would be aliases for `Chart` only with different defaults, e.g. `Points` with only markers by default, and `Line` with only a line by default (if such a thing is possible with the Rust type system, I am too new to Rust to know).

emilk:

See also:

PR #4151, opened 2024-03-09
Added ability to disable highlighting of weekend days in DatePickerPopup.

hiyosilver:
I added a handfull of lines to allow for the red highlight on weekends in `DatePickerPopup` to be disabled.

I tried to follow the rules, but I've never done any kind of PR before, and I'm also not sure if this is at all the kind of thing that is wanted at the moment. If it is not, I'm happy to just have this removed. Just a tiny little addition I would find useful.


emilk (approved):

emilk:
requires rustfmt

hiyosilver:
I thought I had done it properly, but I somehow managed to slip an extra space in there.
I believe that was the only issue and it should work now.

Issue #2332, opened 2022-11-22
Modifier keys being held when a window loses and regains focus aren't detected by InputState::consume_key


Ununoctium117:
**Describe the bug**

In my desktop application, I respond to user commands entered via keystrokes with modifiers. Sometimes, this handling causes the egui window to temporarily lose, and quickly regain, focus. (In particular, I look for Ctrl+Delete and respond by executing a Windows IFileOperation shell command, which causes focus to shift to explorer.exe briefly.) When the window regains focus, if the user wants to perform the same action, they could reasonably still be holding Ctrl and expect that pressing delete would repeat the action. However, `InputState::consume_key(Modifiers::CTRL, Key::Delete)` returns false (in this state, `InputState::consume_key(Modifiers::NONE, Key::Delete)` returns true instead).

The user instead has to release Ctrl and re-press it before `InputState::consume_key(Modifiers::CTRL, Key::Delete)` will return true.

**To Reproduce**

1. With an egui desktop application in focus, press and don't release Ctrl.
2. Switch focus to a different app (perhaps by using the mouse), while still holding Ctrl.
3. Switch focus back to the egui application, while still holding Ctrl.
4. While still holding Ctrl, press some other keyboard input. The key combination isn't correctly detected by `InputState::consume_key`, which instead detects it as though Ctrl were not being held.

**Expected behavior**

Modifier key hold states aren't affected by application focus changes.

**Additional information**
This doesn't reproduce unless you are holding the modifier with the egui app in focus before switching away from it. If you begin holding the modifier key with some other app in focus, and then switch to the egui app, `InputState::consume_key` works as expected.

If this is hard to reproduce I can try to put together a minimal example that demonstrates this behavior.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version: 22H2

Ununoctium117:
I was hoping that #2933 would fix this, but no such luck; I can still repro on egui 0.22.

emilk:
Sounds like https://github.com/emilk/egui/pull/2933 does the opposite - it intentionally clears the modifier keys when eframe gains focus. This is because otherwise you have the problem of holding down a modifier, loosing focus, releasing the modifier, and then when switching back to the eframe app, eframe still things the modifier is being held. This is a classic "stuck modifiers" problem. Fixing this would likely require a deep dive into the `winit` internals on all platforms, and even then I wouldn't count on some key events not being lost.

PR #4157, opened 2024-03-10
Don't clear modifier state on focus change

ming08108:
I believe that the underlying issue that caused the stuck modifier keys was resolved in the 0.29 winit keyboard refactor.

This probably needs to tested on other desktop platforms however since I am only able to test this on windows 11.

* Closes <https://github.com/emilk/egui/issues/2332>


emilk:
Let's try it and see if we regress :)

Do you have a link for the winit-side fix for posterity?

Issue #4152, opened 2024-03-09
rounded corners don't scale properly in the new pan zoom demo


CrazyCraftix:
**Describe the bug**
Rounded corners inside the new pan zoom demo don't scale properly when zooming.

**To Reproduce**
In the demo code, increase the corner radius of the `egui::Frame` from 4.0 to 20.0 (it is barely noticeable otherwise).
Zoom in and out.

**Expected behavior**
When zooming, the corner radius should stay consistent in relation to the rest of the shape.

**Screenshots**
![wrong_corner_scaling](https://github.com/emilk/egui/assets/43807375/5f50a92e-13b2-4f14-bc8c-670dd58ed2d9)

**Additional context**
Multiplying the radius by `transform.scaling` in the demo code seems to fix the issue, however I'd expect this to be handled automatically.
I suppose this was simply overlooked in #3906.
I'd fix it myself but I've barely started using egui, so I'll leave it to the professionals :)

PR #4161, opened 2024-03-11
Fix scaling of rounded corners for rect shape (#4152)

CrazyCraftix:
When scaling an `egui::Shape` of variant `Rect` using the new `transform` function, corner rounding isn't taken into account.

The fix is to multiply the rounding by the scaling factor.

* Closes <https://github.com/emilk/egui/issues/4152>


CrazyCraftix:
I hope I didn't miss anything, I haven't really done this before. I now see I probably should have referenced this PR from my commits instead of the issue. Oh well...

emilk (approved):

PR #4160, opened 2024-03-11
Update `image` crate to 0.25

emilk:
To not produce duplicating deps in Rerun (https://github.com/rerun-io/rerun/pull/5280) I suggest we wait with merging this until these crates have updated to `image` 0.25:
* [x] [`arboard`](https://crates.io/crates/arboard)
* [x] [`gltf`](https://crates.io/crates/gltf)

PR #4164, opened 2024-03-12
Bump  rfd and env_logger

oscargus:
Not sure what the policy is, but would be nice to have the latest versions of these as one may expect most programs using egui wanting the latest versions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,13 +57,13 @@ checksum = "6c8c9b4467d77cacfbc93cee9aa8e7822f6d527c774efdca5f8b3a5280c34847"
 dependencies = [
  "accesskit",
  "accesskit_consumer",
- "async-channel",
+ "async-channel 1.9.0",
  "async-once-cell",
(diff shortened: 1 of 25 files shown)


oscargus:
Found #4160, so this should maybe wait until after that? If not, I can remove the `image` bump from this.

oscargus:
The newest version of rfd has dependencies that require rustc 1.75. Should I bump where required or remove the bump of rfd?

emilk:
There's no explicit policy, but I try to keep the direct dependencies of the published crates reasonably up-to-date.

But `rfd` and `env_logger` are only used by examples, so I feel less need to keep those up-to-date. Why not keep them up to date?
* Sometimes updating adds more 3rd party dependencies and compile time
* Sometimes updating requires a newer MSRV (as you've notices)

emilk:
Closing: main is now on rfd 0.17 and env_logger 0.11. Thanks for the PR!

PR #4169, opened 2024-03-13
Fix crash on `request_animation_frame` when destroying web runner

jprochazk:
Previously, any frames in flight (`requestAnimationFrame`) on web were not being cancelled (`cancelAnimationFrame`) when `WebRunner::destroy` was called. If a user called `destroy`, then immediately removed the canvas from the DOM, eframe could panic with a "failed to find (canvas) element by id" error message.

This PR changes two things:
- The canvas element is directly referenced everywhere it's needed instead of being looked up by `canvas_id`[^1]
- The RAF handle is stored in `WebRunner` and `cancelAnimationFrame` is called on it inside of `WebRunner::destroy`[^2]

[^1]: The WebGL/WGPU backends were already holding onto the canvas (and associated GPU context), so the change is just converting all the `get_element_by_id` lookups to retrieve the canvas from the web runner handle.

[^2]: There is only ever one frame in flight, so we store it directly as a scalar field.

### DIFF
--- a/crates/eframe/src/web/web_painter_glow.rs
+++ b/crates/eframe/src/web/web_painter_glow.rs
@@ -38,8 +38,8 @@ impl WebPainter for WebPainterGlow {
         self.painter.max_texture_side()
     }
 
-    fn canvas(&self) -> HtmlCanvasElement {
-        self.canvas.clone()
+    fn canvas(&self) -> &HtmlCanvasElement {
+        &self.canvas
     }
 
     fn paint_and_update_textures(
--- a/crates/eframe/src/web/web_painter_glow.rs
+++ b/crates/eframe/src/web/web_painter_glow.rs
@@ -10,7 +10,6 @@ use super::web_painter::WebPainter;
 
 pub(crate) struct WebPainterGlow {
     canvas: HtmlCanvasElement,
-    canvas_id: String,
     painter: egui_glow::Painter,
 }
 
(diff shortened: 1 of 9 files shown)


crates/eframe/src/web/web_painter_glow.rs
```diff
@@ -43,8 +38,8 @@ impl WebPainter for WebPainterGlow {
         self.painter.max_texture_side()
     }
 
-    fn canvas_id(&self) -> &str {
-        &self.canvas_id
+    fn canvas(&self) -> HtmlCanvasElement {
+        self.canvas.clone()
```
emilk, line 42: This could return a reference instead, then all `&runner.canvas()` everywhere could just be `runner.canvas()`

crates/eframe/src/web/web_runner.rs
```diff
@@ -24,6 +27,9 @@ pub struct WebRunner {
     /// They have to be in a separate `Rc` so that we don't need to pass them to
     /// the panic handler, since they aren't `Send`.
     events_to_unsubscribe: Rc<RefCell<Vec<EventToUnsubscribe>>>,
+
+    /// Used in `destroy` to cancel a pending frame.
+    raf_id: Cell<Option<i32>>,
```
emilk, line 32: let's not involve the royal air force in this

```suggestion
    request_animation_frame_id: Cell<Option<i32>>,
```

crates/eframe/src/web/web_runner.rs
```diff
@@ -179,23 +191,35 @@ impl WebRunner {
...
-struct IntervalHandle {
-    handle: i32,
-    closure: Closure<dyn FnMut()>,
+pub(super) struct IntervalHandle {
+    pub handle: i32,
+    pub closure: Closure<dyn FnMut()>,
 }
 
-enum EventToUnsubscribe {
+pub(super) enum EventToUnsubscribe {
```
emilk, line 222: do these really need to become public?

emilk (approved):
Much nicer ⭐

Issue #4162, opened 2024-03-12
Ability to pass data / control state of context menu


enomado:
Current API relied on response.rect. When we want to show menu, we associate it with response, e.g. menu call `context_interaction` to get menu events(create, close, stay) **internally**. All this stuff is private and couldnt be accessed outside egui. 

This behavior dont work if we want to track entity user clicked inside response.rect in our own. 

When I have `map, diagram, timeline` - **there is no way to know exact point was clicked on menu creation**. 

I have my own fork with workaround inside:
It is `context_menu` function

```
     let menu_response = egui::menu::MenuRoot::context_interaction(response, root, id);
```
added code is here. we just inject some `inter-frame-object`
```
    let my_state = match &menu_response {
        egui::menu::MenuResponse::Create(_, _) => {
            my_state.store()
        }
        _ => MyState::load()
    };
        
```
and pass `my_state` to the render function
```
bar_state.show(response, |ui| my_context_menu_render(ui, my_state));         
```

But it requires **a lot** of doing `pub(crate) -> pub` replacements. 

Then I forked `menu.rs`, but it couldnt work because `ui.set_menu_state` is private as well. 

1. So my naive API, just to check out the idea, may be something like:
```
response
.menu_clicked(|my_state, menu_response| 
    if menu_response == 'create'{ my_state = row_id}
    if menu_response == 'close'{ clean_resources() }
true
)
.show_menu(|ui, my_state| {...})
```
2. Much easier API would be
```
resp.context_menu(|ui, menu_handle|{
menu_handle. is_just_created() ?
menu_handle. some_handle_i_could_associate the data() ?
menu_handle. is_about_to_close() ?
})
```
Btw, looks like currently we have `ui.close_menu()` as this `menu_handle` germ.  

3. Probably, simplest and most useful API would to just make `ui.menu_state.response` public.

so I could do

resp.context_menu(|ui|{
    if ui.menu_state.response == 'create'{
     memory.save_my_data()
    }
})

(but currently show is earlier that setting the response)

4. Make context_menu and all related stuff  `public` to allow everybody inline it on his own function and add some logic. Current implementation is very easy to do this, except private visibility. 

personally I like variant №4

PS: `MenuResponse::Create(**pos**, id)` is not enough, because map could be scrolled after menu was created. we need ability to user make his own associated data.

PR #4170, opened 2024-03-14
Extend Response with `context_menu_custom` method

WojciechMula:
Method `context_menu_custom` allows to pass set of PointerButtons, that activate the context menu. This let to show menu on clicks by desired set of buttons/

Example:

```rust
use egui::PointerButtonSet;
use egui::PointerButtons;

let left_or_right = PointerButtonSet::from([
    PointerButtons::Primary,
    PointerButtons::Secondary,
]);

let response = ui.add(Label::new("Click me!").sense(Sense::click()));
response.context_menu_custom(left_or_right, |ui| {
    if ui.button("Close the menu").clicked() {
        ui.close_menu();
    }
});
```

### DIFF
--- a/crates/egui/src/data/input.rs
+++ b/crates/egui/src/data/input.rs
@@ -528,8 +528,78 @@ pub enum PointerButton {
     Extra2 = 4,
 }
 
+impl PointerButton {
+    /// Returns the list of all possible buttons.
+    pub fn all() -> [PointerButton; NUM_POINTER_BUTTONS] {
+        [
+            PointerButton::Primary,
+            PointerButton::Secondary,
+            PointerButton::Middle,
+            PointerButton::Extra1,
+            PointerButton::Extra2,
+        ]
+    }
+}
+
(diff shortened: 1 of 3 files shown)


enomado:
Hi there. I, as you, want more control to configure context menu apearing. I made some variants of possible API 

https://github.com/emilk/egui/issues/4162

Your feature could be made if there will be fully custom menu-starting function, which is not dependent on response

emilk:
What is the use case here? You want to activate a menu when both primary and secondary buttons are pressed?

If so, wouldn'ty `if response.primary_clicked() && response.secondary_clicked() { /* show menu */ }` be a better approach?

WojciechMula:
> What is the use case here? You want to activate a menu when both primary and secondary buttons are pressed?
> 
> If so, wouldn'ty `if response.primary_clicked() && response.secondary_clicked() { /* show menu */ }` be a better approach?

Yes, this is my goal, to show menu on some non-standard event. But I see my solution is too restrictive, what if somebody wanted to open menu on key.

Issue #4150, opened 2024-03-09
Expose winit WindowType (X11 feature) in ViewportBuilder


psethwick:
I want to make x11 windows with _NET_WM_WINDOW_TYPE other than the default. Specifically I want to make utitily/dialogs that automatically float in tiling window managers (i3 and the like)

winit already supports this, I think it just needs the bridge built to egui: https://github.com/rust-windowing/winit/blob/master/src/platform/x11.rs#L17

It looks like there are already platform specific options in ViewportBuilder (wayland, macos), so it seems appropriate enough.

I plan to open a PR for this

PR #4175, opened 2024-03-15
Add x11 window type settings to viewport builder

psethwick:
Not sure about the api, currently I've mapped the whole XWindowType enum, but maybe there's something more sensible to do?

* Closes <https://github.com/emilk/egui/issues/4150>

### DIFF
--- a/crates/egui/src/viewport.rs
+++ b/crates/egui/src/viewport.rs
@@ -819,52 +819,54 @@ pub enum WindowLevel {
 
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub enum X11WindowType {
     /// This is a normal, top-level window.
     #[default]
     Normal,
-    
+
     /// A desktop feature. This can include a single window containing desktop icons with the same dimensions as the
     /// screen, allowing the desktop environment to have full control of the desktop, without the need for proxying
     /// root window clicks.
     Desktop,
+
     /// A dock or panel feature. Typically a Window Manager would keep such windows on top of all other windows.
     Dock,
-    
+
     /// Toolbar windows. "Torn off" from the main application.
     Toolbar,
-    
+
     /// Pinnable menu windows. "Torn off" from the main application.
     Menu,
-    
+
     /// A small persistent utility window, such as a palette or toolbox.
     Utility,
-    
+
     /// The window is a splash screen displayed as an application is starting up.
     Splash,
-    
+
     /// This is a dialog window.
     Dialog,
-    
+
     /// A dropdown menu that usually appears when the user clicks on an item in a menu bar.
     /// This property is typically used on override-redirect windows.
     DropdownMenu,
-    
+
     /// A popup menu that usually appears when the user right clicks on an object.
     /// This property is typically used on override-redirect windows.
     PopupMenu,
-    
+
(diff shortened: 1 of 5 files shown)


crates/egui/src/viewport.rs
```diff
@@ -799,6 +817,48 @@ pub enum WindowLevel {
     AlwaysOnTop,
 }
 
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub enum WindowType {
```
emilk, line 822: Better to re-export `winit::platform::x11::XWindowType` instead, no?

psethwick: Yeah I can do that. Do you think I should change it to take multiple
XWindowType's too? I'm not sure I understand the point of that, but it is supported for some reason.

psethwick: done! It does unfortunately make egui depend on winit directly, thoughts? I guess the alternative is an extension trait in egui_winit?

emilk: Oh no, my bad 🤦 
I thought this code was in `egui-winit`
Yeah, we can't have `egui` depend on `winit`

emilk, line 822: let's make it clear this is a X11 thing

```suggestion
pub enum X11WindowType {
```

psethwick: totally agree and done!

crates/egui/src/viewport.rs
```diff
@@ -799,6 +817,60 @@ pub enum WindowLevel {
...
+    /// This property is typically used on override-redirect windows.
+    Combo,
+    
+    /// This indicates the the window is being dragged.
+    /// This property is typically used on override-redirect windows.
+    Dnd,
+    
+    /// This is a normal, top-level window.
+    #[default]
+    Normal,
```
emilk, line 871:
```suggestion
    /// This is a normal, top-level window.
    #[default]
    Normal,
    
    /// A desktop feature. This can include a single window containing desktop icons with the same dimensions as the
    /// screen, allowing the desktop environment to have full control of the desktop, without the need for proxying
    /// root window clicks.
    Desktop,
    /// A dock or panel feature. Typically a Window Manager would keep such windows on top of all other windows.
    Dock,
    
    /// Toolbar windows. "Torn off" from the main application.
    Toolbar,
    
    /// Pinnable menu windows. "Torn off" from the main application.
    Menu,
    
    /// A small persistent utility window, such as a palette or toolbox.
    Utility,
    
    /// The window is a splash screen displayed as an application is starting up.
    Splash,
    
    /// This is a dialog window.
    Dialog,
    
    /// A dropdown menu that usually appears when the user clicks on an item in a menu bar.
    /// This property is typically used on override-redirect windows.
    DropdownMenu,
    
    /// A popup menu that usually appears when the user right clicks on an object.
    /// This property is typically used on override-redirect windows.
    PopupMenu,
    
    /// A tooltip window. Usually used to show additional information when hovering over an object with the cursor.
    /// This property is typically used on override-redirect windows.
    Tooltip,
    
    /// The window is a notification.
    /// This property is typically used on override-redirect windows.
    Notification,
    
    /// This should be used on the windows that are popped up by combo boxes.
    /// This property is typically used on override-redirect windows.
    Combo,
    
    /// This indicates the the window is being dragged.
    /// This property is typically used on override-redirect windows.
    Dnd,
```

PR #4180, opened 2024-03-16
Option to change date picker format

zaaarf:
Simply allows date picker buttons to show other formats than %Y-%m-%d, while keeping that as default to not break compatibility.

I'm not that experienced with Rust, so I was unsure whether you'd prefer `&'a str` or a `String`, let me know if I should change that.

### DIFF
--- a/crates/egui_extras/src/datepicker/button.rs
+++ b/crates/egui_extras/src/datepicker/button.rs
@@ -16,6 +16,7 @@ pub struct DatePickerButton<'a> {
     calendar: bool,
     calendar_week: bool,
     show_icon: bool,
+    format: String,
 }
 
 impl<'a> DatePickerButton<'a> {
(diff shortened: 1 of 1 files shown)


crates/egui_extras/src/datepicker/button.rs
```diff
@@ -73,6 +75,14 @@ impl<'a> DatePickerButton<'a> {
         self.show_icon = show_icon;
         self
     }
+
+    /// Change the format shown on the button. (Default: %Y-%m-%d)
+    /// See [chrono::format::strftime] for valid formats.
+    #[inline]
+    pub fn format(mut self, format: &str) -> Self {
+        self.format = format.to_string();
```
emilk, line 83:
```suggestion
    pub fn format(mut self, format: impl Into<String>) -> Self {
        self.format = format.into();
```

emilk (approved):

zaaarf:
Okay, applied the changes. Why did the CI fail though?

zaaarf:
Seems to work now.

PR #4185, opened 2024-03-18
Added support for different layers for panels

john-sharratt:
This PR adds support for layers in the panels.

The main reason this is needed is when a background image is rendered on the background layer and then the panels are overlaid on top.

### DIFF
--- a/crates/egui/src/containers/panel.rs
+++ b/crates/egui/src/containers/panel.rs
@@ -93,6 +93,7 @@ impl Side {
 pub struct SidePanel {
     side: Side,
     id: Id,
+    layer_id: LayerId,
     frame: Option<Frame>,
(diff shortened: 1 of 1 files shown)


crates/egui/src/containers/panel.rs
```diff
@@ -124,6 +126,13 @@ impl SidePanel {
         }
     }
 
+    /// Redirect shapes to another paint layer.
```
emilk, line 129: It should be noted that this is only for the `show` functions, not `show_inside`

emilk (approved):

emilk:
bump

lucasmerlin:
Since `show` was removed in favor of `show_inside`, I guess we can close this.

PR #4190, opened 2024-03-19
Fix bug in `Context::parent_viewport_id`

### LINKED ISSUE #4084: Display parent viewport id Issues
When you run `examples\multiple_viewports` and open a viewport, the `parent viewport id` is displayed as `FFFF` in all viewports.

Show a problem with the `example`, but it seems like there's a problem with the `context`.

rustbasic:
* Closes #4084 

Fix : Display parent viewport id Issues


emilk (approved):
Good catch!

PR #4194, opened 2024-03-19
egui_extras: update resvg, add feature flag to support text

twitchyliquid64:
* Uses resvg 0.40 instead of 0.37, updating API use as appropriate. In particular, sizing hints are now implemented via the transform parameter, instead of mutating the `size` field on a flattened graph which I don't think we were meant to be doing
 * Adds optional feature flag `svg-text` to enable support for rastering text nodes in SVG trees.

image


crates/egui_extras/Cargo.toml
```diff
@@ -55,6 +55,9 @@ puffin = ["dep:puffin", "egui/puffin"]
 ## Support loading svg images.
 svg = ["resvg"]
 
+## Support rendering text in svg images. Adds a few hundred kb to the binary.
```
emilk, line 58: Does it work on web too?

emilk (approved):
Looks good if tested! Thanks

crates/egui_extras/Cargo.toml
twitchyliquid64: Looks like everything is fine except it cant find the fonts. Does egui bundle the raw font files it uses? if we could hand them to fontdb, that would work.

image

twitchyliquid64:
Fixed the Cranky CI failure, but could use some guidance on the cargo-deny ones if they matter.

crates/egui_extras/Cargo.toml
emilk: Yes, a few .ttf files are included with `include_bytes` in `crates/epaint/src/text/fonts.rs`

twitchyliquid64:
I sent a PR to resvg to bump their version of kurbo to match svgtypes, seemed like a no-op.

crates/egui_extras/Cargo.toml
twitchyliquid64: Any chance we could expose an API to make those bytes available?

emilk: They can be found in `ctx.fonts(|f| { f.lock().fonts.definitions().font_data }`

twitchyliquid64: Hmmm, that requires us having a valid `egui::Context` that has already been run however. Are you okay with me plumbing through context to this method (a public API change), or should we have another way to access the font data that doesn't need a `Context` ?

NidasioAlberto:
Hello! We are developing a sort of diagram editor and we'd love to have support for svg text. I see that now resvg is at version 0.40 and that the double dependency on `kurbo` is no more an issue. Is there anything left to have this feature merged?

emilk:
This PR adds a bunch of duplicated dependencies - if you want it merged it needs some more love (e.g. update `resvg` again so we don't duplicate `ttf-parser`)

emilk:
* Done in https://github.com/emilk/egui/pull/5979 instead

Issue #4196, opened 2024-03-20
On wayland `ViewportCommand::InnerSize` doesn't update the content size


ColinKinloch:
**Describe the bug**

On wayland (mutter) when the window is resized using `ViewportCommand::InnerSize` the viewport size doesn't match the window.

| Drag Resize Handle | Press "Phone Size" | Drag Resize Handle | Press "Desktop Size" |
| - | - | - | - |
| ![Screenshot from 2024-03-20 11-30-33](https://github.com/emilk/egui/assets/334272/19a6d819-4c9b-4f9d-b175-cf3cf8026a33) | ![Screenshot from 2024-03-20 11-28-20](https://github.com/emilk/egui/assets/334272/7ec5b50a-1dfa-4755-aa37-2f220a596c3c) | ![Screenshot from 2024-03-20 11-30-54](https://github.com/emilk/egui/assets/334272/ff5ad227-e962-4796-99e7-61c01279d5e8) | ![Screenshot from 2024-03-20 11-28-35](https://github.com/emilk/egui/assets/334272/01ebb808-df47-4dc5-887e-341732408ea5) |

I suspect this is because winit doesn't send `Resized` events after the resize, the correct way to handle this is applying the value returned from `request_inner_size` to the viewport here:

https://github.com/emilk/egui/blob/d449cb1d488cbc10915a3c720a232c01c1901561/crates/egui-winit/src/lib.rs#L1313-L1322

**To Reproduce**
1. On gnome shell press the Phone Size button, note the window contents are stretched
2. Resize the window, note the window contents are not stretched
3. Repeat

**Expected behavior**
Contents are not stretched

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Fedora 39

https://github.com/emilk/egui/pull/4197 makes it easier to test.

zeta0134:
I just ran into this bug today. Same behavior, this time on Arch+KDE. Only occurs in Wayland, the same device with an x11 session computes the dimensions correctly, even with DPI scaling applied.

ColinKinloch:
The behaviour is documented here:
https://docs.rs/winit/latest/winit/window/struct.Window.html#method.request_inner_size

PR #4197, opened 2024-03-20
egui demo: Add drop down to resize the window

ColinKinloch:
An alternative to the "Phone Size" button useful for testing `ViewportCommand::InnerSize`.

I created this to make it easy to debug https://github.com/emilk/egui/issues/4196 there are more details there.

### DIFF
--- a/crates/egui_demo_app/src/backend_panel.rs
+++ b/crates/egui_demo_app/src/backend_panel.rs
@@ -296,14 +296,28 @@ fn integration_ui(ui: &mut egui::Ui, _frame: &mut eframe::Frame) {
                 }
             }
 
-            if ui
-                .button("📱 Phone Size")
-                .on_hover_text("Resize the window to be small like a phone.")
-                .clicked()
-            {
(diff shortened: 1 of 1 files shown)


crates/egui_demo_app/src/backend_panel.rs
```diff
@@ -296,6 +296,20 @@ fn integration_ui(ui: &mut egui::Ui, _frame: &mut eframe::Frame) {
...
+                .clicked()
+            {
+                let size = egui::vec2(1280.0, 720.0);
+
+                ui.ctx()
+                    .send_viewport_cmd(egui::ViewportCommand::InnerSize(size));
+                ui.ctx()
+                    .send_viewport_cmd(egui::ViewportCommand::Fullscreen(false));
+                ui.close_menu();
+            }
```
emilk, line 311: I vote making the sizing options a drop-down menu so it doesn't use as much space, i.e. `Resize…` -> menu with "Phone sized" and "720p"

emilk (approved):

PR #4200, opened 2024-03-20
Expose state override for `HeaderResponse`

Zeenobit:
I'm trying to create some custom collapsing headers that add additional buttons inside the header itself. Because of this, I load the `CollapsingState` in my special widget manually. But because `HeaderResponse` owns the `Ui` and the `CollapsingState`, there is no way for me to open/close the collapsing header based on response of the inner widget.

Initially, I considered just exposing `state` member of `HeaderResponse`, but that exposes too much of the API at the wrong time, in my opinion.

So instead I found it'd be safer to just expose the open/close API to the response itself, and that's what this PR does.


enomado:
May be, If you upvote this, we could save few man-years https://github.com/emilk/egui/discussions/4181

emilk (approved):

PR #4182, opened 2024-03-17
Fix: Viewport Normal operation & Shaking when Resize

rustbasic:
This will continue to be updated :

Issues :
* Closes #2783 
* Closes #3811 
* Closes #3902 
* Closes #3953
* Closes #3972 
* Closes #4022
* Closes #4070 
* Related #4091
* Closes #4135 
* Related #4173 
* Closes #4178 
* Closes #4179
* Closes #4216 
* Closes #4222 

Pull Requests :
* Related #4040 
* Related #4043 
* Related #4054 
* Related #4071 
* Closes #4088
* Closes #4154
* Related #4214 
* Closes #4217
* Closes #4242 
* Closes #4328 
* Closes #4332 
* Closes #4348 
* Closes #4360
* Closes #4362 
* Closes #4374
* Closes #4388  
* Closes #4412 
* Related #4487 
* Related #4488 
* Related #4508
* Closes #4509

This will continue to be updated :

1. It may not be merged, so use it for reference only.
2. It is always in a usable state, maintaining a state with few bugs.
3. Anyone is free to maintain or improve a portion of this, and Pull Request it.

Based on `examples/test_viewports` and `examples/multiple_viewports`, all fixed to work properly.
`wgpu` only tested `examples/test_viewports`, so additional testing on your part is needed.

PR #4207, opened 2024-03-21
Fix `InnerSize` command not resizing viewport

### LINKED ISSUE #4196: On wayland `ViewportCommand::InnerSize` doesn't update the content size

ColinKinloch:
On some platforms changing window size via `request_inner_size` doesn't trigger a `Resized` event, in this case the viewport is not resized.

I'm new to this project and I think that this MR needs work.

What would the correct way to send the return value of `request_inner_size` to the resize function of the glutin surface?
https://github.com/emilk/egui/blob/769199648db30dadbec21cd1845bc325c5c91024/crates/egui-winit/src/lib.rs#L1314

Would setting `viewport.info.inner_rect` and resizing from that be correct?

* Closes #4196

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -1234,6 +1234,7 @@ impl GlutinWindowContext {
 
             if let Some(window) = &viewport.window {
                 let is_viewport_focused = self.focused_viewport == Some(viewport_id);
+                let inner_size = window.inner_size();
                 egui_winit::process_viewport_commands(
                     egui_ctx,
                     &mut viewport.info,
@@ -1242,6 +1243,11 @@ impl GlutinWindowContext {
                     is_viewport_focused,
                     &mut viewport.screenshot_requested,
                 );
+
+                let new_inner_size = window.inner_size();
+                if inner_size != new_inner_size {
+                    self.resize(viewport_id, new_inner_size);
+                }
             }
         }
 


rustbasic:
@ColinKinloch 

Check if this is not needed if you apply PR #4182.

ColinKinloch:
The issue is still present in https://github.com/emilk/egui/pull/4182

rustbasic:
> The issue is still present in #4182

egui-winit/src/lib.rs 800 line ~ 900 line

```rust
    let has_a_position = match window.is_minimized() {
        None | Some(true) => false,
        Some(false) => true,
    };
```

to

```rust
    let has_a_position = match window.is_minimized() {
        Some(true) => false,
        Some(false) | None => true,
    };
```


It may be because of this part.

yet again,

If you apply PR #4182, please make sure this is not necessary.

Please let me know the results.

rustbasic:
> The issue is still present in #4182

@ColinKinloch 

I changed this a little more.
Could you try this one more time and let me know the results?

* #4211

ColinKinloch:
Closed in favour of #4211

PR #4214, opened 2024-03-23
Add Continuous mode to ctx.memory.options

rustbasic:
Add Continuous mode to ctx.memory.options

### DIFF
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -278,12 +278,22 @@ impl EpiIntegration {
 
         let full_output = self.egui_ctx.run(raw_input, |egui_ctx| {
             if let Some(viewport_ui_cb) = viewport_ui_cb {
-                // Child viewport
+                // Child Deferred Viewport
                 crate::profile_scope!("viewport_callback");
                 viewport_ui_cb(egui_ctx);
             } else {
+                // ROOT Viewport ( with Immediate Viewport )
                 crate::profile_scope!("App::update");
                 app.update(egui_ctx, &mut self.frame);
+            }
+
+            let continuous_mode = egui_ctx.options(|options| options.continuous_mode);
+            if continuous_mode {
+                let viewport_id = egui_ctx.viewport_id();
+                // TODO : Do not recall until the next repaint.
+                // 1000 millis / 60 fps = 16.67 millis
+                self.egui_ctx
+                    .request_repaint_after_for(std::time::Duration::from_millis(8), viewport_id);
             }
         });
 
--- a/crates/egui/src/memory.rs
+++ b/crates/egui/src/memory.rs
@@ -191,6 +191,14 @@ pub struct Options {
     /// Controls the tessellator.
     pub tessellation_options: epaint::TessellationOptions,
 
+    /// If `true`, This will call `egui::Context::request_repaint()` at the end of each frame
+    /// If `false` (default), egui is only updated if are input events (like mouse movements) or there are some animations in the GUI.
+    ///
+    /// ```
+    ///     ui.ctx().options_mut(|options| { options.continuous_mode = true});
+    /// ```
+    pub continuous_mode: bool,
+
     /// If any widget moves or changes id, repaint everything.
     ///
     /// It is recommended you keep this OFF, because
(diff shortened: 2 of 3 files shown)


crates/eframe/src/native/epi_integration.rs
```diff
@@ -278,13 +278,23 @@ impl EpiIntegration {
...
             } else {
+                // ROOT Viewport ( with Immediate Viewport )
                 crate::profile_scope!("App::update");
                 app.update(egui_ctx, &mut self.frame);
             }
+
+            let continuous_mode = egui_ctx.options(|options| options.continuous_mode);
+            if continuous_mode {
+                let viewport_id = egui_ctx.viewport_id();
+                // TODO : Do not recall until the next repaint.
```
YgorSouza, line 293: The contributing guidelines say:

> Add your github handle to the TODO:s you write, e.g: `TODO(emilk): clean this up`

rustbasic: > The contributing guidelines say:
> 
> > Add your github handle to the TODO:s you write, e.g: `TODO(emilk): clean this up`

thank you.

rustbasic: > The contributing guidelines say:
> 
> > Add your github handle to the TODO:s you write, e.g: `TODO(emilk): clean this up`

If you know how to pass the doc test, please let me know.
It doesn't work with the "no_run" option.

crates/egui/src/memory.rs
```diff
@@ -191,6 +191,10 @@ pub struct Options {
     /// Controls the tessellator.
     pub tessellation_options: epaint::TessellationOptions,
 
+    /// If `true`, This will call `egui::Context::request_repaint()` at the end of each frame
+    /// If `false` (default), egui is only updated if are input events (like mouse movements) or there are some animations in the GUI.
+    pub continuous_mode: bool,
```
emilk, line 196: What is the advantage of this compared to just having the user call `request_repaint`?

crates/eframe/src/native/epi_integration.rs
```diff
@@ -278,13 +278,23 @@ impl EpiIntegration {
...
                 app.update(egui_ctx, &mut self.frame);
             }
+
+            let continuous_mode = egui_ctx.options(|options| options.continuous_mode);
+            if continuous_mode {
+                let viewport_id = egui_ctx.viewport_id();
+                // TODO(rustbasic) : Do not recall until the next repaint.
+                // 1000 millis / 60 fps = 16.67 millis
+                self.egui_ctx
+                    .request_repaint_after_for(std::time::Duration::from_millis(8), viewport_id);
```
emilk, line 296: why 8 ms delay?

emilk (changes requested):

crates/egui/src/memory.rs
rustbasic: > What is the advantage of this compared to just having the user call `request_repaint`?

1. Users do not need to worry about request_repaint().
2. This is much more efficient in terms of cput usage, etc.

crates/eframe/src/native/epi_integration.rs
rustbasic: > why 8 ms delay?

It takes about two attempts.
It is the most efficient.
If you do 16.67ms, 60 fps cannot be achieved.

emilk: But why not request an immediate repaint? eframe has vsync anyways

rustbasic: > But why not request an immediate repaint? eframe has vsync anyways



> But why not request an immediate repaint? eframe has vsync anyways

This is an unnecessary waste of CPU usage.
It would be a good idea to apply this and continue to monitor and upgrade.
It can't be perfect from the beginning.

emilk: It's not a waste of CPU, because of vsync.

crates/egui/src/memory.rs
emilk, line 196: Maybe we should just move `enum RunMode {` into `egui` from `crates/egui_demo_app/src/backend_panel.rs`. At least we should copy its documentation.

crates/egui_demo_app/src/backend_panel.rs
```diff
@@ -65,10 +65,11 @@ impl BackendPanel {
         match self.run_mode {
```
emilk, line 65: `self.run_mode` can be replaced with `options.continuous_mode`

PR #4217, opened 2024-03-23
Add `Context::parent_viewport_id_of`

### LINKED ISSUE #4216: ViewportInfo.parent doesn't always store any information.
`ViewportInfo.parent` doesn't always store any information.

and

It would be nice to have `ViewportInfo.this`.

rustbasic:
* Closes #4216 

fix : ViewportInfo.parent doesn't always store any information.

and

addition ViewportInfo.this


crates/egui/src/context.rs
```diff
@@ -599,6 +599,14 @@ impl ContextImpl {
             .unwrap_or(&ViewportId::ROOT)
     }
 
+    /// Return the `ViewportId` of his parent.
+    pub(crate) fn get_parent_viewport_id(&self, viewport_id: ViewportId) -> ViewportId {
```
emilk, line 603: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter


```suggestion
    pub(crate) fn parent_viewport_id_of(&self, viewport_id: ViewportId) -> ViewportId {
```

crates/egui/src/context.rs
```diff
@@ -3057,6 +3065,11 @@ impl Context {
         self.read(|ctx| ctx.parent_viewport_id())
     }
 
+    /// Return the `ViewportId` of his parent.
+    pub fn get_parent_viewport_id(&self, viewport_id: ViewportId) -> ViewportId {
```
emilk, line 3069:
```suggestion
    pub fn parent_viewport_id_of(&self, viewport_id: ViewportId) -> ViewportId {
```

crates/egui/src/data/input.rs
```diff
@@ -176,8 +176,11 @@ pub enum ViewportEvent {
 #[derive(Clone, Debug, Default, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub struct ViewportInfo {
+    /// this viewport, if known.
+    pub this: Option<ViewportId>,
```
emilk, line 180: This is unused?

I think everywhere we have a `ViewportInfo` we have a `ViewportId` for it as the key

emilk (changes requested):

crates/egui/src/data/input.rs
rustbasic: This is used.
Since parent_viewport_id() returns the last viewport_id, it is difficult to use unless it is inside a crate.

rustbasic: > This is unused?
> 
> I think everywhere we have a `ViewportInfo` we have a `ViewportId` for it as the key

no. There are places where viewport_id is missing. It's hard to explain.

crates/egui/src/context.rs
rustbasic: > https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter



> https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

This name does not pass inspection. Should we use a different name?

crates/egui/src/data/input.rs
rustbasic: > Try to explain and/or make this a `ids: ViewportIdPair` which is set in a constructor, instead of sometimes being `None`.

There are parts where this is absolutely necessary, and users will be able to easily access it. (In fact, this may not be easy for beginners to access.)
This is absolutely necessary, as there are cases where there is currently no way to determine the viewport_id.
Without this, every time you create a viewport, you would have to save the viewport_id somewhere and take care of it.

rustbasic: > Try to explain and/or make this a `ids: ViewportIdPair` which is set in a constructor, instead of sometimes being `None`.

If you don't save the viewport_id when creating a viewport, the current window sometimes doesn't know what the current viewport_id is. This is difficult to explain. Currently the viewport_id is unknown, so how can I find the parent_id? This is a different case from examples/demo_viewports.
last viewport_id may not be the current viewport.

rustbasic: > Then add `ids: ViewportIdPair` to `ViewportInfo` (and remove the `Default` bound on it), so users can _depend_ on the ids

`info.this` is absolutely necessary.
If `info.this` is present, `info.parent` does not need to be present.
`info.parent` pre-existed and could not work.

crates/egui/src/context.rs
rustbasic: > https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

what do you want?

parent_viewport_id_of() ? (you're writed here)
viewport_parent_id_of() ? (you're writed title)

crates/egui/src/data/input.rs
emilk, line 180:
```suggestion
    pub this: ViewportId,
```

crates/egui/src/data/input.rs
```diff
@@ -176,8 +176,11 @@ pub enum ViewportEvent {
 #[derive(Clone, Debug, Default, PartialEq)]
```
emilk, line 176: If we should store the `ViewportId` in this `struct`, we should force users to initialize it to the correct thing when they create it via a `ViewportInfo::new`, otherwise it will be wrong.

```suggestion
#[derive(Clone, Debug, PartialEq)]
```

emilk:
Closing this stale draft. If there is still a need for looking up the parent of an arbitrary viewport, please open a fresh PR against current main. Thanks for the PR, and sorry for the slow response

PR #4218, opened 2024-03-23
Don't show URLs when hovering hyperlinks

emilk:
Can be changed with `Style::url_in_tooltip`


crumblingstatue:
What's the rationale for this being the default?
I would expect to know where a hyperlink leads before I click it.
And common tools like web browsers also show the URL on hover.

PR #4242, opened 2024-03-27
For keep the border so we know where `TextEdit` exists.

rustbasic:
For keep the border so we know where `TextEdit` exists.


emilk:
Can you show an before-and-after image of the default egui style with this change?

rustbasic:
This is the image I posted before.

There is a lrectangle line in the input section at the end.


Before :

explain20240312-1

After :

explain20240312-2

emilk:
Seeing screenshots of your heavily customized app is not helpful.

This is what I'm interested in:

Before this PR:
Screenshot 2024-04-21 at 11 59 24


After this PR:
Screenshot 2024-04-21 at 11 58 58


Now I understand what the change is. I can see the benefits (better readability) and the downsides (its uglier)

rustbasic:
Seeing screenshots of your heavily customized app is not helpful.
=> Sorry about that. but, I don't have enough time to even edit images.

There are pros and cons, but I think the advantage of knowing that TextEdit exists there is greater.

Issue #4243, opened 2024-03-27
Provide a way to register native texture in `eframe`


Chaojimengnan:
**Is your feature request related to a problem? Please describe.**

I need to do some custom rendering with framebuffer and texture and then display it via `eframe`, but `eframe` doesn't provide a way to convert the native texture id to `epaint::TextureId`.

I've found `Painter::register_native_texture` but I can't access it through `eframe`.

**Describe the solution you'd like**

1. Pass the `&mut Painter` to the user to be able to register the native texture
2. Provide a function that allows the user to register a native texture.

**Describe alternatives you've considered**

None.

**Additional context**

None.

PR #4246, opened 2024-03-27
Add `register_native_texture` in `eframe::Frame`

Chaojimengnan:
* Closes https://github.com/emilk/egui/issues/4243

### DIFF
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -696,7 +696,7 @@ impl Frame {
     }
 
     /// Register your own [`glow::Texture`],
-    /// and then you can use the returned [`egui::TextureId`] to render your texture with `egui`.
+    /// and then you can use the returned [`egui::TextureId`] to render your texture with [`egui`].
     ///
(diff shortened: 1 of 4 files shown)


crates/eframe/src/epi.rs
```diff
@@ -690,6 +695,15 @@ impl Frame {
         self.gl.as_ref()
     }
 
+    /// Register your own [`glow::Texture`],
+    /// and then you can use the returned [`egui::TextureId`] to render your texture with [`egui`].
+    ///
+    /// This function will take the ownership of your [`glow::Texture`], so please do not delete your [`glow::Texture`] after registering.
+    #[cfg(all(feature = "glow", not(target_arch = "wasm32")))]
+    pub fn register_native_texture(&mut self, native: glow::Texture) -> egui::TextureId {
```
emilk, line 703: let's avoid name clashes in case we add a similar function for wgpu

```suggestion
    pub fn register_native_glow_texture(&mut self, native: glow::Texture) -> egui::TextureId {
```

emilk (approved):
Excellent!

PR #4260, opened 2024-03-28
Add web support for `zoom_factor`

justusdieckmann:
Before, when setting the `zoom_factor`, the website was already enlarged, but the zoom was ignored when calculating the logical window size and mouse position, causing an offset between the actual cursor and selected elements. That is addressed here

### DIFF
--- a/crates/eframe/src/web/app_runner.rs
+++ b/crates/eframe/src/web/app_runner.rs
@@ -183,7 +183,7 @@ impl AppRunner {
     /// The result can be painted later with a call to [`Self::run_and_paint`] or [`Self::paint`].
     pub fn logic(&mut self) {
         super::resize_canvas_to_screen_size(self.canvas(), self.web_options.max_size_points);
-        let canvas_size = super::canvas_size_in_points(self.canvas());
+        let canvas_size = super::canvas_size_in_points(self.canvas(), self.egui_ctx());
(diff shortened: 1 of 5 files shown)


justusdieckmann:
Thank you very much for the library! I did my first bigger project in rust and am really happy with it, especially the ease with which one can create both a native and web application. My prof had some difficulty reading the text, and I can see how one might have problems, especially with the moderate text contrast in dark mode. So I created this PR because I would like to enlarge the website a bit by default

emilk:
The intent of the current behavior is that the web browser is in charge of the zoom-factor/pixels-per-point.

If you launch a native egui app and press cmd-plus, you will zoom in, because egui catches that command and adjust the zoom factor.
If you go to egui.rs you'll notice cmd-plus also works to zoom in, but this time it's the browser that is in control. This means egui just follows the general zoom of the web page. In other words, if you embed an egui/eframe app into a web page and tells the browser to zoom in, egui will act just like any other web element.

Do you want to be able to zoom an egui app independently of the browsers setting?

justusdieckmann:
Exactly, I don't want to override the browsers zoom level, but I would like to be able to adjust the zoom level of the application programmatically

emilk (approved):

emilk:
Makes sense - thanks for the PR!

PR #4269, opened 2024-03-30
Fix continuous repaint on Wayland when TextEdit is focused or IME output is not None

### LINKED ISSUE #4254: On Wayland, app repaints every frame if IME output is set
**Describe the bug**
On Linux with Wayland, when an egui app using eframe has `ui.output(|o| o.ime)` not equal to `None`, the app continuously repaints.

One example of when `ui.output(|o| o.ime)` is not `None` is when an `egui::widgets::TextEdit` is focused because it is set by the widget in order to allow IMEs to determine where the cursor is.

I haven't tested this on non-Linux machines, but WebAssembly apps in browsers on Linux don't seem to have this problem, nor do apps for Windows running in Wine on Linux. Linux with X11 doesn't have this problem either.

**To Reproduce**
1. Clone egui and checkout to the 0.27.0 tag
2. `cargo run -p egui_demo_app`
3. Make sure "Backend" is selected in the top-left corner of the app so that the left pane is visible
4. Set the rendering mode to "Reactive" in the left pane so that the app doesn't repaint unless it has to
5. Focus the text edit in the widget gallery or any other text edit
6. Observe that the "Mean CPU usage" in the left pane constantly updates, meaning the app is repainting every frame

**Expected behavior**
The app should not repaint every frame when the text edit is selected.

 - OS: Arch Linux with GNOME 46
 - Version: Linux unknown 6.8.1-zen1-1-zen <span>#</span>1 ZEN SMP PREEMPT_DYNAMIC Sat, 16 Mar 2024 17:15:23 +0000 x86_64 GNU/Linux

white-axe:
* Closes #4254

Changes egui-winit so that it calls `window.set_ime_cursor_area` when the IME rect changes or the user interacts with the application instead of calling it every time the app is rendered. This works around a winit bug that causes the app to continuously repaint under certain circumstances on Wayland.

Tested on Wayland and on X11 using the text edit in the egui_demo_app - no changes in IME functionality as far as I can tell. Untested on non-Linux platforms.

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -99,7 +99,7 @@ pub struct State {
     accesskit: Option<accesskit_winit::Adapter>,
 
     allow_ime: bool,
-    ime_rect: Option<egui::Rect>,
+    ime_rect_px: Option<egui::Rect>,
 }
 
 impl State {
@@ -140,7 +140,7 @@ impl State {
             accesskit: None,
 
             allow_ime: false,
-            ime_rect: None,
+            ime_rect_px: None,
         };
 
         slf.egui_input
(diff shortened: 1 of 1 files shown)


crates/egui-winit/src/lib.rs
```diff
@@ -817,19 +819,24 @@ impl State {
...
-                winit::dpi::PhysicalPosition {
-                    x: pixels_per_point * rect.min.x,
-                    y: pixels_per_point * rect.min.y,
-                },
-                winit::dpi::PhysicalSize {
-                    width: pixels_per_point * rect.width(),
-                    height: pixels_per_point * rect.height(),
-                },
-            );
+            let rect = ime.rect * pixels_per_point;
```
emilk, line 823: a `Rect` is generally assumed to use `points` as the unit. It doesn't matter much here _rigth now_, but if someone were to try to use `self.ime_rect` for something else (e.g. debug-paint it) they would be in for weird bugs.

So I suggest postposning the multiplication with `pixels_per_point` until the conversion to `PhysicalPositon/PhysicalSize`, just like in the old code

emilk (approved):
Makes sense, but please wait with applying `pixels_per_point`

crates/egui-winit/src/lib.rs
white-axe: Alright, but I changed it to store the `pixels_per_point` as well in case the user changes their screen resolution in-between two times the app renders

emilk: Ah I see, that's what you wanted to capture… then I think the old code was better, if we call it `ime_rect_in_pixels` or similar. This tuple business is a bit over complex

emilk:
Thanks!

Issue #4206, opened 2024-03-21
TextEdit partially ignore clicks when they are not centred enough


abey79:
I noticed that text edit will react, but _not_ activate text cursor, when clicked near its margin. When the click is well centred, the text edit activates properly.

Repro on `master` as of now, with the stock `hello_world` example. Notice how the text cursor shortly appears and immediately disappear after each click, until the last click that occurs closer to the centre:

https://github.com/emilk/egui/assets/49431240/c2d4595b-81f3-48f6-bca9-24f4b4053978

YgorSouza:
That seems to be caused by this block:

https://github.com/emilk/egui/blob/769199648db30dadbec21cd1845bc325c5c91024/crates/egui/src/widgets/text_edit/builder.rs#L396-L398

If I remove the `&& !output.response.lost_focus()`, it seems to work correctly. But I don't know why this condition is there in the first place.

There is this other block that gets focus on pointer down instead of click.

https://github.com/emilk/egui/blob/769199648db30dadbec21cd1845bc325c5c91024/crates/egui/src/widgets/text_edit/builder.rs#L563-L565

So it seems that when you click near the edge, the response's `clicked_elsewhere` is true and it surrenders focus immediately, so the first block does not acquire focus.

rustbasic:
This appears to be intended behavior.
It appears to be so that other functions can be added.
Hold the click for about 0.3 seconds to get focus.

crumblingstatue:
> Is this intended behavior?

I hope it's not intended, because I thought there is something wrong with my system or egui when my text input kept getting unfocused.

rustbasic:
It is inconvenient because the recognition range is narrow.
I think it would be good to broaden the scope of awareness.

PR #4272, opened 2024-03-30
Fix bug in determining wether to remove focus from a widget

emilk:
* Closes https://github.com/emilk/egui/issues/4206

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -1113,8 +1113,6 @@ impl Context {
             changed: false,
         };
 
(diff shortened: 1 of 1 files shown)


crates/egui/src/context.rs
```diff
@@ -1113,8 +1113,6 @@ impl Context {
             changed: false,
         };
 
-        let clicked_elsewhere = res.clicked_elsewhere();
```
emilk, line 1116: The bug was that this was called too early.

Issue #4173, opened 2024-03-15
`eframe` on Windows high CPU usage due to excessive calls to `glutin::context::make_not_current`


gerwin3:
I noticed that my application CPU usage was much higher than expected on Windows. On a similar machine I was seeing 1%-2% usage on Linux, and around 20%-30% on Window while moving the mouse in window (just mouse movement).

After some optimization I got my rendering loop down to less than 1ms and still the CPU usage was too high. I then sampled the CPU with https://github.com/nico-abram/blondie/tree/main and this is the flamegraph that it produces:

As you can see, almost 90% of the time, it is calling the `make_not_current` function, which seems really excessive...

EDIT: I found in the docs [here](https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-wglmakecurrent) that this call flushes the rendering context so probably rendering is just very slow... Still not sure why though.

**Screenshots**

![image](https://github.com/emilk/egui/assets/1304177/0a5d9b27-90e5-437f-adcb-02f9b27364c0)

Another flamegraph with Windows Performance Analyzer (similar results):

![image](https://github.com/emilk/egui/assets/1304177/362c8a29-b484-4a84-8759-25959d72ae4e)

This shows the call stack a bit deeper:
* opengl32d.dll `wglMakeCurrent`
* opengl32d.dll `bMakeNoCurrent`
* opengl32d.dll `wglMakeCurrent`
* opengl32d.dll `wglMakeCurrent`
* opengl32d.dll `wglMakeCurrent`
* KernelBase `SetEvent`

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: egui 0.26

emilk:
I think this was introduced in https://github.com/emilk/egui/pull/3172 when we added support for multiple viewports. When there are multiple viewports we need to switch to which has the "current" gl context, and we unfortunately pay this cost even when there is only one viewport, but that should be fixable.

gerwin3:
Is there a workaround for now (i.e. options to disable multi-viewport)? (Apart from reverting to an older version)

emilk:
This is still an issue on Windows.

If someone wants to work on this, please take a look at
* https://github.com/emilk/egui/pull/4284
* https://github.com/emilk/egui/pull/4296

gerwin3:
[...]

PR #4284, opened 2024-03-30
eframe: Early-out from context switching the `glow` backend

emilk:
* Closes https://github.com/emilk/egui/issues/4173


rustbasic:
@emilk 

After the #4284 update, if you run `\examples\demo_viewports` and open all Viewports,
Viewport graphics are broken.

PR #4288, opened 2024-03-31
type : Pos2 <-> Vec2

rustbasic:
type : Pos2 <-> Vec2


crates/emath/src/pos2.rs
```diff
@@ -257,11 +257,11 @@ impl Add<Vec2> for Pos2 {
 }
 
 impl Sub for Pos2 {
-    type Output = Vec2;
+    type Output = Self;
```
YgorSouza, line 260: This doesn't make much sense. The difference between two points in space should be a vector, mathematically. It's the reason why these two exist as separate types in the first place.

rustbasic: > This doesn't make much sense. The difference between two points in space should be a vector, mathematically. It's the reason why these two exist as separate types in the first place.

Maybe yes. This was automatically requested during the last inspection after the Pull Request.

emilk:
`Pos2` is for positions (points), and `Vec2` is for vectors such as sizes. This is by design.

PR #4290, opened 2024-03-31
The contents of  #4284   have been reverted

### LINKED ISSUE #4289: After the #4284 update, Viewport graphics are broken.
@emilk

After the 
* #4284 
update, 

if you run \examples\demo_viewports and open all Viewports,
Viewport graphics are broken.

egui Version : Master

rustbasic:
The contents of 
* #4284  
have been reverted,

and processed as comments.

* Closes #4289


emilk:
Which commit in https://github.com/emilk/egui/pull/4284/commits caused a problem for you?

rustbasic:
> Which commit in https://github.com/emilk/egui/pull/4284/commits caused a problem for you?

Yes, #4284

rustbasic:
If we don't `return`, we can confirm that there is no problem.

Issue #2887, opened 2023-04-08
Replace shell scripts by xtask


YgorSouza:
[xtasks](https://github.com/matklad/cargo-xtask) are effectively like shell scripts written in Rust. As they only require rustc and cargo, they solve most portability problems (see for example #2859). They also allow using any Rust crates as part of the build automation, so for example they could pull the [wasm-opt](https://crates.io/crates/wasm-opt) crate as a dependency, instead of having this comment instructing the user to install it manually:

https://github.com/emilk/egui/blob/eb0812a9530f77712bf77aad0956f932a3a8f23d/sh/build_demo_web.sh#L81-L82

So it could make the repository easier to work with for new contributors.

PR #4293, opened 2024-03-31
Add xtask crate

YgorSouza:
Replaces only the cargo_deny.sh script for now. Can be expanded over time to replace the other shell and python scripts, so only Rust is needed to work with the repository.

Closes <https://github.com/emilk/egui/issues/2887>
Closes <https://github.com/emilk/egui/issues/4373>

### DIFF
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -4,5 +4,6 @@ edition.workspace = true
 license.workspace = true
 rust-version.workspace = true
 version.workspace = true
+publish = false
 
 [dependencies]
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "xtask"
+edition.workspace = true
+license.workspace = true
+rust-version.workspace = true
+version.workspace = true
+
+[dependencies]
--- a/xtask/src/deny.rs
+++ b/xtask/src/deny.rs
@@ -26,17 +26,18 @@ pub fn deny(args: &[&str]) -> Result<(), DynError> {
         "x86_64-unknown-redox",
     ];
     for target in targets {
-        let status = Command::new("cargo")
-            .args([
-                "deny",
-                "--all-features",
-                "--log-level",
-                "error",
-                "--target",
-                target,
-                "check",
-            ])
-            .status()?;
+        let mut cmd = Command::new("cargo");
+        cmd.args([
+            "deny",
+            "--all-features",
+            "--log-level",
+            "error",
+            "--target",
+            target,
+            "check",
+        ]);
+        super::utils::print_cmd(&cmd);
+        let status = cmd.status()?;
         if !status.success() {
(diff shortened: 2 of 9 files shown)


xtask/src/deny.rs
```diff
@@ -0,0 +1,55 @@
...
+            ])
+            .status()?;
+        if !status.success() {
+            return Err(status.to_string().into());
+        }
+    }
+    Ok(())
+}
+
+fn install_cargo_deny() -> Result<(), DynError> {
```
YgorSouza, line 47: Should we prompt the user to see if they are ok with installing packages? The shell scripts don't do that, but it seems like a good idea. There is an example in miri: <https://github.com/rust-lang/miri/blob/dba35d2be72f4b78343d1a0f0b4737306f310672/cargo-miri/src/util.rs#L181-L204>.

xtask/src/deny.rs
```diff
@@ -0,0 +1,59 @@
...
+//!
+//! Also installs the subcommand if it is not already installed.
+
+use std::process::Command;
+
+use super::DynError;
+
+pub fn deny(args: &[&str]) -> Result<(), DynError> {
+    if !args.is_empty() {
+        eprintln!("Warning: arguments ignored: {args:?}");
```
emilk, line 11: I think this should be a hard error

xtask/src/deny.rs
```diff
@@ -0,0 +1,59 @@
...
+                "--log-level",
+                "error",
+                "--target",
+                target,
+                "check",
+            ])
+            .status()?;
+        if !status.success() {
+            return Err(status.to_string().into());
+        }
```
emilk, line 42: We should print the command before the output like the previous script did (`set -x`) so that a user can re-run the failed command without having to re-run all checks

emilk (approved):
Nice, good idea to move away from shell scripts

xtask/src/deny.rs
```diff
@@ -0,0 +1,60 @@
...
+        cmd.args([
+            "deny",
+            "--all-features",
+            "--log-level",
+            "error",
+            "--target",
+            target,
+            "check",
+        ]);
+        super::utils::print_cmd(&cmd);
```
YgorSouza, line 39: Should I just debug-print the command instead? That would print it like this:

```sh
"cargo" "deny" "--all-features" "--log-level" "error" "--target" "aarch64-apple-darwin" "check"
```

Which isn't how you would normally type it, but it works.

emilk: I prefer the pretty printing that you added!

emilk (approved):

xtask/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "xtask"
+edition.workspace = true
+license.workspace = true
+rust-version.workspace = true
+version.workspace = true
```
emilk, line 6:
```suggestion
version.workspace = true
publish = false
```

PR #4294, opened 2024-03-31
WIP: get rid of clip_rect_margin

emilk:
`clip_rect_margin` is an ugly hack that I rather not have.

This PR sets it to zero.

The margins of a window is now added to the inside of any ScrollArea in that window.


emilk:
@varphone you may wanna test this to see I didn't break your custom window frames

PR #4296, opened 2024-04-01
Only avoid glow context switching on Windows

emilk:
…since it is reportedly broken on Windows

* Early-out added in https://github.com/emilk/egui/pull/4284
* Re-opens https://github.com/emilk/egui/issues/4173 😭 
* Closes https://github.com/emilk/egui/issues/4289
* Closes https://github.com/emilk/egui/pull/4290

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -849,10 +849,18 @@ fn change_gl_context(
 ) {
     crate::profile_function!();
 
-    if let Some(current_gl_context) = current_gl_context {
-        crate::profile_scope!("is_current");
-        if gl_surface.is_current(current_gl_context) {
-            return; // Early-out to save a lot of time.
+    if cfg!(target_os = "macos") {
+        // According to https://github.com/emilk/egui/issues/4289
+        // we cannot do this early-out on Windows,
+        // and we haven't tested Linux, so let's only do it on Mac for now.
+        // TODO(emilk): optimize this for other platforms too.
+        // See https://github.com/emilk/egui/issues/4173
+
+        if let Some(current_gl_context) = current_gl_context {
+            crate::profile_scope!("is_current");
+            if gl_surface.is_current(current_gl_context) {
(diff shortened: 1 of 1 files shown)


emilk:
@rustbasic please test this

YgorSouza:
FWIW I'm not seeing any problems with the `test_viewports` example on Linux, neither on this branch nor on master.

```
OS: Fedora Linux 39 (Workstation Edition) x86_64 
Host: Laptop (12th Gen Intel Core) A6 
Kernel: 6.7.10-200.fc39.x86_64 
Resolution: 2256x1504 
DE: GNOME 45.5 
WM: Mutter 
WM Theme: Adwaita 
Theme: Adwaita-dark [GTK2/3] 
Icons: Adwaita [GTK2/3] 
CPU: 12th Gen Intel i7-1260P (16) @ 4.700GHz 
GPU: Intel Alder Lake-P GT2 [Iris Xe Graphics] 
Memory: 7123MiB / 64008MiB 
```

rustbasic:
As a result of a short test,
It's Normal.

If any other abnormalities are found, we will notify you again.

rustbasic:
> @rustbasic please test this

As a result of a short test,
It's Normal.

If any other abnormalities are found, we will notify you again.

rustbasic:
@YgorSouza

YgorSouza, what do you want to do in `linux` ?

`return` or `not return` ?

emilk:
Thanks @YgorSouza - so sounds this is is a Windows-only problem then.

YgorSouza:
By the way, I just noticed another problem with this example on Linux, but I think it is unrelated to this. When I alt+tab into and out of the application a few times, eventually the viewports start hanging. But I tried the commits that reverts the early return for linux and the one that doesn't, and it seems that both have the same problem.

YgorSouza:
> When I alt+tab into and out of the application a few times, eventually the viewports start hanging

And this also happens with the `wgpu` feature, so definitely not related to this PR. That's the only problem I found on Linux so far.

Issue #4241, opened 2024-03-26
Blurry font rendering in firefox on linux


mash-graz:
Although some older similar issue reports exist, I want to document the `egui` Web rendering differences on firefox vs. chromium on my debian linux gnome wayland machine.

The differences are rather subtle and hard to grasp, but they make it very uncomfortable to work with text in `egui` in browser based applications.

I attach a view example screenshots:

![chromium](https://github.com/emilk/egui/assets/13013641/6db909ce-3277-4428-8922-736358d26a4e)
![chromium2](https://github.com/emilk/egui/assets/13013641/1abf1bf2-290c-4f14-b272-6b1d8942d7a8)
![firefox](https://github.com/emilk/egui/assets/13013641/684a096b-50a1-47b9-99d6-a1f15f779703)
![firefox2](https://github.com/emilk/egui/assets/13013641/add49c37-e766-4e6e-babd-5fc7feb6c7b9)
![side-by-side](https://github.com/emilk/egui/assets/13013641/2cc66e59-0df2-41f2-a9df-9e23be14fc4f)

lnicola:
Same for me, if not worse:

![image](https://github.com/emilk/egui/assets/308347/5bd65c11-defe-437e-81f0-251030e49da5)

Bad gamma, maybe?

CryZe:
@lnicola That's how it's supposed to look. It's just that the default text color is grey, not black and that the text hinting is not particularly great (if it even exists).

emilk:
The problem is that egui's pixel grid is not aligned with the pixel grid of the display:

Firefox:
<img width="739" alt="Screenshot 2024-03-27 at 15 31 18" src="https://github.com/emilk/egui/assets/1148717/22c80335-437d-43e4-96b3-794c5782a0a8">

In Chromium it is better, but it is still wrong vertically:
<img width="501" alt="Screenshot 2024-03-27 at 15 32 13" src="https://github.com/emilk/egui/assets/1148717/296b15b6-1034-4dda-8977-c0d942077167">

(these are blow-ups of your screenshots).

Is suspect we can fix it by tweaking how the canvas element is aligned in its parent somewhere in
https://github.com/emilk/egui/blob/a15e6c21220e593932569558a31de9403e3791d0/crates/eframe/src/web/mod.rs#L127-L130

emilk:
I can reproduce this on chromium on my Mac now, so I'm taking a look

emilk:
[...]

PR #4299, opened 2024-04-01
Fix blurry rendering in some browsers

emilk:
* Closes https://github.com/emilk/egui/issues/4241

I would love some more testers of this.

I'm not sure if we really need the round-to-even code, but I'm hesitant to out-right revert https://github.com/emilk/egui/pull/151 when I cannot reproduce its problem. Keeping it seems quite safe though.

---
# Testing
Checkout the branch and run:

* `./scripts/start_server.sh`
* `./scripts/build_demo_web.sh` and then open `http://localhost:8888/index.html#Rendering`
* `./scripts/build_demo_web.sh --wgpu` and then open `http://localhost:8888/index.html#Rendering`

Check the "Rendering test" that the squares in the pixel alignment test are perfectly sharp, like this:

Screenshot 2024-04-01 at 13 27 20

If it looks something like this, something is WRONG:
Screenshot 2024-04-01 at 13 29 07

Please try it on different zoom levels in different browsers, and if possible on different monitors with different native dpi scaling. Report back the results!

### Mac
I have tested on a high-DPI Mac:
* Chromium (Brave): ✅ Can reproduce problem on `master`, and it's now fixed
* Firefox:  ✅ Can reproduce problem on `master`, and it's now fixed
* Safari: ❌ Can't get it to work; giving up for now

### DIFF
--- a/crates/eframe/src/web/mod.rs
+++ b/crates/eframe/src/web/mod.rs
@@ -141,7 +141,7 @@ fn resize_canvas_to_screen_size(
         y: parent.client_height() as f32,
     };
 
-    if parent_size_points.x <= 0 || parent_size_points.y <= 0 {
+    if parent_size_points.x <= 0.0 || parent_size_points.y <= 0.0 {
         log::error!("The parent element of the egui canvas is {}x{}. Try adding `html, body {{ height: 100%; width: 100% }}` to your CSS!", parent_size_points.x, parent_size_points.y);
     }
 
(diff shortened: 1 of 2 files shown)


jleibs:
Tested both firefox and chrome on arch linux across multiple zoom levels. All looks good.

emilk:
Sounds like we're good then! 😃 
Thanks for testing

mash-graz:
Sorry for the late replay.
But your fix looks fine! (linux, firefox, gome wayland)

rendering

Thanks!

Issue #4295, opened 2024-03-31
Add some way to specify whether scroll_to_rect etc.. should be animated


lucasmerlin:
Egui should allow specifying whether scrolling in a ScrollArea should be animated (and also allow changing the animation duration, although that is not important for my problem)

**Is your feature request related to a problem? Please describe.**

After #4119 scrolling to a certain offset is always animated, this breaks egui_virtual_list, which relied on scroll_to_rect to instantly update the scroll position. This is necessary when the user scrolls up in a infinite scroll list that loads at the top and new content is added at the top (the scroll position is offset by the height of the new content). If the scroll is animated here, the list unexpectedly jumps for the user.

**Describe the solution you'd like**
Maybe add an additional parameter to the scroll functions, a bool or maybe a enum like 

```rust
enum ScrollAnimation {
    None,
    Default,
    Custom(Duration, Easing),
}
```

I also saw a [TODO comment](https://github.com/emilk/egui/blob/3ee4890b9433e8038de4cb498a3a9d09eb0d0c4b/crates/egui/src/containers/scroll_area.rs#L813) saying this should be made configurable via Style, but that would cause an issue for my crate, if I set the scroll animation duration to zero the user might override this at a later point causing the infinite scroll to break. The user could work around this by only updating the style if scroll_to_* is actually called but this wouldn't be ideal. 
Maybe there could be e.g. a scroll_to_rect() with the current signature which uses the value from Style and an additional funciton like scroll_to_rect_animated() that takes the enum from above as an additional argument. 

**Describe alternatives you've considered**
My problem could also be solved by #2783, which would allow me to use scroll_with_delta, which curiously doesn't seem to have an animation (I'm not sure if this is intentional). But due to #2783, scroll_with_delta doesn't work reliably for me.

rustbasic:
Please also refer to the following:

settings_ui -> Style -> Always scroll the only enabled direction

```rust
     ui.style_mut().always_scroll_the_only_direction = false;
```

PR #4303, opened 2024-04-01
Fix `Ui::scroll_with_delta` only scrolling if the `ScrollArea` is focused

### LINKED ISSUE #2783: `scroll_with_delta` doesn't work
**Describe the bug**

`scroll_with_delta` doesn't seem to work.

**To Reproduce**

1. Create an app with `eframe`.
2. Copy the `scroll_with_delta` example from documentation.
3. Click `Scroll down` button.
4. See that nothing happens.

```rust
use eframe::egui::{self, Vec2};

    eframe::run_native(
        "My egui App",
        Box::new(|_cc| Box::new(MyApp {})),
    )
}

struct MyApp {}

impl eframe::App for MyApp {
            let mut scroll_delta = Vec2::ZERO;
            if ui.button("Scroll down").clicked() {
                scroll_delta.y -= 64.0; // move content up
            }
                ui.scroll_with_delta(scroll_delta);
                for i in 0..1000 {
                    ui.label(format!("Item {}", i));
                }
            });
        });
    }
}
```

**Expected behavior**

Content scrolls down.

**Screenshots**

[video](https://user-images.githubusercontent.com/12128106/222928092-28c4f85e-29e7-4b11-9992-f4811a4c41ad.webm)

 - OS: GNU/Linux

lucasmerlin:
This introduces the boolean field force_current_scroll_area to InputState which will be set when scroll_with_delta is called, causing the ScrollArea to skip the check whether it is focused and always consume the smooth scroll delta.

* Closes #2783 
* Related to #4295

### DIFF
--- a/crates/egui/src/frame_state.rs
+++ b/crates/egui/src/frame_state.rs
@@ -41,8 +41,8 @@ pub(crate) struct FrameState {
     /// The current scroll area should scroll to this range (horizontal, vertical).
     pub(crate) scroll_target: [Option<(Rangef, Option<Align>)>; 2],
 
-    /// The current scroll area should scroll by this much (horizontal, vertical).
-    pub(crate) scroll_delta: [Option<f32>; 2],
+    /// The current scroll area should scroll by this much.
+    pub(crate) scroll_delta: Vec2,
 
     #[cfg(feature = "accesskit")]
     pub(crate) accesskit_state: Option<AccessKitFrameState>,
@@ -66,7 +66,7 @@ impl Default for FrameState {
             used_by_panels: Rect::NAN,
             tooltip_state: None,
             scroll_target: [None, None],
-            scroll_delta: [None, None],
+            scroll_delta: Vec2::default(),
             #[cfg(feature = "accesskit")]
             accesskit_state: None,
             highlight_this_frame: Default::default(),
@@ -104,7 +104,7 @@ impl FrameState {
         *used_by_panels = Rect::NOTHING;
         *tooltip_state = None;
         *scroll_target = [None, None];
-        *scroll_delta = [None, None];
+        *scroll_delta = Vec2::default();
 
         #[cfg(debug_assertions)]
         {
--- a/crates/egui/src/frame_state.rs
+++ b/crates/egui/src/frame_state.rs
@@ -38,8 +38,11 @@ pub(crate) struct FrameState {
     /// Initialized to `None` at the start of each frame.
     pub(crate) tooltip_state: Option<TooltipFrameState>,
 
-    /// horizontal, vertical
+    /// The current scroll area should scroll to this range (horizontal, vertical).
     pub(crate) scroll_target: [Option<(Rangef, Option<Align>)>; 2],
+
+    /// The current scroll area should scroll by this much (horizontal, vertical).
+    pub(crate) scroll_delta: [Option<f32>; 2],
 
     #[cfg(feature = "accesskit")]
     pub(crate) accesskit_state: Option<AccessKitFrameState>,
@@ -63,6 +66,7 @@ impl Default for FrameState {
             used_by_panels: Rect::NAN,
             tooltip_state: None,
             scroll_target: [None, None],
+            scroll_delta: [None, None],
             #[cfg(feature = "accesskit")]
             accesskit_state: None,
             highlight_this_frame: Default::default(),
@@ -84,6 +88,7 @@ impl FrameState {
             used_by_panels,
             tooltip_state,
             scroll_target,
+            scroll_delta,
             #[cfg(feature = "accesskit")]
             accesskit_state,
             highlight_this_frame,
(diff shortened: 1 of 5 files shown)


crates/egui/src/ui.rs
```diff
@@ -1091,8 +1091,10 @@ impl Ui {
     /// # });
     /// ```
     pub fn scroll_with_delta(&self, delta: Vec2) {
-        self.ctx()
-            .input_mut(|input| input.smooth_scroll_delta += delta);
+        self.ctx().input_mut(|input| {
+            input.smooth_scroll_delta += delta;
+            input.force_current_scroll_area = true;
```
emilk, line 1096: I think a better fix is to add a `scroll_delta` to `FrameState`, alongside `FrameState::scroll_target`

lucasmerlin: That is indeed much nicer 👍 I've updated my PR

crates/egui/src/frame_state.rs
```diff
@@ -38,9 +38,12 @@ pub(crate) struct FrameState {
     /// Initialized to `None` at the start of each frame.
     pub(crate) tooltip_state: Option<TooltipFrameState>,
 
-    /// horizontal, vertical
+    /// The current scroll area should scroll to this range (horizontal, vertical).
     pub(crate) scroll_target: [Option<(Rangef, Option<Align>)>; 2],
 
+    /// The current scroll area should scroll by this much (horizontal, vertical).
+    pub(crate) scroll_delta: [Option<f32>; 2],
```
emilk, line 45: This could just be a `Vec2`, no?

```suggestion
    pub(crate) scroll_delta: Vec2,
```

lucasmerlin: Updated. I think we could also update scroll_target to be `Option<[(..); 2]>` instead of `[Option<(..)>; 2]`. That would make the code in Prepared::end a bit cleaner. Should I include that change as well?

emilk: No - keep each PR as small as possible please

crates/egui/src/ui.rs
```diff
@@ -1091,8 +1091,9 @@ impl Ui {
     /// # });
     /// ```
     pub fn scroll_with_delta(&self, delta: Vec2) {
-        self.ctx()
-            .input_mut(|input| input.smooth_scroll_delta += delta);
+        self.ctx().frame_state_mut(|state| {
+            state.scroll_delta = delta;
```
emilk, line 1095: Are we sure we don't want `+=` here? In either case, we should document that behavior.

emilk (approved):

crates/egui/src/ui.rs
lucasmerlin: I changed it to +=, I think that makes more sense. Also documented the behavior.

lucasmerlin (commented):
I also noticed I accidentally inverted the behavior of scroll_with_delta, this is fixed now

crates/egui/src/ui.rs
```diff
@@ -1073,6 +1073,8 @@ impl Ui {
     /// A positive Y-value indicates the content is being moved down,
     /// as when swiping down on a touch-screen or track-pad with natural scrolling.
     ///
+    /// If this is called multiple times per frame for the same [`ScrollArea`], the deltas will be combined.
```
emilk, line 1076:
```suggestion
    /// If this is called multiple times per frame for the same [`ScrollArea`], the deltas will be summed.
```

crates/egui_demo_lib/src/demo/scrolling.rs
```diff
@@ -292,6 +293,9 @@ impl super::View for ScrollTo {
         ui.horizontal(|ui| {
             scroll_top |= ui.button("Scroll to top").clicked();
             scroll_bottom |= ui.button("Scroll to bottom").clicked();
+            if ui.button("Scroll down by 64px").clicked() {
+                scroll_delta = Some(Vec2::new(0.0, -64.0));
```
emilk, line 297:
```suggestion
                scroll_delta = Some(64.0 * Vec2::UP); // move contents up
```

crates/egui_demo_lib/src/demo/scrolling.rs
```diff
@@ -292,6 +293,9 @@ impl super::View for ScrollTo {
         ui.horizontal(|ui| {
             scroll_top |= ui.button("Scroll to top").clicked();
             scroll_bottom |= ui.button("Scroll to bottom").clicked();
+            if ui.button("Scroll down by 64px").clicked() {
```
emilk, line 296: It would be nicer demo with two buttons, using ⬆⬇ arrows instead of the text (egui supports those emojis)

emilk (approved):
nice!

crates/egui_demo_lib/src/demo/scrolling.rs
lucasmerlin: Done. Like this?

https://github.com/emilk/egui/assets/8009393/1b8663f5-b06d-4a13-a10a-9041fb4f011c

emilk (approved):

PR #4305, opened 2024-04-02
Make `scroll_to_*` animations configurable

### LINKED ISSUE #4295: Add some way to specify whether scroll_to_rect etc.. should be animated

lucasmerlin:
* Closes #4295

I based this on #4303, I'll rebase once that one gets merged.

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -809,20 +809,25 @@ impl Prepared {
                 .frame_state_mut(|state| state.scroll_target[d].take());
 
             if scroll_enabled[d] {
-                let update = if let Some(target) = scroll_target {
+                if let Some(target) = scroll_target {
+                    let frame_state::ScrollTarget {
+                        range,
+                        align,
+                        animation: animation_update,
+                    } = target;
                     let min = content_ui.min_rect().min[d];
                     let clip_rect = content_ui.clip_rect();
                     let visible_range = min..=min + clip_rect.size()[d];
-                    let (start, end) = (target.range.min, target.range.max);
+                    let (start, end) = (range.min, range.max);
                     let clip_start = clip_rect.min[d];
                     let clip_end = clip_rect.max[d];
                     let mut spacing = ui.spacing().item_spacing[d];
 
-                    let delta = if let Some(align) = target.align {
+                    let delta_update = if let Some(align) = align {
                         let center_factor = align.to_factor();
 
                         let offset =
-                            lerp(target.range, center_factor) - lerp(visible_range, center_factor);
+                            lerp(range, center_factor) - lerp(visible_range, center_factor);
 
                         // Depending on the alignment we need to add or subtract the spacing
                         spacing *= remap(center_factor, 0.0..=1.0, -1.0..=1.0);
@@ -836,15 +841,10 @@ impl Prepared {
                         // Ui is already in view, no need to adjust scroll.
                         0.0
                     };
-                    Some((delta, animation))
-                } else {
-                    None
-                };
 
-                if let Some((delta_update, animation_update)) = update {
                     delta += delta_update;
                     animation = animation_update;
-                }
+                };
 
                 if delta != 0.0 {
                     let target_offset = state.offset[d] + delta;
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -4,7 +4,7 @@ use crate::*;
 
 #[derive(Clone, Copy, Debug)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
-struct ScrollTarget {
+struct ScrollingToTarget {
     animation_time_span: (f64, f64),
     target_offset: f32,
 }
@@ -17,7 +17,7 @@ pub struct State {
     pub offset: Vec2,
 
     /// If set, quickly but smoothly scroll to this target offset.
-    offset_target: [Option<ScrollTarget>; 2],
+    offset_target: [Option<ScrollingToTarget>; 2],
 
     /// Were the scroll bars visible last frame?
     show_scroll: Vec2b,
@@ -860,7 +860,7 @@ impl Prepared {
                         let now = ui.input(|i| i.time);
                         let animation_duration = (delta.abs() / animation.points_per_second)
                             .clamp(animation.duration.min, animation.duration.max);
-                        state.offset_target[d] = Some(ScrollTarget {
+                        state.offset_target[d] = Some(ScrollingToTarget {
                             animation_time_span: (now, now + animation_duration as f64),
                             target_offset,
                         });
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -809,20 +809,20 @@ impl Prepared {
                 .frame_state_mut(|state| state.scroll_target[d].take());
 
             if scroll_enabled[d] {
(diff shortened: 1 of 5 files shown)


lucasmerlin:
@emilk Can you have a look at this one? I think it's a useful addition and it even removes one of your TODOs!

crates/egui/src/frame_state.rs
```diff
@@ -153,7 +153,7 @@ impl FrameState {
         *unused_rect = screen_rect;
         *used_by_panels = Rect::NOTHING;
         *scroll_target = [None, None];
-        *scroll_delta = Vec2::default();
+        *scroll_delta = (Vec2::default(), style::ScrollAnimation::none());
```
emilk, line 156:
```suggestion
        *scroll_delta = Default::default();
```

crates/egui/src/frame_state.rs
```diff
@@ -113,7 +113,7 @@ impl Default for FrameState {
             unused_rect: Rect::NAN,
             used_by_panels: Rect::NAN,
             scroll_target: [None, None],
-            scroll_delta: Vec2::default(),
+            scroll_delta: (Vec2::default(), style::ScrollAnimation::none()),
```
emilk, line 116:
```suggestion
            scroll_delta: Default::default(),
```

crates/egui/src/style.rs
```diff
@@ -692,6 +695,76 @@ impl ScrollStyle {
...
+        Self {
+            points_per_second: 0.0,
+            min_duration: 0.0,
+            max_duration: 0.0,
+        }
+    }
+
+    pub fn duration(t: f32) -> Self {
+        Self {
+            points_per_second: 0.0,
```
emilk, line 728: Using `0.0` here is quite surprising, as it leads to division by zero later. `f32::INFINITE` would make more sense in my head

crates/egui/src/style.rs
```diff
@@ -692,6 +695,76 @@ impl ScrollStyle {
...
+        Self {
+            points_per_second: 1000.0,
+            min_duration: 0.1,
+            max_duration: 0.3,
+        }
+    }
+}
+
+impl ScrollAnimation {
+    pub fn none() -> Self {
```
emilk, line 718: Docstring

crates/egui/src/style.rs
```diff
@@ -692,6 +695,76 @@ impl ScrollStyle {
...
+        Self {
+            points_per_second: 0.0,
+            min_duration: 0.0,
+            max_duration: 0.0,
+        }
+    }
+
+    pub fn duration(t: f32) -> Self {
```
emilk, line 726: Docstring

crates/egui/src/style.rs
```diff
@@ -692,6 +695,76 @@ impl ScrollStyle {
 
 // ----------------------------------------------------------------------------
 
+#[derive(Copy, Clone, Debug, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+#[cfg_attr(feature = "serde", serde(default))]
+pub struct ScrollAnimation {
```
emilk, line 701: docstring

crates/egui/src/frame_state.rs
lucasmerlin: Hmm, I explicitly set it to none since none is not the default. But I think it shouldn't matter since the animation isn't applied unless the Vec2 is not zero, so I'll update it.

crates/egui/src/style.rs
lucasmerlin: Ah, good catch!

crates/egui/src/containers/scroll_area.rs
```diff
@@ -834,11 +835,17 @@ impl Prepared {
                     } else {
                         // Ui is already in view, no need to adjust scroll.
                         0.0
-                    }
+                    };
+                    Some((delta, animation))
```
emilk, line 839: Shouldn't this be

```suggestion
                    Some((delta, target.animation))
```

?

crates/egui/src/containers/scroll_area.rs
```diff
@@ -808,20 +809,20 @@ impl Prepared {
                 .frame_state_mut(|state| state.scroll_target[d].take());
 
             if scroll_enabled[d] {
-                delta += if let Some((target_range, align)) = scroll_target {
+                let update = if let Some(target) = scroll_target {
                     let min = content_ui.min_rect().min[d];
```
emilk, line 813: by using destruction you won't forget any fields, even if more are added in the future

```suggestion
                let update = if let Some(target) = scroll_target {
                    let ScrollTarget { ragne, aling, animation } = target;
                    let min = content_ui.min_rect().min[d];
```

crates/egui/src/containers/scroll_area.rs
lucasmerlin: Ah yes, must have forgotten that when introducing the ScrollTarget struct

crates/egui/src/frame_state.rs
```diff
@@ -40,6 +40,23 @@ pub struct PerLayerState {
     pub widget_with_tooltip: Option<Id>,
 }
 
+#[derive(Clone, Debug)]
+pub struct ScrollTarget {
```
lucasmerlin, line 44: I just noticed that there are now two different ScrollTarget structs in egui, scroll_area::ScrollTarget and frame_state::ScrollTarget, both have somewhat different uses. Maybe this is a bit confusing? Should I maybe rename the scroll_area one to scroll_area::ScrollingToTarget (since it tracks the currently active scroll animation)? Do you have a better naming idea @emilk?

emilk: Yeah, let's avoid nameclashes pelase!

crates/egui/src/frame_state.rs
```diff
@@ -40,6 +40,23 @@ pub struct PerLayerState {
...
+    pub range: Rangef,
+    pub align: Option<Align>,
+    pub animation: style::ScrollAnimation,
```
emilk, line 47: Since this is a public struct it would be good to document these fields a bit

emilk (approved):
Thanks!

PR #4309, opened 2024-04-02
Allow disabling animations on a ScrollArea

lucasmerlin:
Adds a flag to ScrollArea that disables animations for the scroll_to_* functions


emilk:
Some CI failures, otherwise LGTM

PR #4310, opened 2024-04-02
Toggle Button created using Slider

rustbasic:
* Closes #4308

Toggle Button created using Slider.


rustbasic:
@DataTriny 

It seems like it was written without use.
Please try it and let me know what specific changes need to be made.

For example, you can change things with just a click rather than dragging. 
Change the `selected` value like a checkbox.

YgorSouza:
There is already a toggle switch implementation in the [demo app](https://github.com/emilk/egui/blob/2342788973120552c012f028bfda576fea87738b/crates/egui_demo_lib/src/demo/toggle_switch.rs). It is provided as an example of how users can write their own widgets. If egui were to have a toggle switch widget, it would probably be implemented like this. But currently anyone who needs it can just copy those 20 lines of code, and then customize it if they want.

rustbasic:
> There is already a toggle switch implementation in the [demo app](https://github.com/emilk/egui/blob/2342788973120552c012f028bfda576fea87738b/crates/egui_demo_lib/src/demo/toggle_switch.rs). It is provided as an example of how users can write their own widgets. If egui were to have a toggle switch widget, it would probably be implemented like this. But currently anyone who needs it can just copy those 20 lines of code, and then customize it if they want.

Yes, this is just to provide it in an easier and simpler way.

emilk:
As DataTrinity pointed out, this is a bad way of implementing a toggle switch.

At some point I'd like to have a global option to show checkboxes as toggle switches, i.e. like a styling option. But this PR is not it.

rustbasic:
> As DataTrinity pointed out, this is a bad way of implementing a toggle switch.
> 
> At some point I'd like to have a global option to show checkboxes as toggle switches, i.e. like a styling option. But this PR is not it.

Okay. I thought it would be better to have something simple to implement rather than nothing.

PR #4314, opened 2024-04-03
Fix wrong replacement function in deprecation notice of `drag_released*`

sornas:
Quick thing I noticed while updating a crate to egui 0.27.


emilk (approved):
Oopsies :)

PR #4320, opened 2024-04-04
report super/meta key for win/linux/web

psethwick:
Re: #3653

Tested on linux (x11).  I'm sort of trusting that winit/(other backends) handle the other platforms correctly.

this is a non-breaking change, I believe. 

Alternatively (breaking) We _could_ send meta on Mac, too? And then deprecate and eventually remove mac_cmd?

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -426,7 +426,7 @@ impl State {
                 self.egui_input.modifiers.ctrl = ctrl;
                 self.egui_input.modifiers.shift = shift;
                 self.egui_input.modifiers.mac_cmd = cfg!(target_os = "macos") && super_;
-                self.egui_input.modifiers.super_ = if cfg!(target_os = "macos") {
+                self.egui_input.modifiers.meta = if cfg!(target_os = "macos") {
                     false
                 } else {
                     super_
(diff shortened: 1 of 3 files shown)


crates/egui-winit/src/lib.rs
```diff
@@ -426,6 +426,11 @@ impl State {
                 self.egui_input.modifiers.ctrl = ctrl;
                 self.egui_input.modifiers.shift = shift;
                 self.egui_input.modifiers.mac_cmd = cfg!(target_os = "macos") && super_;
+                self.egui_input.modifiers.meta = if cfg!(target_os = "macos") {
+                    false
+                } else {
+                    super_
+                };
```
emilk, line 433: I think this is the path of least surprise, considering this is also how the meta key is defined on web.

```suggestion
                self.egui_input.modifiers.meta = super_;
```

crates/egui/src/data/input.rs
```diff
@@ -925,6 +959,7 @@ impl<'a> ModifierNames<'a> {
             append_if(modifiers.ctrl || modifiers.command, self.ctrl);
             append_if(modifiers.alt, self.alt);
             append_if(modifiers.shift, self.shift);
+            append_if(modifiers.meta, self.meta);
```
emilk, line 962: Does the windows key usually show up last in keyboard shortcuts (e.g. `Shift+Win+L`)?

emilk (changes requested):
Looks good, but I think there are more places `meta` needs to be checked, like `fn matches_logically`

crates/egui/src/data/input.rs
psethwick: I'll find out, I didn't fully internalize what ModifierNames actually does, oops

psethwick: seems like it's first (ref: https://support.microsoft.com/en-us/windows/keyboard-shortcuts-in-windows-dcc61a57-8ff0-cffe-9796-cb9706c75eec) will update

Issue #4324, opened 2024-04-05
Use parent elements ui for the style for popup


alexparlett:
**Is your feature request related to a problem? Please describe.**
Unable to style the popup from the parent (i.e. combox box) due to it using the root ctx and a new ui associated with that, meaning that any style changes made are reset back to default.

**Describe the solution you'd like**
Honour the style from the parent elements ui instead of using the root style.

**Describe alternatives you've considered**
You can manually implement a scope function storing the previous root style, changing it and then restoring if after you function has run.

**Additional context**
````
ui.style_mut().visuals.window_fill = egui::Color32::RED;

egui::ComboBox::from_id_source("window-mode-selector")
                                    .selected_text(format!("{}", self.display_settings.window_mode))
                                    .icon(filled_triangle)
                                    .width(ui.available_width())
                                    .show_ui(ui, |ui| {
                                        ui.selectable_value(
                                            &mut self.display_settings.window_mode,
                                            WindowMode::Fullscreen,
                                            format!("{}", WindowMode::Fullscreen),
                                        );
                                        ui.selectable_value(
                                            &mut self.display_settings.window_mode,
                                            WindowMode::Windowed,
                                            format!("{}", WindowMode::Windowed),
                                        );
                                    });
````

alexparlett:
@emilk  - any view on whether this is something that might be changed at source or should we build our own version of combo boxes and popups.

Just asking as currently we're using a custom fork and I'd rather not diverge from base egui as the dependency.

PR #4325, opened 2024-04-05
Use parent `Ui`s style for popups

alexparlett:
* Closes <https://github.com/emilk/egui/issues/4324>


emilk (approved):
Makes sense - thanks!

PR #4333, opened 2024-04-07
Handling of super keys on other platforms

rustbasic:
Handling of super keys on other platforms


crates/egui/src/data/input.rs
```diff
@@ -651,20 +651,48 @@ impl Modifiers {
...
+    pub fn super_only(&self) -> bool {
+        self.mac_cmd && !(self.alt || self.ctrl || self.shift)
+    }
+
+    /// true if only [`Self::mac_cmd`] is pressed on `MacOs`.
+    #[inline]
+    pub fn mac_cmd_only(&self) -> bool {
+        if !cfg!(target_os = "macos") {
+            return false;
+        }
```
emilk, line 683: Can't use the compile-time `target_os` in egui, since that won't work on wasm32. That's why we have `ctx.os()`

rustbasic: ctx.os() didn't seem to work here.
Let me check again.

emilk: Maybe we don't need this check at all. Let's just define `mac_cmd` to mean `super`?

rustbasic: Changed to

```rust
    pub fn mac_cmd_only(&self) -> bool {
        self.super_only()
    }
```

Issue #3389, opened 2023-09-25
Allow creation of large window that can exceed the monitor size


EvinceMoi:
**Is your feature request related to a problem? Please describe.**
When I set `initial_window_size` in `eframe::NativeOptions` with large value (bigger than the logical size of the monitor), the size is chop by [largest_monitor_point_size](https://github.com/emilk/egui/blob/master/crates/eframe/src/native/epi_integration.rs#L203C14-L217), this blocks some cases like displaying a workspace screenshot on multiple monitors, like #3185, or display high resolution pic/videos in actical size. I'm curious about the reason behind it.

**Describe the solution you'd like**
maybe add a option like `unbounded_size: bool` the disable the max window restriction?

**Describe alternatives you've considered**
N/A

**Additional context**
An simple example(all values are logical)
| monitor | x | y | width | height |mode|
|-|-|-|-|-|-|
|m1|0|0|1080|1920|portrait|
|m2|1080|840|1920|1080|landscape|

with this setup, a window across all screen should have the size of `3000x1920`. Setting `initial_window_size` to `3000x1920` , `initial_window_pos` to `0, 0`, a window at `0, 0` with size `1920x1080` wat created in my case.

lopo12123:
Hello @emilk , is there any progress on this issue? I would like to know what the motivation is for using largest_monitor_point_size, and whether it is possible to provide a bool value for users to choose whether to apply this option.

emilk:
The motivation can be found here:

https://github.com/emilk/egui/blob/2342788973120552c012f028bfda576fea87738b/crates/egui-winit/src/window_settings.rs#L97-L99

And yes, we can make that a `bool`, though I suggest we avoid negations in names and go with `clamp_size_to_monitor_size` or something like that, and default it to `true` only on Linux? 🤷 

PRs welcome!

lopo12123:
Thanks for the answer🙇‍

If the situation on linux is the only thing to consider, then I would try adding this field and then pr.

It may take me four or five days (I have three days off next and I have to go to the company to have multiple monitors for testing)

PR #4337, opened 2024-04-08
Allow users to create viewports larger than monitor on Windows & macOS

lopo12123:
Added clamp_size_to_monitor_size field on ViewportBuilder, which means whether clamp the window's size to monitor's size. (default to `true`)

* Closes https://github.com/emilk/egui/issues/3389

### simple example

```rust
pub struct MyApp {}

impl MyApp {
    pub fn new() -> MyApp {
        MyApp {}
    }
}

impl eframe::App for MyApp {
    fn update(&mut self, ctx: &Context, frame: &mut eframe::Frame) {
        egui::CentralPanel::default()
            .frame(Frame::none().fill(Color32::DARK_GRAY))
            .show(ctx, |ui| {
                if ctx.input(|i| i.key_pressed(Key::Escape)) {
                    ctx.send_viewport_cmd(ViewportCommand::Close);
                }
            });
    }
}

pub fn main() {
    let option = eframe::NativeOptions {
        viewport: ViewportBuilder::default()
            .with_position([10.0, 10.0])
            .with_inner_size([3000.0, 2000.0])
            .with_clamp_size_to_monitor_size(false),
        ..Default::default()
    };

    eframe::run_native(
        "a large window app",
        option,
        Box::new(|ctx| Box::new(MyApp::new())),
    ).unwrap();
}
```

It works on my windows (with 3 monitors), but I don't have a test environment for macos

### DIFF
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -30,7 +30,7 @@ pub fn viewport_builder<E>(
     let inner_size_points = if let Some(mut window_settings) = window_settings {
         // Restore pos/size from previous session
 
-        if with_constraint {
+        if clamp_size_to_monitor_size {
             window_settings.clamp_size_to_sane_values(largest_monitor_point_size(
                 egui_zoom_factor,
                 event_loop,
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -45,7 +45,7 @@ pub fn viewport_builder<E>(
             viewport_builder = viewport_builder.with_position(pos);
         }
 
-        if with_constraint {
+        if clamp_size_to_monitor_size {
             if let Some(initial_window_size) = viewport_builder.inner_size {
                 let initial_window_size = initial_window_size
                     .at_most(largest_monitor_point_size(egui_zoom_factor, event_loop));
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -22,7 +22,8 @@ pub fn viewport_builder<E>(
 
     let clamp_size_to_monitor_size = viewport_builder.clamp_size_to_monitor_size.unwrap_or(
         // On some Linux systems, a window size larger than the monitor causes crashes
-        cfg!(target_os = "linux"));
+        cfg!(target_os = "linux"),
+    );
 
     // Always use the default window size / position on iOS. Trying to restore the previous position
     // causes the window to be shown too small.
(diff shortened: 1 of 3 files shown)


crates/eframe/src/native/epi_integration.rs
```diff
@@ -20,14 +20,22 @@ pub fn viewport_builder<E>(
 
     let mut viewport_builder = native_options.viewport.clone();
 
+    // Allow users to create windows larger than monitor on windows & macos
+    let with_constraint: bool = !cfg!(any(target_os = "windows", target_os = "macos"))
+        || viewport_builder.clamp_size_to_monitor_size.unwrap_or(true);
```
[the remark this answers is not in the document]
lopo12123: In that case, the default behavior on non-linux platforms is inconsistent with previous versions. What I'm thinking about is to have the same behavior as before by default, which will reduce breaking changes.

As for linux, you mentioned that creating a window larger than the monitor will cause a crash, so I blocked creating large windows on linux here. Maybe this behavior should be left to the user to decide.

crates/egui/src/viewport.rs
```diff
@@ -275,6 +275,11 @@ pub struct ViewportBuilder {
     pub min_inner_size: Option<Vec2>,
     pub max_inner_size: Option<Vec2>,
 
+    /// Whether clamp the window's size to monitor's size. Default to `true`.
```
emilk, line 278: It only default to `true` on Linux

lopo12123: updated

crates/eframe/src/native/epi_integration.rs
```diff
@@ -20,14 +20,22 @@ pub fn viewport_builder<E>(
...
+
     // Always use the default window size / position on iOS. Trying to restore the previous position
     // causes the window to be shown too small.
     #[cfg(not(target_os = "ios"))]
     let inner_size_points = if let Some(mut window_settings) = window_settings {
         // Restore pos/size from previous session
 
-        window_settings
-            .clamp_size_to_sane_values(largest_monitor_point_size(egui_zoom_factor, event_loop));
+        if with_constraint {
```
emilk, line 33:
```suggestion
        if clamp_size_to_monitor_size {
```

crates/eframe/src/native/epi_integration.rs
```diff
@@ -37,10 +45,12 @@ pub fn viewport_builder<E>(
             viewport_builder = viewport_builder.with_position(pos);
         }
 
-        if let Some(initial_window_size) = viewport_builder.inner_size {
-            let initial_window_size = initial_window_size
-                .at_most(largest_monitor_point_size(egui_zoom_factor, event_loop));
-            viewport_builder = viewport_builder.with_inner_size(initial_window_size);
+        if with_constraint {
```
emilk, line 48:
```suggestion
        if clamp_size_to_monitor_size {
```

emilk (approved):

emilk:
need `cargo fmt`

PR #4342, opened 2024-04-09
Implement Color32 arithmetic

kikawet:
Implement improvements to combine and mix Color32:

- Implement std::ops::Add trait
- Implement std::ops::Sub trait
- Implement lerp function
- Implement merge function (interpolate at 50%)

### DIFF
--- a/crates/ecolor/src/color32.rs
+++ b/crates/ecolor/src/color32.rs
@@ -235,4 +235,97 @@ impl Color32 {
             a as f32 / 255.0,
         ]
     }
+
+    /// Interpolate two colors linearly, take as alpha the max value from both colors.
+    ///
+    ///```
+    /// use ecolor::Color32;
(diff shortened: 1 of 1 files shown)


kikawet:
To give some feedback about the algorithm choosen refer to this

https://stackoverflow.com/questions/726549/algorithm-for-additive-color-mixing-for-rgb-values

> It depends on what you want, and it can help to see what the results are of different methods.
> 
> If you want
> 
> Red + Black        = Red
> Red + Green        = Yellow
> Red + Green + Blue = White
> Red + White        = White 
> Black + White      = White
> 
> then adding with a clamp works (e.g. min(r1 + r2, 255)) This is more like the light model you've referred to.

emilk:
I'm sorry, but I don't think any of these operations makes any sense on colors with premultiplied alpha.

PR #4343, opened 2024-04-10
`egui`: Change `Ui::allocate_painter` to inherit properties from `Ui`

varphone:
The painter, allocated by `Ui::allocate_painter`, doesn't inherit properties from the `Ui` object, leading to improper widget rendering in `egui`.

Specifically, if the `Ui` object is disabled, the corresponding painter should also be disabled.


emilk (approved):
Good catch!

PR #4348, opened 2024-04-11
When the ScrollBar is set to Visibility::AlwaysVisible, scrolling is always possible & contains pointer

### LINKED ISSUE #2783: `scroll_with_delta` doesn't work

rustbasic:
Issues :
* Closes #2783
* Related #4350

Pull requests :
* Related #4303 
* Closes #4332 

When the ScrollBar is set to Visibility::AlwaysVisible, scrolling is always possible  & contains pointer

This way we can scroll wherever we want at any time.
(e.g. find string)


crates/egui/src/input_state.rs
```diff
@@ -234,7 +234,9 @@ impl InputState {
             // Mouse wheels often go very large steps.
             // A single notch on a logitech mouse wheel connected to a Macbook returns 14.0 raw_scroll_delta.
             // So we smooth it out over several frames for a nicer user experience when scrolling in egui.
-            unprocessed_scroll_delta += raw_scroll_delta;
+
+            // unprocessed_scroll_delta += raw_scroll_delta;
```
emilk, line 238: this looks like a bug

rustbasic: > this looks like a bug

There is no problem if I remove this from Windows.
I left it here to hear 'emilk's opinion on whether it can be removed.

emilk: This comment effectively ignores `raw_scroll_delta` and all the work that went into smoothing it, breaking the existing smooth scrolling of everything except for ScrollAreas (e.g. scroll-to-zoom in a plot).

rustbasic: > This comment effectively ignores `raw_scroll_delta` and all the work that went into smoothing it, breaking the existing smooth scrolling of everything except for ScrollAreas (e.g. scroll-to-zoom in a plot).

Edited as #4412
 
* #4412

Issue #4126, opened 2024-03-02
`SidePanel::right`s with an `exact_width`  have incorrect padding


sobaq:
**Describe the bug**

`SidePanel::right` incorrectly pads the right side of the panel when it has an `exact_width` set.

**To Reproduce**

The following `update` function roughly implements the screenshot below:

```rust
fn update(&mut self, ctx: &egui::Context, _: &mut eframe::Frame) {
    egui::SidePanel::right("right").exact_width(100.).show(ctx, |ui| {
        ui.label("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA");
    });
    egui::SidePanel::left("left").exact_width(100.).show(ctx, |ui| {
        ui.label("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA");
    });
    egui::CentralPanel::default().show(ctx, |ui| { });
}

```

**Screenshots**

<img width=500 src="https://github.com/emilk/egui/assets/109554668/82bdab2b-5c8c-49b6-8289-46cbf583ffaf">

**Desktop (please complete the following information):**
- OS: Linux.

**Additional Information**

As implied, the padding is correct when an `exact_width` is not given.

sobaq:
Closely related to #3006, I wouldn't call it a duplicate but I'd be happy to close this if need be.

RylanYancey:
I am also experiencing this problem! Here is what I'm seeing: 

![Screenshot from 2024-03-10 18-42-53](https://github.com/emilk/egui/assets/76256852/0edb7a83-851a-4d32-be68-3c454947fa1a)

The 'saved moves' panel is being rendered on top of 'Nxf6' and the multiline text widget. Heres' my src:

![Screenshot from 2024-03-10 18-44-39](https://github.com/emilk/egui/assets/76256852/95cc99be-167d-408e-a309-7781d9602f60)

(screenshot cuz my editor is being weird about copying)

I'm calculating what I want the width to be (40% of the screen space for the notes panel, and 10% for the moves panel), then assigning it by using `exact_width()`. However, this results in the aforementioned issue.

Thanks!
    - Rylan Yancey

Issue #3006, opened 2023-05-20
`SidePanel::exact_width` does not product a panel of that size


al2me6:
**Describe the bug**

Side panels are wider than the given `exact_width`.

**To Reproduce**

```rust
struct App;

impl eframe::App for App {
    fn update(&mut self, ctx: &eframe::egui::Context, _frame: &mut eframe::Frame) {
        eframe::egui::SidePanel::right("panel")
            .exact_width(150.)
            .show(ctx, |ui| {
                ui.horizontal_wrapped(|ui| {
                    ui.label("Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.");
                });
                ui.separator();
            });
    }
}

fn main() -> eframe::Result<()> {
    eframe::run_native(
        "ExactWidth",
        eframe::NativeOptions::default(),
        Box::new(|_cc| Box::new(App)),
    )
}
```

**Expected behavior**

The panel is exactly 150 px wide.

**Screenshots**

Observe that the panel is in fact 166 px wide:

![image](https://github.com/emilk/egui/assets/35266431/80b18296-c142-4b58-98de-6f53baedf3ae)

Thus, if the panel is on the right side, UI elements are clipped beyond the window:

![image](https://github.com/emilk/egui/assets/35266431/31c39216-d46f-46f2-b0b5-797679c32ff3)

**Desktop:**
 - OS: Windows 10
 - Browser: Firefox, Chromium-derived
 - Version `eframe 0.21.3`

**Additional context**
N/A

PR #4351, opened 2024-04-12
Fix `Panel` incorrect size

zhatuokun:
* Closes <https://github.com/emilk/egui/issues/4126>
* Closes <https://github.com/emilk/egui/issues/3006>

Previously, `SidePanel`/`TopBottomPanel`  didn't include `inner_margin` when setting the min width/height of `Frame`. As a result, when your `Panel` has `inner_margin`, its size doesn't work as expected.

Before

before

After

after

### DIFF
--- a/crates/egui/src/containers/panel.rs
+++ b/crates/egui/src/containers/panel.rs
@@ -262,7 +262,9 @@ impl SidePanel {
         let frame = frame.unwrap_or_else(|| Frame::side_top_panel(ui.style()));
         let inner_response = frame.show(&mut panel_ui, |ui| {
             ui.set_min_height(ui.max_rect().height()); // Make sure the frame fills the full height
-            ui.set_min_width(width_range.min);
+            ui.set_min_width(
+                width_range.min - (frame.inner_margin.left + frame.inner_margin.right),
+            );
             add_contents(ui)
         });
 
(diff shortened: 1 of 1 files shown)


emilk (commented):
Looks good, but please also update the docstrings of the `min_height, max_height, height_range etc` to explain they refer to the inner size, excluding the margins.

zhatuokun:
> Looks good, but please also update the docstrings of the `min_height, max_height, height_range etc` to explain they refer to the inner size, excluding the margins.

Thanks! Previously, if you called `SidePanel::min_width` to set the minimum width to `150.0`, the previous algorithm would result in the SidePanel's actual minimum width being `150.0 + inner_margin.left + inner_margin.right`. Under the default margins, this would add an extra `16.0` in size, for a final SidePanel actual minimum width of `166.0`. Therefore, I subtract their inner_margin, meaning that when you use `min_width` to set the Panel's minimum width to `150.0`, the Panel's actual size is `150.0`, and the content inside is `150.0 - (inner_margin.left + inner_margin.right)`, which is consistent with the documentation description (`Minimum width of the panel.`). The same applies to other properties.

emilk:
The problem with the existing documentation (and this is my mistake) is that "Minimum width of the panel" is not at all clear wether it is the outer size (size including margins) or the inner size (excluding the margins).

It seems the existing code was also confused, hence the bug. Bugs will keep cropping up unless we define explicitly wether or not the size includes the margins.

zhatuokun:
That's true, for this aspect, perhaps we can refer to the design of CSS `box-sizing`, but that's another issue. For now, due to the two issues I mentioned above,  I suggest merging the changes after I add the document. I will start working on it soon.

emilk (approved):

emilk:
Thanks!

PR #4353, opened 2024-04-12
Added ability to define colors at UV coordinates along a path

murl-digital:
I had to make a couple types not Copy because closures, but it should'nt be a massive deal.

I tried my best to make the API change as non breaking as possible. Anywhere a PathStroke is used, you can just use a normal Stroke instead. As mentioned above, the bezier paths couldn't be copy anymore,  but IMO that's a minor caveat.


crates/epaint/src/stroke.rs
```diff
@@ -52,3 +54,116 @@ impl std::hash::Hash for Stroke {
...
+pub enum ColorMode {
+    /// The entire path is one solid color, this is the default.
+    Solid(Color32),
+
+    /// Provide a callback which takes in a UV coordinate and converts it to a color. The UVs are relative to the bounding box of the path.
+    /// If you need something more custom, consider [`ColorMode::UVBounds`]
+    ///
+    /// **This cannot be serialized**
+    #[cfg_attr(feature = "serde", serde(skip))]
+    UV(Arc<Box<dyn Fn(Pos2) -> Color32 + Send + Sync>>),
```
emilk, line 70: This needs a bit more explanation. Is `Pos2(0, 0)` the top right of the bounding box and `Pos2(1,1,)` the bottom right of it?

crates/epaint/src/stroke.rs
```diff
@@ -52,3 +54,116 @@ impl std::hash::Hash for Stroke {
...
+
+    /// Provide a callback which takes in a UV coordinate and converts it to a color. The UVs are relative to the Rect provided.
```
emilk, line 72: Again, this need more explanation. Isa the given `Pos2` just a coordinate within the `Rect`? So that `pos == rect.left_top()` for a vertex that starts in the left-top corner of the bounding box?

crates/epaint/src/tessellator.rs
```diff
@@ -864,19 +875,58 @@ fn stroke_path(
...
+        .map(|point| OrderedFloat::from(point.pos.y))
+        .min()
+        .map(OrderedFloat::into_inner)
+        .unwrap();
+    let max_y = path
+        .iter()
+        .map(|point| OrderedFloat::from(point.pos.y))
+        .max()
+        .map(OrderedFloat::into_inner)
+        .unwrap();
```
emilk, line 912:
```suggestion
    let bbox = Rect::from_poins(path);
```

crates/epaint/src/tessellator.rs
```diff
@@ -864,19 +875,58 @@ fn stroke_path(
...
+        ColorMode::UV(fun) => {
+            let x = remap_clamp(pos.x, min_x..=max_x, 0.0..=1.0);
+            let y = remap_clamp(pos.y, min_y..=max_y, 0.0..=1.0);
+            fun(pos2(x, y))
+        }
+        ColorMode::UVBounds(rect, fun) => {
+            let x = remap_clamp(pos.x, rect.min.x..=rect.max.x, 0.0..=1.0);
+            let y = remap_clamp(pos.y, rect.min.y..=rect.max.y, 0.0..=1.0);
+            fun(pos2(x, y))
+        }
```
emilk, line 925: I suggest we keep this simple. We can always have constructors to `ColorMode` that does helpful coordinate transforms

```suggestion
        ColorMode::UV(fun) => {
            fun(bbox, pos2(x, y))
        }
```

crates/epaint/src/tessellator.rs
```diff
@@ -864,19 +875,58 @@ fn stroke_path(
...
     path: &[PathPoint],
     path_type: PathType,
-    stroke: Stroke,
+    stroke: &PathStroke,
     out: &mut Mesh,
 ) {
     let n = path.len() as u32;
 
-    if stroke.width <= 0.0 || stroke.color == Color32::TRANSPARENT || n < 2 {
+    if stroke.width <= 0.0 || stroke.color == ColorMode::Solid(Color32::TRANSPARENT) || n < 2 {
```
emilk, line 883: we can add a helper for this

```suggestion
    if stroke.width <= 0.0 || stroke.color == ColorMode::TRANSPARENT || n < 2 {
```

examples/hello_world/src/main.rs
```diff
@@ -52,6 +57,36 @@ impl eframe::App for MyApp {
...
+                stroke.clone(),
+            ));
+
+            ui.painter_at(rect).add(Shape::line_segment(
+                [
+                    rect.left_center() + vec2(25.0, 15.0),
+                    rect.right_center() + vec2(0.0, 15.0),
+                ],
+                stroke.clone(),
+            ));
```
emilk, line 89: This is not the right place for this test/example.

Either add a standalone example, or add a new demo to `egui_demo_lib`

crates/epaint/src/stroke.rs
```diff
@@ -52,3 +54,116 @@ impl std::hash::Hash for Stroke {
         color.hash(state);
     }
 }
+
+/// How paths will be colored.
+#[derive(Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub enum ColorMode {
```
emilk, line 61: Let's move this to its own file. In the future we may very well start using this for the fill color of primitives too (because why not!)

crates/epaint/src/tessellator.rs
```diff
@@ -904,7 +956,7 @@ fn stroke_path(
                 let p = p1.pos;
                 let n = p1.normal;
                 out.colored_vertex(p + n * feathering, color_outer);
-                out.colored_vertex(p, color_inner);
+                out.colored_vertex(p, get_color(color_inner, p));
```
emilk, line 959: This no longer fades out as it gets thinner

crates/epaint/src/tessellator.rs
```diff
@@ -1067,19 +1123,22 @@ fn stroke_path(
...
                 out.colored_vertex(p.pos + radius * p.normal, color);
                 out.colored_vertex(p.pos - radius * p.normal, color);
             }
         } else {
             let radius = stroke.width / 2.0;
             for p in path {
-                out.colored_vertex(p.pos + radius * p.normal, stroke.color);
-                out.colored_vertex(p.pos - radius * p.normal, stroke.color);
+                out.colored_vertex(p.pos + radius * p.normal, get_color(&stroke.color, p.pos));
+                out.colored_vertex(p.pos - radius * p.normal, get_color(&stroke.color, p.pos));
```
emilk, line 1141: The new color code has the potential of slowing down this function.

I'd like you to add a benchmark to `crates/epaint/benches/benchmark.rs` to test this for two cases:
* thin solid line (stroke width < feathering)
* thick solid line (stroke width > feathering)

emilk (commented):
This looks very promising! `ColorMode` has the potential to be useful for more things in the future too, like the fill color for shapes (not in this PR though!)

crates/epaint/src/tessellator.rs
murl-digital: I was considering putting the bounding box in a once_cell::Lazy to potentially save on having to do the bounding box calculations if it's not needed, do you think that's necessary?

crates/epaint/src/tessellator.rs
murl-digital: resolved with 66644820a93a76758da3d8f8f5d7b03ba88f79a9

crates/epaint/src/tessellator.rs
murl-digital: resolved with c9efc240d03842e16716a92d1077d84ee06e7543

crates/epaint/src/tessellator.rs
murl-digital: resolved with 5451e3043796568c4bce0895d01cfd118e2d1f01

crates/epaint/src/tessellator.rs
murl-digital: the simplify suggestion kind of renders this moot, i'll investigate with the new benchmarks later.

(66644820a93a76758da3d8f8f5d7b03ba88f79a9)

crates/epaint/src/stroke.rs
murl-digital: did the thing in 41f819214bfac6bb9da35667ff4a91a0bf2cc5c6

crates/epaint/src/tessellator.rs
murl-digital: resolved with 2d8ab0ce5d44c936657d61f815ca3eb4426febe4

examples/hello_world/src/main.rs
murl-digital: resolved in a8e0e57dc2a4e0cc2933dd3c8bfd146beb88d2de

crates/egui_demo_lib/Cargo.toml
```diff
@@ -50,6 +50,9 @@ chrono = { version = "0.4", optional = true, features = ["js-sys", "wasmbind"] }
 ## Enable this when generating docs.
 document-features = { workspace = true, optional = true }
 serde = { version = "1", optional = true, features = ["derive"] }
+once_cell = "1.19"
+colorgrad = "0.6"
+noise = "0.9"
```
emilk, line 55: No please don't - we already have enough dependencies to compile as it is.

And you can already parse hex colors: https://docs.rs/ecolor/latest/ecolor/macro.hex_color.html

crates/epaint/benches/benchmark.rs
```diff
@@ -72,10 +72,74 @@ fn tessellate_circles(c: &mut Criterion) {
...
+
+        b.iter(|| {
+            let mut mesh = Mesh::default();
+            path.stroke_closed(1.5, &Stroke::new(0.5, Color32::RED).into(), &mut mesh);
+
+            black_box(mesh);
+        });
+    });
+}
+
```
emilk, line 134: What are the results of running these benchmarks before your other code changes, and what were the results after?

murl-digital:
```
=== with changes ===

thick_solid_line        time:   [149.88 ns 149.96 ns 150.04 ns]
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe

thick_large_solid_line  time:   [20.033 µs 20.140 µs 20.340 µs]
  5 (5.00%) high mild
  2 (2.00%) high severe

thin_solid_line         time:   [119.98 ns 120.12 ns 120.27 ns]
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) low mild
  4 (4.00%) high mild

thin_large_solid_line   time:   [14.664 µs 14.675 µs 14.687 µs]
  1 (1.00%) low severe
  1 (1.00%) low mild
  2 (2.00%) high mild
  
=== without changes (run on the current master branch) ===

thick_solid_line        time:   [138.18 ns 138.34 ns 138.51 ns]
                        change: [-7.7412% -7.5862% -7.4276%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe

thick_large_solid_line  time:   [16.391 µs 16.403 µs 16.415 µs]
                        change: [-18.625% -18.275% -18.045%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild

thin_solid_line         time:   [95.293 ns 95.372 ns 95.456 ns]
                        change: [-20.478% -20.365% -20.252%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

thin_large_solid_line   time:   [12.709 µs 12.719 µs 12.729 µs]
                        change: [-13.525% -13.419% -13.315%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) low mild
  1 (1.00%) high mild
```

The other benchmarks also improved, so there is some run to run variance. Overall, in my opinion at least, the overhead is negligible.

crates/egui_demo_lib/Cargo.toml
murl-digital: I can remove the once_cell and colorgrad dependencies, but unless you want me to throw together a noise function for the color demo, I'd like to keep the noise crate in. If that's too much, I can see what I can do.

emilk: Just use a sine wave or something?

crates/epaint/benches/benchmark.rs
emilk: I agree - that's close enough for the benefit this brings. Thanks for running the numbers!

murl-digital: No problem!

crates/egui_demo_lib/Cargo.toml
murl-digital: I'll roll it in now, for some reason color_hex isn't resolving and I have no idea why, maybe it's a silly mistake I made

emilk: Have you enabled the `color-hex` in `ecolor`?

murl-digital: I've tried enabling it in both `egui` and `ecolor`, for some reason it's just refusing to include the color_hex dependency. I think it's a bug on my end, so I'm asking someone else to see if it runs on their machine.

murl-digital: This was shaping up to be a massive mystery, until I saw #4372. Updating my branch fixed the issue. I did have to add back once_cell because color_hex!() can't be used in consts, maybe another PR is in order?

crates/epaint/src/tessellator.rs
```diff
@@ -864,19 +873,29 @@ fn stroke_path(
...
+    if stroke.width <= 0.0 || stroke.color == ColorMode::TRANSPARENT || n < 2 {
         return;
     }
 
     let idx = out.vertices.len() as u32;
 
+    // expand the bounding box to include the thickness of the path
+    // this tiny bit of margin is here because a rect doesn't think it contains a point if a pos' component is on a boundary. why? i have no idea.
+    let bbox = Rect::from_points(&path.iter().map(|p| p.pos).collect::<Vec<Pos2>>())
+        .expand((stroke.width / 2.0) + feathering + 1.0);
```
emilk, line 890: Probably because of rounding errors. If you change the `.contains(x)` test to `distance_to_pos(x) < 0.1` it will probably work. That is better than adding an arbitrary `+ 1.0` here methinks

murl-digital: When doing some debugging prints, the edge case happens when they're dead equal, which `.contains()` doesn't seem to handle. I'll try the distance check and see if that helps at all, though.

emilk: `Rect::contains` should handle it:

```
    pub fn contains(&self, p: Pos2) -> bool {
        self.min.x <= p.x && p.x <= self.max.x && self.min.y <= p.y && p.y <= self.max.y
    }
```

crates/egui_demo_lib/Cargo.toml
emilk: `once_cell` is fine

crates/epaint/src/tessellator.rs
murl-digital: assertion failure is spitting out the following, really strange: `passed rect [[-49.0 -49.0] - [59.0 59.0]] didn't contain point [59.1 4.3]`, so it's exactly 0.1 out. I can just have it be <= instead of just <, but that doesn't feel right.

murl-digital: So, it seems that the error gets larger as the path width increases, I have no idea what's going on, but the problem seems to be cropping up starting at line 971.

murl-digital:
While playing around with trying to remove the `'static` lifetime bound when making a new ColorMode, I found that adding a lifetime bound to the type as a whole seemed to work, however the change bubbles out to all of egui very quickly, and would very much be a breaking change. I think this is something that should be looked into further

emilk (approved):

Issue #4356, opened 2024-04-13
Add the ability for users to pass settings into the Undoer API


valadaptive:
**Is your feature request related to a problem? Please describe.**
As far as I can tell, the undoer API provides a [publicly-constructible `Settings` struct](https://docs.rs/egui/0.27.2/egui/util/undoer/struct.Settings.html), but unless I'm missing something, there is no way to actually create an `Undoer` using it.

**Describe the solution you'd like**
An API that allows you to create an `Undoer` with given settings, something like `Undoer::with_settings(...)` or a builder API for `Settings`.

**Describe alternatives you've considered**
N/A

**Additional context**
Does the API exist somewhere and I'm just not finding it?

PR #4357, opened 2024-04-13
Add a way to specify Undoer settings and construct Undoers more easily

valadaptive:
* Closes https://github.com/emilk/egui/issues/4356

- Add `Undoer::new()`
  - This is necessary to construct an `Undoer` whose `State` parameter doesn't implement `Default`.
- Add `Undoer::with_settings(...)`
  - This is necessary to actually pass settings into the `Undoer`. Without this, API consumers could construct their own `Settings` but not actually do anything with it.

I've refrained from adding any kind of builder API for `Settings` because there are only three options and I don't want to duplicate or move all the documentation onto the builder methods.

### DIFF
--- a/crates/egui/src/util/undoer.rs
+++ b/crates/egui/src/util/undoer.rs
@@ -47,7 +47,7 @@ impl Default for Settings {
 ///
 /// Rule 1) will make sure an undo point is not created until you _stop_ dragging that slider.
 /// Rule 2) will make sure that you will get some undo points even if you are constantly changing the state.
-#[derive(Clone, Default)]
+#[derive(Clone)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub struct Undoer<State> {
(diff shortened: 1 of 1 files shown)


crates/egui/src/util/undoer.rs
```diff
@@ -89,6 +89,23 @@ impl<State> Undoer<State>
...
     State: Clone + PartialEq,
 {
+    pub fn new() -> Self {
+        Self {
+            settings: Settings::default(),
+            undos: VecDeque::new(),
+            redos: Vec::new(),
+            flux: None,
+        }
+    }
```
emilk, line 99: This is better done with an explicit `

```
impl<T> Default for Undoer<T> {
    …
}
```

It's only the derive macro (`#[derive(Default)]`) that fails when `T: !Default`

valadaptive: Done

emilk (approved):

PR #4358, opened 2024-04-13
Improve IME support with new `Event::Ime`

### LINKED ISSUE #4354: i can't input chinese words into singleline properly
[patch.crates-io]
`eframe = { features = [
    "default",
] }`
**Describe the bug**

When using the chinese input method, you can only successfully input once until the single line content is cleared.

**To Reproduce**
1. use chinese input method editor.
2. type some chinese words in the textEdit singleline
3. repeat step 2

**Screenshots**

https://github.com/emilk/egui/assets/73056742/39da1a88-b220-4bf0-9f32-e80756dbd02a

 - OS: windows 11

rustbasic:
* Closes #4354 

Fix: can't repeat input chinese words

AND

For Windows :
ImeEnable
ImeDisable


crates/egui/src/data/input.rs
```diff
@@ -445,6 +445,12 @@ pub enum Event {
...
     Zoom(f32),
 
+    /// IME Enable
+    ImeEnable,
+
+    /// IME Disable
+    ImeDisable,
+
     /// IME composition start.
     CompositionStart,
```
emilk, line 455: Please add docstrings that explain these things. What does it mean to enable IME? How does that differ from `CompositionStart`?

rustbasic: CompositionStart was created by egui,
ImeEnable is what happens by default in Windows (winit?).

in China's IME, characters are sometimes added without CompositionStart (Preedit?).

rustbasic: I think, `Event::CompositionStart` may be `deprecated`.

Alternatively, if there is no confusion, we can use other names instead of winit's `Event::ImeEnable` and 'Event::ImeDisable' to 'Event::CompositionStart` and 'Event::CompositionEnd`.

emilk: Please see if my comited code works well for you

rustbasic: > Please see if my comited code works well for you

Yes, it works fine.

AND,
I would like to save the `TextEdit ID` in the same format as #4374 to prevent the entered text from being copied to another TextEdit.

* Related #4374

emilk (approved):

emilk:
This is getting annoying:
Screenshot 2024-04-22 at 11 24 31

rustbasic:
Check to
eframe/src/web/text_agent.rs

rustbasic:
> This is getting annoying:

sorry about that.
This happens because `rustbasic/egui/master` gets weird.
I don't know how to fix it. Right now, you can't completely delete it and create a new one.

PR #4360, opened 2024-04-14
Include punctuation when double-clicking words to select them

rustbasic:
Determines whether a character is a component of a word.

This is In general, it feels good to use.

This works well even when using UTF8 characters rather than English-speaking characters.

Before : This is a double clicked on a word.

After :  This is a double clicked on a word.


crates/egui/src/text_selection/text_cursor_state.rs
```diff
@@ -291,6 +292,18 @@ pub fn is_word_char(c: char) -> bool {
     c.is_ascii_alphanumeric() || c == '_'
 }
 
+/// Determines whether a character is a component of a string.
+///
+/// Punctuation marks that are commonly used with strings are considered string components.
+pub fn is_string_component(c: char) -> bool {
+    if c == '_' || c == '-' || c == ':' || c == '/' || c == '.' || c == '\\' || c == '@' || c == '#'
```
YgorSouza, line 299: You can use `matches!` instead. See <https://rust-lang.github.io/rust-clippy/master/index.html#/string_lit_chars_any>.

crates/egui/src/text_selection/text_cursor_state.rs
```diff
@@ -291,6 +292,18 @@ pub fn is_word_char(c: char) -> bool {
     c.is_ascii_alphanumeric() || c == '_'
 }
 
+/// Determines whether a character is a component of a string.
+///
+/// Punctuation marks that are commonly used with strings are considered string components.
+pub fn is_string_component(c: char) -> bool {
```
YgorSouza, line 298: This name is confusing. Any `char` can be part of a `String`, so it's not clear what "string" means in this context.

rustbasic: Since I don't speak English, I don't know what would be better.

crates/egui/src/text_selection/text_cursor_state.rs
rustbasic: I don't think it's a place that requires high speed.

I think In general, rust is fast enough, so it's better to think about readability rather than speed.

emilk: `matches!(c, '_' | '-' | ':' | …)` is much more readable than repeating `c ==` over and over again

rustbasic: > `matches!(c, '_' | '-' | ':' | …)` is much more readable than repeating `c ==` over and over again

Okay. Edit has been completed.

in URL, '?' and '=' were also added because they are often used.

crates/egui/src/text_selection/text_cursor_state.rs
emilk: What is it trying to accomplish? Should this be a change to `is_word_char` instead?

rustbasic: A new name was used so as not to affect the existing code.
but, Changed to `is_word_char()`.

rustbasic: > What is it trying to accomplish? Should this be a change to `is_word_char` instead?

A new name was used so as not to affect the existing code.
but, Changed to is_word_char().

emilk (approved):

emilk:
This is not how mac works though - double-clicking a word does NOT include `-` or `.`

Screenshot 2024-05-11 at 00 39 52
Screenshot 2024-05-11 at 00 39 48

rustbasic:
> This is not how mac works though - double-clicking a word does NOT include `-` or `.`

I can only test Windows.
This function `is_word_char()` works fine.
The problem on mac does not seem to be an issue with `is_word_char()`. 

(
It looks like changes to 'is_word_char()' were not applied at compile time.
Wouldn't the changed to 'is_word_char()' have been applied at compile time? 
)

emilk:
If the new code matches the behavior of native windows, then check if `ctx.os() == OperatingSystem::Windows`

rustbasic:
> If the new code matches the behavior of native windows, then check if `ctx.os() == OperatingSystem::Windows`

The `is_word_char()` function simply returns a bool value, so I think it's a good idea to keep it that way.
I think it would be better to fix the `mac` issue later.
Otherwise, please edit as you wish.

emilk:
I want to match how text editing works natively on whatever platform egui runs on.

If double-clicking `foo-bar` selects both `foo` and `bar` on windows, then let's make it work like that, but only on Windows (because on Mac, words don't extend across hyphen on double-click).

I don't know how Windows works though.

rustbasic:
> I want to match how text editing works natively on whatever platform egui runs on.
> 
> If double-clicking `foo-bar` selects both `foo` and `bar` on windows, then let's make it work like that, but only on Windows (because on Mac, words don't extend across hyphen on double-click).
> 
> I don't know how Windows works though.

yes. In `windows` everything works fine. 
Not only `foo-bar`, but also URL, e-mail, path, UTF8 characters (Korean), etc. all work well. 
I think the reason it doesn't work on `mac` has nothing to do with this function.

YgorSouza:
> > I want to match how text editing works natively on whatever platform egui runs on.
> > If double-clicking `foo-bar` selects both `foo` and `bar` on windows, then let's make it work like that, but only on Windows (because on Mac, words don't extend across hyphen on double-click).
> > I don't know how Windows works though.
> 
> yes. In `windows` everything works fine. Not only `foo-bar`, but also URL, e-mail, path, UTF8 characters (Korean), etc. all work well. I think the reason it doesn't work on `mac` has nothing to do with this function.

Which Windows app works like that? Here is Notepad:

https://github.com/emilk/egui/assets/43298013/2f897e66-33a2-4470-b5eb-2a52e42e5037

rustbasic:
> Which Windows app works like that? Here is Notepad:

This was a response to egui_demo_app that applied this commit.
I don't understand the nuances of English.

YgorSouza:
> > Which Windows app works like that? Here is Notepad:
> 
> This was a response to `egui_demo_app` that applied this commit. I don't understand the nuances of English.

What emilk is saying is that egui should work exactly like other programs in each operating system. So you should probably change:

```rs
    if matches!(
        c,
        '_' | '-' | ':' | '/' | '.' | '\\' | '@' | '#' | '?' | '='
    )
```

to 

```rs
    if matches!(c, '_')
```

rustbasic:
> What emilk is saying is that egui should work exactly like other programs in each operating system.

In Notepad on Windows 10, many things such as `.`, `_`, `-`, `;`, `(`, `)` are selected together when double-clicked.
And what is selected when double-clicking is unique to egui and does not need to be copied by other programs.

PR #4362, opened 2024-04-14
Fix: On Windows, Strange IME display after update #4269

rustbasic:
Fix: On windows, Strange IME display after update #4269

* Related #4254
* Related #4269


crates/egui-winit/src/lib.rs
```diff
@@ -820,11 +821,16 @@ impl State {
...
             let pixels_per_point = pixels_per_point(&self.egui_ctx, window);
-            let ime_rect_px = pixels_per_point * ime.rect;
-            if self.ime_rect_px != Some(ime_rect_px)
-                || self.egui_ctx.input(|i| !i.events.is_empty())
-            {
-                self.ime_rect_px = Some(ime_rect_px);
+            let ime_rect_px = ime.rect * pixels_per_point;
+            let mut need_set_ime_cursor_area = true;
+
+            // On Wayland of Linux, repaints every frame Issue : See https://github.com/emilk/egui/pull/4254
```
emilk, line 827: Wasn't this fixed in https://github.com/emilk/egui/pull/4269 ?

crates/egui-winit/src/lib.rs
```diff
@@ -820,11 +821,16 @@ impl State {
...
+            if self.egui_ctx.os() == egui::os::OperatingSystem::Nix {
+                need_set_ime_cursor_area = self.ime_rect_px != Some(ime_rect_px)
+                    && self.egui_ctx.input(|i| !i.events.is_empty());
```
emilk, line 830: Do we need to treat `Nix` specially here? Isn't this logic sound for all OSes?

rustbasic: I think it should only be applied on Linux.
Definitely shouldn't be applied on Windows.

crates/egui-winit/src/lib.rs
rustbasic: Since the problem occurs on Windows, we have moved it down to only apply on Linux.

Issue #2644, opened 2023-01-29
`hex_color!` macro uses `color_hex` crate, but it doesn't seem to import it correctly


Hpmason:
Trying to use the `hex_color!` macro gives the following error:
```
error[E0433]: failed to resolve: use of undeclared crate or module `color_hex`
  --> src\main.rs:29:61
   |
29 |         let frame = egui::containers::Frame::default().fill(hex_color!("#ffffff"));
   |                                                             ^^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `color_hex`
   |
   = note: this error originates in the macro `hex_color` (in Nightly builds, run with -Z macro-backtrace for more info)     

For more information about this error, try `rustc --explain E0433`.
```

From what I can tell, the macro calls another macro from the `color_hex` crate, but there's no guarantee the crate is in scope. Maybe in version 0.19, the `color_hex` crate was reexported so it was always in scope?

https://github.com/emilk/egui/blob/f222ee044edf8beebfaf5dd7be15c9f318f20886/crates/ecolor/src/hex_color_macro.rs#L9-L11

emilk:
Ah, good catch. I guess `ecolor` would need to re-export `color_hex`, and the macro would need to call `#crate::color_hex::color_from_hex`. Maybe? PRs welcome!

PR #4372, opened 2024-04-17
Fix `hex_color!` macro by re-exporting `color_hex` crate from `ecolor`

dataphract:
The `hex_color!` macro currently makes an unqualified reference to the `color_hex` crate, which users may not have in their `Cargo.toml`. This PR re-exports `color_hex` from the root of `ecolor` and changes the macro to use the re-exported path.

* Closes https://github.com/emilk/egui/issues/2644


emilk (approved):
Good catch

PR #4374, opened 2024-04-17
Fix: A symptom of characters being copied when there are multiple TextEdits.

### LINKED ISSUE #4354: i can't input chinese words into singleline properly

rustbasic:
* Related #4358

Fix: A symptom of characters being copied when there are multiple TextEdits.


crates/egui/src/widgets/text_edit/state.rs
```diff
@@ -44,6 +44,10 @@ pub struct TextEditState {
     #[cfg_attr(feature = "serde", serde(skip))]
     pub(crate) has_ime: bool,
 
+    // target ID for IME candidate.
+    #[cfg_attr(feature = "serde", serde(skip))]
+    pub(crate) ime_target_id: Id,
```
emilk, line 49: Won't this always be the same `Id` as the key used to find `TextEditState`?

rustbasic: Yes, it only stores TextEdit ID.                                                      
It's currently working fine without any problems.                                                                                      
If something is wrong or needs to be fixed, please fix it or let me know.

emilk: That the code is working fine is not enough. I need to maintain this, and so I need to understand it. Storing the `Id` here makes no sense to me, since it is already the key.

emilk:
Blocked on https://github.com/emilk/egui/pull/4358

rustbasic:
@emilk 

for example,
This is necessary because after entering the ID, if you select the Password line to enter the password, part of the ID will be copied.

Please review.

PR #4377, opened 2024-04-17
Fix incorrect line breaks

juancampa:
While breaking a paragraph, it was possible to lose line break candidates that could've been used on the next line, causing egui to unnecessarily overrun `wrap.max_width`.

This PR fixes it so that we don't forget about those candidates.

Before:
Note that the window can't resize to the requested width because the text is not wrapping.

https://github.com/emilk/egui/assets/1410520/6430a334-2995-4b40-bc34-8f01923f9f95

After:

https://github.com/emilk/egui/assets/1410520/225fa4cd-cbbb-4a7e-9580-7f1814c05ee7

### DIFF
--- a/crates/epaint/src/text/text_layout.rs
+++ b/crates/epaint/src/text/text_layout.rs
@@ -296,7 +296,7 @@ fn line_break(paragraph: &Paragraph, job: &LayoutJob, out_rows: &mut Vec<Row>, e
                 // Start a new row:
                 row_start_idx = last_kept_index + 1;
                 row_start_x = paragraph.glyphs[row_start_idx].pos.x;
-                row_break_candidates.forget_candiates_before_idx(row_start_idx);
+                row_break_candidates.forget_before_idx(row_start_idx);
             } else {
                 // Found no place to break, so we have to overrun wrap_width.
             }
@@ -944,7 +944,7 @@ impl RowBreakCandidates {
         }
     }
 
-    fn forget_candiates_before_idx(&mut self, index: usize) {
+    fn forget_before_idx(&mut self, index: usize) {
         if self.space.map(|s| s < index).unwrap_or_default() {
             self.space = None;
         }
--- a/crates/epaint/src/text/text_layout.rs
+++ b/crates/epaint/src/text/text_layout.rs
@@ -945,23 +945,31 @@ impl RowBreakCandidates {
     }
 
     fn forget_before_idx(&mut self, index: usize) {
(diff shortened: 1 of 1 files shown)


juancampa:
@emilk in our fork we also removed the last line from here: https://github.com/emilk/egui/blob/99bb0124f5ebbe07c275551402b3c6478162a2ec/crates/epaint/src/text/text_layout.rs#L936-L945

We removed `.or(self.any)` because when break_anywhere is false, we would rather overrun the max_width instead of splitting words into single characters vertically (which looks very broken). Last time I checked, that was the behavior of browsers too, Happy to also apply this change here.

crates/epaint/src/text/text_layout.rs
```diff
@@ -943,6 +943,27 @@ impl RowBreakCandidates {
...
+        if self.dash.map(|s| s < index).unwrap_or_default() {
+            self.dash = None;
+        }
+        if self.punctuation.map(|s| s < index).unwrap_or_default() {
+            self.punctuation = None;
+        }
+        if self.any.map(|s| s < index).unwrap_or_default() {
+            self.any = None;
+        }
+    }
```
emilk, line 966: This is a good trick to ensure we don't miss any member (if we add more in the future)

```suggestion
    fn forget_before_idx(&mut self, index: usize) {
        let Self {
            space,
            cjk,
            pre_cjk,
            dash,
            punctuation,
            any,
        } = self;
        if space.map_or(false, |s| s < index) {
            *space = None;
        }
        if cjk.map_or(false, |s| s < index) {
            *cjk = None;
        }
        if pre_cjk.map_or(false, |s| s < index) {
            *pre_cjk = None;
        }
        if dash.map_or(false, |s| s < index) {
            *dash = None;
        }
        if punctuation.map_or(false, |s| s < index) {
            *punctuation = None;
        }
        if any.map_or(false, |s| s < index) {
            *any = None;
        }
    }
```

emilk (approved):
Good catch!

emilk:
> We removed `.or(self.any)` because when break_anywhere is false, we would rather overrun the max_width instead of splitting words into single characters vertically (which looks very broken). Last time I checked, that was the behavior of browsers too, Happy to also apply this change here.

I can see both behaviors being valuable. The current behavior makes sense when you really _can't_ expand the parent, so you want it as thing as possible. And browser do it sometimes:

Screenshot 2024-04-18 at 10 05 40

Let's consider it for another PR

emilk (approved):

PR #4380, opened 2024-04-19
Expose `ClosestElem` and `PlotConfig`

Narcha:
These two items are needed to implement the `PlotItem` trait (which is already public) when using `PlotGeometry::Rects`.
Specifically, they are used in the return type of `PlotItem::find_closest` and as arguments to `PlotItem::on_hover`, which need to be implemented when using `PlotGeometry::Rects`.


emilk (approved):

Narcha:
Thanks for the quick merge! When can we expect the next release?

Issue #4219, opened 2024-03-23
Update `accesskit_winit`


mwcampbell:
accesskit_winit is on 0.18 now, and we have another refactor on the way which will be released in accesskit_winit 0.19. @emilk Do you need help from us to keep egui-winit up to date, or is there something on our end that's blocking the update?

emilk:
It depends on how much work it is, I guess.

If the change in accesskit when upgrading was just a version bump (no change in code needed), then you could support multiple version of `objc2` etc (`objc = ">=0.4.0 <= 0.5.0"`) which would be the best of two worlds.

Duplicated dependencies are super annoying, as it adds to the already long compile time of `eframe`

DataTriny:
It looks like winit 0.30 is coming soon. I'd suggest we update all at once. We need to move away from icrate as quickly as possible because of its number of compile features.

See https://github.com/rust-windowing/winit/pull/3634

mwcampbell:
#4849 fixed this.

PR #4395, opened 2024-04-23
eframe: Use `objc2` and its framework crates

madsmtm:
These are a replacement to the `objc` and `cocoa` crates.

This PR prevents:
- An extra copy when creating `NSData`
- A memory leak when creating `NSImage`
- A memory leak when creating `NSString`

And is generally a readability improvement.

Note that we define `NSApp` manually for now, the implementation in `objc2-app-kit` is currently suboptimal and wouldn't allow you to check whether the NSApplication has been created or not.

Related: https://github.com/emilk/egui/issues/4219, this should nicely coincide with the Winit `0.30` release.


emilk (approved):
This is great - thank you!

madsmtm:
Thanks for fixing the build!

chgxtony:
May I ask why this version ` use objc2 v0.3.0-beta.3.patch-leaks.3 ` is used instead of an updated version

madsmtm:
I'm not sure what you mean? This uses `objc2 v0.5.1`?

PR #4408, opened 2024-04-25
eframe: update ViewportBuilder.with_icon() documentation

roccoblues:
`None` is no longer correct. To disable the egui icon one has to pass `IconData::default()` now.


emilk (approved):
Good catch!

PR #4410, opened 2024-04-25
Fix : In Windows, the 'egui_demo_app' screen does not appear.

### LINKED ISSUE #4409: After the #4337 update, the 'egui_demo_app' screen does not appear.
In Windows,
When running 'egui_demo_app', the 'egui_demo_app' screen does not appear, and it closes after a while,
After the #4337 update,

And the screen is often broken.

egui version: master
OS: Windows 10

rustbasic:
* Related #4337 
* Closes #4409 

Fix : In Windows, the 'egui_demo_app' screen does not appear After the #4337 update.


crates/eframe/src/native/epi_integration.rs
```diff
@@ -20,10 +20,8 @@ pub fn viewport_builder<E>(
 
     let mut viewport_builder = native_options.viewport.clone();
 
-    let clamp_size_to_monitor_size = viewport_builder.clamp_size_to_monitor_size.unwrap_or(
-        // On some Linux systems, a window size larger than the monitor causes crashes
-        cfg!(target_os = "linux"),
-    );
+    // On some Linux systems, a window size larger than the monitor causes crashes
```
emilk, line 23:
```suggestion
    // On some Linux systems, a window size larger than the monitor causes crashes,
    // and on Windows the window does not appear at all.
```

emilk (approved):

PR #4413, opened 2024-04-25
Replace cargo-cranky with workspace lints

emilk:
Replace `cargo-cranky` (which has served us well) with workspace lints


torokati44:
> (which has served us well)

https://github.com/ericseppanen/cargo-cranky/issues/15 :slightly_smiling_face:

PR #4419, opened 2024-04-26
Treat `Event::PointerGone` as `PointerEvent::Released`

### LINKED ISSUE #4406: Scrollbars and other draggable elements get stuck mouse leaves browser window
**Describe the bug**

This bug is exclusive to wasm.
When dragging a scroll bar and then leaving the browser window with the mouse makes the scrollbar "stuck". It will only be unstuck by clicking into the browser window again, until you click again there will also be no hover effects or any other interaction.

I was able to observe and consistently product this bug on both windows and debian linux.

**To Reproduce**
1. Open egui.rs
2. Open any window with scrolling, for instance the scrolling demo window
3. Click and hold the scroll handle
4. Move your mouse outside the browser window without releasing the mouse button
5. Try to interact with anything else
6. After clicking the scrollbar will jump

**Expected behavior**

I would expect the scroll bar to treat the mouse leaving the window as releasing the mouse button. 

I tested this on both Debian and Ubuntu Linux as well as Windows in Firefox, Chromium and Chrome, the behavior was the same in every browser.
The problem seems to be exclusive to wasm.

**Additional context**

When given pointers i would definitely be willing to track down the cause of the problem. It seems to be quite deep as other draggables such as sliders inhibit similar but not as disrupting problems

Thanks for all your hard work and have a great day ::)

### LINKED ISSUE #4418: Not detecting mouse release after alt-tab on windows
# Environment
App: egui_demo_app (+ other egui-based apps)
OS: Windows 11
Git: 0bc59f57

# Describe the bug
1. Open the 'input test' demo.
3. Click and hold `Sense::drag`
4. Move cursor off egui window
5. Alt-tab to a different window
6. Release mouse
7. Alt-tab back to egui window
8. Move cursor around
9. Mouse is erroneously considered pressed, even though it was released
10. Click mouse on egui window to fix state

# Expected
Mouse should be detected as released once it is released. Other native windows apps seem to do this correctly, e.g. ms paint.

rustbasic:
* Closes #4406
* Closes #4418 

If `Event::PointerGone` occurs, it is treated as `PointerEvent::Released`.


rustbasic:
For correct operation, `PointerButton` must also be stored in each item of `struct InteractionSnapshot`.
This is a lot of work, so it's put on hold.

crates/egui_demo_lib/src/demo/tests.rs
```diff
@@ -456,33 +456,28 @@ fn response_summary(response: &egui::Response, show_hovers: bool) -> String {
         egui::PointerButton::Extra1,
         egui::PointerButton::Extra2,
     ] {
-        let button_suffix = if button == egui::PointerButton::Primary {
-            // Reduce visual clutter in common case:
-            String::default()
-        } else {
-            format!(" by {button:?} button")
-        };
```
emilk, line 464: please revert

emilk (approved):
Makes sense, but please revert unrelated changea

emilk (approved):

Issue #914, opened 2021-11-26
Storage options: selecting where to save app


saona-raimundo:
The current way to have a persistent app is to implement [`epi::App`](https://docs.rs/epi/0.15.0/epi/trait.App.html) and use the feature `persistence`.
In particular, one must implement the method [`save`](https://docs.rs/epi/0.15.0/epi/trait.App.html#method.save).

`save` uses a default location given by:
- `Local Storage` on the web
- [`directories_next::ProjectDirs::data_dir`](https://docs.rs/directories-next/2.0.0/directories_next/struct.ProjectDirs.html#method.data_dir) on native.

But there is no way (as far as I can see) to change the default location.

**Describe the solution you'd like**

I would like to be able to choose where the data is stored. 

**Describe alternatives you've considered**

I propose to exhibit the call to create a [dyn `epi::Storage`](https://docs.rs/epi/0.15.0/epi/trait.Storage.html), so people could change it. 

We could add a method to the [`epi::App`](https://docs.rs/epi/0.15.0/epi/trait.App.html) trait
```rust
fn storage(&self) -> dyn Storage { ... }
```
A default implementation is given by the current behaviour.

**Additional context**

A very simple use case is that I would like to store the app data in a Google drive folder, so I get "persistence" across devices too.

wmte:
> Perhaps the path could be an option in `NativeSettings` ?

I worry this does not address user-selectable paths (ie, a typical File->Save action) which I am trying to unify with the current persistence mechanism.

By [exposing FileStorage](https://github.com/emilk/egui/commit/194507cebd225af587574fa7075739f2de4f15d9) I was able to save to a chosen path on user interaction. In my application I use the default storage behavior as a kind of 'session-resume', but would like to also have manual save and load as I expect users to generate multiple application states.

emilk:
Closed by https://github.com/emilk/egui/pull/4423

PR #4423, opened 2024-04-27
eframe: Add `NativeOptions::persistence_path`

### LINKED ISSUE #4098: android persistence feature
Hello,

It seems `eframe` [directories-next](https://crates.io/crates/directories-next) dependencies can't handle android.

Any idea how we can implement persistence on systems other than Linux, macOS, Redox and Windows?

Thank you.

lucasmerlin:
This allows customizing the persistence path in NativeOptions. Previously, persistence wouldn't work with android because directories-next doesn't support android so eframe would just fail to find a place where it could store its config. 

* Closes #4098 (android users can now specify a path that works with android, by e.g. using app_dirs2, which supports android)

### DIFF
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -362,6 +362,10 @@ pub struct NativeOptions {
     /// Controls whether or not the native window position and size will be
     /// persisted (only if the "persistence" feature is enabled).
     pub persist_window: bool,
+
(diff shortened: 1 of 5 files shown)


YgorSouza:
I think this also addresses  #914.

emilk (approved):
Looks good. We should link to `NativeOptions::persistence_path` from the docs of `App::save` too

Issue #4188, opened 2024-03-18
Using PathShape::line to draw an arc results in jagged edges at larger radii


murl-digital:
**Describe the bug**
This could also serve as a feature request, but currently I have a usecase where I need to draw arcs of circles. I have an implementation that works for smaller radii (sub 25 pixels), but it rapidly gets more jaggy and aliased as the radius increase.

**To Reproduce**
This is the code I'm using, you get the same results if you just iterate over a range of integer degrees.
```
  let circle_radius = 100.0;
  let mut circle_points = Vec::new();
  let visual_angle = 180.0 + correcting_angle;
  let angle_range = 180.0_f32.max(visual_angle)..180.0_f32.min(visual_angle);
  for i in 1..=512 {
      let (sin, cos) = (i as f32).map_range(1.0_f32..512.0_f32, angle_range.clone()).to_radians().sin_cos();
  
      circle_points.push(painter.clip_rect().center() + Vec2::new(circle_radius * sin, circle_radius * cos));
  }
  painter.add(PathShape::line(circle_points, Stroke::new(2.0, Color32::GREEN)));
```

**Screenshots**
![image](https://github.com/emilk/egui/assets/8054329/526ff98a-703f-4b8e-b3d1-5c7f824b76eb)
A side by side example of arcs at 2 different radii

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser Baseview backend
 - Version 0.26

dataphract:
The typical way to render circular arcs is to approximate them using cubic Bezier curves of up to π/2 radians, using multiple curves to construct larger arcs. The computation for a single curve looks like

```rust
    // Center of the circle
    let xc = center.x;
    let yc = center.y;

    // First control point
    let p1 = center + radius * Vec2::new(start_angle.cos(), -start_angle.sin());

    // Last control point
    let p4 = center + radius * Vec2::new(end_angle.cos(), -end_angle.sin());

    let a = p1 - center;
    let b = p4 - center;
    let q1 = a.length_sq();
    let q2 = q1 + a.dot(b);
    let k2 = (4.0 / 3.0) * ((2.0 * q1 * q2).sqrt() - q2) / (a.x * b.y - a.y * b.x);

    let p2 = Pos2::new(xc + a.x - k2 * a.y, yc + a.y + k2 * a.x);
    let p3 = Pos2::new(xc + b.x + k2 * b.y, yc + b.y - k2 * b.x);
```

[...]

PR #4424, opened 2024-04-27
Add Arc and Pie Drawing and Plotting Features

varphone:
## Summary

This PR introduces several new features related to drawing and plotting arcs and pies in the `egui` ecosystem.

## Changes

- `epaint`: Introduced `ArcPieShape`, a new shape that represents an arc or a pie slice with specified radius, angle, fill, and stroke properties.
- `egui`: Added `Painter::{arc, pie}` methods, which simplify the process of drawing arcs and pies.
- `egui_plot`: Introduced three new plot types: `ArcLine`, `Pie`, and `PieChart`. `ArcLine` renders a simple arc line, `Pie` renders a single pie slice, and `PieChart` renders a pie chart with data and labels.
- `egui_demo_lib`: Added demonstrations for `ArcPieShape`, `ArcLine`, `Pie`, and `PieChart`. These demos showcase how to use these new features in a variety of contexts.

## Impact

These changes enhance the capabilities of the `egui` ecosystem by providing users with more options for drawing and plotting.
The new demos also serve as valuable examples for users to learn from.

[screencast-egui_demo_app-2024-04-27.webm](https://github.com/emilk/egui/assets/1274171/f91e9600-aff5-4bba-9c79-f5d2596d1b0e)

PR #4426, opened 2024-04-27
Add easy way to draw circle arcs

murl-digital:
This PR is related to the discussion in #4188, however it doesn't resolve the issue. It does, however, add a way to draw arcs on a circle. A couple things which need discussion:

* AngleIter can, in it's current state, loop infinitely. I'm not sure what the best way to handle this is.
* Using this method, it can sometimes be difficult to get the arc to behave the way you want. For example, drawing a track for a dial requires you to rather unintuitively provide a start and end angle of 225 and -45.
* The documentation could use some better wording, but I'm not sure how best to communicate its quirks.

### DIFF
--- a/crates/egui/src/painter.rs
+++ b/crates/egui/src/painter.rs
@@ -3,11 +3,12 @@ use std::sync::Arc;
 use crate::{
     emath::{Align2, Pos2, Rangef, Rect, Vec2},
     layers::{LayerId, PaintList, ShapeIdx},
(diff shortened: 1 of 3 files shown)


YgorSouza:
Looks like there might be some conflict between this and  #4424.

murl-digital:
> Looks like there might be some conflict between this and #4424.

yep, closing mine in favor of this

PR #4429, opened 2024-04-28
Make sure plot size is positive

### LINKED ISSUE #4425: Context Menus demo crashes on negative width or height
**Describe the bug**

In the web demo, when the plot in the Context Menus demo has negative width or height, the app crashes due to invalid arguments for `f64::clamp`.

**To Reproduce**
1. Open the Context Menus demo
2. Set the width or height of the plot to a negative number by either dragging or manually inputting
3. The app panics with `panicked at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/num/f64.rs:1496:9: min > max, or either was NaN. min = 440.00000059604645, max = -440.00000059604645`

**Expected behavior**

The width or height should be set to 0.

**Desktop:**
 - OS:  Windows
 - Browser:  Google Chrome
 - Version 124.0.6367.91 (Official Build) (64-bit)

rustbasic:
* Closes #4425 

Fix: in Plot, Minimum values for screen protection.

### DIFF
--- a/crates/egui_demo_lib/src/demo/context_menu.rs
+++ b/crates/egui_demo_lib/src/demo/context_menu.rs
@@ -1,4 +1,4 @@
-use egui::Vec2b;
+use egui::{NumExt, Vec2b};
 
 #[derive(Clone, Debug, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
@@ -105,7 +105,10 @@ impl super::View for ContextMenus {
                             .speed(1.0)
                             .prefix("Height: "),
                     );
+                    self.width = self.width.at_least(1.0);
+                    self.height = self.height.at_least(1.0);
                     ui.end_row();
+
                     ui.checkbox(&mut self.show_axes[0], "x-Axis");
                     ui.checkbox(&mut self.show_axes[1], "y-Axis");
                     ui.end_row();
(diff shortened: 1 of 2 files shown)


crates/egui_demo_lib/src/demo/context_menu.rs
```diff
@@ -105,7 +105,10 @@ impl super::View for ContextMenus {
                             .speed(1.0)
                             .prefix("Height: "),
                     );
+                    self.width = self.width.at_least(1.0);
+                    self.height = self.height.at_least(1.0);
```
emilk, line 109: use `DragValue::clamp_range` instead

emilk (changes requested):

crates/egui_plot/src/lib.rs
```diff
@@ -768,6 +768,10 @@ impl Plot {
 
         // Determine position of widget.
         let pos = ui.available_rect_before_wrap().min;
+        // Minimum values for screen protection
+        min_size.x = min_size.x.at_least(64.0);
+        min_size.y = min_size.y.at_least(15.0);
```
emilk, line 773: I don't think we should do this. If the user wants a plot that is `16x32` pixels big, let them have it

rustbasic: > I don't think we should do this. If the user wants a plot that is `16x32` pixels big, let them have it

If it is greater than `1.0`, panic does not occur, so it was set to `1.0`.
If you want to edit it to a different size, please do so.

emilk (approved):

PR #4431, opened 2024-04-28
Add `Options::reduce_texture_memory` to free up RAM

varphone:
## Summary

This PR introduces a new configuration option `reduce_texture_memory` in `egui`.

## Changes

- Added `reduce_texture_memory` option in `egui`. When set to `true`, `egui` will discard the loaded image data after the texture is uploaded to the GPU, reducing memory usage. This is beneficial when handling a large number of images and retaining the image data is unnecessary, potentially saving substantial memory. However, this makes it impossible to serialize the loaded images or render on non-GPU systems. Default is `false`.

## Impact

This new configuration option gives users more control over their memory usage, especially when dealing with a large number or large resolution of images. It allows users to optimize their applications based on their specific needs and constraints.

### DIFF
--- a/crates/egui/src/load/texture_loader.rs
+++ b/crates/egui/src/load/texture_loader.rs
@@ -28,6 +28,17 @@ impl TextureLoader for DefaultTextureLoader {
                     let handle = ctx.load_texture(uri, image, texture_options);
                     let texture = SizedTexture::from_handle(&handle);
                     cache.insert((uri.into(), texture_options), handle);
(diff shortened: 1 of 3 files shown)


emilk (approved):
The feature makes sense. I'm not 100% sold on the name, but don't have a better suggestion either 🤷

emilk:
merge conflict

varphone:
> merge conflict

Merge conflicts have been fixed.

Issue #4434, opened 2024-04-29
egui_plot formatter, spacer etc. should not need to be 'static


Fabus1184:
**Is your feature request related to a problem? Please describe.**
Currently, all methods like `Plot::{x,y}_grid_spacer` or `Plot::{x,y}_axis_formatter` take an `impl Fn... + 'static` which is quite impractical mainly because these methods can easily require some access to `self` or other borrowed data.

**Describe the solution you'd like**
Introduce a lifetime for the `Plot` struct? Since a Plot is rarely stored in another struct, this shouldn't be too much of an issue.

**Describe alternatives you've considered**
Not using borrowed data or some sketchy shit with RefCells

**Additional context**

Basic example:

```rust
struct App {
    data: (),
}

impl App {
    pub fn draw(&self, ctx: &egui::Context) {
        egui::CentralPanel::default().show(ctx, |ui| {
            egui_plot::Plot::new("plot")
                .x_axis_formatter(|_, _, _| format!("{:?}", self.data))
                .show(ui, |_| {});
        });
    }
}
```

This does currently not work because `x_axis_formatter` requires a `'static` lifetime
```rust
pub fn x_axis_formatter(self, fmt: impl Fn(GridMark, usize, &RangeInclusive<f64>) -> String + 'static) -> Self
```

PR #4435, opened 2024-04-30
Introduce lifetime to `egui_plot::Plot` to replace `'static` fields

Fabus1184:
* Closes https://github.com/emilk/egui/issues/4434

Shouldn't break anything, because when all arguments have a 'static lifetime (required without this PR), the Plot is also 'static and can be used like before.


emilk (approved):
nice

emilk:
Needs fix for wasm. You can run `scripts/clippy_wasm.sh` to check locally

Fabus1184:
@emilk can you merge this?

Issue #4442, opened 2024-05-02
Window position creeps between executions on scaled monitors


avery-radmacher:
**Describe the bug**

On monitors with a scale other than 100%, persistent windows are not recreated at the location where they were closed. In the gif below, I am running a shell script to start the demo app right after I manually close it. You can see the creep down and to the right. This screen has a 150% scaling.
![egui_window_position_creep](https://github.com/emilk/egui/assets/45777186/b19267ba-4143-4f77-948c-583638596b46)

Once the window has moved off the monitor, the program can no longer be brought back onto the screen and is very difficult to interact with.

**To Reproduce**

1. You must have a monitor with a scaling other than 100%. `persistence` must be enabled (it is by default).
2. Run `egui_demo_app` (`cargo run --release`)
3. If the window is not already on the scaled monitor, drag it to the middle of the scaled monitor
  a. The further the window's top left corner is from the monitor's top left corner, the greater the effect noticed in step 6 will be. 
4. Close the program.
5. Run `egui_demo_app` again.
6. Notice that the window is in a different location.

**Expected behavior**

The window reappears in the same location where it was closed.

**Desktop (please complete the following information):**
 - OS: Windows 11 Pro
 - Browser: N/A
 - Version: `rust-version = "1.76"`, egui: `version = "0.27.2"`

PR #4443, opened 2024-05-02
Fix: Window position creeps between executions on scaled monitors

avery-radmacher:
* Closes <https://github.com/emilk/egui/issues/4442>
* Refactors active monitor detection so it can be called from multiple locations.

Compare this gif to the one on the issue report.
egui_window_position_no_creep

### Investigation notes

- [`WindowSettings.inner_position_pixels` and `WindowSettings.outer_position_pixels`](https://github.com/emilk/egui/blob/master/crates/egui-winit/src/window_settings.rs#L8-L12) are stored in physical/pixel coordinates.
- `ViewportBuilder::with_position` expects to be passed a position in _logical_ coordinates.
- Prior to this PR, the position was being passed from `WindowSettings` to `with_position` [without any scaling](https://github.com/emilk/egui/blob/master/crates/egui-winit/src/window_settings.rs#L61-L68). This was the root cause of the issue.
- The fix is to first convert the position to logical coordinates, respecting the scaling factor of the active monitor. This requires us to first determine the active monitor, so I factored out some of the logic in [`clamp_pos_to_monitor`](https://github.com/emilk/egui/blob/master/crates/egui-winit/src/window_settings.rs#L130) to find the active monitor.

### DIFF
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -39,7 +39,11 @@ pub fn viewport_builder<E>(
         }
         window_settings.clamp_position_to_monitors(egui_zoom_factor, event_loop);
 
-        viewport_builder = window_settings.initialize_viewport_builder(viewport_builder);
+        viewport_builder = window_settings.initialize_viewport_builder(
+            egui_zoom_factor,
+            event_loop,
+            viewport_builder,
+        );
         window_settings.inner_size_points()
     } else {
         if let Some(pos) = viewport_builder.position {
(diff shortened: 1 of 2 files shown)


emilk (approved):
Thanks for fixing this! Have you tested this with differing `egui_zoom_factor`s and monitor dpi scales too?

avery-radmacher:
I only tested on 100% and 150% monitors, and an `egui_zoom_factor` of 1.0 I will do some more testing this weekend!

avery-radmacher:
More testing is complete. This works on other zoom factors and monitor scales too!

emilk:
Thanks for the additional testing!

PR #4455, opened 2024-05-03
Fix stacked BarChart offsets

### LINKED ISSUE #4454: Stacked bars are offset by a wrong value
**Describe the bug**
I was working on a little app with a stacked bar chart, and came across this possible bug (black lines were added in Paint for illustrative purposes):
filtered

Each bar chart is based upon a `Vec<(f64, f64)>`, which contains pairs of `(value, frequency)`, all vectors have the same length, `value` isn't being used as the `Bar`'s argument intentionally.
```Rust
// declared elsewhere
frequencies: Vec<Vec<(f64, f64)>>,

// plot rendering
let mut charts = vec![];
for (frequencies, label) in self.frequencies.iter().zip(self.labels.iter()) {
    let bars = frequencies
        .iter()
        .enumerate()
        .filter(|(_, (_, f))| f > &0.0)
        .map(|(i, (v, f))| Bar::new(i as f64, *f).width(self.width).name(v))
        .collect();

    let mut chart = BarChart::new(bars).name(&label);
    if !self.vertical {
        chart = chart.horizontal();
    }

    if charts.len() > 0 {
        let other_charts: Vec<_> = charts.iter().collect();
        chart = chart.stack_on(&other_charts);
    }

    charts.push(chart);
}
```
Zero frequencies are filtered out, so the total number of bars will vary. The reason for filtering is that a `Bar` with zero height will be rendered anyway, and to me it looks a little confusing (the plot's aspect ratio is different, but all values are the same):
non-filtered
Here, the offset values are correct, since the number of bars in both charts is the same. 

**To Reproduce**
The easiest way to reproduce this behavior is to modify the demo app by commenting out [this line](https://github.com/emilk/egui/blob/ded8dbd45bc8616475c364878f6b8e2f7293d12c/crates/egui_demo_lib/src/demo/plot_demo.rs#L976):
demo-app

**Expected behavior**
I expect the modified demo app to look like this:
demo-app-fix

**Additional context**
I assume it's a bug, since I never used stacked bar charts before.

rctlmk:
* Closes emilk/egui_plot#23

### DIFF
--- a/crates/egui_plot/src/items/mod.rs
+++ b/crates/egui_plot/src/items/mod.rs
@@ -1654,19 +1654,31 @@ impl BarChart {
     /// Negative values are stacked below other negative values.
     #[inline]
     pub fn stack_on(mut self, others: &[&Self]) -> Self {
-        for (index, bar) in self.bars.iter_mut().enumerate() {
+        for bar in self.bars.iter_mut() {
             let new_base_offset = if bar.value.is_sign_positive() {
                 others
                     .iter()
-                    .filter_map(|other_chart| other_chart.bars.get(index).map(|bar| bar.upper()))
+                    .filter_map(|other_chart| {
+                        other_chart
+                            .bars
+                            .iter()
+                            .find(|other_bar| other_bar.argument == bar.argument)
+                            .map(|other_bar| other_bar.upper())
+                    })
                     .max_by_key(|value| value.ord())
             } else {
                 others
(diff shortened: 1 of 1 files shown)


rctlmk:
It should be noted that this solution doesn't scale well, and with many bars (more than 10000) it becomes really noticeable.

emilk:

See also:

PR #4457, opened 2024-05-04
Disable interaction for `ScrollArea` and `Plot` when UI is disabled

### LINKED ISSUE #4341: `ScrollArea` (and all scoll-able widgets) scroll-able even when not enabled
**Describe the bug**
I try to disable the whole main window when a modal window is visible on top but of the disabled widget on the main window, the `ScrollArea` widgets are still scroll-able.

This issues affects all the scroll-able widgets (`egui-extras::TableBody` for example) not only the `ScrollArea`.

In 0.26 the issue was impacting all the drag-able widgets also but this is fixed in 0.27 and still is in git master.

**To Reproduce**
```Rust
fn main() {
    let mut main_enabled = false;

    let _ = eframe::run_simple_native(
        "Issue test",
        move |ctx, _frame| {
                ui.checkbox(&mut main_enabled, "Enabled");
                ui.separator();
                ui.add_enabled_ui(main_enabled, |ui| {
                        for _ in 0..100 {
                            ui.label("some text");
                        }
                    });
                });
            });
        },
    );
}

```

**Expected behavior**
I expect the scroll-able widgets to not be able to be interacted with at all when disabled.

 - OS: Linux 6.6.25 LTS

varphone:
## Summary

This PR modifies `ScrollArea` and `Plot` to disable their interactions when the UI is disabled.

## Changes

- Interaction with `ScrollArea` in `egui` is disabled when the UI is disabled.
- Interaction with `Plot` in `egui_plot` is disabled when the UI is disabled.
- These changes ensure that `ScrollArea` and `Plot` behave consistently with the rest of the UI, preventing them from responding to user input when the UI is in a disabled state.

## Impact

This PR enhances the consistency of `egui`'s UI behavior by ensuring that all elements, including `ScrollArea` and `Plot`, respect the UI's disabled state. This prevents unexpected interactions when the UI is disabled.

Closes #4341

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -493,6 +493,7 @@ impl ScrollArea {
         } = self;
 
         let ctx = ui.ctx().clone();
+        let scrolling_enabled = scrolling_enabled && ui.is_enabled();
 
         let id_source = id_source.unwrap_or_else(|| Id::new("scroll_area"));
         let id = ui.make_persistent_id(id_source);
--- a/crates/egui_plot/src/lib.rs
+++ b/crates/egui_plot/src/lib.rs
@@ -766,6 +766,11 @@ impl Plot {
             sense,
         } = self;
 
+        // Disable interaction if ui is disabled.
+        let allow_zoom = allow_zoom.and(ui.is_enabled());
+        let allow_drag = allow_drag.and(ui.is_enabled());
+        let allow_scroll = allow_scroll.and(ui.is_enabled());
+
         // Determine position of widget.
         let pos = ui.available_rect_before_wrap().min;
         // Determine size of widget.


crates/egui/src/containers/scroll_area.rs
```diff
@@ -493,6 +493,7 @@ impl ScrollArea {
         } = self;
 
         let ctx = ui.ctx().clone();
+        let scrolling_enabled = scrolling_enabled && ui.is_enabled();
```
Mingun, line 496: If `disabled` means that you cannot change widget properties (like entering text in text fields), then scroll area should not respect this setting, because it does not have properties that reflects application data. Imagine a read-only table widget which you cannot scroll because is is non-editable. I think, that is not what you want.

varphone: In mainstream UI frameworks like Qt and MFC, disabling a parent widget results in all child widgets becoming non-interactive. This includes the scroll bars of the child widgets.

Instead of relying on the disabled state of the scroll area, a read-only table should be implemented as an inherent feature of the table itself. This approach is more direct and explicit, and it avoids potential issues that could arise from disabling the scroll area.

emilk (approved):

Issue #4081, opened 2024-02-21
Universal clipboard shortcuts (`Ctrl`+`C`/`V`/`X`) don't work with non-Latin keyboard layouts


TicClick:
**Describe the bug**

at least on Windows 10 (and I think it happens on MacOS as well) these shortcuts only work in layouts with Latin characters, such as `English (United States)`. this is because the code checking for special shortcuts operates on logical, not physical keys:

- https://github.com/emilk/egui/blame/b8048572e8cc47ef9410b3516456da2a320fcdd2/crates/egui-winit/src/lib.rs#L720
- ```
  2024-02-21 00:54:51.1492981 [TRACE] (1) egui_winit: logical Character("с") -> None,  physical Code(KeyC) -> Some(C)
  ```

(this is present in at least egui v0.25.0, and it doesn't look like it's fixed in the current `master` branch, b8048572e8cc47ef9410b3516456da2a320fcdd2)

**To Reproduce**

try copying any text from a text field while having a keyboard layout without a Latin character on the `C` hardware key

PR #4461, opened 2024-05-05
egui-winit: emit physical key presses when a non-Latin layout is active

TicClick:
resolves https://github.com/emilk/egui/issues/4081 (see discussion starting from https://github.com/emilk/egui/issues/3653#issuecomment-1962740175 for extra context)

this partly restores event-emitting behaviour to the state before #3649, when shortcuts such as `Ctrl` + `C` used to work regardless of the active layout. the difference is that physical keys are only used in case of the logical ones' absence now among the named keys.

while originally I have only limited this to clipboard shortcuts (Ctrl+C/V/X), honestly it felt like a half-assed solution. as a result, I decided to expand this behaviour to all key events to stick to the original logic, in case there are other workflows and hotkeys people rely on or expect to work out of the box. let me know if this is an issue.

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -746,15 +746,19 @@ impl State {
             physical_key
         );
 
-        if let Some(logical_key) = logical_key {
+        // "Logical OR physical key" is a fallback mechanism for keyboard layouts without Latin characters: it lets them
+        // emit events as if the corresponding keys from the Latin layout were pressed. In this case, clipboard shortcuts
+        // are mapped to the physical keys that normally contain C, X, V, etc.
+        // See also: https://github.com/emilk/egui/issues/3653
+        if let Some(active_key) = logical_key.or(physical_key) {
             if pressed {
-                if is_cut_command(self.egui_input.modifiers, logical_key) {
+                if is_cut_command(self.egui_input.modifiers, active_key) {
                     self.egui_input.events.push(egui::Event::Cut);
                     return;
-                } else if is_copy_command(self.egui_input.modifiers, logical_key) {
+                } else if is_copy_command(self.egui_input.modifiers, active_key) {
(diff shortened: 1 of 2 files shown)


crates/egui-winit/src/lib.rs
```diff
@@ -746,15 +746,19 @@ impl State {
             physical_key
         );
 
-        if let Some(logical_key) = logical_key {
+        // "Logical OR physical key" is a fallback mechanism for keyboard layouts without Latin characters: it lets them
+        // emit events as if the corresponding keys from the Latin layout were pressed. In this case, clipboard shortcuts
+        // are mapped to the physical keys that normally contain C, X, V, etc.
+        // See also: https://github.com/emilk/egui/issues/3653
```
emilk, line 752: I assume you have tested this?

I'm a bit surprised that `logical_key` is a `None` in these cases.

crates/egui-winit/src/lib.rs
```diff
@@ -766,7 +770,7 @@ impl State {
             }
 
             self.egui_input.events.push(egui::Event::Key {
-                key: logical_key,
+                key: active_key,
```
emilk, line 773: We should document that `Event::Key::key` is either the logical key (if known), else the physical key.

emilk (commented):
Thanks!

crates/egui-winit/src/lib.rs
TicClick: yup — I myself have a Russian keyboard layout which prompted #4081 in the first place unfortunately

crates/egui-winit/src/lib.rs
TicClick: thanks for highlighting this -- I have added a clarification. if it's not to your liking, feel free to reword

emilk (approved):
Awesome; thank you!

PR #4495, opened 2024-05-14
Move dependencies to workspace

oscargus:
Inspired by:
https://github.com/emilk/egui/blob/44d65f41ac297847ceedc452c6a52299fe17f4a9/Cargo.toml#L65

I took the liberty of removing that comment since I *think* that I got all "relevant" ones (showing up more than once, sort of).


emilk (approved):
Great - thank you!

emilk:
I'd like to eventually move all dependencies to be workspace dependencies, just because it lets me get an easy overview over all the dependencies (which is extra useful when using https://marketplace.visualstudio.com/items?itemName=serayuzgur.crates)

PR #4497, opened 2024-05-14
Introduce dithering to reduce banding

### LINKED ISSUE #4493: Dithering in Fragment Shader
Egui currently doesn't apply any dithering to it's output. That's hardly an issue with the flat default look and feel. However, when drawing gradients or images (which have been resampled) banding becomes a problem.

I think ideally egui would perform dithering in the fragment shader. Given the trivial nature of the shader performing a few more operations on the ALU for some simple dithering should not have a relevant impact on rendering performance. The one exception is potentially when software rasterization.

I'm currently just drawing a noisy texture on top of gradients to mask the problem but it requires a lot more noise to be added to the gradient to make it look smooth (ironically) than doing it in a fragment shader.

I think there is also an argument that this is out of scope for egui. Other than shadows most widgets seem to use solid colors. For the shadows the fall off is quick enough that dithering isn't necessary. Even when rendering images this is likely going to be an edge case and only an issue when rendering images with smooth gradients that either aren't dithered or have been resampled to the extent that the dithering in the image is ineffective.

**Additional context**
I'd be happy to create a PR if dithering is considered to be in scope for egui.

jwagner:
This PR introduces dithering in the egui_glow and egui_wgpu backends to reduce banding artifacts.

It's based on the approach mentioned in #4493 with the small difference that the amount of noise is scaled down slightly to avoid dithering colors that can be represented exactly. This keeps flat surfaces clean.

Exaggerated dithering to show what is happening:
Screenshot from 2024-05-14 19-09-48

Subtle dithering as commited.
Screenshot from 2024-05-14 19-13-40

Closes #4493

### DIFF
--- a/crates/egui-wgpu/src/egui.wgsl
+++ b/crates/egui-wgpu/src/egui.wgsl
@@ -7,12 +7,35 @@ struct VertexOutput {
 };
 
 struct Locals {
-    screen_size: vec2<f32>,
+    screen_size: vec2<f32>, // in points
+    pixels_per_point: f32,
     // Uniform buffers need to be at least 16 bytes in WebGL.
     // See https://github.com/gfx-rs/wgpu/issues/2072
-    _padding: vec2<u32>,
+    _padding: u32,
 };
 @group(0) @binding(0) var<uniform> r_locals: Locals;
+
+
+// -----------------------------------------------
+// Adapted from
+// https://www.shadertoy.com/view/llVGzG
+// Originally presented in:
+// Jimenez 2014, "Next Generation Post-Processing in Call of Duty"
+//
+// A good overview can be found in
+// https://blog.demofox.org/2022/01/01/interleaved-gradient-noise-a-different-kind-of-low-discrepancy-sequence/
+// via https://github.com/rerun-io/rerun/
+fn interleaved_gradient_noise(n: vec2<f32>) -> f32 {
+    let f = 0.06711056 * n.x + 0.00583715 * n.y;
+    return fract(52.9829189 * fract(f));
+}
+
+fn dither_interleaved(rgb: vec3<f32>, levels: f32, frag_coord: vec4<f32>) -> vec3<f32> {
+    var noise = interleaved_gradient_noise(frag_coord.xy);
+    // scale down the noise slightly to ensure flat colors aren't getting dithered
+    noise = (noise - 0.5) * 0.95;
+    return rgb + noise / (levels - 1.0);
+}
 
 // 0-1 linear  from  0-1 sRGB gamma
 fn linear_from_gamma_rgb(srgb: vec3<f32>) -> vec3<f32> {
@@ -78,7 +101,11 @@ fn fs_main_linear_framebuffer(in: VertexOutput) -> @location(0) vec4<f32> {
     let tex_linear = textureSample(r_tex_color, r_tex_sampler, in.tex_coord);
     let tex_gamma = gamma_from_linear_rgba(tex_linear);
     let out_color_gamma = in.color * tex_gamma;
-    return vec4<f32>(linear_from_gamma_rgb(out_color_gamma.rgb), out_color_gamma.a);
+    let out_color_linear = linear_from_gamma_rgb(out_color_gamma.rgb);
+    // Dither the float color down to eight bits to reduce banding.
+    // This step is optional for egui backends.
+    let out_color_dithered = dither_interleaved(out_color_linear, 256.0, in.position);
+    return vec4<f32>(out_color_dithered, out_color_gamma.a);
 }
 
 @fragment
@@ -87,5 +114,6 @@ fn fs_main_gamma_framebuffer(in: VertexOutput) -> @location(0) vec4<f32> {
     let tex_linear = textureSample(r_tex_color, r_tex_sampler, in.tex_coord);
     let tex_gamma = gamma_from_linear_rgba(tex_linear);
     let out_color_gamma = in.color * tex_gamma;
-    return out_color_gamma;
+    let out_color_dithered = vec4<f32>(dither_interleaved(out_color_gamma.rgb, 256.0, in.position * r_locals.pixels_per_point), out_color_gamma.a);
+    return out_color_dithered;
 }
--- a/crates/egui-wgpu/src/renderer.rs
+++ b/crates/egui-wgpu/src/renderer.rs
@@ -132,9 +132,10 @@ impl ScreenDescriptor {
 #[repr(C)]
 struct UniformBuffer {
     screen_size_in_points: [f32; 2],
+    pixels_per_point: f32,
     // Uniform buffers need to be at least 16 bytes in WebGL.
     // See https://github.com/gfx-rs/wgpu/issues/2072
-    _padding: [u32; 2],
+    _padding: u32,
 }
 
 impl PartialEq for UniformBuffer {
@@ -200,6 +201,7 @@ impl Renderer {
             label: Some("egui_uniform_buffer"),
             contents: bytemuck::cast_slice(&[UniformBuffer {
                 screen_size_in_points: [0.0, 0.0],
+                pixels_per_point: 0.,
                 _padding: Default::default(),
             }]),
             usage: wgpu::BufferUsages::UNIFORM | wgpu::BufferUsages::COPY_DST,
@@ -211,7 +213,7 @@ impl Renderer {
                 label: Some("egui_uniform_bind_group_layout"),
                 entries: &[wgpu::BindGroupLayoutEntry {
                     binding: 0,
-                    visibility: wgpu::ShaderStages::VERTEX,
+                    visibility: wgpu::ShaderStages::VERTEX | wgpu::ShaderStages::FRAGMENT,
                     ty: wgpu::BindingType::Buffer {
                         has_dynamic_offset: false,
                         min_binding_size: NonZeroU64::new(std::mem::size_of::<UniformBuffer>() as _),
@@ -363,7 +365,8 @@ impl Renderer {
             // Buffers on wgpu are zero initialized, so this is indeed its current state!
             previous_uniform_buffer_content: UniformBuffer {
                 screen_size_in_points: [0.0, 0.0],
-                _padding: [0, 0],
+                pixels_per_point: 0.,
+                _padding: 0,
             },
             uniform_bind_group,
             texture_bind_group_layout,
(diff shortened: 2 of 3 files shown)


murl-digital (approved):
LGTM, I tested this with the baseview windowing backend, with is glow based. Taking a screenshot of a gradient and zooming all the way in shows me the dithering's there:

image

crates/egui-wgpu/src/egui.wgsl
```diff
@@ -78,7 +101,11 @@ fn fs_main_linear_framebuffer(in: VertexOutput) -> @location(0) vec4<f32> {
     let tex_linear = textureSample(r_tex_color, r_tex_sampler, in.tex_coord);
     let tex_gamma = gamma_from_linear_rgba(tex_linear);
     let out_color_gamma = in.color * tex_gamma;
-    return vec4<f32>(linear_from_gamma_rgb(out_color_gamma.rgb), out_color_gamma.a);
+    let out_color_linear = linear_from_gamma_rgb(out_color_gamma.rgb);
+    // Dither the float color down to eight bits to reduce banding.
+    // This step is optional for egui backends.
+    let out_color_dithered = dither_interleaved(out_color_linear, 256.0, in.position);
```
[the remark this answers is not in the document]
Wumpf: I agree, this should be done on the gamma color. It's important to know the context of when this fragment shader is used: https://github.com/emilk/egui/blob/ff7a3832b68a8ecccf531511d1a17d92821b5430/crates/egui-wgpu/src/renderer.rs#L319 it is enabled in the case of using an srgb converting target. That means whatever float value we emit here will be _upon write_ converted to an srgb value. I.e. as-if the output of this method was wrapped in another call of `gamma_from_linear_rgba`. 

Therefore, we have to make sure that all the meaningful value changes happen on the gamma color.

Note that it would be an entirely different story if we're outputing _anything_ that isn't ending up on a `wgpu::Surface` that is targeting a srgb interpreting monitor/window. Meaning that all of this breaks down for HDR.

crates/egui-wgpu/src/egui.wgsl
```diff
@@ -87,5 +114,6 @@ fn fs_main_gamma_framebuffer(in: VertexOutput) -> @location(0) vec4<f32> {
     let tex_linear = textureSample(r_tex_color, r_tex_sampler, in.tex_coord);
     let tex_gamma = gamma_from_linear_rgba(tex_linear);
     let out_color_gamma = in.color * tex_gamma;
-    return out_color_gamma;
+    let out_color_dithered = vec4<f32>(dither_interleaved(out_color_gamma.rgb, 256.0, in.position * r_locals.pixels_per_point), out_color_gamma.a);
```
Wumpf, line 117: I don't get what `* r_locals.pixels_per_point` is about here. `in.position` is `@builtin(position)` which makes this raw pixel coordinates ranging from 0 to <target size>, see https://www.w3.org/TR/WGSL/#built-in-values-position. So scaling it with eguis points->pixel factor doesn't make any sense since we're starting with pixels anyways.

you also didn't apply this to either glsl or the `fs_main_linear_framebuffer` version. What's the intention here?

crates/egui-wgpu/src/renderer.rs
```diff
@@ -132,9 +132,10 @@ impl ScreenDescriptor {
 #[repr(C)]
 struct UniformBuffer {
     screen_size_in_points: [f32; 2],
+    pixels_per_point: f32,
```
Wumpf, line 135: as per other comment, I don't see why this is needed

crates/egui-wgpu/src/egui.wgsl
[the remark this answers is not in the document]
jwagner: I assumed that the output buffer was linear in that case dithering the linear color is what is wanted. If it's converted back to srgb before storage, it isn't optimal. I'll need to double check that.

Wumpf (changes requested):
great addition, however I think there should be an easy way to disable dithering:
* this is srgb dithering and doesn't make sense if someone wants to use an HDR target (e.g. WebGPU can also support P3 instead of srgb)
* there may be corner cases where a user doesn't actually want to dither everything drawn with egui indiscriminately as the added noise may destroy otherwise sharp corners

crates/egui-wgpu/src/egui.wgsl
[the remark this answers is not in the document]
jwagner: Just saw the comment be @Wumpf bellow, so you are right, this is off.

jwagner:
@Wumpf @emilk for disabling dithering, would a feature on the crate do the job or should this be a runtime option?

emilk:
I think a runtime feature flag is likely superior, as it is simpler, and lets the same user use `egui_wgpu` with different options depending on their render target, which may depend on the adapter capabilities.

Checking a bool in the shader should be fine

jwagner:
I'll do a revision of the PR adding an option and fixing the coordinates in wgpu. I'm not certain where to add it yet. Adding it to native / web options in eframe, and then an implementation specific option in the specific backend (WgpuConfiguration for wgpu , EguiGlow::new for glow) seems about right. If you'd like it to be added another way, just let me know.

jwagner:
I got a first draft of configurable dithering to work but I'm running out of time for today. I changed to state of the PR back to draft for now.

crates/egui-wgpu/src/renderer.rs
jwagner: Removed it now/turned it into the uniform controlling dithering.

jwagner:
I updated and cleaned up the code and will mark the PR as 'Ready for review' again.

#### Testing performed
I tested the demo app using both wgpu and glow, with dithering enabled and disabled, on two different systems (Linux and Mac), as well as on the web using Chrome and Firefox. The dithering works as expected in all cases. While the results from wgpu and glow are not identical, they are very close.

There were some minute differences with dithering disabled as well. So I guess producing identical output between different backends isn't a requirement (and likely not a reasonable goal across systems anyways).

In addition to the demo app, I tested the proposed changes with one of my own applications.

#### Potential Optimizations
egui_glow::Painter::new, RenderState::create and Renderer::new, now all take quite a few arguments. If desired I can change them to take a struct (with a default) instead.

The implementation of the condition is different for wgpu and glow.

The glow implementation makes use of the glsl preprocessor using a define and if, similar to the existing code for SRGB_TEXTURES. This should result in zero shading cost when dithering is disabled.

The wgpu implementation uses a uniform instead. At least as far as I know wgsl doesn't have a preprocessor and I wasn't to keen to introduce some code to process the shader source. Instead it's using a uniform and a condition in the shader as suggested by @emilk. I'm not sure about the optimizations applied in wgpu, let alone it's various backends but this is likely very cheap (coherent branching) but not zero cost in either case.

Lastly it would be possible to make the amount if dithering configurable, for instance for using egui in an application with a bit depth lower than 8 bit. I don't think this is a very common use case though.

Wumpf:
> At least as far as I know wgsl doesn't have a preprocessor and I wasn't to keen to introduce some code to process the shader source.

there's no preprocessor, but this is a great usecase for a pipeline-overridable constant! :) There's an example in the [wgpu 0.20.0 release notes](https://github.com/gfx-rs/wgpu/releases/tag/v0.20.0).
Unfortunately, they got only added in wgpu 0.20 and egui had to back out of 0.20 for the moment. So would be good to leave a todo note towards (and subsequent patch in) https://github.com/emilk/egui/pull/4560
(not a big deal anyways, this kind of static condition is extremely cheap)

jwagner:
> there's no preprocessor, but this is a great usecase for a pipeline-overridable constant! :) There's an example in the [wgpu 0.20.0 release notes](https://github.com/gfx-rs/wgpu/releases/tag/v0.20.0). Unfortunately, they got only added in wgpu 0.20 and egui had to back out of 0.20 for the moment. So would be good to leave a todo note towards (and subsequent patch in) #4560 (not a big deal anyways, this kind of static condition is extremely cheap)

That sounds like a great solution. I'll need to do a bit more digging to figure out what it does in practice. If it's ok, I'll do a follow up on this PR once it is merged.

crates/eframe/src/epi.rs
```diff
@@ -369,6 +369,10 @@ pub struct NativeOptions {
     /// The folder where `eframe` will store the app state. If not set, eframe will get the paths
     /// from [directories_next].
     pub persistence_path: Option<std::path::PathBuf>,
+
+    /// Controls whether colors should be dithered to minimize banding.
+    /// Defaults to true.
+    pub dithering: bool,
```
Wumpf, line 375: I have a bad feeling about making this opt-out rather than opt-in. In theory textures that are already in srgb8 should be fine since the dithering shouldn't affect those (see the `scale down the noise slightly to ensure flat colors aren't getting dithered` bit in the shader(s)). But let's stick with it.

In any case this is one of the many things that should be clarified here. Let's try:

```suggestion
    /// Controls whether to apply dithering to minimize banding artifacts.
    /// 
    /// Dithering assumes an sRGB output and thus will apply noise to any input value that lies between
    /// two 8bit values after applying the sRGB OETF function, i.e. if it's not a whole 8bit value in "gamma space".
    /// This means that only inputs from texture interpolation and vertex colors should be affected in practice.
    ///
    /// Defaults to true.
    pub dithering: bool,
```

I thiiiink this is accurate without being two confusing. Counter-feedback welcome.

crates/eframe/src/epi.rs
```diff
@@ -466,6 +472,10 @@ pub struct WebOptions {
     /// Configures wgpu instance/device/adapter/surface creation and renderloop.
     #[cfg(feature = "wgpu")]
     pub wgpu_options: egui_wgpu::WgpuConfiguration,
+
+    /// Controls whether colors should be dithered to minimize banding.
+    /// Defaults to true.
+    pub dithering: bool,
```
Wumpf, line 478: as above

Wumpf (changes requested):
lgtm except docs!

jwagner:
Sorry for the delayed response. Life happened.

I rebased the branch and applied the suggestions. I'm not sure if the more detailed descriptions should maybe go into the lower level crates (egui_glow, egui-wgpu) rather than eframe since users of those crates are more likely to care about the specifics. 

Another consideration is api compatibility. I guess we don't want to promise that dithering true will always dither to 8 bit. So maybe `_Currently_ dithering assumes an sRGB output and thus will apply noise to any input value that lies between` would be a slightly softer comment. 

Both are minor concerns though so I'd be happy to get the change merged as is.

Thank you for your patience @Wumpf @emilk :)

Wumpf (approved):
lgtm!

Still torn about having it opt-out instead of opt-in, but I wrote earlier to leave it be, so let's just do that (:

now that wgpu 0.20 is in we could use pipeline constants, but that constant buffer value won't hurt anyone 🤷

Wumpf:
Re api compatibility: I think for the renderers to handle hdr output correctly some other stuff needs to happen so not too concerned, whoever looks into that can also do something about that. Worst case they have to break the variable name to make it more obvious

PR #4498, opened 2024-05-15
Fix: Use default features in Image Crate

### LINKED ISSUE #4489: Gif Support
Gif Support

rustbasic:
* Closes #4489 
* Related #4495

Fix: Use default features in Image Crate

Because only `.png` is available after update #4495.

Required to use JPEG, etc.


emilk:
No - eframe only need `png` support, and should not be paying for the compilation of ten other image formats

PR #4505, opened 2024-05-17
Import directly from correct crates rather than the re-exports

oscargus:
Slightly related to #4504 and by noting that there are quite a lot of *-imports in egui. So I decided to get rid of the *-imports which then naturally opens the question: where should things be imported from? As an example, Pos2 is defined in emath, but re-exported from egui, epaint, egui::emath, egui::epaint, etc...

Not claiming to be 100% correct, but this should at least be a step in the right direction when it comes to imports. If one would like to go in that direction...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1340,8 +1340,12 @@ dependencies = [
 name = "egui_plot"
 version = "0.27.2"
 dependencies = [
+ "ahash",
  "document-features",
+ "ecolor",
  "egui",
+ "emath",
+ "epaint",
  "serde",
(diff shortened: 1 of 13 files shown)


lucasmerlin:
I'll close this since egui_plot now lives outside of the egui repo. Feel free to reopen there, but I'm not sure what the reasoning for not using the reexports would be? I think it's nice and convenient, since it means you generally only have to add the `eframe` dependency in order to build a egui app, and everything else you might need (egui, emath, epaint etc...) comes with that.

PR #4509, opened 2024-05-19
Handle the `IME` event first

rustbasic:
Handle the `IME` event first 
There is a need to handle the Ime event first.

Fix Issues : When you press `BackSpace`, the entire text is erased, and when you press another letter, it appears again.
Fix Issues : When you press `Left`, the character being entered will be copied once more.
Fix Issues : When you press `Enter`, `Enter` with `repeat` set is entered and `Enter` is entered twice.
Fix Issues : When you press a key in `IME` mode, `repeat` is often set.

Fix Issues : Since you may be selecting something in the IME, this also disables the `Arrow` keys.

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -1050,18 +1050,18 @@ fn ime_enabled_filter_events(events: Vec<Event>) -> Vec<Event> {
 }
 
 fn ime_front_events(events: Vec<Event>) -> Vec<Event> {
-    let mut ime_first_events: Vec<Event> = Vec::new();
+    let mut ime_front_events: Vec<Event> = Vec::new();
     let mut other_events: Vec<Event> = Vec::new();
 
     for event in events {
         match event {
-            Event::Ime(_) => ime_first_events.push(event),
+            Event::Ime(_) => ime_front_events.push(event),
             _ => other_events.push(event),
         }
     }
 
-    ime_first_events.extend(other_events);
-    ime_first_events
+    ime_front_events.extend(other_events);
+    ime_front_events
 }
 
 // ----------------------------------------------------------------------------
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -1045,19 +1045,12 @@ fn ime_enabled_filter_events(events: &mut Vec<Event>) {
 }
 
 fn ime_front_events(events: &mut Vec<Event>) {
-    let mut ime_front_events: Vec<Event> = Vec::new();
-    let mut other_events: Vec<Event> = Vec::new();
-
-    for event in events.clone() {
-        match event {
-            Event::Ime(_) => ime_front_events.push(event),
-            _ => other_events.push(event),
-        }
-    }
-
-    events.clear();
-    events.extend(ime_front_events);
-    events.extend(other_events);
+    events.sort_by(|a, b| match (a, b) {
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -1025,6 +1032,40 @@ fn events(
...
+    for event in events {
+        match event {
+            Event::Ime(_) => ime_front_events.push(event),
+            _ => other_events.push(event),
+        }
+    }
+
+    ime_front_events.extend(other_events);
+    ime_front_events
+}
```
emilk, line 1065: this could be a `.sort(…)`

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -1025,6 +1032,40 @@ fn events(
...
+            Event::Key { repeat: true, .. }
+            | Event::Key {
+                key: Key::Backspace,
+                ..
+            } => {}
+            _ => filter_events.push(event),
+        }
+    }
+
+    filter_events
```
emilk, line 1049: This could be a `.filter(…).collect(…)` or a `.retain(…)`

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -1025,6 +1032,40 @@ fn events(
...
+fn ime_enabled_filter_events(events: Vec<Event>) -> Vec<Event> {
+    let mut filter_events: Vec<Event> = Vec::new();
+
+    for event in events {
+        match event {
+            Event::Key { repeat: true, .. }
+            | Event::Key {
+                key: Key::Backspace,
+                ..
+            } => {}
```
emilk, line 1044: Explain why in a comment

emilk (changes requested):

crates/egui/src/widgets/text_edit/builder.rs
rustbasic: > this could be a `.sort(…)`

Since there shouldn't be any change in order, it would be better to simply process it without using `sort()`. 
Some modifications have been made.

crates/egui/src/widgets/text_edit/builder.rs
rustbasic: > This could be a `.filter(…).collect(…)` or a `.retain(…)`

It has been modified.

crates/egui/src/widgets/text_edit/builder.rs
emilk: no, explain it IN THE CODE. It is the code that will last.

rustbasic: > no, explain it IN THE CODE. It is the code that will last.

It has been modified.

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -826,7 +826,13 @@ fn events(
 
     let mut any_change = false;
 
-    let events = ui.input(|i| i.filtered_events(&event_filter));
+    let mut events = ui.input(|i| i.filtered_events(&event_filter));
+
+    if state.ime_enabled {
+        ime_enabled_filter_events(&mut events);
+        ime_front_events(&mut events);
```
emilk, line 833:
```suggestion
        
        // Process IME events first:
        events.sort_by_key(|e| !matches!(e, Event::Ime(_)));
```

rustbasic: It has been modified.

PR #4517, opened 2024-05-21
TextEdit hint text styling

zaaarf:
Simply adds a `hint_text_font` (any suggestions for a better name? hint_text_style seemed misleading as it doesn't only accept `TextStyle`) method to the TextEdit builder that allows to set the styling (specifically, a `FontSelection`) for the hint text.

My personal use-case for this was having body-sized hint text in a heading-sized TextEdit, but I'm sure there's more out there.

The PR shouldn't break compatibility, as it's stored as an `Option` that defaults to the `font_id` (which was the only behaviour prior to this) when empty.

It looked too trivial to add something to the actual demo, but I have it locally, so let me know if I should commit that.

Ran the check script locally, had no complaints.

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -662,9 +662,19 @@ impl<'t> TextEdit<'t> {
             if text.as_str().is_empty() && !hint_text.is_empty() {
                 let hint_text_color = ui.visuals().weak_text_color();
                 let galley = if multiline {
-                    hint_text.into_galley(ui, Some(true), desired_inner_size.x, hint_text_font.unwrap_or(font_id.into()))
+                    hint_text.into_galley(
+                        ui,
+                        Some(true),
+                        desired_inner_size.x,
+                        hint_text_font.unwrap_or(font_id.into()),
+                    )
                 } else {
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -189,6 +191,13 @@ impl<'t> TextEdit<'t> {
         self
     }
 
+    /// Forces a specific style for the hint text.
```
emilk, line 194:
```suggestion
    /// Set a specific style for the hint text.
```

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -652,9 +662,19 @@ impl<'t> TextEdit<'t> {
             if text.as_str().is_empty() && !hint_text.is_empty() {
                 let hint_text_color = ui.visuals().weak_text_color();
                 let galley = if multiline {
```
emilk, line 664:
```suggestion
                let hint_text_font_id = hint_text_font.unwrap_or(font_id.into());
                let galley = if multiline {
```

to reduce code duplication a bit

emilk (approved):

PR #4520, opened 2024-05-21
Plot now respects the `interact_radius` set in the UI's style

### LINKED ISSUE #4519: Configurable plot cursor "snap" radius
First, apologies if this has been asked already; I looked but didn't managed to identify an existing issue to track this need.

When viewing the plot, the cursor often jumps a considerable distance to snap to the nearest available point.  Being able to modify this distance (including disabling it altogether, perhaps by zeroing it out) would be helpful for some scenarios where hovering near a point is important.

Currently, the radius appears to be [hard-coded](https://github.com/emilk/egui/blob/8321f64f6e68c5b8a433517e535e22087573f8a4/crates/egui_plot/src/lib.rs#L1669); I'd request a property at the struct level to override.

YgorSouza:
* Closes #4519

### DIFF
--- a/crates/egui_plot/src/lib.rs
+++ b/crates/egui_plot/src/lib.rs
@@ -170,6 +170,7 @@ pub struct Plot<'a> {
 
     show_x: bool,
     show_y: bool,
+    interact_distance: f32,
     label_formatter: LabelFormatter<'a>,
     coordinates_formatter: Option<(Corner, CoordinatesFormatter<'a>)>,
     x_axes: Vec<AxisHints<'a>>, // default x axes
@@ -217,6 +218,7 @@ impl<'a> Plot<'a> {
 
             show_x: true,
             show_y: true,
+            interact_distance: 16.0,
             label_formatter: None,
             coordinates_formatter: None,
             x_axes: vec![AxisHints::new(Axis::X)],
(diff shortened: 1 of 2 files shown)


crates/egui_plot/src/lib.rs
```diff
@@ -1666,7 +1682,7 @@ impl<'a> PreparedPlot<'a> {
             return (Vec::new(), None);
         }
 
-        let interact_radius_sq = (16.0_f32).powi(2);
+        let interact_radius_sq = self.interact_distance.powi(2);
```
emilk, line 1685: It seems to me we should instead just use [`ui.style().interact.interact_radius`](https://docs.rs/egui/latest/egui/style/struct.Interaction.html#structfield.interact_radius) here. One global setting, instead of having to set it on each plot

crates/egui_plot/src/lib.rs
```diff
@@ -1666,7 +1666,7 @@ impl<'a> PreparedPlot<'a> {
             return (Vec::new(), None);
         }
 
-        let interact_radius_sq = (16.0_f32).powi(2);
+        let interact_radius_sq = ui.style().interaction.interact_radius.powi(2);
```
emilk, line 1669: I think this PR works best if you revert all changes except this one

emilk (approved):

emilk:
Thanks!

Issue #4510, opened 2024-05-19
`include_image!` macro should take in an `expr` not a `literal`


ThomasAlban:
The `include_image!` macro, unlike `include_bytes!`, currently takes in a `literal` rather than an `expr` for the file path. This is annoying because it limits what you can do when using the macro inside other macros.

PR #4521, opened 2024-05-21
`include_image!` now accepts expressions

YgorSouza:
Closes https://github.com/emilk/egui/issues/4510

### DIFF
--- a/crates/egui/src/lib.rs
+++ b/crates/egui/src/lib.rs
@@ -507,7 +507,7 @@ pub fn warn_if_debug_build(ui: &mut crate::Ui) {
 /// ```
 #[macro_export]
 macro_rules! include_image {
-    ($path: literal) => {
+    ($path:expr $(,)?) => {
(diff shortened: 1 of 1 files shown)


crates/egui/src/lib.rs
```diff
@@ -507,7 +507,7 @@ pub fn warn_if_debug_build(ui: &mut crate::Ui) {
 /// ```
 #[macro_export]
 macro_rules! include_image {
-    ($path: literal) => {
+    ($path:expr $(,)?) => {
```
emilk, line 510: why the `$(,)?` though?

YgorSouza: I just copied it from [include_bytes](https://doc.rust-lang.org/src/core/macros/mod.rs.html#1393) to be consistent. I'm not sure what it is for.

emilk: 🤷

Issue #3157, opened 2023-07-12
Web demo loses mouse cursor position when dragging off the window


pfgithub:
**Describe the bug**
While dragging in the web demo if the mouse goes off the window egui stops receiving mouse events

**To Reproduce**
1. Go to https://www.egui.rs/
2. Click '3d painting'
3. Click and drag the triangle and move your mouse off the window
4. The triangle stops dragging when the mouse leaves the window

**Expected behavior**
The triangle should keep dragging even with the mouse off the window

**Screenshots**

https://github.com/emilk/egui/assets/6010774/591c2354-efff-49e4-ab01-88dd05a9f11f

The triangle stops rotating when the mouse goes off the screen, even though the mouse is held down

**Desktop (please complete the following information):**
All

**Smartphone (please complete the following information):**
All

**Additional context**
Event listeners registered on document.body will continue to give mousemove/touchmove events even with the cursor off the screen, but eframe only registers event listeners on the canvas element.

https://github.com/emilk/egui/blob/083d61fccdfc9d967a12be09b18d4546e2313580/crates/eframe/src/web/events.rs#L256-L315

This registers event listeners on the canvas element, but after mousedown/touchdown, event listeners need to be registered on the document element in order to continue receiving mousemove events as the mouse goes off the screen.

emilk:
I tried registering `mousemove` on `document.body`, but the events still stop when dragging outside the canvas

pfgithub:
Sorry, I said the wrong thing. It should be on `document` (or `document.documentElement`) not `document.body`. Tetsing in the dev console, this continues to send events while dragging the mouse if it goes off the page (but not if you didn't start the drag on the page)

```js
document.addEventListener("mousemove", e => console.log(e.clientX, e.clientY));
```

emilk:
Nice, that works! I just have to figure out what to do with `mouseleave/Event::PointerGone` now

PR #4522, opened 2024-05-21
Fix: still track mouse when dragging outside web canvas

emilk:
* Closes https://github.com/emilk/egui/issues/3157

If the mouse leaves the canvas when dragging a slider, the slider will still move.

---

To support this, I had to revert https://github.com/emilk/egui/pull/4419
Despite that, I fail to reproduce the two issues it claimed to solve:

* https://github.com/emilk/egui/issues/4406 may have been solved in another way by this PR
* https://github.com/emilk/egui/issues/4418 I cannot reproduce on Mac. If it is still a problem, I think it should be solved by triggering a `PointerEvent::Released` when focus is lost (i.e. on alt-tab), and not on `PointerGone`

### DIFF
--- a/crates/eframe/src/web/events.rs
+++ b/crates/eframe/src/web/events.rs
@@ -284,6 +284,8 @@ pub(crate) fn install_color_scheme_change_event(runner_ref: &WebRunner) -> Resul
 
 pub(crate) fn install_canvas_events(runner_ref: &WebRunner) -> Result<(), JsValue> {
     let canvas = runner_ref.try_lock().unwrap().canvas().clone();
+    let window = web_sys::window().unwrap();
+    let document = window.document().unwrap();
 
     {
         let prevent_default_events = [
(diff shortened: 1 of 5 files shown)


jprochazk:
This works, but it behaves a bit strangely when dragging windows. The window always moves by the mouse's delta, including if the mouse is outside the viewport.

1. Open the web demo
1. Start dragging a window
1. Move the mouse outside of the viewport, e.g. into the browser title bar
1. Move the mouse back into the viewport

Now the mouse is outside of the window it's dragging, but the mouse movement still moves the window. I think one possible fix would be to not apply any mouse delta for a given axis if the mouse is outside of the canvas clientRect on that axis.

For example: Canvas client rect is `top_left=[100 100], bottom_right=[200 200]`, mouse is `x=100, y=50`, moving the mouse to `x=100, y=70` will _not_ move the window 20 pixels in `+y`.

emilk:
@jprochazk I think the problem there is the drag-tracking in `Window` should just store where on the Window the drag started (in local coordinates)

jprochazk (approved):
Nice!

PR #4536, opened 2024-05-24
Use ResizeObserver instead of `resize` event

jprochazk:
Currently, if the size of the canvas element changes independently of the size of the browser window (e.g. due to its parent element shrinking), then no repaints are scheduled.

This PR replaces the `resize` event with a `ResizeObserver`, which ensures that _any_ resize of the canvas element (including those caused by browser window resizes) trigger a repaint. The repaint is done synchronously as part of the resize event, to reduce any potential flickering.

The result seems to pass the rendering tests on most platform+browser combinations. We tested:
- Chrome, Firefox, Safari on macOS
- Chrome, Firefox on Linux (ubuntu and arch, both running wayland)
- Chrome, Firefox on Windows

Firefox still has some antialiasing issues on Linux platforms, but this antialiasing also happens on `master`, so this PR is not a regression there.

The code setting `canvas.style.width` and `canvas.style.height` at the start of `AppRunner::logic` was also removed - the canvas _display_ size is now fully controlled by CSS, e.g. by setting `canvas { width: 100%; height: 100%; }`.

The approach used here is described in https://webglfundamentals.org/webgl/lessons/webgl-resizing-the-canvas.html

Note: The only remaining place where egui updates the style of the canvas it is rendering to is some of the IME/mobile input handling code. Fixing that is out of scope for this PR, and will be done in a followup PR.


crates/eframe/src/web/app_runner.rs
```diff
@@ -184,7 +185,6 @@ impl AppRunner {
     ///
     /// The result can be painted later with a call to [`Self::run_and_paint`] or [`Self::paint`].
     pub fn logic(&mut self) {
-        super::resize_canvas_to_screen_size(self.canvas(), self.web_options.max_size_points);
```
emilk, line 187: Is `max_size_points` no longer respected? If so, we should `#[deprecate]` or delete it, and explain what CSS incantation replaces it

crates/eframe/src/web/events.rs
```diff
@@ -590,3 +591,71 @@ pub(crate) fn install_canvas_events(runner_ref: &WebRunner) -> Result<(), JsValu
...
+
+pub(crate) fn install_resize_observer(runner_ref: &WebRunner) -> Result<(), JsValue> {
+    let closure = Closure::wrap(Box::new({
+        let runner_ref = runner_ref.clone();
+        move |entries: js_sys::Array| {
+            // Only call the wrapped closure if the egui code has not panicked
+            if let Some(mut runner_lock) = runner_ref.try_lock() {
+                runner_lock.needs_repaint.repaint_asap();
+                let canvas = runner_lock.canvas();
+                let (width, height) = get_display_size(&entries).unwrap();
```
emilk, line 603: unwrap 😬 

please match and log error instead

crates/eframe/src/web/events.rs
```diff
@@ -590,3 +591,71 @@ pub(crate) fn install_canvas_events(runner_ref: &WebRunner) -> Result<(), JsValu
...
+            height = size.block_size();
+        }
+    } else {
+        // legacy
+        let content_rect = entry.content_rect();
+        width = content_rect.width();
+        height = content_rect.height();
+    }
+
+    Ok(((width.round() * dpr) as u32, (height.round() * dpr) as u32))
```
emilk, line 660: Are we sure about the rounding _before_ applying dpr? The linked code applies it _after_.

Remember that dpr is often fractional

crates/eframe/src/web/events.rs
```diff
@@ -590,3 +591,71 @@ pub(crate) fn install_canvas_events(runner_ref: &WebRunner) -> Result<(), JsValu
...
+// https://webglfundamentals.org/webgl/lessons/webgl-resizing-the-canvas.html
+fn get_display_size(resize_observer_entries: &js_sys::Array) -> Result<(u32, u32), JsValue> {
+    let width;
+    let height;
+    let mut dpr = web_sys::window().unwrap().device_pixel_ratio();
+
+    let entry: web_sys::ResizeObserverEntry = resize_observer_entries.at(0).dyn_into()?;
+    if JsValue::from_str("devicePixelContentBoxSize").js_in(entry.as_ref()) {
+        // NOTE: Only this path gives the correct answer for most browsers.
+        // Unfortunately this doesn't work perfectly everywhere.
```
emilk, line 633: Did you test on any browser that didn't follow this path?

web_demo/index.html
```diff
@@ -50,6 +50,8 @@
             position: absolute;
             top: 0%;
             left: 50%;
+            width: 100%;
+            height: 100%;
```
emilk, line 54: Is this now required? Then please add a migration guide to the PR description!

crates/eframe/src/web/web_runner.rs
```diff
@@ -28,8 +25,10 @@ pub struct WebRunner {
     /// the panic handler, since they aren't `Send`.
     events_to_unsubscribe: Rc<RefCell<Vec<EventToUnsubscribe>>>,
 
-    /// Used in `destroy` to cancel a pending frame.
-    request_animation_frame_id: Cell<Option<i32>>,
```
emilk, line 32: What is the relevance of the changes to the `request_animation_frame` stuff here? What was wrong with the old code?

jprochazk: The official `requestAnimationFrame` example in wasm bindgen [does the same thing](https://rustwasm.github.io/wasm-bindgen/api/wasm_bindgen/closure/struct.Closure.html#using-fnonce-and-closureonce-with-requestanimationframe). The way I understand it, calling `Closure::forget()` in this case is just unnecessary GC pressure. Unlikely to matter but it was bothering me 😅.

crates/eframe/src/web/events.rs
jprochazk: I don't remember, will re-test

web_demo/index.html
jprochazk: Not required, but by default a canvas element is very small (300x150px). It's now entirely up to the user how they decide to display it, so any `width`/`height` is fine.

crates/eframe/src/web/events.rs
jprochazk: Yep, this was incorrect.

crates/eframe/src/web/events.rs
jprochazk: It seems like most of the code is just to support slightly older versions of Safari. Latest Safari doesn't support `devicePixelContentBoxSize` and it seems to work there: https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry#browser_compatibility

I also commented out all the paths one-by-one to see if they work at all (_except_ the `contentBoxSize` non-array path) and didn't run into any issues.

web_demo/index.html
jprochazk: I'm not really sure where to add a migration guide, because there's nothing to migrate over here yet, because it is still not possible to set `width`/`height` without having it overriden by `egui`. The IME handling still sets the canvas style, but it'd be better to leave that to its own PR. This one is just to get the canvas resize working properly.

crates/eframe/src/web/web_runner.rs
emilk: Could it be moved to a separate PR?

web_demo/index.html
emilk: So a user that has copy-pasted the old index.html would not need to change anything when updating eframe?

jprochazk: They would, but it doesn't yet to the right thing. I guess what I'm asking is does the migration guide go into this PR or the followup that fixes IME?

crates/eframe/src/web/web_runner.rs
jprochazk: https://github.com/emilk/egui/pull/4551

web_demo/index.html
emilk: I guess it can wait 👍

emilk (approved):
This is wonderful!

With [the new alignment test](https://github.com/emilk/egui/pull/4537) I managed to find a case on Mac chromium where the old code doesn't align to the pixel grid on trunk, but this PR fixes it, which is _awesome_. When resizing the chromium browser the old code would also create jitter, which is now gone. So great!

On Mac Firefox the old and new code acts pretty much the same afaict, but with a different _kind_ of jittering behavior when resizing the browser 🤷 

Great job!

web_demo/index.html
```diff
@@ -48,9 +48,10 @@
             margin-left: auto;
             display: block;
             position: absolute;
-            top: 0%;
-            left: 50%;
-            transform: translate(-50%, 0%);
+            top: 0;
```
emilk, line 51: We should have updated the comment for the `canvas` block as well. I don't think it is accurate anymore.

PR #4539, opened 2024-05-24
egui_plot: Force use of min_auto_bounds if auto_bounds is disabled by default.  Fixes  #3893.

### LINKED ISSUE #3893: egui_plot 0.25.0 - No auto_bounds + include_X/Y does not render anything until double click
**Describe the bug**

When trying to manually set the data view port with Include X/Y, the plot will show nothing until double clicked.

This is being done to hide the most recent plot point of data in a historic time series (Which is retrieved every 100ms or so), in order to avoid the appearance of stuttering when data comes in. (Stuttering will appear as soon as `plot.auto_bounds([true, true].into())` is used.

**To Reproduce**
1. Create a thread which adds a plot point every 100ms or so
```rust
let mut plot = Plot::new(d.group_name.clone())
                          .height(space_per_chart)
                          .allow_drag(false)
                          .include_x(disp_now as f64) // Timestamp now - 100.0ms to hide the most recent plot point (To look smooth)
                          .include_x(last_bound) // Timestamp now - 20 seconds
                          .legend(legend.clone());
                      if let Some((min, max)) = &d.bounds {// Data has finite range in Y dir, restrict it
                          plot = plot.include_y(*min);
                          if *max > 0.1 {
                              // 0.0 check
                              plot = plot.include_y(*max);
                          }
                      }
                     // False False - User has to double click to show the chart, and it looks smooth
                    //  True True - Plot stutters as the infrequent data will show
                      plot = plot.auto_bounds([false, false].into());
                      plot.show(col, |f| {
                          for line in lines {
                              f.line(line);
                          }
                      });
```

**Expected behavior**
Plot should show the requested area immediately, rather than wait for user input.

**Screenshots**
With `plot = plot.auto_bounds([false, false].into());` (Buggy)
[Kooha-2024-01-26-10-09-39.webm](https://github.com/emilk/egui/assets/14551965/be21283f-fa5f-4725-9c2e-3100ea2f2623)

With `plot = plot.auto_bounds([true, true].into());` (Stutters, but expected behaviour, hence why no autobounds is being used)
[Kooha-2024-01-26-10-10-42.webm](https://github.com/emilk/egui/assets/14551965/eecf7e63-7a82-45b7-b6bc-aaf70f54d510)

aholtzma-am:
The logic for determining bounds was not using the values supplied by include_x()/include_y() after the initial plot call when auto_bounds() was false. This change forces the use of the manual limits when default_auto_bounds is false.

* Closes <https://github.com/emilk/egui/issues/3893>

### DIFF
--- a/crates/egui_plot/src/lib.rs
+++ b/crates/egui_plot/src/lib.rs
@@ -975,10 +975,10 @@ impl<'a> Plot<'a> {
         }
 
         // Reset bounds to initial bounds if they haven't been modified.
-        if mem.auto_bounds.x {
+        if !default_auto_bounds.x || mem.auto_bounds.x {
             bounds.set_x(&min_auto_bounds);
         }
-        if mem.auto_bounds.y {
+        if !default_auto_bounds.y || mem.auto_bounds.y {
             bounds.set_y(&min_auto_bounds);
         }
 


aholtzma:
Looks like this just needs a label applied.

crates/egui_plot/src/lib.rs
```diff
@@ -975,10 +975,10 @@ impl<'a> Plot<'a> {
         }
 
         // Reset bounds to initial bounds if they haven't been modified.
-        if mem.auto_bounds.x {
+        if !default_auto_bounds.x || mem.auto_bounds.x {
             bounds.set_x(&min_auto_bounds);
```
emilk, line 979: This reads backwards to me now. If the user has said `DON'T apply automatic bounds` then we set the bounds to `min_auto_bounds`

aholtzma-am: The min_auto_bounds includes the bounds manually specified by include_x()/include_y(). So if you turn off auto bounds, and supply bounds via include_xy(), it won't use them except for the first call.

Fabus1184:
Perhaps it would be better to factor this out into two methods, `expand_{x,y}` to expand the bounds to always show the specified value, and `limit_{x,y}` to limit the axis range, both applying regardless of whether auto bounds is active or not.
Of course if the user moves the plot, this should not apply. You could always disallow moving the plot if you don't want that.

emilk:

PR #4546, opened 2024-05-27
Hide toolip when opening `ComboBox` drop-down

### LINKED ISSUE #4338: Hide tooltip when opening `ComboBox` dropdown
* Very similar to https://github.com/emilk/egui/pull/4138

If you set a tooltip for a `ComboBox`, that should be hidden as soon as the user opens the `ComboBox` dropdown, or the tooltip will cover the dropdown.

abey79:
- Fixes #4338

https://github.com/emilk/egui/assets/49431240/73ea87a1-41ad-40b1-b451-d6be2b38c7e0

Tested using `example/hello_world` modified to:
```rust
#![allow(rustdoc::missing_crate_level_docs)] // it's an example

use eframe::egui;

    env_logger::init(); // Log to stderr (if you run with `RUST_LOG=debug`).
        ..Default::default()
    };
    eframe::run_native(
        "My egui App",
        options,
        Box::new(|cc| {
            // This gives us image support:

            Box::<MyApp>::default()
        }),
    )
}

struct MyApp {
    name: String,
    age: u32,
}

impl Default for MyApp {
    fn default() -> Self {
        Self {
            name: "Arthur".to_owned(),
            age: 42,
        }
    }
}

impl eframe::App for MyApp {
            egui::ComboBox::new("combo", "combo box")
                .selected_text(&self.name)
                .show_ui(ui, |ui| {
                    ui.selectable_value(&mut self.name, "Arthur".into(), "Arthur")
                        .on_hover_text("This is Arthur");
                    ui.selectable_value(&mut self.name, "Ford".into(), "Ford")
                        .on_hover_text("This is Ford");
                    ui.selectable_value(&mut self.name, "Trillian".into(), "Trillian")
                        .on_hover_text("This is Trillian");
                })
                .response
                .on_hover_text("This is a combo box");
        });
    }
}
```


crates/egui/src/response.rs
```diff
@@ -571,6 +571,10 @@ impl Response {
             return false;
         }
 
+        if self.ctx.memory(|m| m.any_popup_open()) {
+            return false;
+        }
```
emilk, line 576: Won't this break tooltips for items inside of tooltips? I hovering one of the elements in the ComboBox dropdown list?

abey79: oh you're probably right — testing.

abey79: Fixed

crates/egui/src/containers/combo_box.rs
```diff
@@ -44,6 +44,11 @@ pub struct ComboBox {
 }
 
 impl ComboBox {
+    /// Convert a [`ComboBox`] id to the id used to store it's popup state.
+    pub(crate) fn widget_to_popup_id(widget_id: Id) -> Id {
+        widget_id.with("popup")
+    }
+
```
abey79, line 51: Is there a better place to put this?

crates/egui/src/response.rs
```diff
@@ -571,6 +572,13 @@ impl Response {
             return false;
         }
 
+        if self
+            .ctx
+            .memory(|m| m.is_popup_open(ComboBox::widget_to_popup_id(self.id)))
+        {
```
emilk, line 578: I'd structure this like so instead:

```suggestion
        if ComboBox::is_open(ctx, self.id) {
```

And make `is_open` public

crates/egui/src/response.rs
```diff
@@ -571,6 +572,10 @@ impl Response {
             return false;
         }
 
+        if ComboBox::is_open(&self.ctx, self.id) {
+            return false;
+        }
```
emilk, line 577:
```suggestion
        if ComboBox::is_open(&self.ctx, self.id) {
            return false; // Don't cover the open ComboBox with a tooltip
        }
```

emilk (approved):

emilk:
* Better solved by https://github.com/emilk/egui/pull/4766

PR #4154, opened 2024-03-10
Prevents slight encroachment on the title outer line in egui::Window.

rustbasic:
Prevents slight encroachment on the title outer line in egui::Window.

You can easy checkable when title has a background color.

PR #4547, opened 2024-05-27
Prevents slight encroachment on the title outer line in egui::Window.

rustbasic:
Prevents slight encroachment on the title outer line in egui::Window.

* Related #4154


emilk:
As I've said before: if you're changing the visuals of egui, you must have a before/after screenshot of the demo app

PR #4549, opened 2024-05-27
Hide all other series when alt-clicking in the legend

### LINKED ISSUE #4548: egui_plot legend: add shortcut to hide all series but this one
This could be achieved e.g. by alt-clicking on a series name.

abey79:
https://github.com/emilk/egui/assets/49431240/75d32f53-4c1c-4713-b25e-e1787e465a48

* Closes #4548

### DIFF
--- a/crates/egui_plot/src/legend.rs
+++ b/crates/egui_plot/src/legend.rs
@@ -99,11 +99,11 @@ impl LegendEntry {
         }
     }
 
-    fn ui(&mut self, ui: &mut Ui, text: String, text_style: &TextStyle) -> Response {
+    fn ui(&self, ui: &mut Ui, text: String, text_style: &TextStyle) -> Response {
         let Self {
             color,
             checked,
-            hovered,
+            hovered: _,
         } = self;
 
         let font_id = text_style.resolve(ui.style());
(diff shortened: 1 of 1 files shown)


crates/egui_plot/src/legend.rs
```diff
@@ -267,14 +267,43 @@ impl Widget for &mut LegendWidget {
...
-                        entries
+                        let mut focus_on_item = None;
+
+                        let response_union = entries
                             .iter_mut()
-                            .map(|(name, entry)| entry.ui(ui, name.clone(), &config.text_style))
+                            .map(|(name, entry)| {
+                                let response = entry.ui(ui, name.clone(), &config.text_style);
+
+                                if response.clicked() && ui.input(|r| r.modifiers.alt) {
```
emilk, line 277: Where is the normal clicking handled? We shouldn't spread out the different click-handling to different places

abey79: good call. i'll refactor some stuff.

emilk (approved):

emilk:
Where can we document this feature? Would a tooltip when hovering the legend be too annoying?

abey79:
> Where can we document this feature? Would a tooltip when hovering the legend be too annoying?

I've pondered about this and decided against the tooltip—too annoying and difficult to override for the host application. I guess I can add a feature flag for this in `Legend`, such that the functionality appears in the docs. I'd argue that the host application is responsible for controlling what the user can do and provide help about it (e.g. also double click to reset, etc.).

For rerun, I have it in my todos to add something in the space view help.

PR #4551, opened 2024-05-27
Fix: Don't `.forget()` RAF closure

jprochazk:
The closure is now stored in `WebRunner` and dropped in the next `request_animation_frame` instead of being "leaked" via `forget`


crates/eframe/src/web/web_runner.rs
```diff
@@ -193,20 +191,47 @@ impl WebRunner {
...
+            return Ok(());
+        }
+
         let window = web_sys::window().unwrap();
         let closure = Closure::once({
             let runner_ref = self.clone();
-            move || events::paint_and_schedule(&runner_ref)
+            move || {
+                // we can paint now, so clear the animation frame
+                // this drop the `closure` and allows another
```
emilk, line 204:
```suggestion
                // We can paint now, so clear the animation frame.
                // This drops the `closure` and allows another
```

I think?

crates/eframe/src/web/web_runner.rs
```diff
@@ -193,20 +191,47 @@ impl WebRunner {
...
+                // animation frame to be scheduled
+                let _ = runner_ref.frame.take();
```
emilk, line 206:
```suggestion
                runner_ref.frame = None;
```

crates/eframe/src/web/web_runner.rs
```diff
@@ -115,9 +112,10 @@ impl WebRunner {
     pub fn destroy(&self) {
         self.unsubscribe_from_all_events();
 
-        if let Some(id) = self.request_animation_frame_id.get() {
+        if let Some(frame) = self.frame.take() {
             let window = web_sys::window().unwrap();
-            window.cancel_animation_frame(id).ok();
+            window.cancel_animation_frame(frame.id).ok();
+            drop(frame.closure);
```
emilk, line 118: this happens implicitly anyway

```suggestion
```

emilk (approved):
Thanks for breaking this out!

crates/eframe/src/web/web_runner.rs
jprochazk: This doesn't work because it's an `Rc<RefCell<Option<...>>>`

PR #4555, opened 2024-05-28
Ability to hide rulers of cursor on plots

Tombleron:
Sometimes rulers can be distracting and unnecessary. Linked cursors will still be displayed, but since their visibility is already configurable, it's not an issue.


crates/egui_plot/src/lib.rs
```diff
@@ -599,6 +601,15 @@ impl<'a> Plot<'a> {
         self
     }
 
+    /// Show or hide the cursor lines.
+    ///
+    /// Default: `true`.
+    #[inline]
+    pub fn enable_cursor_drawing(mut self, enable: bool) -> Self {
+        self.enable_cursor_drawing = enable;
```
emilk, line 609: I think `show_cursor` is a shorter and nicer name

PR #4556, opened 2024-05-28
Add support for text truncation to `egui::Style`

### LINKED ISSUE #4473: `Style` feature to have label truncate by default
In some context, it would be useful to have `ui.label(txt)` defaults to truncate, such as we don't have to change them all for `ui.add(egui::Label::new(txt).truncate(true))`. Likewise, it would be nice to have a `ui.label_truncated()` helper (though each might be suited for different situations).

As per:
- https://github.com/rerun-io/rerun/issues/4161#issuecomment-2089947539

abey79:
* Closes #4473

This PR introduce `Style::wrap_mode`, which adds support for text truncation in addition to text wrapping. This PR also update some width calculation of the ComboBox.

#### Core

- Add `egui::TextWrapMode` (pure enum with `Extend`, `Wrap`, `Truncate`)
- Add `Style::wrap_mode: Option<tTextWrapMode>`
- **DEPRECATED**: `Style::wrap`, use `Style::wrap_mode` instead.
- Add `Ui::wrap_mode()` to return the wrap mode to use in the current ui. If specified in `Style`, return it. Otherwise, return `TextWrapMode::Wrap` for vertical layout and wrapping horizontal layout, and `TextWrapMode::Extend` otherwise.
- **DEPRECATED**: `Ui::wrap_text()`, use `Ui::wrap_mode` instead.

#### Widget

- Update the width calculation of the `ComboBox` button (_not_ its popup menu).
  - Now, `ComboBox::width()` (defaulting to `Spacing::combo_width`) is always considered a minimum width and will extend the `Ui`, regardless of the selected text width and wrap mode.
  - Introduce `ComboBox::wrap_mode`, which overrides `Ui::wrap_mode` for the selected text layout.
    - Note: since `ComboBox` uses `ui.horizontal` internally, the default wrap mode is always `TextWrapMode::Extend`, regardless of the caller's `Ui`'s layout.
  - The `ComboBox` button no longer extend to `ui.available_width()` with wrapping is enabled.
  - **BREAKING**: `ComboBox::wrap()` no longer has a `bool` argument and is now a short-hand for `ComboBox::wrap_mode(TextWrapMode::Wrap)`.
  - Added `ComboBox::truncate()` as short-hand for `ComboBox::wrap_mode(TextWrapMode::Truncate)`.
- Update `Label`
  - Add `Label::wrap_mode()` to specify the text wrap mode.
  - **BREAKING**:  `Label::wrap()` no longer has a `bool` argument and is now a short-hand for `Label::wrap_mode(TextWrapMode::Wrap)`.
  - **BREAKING**:  `Label::truncate()` no longer has a `bool` argument and is now a short-hand for `Label::wrap_mode(TextWrapMode::Truncate)`.
- Update `Button`
  - Add `Button::wrap_mode()` to specify the text wrap mode.
  - **BREAKING**:  `Button::wrap()` no longer has a `bool` argument and is now a short-hand for `Button::wrap_mode(TextWrapMode::Wrap)`.
  - Added `Button::truncate()` as short-hand for `Button::wrap_mode(TextWrapMode::Truncate)`.

#### Low-level

- **BREAKING**: `WidgetText::into_galley()` now takes an `Option<TextWrapMode>` instead of a `Option<bool>` argument.
- **BREAKING**: `WidgetText::into_galley_impl(()` now takes a `TextWrapping` argument instead of `wrap: bool` and `availalbe_width: f32` arguments.

### DIFF
--- a/crates/egui/src/containers/combo_box.rs
+++ b/crates/egui/src/containers/combo_box.rs
@@ -148,14 +148,6 @@ impl ComboBox {
         self
     }
 
-    /// Controls whether text wrap is used for the selected text.
-    #[deprecated = "Use `wrap_mode` instead"]
-    #[inline]
-    pub fn wrap(mut self, wrap: bool) -> Self {
-        self.wrap_mode = if wrap { Some(TextWrapMode::Wrap) } else { None };
-        self
-    }
-
     /// Controls the wrap mode used for the selected text.
     ///
     /// By default, [`Ui::wrap_mode`] will be used, which can be overridden with [`Style::wrap_mode`].
@@ -164,6 +156,21 @@ impl ComboBox {
     #[inline]
     pub fn wrap_mode(mut self, wrap_mode: TextWrapMode) -> Self {
         self.wrap_mode = Some(wrap_mode);
+        self
+    }
+
+    /// Set [`Self::wrap_mode`] to [`TextWrapMode::Wrap`].
+    #[inline]
+    pub fn wrap(mut self) -> Self {
+        self.wrap_mode = Some(TextWrapMode::Wrap);
+
+        self
+    }
+
+    /// Set [`Self::wrap_mode`] to [`TextWrapMode::Truncate`].
+    #[inline]
+    pub fn truncate(mut self) -> Self {
+        self.wrap_mode = Some(TextWrapMode::Truncate);
         self
     }
 
--- a/crates/egui/src/containers/combo_box.rs
+++ b/crates/egui/src/containers/combo_box.rs
@@ -158,7 +158,7 @@ impl ComboBox {
 
     /// Controls the wrap mode used for the selected text.
     ///
-    /// By default, the text will extend the [`Ui`] boundary. This default behavior can be
+    /// By default, [`Ui::wrap_mode`] will be used, which can be
     /// overridden with [`Style::wrap_mode`].
     ///
     /// Note that any `\n` in the text will always produce a new line.
--- a/crates/egui/src/containers/combo_box.rs
+++ b/crates/egui/src/containers/combo_box.rs
@@ -373,9 +373,7 @@ fn combo_box_dyn<'c, R>(
                     // result in labels that wrap very early.
                     // Instead, we turn it off by default so that the labels
                     // expand the width of the menu.
-                    if ui.style().wrap_mode == Some(TextWrapMode::Wrap) {
-                        ui.style_mut().wrap_mode = None;
-                    }
+                    ui.style_mut().wrap_mode = Some(TextWrapMode::Extend);
                     menu_contents(ui)
                 })
                 .inner
--- a/crates/egui/src/containers/combo_box.rs
+++ b/crates/egui/src/containers/combo_box.rs
@@ -1,3 +1,4 @@
+use epaint::text::TextWrapMode;
 use epaint::Shape;
 
 use crate::{style::WidgetVisuals, *};
@@ -5,7 +6,7 @@ use crate::{style::WidgetVisuals, *};
 #[allow(unused_imports)] // Documentation
 use crate::style::Spacing;
 
-/// Indicate whether or not a popup will be shown above or below the box.
+/// Indicate whether a popup will be shown above or below the box.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum AboveOrBelow {
     Above,
@@ -40,7 +41,7 @@ pub struct ComboBox {
     width: Option<f32>,
     height: Option<f32>,
     icon: Option<IconPainter>,
-    wrap_enabled: bool,
+    wrap_mode: Option<TextWrapMode>,
 }
 
 impl ComboBox {
@@ -53,7 +54,7 @@ impl ComboBox {
             width: None,
             height: None,
             icon: None,
-            wrap_enabled: false,
+            wrap_mode: None,
         }
     }
 
@@ -67,7 +68,7 @@ impl ComboBox {
             width: None,
             height: None,
             icon: None,
-            wrap_enabled: false,
+            wrap_mode: None,
         }
     }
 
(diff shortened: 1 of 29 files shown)


crates/egui/src/containers/combo_box.rs
```diff
@@ -371,7 +374,9 @@ fn combo_box_dyn<'c, R>(
                     // result in labels that wrap very early.
                     // Instead, we turn it off by default so that the labels
                     // expand the width of the menu.
-                    ui.style_mut().wrap = Some(false);
+                    if ui.style().wrap_mode == Some(TextWrapMode::Wrap) {
+                        ui.style_mut().wrap_mode = None;
+                    }
```
emilk, line 379: The comment no longer matches the behavior.

Why isn't this just 
```suggestion
                    ui.style_mut().wrap_mode = Some(TextWrapMode::Extend);
```
?

crates/egui/src/style.rs
```diff
@@ -182,15 +182,25 @@ pub struct Style {
...
+    #[deprecated = "Use wrap_mode instead"]
     pub wrap: Option<bool>,
 
+    /// If set, labels, buttons, etc. will use this to determine whether to wrap or truncate the
+    /// text at the right edge of the [`Ui`] they are in, or to extend it. By default, this is
+    /// `None`.
+    ///
+    /// * `None`: follow layout (with may wrap)
+    /// * `Some(mode)`: use the specified mode as default
+    pub wrap_mode: Option<crate::text::TextWrapMode>,
```
emilk, line 202: I think `TextWrapMode` should be exported top-level

```suggestion
    pub wrap_mode: Option<crate::TextWrapMode>,
```

crates/egui/src/ui.rs
```diff
@@ -329,20 +330,44 @@ impl Ui {
...
-    /// Should text wrap in this [`Ui`]?
+    /// Which wrap mode should the text use in this [`Ui`]?
     ///
-    /// This is determined first by [`Style::wrap`], and then by the layout of this [`Ui`].
-    pub fn wrap_text(&self) -> bool {
+    /// This is determined first by [`Style::wrap_mode`], and then by the layout of this [`Ui`].
+    pub fn wrap_mode(&self) -> TextWrapMode {
+        // `wrap` handling for backward compatibility
+        #[allow(deprecated)]
         if let Some(wrap) = self.style.wrap {
```
emilk, line 339: We should start with the new (`wrap_mode`) and only use the deprecated items as the fallback

crates/egui/src/widgets/label.rs
```diff
@@ -9,10 +9,11 @@ use self::text_selection::LabelSelectionState;
 /// Usually it is more convenient to use [`Ui::label`].
 ///
 /// ```
+/// # use egui::text::TextWrapMode;
 /// # egui::__run_test_ui(|ui| {
 /// ui.label("Equivalent");
 /// ui.add(egui::Label::new("Equivalent"));
-/// ui.add(egui::Label::new("With Options").wrap(false));
+/// ui.add(egui::Label::new("With Options").wrap_mode(TextWrapMode::Truncate));
```
emilk, line 16: This becomes quite verbose.

I think we should consider having two helpers:

* `.truncate() == .wrap_mode(TextWrapMode::Truncate)`
* `.wrap() == .wrap_mode(TextWrapMode::Wrap)`

They then no longer have any `bool` arguments, which could potentially make migration a bit more confusing for users, but the end API is pretty nice:

`ui.add(egui::Label::new("Foo").wrap())`

crates/egui_demo_app/src/backend_panel.rs
```diff
@@ -180,7 +180,7 @@ fn integration_ui(ui: &mut egui::Ui, _frame: &mut eframe::Frame) {
 
     #[cfg(target_arch = "wasm32")]
     ui.collapsing("Web info (location)", |ui| {
-        ui.style_mut().wrap = Some(false);
+        ui.style_mut().wrap_mode = Some(egui::text::TextWrapMode::Extend);
```
emilk, line 183: here and elsewhere
```suggestion
        ui.style_mut().wrap_mode = Some(egui::TextWrapMode::Extend);
```

crates/epaint/src/text/text_layout_types.rs
```diff
@@ -319,6 +319,17 @@ impl TextFormat {
...
 
+/// How to wrap and elide text.
+///
+/// This enum is used in high-level APIs where providing a [`TextWrapping`] is too verbose.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub enum TextWrapMode {
+    Extend,
+    Wrap,
+    Truncate,
```
emilk, line 330: These should have docstrings to explain them fully

emilk (approved):
Looks great!

crates/egui/src/containers/combo_box.rs
```diff
@@ -148,10 +149,23 @@ impl ComboBox {
...
     pub fn wrap(mut self, wrap: bool) -> Self {
-        self.wrap_enabled = wrap;
+        self.wrap_mode = if wrap { Some(TextWrapMode::Wrap) } else { None };
+        self
+    }
+
+    /// Controls the wrap mode used for the selected text.
+    ///
+    /// By default, the text will extend the [`Ui`] boundary. This default behavior can be
+    /// overridden with [`Style::wrap_mode`].
```
emilk, line 163:
```suggestion
    /// By default, [`Ui::wrap_mode`] will be used, which can be
    /// overridden with [`Style::wrap_mode`].
```

crates/egui/src/style.rs
abey79: 👍🏻

crates/epaint/src/text/text_layout_types.rs
abey79: That was very much my intention 🤦🏻

crates/egui/src/widgets/label.rs
abey79: Totally agree with that—was reluctant because of the intrusive breaking change, but the end result is worth it!

Issue #3974, opened 2024-02-05
egui menus: remember size from previous frame


emilk:
egui menus use a justified layout, so that buttons fill the full width.
This means we currently have to pick a width up front and then fit all buttons to that.

It would be nicer to be able to dynamically size the menu based on the contents in the first frame.

Issue #4535, opened 2024-05-24
Sizing-pass flag


emilk:
### Problem
The first frame we shown an Area or a Grid we only measure the size of everything, but don't show anything to the user. This the first frame becomes a size-measuring pass. This is a poor-mans two-pass mode.

A common problem is "How wide do we need to make this area"? (e.g. a menu).
We usually want to set a max width of e.g. 400 points, at which we should start wrapping.
But if we can keep things more narrow, that's better.

Now consider a widget that expands to fill its parent, e.g. a `Separator`.
It will the first frame expand to the full max width, and then the min width = max width, and so we will always expand to the maximum allowed with. This is bad, causing menus to become overly wide.

For more details, see https://github.com/emilk/egui/issues/843

### Solution
We set a simple boolean flag for the sizing pass.
Widgets that expand to fill the max_width would first check this flag, and if set, they would _not_ expand, but only allocate the minimum size they actually need.
When the sizing pass finishes, we store the final size in the `Area` state (or whatever), and then next frame we have a properly sized area.

This flag would be passed down to child UIs.

### Would close
* https://github.com/emilk/egui/issues/3974
* #4452 
* https://github.com/rerun-io/rerun/issues/6246

### Other related
* #4471
* https://github.com/emilk/egui/issues/843
* https://github.com/emilk/egui/issues/4378

PR #4557, opened 2024-05-28
Add `Ui::is_sizing_pass` for better size estimation of `Area`s, and menus in particular

emilk:
* Part of https://github.com/emilk/egui/issues/4535
* Closes https://github.com/emilk/egui/issues/3974

This adds a special `sizing_pass` mode to `Ui`, in which we have no centered or justified layouts, and everything is hidden. This is used by `Area` to use the first frame to measure the size of its contents so that it can then set the perfectly correct size the subsequent frames.

For menus, where buttons are justified (span the full width), this finally the problem of auto-sizing. Before you would have to pick a width manually, and all buttons would expand to that width. If it was too wide, it looked weird. If it was too narrow, text would wrap. Now all menus are exactly the width they need to be. By default menus will wrap at `Spacing::menu_width`.

This affects all situations when you have something that should be as small as possible, but still span the full width/height of the parent. For instance: the `egui::Separator` widget now checks the `ui.is_sizing_pass` flag before deciding on a size. In the sizing pass a horizontal separator is always 0 wide, and only in subsequent passes will it span the full width.

### DIFF
--- a/crates/egui/src/containers/area.rs
+++ b/crates/egui/src/containers/area.rs
@@ -165,7 +165,7 @@ impl Area {
         self
     }
 
-    /// The size used for the [`Ui::max_size`] the first frame.
+    /// The size used for the [`Ui::max_rect`] the first frame.
     ///
     /// Text will wrap at this width, and images that expand to fill the available space
     /// will expand to this size.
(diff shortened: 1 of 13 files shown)


crates/egui/src/containers/area.rs
```diff
@@ -461,10 +508,15 @@ impl Prepared {
...
-            temporarily_invisible: _,
+            sizing_pass: _,
         } = self;
 
-        state.size = content_ui.min_size();
+        // We round up to an integer point size.
+        // This avoid rounding errors, especially important after a sizing pass.
+        // If during the sizing pass we measure our width to `123.45` and
+        // then try to wrap to exactly that next frame,
+        // we may accidentally wrapping the last letter of some text.
```
emilk, line 518:
```suggestion
        // we may accidentally wrap the last letter of some text.
```

emilk:
Contents of context menus sometimes wrap too early, and sometimes continuously expand…

PR #4559, opened 2024-05-28
Revert update to wgpu 0.20 => downgrade to wgpu 0.19.1

emilk:
0.20 has a bunch of bugs that will be fixed by:

* https://github.com/gfx-rs/wgpu/pull/5681

At Rerun, we don't want to wait for the wgpu 0.20.1 patch release before we update egui, so we will temporarily downgrade to wgpu 0.19

After reverting I'll open a new PR that will update to 0.20 again, with the intention of merging that once 0.20.1 is released.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -108,7 +108,7 @@ jobs:
       - name: wasm-bindgen
         uses: jetli/wasm-bindgen-action@v0.1.0
         with:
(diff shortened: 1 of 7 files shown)


emilk:
If you want to use wgpu 0.20.0 now, you can track this branch: https://github.com/emilk/egui/pull/4560

Issue #4479, opened 2024-05-11
Allow for manual placement of the canvas element


AlexanderSchuetz97:
**Is your feature request related to a problem? Please describe.**
For Wasm/Web, I want to integrate egui into an existing large single page web application written in javascript/typescript. 
egui should only render stuff inside a canvas. Placement of that canvas is very cumbersome as
egui runs code in its update loop that just assumes that egui takes care of the entire screen.

**Describe the solution you'd like**
An option for me prevent egui from touching the css style of the canvas element at all.
Ideally egui should look at the canvas size in its draw routine in this mode and just "accept" that this is the size it has to use.

**Describe alternatives you've considered**
Wrap the canvas in a div, and overwrite the css style in the div. This is however very wonky as 
if I do not set the size of the div to absolute pixel values for every execution of the update routine the canvas grows larger.
This prevents me from using %width or %height stuff in the div.

My current workaround is not pretty but works:
```
<div style="height: 500px; width: 500px; position: relative !important;">
    <canvas id="the_canvas_id"></canvas>
</div>
```
the "position: relative !important" has to be set to overwrite the "position: absolute" that egui sets on the canvas element.
and the sizes have to be set for the update loop to be able to detect. Again if I set the sizes to a percentage then
the canvas will just add 1 pixel each time the update loop is ran. (rounding issue?)

I would prefer to not have to do this at all and just set some flag in WebOptions that placement of the canvas 
is "manual" and egui should not mess with the style tag of the canvas at all. Then I could just set all
the stuff on the canvas directly and not abuse a div for this. Plus hopefully the canvas size increase issue would go away
when using %height %width.

Setting the size in the WebOptions::max_size_points is not really helping either since the canvas may move/resize
depending on what the user does and there doesn't appear to be a way of changing these values again after starting
the WebRunner.

**Additional context**
I am not using trunk, as my goal is to integrate egui into an existing application. I am using wasm-pack. This should not
have any impact tho as the only thing thats different is that the entrypoint is not called main but some other function 
that the javascript code calls with the canvas id.
The rest of the "app" code is for now identical to emilk/eframe_template.

jprochazk:
There are two reasons why this is happening:

[...]

Issue #4060, opened 2024-02-16
TextAgent assumes canvas is aligned to the top


canselcik:
https://github.com/emilk/egui/blob/f7fc3b0154bfef4f7f787f0eab21c48fe497b583/crates/eframe/src/web/text_agent.rs#L153-L154

There seems to be no way to disable this. Any input to the canvas makes the canvas get a `position: absolute, top: 0%`, breaking layout.

Maybe `mouseup` shouldn't result in `text_agent::update_text_agent(runner)` getting called.

Maybe eframe should save some of the original style attributes of the canvas and then restore them instead of making this assumption.

emilk:
Web stuff is not really my skill set, so any help here would be appreciated.

canselcik:
> Web stuff is not really my skill set, so any help here would be appreciated.

Would be happy to help. Am I understanding correctly that this is for IME support? Was this behind a feature flag in the past?

If this was for IME support, what's the intention behind changing the position of the canvas element at all? At a quick glance, it appears the intention would be to slide it up and down to accommodate the keyboard that would slide in on a mobile device but my assumption would be that the keyboard appearing would already have that effect.

canselcik:
https://github.com/canselcik/egui/commit/be66734dd1ec4cbabe316de25e3642661b8644b2
How about something like this?

Since it is non-trivial accessing some configuration in that closure, I figured this would be a compromise-free option.

If the canvas isn't flush to the upper left corner of the page, pushing it up like it is currently done will result in potentially broken page layout. So if the canvas is flush to the upper left corner, we do what is currently done and then set an attribute on the canvas. We prefix it with `data-` as custom attributes doing that is allowed by the HTML5 spec. When the on screen keyboard is gone, we check for that custom attribute, if it is set, we unset it and "restore" the canvas back to the upper left corner of the page.

jprochazk:
The current workaround is to add some CSS to the canvas:
```css
canvas {
  position: initial !important;
  top: unset !important;
}
```

Ideally, egui would never set the style of the canvas element for any reason.

Issue #3618, opened 2023-11-24
egui_text_agent is not removed from the dom when WebRunner.destroy() is called. 


LiamPClancy:
**Describe the bug**
I'm loading a WASM package into a a react app, it's using egui to render the plot and rust to do logic. 
when calling wasmHandle.start(...)
it adds the text agent with 
super::text_agent::install_text_agent(self)?;

but when calling 
wasmHandle.destroy()
the text_agent is not removed from the dom,
This is creating an issue if the wasmHandle.start(...) is called again. 

my use case is that when someone logs out the wasmHandle is destroyed, and if they log back in again then .start is called. 
i'm handling this by removing the   egui_text_agent when the user logs out, but it'd be better i think if it was part of the destroy method. 

I'm thinking to add 

```
pub fn remove_text_agent() -> Result<(), JsValue> {
    if let Some(element) = web_sys::window()
        .unwrap()
        .document()
        .unwrap()
        .get_element_by_id(AGENT_ID)
    {
        if let Some(parent) = element.patent_node() {
            patent.remove_child(&element)?;
        }
    };

    Ok(());
}
```

to the text_agent.rs

and then modifying webRunner destroy with a call to remove_text_agent()

```
pub fn destroy(&self) {
        self.unsubscribe_from_all_events();
        super::text_agent::remove_text_agent()?;
        if let Some(runner) = self.runner.replace(None) {
            runner.destroy();
        }
    }

```

Happy to make a PR if you think the above is the correct approach?

emilk:
Sounds good - please make a PR!

Issue #2231, opened 2022-11-03
egui eframe TextEdit on mobile


Philipp-Sc:
Hi, thank you for creating this awesome repository. It's a lot of fun!

Is there a way to disable the following behavior?

On Mobile when the TextEdit gets focus (via touch, or click on the TextEdit area), the keyboard pops up AND the position of the Canvas is pushed up 30 to 45% leaving a gray space between the canvas and keyboard.

Am I missing something?

This can be reproduced by using this code [here](https://github.com/Philipp-Sc/librelearning/tree/0e7ad454f18fcf200b4a446a6b2246f6c409da56), ~~a live version can be found [here](https://www.philipp-schluetermann.de/librelearning/index.html).~~ (live version now contains a workaround)

Relevant code:

[eframe/src/web/text_agent.rs](https://github.com/emilk/egui/blob/041f2e64bac778c9095fbf4316dc1f7c7bceb670/crates/eframe/src/web/text_agent.rs)

```rust
            // Move up canvas so that text edit is shown at ~30% of screen height.
            // Only on touch screens, when keyboard popups.
            if let Some(latest_touch_pos) = runner.input.latest_touch_pos {
                let window_height = window.inner_height().ok()?.as_f64()? as f32;
                let current_rel = latest_touch_pos.y / window_height;

                // estimated amount of screen covered by keyboard
                let keyboard_fraction = 0.5;

                if current_rel > keyboard_fraction {
                    // below the keyboard

                    let target_rel = 0.3;

                    // Note: `delta` is negative, since we are moving the canvas UP
                    let delta = target_rel - current_rel;

                    let delta = delta.max(-keyboard_fraction); // Don't move it crazy much

                    let new_pos_percent = format!("{}%", (delta * 100.0).round());

                    canvas_style.set_property("position", "absolute").ok()?;
                    canvas_style.set_property("top", &new_pos_percent).ok()?;
                }

```

I am using the latest egui/eframe release: 0.19.0

Thank you in advance for investigating this.
Philipp-Sc

Philipp-Sc:
I updated my code to use the latest code from GitHub. The issue persists.

Philipp-Sc:
I added a workaround, to debug this issue, it can be reproduced using the following rev:
https://github.com/Philipp-Sc/librelearning/tree/0e7ad454f18fcf200b4a446a6b2246f6c409da56

The workaround consists of adding a CSS class with ```"top: 0% !important"``` to the canvas that enforces the position.
[...]

PR #4561, opened 2024-05-28
Improve web text agent

jprochazk:
- Closes https://github.com/emilk/egui/issues/4060 - no longer aligned to top
- Closes https://github.com/emilk/egui/issues/4479 - `canvas.style` is not set anywhere anymore
- Closes https://github.com/emilk/egui/issues/2231 - same as #4060
- Closes https://github.com/emilk/egui/issues/3618 - there is now one `<input>` per `eframe` app, and it's removed transitively by `WebRunner::destroy -> AppRunner::drop -> TextAgent::drop`

This PR improves the text agent to make fewer assumptions about how `egui` is embedded into the page:
- Text agent no longer sets the canvas position
- There is now a text agent for each instance of `WebRunner`
- The input element is now moved to the correct position, so the OS can display the IME window in the correct place. Before it would typically be outside of the viewport

The best way to test this is to build & server the web demo locally:
```
scripts/build_demo_web.sh && scripts/start_server.sh
```

Then open the EasyMark editor, and try using IME to input some emojis: http://localhost:8888/#EasyMarkEditor

To open the emoji keyboard use:
- <kbd>win + .</kbd> on Windows
- <kbd>ctrl + cmd + space</kbd> on Mac

Tested on:
- [x] Windows
- [x] Linux
- [x] MacOS
- [x] Android
- [x] iOS

## Migration guide

The canvas no longer controls its own size/position on the page. This means that those properties can now be controlled entirely via HTML and CSS, and multiple separate `eframe` apps can coexist better on a single page.

To match the old behavior, set the `canvas` width and height to 100% of the `body` element:

```html
<html>
  <body>
    <canvas></canvas>
  </body>
</html>
```

```css
/* remove default margins and use full viewport */
html, body {
  margin: 0;
  width: 100%;
  height: 100%;
}

canvas {
  /* match parent element size */
  width: 100%;
  height: 100%;
}
```

Note that there is no need to set `position: absolute`/`left: 50%; transform: translateX(-50%)`/etc., and setting those properties may poorly affect the sharpness of `egui`-rendered text.

Because `eframe` no longer updates the canvas style in any way, it also means that on mobile, the canvas no longer collapses upwards to make space for a mobile keyboard. This should be solved in other ways: https://github.com/emilk/egui/issues/4572

### DIFF
--- a/crates/eframe/src/web/text_agent.rs
+++ b/crates/eframe/src/web/text_agent.rs
@@ -1,237 +1,177 @@
-//! The text agent is an `<input>` element used to trigger
-//! mobile keyboard and IME input.
-//!
-use std::{cell::Cell, rc::Rc};
+//! The text agent is a hidden `<input>` element used to capture
+//! IME and mobile keyboard input events.
+
+use std::cell::Cell;
 
 use wasm_bindgen::prelude::*;
 
 use super::{AppRunner, WebRunner};
 
-static AGENT_ID: &str = "egui_text_agent";
-
-pub fn text_agent() -> web_sys::HtmlInputElement {
-    web_sys::window()
-        .unwrap()
-        .document()
-        .unwrap()
-        .get_element_by_id(AGENT_ID)
-        .unwrap()
-        .dyn_into()
-        .unwrap()
+pub struct TextAgent {
+    input: web_sys::HtmlInputElement,
+    prev_ime_output: Cell<Option<egui::output::IMEOutput>>,
 }
 
-/// Text event handler,
-pub fn install_text_agent(runner_ref: &WebRunner) -> Result<(), JsValue> {
-    let window = web_sys::window().unwrap();
-    let document = window.document().unwrap();
-    let body = document.body().expect("document should have a body");
-    let input = document
-        .create_element("input")?
-        .dyn_into::<web_sys::HtmlInputElement>()?;
-    let input = std::rc::Rc::new(input);
-    input.set_id(AGENT_ID);
-    let is_composing = Rc::new(Cell::new(false));
-    {
+impl TextAgent {
+    /// Attach the agent to the document.
+    pub fn attach(runner_ref: &WebRunner) -> Result<Self, JsValue> {
+        let document = web_sys::window().unwrap().document().unwrap();
+
+        // create an `<input>` element
+        let input = document
+            .create_element("input")?
+            .dyn_into::<web_sys::HtmlInputElement>()?;
+        input.set_type("text");
+
+        // append it to `<body>` and hide it outside of the viewport
         let style = input.style();
-        // Transparent
-        style.set_property("opacity", "0").unwrap();
-        // Hide under canvas
-        style.set_property("z-index", "-1").unwrap();
-    }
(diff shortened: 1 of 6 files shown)


Mm2PL:
I was asked to test this, works decently with ibus however there is a slight issue with the input being selected.
When I click-drag my mouse out of the input (like in selecting for example): I cannot use IME.
https://github.com/emilk/egui/assets/25011746/d88877ea-72b5-4c9b-8dc8-f87e4dfada00

jprochazk:
> I was asked to test this, works decently with ibus however there is a slight issue with the input being selected.

Thanks! I opened an issue for that: https://github.com/emilk/egui/issues/4563

jprochazk:
At the very least there don't seem to be any regressions here, and it's definitely improved some things, but also revealed just how broken input handling is on mobile web.

emilk:
Tested on iOS Safari, and when I click a text field that is in the lower half of the screen, the on-screen keyboard will now cover that text field, making it impossible to see what I'm typing. On `master`, the egui canvas would temporarily move up to prevent this problem.

So this is a regression that we should note down in an issue, ideally together with some ideas on how to solve it.

Still, I think a regression on mobile text input (which is already pretty broken anyways) is worth it in the interest of improving desktop web.

crates/eframe/src/web/text_agent.rs
```diff
@@ -1,237 +1,177 @@
...
+    }
 
-    Ok(())
-}
+    pub fn move_to(
+        &self,
+        ime: Option<egui::output::IMEOutput>,
+        canvas: &web_sys::HtmlCanvasElement,
+    ) -> Result<(), JsValue> {
+        // don't move the text agent on mobile
```
emilk, line 102: why?

emilk:
I'm testing bringing up the emoji editor on Mac on desktop chromium, and it will only _sometimes_ track the text field I have selected, but quite often go to the top-left corner.

I am testing using the TextEdit field in the WidgetGallery

crates/eframe/src/web/text_agent.rs
```diff
@@ -1,237 +1,177 @@
...
+        } else {
+            self.blur();
+        }
+    }
 
-                    // Note: `delta` is negative, since we are moving the canvas UP
-                    let delta = target_rel - current_rel;
+    pub fn has_focus(&self) -> bool {
+        let active_element = super::focused_element();
+        self.input.clone().dyn_into().ok() == active_element
```
emilk, line 136: This will return `true` if the left side errors (i.e. `.ok()` results in `None`)

emilk (approved):
So nice to get this cleaned up ❤️ 

But please open an issue describing the regression on mobile.

Are there any other likely regressions?

jprochazk:
Opened a bunch of issues and added them to
- https://github.com/emilk/egui/issues/4500

crates/eframe/src/web/text_agent.rs
jprochazk: The `<input>` is moved to allow IME "popups" to follow the text cursor, but there's no need for that on mobile. Updated the comment

PR #4562, opened 2024-05-28
Use Arc instead of Vec for cheaper PlotPoints clone

trueb2:
This PR makes it cheaper to create Line Plots. On each pass, I noticed that the PlotPoints::new interface used to copy 10s-100s of MiB, when we could be doing a cheap `Arc<[PlotPoint]>` clone instead of an expensive `Vec<PlotPoint>` clone.

It is related to https://github.com/emilk/egui/issues/1485, but as discussion has referenced, there are several aspects degrading rendering performance and this is not a silver bullet.

Construction is slightly different
```rust
let line_points =  x.zip(y)
    .map(|(x, y)| [x, y].into())
    .collect::<Vec<_>>()
    .into(); // Use the impl From<Vec<T>> for Arc<[T]>
 let points = vec![line_points];
 ```

Usage looks identical.
```rust
for series in points {
  ui.line(Line::new(PlotPoints::new(series.clone())));
}
```

If ExplicitGenerator were Clone, then we would be able to have PlotPoints be Clone.

### DIFF
--- a/crates/egui_plot/src/items/values.rs
+++ b/crates/egui_plot/src/items/values.rs
@@ -1,4 +1,7 @@
-use std::ops::{Bound, RangeBounds, RangeInclusive};
+use std::{
+    ops::{Bound, RangeBounds, RangeInclusive},
+    sync::Arc,
+};
 
 use egui::{Pos2, Shape, Stroke, Vec2};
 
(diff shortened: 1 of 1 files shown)


emilk:

See also:

Issue #2784, opened 2023-03-04
Log axes for plot


hacknus:
**Is your feature request related to a problem? Please describe.**
I want to be able to plot data on a logarithmic scale. Either just x, just y or both axes.

**Describe the solution you'd like**
It would be easiest to have an implementation that would only need one call in the constructor:
```rust
let my_plot = Plot::new("My Plot")
                    .height(height)
                    .width(width)
                    .log_axes([false, self.log_mode])
                    .legend(Legend::default());
```
to enable the log axes.

**Describe alternatives you've considered**
None

**Additional context**
I already started [here](https://github.com/hacknus/egui/tree/log_plot). 

<img width="358" alt="Bildschirm­foto 2023-03-04 um 23 56 16" src="https://user-images.githubusercontent.com/33124824/222932333-5f7b3e8c-0c30-4d85-abc1-5c919bc138cb.png">

Specifically, I started by adding the function 
```rust
pub fn log_axes(mut self, log_axes: [bool; 2]) -> Self {
    self.log_axes = log_axes;
    if self.log_axes[0] && self.log_axes[1] {
        let label_fmt = move |s: &str, val: &PlotPoint| {
            format!(
                "{}\n{:4.2}\n{:4.2}",
                s,
                10.0_f64.powf(val.x),
                10.0_f64.powf(val.y)
            )
        };
        let ax_fmt =
            move |x: f64, _range: &RangeInclusive<f64>| format!("{:4.2}", 10.0_f64.powf(x));
        self.label_formatter(label_fmt)
            .x_axis_formatter(ax_fmt)
            .y_axis_formatter(ax_fmt)
            .x_grid_spacer(logarithmic_grid_spacer(10))
            .y_grid_spacer(logarithmic_grid_spacer(10))
    } else if self.log_axes[0] {
        let label_fmt = move |s: &str, val: &PlotPoint| {
            format!("{}\n{:4.2}\n{:4.2}", s, 10.0_f64.powf(val.x), val.y)
        };
        let ax_fmt =
            move |y: f64, _range: &RangeInclusive<f64>| format!("{:4.2}", 10.0_f64.powf(y));
        self.label_formatter(label_fmt)
            .x_axis_formatter(ax_fmt)
            .x_grid_spacer(logarithmic_grid_spacer(10))
    } else if self.log_axes[1] {
        let label_fmt = move |s: &str, val: &PlotPoint| {
[...]

PR #4564, opened 2024-05-28
Example: semi-log and log-log plots

YgorSouza:
This has been requested a few times, such as #2784 and #4553.

As it is difficult to come up with an API that is clear and covers everyone's needs, I opted for an example for now, so people can modify it as they please. Eventually parts of the example can be moved to the egui_plot crate if desired.

### DIFF
--- a/examples/plot_log_scale/src/main.rs
+++ b/examples/plot_log_scale/src/main.rs
@@ -13,7 +13,7 @@ fn main() -> Result<(), eframe::Error> {
     eframe::run_native(
         "Plot",
         options,
-        Box::new(|_cc| Box::<PlotExample>::default()),
+        Box::new(|_cc| Ok(Box::<PlotExample>::default())),
     )
 }
 
--- a/examples/plot_log_scale/src/main.rs
+++ b/examples/plot_log_scale/src/main.rs
@@ -144,8 +144,8 @@ fn log_axis_formatter(gm: GridMark, max_size: usize, _bounds: &RangeInclusive<f6
     let size = digits + min_precision + 1;
     let value = 10.0f64.powf(gm.value);
     if size < max_size {
-        let precision = max_size.saturating_sub(digits + 1);
-        format!("{value:.precision$}")
+        let precision = max_size.saturating_sub(digits + 1).max(1);
+        egui::emath::format_with_decimals_in_range(value, 1..=precision)
     } else {
         let exp_digits = (digits as f64).log10() as usize;
         let precision = max_size.saturating_sub(exp_digits).saturating_sub(3);
--- a/examples/plot_log_scale/src/main.rs
+++ b/examples/plot_log_scale/src/main.rs
@@ -119,9 +119,16 @@ fn log_axis_spacer(input: GridInput) -> Vec<GridMark> {
     let (min, max) = input.bounds;
     let mut marks = vec![];
     for i in min.floor() as i32..=max.ceil() as i32 {
-        marks.extend((1..10).map(|j| GridMark {
-            value: i as f64 + (j as f64).log10(),
-            step_size: if j == 1 { 1.0 } else { 0.1 },
+        marks.extend((10..100).map(|j| {
+            let value = i as f64 + (j as f64).log10() - 1.0;
+            let step_size = if j == 10 {
+                1.0
+            } else if j % 10 == 0 {
+                0.1
+            } else {
+                0.01
+            };
+            GridMark { value, step_size }
         }));
     }
     marks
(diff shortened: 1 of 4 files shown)


examples/plot_log_scale/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+            }
+            plot.show(ui, |plot_ui| {
+                for signal in &self.signals {
+                    let points: Vec<_> = signal
+                        .points
+                        .iter()
+                        .copied()
+                        .map(|[x, y]| {
+                            let x = if log_x { x.log10() } else { x };
+                            let y = if log_y { y.log10() } else { y };
```
hyumo, line 106: when x or y is negative, `log10` is not valid and returns `NaN`

YgorSouza: Yes, normally you can't represent non-positive numbers on a log scale. Apparently there is a modified transform that allows that, but I think this starts to get too complicated for this example.

It is also noteworthy that the infinity value produced when we take the log of 0 causes the autobounds to stop working (it does not seem to be bothered by the NaNs though), so that is something that could be looked into, but in a separate PR.

hyumo: Yeah, thx, I am basically trying to do a bode plot, I will try to understand the example a bit and maybe do the transform on my end
https://en.wikipedia.org/wiki/Bode_plot

YgorSouza: Yeah, for a Bode plot the y axis is in dB, so you don't need the log scale for it.

examples/plot_log_scale/src/main.rs
```diff
@@ -0,0 +1,154 @@
...
+            });
+        });
+    }
+}
+
+#[allow(clippy::needless_pass_by_value)]
+fn log_axis_spacer(input: GridInput) -> Vec<GridMark> {
+    let (min, max) = input.bounds;
+    let mut marks = vec![];
+    for i in min.floor() as i32..=max.ceil() as i32 {
```
YgorSouza, line 121: I think the only thing missing is to make this spacer more general. Right now it only works well if the number of decades visible on the axis is between 0.5 and 15, roughly. It is a decent range for a log plot, but it would be nice if it could have the "infinite" zoom-in and zoom-out like the default grid. I haven't been able to figure that out yet.

ForsakenHarmony: I had implemented it on top of the original plot, here's what I came up with

```
pub fn log_grid_spacer(log_base: i64) -> GridSpacer {
    let log_base = log_base as f64;
    let step_sizes = move |input: GridInput| -> Vec<GridMark> {
        // The distance between two of the thinnest grid lines is "rounded" up
        // to the next-bigger power of base
        if input.log {
            let smallest_power = last_power(input.bounds.0, log_base)
                .log(log_base)
                .ceil()
                .max(0.0) as i32;
            let largest_power = next_power(input.bounds.1, log_base).log(log_base).ceil() as i32;

            let mut steps = vec![];
            for power in smallest_power..largest_power {
                let step_size = log_base.powi(power);
                let next_step_size = log_base.powi(power + 1);

                let bounds = (
                    step_size.max(input.bounds.0),
                    next_step_size.min(input.bounds.1),
                );

                if bounds.0 >= bounds.1 {
                    dbg!("oopsie");
                }

                fill_marks_between(&mut steps, step_size, bounds);
            }
            steps
        } else {
            let smallest_visible_unit = next_power(input.base_step_size, log_base);

            let step_sizes = [
                smallest_visible_unit,
                smallest_visible_unit * log_base,
                smallest_visible_unit * log_base * log_base,
            ];

            generate_marks(step_sizes, input.bounds)
        }
    };

    Box::new(step_sizes)
}

fn next_power(value: f64, base: f64) -> f64 {
    assert_ne!(value, 0.0); // can be negative (typical for Y axis)
    base.powi(value.abs().log(base).ceil() as i32)
}

fn last_power(value: f64, base: f64) -> f64 {
    assert_ne!(value, 0.0); // can be negative (typical for Y axis)
    base.powi(value.abs().log(base).floor() as i32)
}
```

YgorSouza: Thanks, I'll see if I can adapt this to what I was trying to do (since some of these egui_plot functions are private). I'm still not sure if this is better off as an example or as part of the API. Either way feels more complicated than it should be with all the options people might need.

foxcob:
This is something I was considering writing myself, but I was looking at a way to map the pixel coordinates to value through an arbitrary function (and back).  This accomplishes much of the same thing while keeping the actual plot linear.   What's your estimate on this being ready?

YgorSouza:
> This is something I was considering writing myself, but I was looking at a way to map the pixel coordinates to value through an arbitrary function (and back). This accomplishes much of the same thing while keeping the actual plot linear. What's your estimate on this being ready?

I'll see if I can look into it again this weekend. I think it's really only missing the grid. I was trying to come up with some code that is simple enough to be a useful example, but the code that generalizes to any range and zoom level is kind of hard to read and reason about.

Also, egui_plot is kind of in a limbo at the moment as it looks for new maintainers, so maybe it's better not to make too many PRs until that is sorted out.

emilk:

PR #4581, opened 2024-05-29
Bug fix: ui opacity and gray-out not affecting strokes

emilk:
Bug introduced in https://github.com/emilk/egui/pull/4353


murl-digital:
@emilk I don't think you don't need to put a box in an arc, because arc is backed by a box already

PR #4588, opened 2024-05-30
Introduce `UiStack`

### LINKED ISSUE #4534: Store a Ui/Frame-stack
Sometimes you want to answer the question "What is this Ui part of?". For instance: sometimes you want to paint the background of a widget to extend to the edges of the panel it is in, but that panel may be several `Ui`s removed from the current widget.

What you want is access to a stack of parent Ui frames, with some semantics.

Each stack should contain:

* `Id`
* `Frame` (inner & outer margins, stroke, …)
* `min_size` and `max_size` (with/without margins)
* Type (Panel, Area, Group, …)

We would push onto this stack each time we create a new `Ui`.

This means the root of the stack is created by `Ui::new` (most often created by `Area/Window` and `Panel`).
Then things like `ui.group`, `ui.horizontal`, `egui::Frame` etc would push onto this stack before calling into its contents, and then pop afterwards.

Working name: `StackFrame`.

Then we add `impl Ui { pub fn stack(&self) -> &Stack { … } }` and then users can walk the stack to implement advanced styling.

### Implementation
`Ui::new` creates the stack as a `Rc<RefCell<Stack>>` and hands it to its children.

### Future work: columns
We can use this architecture to implement columnar layouts, allowing aligning widgets that share an ancestor.
Each `StackFrame` has an optional field for column widths.

An `egui::Grid` would create such a special stack frame.
When adding a widget, we would walk up the stack to find which column we are in, and align to that, then register if we overflowed that width.
`ui.end_row()` would reset a column=0.

Maybe this should be a separate stack?  Probably.

Anyway, this requires more design, so let's not discuss this further here.

abey79:
* Closes #4534

WIP

TODO:
- [ ] bench mark
- [ ] add example to demo

### DIFF
--- a/crates/egui/src/ui.rs
+++ b/crates/egui/src/ui.rs
@@ -67,6 +67,9 @@ pub struct Ui {
 
     /// Indicates whether this Ui belongs to a Menu.
     menu_state: Option<Arc<RwLock<MenuState>>>,
+
+    /// The [`UiStack`] for this [`Ui`].
+    stack: Arc<UiStack>,
 }
 
 impl Ui {
@@ -77,17 +80,36 @@ impl Ui {
     ///
     /// Normally you would not use this directly, but instead use
     /// [`SidePanel`], [`TopBottomPanel`], [`CentralPanel`], [`Window`] or [`Area`].
-    pub fn new(ctx: Context, layer_id: LayerId, id: Id, max_rect: Rect, clip_rect: Rect) -> Self {
+    pub fn new(
+        ctx: Context,
+        layer_id: LayerId,
+        id: Id,
+        max_rect: Rect,
+        clip_rect: Rect,
+        stack_frame_data: UiStackInfo,
+    ) -> Self {
         let style = ctx.style();
+        let layout = Layout::default();
+        let placer = Placer::new(max_rect, layout);
+        let stack_frame = UiStack {
+            id,
+            layout_direction: layout.main_dir,
+            kind: stack_frame_data.kind,
+            frame: stack_frame_data.frame,
+            parent: None,
+            min_rect: placer.min_rect(),
+            max_rect: placer.max_rect(),
+        };
         let ui = Ui {
             id,
             next_auto_id_source: id.with("auto").value(),
             painter: Painter::new(ctx, layer_id, clip_rect),
             style,
-            placer: Placer::new(max_rect, Layout::default()),
+            placer,
             enabled: true,
             sizing_pass: false,
             menu_state: None,
+            stack: Arc::new(stack_frame),
         };
 
         // Register in the widget stack early, to ensure we are behind all widgets we contain:
@@ -105,16 +127,25 @@ impl Ui {
     }
 
     /// Create a new [`Ui`] at a specific region.
-    pub fn child_ui(&mut self, max_rect: Rect, layout: Layout) -> Self {
-        self.child_ui_with_id_source(max_rect, layout, "child")
+    ///
+    /// Note: calling this function twice from the same [`Ui`] will create a conflict of id. Use
+    /// [`Self::scope`] if needed.
+    pub fn child_ui(
+        &mut self,
+        max_rect: Rect,
+        layout: Layout,
+        stack_frame_data: Option<UiStackInfo>,
+    ) -> Self {
+        self.child_ui_with_id_source_and_frame_data(max_rect, layout, "child", stack_frame_data)
     }
 
     /// Create a new [`Ui`] at a specific region with a specific id.
-    pub fn child_ui_with_id_source(
+    pub fn child_ui_with_id_source_and_frame_data(
         &mut self,
         max_rect: Rect,
         mut layout: Layout,
         id_source: impl Hash,
+        stack_frame_data: Option<UiStackInfo>,
     ) -> Self {
         if self.sizing_pass {
             // During the sizing pass we want widgets to use up as little space as possible,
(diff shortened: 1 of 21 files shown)


abey79:
benchmark for `main`, "change" are w.r.t this PR.

```
demo_no_tessellate      time:   [100.03 µs 100.27 µs 100.55 µs]
                        change: [-0.9036% -0.4831% +0.0081%] (p = 0.04 < 0.05)
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe

demo_only_tessellate    time:   [63.612 µs 63.759 µs 63.910 µs]
                        change: [+0.0594% +0.6954% +1.3806%] (p = 0.04 < 0.05)
  3 (3.00%) high mild
  1 (1.00%) high severe

label &str              time:   [553.38 ns 618.15 ns 757.60 ns]
                        change: [-18.927% -3.9587% +12.659%] (p = 0.71 > 0.05)
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe

label format!           time:   [607.36 ns 723.32 ns 973.61 ns]
                        change: [-34.326% -6.3731% +32.513%] (p = 0.78 > 0.05)
Found 10 outliers among 100 measurements (10.00%)
  6 (6.00%) high mild
  4 (4.00%) high severe

Painter::rect           time:   [31.272 ns 31.625 ns 31.942 ns]
                        change: [-4.8340% -3.4710% -2.1191%] (p = 0.00 < 0.05)
                        Performance has improved.

text_layout_uncached    time:   [62.191 µs 62.366 µs 62.584 µs]
                        change: [+0.1487% +0.3898% +0.6197%] (p = 0.00 < 0.05)
  1 (1.00%) low mild
  2 (2.00%) high mild
  4 (4.00%) high severe

text_layout_cached      time:   [209.13 ns 209.47 ns 209.86 ns]
                        change: [-0.0661% +0.1518% +0.3733%] (p = 0.17 > 0.05)
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild

tessellate_text         time:   [8.3576 µs 8.4386 µs 8.5532 µs]
                        change: [+0.0148% +0.6121% +1.5481%] (p = 0.10 > 0.05)
Found 16 outliers among 100 measurements (16.00%)
  1 (1.00%) low mild
  8 (8.00%) high mild
  7 (7.00%) high severe
```

crates/egui/src/ui.rs
```diff
@@ -258,6 +303,11 @@ impl Ui {
         &mut self.style_mut().visuals
     }
 
+    /// Get a reference to this [`Ui`]'s [`UiStack`].
+    pub fn stack(&self) -> Arc<UiStack> {
+        self.stack.clone()
```
emilk, line 308:
```suggestion
    #[inline]
    pub fn stack(&self) -> &Arc<UiStack> {
        &self.stack
```

crates/egui/src/ui_stack.rs
```diff
@@ -0,0 +1,157 @@
...
+    /// Is this a root [`crate::Ui`], i.e. created with [`crate::Ui::new()`]?
+    #[inline]
+    pub fn is_root_ui(&self) -> bool {
+        self.parent.is_none()
+    }
+
+    /// This this [`crate::Ui`] a [`crate::Frame`] with a visible stroke?
+    #[inline]
+    pub fn has_visible_frame(&self) -> bool {
+        self.frame.stroke.width > 0.0 && self.frame.stroke.color != Color32::TRANSPARENT
```
emilk, line 118:
```suggestion
        !self.frame.stroke.is_empty()
```

crates/egui/src/ui_stack.rs
```diff
@@ -0,0 +1,157 @@
...
+    /// Return an iterator that walks the stack from this node to the root.
+    #[allow(clippy::iter_without_into_iter)]
+    pub fn iter(&self) -> UiStackIterator {
+        UiStackIterator {
+            next: Some(Arc::new(self.clone())),
+        }
+    }
+
+    /// Check if this node is or is contained in a [`crate::Ui`] of a specific kind.
+    pub fn contained_id(&self, kind: UiKind) -> bool {
```
emilk, line 133:
```suggestion
    pub fn contained_in(&self, kind: UiKind) -> bool {
```

crates/egui/src/ui_stack.rs
```diff
@@ -0,0 +1,157 @@
...
+/// See [`UiStack::iter`].
+pub struct UiStackIterator {
+    next: Option<Arc<UiStack>>,
+}
+
+impl Iterator for UiStackIterator {
+    type Item = Arc<UiStack>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let current = self.next.clone();
```
emilk, line 151: If you use references (and lifetimes) we can avoid this `.clone()`. I think this is likely to be on the hot path of quite a few layout thing in the future.

crates/egui/src/ui_stack.rs
```diff
@@ -0,0 +1,157 @@
...
+///
+/// See [`UiStack::iter`].
+pub struct UiStackIterator {
+    next: Option<Arc<UiStack>>,
+}
+
+impl Iterator for UiStackIterator {
+    type Item = Arc<UiStack>;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
emilk, line 150:
```suggestion
    #[inline]
    fn next(&mut self) -> Option<Self::Item> {
```

crates/egui_extras/src/layout.rs
```diff
@@ -195,9 +195,15 @@ impl<'l> StripLayout<'l> {
...
-            self.ui
-                .child_ui_with_id_source(rect, self.cell_layout, child_ui_id_source);
+        let mut child_ui = self.ui.child_ui_with_id_source(
+            rect,
+            self.cell_layout,
+            child_ui_id_source,
+            Some(egui::UiStackInfo {
+                kind: Some(egui::UiKind::TableCell),
+                frame: egui::Frame::default(),
+            }),
```
emilk, line 205: We could make this and other places shorter with a helper

```suggestion
            Some(egui::UiStackInfo::new(egui::UiKind::TableCell)),
```

examples/hello_world/src/main.rs
```diff
@@ -53,6 +53,12 @@ impl eframe::App for MyApp {
             ui.image(egui::include_image!(
                 "../../../crates/egui/assets/ferris.png"
             ));
+
+            ui.scope(|ui| {
+                ui.scope(|ui| {
+                    ui.label(format!("{:?}", ui.stack()));
+                });
+            });
```
emilk, line 61: Doesn't belong here

emilk (approved):
Looks good!

examples/hello_world/src/main.rs
abey79: whoops

Issue #1010, opened 2021-12-28
Context menu on hover / clickable tooltips


emilk:
It would be great to be able to bring up a context menu just by hovering an element. It would act similar to a tooltip, but a tooltip you can interact with (it won't go away when you move the mouse).

mankinskin:
For anyone looking into implementing this, this would probably be an abstraction over the trigger mechanism for bringing up a stationary menu, i.e. a menu created at some other element (represented by `response`):

https://github.com/emilk/egui/blob/2d2022fb7290bf4fcca373ad3a0dc98b680a9203/egui/src/menu.rs#L284-L308

the menu doesn't really have a mechanism for configuring it yet, but I suppose this could be easily added with some enums and a `MenuConfig` struct or something like that.

apexys:
It would also be interesting to be able to open a context menu with a left click or through some other action programmatically.

emilk:
This could also be implemented as "Tooltips stay put when you move the mouse onto them"

YgorSouza:
This could also be seen as a `Window` that is anchored to a `Response` and closes when you click outside it, like the solution proposed for #4463. This is called a `Popup` in other UI libraries, but egui already uses popup to refer to the ComboBox drawer.

PR #4596, opened 2024-05-31
Support interactive widgets in tooltips

emilk:
* Closes https://github.com/emilk/egui/issues/1010

### In short
You can now put interactive widgets, like buttons and hyperlinks, in an tooltip using `on_hover_ui`. If you do, the tooltip will stay open as long as the user hovers it.

There is a new demo for this in the egui demo app (egui.rs):

interactive-tooltips

### Design
Tooltips can now contain interactive widgets, such as buttons and links.
If they do, they will stay open when the user moves their pointer over them.

Widgets that do not contain interactive widgets disappear as soon as you no longer hover the underlying widget, just like before. This is so that they won't annoy the user.

To ensure not all tooltips with text in them are considered interactive, `selectable_labels` is `false` for tooltips contents by default. If you want selectable text in tooltips, either change the `selectable_labels` setting, or use `Label::selectable`.

```rs
ui.label("Hover me").on_hover_ui(|ui| {
    ui.style_mut().interaction.selectable_labels = true;
    ui.label("This text can be selected.");

    ui.add(egui::Label::new("This too.").selectable(true));
});
```

### Changes
* Layers in `Order::Tooltip` can now be interacted with

### DIFF
--- a/crates/egui_demo_lib/src/demo/misc_demo_window.rs
+++ b/crates/egui_demo_lib/src/demo/misc_demo_window.rs
@@ -264,7 +264,7 @@ impl Widgets {
                 ui.label("This tooltip was created with");
                 ui.monospace(".on_hover_ui(…)");
             });
-            let _ = ui.button("A button you can never press");
+            let _ = ui.button("A button you can press, but that does nothing.");
         };
         let disabled_tooltip_ui = |ui: &mut Ui| {
             ui.heading("Different tooltip when widget is disabled");
--- a/crates/egui_demo_lib/src/demo/misc_demo_window.rs
+++ b/crates/egui_demo_lib/src/demo/misc_demo_window.rs
@@ -233,7 +233,6 @@ fn label_ui(ui: &mut egui::Ui) {
 #[cfg_attr(feature = "serde", serde(default))]
 pub struct Widgets {
     angle: f32,
-    enabled: bool,
     password: String,
 }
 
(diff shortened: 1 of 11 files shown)


YgorSouza:
Being able to just copy text from a tooltip is super useful too. I do it all the time when coding:

image

I think it's more about having an easy way to toggle this "interactive tooltip" feature on and off for individual tooltips. Sometimes you really want it, and other times you really don't.

rustbasic:
My opinion is :
1. There is a way to provide both the current tooltip and a new interactive tooltip.
or
2. here is a way to select between the existing tooltip and the new interactive tooltip as an option.

YgorSouza:
Should probably update the text on this button in the demo:

https://github.com/emilk/egui/blob/78dfdb3684259ae1034c42c50473f353215389e3/crates/egui_demo_lib/src/demo/misc_demo_window.rs#L261-L268

emilk:
Good catch!

crates/egui_demo_lib/src/demo/misc_demo_window.rs
```diff
@@ -264,7 +264,7 @@ impl Widgets {
                 ui.label("This tooltip was created with");
                 ui.monospace(".on_hover_ui(…)");
             });
-            let _ = ui.button("A button you can never press");
+            let _ = ui.button("A button you can press, but that does nothing.");
```
YgorSouza, line 267: Since this demo already has an interactive and a non-interactive tooltip, maybe it could highlight the difference in the text as well, like "This tooltip stays open when the pointer is moved inside it or towards it, because it contains interactive widgets", and "This tooltip closes immediately when the pointer is moved, because it only contains non-interactive widgets" or something like that.

emilk:
Follow-up issue:
* https://github.com/emilk/egui/issues/4607

PR #4602, opened 2024-06-01
Add `auto_shrink()` function to `egui::Window`

rustbasic:
Add `auto_shrink()` function to `egui::Window`

### DIFF
--- a/crates/egui/src/containers/window.rs
+++ b/crates/egui/src/containers/window.rs
@@ -357,8 +357,6 @@ impl<'open> Window<'open> {
     }
 
     /// If there is [`ScrollArea`], For each axis, should the containing area shrink if the content is small?
-    ///
-    /// Show the [`ScrollArea::auto_shrink`]
     #[inline]
     pub fn auto_shrink(mut self, auto_shrink: impl Into<Vec2b>) -> Self {
         self.scroll = self.scroll.auto_shrink(auto_shrink);
--- a/crates/egui/src/containers/window.rs
+++ b/crates/egui/src/containers/window.rs
@@ -59,7 +59,7 @@ impl<'open> Window<'open> {
                 .with_stroke(false)
                 .min_size([96.0, 32.0])
                 .default_size([340.0, 420.0]), // Default inner size of a window
-            scroll: ScrollArea::neither(),
+            scroll: ScrollArea::neither().auto_shrink(false),
             collapsible: true,
             default_open: true,
             with_title_bar: true,
--- a/crates/egui/src/containers/window.rs
+++ b/crates/egui/src/containers/window.rs
@@ -356,7 +356,7 @@ impl<'open> Window<'open> {
         self
     }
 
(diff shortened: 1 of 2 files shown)


crates/egui/src/containers/window.rs
```diff
@@ -356,6 +356,15 @@ impl<'open> Window<'open> {
         self
     }
 
+    /// If there is [`ScrollArea`], For each axis, should the containing area shrink if the content is small?
+    ///
+    /// Show the [`ScrollArea::auto_shrink`]
+    #[inline]
+    pub fn auto_shrink(mut self, auto_shrink: impl Into<Vec2b>) -> Self {
+        self.scroll = self.scroll.auto_shrink(auto_shrink);
```
emilk, line 364: From the docs of `ScrollArea::auto_shrink`:

> * If `true`, egui will add blank space outside the scroll area.
> * If `false`, egui will add blank space inside the scroll area.

For a window we never want blank space outside the scroll area, so we should just change the default of `Window::scroll` to always have `.auto_shrink(false)`

rustbasic: I didn't fully understand what you meant.

This doesn't seem to be clearly explained in `ScrollArea::auto_shrink`.


If we want to change the default value of `egui::Window`,

1. we can either change the default value of `ScrollArea::auto_shrink` to `false`,

OR

2. add an `auto_shrink` variable to `egui::Window`.

I am unsure of which option is the best approach.

emilk: In `Window::new` do `scroll: ScrollArea::neither().auto_shrink(false)`

crates/egui/src/containers/window.rs
```diff
@@ -356,6 +356,15 @@ impl<'open> Window<'open> {
         self
     }
 
+    /// If there is [`ScrollArea`], For each axis, should the containing area shrink if the content is small?
+    ///
+    /// Show the [`ScrollArea::auto_shrink`]
+    #[inline]
+    pub fn auto_shrink(mut self, auto_shrink: impl Into<Vec2b>) -> Self {
```
emilk, line 363: and now you can remove this

crates/egui/src/containers/window.rs
```diff
@@ -356,6 +356,13 @@ impl<'open> Window<'open> {
         self
     }
 
+    /// If there is [`ScrollArea`], For each axis, should the containing area shrink if the content is small?
+    #[inline]
+    pub fn auto_shrink(mut self, auto_shrink: impl Into<Vec2b>) -> Self {
+        self.scroll = self.scroll.auto_shrink(auto_shrink);
+        self
+    }
+
```
emilk, line 365:
```suggestion
```

emilk (changes requested):

crates/egui/src/containers/window.rs
rustbasic: The new code you suggested is the same as the original one.  There seems to be no changes in the suggestion.

PR #4606, opened 2024-06-03
Update README to maintained Tetra integration

trevyn:
Original integration https://github.com/tesselode/egui-tetra is archived, so I forked and updated for latest `egui` and `tetra`.


emilk:
Thasnks!

PR #4608, opened 2024-06-04
async functionality

uglyoldbob:
This pull request adds some async functionality to egui. I have a library I have been working on that uses the modifications at [this repository](https://github.com/uglyoldbob/egui-multiwin-async).
I wanted to get some feedback on how reasonable my implementation is.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -386,6 +386,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-winit"
+version = "0.2.1"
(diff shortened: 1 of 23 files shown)


emilk:
Can you give a high-level overview what this is about?

uglyoldbob:
This adds functionality to allow creating an entirely async gui application based on egui. I created an async version of egui glow, and tweaked egui a bit to add some of the relevant needed async functions.

emilk:
I'm sorry, but I don't want to maintain 6kloc more of code. I think you should consider publishing this a 3rd party repository instead!

Issue #4349, opened 2024-04-11
Window contents can sometimes be affected by mouse interaction outside the window


s-nie:
egui version 0.27.2, also present on master

**Describe the bug**

Interactions outside a window can sometimes interact with elements inside the window if they allocate enough space.

**To Reproduce**

```rust
use eframe::{
    egui::{self, vec2},
    run_native, App,
};

struct TestApp;

impl App for TestApp {
    fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
        ctx.set_debug_on_hover(true);
        egui::Window::new("test")
            .scroll(true)
            .resizable(true)
            .title_bar(false)
            .show(ctx, |ui| {
                let _ = ui.allocate_space(vec2(1000.0, 100.0));
                ui.label("hello");
            });
    }
}

fn main() {
    run_native(
        "Test",
        Default::default(),
        Box::new(|_cc| Ok(Box::new(TestApp))),
    )
    .unwrap();
}
```

**Expected behavior**

The drag affects whatever is inside the window.

**Screenshots**

https://github.com/emilk/egui/assets/73159570/179168fe-465c-4bb9-96d7-f03f945d97a9

rustbasic:
@emilk

It would be good for emilk to fix this.

This is a problem that has existed for a long time.

The area of `egui::Window` is recognized as large, so the buttons below that area cannot be clicked.

It will be easier to check the problem using the example below.

```rust
struct TestApp;

impl eframe::App for TestApp {
    fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
        egui::Window::new("test")
            .scroll(true)
            .resizable(true)
            .show(ctx, |ui| {
                ui.collapsing(
                    "looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong",
                    |_ui| {},
                );
            });

        egui::CentralPanel::default().show(ctx, |ui| {
            ctx.options_mut(|options| {
                options.tessellation_options.debug_paint_clip_rects = true;
            });
            ctx.settings_ui(ui);
        });
    }
}

fn main() {
    let _result = eframe::run_native(
        "Test",
        Default::default(),
        Box::new(|_cc| Box::new(TestApp)),
    );
}
```

rustbasic:
@emilk

I will solve this and submit a Pull Request.

s-nie:
Thanks for taking a shot at this. I tried it and it does fix the issue.

[...]

PR #4611, opened 2024-06-05
ScrollArea: Prevent drag interaction outside the area

s-nie:
* Closes https://github.com/emilk/egui/issues/4349

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -473,6 +473,9 @@ struct Prepared {
     scrolling_enabled: bool,
     stick_to_end: Vec2b,
     animated: bool,
+
+    /// Id of the drag area. We store this until we know the size of the content to interact with.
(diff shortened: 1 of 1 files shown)


emilk (changes requested):
I tested it, and it creates a bunch of bugs.

Try clicking around on the right side panel in the egui demo app, for instance.

I think the problem is that the late `interact` call means the interaction is _on top_ of the contents, while it should be _below_. So my idea wasn't great.

Better to store the final size for the next frame afterall

s-nie:
>I think the problem is that the late interact call means the interaction is on top of the contents, while it should be below. 

Right, I had not considered this. I changed the approach in https://github.com/emilk/egui/pull/4611/commits/4e1ea29c97ecfd7bdc026fdacecac8f2324b6679, the demo app works fine for me.

emilk (approved):

emilk:
Thanks!

PR #4617, opened 2024-06-05
Add tags to `UiStack`

emilk:
You can now set custom tags on the `UiStack`. This allows you to write code that is situationally aware at runtime. For instance, you could decide wether or not a label should truncate its text depending on what part of your ui it is in, without having to pass that info down via the callstack.


emilk:
I should note in a comment that the tags only live for the duration of the current frame

PR #4620, opened 2024-06-05
Gif support

### LINKED ISSUE #4489: Gif Support

JustFrederik:
* Previous PR: #3951 
* Closes #4489


YgorSouza:
Could add `Closes #4489` to the description so we don't forget to close the issue afterward.

crates/egui/src/widgets/image.rs
```diff
@@ -769,3 +785,28 @@ pub fn paint_texture_at(
         }
     }
 }
+
+fn get_index(ctx: &Context, uri: &str) -> usize {
+    let now = ctx.input(|i| Duration::from_secs_f64(i.time));
+    let durations: Option<Arc<Vec<Duration>>> =
+        ctx.data(|data| data.get_temp(Id::new(format!("{uri}-index"))));
```
YgorSouza, line 792: Maybe you could use `Id::new((uri, "index"))` to avoid allocating a `String` each time? Or maybe even an enum type to generate image data Ids.

```rs
#[derive(Clone, Copy, PartialEq, Eq, Hash)]
pub enum ImageDataId {
    Index,
}

impl ImageDataId {
    #[inline]
    pub fn id(self, url: &str) -> Id {
        Id::new((self.type_id(), self, url))
    }
}
```

And then

```rs
ImageDataId::Index.id(url)
```

So there is less risk of bugs due to using different Ids in different parts of the code. Maybe that's overkill though.

JustFrederik: @YgorSouza I didn't see the point of using a enum. Is there a reason why enum & not a struct?

crates/egui/src/widgets/image.rs
```diff
@@ -288,8 +288,24 @@ impl<'a> Image<'a> {
     }
 
     #[inline]
-    pub fn source(&self) -> &ImageSource<'a> {
-        &self.source
+    pub fn source(&'a self, ctx: &Context) -> ImageSource<'a> {
+        match &self.source {
+            ImageSource::Bytes { uri, .. } => match uri.starts_with("gif://") {
```
emilk, line 293: Instead of using a special `gif://` URI, you can check for the GIF magic header, i.e. 

```rs
let is_gif = bytes.starts_with(b"GIF87a") || bytes.starts_with(b"GIF89a");
```

That means a user can use `include_image!` for gifs too, with no need for the special `include_gif!` macro

crates/egui/src/widgets/image.rs
```diff
@@ -288,8 +288,24 @@ impl<'a> Image<'a> {
...
+                true => Some(uri),
+                false => None,
+            },
+            _ => None,
+        }
+        .map(|v| format!("{}-{}", v, get_index(ctx, v)))
```
emilk, line 299: This special format of `{uri}-{frame_index}` should be better documented.

I suggest adding two functions to this file: one for encoding it, one for decoding it (the latter to be used by `gif_loader.rs`).

crates/egui/src/widgets/image.rs
```diff
@@ -288,8 +288,24 @@ impl<'a> Image<'a> {
...
     #[inline]
-    pub fn source(&self) -> &ImageSource<'a> {
-        &self.source
+    pub fn source(&'a self, ctx: &Context) -> ImageSource<'a> {
+        match &self.source {
+            ImageSource::Bytes { uri, .. } => match uri.starts_with("gif://") {
+                true => Some(uri),
+                false => None,
+            },
+            _ => None,
```
emilk, line 297: We should be able to handle `ImageSource::Uri` here too, for showing images from e.g. a HTTP url (but we don't need to squeeze it in to this PR)

crates/egui/src/widgets/image.rs
```diff
@@ -288,8 +288,24 @@ impl<'a> Image<'a> {
...
+        .map(|v| format!("{}-{}", v, get_index(ctx, v)))
+        .map(|v| match &self.source {
+            ImageSource::Uri(_) => ImageSource::Uri(Cow::Owned(v)),
+            ImageSource::Texture(v) => ImageSource::Texture(*v),
+            ImageSource::Bytes { bytes, .. } => ImageSource::Bytes {
+                uri: Cow::Owned(v),
+                bytes: bytes.clone(),
+            },
+        })
+        .unwrap_or(self.source.clone())
```
emilk, line 308: nit: this function is hard to follow imho. Something like this should work:

```rs
match &self.source {
    ImageSource::Uri(uri) if is_gif_uri(uri) => {
        let frame_uri = format!("{}-{}", v, gif_frame_id(ctx, v));
        ImageSource::Uri(frame_uri)
    }

    ImageSource::Bytes { uri, bytes } if is_gif_uri(uri) || has_gif_magic_header(bytes) => {
        let frame_uri = format!("{}-{}", v, get_index(ctx, v));
        ImageSource::Bytes {
            uri: Cow::Owned(frame_uri),
            bytes: bytes.clone(),
        },
    }

    _ => self.source.clone()
}
```

crates/egui/src/widgets/image.rs
```diff
@@ -769,3 +785,39 @@ pub fn paint_texture_at(
...
+        let mut index = 0;
+        for (i, duration) in durations.iter().enumerate() {
+            cumulative_duration += duration.as_millis();
+            if cumulative_duration >= pos {
+                index = i;
+                break;
+            }
+        }
+        if let Some(duration) = durations.get(index) {
+            ctx.request_repaint_after(*duration);
```
emilk, line 807: This is likely to lead to us skipping frames.

If the frames are 20ms each, and we're at time=6ms, then we should schedule a repaint after 14ms.

crates/egui/src/widgets/image.rs
```diff
@@ -769,3 +785,39 @@ pub fn paint_texture_at(
...
+        }
+        index
+    } else {
+        0
+    }
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct ImageDataIdIndex;
```
emilk, line 816: I suggest this:


```suggestion
/// The number of frames in a GIF, and their individual durations.
#[derive(Clone)]
pub struct GifFrameDurations(pub Arc<Vec<Duration>>);
```

and then you put that into the data store, meaning you know that you put the exact correct type into the data store that you later pull out.

crates/egui/src/widgets/image.rs
```diff
@@ -769,3 +785,39 @@ pub fn paint_texture_at(
         }
     }
 }
+
+fn get_index(ctx: &Context, uri: &str) -> usize {
```
emilk, line 789:
```suggestion
fn gif_frame_index(ctx: &Context, uri: &str) -> usize {
```

crates/egui/src/widgets/image.rs
```diff
@@ -769,3 +785,39 @@ pub fn paint_texture_at(
...
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct ImageDataIdIndex;
+
+impl ImageDataIdIndex {
+    #[inline]
+    pub fn id(self, uri: &str) -> Id {
+        Id::new((std::any::TypeId::of::<Self>(), self, uri))
+    }
+}
```
emilk, line 823:
```suggestion
```

If you use my `GifFrameDurations` idea for storing the data in a typesafe way, the `IdTypeMap` will add the type as part of the key for you, so you can just use `Id::new(uri)` as the key

crates/egui_extras/src/loaders/gif_loader.rs
```diff
@@ -0,0 +1,139 @@
...
+    ColorImage, ImageDataIdIndex,
+};
+use image::AnimationDecoder as _;
+use std::{io::Cursor, mem::size_of, sync::Arc, time::Duration};
+
+/// Array of Frames and the duration for how long each frame should be shown
+#[derive(Debug, Clone)]
+pub struct AnimatedImage {
+    frames: Vec<Arc<ColorImage>>,
+    delays: Arc<Vec<Duration>>,
```
emilk, line 14:
```suggestion
    frame_durations: GifFrameDurations,
```

crates/egui_extras/src/loaders/gif_loader.rs
```diff
@@ -0,0 +1,139 @@
...
+                .iter()
+                .map(|image| {
+                    image.pixels.len() * size_of::<egui::Color32>() + size_of::<Duration>()
+                })
+                .sum::<usize>()
+    }
+
+    /// Gets image at index
+    pub fn get_image(&self, index: usize) -> Arc<ColorImage> {
+        self.frames.get(index % self.frames.len()).cloned().unwrap()
```
emilk, line 31: `.get(…).unwrap()` is just `[…]` with extra steps
```suggestion
        self.frames[index % self.frames.len()].clone()
```

crates/egui_extras/src/loaders/gif_loader.rs
```diff
@@ -0,0 +1,139 @@
...
+    fn load(&self, ctx: &egui::Context, uri_data: &str, _: SizeHint) -> ImageLoadResult {
+        if !is_supported_uri(uri_data) {
+            return Err(LoadError::NotSupported);
+        }
+        let (uri, index) = uri_data
+            .rsplit_once('-')
+            .ok_or(LoadError::Loading("No -{index} at end of uri".to_owned()))?;
+        let index: usize = index
+            .parse()
+            .map_err(|_err| LoadError::Loading("Failed to parse index".to_owned()))?;
```
emilk, line 86: This is what I suggest putting into egui proper, so that the encoding of this scheme is next to the decoding of it

emilk (commented):
This is looking very promising! A a great test gif too :)

crates/egui/src/widgets/image.rs
emilk: I would also consider using `#` here for the separator. That would be more in line with the `#fragment` part of the URI spec: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier

JustFrederik:
@emilk every requested change is implemented. I didn't test the file_loader, but it should behave the exactly like the ehttp loader which works.

crates/egui_extras/src/loaders/gif_loader.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+impl ImageLoader for GifLoader {
+    fn id(&self) -> &str {
+        Self::ID
+    }
+
+    fn load(&self, ctx: &egui::Context, uri_data: &str, _: SizeHint) -> ImageLoadResult {
+        let uri_index = decode_gif_uri(uri_data).map_err(|e| LoadError::Loading(e.to_owned()));
+        let uri = uri_index.as_ref().map(|v| v.0).unwrap_or(uri_data);
```
emilk, line 76: Let's use clear names

```suggestion
    fn load(&self, ctx: &egui::Context, frame_uri: &str, _: SizeHint) -> ImageLoadResult {
        let (image_uri, frame_index) = decode_gif_uri(frame_uri).map_err(|_| Err(LoadError::NotSupported))?;
```

crates/egui_extras/src/loaders/gif_loader.rs
```diff
@@ -0,0 +1,134 @@
...
+        if let Some(entry) = cache.get(uri).cloned() {
+            let index = uri_index?.1;
+            match entry {
+                Ok(image) => Ok(ImagePoll::Ready {
+                    image: image.get_image(index),
+                }),
+                Err(err) => Err(LoadError::Loading(err)),
+            }
+        } else {
+            match ctx.try_load_bytes(uri_data) {
```
emilk, line 87:
```suggestion
            match ctx.try_load_bytes(image_uri) {
```

emilk (changes requested):

crates/egui/src/widgets/image.rs
```diff
@@ -40,6 +40,8 @@ use crate::{
 ///     .paint_at(ui, rect);
 /// # });
 /// ```
+///
+const RENDER_TIME: Duration = Duration::from_millis(6);
```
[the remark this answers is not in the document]
JustFrederik: I created it because of
`This is likely to lead to us skipping frames.

If the frames are 20ms each, and we're at time=6ms, then we should schedule a repaint after 14ms.`

I thought adding a magic number in the code that might change in the future wouldn't make that much sense

JustFrederik:
@emilk is this ok?

JustFrederik:
@emilk ??

crates/egui_extras/src/loaders/gif_loader.rs
```diff
@@ -0,0 +1,135 @@
...
+    fn id(&self) -> &str {
+        Self::ID
+    }
+
+    fn load(&self, ctx: &egui::Context, frame_uri: &str, _: SizeHint) -> ImageLoadResult {
+        let uri_index = decode_gif_uri(frame_uri).map_err(|e| LoadError::Loading(e.to_owned()));
+        let image_uri = uri_index.as_ref().map(|v| v.0).unwrap_or(frame_uri);
+        let mut cache = self.cache.lock();
+        if let Some(entry) = cache.get(image_uri).cloned() {
+            let index = uri_index?.1;
```
emilk, line 79: It is unclear wether or not this code is intended to support only URI:s pointing to specific frames, or if it should _also_ support URI:s pointing to the whole gif.

If the latter, please add comments explaining what that means and should result in.
If the format, please simplify the code by unwrapping failure to parse the frame index right away:

Assuming this code should 

```suggestion
        let (image_uri, frame_index) = decode_gif_uri(frame_uri).map_err(|e| LoadError::Loading(e.to_owned()))?;
        let mut cache = self.cache.lock();
        if let Some(entry) = cache.get(image_uri).cloned() {
```

crates/egui_extras/src/loaders/gif_loader.rs
```diff
@@ -0,0 +1,135 @@
...
+            match entry {
+                Ok(image) => Ok(ImagePoll::Ready {
+                    image: image.get_image(index),
+                }),
+                Err(err) => Err(LoadError::Loading(err)),
+            }
+        } else {
+            match ctx.try_load_bytes(image_uri) {
+                Ok(BytesPoll::Ready { bytes, .. }) => {
+                    let is_gif = bytes.starts_with(b"GIF87a") || bytes.starts_with(b"GIF89a");
```
emilk, line 89: Call `has_gif_magic_header` instead of repeating code?

crates/egui_extras/src/loaders/gif_loader.rs
```diff
@@ -0,0 +1,135 @@
...
+                    if !is_gif {
+                        return Err(LoadError::NotSupported);
+                    }
+                    let index = uri_index?.1;
```
emilk, line 93: "index" is very generic. Index of what? Something like "frame_index" is more descriptive

crates/egui_extras/src/loaders/gif_loader.rs
```diff
@@ -0,0 +1,135 @@
...
+#[derive(Default)]
+pub struct GifLoader {
+    cache: Mutex<HashMap<String, Entry>>,
+}
+
+impl GifLoader {
+    pub const ID: &'static str = egui::generate_loader_id!(GifLoader);
+}
+
+pub fn gif_to_sources(data: Bytes) -> Result<AnimatedImage, String> {
```
emilk, line 46: This could be a member of `AnimatedImage`, e.g. `AnimatedImage::load_gif`

crates/egui_extras/src/loaders/gif_loader.rs
```diff
@@ -0,0 +1,135 @@
...
+    let decoder = image::codecs::gif::GifDecoder::new(Cursor::new(data))
+        .map_err(|_err| "Couldnt decode gif".to_owned())?;
+    let mut images = vec![];
+    let mut durations = vec![];
+    for frame in decoder.into_frames() {
+        let frame = frame.map_err(|_err| "Couldnt decode gif".to_owned())?;
```
emilk, line 52: Don't throw away an error message

```suggestion
        let frame = frame.map_err(|err| format!("Failed to decode gif: {err}"))?;
```

crates/egui_extras/src/loaders/gif_loader.rs
```diff
@@ -0,0 +1,135 @@
...
+    cache: Mutex<HashMap<String, Entry>>,
+}
+
+impl GifLoader {
+    pub const ID: &'static str = egui::generate_loader_id!(GifLoader);
+}
+
+pub fn gif_to_sources(data: Bytes) -> Result<AnimatedImage, String> {
+    let decoder = image::codecs::gif::GifDecoder::new(Cursor::new(data))
+        .map_err(|_err| "Couldnt decode gif".to_owned())?;
```
emilk, line 48: Don't throw away the error

crates/egui/src/widgets/image.rs
```diff
@@ -769,3 +782,56 @@ pub fn paint_texture_at(
...
+    Ok((uri, index))
+}
+
+/// checks if uri is a gif file or starts with gif://
+fn is_gif_uri(uri: &str) -> bool {
+    uri.ends_with(".gif")
+}
+
+/// checks if bytes are gifs
+fn has_gif_magic_header(bytes: &Bytes) -> bool {
```
emilk, line 808:
```suggestion
fn has_gif_magic_header(bytes: &[u8]) -> bool {
```

crates/egui/src/widgets/image.rs
```diff
@@ -769,3 +782,56 @@ pub fn paint_texture_at(
...
+    let (uri, index) = uri
+        .rsplit_once('#')
+        .ok_or("Failed to find index seperator '#'")?;
+    let index: usize = index.parse().map_err(|_err| "Failed to parse index")?;
+    Ok((uri, index))
+}
+
+/// checks if uri is a gif file or starts with gif://
+fn is_gif_uri(uri: &str) -> bool {
+    uri.ends_with(".gif")
```
emilk, line 804: misleading docstring.

Also: should this handle `foo.gif#42` ?

crates/egui/src/widgets/image.rs
```diff
@@ -769,3 +782,56 @@ pub fn paint_texture_at(
...
+    let now = ctx.input(|i| Duration::from_secs_f64(i.time));
+
+    let durations: Option<GifFrameDurations> = ctx.data(|data| data.get_temp(Id::new(uri)));
+    if let Some(durations) = durations {
+        let frames: Duration = durations.0.iter().sum();
+        let pos_ms = now.as_millis() % frames.as_millis().max(1);
+        let mut cumulative_ms = 0;
+        for (i, duration) in durations.0.iter().enumerate() {
+            cumulative_ms += duration.as_millis();
+            if pos_ms <= cumulative_ms {
```
emilk, line 823:
```suggestion
            if pos_ms < cumulative_ms {
```

emilk (commented):
Veery close now - just a few cleanups and we should be good to go

emilk:
btw, as I mention in the PR template, opening a PR from your `master` branch is problematic. For one, it makes it much harder for me to test it.

crates/egui_extras/src/loaders/gif_loader.rs
JustFrederik: LoadError::Loading can't be used because there is no check if the image is supported before. I changed it so if the pattern isn't matched it isn't a supported image

JustFrederik:
@emilk Im sorry. I will keep that in mind for the next pull request

crates/egui/src/widgets/image.rs
```diff
@@ -769,3 +782,58 @@ pub fn paint_texture_at(
...
+/// extracts uri and frame index
+/// # Errors
+/// Will return `Err` if `uri` does not match pattern {uri}-{frame_index}
+pub fn decode_gif_uri(uri: &str) -> Result<(&str, usize), String> {
+    let (uri, index) = uri
+        .rsplit_once('#')
+        .ok_or("Failed to find index separator '#'")?;
+    let index: usize = index
+        .parse()
+        .map_err(|_err| format!("Failed to parse index: \"{index}\""))?;
```
emilk, line 800:
```suggestion
        .map_err(|_err| format!("Failed to parse gif frame index: {index:?} is not an integer"))?;
```

crates/egui_extras/src/loaders/gif_loader.rs
```diff
@@ -0,0 +1,134 @@
...
+
+/// Array of Frames and the duration for how long each frame should be shown
+#[derive(Debug, Clone)]
+pub struct AnimatedImage {
+    frames: Vec<Arc<ColorImage>>,
+    frame_durations: GifFrameDurations,
+}
+
+impl AnimatedImage {
+    fn load_gif(data: Bytes) -> Result<Self, String> {
```
emilk, line 19:
```suggestion
    fn load_gif(data: &[u8]) -> Result<Self, String> {
```

emilk (approved):
Code looks good!

I just need to give it a whirl before merging

crates/egui_extras/Cargo.toml
```diff
@@ -49,6 +49,9 @@ http = ["dep:ehttp"]
 ## ```
 image = ["dep:image"]
 
+## Support loading gif images.
+gif = ["image", "image/gif"]
```
emilk, line 53: Please move up to keep the features sorted by name

examples/images/Cargo.toml
```diff
@@ -16,7 +16,7 @@ eframe = { workspace = true, features = [
   "default",
   "__screenshot", # __screenshot is so we can dump a screenshot using EFRAME_SCREENSHOT_TO
 ] }
-egui_extras = { workspace = true, features = ["default", "all_loaders"] }
+egui_extras = { workspace = true, features = ["default", "all_loaders", "gif"] }
```
emilk, line 19: I suggest we add `gif` to `all_loaders` instead

examples/images/src/main.rs
```diff
@@ -27,6 +27,7 @@ impl eframe::App for MyApp {
     fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
         egui::CentralPanel::default().show(ctx, |ui| {
             egui::ScrollArea::both().show(ui, |ui| {
+                ui.image(egui::include_image!("ferris.gif"));
```
emilk, line 30: It's super cool that this Just Works™️

emilk (approved):
Awesome work - this is really cool!

JustFrederik:
should be ready to merge now

emilk:
Thanks for going through all the review rounds on this!

PR #4627, opened 2024-06-06
Add `Color32::lerp_to_gamma`

abey79:
Add `lerp_to_gamma` utility function to `Color32`

### DIFF
--- a/crates/ecolor/src/color32.rs
+++ b/crates/ecolor/src/color32.rs
@@ -241,10 +241,10 @@ impl Color32 {
         use emath::lerp;
 
         Self::from_rgba_premultiplied(
-            lerp((self[0] as f32)..=(other[0] as f32), t).round() as u8,
-            lerp((self[1] as f32)..=(other[1] as f32), t).round() as u8,
-            lerp((self[2] as f32)..=(other[2] as f32), t).round() as u8,
-            lerp((self[3] as f32)..=(other[3] as f32), t).round() as u8,
+            fast_round(lerp((self[0] as f32)..=(other[0] as f32), t)),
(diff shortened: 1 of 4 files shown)


crates/ecolor/src/color32.rs
```diff
@@ -235,4 +235,16 @@ impl Color32 {
             a as f32 / 255.0,
         ]
     }
+
+    /// Lerp this color towards `other` by `t` in gamma space.
```
abey79, line 239: Docstring kindly provided by Copilot. I'm still largely confused by the linear vs. gamma distinction.

crates/ecolor/src/color32.rs
```diff
@@ -235,4 +235,16 @@ impl Color32 {
...
+    pub fn lerp_to_gamma(&self, other: Self, t: f32) -> Self {
+        use emath::lerp;
+
+        Self::from_rgba_premultiplied(
+            lerp((self[0] as f32)..=(other[0] as f32), t).round() as u8,
+            lerp((self[1] as f32)..=(other[1] as f32), t).round() as u8,
+            lerp((self[2] as f32)..=(other[2] as f32), t).round() as u8,
+            lerp((self[3] as f32)..=(other[3] as f32), t).round() as u8,
```
emilk, line 247: While we're at it (`.round()` is a lot slower, last I checked)

```suggestion
            fast_round(lerp((self[0] as f32)..=(other[0] as f32), t)),
            fast_round(lerp((self[1] as f32)..=(other[1] as f32), t)),
            fast_round(lerp((self[2] as f32)..=(other[2] as f32), t)),
            fast_round(lerp((self[3] as f32)..=(other[3] as f32), t)),
```

emilk (approved):

PR #4630, opened 2024-06-06
Easing functions

emilk:
This adds most of the "standard" easing functions from https://easings.net/ to `emath::easing`, and adds helpers in `egui` for using them.

In particular there is now `ctx.animate_bool_with_easing` and `ctx.animate_bool_responsive`, that uses a cubic easing function.

All animations in egui now uses cubic ease-out, for a more responsive feeling (fast at the start, slower towards the end).


aevyrie:
Closes #2717

Issue #4039, opened 2024-02-13
Vertical Centered expands to infinity


DasLixou:
**Describe the bug**
When having two vertical_centered stacks horizontally next to each other, they grow bigger on every mouse move. (could be a duplicate of #2815 but it's described differently so I'm not sure)

**To Reproduce**
```rs
ui.horizontal(|ui| {
    ui.vertical_centered(|ui| {
        ui.label("Volume");
        ui.label("Text");
    });
    ui.vertical_centered(|ui| {
        ui.label("Pan");
        ui.label("Other Text");
    });
}
```

**Expected behavior**
Having the minimal size of them next by each other.

**Screenshots**
https://github.com/emilk/egui/assets/82600264/fbe3173e-31f8-4b05-8ba8-b6b71dbe1202

DasLixou:
> It would be a little better if you added `.scroll2([true, true])` to the options of egui::Window.
> 
> ```rust
>                  .scroll2([true, true])
> ```

why? I don't want to scroll behind my content, it should be in the small window

rustbasic:
I mean temporary.

```rust
                    ui.horizontal(|ui| {
                        ui.vertical_centered(|ui| {
                            ui.label("Volume");
                            ui.label("Text");
                        });
                    });
                    ui.horizontal(|ui| {
                        ui.vertical_centered(|ui| {
                            ui.label("Pan");
                            ui.label("Other Text");
                        });
                    });
```

DasLixou:
But scrolling wouldn't help because it would grow larger every time I update, so also on scroll.
And having two horizontal things there would just make it stacked vertically, right? meh.

rustbasic:
> But scrolling wouldn't help because it would grow larger every time I update, so also on scroll.
> And having two horizontal things there would just make it stacked vertically, right? meh.

Yes, you will have to find a temporary solution and use it until the bug is fixed.

mattfbacon:
Would [Ui::columns](https://docs.rs/egui/latest/egui/struct.Ui.html#method.columns) be a solution here?

lucasmerlin:
This is just not something that egui's single pass layout supports. The first `vertical_centered` will take the full available width so it can center the contents, so the second `vertical_centered` will grow the parent ui.

There are many solutions to this problem though like `Ui::columns`, `egui_extras::StripBuilder`, `egui_flex`.

Issue #3449, opened 2023-10-08
plots no longer start next line after filling available horizontal space since 0.23


rukai:
Possibly related to https://github.com/emilk/egui/issues/3054

**Describe the bug**
In egui 0.22 plots put in a horizontal_wrapped ui wrap horizontally as expected.
![image](https://github.com/emilk/egui/assets/5120858/4220c257-9791-4a51-9488-c63d1a8cc2e6)

Since egui 0.23 plots now continue on forever without wrapping, they are instead pushed off screen.
![image](https://github.com/emilk/egui/assets/5120858/7817a327-57c2-4a4e-8c64-aeaecfabde5a)

**To Reproduce**
Put this code in the eframe template
```rust
            ui.horizontal_wrapped(|ui| {
                for i in 0..10 {
                    let plot = Plot::new(format!("foo{i}"))
                        .legend(Legend::default())
                        .width(500.0)
                        .height(250.0)
                        .allow_scroll(false);

                    plot.show(ui, |_| {});
                }
            });
```

**Desktop (please complete the following information):**
* Occurs on Arch linux Desktop
* Occurs on  firefox web

YgorSouza:
For the record, it works correctly if you use `add_sized`.

```rs
ui.horizontal_wrapped(|ui| {
    for i in 0..10 {
        let width = 500.0;
        let height = 250.0;
        let plot = Plot::new(format!("foo{i}"))
            .legend(Legend::default())
            .width(width)
            .height(height)
            .allow_scroll(false);
        ui.add_sized([width, height], |ui: &mut egui::Ui| {
            plot.show(ui, |_| {}).response
        });
    }
});
```

There's still a bug, of course, but that's an easy workaround if anyone needs it, and could be a clue for how to fix it in the plot itself.

PR #4631, opened 2024-06-06
Experiment with a new `Group` container

emilk:
Using the new sizing pass logic, we can now support centering of a group of widgets, something we couldn't support before.

This needs more work before it would be ready to use though.

image

### DIFF
--- a/crates/egui/src/containers/group.rs
+++ b/crates/egui/src/containers/group.rs
@@ -1,0 +1,159 @@
+//! Frame container
+
+use crate::{layers::ShapeIdx, *};
+use epaint::*;
+
+/// A group of widgets with a unique id
+/// that can be used in centered layouts.
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+#[must_use = "You should call .show()"]
+pub struct Group {
+    id_source: Id,
+    frame: Frame,
+}
+
+impl Group {
+    pub fn new(id_source: impl Into<Id>) -> Self {
+        Self {
+            id_source: id_source.into(),
+            frame: Frame::default(),
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


crates/egui/src/containers/group.rs
```diff
@@ -0,0 +1,159 @@
...
+
+use crate::{layers::ShapeIdx, *};
+use epaint::*;
+
+/// A group of widgets with a unique id
+/// that can be used in centered layouts.
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+#[must_use = "You should call .show()"]
+pub struct Group {
```
YgorSouza, line 11: Is `Ui::group` going to be changed to use this? Or maybe renamed to avoid confusion?

YgorSouza:
I made the following change on this branch to see if it fixes #3449:

```diff
M crates/egui_plot/src/lib.rs
@@ -723,7 +723,9 @@ impl<'a> Plot<'a> {
         ui: &mut Ui,
         build_fn: impl FnOnce(&mut PlotUi) -> R + 'a,
     ) -> PlotResponse<R> {
-        self.show_dyn(ui, Box::new(build_fn))
+        Group::new(self.id_source.with("group"))
+            .show(ui, |ui| self.show_dyn(ui, Box::new(build_fn)))
+            .inner
     }
```

And it does, so that's nice.

crates/egui/src/containers/group.rs
emilk: I don't know yet. This is all very much related to #4634 in that I would like a simple way to create a new `Ui` which has opt-in support for the new "grouping" behavior of this PR, as well as a frame (like `ui.group` does).

DasLixou:
Would this also fix #4039 ?

emilk:
> Would this also fix #4039 ?

yup, but with a frame of nothingness before the content show up

PR #4632, opened 2024-06-06
Improve behavior of plot auto-bounds with reduced data

### LINKED ISSUE #2307: Plot bounds broken when all y values are the same
**Describe the bug**
When all y values are identically the same, the calculated y bounds will be considered invalid and the default y-range of [0, 1] will be used instead. The same bug applies to x as well.

**To Reproduce**
I've created an example application here:

https://github.com/nknotts/egui/blob/added-plot-bounds-example/examples/plot_bounds/src/main.rs

The gist is

```
            let x_range = 0..1000;
            let single2: PlotPoints = x_range
                .clone()
                .map(|i| {
                    let x = i as f64 * 0.01;
                    [x, 2.0]
                })
                .collect();

            ui.label("All values == 2, completely missing");
            Plot::new("plot3")
                .allow_drag(false)
                .allow_zoom(false)
                .allow_scroll(false)
                .height(row_height)
                .show(ui, |plot_ui| plot_ui.line(Line::new(single2)));
```

Where determining the bounds, there is

https://github.com/emilk/egui/blob/a1f6f6821366c95b36b43e83ba67bc7e8645e14a/crates/egui/src/widgets/plot/mod.rs#L831

```
        let auto_x = !bounds_modified.x && (!min_auto_bounds.is_valid_x() || auto_bounds.x);
        let auto_y = !bounds_modified.y && (!min_auto_bounds.is_valid_y() || auto_bounds.y);
```

Ultimately there is a call to

https://github.com/emilk/egui/blob/a1f6f6821366c95b36b43e83ba67bc7e8645e14a/crates/egui/src/widgets/plot/transform.rs#L52

```
    pub fn is_valid(&self) -> bool {
        self.is_finite() && self.width() > 0.0 && self.height() > 0.0
    }
```

where, in this case, the height will be `== 0.0` since all values are the same. Since the y bounds are not valid, the default bounds of `[0, 1]` are used.

**Expected behavior**
An appropriate bound to be calculated for this case.

**Screenshots**
Screen Shot 2022-11-14 at 1 33 45 PM

 - OS: macOS
 - Version 12.6

**Additional context**

### LINKED ISSUE #3808: "Double-click to reset view" for single-point plots doesn't work
See following screencast:

https://github.com/emilk/egui/assets/2910679/3916a979-304e-4536-935a-042a09921d4c

abey79:
* Fixes #3808
* Fixes #2307

This PR improves the behaviour of auto-bounds with data that:
- is a single point
- where all X values are the same (e.g. vertical line)
- where all Y values are the same (e.g. horizontal line)

In all case, the auto-bound now aim to center on the data. For span, when available, it use the same as the other axis. If the data range of the other axis is also degenerate, then it defaults to +/- 1.0.

https://github.com/emilk/egui/assets/49431240/a62d2b5b-7856-4415-8534-83dc58cfac98

<details>
<summary>Test code</summary>

```rust
#![allow(rustdoc::missing_crate_level_docs)] // it's an example

use eframe::egui;
use egui_plot::{Legend, Line, Plot, PlotPoints, Points};

    env_logger::init(); // Log to stderr (if you run with `RUST_LOG=debug`).

        viewport: egui::ViewportBuilder::default().with_inner_size([350.0, 200.0]),
        ..Default::default()
    };
    eframe::run_native(
        "My egui App with a plot",
        options,
    )
}

#[derive(Default)]
struct MyApp {}

impl eframe::App for MyApp {
        let mut plot_rect = None;
            if ui.button("Save Plot").clicked() {
                ctx.send_viewport_cmd(egui::ViewportCommand::Screenshot);
            }

            let my_plot = Plot::new("My Plot").legend(Legend::default());

            // let's create a dummy line in the plot
            let inner = my_plot.show(ui, |plot_ui| {
                plot_ui.line(
                    Line::new(PlotPoints::from(vec![
                        [0.0, 10.0],
                        [2.0, 10.0],
                        [3.0, 10.0],
                    ]))
                    .name("y = 10.0"),
                );

                plot_ui.line(
                    Line::new(PlotPoints::from(vec![
                        [10.0, 10.0],
                        [10.0, 11.0],
                        [10.0, 12.0],
                    ]))
                    .name("x = 10.0"),
                );
                plot_ui.points(
                    Points::new(PlotPoints::from(vec![[5.0, 5.0]]))
                        .name("(5,5)")
                        .radius(3.0),
                );
                plot_ui.points(
                    Points::new(PlotPoints::from(vec![[5.0, 7.0]]))
                        .name("(5,7)")
                        .radius(3.0),
                );
            });
            // Remember the position of the plot
            plot_rect = Some(inner.response.rect);
        });

        // Check for returned screenshot:
        let screenshot = ctx.input(|i| {
            for event in &i.raw.events {
                if let egui::Event::Screenshot { image, .. } = event {
                    return Some(image.clone());
                }
            }
            None
        });

        if let (Some(screenshot), Some(plot_location)) = (screenshot, plot_rect) {
            if let Some(mut path) = rfd::FileDialog::new().save_file() {
                path.set_extension("png");

                // for a full size application, we should put this in a different thread,
                // so that the GUI doesn't lag during saving

                let pixels_per_point = ctx.pixels_per_point();
                let plot = screenshot.region(&plot_location, Some(pixels_per_point));
                // save the plot to png
                image::save_buffer(
                    &path,
                    plot.as_raw(),
                    plot.width() as u32,
                    plot.height() as u32,
                    image::ColorType::Rgba8,
                )
                .unwrap();
                eprintln!("Image saved to {path:?}.");
            }
        }
    }
}
```

</details>

### DIFF
--- a/crates/egui_plot/src/transform.rs
+++ b/crates/egui_plot/src/transform.rs
@@ -253,6 +253,8 @@ pub struct PlotTransform {
 
 impl PlotTransform {
     pub fn new(frame: Rect, bounds: PlotBounds, x_centered: bool, y_centered: bool) -> Self {
+        // Since the current Y bounds an affect the final X bounds and vice versa, we need to keep
+        // the original version of the `bounds` before we start modifying it.
         let mut new_bounds = bounds;
 
         // Sanitize bounds.


crates/egui_plot/src/transform.rs
```diff
@@ -240,26 +252,51 @@ pub struct PlotTransform {
...
-    pub fn new(frame: Rect, mut bounds: PlotBounds, x_centered: bool, y_centered: bool) -> Self {
-        // Make sure they are not empty.
-        if !bounds.is_valid_x() {
-            bounds.set_x(&PlotBounds::new_symmetrical(1.0));
-        }
-        if !bounds.is_valid_y() {
-            bounds.set_y(&PlotBounds::new_symmetrical(1.0));
-        }
+    pub fn new(frame: Rect, bounds: PlotBounds, x_centered: bool, y_centered: bool) -> Self {
+        let mut new_bounds = bounds;
```
emilk, line 256: alternative 🤷 

```suggestion
    pub fn new(frame: Rect, mut bounds: PlotBounds, x_centered: bool, y_centered: bool) -> Self {
```

emilk (approved):

crates/egui_plot/src/transform.rs
abey79: Nope! 😱  I need the original, unmodified `bounds` throughout, otherwise the modifications for the X axis interact badly with the modifications for the Y axis. Took me the longest time to figure that out.

Issue #4607, opened 2024-06-03
Unify menus with tooltips/popups


emilk:
egui menus (`Ui::menu_button`) and context menus (`Response::context_menu`) are implemented using the `egui::menu` module.

There are also tooltips (`Response::on_hover_ui` etc) and `ComboBox` popups, both implemented using `egui::containers::popup`.

It would be nice to unify these systems.

Things I would like to accomplish:

* Being able to open a popup on:
  * hover (tooltip)
  * click (menu)
  * right-click or long-press (context menu)
* Nest sub-menus everywhere (currently not possible in tooltips)
* Call `ui.context_menu` to close a tooltip and popup (currently only works for menus)

Umatriz:
i'd like to try to implement this.I also would like to

Umatriz:
I would like to try to implement this. But I also want to be able to change the way `egui::containers::popup`s are closed. Because for now if I add any kind of checkboxed into it or anything else that requires some kind of interaction the popup will immediately close as soon as I change something (mark a checkbox or press a button). I think it's better to let user to take control of it. Simitarly to the `ui.menu_button` and it's `ui.close_menu`.

Umatriz:
I think I will make a PR and write all my thoughts about this isssue.

emilk:
Nice! I suspect there is a lot in `menu.rs` that can be simplified too

xangelix:
Would greatly appreciate condition flexibility for the helpers on Response, e.g. https://github.com/emilk/egui/discussions/3452

There are also a few spots in the code base that should probably be merged into whatever solution is chosen for this: e.g. https://github.com/emilk/egui/blob/master/crates/egui/src/widgets/color_picker.rs#L498

lucasmerlin:
My plan is:
- add new `Popup` wich handles the differnet popup / tooltips sizing and positioning options
- add new `Tooltip` struct which builds on top of `Popup` and handles all the tooltip logic
- update the menu to use `Popup` and make it work in any popup / anywhere



We should also pay attantion that the new popup/tooltip/menu plays together nicely with the modal. Maybe the open state handling could even be shared between popups and modals? Seems like it should be pretty similar

PR #4636, opened 2024-06-07
Add `PopupCloseBehavior`

Umatriz:
This PR adds `PopupCloseBehavior` to improve state of the <https://github.com/emilk/egui/issues/4607>

`PopupCloseBehavior` determines when popup will be closed.
- `CloseOnClick` popup will be closed if the click happens anywhere even in the popup's body
- `CloseOnClickAway` popup will be closed if the click happens somewhere else but in the popup's body.

It also adds a test in the demo app which contains several popups examples.

---

My ideas about <https://github.com/emilk/egui/issues/4607> is to make every tooltip and popup a menu. So it will provide more control over popups and tooltips (you will be able to close a popup by calling something similar to the `ui.close_menu` if you need to). You won't need to manually handle it's opening. And also will allow to have multiple popups opened. That means you can have a popup inside a popup. And it will also lead to the easier creation of the popups. (should we create a tracking issue to track changes because to me it seems like a huge amount of changes to be done?)

---

- Improvements on <https://github.com/emilk/egui/issues/4607>


crates/egui/src/containers/popup.rs
```diff
@@ -253,18 +253,32 @@ pub fn was_tooltip_open_last_frame(ctx: &Context, widget_id: Id) -> bool {
     })
 }
 
+/// Determines popup's close behavior
+#[derive(Clone, Copy)]
+pub enum PopupCloseBehavior {
+    /// Popup will be closed on click
+    /// It is used in the [`ComboBox`]
```
emilk, line 260:
```suggestion
    /// Popup will be closed on click anywhere, inside or outside the popup.
    ///
    /// It is used in [`ComboBox`].
```

crates/egui/src/containers/popup.rs
```diff
@@ -253,18 +253,32 @@ pub fn was_tooltip_open_last_frame(ctx: &Context, widget_id: Id) -> bool {
     })
 }
 
+/// Determines popup's close behavior
+#[derive(Clone, Copy)]
+pub enum PopupCloseBehavior {
```
emilk, line 258: For completeness, I suggest we also add a third alternative called e.g. `IgnoreClicks` which ignores any clicks. You can only close it by pressing Escape or calling `Memory::close_popup` (and maybe later with `Ui::close_menu`).

----

Alternatively we replace this `enum PopupCloseBehavior` with 

```rs
struct PopupOptions {
    pub close_on_click_inside: bool,
    pub close_on_click_outside: bool,
    pub close_on_escape: bool,
}

impl PopupOptions {
    pub const CLOSE_ON_CLICK_ANYWHERE: Self = …;
    pub const CLOSE_ON_CLICK_INSIDE: Self = …;
}
```

That makes it much easier to expand with more options in the future

crates/egui/src/containers/popup.rs
```diff
@@ -253,18 +253,32 @@ pub fn was_tooltip_open_last_frame(ctx: &Context, widget_id: Id) -> bool {
...
+    /// Popup will be closed on click
+    /// It is used in the [`ComboBox`]
+    CloseOnClick,
+
+    /// Popup will be closed if the click happened somewhere else
+    /// but in the popup's body
+    CloseOnClickAway,
```
emilk, line 265:
```suggestion
    CloseOnClickOutside,
```

crates/egui/src/containers/popup.rs
```diff
@@ -327,13 +342,19 @@ pub fn popup_above_or_below_widget<R>(
                         .inner
                     })
                     .inner
-            })
-            .inner;
+            });
+
+        let is_close = match close_behavior {
```
emilk, line 347:
```suggestion
        let should_close = match close_behavior {
```

("is close" sounds like it is about proximity)

emilk (changes requested):
Good idea!

emilk (approved):

emilk:
CI is red though

Umatriz:
@emilk I fixed the doc test so now everything should be fine

Issue #3656, opened 2023-11-28
Zooming out crashes `egui_plot`


danielfaust:
**Describe the bug**

In the demo located at https://www.egui.rs/#Demo, when fully zooming out from the "Custom Axes" plot example in the "Plot" window, egui will crash with a "The egui app has crashed." message

**To Reproduce**

1. Go to https://www.egui.rs/#Demo
2. Click on Plot in the left pane
3. Click on Custom Axes in the Plot window
4. Fully scroll out with the mouse wheel while pressing Ctrl

**Desktop**

 - Windows 11
 - Google Chrome
 - Version 119.0.6045.160 (Offizieller Build) (64-Bit)

**Desktop**

 - Ubuntu 22.04
 - Firefox
 - Version 120.0 (64-Bit) canonical-002-1.0 (Mozilla Firefox Snap for Ubuntu)

**Smartphone**

Not reproducible due to pinch and zoom not being able to fully zoom out

Firepal:
Relevant: https://github.com/emilk/egui/blob/8d4de866d4da1835b31e85f9068a5257e6ccbccb/crates/egui_demo_lib/src/demo/plot_demo.rs#L478-L510

It's very much possible that this is trying to draw many, many grid marks

emilk:
Very similar to:
* https://github.com/emilk/egui/issues/3462

Tweoss:
It does seem like the custom axes demo tries to draw many grid marks when zoomed out.

In the normal plot demo, when zooming out, I get the below. Presumably, NaN is compared against another floating point number, and the inequality doesn't hold in line 1916. 
https://github.com/emilk/egui/blob/520b28ce22a0daa20e13767af6061e9e40796f2a/crates/egui_plot/src/lib.rs#L1915-L1925

![Screenshot 2024-02-01 at 10 22 17](https://github.com/emilk/egui/assets/70821802/302e550d-b980-4715-b4a9-f601de93d89e)
![Screenshot 2024-02-01 at 10 22 24](https://github.com/emilk/egui/assets/70821802/713db564-5aa1-4080-85b2-9b8968dc6b7b)

danielfaust:
I can no longer find any plots in the demo. Have they been removed?

Issue #3462, opened 2023-10-12
Panic when zooming out a plot too far


c-git:
I wasn't able to check if it is present on main by reproducing it but the [code that causes](https://github.com/emilk/egui/blob/fd75adb3a0e5f8ca373ac7e5e03f750e690ae87b/crates/egui_plot/src/lib.rs#L1877) the panic is still there.

**Describe the bug**

I was experimenting with zooming using the right mouse button. And while doing so seems I hit an edge case and the program suddenly closed. Checking the console I discovered that it had panicked. Seem I managed to accidentally try to zoom by exactly 0.0 and there is an assert to check that the amount is not 0.0 hence it crashed. While this is hard to recreate it is not impossible, might be worth adding a check before calling [`next_power`](https://github.com/emilk/egui/blob/master/crates/egui_plot/src/lib.rs#L1877) so that it doesn't panic but instead does nothing as that seems to make sense for a zoom by 0.

Error message:

```
thread 'main' panicked at /home/one/.cargo/registry/src/index.crates.io-6f17d22bba15001f/egui_plot-0.23.0/src/lib.rs:1877:5:
assertion `left != right` failed
  left: -0.0
 right: 0.0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**To Reproduce**
Given the nature of the bug I was unable to reproduce it as it requires zooming on a plot with the right mouse button and getting exactly 0 for the amount of zoom.

**Expected behavior**

I think nothing should happen instead of a panic as the user may trigger this (even though it is unlikely)

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04.01
 - Browser: N/A running native when crashed
 - Version: 23

Its-Just-Nans:
I think this could be closes with #4637 ?!

emilk:
> I think this could be closes with #4637 ?!

No - I can still reproduce on egui.rs after zooming out for a minute or so

PR #4637, opened 2024-06-07
egui_plot: use `f64` for translate

Its-Just-Nans:
Can fix translating with high zoom out https://github.com/emilk/egui/issues/3462

* Maybe related to https://github.com/emilk/egui/issues/3656


Its-Just-Nans:
image

nevertheless did someone tried to zoom out more than `f64::MAX` ?

emilk (approved):
Thanks!

Issue #4610, opened 2024-06-05
`DragValue` with `min_decimals` set to 1 formats `0.0` as `0`


Skgland:
**Describe the bug**
With `min_decimals` set to one all values should be displayed with at least one decimal,
but `0` is displayed instead of `0.0`

**To Reproduce**
Create a UI containing a `DragValue` with min_decimals set to 1 and with a current value of 0.0.
e.g. 
`ui.add(egui::DragValue::new(&mut 0.0).min_decimals(1))`

**Expected behavior**
`min_decimals` is honored and `0.0` is displayed.

**Screenshots**

![image](https://github.com/emilk/egui/assets/3877590/225910ff-899b-40e8-80e0-aef993a8eb19)

**Desktop (please complete the following information):**
 - OS: Ubuntu Jammy 22.04 under WSL 2/ Debian Bullseye
 - Version `0.25.0`

**Additional context**

This is probably due to this special case:
<https://github.com/emilk/egui/blob/master/crates/egui/src/widgets/drag_value.rs#L456-L458>

YgorSouza:
It looks like this special case was added in https://github.com/emilk/egui/commit/ebc2486d22355eb6ef1f4827f4304210969b60eb#diff-3878bfa9f60333dd78982f3697643c6755ff8378f18b54fff03973c1eff5fcc8L159-R173 more than 3 years ago, and kept this way ever since. But I don't know what is the reasoning behind it.

PR #4639, opened 2024-06-07
Remove special case for 0 in DragValue default formatter

YgorSouza:
* Closes <https://github.com/emilk/egui/issues/4610>

### DIFF
--- a/crates/egui/src/widgets/drag_value.rs
+++ b/crates/egui/src/widgets/drag_value.rs
@@ -452,13 +452,7 @@ impl<'a> Widget for DragValue<'a> {
 
         let value_text = match custom_formatter {
             Some(custom_formatter) => custom_formatter(value, auto_decimals..=max_decimals),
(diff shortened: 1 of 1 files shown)


YgorSouza:
This special case was added in https://github.com/emilk/egui/commit/ebc2486d22355eb6ef1f4827f4304210969b60eb#diff-3878bfa9f60333dd78982f3697643c6755ff8378f18b54fff03973c1eff5fcc8L159-R173, but I don't see any explanation for it, neither in the code nor in the commit message, and removing it doesn't seem to cause any problems, so I'm assuming it was a mistake. If there is a justification for it, I can change this PR to add a comment to the code instead, so no one questions it again.

https://github.com/emilk/egui/assets/43298013/c38aebe1-c6c0-4251-82e5-69cb9aeaa3dd

emilk (approved):
Makes sense

PR #4641, opened 2024-06-07
`egui_extras`: Make `serde` an optional feature

Dinnerbone:
All the other crates in egui have serde as an optional dependency - which is great! But sadly egui_extras unconditionally includes it, which adds a bunch of code to stuff that may not care for it. This PR gates serde support behind a new `serde` feature.

This is a breaking change; if that's undesirable then we can add it as a default feature instead, though that wouldn't match any of the other crates.


emilk (approved):
Thanks!

emilk:
This caused a problem:
* https://github.com/emilk/egui/issues/4771

Issue #4642, opened 2024-06-08
Custom default replacement glyphs for font


Warpten:
**Is your feature request related to a problem? Please describe.**

There is currently no way to define a custom replacement glyph for user-provided fonts. For example, the font "MaterialSymbolsSharp" can't be used with egui because it does not have the replacement glyphs egui is hardcoded to look for. The font itself seems to load fine; however as soon as I try to use the font the code panics here:

![](https://i.imgur.com/nP4YBxa.png)

**Describe the solution you'd like**
Add user-specified overrides to `FontTweak` that replace the glyphs that the library looks for.

**Describe alternatives you've considered**
Besides using another font, there isn't really any option. Even using font fallbacks (for example with `RichText::append_to`), egui still tries to load the font and panics before falling back.

**Additional context**
Seen on egui 0.27.2 on a custom vulkan backend (although that is unrelated to the issue at hand).

Usage:
```rust
// This happens on app start, looks through ./assets/fonts and loads everything:
def.font_data.insert(
    font_name.to_owned(), // MaterialSymbolsSharp
    FontData::from_owned(file_data.unwrap())
);

def.families.entry(value.clone())
    .and_modify(move |value| value.push(font_name.to_owned()))
    .or_insert(vec![font_name.to_owned()]);

def.families.insert(FontFamily::Name(font_name.into()), vec![font_name.to_owned()]);

// Later in the UI
let mut label_job = LayoutJob::default();
RichText::new("HOME")
    .family(FontFamily::Name("MaterialSymbolsSharp".into()))
    .append_to(&mut label_job, &Style::default(), FontSelection::Default, egui::Align::Min);
RichText::new("Home")
    .family(FontFamily::Name("Roboto-Light".into()))
    .append_to(&mut label_job, &Style::default(), FontSelection::Default, egui::Align::Min);

ui.selectable_value(&mut self.active_tab, Tab::Home, label_job);
```

Warpten:
Of course I find #3526 just after creating this issue

PR #4644, opened 2024-06-09
Try to improve support for .notdef glyphs through `FontTweak`

### LINKED ISSUE #3526: Can't load Material Design font because of missing missing glyph
1. Try to load the font file called `MaterialIcons-Regular.ttf` located in [this giant zip](https://github.com/google/material-design-icons/releases/tag/4.0.0).
2. Get error `Failed to find replacement characters '◻' or '?'` that originates from epaint-0.23.0/src/text/font.rs:364:17.

This is the code that I used to load the font file:

```rust
        let mut fonts = FontDefinitions::default();
        fonts.font_data.insert(
            Self::FONT_ICONS.to_owned(),
            FontData::from_static(include_bytes!(
                "../../../res/fonts/material-design-4.0/MaterialIcons-Regular.ttf"
            )),
        );

        fonts
            .families
            .entry(FontFamily::Name(Self::FONT_ICONS.into()))
            .or_default()
            .insert(0, Self::FONT_ICONS.to_owned());
```

and this is an excerpt from the egui/epaint font.rs file:

```rust
        const PRIMARY_REPLACEMENT_CHAR: char = '◻'; // white medium square
        const FALLBACK_REPLACEMENT_CHAR: char = '?'; // fallback for the fallback

        let replacement_glyph = slf
            .glyph_info_no_cache_or_fallback(PRIMARY_REPLACEMENT_CHAR)
            .or_else(|| slf.glyph_info_no_cache_or_fallback(FALLBACK_REPLACEMENT_CHAR))
            .unwrap_or_else(|| {
                panic!(
                    "Failed to find replacement characters {PRIMARY_REPLACEMENT_CHAR:?} or {FALLBACK_REPLACEMENT_CHAR:?}"
                )
            });
```

I think that this font family represents an edge case, because it's icons rather than letters/numbers/symbols, which means that nobody will ever use this font without knowing exactly which code points to use, which means that the concept of a replacement glyph shouldn't apply.

Clearly, it would have been better if the authors of this font file had included a glyph for the white medium square, but since they didn't, and I'd like to use the font file as-is, I don't have a good workaround at this point.

Warpten:
The idea consists of adding `fallbacks` to the `FontTweak` structure, which then get used in egui's `Font` (actually a wrapper over multiple `FontImpl`) to identify fallback characters. We look through each wrapped `FontImpl`, stopping in the first fallback character we found, defaulting to the empty glyph if none is found.

Example output when replacing the user font by `MaterialSymbolsSharp` in the egui custom font example:
image

```rust
impl eframe::App for MyApp {
            ui.heading("\u{E834} egui using custom fonts");
            ui.text_edit_multiline(&mut self.text);
        });
    }
}
```

Note that as it is (mostly due to my very limited understanding of fonts in general), this isn't enough to load .notdef properly, even with a user-definied fallback character set. I suspect this however has to do with `ttf_parser` seemingly not actually loading .notdef.

* Closes <https://github.com/emilk/egui/issues/3526> (this one was technically originally closed by <https://github.com/emilk/egui/pull/4542>)
* See <https://github.com/emilk/egui/issues/4642>

### DIFF
--- a/crates/epaint/src/text/font.rs
+++ b/crates/epaint/src/text/font.rs
@@ -78,6 +78,8 @@ pub struct FontImpl {
     pixels_per_point: f32,
     glyph_info_cache: RwLock<ahash::HashMap<char, GlyphInfo>>, // TODO(emilk): standard Mutex
     atlas: Arc<Mutex<TextureAtlas>>,
+
+    fallbacks: [char; 2],
 }
 
 impl FontImpl {
@@ -132,6 +134,7 @@ impl FontImpl {
             pixels_per_point,
             glyph_info_cache: Default::default(),
             atlas,
+            fallbacks: tweak.fallbacks.unwrap_or(['◻', '?']),
         }
     }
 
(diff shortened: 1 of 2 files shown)


crates/epaint/src/text/font.rs
```diff
@@ -434,7 +440,7 @@ impl Font {
 
     /// Can we display this glyph?
     pub fn has_glyph(&mut self, c: char) -> bool {
-        self.glyph_info(c) != self.replacement_glyph // TODO(emilk): this is a false negative if the user asks about the replacement character itself 🤦‍♂️
+        self.glyph_info_no_cache_or_fallback(c).is_some()
```
Warpten, line 443: This seemed a nice "bonus" to-do solving, slightly unrelated to the issue at hand; I'm suspecting there's a reason you chose not to do this in the first place?

emilk:
When would you want to have a different replacement char?

Warpten:
Whenever the default replacement character of a font is neither the empty square nor that font has a question mark, which is the case for Material Symbols Sharp:

image

crates/epaint/src/text/font.rs
```diff
@@ -363,20 +366,23 @@ impl Font {
...
+            .fonts
+            .iter()
+            .enumerate()
+            .find_map(|(font_index, font_impl)| {
+                font_impl.fallbacks.iter().find_map(|chr| {
+                    font_impl
+                        .glyph_info(*chr)
+                        .map(|glyph_info| (font_index, glyph_info))
+                })
+            })
```
emilk, line 380: Why not call `glyph_info_no_cache_or_fallback` here like before?

crates/epaint/src/text/font.rs
emilk: Well, `glyph_info_no_cache_or_fallback` doesn't use the cache, so it is marginally slower.

emilk:
Closing: the font code was rewritten in #8504 and #8512, which handle `.notdef` glyphs and fallback fonts explicitly, so this no longer applies. Thanks for the PR!

PR #4647, opened 2024-06-10
Fix custom_window_frame example

Gusted:
Using `dragged_by` caused the `StartDrag` command to be spammed, which in turn caused that you couldn't stop dragging the frame because the window was repeatly being told to start dragging. Use `drag_started_by` which will only send the `StartDrag` command once at the begin.


rustbasic:
yes. It would be better to use `drag_started_by()`.  
On Windows, there is no problem using either `drag_by()` or `drag_started_by()`.

emilk:
Works on Mac.

Has this been tested on Linux and Windows?

PR #4648, opened 2024-06-10
Don't hover things under tooltips

emilk:
The removed code was added in https://github.com/emilk/egui/pull/1240 to fix https://github.com/emilk/egui/issues/1239 but a better fix for that was added in https://github.com/emilk/egui/pull/4623

So: tooltips can now have interactive stuff in them, so when hovering a tooltip we should not hover things below them.


emilk:
Actually, this will break the current behavior of `Window::interactable` and `Area::interactable`. I'll do a better fix

PR #4656, opened 2024-06-13
Fix dragging of `custom_window_frame` example on Windows

rustbasic:
* Related #4592 
* Closes #4647 

Fix dragging of custom_window_frame on Windows (re-edited)


emilk:
Have you tested this on X11?

rustbasic:
> Have you tested this on X11?

no.
Tested on Windows.
Since there doesn't seem to be any particular problem with `has_focus()`,
I don't expect any problems elsewhere.

ardocrat:
Tested on Linux, seems like we have a problem with app window resize as we as with dragging, egui is not receiving event on mouse button release.

PR #4658, opened 2024-06-14
Fix: `ScrollArea` repeatedly appears and disappears quickly issue.

rustbasic:
If the mouse is at a specific position above the `ScrollArea` (in TextEdit?), the `ScrollArea` repeatedly appears and disappears quickly.
You can more easily find the issue if you set `Full bar width` to the maximum value of 32.

If there is a better way to fix this than the solution in this Pull Request, please revise it accordingly.

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -967,9 +967,9 @@ impl Prepared {
             // left/rigth of a vertical scroll (d==1).
             let mut cross = if scroll_style.floating {
                 let max_bar_rect = if d == 0 {
-                    outer_rect.with_min_y(outer_rect.max.y - scroll_style.bar_width)
+                    outer_rect.with_min_y(outer_rect.max.y - scroll_style.bar_width - outer_margin)
                 } else {
-                    outer_rect.with_min_x(outer_rect.max.x - scroll_style.bar_width)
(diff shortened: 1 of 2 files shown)


emilk:
Please give a repro of the problem so I can test this

rustbasic:
> Please give a repro of the problem so I can test this

It is difficult to capture, so I will explain how to reproduce it instead.

To make it easier to reproduce the bug,
Open `settings_ui` in `egui_demo_app`, go to `Spacing` -> `Scroll Area` -> `Details`,
Select `Floating`,
And set `Full bar width` to `32`.

To easily reproduce the bug, open a window containing a `ScrollArea` and repeatedly hover your mouse pointer over the `ScrollArea`.

emilk:
I fail to reproduce any problem. Do you have a video of it?

rustbasic:
> I fail to reproduce any problem. Do you have a video of it?


https://github.com/rustbasic/rustbasic/blob/main/egui%20demo%20app%202024-06-25%2017-26-02.mp4

When I go to the `font book` from the 2/3rd point of the video to the end, the 'Scroll Bar' keeps appearing and disappearing, but in the video, it appears to be blurry due to a save fps issue.

Regardless of the specific conditions, the same symptom appears in all `Scroll Bars`.

You can tell because it blinks once every few dozen times.
If you want to keep blinking, you have to stop the mouse right away when it blinks, so it will keep blinking, so you have to do it hundreds of times.

PR #4659, opened 2024-06-14
egui_extras: feature `svg_text`

xNWP:
# Added
- Adds svg text rendering support via the `svg_text` feature flag.

# Changed
- Updates resvg from `0.37` to `0.42`.

### DIFF
--- a/crates/egui_extras/src/image.rs
+++ b/crates/egui_extras/src/image.rs
@@ -239,7 +239,9 @@ pub fn load_svg_bytes_with_size(
 
     crate::profile_function!();
 
+    #[allow(unused_mut)]
     let mut opt = Options::default();
+
     #[cfg(feature = "svg_text")]
     opt.fontdb_mut().load_system_fonts();
 
--- a/crates/egui_extras/src/image.rs
+++ b/crates/egui_extras/src/image.rs
@@ -235,45 +235,47 @@ pub fn load_svg_bytes_with_size(
     size_hint: Option<SizeHint>,
 ) -> Result<egui::ColorImage, String> {
     use resvg::tiny_skia::{IntSize, Pixmap};
-    use resvg::usvg::{Options, Tree, TreeParsing};
+    use resvg::usvg::{Options, Transform, Tree};
 
     crate::profile_function!();
-    let opt = Options::default();
 
-    let mut rtree = Tree::from_data(svg_bytes, &opt).map_err(|err| err.to_string())?;
+    let mut opt = Options::default();
+    #[cfg(feature = "svg_text")]
+    opt.fontdb_mut().load_system_fonts();
 
-    let mut size = rtree.size.to_int_size();
-    match size_hint {
-        None => (),
-        Some(SizeHint::Size(w, h)) => {
-            size = size.scale_to(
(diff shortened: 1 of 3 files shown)


crates/egui_extras/src/image.rs
```diff
@@ -235,45 +235,49 @@ pub fn load_svg_bytes_with_size(
...
-            size = size
-                .scale_to_height(h)
-                .ok_or_else(|| format!("Failed to scale SVG to height {h}"))?;
-        }
-        Some(SizeHint::Width(w)) => {
-            size = size
-                .scale_to_width(w)
-                .ok_or_else(|| format!("Failed to scale SVG to width {w}"))?;
-        }
+    #[allow(unused_mut)]
```
murl-digital, line 242: is this here because clippy gets confused?

murl-digital (commented):
that allow(unused_mut) caught my eye but otherwise lgtm

xNWP:
> that allow(unused_mut) caught my eye but otherwise lgtm
> 
> edit: i should really read commit messages first, could you put in a comment explaining it?

commented!

murl-digital:
i think for demonstration purposes (and to make testing easier in the future), it'd be a good idea to add an example that shows text in svgs in action

emilk (approved):
I agree that an example would be nice, but LGTM anyway.

Would this text rendering work on web?

xNWP:
> Would this text rendering work on web?

I was wondering the same thing shortly after making the PR 😅, I'm not great at web dev stuff but I'll take a look at it now and test some things.

> i think for demonstration purposes (and to make testing easier in the future), it'd be a good idea to add an example that shows text in svgs in action

after checking the web stuff I'll also look at how y'all test code and see what I can do 😊

xNWP:
> Would this text rendering work on web?

Short answer, no.
Long answer is that the main workhorse of this PR is the `load_system_fonts` function on the Font Database type that `resvg` re-exports from the `fontdb` crate (https://github.com/RazrFalcon/fontdb/blob/fd24f8c9d7b57f30a6f6ba560c697a7854c94858/src/lib.rs#L352). This function is just a conditional compilation switch based on the target os which simply does nothing for wasm. I looked into how support could be added to this crate, however it looks like the way to do it would be the Local Font Access API which has little to no browser support (only experimental Chrome + Edge support) https://developer.mozilla.org/en-US/docs/Web/API/Local_Font_Access_API

emilk:
Thanks for testing web. Please document your findings, i.e. document that the `svg_text` won't work on web, and why (linking to relevant issues, if any).

We should also figure out a solution to the duplicated dependencies.

xNWP:
Disregard this PR (for now).

Apologies for the delay in this, adhd is a bitch. But while looking at this again I realized that I mistakenly have it set up in a way that the system fonts are being loaded EVERY single egui loads a new svg, instead it should be loaded once (during the install image loaders call likely), and that font database should be reused by the SVG loader. I also realized that I could likely solve the problem of no fonts on the web by allowing you to load/embed the font data directly (moving the problem onto the egui user). When I find the time to reimplement this in a better way I'll open a new PR and reference this one :)

PR #4661, opened 2024-06-14
Replace `directories-next` dependency with `directories`

crumblingstatue:
`directories-next` was created because `directories` was not maintained at the time. However, `directories` has gotten active maintainership since, and it has received more updates than `directories-next`.

`directories` also has more recent downloads than its `next` counterpart, so it might make sense to switch to it, to avoid unnecessary duplicate dependencies, where a project depends transitively on both `directories` and `directories-next`.

The main question is whether we depend on any specific behavior from `directories-next`.


deny.toml
```diff
@@ -52,7 +52,7 @@ skip = [
   { name = "libloading" },     # wgpu-hal itself depends on 0.8 while some of its dependencies, like ash and d3d12, depend on 0.7
   { name = "memoffset" },      # tiny dependency
   { name = "quick-xml" },      # old version via wayland-scanner
-  { name = "redox_syscall" },  # old version via directories-next
+  { name = "redox_syscall" },  # old version via winit
```
crumblingstatue, line 55: I'm not exactly sure what is going on in this file, but with this change, `winit` is the crate referring to the oldest version of `redox_syscall` (0.3.5)

emilk (approved):
Makes sense!

PR #4662, opened 2024-06-14
Add legend formatter to egui_plot

rickyelopez:
Small change to introduce the ability to provide a function which (if provided) will be used to format the text used in the legend.

The motivation for this is to be able to plot enumerated signals in which the specific enum name is included in the name of the point/line (eg. `enum_name::enumeration`). Without this change there is no way (as far as I can tell) to get all of the points/lines into the same entry in the legend, so you end up with a legend entry for each possible enumeration. There may be a better way to accomplish this that I'm not seeing, though.

### DIFF
--- a/crates/egui_plot/src/legend.rs
+++ b/crates/egui_plot/src/legend.rs
@@ -181,6 +181,7 @@ impl LegendWidget {
         config: Legend,
         items: &[Box<dyn PlotItem>],
         hidden_items: &ahash::HashSet<String>, // Existing hiddent items in the plot memory.
+        formatter: Option<&LegendFormatterFn>,
     ) -> Option<Self> {
         // If `config.hidden_items` is not `None`, it is used.
         let hidden_items = config.hidden_items.as_ref().unwrap_or(hidden_items);
@@ -192,8 +193,12 @@ impl LegendWidget {
             .iter()
             .filter(|item| !item.name().is_empty())
             .for_each(|item| {
+                let name = formatter.map_or_else(
+                    || item.name().to_owned(),
+                    |formatter| formatter(item.name()).to_owned(),
+                );
                 entries
-                    .entry(item.name().to_owned())
+                    .entry(name.clone())
                     .and_modify(|entry| {
                         if entry.color != item.color() {
                             // Multiple items with different colors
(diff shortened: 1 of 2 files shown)


rickyelopez:
@emilk could you please take a look at this when you get a chance? Thanks!

emilk:
I don't understand your motivation. Why pass a function to change the names, instead of just using different names from the beginning?

Can you maybe show some code to explain?

rickyelopez:
I think this is reasonable regardless of my usecase, but let me try to give an example. Maybe you will have a better idea.

Let's say the data I want to plot is represented by an enum:
```rust
enum TestEnum {
    Val1,
    Val2,
    Val3,
    etc...
}
```
where each enumeration is mapped to an integer representation (i.e. `Val1` is represented by `1`).
Then, each point to plot is effectively a `(Time, (TestEnum, Value))` tuple. The data to plot might look like:
```rust
let data: (u32, (TestEnum, u32)) = vec![
    (0, (TestEnum::Val1, 1)),
    (1, (TestEnum::Val1, 1)),
    (2, (TestEnum::Val2, 2)),
    (3, (TestEnum::Val3, 3)),
    (4, (TestEnum::Val4, 4)),
]
```

I want to see the enumeration name when hovering over the point, so the way that I have been doing it so far is to create Points with the time and enum value, and then setting the name for each point to a stringified representation of the enumeration.

This works ok for showing the data on the plot. I would actually prefer if there was a way to change the Y axis from showing floats to the actual enumerations, but it doesn't seem like there's a way to do that (at least that I've found so far).

The problem is that the points now all have different names (i.e. `TestEnum::Val1`, `TestEnum::Val2`, etc.) so that the names show in the hover, but that results in the legend showing all of the enumerations as different plots. With the `legend_formatter` I've added here, I can just split on the `::` and drop the second part, effectively grouping each of the points into its enum regardless of the specific enumeration.

Continuing with the above example, the first entry in the `data` vec would result in a Point with time 0 and value 1, with the name `TestEnum::Val1` shown on hover, but it would still show up in the legend under `TestEnum` rather than `TestEnum::Val1` thanks to the modifications performed in the `legend_formatter` function. The rest of the points would then also end up under the `TestEnum` group in the legend.

Does this make sense? Like I said, there may be a better way to do this. I feel like the real solution (which would address not only this issue, but also the one mentioned above regarding the Y axis) would be to implement a new Widget such as `EnumeratedPlot` which would support y-axes with data types other than floats, and which could also support the legend behavior I'm looking for. That seems like a much bigger task, though, and I feel like this `legend_formatter` could also be beneficial in general.

emilk:

PR #4664, opened 2024-06-15
Fix: ScrollArea repeatedly appears and disappears quickly issue. + For `scroll_with_delta()`

rustbasic:
Fix: ScrollArea repeatedly appears and disappears quickly issue. (#4658)

* Closes #4658 

If the mouse is at a specific position above the ScrollArea (in TextEdit?), the ScrollArea repeatedly appears and disappears quickly.
You can more easily find the issue if you set Full bar width to the maximum value of 32.

If there is a better way to fix this than the solution in this Pull Request, please revise it accordingly.

AND 

* Closes #4584 

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -979,8 +979,7 @@ impl Prepared {
             // top/bottom of a horizontal scroll (d==0).
             // left/rigth of a vertical scroll (d==1).
             let cross = if scroll_style.floating {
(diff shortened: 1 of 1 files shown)


emilk:
Don't open multiple PRs with the same functionality. You have three options:

* Wait for https://github.com/emilk/egui/pull/4658 to be merged
* Use https://github.com/emilk/egui/pull/4658 as a merge-target for this PR
* Remove the duplicated commits

PR #4667, opened 2024-06-16
Drag n Drop - WEB - stop reading dropped file

### LINKED ISSUE #4654: Drag and Drop: read incrementally

When dropping a large file into `egui` on the web, `eframe` always reads the whole file---even if file drop events aren't handled. Drag a particularly large file into egui, and eframe will try to load the whole thing into memory. This isn't ideal.

The code below is problematic:

https://github.com/emilk/egui/blob/ca36f3df6319cbba0e2ab4366d51d97caa9e85b6/crates/eframe/src/web/events.rs#L563-L589

It would be nice to read dropped files in parts, and only on demand. Seems that [`Blob.slice`](https://developer.mozilla.org/en-US/docs/Web/API/Blob/slice) is how JS usually does it, but my knowledge of web is limited.

Its-Just-Nans:
* Closes <https://github.com/emilk/egui/issues/4654> by not reading files and giving the user a ObjectURL


crates/egui/src/data/input.rs
```diff
@@ -336,7 +336,7 @@ pub struct DroppedFile {
     pub last_modified: Option<std::time::SystemTime>,
 
     /// Set by the `eframe` web backend.
-    pub bytes: Option<std::sync::Arc<[u8]>>,
+    pub stream_url: Option<String>,
```
emilk, line 339: How is this supposed to be used?

Its-Just-Nans: this create a url

Then the user can fetch() the file

Maybe i should provide a helper function to retreive the data ?

emilk: And please make sure the `egui_example_app` demonstrates how to download the file from this URL

emilk: A helper function would be great yes.

emilk:
I'll convert this to draft until it is green

crates/egui/src/data/input.rs
Its-Just-Nans: image

Its-Just-Nans:
image

I think it's ready for review now

crates/eframe/src/web/mod.rs
```diff
@@ -246,3 +246,43 @@ pub fn percent_decode(s: &str) -> String {
         .decode_utf8_lossy()
         .to_string()
 }
+
+use js_sys::Uint8Array;
+use log::{debug, error};
+use std::sync::Arc;
+use wasm_bindgen::JsCast;
+use wasm_bindgen_futures::JsFuture;
+use web_sys::window;
```
emilk, line 255: Move them to inside the `fn`. That way they will move with the function if it has to move file

crates/eframe/src/web/events.rs
```diff
@@ -566,34 +566,22 @@ pub(crate) fn install_canvas_events(runner_ref: &WebRunner) -> Result<(), JsValu
...
-                                            runner_lock.needs_repaint.repaint_asap();
-                                        }
-                                    }
-                                    Err(err) => {
-                                        log::error!("Failed to read file: {:?}", err);
-                                    }
-                                }
+                                if let Some(mut runner_lock) = runner_ref.try_lock() {
+                                    let stream_url =
+                                        web_sys::Url::create_object_url_with_blob(&file).ok();
```
emilk, line 575: According to https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static this allocates an object that lives for the duration of `document`, meaning the memory will stay allocated:

>  for optimal performance and memory usage, if there are safe times when you can explicitly unload them, you should do so.

This feels like a new foot fgun: drag dropping many files will now slowly eat away your RAM

Its-Just-Nans: So i guess i could revoke_object_url in get_data 

And also implement the Drop trait to drop the url ?

emilk: Difficult.

One idea: `eframe::web` automatically calls `revoke_object_url` at the end of the frame unless the user has called `eframe::web::get_data`. This could be communicated with e.g. an `Arc<AtomicBool>`.

crates/egui/src/data/input.rs
emilk, line 339: I wonder if this should just be a callback instead ("call this to read the data"). One complicating factor is that it is `async`

Its-Just-Nans: Yes, quite difficult

Its-Just-Nans:
I've added the bool to drop

But the file reading is not working anymore..

emilk:
> But the file reading is not working anymore..

I'll convert it to a draft in the meantime

emilk:
Closing: this landed in #8354, which stores the `web_sys::File` in `DroppedFile` instead of reading it eagerly. Thanks for the PR!

PR #4671, opened 2024-06-18
Make `Debug` format of `Vec2/Pos2/Rot2` respect user precision

### LINKED ISSUE #4665: Nitpick: give at least two decimals of information for a pos2 in the Debug trait
As Emil has written, debugging layout in egui can be tricky. Extra tricky is when working with Pos2 in a coordinate system normalized to 0..=1. In this case, the debugging output for Pos2 formatted as `{:.1}` gives too little information. This is especially apparent when writing tests:
```
  left: [0.2 0.2]
 right: [0.2 0.2]
```
I would humbly request for the Debug trait to give at least 2 decimals of the x/y values.

murl-digital:
* closes #4665

pretty self explanatory, i opted for 3 instead of 2 since i think that's what display does

### DIFF
--- a/crates/emath/src/pos2.rs
+++ b/crates/emath/src/pos2.rs
@@ -319,7 +319,11 @@ impl Div<f32> for Pos2 {
 
 impl fmt::Debug for Pos2 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "[{:.3} {:.3}]", self.x, self.y)
+        if let Some(precision) = f.precision() {
+            write!(f, "[{1:.0$} {2:.0$}]", precision, self.x, self.y)
+        } else {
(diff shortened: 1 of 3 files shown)


emilk:
Almost always `Pos2` and `Vec2` is in in the unit of ui points, where 1 decimal is enough.
Is there a way to make `{pos:.3}` work for users that want to opt-in to more decimals?

murl-digital:
i could put it behind a feature gate, that wouldn't be hard to do at all

Mingun:
I think it should be possible to respect the formatting flags that user can provide like he can do that for `f64`

murl-digital:
> I think it should be possible to respect the formatting flags that user can provide like he can do that for `f64`

it is indeed possible, was a pretty quick change to respect it

emilk (approved):

PR #4673, opened 2024-06-18
Add `Rect::scale_from_center`

zkrising:
I find myself wanting this API quite a lot, and I imagine it'll probably be useful for others.

# What?

This PR adds `Rect::scale` and `Rect::scale2` functions, which work a lot like `expand`, but instead multiply by a scale.

i.e.
```rs
rect.scale(2.0); // rect is 2x as big, still in same center
rect.scale2(vec2(1.5, 2.0)); // rect is 1.5x as big on x axis, 2.0x as big on y axis. still in same center
```

# Why?

Before this you either had to write this yourself or use a `expand` in a cumbersome way:
```rs
rect.expand2(vec2(rect.width() * scale.x / 2.0, rect.height() * scale.y / 2.0));
```

I find myself wanting to scale things up by a factor frequently enough, and it seems like a useful addition to have a multiply-based variant of `expand`.

I realise this is pretty minor, but it seems useful enough to me!

### DIFF
--- a/crates/emath/src/rect.rs
+++ b/crates/emath/src/rect.rs
@@ -193,13 +193,13 @@ impl Rect {
 
     /// Scale up by this factor in each direction, keeping the center
     #[must_use]
-    pub fn scale(self, amnt: f32) -> Self {
-        self.scale2(Vec2::splat(amnt))
+    pub fn scale_from_center(self, amnt: f32) -> Self {
+        self.scale_from_center2(Vec2::splat(amnt))
     }
 
     /// Scale up by this factor in each direction, keeping the center
     #[must_use]
-    pub fn scale2(self, scale: Vec2) -> Self {
+    pub fn scale_from_center2(self, scale: Vec2) -> Self {
         Self::from_center_size(self.center(), self.size() * scale)
     }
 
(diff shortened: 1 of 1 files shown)


murl-digital (approved):

crates/emath/src/rect.rs
```diff
@@ -191,6 +191,18 @@ impl Rect {
...
 
+    /// Scale up by this factor in each direction, keeping the center
+    #[must_use]
+    pub fn scale(self, amnt: f32) -> Self {
+        self.scale2(Vec2::splat(amnt))
+    }
+
+    /// Scale up by this factor in each direction, keeping the center
+    #[must_use]
+    pub fn scale2(self, scale: Vec2) -> Self {
```
emilk, line 202: I think these should be called `scale_from_center[2]`, or take an `Align2` to specify which corner is the fixpoint

zkrising: Renamed to `scale_from_center`. I don't think taking an `Align2` would be particularly useful (I can't think of a use case for wanting to scale up around non-center?), but I might be wrong on that.

emilk (approved):

zkrising:
Sweet, thanks man!

PR #4680, opened 2024-06-20
Add Into bound into id methods

creativcoder:
I've felt the need for this several times, and assigning ids adds a little bit of friction by having the user call into() when passing a `String` or `&str`.

This PR adds the `Into` trait bound on the `id` method that reduces this small friction from the end user.

### DIFF
--- a/crates/egui/src/containers/area.rs
+++ b/crates/egui/src/containers/area.rs
@@ -127,8 +127,8 @@ impl Area {
     ///
     /// The `id` must be globally unique.
     #[inline]
-    pub fn id(mut self, id: Id) -> Self {
-        self.id = id;
+    pub fn id(mut self, id: impl Into<Id>) -> Self {
+        self.id = id.into();
         self
     }
 
--- a/crates/egui/src/containers/resize.rs
+++ b/crates/egui/src/containers/resize.rs
@@ -61,8 +61,8 @@ impl Default for Resize {
 impl Resize {
     /// Assign an explicit and globally unique id.
     #[inline]
-    pub fn id(mut self, id: Id) -> Self {
-        self.id = Some(id);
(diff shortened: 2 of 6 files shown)


emilk:
I haven't written this down before (will do after writing this here first), but there is a reason for the current design:

An `Id` is supposed to be globally unique. This is ensured by having `Ui` construct a tree of _locally_ unique "ID sources" (anything implementing `Hash`, but often string literals in practice).

So when constructing a child-region of some kind (`Grid`, `CollapsingHeader`, …) you specify a _locally_ unique ID _source_. When constructing a top-level thing (e.g. an `Area` or `Window`) you specify a globally unique `Id`.

I want to make this somewhat clear in the API (though obviously I haven't succeeded) so the rough idea is this: The locally unique ID sources are more common, so we optimize the ergonomics for those. So for those we take anything `impl Hash`. For the globally unique `Id`s however, we force the user to explicitly type `Id::new(…)` or `Id::from(…)` or `ui.id().with(…)` so they think through the promise that they make to egui.

Does that make sense?

--- 

For the plot items the only use of the `Id` is to match against `PlotResponse::hovered_plot_item`, which is an `Option<Id>`, so you want an `Id` there somehow anyway, I believe.

creativcoder:
Ah I see

It's a little less clear that we want to create a unique global id when we have to create them from `Id::new()` or `Id::from()`, because I still ended up creating an `Id` passing a string lying around in my scope in one of my codebase.

Maybe a change in method name like `global_id` for containers and top level ui elements would make the intention more clear. Does that work?

Also a few places, some of the methods that assign an id have different names like: [collapsing_header](https://docs.rs/egui/latest/egui/containers/collapsing_header/struct.CollapsingHeader.html#method.id_source), they can be renamed to just `id()` like others.

Issue #4682, opened 2024-06-20
Example hello_world_par does not compile on Ubuntu


MikiGrit:
**Describe the bug**

I'm on Ubuntu 22.04 and attempt to compile `hello_world_par` example ends with error about platform not supported by winit. All other examples run without any issues. The same example compiles on Windows 11 also without any issues.

**To Reproduce**
```bash
git clone https://github.com/emilk/egui
cd egui/examples/hello_world_par
cargo run
```

**Expected behavior**

Example will compile and run.

**Screenshots**

Following is `cargo run` output shortened to show only error related part:
```bash
   Compiling winit v0.29.10
error: The platform you're compiling for is not supported by winit
  --> /home/miki/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.29.10/src/platform_impl/mod.rs:67:1
   |
67 | compile_error!("The platform you're compiling for is not supported by winit");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0432]: unresolved import `self::platform`
  --> /home/miki/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.29.10/src/platform_impl/mod.rs:26:15
   |
26 | pub use self::platform::*;
   |               ^^^^^^^^ could not find `platform` in `self`

error[E0432]: unresolved import `crate::platform_impl::PlatformIcon`
 --> /home/miki/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.29.10/src/icon.rs:1:5
  |
1 | use crate::platform_impl::PlatformIcon;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `PlatformIcon` in `platform_impl`

error[E0433]: failed to resolve: could not find `DeviceId` in `platform_impl`
   --> /home/miki/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.29.10/src/event.rs:616:42
    |
616 |         DeviceId(unsafe { platform_impl::DeviceId::dummy() })
    |                                          ^^^^^^^^ could not find `DeviceId` in `platform_impl`

error[E0433]: failed to resolve: could not find `WindowId` in `platform_impl`
[...]

PR #4684, opened 2024-06-20
Fix hello_world_par example on Linux

YgorSouza:
This example does not use the default features from eframe in order to avoid accesskit, which panics when run from multiple threads, so it must manually enable the other default features in order to compile correctly on Linux.

* Closes <https://github.com/emilk/egui/issues/4682>

### DIFF
--- a/examples/hello_world_par/Cargo.toml
+++ b/examples/hello_world_par/Cargo.toml
@@ -15,9 +15,16 @@ workspace = true
 eframe = { workspace = true, default-features = false, features = [
     # accesskit struggles with threading
     "default_fonts",
+    "wayland",
+    "x11",
     "wgpu",
 ] }
 env_logger = { version = "0.10", default-features = false, features = [
     "auto-color",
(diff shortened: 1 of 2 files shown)


examples/hello_world_par/Cargo.toml
```diff
@@ -15,9 +15,16 @@ workspace = true
...
     "wgpu",
 ] }
 env_logger = { version = "0.10", default-features = false, features = [
     "auto-color",
     "humantime",
 ] }
+# This is normally enabled by eframe/default, which is not being used here
+# because of accesskit, as mentioned above
+winit = { workspace = true, features = [
+    "default"
```
YgorSouza, line 29: The example works without this, but the window frame looks different, so I added it to make it match the other examples.

#### With winit/default

Screenshot from 2024-06-20 20-12-44

#### Without winit/default

Screenshot from 2024-06-20 20-13-41

emilk (approved):

PR #4689, opened 2024-06-22
doc(emath): Add `top_left` as an alias for `left_top`, etc.

zkldi:
# What

Adds `#[doc(alias = "top_left")]` as an alias for `left_top`, and so on for `right_top`, `right_bottom`, `left_bottom`. 

# Why

Extremely minor doc-only change, but I keep going to type "top left" to look for the top left of a rectangle.
I'm unsure whether this is just a british-english thing or an english-in-general thing, but `top left corner` is far more common than `left top corner`.
These doc aliases don't conflict with anything, and mean that rust-analyzer will suggest the correct function when I search for the wrong thing.

This improves ergonomics and discoverability in my opinion, even if not by much.

### DIFF
--- a/crates/emath/src/rect.rs
+++ b/crates/emath/src/rect.rs
@@ -542,6 +542,7 @@ impl Rect {
     }
 
     #[inline(always)]
+    #[doc(alias = "top_left")]
     pub fn left_top(&self) -> Pos2 {
         pos2(self.left(), self.top())
     }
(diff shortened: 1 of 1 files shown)


emilk (approved):
Nice, makes sense!

The reason for the order is for it to be consistently X,Y order everywhere, even though I agree "right_top" sounds weird in english

Issue #4128, opened 2024-03-03
Pan Zoom demo widgets are layered incorrectly


YgorSouza:
**Describe the bug**

![image](https://github.com/emilk/egui/assets/43298013/92d64bb0-dc09-4fe5-b380-195a7a9404b1)

**Desktop (please complete the following information):**
 - OS: Windows 11, Fedora 39
 - Browser: Firefox

ToppDev:
maybe related. The pan zoom items can be interacted with, even outside of their window. (tested on https://egui.rs)

https://github.com/emilk/egui/assets/17766093/22a70021-cfc4-4e88-b2cc-928f77c6cdd3

markusdd:
closing #4448 in favor of this, thank @YgorSouza

Creative0708:
This bug seems to be a side effect of a workaround to get the pan zoom widgets to display above the window itself.

https://github.com/emilk/egui/blob/2c00cb39912e1043e883bacfeaec5806b51b7841/crates/egui_demo_lib/src/demo/pan_zoom.rs#L113-L132

The areas that hold the widgets are given a layer order of `Order::Foreground`, which displays above _everything_ that is `Order::Middle`, `Order::Background`, etc. The demo windows are `Order::Middle`, so they render behind the widgets. The comments in the code snippet mention this.

If the widgets were given `Order::Middle` instead, moving the pan zoom window to the front (such as by clicking) would bring the widgets behind the pan zoom window:

https://github.com/emilk/egui/assets/104604363/dd1c183b-995d-4e75-b06e-c0ff3f9b0ad5

I'm not sure how this would be fixed without support for, say, parent/child layers. Possible feature request maybe?

YgorSouza:
Actually, I just noticed this was already discussed in the PR that implemented the demo: <https://github.com/emilk/egui/pull/3906#issuecomment-1949899356>.

PR #4690, opened 2024-06-23
Allow setting a layer as a sublayer of another

YgorSouza:
When the layers are reordered at the end of the frame, the sublayers are
placed directly above their respective parents. This allows having Areas
inside Windows, e.g., for the pan-zoom container.

* Closes <https://github.com/emilk/egui/issues/4128>

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -2374,6 +2374,17 @@ impl Context {
         self.memory_mut(|mem| mem.areas_mut().move_to_top(layer_id));
     }
 
+    /// Mark the `child` layer as a sublayer of `parent`.
+    ///
+    /// Sublayers are moved directly above the parent layer at the end of the frame. This is mainly
+    /// intended for adding a new [`Area`] inside a [`Window`].
+    ///
+    /// This currently only supports one level of nesting. If `parent` is a sublayer of another
+    /// layer, the behavior is unspecified.
+    pub fn set_sublayer(&self, parent: LayerId, child: LayerId) {
+        self.memory_mut(|mem| mem.areas_mut().set_sublayer(parent, child));
+    }
+
     /// Retrieve the [`LayerId`] of the top level windows.
     pub fn top_layer_id(&self) -> Option<LayerId> {
         self.memory(|mem| mem.areas().top_layer_id(Order::Middle))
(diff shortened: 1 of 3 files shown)


YgorSouza:
I don't know if this is the best API, but it is relatively simple and seems to solve the problem of the pan zoom demo being layered above everything.

https://github.com/emilk/egui/assets/43298013/b910950f-7362-4ba0-881c-10a63a5f2630



It does not solve the other problem mentioned in <https://github.com/emilk/egui/issues/4128#issuecomment-2001977126>.

crates/egui/src/context.rs
```diff
@@ -2374,6 +2374,17 @@ impl Context {
...
     }
 
+    /// Mark the `child` layer as a sublayer of `parent`.
+    ///
+    /// Sublayers are moved directly above the parent layer at the end of the frame. This is mainly
+    /// intended for adding a new [`Area`] inside a [`Window`].
+    ///
+    /// This currently only supports one level of nesting. If `parent` is a sublayer of another
+    /// layer, the behavior is unspecified.
+    pub fn set_sublayer(&self, parent: LayerId, child: LayerId) {
```
YgorSouza, line 2384: Maybe this could be called something like `set_parent_window` to be clearer? Technically the parent doesn't have to be a Window, but that is the intended use case, and a more complicated hierarchy would not work well with the current reordering algorithm. I'm not really sure what to call it.

emilk: I like the current name, but we should document that this is a sticky setting, and therefor there should also be some way to clear it. Or we make it an immediate-mode function instead and clear it at the end of each frame

crates/egui/src/memory.rs
```diff
@@ -1042,27 +1047,60 @@ impl Areas {
...
             wants_to_be_on_top,
+            sublayers,
             ..
         } = self;
 
         std::mem::swap(visible_last_frame, visible_current_frame);
         visible_current_frame.clear();
         order.sort_by_key(|layer| (layer.order, wants_to_be_on_top.contains(layer)));
         wants_to_be_on_top.clear();
+        for (parent, children) in sublayers {
```
emilk, line 1089:
```suggestion

        // For all layers with sublayers, put the sublayers directly after the parent layer:
        for (parent, children) in sublayers {
```

emilk (commented):
I like this direction!

crates/egui/src/context.rs
YgorSouza: I actually meant for it to be immediate, like the wants_to_be_on_top set. I'm going to fix that.

emilk (approved):

emilk:
Thanks!

emilk (approved):

PR #4695, opened 2024-06-23
add code

Its-Just-Nans:
* Closes <https://github.com/emilk/egui/issues/3462>

#3462 panics when when hit the bound like `f64::MAX` 

This PR "blocks" zoom if to big

Linked to
- comment in https://github.com/emilk/egui/pull/4637#issuecomment-2154848767 
- #3462

### DIFF
--- a/crates/egui_plot/src/transform.rs
+++ b/crates/egui_plot/src/transform.rs
@@ -117,14 +117,22 @@ impl PlotBounds {
 
     #[inline]
     pub fn expand_x(&mut self, pad: f64) {
-        self.min[0] -= pad;
-        self.max[0] += pad;
+        if (self.min[0] - pad).is_finite() {
+            self.min[0] -= pad;
+        }
+        if (self.max[0] + pad).is_finite() {
+            self.max[0] += pad;
+        }
     }
 
     #[inline]
     pub fn expand_y(&mut self, pad: f64) {
-        self.min[1] -= pad;
-        self.max[1] += pad;
+        if (self.min[1] - pad).is_finite() {
+            self.min[1] -= pad;
+        }
+        if (self.max[1] + pad).is_finite() {
+            self.max[1] += pad;
+        }
     }
 
     #[inline]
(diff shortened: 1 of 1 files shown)


crates/egui_plot/src/transform.rs
```diff
@@ -117,14 +117,22 @@ impl PlotBounds {
...
     #[inline]
     pub fn expand_x(&mut self, pad: f64) {
-        self.min[0] -= pad;
-        self.max[0] += pad;
+        if (self.min[0] - pad).is_finite() {
+            self.min[0] -= pad;
+        }
+        if (self.max[0] + pad).is_finite() {
+            self.max[0] += pad;
+        }
```
emilk, line 125: Alternatively:

```
self.min[0] = (self.min[0] - pad).clamp(f64::MIN, f64::MAX);
self.max[0] = (self.max[0] + pad).clamp(f64::MIN, f64::MAX);
```

or add a `self.clamp_to_finite();` helper to call at the end of these functions that does


```
fn clamp_to_finite(&mut self) {
        for d in 0..2 {
			self.min[d] = self.min[d].clamp(f64::MIN, f64::MAX);
			self.max[d] = self.max[d].clamp(f64::MIN, f64::MAX);
        }
}
```

Its-Just-Nans: Hi,

I don't think so
The goal of these if is to ensure no propagation of `NaN`

The docs said if NaN is input of clamp, it will return NaN - a problem

emilk:
```rs
    pub fn expand_x(&mut self, pad: f64) {
        if pad.is_finite() {
		        self.min[0] -= pad;
		        self.max[0] += pad;
				self.clamp_to_finite();
		}
	}


	fn clamp_to_finite(&mut self) {
	        for d in 0..2 {
				self.min[d] = self.min[d].clamp(f64::MIN, f64::MAX);
	            if self.min[d].is_nan() { self.min[d] = 0.0; }
	
				self.max[d] = self.max[d].clamp(f64::MIN, f64::MAX);
	            if self.max[d].is_nan() { self.max[d] = 0.0; }
	        }
	}
```

Its-Just-Nans:
Zoom is lockec but when zoom in, it's not working properly

crates/egui_plot/src/transform.rs
```diff
@@ -115,16 +115,39 @@ impl PlotBounds {
         self.max[1] = self.max[1].max(y);
     }
 
+    #[inline]
+    fn clamp_to_finite(&mut self) {
+        for d in 0..2 {
+            self.min[d] = self.min[d].clamp(f64::MIN, f64::MAX);
+            if self.min[d].is_nan() {
+                println!("min[d] is nan");
```
emilk, line 123: rm

Its-Just-Nans: ups

removed!

Issue #3657, opened 2023-11-28
`any_popup_open()` remains `true` even when there are no popups


juancampa:
**Describe the bug**

If a popup that is being shown is removed without calling `Memory::close_popup`, egui will think that the popup is still being shown, for example
`Memory::any_popup_open()` returns `true`.

Video of the problem:

https://github.com/emilk/egui/assets/1410520/48113d12-09d9-4dba-9c80-f0be2a552ae4

You'd think the solution is to always call `close_popup()` but in some cases that's not practical, for example, the app stops rendering that part of the UI.

Proposed solution: clear `Memory::popup` at the end of a frame if `is_popup_open` was not called with the id of the current one (assume the popup was "abandoned"). Probably rename the function to something that reflects this behavior, e.g. `retain_popup`, `check_and_keep_popup_open`?

**To Reproduce**
1. Open a popup (e.g. color picker, drop-down, etc)
2. Without manually closing the popup, stop rendering it (i.e. don't hit that code path at all)
3. `Memory::any_popup_open()` is `true`

**Expected behavior**

`Memory::any_popup_open()` is `false`

**Screenshots**

**Desktop (please complete the following information):**
 - Version 0.23.0

juancampa:
The reason I would like `any_popup_open` to accurately reflect the state of popups is because I want to prevent some interactions while popups or menus are open. Users expect to click anywhere to close a popup.

PR #4697, opened 2024-06-23
Reset `Memory::popup` to `None` if a popup was abandoned

juancampa:
* Closes https://github.com/emilk/egui/issues/3657

## BEFORE

https://github.com/emilk/egui/assets/1410520/d35168ae-cde8-479a-bb4a-79873b5ce685

## AFTER

https://github.com/emilk/egui/assets/1410520/e9f367d8-38f0-4be1-b3d9-fda505c7b625

Notice that `WRONG` flashes for one frame. That's expected because on the frame the combo box disappears, `any_popup_open` still returns true, since we don't know it has disappeared until the end of the frame.

### DIFF
--- a/crates/egui/src/memory.rs
+++ b/crates/egui/src/memory.rs
@@ -2,6 +2,7 @@
 
 use ahash::{HashMap, HashSet};
 use epaint::emath::TSTransform;
+use std::sync::atomic::AtomicBool;
 
 use crate::{
     area, vec2, EventFilter, Id, IdMap, LayerId, Order, Pos2, Rangef, RawInput, Rect, Style, Vec2,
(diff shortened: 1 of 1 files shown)


crates/egui/src/memory.rs
```diff
@@ -80,8 +81,9 @@ pub struct Memory {
 
     /// Which popup-window is open (if any)?
     /// Could be a combo box, color picker, menu etc.
+    /// The bool is used to detect if a popup stopped showing before calling close_popup
     #[cfg_attr(feature = "persistence", serde(skip))]
-    popup: Option<Id>,
+    popup: Option<(Id, std::sync::Arc<AtomicBool>)>,
```
juancampa, line 86: I'm using an `AtomicBool` here because I didn't want to change `is_popup_open` to take `&mut self`.

An alternative is to add another function, something like `retain_popup(&mut self)` that explicitly sets a `retained` bool to true. But would will be a breaking change and existing apps will need to call it or they popups will last for one frame.

I'm open to suggestions.

emilk: it's not at all obvious from reading this what the bool is for though., or why it is behind an `Arc`.

It seems to answer the question "Was `is_popup_open` called for this this frame?", i.e. "Does anyone care that this is open?".

I suggest making a `struct` with a name field so the motivation behind this can be documented.

We should also document that if `is_popup_open` is not called during a frame, the popup auto-closes.
Since this in fact makes it change state, I think making it take a `&mut` makes sense.

So I suggest

```
struct OpenPopup {
    id: Id,
    shown_this_frame: bool,
}

…

popup: Option<OpenPopup>,
```

juancampa:
Closing in favor of https://github.com/emilk/egui/pull/5814

PR #4707, opened 2024-06-25
add forget_image function to Image

JustFrederik:
Image::source returns a modified ImageSource when the image is a gif.

### DIFF
--- a/crates/egui/src/widgets/image.rs
+++ b/crates/egui/src/widgets/image.rs
@@ -289,6 +289,18 @@ impl<'a> Image<'a> {
     }
 
     #[inline]
+    pub fn forget_image(&self, ctx: &Context) {
+        let uri = match &self.source {
+            ImageSource::Uri(uri) => Some(uri),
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/image.rs
```diff
@@ -288,6 +288,18 @@ impl<'a> Image<'a> {
...
     }
 
+    #[inline]
+    pub fn forget_image(&self, ctx: &Context) {
+        let uri = match &self.source {
+            ImageSource::Uri(uri) => Some(uri),
+            ImageSource::Bytes { uri, .. } => Some(uri),
+            _ => None,
+        };
+        if let Some(uri) = uri {
```
emilk, line 298:
```suggestion
        if let Some(uri) = self.source.uri() {
```

emilk (commented):
Can you explain why would you want this? An `Image` is not meant to be kept around - it is just a builder to be thrown out the same frame, at least usually.

PR #4713, opened 2024-06-27
Cancel DragValue edit if Escape is pressed

YgorSouza:
Since Escape takes focus away from the text field, it makes more sense for it to cancel the change rather than confirm it.

This only matters if `update_while_editing` is set to false, of course.


emilk (approved):
Nice!

PR #4716, opened 2024-06-27
Add basic test for egui accesskit output.

Wcubed:
For a project I am currently attempting to use the acceskit output of egui to preform end-to-end integration tests of my application. However, the output is currently quite basic (buttons for example don't report that they are disabled).

Before adding new features to the acceskit integration, I thought it would be a good idea to add some testing to egui's accesskit output.
I have started with a simple test for button texts, to get feedback on whether this is the correct direction to go in.

### DIFF
--- a/crates/egui/tests/accesskit.rs
+++ b/crates/egui/tests/accesskit.rs
@@ -3,6 +3,38 @@
 use accesskit::Role;
 use egui::{Context, RawInput};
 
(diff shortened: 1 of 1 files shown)


emilk (approved):
This is great; thank you! And I am too interested in using accesskit for this purpose, so I'll be following your work :)

Issue #4563, opened 2024-05-28
Input not focused when selecting text


jprochazk:
Repro:
1. Open Demos -> Widget Gallery on the [web demo](https://egui.rs)
2. Run `setInterval(() => console.log(document.activeElement), 1000)` in a devtools console
3. Type something into the first text field
4. Click somewhere outside of the text field to blur it
5. Select the text by dragging over it - make sure the text field is blurred!

Expected behavior: `runner.mutable_text_under_cursor` is set to `true`, causing the `<input>` element to be the `document.activeElement`
Actual behavior: `runner.mutable_text_under_cursor` is `false`, and `document.activeElement == document.body`

PR #4718, opened 2024-06-27
eframe web: ignore keyboard events unless canvas has focus

emilk:
* Fixes https://github.com/rerun-io/rerun/issues/6638
* Related? https://github.com/emilk/egui/issues/4563

This improves how an eframe canvas works inside of a larger web page, and how it works when there are multiple eframe apps in the same page.

`eframe` will set `tabindex="0"` on the canvas automatically, making it focusable.
It will also set `outline: none` on the CSS, so the focused canvas won't have an ugly outline.

## Breaking changes
You may wanna add this to your `index.html` to give the canvas focus on startup:
```js
document.getElementById("the_canvas_id").focus();
```

## Test setup
```sh
./scripts/build_demo_web.sh
./scripts/start_server.sh
open http://localhost:8888/multiple_apps.html
```
Then open the "Input Event History" and "Text Edit" windows

## Tested
* Chromium
	* [x] drag-and-drop of files
	* Test both when a `TextEdit` is focused and when it is not:
	  * [x] `Event::Key`
	  * [x] `Event::Text`
	  * [x] copy-cut-paste
	  * [x] Wheel scroll
	* [x] `Event::PointerGone`
	* [x] Mouse drag
	* [x] Mouse click
	* [x] Mouse right-click
	* [x] Defocus all eframe canvas, and then start typing text
* [x] Firefox (all of the above)
* [x] Desktop Safari (all of the above)
* [x] Mobile Safari

## Future work (pre-existing issues)
* https://github.com/emilk/egui/issues/4723
* https://github.com/emilk/egui/issues/4724
* https://github.com/emilk/egui/issues/4725
* https://github.com/emilk/egui/issues/4726


jleibs:
This solves half of https://github.com/rerun-io/rerun/issues/6638

In a notebook context, Rerun no longer gets events when typing in notebook cells. However, it now fails to respect usage of any keyboard shortcuts at all.

Selecting a text-entry box works, but as soon as you leave a text-entry context it appears all events only get processed by the outer notebook environment.

jprochazk:
Right click does not work in a notebook context (both classic and lab). For example, right click on a view in the blueprint panel should bring up its context menu, but instead it always opens the notebook's context menu

emilk:
@jprochazk please see if the latest commit ([616fa2b](https://github.com/emilk/egui/pull/4718/commits/616fa2b371c92cb679da9736684c0194b5d16e0a)) fixes it

jprochazk:
I can confirm it's fixed with that commit.

PR #4719, opened 2024-06-27
Introduce type parameter for plot points

FabianLippoldZR:
In order to supply custom data to plot points for use mainly in the label_formatter this introduces a new field which stores userdata.

Hovering over a plot point | Hovering over anything else
-|-
image|image

Of course this requires PlotPoint to have a generic type parameter which then requires lots of other structs to be generic too.

The userdata field is optional, since a PlotPoint can be a Point supplied by the user when constructing a plot item, but it can also be generated by the library.

A demo for this is implemented as you can see in the screenshot.

Of course I would like there to be an easier way to do this, but currently I'm not aware of any.

### DIFF
--- a/crates/egui_demo_lib/src/demo/plot_demo.rs
+++ b/crates/egui_demo_lib/src/demo/plot_demo.rs
@@ -21,6 +21,7 @@ enum Panel {
     Interaction,
     CustomAxes,
     LinkedAxes,
+    Userdata,
 }
 
 impl Default for Panel {
@@ -41,6 +42,7 @@ pub struct PlotDemo {
     interaction_demo: InteractionDemo,
     custom_axes_demo: CustomAxesDemo,
     linked_axes_demo: LinkedAxesDemo,
+    userdata_demo: UserdataDemo,
     open_panel: Panel,
 }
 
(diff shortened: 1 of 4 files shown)


Mingun:
At least, you can make generic parameters optional (`<T = ()>`)

FabianLippold:
Yes thats a good Idea, but the defaults only seem to apply when you use \<T> and not T itself.

See https://github.com/rust-lang/rust/issues/36980.

image
image

emilk (changes requested):
I'm not a big fan of adding templates to everything. It not only makes the core more complex, it also means monomorphization will duplicate the code in the binary, ballooning e.g. the .wasm size.

If you want to attach user-data to points, I suggest you report `Points::id` together with the point index (usize) to the user. The user can then look up the data they need with a hashmap or similar.

FabianLippold:
@emilk I tried it and it works with only a few modifications to the library.
But it is quite cumbersome to use, since you need an Arc<Mutex<...>> for the label_formatter and the plot_ui to be able to use the same HashMap. This also requires you to create or re-fill a HashMap on every draw...

Is this what you had in mind or is there a simpler way to do this?

emilk:

PR #4720, opened 2024-06-27
Add `Image::uri()`

rustbasic:
`uri()` for image

The advantages of this are :

1. No need for `ctx()` and `source()`
2. Displays `name.gif#0` as `name.gif`

### DIFF
--- a/crates/egui/src/widgets/image.rs
+++ b/crates/egui/src/widgets/image.rs
@@ -283,7 +283,9 @@ impl<'a> Image<'a> {
         }
     }
 
-    /// Get the `uri` that this image.
+    /// Returns the URI of the image.
+    ///
+    /// For GIFs, returns the URI without the frame number.
     #[inline]
     pub fn uri(&self) -> Option<&str> {
         let uri = self.source.uri()?;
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/image.rs
```diff
@@ -283,6 +283,18 @@ impl<'a> Image<'a> {
...
+    #[inline]
+    pub fn uri(&self) -> Option<&str> {
+        let mut uri = self.source.uri();
+
+        if is_gif_uri(uri.unwrap_or_default()) {
+            if let Ok((gif_uri, _index)) = decode_gif_uri(uri.unwrap_or_default()) {
+                uri = Some(gif_uri);
+            }
+        }
+        uri
```
emilk, line 295:
```suggestion
        let uri = self.source.uri()?;
        
        if let Ok((gif_uri, _index)) = decode_gif_uri(uri) {
            Some(gif_uri)
        } else {
            Some(uri)
        }
```

emilk: And add a docstring for it explaining the gif stuff

emilk (changes requested):

emilk (approved):

PR #4728, opened 2024-06-28
Add `clamp_to_range` option to DragValue, rename `clamp_range` to `range` (deprecating the former)

Wumpf:
Adds the ability to have `DragValue` not clamp values its presented with and instead apply clamping only once there's any user input.

In action:

https://github.com/emilk/egui/assets/1220815/af42fd67-86d0-4364-8ae6-48a2ec15646a

Alternative name could be `only_clamp_on_change`, not entirely certain which one is better 🤔

### DIFF
--- a/crates/egui/src/widgets/drag_value.rs
+++ b/crates/egui/src/widgets/drag_value.rs
@@ -89,9 +89,8 @@ impl<'a> DragValue<'a> {
 
     /// Sets valid range for the value.
     ///
-    /// By default, the drag value does not show values outside this range,
-    /// If you want don't want clamp incoming and outgoing values, disable this behavior using [`Slider::clamp_to_range`].
-    /// Setting it to `false` will allows users to enter values outside the range by clicking the value and editing it.
+    /// By default all values are clamped to this range, even when not interacted with.
+    /// You can change this behavior by passing `false` to [`Slider::clamp_to_range`].
     #[deprecated = "Use `range` instead"]
     #[inline]
     pub fn clamp_range<Num: emath::Numeric>(mut self, range: RangeInclusive<Num>) -> Self {
@@ -101,16 +100,15 @@ impl<'a> DragValue<'a> {
 
     /// Sets valid range for dragging the value.
     ///
-    /// By default, the drag value does not show values outside this range,
-    /// If you want don't want clamp incoming and outgoing values, disable this behavior using [`Slider::clamp_to_range`].
-    /// Setting it to `false` will allows users to enter values outside the range by clicking the value and editing it.
+    /// By default all values are clamped to this range, even when not interacted with.
+    /// You can change this behavior by passing `false` to [`Slider::clamp_to_range`].
     #[inline]
     pub fn range<Num: emath::Numeric>(mut self, range: RangeInclusive<Num>) -> Self {
         self.range = range.start().to_f64()..=range.end().to_f64();
         self
     }
 
-    /// If set to `true`, all incoming and outgoing values will be clamped to the slider range.
+    /// If set to `true`, all incoming and outgoing values will be clamped to the sliding [`Self::range`] (if any).
     ///
     /// If set to `false`, a value outside of the range that is set programmatically or by user input will not be changed.
     /// Dragging will be restricted to the range regardless of this setting.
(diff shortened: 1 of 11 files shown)


emilk (changes requested):
I like the feature, but am not sold on the API. Here are three alternatives:

### A: follow Slider
`egui::Slider` always takes a range, but doesn't clamp the input unless the user also calls `.clamp_to_range(true)`. We could do the same with `DragValue` (changing its behavior) and have something like `DragValue::new(…).range(…).clamp_to_range(true)`, thus more closely mirroring the `Slider` API. It would mean renaming `clamp_range` to `range` (with a deprecation notice) and changing the default behavior (current default: clamping all input, new default 

### B: have two range functions
`DragValue` doesn't have a range to clamp to by default. This means `DragValue::new(…).only_clamp_on_input(…)` is non-sensical.

If we split `DragValue::clamp_range` into two different functions, e.g. `DragValue::clamp_io_range` and `clamp_output_range`, we force the user to make a choice of what behavior they want.

### C: `clamp_range` with enum
Another alternative is `.clamp_range(a..=b, egui::ClampBehavior::Always)` vs `.clamp_range(a..=b, egui::ClampBehavior::OnEdit)` or similar. We could use the same enum in `Slider`


----

Out of the three alternatives, I'm leaning weakly towards A. WDYT?

Wumpf:
👍 agreed! A sounds best by far to me

Wumpf:
uh the slider docs are actually inconsistent

```
/// By default, the slider can still show values outside this range,
/// and still allows users to enter values outside the range by clicking the slider value and editing it.
/// If you want to clamp incoming and outgoing values, use [`Slider::clamp_to_range`].
```

but then
```
    /// If set to `true`, all incoming and outgoing values will be clamped to the slider range.
    /// Default: `true`.
    #[inline]
    pub fn clamp_to_range(mut self, clamp_to_range: bool) -> Self {
```

`clamp_to_range` does indeed default to `true`. So the slider's docs aren't accurate.
Fixing the docs and following suite on the `DragValue`, making it clamp by default

Wumpf:
tested against egui demo and updated rerun version

crates/egui/src/widgets/drag_value.rs
```diff
@@ -87,10 +87,37 @@ impl<'a> DragValue<'a> {
...
+    pub fn clamp_range<Num: emath::Numeric>(mut self, range: RangeInclusive<Num>) -> Self {
+        self.range = range.start().to_f64()..=range.end().to_f64();
+        self
+    }
+
+    /// Sets valid range for dragging the value.
+    ///
+    /// By default, the drag value does not show values outside this range,
+    /// If you want don't want clamp incoming and outgoing values, disable this behavior using [`Slider::clamp_to_range`].
+    /// Setting it to `false` will allows users to enter values outside the range by clicking the value and editing it.
```
emilk, line 106:
```suggestion
    /// By default all values are clamped to this range, even when not interacted with.
    /// You can change this behavior by passing `false` to [`Slider::clamp_to_range`].
```

crates/egui/src/widgets/drag_value.rs
```diff
@@ -87,10 +87,37 @@ impl<'a> DragValue<'a> {
...
+    #[inline]
+    pub fn range<Num: emath::Numeric>(mut self, range: RangeInclusive<Num>) -> Self {
+        self.range = range.start().to_f64()..=range.end().to_f64();
+        self
+    }
+
+    /// If set to `true`, all incoming and outgoing values will be clamped to the slider range.
```
emilk, line 113:
```suggestion
    /// If set to `true`, all incoming and outgoing values will be clamped to the slider [`Self::range`] (if any).
```

crates/egui/src/widgets/slider.rs
```diff
@@ -48,9 +48,9 @@ pub enum SliderOrientation {
 /// Control a number with a slider.
 ///
 /// The slider range defines the values you get when pulling the slider to the far edges.
-/// By default, the slider can still show values outside this range,
-/// and still allows users to enter values outside the range by clicking the slider value and editing it.
-/// If you want to clamp incoming and outgoing values, use [`Slider::clamp_to_range`].
+/// By default, the slider does not show values outside this range,
+/// If you want don't want clamp incoming and outgoing values, disable this behavior using [`Slider::clamp_to_range`].
+/// Setting it to `false` will allows users to enter values outside the range by clicking the slider value and editing it.
```
emilk, line 53:
```suggestion
    /// By default all values are clamped to this range, even when not interacted with.
    /// You can change this behavior by passing `false` to [`Slider::clamp_to_range`].
```

emilk (approved):
Awesome!

Issue #4490, opened 2024-05-12
Custom light & dark themes on system theme change


kernelkind:
**Is your feature request related to a problem? Please describe.**
Currently, when the system theme changes, the `Visuals` theme is switched to the hard-coded `egui::Visuals::dark()` or `egui::Visuals::light()`. I would like for there to be an option to use custom `Visuals` for dark and light themes.

This seems to be where the system theme change is detected and the egui visuals are set:
https://github.com/emilk/egui/blob/c3f386aa301f26106397c4e14434bd5a734ba6b6/crates/eframe/src/native/epi_integration.rs#L271

And this seems to be where the hard-coded `Visuals` are set:
https://github.com/emilk/egui/blob/c3f386aa301f26106397c4e14434bd5a734ba6b6/crates/eframe/src/epi.rs#L501-L506

**Describe the solution you'd like**
There should be some way to pass a `Visuals` that corresponds to the dark theme, and another `Visuals` which corresponds to the light theme. Then, when a theme change is needed, those custom themes would be used instead of the hard-coded `egui::Visuals::dark()` or `egui::Visuals::light()`. 

**Describe alternatives you've considered**
None

**Additional context**
I am working on https://github.com/damus-io/notedeck and we would like to support custom theming at some point. It's not a high priority at the moment, and I would be open to submitting a PR for this feature in the future.

emilk:
100% agree with this.

The global style is stored in `Options::style`, and I suggest we split that into three fields:

```rs
struct Options {
    pub dark_mode_style: Arc<Style>,
    pub light_mode_style: Arc<Style>,
    pub theme: Theme, // Dark or Light
    …
}
```

with `enum Theme { Dark, Light }` (move the one in `eframe` into `egui`).

PRs welcome!

rustbasic:
Please also refer to issue #4312 while working.

* Related #4312

bash:
@rustbasic Do you have an minimal example that I can run to verify that visuals_mut works as expected for you?

rustbasic:
> @rustbasic Do you have an minimal example that I can run to verify that visuals_mut works as expected for you?

If you call `dark()` or `light()` to change the theme, both will be initialized.
I think you need to save the settings for both of them separately so that they don't get initialized.

If you do it like the example, once the settings are set, it will be initialized when you change the theme.

Since the recent update, even the settings are initialized once.

Since `text_cursor` is easy to check with your eyes, I used `text_cursor` in the example.

```
[...]

PR #4744, opened 2024-06-30
Make Light & Dark Visuals Customizable When Following The System Theme

bash:
* Closes <https://github.com/emilk/egui/issues/4490>

---

Unfortunately, this PR contains a bunch of breaking changes because `Context` no longer has one style, but two. I could try to add some of the methods back if that's desired.

The most subtle change is probably that `style_mut` mutates both the dark and the light style (which from the usage in egui itself felt like the right choice but might be surprising to users).

I decided to deviate a bit from the data structure suggested in the linked issue.
Instead of this:
```rust
pub theme: Theme, // Dark or Light
pub follow_system_theme: bool, // Change [`Self::theme`] based on `RawInput::system_theme`?
```

I decided to add a `ThemePreference` enum and track the current system theme separately.
This has a couple of benefits:
* The user's theme choice is not magically overwritten on the next frame.
* A widget for changing the theme preference only needs to know the `ThemePreference` and not two values.
* Persisting the `theme_preference` is fine (as opposed to persisting the `theme` field which may actually be the system theme).

The `small_toggle_button` currently only toggles between dark and light (so you can never get back to following the system). I think it's easy to improve on this in a follow-up PR :)
I made the function `pub(crate)` for now because it should eventually be a method on `ThemePreference`, not `Theme`.

To showcase the new capabilities I added a new example that uses different "accent" colors in dark and light mode:

A screenshot of egui's widget gallery demo in dark mode using a purple accent color instead of the default blue accent

A screenshot of egui's widget gallery demo in light mode using a green accent color instead of the default blue accent


crates/egui/src/context.rs
```diff
@@ -1610,30 +1634,93 @@ impl Context {
...
+    /// The [`Style`] used by all new windows, panels etc. in dark mode.
+    /// Use [`Self::set_theme`] to choose between dark and light mode.
     ///
-    /// You can also change this using [`Self::style_mut`]
+    /// You can also change this using [`Self::dark_style_mut`]
     ///
     /// You can use [`Ui::style_mut`] to change the style of a single [`Ui`].
-    pub fn set_style(&self, style: impl Into<Arc<Style>>) {
-        self.options_mut(|opt| opt.style = style.into());
+    pub fn set_dark_style(&self, style: impl Into<std::sync::Arc<crate::Style>>) {
```
emilk, line 1668: To avoid having too large of an API surface, I suggest we combine `set_dark_style` and `set_light_style` with `ctx.set_style(theme, style)`. Same with the getters.

bash: Good idea! I had to rename the current `style` method to `active_style` as a consequence (this broke a lot of places, but I think `active_style` is a lot clearer than just `style`.
I also had to rename the currently named `style_mut` (which mutates both dark and light style) to `styles_mut`.

crates/egui/src/memory.rs
```diff
@@ -168,24 +168,28 @@ impl FocusDirection {
...
+    pub theme_preference: ThemePreference,
 
-    /// Which theme to use in case [`Self::follow_system_theme`] is set
+    /// Which theme to use in case [`Self::theme_preference`] is [`ThemePreference::System`]
     /// and egui fails to detect the system theme.
     ///
     /// Default: [`crate::Theme::Dark`].
     pub fallback_theme: Theme,
 
     /// Used to detect changes in system theme
```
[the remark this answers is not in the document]
bash: Yes, the deciding thing for the theme is ThemePreference which has both a `Dark` and a `Light` variant for forcing a theme.

If ThemePreference is set to `System`, then `fallback_theme` is the fallback in case we don't have a system theme [source](https://github.com/bash/egui/blob/2ea927dbac6a3102381e86d17e0b54f23e2905f6/crates/egui/src/memory/mod.rs#L313-L319).

crates/egui/src/containers/panel.rs
```diff
@@ -986,12 +986,12 @@ impl TopBottomPanel {
             let collapsed_height = PanelState::load(ctx, collapsed_panel.id)
                 .map(|state| state.rect.height())
                 .or(collapsed_panel.default_height)
-                .unwrap_or_else(|| ctx.style().spacing.interact_size.y);
+                .unwrap_or_else(|| ctx.active_style().spacing.interact_size.y);
```
emilk, line 989: I think we should go back to calling this just `style` - this produces a lot of diff for little gain.

crates/egui/src/context.rs
```diff
@@ -1612,31 +1612,87 @@ impl Context {
...
     /// ```
     /// # let mut ctx = egui::Context::default();
-    /// ctx.style_mut(|style| {
+    /// ctx.styles_mut(|style| {
     ///     style.spacing.item_spacing = egui::vec2(10.0, 20.0);
     /// });
     /// ```
-    pub fn style_mut(&self, mutate_style: impl FnOnce(&mut Style)) {
-        self.options_mut(|opt| mutate_style(std::sync::Arc::make_mut(&mut opt.style)));
+    pub fn styles_mut(&self, mut mutate_style: impl FnMut(&mut Style)) {
```
emilk, line 1653: The naming of this is a bit too close to `style_mut` imho. Maybe call it `all_styles_mut`?

crates/egui/src/context.rs
```diff
@@ -1612,31 +1612,87 @@ impl Context {
...
+    /// Mutate the [`Style`] used by all subsequent windows, panels etc in dark mode.
+    ///
+    /// Example:
+    /// ```
+    /// # let mut ctx = egui::Context::default();
+    /// ctx.style_mut(egui::Theme::Dark, |style| {
+    ///     style.spacing.item_spacing = egui::vec2(10.0, 20.0);
+    /// });
+    /// ```
+    pub fn style_mut(&self, theme: Theme, mutate_style: impl FnOnce(&mut Style)) {
```
emilk, line 1677: In the interest of minimizing breaking changes I suggest we keep the old `style_mut` to mutate the current active style, and then we can have a `style_mut_of` (or something) for mutating a specific theme selected by argument

crates/egui/src/context.rs
```diff
@@ -1612,31 +1612,87 @@ impl Context {
...
-        self.options_mut(|opt| mutate_style(std::sync::Arc::make_mut(&mut opt.style)));
+    pub fn styles_mut(&self, mut mutate_style: impl FnMut(&mut Style)) {
+        self.options_mut(|opt| {
+            mutate_style(std::sync::Arc::make_mut(&mut opt.dark_style));
+            mutate_style(std::sync::Arc::make_mut(&mut opt.light_style));
+        });
+    }
+
+    /// The [`Style`] used by all subsequent windows, panels etc.
+    pub fn style(&self, theme: Theme) -> Arc<Style> {
```
emilk, line 1661: As noted elsewhere, the `active_style` produces a big diff (=a lot of breaking changes). I suggest we instead call this `style_of`. I suspect this will be less used anyway, so keeping the often used function name shorter makes sense.

bash: I have additionally added back `set_style` (which sets the currently active style). Should I also add back `set_visuals` now that we have the `_of` suffix?

bash:
All ready for re-review :)

crates/egui/src/memory/mod.rs
```diff
@@ -168,24 +168,29 @@ impl FocusDirection {
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 #[cfg_attr(feature = "serde", serde(default))]
 pub struct Options {
-    /// The default style for new [`Ui`](crate::Ui):s.
+    /// The default style for new [`Ui`](crate::Ui):s in dark mode.
     #[cfg_attr(feature = "serde", serde(skip))]
```
emilk, line 172: We're inconsistently only serializing the light style. I suggest we serialize neither (like before)

crates/egui/src/memory/mod.rs
```diff
@@ -168,24 +168,29 @@ impl FocusDirection {
...
-    pub(crate) style: std::sync::Arc<Style>,
+    pub(crate) dark_style: std::sync::Arc<Style>,
 
-    /// Whether to update the visuals according to the system theme or not.
+    /// The default style for new [`Ui`](crate::Ui):s in light mode.
+    pub(crate) light_style: std::sync::Arc<Style>,
```
emilk, line 176:
```suggestion
    pub light_style: std::sync::Arc<Style>,
```

crates/egui/src/memory/mod.rs
```diff
@@ -168,24 +168,29 @@ impl FocusDirection {
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 #[cfg_attr(feature = "serde", serde(default))]
 pub struct Options {
-    /// The default style for new [`Ui`](crate::Ui):s.
+    /// The default style for new [`Ui`](crate::Ui):s in dark mode.
     #[cfg_attr(feature = "serde", serde(skip))]
-    pub(crate) style: std::sync::Arc<Style>,
+    pub(crate) dark_style: std::sync::Arc<Style>,
```
emilk, line 173:
```suggestion
    pub dark_style: std::sync::Arc<Style>,
```

emilk (approved):

PR #4748, opened 2024-07-01
Add a menu button with text and image 

NicolasBircksZR:
In order to make it possible that all buttons can have the same style with a text and an image I created a `menu_text_image_button`.

It works the same way as all the other menu buttons.

### DIFF
--- a/crates/egui/src/menu.rs
+++ b/crates/egui/src/menu.rs
@@ -118,6 +118,24 @@ pub fn menu_custom_button<R>(
     stationary_menu_button_impl(ui, button, Box::new(add_contents))
 }
 
+/// Construct a top level menu with an image in a menu bar. This would be e.g. "File", "Edit" etc.
+///
+/// Responds to primary clicks.
+///
+/// Returns `None` if the menu is not open.
+#[deprecated = "Use `menu_custom_button` instead"]
+pub fn menu_image_button<R>(
+    ui: &mut Ui,
+    image_button: ImageButton<'_>,
+    add_contents: impl FnOnce(&mut Ui) -> R,
+) -> InnerResponse<Option<R>> {
+    stationary_menu_button_impl(
+        ui,
+        Button::image(image_button.image),
+        Box::new(add_contents),
+    )
+}
+
 /// Construct a nested sub menu in another menu.
 ///
 /// Opens on hover.
--- a/crates/egui/src/menu.rs
+++ b/crates/egui/src/menu.rs
@@ -118,6 +118,19 @@ pub fn menu_image_button<R>(
     stationary_menu_image_impl(ui, image_button, Box::new(add_contents))
 }
 
+/// Construct a top level menu with an image and a text in a menu bar. This would be e.g. "File", "Edit" etc.
+///
+/// Responds to primary clicks.
+///
+/// Returns `None` if the menu is not open.
+pub fn menu_text_image_button<R>(
+    ui: &mut Ui,
+    image_button: Button<'_>,
+    add_contents: impl FnOnce(&mut Ui) -> R,
+) -> InnerResponse<Option<R>> {
+    stationary_menu_text_image_impl(ui, image_button, Box::new(add_contents))
+}
+
 /// Construct a nested sub menu in another menu.
 ///
 /// Opens on hover.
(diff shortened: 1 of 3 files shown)


crates/egui/src/ui.rs
```diff
@@ -2567,6 +2567,38 @@ impl Ui {
...
+    /// See also: [`Self::close_menu`] and [`Response::context_menu`].
+    #[inline]
+    pub fn menu_text_image_button<'a, R>(
+        &mut self,
+        title: impl Into<WidgetText>,
+        image: impl Into<Image<'a>>,
+        add_contents: impl FnOnce(&mut Ui) -> R,
+    ) -> InnerResponse<Option<R>> {
+        if let Some(menu_state) = self.menu_state.clone() {
+            menu::submenu_button(self, menu_state, String::new(), add_contents)
```
emilk, line 2597: you should pass in the title here instead of `String::new`

crates/egui/src/menu.rs
```diff
@@ -230,6 +243,24 @@ fn stationary_menu_image_impl<'c, R>(
     InnerResponse::new(inner.map(|r| r.inner), button_response)
 }
 
+/// Build a top level menu with an text and image button.
+///
+/// Responds to primary clicks.
+fn stationary_menu_text_image_impl<'c, R>(
+    ui: &mut Ui,
+    image_button: Button<'_>,
```
emilk, line 251:
```suggestion
/// Build a top level menu with an text and button.
///
/// Responds to primary clicks.
fn stationary_menu_text_image_impl<'c, R>(
    ui: &mut Ui,
    button: Button<'_>,
```

crates/egui/src/menu.rs
```diff
@@ -230,6 +243,24 @@ fn stationary_menu_image_impl<'c, R>(
     InnerResponse::new(inner.map(|r| r.inner), button_response)
 }
 
+/// Build a top level menu with an text and image button.
+///
+/// Responds to primary clicks.
+fn stationary_menu_text_image_impl<'c, R>(
```
emilk, line 249: I'm not sure we need `stationary_menu_image_impl` anymore. A normal `Button` supports both text and image, so the `ImageButton` versions of these functions shouldn't be needed anymore

crates/egui/src/ui.rs
```diff
@@ -2564,7 +2564,39 @@ impl Ui {
...
+    ///     });
+    /// });
+    /// ```
+    ///
+    /// See also: [`Self::close_menu`] and [`Response::context_menu`].
+    #[inline]
+    pub fn menu_text_image_button<'a, R>(
+        &mut self,
+        title: impl Into<WidgetText>,
+        image: impl Into<Image<'a>>,
```
emilk, line 2593: Let's put them in the same order as they will appear (image left of button)

```suggestion
    pub fn menu_image_text_button<'a, R>(
        &mut self,
        image: impl Into<Image<'a>>,
        title: impl Into<WidgetText>,
```

NicolasBircksZR:
@emilk Updated it - does this work now as you wanted?

crates/egui/src/ui.rs
```diff
@@ -2564,7 +2564,39 @@ impl Ui {
...
+    /// let title = "My Menu";
+    ///
+    /// ui.menu_text_image_button(title, img, |ui| {
+    ///     ui.menu_button("My sub-menu", |ui| {
+    ///         if ui.button("Close the menu").clicked() {
+    ///             ui.close_menu();
+    ///         }
+    ///     });
+    /// });
+    /// ```
```
emilk, line 2586: This doctest is now outdated. You can use `egui::__run_test_ui` to actually run your doctests; see for instance

https://github.com/bash/egui/blob/495d72397b55d152597736cf5deece23fb5f342d/crates/egui/src/ui.rs#L19-L30

crates/egui/src/menu.rs
```diff
@@ -105,17 +105,17 @@ pub fn menu_button<R>(
     stationary_menu_impl(ui, title, Box::new(add_contents))
 }
 
-/// Construct a top level menu with an image in a menu bar. This would be e.g. "File", "Edit" etc.
+/// Construct a top level menu with a custom button in a menu bar.
 ///
 /// Responds to primary clicks.
 ///
 /// Returns `None` if the menu is not open.
-pub fn menu_image_button<R>(
```
emilk, line 113: Let's keep the old `menu_image_button` as well, but `#[deprecated]`, and forward it to `menu_custom_button`, so we don't break users code.

You may need to make `ImageButton::image` `pub(crate)` in order to turn it into a `Button`.

emilk (changes requested):

emilk (approved):

emilk:
Thanks!

PR #4751, opened 2024-07-01
Prevent `TextEdit` widgets from sending fake primary clicks

Aliremu:
This prevents TextEdit widgets from sending fake primary clicks when a user types Space / Enter. Small change but having Space / Enter send `OutputEvent::ValueChanged` instead of `OutputEvent::Clicked` makes more sense I believe.

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -539,6 +539,9 @@ impl<'t> TextEdit<'t> {
             Sense::hover()
         };
         let mut response = ui.interact(outer_rect, id, sense);
+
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -539,6 +539,9 @@ impl<'t> TextEdit<'t> {
             Sense::hover()
         };
         let mut response = ui.interact(outer_rect, id, sense);
+
+        response.fake_primary_click = false; // remove fake primary clicks to prevent sending OutputEvent::Clicked
```
emilk, line 543:
```suggestion
        response.fake_primary_click = false; // Don't sent `OutputEvent::Clicked` when a user presses the space bar
```

emilk (approved):

PR #4770, opened 2024-07-03
Fix: ScrollArea

rustbasic:
This is needed to resolve the following explanation, which cannot be resolved by #4754.

                // Move the scrollbar so it is visible. This is needed in some cases.
                // For instance:
                // * When we have a vertical-only scroll area in a top level panel,
                //   and that panel is not wide enough for the contents.
                // * When one ScrollArea is nested inside another, and the outer
                //   is scrolled so that the scroll-bars of the inner ScrollArea (us)
                //   is outside the clip rectangle.
                // Really this should use the tighter clip_rect that ignores clip_rect_margin, but we don't store that.
                // clip_rect_margin is quite a hack. It would be nice to get rid of it.


emilk:
I feel we need to add some proper tests for `ScrollArea`, or we will just regress old bugs.

Also: please explain what a PR is supposed to fix.

PR #4775, opened 2024-07-03
Check if the mouse / touch event is above the canvas via element_from_point

### LINKED ISSUE #4752: touchend listener on document prevents listening for clicks on dom elements (web)
**Describe the bug**
Egui is registering a `touchend` event listener on the document and is calling preventdefault on the event. This prevents any touch interactions with dom elements shown on top of the canvas.

**To Reproduce**

Paste the following in the console on egui.rs (to create a simple html dialog):

```js
let dialog = document.createElement("dialog");
let button = document.createElement("button");
button.innerHTML = "Hello";
dialog.appendChild(button);
button.onclick = () => alert("button click!!");
document.body.appendChild(dialog);
dialog.show()
```
Clicking the button now does nothing. (Make sure you enabled the device toolbar and set it to e.g. iPhone SE)

If we forcefully remove the egui touchend listener from document via
```js
document.removeEventListener("touchend", getEventListeners(document).touchend[0].listener)
```
the button will become clickable.

**Expected behavior**
Egui shouldn't interfere with other dom elements. Maybe the touchend listener doesn't need preventDefault or it could be made configurable.

**Additional context**

The listener is added here: https://github.com/emilk/egui/blob/af404fec635ddd6390e2caeb3313cb9003b54835/crates/eframe/src/web/events.rs#L92

* Related issue: https://github.com/emilk/egui/issues/3157

lucasmerlin:
* Closes #4752 

To test, start the web demo and follow the repro steps from #4752

### DIFF
--- a/crates/eframe/src/web/events.rs
+++ b/crates/eframe/src/web/events.rs
@@ -1,6 +1,5 @@
-use web_sys::EventTarget;
-
 use super::*;
+use web_sys::{window, EventTarget};
 
 // TODO(emilk): there are more calls to `prevent_default` and `stop_propagaton`
 // than what is probably needed.
@@ -422,8 +421,17 @@ fn install_mousedown(runner_ref: &WebRunner, target: &EventTarget) -> Result<(),
 }
 
 /// Returns true if the cursor is above the canvas, or if we're dragging something.
-fn is_interested_in_pointer_event(egui_ctx: &egui::Context, pos: egui::Pos2) -> bool {
-    egui_ctx.input(|i| i.screen_rect().contains(pos) || i.pointer.any_down() || i.any_touches())
+/// Pass in the position in browser viewport coordinates (usually event.clientX/Y).
+fn is_interested_in_pointer_event(runner: &AppRunner, pos: egui::Pos2) -> bool {
+    window()
+        .unwrap()
+        .document()
+        .unwrap()
+        .element_from_point(pos.x, pos.y)
+        .is_some_and(|element| element.eq(runner.canvas()))
+        || runner
+            .egui_ctx()
+            .input(|i| i.pointer.any_down() || i.any_touches())
 }
 
 fn install_mousemove(runner_ref: &WebRunner, target: &EventTarget) -> Result<(), JsValue> {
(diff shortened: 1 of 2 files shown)


emilk (commented):
Have you tested this with scrolling, to make sure we're not confusing coordinate systems?

lucasmerlin:
> Have you tested this with scrolling, to make sure we're not confusing coordinate systems? (`page_x` vs `client_x`)


https://github.com/emilk/egui/assets/8009393/309e613b-30ab-4916-a566-abdfd0240994


I also checked the [clientX/Y](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/clientX) and [elementFromPoint](https://developer.mozilla.org/en-US/docs/Web/API/Document/elementFromPoint) documentation which both state that they take coordinates relative to the viewport.

crates/eframe/src/web/events.rs
```diff
@@ -422,8 +421,17 @@ fn install_mousedown(runner_ref: &WebRunner, target: &EventTarget) -> Result<(),
...
+fn is_interested_in_pointer_event(runner: &AppRunner, pos: egui::Pos2) -> bool {
+    window()
+        .unwrap()
+        .document()
+        .unwrap()
+        .element_from_point(pos.x, pos.y)
+        .is_some_and(|element| element.eq(runner.canvas()))
+        || runner
+            .egui_ctx()
+            .input(|i| i.pointer.any_down() || i.any_touches())
```
emilk, line 434: This is very difficult to parse. Naming intermediates helps

```suggestion
    let document = web_sys::window().unwrap().document().unwrap();
    let hovering_canvas = document
        .element_from_point(pos.x, pos.y)
        .is_some_and(|element| element.eq(runner.canvas()));
    let pointer_down = runner
        .egui_ctx()
        .input(|i| i.pointer.any_down() || i.any_touches())
    
    is_hovering_canvas || pointer_down
```

emilk (approved):

crates/eframe/src/web/events.rs
lucasmerlin: Much better!

Issue #4746, opened 2024-07-01
`WebRunner::start` should accept canvas object


jprochazk:
It would be nice if it was possible to initialize `eframe` on web without giving the canvas an `id` and appending it to the DOM. It should be possible if the `WebRunner::start` accepted an `HTMLCanvasElement`, and passed it along directly:

https://github.com/emilk/egui/blob/d2717180656d640e606a46932718dbef59fd5ef7/crates/eframe/src/web/web_runner.rs#L58-L63

Right now for apps using `eframe` with multiple canvases on the same page, each one needs to be given a unique ID. The painter already holds onto the canvas element directly rather than its ID, so this feels like an obvious and easy change.

emilk:
Context: https://github.com/rerun-io/rerun/pull/6680#discussion_r1659408887

PR #4780, opened 2024-07-04
Use canvas directly

jprochazk:
* Closes https://github.com/emilk/egui/issues/4746

## Migration guide

`WebRunner::start` now accepts a `web_sys::HtmlCanvasElement` instead of its ID.

If you are responsible for creating a canvas and assigning it an ID, you no longer need to assign it an ID, and can instead directly pass the canvas to `WebRunner`:

```js
let canvas = document.createElement("canvas");
body.appendChild(canvas);

let handle = new WebHandle();
handle.start(canvas);
```

If you already have a canvas in your HTML, you are no longer forced to assign it an ID. You can now use any method you prefer to retrieve it from the DOM.

```html
<canvas id="the_canvas"></canvas>
<script>
let handle = new WebHandle();
handle.start(the_canvas);
</script>
```


emilk:
I think we should get this over the finish line.

Let's not bother with backwards compatibility (i.e. handling either an id or an element).

jprochazk:
CI failure seems to be coming from main

crates/eframe/src/web/mod.rs
```diff
@@ -131,7 +131,7 @@ fn get_canvas_element_by_id(canvas_id: &str) -> Option<web_sys::HtmlCanvasElemen
 fn get_canvas_element_by_id_or_die(canvas_id: &str) -> web_sys::HtmlCanvasElement {
     get_canvas_element_by_id(canvas_id)
         .unwrap_or_else(|| panic!("Failed to find canvas with id {canvas_id:?}"))
-}
+} */
```
emilk, line 134: rm dead code

emilk (approved):

Issue #4771, opened 2024-07-03
Error compiling egui-extras


hacknus:
I tried the new version of egui (0.28) but egui-extras seems to fail compiling with the following messages:

in `synthax_highlighting.rs
`
error[E0277]: the trait bound `CodeTheme: SerializableAny` is not satisfied
   --> /Users/linus/.cargo/registry/src/index.crates.io-6f17d22bba15001f/egui_extras-0.28.0/src/syntax_highlighting.rs:158:19
    |
158 |                 d.get_persisted(egui::Id::new("dark"))
    |                   ^^^^^^^^^^^^^ the trait `serde::ser::Serialize` is not implemented for `CodeTheme`, which is required by `CodeTheme: SerializableAny`
    |
    = note: required for `CodeTheme` to implement `SerializableAny`
note: required by a bound in `IdTypeMap::get_persisted`
   --> /Users/linus/.cargo/registry/src/index.crates.io-6f17d22bba15001f/egui-0.28.0/src/util/id_type_map.rs:397:29
    |
397 |     pub fn get_persisted<T: SerializableAny>(&mut self, id: Id) -> Option<T> {
    |                             ^^^^^^^^^^^^^^^ required by this bound in `IdTypeMap::get_persisted`
`

and in `table.rs`:
`
error[E0277]: the trait bound `TableState: SerializableAny` is not satisfied
   --> /Users/linus/.cargo/registry/src/index.crates.io-6f17d22bba15001f/egui_extras-0.28.0/src/table.rs:547:45
    |
547 |             .data_mut(|d| d.get_persisted::<Self>(state_id))
    |                             -------------   ^^^^ the trait `serde::ser::Serialize` is not implemented for `TableState`, which is required by `TableState: SerializableAny`
    |                             |
    |                             required by a bound introduced by this call
    |
    = note: required for `TableState` to implement `SerializableAny`
note: required by a bound in `IdTypeMap::get_persisted`
   --> /Users/linus/.cargo/registry/src/index.crates.io-6f17d22bba15001f/egui-0.28.0/src/util/id_type_map.rs:397:29
    |
397 |     pub fn get_persisted<T: SerializableAny>(&mut self, id: Id) -> Option<T> {
    |                             ^^^^^^^^^^^^^^^ required by this bound in `IdTypeMap::get_persisted`

`

My `cargo.toml` looks like this:

`
[dependencies]
eframe = { version = "0.28", features = ["persistence"] }
egui_extras = { version = "0.28" }
[...]

PR #4786, opened 2024-07-05
Make `serde` a default (opt-out) feature of `egui_extras`

emilk:
This is a workaround for https://github.com/emilk/egui/issues/4771 to make it less likely to bite users

### DIFF
--- a/crates/egui_extras/Cargo.toml
+++ b/crates/egui_extras/Cargo.toml
@@ -27,7 +27,7 @@ all-features = true
 
 
 [features]
-default = ["dep:mime_guess2"]
+default = ["dep:mime_guess2", "serde"]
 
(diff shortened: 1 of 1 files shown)


torokati44:
It would have been nice to mention this in the `v0.28.1` changelog explicitly, as we had to take explicit steps to avoid the slight indirect dependency bloat (https://github.com/ruffle-rs/ruffle/pull/17000). Thank you for the ping on the PR that just got reverted, at least I noticed it this way, since I was following that.

emilk:
It is in the `egui_extras` changelog: https://github.com/emilk/egui/blob/master/crates/egui_extras/CHANGELOG.md#0281---2024-07-05

torokati44:
Fair enough - I was only looking at the main `egui`/`eframe` release on the GH web UI.

PR #4790, opened 2024-07-05
Remove dead code (`EpiIntegration::frame_start`)

rustbasic:
`frame_start` appears to be deprecated now.
( Am I wrong? )

It would be a good idea to add `#[allow(dead_code)]` or remove `frame_start`.


emilk:
yeah it's dead can can be removed

PR #4791, opened 2024-07-06
Improved Behavior of Nested ScrollArea

rustbasic:
**Issue:** 
When a `ScrollArea` is nested with both Horizontal and Vertical scrolling, the `outer_rect` can become very large due to the content, or when the content is moved. This can cause some issues.

**Improvement:**  
There's a part that handles this with a comment, which simplified and made to work better. 

* Related #4658
* Related #4754 
* Related #4770


emilk:
As I said in https://github.com/emilk/egui/pull/4770 and https://github.com/emilk/egui/pull/4658:

We need a proper test for this, otherwise we will just go around in circles. `ScrollArea` is getting way to complex.

The test could be a manual one (in https://github.com/emilk/egui/blob/master/tests/) for now.

rustbasic:
> As I said in #4770 and #4658:
> 
> We need a proper test for this, otherwise we will just go around in circles. `ScrollArea` is getting way to complex.
> 
> The test could be a manual one (in https://github.com/emilk/egui/blob/master/tests/) for now.

This is as simplified as possible. I tested it for a long time without any problems. Please review.

rustbasic:
completed.

PR #4794, opened 2024-07-06
Handle the IME event first in `TextEdit` to fix some bugs

rustbasic:
Handle the `IME` event first 
There is a need to handle the Ime event first.

Fix Issues : When you press `BackSpace`, the entire text is erased, and when you press another letter, it appears again.
Fix Issues : When you press `Left`, the character being entered will be copied once more.
Fix Issues : When you press `Enter`, `Enter` with `repeat` set is entered and `Enter` is entered twice.
Fix Issues : When you press a key in `IME` mode, `repeat` is often set.

Fix Issues : Since you may be selecting something in the IME, this also disables the `Arrow` keys.

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -425,7 +425,7 @@ impl<'t> TextEdit<'t> {
                         frame_rect,
                         visuals.rounding,
                         ui.visuals().extreme_bg_color,
-                        ui.visuals().widgets.noninteractive.bg_stroke, // TODO(emilk): we want to show something here, or a text-edit field doesn't "pop".
+                        visuals.bg_stroke, // TODO(emilk): we want to show something here, or a text-edit field doesn't "pop".
                     )
                 }
             } else {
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -425,7 +425,7 @@ impl<'t> TextEdit<'t> {
                         frame_rect,
                         visuals.rounding,
                         ui.visuals().extreme_bg_color,
-                        visuals.bg_stroke, // TODO(emilk): we want to show something here, or a text-edit field doesn't "pop".
+                        ui.visuals().widgets.noninteractive.bg_stroke, // TODO(emilk): we want to show something here, or a text-edit field doesn't "pop".
```
emilk, line 428: Remove unrelated style changes

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -1055,6 +1062,27 @@ fn events(
 
 // ----------------------------------------------------------------------------
 
+fn ime_enabled_filter_events(events: &mut Vec<Event>) {
```
emilk, line 1065:
```suggestion
fn remove_ime_incompatible_events(events: &mut Vec<Event>) {
```

emilk (changes requested):

emilk (approved):

PR #4803, opened 2024-07-08
``egui-wgpu``: Ensure that ``WgpuConfiguration`` is ``Send + Sync``

murl-digital:
As far as I can tell, there's no reason that this shouldn't be the case

### DIFF
--- a/crates/egui-wgpu/src/lib.rs
+++ b/crates/egui-wgpu/src/lib.rs
@@ -230,7 +230,8 @@ pub struct WgpuConfiguration {
     pub supported_backends: wgpu::Backends,
 
     /// Configuration passed on device request, given an adapter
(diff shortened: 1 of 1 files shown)


emilk (approved):
Makes sense, but please add a test too so we don't regress! You can follow this template: https://github.com/emilk/egui/blob/f9476e0a00a6622d2f01b56964bba9e97fd065ed/crates/egui/src/context.rs#L3468-L3473

murl-digital:
> Makes sense, but please add a test too so we don't regress! You can follow this template: https://github.com/emilk/egui/blob/f9476e0a00a6622d2f01b56964bba9e97fd065ed/crates/egui/src/context.rs#L3468-L3473

done!

emilk (approved):

PR #4805, opened 2024-07-08
Add Theme Switch Widget

bash:
I got a *juust* a bit sidetracked while working on #4490 because I realized that the global dark/light mode toggle would be missing a "follow system" option.

My solution is a custom switch widget with three options: follow system, dark and light.

Screencast from 2024-07-08 21-46-32

(follow system is disabled for now but is supported by the widget)

The icons are hand-drawn because:
1. I couldn't for the life of me figure out how to properly center the icons from the icon font.
1. The sun, moon and gear icons in the built-in icon font don't look tooo nice on small sizes.
1. I can smoothly animate them between different sizes
1. Drawing the icons by hand is fun? 

I tested the widget on my Mac with VoiceOver on to ensure that the widget is announced as `Theme, radio group` and each radio button is announced as `<label>, <n> out of 3`.

The widget and the accompanying `ThemePreference` type are both `pub(crate)` for now, but I intend to make them public in #4744.


bash:
I forgot to mention that I'd be happy to split this up into smaller PRs if needed :)

Some individual parts that could be extracted:
* Return value from `with_accessibility_parent`
* arc drawing could be moved to epaint and exposed as public API
* general shape rotation would remove the need for my `rotated_rect` workaround.

AurevoirXavier:
Just curious, how did you obtain the current system theme in egui?

bash:
@AurevoirXavier egui itself currently doesn't know about the system theme. It relies on `eframe` (or your integration of choice) to set the `Visuals` to match the system theme (controlled by [`follow_system_theme`]).

The way that the switch widget recognizes dark vs. light mode currently is by checking `Visuals.dark_mode`.

If you want to know the system theme and are using eframe then you can get [`system_theme`]  via `frame.info()`.

[`system_theme`]: https://docs.rs/eframe/latest/eframe/struct.IntegrationInfo.html#structfield.system_theme
[`follow_system_theme`]: https://docs.rs/eframe/latest/eframe/struct.NativeOptions.html#structfield.follow_system_theme

emilk:
This is a _very_ pretty theme selector, but I worry it is a bit overkill for the default one in egui for several reasons:

A) it adds a lot of code to maintain
B) it adds a lot of code => increase .wasm size
C) the style of the buttons don't fit in with the rest of egui's default theme (they are too pretty)

I suggest you create this as a 3rd party crate instead, and for egui make a much simpler drop-down menu.

bash:
I'm closing this as I have published a crate for this and extracted the "salvageable" parts into two separate PRs.

Issue #1713, opened 2022-06-03
Create an Undoer Example in the egui Demo App


zicklag:
**Is your feature request related to a problem? Please describe.**
I didn't know about the awesome undoer utility, and almost wrote my own before stumbling upon it by chance while looking for something else in the egui docs.

**Describe the solution you'd like**
We could have a new demo window in the egui Demo App that just has a text box, and maybe an extra text-box or radio button, that allows you to undo by clicking a button or using a keyboard shortcut.

**Describe alternatives you've considered**
Somehow making the utility more visible in the API docs might be a good idea in addition or alternatively to the demo window.

**Additional context**
I often use the egui demo app as kind of an examples index / "feature list" so I think having the undoer featured in a demo would help increase its visibility, and I've found the feature very useful for my app. It made implementing undo painless, when it could have been quite some trouble.

emilk:
Good idea! Feel free to contribute a PR

zicklag:
OK, cool, I'll do that. :+1:

PR #4811, opened 2024-07-10
Add undo/redo demo.

Wcubed:
* Closes https://github.com/emilk/egui/issues/1713 

I almost went to implement my own undo/redo system, and then found the egui undoer.
Went to make a small demo to test for myself how it worked, and then found the linked issue.
So here is a tweaked version of that :)


emilk (approved):
Thanks!

PR #4814, opened 2024-07-11
eframe: Improve `glow` context switching

rustbasic:
Improved `change_gl_context()`

**Before:** 
Create a new `not_current_glcontext`. 

**After:**
If `not_current_glcontext` exists, apply it. Otherwise, create it. 

This will make the program smoother when dragging, etc.


emilk (commented):
Looks good! Have you tested this with multi-viewports?

rustbasic:
> Looks good! Have you tested this with multi-viewports?

yes. It does not affect the current content.

emilk (approved):

PR #4815, opened 2024-07-11
epaint: add some missing inline docs

WojciechMula:
Document a few `Mesh` methods.


emilk (approved):
Nice, especially documenting the panics 👍

PR #4817, opened 2024-07-11
Change default `max_scroll_height` of `egui::Table` to `f32::INFINITY`

abey79:
`egui::Table` has a weirdly specific default `max_scroll_height` set to 800.0. This means that even with `vscroll(true)` and `auto_shrink([_, false])`, the table will not expend to the full available height. This PR changes this value to `f32::INFINITY`. This makes it consistent with the corresponding default value of `ScrollArea::max_size.y`, which is where that `max_scroll_height` ends up being used.


emilk (approved):
Makes sense to match `ScrollArea`

PR #4819, opened 2024-07-12
IME visibility improvement and visible selection

rustbasic:
**IME visibility improvement and visible selection**

I have made the IME indicator visible,
and I have improved it to the best of our ability so far.
However, we still need to continue improving it in the future.

Fix Issues: When `Multiline`, IME Rect is displayed far away.
Fix Issues: The IME is displayed once and after it is entered, the IME is not displayed.
Fix Issues: IME is not updated at the end of line when using `Singleline`.

* Related #4254
* Related #4269
* Related #4362 

**Remaining Issue :**
In `Deferred Viewport`, no 'input event' comes in `handle_platform_output()`. 
In `Immediated Viewport`, the `ROOT`'s 'input event' comes in `handle_platform_output()`.

Therefore, this is the best solution for now.


crates/egui/src/style.rs
```diff
@@ -751,6 +751,9 @@ pub struct TextCursorStyle {
 
     /// When blinking, this is how long the cursor is invisible.
     pub off_duration: f32,
+
+    /// Indicates whether the IME is visible.
+    pub ime_visible: bool,
```
emilk, line 756: Why would anyone want to turn this off?

crates/egui-winit/src/lib.rs
```diff
@@ -836,26 +836,25 @@ impl State {
...
-            let pixels_per_point = pixels_per_point(&self.egui_ctx, window);
-            let ime_rect_px = pixels_per_point * ime.rect;
-            if self.ime_rect_px != Some(ime_rect_px)
-                || self.egui_ctx.input(|i| !i.events.is_empty())
-            {
-                self.ime_rect_px = Some(ime_rect_px);
+            let ime_rect = if ime.visible {
+                ime.cursor_rect
+            } else {
+                self.egui_ctx.screen_rect()
```
emilk, line 842: It's still visible though, right? Just applied to the whole screen instead of just the text cursor?

crates/egui/src/style.rs
rustbasic: You can enter text even if the IME is not displayed.
When an IME is displayed, it often doesn't look good.

crates/egui-winit/src/lib.rs
rustbasic: IME is not displayed within the specified area. In other words, the IME is not displayed within the program and is not visible.
(I'm not sure if this is only on Windows, but my current method is the best.)

In the end, we can control it according to own needs by selecting visibility, etc. depending on the situation.

Issue #3453, opened 2023-10-11
SVG images are blurry when `ctx.pixel_per_point` is not equal to 1.0


mkrueger:
New API:
<img width="133" alt="grafik" src="https://github.com/emilk/egui/assets/341098/89ff6b18-a320-4076-819c-f854d44df9fb">

Old API :
<img width="84" alt="grafik" src="https://github.com/emilk/egui/assets/341098/f49d06fe-e0a5-49f0-9ff4-5c95b9d78e1f">

It's a 4k issue - somewhere internal, the pixels per point don't get valued.

My usage is:
` ui.add(ImageButton::new(icon.clone().fit_to_exact_size(size_points).tint(tint)))`

the icon is loaded with:
` ` `
pub static ref DOCK_LEFT_SVG: Image<'static> = Image::new(egui::include_image!("../../data/icons/dock_left.svg"));
` ` ` 

The file is:
` ` `
<svg width="56" height="56" fill="#ffffff" xmlns="http://www.w3.org/2000/svg" viewBox="0 -960 960 960"><path d="M180-120q-24.75 0-42.375-17.625T120-180v-600q0-24.75 17.625-42.375T180-840h600q24.75 0 42.375 17.625T840-780v600q0 24.75-17.625 42.375T780-120H180Zm207-60h393v-600H387v600Z"/></svg>
` ` ` 

I had the issue with the old API as well - the solution was to double the width/height in the SVG and it looked sharp.
fit_to_exact_size needs to value the pixels_per_point setting - or did I do something wrong on my side?. When I double the size_points the image becomes larger - not sharper btw.

AurevoirXavier:
Still got this in the latest version.

passbyval:
Switching to `TextureOptions::NEAREST` has improved the image sharpness for me:
```rs
ImageButton::new(
    egui::Image::new(img_src)
        .max_size(size)
        .maintain_aspect_ratio(true)
        .texture_options(TextureOptions::NEAREST)
        .rounding(4.0),
),
```

emilk:
We need to take `ctx.pixels_per_point()` into account somewhere in https://github.com/emilk/egui/blob/8bf498d2bf849892ec0c3fa49fea4c3c504a06fe/crates/egui_extras/src/loaders/svg_loader.rs#L53-L55 and/or https://github.com/emilk/egui/blob/8bf498d2bf849892ec0c3fa49fea4c3c504a06fe/crates/egui_extras/src/image.rs#L226-L237

[...]

PR #4823, opened 2024-07-13
Make sure SVGs are crisp

AurevoirXavier:
* Closes https://github.com/emilk/egui/issues/3453

I'm fairly new to egui.

I read the code, but I didn't follow the approach mentioned in https://github.com/emilk/egui/issues/3453#issuecomment-2208255433.

I believe this is an easier way to achieve that, though I'm not certain if it's the best method.

image

I get really nice svg with this patch.

@emilk Can you please take a look? I really need this!

### DIFF
--- a/crates/egui/src/load.rs
+++ b/crates/egui/src/load.rs
@@ -123,6 +123,7 @@ pub type Result<T, E = LoadError> = std::result::Result<T, E>;
 /// Given as a hint for image loading requests.
 ///
 /// Used mostly for rendering SVG:s to a good size.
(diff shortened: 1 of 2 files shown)


AurevoirXavier:
It's interesting that some of the indents are tabs while others are 4-spaces.

Would you like me to fix that?

emilk (approved):
Looks good!

But the `(false, false) => ` branch in `fn hint` should also fall back to `Scale(pixels_per_point)`.

It would also be good to add a comment to `SizeHint` that its units are in image texels, and that `pixels_per_point` has already been taken into account.

PR #4832, opened 2024-07-16
Fix: The viewport stops working when the program is minimized.                        

### LINKED ISSUE #3972: When the program is minimized, the viewport stops working.

### LINKED ISSUE #4772: On Linux, the window doesn't close unless it's unminimized
**Describe the bug**

On Linux, if you close an eframe window when it's minimized, nothing will happen until the window is un-minimized, at which point the window will close:

https://github.com/emilk/egui/assets/79560998/6d77de84-d678-459e-953f-e117ada21f51

This appears to happen on both the glow and wgpu backends.

**To Reproduce**
1. Run any eframe demo app on Linux
2. Try to close the window while it's minimized, either using the "Close" option in the taskbar's context menu, or Ctrl+C
3. Observe that the window is not closed
4. Un-minimize the window
5. Observe that it closes itself

**Expected behavior**

The window closes immediately when it's told to, instead of when it's visible

 - OS: Fedora 40 (X11)
 - Version: v0.28.0

rustbasic:
Fix: The viewport stops working when the program is minimized.                        
                                                                                      
**Issue :**                                                                         
                                                                                      
The viewport stops working when the program is minimized.                             
                                                                                      
* Closes #3972                                                                        
* Closes #4772
                                                                                      
**Solution :**                                                                      
                                                                                      
When `request_redraw()` is performed in Minimized state, the occasional screen tearing phenomenon has disappeared.
( Probably expected to be the effect of #4814 )                                       
                                                                                      
To address the issue of the `Immediate Viewport` not updating in Minimized state, we can call `request_redraw()`.

### DIFF
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -189,13 +189,8 @@ fn run_and_return(event_loop: &mut EventLoop<UserEvent>, mut winit_app: impl Win
 
             if let Some(window) = winit_app.window(*window_id) {
                 log::trace!("request_redraw for {window_id:?}");
-                let is_minimized = window.is_minimized().unwrap_or(false);
-                if is_minimized {
-                    false
-                } else {
-                    window.request_redraw();
-                    true
-                }
+                window.request_redraw();
+                true
             } else {
                 log::trace!("No window found for {window_id:?}");
                 false
@@ -347,13 +342,8 @@ fn run_and_exit(
 
             if let Some(window) = winit_app.window(*window_id) {
                 log::trace!("request_redraw for {window_id:?}");
-                let is_minimized = window.is_minimized().unwrap_or(false);
-                if is_minimized {
-                    false
-                } else {
-                    window.request_redraw();
-                    true
-                }
+                window.request_redraw();
+                true
             } else {
                 log::trace!("No window found for {window_id:?}");
                 false


crates/eframe/src/native/run.rs
```diff
@@ -189,13 +189,8 @@ fn run_and_return(event_loop: &mut EventLoop<UserEvent>, mut winit_app: impl Win
...
                 log::trace!("request_redraw for {window_id:?}");
-                let is_minimized = window.is_minimized().unwrap_or(false);
-                if is_minimized {
-                    false
-                } else {
-                    window.request_redraw();
-                    true
-                }
+                window.request_redraw();
+                true
```
emilk, line 193: This reverts the changes you added in https://github.com/emilk/egui/pull/3985

rustbasic: ne. gyesog `request_repaint()` leul hochulhaji anhneundamyeon gwaenchanhseubnida.
44 / 5,000
Yes, it's ok as long as you don't keep calling `request_repaint()`.

rustbasic: If someone want to keep calling `request_repaint()` but not call it when `minimized`, someone can do the following.
It works fine.

```
            let is_maximized = ui.input(|i| i.viewport().maximized.unwrap_or(false));
            if !is_minimized {
                ui.ctx().request_repaint();
            }
```

PR #4836, opened 2024-07-17
Introduce `ArcPieShape` for Simplified Arc and Pie Rendering

varphone:
## Summary

This PR introduces several new features related to drawing arcs and pies in the `egui` ecosystem.

## Changes

- `epaint`: Introduce `ArcPieShape` for Simplified Arc and Pie Rendering.   
- `egui`: Enhance `Painter` with `arc` and `pie` Methods.
- `egui_demo_lib`: Added demonstrations for `ArcPieShape`,  These demos showcase how to use these new features in a variety of contexts.

## Screenshot

egui_demo_app 2024-07-17 165537


David-OConnor:
Arcs would be great!

emilk:
This could be implemented with a `PathShape::circle_arc` constructor instead, which would require a lot less code (which means less code to maintain).

What motivates this addition? Are circular arcs really that common in GUIs?

David-OConnor:
Check the circle screenshot here: https://github.com/David-OConnor/plascad

I rolled one based on varphone's work. Good tip on circle_arc! I'll adapt it to use that.

Note that I ended up doing some hacks to get the concave fill to work; the actual line arc is straightforward, but the fill is kind of awk. (Breaking it up into small segments that don't glitch, drawing a circle with the background color with radius of the arc's inner radius etc.

David-OConnor:
@emilk I concur with your edit. The big/general fix is for filling concave shapes; that would be a bigger win than a specialty arc!

emilk:
The current tessellator for filled polygons is doing a [triangle fan](https://docs.safe.com/fme/html/FME-Form-Documentation/FME-ReadersWriters/!FME_Geometry/IFMETriangleFan.htm) from the first vertex:

https://github.com/micmonay/egui/blob/fdc6658517c794feb194f78b7ecab0d75f3f259b/crates/epaint/src/tessellator.rs#L732-L787

This means that you could probably use `PathShape` if the first vertex is the center of the filled pie. Of course, that would rely on an implementation detail, but we could codify that if it works.

David-OConnor:
Makes sense! The issues arise  when there's no pie center! (eg a filled arc segment)

Skgland:
> What motivates this addition? Are circular arcs really that common in GUIs?

Depends on the use case, but I think gauges, circular progress bars, and (multi layer) pie/donut charts are common enough.

E.g.

PR #4838, opened 2024-07-17
Apply `TextWrapMode::Extend` in `TextEdit`

### LINKED ISSUE #4837: Text goes to newline without pressing enter
**Describe the bug**
When using Multiline widget, the text goes to next line as soon as the current line fills up, it shouldn't happen until unless user press enter, it must render the text in same line as it is default behaviour in text editors, browsers, etc.

**To Reproduce**
```rust
use eframe::egui;

fn main() -> eframe::Result {
        ..Default::default()
    };
    eframe::run_native(
        "My egui App",
        options,
    )
}

#[derive(Default)]
struct MyApp {
    text: String,
}

impl eframe::App for MyApp {
            egui::ScrollArea::both()
                .max_width(f32::INFINITY)
                .show(ui, |ui| {
                    ui.add(
                        egui::TextEdit::multiline(&mut self.text)
                            .desired_width(f32::INFINITY)
                            .desired_rows(12)
                            .lock_focus(true)
                            .code_editor(),
                    );
                });
        });
    }
}
```

**Expected behavior**
As soon as the text overflows the current widget width, it should continue in the same line by expanding and create a scrollbar.

**Screenshots**
image

 - OS: Windows 11
 
**Additional context**
I tested it using `glow` OpenGL as renderer

rustbasic:
Apply `TextWrapMode::Extend` in `TextEdit`

* Closes #4837

I'm not sure if this is the correct approach.


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -481,6 +481,14 @@ impl<'t> TextEdit<'t> {
...
         let available_width = (ui.available_width() - margin.sum().x).at_least(MIN_WIDTH);
         let desired_width = desired_width.unwrap_or_else(|| ui.spacing().text_edit_width);
+
+        let wrap_mode = if ui.style().wrap_mode.is_some() {
+            ui.style().wrap_mode
+        } else if multiline {
+            Some(TextWrapMode::Wrap)
+        } else {
+            Some(TextWrapMode::Extend)
+        };
```
emilk, line 491:
```suggestion
        let wrap_mode = ui.style().wrap_mode.unwrap_or(if multiline {
            TextWrapMode::Wrap
        } else {
            TextWrapMode::Extend
        });
```

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -491,6 +497,11 @@ impl<'t> TextEdit<'t> {
         let mut default_layouter = move |ui: &Ui, text: &str, wrap_width: f32| {
             let text = mask_if_password(password, text);
             let layout_job = if multiline {
+                let wrap_width = if wrap_mode == TextWrapMode::Extend {
+                    f32::INFINITY
+                } else {
+                    wrap_width
+                };
```
emilk, line 504: This logic is now only run for the `default_layouter`, while custom layouters gets the previous `wrap_width` which ignores `wrap_mode`.

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -670,17 +678,12 @@ impl<'t> TextEdit<'t> {
...
                         hint_text_font_id,
                     )
                 } else {
-                    hint_text.into_galley(
-                        ui,
-                        Some(TextWrapMode::Extend),
-                        f32::INFINITY,
-                        hint_text_font_id,
-                    )
+                    hint_text.into_galley(ui, Some(wrap_mode), f32::INFINITY, hint_text_font_id)
```
emilk, line 686: I think single-line text-edits should always use `Extend`, otherwise they are not single-line

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -481,7 +481,15 @@ impl<'t> TextEdit<'t> {
         const MIN_WIDTH: f32 = 24.0; // Never make a [`TextEdit`] more narrow than this.
         let available_width = (ui.available_width() - margin.sum().x).at_least(MIN_WIDTH);
         let desired_width = desired_width.unwrap_or_else(|| ui.spacing().text_edit_width);
-        let wrap_width = if ui.layout().horizontal_justify() {
+
+        let wrap_mode = ui.style().wrap_mode.unwrap_or(if multiline {
```
emilk, line 485: Should we really use `TextWrapMode::Truncate` if set? I don't think that will work very well

rustbasic: > TextWrapMode::Truncate

I made some changes, but I'm not sure if the changes will work everywhere.

I haven't reviewed the `TextWrapMode::Truncate` issue.
I'm not sure how `TextWrapMode::Truncate` is supposed to work with `singleline()` and `multiline()`, and what it's intended to do.
How about handling the `TextWrapMode::Truncate` issue in a new commit?

You can drop this and create a new one.

PR #4840, opened 2024-07-17
Avoid some `Id` clashes by seeding auto-ids with child id

ironpeak:
I was having trouble with id collisions and was not able to resolve it using `push_id` and `child_ui_with_id_source`.

When investigating the issue I found https://github.com/emilk/egui/pull/2262 which matched the issues I had so I forked egui and implemented the changes from that PR for the latest version.

It solved the issue for me.

I did not notice any regressions in my project or the egui web viewer.


emilk (approved):
Let's try it!

emilk (approved):

PR #4846, opened 2024-07-18
Clarified `eframe::run_simple_native()` persistence

tpstevens:

Clarified that `eframe::run_simple_native()` does in fact save some state (if persistence is enabled) but does not allow user to save state.


emilk (approved):
Excellent!

PR #4847, opened 2024-07-18
upgrade to wgpu 22.0.0

KeKsBoTer:
wgpu changelog: https://github.com/gfx-rs/wgpu/releases/tag/v22.0.0

### DIFF
--- a/crates/egui-wgpu/src/renderer.rs
+++ b/crates/egui-wgpu/src/renderer.rs
@@ -345,7 +345,7 @@ impl Renderer {
                     compilation_options: wgpu::PipelineCompilationOptions::default()
                 }),
                 multiview: None,
-                cache:None,
+                cache: None,
             }
         )
         };
--- a/crates/egui-wgpu/src/renderer.rs
+++ b/crates/egui-wgpu/src/renderer.rs
@@ -345,6 +345,7 @@ impl Renderer {
                     compilation_options: wgpu::PipelineCompilationOptions::default()
                 }),
                 multiview: None,
+                cache:None,
             }
         )
         };
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1799,9 +1799,9 @@ dependencies = [
 
 [[package]]
 name = "glutin"
-version = "0.31.2"
+version = "0.31.3"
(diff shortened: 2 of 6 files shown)


crates/egui-wgpu/src/renderer.rs
```diff
@@ -345,6 +345,7 @@ impl Renderer {
                     compilation_options: wgpu::PipelineCompilationOptions::default()
                 }),
                 multiview: None,
+                cache:None,
```
torokati44, line 348: Huh, how come the format checker doesn't complain about the lack of space here...? 🤔

Wumpf: strange indeed, I poked it to format that file directly and it didn't care. hum.

Wumpf:
Now that lifetimes on passes are more relaxed there's probably room for some follow-ups we can do to make the render callback system easier to use. Needs looking into & experimentation!
(meant to do that before wgpu 22 goes out as an exploration of wgpu's changes there, but things rarely go as planned :))

Wumpf (approved):

John-Nagle:
It looks like wgpu-egui was changed without changing its version:

https://github.com/emilk/egui/commit/56df31ab489312014d57247db438ecd7189a710a

So that can't be pushed to crates.io, where the version must change if the content does.
So [egui-wgpu at crates.io](https://crates.io/crates/egui-wgpu) 0.28.1, last updated 14 days ago, is out of sync with Github master. I think.

Wumpf:
nothing unusual there: egui & eframe also has changed since 0.28.1. There's a few repos that immediately change the version after release because _technically_ the first change after a release if for an unreleased version, but this repo is not one of them.
There's going to be a 0.29.0 at some point published to crates.io containing that change, until then applications that want it will have to point to a commit hash instead

John-Nagle:
OK. So there's a workaround on the egui side.

I'm currently stuck due to wgpu-profiler problems. [That can't even be built against current WGPU without another WGPU fix.](https://github.com/Wumpf/wgpu-profiler/pull/75) Looks like a WGPU 22.1.x will be needed.

Issue #4569, opened 2024-05-29
On iOS safari, clicking a text field won't always bring up the on-screen keyboard


emilk:
Go to egui.rs and open the "Text Edit" demo. Click the text input box, and the box will flash white, but nothing else happens. If you keep clicking it many times, eventually you will get a text cursor _but still no on-screen keyboard_. Only on the _NEXT_ click will the on-screen-keyboard show up.

emilk:
I'm not sure how to trigger the mobile keyboard on web. Maybe we need an invisible `<input>` element covering every `TextEdit` in the eframe app, so that when the user taps it, it gets focus. At the moment we only have one such `<input>` element (`TextAgent`), and it is just one pixel big.

emilk:
Closed by https://github.com/emilk/egui/pull/4855

PR #4848, opened 2024-07-18
Force canvas/text input focus on touch for iOS web browsers

BKSalman:
Related to #4569

- moved `is_mobile` from `eframe/src/web/text_agent.rs` to `eframe/src/web/mod.rs` and used it in `eframe/src/web/events.rs`
- added event listener for `pointerdown` event
- added event listener for `pointerup` event and forced the canvas to be focused when the event is received and therefore the on-screen keyboard appearing

This PR tries to fix iOS browsers issue where it doesn't focus the canvas/text input on touch and therefore not showing the on-screen keyboard. It does it by force focusing the canvas on `pointerup`

### DIFF
--- a/crates/eframe/src/web/events.rs
+++ b/crates/eframe/src/web/events.rs
@@ -77,11 +77,11 @@ pub(crate) fn install_event_handlers(runner_ref: &WebRunner) -> Result<(), JsVal
     // so we check if we have focus inside of the handler.
     install_copy_cut_paste(runner_ref, &document)?;
 
-    install_mousedown(runner_ref, &canvas)?;
     // Use `document` here to notice if the user releases a drag outside of the canvas:
     // See https://github.com/emilk/egui/issues/3157
     install_mousemove(runner_ref, &document)?;
-    install_mouseup(runner_ref, &document)?;
+    install_pointerup(runner_ref, &document)?;
+    install_pointerdown(runner_ref, &canvas)?;
     install_mouseleave(runner_ref, &canvas)?;
 
     install_touchstart(runner_ref, &canvas)?;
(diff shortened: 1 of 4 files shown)


ejjonny:
This is a big step forward for egui mobile web, ty

crates/eframe/src/web/events.rs
```diff
@@ -77,11 +77,16 @@ pub(crate) fn install_event_handlers(runner_ref: &WebRunner) -> Result<(), JsVal
...
     // See https://github.com/emilk/egui/issues/3157
     install_mousemove(runner_ref, &document)?;
-    install_mouseup(runner_ref, &document)?;
+    if is_mobile() {
+        install_pointerup(runner_ref, &canvas)?;
+        install_pointerdown(runner_ref, &canvas)?;
+    } else {
+        install_mouseup(runner_ref, &document)?;
+        install_mousedown(runner_ref, &canvas)?;
+    }
```
emilk, line 89: Please add a comment here explaining why we do different things on mobile and web. In particular, why doesn't `pointerup/down` work on desktop?

emilk (approved):
I tested it on iOS and it definitely helps https://github.com/emilk/egui/issues/4569, but does not completely fix it. Still a step in the right direction ⭐

crates/eframe/src/web/events.rs
BKSalman: I did this because I'm not sure how to properly test it and make sure it works properly for other things as well, but from my naive testing, it looks like it works just fine

jprochazk: If you change the `install_pointerup` to target `document`, does it still work?

BKSalman: I already changed it locally, yes it works, will push it in a bit

jprochazk: Okay, in that case I think we can remove the whole branch altogether, and leave only `install_pointer*`

BKSalman: ok, review latest commit please (I force pushed)

jprochazk (approved):

Issue #4709, opened 2024-06-26
eframe: Switch to winit `ApplicationManager` api


expenses:
The next version of winit will remove the `run` APIs: https://github.com/rust-windowing/winit/pull/3683. This pretty extensively breaks code like https://github.com/emilk/egui/blob/93d458bb68a5c6e1fd05cbebfd64701463e7ab61/crates/eframe/src/native/run.rs#L73-L208. I'm not sure how much effort it will be to fix, but it might make sense to do that sooner-rather-than-later.

emilk:
* Blocked by https://github.com/emilk/egui/issues/4437

Issue #4437, opened 2024-04-30
Update to `winit` 0.30


emilk:
Also needed:
* https://github.com/emilk/egui/issues/4219

lylythechosenone:
Updating to wgpu 0.20.0 should be an important goal here as well. Not sure if that should be a separate issue.

torokati44:
That's already done: https://github.com/emilk/egui/pull/4433

AndriBaal:
I updated winit to 0.30.0 and was able to successfully run the hello-world example successfully with wgpu.

Here is a summary of the changes I made:
- Bump accesskit to version 0.14.0
- Bump winit to 0.30.0
- Partially done the glow integration
- Roughly implemented accesskit
- Implemented winit / wgpu

Here is a summary of things that need to be done:
- Integration with glow. We have to wait until glutin-winit updates to winit 0.30.0
- Not all `accesskit_winit::Event` are being processed, temporarily I have marked them with todo! (found in winit_integration.rs)
- Move from deprecated `winit::EventLoop::run_on_demand` to new Application trait
- Documentation and testing
- Code review

I have madew a pull request with my changes here: #4466

emilk:
We have two competing PRs, neither of which are ready to be merged:
* https://github.com/emilk/egui/pull/4466
* https://github.com/emilk/egui/pull/4702

PR #4702, opened 2024-06-24
update winit to 0.30.3

j-axa:
I liked the new way winit 0.30 handles the eventloop/window creation.

I updated the cargo.toml of relevant crates and fixed the errors that popped up.

It works for my usecase and the demo app seems to work but I can't verify further than that.

PR #4849, opened 2024-07-18
Upgrade winit to 0.30.2

ArthurBrussee:
* Closes https://github.com/emilk/egui/issues/1918
* Closes https://github.com/emilk/egui/issues/4437
* Closes https://github.com/emilk/egui/issues/4709

Hiya,

I need new winit for a specific fix for a android_native_actvity. There are already two PRs, but both don't seem to have a lot of movement, or are entirely complete:

https://github.com/emilk/egui/pull/4466
Seems to have gone stale & is missing some bits.

https://github.com/emilk/egui/pull/4702
Also seems stale (if less so), and is missing a refactor to run_on_demand. I also *think* the accesskit integration has a mistake and can't be enabled. I've marked them as a co-author on this as I started from this branch. (I think! Haven't done that on git before...).

Sorry for the wall of text but just dumping some details / thoughts here:

- There's an issue with creating child windows in winit 0.30.1 and up on macOS. The multiple_viewports, "create immediate viewport" example crashes on anything later 0.30.1, with a stack overflow in unsafe code. I've create [a winit issue](https://github.com/rust-windowing/winit/issues/3800), it *might* already be fixed in 0.31.0 but I can't test as 0.31 will likely require another refactoring. For now I have just pinned things to 0.30.0 exatly.

- Winit has deprecated run_on_demand, instead requiring the ApplicationHandler interface. In 0.31.0 run_on_demand is removed. I've refactored both the integration and the WinitApp trait to follow this pattern. I've left user_events a bit more opaque, as it seems 0.31.0 is doing a rework of UserEvents too.

- I've used the new lazy init approach for access kit from this branch https://github.com/mwcampbell/egui/tree/accesskit-new-lazy-init and marked Matt as co-author, thanks Matt!

- There was very similair but not quite the same code for run_and_return and run_and_exit. I've merged them, but looking at the github issues graveyard it seems vey finnicky. I *hope* this is more robust than before but it's a bit scary.

- when receiving new_events this also used to check the redraw timing dictionary. That doesn't seem necesarry so left this out, but that is a slight behaviour change?

- I have reeneabled serial_windows on macOS. I wondered whether it was fixed after this PR and does seem to be! However, even before this PR it seems to work, so maybe winit has sorted things out before that... Windows also works fine now without the extra hack.

- I've done a very basic test of AccessKit on Windows and screen reader seems ok but I'm really not knowleadgable enough to say whether it's all good or not.

- I've tested cargo tests & all examples on Windows & macOS, and ran a basic Android app. Still, testing native platforms is wel... hard so if anyone can test linux / iOs / older mac versions / windows 10 would probably be a good idea!

- For consistencys sake I've made all event like functions in WinitApp return a `Result<EventResult>`. There's quite a bit of Ok-wrapping now, maybe too annoying? Not sure.

Thank you for having a look!

# Tested on
* [x] macOS
* [x] Windows
* [x] Wayland (thanks [SiebenCorgie](https://github.com/SiebenCorgie))
* [x] X11 (thanks [crumblingstatue](https://github.com/crumblingstatue)!,  [SiebenCorgie](https://github.com/SiebenCorgie))

# TODO
* [x] Fix "follow system theme" not working on initial startup (winit issue, pinning to 0.30.2 for now).
* [x] Fix `request_repaint_after`


Cargo.toml
```diff
@@ -93,8 +93,9 @@ wgpu = { version = "22.0.0", default-features = false, features = [
     # Make the renderer `Sync` even on wasm32, because it makes the code simpler:
     "fragile-send-sync-non-atomic-wasm",
 ] }
-winit = { version = "0.29.4", default-features = false }
 
+# Version 0.30.1 and up unfortunately crash on macOS, due to https://github.com/rust-windowing/winit/pull/3684
```
emilk, line 97: Should link to https://github.com/rust-windowing/winit/issues/3800 too/instead

Cargo.toml
```diff
@@ -93,8 +93,9 @@ wgpu = { version = "22.0.0", default-features = false, features = [
...
+winit = { version = "=0.30.0", default-features = false }
```
emilk, line 98: Unfortunately, by pinning to 0.30.0 we have the problem that https://github.com/rust-windowing/winit/pull/3684 solves, which is crashes during shutdown.

Running this branch gives me this one exit:
image

examples/serial_windows/src/main.rs
```diff
@@ -59,10 +52,6 @@ impl eframe::App for MyApp {
             };
             ui.label(label_text);
 
-            if ctx.os() == egui::os::OperatingSystem::Mac {
-                ui.label("This example doesn't work on Mac!");
-            }
-
```
emilk, line 65: Finally fixed!

crates/egui-winit/Cargo.toml
```diff
@@ -62,14 +62,18 @@ egui = { workspace = true, default-features = false, features = ["log"] }
 
 ahash.workspace = true
 log.workspace = true
+
+# wayland-cursor shoud enable the fs feature as it otherwise doesn't compile.
+# Instead do this manually for now.
```
emilk, line 67: Is there an relevant issue on `wayland-cursor` we can link to?

ArthurBrussee: Tried to understand this a bit better - it's an issue where wayland-cursor relied on wayland-backend to enable nix/fs. Somehow winit uses an older wayland-backend than wayland-cursor, leading to this problem.

This is moot in the future after https://github.com/Smithay/wayland-rs/commit/edd0f60d0baf09604553525c2636df5d6ba05d44, as rustix/shm enables rustix/fs as well.

Have a made a note of that here, quite a mess!

Cargo.toml
ArthurBrussee: Ah crap :/ That weirdly doesn't repro for me - either by closing windows or ctrl-c'ing but well, it's definitely missing the bugfix so makes sense....

I tried some more workarounds for 0.30.4 but really can't find anything that prevents the crash. I guess that leaves a few options:

- Go with 0.30.0 and crash on exit on mac
- Go with 0.30.4 and crash on immediate viewports (perhaps with an error)
- Rejig immediate viewports to be more like deferred somehow
- Wait for a patched winit release / workaround to be available

Lmk your thoughts!

emilk: >Go with 0.30.0 and crash on exit on mac

Having eframe always crash on exit on Mac is not an option imho.

> Go with 0.30.4 and crash on immediate viewports (perhaps with an error)

This is better, but not great. Users _can_ switch to using deferred viewports if they work. We could guide users towards it gently (with a log message) or harshly (with `#[cfg(not(target_os = "macos"))]` on `show_viewport_immediate`)

> Rejig immediate viewports to be more like deferred somehow

If we can circumvent the bug somehow that would be great, but seems unlikely?

> Wait for a patched winit release / workaround to be available

Waiting won't give us much unless someone volunteers to work on https://github.com/rust-windowing/winit/issues/3800

ArthurBrussee: Thank you for that! All good points.

Good news though :) I gave it one last go as you're right someone has to volunteer, and noticed this time that the crash complains about a reference which just so happens to come from an unsafe block not too far up in the call stack from egui....

Integrations register a callback for immediate viewports. When registering this gets an &ActiveEventLoop, then uses unsafe to cast away the lifetime, and store that pointer in the callbak. That's really not allowed! At any point there might be an aliased mutable reference, and the comment about how the lifetime outlives the callback doesnt really make sense to me - winit is free to do what it wants! The value could even just be moved in the meantime? This probably worked by happenstance on older winit versions.

So, instead I now make a tiny thread_local context which stores the pointer for the duration of the painting. I'm not very happy with it, this still leaves an unsafe block, but still seems much easier to reason about. See the safety comment for some justification, though I've never PR'ed unsafe code before so beware :) 

To really get rid of unsafe require a bigger rework that would get rid of this depencency injection shenannigans entirely I think.

emilk: Nicely done!

emilk:
I found a couple of issues during testing; added them under the `TODO` heading in the PR description. I'll do some more testing and code review soon™️. I'd like to get this in as soon as possible!

ArthurBrussee:
Wooh awesome! Thank you for doing some testing, good catch on request_repaint_after! I missed a pretty crucial check for redraw in the 'new_events' callback when porting.

I'm not sure about follow system theme. On my macbook anyway the window seems to change theme when I change the system theme. Is this broken on windows/linux? Or, pre-Sonoma? Or broken in a different way?

emilk:
> I'm not sure about follow system theme. On my macbook anyway the window seems to change theme when I change the system theme. Is this broken on windows/linux? Or, pre-Sonoma? Or broken in a different way?

Changes to the system theme are reported correctaly to egui, but not the initial system theme. If I set my system theme to bright and start `egui_demo_app`, the initial theme is dark (the default).

crates/eframe/src/native/glow_integration.rs
```diff
@@ -723,13 +705,12 @@ impl GlowWinitRunning {
...
-                    .as_ref()
-                    .expect("failed to get current context to swap buffers"),
-            ) {
-                log::error!("swap_buffers failed: {err}");
-            }
+            let context = current_gl_context.as_ref().ok_or(std::convert::Into::<
+                egui_glow::PainterError,
+            >::into(
+                "failed to get current context to swap buffers".to_owned(),
+            ))?;
```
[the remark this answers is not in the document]
ArthurBrussee: Oh yes nice!

crates/eframe/src/native/run.rs
```diff
@@ -59,324 +52,268 @@ fn with_event_loop<R>(
...
-                } else {
-                    log::trace!("Got outdated UserEvent::RequestRepaint");
-                    EventResult::Wait // old request - we've already repainted
+                    EventResult::RepaintNow(window_id) => {
+                        log::trace!("RepaintNow of {window_id:?}",);
+                        if cfg!(target_os = "windows") {
+                            // Fix flickering on Windows, see https://github.com/emilk/egui/pull/2280
+                            self.windows_next_repaint_times.remove(&window_id);
+
+                            let _ = self.winit_app.run_ui_and_paint(event_loop, window_id);
```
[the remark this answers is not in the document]
ArthurBrussee: Hm yes - have folded it into event_result now so it goes through the same shutdown path as other errors.

crates/egui-winit/Cargo.toml
```diff
@@ -51,25 +51,33 @@ puffin = ["dep:puffin", "egui/puffin"]
 ## Allow serialization of [`WindowSettings`] using [`serde`](https://docs.rs/serde).
 serde = ["egui/serde", "dep:serde"]
 
+# wayland-cursor should enable the fs feature as it otherwise doesn't compile.
+# This is fixed upstream post https://github.com/Smithay/wayland-rs/commit/edd0f60d0baf09604553525c2636df5d6ba05d44 or v0.31.1
+# so should be removed when winit upgrades past that version.
+# Enable this feature manually for now (needed in both wayand and x11 backends).
```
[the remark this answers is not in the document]
ArthurBrussee: Oh yes that's nicer!

emilk:
I think pinning to 0.30.2 until https://github.com/rust-windowing/winit/issues/3837 / https://github.com/rust-windowing/winit/pull/3838 is resolved is the best option, especially since winit 0.30.5 is likely to be released before then next egui/eframe release. I don't see anything in the changelog entries for 0.30.3 or 0.30.4 that we need: https://docs.rs/winit/latest/winit/changelog/index.html

emilk:
I've tested this on macOS, but we need some testing on Windows, X11 and Wayland too.

A few things to test:
* `cargo -p custom_window_frame`
* `cargo -p egui_demo_app`
* `cargo -p egui_demo_app -F wgpu`
* `cargo -p egui_demo_app -F wgpu`
* `cargo -p multiple_viewports`

ArthurBrussee:
Thanks for the fixes! 

Done some more testing on Windows now (including all the above projects), and all seems to behave. (though there's a weird amount of latency for `cargo -p egui_demo_app -F wgpu` but it seems that's there both before and after the patch?)

I can try a bit on WSL later but even before the winit patches it had all kinds of sketchy behaviour & crashes, so it won't say that much. Would be good to have someone who daily drives linux to test.

crumblingstatue:
@emilk Everything seems to be working on Linux X11 (openbox), apart from resizing in the egui demo app when the window is maximized.
Not sure if it's intended to work or not when the window is maximized. But the resizes seem to be ignored when the window is maximized.

SiebenCorgie:
> I've tested this on macOS, but we need some testing on Windows, X11 and Wayland too.


I tested it on Wayland (Cosmic desktop and Gnome-46) as well as X11 on Gnome-46. Everything seems to work. 
The only thing I found was some _strange_ decoration on Gnome+Wayland for the window controls. But it also happens on `master`, so I'm confident that this not an issue related to the upgrade.

Screenshot

torokati44:
Oh that's just the result of GNOME folks _insisting_ that window decorations _have_ to be done by the clients (applications), and not by the server (the compositor / window manager), lest they may be _too consistent_. :roll_eyes: Window decorations look normal on KDE.

See: https://github.com/rust-windowing/winit/issues/1967

emilk:
Awesome - thanks for the help with the testing! Let's go 🚀

mwcampbell:
I just did a quick test of accessibility with a screen reader on Windows, and it looks good. Thanks @ArthurBrussee for getting this across the finish line.

torokati44:
> Sounds good! I've pinned to 0.30.2 with a comment, will keep an eye on the 0.30.5.

Soon! :tm: @ArthurBrussee 
https://github.com/rust-windowing/winit/pull/3849

torokati44:
'Tis out!

PR #4851, opened 2024-07-19
Fix iOS compilation of eframe

ardocrat:
Fixed changed the name of function `WinitApp::get_window_winit_id` to `WinitApp::window_id_from_viewport_id` and missed `egui::ViewportId` import.


emilk (approved):
Thanks! It would be greatly appreciated if someone added an iOS build to CI to catch this in the future

PR #4853, opened 2024-07-20
Move default fonts to new crate `epaint_default_fonts`

### LINKED ISSUE #2321: Licenses for default embedded fonts not included in `epaint` crate license
The `epaint` crate uses license `MIT OR Apache-2.0` but when one uses it with the default features or the `default_fonts` features a set of default fonts are embedded in the binary that have additional licenses.

This makes it so tools like [cargo-deny](https://github.com/EmbarkStudios/cargo-deny) (license compliance) and [cargo-about](https://github.com/EmbarkStudios/cargo-about/) (license attribution) can't see and validate/include the required licenses.

These are the licenses used for the default fonts:

- :heavy_check_mark: `fonts/emoji-icon-font.ttf` - license: MIT
- :heavy_check_mark: `fonts/Hack-Regular.ttf` - license: MIT
- :x: `fonts/NotoEmoji-Regular.ttf` - license: [`OFL-1.1`](https://spdx.org/licenses/OFL-1.1.html)
- :x: `fonts/Ubuntu-Light.ttf` - license: ["UBUNTU FONT LICENSE 1.0"](https://tldrlegal.com/license/ubuntu-font-license,-1.0)
    - haven't found an SPDX identifier for this in the [spec](https://spdx.org/licenses/)
    - so maybe we have make up our own ([SPDX supports it](https://spdx.github.io/spdx-spec/SPDX-license-expressions/))? like `LicenseRef-UFL-1.0`. 
    - what do you think @Jake-Shadle ?

So to be fully correct in the license definition of the `epaint` crate I believe we would have to specify to license as:

`(MIT OR Apache-2.0) AND OFL-1.1 AND LicenseRef-UFL-1.0`.

A recent and good previous example for another crate was https://github.com/dtolnay/unicode-ident/pull/9.

A drawback of that however would be that if you use the crate with `default-features = false` and not with `features = ["default_fonts"]` tools wouldn't know that these licenses are then not required.

So believe the most clean solution for everyone would be to split out the default fonts into a separate crate (like `epaint-default-fonts`) that has this full license, a crate that just does `include_bytes!` on the fonts and exposes global variable with them. And then in `epaint` we can keep  `MIT OR Apache-2.0` license and include the `epaint-default-fonts` optionally in that. 

Tools like cargo-deny and cargo-about would fully understand this and then only require that the `OFL-1.1` and `LicenseRef-UFL-1.0` are on the allow list if the optional default fonts are included.

What do you think about such an approach @emilk ?

alex-pinkus:
This allows license checking tools to omit the OFL and UFL licenses when `default_fonts` are turned off.

There was some discussion of versioning on the original issue; I have chosen to label this version as `0.28.1` to match the other crates. Happy to adjust the version as needed.

* Closes <https://github.com/emilk/egui/issues/2321>


crates/epaint_default_fonts/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# `epaint_default_fonts` changelog
+All notable changes to the `epaint_default_fonts` crate will be documented in this file.
+
+This file is updated upon each release.
+Changes since the last release can be found at <https://github.com/emilk/egui/compare/latest...HEAD> or by running the `scripts/generate_changelog.py` script.
```
emilk, line 5: Then we need to also add `epaint_default_fonts` to the `crate_names` field in `scripts/generate_changelog.py`

crates/epaint_default_fonts/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+//! A library containing built-in fonts for `epaint`, embedded as bytes.
+//!
+//! This is intended to be consumed through the `epaint` crate.
+
+pub const HACK_REGULAR: &[u8] = include_bytes!("../fonts/Hack-Regular.ttf");
+pub const NOTO_EMOJI_REGULAR: &[u8] = include_bytes!("../fonts/NotoEmoji-Regular.ttf");
+pub const UBUNTU_LIGHT: &[u8] = include_bytes!("../fonts/Ubuntu-Light.ttf");
+pub const EMOJI_ICON: &[u8] = include_bytes!("../fonts/emoji-icon-font.ttf");
```
emilk, line 8: For bonus points, add some nice docstrings to these :) But only if you want

emilk (approved):
Very thorough ⭐

emilk:
We should also add `epaint_default_fonts` to https://github.com/emilk/egui/blob/64d5f948e43b4be93e3a4541c87dcce74c948df4/.github/workflows/labels.yml#L32

crates/epaint_default_fonts/src/lib.rs
alex-pinkus: I'll see what I can come up with without having much context on these choices :)

alex-pinkus:
Ok, comments addressed!

emilk:
Thanks!

PR #4855, opened 2024-07-21
Fix virtual keyboard on (mobile) web

micmonay:
Hello,

I have made several corrections to stabilize the virtual keyboard on Android and IOS (Chrome and Safari).

I don't know if these corrections can have a negative impact in certain situations, but at the moment they don't cause me any problems.
I'll be happy to answer any questions you may have about these fixes.
These fixes correct several issues with the display of the virtual keyboard, particularly since update 0.28, which can be reproduced on the egui demo site.
We hope to be able to help you.

Thanks a lot for your work, I'm having a lot of fun with egui :)


crates/eframe/src/web/events.rs
```diff
@@ -564,6 +564,11 @@ fn install_touchend(runner_ref: &WebRunner, target: &EventTarget) -> Result<(),
                 runner.needs_repaint.repaint_asap();
                 event.stop_propagation();
                 event.prevent_default();
+
+                if runner.text_agent.has_focus() {
+                    runner.text_agent.set_focus(false);
+                    runner.text_agent.set_focus(true);
+                }
```
Mingun, line 571: What the intention to do that? If this is some workaround it should be documented explicitly, because it could introduce unintended behaviour in other places and if such behaviour would be found and traced down to this lines is should be obvious why they are here.

crates/eframe/src/web/text_agent.rs
```diff
@@ -39,6 +44,12 @@ impl TextAgent {
             let input = input.clone();
             move |event: web_sys::InputEvent, runner: &mut AppRunner| {
                 let text = input.value();
+                // Fix android virtual keyboard Gboard
+                // This removes the virtual keyboard's suggestion.
+                if !event.is_composing(){
+                    input.blur().unwrap();
+                    input.focus().unwrap();
```
emilk, line 51: better to not crash on unforseen problems

```suggestion
                    input.blur().ok();
                    input.focus().ok();
```

crates/eframe/src/web/text_agent.rs
```diff
@@ -119,14 +125,20 @@ impl TextAgent {
 
         let Some(ime) = ime else { return Ok(()) };
 
-        let canvas_rect = super::canvas_content_rect(canvas);
-        let cursor_rect = ime.cursor_rect.translate(canvas_rect.min.to_vec2());
+        //let canvas_rect = super::canvas_content_rect(canvas);
+        //let cursor_rect = ime.cursor_rect.translate(canvas_rect.min.to_vec2());
```
emilk, line 129: This is needed for when the canvas is not in the top left. Try `web_demo/multiple_apps.html`, for instance

crates/eframe/src/web/text_agent.rs
```diff
@@ -103,14 +114,9 @@ impl TextAgent {
...
+        zoom_factor: f32,
     ) -> Result<(), JsValue> {
-        // Mobile keyboards don't follow the text input it's writing to,
-        // instead typically being fixed in place on the bottom of the screen,
-        // so don't bother moving the text agent on mobile.
-        if is_mobile() {
-            return Ok(());
-        }
-
+        // Don't move the text agent if focus
```
emilk, line 119: What is this comment for?

emilk (changes requested):
Thanks for working on this 🙏

crates/eframe/src/web/mod.rs
```diff
@@ -136,10 +136,36 @@ fn get_canvas_element_by_id_or_die(canvas_id: &str) -> web_sys::HtmlCanvasElemen
 /// Returns the canvas in client coordinates.
 fn canvas_content_rect(canvas: &web_sys::HtmlCanvasElement) -> egui::Rect {
     let bounding_rect = canvas.get_bounding_client_rect();
+    // Need to fix Safari with vkeyboard open (flapping canvas position)
+    let rect_body = match web_sys::window()
+        .and_then(|w| w.document())
+        .and_then(|d| d.body())
+    {
+        None => egui::Rect::everything_above(0.),
```
emilk, line 144: This seems like a bad fallback since it will produce infinities. When do you expect we will get here?

crates/eframe/src/web/mod.rs
```diff
@@ -136,10 +136,36 @@ fn get_canvas_element_by_id_or_die(canvas_id: &str) -> web_sys::HtmlCanvasElemen
...
     let mut rect = egui::Rect::from_min_max(
-        egui::pos2(bounding_rect.left() as f32, bounding_rect.top() as f32),
-        egui::pos2(bounding_rect.right() as f32, bounding_rect.bottom() as f32),
+        egui::pos2(
+            bounding_rect.left() as f32 - rect_body.left(),
+            bounding_rect.top() as f32 - rect_body.top(),
+        ),
+        egui::pos2(
+            bounding_rect.right() as f32 - rect_body.right(),
+            bounding_rect.bottom() as f32 - rect_body.bottom(),
```
emilk, line 167: This makes no sense to me. What are you trying to accomplish? Maybe you are looking for `Rect::translate`?

micmonay:
Hello,

I have carried out several tests of my own and they all seem to work.
Do you need me to make any other corrections or do you have any other questions?

Thanks again for your library

crates/eframe/src/web/text_agent.rs
```diff
@@ -119,14 +124,27 @@ impl TextAgent {
...
         let Some(ime) = ime else { return Ok(()) };
 
-        let canvas_rect = super::canvas_content_rect(canvas);
+        let mut canvas_rect = super::canvas_content_rect(canvas);
+        // Fix for safari with virtual keyboard flapping position
+        if web_sys::window()
+            .and_then(|w| w.navigator().user_agent().ok())
+            .map_or(false, |ua| ua.contains("Safari"))
+        {
+            canvas_rect.min.y = canvas.offset_top() as f32;
```
emilk, line 133:
```suggestion
        // Fix for safari with virtual keyboard flapping position
        if is_mobile_safari() {
            canvas_rect.min.y = canvas.offset_top() as f32;
```

with

```rust
fn is_mobile_safari() -> bool {
    (|| {
        let user_agent = web_sys::window()?.navigator().user_agent().ok()?;
        let is_ios = user_agent.contains("iPhone") || user_agent.contains("iPad") || user_agent.contains("iPod");
        let is_safari = user_agent.contains("Safari");
        Some(is_ios && is_safari)
    })().unwrap_or(false)
}
```

v-kat:
@emilk do you think you could give this another review? I'm trying to hackily copy text_agent behavior into bevy in https://github.com/mvlabat/bevy_egui/pull/279 and this change would help.

emilk (approved):

emilk:
Need a rebase (and retest!) on top of https://github.com/emilk/egui/pull/4848

emilk:
You should be able to fix this branch with `git rebase master -i` and selecting only the relevant commits, or undo your `rebase` and do a `merge` instead

micmonay:
I think I succeeded, I'm sorry for the amount of commits and the repetition of texts... :/ 
I've tested the last commit, it seems OK. The script check went through.
I'm not very proficient with GIT outside the basics, but I'll do better next time. 
Let me know if I need to make any other corrections.

emilk:
I tested it again now, and on iOS Safari it now works _great_. Awesome work!

Issue #3959, opened 2024-02-03
Crash when changing a viewport aspect, that requires window recreation


AnotherNathan:
**Description**
When changing something on a viewport, that requires the recreation of the viewport window in order to apply the change, the application crashes.

I have only observed it on the glow backend, since the wgpu crashes immediately for me.

Sometimes I am getting a stack trace with a handle error and other times the application simply closes without any additional output.

The provided example changes, whether the viewport should show the close button or not (which internally requires the recreation of the viewport window).

<details><summary>Stacktrace</summary>
<br>

```
thread 'main' panicked at C:\Users\Nathan\.cargo\git\checkouts\egui-5e4507fa4153be06\114f820\crates\eframe\src\native\glow_integration.rs:815:69: 
called `Result::unwrap()` on an `Err` value: Error { raw_code: Some(6), raw_os_message: Some("Das Handle ist ungültig. (os error 6)"), kind: Misc }
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library\std\src\panicking.rs:645
   1: core::panicking::panic_fmt
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library\core\src\panicking.rs:72
   2: core::result::unwrap_failed
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library\core\src\result.rs:1653
   3: enum2$<core::result::Result<enum2$<glutin::context::PossiblyCurrentContext>,glutin::error::Error> >::unwrap<enum2$<glutin::context::PossiblyCurrentContext>,glutin::error::Error>
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112\library\core\src\result.rs:1077
   4: eframe::native::glow_integration::change_gl_context
             at C:\Users\Nathan\.cargo\git\checkouts\egui-5e4507fa4153be06\114f820\crates\eframe\src\native\glow_integration.rs:815
   5: eframe::native::glow_integration::GlowWinitRunning::run_ui_and_paint
             at C:\Users\Nathan\.cargo\git\checkouts\egui-5e4507fa4153be06\114f820\crates\eframe\src\native\glow_integration.rs:621
   6: eframe::native::glow_integration::impl$1::run_ui_and_paint
             at C:\Users\Nathan\.cargo\git\checkouts\egui-5e4507fa4153be06\114f820\crates\eframe\src\native\glow_integration.rs:402
[...]

PR #4868, opened 2024-07-23
Fix: panic when dragging window between monitors of different pixels_per_point

### LINKED ISSUE #4178: There is on very rare occasions a case where a panic occurred

### LINKED ISSUE #4179: There is on very rare occasions a case where a panic occurred at line 152 of layers.rs

rustbasic:
Fix: panic when dragging window between monitors of different pixels_per_point

This will continue to help us as we develop `egui`.
I hope you agree with my defense of `panic`.

* Related #3959
* Related #4088

* Closes #4178
* Closes #4179

There is also a way to add log if necessary.
```
                log::debug!("Anti-panic behavior occurs");
```


crates/egui/src/layers.rs
```diff
@@ -153,6 +153,12 @@ impl PaintList {
     /// and then later setting it using `paint_list.set(idx, cr, frame);`.
     #[inline(always)]
     pub fn set(&mut self, idx: ShapeIdx, clip_rect: Rect, shape: Shape) {
+        if self.0.len() <= idx.0 {
+            #[cfg(feature = "log")]
+            log::warn!("Index {} is out of bounds for PaintList", idx.0);
+            self.add(clip_rect, Shape::Noop);
```
emilk, line 159:
```suggestion
            return;
```

crates/egui/src/context.rs
```diff
@@ -2418,12 +2418,14 @@ impl Context {
 
         self.write(|ctx| {
             let tessellation_options = ctx.memory.options.tessellation_options;
-            let texture_atlas = ctx
-                .fonts
-                .get(&pixels_per_point.into())
+            let texture_atlas = if let Some(fonts) = ctx.fonts.get(&pixels_per_point.into()) { fonts.texture_atlas() } else {
```
emilk, line 2421:
```suggestion
            let texture_atlas = if let Some(fonts) = ctx.fonts.get(&pixels_per_point.into()) {
            	fonts.texture_atlas()
           	} else {
```

crates/egui/src/context.rs
```diff
@@ -2418,12 +2418,14 @@ impl Context {
...
+                #[cfg(feature = "log")]
+                log::warn!("No font size matching {pixels_per_point} pixels per point found.");
+                ctx.fonts.iter().next()
                 .expect("tessellate called with a different pixels_per_point than the font atlas was created with. \
-                         You should use egui::FullOutput::pixels_per_point when tessellating.")
-                .texture_atlas();
+                You should use egui::FullOutput::pixels_per_point when tessellating.")
```
emilk, line 2426: This error message is wrong now, it should be "No fonts loaded"

emilk (changes requested):

emilk (approved):

PR #4872, opened 2024-07-25
Allow setting a fill color for ProgressBar track

andrewbrey:
Here is the demo widget gallery with the `ProgressBar::track_fill` set to dark blue:

```rust
let progress_bar = egui::ProgressBar::new(progress)
    .show_percentage()
    .track_fill(egui::Color32::DARK_BLUE)
    .animate(*animate_progress_bar);
```

track

### DIFF
--- a/crates/egui/src/widgets/progress_bar.rs
+++ b/crates/egui/src/widgets/progress_bar.rs
@@ -15,6 +15,7 @@ pub struct ProgressBar {
     desired_height: Option<f32>,
     text: Option<ProgressBarText>,
     fill: Option<Color32>,
+    track_fill: Option<Color32>,
     animate: bool,
     rounding: Option<Rounding>,
 }
@@ -28,6 +29,7 @@ impl ProgressBar {
             desired_height: None,
             text: None,
             fill: None,
+            track_fill: None,
             animate: false,
             rounding: None,
         }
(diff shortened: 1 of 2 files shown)


crates/egui/src/widgets/progress_bar.rs
```diff
@@ -15,6 +15,7 @@ pub struct ProgressBar {
     desired_height: Option<f32>,
     text: Option<ProgressBarText>,
     fill: Option<Color32>,
```
bircni, line 17: What about calling the current fill `indicator_color` this could make it more clear?

crates/egui/src/widgets/progress_bar.rs
```diff
@@ -15,6 +15,7 @@ pub struct ProgressBar {
...
+    track_fill: Option<Color32>,
```
bircni, line 18: then renaming this to `track_color` would make it clear

bircni (changes requested):

crates/egui/src/widgets/progress_bar.rs
andrewbrey: The reason for not doing something like this was because that would be a breaking change to the current API, so I opted for a less disruptive name choice. If the maintainers of egui are ok with a breaking change to the `ProgressBar` API, I think using names like the ones you suggest would be nice as well :)

bircni: I think this would make it more clear for everyone!
@emilk what do you think?

PR #4875, opened 2024-07-26
Improved `filter` in `FontBook`

rustbasic:
Improved `filter` in `FontBook`

Issues: `*filter == chr.to_string()` doesn't seem to make sense.

### DIFF
--- a/crates/egui_demo_lib/src/demo/font_book.rs
+++ b/crates/egui_demo_lib/src/demo/font_book.rs
@@ -79,7 +79,11 @@ impl crate::View for FontBook {
                 ui.spacing_mut().item_spacing = egui::Vec2::splat(2.0);
 
                 for (&chr, name) in named_chars {
-                    if filter.is_empty() || name.contains(filter) || *filter == chr.to_string() {
+                    if filter.is_empty()
+                        || name.contains(filter)
+                        || format!("{:x}", chr as u32).starts_with(filter)
+                        || filter.contains(chr)
+                    {
                         let button = egui::Button::new(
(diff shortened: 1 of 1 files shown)


crates/egui_demo_lib/src/demo/font_book.rs
```diff
@@ -79,7 +79,11 @@ impl crate::View for FontBook {
                 ui.spacing_mut().item_spacing = egui::Vec2::splat(2.0);
 
                 for (&chr, name) in named_chars {
-                    if filter.is_empty() || name.contains(filter) || *filter == chr.to_string() {
+                    if filter.is_empty()
+                        || name.contains(filter)
+                        || format!("{:x}", chr as u32).starts_with(filter)
+                        || filter.contains(chr)
```
emilk, line 85: `*filter == chr.to_string()` is for entering a specific glyph (e.g. `Å`) to find it.

`filter.contains(chr)` should work for that too, but entering "infiity" will now include all the letters that make up the word "infinity", which seems a bit much?

Maybe it would make sense to pull this out to its own function with a few small unit tests

rustbasic: > `*filter == chr.to_string()` is for entering a specific glyph (e.g. `Å`) to find it.
> 
> `filter.contains(chr)` should work for that too, but entering "infiity" will now include all the letters that make up the word "infinity", which seems a bit much?
> 
> Maybe it would make sense to pull this out to its own function with a few small unit tests

It has the same effect as `*filter == chr.to_string()`.

When you type "infiity", the intention is to include all the letters that make up the word "infinity". 
That was convenient to use.

Issue #4881, opened 2024-07-27
Regression: dragging value of slider no longer supports out of range values


ishbosamiya:
**Describe the bug**
Dragging the value part of the slider in `0.27` allowed out of range values when using `egui::Slider::clamp_to_range(false)`. In `0.28`, dragging forces the value to be in the slider's range. Setting a value out of range via keyboard input works as expected.

**To Reproduce**
* Create a slider with `egui::Slider::clamp_to_range` set to `false`.
* Try to drag the value part of the slider outside the range of the slider.

**Expected behavior**
Dragging the value part of the slider should allow going out of the range of the slider if `egui::Slider::clamp_to_range` is set to `false`.

**Screenshots**

https://github.com/user-attachments/assets/d9e98187-fdd6-4584-b761-49498ecf571d

**Additional context**
It seems like the changes introduced by https://github.com/emilk/egui/pull/4728 caused this change in behavior.

PR #4882, opened 2024-07-27
Add `clamp_to_range_on_drag` option to `DragValue` and `clamp_to_range_on_value_drag` option to `Slider`

ishbosamiya:
Added `egui::DragValue::clamp_to_range_on_drag` and `egui::Slider::clamp_to_range_on_value_drag` to allow the user to determine if the value should be clamped while dragging.

The default for `egui::DragValue::clamp_to_range_on_drag` and `egui::Slider::clamp_to_range_on_value_drag` is set to `true`. This means the default behavior mimics `0.28`.

If the default behavior must match `0.27`, `egui::DragValue::clamp_to_range_on_drag` should be `true` and `egui::Slider::clamp_to_range_on_value_drag` should be `false`.

* Closes https://github.com/emilk/egui/issues/4881

https://github.com/user-attachments/assets/1ec81142-8ab3-4eb2-a9c9-afb61426dffc

### DIFF
--- a/crates/egui/src/widgets/drag_value.rs
+++ b/crates/egui/src/widgets/drag_value.rs
@@ -39,7 +39,6 @@ pub struct DragValue<'a> {
     suffix: String,
     range: RangeInclusive<f64>,
     clamp_to_range: bool,
-    clamp_to_range_on_drag: bool,
     min_decimals: usize,
     max_decimals: Option<usize>,
     custom_formatter: Option<NumFormatter<'a>>,
@@ -71,7 +70,6 @@ impl<'a> DragValue<'a> {
             suffix: Default::default(),
             range: f64::NEG_INFINITY..=f64::INFINITY,
             clamp_to_range: true,
-            clamp_to_range_on_drag: true,
             min_decimals: 0,
             max_decimals: None,
             custom_formatter: None,
@@ -117,20 +115,6 @@ impl<'a> DragValue<'a> {
     #[inline]
     pub fn clamp_to_range(mut self, clamp_to_range: bool) -> Self {
         self.clamp_to_range = clamp_to_range;
-        self
-    }
-
-    /// If set to `true`, dragging will clamp the value to the sliding
-    /// [`Self::range`] (if any).
-    ///
-    /// If set to `false`, the value can be dragged outside of the
-    /// sliding [`Self::range`] (if [`Self::clamp_to_range`] is also
-    /// `false`).
-    ///
-    /// Default: `true`.
-    #[inline]
-    pub fn clamp_to_range_on_drag(mut self, clamp_to_range_on_drag: bool) -> Self {
-        self.clamp_to_range_on_drag = clamp_to_range_on_drag;
         self
     }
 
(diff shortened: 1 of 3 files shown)


emilk:
Do we really need both this options?
It seems to me that `clamp_to_range` should clamp both drags and keyboard edits instead.

ishbosamiya:
I don't think both these options are required. This PR was made assuming that the regression was intentional, it seems like it is not.

I will update the PR soon to use just `clamp_to_range`.

ishbosamiya:
With the latest update, the behavior now mimics `0.27` where `clamp_to_range` is used to determine if the value should be clamped while dragging instead of always clamping.

The PR title will need to be updated, but I am not sure about the guidelines for this in `egui`, so I am leaving it up to you to edit when merging.

emilk:
I'm trying to refresh my memory of the last PR: https://github.com/emilk/egui/pull/4728 by @Wumpf 

It seems like we want is three modes:

A) no clamping
B) clamping only user input (dragging or editing text), but leave existing value intact
C) always clamp

The difference between mode B and C is:

```rs
let mut x = 42.0;
ui.add(DragValue::new(&mut x).range(0.0..=1.0));
// What will `x` be here?
```

I _think_ the intention of https://github.com/emilk/egui/pull/4728 was that you get the different modes via:

* A): don't call `.range()`
* B) call `.range()` and `.clamp_to_range(false)`
* C) call `.range()`

At least, that's how I think both `Slider` and `DragValue` should work (it seems like the least surprising behavior). But obviously that's not how `DragValue` currently works, because it checks `clamp_to_range` too much. I believe the correct fix is to remove these `if`s:

https://github.com/emilk/egui/blob/a48b894917f27d58f532e5ac47fced3d27c42d99/crates/egui/src/widgets/drag_value.rs#L504

https://github.com/emilk/egui/blob/a48b894917f27d58f532e5ac47fced3d27c42d99/crates/egui/src/widgets/drag_value.rs#L540

Ping @Wumpf

Wumpf:
Went through it again and can confirm the behavior of the `Slider` as described by your comment. It is also reflected in the doc strings just fine.

On how to fix it:
The effect of clamping programmatic input if  `clamp_to_range` is `false` is done early on here https://github.com/emilk/egui/blob/a48b894917f27d58f532e5ac47fced3d27c42d99/crates/egui/src/widgets/drag_value.rs#L473
So afterwards, all clamping is about user input. Since we earlier asserted that user input is always clamped against the set range (which may be infinite in either direction, thus not clamping!), we have to apply clamping to range indiscriminately to user input, but not if no user input occurred. Both pieces of code linked by @emilk are about keyboard user input and indeed this means that there we should apply clamping always.
The last piece where we have to clamp to `range` is when dragging. That happens here https://github.com/emilk/egui/blob/a48b894917f27d58f532e5ac47fced3d27c42d99/crates/egui/src/widgets/drag_value.rs#L615 and already does the right thing

So, yes! What @emilk said :)

Furthermore, documentation on clamping should be equalized between `Slider` and `DragValue`, as in both cases the semantics should be the exact same after this change.

emilk:
Then I'll make a new PR to get us the desired behavior

PR #4889, opened 2024-07-30
Fix: hint text follows the alignment set on the `TextEdit`

PrimmR:
Updated the `show_content` method of the `TextEdit` widget so that the requested text alignment (if any) is also applied to the hint text.

Currently, hint text is always aligned to the top left of the widget, which may be inconsistent with the actual location of text entered by the user. I think it's more intuitive for the hint text to be positioned as if the user had entered the same text themselves.

Here are some comparisons of the difference between how the hint text and the entered text looks, before and after this change:

### Previous Behaviour
Prev

### Updated Behaviour
Update

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -682,7 +682,11 @@ impl<'t> TextEdit<'t> {
                         hint_text_font_id,
                     )
                 };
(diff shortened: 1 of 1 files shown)


PrimmR:
And all the checks seem to pass, great ^^

emilk (approved):
Makes sense!

PR #4893, opened 2024-07-31

rustbasic:

Pull requests :
* Related #4303 
* Related #4305
* Related #4584

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -793,15 +793,7 @@ impl Prepared {
 
         let content_size = content_ui.min_size();
 
-        let scroll_delta = content_ui
-            .ctx()
-            .frame_state_mut(|state| std::mem::take(&mut state.scroll_delta));
-
         for d in 0..2 {
-            // FrameState::scroll_delta is inverted from the way we apply the delta, so we need to negate it.
-            let mut delta = -scroll_delta.0[d];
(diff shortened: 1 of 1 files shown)


crates/egui/src/containers/scroll_area.rs
```diff
@@ -793,22 +793,25 @@ impl Prepared {
...
-            let mut animation = scroll_delta.1;
-
             // We always take both scroll targets regardless of which scroll axes are enabled. This
             // is to avoid them leaking to other scroll areas.
             let scroll_target = content_ui
                 .ctx()
                 .frame_state_mut(|state| state.scroll_target[d].take());
 
             if scroll_enabled[d] {
+                let (scroll_delta_0, scroll_delta_1) = content_ui.ctx().frame_state_mut(|state| {
```
bircni, line 804: This could also be solved by using less lines of code if wanted
```rust
let (mut delta, mut animation) = content_ui.ctx().frame_state_mut(|state| {
                    (
                        -std::mem::take(&mut state.scroll_delta.0[d]),
                        std::mem::take(&mut state.scroll_delta.1),
                    )
                });
```

rustbasic: Using `-` here is bad for readability.

It's okay to do `let (scroll_delta, scroll_animation)`, but since it's only used temporarily, it's not a big problem.

bircni: you are right

PR #4436, opened 2024-04-30
IME for chinese

rustbasic:
* Completed.

* Closes #4430

IME for chinese

PR #4896, opened 2024-08-01
Fix: Ensures correct IME behavior when the text input area gains or loses focus.

rustbasic:
Fix: Ensures correct IME behavior when the text input area gains or loses focus.

Fix: Handling `state.ime_enabled` in multiple `TextEdit`.
Fix: A symptom of characters being copied when there are multiple TextEdits.

* Related #4137
* Related #4358 
* Closes #4374
* Related #4436
* Related #4794 
* Related #4908 

* Related #5008

Fix Issues: When focus is moved elsewhere, you must set `state.ime_enabled = false`, otherwise the IME will have problems when focus returns.

Fix Issues: A symptom of characters being copied when there are multiple TextEdits.
Deletes all current `IME events`, preventing them from being copied to `other TextEdits`, without saving the `TextEdit ID`, 

( Related Issues: Some `LINUX` seem to trigger an IME enable event on startup. So, when we gained focus, we do `state.ime_enabled = false`. )

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -748,6 +748,10 @@ impl<'t> TextEdit<'t> {
 
         if state.ime_enabled && response.lost_focus() {
             state.ime_enabled = false;
+            if let Some(mut ccursor_range) = state.cursor.char_range() {
+                ccursor_range.secondary.index = ccursor_range.primary.index;
+                state.cursor.set_char_range(Some(ccursor_range));
+            }
         }
 
         state.clone().store(ui.ctx(), id);
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -1005,6 +1005,8 @@ fn events(
                     if text_mark == "\n" || text_mark == "\r" {
                         None
                     } else {
+                        state.ime_enabled = false;
+
                         // Empty prediction can be produced when user press backspace
                         // or escape during IME, so we clear current text.
                         let mut ccursor = text.delete_selected(&cursor_range);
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -746,6 +746,10 @@ impl<'t> TextEdit<'t> {
             }
         }
 
+        if state.ime_enabled && response.lost_focus() {
```
bircni, line 749: Couldn`t you just leave the `state.ime_enabled` ? Because if its true its getting set to false and else it is already false, so to me it seems like an unnecessary condition

rustbasic: > Couldn`t you just leave the `state.ime_enabled` ? Because if its true its getting set to false and else it is already false, so to me it seems like an unnecessary condition

When focus is moved elsewhere, you must set `ime_enabled` to 'false`, otherwise the IME will have problems when focus returns.

bircni: sure the idea was the following
```rust
if response.lost_focus() {
state.ime_enabled = false;
}
```

rustbasic: > sure the idea was the following
> 
> ```rust
> if response.lost_focus() {
> state.ime_enabled = false;
> }
> ```

Yes, I thought about this too. If ime_enabled is not 'true`, there is no need to check lost_focus() , so there is a difference in speed and readability. Let's wait for emilk's opinion on which one is better.

thank you.

bircni: Ok but if so it would be good to add a comment there anyway

emilk (approved):

PR #4897, opened 2024-08-01
add missing self. (iOS target)

frederik-uni:
I tried building the frame_template with `cargo bundle` for iOS. didn't work so I fixed it.


lucasmerlin:
Ah I just tested if this workaround is still necessary and it doesn't seem like it is, so I've removed it in #4898

PR #4898, opened 2024-08-01
Add ios pipeline

lucasmerlin:
This PR
- adds a pipeline to check the ios build
- removes the iOS WaitUntil workaround, which doesn't seem to be necessary anymore after the winit update (and caused the build for iOS to fail again because of a missing self

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,4 +1,4 @@
-on: [push, pull_request]
+on: [ push, pull_request ]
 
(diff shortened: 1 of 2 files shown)


emilk (approved):
Thanks! Looks good, but CI is red

lucasmerlin:
@emilk finally got around to fixing the lint warnings, sorry it took so long

emilk (approved):
Thanks!

PR #4901, opened 2024-08-01
Function : Returns the `FontId` corresponding to the given `TextStyle`

rustbasic:
Function : Returns the `FontId` corresponding to the given `TextStyle`

Before : hard
```
    let font_id = TextStyle::Button.resolve(ui.style());
```

After : easy
```
    let font_id = ui.text_style_font_id(&TextStyle::Button);
```

I'm not sure what would be a good name for the function.

### DIFF
--- a/crates/egui/src/ui.rs
+++ b/crates/egui/src/ui.rs
@@ -545,7 +545,7 @@ impl Ui {
     }
 
     /// Returns the `FontId` corresponding to the given `TextStyle`.
-    pub fn font_id_by_text_style(&self, style: &TextStyle) -> FontId {
+    pub fn text_style_font_id(&self, style: &TextStyle) -> FontId {
         style.resolve(&self.style)
     }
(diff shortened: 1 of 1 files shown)


crates/egui/src/ui.rs
```diff
@@ -544,6 +544,11 @@ impl Ui {
         self.painter().layer_id()
     }
 
+    /// Returns the `FontId` corresponding to the given `TextStyle`.
+    pub fn font_id_by_text_style(&self, style: &TextStyle) -> FontId {
```
bircni, line 548: I would suggest to call it `font_id()` or `text_style_font_id` this would match with `text_style_height`

rustbasic: thank you.
Both `font_id()` and `text_style_font_id()` looks good.
I'll have to wait for emilk's opinion.

Issue #4884, opened 2024-07-28
Replace `directories` dependency


valadaptive:
**Describe the bug**

The maintainer of the `directories` crate has [intentionally added an MPL-licensed dependency](https://github.com/dirs-dev/dirs-sys-rs/issues/21) and refuses to remove it. While the MPL isn't technically a viral license, it would still be nice to remove it to avoid any potential legal trouble.

A potential candidate is [etcetera](https://github.com/lunacookies/etcetera), but it appears to have different behavior on macOS with regards to the bundle identifier.

YgorSouza:
I wonder if eframe shouldn't just copy the part of the code it needs from one of these crates. It's just one folder anyway. Might contribute towards #4481.

PR #4904, opened 2024-08-02
Remove the `directories` dependency

YgorSouza:
eframe now has its own logic to find the storage_dir to persist the app when the persistence feature is enabled, instead of using the directories crate. The directory should be the same as before (verified with a unit test).

* Closes <https://github.com/emilk/egui/issues/4884>

### DIFF
--- a/crates/eframe/Cargo.toml
+++ b/crates/eframe/Cargo.toml
@@ -57,12 +57,15 @@ android-native-activity = ["egui-winit/android-native-activity"]
 ## If you plan on specifying your own fonts you may disable this feature.
 default_fonts = ["egui/default_fonts"]
 
+# Enable the directories dependency for a more precise file path for the persistence feature.
+# Not needed if you provide your own path, or if the simpler default path resolvers are good enough.
+directories = ["dep:directories"]
+
 ## Use [`glow`](https://github.com/grovesNL/glow) for painting, via [`egui_glow`](https://github.com/emilk/egui/tree/master/crates/egui_glow).
 glow = ["dep:egui_glow", "dep:glow", "dep:glutin-winit", "dep:glutin"]
 
 ## Enable saving app state to disk.
 persistence = [
-  "directories",
   "egui-winit/serde",
   "egui/persistence",
   "ron",
--- a/crates/eframe/Cargo.toml
+++ b/crates/eframe/Cargo.toml
@@ -57,15 +57,12 @@ android-native-activity = ["egui-winit/android-native-activity"]
 ## If you plan on specifying your own fonts you may disable this feature.
 default_fonts = ["egui/default_fonts"]
 
-## Enable the directories dependency for a more precise file path for the persistence feature.
-## Not needed if you provide your own path, or if the simpler default path resolvers are good enough.
-directories = ["dep:directories"]
-
 ## Use [`glow`](https://github.com/grovesNL/glow) for painting, via [`egui_glow`](https://github.com/emilk/egui/tree/master/crates/egui_glow).
 glow = ["dep:egui_glow", "dep:glow", "dep:glutin-winit", "dep:glutin"]
 
 ## Enable saving app state to disk.
 persistence = [
+  "dep:home",
   "egui-winit/serde",
   "egui/persistence",
   "ron",
@@ -150,7 +147,6 @@ image = { workspace = true, features = ["png"] } # Needed for app icon
 winit = { workspace = true, default-features = false, features = ["rwh_06"] }
 
 # optional native:
-directories = { version = "5", optional = true }
 egui-wgpu = { workspace = true, optional = true, features = [
   "winit",
 ] } # if wgpu is used, use it with winit
@@ -160,6 +156,7 @@ pollster = { version = "0.3", optional = true } # needed for wgpu
 # this can be done at the same time we expose x11/wayland features of winit crate.
 glutin = { workspace = true, optional = true }
 glutin-winit = { workspace = true, optional = true }
+home = { workspace = true, optional = true }
 puffin = { workspace = true, optional = true }
 wgpu = { workspace = true, optional = true, features = [
   # Let's enable some backends so that users can use `eframe` out-of-the-box
@@ -187,6 +184,7 @@ objc2-app-kit = { version = "0.2.0", features = [
 # windows:
 [target.'cfg(any(target_os = "windows"))'.dependencies]
 winapi = { version = "0.3.9", features = ["winuser"] }
+windows-sys = { workspace = true, features = ["Win32_Foundation", "Win32_UI_Shell", "Win32_System_Com"] }
 
 # -------------------------------------------
 # web:
@@ -255,3 +253,7 @@ wgpu = { workspace = true, optional = true, features = [
   # without having to explicitly opt-in to backends
   "webgpu",
 ] }
+
+# Native dev dependencies for testing
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+directories = "5"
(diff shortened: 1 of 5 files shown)


bircni:
Good idea - this should be documented somewhere for users who will search for it

YgorSouza:
> Good idea - this should be documented somewhere for users who will search for it

I fixed the comment style in the Cargo.toml. Now it should be automatically added to the Feature Flags section in the [docs](https://docs.rs/eframe/0.28.1/eframe/#feature-flags), thanks to the document-features crate. Thank you.

crates/eframe/src/native/file_storage.rs
```diff
@@ -10,15 +10,60 @@ use std::{
...
 /// or the title argument to [`crate::run_native`].
 ///
-/// On native the path is picked using [`directories::ProjectDirs::data_dir`](https://docs.rs/directories/5.0.1/directories/struct.ProjectDirs.html#method.data_dir) which is:
+/// On native, if the `directories` feature is enabled, the path is picked using
+/// [`directories::ProjectDirs::data_dir`](https://docs.rs/directories/5.0.1/directories/struct.ProjectDirs.html#method.data_dir)
+/// which is:
 /// * Linux:   `/home/UserName/.local/share/APP_ID`
 /// * macOS:   `/Users/UserName/Library/Application Support/APP_ID`
-/// * Windows: `C:\Users\UserName\AppData\Roaming\APP_ID`
+/// * Windows: `C:\Users\UserName\AppData\Roaming\APP_ID\data`
```
YgorSouza, line 18: Even directories's own documentation is wrong about this. But the `data` folder is being appended right [here](https://github.com/dirs-dev/directories-rs/blob/da65f064a8bb18d790600433fadeed17a54c793a/src/win.rs#L77).

crates/eframe/Cargo.toml
```diff
@@ -58,12 +58,15 @@ android-native-activity = ["egui-winit/android-native-activity"]
 ## If you plan on specifying your own fonts you may disable this feature.
 default_fonts = ["egui/default_fonts"]
 
+## Enable the directories dependency for a more precise file path for the persistence feature.
+## Not needed if you provide your own path, or if the simpler default path resolvers are good enough.
+directories = ["dep:directories"]
```
emilk, line 63: Should this be a default feature to avoid breaking changes?

How often does the naive approach fail, and `directories` be required?

YgorSouza: The problem with making it default is that people using eframe without persistence would have the directories dependency even though it isn't being used for anything, unless they pass `--no-default-features`.

This naive function relies on default environment variables, so it will fail on custom environments that don't have those variables. I'm not sure how often that comes up for eframe users.

Another option would be to use the [home](https://crates.io/crates/home) crate (which eframe already depends on indirectly) to get the paths on Linux and macOS, and a modified version of [this code](https://github.com/rust-lang/cargo/blob/f2d1e37b56529f5200c3861be8b3d1c24f62c135/crates/home/src/windows.rs#L20-L42) to get the appdata path on Windows. I think that would completely reproduce what directories does, and then it could be removed or left as a dev dependency for the unit test.

crates/eframe/src/native/file_storage.rs
```diff
@@ -10,15 +10,60 @@ use std::{
...
+    use std::env::var_os;
+    match OS::from_target_os() {
+        OS::Nix => var_os("XDG_DATA_HOME")
+            .map(PathBuf::from)
+            .or_else(|| var_os("HOME").map(|s| PathBuf::from(s).join(".local").join("share")))
+            .map(|p| {
+                p.join(
+                    app_id
+                        .to_lowercase()
+                        .replace(|c: char| c.is_ascii_whitespace(), ""),
```
emilk, line 53: This doesn't seems enough to sanitize the path 😬 

https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names

I suggest we replace all characters not in `[a-zA-Z0-9_-]`

YgorSouza: This is actually just replicating what the directories crate does.

The Linux impl changes the text to lowercase and removes spaces:

<https://github.com/dirs-dev/directories-rs/blob/da65f064a8bb18d790600433fadeed17a54c793a/src/lin.rs#L92-L110>

The macOS impl changes the spaces to dashes:

<https://github.com/dirs-dev/directories-rs/blob/da65f064a8bb18d790600433fadeed17a54c793a/src/mac.rs#L91-L98>

And the Windows impl doesn't change anything:

<https://github.com/dirs-dev/directories-rs/blob/da65f064a8bb18d790600433fadeed17a54c793a/src/win.rs#L98-L100>

I suppose we could be more strict, but for now I'm just trying to remove the dependency with minimal changes to the functionality.

crates/eframe/Cargo.toml
YgorSouza: I made the change to just replicate what directories does entirely instead of using the naive approach. If it's too much unsafe/windows-specific code to be in eframe, we could use the etcetera crate instead, as suggested in the issue. Although it is a bit outdated with respect to the home crate at the moment.

crates/eframe/src/native/file_storage.rs
```diff
@@ -10,13 +10,80 @@ use std::{
...
+    use windows_sys::Win32::UI::Shell::{
+        FOLDERID_RoamingAppData, SHGetKnownFolderPath, KF_FLAG_DONT_VERIFY,
+    };
+
+    extern "C" {
+        fn wcslen(buf: *const u16) -> usize;
+    }
+    unsafe {
+        let mut path = ptr::null_mut();
+        match SHGetKnownFolderPath(
```
YgorSouza, line 63: Could also use the [known-folders](https://crates.io/crates/known-folders) crate, which has this function and nothing else. But it's still another dependency, which might end up pulling a different version of windows-sys, so I don't know if it's worth it.

emilk (approved):
Thanks!

Ultimately it would be nice if someone published all this as its own crate to reduce the amount of code in `eframe` :)

bircni:
@emilk which part do you want to be extracted?

emilk:
Basically all the code that was added in this PR :)

bircni:
Just started a bit
@YgorSouza if you want I'd be happy to get some help via PR or whatever :-)
https://github.com/bircni/file_storage

Mingun:
Issue tracker should be different in this line:
https://github.com/bircni/file_storage/blob/1fbec8ed8c3d853a99cde7bab33a7f17eaf6bc79/src/os.rs#L73

bircni:
The question is what to do with the `os.rs` ?
Put it in the crate too or get it via egui as a dependency

YgorSouza:
Do we really want to make a new crate just for that? I mean, previously eframe was getting all this code from the `directories` crate, but the linked issue asked to remove it because the author added an MPL-2.0 dependency that could be replaced by a single line of code, and refuses to remove it. Note that egui accepts the MPL-2.0 license, and there are 5 other dependencies with this license in the lockfile. The suggested alternative was the `etcetera` crate, which does the same thing as `directories`, but as it isn't updated often, I figured it would cause duplicated dependencies in the long run (it already uses a different version of `windows-sys` than `winit` does, but so does `directories`), so I just put the code directly in eframe. The only non-trivial bit of code is the `roaming_appdata` function for Windows, which can be replaced by the `known-folders` crate as I mentioned above. So if we don't want to have this code in eframe, there are plenty of existing options already: either use `etcetera` for everything, or use `known-folders` for the Windows part, or just revert this PR and go back to `directories`. I feel like rewriting everything in a new crate would just be redundant.

PR #4906, opened 2024-08-02
Add the nursery lint to the project

bircni:
I wanted to try to enable the clippy::nursery lint -> these are the things i modified


crates/eframe/src/native/file_storage.rs
```diff
@@ -56,21 +56,24 @@ impl FileStorage {
...
     pub fn from_app_id(app_id: &str) -> Option<Self> {
         crate::profile_function!(app_id);
-        if let Some(data_dir) = storage_dir(app_id) {
-            if let Err(err) = std::fs::create_dir_all(&data_dir) {
-                log::warn!(
-                    "Saving disabled: Failed to create app path at {:?}: {}",
-                    data_dir,
-                    err
-                );
+        storage_dir(app_id).map_or_else(
```
ItsEthra, line 59: This `map_or_else` doesn't make sense to me here, I think a better way would be to use guarded let-else to return early

crates/eframe/src/native/glow_integration.rs
```diff
@@ -363,11 +363,10 @@ impl WinitApp for GlowWinitApp {
...
         let viewport_id = *glutin.viewport_from_window.get(&window_id)?;
-        if let Some(viewport) = glutin.viewports.get(&viewport_id) {
-            viewport.window.clone()
-        } else {
-            None
-        }
+        glutin
+            .viewports
+            .get(&viewport_id)
+            .map_or_else(|| None, |viewport| viewport.window.clone())
```
ItsEthra, line 369: This is `map_or`

crates/eframe/src/native/glow_integration.rs
```diff
@@ -867,16 +866,17 @@ fn change_gl_context(
...
-        not_current_context
-    } else {
-        crate::profile_scope!("make_not_current");
-        current_gl_context
-            .take()
-            .unwrap()
-            .make_not_current()
-            .unwrap()
-    };
+    let not_current = not_current_gl_context.take().map_or_else(
```
ItsEthra, line 869: This is unwrap_or_else

crates/eframe/src/native/wgpu_integration.rs
```diff
@@ -372,11 +372,10 @@ impl WinitApp for WgpuWinitApp {
     ) -> Result<EventResult> {
         self.initialized_all_windows(event_loop);
 
-        if let Some(running) = &mut self.running {
-            running.run_ui_and_paint(window_id)
-        } else {
-            Ok(EventResult::Wait)
-        }
+        self.running.as_mut().map_or_else(
```
ItsEthra, line 375: This is map_or

ItsEthra (commented):
note: i'm not a maintainer

crates/eframe/src/native/glow_integration.rs
bircni: and_then is better here

crates/egui-winit/src/lib.rs
```diff
@@ -188,7 +188,7 @@ impl State {
     }
 
     /// Returns [`false`] or the last value that [`Window::set_ime_allowed()`] was called with, used for debouncing.
-    pub fn allow_ime(&self) -> bool {
+    pub const fn allow_ime(&self) -> bool {
```
YgorSouza, line 191: Is there any point in having const methods in a type that doesn't have a const constructor? No one will be able to call them in a const context anyway, and any internal refactoring might lead to breaking changes.

bircni:
Update:
It seems like everything works except one error which I couldn't find where it comes from:

```shell
➜  egui git:(add-clippy-nursery) ✗ cargo check --quiet -p egui_glow --all-features                                                                                                                                     
error[E0277]: the trait bound `ActiveEventLoop: raw_window_handle::borrowed::HasDisplayHandle` is not satisfied
  --> crates/egui_glow/src/winit.rs:43:13
   |
43 |             event_loop,
   |             ^^^^^^^^^^ the trait `raw_window_handle::borrowed::HasDisplayHandle` is not implemented for `ActiveEventLoop`
   |
   = help: the following other types implement trait `raw_window_handle::borrowed::HasDisplayHandle`:
             raw_window_handle::borrowed::DisplayHandle<'a>
             &H
             &mut H
   = note: required for the cast from `&ActiveEventLoop` to `&dyn raw_window_handle::borrowed::HasDisplayHandle`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `egui_glow` (lib) due to 1 previous error
```

YgorSouza:
> Update: It seems like everything works except one error which I couldn't find where it comes from:
> 

This bug is also on master. egui_glow is missing the `winit/rwh_06` feature, so it does not build correctly on its own. It builds as part of eframe, because eframe enables the feature.

https://github.com/emilk/egui/blob/d856f7b3a58fc65a81ff9657f5c6e1ab462b7bb6/crates/eframe/Cargo.toml#L148

It should be just a matter of adding it to egui_glow as well, but I'm not sure under which conditions it should be enabled.

bircni:
> It should be just a matter of adding it to egui_glow as well, but I'm not sure under which conditions it should be enabled.

@YgorSouza can confirm  - added winit to egui-glow and it works properly!

crates/eframe/src/native/winit_integration.rs
```diff
@@ -129,7 +129,7 @@ pub fn system_theme(window: &Window, options: &crate::NativeOptions) -> Option<c
 }
 
 #[cfg(feature = "accesskit")]
-pub(crate) fn on_accesskit_window_event(
+pub fn on_accesskit_window_event(
```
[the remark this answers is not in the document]
bircni: Gonna close this -> only was a POC
Clippy told that a pub(crate) is unnecessary here

PR #4915, opened 2024-08-04
iOS safe area

frederik-uni:
Before:
Screenshot 2024-08-04 at 23 24 07
After:
Screenshot 2024-08-04 at 23 22 51

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1311,6 +1311,7 @@ dependencies = [
  "image",
  "log",
  "mime_guess2",
+ "objc",
  "puffin",
  "resvg",
  "serde",
--- a/crates/egui_extras/Cargo.toml
+++ b/crates/egui_extras/Cargo.toml
@@ -66,6 +66,7 @@ svg = ["resvg"]
 ## Enable better syntax highlighting using [`syntect`](https://docs.rs/syntect).
 syntect = ["dep:syntect"]
 
+ios-safe-area = ["dep:objc"]
 
 [dependencies]
 egui = { workspace = true, default-features = false }
@@ -100,6 +101,9 @@ puffin = { workspace = true, optional = true }
 syntect = { version = "5", optional = true, default-features = false, features = [
   "default-fancy",
 ] }
+
+# ios
+objc = { version = "0.2", optional = true }
 
 # svg feature
 resvg = { version = "0.37", optional = true, default-features = false }
--- a/crates/egui_extras/src/lib.rs
+++ b/crates/egui_extras/src/lib.rs
@@ -79,3 +79,5 @@ macro_rules! log_or_panic {
     }};
 }
 pub(crate) use log_or_panic;
+#[cfg(feature = "ios-safe-area")]
+pub mod safe_area;
--- a/crates/egui_extras/src/safe_area.rs
+++ b/crates/egui_extras/src/safe_area.rs
@@ -1,0 +1,39 @@
+use egui::{vec2, Sense, Widget};
+use objc::runtime::Object;
+use objc::{class, msg_send, sel, sel_impl};
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct UIEdgeInsets {
+    top: f64,
+    left: f64,
+    bottom: f64,
+    right: f64,
+}
+
+#[allow(unsafe_code)]
+/// Gets the ios safe area
+/// A safe area defines the area within a view that isn’t covered by a navigation bar, tab bar, toolbar, or other views a window might provide. Safe areas are essential for avoiding a device’s interactive and display features, like Dynamic Island on iPhone or the camera housing on some Mac models. For developer guidance, see Positioning content relative to the safe area.
+pub fn get_ios_safe_area_insets() -> UIEdgeInsets {
+    unsafe {
+        let shared_application: *mut Object = msg_send![class!(UIApplication), sharedApplication];
+        let windows: *mut Object = msg_send![shared_application, windows];
+        let first_object: *mut Object = msg_send![windows, firstObject];
+        let safe_area_insets: UIEdgeInsets = msg_send![first_object, safeAreaInsets];
+        safe_area_insets
+    }
+}
+
+/// Adds spacing at the top on ios devices
+#[derive(Default)]
+pub struct IosSafeArea;
+
+impl Widget for IosSafeArea {
+    fn ui(self, ui: &mut egui::Ui) -> egui::Response {
+        let (id, rect) = ui.allocate_space(vec2(
+            ui.available_width(),
+            get_ios_safe_area_insets().top as f32,
+        ));
+        ui.interact(rect, id, Sense::hover())
+    }
+}


emilk:
I think this should be part of `eframe`, i.e. so you can call `eframe::safe_area_top(ctx)` to get how many ui points from the top it is safe to put widgets.

frederik-uni:
@emilk so, I move the safe area insets function to eframe, but what about the safe area widget. I think it would be better to use something like this instead of a widget:

```rs
fn safe_area<R>(ui: &mut Ui, add_contents: impl FnOnce(&mut Ui) -> R) -> R {
    let insets = get_ios_safe_area_insets();
    if let Some(insets) = insets {
        let next = insets.top - ui.cursor().max.y as f64;
        if next > 0. {
            let (id, rect) = ui.allocate_space(vec2(ui.available_width(), insets.top as f32));
            ui.interact(rect, id, Sense::hover());
        }
        ui.set_max_height(ui.available_height() - insets.bottom);
    }
    add_contents(ui)
}
```

where should I put it? should this be part of egui/egui_extras. should this & the safe area insets function be hidden behind a feature or should it be enabled for every iOS build? its kinda essential and probably needed in every iOS build anyway.

emilk:
One approach would be to modify `FrameState::available_rect` so that is is the safe area. That would mean `egui::CentralPanel` (and other panels) would start within that smaller rect. However, one should still paint some sort of background color outside that.

lucasmerlin:
Seems winit will soon have a safe area api: https://github.com/rust-windowing/winit/pull/3890
Maybe we can use that one instead? Then other platforms will also be supported once support is added to winit (https://github.com/rust-windowing/winit/issues/3910)

frederik-uni:
Im not quite sure when which events fires in winit. I guessed. theoretically WindowEvent::Resized(_)  & ScaleFactorChanged should be enough, but im not sure if the events fire when the app is minimized or repopend in split screen or if the device was flipped while the screen was of so included Focused(true) & Occluded(false).

Im not quite sure when the safe area gets merged into winit as it had no updates the last 3 weeks. Migrating from my implementation to the winit implementation should be quite easy anyway(`self.egui_input_mut().safe_area = Some(egui::SafeArea::from(safe_area::get_ios_safe_area_insets()));` would need to be changed to `self.egui_input_mut().safe_area = Some(egui::SafeArea::from(window.get_safe_area()))`

lucasmerlin:
The winit PR was approved and is added to the 0.31 milestone so I think it makes sense to just wait for that and then update your PR with the winit implementation

frederik-uni:
@emilk the code was merged so once the version bump is in crates.io & this project is updated, this can be merged without change

hacknus:
Cool, Can this be merged now?

lucasmerlin:
Sorry, I didn't expect the 0.31 release to take so long. I'm fine with merging the objc approach. Will look into it next week.

lucasmerlin:
I've merged master and reverted the commit relying on winit's safe area implementation, and tested it with my [egui ios demo](https://github.com/lucasmerlin/egui-apple-example), works well!

https://github.com/user-attachments/assets/decf91ca-d33f-40d1-8ad2-7ee19c5baafe

Only thing I noticed is that windows and popups can extend past the safe area, which I don't think it should. It could mean e.g. a combobox popup is partially hidden behind the bottom bar thingy.

crates/egui/src/pass_state.rs
```diff
@@ -258,7 +260,11 @@ impl Default for PassState {
 }
 
 impl PassState {
-    pub(crate) fn begin_pass(&mut self, screen_rect: Rect) {
+    pub(crate) fn begin_pass(&mut self, screen_rect: Rect, safe_area: SafeArea) {
+        let screen_rect = Rect::from_min_max(
+            screen_rect.min + emath::vec2(safe_area.left, safe_area.top),
+            screen_rect.max - emath::vec2(safe_area.right, safe_area.bottom),
+        );
```
emilk, line 267: Can we just pass in a pre-shrunk `screen-rect` here instead? That should then fix popups ending up behind the "safe area"

lucasmerlin:
https://github.com/user-attachments/assets/d21246c1-b5e0-4c1e-800d-523fe9960909

Works really well now!

I've updated the way safe area is handled, it now maches screen_rect in that it's only updated if it's `Some`. Also I made the InputState::screen_rect field private and added a getter fn that respects the safe area and added screen_rect_including_unsafe_area which returns the unmodified screen rect. Not sure on the naming but I couldn't come up with something nicer.

crates/egui/src/pass_state.rs
lucasmerlin: Ah, I think my implementation is a bit different from what you had in mind, but I think it makes sense to store the safe area and originial screen_rect in egui somewhere, since in some apps you might want to know both so you can draw some widgets extending past the safe area.

crates/egui/src/data/input.rs
```diff
@@ -7,6 +7,27 @@ use crate::{
...
 };
 
+/// Safe area insets of the screen. I.e., area not covered by the status bar, navigation bar, etc.
+#[derive(Debug, PartialEq, Copy, Clone, Default)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub struct SafeArea {
+    pub top: f32,
+    pub right: f32,
+    pub bottom: f32,
+    pub left: f32,
```
emilk, line 17: This needs better docstring.

Does the struct describe the bounds of the safe area? If so, why isn't it a `Rect`?

Or is this describing the outer margins by which you need to shrink in order to make the screen rect safe? If so, can  we use `MarginF32` ?

crates/egui/src/input_state/mod.rs
```diff
@@ -447,11 +460,30 @@ impl InputState {
...
     }
 
+    /// Which part of the screen does egui cover?
+    ///
+    /// Returns the `RawInput::screen_rect` - [`Self::safe_area`].
+    /// To get the full `screen_rect`, use [`Self::screen_rect_including_unsafe_area`]
     #[inline(always)]
     pub fn screen_rect(&self) -> Rect {
+        self.screen_rect - self.safe_area
+    }
```
emilk, line 470: That `self.screen_rect != self.screen_rect()` is too confusing. Use different words for different things please

lucasmerlin: I think there are three options but none are great:
1.  keep `screen_rect` and have it return the "safe area rect"
	  - needs some new name for the rect that includes the "unsafe area"
	    - which is hard because the screen is the most outer rect of a computer. Maybe `outer_screen_rect`? 😕
	  - no breaking changes for egui apps, since you'd want to use `screen_rect` in most places

2. deprecate `screen_rect` and introduce two new terms 
  	- e.g. `outer_rect` and `inner_rect` or `viewport_rect` and `content_rect`
 
3. keep `screen_rect`, having it return the "unsafe outer rect"
	  - inner rect could be `content_rect` or `inner_rect`
	  - naming would make sense
	  - no breaking changes _but_ most existing code using `screen_rect` would probably be wrong without any warning, since it should be using the screen_rect when it should be using `content_rect`
	  
I think I prefer option 2 since it seems like the most "correct" thing to do

irh:
@lucasmerlin I've gone ahead and tried to address the open points of feedback on a branch here: https://github.com/irh/egui/tree/ios-safe-area. 

It would be great if we could get this landed, should I open a new PR or would you prefer to pull my changes into this one?

lucasmerlin:
@irh Nice! You should create a new one, I don't think you can update this one

irh:
> @irh Nice! You should create a new one, I don't think you can update this one

@lucasmerlin Ok great, new PR here: #7578.

PR #4930, opened 2024-08-06
Allow for synchronizing between two `Context`s by serializing their entire contents

DouglasDwyer:
## What this PR does

This PR implements the ability to serialize/deserialize (almost) the entirety of a `Context` via a new feature called `snapshot`. This includes the `FontDefinitions`, `Memory`, `Style`, `ViewportState`s, and more. This allows for *synchronizing* state between two separate `Context` objects, so that one `Context` mirrors the other. It also introduces a way to incrementally track when a `Context`'s state changes, so that big things like `FontDefinitions` don't need to be serialized every time.

**Note that this is *different* than the `persistence` feature, which only allows serializing a subset of `Memory`. With `snapshot`, everything that can be serialized from a `Context` will be. The generated `ContextSnapshot` can then be deserialized and applied to a different `Context` to make the internal states match.**

## Why I want this

I am building a WASM-based modding system for my game engine. I would like a way for WASM plugins to draw UI, and `egui` seems like a great choice. Because multiple WASM modules and the host game engine may all be drawing UI elements, they need to share an `egui::Context`. That's impossible, since each WASM module has a separate address space. To circumvent this, whenever a WASM plugin needs to draw UI, I serialize the entirety of the `egui::Context` into WASM memory from the host. When the WASM module is done, the `egui::Context` is copied back to the host. You can see [a working example here](https://github.com/DouglasDwyer/egui_wings).

## High-level changes

- Created a `snapshot` features with allows for serializing/deserializing `Context`s when enabled.
- Several new methods are added to `Context` for this purpose - `snapshot_for`, which generates an object that can be serialized, and `apply_snapshot`, which allows for consuming the deserialized result.
- A new private `snapshot` module is added inside the `context` module. This module implements a fair amount of custom serialization/deserialization logic. This custom logic is necessary to override some of the serialization behavior that the `persistence` feature uses - for example, the `persistence` feature will `serde(skip)` serializing an `Area`'s size, but here the size should be serialized.
- `#[derive(serde::Deserialize, serde::Serialize)]` has been added to some extra internal structures so that they may be serialized.
- Things that are not serialized within a context include plugins and `Memory::data`, since those use type erasure and serialization is impossible.

## Comments

I want to start off by saying that `egui` is very well-designed! It took only a few days to get WASM-egui integration working; the fact that this is possible speaks to the simplicity and robustness of the `egui` API.

That said, the serialization implementation here is somewhat specific to my needs. If this is too much to merge, then I would be happy to explore alternative solutions. Ultimately, I could keep the context serialization logic in my own crate - but I would need a way to access some of `egui`'s remaining internal state, including:

- The raw hashmaps of `Vec<ClippedShape>` that `GraphicLayers` internally stores
- The `interact_widgets`, `hits`, and other crate-private members of `ViewportState`
- The map of `ViewportState`s stored within a `Context`
- and so on

I chose to make this PR initially - just implementing the snapshot logic directly in `egui` - so that this internal state wouldn't need to be exposed. Either way, I would greatly appreciate if we could find a way to support my use case. It would be a great benefit to the Rust game engine community :)

### DIFF
--- a/crates/egui/Cargo.toml
+++ b/crates/egui/Cargo.toml
@@ -74,6 +74,9 @@ rayon = ["epaint/rayon"]
 ## Allow serialization using [`serde`](https://docs.rs/serde).
 serde = ["dep:serde", "epaint/serde", "accesskit?/serde"]
 
+## Allow creating instant, serializable snapshots of a context
+snapshot = ["serde", "epaint/serde", "ron"]
+
 ## Change Vertex layout to be compatible with unity
 unity = ["epaint/unity"]
 
--- a/crates/egui/src/context/mod.rs
+++ b/crates/egui/src/context/mod.rs
@@ -7,6 +7,9 @@ use epaint::{
     emath::TSTransform, mutex::*, stats::*, text::Fonts, util::OrderedFloat, TessellationOptions, *,
 };
 
+#[cfg(feature = "snapshot")]
+use snapshot::{MemorySnapshot, OptionsSnapshot, ViewportStateSnapshot};
+
 use crate::{
     animation_manager::AnimationManager,
     data::output::PlatformOutput,
@@ -23,6 +26,12 @@ use crate::{
 };
 
 use self::{hit_test::WidgetHits, interaction::InteractionSnapshot};
+
+#[cfg(feature = "snapshot")]
+pub use self::snapshot::{ContextShapshotBorrow, ContextSnapshot, ContextSnapshotDeltas};
+
+#[cfg(feature = "snapshot")]
+mod snapshot;
 
 /// Information given to the backend about when it is time to repaint the ui.
 ///
@@ -397,11 +406,21 @@ struct ContextImpl {
     #[cfg(feature = "accesskit")]
     is_accesskit_enabled: bool,
 
+    /// Tracks the changes that have occurred to a context to allow
+    /// for partial synchronization with [`ContextSnapshot`]s.
+    #[cfg(feature = "snapshot")]
+    snapshot_deltas: ContextSnapshotDeltas,
+
     loaders: Arc<Loaders>,
 }
 
 impl ContextImpl {
     fn begin_frame_mut(&mut self, mut new_raw_input: RawInput) {
+        #[cfg(feature = "snapshot")]
+        {
+            self.snapshot_deltas.frame_count = self.snapshot_deltas.frame_count.wrapping_add(1);
+        }
+
         let viewport_id = new_raw_input.viewport_id;
         let parent_id = new_raw_input
             .viewports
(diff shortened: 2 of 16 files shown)


geolehmann:
Woah, cool! This is something I was looking into too. I ended up with drawing the egui UI from a WASM plugin in a separate canvas with transparent background above the main canvas, and using some JS functions to exchange state etc. - but your solution looks really great.

juancampa:
Interesting approach!

I've been also looking into something like this myself. I want to allow hot-reloading of parts of the UI to speed up the iteration cycle of [our IDE](https://membrane.io).

I have a branch where two WASM modules share memory by [making the WASM memory a `SharedArrayBuffer`](https://rustwasm.github.io/docs/wasm-bindgen/examples/raytrace.html). One is the "host", running an `eframe` app, and the second one is the "plugin" which only knows about `egui`.

My plan was to 🪄 _somehow_ ✨ pass a `*mut Ui` from the host to the plugin so that it would render stuff, interact, etc. as if the plugin was just another function. In theory it should work if both modules end up with the same memory layout for all structs, which admittedly is a big if.

For this to work the "plugin" would have to be compiled with "Position Independent Code" as described in this [Dynamic Linking writeup](https://github.com/WebAssembly/tool-conventions/blob/main/DynamicLinking.md), and then the host would have to reserve enough memory for the plugin to use, hold all the static data, etc.

It was way more complicated than I originally thought so I've put it in the back burner for the time being.

Anyway, sorry for the long rant.

I do have a question though, does it support interacting with the UI? How does it do it?

DouglasDwyer:
> I do have a question though, does it support interacting with the UI? How does it do it?

For my use case, I have a host (a native app) and a guest (the WASM plugin). The host handles user input with the `egui` context the normal way. Since the WASM plugin "shares" the context with the host, it is able to support user interactions automatically, by virtue of the host supplying/updating the `InputState` each frame.

geolehmann:
Under [egui_wasm_plugins](https://github.com/geolehmann/egui_wasm_plugins) I try to make a minimal template for running egui on the web using the PR with the host (under **src/main.rs**) providing the context and the plugin (under **plugin/src/lib.rs**) drawing some Ui elements. Currently I encounter a strange error which I try to debug:

```
Error occurred:
InvalidModule("Module intrinsic __wings_alloc_marshal_buffer had invalid signature: FuncType { params: [I32], results: [I32], name: Some(\"__wings_alloc_marshal_buffer\") }  and FuncType { params: [I32], results: [I32], name: None }")
```

I think it has to do with the fact that due to the wasm32-unknown-unknown target I had to exchange the **wasm_runtime_layer** runtime with **js_wasm_runtime_layer** in the host...

DouglasDwyer:
@geolehmann ah, looks like you found a tiny bug in one of my other crates (`wasm_runtime_layer`). It was including the debug string `name` when comparing equality of `FuncType`s, which shouldn't be the case. I have fixed it by releasing `wasm_runtime_layer v0.4.1`; if you have further issues using any of those libraries please open a PR on the relevant repository.

emilk:
> the serialization implementation here is somewhat specific to my needs. If this is too much to merge, then I would be happy to explore alternative solutions.

Unfortunately I think this is too much to merge. It is a lot of new code to maintain, and will slow down further internal changes to egui.

> Ultimately, I could keep the context serialization logic in my own crate - but I would need a way to access some of egui's remaining internal state

I think this is a better approach, to make more of the internals of egui public. Feel free to open a PR adding `pub` to stuff.

DouglasDwyer:
No problem, I totally get it! I'll work on the other approach :)

PR #4933, opened 2024-08-07
`egui_extras`: Fix file mime from path (wrong feature name)

rustbasic:
Fix: features "mime_guess" to "file"

typo: 1

Is this what you intended?
If you intended to add `mime_guess` to `cargo.toml`, please drop it.


emilk (approved):
Nice

PR #4934, opened 2024-08-07
Refactor: use `if let` instead of `for` on `Option`s

rustbasic:
This is recommended by `rust-analyzer`

It might be a good idea to fix this to appease `rust-analyzer`.


bircni (approved):
Looks good!

emilk (commented):
Since these are `Option`s, an `if let` would make more sense

Issue #4895, opened 2024-07-31
touchscreen click events are not detected properly (on Linux)


girtsf:
**Describe the bug**

With any widget, such as a `Button` or  `Label`, `.clicked()` events are not recognized when the screen is tapped once. Instead, a single tap triggers `long_touched()` after a delay. Tapping twice in a quick succession generates a `clicked()` followed by a `long_touched()`.

**To Reproduce**

`Cargo.toml`:

```
[package]
name = "repro"
version = "0.1.0"
edition = "2021"

[dependencies]
egui = { git = "https://github.com/emilk/egui", branch = "master" }
eframe = { git = "https://github.com/emilk/egui", branch = "master" }
```

`src/main.rs`:

```
#[derive(Default)]
struct App {
    log: String,
}

impl eframe::App for App {
    fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
        egui::CentralPanel::default().show(ctx, |ui| {
            let button = ui.button("click me");
            if button.clicked() {
                self.log.push_str("clicked()\n");
            }
            if button.long_touched() {
                self.log.push_str("long_touched()\n");
            }
            ui.label(&self.log);
        });
    }
}

fn main() {
    eframe::run_native(
        "test",
        eframe::NativeOptions::default(),
        Box::new(|_cc| Ok(Box::new(App::default()))),
    )
    .unwrap();
}
```

1. `cargo run` the above code
2. tap on "click me" button

**Expected behavior**
`clicked()` gets shown on the screen.

**Screenshots**
![Screenshot from 2024-07-31 14-48-42](https://github.com/user-attachments/assets/c9fe5eb0-071d-4063-9287-fffe384b3098)

This can also be reproduced by `egui_demo_app`, built from latest `master` (d856f7b3a58fc65a81ff9657f5c6e1ab462b7bb6), by opening "Input Test" and tapping on "Sense::click" label:

![Screenshot from 2024-07-31 14-53-00](https://github.com/user-attachments/assets/b0723de5-e6b1-438e-b85d-f673480d5595)

**Desktop:**
 - OS: Ubuntu 24.04
 - Device:  Lenovo Yoga 7 2-in-1 14IML9
 - Wayland

(I tested on Windows, on a slightly different touchscreen-equipped Lenovo laptop, and it works correctly there.)

**Additional context**

This might be related to the observations found in #4668 ? 

[...]

PR #4939, opened 2024-08-08
Unpin & upgrade winit to 0.30.5

ArthurBrussee:
This updates winit to 0.30.5. 

https://github.com/emilk/egui/pull/4849 Had to pin the version to 0.30.2, as a Winit patch changed the behavior of selecting a theme. Winit 0.30.5 reverts this, so we could stick with `window.theme()`, but the newly added `ActiveEventLoop::system_theme` is more like what egui wants anyway, as individual windows can have theme overrides.

Also bump `smithay-clipboard` to prevent some now duplicate dependencies.


girtsf:
Tested and can confirm that this fixes touchscreen release issues (#4895).

emilk (approved):
Awesome!

PR #4942, opened 2024-08-09
Add `Options::input_options` for click-delay etc

girtsf:
This takes 3 hardcoded constants from `input_state.rs` and puts them in a `InputOptions` struct that then gets added to `Options`. This allows adjusting these values at runtime, for example, to increase `MAX_CLICK_DIST` for touchscreen usage.

### DIFF
--- a/crates/egui/src/input_state/mod.rs
+++ b/crates/egui/src/input_state/mod.rs
@@ -17,19 +17,77 @@ pub use crate::Key;
 pub use touch_state::MultiTouchInfo;
 use touch_state::TouchState;
 
-/// If the pointer moves more than this, it won't become a click (but it is still a drag)
-const MAX_CLICK_DIST: f32 = 6.0; // TODO(emilk): move to settings
+/// Options for input state handling.
+#[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub struct InputOptions {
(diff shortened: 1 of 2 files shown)


girtsf:
Background on the PR: I'm working on a UI for a touchscreen device and I've found that the default `MAX_CLICK_DIST` of 6 pixels doesn't work well - most finger taps will "slide" on the screen more than 6 pixels and thus won't register as clicks. Setting the value to a higher number makes the interface less finicky.

Thoughts on this? I don't quite like that I had to stash `InputOptions` in `PointerState` and `InputState` so that functions such as `wants_repaint_after` and `is_long_press` can do their calculations. Better ideas? Or is this acceptable?

(Maybe outside the scope of this PR, but maybe we could consider separate config values for mouse vs touch? The latter is much less precise. From a quick glance, this would require changing some of the plumbing that translates touch events into `PointerButton` events.)

crates/egui/src/input_state.rs
```diff
@@ -794,6 +859,9 @@ pub struct PointerState {
 
     /// All button events that occurred this frame
     pub(crate) pointer_events: Vec<PointerEvent>,
+
+    /// Input state management configuration.
+    input_options: InputOptions,
```
emilk, line 864: We should note here that this is copied from `egui::Options` at the start of each frame, and is only here for convenience

emilk (commented):
I like this approach better than the alternative in https://github.com/emilk/egui/pull/4986

emilk (approved):

Issue #4776, opened 2024-07-03
1-pixel separator lines are feathered


valadaptive:
**Describe the bug**

Separator lines which are 1 pixel wide, such as those used by default in panels and the `Separator` widget, are rendered with their *centers* at a pixel coordinate rather than their *edges* at a pixel coordinate. This leads to them appearing as 2-pixel translucent lines rather than 1-pixel opaque lines.

This was discussed somewhat in https://github.com/emilk/egui/issues/1322, but I think it's worth revisiting. Right now, 1-pixel lines are very heavily used in egui, and this behavior can lead to some unintuitive and visually unappealing results (for instance, this contributes to https://github.com/emilk/egui/issues/4773; the 1-pixel border of the panel feathers upwards, merging with the vertical separator above). If a 2-pixel transparent border is more aesthetically appealing, it should be explicitly represented as such for layout purposes.

I think this lies more on the "bug" side than the "feature" side because the code that draws these separator lines, for both separators and panels, *does* call `round_to_pixel`--it's just that the rounding value is incorrect for lines with odd widths.

**To Reproduce**

This can be seen in any egui application that uses panels or separators. For instance, all of these 2-pixel medium-grey outlines in the demo app are *supposed* to be 1-pixel light grey outlines:
![image](https://github.com/emilk/egui/assets/79560998/a448b7cf-59ab-4d77-a715-9cbd46a52561)

**Expected behavior**

Lines which are 1 pixel (as a unit) wide should, at 1 pixel per point, be 1 physical pixel wide. It's not as simple as always rounding up or down-- for e.g. a border around a rectangle, you may want to round "outwards" or "inwards", which rounds differently depending on which side of the rectangle you're on. It may be worth looking into how CSS does this (which ties into https://github.com/emilk/egui/issues/4019).

juancampa:
I spent some time on this today with promising results. Hoping to have a PR ready soon.

Issue #4019, opened 2024-02-10
CSS-like `Border`


emilk:
* Part of https://github.com/emilk/egui/issues/3284

As a first step towards more powerful styling, I want to change how `Frame` works to be slightly more like the CSS box model.

In particular, the frame border stroke width should be counted as part of the width of the frame. This should let us get rid of the ugly `clip_rect_margin`.

The plan is to use the new `Frame` for essentially all widgets (except maybe `Label`). So a `Button` would ask the `Style` for a `Frame`, then use that for both its sizing and color calculations.

Here is the proposed outline

```rs
/// A frame around some content, including margin and border colors.
/// 
/// The total (outer) size of a frame is
/// `content_size + inner_margin + border.stroke.width + outer_margin`.
/// 
/// Everything within the border stroke is filled with the border fill color (if any).
/// 
/// ```
/// +---------------^-------------------------------+
/// |               | outer_margin                  |
/// |   +-----------v----^-----------------------+  |
/// |   |                | border stroke width   |  |
/// |   |  +-------------v---^----------------+  |  |
/// |   |  |                 | inner_margin   |  |  |
/// |   |  |  +--------------v-------------+  |  |  |
/// |   |  |  |                            |  |  |  |
/// |   |  |  |                            |  |  |  |
/// |   |  |  | content rect               |  |  |  |
/// |   |  |  +----------------------------+  |  |  |
/// |   |  |                                  |  |  |
/// |   |  +----------------------------------+  |  |
/// |   |       widget rect                      |  |
/// |   +----------------------------------------+  |
/// |           outer rect                          |
/// +-----------------------------------------------+
/// ```
/// 
/// `widget rect` is the interactive part of the widget (what sense clicks etc),
/// and is what is returned by [`Response::rect`].
struct Frame {
    /// Inner spacing. Called `padding` in CSS.
    pub inner_margin: Margin,

    /// Stroke, fill color etc.
    ///
    /// The width of the border is counted as part of the size of the frame.
    pub border: Border,

    /// Optional drop-shadow.
    /// Does not affect margins or sizes.
    pub shadow: Shadow,

    /// Outer spacing. Called `margin` in CSS.
    /// 
    /// This does NOT do "Margin Collapse" like in CSS,
    /// i.e. when placing two frames next to each other,
    /// the distance between their borders is the SUM
    /// of their other margin.
    /// In CSS the distance would be the MAX of their outer margins.
    ///
    /// Supporting margin collapse is difficult, and would 
[...]

Issue #3284, opened 2023-08-28
CSS-like styling


emilk:
Some half-finished ideas around how to improve the styling and theming story for egui.

# Background

Styling for egui is currently supplied by `egui::Style` which controls spacing, colors, etc for the whole of egui. There is no convenient way of changing the syling of a portion of the UI, except for changing out or modifying the `Style` temporarily, and then changing it back.

We would like to have a system that can support CSS-like selectors, so that users can easily style their ui based on the *Style Modifiers* (see below):

It would be very beneficial if such styling could be set in a single text file and live-loaded.

# Proposal

## Style modifiers

Here are some things that could influence the style of a widget:

- widget type (button, slider, label, …)
- interact state (disabled, inactive, active, hovered, active)
- text modifier (header, small, weak, strong, code, …)
- per-`Ui` identifier (”settings_panel”)
- per-widget identifier (”exit_button”)

For instance, a user may want to change the sizes of all buttons within the `"settings_panel"`.

The per-`Ui` identifier would need to be a hierarchial stack, so the query to a theme would be something like:

> Give me the WidgetStyle for a ************button************ that is ******hovered****** that is nested in a **“options”→”internals”**
> 

We could also consider having dark/light mode as a modifier, allowing users to specify both variants in one theme file.

## WidgetStyle

Let’s start with this:

```rust
pub struct WidgetStyle {
    /// Background color, stroke, margin, and shadow.
    pub frame: Frame,
  
    /// What font to use and at what size.
    pub text: TextStyle,
  
    /// Color and width of e.g. checkbox checkmark.
    /// Also text color.
    pub stroke: Stroke,
}

pub struct TextStyle {
    pub font: FontId,
    pub underlined: bool,
    …
}
```

If each widget as given a `WidgetStyle` it could then use it both for sizing (`frame` margins and font size) and its visual styling. The current theme would select a `WidgetStyle` based on some given style modifiers.

Because the `WidgetStyle` affects the widget size, we need to know the `WidgetStyle` *before* we place it, and therefore before we know wether or not the user is interacting with it. This can be solved by querying for the style *twice* (and ignoring the margins and font size the second time, assuming them to be the same). This can be optimized out for the common case where the widget is not hover or bneeing interacted with.

`WidgetStyle` would be used by all built-in widgets (button, checkbox, slider, …) but also each `Window` and `Ui`.

## Example

```rust
fn button_ui(ui: &mut Ui, text: &str) {
    let style = ui.style_of_interactive("button");
[...]

PR #4943, opened 2024-08-10
Fix blurry lines

juancampa:
* Closes <https://github.com/emilk/egui/issues/4776>

I've been meaning to look into this for a while but finally bit the bullet this week. Contrary to what I initially thought, the problem of blurry lines is unrelated to feathering because it also happens with feathering disabled.

The root cause is that lines tend to land on pixel boundaries, and because of that, frequently used strokes (e.g. 1pt), end up partially covering pixels. This is especially noticeable on 1ppp displays.

There were a couple of things to fix, namely: individual lines like separators and indents but also shape strokes (e.g. Frame).

Lines were easy, I just made sure we round them to the nearest pixel _center_, instead of the nearest pixel boundary.

Strokes were a little more complicated. To illustrate why, here’s an example: if we're rendering a 5x5 rect (black fill, red stroke), we would expect to see something like this:

Screenshot 2024-08-11 at 15 01 41

The fill and the stroke to cover entire pixels. Instead, egui was painting the stroke partially inside and partially outside, centered around the shape’s path (blue line):

Screenshot 2024-08-11 at 15 00 57

Both methods are valid for different use-cases but the first one is what  we’d typically want for UIs to feel crisp and pixel perfect. It's also how CSS borders work (related to #4019 and #3284).

Luckily, we can use the normal computed for each `PathPoint` to adjust the location of the stroke to be outside, inside, or in the middle. These also are the 3 types of strokes available in tools like Photoshop.

This PR introduces an enum `StrokeKind` which determines if a `PathStroke` should be tessellated outside, inside, or _on_ the path itself. Where "outside" is defined by the directions normals point to.

Tessellator will now use `StrokeKind::Outside` for closed shapes like rect, ellipse, etc. And `StrokeKind::Middle` for the rest since there's no meaningful "outside" concept for open paths. This PR doesn't expose `StrokeKind` to user-land, but we can implement that later so that users can render shapes and decide where to place the stroke.

### Strokes test
(blue lines represent the size of the rect being rendered)

`Stroke::Middle` (current behavior, 1px and 3px are blurry)
Screenshot 2024-08-09 at 23 55 48

`Stroke::Outside` (proposed default behavior for closed paths)
Screenshot 2024-08-09 at 23 51 55

`Stroke::Inside` (for completeness but unused at the moment)
Screenshot 2024-08-09 at 23 54 49

### Demo App
The best way to review this PR is to run the demo on a 1ppp display, especially to test hover effects. Everything should look crisper. Also run it in a higher dpi screen to test that nothing broke 🙏.

Before:
egui_old

After (notice the sharper lines):
egui_new

### DIFF
--- a/crates/egui/src/containers/frame.rs
+++ b/crates/egui/src/containers/frame.rs
@@ -89,7 +89,6 @@ impl Frame {
         Self {
             inner_margin: Margin::symmetric(8.0, 2.0),
             fill: style.visuals.panel_fill,
-            // outer_margin: Margin::same(1.0),
             ..Default::default()
         }
     }
--- a/crates/egui/src/containers/frame.rs
+++ b/crates/egui/src/containers/frame.rs
@@ -89,6 +89,7 @@ impl Frame {
         Self {
             inner_margin: Margin::symmetric(8.0, 2.0),
             fill: style.visuals.panel_fill,
+            // outer_margin: Margin::same(1.0),
             ..Default::default()
         }
     }
--- a/crates/egui/src/containers/panel.rs
+++ b/crates/egui/src/containers/panel.rs
@@ -342,7 +342,7 @@ impl SidePanel {
             // In the meantime: nudge the line so its inside the panel, so it won't be covered by neighboring panel
             // (hence the shrink).
             let resize_x = side.opposite().side_x(rect.shrink(1.0));
-            let resize_x = ui.painter().round_to_pixel(resize_x);
+            let resize_x = ui.painter().round_to_pixel_center(resize_x);
             ui.painter().vline(resize_x, panel_rect.y_range(), stroke);
         }
 
@@ -831,7 +831,7 @@ impl TopBottomPanel {
             // In the meantime: nudge the line so its inside the panel, so it won't be covered by neighboring panel
             // (hence the shrink).
             let resize_y = side.opposite().side_y(rect.shrink(1.0));
-            let resize_y = ui.painter().round_to_pixel(resize_y);
+            let resize_y = ui.painter().round_to_pixel_center(resize_y);
             ui.painter().hline(panel_rect.x_range(), resize_y, stroke);
         }
 
--- a/crates/egui/src/containers/panel.rs
+++ b/crates/egui/src/containers/panel.rs
@@ -325,6 +325,9 @@ impl SidePanel {
             ui.ctx().set_cursor_icon(cursor_icon);
         }
 
+        // Keep this rect snapped so that panel content can be pixel-perfect
+        let rect = ui.painter().round_rect_to_pixels(rect);
+
         PanelState { rect }.store(ui.ctx(), id);
 
         {
@@ -339,10 +342,12 @@ impl SidePanel {
                 Stroke::NONE
             };
             // TODO(emilk): draw line on top of all panels in this ui when https://github.com/emilk/egui/issues/1516 is done
-            // In the meantime: nudge the line so its inside the panel, so it won't be covered by neighboring panel
-            // (hence the shrink).
-            let resize_x = side.opposite().side_x(rect.shrink(1.0));
+            let resize_x = side.opposite().side_x(rect);
             let resize_x = ui.painter().round_to_pixel_center(resize_x);
+
+            // We want the line exactly on the last pixel but rust rounds away from zero so we bring it back a bit for
+            // left-side panels
+            let resize_x = resize_x - if side == Side::Left { 1.0 } else { 0.0 };
             ui.painter().vline(resize_x, panel_rect.y_range(), stroke);
         }
 
@@ -814,6 +819,9 @@ impl TopBottomPanel {
             ui.ctx().set_cursor_icon(cursor_icon);
         }
 
+        // Keep this rect snapped so that panel content can be pixel-perfect
+        let rect = ui.painter().round_rect_to_pixels(rect);
+
         PanelState { rect }.store(ui.ctx(), id);
 
         {
@@ -828,10 +836,12 @@ impl TopBottomPanel {
                 Stroke::NONE
             };
             // TODO(emilk): draw line on top of all panels in this ui when https://github.com/emilk/egui/issues/1516 is done
-            // In the meantime: nudge the line so its inside the panel, so it won't be covered by neighboring panel
-            // (hence the shrink).
-            let resize_y = side.opposite().side_y(rect.shrink(1.0));
+            let resize_y = side.opposite().side_y(rect);
             let resize_y = ui.painter().round_to_pixel_center(resize_y);
+
+            // We want the line exactly on the last pixel but rust rounds away from zero so we bring it back a bit for
+            // top-side panels
+            let resize_y = resize_y - if side == TopBottomSide::Top { 1.0 } else { 0.0 };
             ui.painter().hline(panel_rect.x_range(), resize_y, stroke);
         }
 
--- a/crates/egui/src/containers/panel.rs
+++ b/crates/egui/src/containers/panel.rs
@@ -343,6 +343,8 @@ impl SidePanel {
             };
             // TODO(emilk): draw line on top of all panels in this ui when https://github.com/emilk/egui/issues/1516 is done
             let resize_x = side.opposite().side_x(rect);
+
+            // This makes it pixel-perfect for odd-sized strokes (width=1.0, width=3.0, etc)
             let resize_x = ui.painter().round_to_pixel_center(resize_x);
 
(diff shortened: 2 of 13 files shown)


juancampa:
The main challenge I'm seeing at the moment is that if the rasterization of the fill doesn't exactly match the rasterization of the stroke, you'll see a gap between the two 🤔. This only happens when rounding is enabled afaict. 

Screenshot 2024-08-10 at 00 14 58

crates/egui/src/containers/window.rs
```diff
@@ -436,9 +436,6 @@ impl<'open> Window<'open> {
         let mut window_frame = frame.unwrap_or_else(|| Frame::window(&ctx.style()));
         // Keep the original inner margin for later use
         let window_margin = window_frame.inner_margin;
-        let border_padding = window_frame.stroke.width / 2.0;
-        // Add border padding to the inner margin to prevent it from covering the contents
-        window_frame.inner_margin += border_padding;
```
juancampa, line 441: As an added benefit, spacing calculations for windows can be simplified because the stroke is 100% outside, whereas before it was half in, half out.

polina4096:
There is a visible gap where the line ends on the "after" screen shot (above the window cross).
Screenshot 2024-08-12 at 13 33 53

juancampa:
Screenshot 2024-08-12 at 15 45 44


> There is a visible gap where the line ends on the "after" screen shot

Good catch. Fixed!

invisageable:
Thanks the egoat!

crates/egui/src/containers/frame.rs
```diff
@@ -89,6 +89,7 @@ impl Frame {
         Self {
             inner_margin: Margin::symmetric(8.0, 2.0),
             fill: style.visuals.panel_fill,
+            // outer_margin: Margin::same(1.0),
```
emilk, line 92:
```suggestion
```

crates/egui/src/containers/panel.rs
```diff
@@ -339,10 +342,12 @@ impl SidePanel {
                 Stroke::NONE
             };
             // TODO(emilk): draw line on top of all panels in this ui when https://github.com/emilk/egui/issues/1516 is done
-            // In the meantime: nudge the line so its inside the panel, so it won't be covered by neighboring panel
-            // (hence the shrink).
-            let resize_x = side.opposite().side_x(rect.shrink(1.0));
-            let resize_x = ui.painter().round_to_pixel(resize_x);
+            let resize_x = side.opposite().side_x(rect);
+            let resize_x = ui.painter().round_to_pixel_center(resize_x);
```
emilk, line 346: It should maybe be noted somewhere that this only makes sense for odd-width strokes (e.g. width=1, width=3 etc)

crates/egui/src/context.rs
```diff
@@ -1701,26 +1701,42 @@ impl Context {
         });
     }
 
-    /// Useful for pixel-perfect rendering
+    /// Useful for pixel-perfect rendering of lines
```
emilk, line 1704:
```suggestion
    /// Useful for pixel-perfect rendering of lines that are one pixel wide
```

crates/egui/src/context.rs
```diff
@@ -1701,26 +1701,42 @@ impl Context {
...
+    #[inline]
+    pub(crate) fn round_to_pixel_center(&self, point: f32) -> f32 {
+        let pixels_per_point = self.pixels_per_point();
+        ((point * pixels_per_point - 0.5).round() + 0.5) / pixels_per_point
+    }
+
+    /// Useful for pixel-perfect rendering of lines
```
emilk, line 1711:
```suggestion
    /// Useful for pixel-perfect rendering of lines that are one pixel wide
```

crates/egui/src/painter.rs
```diff
@@ -158,7 +158,19 @@ impl Painter {
         self.clip_rect = clip_rect;
     }
 
-    /// Useful for pixel-perfect rendering.
+    /// Useful for pixel-perfect rendering of strokes.
```
emilk, line 161:
```suggestion
    /// Useful for pixel-perfect rendering of strokes that are one pixel wide
```

crates/egui/src/painter.rs
```diff
@@ -158,7 +158,19 @@ impl Painter {
...
+    #[inline]
+    pub fn round_to_pixel_center(&self, point: f32) -> f32 {
+        self.ctx().round_to_pixel_center(point)
+    }
+
+    /// Useful for pixel-perfect rendering of strokes.
```
emilk, line 167:
```suggestion
    /// Useful for pixel-perfect rendering of strokes that are one pixel wide
```

crates/epaint/src/stroke.rs
```diff
@@ -55,6 +55,20 @@ impl std::hash::Hash for Stroke {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub(crate) enum StrokeKind {
+    Outside,
+    Inside,
+    Middle,
+}
```
emilk, line 64: Please add docstrings to this enum and the variants

crates/epaint/src/stroke.rs
```diff
@@ -63,20 +77,23 @@ impl std::hash::Hash for Stroke {
 pub struct PathStroke {
     pub width: f32,
     pub color: ColorMode,
+    pub(crate) kind: StrokeKind,
```
emilk, line 80: `PathStroke` is public, so I think `StrokeKind` should be too

crates/epaint/src/stroke.rs
```diff
@@ -91,6 +108,28 @@ impl PathStroke {
...
+    }
+
+    pub fn outside(self) -> Self {
+        Self {
+            kind: StrokeKind::Outside,
+            ..self
+        }
+    }
+
+    pub fn inside(self) -> Self {
```
emilk, line 129: docstrings

crates/epaint/src/tessellator.rs
```diff
@@ -868,6 +866,21 @@ fn fill_closed_path_with_uv(
     }
 }
 
+/// Translate a point according to the stroke kind.
+fn translate_stroke_point(p: &PathPoint, stroke: &PathStroke) -> PathPoint {
+    match stroke.kind {
+        stroke::StrokeKind::Middle => p.clone(),
```
emilk, line 872: nit: we should add `#[derive(Copy)]` on `PathPoint` so it is obvious that this is cheap

```suggestion
        stroke::StrokeKind::Middle => *p,
```

emilk (commented):
This is great - thanks for fixing this!

----

I think we need to fix the gap between stroke and fill - it is very noticable:

gap


Compared to `master`:
no-gap

crates/epaint/src/tessellator.rs
```diff
@@ -885,8 +898,13 @@ fn stroke_path(
     let idx = out.vertices.len() as u32;
 
     // expand the bounding box to include the thickness of the path
-    let bbox = Rect::from_points(&path.iter().map(|p| p.pos).collect::<Vec<Pos2>>())
-        .expand((stroke.width / 2.0) + feathering);
+    let bbox = Rect::from_points(
+        &path
+            .iter()
+            .map(|p| translate_stroke_point(p, stroke).pos)
```
emilk, line 904: I think running `translate_stroke_point` once on all the points (mutating the input) would be simpler, and also fix the gap between the stroke and the fill

juancampa: That did simplify things but it didn't completely close the gap. The issue is that feathering is slightly shrinking both the fill and stroke in opposite directions. The gap does disappear when feathering is disabled. I haven't looked into it but I'm wondering if you have any suggestions.

1px feathering
Screenshot 2024-08-28 at 21 49 07

5px feathering
Screenshot 2024-08-28 at 21 55 09

No feathering
Screenshot 2024-08-28 at 21 54 36

juancampa: I spent some time thinking about this and landed on: the fill should feather to the stroke color (instead of transparent), and vice versa.

I first thought that it would be okay to just grow the fill but that wouldn't work if the stroke is translucent.

juancampa:
It turns out that feathering the _fill_ to the stroke color and keeping the stroke feathering transparent takes us 99.9% there.

Screenshot 2024-08-29 at 11 35 31

emilk (approved):
This is an amazing improvement on low-dpi screens ❤️

emilk:
Related:
* https://github.com/emilk/egui/issues/5164

PR #4949, opened 2024-08-12
Update all examples, all tests, and `egui_demo_app` to use `env_logger 0.11`

LikeLakers2:

This PR updates all examples, all tests, and the `egui_demo_app` crate to use the latest version of env_logger, `0.11`. This helps to clean up the list of outdated crates listed when running `cargo outdated --root-deps-only --workspace`.

To ensure I didn't break anything, I ran `cargo test` both before and after the changes, and observed no changes in the output. Additionally, the CI run for my commit finished successfully: https://github.com/LikeLakers2/egui/actions/runs/10347887485

(P.S. I recommend updating your `Cargo.lock` at some point, as `cargo-outdated` also lists a bunch of crates that have new minor/patch versions.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -204,10 +204,53 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
-name = "anstyle"
-version = "1.0.4"
+name = "anstream"
+version = "0.6.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7079075b41f533b8c61d2a4d073c4676e1f8b249ff94a393b0595db304e0dd87"
+checksum = "64e15c1ab1f89faffbf04a634d5e1962e9074f2741eef6d97f3c4e322426d526"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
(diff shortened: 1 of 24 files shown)


bircni (approved):
I additionally ran `scripts/check.sh` - seems to work!

LikeLakers2:
@bircni Thanks! I'm on windows so I couldn't run that shell script - which is why I pointed to the Github Action run, because I was pretty sure it'd run `scripts/check.sh`. :)

bircni:
The script is for Linux only 😊

Cargo.lock
```diff
@@ -4172,6 +4237,12 @@ dependencies = [
  "tiny-skia-path",
 ]
 
+[[package]]
+name = "utf8parse"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "06abde3611657adf66d383f00b093d7faecc7fa57071cce2578660c9f1010821"
+
```
emilk, line 4245: This adds a lot of dependencies - is there a way to avoid them with feature-flags on `env_logger`, or by using an alternative crate?

LikeLakers2: I don't know. But given that these are examples and tests (things which are rarely compiled in full, or rarely compiled at all, especially compared to `egui` itself), I don't think the extra dependencies matter.

LikeLakers2:
Is there anything blocking this from being merged?

LikeLakers2:
It's been over a year with no response. As I genuinely cannot be bothered to continue checking this PR, plus the merge conflicts happening, I'm closing this PR.

PR #4833, opened 2024-07-16
Upgrade Glow to 0.14

victoryforphil:
# Reason for Change
By updating `glow` to `0.14`, this will pull a multiude of beneficial changes, such as [Make debug message callback Send + Sync.
](https://github.com/grovesNL/glow/commit/edfd08e2a941f4c3db6a3ba6cf84865f7c5365ff) which is important when drawing `three-d` windows in a glow context. 

Glow update: https://github.com/grovesNL/glow/pull/296

# Description of Changes
- Bump `glow` to `0.14`
- No breaking changes notices from upgrade.
# Verification
- Ran `./scripts/check` ✅  
- Been running these changes manually on a in-use project

* Closes: N/A (can make a ticket if requested)
* [x] I have followed the instructions in the PR template

## Labels:
`dependencies`

PR #4952, opened 2024-08-12
Update `glow` to 0.14

bircni:
Before making this PR, I did take notice of a similar PR, https://github.com/emilk/egui/pull/4833, but as it appears to be abandoned, I decided to make this PR.

**Missing**
One of the checks doesn't pass as wgpu still uses glow `0.13.1`

```shell
cargo deny --all-features --log-level error --target aarch64-apple-darwin check
```

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -59,6 +59,8 @@ skip = [
   { name = "time" },           # old version pulled in by unmaintianed crate 'chrono'
   { name = "windows-core" },   # old version via accesskit_windows
   { name = "windows" },        # old version via accesskit_windows
+  { name = "glow" },           # ignore glow. See #4952
+
 ]
 skip-tree = [
   { name = "criterion" },     # dev-dependency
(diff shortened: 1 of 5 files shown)


bircni:
glow is already on the main branch of wgpu so we just need to wait a bit
https://github.com/gfx-rs/wgpu/blob/7c917abf525b5c32c0b7345ec31788e316afa0cc/Cargo.toml#L162

Wumpf:
As mentioned here https://github.com/gfx-rs/wgpu/issues/6240#issuecomment-2337846087, this shouldn't be blocked on wgpu update. Just add a cargo deny exception here for the time being - the usecase of having both wgpu and glow backends of eframe enabled doesn't really exist, no?

bircni:
@Wumpf you are right gonna add this asap

deny.toml
```diff
@@ -59,6 +59,8 @@ skip = [
   { name = "time" },           # old version pulled in by unmaintianed crate 'chrono'
   { name = "windows-core" },   # old version via accesskit_windows
   { name = "windows" },        # old version via accesskit_windows
+  { name = "glow" },           # ignore glow. See #4952
```
emilk, line 62:
```suggestion
  { name = "glow" },           # wgpu uses an old `glow`, but realistically no one uses _both_ `egui_wgpu` and `egui_glow`, so we won't get a duplicate dependency
```

emilk (approved):

PR #4954, opened 2024-08-13
POC: Bump rust-version to `1.78`

bircni:
As mentioned in the last update https://github.com/emilk/egui/pull/4411 it's ok to be 2 versions behind and I think this would be useful.
To make this work I also updated the wasm version from 0.2.92 to 0.2.93

* [x] I ran the script in `./scripts/check.sh`

### DIFF
--- a/.github/workflows/deploy_web_demo.yml
+++ b/.github/workflows/deploy_web_demo.yml
@@ -11,7 +11,6 @@ on:
 #   release:
 #     types: ["published"]
 
-
 permissions:
   contents: write # for committing to gh-pages branch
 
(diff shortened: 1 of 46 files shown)


bircni:
To fix #4961 I also bumped web-sys to the newer version

oscargus:
Isn't there a point at keeping the MSRV as low as possible? That is, if needed, from dependencies or language features, bump it, but not just for the sake of it?

(In my project, we had some reports after egui required 1.76 as it could not be built on some platforms.)

Then it of course makes sense to test using the minimum version as well as the most recent (and possibly beta and nightly to catch upcoming problems).

bircni:
@oscargus yeah you are right first it was just like a poc
Maybe i gonna set this as a draft for later and put the web_sys fix into another pr

PR #4975, opened 2024-08-17
Add `Label::halign`

rustbasic:
Function to specify `Align` to `Label`

There are cases where you want to display `Label` on the left or right regardless of other `Layout` specifications.

Before : Note the part showing the rust source code.

After : Note the part showing the rust source code.

### DIFF
--- a/crates/egui/src/widgets/label.rs
+++ b/crates/egui/src/widgets/label.rs
@@ -26,6 +26,7 @@ pub struct Label {
     wrap_mode: Option<TextWrapMode>,
     sense: Option<Sense>,
     selectable: Option<bool>,
+    align: Option<Align>,
 }
 
 impl Label {
@@ -35,6 +36,7 @@ impl Label {
             wrap_mode: None,
             sense: None,
             selectable: None,
+            align: None,
         }
     }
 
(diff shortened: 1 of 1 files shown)


bircni:
maybe change the function name to text_align as this only aligns the text inside the label not the label itself

rustbasic:
> maybe change the function name to text_align as this only aligns the text inside the label not the label itself

Yes, that would be great too. Let's wait for emilk's review.

crates/egui/src/widgets/label.rs
```diff
@@ -76,6 +78,13 @@ impl Label {
         self
     }
 
+    /// Set [`Self::align`] to [`Align`].
```
emilk, line 81: This doesn't explain anything

crates/egui/src/widgets/label.rs
```diff
@@ -76,6 +78,13 @@ impl Label {
...
+    #[inline]
+    pub fn align(mut self, align: Align) -> Self {
+        self.align = Some(align);
```
emilk, line 84: `halign` is probably a better name

emilk (changes requested):

crates/egui/src/widgets/label.rs
```diff
@@ -76,6 +78,15 @@ impl Label {
         self
     }
 
+    /// Sets the horizontal alignment of the Label to the given `Align` value.
+    ///
+    /// Set [`Self::halign`] to [`Align`].
```
emilk, line 83:
```suggestion
```

PR #4978, opened 2024-08-18
No more wildcard imports and general import cleanup

oscargus:

Not sure if wildcard imports are wanted or not, but this removes them and also changes it to warnings.

In addition, there are a number of fixes where the "correct" import location is used for `emath` and `Color32`. That is, the re-exports are not used. Again, I do not know the stance of this, but to keep things consistent in the code base and possibly simplify deprecating the re-exports etc now that there are multiple crates.

Edit: Seems only half-finished, but any input if I should finish it is appreciated.


oscargus:
Closing and adding parts in separate PRs. Will cause too many conflicts and probably wildcard imports are wanted every now and then...

PR #4981, opened 2024-08-18
Always import emath directly when already a direct dependency

oscargus:

A bit of cleaning up in the import/re-export jungle. Now, no wildcard imports from emath and if emath is already a direct dependency, import from emath, not crate::emath or crate::epaint::emath or ...

In some instances I have also moved a some epaint stuff, but the focus was on emath.


emilk (approved):

emilk:
Sorry, this was forgotten, and now there is _a lot_ of merge conflicts 😬 
Please reopen if you feel strongly about it

PR #4986, opened 2024-08-19
Add max click timings & distance to `Options`

Azorlogh:

Moves the click timing & distance constants to the Options struct so they can be configured by users.

Note: This need came up as part of my investigation of https://github.com/mvlabat/bevy_egui/issues/302. Essentially, on some platforms we are not guaranteed to have correct timekeeping so it's useful to be able to disable the max click duration in this case.

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -489,7 +489,11 @@ impl ContextImpl {
                 &viewport.prev_frame.widgets,
                 &viewport.hits,
                 &viewport.input,
-                self.memory.interaction_mut(),
+                self.memory
+                    .interactions
+                    .entry(self.memory.viewport_id)
+                    .or_default(),
+                &self.memory.options,
             );
         }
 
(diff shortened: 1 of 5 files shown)


crates/egui/src/context.rs
```diff
@@ -489,7 +489,11 @@ impl ContextImpl {
                 &viewport.prev_frame.widgets,
                 &viewport.hits,
                 &viewport.input,
-                self.memory.interaction_mut(),
+                self.memory
+                    .interactions
+                    .entry(self.memory.viewport_id)
+                    .or_default(),
+                &self.memory.options,
```
Azorlogh, line 496: This isn't super pretty, but otherwise it conflicts with the mut access of `memory` above

crates/egui/src/ui.rs
```diff
@@ -2721,7 +2721,7 @@ fn register_rect(ui: &Ui, rect: Rect) {
         return;
     }
 
-    let is_clicking = ui.input(|i| i.pointer.could_any_button_be_click());
+    let is_clicking = ui.memory(|m| ui.input(|i| i.pointer.could_any_button_be_click(&m.options)));
```
emilk, line 2724: This deadlocks (try holding down all modifier keys in the demo app)

emilk (changes requested):
Makes sense, but we need to fix that deadlock

Azorlogh:
> There is an alternative PR for this in #4942 which I think uses a slightly better approach, and avoids the deadlock

Oh I totally missed it! Yeah I agree that other one is a lot cleaner :ok_hand: I'll close this one

Issue #4941, opened 2024-08-09
`TextEdit` requires double touch to focus on Android


Zoxc:
`TextEdit` requires double touch now to focus on Android in 0.28.1. Other however elements still correctly work with a single touch however.

This still occurs on `1f6ae49a5f6bf43a869c215dea0d3028be8d742a`.

Zoxc:
It looks like this is because `TextEdit` does not request focus when clicking. Clicking with a mouse counts as a drag which does request focus, but a touch doesn't count as a drag.

emilk:
This should now be fixed on egui.rs

PR #4991, opened 2024-08-21
Request focus on a TextEdit when clicked

Zoxc:
This request focus on a TextEdit when clicked, so that touch events on Android can now focus it.

This looks like a reasonable fix for https://github.com/emilk/egui/issues/4941.


rustbasic:
I guess this won't be needed once #4865 is applied.

emilk (approved):

PR #4997, opened 2024-08-24
Remove the `accesskit` feature on egui

### LINKED ISSUE #3137: Remove `accesskit` feature in `egui`, and always depend on it
It seems that by giving into the request to disable AccessKit by default in the base egui crate, I have set a confusing precedent, as we can see in [this comment](https://github.com/slint-ui/slint/pull/3046#issuecomment-1623675120). If there are practical reasons to continue to disable AccessKit by default in the base egui crate, I will do what I can to address those. But my position is now that accessibility should be the default at all layers of the stack, and it should be something that developers have to work to disable.

DataTriny:
Closes #3137

### DIFF
--- a/crates/eframe/Cargo.toml
+++ b/crates/eframe/Cargo.toml
@@ -42,7 +42,7 @@ default = [
 ]
 
 ## Enable platform accessibility API implementations through [AccessKit](https://accesskit.dev/).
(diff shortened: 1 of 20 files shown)


lucasmerlin:
Closed via https://github.com/emilk/egui/pull/7701

PR #5010, opened 2024-08-26
Use `log` crate instead of `eprintln` & remove some unwraps

bircni:
- I fixed the TODO to use the `log` crate instead of `eprintln`
- Set the rust-version in the `scripts/check.sh` to the same as egui is on
- I made xtask use anyhow to remove some unwraps 

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,9 +74,10 @@ backtrace = "0.3"
 bytemuck = "1.7.2"
 criterion = { version = "0.5.1", default-features = false }
 document-features = " 0.2.8"
-glow = "0.13"
+glow = "0.14"
 glutin = "0.32.0"
 glutin-winit = "0.5.0"
+home = "0.5.9"
 image = { version = "0.25", default-features = false }
 log = { version = "0.4", features = ["std"] }
 nohash-hasher = "0.2"
@@ -95,6 +96,7 @@ wgpu = { version = "22.1.0", default-features = false, features = [
     # Make the renderer `Sync` even on wasm32, because it makes the code simpler:
     "fragile-send-sync-non-atomic-wasm",
 ] }
+windows-sys = "0.52"
 winit = { version = "0.30.5", default-features = false }
 
 [workspace.lints.rust]
@@ -118,6 +120,7 @@ unused_qualifications = "allow"
 [workspace.lints.rustdoc]
 all = "warn"
 missing_crate_level_docs = "warn"
+broken_intra_doc_links = "warn"
 
 # See also clippy.toml
 [workspace.lints.clippy]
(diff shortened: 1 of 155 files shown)


Cargo.toml
```diff
@@ -249,12 +249,12 @@ useless_transmute = "warn"
 verbose_file_reads = "warn"
 wildcard_dependencies = "warn"
 zero_sized_map_values = "warn"
+print_stderr = "warn"
```
YgorSouza, line 252: This list is sorted alphabetically.

xtask/src/utils.rs
```diff
@@ -30,16 +28,16 @@ pub fn ask_to_run(mut cmd: Command, ask: bool, reason: &str) -> Result<(), DynEr
...
         match buf.trim().to_lowercase().as_ref() {
             "" | "y" | "yes" => {}
-            "n" | "no" => return Err("Aborting as per your request".into()),
-            a => return Err(format!("Invalid answer `{a}`").into()),
+            "n" | "no" => anyhow::bail!("Aborting as per your request"),
+            a => anyhow::bail!(format!("Invalid answer `{a}`")),
         };
     } else {
-        eprintln!("Running `{cmd:?}` to {reason}.");
+        log::info!("Running `{cmd:?}` to {reason}.");
```
YgorSouza, line 35: Don't know if this makes sense. xtask is supposed to be just like a shell script, but in Rust. The fewer bells and whistles it has, the better.

bircni: you are right removed that and replaced it with println

bircni:
@emilk Could you have a look at it?

examples/puffin_profiler/Cargo.toml
```diff
@@ -27,3 +27,4 @@ env_logger = { version = "0.10", default-features = false, features = [
 ] }
 puffin = "0.19"
 puffin_http = "0.16"
+log = { workspace = true }
```
emilk, line 30: nit: sort dependencies

examples/puffin_profiler/src/main.rs
```diff
@@ -153,7 +153,7 @@ fn start_puffin_server() {
 
     match puffin_http::Server::new("127.0.0.1:8585") {
         Ok(puffin_server) => {
-            eprintln!("Run:  cargo install puffin_viewer && puffin_viewer --url 127.0.0.1:8585");
+            log::info!("Run:  cargo install puffin_viewer && puffin_viewer --url 127.0.0.1:8585");
```
emilk, line 156: This means the user won't see the message unless they have `RUST_LOG=info` env-var set.

Either revert, or add this to before the call to `env_logger::init`:

```rs
// Show log messages by default:
let rust_log = std::env::var("RUST_LOG").unwrap_or_else(|_| "info".to_owned());
std::env::set_var("RUST_LOG", rust_log);
```

emilk (approved):

PR #5011, opened 2024-08-27
Remove reference to egui::Color32.

tiaanl:
As most of the code refers to types in epaint, it makes sense to use the Color32 alias from epaint as well.


Wumpf (approved):
sure, same to me either way 🤷

PR #5030, opened 2024-08-29
Enable rustdoc `generate-link-to-definition` feature on docs.rs

GuillaumeGomez:
You can see this feature in action [here](https://docs.rs/sysinfo/latest/src/sysinfo/common/system.rs.html#46) or on any of dtolnay's crates and many others. I found myself going through your project code recently on docs.rs and I was a bit sad I couldn't have this feature enabled. This should fix it at next release. :)

### DIFF
--- a/crates/ecolor/Cargo.toml
+++ b/crates/ecolor/Cargo.toml
@@ -21,6 +21,7 @@ workspace = true
 
 [package.metadata.docs.rs]
 all-features = true
+rustdoc-args = ["--generate-link-to-definition"]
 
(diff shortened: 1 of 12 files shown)


emilk:
Very cool feature; I didn't know about this!

Why is it opt-in and not enabled by default?

GuillaumeGomez:
Not stable and likely won't be before a long time. There are some disagreements in the rustdoc about how we should implement this feature or whether or not we should even have this feature in the first place. "politics"

PR #5038, opened 2024-08-30
build(deps): Replace `winapi` by `windows-sys`

caspermeijn:
`winapi` has not been updated for four years. `windows-sys` seems to have a good maintainer. `winit` also uses `windows-sys`.

The Windows API calls are not changed. It just uses a different bindings library.

### Alternative

I tried to remove the Windows API calls from this crate. As stated in a comment, `winit` can do some of the work, but I was unable to fix `winit` enough to use that. My poor knowledge of Windows API doesn't help with that.

```
// We would get fairly far already with winit's `set_window_icon` (which is exposed to eframe) actually!
// However, it only sets ICON_SMALL, i.e. doesn't allow us to set a higher resolution icon for the task bar.
// Also, there is scaling issues, detailed below.
```

### DIFF
--- a/crates/eframe/Cargo.toml
+++ b/crates/eframe/Cargo.toml
@@ -184,8 +184,13 @@ objc2-app-kit = { version = "0.2.0", features = [
 
 # windows:
 [target.'cfg(any(target_os = "windows"))'.dependencies]
-winapi = { version = "0.3.9", features = ["winuser"] }
-windows-sys = { workspace = true, features = ["Win32_Foundation", "Win32_UI_Shell", "Win32_System_Com"] }
+windows-sys = { workspace = true, features = [
+  "Win32_Foundation",
+  "Win32_System_Com",
+  "Win32_UI_Input_KeyboardAndMouse",
+  "Win32_UI_Shell",
+  "Win32_UI_WindowsAndMessaging",
+] }
 
 # -------------------------------------------
 # web:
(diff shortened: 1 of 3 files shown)


crates/eframe/Cargo.toml
```diff
@@ -183,7 +183,10 @@ objc2-app-kit = { version = "0.2.0", features = [
 
 # windows:
 [target.'cfg(any(target_os = "windows"))'.dependencies]
-winapi = { version = "0.3.9", features = ["winuser"] }
+windows-sys = { version = "0.59", features = [
```
YgorSouza, line 186: winit is still on 0.52. Might be better to sync with them.

caspermeijn: I see `cargo-deny` also complains about that. Version 0.52 has a less ergonomic API. I added a separate commit to make the difference visible. I will look into what it takes to update `winit` to version 0.59.



```
cargo deny --all-features --log-level error --target i686-pc-windows-gnu check 
error[duplicate]: found 2 duplicate entries for crate 'windows-sys'
    ┌─ C:\Users\caspe\Documents\GitHub\egui/Cargo.lock:275:1
    │
275 │ ╭ windows-sys 0.52.0 registry+https://github.com/rust-lang/crates.io-index
276 │ │ windows-sys 0.59.0 registry+https://github.com/rust-lang/crates.io-index
    │ ╰────────────────────────────────────────────────────────────────────────
```

emilk (approved):

Issue #4987, opened 2024-08-19
Recent regression in TextEdit assuming the canvas is aligned to the top


canselcik:
This is with the latest release, using eframe. It seems https://github.com/emilk/egui/issues/4060 came back in a new form.

**To Reproduce**
1- Call `ui.text_edit_singleline(...)`
2- Click the text edit, once focus is gained, the canvas will be aligned to the top.

**Expected behavior**
Canvas isn't moved.

**Screenshots**
Before focus:
<img width="165" alt="image" src="https://github.com/user-attachments/assets/c283a0fe-3b9f-4c51-859f-c7242e597016">
After focus:
<img width="84" alt="image" src="https://github.com/user-attachments/assets/ba8a1285-1123-45ac-9f88-334c79eee30b">

**Desktop (please complete the following information):**
 - OS:  OSX
 - Browser: Chromium
 - Version 127

**Smartphone (please complete the following information):**
Happens on Android Chrome as well.

**Additional context**
It is interesting because the workaround we had devised in https://github.com/emilk/egui/issues/4060 is no longer protecting against this. In fact the CSS seems entirely unmodified and the attributes of the canvas are untouched as well, tested this with `break on access` etc.

This was the workaround:
```css
canvas {
  position: initial !important;
  top: unset !important;
}
```
I suspect this is related to some recent IME changes.

---

Had a chance to explore a little bit more. Commenting out https://github.com/emilk/egui/blob/9a1e358a144b5d2af9d03a80257c34883f57cf0b/crates/eframe/src/web/text_agent.rs#L34 fixes the symptom.

So if we don't add the input for the TextAgent into the DOM. I will continue to narrow down further.

canselcik:
Looks like adding `style.set_property("display", "none")?` on the input we create takes care of the problem as well, and unlike not appending it to the body, this approach probably won't be breaking the IME features.

Another thing to consider is perhaps to put it in Shadow DOM. Not sure if that will work with the IME features, and it will require enabling `ShadowRootInit`, `ShadowRootMode` and `ShadowRoot` features on `web-sys`.

canselcik:
Closing with this comment: https://github.com/emilk/egui/pull/5041#issuecomment-2327561929
egui's current handling is good, if you have a similar issue, I'd recommend making sure you don't have `overflow: hidden` in the body or in some parent element.

PR #5041, opened 2024-08-30
`display:none` on Input inserted by `TextAgent`

canselcik:
* Closes https://github.com/emilk/egui/issues/4987

There isn't much to discuss as the change is very limited in scope. Doing this seems to break nothing for me, even on a mobile device while it fixes the regression explained in the linked issue.

### DIFF
--- a/crates/eframe/src/web/text_agent.rs
+++ b/crates/eframe/src/web/text_agent.rs
@@ -34,6 +34,7 @@ impl TextAgent {
         style.set_property("height", "1px")?;
         style.set_property("caret-color", "transparent")?;
         style.set_property("position", "absolute")?;
+        style.set_property("display", "none")?;
         style.set_property("top", "0")?;
         style.set_property("left", "0")?;
         document.body().unwrap().append_child(&input)?;


crates/eframe/src/web/text_agent.rs
```diff
@@ -34,6 +34,7 @@ impl TextAgent {
         style.set_property("height", "1px")?;
         style.set_property("caret-color", "transparent")?;
         style.set_property("position", "absolute")?;
+        style.set_property("display", "none")?;
```
emilk, line 37: Please adda comment explaining why we have this line of code here, so we don't accidentally remove it again!

emilk (approved):
Thanks!

crates/eframe/src/web/text_agent.rs
canselcik: Will do, it seems we didn't remove it since the last time anyway. The issue was previously with the canvas, this time it is the input we add.

I am going to test it a little bit more today, when added as a "home screen app" on Android, I noticed the behavior popped up again somehow, I want to confirm if that was an older cached version or another corner case.

canselcik: Plot thickens: https://chromestatus.com/feature/5093566007214080

So the `display: none` fixes the issue on non-touch devices. On touch devices we still seem to have the same thing happening with the following coinciding with it, trying without the events added now.

canselcik: So it seems if touch events are not enabled (via developer tools or by actually being on a touch-capable device) when the app is getting loaded, but afterwards, this doesn't happen at all.

The `display: none` fix seems to be effective to have this issue not happen on non-touch devices.
But with or without `display: none`, it continues to happen on touch devices, even with zero user interaction. The bug that's addressed by `display: none` originates from the `TextAgent` inputbox and does not show its head unless user interaction takes place. However the other one seems to happen regardless of user interaction.

Will continue to investigate. I think we should merge this (with my added comment soon) and I will try to see where the non-user interaction requiring version of this behavior stems from.

canselcik:
This whole thing is definitely some sort of Chrome bug or idiosyncrasy depending on how one likes to look at it. One good thing about it is that it only happens if you do `overflow: hidden` on the body, which I can live without. I'll close with this comment and keep it around for reference in case someone else also hits this strange behavior.

egui's current TextAgent (with our last modification that's already merged) is not causing this issue, the added eventListeners passive or not are also cleared. Even though they are all working together to have this behavior it happens if and only if the body has `overflow: hidden`, which is easy to avoid having.

PR #5043, opened 2024-08-31
This part is unnecessary in wgpu_integration.rs

rustbasic:
This part is unnecessary, as we call  `remove_viewports_not_in()` in the `handle_viewport_output()` function right before it.

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -552,6 +552,44 @@ impl GlowWinitRunning {
             (raw_input, viewport_ui_cb)
         };
 
+        let clear_color = self
+            .app
+            .clear_color(&self.integration.egui_ctx.style().visuals);
+
+        let has_many_viewports = self.glutin.borrow().viewports.len() > 1;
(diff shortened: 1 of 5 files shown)


crates/eframe/src/native/glow_integration.rs
```diff
@@ -552,44 +552,6 @@ impl GlowWinitRunning {
...
-            {
-                frame_timer.pause();
-                change_gl_context(current_gl_context, not_current_gl_context, gl_surface);
-                frame_timer.resume();
-            }
-
-            self.painter
-                .borrow()
-                .clear(screen_size_in_pixels, clear_color);
-        }
```
emilk, line 591: Doesn't this break `test_inline_glow_paint`?

rustbasic: > Doesn't this break `test_inline_glow_paint`?

Yes, no problem. 
(The plan was to PR wgpu and glow separately.)

rustbasic: > Doesn't this break `test_inline_glow_paint`?

As planned, I will test glow separately and if there are no issues, I will PR again.

PR #5050, opened 2024-09-01
Provides Rect::new()

rustbasic:
I've sometimes wondered why there's no Rect::new() , but I see no reason why it shouldn't be provided.

### DIFF
--- a/crates/emath/src/rect.rs
+++ b/crates/emath/src/rect.rs
@@ -70,6 +70,14 @@ impl Rect {
     };
 
     #[inline(always)]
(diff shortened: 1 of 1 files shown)


Mingun:
I think, because it may be unclear, what the numbers mean. It is:
- position of two corners?
- position of a corner and size?
- position of a center and size?

Because fields of `Rect` are public, constructing struct directly allows to write self-documenting code.

emilk:
`Rect::from_min_max` already covers this, and better

PR #5054, opened 2024-09-01
Interactive `Ui`:s: add `UiBuilder::sense` and `Ui::response`

### LINKED ISSUE #5053: `Context::interact_bg` breaks focus
**Describe the bug**

Calling `ui.interact_bg(Sense::click())` somewhere in an egui applicaiton makes it impossible to focus any elements added after the call.

https://github.com/user-attachments/assets/588c2aef-005b-491a-acb1-5578deb1aa44

The label in the video is in a child ui where `interact_bg` is called. I just keep pressing tab. 
I'd expect the focus to jump between the first button, the child scope and the second button with each tab press. 
Instead the scope seems to receive focus for a single frame and then loose it, causing the first button to be focused again on the next tab press.

**To Reproduce**
Example code: 
```rust
use eframe::egui;
use eframe::NativeOptions;
use egui::{CentralPanel, Sense};

pub fn main() -> eframe::Result {
    eframe::run_simple_native("focus test", NativeOptions::default(), |ctx, _frame| {
            ui.button("I can focus");

            ui.push_id("focus me", |ui| {
                ui.label("I can focus for a single frame");
                let response = ui.interact_bg(Sense::click());
                let t = if response.has_focus() {
                    "has focus"
                } else {
                    "doesn't have focus"
                };
                ui.label(t);
            });

            ui.button("I can't focus :(");
        });
    })
}

```

**Expected behavior**
Focus jumps between all focusable elements.

 - OS: macOS 14.6.1

lucasmerlin:
* Closes #5053 

This fixes #5053 by adding a Sense parameter to UiBuilder, using that in Context::create_widget, so the Widget is registered with the right Sense / focusable. Additionally, I've added a ignore_focus param to create_widget, so the focus isn't surrendered / reregistered on Ui::interact_bg. 

The example from #5053 now works correctly: 

https://github.com/user-attachments/assets/a8a04b5e-7635-4e05-9ed8-e17d64910a35

<details><summary>Updated example code</summary>
<p>

```rust
            ui.button("I can focus");

            ui.scope_builder(
                UiBuilder::new()
                    .sense(Sense::click())
                    .id_source("focus_test"),
                |ui| {
                    ui.label("I can focus for a single frame");
                    let response = ui.interact_bg();
                    let t = if response.has_focus() {
                        "has focus"
                    } else {
                        "doesn't have focus"
                    };
                    ui.label(t);
                },
            );

            ui.button("I can't focus :(");
```

</p>
</details> 

---

Also, I've added `Ui::interact_scope` to make it easier to read a Ui's response in advance, without having to know about the internals of how the Ui Ids get created.

This makes it really easy to created interactive container elements or custom buttons, without having to use Galleys or Painter::add(Shape::Noop) to style based on the interaction.

<details><summary>
Example usage to create a simple button
</summary>
<p>

```rust
use eframe::egui;
use eframe::egui::{Frame, InnerResponse, Label, RichText, UiBuilder, Widget};
use eframe::NativeOptions;
use egui::{CentralPanel, Sense, WidgetInfo};

pub fn main() -> eframe::Result {
    eframe::run_simple_native("focus test", NativeOptions::default(), |ctx, _frame| {
            ui.button("Regular egui Button");
            custom_button(ui, |ui| {
                ui.label("Custom Button");
            });

            if custom_button(ui, |ui| {
                ui.label("You can even have buttons inside buttons:");

                if ui.button("button inside button").clicked() {
                    println!("Button inside button clicked!");
                }
            })
            .response
            .clicked()
            {
                println!("Custom button clicked!");
            }
        });
    })
}

fn custom_button<R>(
    ui: &mut egui::Ui,
    content: impl FnOnce(&mut egui::Ui) -> R,
) -> InnerResponse<R> {
    let auto_id = ui.next_auto_id();
    ui.skip_ahead_auto_ids(1);
    let response = ui.interact_scope(
        Sense::click(),
        UiBuilder::new().id_source(auto_id),
        |ui, response| {
            ui.style_mut().interaction.selectable_labels = false;
            let visuals = response
                .map(|r| ui.style().interact(&r))
                .unwrap_or(&ui.visuals().noninteractive());
            let text_color = visuals.text_color();

            Frame::none()
                .fill(visuals.bg_fill)
                .stroke(visuals.bg_stroke)
                .rounding(visuals.rounding)
                .inner_margin(ui.spacing().button_padding)
                .show(ui, |ui| {
                    ui.visuals_mut().override_text_color = Some(text_color);
                    content(ui)
                })
                .inner
        },
    );

    response
        .response
        .widget_info(|| WidgetInfo::new(egui::WidgetType::Button));

    response
}
```

</p>
</details> 

https://github.com/user-attachments/assets/281bd65f-f616-4621-9764-18fd0d07698b

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -1050,7 +1050,7 @@ impl Context {
     ///
     /// If the widget already exists, its state (sense, Rect, etc) will be updated.
     #[allow(clippy::too_many_arguments)]
-    pub(crate) fn create_widget(&self, w: WidgetRect) -> Response {
+    pub(crate) fn create_widget(&self, w: WidgetRect, ignore_focus: bool) -> Response {
         // Remember this widget
         self.write(|ctx| {
             let viewport = ctx.viewport();
@@ -1060,12 +1060,12 @@ impl Context {
             // but also to know when we have reached the widget we are checking for cover.
             viewport.this_frame.widgets.insert(w.layer_id, w);
 
-            if w.sense.focusable {
+            if w.sense.focusable && !ignore_focus {
                 ctx.memory.interested_in_focus(w.id);
             }
         });
 
-        if !w.enabled || !w.sense.focusable || !w.layer_id.allow_interaction() {
+        if !w.enabled || !w.sense.focusable || !w.layer_id.allow_interaction() && !ignore_focus {
             // Not interested or allowed input:
             self.memory_mut(|mem| mem.surrender_focus(w.id));
         }
@@ -1078,7 +1078,7 @@ impl Context {
         let res = self.get_response(w);
 
         #[cfg(feature = "accesskit")]
-        if w.sense.focusable {
+        if w.sense.focusable && !ignore_focus {
             // Make sure anything that can receive focus has an AccessKit node.
             // TODO(mwcampbell): For nodes that are filled from widget info,
             // some information is written to the node twice.
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -1051,7 +1051,7 @@ impl Context {
     /// If the widget already exists, its state (sense, Rect, etc) will be updated.
     ///
     /// `allow_focus` should usually be true, unless you call this function multiple times with the
-    /// same widget, then `allow_focus` should only be true once (like in [`Ui::new`] (true) and [`Ui::interact_bg`] (false)).
+    /// same widget, then `allow_focus` should only be true once (like in [`Ui::new`] (true) and [`Ui::remember_min_rect`] (false)).
     #[allow(clippy::too_many_arguments)]
     pub(crate) fn create_widget(&self, w: WidgetRect, allow_focus: bool) -> Response {
         // Remember this widget
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -1117,7 +1117,7 @@ impl Context {
     }
 
     /// Do all interaction for an existing widget, without (re-)registering it.
-    fn get_response(&self, widget_rect: WidgetRect) -> Response {
+    pub(crate) fn get_response(&self, widget_rect: WidgetRect) -> Response {
         let WidgetRect {
             id,
             layer_id,
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -1049,8 +1049,11 @@ impl Context {
     /// You should use [`Ui::interact`] instead.
     ///
     /// If the widget already exists, its state (sense, Rect, etc) will be updated.
+    ///
+    /// `allow_focus` should usually be true, unless you call this function multiple times with the
+    /// same widget, then `allow_focus` should only be true once (like in [`Ui::new`] (true) and [`Ui::interact_bg`] (false)).
     #[allow(clippy::too_many_arguments)]
-    pub(crate) fn create_widget(&self, w: WidgetRect, ignore_focus: bool) -> Response {
+    pub(crate) fn create_widget(&self, w: WidgetRect, allow_focus: bool) -> Response {
         // Remember this widget
         self.write(|ctx| {
             let viewport = ctx.viewport();
@@ -1060,12 +1063,12 @@ impl Context {
             // but also to know when we have reached the widget we are checking for cover.
             viewport.this_frame.widgets.insert(w.layer_id, w);
 
-            if w.sense.focusable && !ignore_focus {
+            if allow_focus && w.sense.focusable {
                 ctx.memory.interested_in_focus(w.id);
             }
         });
 
-        if !w.enabled || !w.sense.focusable || !w.layer_id.allow_interaction() && !ignore_focus {
+        if allow_focus && (!w.enabled || !w.sense.focusable || !w.layer_id.allow_interaction()) {
             // Not interested or allowed input:
             self.memory_mut(|mem| mem.surrender_focus(w.id));
         }
@@ -1078,7 +1081,7 @@ impl Context {
         let res = self.get_response(w);
 
         #[cfg(feature = "accesskit")]
-        if w.sense.focusable && !ignore_focus {
+        if allow_focus && w.sense.focusable {
             // Make sure anything that can receive focus has an AccessKit node.
             // TODO(mwcampbell): For nodes that are filled from widget info,
             // some information is written to the node twice.
--- a/crates/egui/src/ui.rs
+++ b/crates/egui/src/ui.rs
@@ -93,6 +93,9 @@ pub struct Ui {
     /// The sense for the ui background.
     /// It will be used for [Ui::interact_bg].
     sense: Sense,
+
+    /// Whether [`Ui::interact_bg`] should be called when the [`Ui`] is dropped.
+    should_interact_bg_on_drop: bool,
 }
 
 impl Ui {
@@ -149,6 +152,7 @@ impl Ui {
             menu_state: None,
             stack: Arc::new(ui_stack),
             sense,
+            should_interact_bg_on_drop: true,
         };
 
         // Register in the widget stack early, to ensure we are behind all widgets we contain:
@@ -282,6 +286,7 @@ impl Ui {
             menu_state: self.menu_state.clone(),
             stack: Arc::new(ui_stack),
             sense,
+            should_interact_bg_on_drop: true,
         };
 
         // Register in the widget stack early, to ensure we are behind all widgets we contain:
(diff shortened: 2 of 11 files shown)


lucasmerlin:
@emilk I wonder if it would make sense to return a Response based on Ui::interact_bg from Ui::scope_dyn? Then the Ui responses would contain the right interactions based on UiBuilder::sense. 
Like this: 

```rust
    /// Create a child, add content to it, and then allocate only what was used in the parent `Ui`.
    pub fn scope_dyn<'c, R>(
        &mut self,
        ui_builder: UiBuilder,
        add_contents: Box<dyn FnOnce(&mut Ui) -> R + 'c>,
    ) -> InnerResponse<R> {
        let next_auto_id_source = self.next_auto_id_source;
        let mut child_ui = self.new_child(ui_builder);
        self.next_auto_id_source = next_auto_id_source; // HACK: we want `scope` to only increment this once, so that `ui.scope` is equivalent to `ui.allocate_space`.
        let ret = add_contents(&mut child_ui);
        self.allocate_rect(child_ui.min_rect(), Sense::hover());
        let response = child_ui.interact_bg();
        InnerResponse::new(ret, response)
    }
```

crates/egui/src/context.rs
```diff
@@ -1060,12 +1060,12 @@ impl Context {
...
             viewport.this_frame.widgets.insert(w.layer_id, w);
 
-            if w.sense.focusable {
+            if w.sense.focusable && !ignore_focus {
                 ctx.memory.interested_in_focus(w.id);
             }
         });
 
-        if !w.enabled || !w.sense.focusable || !w.layer_id.allow_interaction() {
+        if !w.enabled || !w.sense.focusable || !w.layer_id.allow_interaction() && !ignore_focus {
```
emilk, line 1068: Is this the correct logic? Remember that `&&` binds harder than `||`

crates/egui/src/context.rs
```diff
@@ -1060,12 +1060,12 @@ impl Context {
             // but also to know when we have reached the widget we are checking for cover.
             viewport.this_frame.widgets.insert(w.layer_id, w);
 
-            if w.sense.focusable {
+            if w.sense.focusable && !ignore_focus {
```
emilk, line 1063: Could we just set `w.sense.focusable = false` instead of having an extra parameter?

crates/egui/src/context.rs
```diff
@@ -1050,7 +1050,7 @@ impl Context {
     ///
     /// If the widget already exists, its state (sense, Rect, etc) will be updated.
     #[allow(clippy::too_many_arguments)]
-    pub(crate) fn create_widget(&self, w: WidgetRect) -> Response {
+    pub(crate) fn create_widget(&self, w: WidgetRect, ignore_focus: bool) -> Response {
```
emilk, line 1053: Negations in booleans is a source of confusion. Better to invert it and call it `allow_focus`

emilk (commented):
This is cool stuff!

I'd love to have an official example or test of this though.

And yes, I think having `scope_dyn` actually return a correct response based on `UiBuilder::sense` is better than adding a new `interact_scope` function

lucasmerlin:
> And yes, I think having `scope_dyn` actually return a correct response based on `UiBuilder::sense` is better than adding a new `interact_scope` function

The interact_scope function would still be required to read the response early and pass it to the ui closure. 
But I think we could also

- always read the response from ctx when a new Ui is created 
- store the response in the Ui struct 
- add Ui::read_response to access the stored response
- Call Ui::interact_bg when a Ui is dropped, so it automatically works with all kinds of child uis, not only scopes

Then the new Ui::interact_scope wouldn't be needed anymore and we could read the response early in all kinds of Uis.

lucasmerlin:
@emilk I have updated the PR to implement the ideas outlined above, I think I like this even more. 

I was able to implement this without having to store the response in the Ui struct, by inversing the logic in Ui::read_response, making this even cleaner.

Also, I added a small example: 

https://github.com/user-attachments/assets/772662db-9f29-4c26-bef0-fba8df6d8233

crates/egui/src/ui.rs
```diff
@@ -970,13 +994,59 @@ impl Ui {
...
+    }
+
     /// Interact with the background of this [`Ui`],
     /// i.e. behind all the widgets.
     ///
     /// The rectangle of the [`Response`] (and interactive area) will be [`Self::min_rect`].
-    pub fn interact_bg(&self, sense: Sense) -> Response {
+    /// You can customize the [`Sense`] via [`UiBuilder::sense`].
+    // This is marked as deprecated for public use but still makes sense to use internally.
+    #[deprecated = "Use Ui::read_response instead"]
```
emilk, line 1030:
```suggestion
    #[deprecated = "Use Ui::response instead"]
```

crates/egui/src/ui.rs
```diff
@@ -2811,6 +2884,10 @@ impl Ui {
 #[cfg(debug_assertions)]
 impl Drop for Ui {
     fn drop(&mut self) {
+        if self.should_interact_bg_on_drop {
+            #[allow(deprecated)]
+            self.interact_bg();
```
emilk, line 2889: a why-comment would be useful here

```suggestion
            // Register our final `min_rect`
            self.interact_bg();
```

crates/egui/src/ui_builder.rs
```diff
@@ -116,4 +117,11 @@ impl UiBuilder {
         self.style = Some(style.into());
         self
     }
+
+    /// Sense of the Ui. Should be the same as the one passed to [`Ui::interact_bg`]
```
emilk, line 121:
```suggestion
    /// Set if you want sense clicks and/or drags.
    ///
    /// The response can be read with [`Ui::response`].
```

emilk (approved):
This is utterly awesome ⭐ 

This is something I've been wanting to do in egui for a long time

crates/egui/src/ui.rs
```diff
@@ -2142,7 +2212,10 @@ impl Ui {
         let mut child_ui = self.new_child(ui_builder);
         self.next_auto_id_salt = next_auto_id_salt; // HACK: we want `scope` to only increment this once, so that `ui.scope` is equivalent to `ui.allocate_space`.
         let ret = add_contents(&mut child_ui);
-        let response = self.allocate_rect(child_ui.min_rect(), Sense::hover());
+        #[allow(deprecated)]
+        let response = child_ui.interact_bg();
+        child_ui.should_interact_bg_on_drop = false;
```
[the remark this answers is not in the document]
lucasmerlin: Yeah, I think it's not very elegant but I couldn't come up with a better way. I'll add a comment clarifying that it's an optimization.

crates/egui/src/ui.rs
```diff
@@ -2142,7 +2212,10 @@ impl Ui {
...
+        self.allocate_rect(child_ui.min_rect(), Sense::hover());
```
lucasmerlin, line 2223: I also wonder if we should replace this with `Ui::advance_cursor_after_rect` since we don't need this response?

crates/egui/src/ui.rs
```diff
@@ -2811,6 +2884,10 @@ impl Ui {
 #[cfg(debug_assertions)]
```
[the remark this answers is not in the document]
lucasmerlin: Oops, good catch!

crates/egui/src/ui.rs
emilk: Yeah, makes sense

emilk (approved):

crates/egui/src/ui.rs
emilk: Fixed in https://github.com/emilk/egui/pull/5130

emilk:
The widget id used for the `Ui` is `Ui::id`, which has a problem: is not necessarily unique.

We should instead use an `Id` that is globally unique, i.e. based on where in the hierarchy the `Ui` is, i.e. based on `next_auto_id_salt`. I believe this means we need to store this in `Ui` so we can use the same `Id` in `remember_min_rect`.

To see the problem with the current code, just apply this diff:

``` diff
--- a/crates/egui_demo_lib/src/demo/interactive_container.rs
+++ b/crates/egui_demo_lib/src/demo/interactive_container.rs
@@ -21,6 +21,7 @@ impl crate::Demo for InteractiveContainerDemo {
             .show(ctx, |ui| {
                 use crate::View as _;
                 self.ui(ui);
+                self.ui(ui);
             });
     }
 }
```

David-OConnor:
I am experiencing this when using egui-file-dialog and WGPU together, upon opening a file dialog. I do not get this error with EFRAME directly (no WGPU)

PR #5056, opened 2024-09-01
Conditionally propagate web events using a filter in WebOptions

liamrosenfeld:
Currently egui will prevent all web events from propagating. This causes issues in contexts where you are using egui in a larger web context that wants to receive events that egui does not directly respond to. For example, currently using egui in a VSCode extension will block all app hotkeys, such as saving and opening the panel.

This adds a closure to `WebOptions` that takes in a reference to the egui event that is generated from a web event and returns if the corresponding web event should be propagated or not. The default for it is to always return false.

Alternatives I considered were:
1. Having the propagation filter be a property of the focus in memory. That way it could be configured by the view currently selected. I opted away from that because I wanted to avoid lowering eframe implementation specific stuff into egui.
2. Having events contain a `web_propagate` flag that could be set when handling them. However, that would not be compatible with the current system of egui events being handled outside of the web event handler.

I just recently started using egui so I am not sure how idiomatic my approach here is. I would be happy to switch this over to a different architecture if there are suggestions.

### DIFF
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -456,6 +456,10 @@ pub struct WebOptions {
     ///
     /// Defaults to true.
     pub dithering: bool,
+
+    /// If the web event corresponding to an egui event should be propagated
+    /// to the rest of the web page.
+    pub should_propagate_event: Box<dyn Fn(&egui::Event) -> bool>,
 }
 
 #[cfg(target_arch = "wasm32")]
@@ -471,6 +475,8 @@ impl Default for WebOptions {
             wgpu_options: egui_wgpu::WgpuConfiguration::default(),
 
             dithering: true,
+
+            should_propagate_event: Box::new(|_| false),
         }
     }
 }
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -459,6 +459,10 @@ pub struct WebOptions {
 
     /// If the web event corresponding to an egui event should be propagated
     /// to the rest of the web page.
+    ///
+    /// The default is `false`, meaning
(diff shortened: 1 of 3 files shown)


crates/eframe/src/web/events.rs
```diff
@@ -138,14 +138,18 @@ fn install_keydown(runner_ref: &WebRunner, target: &EventTarget) -> Result<(), J
...
-                    runner.input.raw.events.push(egui::Event::Text(text));
+                    let egui_event = egui::Event::Text(text);
+                    let should_propagate = (runner.web_options.should_propagate_event)(&egui_event);
+                    runner.input.raw.events.push(egui_event);
                     runner.needs_repaint.repaint_asap();
 
                     // If this is indeed text, then prevent any other action.
                     event.prevent_default();
 
                     // Assume egui uses all key events, and don't let them propagate to parent elements.
```
emilk, line 149: This comment should now be updated, here and everywhere

crates/eframe/src/epi.rs
```diff
@@ -456,6 +456,10 @@ pub struct WebOptions {
     ///
     /// Defaults to true.
     pub dithering: bool,
+
+    /// If the web event corresponding to an egui event should be propagated
+    /// to the rest of the web page.
```
emilk, line 461:
```suggestion
    /// to the rest of the web page.
    ///
    /// The default is `false`, meaning
    /// [`stopPropagation`](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation)
    /// is called on every event.
```

emilk (approved):

liamrosenfeld:
I updated the comments. I also added the same comments to other occurrences of the check for consistency.

Issue #4976, opened 2024-08-17
Ability to skip rendering a frame / ask for second pass


ficoos:
**Is your feature request related to a problem? Please describe.**

I would like to be able to specifically skip a frame because I know it will cause a graphical stutter due to sizing layouting.
Being able to do `ctx.skip_frame()` or something similar to prompt the system to skip displaying the next frame and immediately request another repaint.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

emilk:
There's some trickyness to this to figure out:

* We need to avoid an infinite loop, if `skip_frame` is repeatedly called. We should probably only ever skip one frame.
* We need to handle the output of the skipped frame, just skip the rendering
* Should we skip painting and request a new animation frame, or just call `App::update` (or viewport callback) again?

ficoos:
Avoiding an infinite loop is a non-issue IMHO because you can do any number of things to cause an infinite loop when programming. It is also difficult to detect if the skipping is due to the same cause or two different causes that happened to be in consecutive frames. There could maybe be a warning if it is called consecutively in debug by putting this information in a temp.

I think that the simplest thing would be to just skip painting and request a new frame. This makes the semantics almost identical to if the frame had been presented (the current behaviour) which would simplify the integration and backward compatibility.

The major issues with this that I see is actually that building this in to Widgets has a global effect (some operations can cause the UI to take twice as long to update) which some users may prefer jitter to lost frames.

This means it should probably be configurable whether to respect skip_frame requests.

Issue #4378, opened 2024-04-18
Improving layouting


emilk:
## Understanding the problem
One of the main drawback of single-pass immediate mode GUIs is layouting. There is a cyclical dependency between position, size, and layout. This can be illustrated explained by an example:

Say you want to show a small dialog window in the center of the screen. To position the window we must first know the size of it. To know the size of it we must first layout the contents of the window. In immediate mode, the layout code also checks for interaction ("was the OK button clicked?"), thus we need to know _where_ to place the OK button before doing the layout. So we need to know _where_ to put the window before we can layout the contents of the window, but only after laying out the contents of the window can we know the size of the window, and thus where to place it so that it is centered on the screen.

The crux of the issue is really knowing the size of a thing before you can lay it out properly. So, how can we know the size of a thing before calling the layout code for it?

There are simple _atomic_ widgets like `egui::Button` that calculates its size based on the text it contains and the border. This is where egui "bottoms out". These atomic widgets can be properly centered, because they only result in a single call to `ui.allocate*`.

Whenever you have a group of widgets though, you have a problem.

## Figuring out sizes before layout

So how can we know the size of a thing before doing the layout?

There are a few different ways:

A) Using fixed size elements (e.g. decide that the dialog window is always exactly 300x200 no matter what it contains)
B) Remember sizes of elements from previous frames (first frame calculate sizes, all other frames: position things based on those sizes). egui uses this strategy for some stuff (like `Window`, `Grid`, `Table`, …)
C) An API for calculating the size of an element before adding it (https://github.com/emilk/egui/issues/606 - has potential for exponential slowdowns)
D) Multi-pass immediate mode ([rejected](https://github.com/emilk/egui/issues/843#issuecomment-1886968264))

I think B) is the most interesting one to dig deeper into.

Whenever the B) strategy is used, one should be careful not to show the widget during the "layout" frame. For instance, a centered `egui::Window` is invisible for the first frame.

The B) strategy also fails if the thing changes size over time. Though it is self correcting, it has a frame-delay. So in the original example: if the dialog windows grows it will shift to the right for one frame, then re-center then next frame. This will look ugly.

## Improving layout given sizes

Once you have the size you should be able to apply any advanced layouting technique. For instance using:

* Explicit math
* `egui_extras::Strip`
*  https://crates.io/crates/taffy

Here we have a lot of opportunity for improvement in egui.

YgorSouza:
[...]

PR #5059, opened 2024-09-02
Add `Context::request_discard`

emilk:
* Closes https://github.com/emilk/egui/issues/4976
* Part of #4378 
* Implements parts of #843

### Background
Some widgets (like `Grid` and `Table`) needs to know the width of future elements in order to properly size themselves. For instance, the width of the first column of a grid may not be known until all rows of the grid has been added, at which point it is too late. Therefore these widgets store sizes from the previous frame. This leads to "first-frame jitter", were the content is placed in the wrong place for one frame, before being accurately laid out in subsequent frames.

### What
This PR adds the function `ctx.request_discard` which discards the visual output and does another _pass_, i.e. calls the whole app UI code once again (in eframe this means calling `App::update` again). This will thus discard the shapes produced by the wrongly placed widgets, and replace it with new shapes. Note that only the visual output is discarded - all other output events are accumulated.

Calling `ctx.request_discard` should only be done in very rare circumstances, e.g. when a `Grid` is first shown. Calling it every frame will mean the UI code will become unnecessarily slow.

Two safe-guards are in place:

* `Options::max_passes` is by default 2, meaning egui will never do more than 2 passes even if `request_discard` is called on every pass
* If multiple passes is done for multiple frames in a row, a warning will be printed on the screen in debug builds:

image

### Breaking changes
A bunch of things that had "frame" in the name now has "pass" in them instead:

* Functions called `begin_frame` and `end_frame` are now called `begin_pass` and `end_pass`
* `FrameState` is now `PassState`
* etc

### TODO
* [x] Figure out good names for everything (`ctx.request_discard`)
* [x] Add API to query if we're gonna repeat this frame (to early-out from expensive rendering)
* [x] Clear up naming confusion (pass vs frame) e.g. for `FrameState`
* [x] Figure out when to call this
* [x] Show warning on screen when there are several frames in a row with multiple passes
* [x] Document
* [x] Default on or off?
* [x] Change `Context::frame_nr` name/docs
* [x] Rename `Context::begin_frame/end_frame` and deprecate the old ones
* [x] Test with Rerun
* [x] Document breaking changes


juancampa:
Very simple and elegant solution. I appreciate the guardrails too (frame limit and warning) 👌

Issue #2152, opened 2022-10-14
Allow the use of non-static App


Gnurfos:
I have some struct offering a view over something, not owning but storing a reference to that something:

```rust
struct Reader<'a> {
    buffer: &'a[u8],
}

impl<'a> Reader<'a> {
    pub fn get(&self, i: usize) -> u8 {
        self.buffer[i]
    }
}
```

I'd like to use that reader to run an egui app:

```rust
struct ViewerApp<'a> {
    reader: Reader<'a>,
}

impl<'a> eframe::App for ViewerApp<'a> {
    fn update(&mut self, ctx: &egui::Context, _: &mut eframe::Frame) {
        eframe::egui::CentralPanel::default()
            .frame(eframe::egui::Frame::none())
            .show(ctx, |ui| {
                ui.label(format!("Buffer slot 0: {:?}", self.reader.get(0)));
            });
    }
}
```
However, from the way `AppCreator` is declared, I think it can only accept things with `'static` lifetimes:
```rust
fn main() {
    let buffer = vec![0u8; 100];
    let reader = Reader { buffer: &buffer };
    visualize(reader);
}

fn visualize<'a>(reader: Reader<'a>) {
    eframe::run_native(
        "Buffer visualizer",
        eframe::NativeOptions::default(),
        Box::new(|_| Box::new(ViewerApp { reader })),
    );
}
```
Fails with:
>  |         Box::new(|_| Box::new(ViewerApp { reader })),
>  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cast requires that `'a` must outlive `'static`

I tried various workarounds but the only way I could "solve" this was by doing:
```rust
    let reader = unsafe {
        std::mem::transmute::<_, Reader<'static>>(reader)
    };
```
to "cast away" the lifetime, assuming that the items are alive at least until run_native returns.
It appears to work, but I'm not sure it's safe in every situation.

Ideally, run_native would accept something with lifetimes, so that the code above just works.
Alternatively, but I'm just guessing, this could be achieved by replacing AppCreator by a trait other than FnOnce.

I have no idea if either of these is feasible, or if that would pollute the api in unacceptable ways.

Speak2Erase:
Take a look at `Box::leak`?

Gnurfos:
I think I tried but it doesn't change the lifetime issue. And leaking is not great, now that run_native can return and be called several times in an app.

[...]

PR #5060, opened 2024-09-02
Allow non-`static` `eframe::App` lifetime

timstr:
I love egui! Thank you Emil <3

This request specifically enables an `eframe::App` which stores a lifetime.

In general, I believe this is necessary because `eframe::App` currently does not seem to provide a good place to allocate and then borrow from long-lived data between `update()` calls. To attempt to borrow such long-lived data from a field of the `App` itself would be to create a self-referential struct. A hacky alternative is to allocate long-lived data with `Box::leak`, but that's a code smell and would cause problems if a program ever creates multiple Apps.

As a more specific motivating example, I am developing with the [inkwell](https://github.com/TheDan64/inkwell/) crate which requires creating a `inkwell::context::Context` instance which is then borrowed from by a bazillion things with a dedicated `'ctx` lifetime. I need such a `inkwell::context::Context` for the duration of my `eframe::App` but I can't store it as a field of the app. The most natural solution to me is to simply to lift the inkwell context outside of the App and borrow from it, but that currently fails because the AppCreator implicitly has a `'static` lifetime requirement due to the use of `dyn` trait objects.

Here is a simpler, self-contained motivating example adapted from the current [hello world example](https://docs.rs/eframe/latest/eframe/):

```rust
use eframe::egui;

struct LongLivedThing {
    message: String,
}

fn main() {
    let long_lived_thing = LongLivedThing {
        message: "Hello World!".to_string(),
    };

    eframe::run_native(
        "My egui App",
        native_options,
        Box::new(|cc| Ok(Box::new(MyEguiApp::new(cc, &long_lived_thing)))),
        //                                           ^^^^^^^^^^^^^^^^^
        //                                           BORROWING from long_lived_thing in App
    );
}

struct MyEguiApp<'a> {
    long_lived_thing: &'a LongLivedThing,
}

impl<'a> MyEguiApp<'a> {
    fn new(cc: &eframe::CreationContext<'_>, long_lived_thing: &'a LongLivedThing) -> Self {
        MyEguiApp { long_lived_thing }
    }
}

impl<'a> eframe::App for MyEguiApp<'a> {
            ui.heading(&self.long_lived_thing.message);
        });
    }
}
```

This currently fails to compile with:
```plaintext
error[E0597]: `long_lived_thing` does not live long enough
  --> src/main.rs:16:55
   |
16 |         Box::new(|cc| Ok(Box::new(MyEguiApp::new(cc, &long_lived_thing)))),
   |         ----------------------------------------------^^^^^^^^^^^^^^^^----
   |         |        |                                    |
   |         |        |                                    borrowed value does not live long enough
   |         |        value captured here
   |         cast requires that `long_lived_thing` is borrowed for `'static`
17 |     );
18 | }
   | - `long_lived_thing` dropped here while still borrowed
   |
   = note: due to object lifetime defaults, `Box<dyn for<'a, 'b> FnOnce(&'a CreationContext<'b>) -> Result<Box<dyn App>, Box<dyn std::error::Error + Send + Sync>>>` actually means `Box<(dyn for<'a, 'b> FnOnce(&'a CreationContext<'b>) -> Result<Box<dyn App>, Box<dyn std::error::Error + Send + Sync>> + 'static)>`
```

With the added lifetimes in this request, I'm able to compile and run this as expected on Ubuntu + Wayland. I see the CI has been emailing me about some build failures and I'll do what I can to address those. Currently running the check.sh script as well.

This is intended to resolve https://github.com/emilk/egui/issues/2152

* Closes https://github.com/emilk/egui/issues/2152

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -48,24 +48,24 @@ use super::{
 // ----------------------------------------------------------------------------
 // Types:
 
-pub struct GlowWinitApp<'a> {
+pub struct GlowWinitApp<'app> {
     repaint_proxy: Arc<egui::mutex::Mutex<EventLoopProxy<UserEvent>>>,
     app_name: String,
     native_options: NativeOptions,
-    running: Option<GlowWinitRunning<'a>>,
+    running: Option<GlowWinitRunning<'app>>,
 
     // Note that since this `AppCreator` is FnOnce we are currently unable to support
     // re-initializing the `GlowWinitRunning` state on Android if the application
     // suspends and resumes.
-    app_creator: Option<AppCreator<'a>>,
+    app_creator: Option<AppCreator<'app>>,
 }
 
 /// State that is initialized when the application is first starts running via
 /// a Resumed event. On Android this ensures that any graphics state is only
 /// initialized once the application has an associated `SurfaceView`.
-struct GlowWinitRunning<'a> {
+struct GlowWinitRunning<'app> {
     integration: EpiIntegration,
-    app: Box<dyn 'a + App>,
+    app: Box<dyn 'app + App>,
 
     // These needs to be shared with the immediate viewport renderer, hence the Rc/Arc/RefCells:
     glutin: Rc<RefCell<GlutinWindowContext>>,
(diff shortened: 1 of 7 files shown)


crates/eframe/src/native/glow_integration.rs
```diff
@@ -48,24 +48,24 @@ use super::{
 // ----------------------------------------------------------------------------
 // Types:
 
-pub struct GlowWinitApp {
+pub struct GlowWinitApp<'a> {
```
emilk, line 51: Let's be explicit and call this lifetime `'app` here and everywhere

emilk (approved):

crates/eframe/src/native/glow_integration.rs
timstr: Good idea, done.

Issue #5086, opened 2024-09-07
optimize Color32::from_rgba_unmultiplied


veyh:
I noticed displaying images with alpha channels was a bit slower than I expected so I attached a profiler and saw that a considerable amount of time was spent in `Color32::from_rgba_unmultiplied` (though it shows up as `ColorImage::from_rgba_unmultiplied` because it's inlined).

![image](https://i.imgur.com/mgqlM5e.png)

So I looked at the code and basically the [else branch](https://github.com/emilk/egui/blob/master/crates/ecolor/src/color32.rs#L97) is the problem. But since were working with a limited number of values, couldn't we just compute them in advance and use a lookup table instead?

I tried this out and saw a massive performance boost as a result. So I was just wondering if you see anything wrong with it and whether you'd like me to submit a PR (or you can just take these changes and apply them yourself).

This is basically what I did

```rust
use once_cell::sync::Lazy;

static GAMMA_U8_FROM_LINEAR_F32_LUT: Lazy<[u8; 256 * 256]> = Lazy::new(|| {
    let mut values: [u8; 256 * 256] = [0; 256 * 256];

    for color in 0..=255 {
        for alpha in 0..=255 {
            values[(color * 256) + alpha] = gamma_u8_from_linear_f32(
                linear_f32_from_gamma_u8(color as u8) *
                linear_f32_from_linear_u8(alpha as u8)
            );
        }
    }

    values
});

impl Color32 {
    pub fn from_rgba_unmultiplied(r: u8, g: u8, b: u8, a: u8) -> Self {
        let r = unsafe {
            *GAMMA_U8_FROM_LINEAR_F32_LUT.get_unchecked(((r as usize) * 256) + (a as usize))
        };

        let g = unsafe {
            *GAMMA_U8_FROM_LINEAR_F32_LUT.get_unchecked(((g as usize) * 256) + (a as usize))
        };

        let b = unsafe {
            *GAMMA_U8_FROM_LINEAR_F32_LUT.get_unchecked(((b as usize) * 256) + (a as usize))
        };

        Self::from_rgba_premultiplied(r, g, b, a)
    }
}
```

YgorSouza:
I made a version that pulls the lookup table from a file at compile time instead, and uses just regular indexing without unsafe, to see how it goes. It seems to be as fast as the `0` and `255` cases from the current version.

veyh:
Looks good!

PR #5088, opened 2024-09-07
Use precomputed lookup table in Color32::from_rgba_unmultiplied

YgorSouza:
Improves performances significantly (about 40 times) according to the benchmarks.

* Closes <https://github.com/emilk/egui/issues/5086>


YgorSouza:
Before the lookup table:

```
from_rgba_unmultiplied_0
                        time:   [148.77 ps 149.04 ps 149.36 ps]

from_rgba_unmultiplied_100
                        time:   [63.772 ns 63.793 ns 63.813 ns]

from_rgba_unmultiplied_255
                        time:   [147.14 ps 147.31 ps 147.48 ps]
```

After the lookup table:

```
from_rgba_unmultiplied_0
                        time:   [137.89 ps 139.49 ps 141.31 ps]
                        change: [-4.1652% -3.2486% -2.3279%] (p = 0.00 < 0.05)
                        Performance has improved.

from_rgba_unmultiplied_100
                        time:   [128.12 ps 128.64 ps 129.17 ps]
                        change: [-99.799% -99.798% -99.797%] (p = 0.00 < 0.05)
                        Performance has improved.

from_rgba_unmultiplied_255
                        time:   [136.86 ps 137.31 ps 137.72 ps]
                        change: [-7.7432% -7.2489% -6.7624%] (p = 0.00 < 0.05)
                        Performance has improved.
```

YgorSouza:
Note that this will increase the binary size, of course, which is bad for WASM in particular. The alternative would be to build the lookup table at runtime, as suggested in the issue. The downsides are that it needs a dependency (`once_cell`, which is already used in a lot of places and can be removed once the MSRV moves to 1.80) and that the first call would be very slow.

Yet another possibility would be to build a lookup table only for `linear_f32_from_gamma_u8` instead, which would be only 1KB, and could be used both ways (with `partition_point`). `powf` seems to be the bottleneck according to the profile posted in the issue, so maybe that would be good enough. I'm going to give it a try.

veyh:
Isn't building the lookup table on the fly be pretty much the same as (currently) displaying a 256x256 image with transparency? I would think it's not really _that_ slow.

Alternatively, the 1KB table could be used for WASM and the 64KB table for other platforms? (If there is enough of a difference between the two.)

YgorSouza:
> Yet another possibility would be to build a lookup table only for `linear_f32_from_gamma_u8` instead, which would be only 1KB, and could be used both ways (with `partition_point`). `powf` seems to be the bottleneck according to the profile posted in the issue, so maybe that would be good enough. I'm going to give it a try.

So I tried reverting the previous change and adding this:

```rs
#[inline(always)]
pub fn linear_f32_from_gamma_u8(s: u8) -> f32 {
    lookup::LINEAR_F32_FROM_GAMMA_U8[usize::from(s)]
}

#[inline(always)]
pub fn gamma_u8_from_linear_f32(l: f32) -> u8 {
    let threshold = (l - 1e-4).clamp(0.0, 1.0);
    let lower_bound = (threshold * 255.0) as u8;
    (lower_bound..u8::MAX)
        .find(|&s| linear_f32_from_gamma_u8(s) >= threshold)
        .unwrap_or(u8::MAX)
}
```

With a `[f32; 256]` array I generated and pasted directly in the code. The `- 1e-4` is needed for the unit tests to pass, because of rounding errors. I got this result on the bench:

```
from_rgba_unmultiplied_0
                        time:   [127.57 ps 127.86 ps 128.14 ps]

from_rgba_unmultiplied_100
                        time:   [15.520 ns 15.545 ns 15.570 ns]

from_rgba_unmultiplied_255
                        time:   [134.04 ps 134.38 ps 134.68 ps]
```

Which is only about 4 times faster than the current implementation on master, and a lot slower than the version with the big lookup table that is currently in this PR. I tried to use `partition_point` instead of a linear search, but it was even slower. I'm going to see if I can come up with something better, otherwise I'll just leave it as it is currently, or try the `once_cell` approach.

Also, I realized that my benchmarks aren't right, because I'm not using `black_box` on the inputs, which is allowing the compiler to do more work at compile time than it's supposed to. I'm going to fix that and edit all the benchmarks I've posted so far with the corrected values.

YgorSouza:
I added the `once_cell` version. The benchmark went up to 2.9ns (from 1.5ns with the array in program memory). I guess dereferencing the `Lazy` isn't completely free. It's still 20 times faster than the current implementation. I left it as a separate commit so it can be easily reverted depending on which tradeoff is preferred. And I think I'm not going to bother with the f32 lookup table for the gamma value I mentioned above. Too finicky, and I couldn't get it to perform well enough to be worth the trouble.

crates/ecolor/src/color32.rs
```diff
@@ -95,21 +105,17 @@ impl Color32 {
...
-            let g_lin = linear_f32_from_gamma_u8(g);
-            let b_lin = linear_f32_from_gamma_u8(b);
-            let a_lin = linear_f32_from_linear_u8(a);
-
-            let r = gamma_u8_from_linear_f32(r_lin * a_lin);
-            let g = gamma_u8_from_linear_f32(g_lin * a_lin);
-            let b = gamma_u8_from_linear_f32(b_lin * a_lin);
-
-            Self::from_rgba_premultiplied(r, g, b, a)
+        match a {
```
veyh, line 108: Might be worth checking if skipping the match for the 0/255 cases is actually faster. At least on my machine, it is.

With match
```
from_rgba_unmultiplied_100
                        time:   [2.0122 ns 2.0443 ns 2.0813 ns]
```

Just using the table
```
from_rgba_unmultiplied_100
                        time:   [1.6584 ns 1.6675 ns 1.6772 ns]
                        change: [-18.696% -17.776% -16.940%] (p = 0.00 < 0.05)
                        Performance has improved.
```

YgorSouza: Yes, but the common cases become slower, so I decided to just leave it as it was before.

veyh: Yeah, that is true.

crates/ecolor/src/color32.rs
```diff
@@ -1,6 +1,16 @@
...
-    fast_round, gamma_u8_from_linear_f32, linear_f32_from_gamma_u8, linear_f32_from_linear_u8, Rgba,
-};
+use once_cell::sync::Lazy;
+
+use crate::{fast_round, linear_f32_from_linear_u8, Rgba};
+
+static FROM_UNMULTIPLIED_LUT: Lazy<[u8; 256 * 256]> = Lazy::new(|| {
+    use crate::{gamma_u8_from_linear_f32, linear_f32_from_gamma_u8};
+    core::array::from_fn(|i| {
+        let [value, alpha] = (i as u16).to_be_bytes();
```
emilk, line 8: I doubt it will affect performance at all, but little-endian feels like a more natural choice (considering basically all CPUs are LE these days)

crates/epaint/benches/benchmark.rs
```diff
@@ -223,6 +223,36 @@ fn thin_large_line_uv(c: &mut Criterion) {
...
+        b.iter(|| {
+            let color = ecolor::Color32::from_rgba_unmultiplied(r, g, bl, a);
+            black_box(color);
+        });
+    });
+}
+
+fn from_rgba_unmultiplied_100(c: &mut Criterion) {
+    c.bench_function("from_rgba_unmultiplied_100", move |b| {
+        let [r, g, bl, a] = black_box([20u8, 20, 20, 100]);
```
emilk, line 238: Thanks for adding a proper benchmark for this! I think a more fair benchmark would test a bunch of _different_ color values though. Now it will hit the same place in the LUT every time, so it will always be a cache hit.

emilk (approved):
Thanks for working on this! I think it is worth avoiding adding 64kB to the .wasm size.

I think you should be able to replace `once_cell` with [`std::cell::OnceCell`](https://doc.rust-lang.org/std/cell/struct.OnceCell.html#method.get_or_init)

YgorSouza:
Updated benchmarks:

(Note that this now refers to the time to call the function 1000 times with different values)

### Without LUT
```
from_rgba_unmultiplied_0
                        time:   [416.57 ns 419.80 ns 422.97 ns]

from_rgba_unmultiplied_other
                        time:   [75.650 µs 75.769 µs 75.856 µs]

from_rgba_unmultiplied_255
                        time:   [1.0988 µs 1.1004 µs 1.1020 µs]
```

### With LUT
```
from_rgba_unmultiplied_0
                        time:   [501.18 ns 501.74 ns 502.29 ns]
                        change: [+20.148% +21.106% +21.946%] (p = 0.00 < 0.05)
                        Performance has regressed.

from_rgba_unmultiplied_other
                        time:   [1.6264 µs 1.6273 µs 1.6281 µs]
                        change: [-97.868% -97.856% -97.846%] (p = 0.00 < 0.05)
                        Performance has improved.

from_rgba_unmultiplied_255
                        time:   [1.1058 µs 1.1067 µs 1.1075 µs]
                        change: [+0.2530% +0.4257% +0.5968%] (p = 0.00 < 0.05)
```

The 0 case is slightly slower now, not sure why. But it is still way faster than the others.

YgorSouza:
> Thanks for working on this! I think it is worth avoiding adding 64kB to the .wasm size.
> 
> I think you should be able to replace `once_cell` with [`std::cell::OnceCell`](https://doc.rust-lang.org/std/cell/struct.OnceCell.html#method.get_or_init)

`OnceCell` can't be used in `static` (not thread-safe), but I used `OnceLock`. It's actually twice as fast as the once_cell crate too.

emilk (approved):

Issue #3549, opened 2023-11-13
Use Style{override_font_id} to specify the font for syntax_highlighting::highlight()


srgfrog:
Currently the FontId when creating a TextArea for editing code, using syntect is hardcoded to egui::FontId::monospace(12.0). At the moment, I'm using the workaround 

```let mut layouter = |ui: &egui::Ui, string: &str, wrap_width: f32| {
    let theme = egui_extras::syntax_highlighting::CodeTheme::default();
    let mut layout_job =
        egui_extras::syntax_highlighting::highlight(ui.ctx(), &theme, string, "rs");
    layout_job.wrap.max_width = wrap_width;
    layout_job.sections = layout_job 
        .sections 
        .iter()
        .map(|layout_section| {
            let mut section = layout_section.clone();
            section.format.font_id = FontId::monospace(16.0);
            section
        })
        .collect();
    ui.fonts(|f| f.layout_job(layout_job))
};
```

I tried earlier using 
```
egui_extras::syntax_highlighting::CodeTheme::from_style(mystyle);
```
but ```mystyle.override_font_id = Some(FontId::monospace(16.0));``` is ignored.

Looking at the code, enabling the Style{override_font_id} field for the highlight() function means adding a field to the CodeTheme{} and modifying fn highlight_impl() to use the specified font.

Please let me know if I have overlooked an easier way.

I don't have write access to the repository, so I think I can't open a pull request. I've attached a ```git diff``` for the changes. Please let me know if there is a better way. Thank you.
[syntax_highlighting_git_diff.txt](https://github.com/emilk/egui/files/13342967/syntax_highlighting_git_diff.txt)

PR #5090, opened 2024-09-08
Use Style's font size in egui_extras::syntax_highlighting

lampsitter:
* Closes https://github.com/emilk/egui/issues/3549

The syntax highlighting font size was always hardcoded to 12 or 10 depending on what case it was hitting (so not consistent). This is particularly noticeable when you increase the font size to something larger for the rest of the ui.

With this the default monospace font size is used by default.

Since the issue is closely related to  #3549 I decided to implement the ability to use override_font_id too.

## Visualized

Default monospace is set to 15 in all the pictures

Before/After without syntect:
normal

Before/after _with_ syntect:
syntect

Font override after without/with syntect (monospace = 20):
override

### Breaking changes

- `CodeTheme::dark` and `CodeTheme::light` takes in the font size
- `CodeTheme::from_memory` takes in `Style`
- `highlight` function takes in `Style`

### DIFF
--- a/crates/egui_extras/src/syntax_highlighting.rs
+++ b/crates/egui_extras/src/syntax_highlighting.rs
@@ -15,14 +15,24 @@ pub fn code_view_ui(
     code: &str,
     language: &str,
 ) -> egui::Response {
-    let layout_job = highlight(ui, theme, code, language);
+    let layout_job = highlight(ui.ctx(), ui.style(), theme, code, language);
     ui.add(egui::Label::new(layout_job).selectable(true))
 }
 
 /// Add syntax highlighting to a code string.
 ///
 /// The results are memoized, so you can call this every frame without performance penalty.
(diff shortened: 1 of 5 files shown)


crates/egui_extras/src/syntax_highlighting.rs
```diff
@@ -128,47 +153,84 @@ pub struct CodeTheme {
...
 
+    /// ### Example
+    ///
+    /// ```
+    /// # egui::__run_test_ui(|ui| {
+    /// use egui_extras::syntax_highlighting::CodeTheme;
+    /// let theme = CodeTheme::dark(12.0);
+    /// # });
+    /// ```
+    pub fn dark(font_size: f32) -> Self {
```
lampsitter, line 192: Added doc example to ease the impact of the breaking change

emilk (approved):

Issue #5079, opened 2024-09-06
If the contents of the popup window are incorrectly wrapped, it will continue to be displayed incorrectly wrapped.


rustbasic:
If the contents of a popup window displayed as `on_hover_text_at_pointer()`  are incorrectly wrapped(change line) due to other influences, they will not be reset and will continue to be displayed incorrectly wrapped(change line).

( I don't have any plans to work on this yet. )

![20240907-bug1](https://github.com/user-attachments/assets/40a07848-d0aa-437c-a402-ffdeeefa3635)

rustbasic:
#5076 #5077 After update, Wrap are happening more and more strangely. 
In case of `on_hover_text_at_pointer()`, more and more Wrap are added every time the mouse is moved.

rustbasic:
Someone, please refer to #5093 and fix the bug correctly by referring to this part.
I'm working on something else so I can't look into it in detail.

PR #5093, opened 2024-09-09
Fix: more and more Wrap are added every time the mouse is moved.

### LINKED ISSUE #5084: #5077 After update, more and more Wrap are added every time the mouse is moved.
#5076 #5077 After update (May be #5077), 
wrap(line breaks) are happening more and more strangely.
In case of `on_hover_text_at_pointer()`, more and more Wrap(line breaks) are added every time the mouse is moved.

rustbasic:
Fix Issues: more and more Wrap are added every time the mouse is moved.

* Related #5077
* Related #5079
* Closes #5084

`Here is an example of how this would solve the issue.`

Adding a value greater than `2.0` will prevent the `popup(=tooltip)` size from changing when you move the mouse.
If you add `16.0` ~ `20.0`, tooltips that were affected by other `tooltip`s will return to their original size.
If you add `25.0` or more, the size of the `tooltip` will continue to grow until there is no wrap.
May be, It seems to be affected by `line size` and `inner margin` of `tooltip`.

### DIFF
--- a/crates/epaint/src/text/text_layout_types.rs
+++ b/crates/epaint/src/text/text_layout_types.rs
@@ -182,7 +182,7 @@ impl LayoutJob {
             // On a previous pass we may have rounded down by at most 0.5 and reported that as a width.
             // egui may then set that width as the max width for subsequent frames, and it is important
             // that we then don't wrap earlier.
-            self.wrap.max_width + 2.0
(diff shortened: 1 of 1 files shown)


emilk:
This is not the right way to fix this

PR #5107, opened 2024-09-13
Fix: Popups do not appear in certain situations.

### LINKED ISSUE #5080: After using about 45,000 popups, the popups will not be displayed.
In `FontBook`,
If you use a font with about 15,000 characters, the popup (`on_hover_ui_at_pointer()`) is displayed properly.
However, if you use a font with about 45,000 characters, the popup is not displayed.

When the popup is not displayed, the popup of the entire program is not displayed.
If you close `FontBook`, the popup is displayed again.

rustbasic:
Fix: Popups do not appear in certain situations.

* Closes #5080

The root cause is that `last_move_time` is not updated in certain situations (slow situations).

This might also be useful for drag etc.

```
        if clicked_more_recently_than_moved && (self.clicked() || self.is_pointer_button_down_on())
```

Alternatively, we can apply a specific time (0.5 to 1.0 seconds).

```
        if clicked_more_recently_than_moved && time_since_last_click < 1.0
```

### DIFF
--- a/crates/egui/src/input_state/mod.rs
+++ b/crates/egui/src/input_state/mod.rs
@@ -932,6 +932,7 @@ impl PointerState {
                             press_origin.distance(pos) > self.input_options.max_click_dist;
                     }
 
+                    self.last_move_time = time;
                     self.pointer_events.push(PointerEvent::Moved(pos));
                 }
                 Event::PointerButton {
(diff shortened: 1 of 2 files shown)


crates/egui/src/input_state/mod.rs
```diff
@@ -1028,15 +1029,6 @@ impl PointerState {
...
-        self.velocity = if self.pos_history.len() >= 3 && self.pos_history.duration() > 0.01 {
-            self.pos_history.velocity().unwrap_or_default()
-        } else {
-            Vec2::default()
-        };
-        if self.velocity != Vec2::ZERO {
-            self.last_move_time = time;
-        }
-
         self.direction = self.pos_history.velocity().unwrap_or_default().normalized();
```
rustbasic, line 1032: The root cause is that last_move_time is not updated in certain situations (slow situations).

PR #5108, opened 2024-09-13
early out in `should_show_hover_ui()`

rustbasic:
`should_show_hover_ui()` only works on hover, so it's better to come out early.
Is there something different from what you intended or is there a problem?

### DIFF
--- a/crates/egui/src/response.rs
+++ b/crates/egui/src/response.rs
@@ -599,6 +599,9 @@ impl Response {
         if self.ctx.memory(|mem| mem.everything_is_visible()) {
             return true;
         }
+        if !self.hovered() {
+            return false;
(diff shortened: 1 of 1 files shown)


lucasmerlin:
I think this is already handled in line 720: 
```rust
        // Fast early-outs:
        if self.enabled {
            if !self.hovered || !self.ctx.input(|i| i.pointer.has_pointer()) {
                return false;
            }
        } else if !self.ctx.rect_contains_pointer(self.layer_id, self.rect) {
            return false;
        }
```
And I believe the reason the check is only done if self.enabled is that otherwise `on_disabled_hover_ui` wouldn't work, because things that aren't enabled don't get the hover event

PR #5109, opened 2024-09-14

### LINKED ISSUE #5289: When moving with the animation function of `ScrollArea`, it does not move to the exact location.
When moving with the animation function of `ScrollArea`, it does not move to the exact location.

rustbasic:

Pull requests :
* Related #4303 
* Related #4305
* Related #4584
* Closes #5289
* Closes #5290


lucasmerlin:
The current behavior matches the behavior documented in scroll_with_delta: 

> Scroll this many points in the given direction, in the parent [`crate::ScrollArea`].

I think the current behavior makes sense and if you want to scroll the parent ScrollArea it should be possible by calling scroll_with_delta outside of the child ScrollArea.

rustbasic:
> I think the current behavior makes sense and if you want to scroll the parent ScrollArea it should be possible by calling scroll_with_delta outside of the child ScrollArea.

It doesn't work when horizontal() and vertical() are separated.
Because the value is retrieved with take() and the value disappears,


```
ui.horizontal() {
  ui.vertical() {
    ScrollArea::vertical() {
      ui.label()
      ScrollArea::horizontal() {
      }
    }
  }
}
```

For example, if each line in TextEdit is numbered.

Issue #4929, opened 2024-08-06
Improve vertical centering of text (e.g. in buttons)


abey79:
The optical vertical alignment of text is not perfect with button and selectable label, especially with the Inter font we use at rerun. The text here should be ~1-2px lower compared to the background. Unfortunately, we dont have separate control on the top/bottom padding there.

<img width="456" alt="image" src="https://github.com/user-attachments/assets/92602347-802f-4adf-94a3-8032c4babfca">

I'm not sure if there is something in the font metadata that could be used to adjust alignment.

An escape hatch would be provided by making `button_padding` a `Margin` instead of a `Vec2`.

emilk:
You can manually tweak this when you set a font: https://docs.rs/epaint/latest/epaint/text/struct.FontTweak.html

…but we should be able to do better by default too

PR #5117, opened 2024-09-17
Center-align all text vertically

emilk:
* Closes https://github.com/emilk/egui/issues/4929
* Builds on top of https://github.com/emilk/egui/pull/2724 by @lictex (ptal!)
* Implement `Center` and `Max` vertical text alignment properly
* Change default vertical alignment of text to centering

The end result is that text centers better in buttons and other places, especially when mixing in emojis.
Before, mixing text of different heights (e.g. emojis and latin text) in a label or button would cause the text to jump vertically.

## Before
This is `master`, with custom `FontTweak` to move fonts up and down:
image
image

## After
This PR, with the default (zero) `FontTweak`

image

image

### DIFF
--- a/crates/egui/src/style.rs
+++ b/crates/egui/src/style.rs
@@ -199,7 +199,7 @@ pub struct Style {
 
     /// How to vertically align text.
     ///
-    /// Default depends on layout.
+    /// Set to `None` to use align that depends on the current layout.
     pub override_text_valign: Option<Align>,
 
     /// The [`FontFamily`] and size you want to use for a specific [`TextStyle`].
@@ -1207,7 +1207,7 @@ impl Default for Style {
         Self {
             override_font_id: None,
             override_text_style: None,
-            override_text_valign: None,
+            override_text_valign: Some(Align::Center),
             text_styles: default_text_styles(),
             drag_value_text_style: TextStyle::Button,
             number_formatter: NumberFormatter(Arc::new(emath::format_with_decimals_in_range)),
--- a/crates/egui/src/style.rs
+++ b/crates/egui/src/style.rs
@@ -4,6 +4,7 @@
 
 use std::{collections::BTreeMap, ops::RangeInclusive, sync::Arc};
 
+use emath::Align;
 use epaint::{text::FontTweak, Rounding, Shadow, Stroke};
 
 use crate::{
@@ -195,6 +196,11 @@ pub struct Style {
     /// On most widgets you can also set an explicit text style,
     /// which will take precedence over this.
     pub override_font_id: Option<FontId>,
+
+    /// How to vertically align text.
+    ///
+    /// Default depends on layout.
+    pub override_text_valign: Option<Align>,
 
     /// The [`FontFamily`] and size you want to use for a specific [`TextStyle`].
     ///
@@ -1201,6 +1207,7 @@ impl Default for Style {
         Self {
             override_font_id: None,
             override_text_style: None,
+            override_text_valign: None,
             text_styles: default_text_styles(),
             drag_value_text_style: TextStyle::Button,
             number_formatter: NumberFormatter(Arc::new(emath::format_with_decimals_in_range)),
@@ -1501,6 +1508,7 @@ impl Style {
         let Self {
             override_font_id,
             override_text_style,
+            override_text_valign,
             text_styles,
             drag_value_text_style,
             number_formatter: _, // can't change callbacks in the UI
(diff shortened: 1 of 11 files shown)


lictex:
ah mixing fonts is always that hard.. :(

but this looks like still break things in some rare cases
|  |  |
| -- | -- |
|image| master (baseline aligned)<br>`sourcehansans` is higher than `cantarell` and it pushes the baseline down,<br>while characters are still aligned properly |
|image| this pr (center aligned)<br>...misaligned,<br> seems that glyph center of some fonts is just not the visual center |

----

> Before, mixing text of different heights (e.g. emojis and latin text) in a label or button would cause the text to jump vertically.

wondering if sth like below could make things better

```diff
diff --git a/crates/epaint/src/text/font.rs b/crates/epaint/src/text/font.rs
index 5a119cdc..3135d870 100644
--- a/crates/epaint/src/text/font.rs
+++ b/crates/epaint/src/text/font.rs
@@ -473,6 +473,10 @@ impl Font {
         }
         None
     }
+
+    pub fn ascent(&self) -> f32 {
+        self.fonts.first().map_or(0., |f| f.ascent())
+    }
 }
 
 /// Code points that will always be invisible (zero width).
diff --git a/crates/epaint/src/text/text_layout.rs b/crates/epaint/src/text/text_layout.rs
index 4fa3ba49..4cd22dcb 100644
--- a/crates/epaint/src/text/text_layout.rs
+++ b/crates/epaint/src/text/text_layout.rs
@@ -172,7 +172,7 @@ fn layout_section(
                 chr,
                 pos: pos2(paragraph.cursor_x, f32::NAN),
                 size: vec2(glyph_info.advance_width, line_height),
-                ascent: font_impl.map_or(0.0, |font| font.ascent()), // Failure to find the font here would be weird
+                ascent: font.ascent(),
                 uv_rect: glyph_info.uv_rect,
                 section_index,
             });
```

this way only metrics from the primary (first) font are used during layout

emilk:
It doesn't really help. I think the problem is mixing fonts that don't properly center the text in them.

We could center on the _visual_ bounds, but the problem with that would be that text would move vertically when you add new letters to a `TextEdit`.

I think the code in this PR is still an improvement. You can tweak the problematic fonts on load with `FontTweak::y_offset_factor`

lictex:
> but the problem with that would be that text would move vertically when you add new letters to a TextEdit.

with a fixed ascent value i don t really get how this *move* could happen..

did a quick test and the results are:

| `master` | `master` with diff above |
| ---- | ---- |
|image|image|
|<video src="https://github.com/user-attachments/assets/66b0235d-e864-436d-bee4-26a08e00fcd5"/>|<video src="https://github.com/user-attachments/assets/0c3b3108-5826-4e29-a585-1402ca639420"/>|

emilk:
You misunderstand me – your diff does not move the text, but it also does not solve all centering issues. Here is egui with this PR plus you diff, using Cantarell:

image

Not great.

lictex:
:open_mouth: i really got some different results.. 
```diff
+        font_data.insert(
+            "Cantarell".to_owned(),
+            FontData::from_static(include_bytes!("/usr/share/fonts/truetype/Cantarell-VF.otf")),
+        );
         families.insert(
             FontFamily::Proportional,
             vec![
-                "Ubuntu-Light".to_owned(),
+                "Cantarell".to_owned(),
                 "NotoEmoji-Regular".to_owned(),
                 "emoji-icon-font".to_owned(),
```
image

----

and if i remove all the `FontTweak` on default fonts..
image
while it doesn t look that good, it renders just like in my web browser (msedge):
image
which should be somewhat *correct*

emilk:
Good point… I'll try out your ascent idea in a separate PR then and hold off on this one for a little bit

emilk:
This is starting to look good now! This is using default (zero) `FontTweak` for everything (including emoji fonts):

With default egui fonts:
Screenshot 2024-09-19 at 08 56 33

With Cantarell and NotoSansCJK:
image

emilk:
## `master`, with a bunch of custom `FontTweak`:
old-default

## This PR, with default `FontTweak` (except for scaling emojis):
new-centering

emilk:
There is a big downside to using `valign = Center` though: mixing fonts of different heights (e.g. Heading and Body) looks bad:

Screenshot 2024-09-19 at 09 33 22

I think there should be a solution to this though, which looks something like this:

* Revert to using `valign = Bottom`
* Always center glyphs within the same `Font` (and size), regardless of `valign`
* Only apply valign to the difference in height between the largest `Font` height the the height of the current glyph

I guess I need to do some more experiments…

emilk:
Ok, I think this is good now!

emilk:
Ok, I think this is good now!

PR #5122, opened 2024-09-18
Update sampler along with texture on wgpu backend

### LINKED ISSUE #5121: Partial texture updates will update the sampler in `egui_glow` but not `egui_wgpu`
**Describe the bug**

When using `set_partial` to update a texture, you provide a `TextureOptions` as the third argument. In the `egui_glow` backend, these options are applied to the texture. However, in `egui_wgpu`, they are ignored entirely, and are only updated when you use `set` to *fully* replace the texture.

**To Reproduce**
1. Call `ctx.load_texture` with a new `ColorImage` and some `TextureOptions`
2. Call `set_partial` on the returned `TextureHandle` with some different `TextureOptions` (e.g. switch from linear to nearest-neighbor interpolation)
3. Render the texture via an `Image` widget
4. If you're using the `glow` backend, it'll appear using the new `TextureOptions` passed via `set_partial`. If you're using the `wgpu` backend, it'll appear using the original `TextureOptions` passed via `load_texture`.

**Expected behavior**
Calling `set_partial` should update the `TextureOptions` in both cases. Alternatively, it could be separated into different functions for updating the texture data and the texture options.

valadaptive:
* Closes #5121

This unifies the code paths in `update_texture` somewhat, so that the texture sampler and bind group are always replaced.

Not sure whether removing and reinserting the texture from and into the `textures` map, or creating a new bind group, has much of a performance impact. An alternative, as described in #5121, would be to split the functionality for updating a texture's data from updating its options, so that we don't have to unconditionally update the bind group (or do something like store the options to check if they're changed).

### DIFF
--- a/crates/egui-wgpu/src/renderer.rs
+++ b/crates/egui-wgpu/src/renderer.rs
@@ -164,13 +164,14 @@ struct SlicedBuffer {
 }
 
 pub struct Texture {
-    /// The texture may be None if the `TextureId` is just a handle to a user-provided sampler.
+    /// The texture may be None if the `TextureId` is just a handle to a user-provided bind-group.
     pub texture: Option<wgpu::Texture>,
 
     /// Bindgroup for the texture + sampler.
     pub bind_group: wgpu::BindGroup,
 
-    /// Options describing the sampler used in the bind group. This may be None if the `TextureId` is just a handle to a user-provided sampler.
+    /// Options describing the sampler used in the bind group. This may be None if the `TextureId`
+    /// is just a handle to a user-provided bind-group.
     pub options: Option<epaint::textures::TextureOptions>,
 }
 
--- a/crates/egui-wgpu/src/renderer.rs
+++ b/crates/egui-wgpu/src/renderer.rs
@@ -163,6 +163,17 @@ struct SlicedBuffer {
     capacity: wgpu::BufferAddress,
 }
 
+pub struct Texture {
+    /// The texture may be None if the `TextureId` is just a handle to a user-provided sampler.
+    pub texture: Option<wgpu::Texture>,
+
+    /// Bindgroup for the texture + sampler.
+    pub bind_group: wgpu::BindGroup,
+
+    /// Options describing the sampler used in the bind group. This may be None if the `TextureId` is just a handle to a user-provided sampler.
+    pub options: Option<epaint::textures::TextureOptions>,
+}
+
 /// Renderer for a egui based GUI.
 pub struct Renderer {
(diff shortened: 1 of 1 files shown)


Wumpf (changes requested):
thanks for finding and fixing this!

> Not sure whether removing and reinserting the texture from and into the textures map, or creating a new bind group, has much of a performance impact.

I'm not worried about removing and inserting the texture as this is a fairly fast operation. But re-creating the bind-group every time is indeed a bit worrisome to me since this is an extra resource that has to be taken care of by wgpu and the driver. (On the web this even needs to wait garbage collection if the old one isn't explicitly destroyed).

I figure solving this shouldn't be too hard: We just store something like this on the hashmap
```
struct Texture {
    /// The texture may be None if the `TextureId` is just a handle to a user-provided sampler.
    texture: Option<wgpu::Texture>,
    
    /// Bindgroup for the texture + sampler.
    bind_group: wgpu::BindGroup,
    
    /// Options describing the sampler used in the bind group.
    options: epaint::textures::TextureOptions,
}
```
and recreate fetch sampler & recreate bindgroup only if the options have a mismatch.

Oh and while you're on it: let's call `destroy` if we discard a bind_group.
It doesn't make a difference on native, but right now wgpu doesn't call that for you on object death when being on WebGPU.
Same on `free_texture` for the texture and bindgroup. (calling destroy invalidates every wgpu object that uses the resource at that point and hasn't submitted its operations yet, but `free_texture` is `&mut` and `render` is `&self`, so we should be good!)

valadaptive:
I've updated the code to store the `TextureOptions` along with the other texture info, and only update the bind group if they differ. I've kept the remove/reinsert from the map, since the borrow checker makes us choose between duplicating the code for creating a bind group or moving the data out of the map.

> Oh and while you're on it: let's call `destroy` if we discard a bind_group. It doesn't make a difference on native, but right now wgpu doesn't call that for you on object death when being on WebGPU. This is relevant because all WebGPU objects live in Javascript land and thus get only destroyed once the garbage collector runs over them. Same on `free_texture` for the texture and bindgroup. (calling destroy invalidates every wgpu object that uses the resource at that point and hasn't submitted its operations yet, but `free_texture` is `&mut` and `render` is `&self`, so we should be good!)

I've also made it so that `free_texture` `destroy()`s the texture. As far as I can tell, `BindGroup` doesn't have a `destroy` method?

Wumpf (approved):
oh, interesting! Wasn't aware that `BindGroup` has no `destroy`.
Small nit on doc string otherwise good to go imho. Thanks for taking care of everything!

PR #5126, opened 2024-09-18
Use the largest ascent as the ascent for a font

emilk:
* Closes https://github.com/emilk/egui/issues/4929
* Builds on top of https://github.com/emilk/egui/pull/2724 by @lictex (ptal!)

I think we need both this _and_ some sort of centering (like in https://github.com/emilk/egui/pull/5117) – I'll experiment more. Maybe we should to take the `FontImpl` ascent _and_ the `Font` ascent into account when centering 🤔

### DIFF
--- a/crates/epaint/src/text/font.rs
+++ b/crates/epaint/src/text/font.rs
@@ -468,6 +468,18 @@ impl Font {
         (Some(font_impl), glyph_info)
     }
 
+    pub(crate) fn ascent(&self) -> f32 {
+        if self.fonts.is_empty() {
+            self.row_height
+        } else {
+            let mut max_ascent = 0.0;
+            for font in &self.fonts {
+                max_ascent = f32::max(max_ascent, font.ascent());
+            }
+            max_ascent
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


lictex:
https://github.com/emilk/egui/blob/f4ed394a85fdce6a141fab20002554442c8b69aa/crates/epaint/src/text/font.rs#L355-L359
https://github.com/emilk/egui/blob/f4ed394a85fdce6a141fab20002554442c8b69aa/crates/epaint/src/text/font.rs#L422-L426

row height should be taken from the same font as the ascent, otherwise rows could be clipped vertically.
doing this will make all rows higher but text should be naturally centered again then.

imo whether to use metrics from the primary font (`font[0]`) or the largest font is more like a design choice that, is the *fallback font* important enough to massively affect the layout process?

> probably offtopic:
> `FontTweak::baseline_offset_factor` (used to manually shift rows as a last resort) is getting a bit confusing. maybe it should be moved to `Font` instead

emilk:
I've returned to a mixed approach in https://github.com/emilk/egui/pull/5117 now - it seems to work very well (finally!)

PR #5131, opened 2024-09-19
Add PR preview deployments

lucasmerlin:
This adds preview deployments that will deploy a version of egui_demo_app for each pull request, making things easier to review / test.

Some notes on security:
The preview deployment is split in two workflows, preview_build and preview_deploy.
`preview_build` runs on pull_request, so it won't have any access to the repositories secrets, so it is safe to
build / execute untrusted code.
`preview_deploy` has access to the repositories secrets (so it can push to the pr preview repo) but won't run
any untrusted code (it will just extract the build artifact and push it to the pages branch where it will
automatically be deployed).

To set this up, a DEPLOY_KEY secret needs to be added, which allows the action to push the compiled artifacts into this repository: https://github.com/egui-pr-preview/pr
The deploy key is the private key part of a key generated via ssh-keygen. The public key is set as a deploy key in that repo.
I have created the repo on a separate github org, so it won't be directly associated with emil or egui in case someone pushes something naughty. 

I have set this up in my fork of egui to show how this works:
- I created a PR: https://github.com/lucasmerlin/egui/pull/2
- The code will be compiled and pushed to the egui-pr-preview/pr repo and deployed via github pages
- The bot leaves a comment on the pr with a link to the preview
- The preview is available at https://egui-pr-preview.github.io/pr/2-pr-preview-demo/
   (It's unfortunately only available a couple seconds after the bot writes the comment, because the pages deployment action is run independently on the other repository)
- Once the PR is merged / closed the preview will be cleaned up (unfortunately the empty folder will remain, it seems like it's not possible to remove that via the JamesIves/github-pages-deploy-action action I use, but I don't think that it's a big issue)

I'll leave the PR in draft until the DEPLOY_KEY is set up


.github/workflows/preview_build.yml
```diff
@@ -0,0 +1,52 @@
+# This action builds and deploys egui_demo_app on each pull request created
+# Security notes:
+# The preview deployment is split in two workflows, preview_build and preview_deploy.
+# `preview_build` runs on pull_request, so it won't have any access to the repositories secrets, so it is safe to
+# build / execute untrusted code.
+# `preview_deploy` has access to the repositories secrets (so it can push to the pr preview repo) but won't run
+# any untrusted code (it will just extract the build artifact and push it to the pages branch where it will
+# automatically be deployed).
```
emilk, line 8: Very nice!

emilk: …but I guess it doesn't prevent a malicious user from simply changing the `.yml` in the PR.

But there should be some way to configure it so that it is always the `.yml` from `master` that is run? 🤔

.github/workflows/preview_cleanup.yml
```diff
@@ -0,0 +1,31 @@
...
+        uses: JamesIves/github-pages-deploy-action@v4
+        with:
+          folder: empty_dir
+          repository-name: egui-pr-preview/pr
+          branch: 'main'
+          clean: true
+          target-folder: ${{ env.URL_SLUG }}
+          ssh-key: ${{ secrets.DEPLOY_KEY }}
+          commit-message: "Remove preview for PR ${{ env.URL_SLUG }}"
+          single-commit: true
```
emilk, line 31: nit: hot tip is to configure your editor to

A) ensure a final newline on save
B) trim trailing whitespace on save

.github/workflows/preview_build.yml
lucasmerlin: The pull_request trigger will run workflows from the pr branch, so yes, they could modify the yml. But the workflow won't get any access to secrets and will only get a read only GITHUB_TOKEN, so it wouldn't be a problem. 

Worst thing they could do is change the output to meta.json to point at a different branch and overwrite someone elses preview deployment that way 🤔

It's possible to run the workflow from master via [pull_request_target](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target) but then it would be unsafe to build the untrusted code since workflows using that have access to secrets.

.github/workflows/preview_cleanup.yml
lucasmerlin: Maybe we can add a https://editorconfig.org/ to egui so everyone has the same config

emilk (approved):

Spiral-Memory:
Hey @lucasmerlin ,

Can we set it up so that a PR is deployed only if it has been approved? We are currently using a similar build workflow that triggers with pull_request instead of pull_request_review and uploads artifacts.

Then, with workflow_run, we'll deploy to the same repository under the gh-pages branch. This will work fine, but we want to introduce another safety check: the deployment should only happen if a maintainer or collaborator has approved the changes. I think workflow_run does have write permissions, but the pull_request_review trigger doesn't have write permissions.

Can we do that? If so, please let me know.

Spiral-Memory:
Hey @lucasmerlin ,

Can we set it up so that a PR is deployed only if it has been approved? We are currently using a similar build workflow that triggers with pull_request instead of pull_request_target and uploads artifacts.

Then, with on workflow_run, we'll deploy to the same repository under the gh-pages branch. This will work fine, but we want to introduce another safety check: the deployment should only happen if a maintainer or collaborator has approved the changes. I think on workflow_run does have write permissions, but the pull_request_review trigger doesn't have write permissions.

Can we do that? If so, please let me know.

lucasmerlin:
Hi @Spiral-Memory, I think this should be possible somehow but I haven't looked into that. In the workflow_run you should be able check whether the PR has been approved via the API and you should also be able to trigger the workflow run task via a pull_request_review task. I think this way you should be able to wire this up.

Spiral-Memory:
Yep, i was thinking about the same approach, thanks for your help / confirmation 

I'll try implementing that and let you know if it works 

Thanks !

Issue #5132, opened 2024-09-19
Getting PERF_WARNING: request_discard has been called X frames in a row


kjarosh:
**Describe the bug**

After updating to the latest master, the following warning started to appear:

<image src="https://github.com/user-attachments/assets/55284d3f-11aa-40e5-ab14-16c6639446fe" width="300"/>

**To Reproduce**
1. Open Ruffle (the newest code from master https://github.com/ruffle-rs/ruffle, should be available tomorrow as a nightly release)
2. Expand "Movie Parameters"
3. Move your mouse

**Expected behavior**

No warning

**Desktop (please complete the following information):**
 - OS: Fedora 
 - Browser N/A 
 - Version 40 

**Additional context**

That started happening after upgrading from f4697bc007447c6c2674beb4e25f599fb7afa093 to e0f0b7f47f4094967530214aec060b9f1ce5954e. Possibly related to https://github.com/emilk/egui/pull/5059

Here's the related code in Ruffle: https://github.com/ruffle-rs/ruffle/blob/309d600c6daa400ac02d8aff5ee60d0cda670f6c/desktop/src/gui/dialogs/open_dialog.rs#L537. It started happening after https://github.com/ruffle-rs/ruffle/pull/17987

CC @torokati44

lucasmerlin:
This seems to be caused by the empty Grid. I think because nothing is added the Grid doesn't save it's state, so it thinks it's always the first frame and requests a discard. This is definitely a egui bug, thanks for reporting!

PR #5133, opened 2024-09-19
Fix empty grids repeatedly requesting a discard

lucasmerlin:
* Closes <https://github.com/emilk/egui/issues/5132>

This does cause a slight difference in the amount of space allocated by a empty grid, is this a problem?
Before:

image

After:

image


emilk (approved):
Looks good - I don't think the minor size difference is worth losing sleep over

PR #5140, opened 2024-09-20
Improvement: Reduce the area covered by the mouse cursor in `on_hover_at_pointer()`.

rustbasic:
Improvement: Reduce the area covered by the mouse cursor in `on_hover_at_pointer()`.

### DIFF
--- a/crates/egui/src/containers/popup.rs
+++ b/crates/egui/src/containers/popup.rs
@@ -87,7 +87,7 @@ pub fn show_tooltip_at_pointer<R>(
 
         // Add a small exclusion zone around the pointer to avoid tooltips
         // covering what we're hovering over.
-        let mut exclusion_rect = Rect::from_center_size(pointer_pos, Vec2::splat(24.0));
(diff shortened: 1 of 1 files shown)


crates/egui/src/containers/popup.rs
```diff
@@ -87,7 +87,8 @@ pub fn show_tooltip_at_pointer<R>(
 
         // Add a small exclusion zone around the pointer to avoid tooltips
         // covering what we're hovering over.
-        let mut exclusion_rect = Rect::from_center_size(pointer_pos, Vec2::splat(24.0));
+        let exclusion_size = Vec2::splat(32.0 * ctx.native_pixels_per_point().unwrap_or(1.0));
```
emilk, line 90: Why would the pixel density of the display matter here?

rustbasic: > Why would the pixel density of the display matter here?

On Windows, the mouse cursor size varies depending on `native_pixels_per_point`.
I don't know if the mouse cursor size varies on other OSes.

PR #5141, opened 2024-09-20
Improvement: Simple structure of FontBook

rustbasic:
Improvement: Simple structure of FontBook

1. Simple structure of FontBook
2. Fix for the warning "epaint texture atlas overflowed!"


emilk:
Closing: the font book was rewritten in #8504 and #8512, which cover both of these changes. Thanks for the PR!

PR #5146, opened 2024-09-22
Add support for mipmap textures.

nolanderc:

Adds support for mipmaps in the `glow` backend.

Should be possible to implement for `wgpu` in the future as well, but requires a custom compute kernel.


emilk (approved):
Nice!

PR #5149, opened 2024-09-23
Wgpu render pass on paint callback has now static lifetime

Wumpf:
A very common usability issue on egui-wgpu callbacks is that `paint` can't access any data that doesn't strictly outlive the callback resources' data. E.g. if the callback resources have an `Arc` to some resource manager, you can't easily pull out resources since you statically needed to ensure that those resource references outlived the renderpass, whose lifetime was only constrained to the callback resources themselves.

Wgpu 22 no longer has this restriction! Its (render/compute-)passes take care of the lifetime of any passed resource internally. The lifetime constraint is _still_ opt-out since it protects from a common runtime error of adding commands/passes on the parent encoder while a previously created pass wasn't closed yet.
This is not a concern in egui-wgpu since the paint method where we have to access the render pass doesn't even have access to the encoder!

### DIFF
--- a/crates/egui-wgpu/src/renderer.rs
+++ b/crates/egui-wgpu/src/renderer.rs
@@ -112,11 +112,11 @@ pub trait CallbackTrait: Send + Sync {
     ///
     /// It is given access to the [`wgpu::RenderPass`] so that it can issue draw commands
     /// into the same [`wgpu::RenderPass`] that is used for all other egui elements.
-    fn paint<'a>(
-        &'a self,
+    fn paint(
+        &self,
         info: PaintCallbackInfo,
         render_pass: &mut wgpu::RenderPass<'static>,
-        callback_resources: &'a CallbackResources,
+        callback_resources: &CallbackResources,
     );
 }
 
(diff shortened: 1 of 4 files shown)


crates/egui-wgpu/src/renderer.rs
```diff
@@ -408,10 +408,16 @@ impl Renderer {
     }
 
     /// Executes the egui renderer onto an existing wgpu renderpass.
-    pub fn render<'rp>(
-        &'rp self,
-        render_pass: &mut wgpu::RenderPass<'rp>,
-        paint_jobs: &'rp [epaint::ClippedPrimitive],
+    ///
+    /// Note that the lifetime of `render_pass` is `'static` which requires a call to `wgpu::RenderPass::forget_lifetime`.
```
emilk, line 412:
```suggestion
    /// Note that the lifetime of `render_pass` is `'static` which requires a call to [`wgpu::RenderPass::forget_lifetime`].
```

crates/egui-wgpu/src/renderer.rs
```diff
@@ -115,7 +115,7 @@ pub trait CallbackTrait: Send + Sync {
     fn paint<'a>(
         &'a self,
         info: PaintCallbackInfo,
-        render_pass: &mut wgpu::RenderPass<'a>,
+        render_pass: &mut wgpu::RenderPass<'static>,
         callback_resources: &'a CallbackResources,
```
Wumpf, line 119: huh, does this even need the explicit lifetimes at all now?

emilk (approved):
nice

crates/egui-wgpu/src/renderer.rs
Wumpf: the answer is no!

torokati44:
May I request some assistance for making https://github.com/ruffle-rs/ruffle/pull/18082 work, after this change?

Wumpf:
@torokati44 should be fairly straight forward: Just remove the lifetimes on paint. Since all changes are strictly lessenings of previous lifetime constraints it should be working out of the box once you adjusted the callback method's signature. Let me know if you have any concrete issues and I can have a look!

PR #5157, opened 2024-09-24
allow(dead_code)

rustbasic:
allow(dead_code)


lucasmerlin:
Why is this needed? Are you getting a lint error?

If DummyStorage isn't exported and isn't used anywhere, maybe we can just remove it?

bircni:
@lucasmerlin I also already wondered but maybe there is something we miss...
I don't get a linting error...

rustbasic:
This is required for the latest version of the Rust compiler.

PR #5158, opened 2024-09-24
Enhance Github Workflows

bircni:
In order to make the workflows more readable (and faster). I moved the docs and test to another job.


bircni:
Makes the response time even worse :-(

PR #5159, opened 2024-09-24
egui_extras: Add `TableBuilder::animate_scrolling`

ecpost:
This will allow disabling the animation that occurs during `scroll_to_row` calls.  Follows the same coding style as the other scroll options in `TableBuilder`.

### DIFF
--- a/crates/egui_extras/src/table.rs
+++ b/crates/egui_extras/src/table.rs
@@ -415,7 +415,7 @@ impl<'a> TableBuilder<'a> {
     ///
     /// Default: `true`.
     #[inline]
-    pub fn animated(mut self, animated: bool) -> Self {
+    pub fn animate_scrolling(mut self, animated: bool) -> Self {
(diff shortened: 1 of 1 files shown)


crates/egui_extras/src/table.rs
```diff
@@ -409,6 +411,15 @@ impl<'a> TableBuilder<'a> {
         self
     }
 
+    /// Should the scroll area animate `scroll_to_*` functions?
+    ///
+    /// Default: `true`.
+    #[inline]
+    pub fn animated(mut self, animated: bool) -> Self {
+        self.scroll_options.animated = animated;
```
emilk, line 419: Let's call this `animate_scrolling` to be more explicit

```suggestion
    pub fn animate_scrolling(mut self, animated: bool) -> Self {
        self.scroll_options.animated = animated;
```

emilk (approved):

Issue #4928, opened 2024-08-06
Add `ui.align_cursor()` to for the cursor to be aligned with physical pixels


abey79:
Unaligned cursor lead to various visual glitches. It would be nice to have an API to "fix" the cursor position. This would avoid using code like this:

```
ui.add_space(-ui.cursor().min.y.fract());
```

which is not robust to layout orientation and assumes 1:1 physical pixels.

Related:
- https://github.com/emilk/egui/issues/4927

emilk:
There are three alternatives:
* Round widgets to physical pixels
* Round widgets to logical ui points
* No rounding

I'd like to understand a bit more what the trade-offs are between these.

Regardless of how/if we round widgets we still want to round some visual elements to physical pixels to make them crisper (text, thin lines, etc).

abey79:
> I'd like a repro for this

```rust
use eframe::egui;

fn main() -> eframe::Result {
    env_logger::init(); // Log to stderr (if you run with `RUST_LOG=debug`).
    let options = eframe::NativeOptions {
        viewport: egui::ViewportBuilder::default().with_inner_size([320.0, 240.0]),
        ..Default::default()
    };
    eframe::run_native(
        "My egui App",
        options,
        Box::new(|cc| {
            // This gives us image support:
            egui_extras::install_image_loaders(&cc.egui_ctx);

            Ok(Box::new(MyApp))
        }),
    )
}

struct MyApp;

impl eframe::App for MyApp {
    fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
        egui::CentralPanel::default().show(ctx, |ui| {
            // unaligned cursor
            ui.add_space(0.1234);

            ui.spacing_mut().item_spacing.y = 0.0;

            for _ in 0..5 {
                let (rect, _) =
                    ui.allocate_exact_size(egui::vec2(100.0, 20.0), egui::Sense::hover());
                ui.painter()
                    .rect_filled(rect, 0.0, ui.style().visuals.selection.bg_fill);
            }
        });
    }
}
```

<img width="432" alt="image" src="https://github.com/user-attachments/assets/6dfbd92d-a2a1-4fe2-80ad-8225d1b56da8">

abey79:
> There are three alternatives:
> * Round widgets to physical pixels
> * Round widgets to logical ui points
> * Only round when `pixels_per_point` is an integer
> * No rounding

[...]

Issue #4927, opened 2024-08-06
Debug tool to highlight unaligned cursor coordinates


abey79:
When `ui.cursor()` is set to fractional coordinates that are not aligned to physical pixels, all sorts of visual issues ensue. For example, `ui.button()` and `ui.selectable_*()` have their text oddly vertically aligned on the background (causing [this](https://github.com/rerun-io/rerun/pull/7064#issuecomment-2270865953)).

It would be very useful to have a tool that displays a big fat point whenever the cursor is unaligned, so we can easily spot the thing that created that misalignment in the first place.

emilk:
See my comment on https://github.com/emilk/egui/issues/4928#issuecomment-2353528354 about the need to decide what we want to align to (ui points of physical pixels)

PR #5161, opened 2024-09-25
Prevent text shrinking in tooltips; round wrap-width to integer

emilk:
* Closes https://github.com/emilk/egui/pull/5106
* Closes https://github.com/emilk/egui/issues/5084

Protect against rounding errors in egui layout code.

Say the user asks to wrap at width 200.0.
The text layout wraps, and reports that the final width was 196.0 points.
This than trickles up the `Ui` chain and gets stored as the width for a tooltip (say).
On the next frame, this is then set as the max width for the tooltip,
and we end up calling the text layout code again, this time with a wrap width of 196.0.
Except, somewhere in the `Ui` chain with added margins etc, a rounding error was introduced,
so that we actually set a wrap-width of 195.9997 instead.
Now the text that fit perfectly at 196.0 needs to wrap one word earlier,
and so the text re-wraps and reports a new width of 185.0 points.
And then the cycle continues.

So this PR limits the text wrap-width to be an integer.

Related issues:
* https://github.com/emilk/egui/issues/4927
* https://github.com/emilk/egui/issues/4928
* https://github.com/emilk/egui/issues/5163

--- 

Pleas test this @rustbasic

### DIFF
--- a/crates/epaint/src/text/fonts.rs
+++ b/crates/epaint/src/text/fonts.rs
@@ -697,6 +697,7 @@ impl GalleyCache {
             // * https://github.com/emilk/egui/issues/4927
             // * https://github.com/emilk/egui/issues/4928
             // * https://github.com/emilk/egui/issues/5084
+            // * https://github.com/emilk/egui/issues/5163
 
             job.wrap.max_width = job.wrap.max_width.round();
         }
--- a/crates/epaint/src/text/fonts.rs
+++ b/crates/epaint/src/text/fonts.rs
@@ -676,7 +676,31 @@ struct GalleyCache {
 }
 
 impl GalleyCache {
-    fn layout(&mut self, fonts: &mut FontsImpl, job: LayoutJob) -> Arc<Galley> {
+    fn layout(&mut self, fonts: &mut FontsImpl, mut job: LayoutJob) -> Arc<Galley> {
+        if job.wrap.max_width.is_finite() {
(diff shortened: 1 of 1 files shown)


crates/epaint/src/text/fonts.rs
```diff
@@ -676,7 +676,31 @@ struct GalleyCache {
...
+            // Now the text that fit perfrectly at 196.0 needs to wrap one word earlier,
+            // and so the text re-wraps and reports a new width of 185.0 points.
+            // And then the cycle continues.
+
+            // So we limit max_width to integers.
+
+            // Related issues:
+            // * https://github.com/emilk/egui/issues/4927
+            // * https://github.com/emilk/egui/issues/4928
+            // * https://github.com/emilk/egui/issues/5084
```
emilk, line 699:
```suggestion
            // * https://github.com/emilk/egui/issues/5084
            // * https://github.com/emilk/egui/issues/5163
```

rustbasic:
Dear emilk.

After a brief test, it works well.
Even with multiple popups, the size doesn't change, so I think we can close #5079.
I will let you know if I find any other issues.

Thank you, emilk.

KaranJain21:
This is happening in egui 0.29.1:
[Screencast from 2024-12-05 10-50-58.webm](https://github.com/user-attachments/assets/268ab962-1968-4384-9f17-27e782ec36f9)

PR #5166, opened 2024-09-25
Add egui testing library

### LINKED ISSUE #3491: Integration testing/event injection
I didn't find _anything_ about this in this whole project which is odd so I think the following should be considered a feature request.

There's currently no documented (or even possible?) way to do integration testing/event injection/instrumentation for the purposes of integration testing. It's probably possible to abuse the [new accessibility features](#167) for that like in other frameworks but it would still be good to have an official way for doing this for people that want to test their UIs.

### LINKED ISSUE #3926: Improved unit/regression test support

To improve regression testing, it would be nice if there is was a mechanism to "inject" mouse and keyboard presses from code. (I guess there may be, just that it is not that simple to figure out how.)

Ideally a testing framework to make it easier to test egui applications. However, a clear described mechanism to force certain events would be a first step.

**Additional context**

This is somewhat related to https://github.com/emilk/egui/issues/1129 (we are using egui-skia to generate test images, but can only test features triggered by messages, not "pressing" keys/mouse buttons in suitable areas).

lucasmerlin:
- closes #3491 
- closes #3926

This adds a testing library to egui based on [kittest](https://github.com/rerun-io/kittest). Kittest is a new [AccessKit](https://github.com/AccessKit/accesskit/)-based testing library. The api is inspired by the js [testing-library](https://testing-library.com/) where the idea is also to query the dom based on accessibility attributes. 
We made kittest with egui in mind but it should work with any rust gui framework with AccessKit support.

It currently has support for:
- running the egui app, frame by frame
- building the AccessKit tree
- ergonomic queries via kittest
  - via e.g. get_by_name, get_by_role
- simulating events based on the accesskit node id
- creating arbitrary events based on Harness::input_mut
- rendering screenshots via wgpu
- snapshot tests with these screenshots

A simple test looks like this: 
```rust
fn main() {
    let mut checked = false;
    let app = |ctx: &Context| {
            ui.checkbox(&mut checked, "Check me!");
        });
    };

    let mut harness = Harness::builder().with_size(egui::Vec2::new(200.0, 100.0)).build(app);
    
    let checkbox = harness.get_by_name("Check me!");
    assert_eq!(checkbox.toggled(), Some(Toggled::False));
    checkbox.click();
    
    harness.run();

    let checkbox = harness.get_by_name("Check me!");
    assert_eq!(checkbox.toggled(), Some(Toggled::True));

    // You can even render the ui and do image snapshot tests
    #[cfg(all(feature = "wgpu", feature = "snapshot"))]
    egui_kittest::image_snapshot(&egui_kittest::wgpu::TestRenderer::new().render(&harness), "readme_example");
}
```

~Since getting wgpu to run in ci is a hassle, I'm taking another shot at creating a software renderer for egui (ideally without a huge dependency like skia)~ (this didn't work as well as I hoped and it turns out in CI you can just run tests on a mac runner which comes with a real GPU)
 
Here is a example of a failed snapshot test in ci, it will say which snapshot failed and upload an artifact with the before / after and diff images:
https://github.com/emilk/egui/actions/runs/11183049487/job/31090724606?pr=5166


emilk:
After some discussion we came to these conclusions:

* Investigate using `git lfs` for storing the `.png` snapshot files so we don't bloat the git repository. Try to set it up in such a way that `git clone` + `cargo test` still works for contributors without `lfs`.
* Try to set up a proper software wgpu backend that works on CI, same as the `wgpu` repository uses. This will be the most powerful and accurate, and allow us to test more than just egui (i.e. the Rerun Viewer)
* Move the accesskit-specific part of the integration to its own repository and eventually invite collaborators from other GUI toolkits to collaborate on it

.github/workflows/rust.yml
```diff
@@ -222,3 +218,36 @@ jobs:
...
+      - name: Set up cargo cache
+        uses: Swatinem/rust-cache@v2
+
+      - name: Run tests
+        # TODO(lucasmerlin): Enable --all-features (currently this breaks the rendering in the tests because of the `unity` feature)
+        run: cargo test
+
+      - name: Run doc-tests
+        # TODO(lucasmerlin): Enable --all-features (currently this breaks the rendering in the tests because of the `unity` feature)
+        run: cargo test --doc
```
lucasmerlin, line 246: @emilk the `unity` feature completely breaks rendering with egui_wgpu if enabled, so we can't use --all-features anymore. Do you have an idea how we could work around that? 
My only idea is to manually enable all features except the unity one but I really don't like that.

Or we add a `no_unity` feature that overwrites the unity feature but that seems terrible as well.

crates/egui_kittest/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "egui_kittest"
+edition.workspace = true
+license.workspace = true
+rust-version.workspace = true
+version.workspace = true
```
emilk, line 6: This lacks a lot of stuff, like:

* authors
* description
* readme
* categories
* keywords
* include

See any other `Cargo.toml` for more

crates/egui_kittest/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[features]
+wgpu = ["dep:egui-wgpu", "dep:pollster", "dep:image"]
+snapshot = ["dep:dify"]
+
+default = ["image?/png"]
```
emilk, line 14: All crates in this repository uses [document-features](https://crates.io/crates/document-features) to document their features. Please do the same here!

crates/egui_kittest/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+kittest = { git = "https://github.com/rerun-io/kittest", version = "0.1", branch = "main" }
+egui = { workspace = true, features = ["accesskit"] }
+
+# wgpu dependencies
+egui-wgpu = { workspace = true, optional = true }
+pollster = { version = "0.3", optional = true }
+image = { workspace = true, optional = true }
+
+# snapshot dependencies
+dify = { version = "0.7.0", optional = true, default-features = false }
```
emilk, line 27: Add new deps as workspace dependencies under the main `Cargo.toml` (accesskit_consumer, kittest, pollster, dify, …)

crates/egui_kittest/README.md
```diff
@@ -0,0 +1,35 @@
+# egui_kittest 
+
+Ui testing library for egui, based on [kittest](https://github.com/rerun-io/kittest) (a AccessKit based testing library).
```
emilk, line 3:
```suggestion
Ui testing library for egui, based on [kittest](https://github.com/rerun-io/kittest) (an [AccessKit](https://github.com/AccessKit/accesskit) based testing library).
```

crates/egui_kittest/src/builder.rs
```diff
@@ -0,0 +1,54 @@
...
+            screen_rect: Rect::from_min_size(Pos2::ZERO, Vec2::new(800.0, 600.0)),
+            dpi: 1.0,
+        }
+    }
+}
+
+impl HarnessBuilder {
+    /// Set the size of the window.
+    #[inline]
+    pub fn with_size(mut self, size: Vec2) -> Self {
```
emilk, line 22: slightly nice ergonomics (can pass in `[800.0, 600.0]` as an argument, for instance)

```suggestion
    pub fn with_size(mut self, size: impl Into<Vec2>) -> Self {
```

crates/egui_kittest/src/event.rs
```diff
@@ -0,0 +1,179 @@
...
+use kittest::{ElementState, MouseButton, SimulatedEvent};
+
+#[derive(Default)]
+pub(crate) struct EventState {
+    modifiers: Modifiers,
+    last_mouse_pos: Pos2,
+}
+
+impl EventState {
+    pub fn kittest_event_to_egui(&mut self, event: kittest::Event) -> Option<egui::Event> {
```
emilk, line 12: Does this really need `&mut self`?

crates/egui_kittest/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+        let ctx = egui::Context::default();
+        ctx.enable_accesskit();
+        let mut input = egui::RawInput {
+            screen_rect: Some(builder.screen_rect),
+            ..Default::default()
+        };
+        let viewport = input.viewports.get_mut(&ViewportId::ROOT).unwrap();
+        viewport.native_pixels_per_point = Some(builder.dpi);
+
+        let mut output = ctx.run(input.clone(), &mut app);
```
emilk, line 56: Maybe explain why we need to run it once here

crates/egui_kittest/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// prefer using [`HarnessBuilder::with_dpi`].
+    #[inline]
+    pub fn set_dpi(&mut self, dpi: f32) -> &mut Self {
+        self.ctx.set_pixels_per_point(dpi);
+        self
+    }
+
+    /// Run a frame.
+    /// This will call the app closure with the current context and update the Harness.
+    pub fn run(&mut self) {
```
emilk, line 119: As mentioned before, I think we should call this `step` and then have a separate `run` that for now calls `step` twice (but in the future would run until "stable")

crates/egui_kittest/src/snapshot.rs
```diff
@@ -0,0 +1,122 @@
+use image::ImageError;
+use std::fmt::Display;
+use std::io::ErrorKind;
+use std::path::{Path, PathBuf};
+
+#[derive(Debug)]
+pub enum SnapshotError {
+    Diff {
+        diff: i32,
```
emilk, line 9: What's the unit here?

crates/egui_kittest/src/snapshot.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    if previous.dimensions() != current.dimensions() {
+        maybe_update_snapshot(&path, current);
+        return Err(SnapshotError::SizeMismatch {
+            expected: previous.dimensions(),
+            actual: current.dimensions(),
+        });
+    }
+
+    let result = dify::diff::get_results(previous, current.clone(), 0.1, true, None, &None, &None);
```
emilk, line 85: What's `0.1` here? You can name it to make the code a bit more readable

```suggestion
    let tolerance = 0.1; // Chosen because blah blah blah
    let result = dify::diff::get_results(previous, current.clone(), tolerance, true, None, &None, &None);
```

emilk, line 85: And separately - have you tested tolerance=0?

crates/egui_kittest/src/snapshot.rs
```diff
@@ -0,0 +1,122 @@
...
+pub fn try_image_snapshot(current: &image::RgbaImage, name: &str) -> Result<(), SnapshotError> {
+    let snapshots_path = Path::new("tests/snapshots");
+
+    let path = snapshots_path.join(format!("{name}.png"));
+    std::fs::create_dir_all(path.parent().unwrap()).ok();
+
+    let diff_path = snapshots_path.join(format!("{name}.diff.png"));
+    let current_path = snapshots_path.join(format!("{name}.new.png"));
+
+    current.save(&current_path).unwrap();
```
emilk, line 67: Maybe failure to write images to disk should also result in an error, rather than a panic.

In either case, it should include the problematic path in the error message

crates/egui_kittest/src/wgpu.rs
```diff
@@ -0,0 +1,126 @@
...
+        let mut encoder = self
+            .device
+            .create_command_encoder(&wgpu::CommandEncoderDescriptor {
+                label: Some("Egui Command Encoder"),
+            });
+
+        let size = harness.ctx.screen_rect().size() * harness.ctx.pixels_per_point();
+        let screen = ScreenDescriptor {
+            pixels_per_point: harness.ctx.pixels_per_point(),
+            size_in_pixels: [size.x as u32, size.y as u32],
```
emilk, line 65:
```suggestion
            size_in_pixels: [size.x.round() as u32, size.y.round() as u32],
```

crates/egui_kittest/src/texture_to_bytes.rs
```diff
@@ -0,0 +1,83 @@
+use egui_wgpu::wgpu;
+use egui_wgpu::wgpu::{Device, Extent3d, Queue, Texture};
+use image::RgbaImage;
+use std::iter;
+use std::mem::size_of;
+use std::sync::mpsc::channel;
+
+pub(crate) fn texture_to_bytes(device: &Device, queue: &Queue, texture: &Texture) -> RgbaImage {
```
emilk, line 8: A bit confusing name imho. Could be called `read_texture` or `image_from_texture`

.gitattributes
```diff
@@ -1,2 +1,3 @@
 * text=auto eol=lf
 Cargo.lock linguist-generated=false
+**/tests/snapshots/**/*.png filter=lfs diff=lfs merge=lfs -text
```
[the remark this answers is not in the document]
lucasmerlin: I think so, but I think we should do that in a separate PR. And it should also include gifs, which seem to be the biggest files in the repo. And maybe also the entirety of the media folder.

crates/egui_kittest/src/event.rs
lucasmerlin: Yes, because we need to somehow track the modifiers and last mouse position. I aligned the kittest events with winits event types, so the mouse click events don't contain the mouse position.

crates/egui_demo_lib/tests/snapshots/demos/Tooltips.png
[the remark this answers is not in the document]
lucasmerlin: Ah, that's because this one has `constrain(false)`. No idea how I could fix that though without changing constrain to true 🤔

crates/egui_kittest/src/snapshot.rs
lucasmerlin: I've changed it to 0, but I think we should at least have someone test this on linux and windows. I can test on windows but someone else will need to test it for linux. Preferably once with the vulkan and once with the gl backend.

crates/egui_demo_lib/src/demo/widget_gallery.rs
```diff
@@ -285,3 +285,38 @@ fn doc_link_label_with_crate<'a>(
...
+                demo.ui(ui);
+            });
+        };
+        let mut harness = Harness::builder()
+            .with_size(Vec2::new(380.0, 550.0))
+            .with_dpi(2.0)
+            .build(app);
+
+        // The first and second frames are slightly different, so we take the second frame
+        harness.run();
```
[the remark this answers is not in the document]
lucasmerlin: Done!

lucasmerlin:
I've changed everything you mentioned and also added `Harness::wgpu_snapshot` as a convenicence.

Also, I've disabled dithering since it caused minor differences between frames. This made the pngs much smaller!

Issue #5160, opened 2024-09-24
hex_color!() macro is not const


Ora9:
It would be nice if `hex_color!()` macro would be `const` to be used in constant declaration and to define colors at compile time

Btw, the doc says that it's the case on [`Color32::from_hex` method](https://docs.rs/egui/latest/egui/struct.Color32.html#method.from_hex) : 
« To parse hex colors at compile-time (e.g. for use in const contexts) use the macro [crate::hex_color!](https://docs.rs/egui/latest/egui/macro.hex_color.html) instead. »

But when trying on a test project with the latest egui git master, i get this error :
```
error[E0015]: cannot call non-const fn `Color32::from_rgba_unmultiplied` in constants
 --> src/main.rs:1:30
  |
1 | const TEST: egui::Color32 = egui::hex_color!("#ff0202");
  |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: calls in constants are limited to constant functions, tuple structs and tuple variants
  = note: this error originates in the macro `egui::hex_color` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0015`.
```

This is related to #1596, where [this comment](https://github.com/emilk/egui/pull/1596#issuecomment-1124187465) was saying that it wasn't possible at the time to make the macro const due to floating-point conversions. And that this could be soon fixed
But this was in 2022, and looking at the code ([here](https://github.com/emilk/egui/blob/1603f0581882c4ca299db1d6efdba9148ebcddb6/crates/ecolor/src/hex_color_macro.rs#L14) and [here](https://github.com/emilk/egui/blob/1603f0581882c4ca299db1d6efdba9148ebcddb6/crates/ecolor/src/color32.rs#L95) there are some functions that are not const yet (`linear_f32_from_gamma_u8` and friends)
I'm quite new to rust, but is it still not possible to make the macro const ? That would be really handy

YgorSouza:
Const floating-point arithmetic is going to be stabilized in Rust 1.82, which comes out in 3 weeks. But egui is still targeting 1.76 for now, and likely won't switch immediately. When it does switch, I suppose a const version of `from_rgba_unmultiplied` could be written to make the macro const. The current implementation cannot be made const because it uses a lookup table to improve performance, and that can only be initialized at runtime.

Until then, maybe the documentation could be updated to reflect the fact that the macro is not actually const, but only has the advantage of generating a compilation error if the string literal is invalid.

[...]

PR #5169, opened 2024-09-25
Document the fact that the hex_color macro is not const

YgorSouza:
It cannot be made const with the current version of Rust, and that is counterintuitive since it does compile-time checks, so we make that clear in the documentation. It might be possible to make it const once MSRV is bumped to 1.82.

* See <https://github.com/emilk/egui/issues/5160>


emilk (approved):
Thanks!

PR #5175, opened 2024-09-27
Update ScrollArea drag velocity when drag stopped

### LINKED ISSUE #5174: When dragging to scroll, spurious drag velocity is applied
**Describe the bug**
When dragging to scroll in a `ScrollArea`, egui applies the drag velocity if the user releases their finger/mouse button mid-drag. However, this velocity is only updated when a drag actually takes place. If the cursor decelerates quickly, no drag events will be registered because it's not moving, and the velocity of the previous drag event will be used:

https://github.com/user-attachments/assets/0f2928b2-720f-4360-87f2-8346971f8d91

**To Reproduce**
1. Open the egui demo app
2. Open a demo that lets you drag to scroll, e.g. the "Scroll a lot of lines" tab in the "Scrolling" demo, the "Misc Demos" window, or the "EasyMark Editor".
3. Drag a scrollable area quickly, then decelerate your dragging movement rapidly.
4. Release the mouse button and observe that the previous velocity is kept even if the cursor is currently not moving.

**Expected behavior**
No further scrolling should occur, as the cursor is completely still and hence has no velocity.

valadaptive:
Fixes #5174.

The drag velocity was not being updated unless the cursor counted as "dragging", which only happens when it's in motion. This effectively guarantees that the drag velocity will never be zero, even if the cursor is not moving, and results in spurious scroll velocity being applied when the cursor is released.

Instead, we update the velocity only when the drag is stopped, which is when the kinetic scrolling actually needs to begin. Note that we immediately *apply* the scroll velocity on the same frame that we first set it, to avoid a 1-frame gap where the scroll area doesn't move.

I believe that *not* setting `scroll_stuck_to_end` and `offset_target` when the drag is released is the correct thing to do, as they should apply immediately once the user stops dragging. Should we maybe clear the drag velocity instead if `scroll_stuck_to_end` is true or `offset_target` exists?

* Closes #5174

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -621,20 +621,35 @@ impl ScrollArea {
                 .interact_rect
                 .map(|rect| ui.interact(rect, id.with("area"), Sense::drag()));
 
-            if content_response_option.map(|response| response.dragged()) == Some(true) {
+            if content_response_option
+                .as_ref()
+                .is_some_and(|response| response.dragged())
+            {
                 for d in 0..2 {
                     if scroll_enabled[d] {
                         ui.input(|input| {
                             state.offset[d] -= input.pointer.delta()[d];
-                            state.vel[d] = input.pointer.velocity()[d];
                         });
                         state.scroll_stuck_to_end[d] = false;
                         state.offset_target[d] = None;
-                    } else {
-                        state.vel[d] = 0.0;
                     }
                 }
             } else {
+                // Apply the cursor velocity to the scroll area when the user releases the drag.
+                if content_response_option
+                    .as_ref()
+                    .is_some_and(|response| response.drag_stopped())
+                {
+                    for d in 0..2 {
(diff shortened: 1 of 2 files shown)


crates/egui/src/containers/scroll_area.rs
```diff
@@ -621,20 +621,35 @@ impl ScrollArea {
...
+                {
+                    for d in 0..2 {
+                        if scroll_enabled[d] {
+                            ui.input(|input| {
+                                state.vel[d] = input.pointer.velocity()[d];
+                            });
+                        } else {
+                            state.vel[d] = 0.0;
+                        }
+                    }
```
emilk, line 651: It would be nice if we could write this as 

```suggestion
                    state.vel = scroll_enabled.to_vec2() * input.pointer.velocity();
```

(requires implementing adding a `fn to_vec2(&self)` to `Vec2b`)

emilk (approved):
Makes sense!

PR #5180, opened 2024-09-28
Preserves the text cursor position when right-clicking in TextEdit

rustbasic:
Preserves the text cursor position when right-clicking in TextEdit

**Advantages:**
1. Right-clicking allows you to maintain the cursor position or selection while gaining focus.
2. Right-dragging lets you specify a region starting from the current cursor position.

* Closes #4865


emilk (commented):
How macOS works:

* right-click a word, and that word is selected
* right-click an existing selection, and that selection remains

PR #5186, opened 2024-09-29
Fix issue where nested side panels can't be resized larger than their contents

teddemunnik:
This fixes an issue where a panel could not be resized to a size larger than it's content layout rect when using nested panels. The issue shows both when nesting panels in panels, and when nesting panels in a window. I have created a github project as a repro here: 
https://github.com/teddemunnik/egui_inner_panel_repro/blob/master/src/main.rs

https://github.com/user-attachments/assets/f812b7eb-1f24-4d43-8396-eec31dea3794

https://github.com/user-attachments/assets/a84a4a21-117a-45a9-be0a-a932dc83b1e4

### DIFF
--- a/crates/egui/src/containers/panel.rs
+++ b/crates/egui/src/containers/panel.rs
@@ -284,21 +284,19 @@ impl SidePanel {
             add_contents(ui)
         });
 
-        let rect = inner_response.response.rect;
-
         {
             let mut cursor = ui.cursor();
             match side {
                 Side::Left => {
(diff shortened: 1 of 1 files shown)


emilk:
When testing, let's make sure this doesn't break the auto-sizing of panels

teddemunnik:
> When testing, let's make sure this doesn't break the auto-sizing of panels

Probably not! My bad, also found the docs regarding this behavior on panels when looking for the auto sizing behavior as you mentioned it

Issue #5008, opened 2024-08-26
eframe Linux: Backspace & arrow keys are completely broken


StarStarJ:
I use the latest master from egui (a9a6e0c2f223419d52a90cb3d40e211810caf1ee), and run the egui_demo_app on x11 linux (no IME). I can't use arrows or backspace in a text edit

Maybe it is related to this issue:
* #4908

**To Reproduce**
Simply start the demo app (native) and the keys wont work

**Desktop (please complete the following information):**
 - OS: debian sid x11 using KDE

rustbasic:
Could you check if applying #4896 fixes the issue?

* #4896

StarStarJ:
I tried 6ef2ef55c4e662762d47ca6639fed5a522cc1159,
which does **not** fix the issue.

rustbasic:
> I tried your branch on rev [6ef2ef5](https://github.com/emilk/egui/commit/6ef2ef55c4e662762d47ca6639fed5a522cc1159), which does **not** fix the issue.

It's strange. The issue only happens when `state.ime_enabled` is `true`. So applying #4896 to set `state.ime_enabled` to `false` when gaining or losing focus should fix the issue, but it's not working, which is strange.

rustbasic:
Dear `emilk`.
#4896 should be applied separately from the resolution of this issue.

StarStarJ:
You are right, it's most likely not related to your merge request

`[crates/egui/src/widgets/text_edit/builder.rs:750:9] state.ime_enabled = true`
ime_enabled is true, seems like winit sends that event. 

maybe a regression there:
https://github.com/rust-windowing/winit/issues/3092

Sadly it affects egui too

rustbasic:
> You are right, it's most likely not related to your merge request
> 
> `[crates/egui/src/widgets/text_edit/builder.rs:750:9] state.ime_enabled = true` ime_enabled is true, seems like winit sends that event.
> 
> maybe a regression there: [rust-windowing/winit#3092](https://github.com/rust-windowing/winit/issues/3092)
> 
> Sadly it affects egui too

Yes. There seem to be a lot of bug reports related to IME in `winit` + `LINUX`. I created a separate option related to `ime_enable`, but it was not approved. When I have time, I will commit a separate option again.

rustbasic:
@StarStarJ 
#4896 has been applied to `egui Master version`, is it still in a weird state?

StarStarJ:
Yes, still buggy.
[...]

PR #5188, opened 2024-09-30
Fix backspace/arrow keys in TextEdit on Linux

emilk:
* Closes https://github.com/emilk/egui/issues/5008
* Closes https://github.com/emilk/egui/pull/5182
* Bug introduced in https://github.com/emilk/egui/pull/4912

I suspect this will make IME no longer work on Linux, though I don't know if it ever worked.
I rather have backspace/arrows working though.

Please help test this (I don't have Linux!)

# Tested on
* [x] Mac
* [ ] Linux Wayland
* [x] Linux X11

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -329,39 +329,43 @@ impl State {
             }
 
             WindowEvent::Ime(ime) => {
-                // on Mac even Cmd-C is pressed during ime, a `c` is pushed to Preedit.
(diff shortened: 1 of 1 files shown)


crumblingstatue:
I tested it on Linux X11, and backspace and arrow keys are working in text inputs.

YgorSouza:
Backspace and arrow keys working in Wayland as well, although they already worked before, and so did IME (e.g., compose key), which now no longer works of course.

PR #5198, opened 2024-09-30
Re-enable IME support on Linux

YgorSouza:
Reverts #5188 and adds a different fix to restore IME on Linux without breaking the backspace and arrow keys.

* Closes https://github.com/emilk/egui/issues/5544
* Closes https://github.com/emilk/egui/pull/5198

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -329,43 +329,39 @@ impl State {
             }
 
             WindowEvent::Ime(ime) => {
-                if cfg!(target_os = "linux") {
-                    // We ignore IME events on linux, because of https://github.com/emilk/egui/issues/5008
-                } else {
-                    // on Mac even Cmd-C is pressed during ime, a `c` is pushed to Preedit.
-                    // So no need to check is_mac_cmd.
-                    //
-                    // How winit produce `Ime::Enabled` and `Ime::Disabled` differs in MacOS
-                    // and Windows.
-                    //
-                    // - On Windows, before and after each Commit will produce an Enable/Disabled
-                    // event.
-                    // - On MacOS, only when user explicit enable/disable ime. No Disabled
-                    // after Commit.
-                    //
-                    // We use input_method_editor_started to manually insert CompositionStart
-                    // between Commits.
-                    match ime {
-                        winit::event::Ime::Enabled => {
-                            self.ime_event_enable();
-                        }
-                        winit::event::Ime::Preedit(text, Some(_cursor)) => {
-                            self.ime_event_enable();
-                            self.egui_input
-                                .events
-                                .push(egui::Event::Ime(egui::ImeEvent::Preedit(text.clone())));
-                        }
-                        winit::event::Ime::Commit(text) => {
-                            self.egui_input
-                                .events
-                                .push(egui::Event::Ime(egui::ImeEvent::Commit(text.clone())));
-                            self.ime_event_disable();
-                        }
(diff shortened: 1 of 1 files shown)


YgorSouza:
@rustbasic @crumblingstatue can you test this?

I've tested with Fedora 40/Gnome 46 with both the regular (Wayland) session and the XOrg session, with the compose key, the Unicode input (Ctrl+Shift+U) and the Chinese (Intelligent Pinyin) input source. Everything seems to work, and the backspace and arrow keys work on both as well.

crumblingstatue:
I can't test IME, because fcitx (& pals) somehow break my X11 setup (don't ask me how), but I can confirm that backspace and arrow keys work.

rustbasic:
Dear `emilk` & Dear `YgorSouza`

My opinion is,

* #5188
It seems like it's impossible to input languages ​​other than English that require an IME on Linux.

* #5182
For non-English users using Linux, I think #5182 is the right way, because we can control turning it on/off as needed.
If there are any parts we don't like, we can just edit them.

* #5198
#5198 It should work fine on Windows, as there are no changes. 
I'm not sure if it will work well on some `Linux + Chinese`.

YgorSouza:
> I'm not sure if it will work well on some Linux + Chinese

Is there anyone who can test this? I mean, it seems fine to me, but I'm not Chinese so I don't know.

image

https://github.com/emilk/egui/pull/5182 looks like it would require a major release? It's adding fields to structs, so I guess that's a breaking change. But if we can get the IME to work correctly on all platforms, maybe it won't be needed? And egui-winit already has an `allow_ime` field, so we could block IME events at this level, if needed, without involving the `TextEdit` or any other widgets.

https://github.com/emilk/egui/blob/ce744e6f7a22412cecf1cf2ea8678344bfb56489/crates/egui-winit/src/lib.rs#L844-L849

rustbasic:
`allow_ime` turns the entire IME function of the OS ON/OFF, so it is not the way we expect.

As you may have noticed, 
some Linux and some Chinese input IMEs have a problem with `IME::enable` that persists from the program startup,
so it is always recognized as IME input state.
It seems necessary to directly control ON/OFF.

YgorSouza:
The change in this PR ignores the Enabled event on Linux, where on X11 it seems to just mean that the window supports IME (and sometimes it can't even be turned off). That should hopefully be enough to prevent egui from going into IME mode in applications where IME isn't used at all. It is only enabled in the Preedit event, which contains a string and a cursor range, so we're definitely using IME if we get that event, since winit can't just invent this data out of nowhere.

If this still causes problems, I think we need an issue explaining what those problems are and on which systems they occur, so we can think about a solution that doesn't break anything for anyone else.

emilk (approved):
Code LGTM, but we need to test this properly:

* [x] X11 without IME
* [ ] X11 with IME
* [ ] Wayland without IME
* [ ] Wayland with IME

rustbasic:
Another way is to apply #5198, and wait for people's test results.

iamazy:
> Code LGTM, but we need to test this properly:
> 
> * [ ]  X11 without IME
> * [ ]  X11 with IME
> * [ ]  Wayland without IME
> * [ ]  Wayland with IME

I test in `X11 without IME` and `X11 with IME`, it works well.

tmvkrpxl0:
I tried it out on Wayland with fcitx5 running `custom_font` example. Korean input works well, although I had to add custom font that supports Korean font.

emilk:
Thanks for the help testing!

Issue #5183, opened 2024-09-28
Allow easier setting of background color for TextEdit


AlexanderSchuetz97:
**Is your feature request related to a problem? Please describe.**
Setting background color on TextEdit is pain. 
I require a few text edits to have different background colors on a single GUI.
for example one currently has to do something like this to set the background color of a single text edit:
```rust
  let old_color = ui.visuals_mut().extreme_bg_color;
  ui.visuals_mut().extreme_bg_color = RED;
  ui.add(egui::TextEdit::singleline(&mut self.my_awesome_string));
  ui.visuals_mut().extreme_bg_color = old_color;
```

**Describe the solution you'd like**
```rust
  ui.add(egui::TextEdit::singleline(&mut self.my_awesome_string).background_color(RED));
```

**Describe alternatives you've considered**
Workaround as described in Feature Request is what I currently use.

**Additional context**
RED is a constant Color32 in the examples.

After looking at the TextEdit's show implementation I think I would be capable of implementing this myself and make a PR if you do not have the time to implement it yourself. However I would only make this effort if you actually want this "feature". Please let me know what you think

YgorSouza:
For the record, you can also use [Ui::scope](https://docs.rs/egui/0.29.0/egui/struct.Ui.html#method.scope) for that.

```rs
ui.scope(|ui| {
    ui.visuals_mut().extreme_bg_color = Color32::RED;
    ui.text_edit_singleline(&mut self.name);
});
```

PR #5203, opened 2024-10-01
Allow easier setting of background color for TextEdit

bircni:
* Closes <https://github.com/emilk/egui/issues/5183>

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -203,7 +203,8 @@ impl<'t> TextEdit<'t> {
         self
     }
 
-    /// Set the background color of the [`TextEdit`].
+    /// Set the background color of the [`TextEdit`]. The default is [`crate::Visuals::extreme_bg_color`].
+    // TODO: remove this once #3284 is implemented
     #[inline]
     pub fn background_color(mut self, color: Color32) -> Self {
         self.background_color = Some(color);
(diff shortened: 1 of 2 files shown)


lucasmerlin:
Is there a reason why one would often need to change the background color of TextEdits?
Someone could argue that if we have this functionality for the background, we should also add functions to customize the font, size, padding, etc... I think this could get out of hand quickly.
If something like this is desirable be useful I think we should come up with some way to generalize this and add this to all widgets (and I think then this is related to #3284)

AlexanderSchuetz97:
@lucasmerlin 
I require this for 2 reasons:
Really highlight text edits in red if they contain invaluid data that the human must change.

I use gray background for every other textbox because with white background users didnt even realize the textbox exists until they moved their cursor over them.

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -201,6 +203,13 @@ impl<'t> TextEdit<'t> {
         self
     }
 
+    /// Set the background color of the [`TextEdit`].
```
lucasmerlin, line 206: Maybe you can add a comment that by default this comes from `Visuals::extreme_bg_color`?

lucasmerlin (approved):
I talked with @emilk and we decided that since extreme_bg_color is kind of obscure it makes sense to add this until #3284 is implemented

bircni:
Okay! I will also add a comment to refer to the issue you mentioned

lucasmerlin:
Thank you!

PR #5204, opened 2024-10-01
egui-wgpu: Fix docs.rs build

### LINKED ISSUE #5202: egui-wgpu docs build is failing since 0.29
[heres the build log](https://docs.rs/crate/egui-wgpu/0.29.0/builds/1371185), seems like the winit feature for raw window handle is missing

lucasmerlin:
This enables the rwh_06 feature in winit which is required to correctly build egui-wgpu
- fixes #5202

### DIFF
--- a/crates/egui-wgpu/Cargo.toml
+++ b/crates/egui-wgpu/Cargo.toml
@@ -37,7 +37,7 @@ default = ["fragile-send-sync-non-atomic-wasm"]
 puffin = ["dep:puffin"]
 
 ## Enable [`winit`](https://docs.rs/winit) integration. On Linux, requires either `wayland` or `x11`
-winit = ["dep:winit"]
+winit = ["dep:winit", "winit/rwh_06"]
 
 ## Enables Wayland support for winit.
 wayland = ["winit?/wayland"]


crates/egui-wgpu/Cargo.toml
```diff
@@ -37,7 +37,7 @@ default = ["fragile-send-sync-non-atomic-wasm"]
 puffin = ["dep:puffin"]
 
 ## Enable [`winit`](https://docs.rs/winit) integration. On Linux, requires either `wayland` or `x11`
-winit = ["dep:winit"]
+winit = ["dep:winit", "winit/rwh_06"]
```
YgorSouza, line 40: Could do the same as eframe for consistency, although it's the same result in the end.

https://github.com/emilk/egui/blob/24cd711f5b9fce1e0ee28dbe518651b1a84899f8/crates/eframe/Cargo.toml#L149

lucasmerlin: Looking like it's already enabled for eframe? 🤔

Wumpf (approved):
thanks! 👍

PR #5208, opened 2024-10-02
Fix: Popup incorrectly calculates possible min width of popup causing panic

### LINKED ISSUE #5206: Popup incorrectly calculates possible min width of popup causing panic
Debug build panics due to incorrect calculation here:
https://github.com/emilk/egui/blob/24cd711f5b9fce1e0ee28dbe518651b1a84899f8/crates/egui/src/containers/popup.rs#L410

and later panic is caused inside set_min_width debug assert.

**To Reproduce**
Create button with very small width and create popup for it:
```rs
            let button = ui.add(egui::Button::new(""));
            let popup_id = ui.id().with("test");

            if button.clicked() {
                ui.memory_mut(|memory| memory.open_popup(popup_id));
            }
            egui::popup_above_or_below_widget(
                ui,
                popup_id,
                &button,
                AboveOrBelow::Above,
                egui::PopupCloseBehavior::CloseOnClickOutside,
                |ui| {
                    ui.label("Test?");
                },
            );
```

**Expected behavior**
Correctly shows popup.

**Additional information**
Currently tested with egui 0.28, but from source code seems that the issue still persists in 0.29.

rustbasic:
Fix: Popup incorrectly calculates possible min width of popup causing panic

* Closes #5206

### DIFF
--- a/crates/egui/src/containers/popup.rs
+++ b/crates/egui/src/containers/popup.rs
@@ -407,7 +407,7 @@ pub fn popup_above_or_below_widget<R>(
 
     let frame = Frame::popup(parent_ui.style());
     let frame_margin = frame.total_margin();
-    let inner_width = widget_response.rect.width() - frame_margin.sum().x;
+    let inner_width = (widget_response.rect.width() - frame_margin.sum().x).max(0.0);
 
     parent_ui.ctx().pass_state_mut(|fs| {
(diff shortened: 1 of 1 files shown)


bircni (approved):
lgtm

emilk (approved):

PR #5211, opened 2024-10-02
iOS: Support putting UI next to the dynamic island

### LINKED ISSUE #3547: Black bar at bottom on iOS targets
**Describe the bug**

There's a black bar at the bottom of the screen when running egui projects as an iOS app. In landscape orientation this shows up on the side instead.

**To Reproduce**

1. Set up an iOS Xcode project that runs an egui example. (if more details are necessary I can describe how this is done)
2. Use a physical iPhone device as a run destination. (a Mac is needed for this. Alternatively you can use MacInCloud and Diawi, and then download the IPA through your iPhone)

**Expected behavior**

I expected the UI to show correctly (i.e. no black bar at the bottom).

**Screenshots**

See https://github.com/emilk/egui/issues/3117#issuecomment-1694711382

**Smartphone (please complete the following information):**
 - Device: iPhone 12
 - OS: iOS 17.0.3

frederik-uni:
winit::Window::inner_size returns size of safe area on iOS. use winit::Window::outer_size on iOS
The dimensions of outer_size include the title bar and borders, but as far as I know there is no way to actually display the title bar or borders on iOS so it should be fine.

* Closes #3547 

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -36,10 +36,14 @@ use winit::{
 };
 
 pub fn screen_size_in_pixels(window: &Window) -> egui::Vec2 {
-    #[cfg(target_os = "ios")]
-    let size = window.outer_size();
-    #[cfg(not(target_os = "ios"))]
-    let size = window.inner_size();
+    let size = if cfg!(target_os = "ios") {
+        // `outer_size` Includes the area behind the "dynamic island".
+        // It is up to the eframe user to make sure the dynamic island doesn't cover anything important.
+        // That will be easier once https://github.com/rust-windowing/winit/pull/3890 lands 
+        window.outer_size();
+    } else {
+		window.inner_size()
+	};
     egui::vec2(size.width as f32, size.height as f32)
 }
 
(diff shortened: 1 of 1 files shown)


lucasmerlin (approved):
I think this makes sense. Now we will still have to manually work around the safe area (until this is done: https://github.com/rust-windowing/winit/pull/3890) but at least egui will render on the whole screen

frederik-uni:
@lucasmerlin I updated my safe area implementation so it is actually usable. it could be merged and the get_safe_area_insets could be replaced while upgrading winit in the future

frederik-uni:
@emilk is this alright?

crates/egui-winit/src/lib.rs
```diff
@@ -36,6 +36,9 @@ use winit::{
 };
 
 pub fn screen_size_in_pixels(window: &Window) -> egui::Vec2 {
+    #[cfg(target_os = "ios")]
+    let size = window.outer_size();
+    #[cfg(not(target_os = "ios"))]
     let size = window.inner_size();
```
emilk, line 42: We should explain this better. Does this make sense?

```suggestion
    let size = if cfg!(target_os = "ios") {
        // `outer_size` Includes the area behind the "dynamic island".
        // It is up to the eframe user to make sure the dynamic island doesn't cover anything important.
        // That will be easier once https://github.com/rust-windowing/winit/pull/3890 lands 
        window.outer_size();
    } else {
		window.inner_size()
	};
```

emilk (approved):

emilk (approved):

Issue #5210, opened 2024-10-02
EXC_BAD_ACCESS in `Color32::from_rgba_unmultiplied` on iOS


lucasmerlin:
When I'm running my app for iOS with a debug build I'm now getting a EXC_BAD_ACCESS (Maybe this is a stackoverflow?). I guess this came with https://github.com/emilk/egui/pull/5088, it seems like [u8; 65536] crosses some limit? When running a release build, everything works as expected. 

<img width="739" alt="image" src="https://github.com/user-attachments/assets/7f093993-29c1-4258-9601-ca4daac2b82c">

emilk:
Maybe we just need to put it on the heap, i.e. use a `Vec` or `Box<[u8; 256 * 256]>` instead in the `OnceLock`

PR #5212, opened 2024-10-02
Use boxed slice for lookup table to avoid stack overflow

YgorSouza:
* Closes <https://github.com/emilk/egui/issues/5210>

### DIFF
--- a/crates/ecolor/src/color32.rs
+++ b/crates/ecolor/src/color32.rs
@@ -111,7 +111,7 @@ impl Color32 {
                 static LOOKUP_TABLE: OnceLock<Box<[u8]>> = OnceLock::new();
                 let lut = LOOKUP_TABLE.get_or_init(|| {
                     use crate::{gamma_u8_from_linear_f32, linear_f32_from_gamma_u8};
-                    (0..u16::MAX)
+                    (0..=u16::MAX)
                         .map(|i| {
                             let [value, alpha] = i.to_ne_bytes();
                             let value_lin = linear_f32_from_gamma_u8(value);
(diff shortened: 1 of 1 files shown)


YgorSouza:
@lucasmerlin can you test?

This is slightly slower according to the benchmark (about 25%), but it should hopefully avoid stack problems. I'm actually not sure where the compiler was putting the array before. I heard something about objects always being created in the stack and then moved to the heap or to where they are supposed to be (although sometimes this is optimized away), so I went with a boxed slice rather than a boxed array to be sure. I think this is called the "placement new" issue?

lucasmerlin (approved):
Awesome, just tried it and it fixes the problem! 

The following also worked, and I think it's a bit cleaner (and closer to your original implementation):
```rust
    static LOOKUP_TABLE: OnceLock<Box<[u8; 256 * 256]>> = OnceLock::new();
    let lut = LOOKUP_TABLE.get_or_init(|| {
        use crate::{gamma_u8_from_linear_f32, linear_f32_from_gamma_u8};
        Box::new(core::array::from_fn(|i| {
            let [value, alpha] = (i as u16).to_ne_bytes();
            let value_lin = linear_f32_from_gamma_u8(value);
            let alpha_lin = linear_f32_from_linear_u8(alpha);
            gamma_u8_from_linear_f32(value_lin * alpha_lin)
        }))
    });
```
Is there a reason you don't use a `[u8; 256 * 256]` anymore?

YgorSouza:
> Is there a reason you don't use a `[u8; 256 * 256]` anymore?

I wasn't sure it would work, as I mentioned above. I thought the array would be constructed in the stack and then moved to the heap without optimizations, so it would be the same problem in the end.

PR #5214, opened 2024-10-02
enable `manual_range_contains` lint

bircni:
Enabled the manual_range_contains lint 


rustbasic:
I don't like the restriction of the `manual_range_contains` lint.

bircni:
I think this makes the code more readable I Like it

rustbasic:
> I think this makes the code more readable I Like it

Yes, it improves readability, but I think it would be better to allow it.
Freedom is being reduced more and more.
just my opinion.

emilk:
No, I don't like it

PR #5219, opened 2024-10-03
Default in egui::Rect

rustbasic:
Default in egui::Rect

I just used it even though it was inconvenient, but I think it would be nice if it was supported by default.


emilk (commented):
No; read the comments above it

PR #5226, opened 2024-10-06
Free textures after submitting queue instead of before with wgpu renderer

### LINKED ISSUE #5224: eframe with wgpu panics when unloading texture in same frame as rendering it
**Describe the bug**
When using eframe with the wgpu renderer the program may panic, if you drop a `TextureHandle` in the same frame as it is used for rendering. The panic happens when `egui_wgpu::winit::Painter::paint_and_update_textures` submits the `wgpu::Queue` in [crates/egui-wgpu/src/winit.rs:688](https://github.com/emilk/egui/blob/0f2b427ff4c0a8c68f6622ec7d0afb7ba7e71bba/crates/egui-wgpu/src/winit.rs#L688). The unloading of the texture happens _before_ that in [line 673](https://github.com/emilk/egui/blob/0f2b427ff4c0a8c68f6622ec7d0afb7ba7e71bba/crates/egui-wgpu/src/winit.rs#L673). I barely understand any of the code in that function, but if I just move the texture unloading to after the queue submitting (e.g. [line 693](https://github.com/emilk/egui/blob/0f2b427ff4c0a8c68f6622ec7d0afb7ba7e71bba/crates/egui-wgpu/src/winit.rs#L693)), my code doesn't crash anymore.
If this is the correct fix, this means that destroying a texture is not a queued command. I tried to find out about that, but only found:
- The [documentation](https://docs.rs/wgpu/latest/wgpu/struct.Texture.html#method.destroy) of `Texture::destroy`: "Destroy the associated native resources as soon as possible."
- [This issue](https://github.com/gfx-rs/wgpu/issues/964) with associated PR which covers the topic of destroying buffers and textures, and _may_ imply my above assumption.
If this is not the correct fix, there may be a bug in wgpu itself.

**To Reproduce**
1. Run the following eframe program with the egui-wgpu feature enabled:
```rust
use eframe::{
    self,
    egui,
};

struct MyApp {}

impl eframe::App for MyApp {
    fn update(&mut self, context: &egui::Context, _frame: &mut eframe::Frame) {
        // load texture
        let texture = context.load_texture("my image", egui::ColorImage::example(), Default::default());

        // use texture
        egui::CentralPanel::default().show(context, |ui| {
            ui.image(egui::ImageSource::Texture((&texture).into()));
        });

        context.request_repaint();
        // texture is destroyed
    }
}

fn main() -> eframe::Result {
    let native_options = eframe::NativeOptions {
        ..Default::default()
    };
    eframe::run_native("My App", native_options, Box::new(|_| Ok(Box::new(MyApp{}))))
}
```
2. See the panic with the following message:
```
Error in Queue::submit: Validation Error

Caused by:
  Texture with 'egui_texid_Managed(1)' label has been destroyed
```

**Expected behavior**
The example picture should be shown in a window and the code should not panic.

**Additional context**
I encountered this issue in a project in which I want to display the image files in a folder as described in #5200. Since I may not be able to hold all images in memory, I unload them regularly depending on the selection of the user. If I scroll through the images too quickly, the crash occurs.
In my project the crash happens quite randomly, but with the example program above I can reliably produce a panic on frame 1. I'm not sure if this will work on every machine though.

Rusty-Cube:
* Closes #5224

I'm unfamiliar with wgpu, so I'd like someone to confirm, that calling `wgpu::Texture` _after_ `wgpu::Queue::submit` is in fact the right thing to do.

### DIFF
--- a/crates/egui-wgpu/src/winit.rs
+++ b/crates/egui-wgpu/src/winit.rs
@@ -684,6 +684,9 @@ impl Painter {
             vsync_sec += start.elapsed().as_secs_f32();
         };
 
+        // Free textures marked for destruction **after** queue submit since they might still be used in the current frame.
+        // Calling `wgpu::Texture::destroy` on a texture that is still in use would invalidate the command buffer(s) it is used in.
+        // However, once we called `wgpu::Queue::submit`, it is up for wgpu to determine how long the underlying gpu resource has to live.
         {
             let mut renderer = render_state.renderer.write();
             for id in &textures_delta.free {
--- a/crates/egui-wgpu/src/winit.rs
+++ b/crates/egui-wgpu/src/winit.rs
@@ -668,13 +668,6 @@ impl Painter {
             );
         }
 
-        {
-            let mut renderer = render_state.renderer.write();
-            for id in &textures_delta.free {
-                renderer.free_texture(id);
-            }
-        }
-
         let encoded = {
             crate::profile_scope!("CommandEncoder::finish");
             encoder.finish()
@@ -690,6 +683,13 @@ impl Painter {
                 .submit(user_cmd_bufs.into_iter().chain([encoded]));
             vsync_sec += start.elapsed().as_secs_f32();
         };
+
+        {
+            let mut renderer = render_state.renderer.write();
+            for id in &textures_delta.free {
+                renderer.free_texture(id);
+            }
+        }
 
         let screenshot = if capture {
             self.screen_capture_state


lucasmerlin:
> I'd like someone to confirm, that calling wgpu::Texture after wgpu::Queue::submit is in fact the right thing to do.

@Wumpf should be able to answer that question

crates/egui-wgpu/src/winit.rs
```diff
@@ -691,6 +684,13 @@ impl Painter {
             vsync_sec += start.elapsed().as_secs_f32();
         };
 
+        {
```
Wumpf, line 687:
```suggestion
        // Free textures marked for destruction **after** queue submit since they might still be used in the current frame.
        // Calling `wgpu::Texture::destroy` on a texture that is still in use would invalidate the command buffer(s) it is used in.
        // However, once we called `wgpu::Queue::submit`, it is up for wgpu to determine how long the underlying gpu resource has to live.
        {
```

Wumpf (approved):
makes sense to me! 👍 
I added a comment with some fineprint to further motivate the change

PR #5240, opened 2024-10-09
Report egui::Window contents as children to accesskit

Wcubed:
Previously, all widgets would be listed in accesskit as children of the toplevel window.
With this change, they will be reported as child of the `egui::Window` they are in, Which should increase parseability of the ui for screenreaders and integration tests.

Added an accesskit test to check that it is indeed working.

### DIFF
--- a/crates/egui/src/containers/area.rs
+++ b/crates/egui/src/containers/area.rs
@@ -569,6 +569,14 @@ impl Prepared {
         ui
     }
 
+    pub(crate) fn with_widget_info(&self, make_info: impl Fn() -> crate::WidgetInfo) {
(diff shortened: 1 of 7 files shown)


lucasmerlin:
Awesome! 
I also stumbled upon this in my work on #5166, I think it'd actually make sense if all children of a ui are automatically added as a child of that ui in accesskit? So e.g. if you create a `ui.group()` with a button, that button but be child of the group? 

Also, once #5166 is done we should be able to simplify the accesskit tests using kittest

Wcubed:
That was indeed my idea as well 😄.

emilk (approved):

PR #5247, opened 2024-10-11
Expose center position in `MultiTouchInfo`

lucasmerlin:
Expose the center / average position in MultiTouchInfo

### DIFF
--- a/crates/egui/src/input_state/touch_state.rs
+++ b/crates/egui/src/input_state/touch_state.rs
@@ -60,6 +60,9 @@ pub struct MultiTouchInfo {
     /// Note 2: Just increasing the physical pressure without actually moving the finger may not
     /// necessarily lead to a change of this value.
     pub force: f32,
+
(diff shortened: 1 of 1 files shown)


crates/egui/src/input_state/touch_state.rs
```diff
@@ -60,6 +60,9 @@ pub struct MultiTouchInfo {
     /// Note 2: Just increasing the physical pressure without actually moving the finger may not
     /// necessarily lead to a change of this value.
     pub force: f32,
+
+    /// Center position of the current gesture (average of all touch points).
+    pub center_pos: Pos2,
```
emilk, line 65: nit: I think this should come earlier in the struct. Putting it last means it is the last thing users will see in the docs.

emilk (approved):

crates/egui/src/input_state/touch_state.rs
lucasmerlin: Moved it below start_pos

PR #5250, opened 2024-10-11
Fix: Speed ​​issues on Windows when immediate viewport is on another screen

### LINKED ISSUE #5249: Speed ​​issues on Windows when `immediate viewport` is on another screen
OS: Windows10
egui: Master & All

If the `Immediate Viewport` created by a program on Windows is on a different screen than the program itself, the speed will decrease to about 30-40 FPS.

rustbasic:
Fix Issues : Speed ​​issues on Windows when immediate viewport is on another screen

On Windows, we should use `EventResult::RepaintNow()` in `Immediate Viewport` to avoid FPS slowdown.

Improvement : In `glow` and `wgpu`, we are removing `egui_ctx` and only using `integration.egui_ctx`.
This simplifies the structure and reduces the possibility of bugs.

* Related #4022
* Closes #5249

### DIFF
--- a/crates/eframe/src/native/wgpu_integration.rs
+++ b/crates/eframe/src/native/wgpu_integration.rs
@@ -65,7 +65,6 @@ struct WgpuWinitRunning<'app> {
 ///
 /// Wrapped in an `Rc<RefCell<…>>` so it can be re-entrantly shared via a weak-pointer.
 pub struct SharedState {
-    egui_ctx: egui::Context,
     viewports: Viewports,
     painter: egui_wgpu::winit::Painter,
     viewport_from_window: HashMap<WindowId, ViewportId>,
@@ -134,8 +133,8 @@ impl<'app> WgpuWinitApp<'app> {
 
         for viewport in viewports.values_mut() {
             viewport.initialize_window(
-                event_loop,
                 &running.integration.egui_ctx,
+                event_loop,
                 viewport_from_window,
                 painter,
             );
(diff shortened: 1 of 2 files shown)


crates/eframe/src/native/wgpu_integration.rs
```diff
@@ -702,11 +702,6 @@ impl<'app> WgpuWinitRunning<'app> {
             viewport_from_window,
         );
 
-        // Prune dead viewports:
-        viewports.retain(|id, _| active_viewports_ids.contains(id));
-        viewport_from_window.retain(|_, id| active_viewports_ids.contains(id));
-        painter.gc_viewports(&active_viewports_ids);
```
emilk, line 708: What happened with this code?

crates/eframe/src/native/wgpu_integration.rs
```diff
@@ -954,8 +949,6 @@ fn render_immediate_viewport(
         input
     };
 
-    let egui_ctx = shared.borrow().egui_ctx.clone();
```
emilk, line 957: What is this `borrow()` that caused performance problems?

crates/eframe/src/native/wgpu_integration.rs
rustbasic: > What happened with this code?

It is not necessary as it is handled in `handle_viewport_output()`.

crates/eframe/src/native/wgpu_integration.rs
rustbasic: Improvement : In glow and wgpu, we are removing egui_ctx and only using integration.egui_ctx.

Fix Issues: On Windows, we should use EventResult::RepaintNow() in Immediate Viewport to avoid FPS slowdown.

PR #5257, opened 2024-10-13
Expand max font atlas size from 8k to 16k

### LINKED ISSUE #5256: epaint Texture Atlas Overflow with Large Font Character Sets
When using fonts with an average of 50,000 characters,
'epaint texture atlas overflowed!' may be printed and cause problems.
It is necessary to expand the max value related to texture.

rustbasic:
When using fonts with an average of 50,000 characters,
'epaint texture atlas overflowed!' may be printed and cause problems.
It is necessary to expand the max value related to texture.

* Closes #5256

### DIFF
--- a/crates/epaint/src/texture_atlas.rs
+++ b/crates/epaint/src/texture_atlas.rs
@@ -159,8 +159,8 @@ impl TextureAtlas {
     }
 
     fn max_height(&self) -> usize {
-        // the initial width is likely the max texture side size
-        self.image.width()
+        // likely the max texture height
+        self.image.height().max(self.image.width())
     }
 
     /// When this get high, it might be time to clear and start over!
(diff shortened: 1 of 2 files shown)


crates/epaint/src/texture_atlas.rs
```diff
@@ -159,8 +159,8 @@ impl TextureAtlas {
     }
 
     fn max_height(&self) -> usize {
-        // the initial width is likely the max texture side size
-        self.image.width()
+        // likely the max texture height
+        self.image.height().max(self.image.width())
```
emilk, line 163:
```suggestion
        // the initial width is set to the max size
        self.image.height().max(self.image.width())
```

emilk (approved):

Issue #5255, opened 2024-10-12
`Ctrl+Shift+Z` should restore data to the state that was before like `Ctrl+Y` does


AngelicosPhosphoros:
**Is your feature request related to a problem? Please describe.**

Well, I often use keybinding `Ctrl+Shift+Z` to restore text state after I accidentally undid too much changes using `Ctrl+Z`. In egui controls, e.g. text edit, `Ctrl+Shift+Z` doesn't behave like rest of such inputs in my system (Windows 10).

**Describe the solution you'd like**
`Ctrl+Shift+Z` keybinding should do redo operation, like `Ctrl+Y` does.

**Describe alternatives you've considered**
Do nothing and tell users to use `Ctrl+Y`. It works on egui controls.

**Additional context**
Well, I use computers the biggest part of my life and always used `Ctrl+Shift+Z` to cancel `Ctrl+Z`. Only when opening this issue, I checked if this is actually standardized and found that there exists a `Ctrl+Y` keybinding. I think, it is common situation so supporting both `Ctrl+Y` and `Ctrl+Shift+Z` would be convenient for users.

Maybe, it is a wrong repository to report this and I should report it to winit, I don't really know.

YgorSouza:
No, this has nothing to do with winit, and it's actually a bug in egui. The Ctrl+Shift+Z shortcut is supposed to work as redo as well:

https://github.com/emilk/egui/blob/23728e145ec52bd1193f6f0123973763de4dbb3d/crates/egui/src/widgets/text_edit/builder.rs#L997-L1015

But it doesn't, because the previous match arm (undo) is also triggered by this same key combination:

https://github.com/emilk/egui/blob/23728e145ec52bd1193f6f0123973763de4dbb3d/crates/egui/src/widgets/text_edit/builder.rs#L980-L996

This `matches_logically` method is supposed to work like this, at least according to its doctests:

https://github.com/emilk/egui/blob/23728e145ec52bd1193f6f0123973763de4dbb3d/crates/egui/src/data/input.rs#L769-L786

But it is being used incorrectly here, causing both Ctrl+Z and Ctrl+Shift+Z to work as undo.

Mingun:
This code looks awful
https://github.com/emilk/egui/blob/23728e145ec52bd1193f6f0123973763de4dbb3d/crates/egui/src/widgets/text_edit/builder.rs#L1002-L1004
[...]

PR #5258, opened 2024-10-13
Fix Ctrl+Shift+Z redo shortcut

YgorSouza:
This shortcut was previously triggering the Undo action due to the matches_logically method ignoring the state of the Shift key. This was solved by simply inverting the order of the undo and redo arms, so the undo is not matched if the shortcut corresponds to redo.

* Closes <https://github.com/emilk/egui/issues/5255>

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -977,23 +977,7 @@ fn events(
                     break;
                 }
             }
-            Event::Key {
-                key: Key::Z,
-                pressed: true,
-                modifiers,
-                ..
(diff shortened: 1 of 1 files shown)


YgorSouza:
Not sure if this is the best solution, or if it would be better to just use `matches_exact` instead. I don't know if there would be any undesirable side effects on different keyboards.

bircni (approved):
lgtm

maybe it would be good to use [KeyboardShortcut](https://docs.rs/egui/0.29.1/egui/struct.KeyboardShortcut.html) here

emilk (approved):
Good catch

PR #5262, opened 2024-10-14
Fix: Layout does not consider the sizing_pass field of `UiBuilder`

zhatuokun:
During the sizing pass we want widgets to use up as little space as possible, but it does not consider the sizing_pass field of `UiBuilder`.


lucasmerlin (approved):

emilk:
Good catch!

PR #5265, opened 2024-10-14
Compile Caching

bircni:
Added compile caching with sccache to github workflows to make them faster

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -88,6 +88,8 @@ jobs:
   check_wasm:
     name: Check wasm32 + wasm-bindgen
     runs-on: ubuntu-22.04
+    env:
+      RUSTC_WRAPPER: ""
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/rust.yml
```diff
@@ -124,6 +134,10 @@ jobs:
 
       - name: Set up cargo cache
         uses: Swatinem/rust-cache@v2
+
+      - name: Run sccache-cache
+        uses: mozilla-actions/sccache-action@v0.0.6
+
```
lucasmerlin, line 140: Does it make sense to use two different build caches?

lucasmerlin (commented):
How much does the pipeline time decrease with these changes?

emilk (commented):
We already have https://github.com/Swatinem/rust-cache set up for the dependencies - how is this different?

bircni:
When using local cache for local builds this could improve build times massively otherwise here it seems like we unfortunately dont have any improvements -> will have a look at how to improve local compile caching!

PR #5275, opened 2024-10-16
Fix textedit intrinsic size metric

lucasmerlin:
Since textedit is doing the justify layout calculation itself, we need to report the original desired_size as intrinsic size, instead of the value passed to allocate_space.

I wonder though, is it still necessary that the TextEdit does the justify calculation itself instead of relying on the ui layout to do it? As far as I understand it, justify should be handled by the ui.allocate_space call.

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -525,13 +525,13 @@ impl<'t> TextEdit<'t> {
 
         let mut galley = layouter(ui, text.as_str(), wrap_width);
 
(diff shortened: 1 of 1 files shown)


emilk:
> I wonder though, is it still necessary that the TextEdit does the justify calculation itself instead of relying on the ui layout to do it? As far as I understand it, justify should be handled by the ui.allocate_space call.

I think `ui.allocate_space` should be able to handle it, unless I'm missing something. Could just be legacy.

emilk (approved):

PR #5276, opened 2024-10-16
Put font data into Arc to reduce memory consumption

StarStarJ:
egui never accesses the `FontDefinitions`' member fields mutably, except in `fonts_tweak_ui` where it cloned the `FontDefinitions` object anyway.

This patch reduces system memory consumption for shared font definitions.
And also removes some overhead from copying (e.g. for the per `pixel_per_points` font atlas)

Also it allows to keep a copy of the font definitions outside of egui.

In my App that uses international fonts:
Before:
image

New:
image

Note: If `Arc` is not wanted, then it could ofc be abstracted away.

I know this is quite a breaking change API wise, but would like to hear your opinion.

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -597,7 +597,9 @@ impl ContextImpl {
                         FontPriority::Lowest => fam.push(font.name.clone()),
                     }
                 }
-                self.font_definitions.font_data.insert(font.name, font.data);
+                self.font_definitions
+                    .font_data
+                    .insert(font.name, Arc::new(font.data));
             }
 
             #[cfg(feature = "log")]
@@ -2940,7 +2942,9 @@ impl Context {
 
         for (name, data) in &mut font_definitions.font_data {
             ui.collapsing(name, |ui| {
-                if data.tweak.ui(ui).changed() {
+                let mut tweak = data.tweak;
+                if tweak.ui(ui).changed() {
+                    Arc::make_mut(data).tweak = tweak;
                     changed = true;
                 }
             });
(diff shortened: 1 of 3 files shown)


StarStarJ:
E.g. it keeps a copy in `ContextImpl` and copies it into the fonts per `pixels_per_point`:
https://github.com/emilk/egui/blob/707cd03357f926d040acb241a5c42d8e91a83d8e/crates/egui/src/context.rs#L599

from:
https://github.com/emilk/egui/blob/707cd03357f926d040acb241a5c42d8e91a83d8e/crates/egui/src/context.rs#L587-L601

emilk (commented):
I think a simpler and smaller change would be

```diff
@@ -243,7 +243,7 @@ pub struct FontDefinitions {
     /// List of font names and their definitions.
     ///
     /// `epaint` has built-in-default for these, but you can override them if you like.
-    pub font_data: BTreeMap<String, FontData>,
+    pub font_data: BTreeMap<String, Arc<FontData>>,
```

StarStarJ:
> I think a simpler and smaller change would be
> 
> ```diff
> @@ -243,7 +243,7 @@ pub struct FontDefinitions {
>      /// List of font names and their definitions.
>      ///
>      /// `epaint` has built-in-default for these, but you can override them if you like.
> -    pub font_data: BTreeMap<String, FontData>,
> +    pub font_data: BTreeMap<String, Arc<FontData>>,
> ```

Sounds good aswell

crates/egui/src/context.rs
```diff
@@ -2940,7 +2942,13 @@ impl Context {
...
         for (name, data) in &mut font_definitions.font_data {
             ui.collapsing(name, |ui| {
-                if data.tweak.ui(ui).changed() {
+                let mut tweak = data.tweak;
+                if tweak.ui(ui).changed() {
+                    *data = Arc::new(FontData {
+                        font: data.font.clone(),
+                        index: data.index,
+                        tweak,
+                    });
```
emilk, line 2951:
```suggestion
                if tweak.ui(ui).changed() {
                    Arc::make_mut(data).tweak = tweak;
```

emilk (approved):
Much better!

PR #5285, opened 2024-10-18
Drop for Painter in egui_glow

rustbasic:
Drop for Painter in egui_glow


emilk (commented):
hmm 🤔 

At this point we could just destroy in the destructor, with no need for an explicit `.destroy()` function

Issue #3734, opened 2023-12-24
ScrollArea's scrollbar goes past frame's borders


tamewild:
**Describe the bug**
When using `show_rows` with a high amounts of rows, the scrollbar can go past the frame's borders. The part of the scrollbar outside of the frame's borders can't be interacted with. Problem is exacerbated with even more rows + higher row height.

When the window is taller, the problem seems to lessen but it doesn't help that much if you have a high enough row count.

Image of the overflowed scrollbar at the top of the frame:
![image](https://github.com/emilk/egui/assets/138316124/89540e79-7d1c-45dc-845d-78d4146d75de)

Image of the overflowed scrollbar at the bottom of the frame:
![image](https://github.com/emilk/egui/assets/138316124/9354d2e0-28d9-4bf5-9f8b-9405044b1022)

**To Reproduce**
Using this code on the latest master:
```rust
use eframe::egui::{CentralPanel, ScrollArea};
use eframe::{egui, NativeOptions};
use eframe::egui::scroll_area::ScrollBarVisibility;
use eframe::egui::style::ScrollStyle;

fn main() {
    eframe::run_simple_native("", NativeOptions::default(), |ctx, frame| {
        ctx.style_mut(|x| x.spacing.scroll = ScrollStyle::solid());
        CentralPanel::default().show(ctx, |ui| {
            egui::Frame::canvas(ui.style())
                .show(ui, |ui| {
                    ScrollArea::vertical()
                        .scroll_bar_visibility(ScrollBarVisibility::AlwaysVisible)
                        .auto_shrink([false, false])
                        .show_rows(ui, 1000.0, 10000, |_, _| {});
                });
        });
    }).unwrap();
}
```

**Expected behavior**
The scrollbar should remain in the frame without causing visual incongruencies

**Additional context**
This was tested on the latest master

gilbertoalexsantos:
This issue can also be reproduced in the egui demo.

I created a PR to address it: https://github.com/emilk/egui/pull/5286

PR #5286, opened 2024-10-18
Fix scroll handle extending outside of `ScrollArea`

gilbertoalexsantos:
* Closes https://github.com/emilk/egui/issues/3734

## Info

This PR addresses an issue where resizing a scroll handle can lead to unwanted overlap.

It is also happening on the egui-demo.

Screenshot 2024-10-18 at 17 35 25

## Cause

*Note: The following explanation assumes a vertical scroll; however, the logic applies equally to horizontal scrolling.*

When the scroll handle is positioned at the top or bottom of the scroll area and the handle is resized to fit the minimum handle size, there is a risk of overlap. This occurs if the handle’s new size extends beyond the bounds of the scroll area.

## Proposed Solution

1.	Check whether increasing the handle size will cause it to overlap with the scroll area.
2.	If an overlap is detected, adjust the handle’s center position by the overlap amount, moving it towards the center of the scroll area.


lucasmerlin:
While this is a improvement, now it will look like you're at the top while already scrolled done by some amount:

https://github.com/user-attachments/assets/05fc1505-f5b4-48b8-aeec-979e16d6051e

Ideally, the scrollbar position would just be calculated correctly, taking the min size into account.

gilbertoalexsantos:
@lucasmerlin 

Oh, you're totally right. I didn't notice this behaviour 🥲

I'll work on a fix for this ☺️

gilbertoalexsantos:
@lucasmerlin 

I did an update to the PR.

1º - I extracted the handle_rect calculation to a function
2º - When calculating the handle position, I take into consideration the handle size (by subtracting it from scroll_bar size)


https://github.com/user-attachments/assets/75a06de9-e9e1-4057-9a71-3c13b3896b28

lucasmerlin (approved):
Awesome, it's working perfectly now, thank you for this! 🚀 I think we can merge this once the pipeline succeeds.

gilbertoalexsantos:
I did two more commits:
1º - Applying rustfmt (ops)
2º - Removing unused import

lucasmerlin:
Seems like you'll need to do another format. Usually it makes more sense to run rustfmt last after any other changes

gilbertoalexsantos:
Indeed.

I'm using Rust Hover, and for some reason, the format is different than: cargo fmt. Maybe they are using some format of their own.

lucasmerlin:
Assuming you mean RustRover, there is a setting to use rustfmt instead: 

image

Not sure why they don't enable this by default 🤔

gilbertoalexsantos:
Oh... That is nice. I'll toggle it. Thank you ^^

lucasmerlin:
Sorry, totally forgot about this. I'll merge it now, thank you again!

PR #5290, opened 2024-10-20
Fix: When using the animation feature of the ScrollArea, it does not move to the correct location.

### LINKED ISSUE #5289: When moving with the animation function of `ScrollArea`, it does not move to the exact location.

rustbasic:
Fix: When using the animation feature of the ScrollArea, it does not move to the correct location.

There is a part that is interfering incorrectly during the animation process.

* Closes #5289

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -883,6 +883,8 @@ impl Prepared {
 
                     if !animated {
                         state.offset[d] = target_offset;
+                    } else if let Some(_scroll_to_target) = &mut state.offset_target[d] {
+                        // Probably in the middle of animation process.
                     } else {
                         // The further we scroll, the more time we take.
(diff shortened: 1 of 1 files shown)


lucasmerlin:
But this would mean that if a scroll animation is in progress and the user clicks e.g. a scroll top button that would just not register. I think the current behaviour, while not ideal, is definitely more correct than with your changes. 

You can also see it in the web demo. I put the scroll animation duration to 1s to make it easy to see. 
With master if I quickly click scroll down and then scroll up while it hasn't completed the animation yet it will scroll back up: 

https://github.com/user-attachments/assets/5363cf4c-535c-4b1a-947c-77a3de4d77c4

With your changes it will not scroll back up:

https://github.com/user-attachments/assets/ed728d43-31bb-4f0c-b83f-ab20fe0e57f5

Issue #5297, opened 2024-10-22
Snapshot tests fail with`unity` feature


lucasmerlin:
**Describe the bug**
Running the tests with cargo test --workspace --all-features currently fails because the unity feature isn't compatible with egui_wgpu. This means we currently aren't running tests with all possible features.

**Expected behavior**
Should be possible to run tests with --all-features

This was broken in #5166 (intentionally, in order to not make that PR even bigger)

PR #5299, opened 2024-10-22
Fix cargo test in check.sh

lucasmerlin:
The check.sh script was broken in #5166, this fixes it
* Related to #5297 

### DIFF
--- a/scripts/check.sh
+++ b/scripts/check.sh
@@ -29,7 +29,7 @@ cargo check --quiet  --all-targets
 cargo check --quiet  --all-targets --all-features
 cargo check --quiet  -p egui_demo_app --lib --target wasm32-unknown-unknown
 cargo check --quiet  -p egui_demo_app --lib --target wasm32-unknown-unknown --all-features
-# TODO(lucasmerlin) re-enable --all-features once the tests work with the unity feature
+# TODO(#5297) re-enable --all-features once the tests work with the unity feature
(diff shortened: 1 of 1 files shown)


scripts/check.sh
```diff
@@ -29,7 +29,8 @@ cargo check --quiet  --all-targets
 cargo check --quiet  --all-targets --all-features
 cargo check --quiet  -p egui_demo_app --lib --target wasm32-unknown-unknown
 cargo check --quiet  -p egui_demo_app --lib --target wasm32-unknown-unknown --all-features
-cargo test  --quiet --all-targets --all-features
+# TODO(lucasmerlin) re-enable --all-features once the tests work with the unity feature
```
emilk, line 32: We should consider doing something like this so we can find the relevant places in the code from the issue number. We do this in Rerun, and it's quite useful. The GitHub extension for vs-code even correctly interprets it as an issue number to the right repository, with a preview on hover (!)

```suggestion
# TODO(#5297) re-enable --all-features once the tests work with the unity feature
```

emilk (approved):

scripts/check.sh
lucasmerlin: Would be awesome if this could also automatically show up in the relevant issue, like it does when you mention an issue in a comment. But I guess they don't because that could have a lot of false positives?

PR #5232, opened 2024-10-08
Fix: `Sides` did not apply the layout position correctly.

zhatuokun:
Run this code

```rs
        CentralPanel::default().show(ctx, |ui| {
            Sides::new().show(
                ui,
                |ui| {
                    ui.label("1");
                },
                |ui| {
                    ui.label("2");
                },
            );
            Sides::new().show(
                ui,
                |ui| {
                    ui.label("11");
                },
                |ui| {
                    ui.label("22");
                },
            );
            Sides::new().show(
                ui,
                |ui| {
                    ui.label("111");
                },
                |ui| {
                    ui.label("222");
                },
            );
        });
```

Before

before

After

after

PR #5300, opened 2024-10-23
Revert "Fix: `Sides` did not apply the layout position correctly."

emilk:
Reverts emilk/egui#5232

I should have tested it first. `cursor` can contain infinites. There is a better fix to be found @zhatuokun


zhatuokun:
That's true, so just use `available_rect_before_wrap`.

PR #5301, opened 2024-10-23
Add `Harness::new_ui`, `Harness::fit_contents`

lucasmerlin:
This adds a `Harness::new_ui`, which accepts a Ui closure and shows the ui in a central panel. One big benefit is that this allows us to add a fit_contents method that can run the ui closure with a sizing pass and resize the "screen" based on the content size.

I also used this to add a snapshot test for the rendering_test at different scales.


crates/egui_demo_lib/tests/snapshots/widget_gallery.png
lucasmerlin, line 1: I wonder, should we add some margin by default? I think no margin is more correct, since it is actually what the ui is but more margin would make the screenshots look nicer. Also, should I change the background to be transparent?

crates/egui_demo_lib/src/demo/widget_gallery.rs
```diff
@@ -300,15 +300,12 @@ mod tests {
...
         };
-        let app = |ctx: &Context| {
-            CentralPanel::default().show(ctx, |ui| {
-                demo.ui(ui);
-            });
-        };
-        let harness = Harness::builder()
-            .with_size(Vec2::new(380.0, 550.0))
+        let mut harness = Harness::builder()
             .with_dpi(2.0)
```
emilk, line 304: I should have caught this in an earlier PR, but this should be called `with_pixels_per_point`. `dpi` means `Dots per inch`, and is an absolute measurement of pixel density

crates/egui_demo_lib/tests/snapshots/widget_gallery.png
emilk: Once https://github.com/emilk/egui/issues/4019 is done, widget should just perfectly kiss the border, so we need some way to see that.

So I think it makes sense to add a transparent margin. The opaque background then signifies the rectangle of the `Ui`, but we can still see wether or not a widget paints outside of that area (which could be a bug)

crates/egui_kittest/src/harness_kind.rs
```diff
@@ -0,0 +1,72 @@
...
+        match self {
+            AppKind::Context(f) => {
+                f(ctx);
+                None
+            }
+            AppKind::Ui(f) => Some(Self::run_ui(f, ctx, false)),
+        }
+    }
+
+    pub fn run_sizing_pass(&mut self, ctx: &egui::Context) -> Option<egui::Response> {
```
emilk, line 46: This should have a docstring

crates/egui_kittest/src/harness_kind.rs
```diff
@@ -0,0 +1,72 @@
...
+        match self {
+            AppKind::Context(f) => {
+                f(ctx);
```
emilk, line 49: Maybe we should panic here, since we're not setting any sizing-pass specialness in the `Context` case?

emilk (approved):

crates/egui_kittest/src/harness_kind.rs
lucasmerlin: Oops, AppKind doesn't even need to be public, I've made it private.

crates/egui_kittest/src/harness_kind.rs
lucasmerlin: I added an assert in #5313

crates/egui_demo_lib/tests/snapshots/widget_gallery.png
lucasmerlin: Makes sense! I added an 8px margin with transparent bg and you can immedeately see that the separator is one px too wide:

image

Also, I noticed that the snapshot ignores alpha, meaning a completely transparent is the same as a completely black one and the snapshot test will succeed. I'll open an issue in dify

PR #5309, opened 2024-10-27
Impl from Box<str> for WidgetText, RichText

dimtpap:
`Box<str>` is an immutable heap-allocated string slice.
This PR makes it more convenient to use them in labels for example.

Before this PR
```rust
let text: Box<str> = "Hello".into();
ui.label(text.into_string());

let text_ref: &Box<str> = &"Hello".into();
ui.label(text_ref.clone().into_string());
// or
ui.label(text_ref.as_ref());
// or
ui.label(&**text_ref);
```
After this PR
```rust
let text: Box<str> = "Hello".into();
ui.label(text);

let text_ref: &Box<str> = &"Hello".into();
ui.label(text_ref);
```

### DIFF
--- a/crates/egui/src/widget_text.rs
+++ b/crates/egui/src/widget_text.rs
@@ -67,6 +67,27 @@ impl From<String> for RichText {
     }
 }
 
+impl From<&Box<str>> for RichText {
+    #[inline]
+    fn from(text: &Box<str>) -> Self {
+        Self::new(text.clone())
+    }
+}
+
+impl From<&mut Box<str>> for RichText {
+    #[inline]
+    fn from(text: &mut Box<str>) -> Self {
+        Self::new(text.clone())
+    }
+}
+
+impl From<Box<str>> for RichText {
+    #[inline]
+    fn from(text: Box<str>) -> Self {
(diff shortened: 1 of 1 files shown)


crates/egui/src/widget_text.rs
```diff
@@ -67,6 +67,27 @@ impl From<String> for RichText {
...
+    }
+}
+
+impl From<&mut Box<str>> for RichText {
+    #[inline]
+    fn from(text: &mut Box<str>) -> Self {
+        Self::new(text.clone())
+    }
+}
+
```
lucasmerlin, line 83: Is there a reason `&mut Box<str>` is implemented for RichText but not for WidgetText?

lucasmerlin (approved):
Nice!

crates/egui/src/widget_text.rs
dimtpap: I followed what already exists for `String`.

I believe Rust will [coerce](https://doc.rust-lang.org/reference/type-coercions.html) &mut anyway.

emilk (approved):

PR #5310, opened 2024-10-27
Respect `Style::override_font_id` in `RichText`

MStarha:
Currently `Style::override_font_id` does not affect widgets with their `TextStyle` like `Button`s, `ComboBox`es, `CheckBox`es, etc. This PR fixes this behavior such that all widgets using default configuration of `RichText` are affected by this setting.


emilk (approved):
Good catch

PR #5316, opened 2024-10-28
egui-winit: Remove implicit `accesskit_winit` feature

waywardmonkeys:
This is controlled by the `accesskit` feature and nothing needs the implicit `accesskit_winit` feature, so use `dep:` syntax to prevent it from being created.


emilk (approved):
oops!

PR #5319, opened 2024-10-29
Add option to initialize on existing wgpu setup

ArthurBrussee:
When mixing and matching eframe with other wgpu applications (https://github.com/tracel-ai/burn in my case), it can be helpful to use an existing wgpu setup to initialize eframe with. This PR changes the WpuConfiguration (in a non-backwards compat way :/), to either take some options how to create a wgpu setup, or an existing wgpu setup (consisting of an instance, adapter, device and queue). 

### DIFF
--- a/crates/egui-wgpu/src/lib.rs
+++ b/crates/egui-wgpu/src/lib.rs
@@ -347,7 +347,7 @@ impl Default for WgpuConfiguration {
 
             // By default, create a new wgpu setup. This will create a new instance, adapter, device and queue.
             // This will create an instance for the supported backends (which can be configured by
-            // `WGPU_BACKEND`), and will pick an adapter by iterateing adapters based on their power preference. The power
+            // `WGPU_BACKEND`), and will pick an adapter by iterating adapters based on their power preference. The power
             // preference can also be configured by `WGPU_POWER_PREF`.
             wgpu_setup: WgpuSetup::CreateNew {
                 // Add GL backend, primarily because WebGPU is not stable enough yet.
--- a/crates/egui-wgpu/src/lib.rs
+++ b/crates/egui-wgpu/src/lib.rs
@@ -24,6 +24,7 @@ pub use wgpu;
 mod renderer;
 
 pub use renderer::*;
+use wgpu::{Adapter, Device, Instance, Queue};
 
 /// Module for painting [`egui`](https://github.com/emilk/egui) with [`wgpu`] on [`winit`].
 #[cfg(feature = "winit")]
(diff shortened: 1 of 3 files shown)


crates/egui-wgpu/src/lib.rs
```diff
@@ -264,59 +320,60 @@ fn wgpu_config_impl_send_sync() {
...
-
             present_mode: wgpu::PresentMode::AutoVsync,
 
             desired_maximum_frame_latency: None,
 
-            power_preference: wgpu::util::power_preference_from_env()
-                .unwrap_or(wgpu::PowerPreference::HighPerformance),
+            // Use "standard" wgpu adapter selection by default, which iterates adapters
+            // based on their power preference. The power preferenc can be configured from
+            // env variables.
```
Wumpf, line 348: nit: it also picks a device depending on the device descriptors etc., undersells it a bit by only talking about the adapter selection

Wumpf (changes requested):
Nice! I really like having this hook in there.
Comments are entirely about naming & docs, approach & structure looks good to me!

crates/egui-wgpu/src/lib.rs
```diff
@@ -215,27 +235,63 @@ pub enum SurfaceErrorAction {
...
+                .finish_non_exhaustive(),
+        }
+    }
+}
+
 /// Configuration for using wgpu with eframe or the egui-wgpu winit feature.
 ///
 /// This can also be configured with the environment variables:
 /// * `WGPU_BACKEND`: `vulkan`, `dx11`, `dx12`, `metal`, `opengl`, `webgpu`
 /// * `WGPU_POWER_PREF`: `low`, `high` or `none`
```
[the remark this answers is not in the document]
ArthurBrussee: I've moved this closer to where they are actually used. Maybe they're a bit too hidden now, but idk!

ArthurBrussee:
Thanks for the speedy review :D Much better names now, definitely was a bit too quick as I adopted this from existing adapter selection code.

crates/egui-wgpu/src/lib.rs
```diff
@@ -343,10 +345,11 @@ impl Default for WgpuConfiguration {
 
             desired_maximum_frame_latency: None,
 
-            // Use "standard" wgpu adapter selection by default, which iterates adapters
-            // based on their power preference. The power preferenc can be configured from
-            // env variables.
-            wgpu_device_setup: WgpuDeviceSetup::Standard {
+            // By default, create a new wgpu setup. This will create a new instance, adapter, device and queue.
+            // This will create an instance for the supported backends (which can be configured by
+            // `WGPU_BACKEND`), and will pick an adapter by iterateing adapters based on their power preference. The power
```
Wumpf, line 350: typo
```suggestion
            // `WGPU_BACKEND`), and will pick an adapter by iterating adapters based on their power preference. The power
```

Wumpf (approved):
thanks again!

PR #5320, opened 2024-10-29
Move all existing .png images to git LGS

emilk:
* Sibling PR: https://github.com/emilk/egui/pull/5321


lucasmerlin:
I think we should also add anything in the media folder to lfs

PR #5322, opened 2024-10-29
Update MSRV from 1.76 to 1.77

xangelix:
* [X] I have followed the instructions in the PR template

I am preparing a separate PR that adds support for JXL with `jxl-oxide`, which is unlikely to be added to the `image` crate anytime soon (more context will be provided in that PR).

`jxl-oxide` makes use of the [`array::each_mut`](https://doc.rust-lang.org/stable/std/primitive.array.html#method.each_mut) API which was stabilized in 1.77, which is the motivation for this MSRV bump.

Rust 1.77 was officially released to stable on 21 March, 2024.

PR #5325, opened 2024-10-29
Add basic JXL image support

xangelix:

### Blocked by:
- [x] https://github.com/emilk/egui/pull/5322 `jxl-oxide` makes use of the [`array::each_mut`](https://doc.rust-lang.org/stable/std/primitive.array.html#method.each_mut) API 
- [x] https://github.com/emilk/egui/pull/5324
- [x] [`jxl-oxide`](https://github.com/tirr-c/jxl-oxide) v0.10 --> Git source then needs to be replaced

### Rationale
The `image` crate maintainer has made it clear that they do not intend to support JXL at this time, and likely won't be in the future either, see: https://github.com/image-rs/image/issues/1979

For this reason, the `jxl-oxide` crate has just recently added support for the `image` crate format, see: https://github.com/tirr-c/jxl-oxide/issues/162

JXL is a leading format (with yes, much less adoption than others) including various features that make it a compelling option. ~~`jxl-oxide` is an appropriate crate choice as it has a history of being very well maintained, and (at least a couple months ago?) is the likely pickup for Chromium and Firefox support in the next year or so. I've lost the source on this one, but remember either the Chromium or Firefox devs officially stating they were unwilling to merge the C library in, but suggested continuing this rust port.~~ The eventual crate being worked on to be implemented in Chromium and Firefox is https://github.com/libjxl/jxl-rs, however, `jxl-oxide` is still receiving active updates.

This is also especially interesting to implement as it allows high quality jxl support through WASM, irrespective of browser jxl support (which is poor at the moment).

~~This PR is a draft until `jxl-oxide`'s v0.10 release, which adds built-in `image` crate compatibility. The other linked PRs should also come first.~~ v0.10 was released https://github.com/tirr-c/jxl-oxide/releases/tag/0.10.0

### Todo/Further Considerations

- [ ] https://github.com/emilk/egui/issues/5341 ~Possibly should await any further image crate support detection changes. JXL magic byte detection could be added if that is the route we want to go down.
- [ ] `jxl-oxide` uses rayon optionally to speed up decode. Is it fine leaving this feature for the caller to specify in their own Cargo.toml? (similar to how the `image` crate features are currently set up)
- [ ] Should this PR be expanded to include more JXL features? Or (more likely) worry about basic support for now, and do larger refactoring of image loaders in a new PR? (Multiple image formats support progressive decode)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2201,16 +2201,18 @@ dependencies = [
 
 [[package]]
 name = "jxl-bitstream"
-version = "0.5.0-alpha.0"
-source = "git+https://github.com/tirr-c/jxl-oxide.git#229c223af49de96247a4e5548470ce2753912f3e"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3c84490d13b6303b007b63b6ffea9fa82ff5a4c33b0b2dc673e6f4556f11a2a1"
 dependencies = [
  "tracing",
 ]
 
 [[package]]
 name = "jxl-coding"
-version = "0.5.0-alpha.0"
-source = "git+https://github.com/tirr-c/jxl-oxide.git#229c223af49de96247a4e5548470ce2753912f3e"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0dd7ffdab0c48e989f23a8bd6113d88bd243ae45c7871e90cfdcb6997eacbfb2"
 dependencies = [
  "jxl-bitstream",
  "tracing",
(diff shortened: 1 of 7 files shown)


lucasmerlin:
Great work! Unfortunately I think jpeg xl is a bit too niche in order to be included in egui. You should instead release it as a 3rd party crate (e.g. egui_jxl) and add it to [the wiki](https://github.com/emilk/egui/wiki/3rd-party-egui-crates) so other people can find it!

xangelix:
Published to https://crates.io/crates/egui-jxl after all these years for anyone who finds this PR :P

Supports pure rust progressive decoding, based on the same rust crate used in major browsers: https://github.com/libjxl/jxl-rs

Issue #5293, opened 2024-10-21
Copy (CTRL+C and ui.output_mut) crash if HTTP


itsNiccoloSabatini:
## What happened
After the update for egui 0.29.1 i noticed that whenever i try to copy something in HTTP the WASM compiled version has a fatal crash at runtime. This does not happen when i use HTTPS.

## When does it happen 
**No matter if the OS is Windows or Linux it crashes in both.**
whenever i try to copy some text from the WASM page but in HTTP protocol, in 0.27.2 it just wouldn't have copied anything but since when the web-sys-unstable-apis aren't needed anymore, and something changed about how the copy to clipboard works, the copy in a HTTP page makes everything crash.

this happens in every time if i press CTRL+C or i use the code macro in HTTP:
```rust
ui.output_mut(|o| {
    debug!("Copying API key to clipboard: {:?}", key);
    o.copied_text.clone_from(&key.to_string());
});
```

## Image of the error in Chromium
![Schermata del 2024-10-21 16-02-02](https://github.com/user-attachments/assets/ca438af3-efdd-4316-904c-6590edd9703b)

## Image of the error in Firefox
![immagine](https://github.com/user-attachments/assets/c3382372-e384-40e6-aed8-39a086346e95)

lucasmerlin:
The clipboard api (which I guess is used now) only works in secure contexts ([explained here](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#security_considerations)). So when accessing an egui app over http on an address other than localhost this will fail. But of course egui shouldn't crash but return some kind of error result so the failure can be caught gracefully.

emilk:
I don't have this problem on https://www.egui.rs/ - do you? If not, I would suspect this is a wasm-bindgen mismatch on your end, or something similar

lucasmerlin:
> I don't have this problem on [egui.rs](https://www.egui.rs/) - do you? If not, I would suspect this is a wasm-bindgen mismatch on your end, or something similar

It should only happen when accessing an egui app via http (not https) on domains other than localhost, since the clipboard api is not available there. So this is a bit of an edge case, but we should still try to handle this without crashing

YgorSouza:
I was able to reproduce the issue locally with the demo by navigating to my machine's LAN IP ( `192.168...` as returned by e.g. `ifconfig` on Linux) instead of `127.0.0.1`. Then it considers that it is not a secure context and copying text panics. My PR changes it to just log an error and return instead, as it was before.

[...]

PR #5326, opened 2024-10-30
Prevent panic when copying text outside of a secure context

YgorSouza:
* Closes <https://github.com/emilk/egui/issues/5293>

### DIFF
--- a/crates/eframe/src/web/mod.rs
+++ b/crates/eframe/src/web/mod.rs
@@ -174,6 +174,13 @@ fn set_cursor_icon(cursor: egui::CursorIcon) -> Option<()> {
 /// Set the clipboard text.
 fn set_clipboard_text(s: &str) {
     if let Some(window) = web_sys::window() {
+        if !window.is_secure_context() {
+            log::error!(
+                "Clipboard is not available because we are not in a secure context. \
+                See https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts"
+            );
+            return;
+        }
         let promise = window.navigator().clipboard().write_text(s);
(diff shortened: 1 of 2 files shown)


crates/eframe/src/web/mod.rs
```diff
@@ -174,6 +174,13 @@ fn set_cursor_icon(cursor: egui::CursorIcon) -> Option<()> {
 /// Set the clipboard text.
 fn set_clipboard_text(s: &str) {
     if let Some(window) = web_sys::window() {
+        if !window.is_secure_context() {
+            log::error!(
```
rustbasic, line 178:
```suggestion
            log::debug!(
```

lucasmerlin: I think it's fine to log at least a warning here. Something is clearly wrong when this occurs and the developer should have a chance to see this error message so they know what's wrong.

lucasmerlin:
Thank you!

Maybe we should also update the [copy_text](https://docs.rs/egui/latest/egui/struct.Context.html#method.copy_text) docs to mention that this might not work for wasm builds outside of secure contexts / when the app is deployed without https?

emilk (approved):

YgorSouza:
> Thank you!
> 
> Maybe we should also update the [copy_text](https://docs.rs/egui/latest/egui/struct.Context.html#method.copy_text) docs to mention that this might not work for wasm builds outside of secure contexts / when the app is deployed without https?

I added it, although I don't know if this is the right place for that, since it doesn't really have anything to do with egui, it's just how wasm works. But I guess at least it is hard to miss.

PR #5328, opened 2024-10-30
Fix documentation for file drag and drop

valadaptive:
The `RawInput::dropped_files` documentation referred to `NativeOptions`, but the `drag_and_drop` option was moved to `ViewportBuilder` and is now enabled by default.

Note that drag-and-drop is not *explicitly* enabled by default here, but rather is enabled by default in winit. We only override winit's default if the API consumer specifies it. It might be a good idea in general to provide explicit defaults for everything in `ViewportBuilder` so we're not relying on winit's behavior, but that's best left for another PR.

* Closes N/A

### DIFF
--- a/crates/egui/src/data/input.rs
+++ b/crates/egui/src/data/input.rs
@@ -68,8 +68,8 @@ pub struct RawInput {
 
     /// Dragged files dropped into egui.
     ///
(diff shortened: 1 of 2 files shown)


emilk (approved):
Thanks!

PR #5330, opened 2024-10-30
wgpu: Bump to wgpu 23.0.0 and wasm-bindgen to 0.2.95

torokati44:


ArthurBrussee:
Hah race condition, did the same. For the cargo deny errors, you can get some inspiration of my branch

https://github.com/ArthurBrussee/egui/tree/wgpu-upgrade

crates/egui_demo_app/src/apps/custom3d_wgpu.rs
```diff
@@ -47,13 +47,13 @@ impl Custom3d {
             layout: Some(&pipeline_layout),
             vertex: wgpu::VertexState {
                 module: &shader,
-                entry_point: "vs_main",
+                entry_point: Some("vs_main"),
```
Wumpf, line 50: Let's use that new feature and just pass in `None` because why not :)
(works here but not inside of egui-wgpu renderer.rs)

Wumpf (approved):
I can have a look at the dependency bonanza later

There's also changes I want to make to web backend I want to make - there's now utilities in wgpu that make WebGPU detection _a lot_ saner and faster

torokati44:
Whoo, thanks a lot for sorting out the dependency situation, and finishing this up so quickly!

PR #5336, opened 2024-11-01
Exclude icon.png from lfs

### LINKED ISSUE #5331: Building from git fails after #5320
#5320 moved all pngs to lfs, now when using eframe as a git dependency, the egui icon doesn't load because the repo isn't cloned with lfs. So we should make a lfs exception for the [egui icon](https://github.com/emilk/egui/blob/5bfff316c9818b3c140d02bb6cdc488556d46ab7/crates/eframe/data/icon.png#L2).

```bash
cat icon.png 
version https://git-lfs.github.com/spec/v1
oid sha256:31bbc7de655ccdf16efb6adb9959b5c6b9318befa0cdae88499b072549a55e3f
size 17166
```

lucasmerlin:
* Closes #5331 


emilk (approved):
Oops! Thanks 😅

PR #5339, opened 2024-11-01
Fix some typos

lucasmerlin:
The spell check pipeline in #5313 suddenly failed, this fixes these typos and some more found via my IDEs spell checker tool

### DIFF
--- a/crates/egui/src/util/cache.rs
+++ b/crates/egui/src/util/cache.rs
@@ -39,7 +39,7 @@ impl<Value, Computer> FrameCache<Value, Computer> {
     }
 
     /// Must be called once per frame to clear the cache.
-    pub fn evice_cache(&mut self) {
+    pub fn evict_cache(&mut self) {
         let current_generation = self.generation;
         self.cache.retain(|_key, cached| {
             cached.0 == current_generation // only keep those that were used this frame
@@ -89,7 +89,7 @@ impl<Value: 'static + Send + Sync, Computer: 'static + Send + Sync> CacheTrait
     for FrameCache<Value, Computer>
 {
     fn update(&mut self) {
-        self.evice_cache();
+        self.evict_cache();
     }
 
     fn len(&self) -> usize {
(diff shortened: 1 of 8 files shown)


crates/egui/src/util/cache.rs
```diff
@@ -39,7 +39,7 @@ impl<Value, Computer> FrameCache<Value, Computer> {
     }
 
     /// Must be called once per frame to clear the cache.
-    pub fn evice_cache(&mut self) {
+    pub fn evict_cache(&mut self) {
         let current_generation = self.generation;
         self.cache.retain(|_key, cached| {
             cached.0 == current_generation // only keep those that were used this frame
```
lucasmerlin, line 45: This is a breaking change but as far as I can tell this is only called in CacheTrait::update and no external code calls this ([confirmed via github code search](https://github.com/search?q=%2F%5CWevice_cache%2F&type=code))

crates/egui/src/widgets/checkbox.rs
```diff
@@ -38,7 +38,7 @@ impl<'a> Checkbox<'a> {
 
     /// Display an indeterminate state (neither checked nor unchecked)
     ///
-    /// This only affects the checkbox's appearance. It will still toggle its boolean value when
+    /// This only affects the checkboxes appearance. It will still toggle its boolean value when
```
YgorSouza, line 41: This one isn't a typo.

crates/epaint/src/tessellator.rs
```diff
@@ -2029,8 +2029,8 @@ impl Tessellator {
         use rayon::prelude::*;
 
         // We only parallelize large/slow stuff, because each tessellation job
-        // will allocate a new Mesh, and so it creates a lot of extra memory framentation
-        // and callocations that is only worth it for large shapes.
+        // will allocate a new Mesh, and so it creates a lot of extra memory fragmentation
+        // and call locations that is only worth it for large shapes.
```
YgorSouza, line 2033: I think this is supposed to be "allocations".

crates/egui/src/widgets/checkbox.rs
lucasmerlin: Huh, the grammar check marked it as wrong but I guess the original version was correct. I'll revert it

crates/epaint/src/tessellator.rs
lucasmerlin: Ah, that makes much more sense

Issue #5341, opened 2024-11-02
Some image urls don't load after #5324


lucasmerlin:
**Describe the bug**
#5324 introduced stricter image uri and mime detection. While this makes sense for e.g. file urls where the extensions are likely correct, I don't think we should check extensions for http urls, as they are not reliable at all. 

Some imaginary cases I can think of where the new checks would break:
- Query params, e.g. to change the image size
  - e.g. http://example.com/myimage.png?width=200
  - also common, presigned s3 urls: https://my-bucket.s3-provider.com/img.webp?X-Amz-Algorithm=AWS4-HMAC-SHA256[...]
- Hash fragments, often used for cache busting
  - e.g. http://example.com/myimage.png#1234 
- Urls that look like they have an an extension but it's not an extension at all
  - e.g. http://example.com/api/user_images?username=lucas.merlin (.merlin would be detected as an extension with our logic)
- Urls that have a wrong extension
  - e.g. http://example.com/render_html_to_image/my_document.html (which would return my_document rendered as an image)

Some of these cases might be bad api design. If you have full control over the backend that might not be a problem but if you e.g. build a chat app where people can send an image url and the app tries to load the image preview from the external url, you would want to support as many urls as possible. 

We could try stripping the query params and fragments when checking for the extensions, but I don't think the extra complexity is worth it, considering that this is optimizing the failure case.

The mime check is probably more reliable, but might not always be true. When uploading an image to s3 and you forget to set the mime type, it will have `application/octet-stream`. I believe, when showing such an image in a `<img>` tag, the browser will still show it successfully (but I'm not 100% sure). 
Since we already have the image downloaded when checking the mime type, we might as well just get the type via `image::guess_format`.

Also, we aren't even using the mime type when loading the image, instead relying on the image crate to guess the format, meaning currently `image` has to guess the type twice (once when we call 'image::guess_type' and once more when we actually decode the image).

I think we should at least make this behavior configurable, and turn it off by default. 

**To Reproduce**
1. Run `cargo run -p egui_demo_app --features image_viewer` from the egui repo on the current master
2. Open the image viewer app
3. Try opening e.g. https://images.pexels.com/photos/45201/kitty-cat-kitten-pet-45201.jpeg?auto=compress&cs=tinysrgb&w=1260&h=750&dpr=2  (copied from here: https://www.pexels.com/photo/white-and-grey-kitten-on-brown-and-black-leopard-print-textile-45201/)
4. Notice how it won't load

**Expected behavior**
[...]

PR #5344, opened 2024-11-03
Fix: How to check the image extension from Uri.

rustbasic:
Fix: How to check the image extension from Uri.

* Related #5324
* Related #5341

It addresses most of what was discussed in #5341, but I'm not sure if it addresses everything.

### DIFF
--- a/crates/egui_extras/src/loaders/image_loader.rs
+++ b/crates/egui_extras/src/loaders/image_loader.rs
@@ -5,7 +5,7 @@ use egui::{
     ColorImage,
 };
 use image::ImageFormat;
-use std::{mem::size_of, path::Path, sync::Arc};
+use std::{mem::size_of, sync::Arc};
 
 type Entry = Result<Arc<ColorImage>, String>;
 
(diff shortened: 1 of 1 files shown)


xangelix:
Thank you for the PR and apologies again that my previous changes have caused issues for you, though I have concerns that this PR would still fail in some web URI scenarios.

Also, if this were to be the direction we go, you'd almost certainly want to do something like this to avoid having to keep extensions in sync with the image crate manually :)

```rs
    ImageFormat::all()
        .filter(ImageFormat::reading_enabled)
        .flat_map(ImageFormat::extensions_str)
        .any(|format_ext| uri.contains(*format_ext))
```

rustbasic:
> Thank you for the PR to resolve an issue I introduced and apologies that my previous changes have caused issues for you, though I have concerns that this PR would still fail in some web URI scenarios.
> 
> Also, if this were to be the direction we go, you'd almost certainly want to do something like this to avoid having to keep extensions in sync with the image crate manually :)
> 
> ```rust
>     ImageFormat::all()
>         .filter(ImageFormat::reading_enabled)
>         .flat_map(ImageFormat::extensions_str)
>         .any(|format_ext| uri.contains(*format_ext))
> ```

I have modified this PR as you said.
If you want to make a PR, I will close this PR once you submit it.

crates/egui_extras/src/loaders/image_loader.rs
```diff
@@ -19,16 +19,11 @@ impl ImageCrateLoader {
 }
 
 fn is_supported_uri(uri: &str) -> bool {
-    let Some(ext) = Path::new(uri).extension().and_then(|ext| ext.to_str()) else {
-        // `true` because if there's no extension, assume that we support it
-        return true;
-    };
-
     // Uses only the enabled image crate features
```
rustbasic, line 22: In most cases it's fine, but I'm not sure if it's a good idea to remove this part.

emilk:
Closed by https://github.com/emilk/egui/pull/5431

PR #5253, opened 2024-10-12
Fix: `ctx.set_fonts()` Doesn't Apply Fonts with Same Name

rustbasic:
If a font file named `my_font` is set using `ctx.set_fonts()` and then another font file with the same name is set again using `ctx.set_fonts()`, it won't be applied.

1. Additionally, it's recommended to apply fonts immediately using `ctx.set_fonts()`.
2. This will also allow you to reset the fonts and texture.

* Related #5233
* Closes #5252

Issue #5233, opened 2024-10-08
Automatically load system fonts


emilk:
Currently egui only supports glyphs available in fonts that has been explicitly installed with `ctx.set_fonts`.

It would be great if there was a way for `egui` to tell it's integration "The user wanted to show a Chinese glyph, but no current font has support for that" and then have `eframe` respond by automatically finding such a font on the system, e.g. by downloading it.

haixuanTao:
FYI, some country like China has a GFW that can limit access to western website like google.font that can make dynamically loaded asset fail to load. 🥹

GamingLiamStudios:
I've encountered an issue trying to implement this in my own application. Much like in #7325, I've used `fontconfig` /`DirectWrite` to fetch the System fonts and load them in at program start. However since I may encounter many different languages in my app, I end up loading fonts until every Unicode Codepoint possible is covered (a use-case explicitly allowed by `fontconfig`). This leads to a large amount of fonts to be loaded, and hence a large amount of memory to be utilized (e.g. 200MB on Linux, 800MB on Windows). As they also mentioned, this is linked to how egui expects all fonts to be loaded in memory (via the `FontDefinitions` struct).

Their offered solution however isn't generic enough imho. I think a better solution would be if instead there was a `Context::set_font_loader` function. This would be set by either the implementation (eframe) or the user, and would function much like `Context::set_immediate_viewport_renderer`. A `Fn`/`FnMut` callback would be provided, and called whenever a glyph is not found within a `FontFamily`. 
This would easily allow the implementation to fetch/load another font that could contain this glyph when required, lowering the memory usage by the app and simplifying the process for implementing this Issue (#5233) in `eframe`. Unfortunately this can cause issues to do with blocking the render, so would likely require some care to handle properly.

I'm not sure if some of the other work being done with #5784 would make this redundant, but I thought it would be good to just throw the idea out there, as I do plan to (attempt to) implement this myself for my own project, and may make a PR in the future if that turns out well.

PR #5228, opened 2024-10-07
Add font

frederik-uni:
make it easier to add fonts. 

For example if I want to add a custom FontFamily or if the user wants to add a Chinese fallback 
* [x] I have followed the instructions in the PR template

PR #5345, opened 2024-11-03
Fix: ctx.set_fonts() Doesn't Apply Fonts with Same Name

### LINKED ISSUE #5252: `ctx.set_fonts()` Doesn't Apply Fonts with Same Name
If a font file named `my_font` is set using `ctx.set_fonts()` and then another font file with the same name is set again using `ctx.set_fonts()`, it won't be applied.

Additionally, it's recommended to apply fonts immediately using `ctx.set_fonts()`.

rustbasic:
Fix: ctx.set_fonts() Doesn't Apply Fonts with Same Name

If a font file named `my_font` is set using `ctx.set_fonts()` and then another font file with the same name is set again using `ctx.set_fonts()`, it won't be applied.

1. Additionally, it's recommended to apply fonts immediately using `ctx.set_fonts()`.
2. This will also allow to `reset` the `fonts` and `texture`.

* Related #5233
* Closes #5252
* Related #5253
* Related #5228
* Related #5276

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -567,7 +567,7 @@ impl ContextImpl {
     }
 
     /// Load fonts unless already loaded.
-    fn update_fonts_mut(&mut self) {
+    pub(crate) fn update_fonts_mut(&mut self) {
         crate::profile_function!();
 
         let input = &self.viewport().input;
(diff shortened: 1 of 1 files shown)


lalala-233:
I agree with the suggestion to remove the comparison check in `set_fonts`, as it seems unnecessary in most cases and could improve performance. 

However, I'm not entirely convinced about the need for real-time updates (applying fonts immediately). While real-time updates might be useful in some scenarios, the current behavior of applying changes in the next frame is generally sufficient.

emilk:
Closing: the font code was rewritten in #8504 and #8512, so this diff no longer applies. If #5252 still reproduces on current main, please open a fresh PR. Thanks for the PR, and sorry for the slow response

PR #5347, opened 2024-11-03
Fix: Ignoring Characters with ALT Key

### LINKED ISSUE #5338: consume_key and consume_shortcut with input field
**Describe the bug**
When a keyboard shortcut is consumed, the key will inserted into an active edit field anyway.

In the below  example, there is an edit field to type and independent of this, the keyboard shortcut ALT-B will be intercepted. In the example, the keyboard shortcut is processed before the edit field is displayed. Pressing ALT-B will trigger the action but will also insert the letter 'b' into the edit field.

**To Reproduce**

```rust
use eframe::egui;

fn main() -> eframe::Result {
    eframe::run_native(
        "xyz",
    )
}

#[derive(Default)]
pub struct MyApp {
    content: String,
    show_label: bool,
}

impl eframe::App for MyApp {
            if ctx.input_mut(|i| i.consume_key(egui::Modifiers::ALT, egui::Key::B)) {
                self.show_label = true;
            };
            ui.label("Type a few chars and then press ALT-B:");
            ui.text_edit_singleline(&mut self.content);
            if self.show_label {
                ui.label("You pressed ALT-B. Note that the char 'b' is inserted into the edit field, even though consume_key has been called.");
            }
        });
    }
}
```

**Expected behavior**

I would expect that if the key is consumed, i.e. that it will not be interpreted by the text edit. This is also what the function description indicates: "Check for a key press. If found, `true` is returned and the key pressed is consumed, so that this will only return `true` once.".

**Screenshots**

Bildschirmfoto_2024-11-01_16-57-03

Bildschirmfoto_2024-11-01_16-57-32

 - OS: ArchLinux

rustbasic:
Fix: Ignoring Characters with ALT Key

When you press the ALT key and a character at the same time, the character is ignored.

* Closes #5338

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -799,6 +799,7 @@ impl State {
                 // Also make sure the key is pressed (not released). On Linux, text might
                 // contain some data even when the key is released.
                 let is_cmd = self.egui_input.modifiers.ctrl
+                    || self.egui_input.modifiers.alt
                     || self.egui_input.modifiers.command
                     || self.egui_input.modifiers.mac_cmd;
                 if pressed && !is_cmd {


emilk:
On Mac (and probably other OSes), `Alt/Option` is used to input special characters that will now be impossible to write

PR #5350, opened 2024-11-04
chore: update arboard

Gae24:
While it was updated on the lock file, Cargo.toml was still on 3.3.


emilk:
that's fine - we don't need to update the Cargo.toml unless we depend on a newer version

PR #5362, opened 2024-11-10
Create gray -> grey doc aliases

MeGaGiGaGon:
As someone who uses "grey" instead of "gray", it is annoying that my autocomplete can never find any of the "gray" color related things, so this adds doc aliases for that.


lucasmerlin (approved):
Nice!

PR #5364, opened 2024-11-11
Fix vector normalization in epaint/tessellator/add_line_loop.

PangolinGiLA:
In epaint/tessellator/add_line_loop normal vectors of added points were not normalized correctly, they were divided by length squared, what caused their length to explode for some points close together.

Before:
before

After:
after

### DIFF
--- a/crates/epaint/src/tessellator.rs
+++ b/crates/epaint/src/tessellator.rs
@@ -470,7 +470,7 @@ impl Path {
                 self.add_point(points[i], n1c / n1c.length_sq());
             } else {
                 // miter join
-                self.add_point(points[i], normal / length_sq);
+                self.add_point(points[i], normal.normalized());
             }
 
             n0 = n1;


lucasmerlin:
Seems like this causes a couple small differences in the snapshot tests, the biggest one is visible in the rendering test: 
(the thinner one is the new, updated snapshot)

https://github.com/user-attachments/assets/c1a91aee-62ba-43bd-ac2e-6c37cd8b3332

And some minor differences in the bezier test: 
(the one where the top green line is more faint is the new one)
Also, notice how the triangle to close the window seems slightly worse antialiased.

https://github.com/user-attachments/assets/dbdd314a-3029-40e8-bb5b-0a6638b7dd88


Not sure which versions are more correct though, just wanted to share the differences I noted. 

Here is a link to the snapshot test results: https://github.com/emilk/egui/actions/runs/11778645510/artifacts/2171312779

@PangolinGiLA You'll need to run the tests with `UPDATE_SNAPSHOTS=true` to update the snapshots before we can merge this

crates/epaint/src/tessellator.rs
```diff
@@ -470,7 +470,7 @@ impl Path {
                 self.add_point(points[i], n1c / n1c.length_sq());
             } else {
                 // miter join
-                self.add_point(points[i], normal / length_sq);
+                self.add_point(points[i], normal.normalized());
```
emilk, line 473: I believe the fix is wrong. Dividing with length-squared is what gets the correct widths for a miter-join. If that sometimes causes numerical problems we should either fix those narrow cases (if angle < …) or switch from miter-joins

emilk (changes requested):

crates/epaint/src/tessellator.rs
PangolinGiLA: Oh sorry, I didn't realize that, I just changed it thinking normal vectors *always* have unit length, and this fixed my problem with rendering.

The problem appeared when I tried to render points distributed this way (this is the bottom part of the image above): 
points

PangolinGiLA: Oh I just read the issue linked in the code. I guess i should have done a little bit more research before opening the PR.

emilk:
I would love to have a proper solution to this problem, but this PR isn't it

Issue #5359, opened 2024-11-08
Disabled widgets prevent focus from going to next widget


lucasmerlin:
**Describe the bug**

Disabled widgets currently make it impossible to focus all widgets by pressing tab. 

In this video I keep hitting tab:

https://github.com/user-attachments/assets/3af72e3e-cfd5-4047-ac9c-a0f49d386841

Notice how `Button 3` never receives focus.

**To Reproduce**
Minimal example:
```rust
    eframe::run_simple_native("My egui App", options, move |ctx, _frame| {
        egui::CentralPanel::default().show(ctx, |ui| {
            ui.button("Button 1");
            ui.button("Button 2");

            ui.add_enabled(false, Button::new("Button Disabled"));

            ui.button("Button 3");
        });
    })
```

**Expected behavior**
It should be possible to focus the widget after the disabled widget

PR #5370, opened 2024-11-14
Fix disabled widgets "eating" focus

lucasmerlin:
- fixes https://github.com/emilk/egui/issues/5359

For the test I added a `Harness::press_key` function. We should eventually add these to kittest, probably via a trait one can implement for the `Harness` but for now this should do.

### DIFF
--- a/crates/egui/tests/regression_tests.rs
+++ b/crates/egui/tests/regression_tests.rs
@@ -1,0 +1,30 @@
+use egui::Button;
+use egui_kittest::kittest::Queryable;
+use egui_kittest::Harness;
+
+#[test]
+pub fn focus_should_skip_over_disabled_buttons() {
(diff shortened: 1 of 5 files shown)


crates/egui/tests/regression_tests.rs
```diff
@@ -0,0 +1,30 @@
...
+    harness.run();
+
+    let button_3 = harness.get_by_name("Button 3");
+    assert!(button_3.is_focused());
+
+    harness.press_key(egui::Key::Tab);
+    harness.run();
+
+    let button_1 = harness.get_by_name("Button 1");
+    assert!(button_1.is_focused());
```
emilk, line 29: I love it

emilk (approved):

PR #5372, opened 2024-11-14
Update accesskit to 0.17

lucasmerlin:
Updates accesskit and kittest. 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,9 +20,9 @@ checksum = "c71b1793ee61086797f5c80b6efa2b8ffa6d5dd703f118545808a7f2e27f7046"
 
 [[package]]
 name = "accesskit"
-version = "0.16.3"
+version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "99b76d84ee70e30a4a7e39ab9018e2b17a6a09e31084176cc7c0b2dec036ba45"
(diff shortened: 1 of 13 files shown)


lucasmerlin:
@mwcampbell the update was straightforward, but maybe you can review this in case I missed anything? 

Especially around the `set_default_action_verb`, I just removed those calls but am not 100% sure if I need to do anything else there.

mwcampbell:
I see you dropped two calls to `set_default_action_verb`:

* `builder.set_default_action_verb(accesskit::DefaultActionVerb::Focus);`: This one is no longer necessary.
* `builder.set_default_action_verb(accesskit::DefaultActionVerb::Click);`: The if statement surrounding this needs to be re-added, but replace this statement with: `builder.add_action(accesskit::Action::Click);`.

The other problem is with the part where you use `builder.set_label`. If the role of the new node is `Role::Label`, you need to use `builder.set_value` instead.

lucasmerlin:
@mwcampbell thank you for the review, I've updated the PR accordingly!

mwcampbell (approved):
LGTM now. I also did a quick test with a screen reader, and all looks right.

mwcampbell:
Just checking: are you waiting on something else from me before merging? Thanks.

bircni (approved):
lgtm 👍

Issue #5379, opened 2024-11-17
Incorrect pointer position during changing zoom factor


timstr:
Thank you again for this wonderful library!

**Describe the bug**
Using the Winit backend, when modifying egui's zoom factor while querying the pointer position, egui simply uses the computed position from the last time the cursor was moved is used. However, this computed position bakes in the past zoom factor, meaning that the current position is scaled incorrectly relative to the top left corner of the window, until the cursor is moved again. Given that the screen rect is always up-to-date with the current zoom factor, this can lead to surprises when doing math that combines both the cursor position and the screen rect.

**To Reproduce**
Here's a quick demo that allows the egui zoom factor to be changed by scrolling and which displays the pointer interact position using a red circle. This is adapted from the eframe native demo at https://docs.rs/eframe/0.29.1/eframe/index.html#usage-native

```rust
impl eframe::App for MyEguiApp {
    fn update(&mut self, ctx: &egui::Context, _: &mut eframe::Frame) {
        egui::CentralPanel::default().show(ctx, |ui| {
            ui.heading("Hello World!");

            // START OF DEMO ====================================
            let (scroll, pointer_pos) =
                ui.input(|i| (i.smooth_scroll_delta, i.pointer.interact_pos()));

            ui.ctx().set_zoom_factor(
                (ui.ctx().zoom_factor() * (scroll.y * 0.01).exp()).clamp(0.25, 4.0),
            );

            if let Some(pos) = pointer_pos {
                ui.painter()
                    .circle_stroke(pos, 10.0, egui::Stroke::new(4.0, egui::Color32::RED));
            }
            // END OF DEMO ====================================
        });
    }
}
```

Here is a screen recording showing how egui's stored pointer position appears to move with the scale and differs from the actual pointer position:

[Screencast from 2024-11-16 04:09:08 PM.webm](https://github.com/user-attachments/assets/da560e73-94f3-42a4-8a64-6cb09ac3eb33)

**Expected behavior**

I would expect the pointer position given by egui to correspond to the visible location of the OS cursor at all times. Here I've written a quick patch that achieves this by deferring the zoom factor calculations to be per-frame, which I'll put up for review shortly.

[Screencast from 2024-11-16 04:13:54 PM.webm](https://github.com/user-attachments/assets/86445c41-b4c1-462c-8b35-e962e69e2264)

**Why this matters to me at all**

[...]

PR #5380, opened 2024-11-17
Always compute pointer position with current zoom

timstr:
This addresses out-of-date pointer positions during changing zoom factors by performing the same zoom-based scaling per-frame when the current zoom factor is known, rather than baking in the zoom factor during the most recent pointer move.

Before:

[Screencast from 2024-11-16 04:09:08 PM.webm](https://github.com/user-attachments/assets/36976437-d79e-4d71-a9e9-ca7aca9d88a7)

After:

[Screencast from 2024-11-16 04:13:54 PM.webm](https://github.com/user-attachments/assets/aecfa5b1-7242-49ac-85c9-42e35002b6b3)

See https://github.com/emilk/egui/issues/5379 for discussion and more details

* Closes https://github.com/emilk/egui/issues/5379

### DIFF
--- a/crates/egui/src/input_state/mod.rs
+++ b/crates/egui/src/input_state/mod.rs
@@ -800,6 +800,9 @@ pub struct PointerState {
     /// Latest known time
     time: f64,
 
+    /// Latest known zoom factor
+    zoom_factor: f32,
+
     // Consider a finger tapping a touch screen.
     // What position should we report?
     // The location of the touch, or `None`, because the finger is gone?
@@ -876,7 +879,8 @@ pub struct PointerState {
 impl Default for PointerState {
     fn default() -> Self {
         Self {
-            time: -f64::INFINITY,
+            time: f64::NEG_INFINITY,
+            zoom_factor: f32::NEG_INFINITY,
             latest_pos: None,
             interact_pos: None,
             delta: Vec2::ZERO,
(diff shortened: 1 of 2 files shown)


crates/egui/src/input_state/mod.rs
```diff
@@ -1097,13 +1102,13 @@ impl PointerState {
...
     pub fn latest_pos(&self) -> Option<Pos2> {
-        self.latest_pos
+        self.latest_pos.map(|p| p / self.zoom_factor)
     }
 
     /// If it is a good idea to show a tooltip, where is pointer?
     #[inline(always)]
     pub fn hover_pos(&self) -> Option<Pos2> {
-        self.latest_pos
+        self.latest_pos.map(|p| p / self.zoom_factor)
```
emilk, line 1111: This all feels like the wrong solution to me.

In egui, everything is in the same global coordinate space called _ui points_.

If `self.latest_post` needs dividing with `zoom_factor`, that's a bug in how `latest_pos` is set

emilk (changes requested):

crates/egui/src/input_state/mod.rs
timstr: Thanks for your eyes. Indeed my actual problem was already solved, I just hadn't looked around enough. Closing this MR and the linked issue now.

PR #5391, opened 2024-11-20
forward x11 and wayland features to glutin

e00E:
eframe has features for selecting between x11 and wayland. eframe does not forward the features to glutin. This makes glutin always compile with both backends enabled. This change forwards the feature. This allows users of egui to compile less dependencies when they only need one of x11, wayland.

To understand this change, read the glutin Cargo.toml [1] and the glutin build.rs [2]. You always have to enable glutin's glx feature with the x11 feature. The other default features (egl, wgl) stay enabled. This is intentional so that everything continues to work as before. We could further minimize when egl and wgl are enabled, but that is not part of this change. There is little reason to do so because those feature already only add dependencies when you compile glutin for the right platform (for example wgl on windows).

[1] https://github.com/rust-windowing/glutin/blob/v0.32.1/glutin/Cargo.toml
[2] https://github.com/rust-windowing/glutin/blob/v0.32.1/glutin/build.rs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,8 +78,8 @@ criterion = { version = "0.5.1", default-features = false }
 dify = { version = "0.7", default-features = false }
 document-features = "0.2.10"
 glow = "0.14"
-glutin = "0.32.0"
-glutin-winit = "0.5.0"
+glutin = { version = "0.32.0", default-features = false }
+glutin-winit = { version = "0.5.0", default-features = false }
 home = "0.5.9"
 image = { version = "0.25", default-features = false }
 kittest = { git = "https://github.com/rerun-io/kittest", version = "0.1", branch = "main" }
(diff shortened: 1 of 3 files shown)


emilk (approved):

emilk:
I suspect this PR is the cause of a new problem:

https://github.com/emilk/eframe_template/actions/runs/12357896151/job/34487194281

emilk:
Confirmed that reverting this fixes eframe_template:
* https://github.com/emilk/eframe_template/pull/170

emilk:
EDIT: the fix is to add "wayland" as a feature AND update to Rust 1.81

fluxxcode:
> EDIT: the fix is to add "wayland" as a feature AND update to Rust 1.81

In the changelog and release the change is listed as a normal change. But this should be a breaking change because it is now required to activate this feature. Otherwise there is a lot of confusion if Linux is no longer supported after an eframe update with default features disabled.

emilk:
@fluxxcode I've updated https://github.com/emilk/egui/releases/tag/0.30.0 with a "BREAKING: " prefix

PR #5398, opened 2024-11-23
Fix all links to the demo website

tguichaoua:

The demo website use the hash part of the url to open a specific demo. The website expect capitalized hash (`#Demo`) while all the links in the docs use lower case hash (`#demo`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ an SDK for visualizing streams of multimodal data.
 
 ---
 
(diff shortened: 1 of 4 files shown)


emilk:
Thanks, but I think a nicer fix would be to make the hash-matching case insensitive

tguichaoua:
Close in favor of #5446

Issue #2041, opened 2022-09-12
Caps Lock Key/Modifier


parker-codes:
**Is your feature request related to a problem? Please describe.**

Provide more visual context to a user as they are typing when Caps Lock is on. One example might be password inputs where a warning message could prevent confusion from accidental caps.

**Describe the solution you'd like**

I'd like to either be able to check the modifier state or listen to an on/off event. Then the UI could style the hint however I want.

njust:
You can already check the modifiers via the input state.
https://docs.rs/egui/0.19.0/egui/struct.InputState.html
https://docs.rs/egui/0.19.0/egui/struct.Modifiers.html

parker-codes:
Thank you, @njust. What I'm specifically looking for though is the Caps Lock modifier, which is not yet available.

emilk:
Closed in favor of https://github.com/emilk/egui/issues/3653

PR #5400, opened 2024-11-24
Add some Keys

### LINKED ISSUE #3636: Need to Distinguish Left and Right Keys for "Ctrl, Alt, Shift" on the Keyboard
Subject: Need to Distinguish Left and Right Keys for Ctrl, Alt, Shift on the Keyboard

Dear Emilk,

It seems that while it's possible to detect the "left Ctrl" and "left Alt" on the keyboard,
there is no way to determine whether the "right Ctrl" or "right Alt" has been pressed.
In the case of "Shift", even if the right Shift is pressed, it is recognized as Shift.

It would be better if there is a way to differentiate between left and right for "Ctrl, Alt, and Shift".

This is particularly essential for regions where the "right Alt" key or "right Ctrl" key is commonly used
as a language switch key, not just in English-speaking areas.

Thanks, Emilk.

Library: egui (all versions)
OS: windows 10

rustbasic:
Add some Keys :

    CapsLock,
    NumLock,
    ScrollLock,

    Alt,
    AltRight,
    Control,
    ControlRight,
    Shift,
    ShiftRight,

* Related #2041
* Closes #3636
* Related #3653


JRoper18:
Hey, I'd love this feature. Any hope of getting it merged in?

crates/egui/src/data/key.rs
```diff
@@ -25,6 +25,17 @@ pub enum Key {
...
+    CapsLock,
+    NumLock,
+    ScrollLock,
+
+    Alt,
+    AltRight,
+    Control,
+    ControlRight,
+    Shift,
+    ShiftRight,
```
emilk, line 37: A) it's weird that there is both `Alt` and `AltRight`, but not `AltLeft`
B) on the PR preview (https://egui-pr-preview.github.io/pr/5400-patch150/) the `Right` versions doesn't work at all.

I think we either should just have a single `Alt`, OR we should have `AltLeft` and `AltRight`, and make sure it works everywhere

emilk:
This doesn't work at all in the web demo at <https://egui-pr-preview.github.io/pr/5400-patch150/>:
image

rustbasic:
While it might not work well on the web because the `browser` takes `focus`, this isn't the issue with this commit.
We'll need an additional commit for the `modifiers`.

Ripper53:
I would really benefit from this feature. I am currently trying to create a key remapping application, and need access to the precise physical keys that are pressed, ie. `LeftShift`, `RightShift`, `LeftControl`, `RightControl`, `LeftAlt`, `RightAlt`. Right now, I just map any shift key onto the left shift. I do not need browser support for my application, only Windows as of now.

An update on this works progression and if it is still being worked on would be appreciated.

rustbasic:
@Ripper53

This PR was drafted by emilk and will not be progressed further.
Please use this PR as a reference and modify egui directly for your use.
It would be a good idea for you to refer to this PR and try creating a new, improved PR yourself.

emilk:
Closing: left/right modifier keys landed in #8127. The lock keys are still tracked in #3636. Thanks for the PR!

PR #5403, opened 2024-11-25
Add localization example

fewensa:
* Closes <https://github.com/emilk/egui/issues/136>

Related:
- https://github.com/emilk/egui/issues/136
- https://github.com/emilk/egui/discussions/1751
- https://github.com/fewensa/egui-i18n/pull/1

This just an example to show my idea, From the previous communication records, it seems that there is no clear support plan for multiple languages. If you think this solution is feasible, please let me know, I will migrate all the code bases of [egui-i18n](https://github.com/fewensa/egui-i18n) to egui and provide a deeper integration. @emilk 

more examples:

- [classic](https://github.com/fewensa/egui-i18n/tree/main/examples/classic)
- [fluent](https://github.com/fewensa/egui-i18n/tree/main/examples/fluent)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1239,6 +1239,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "egui-i18n"
+version = "0.1.0"
+source = "git+https://github.com/fewensa/egui-i18n#a1afe3cc6244a7edbe6135fcde0958c0ee6d9d56"
+dependencies = [
+ "log",
+ "once_cell",
+]
+
+[[package]]
 name = "egui-wgpu"
 version = "0.29.1"
 dependencies = [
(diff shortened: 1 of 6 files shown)


lucasmerlin:
egui_i18n looks great! But unfortunately including this example would introduce a circular dependency, where when we make a breaking change to egui, we'd have to first update egui-i18n to fix the breakage and then update the example with the new version before we could have a successful CI. 

Including egui-i18n in egui would of course fix it, but I think it works great as a third party crate, and it would be a lot of extra code to maintain. Instead you should add it to https://github.com/emilk/egui/wiki/3rd-party-egui-crates so people can find it! 

I also suggest adding a bit more to the readme, right now it looks very empty, and also add a license, so people can legally use it.

PR #5392, opened 2024-11-20
Document justification for FallbackEgl

e00E:
The previous link does not explain why we chose FallbackEgl. This is a better link.

PR #5408, opened 2024-11-26
fix accidental change of FallbackEgl to PreferEgl

e00E:
I accidentally changed this in a previous commit when I meant to only change the comment above it.

https://github.com/emilk/egui/pull/5392#discussion_r1859383653


lucasmerlin (approved):
Thanks!

PR #5410, opened 2024-11-28
Fix cargo machete

lucasmerlin:

cargo machete depends on cargo-platform which seems to bumped it's msrv. Installing via --locked should fix this for now. I think it's fine to do this manually instead of using the cargo action since it's so simple and the action we used basically did the same (without --locked)


emilk (approved):
Nice, thanks

PR #5411, opened 2024-11-28
Optimize editing long text by caching each paragraph

afishhh:
## What
(written by @emilk)
When editing long text (thousands of line), egui would previously re-layout the entire text on each edit. This could be slow.

With this PR, we instead split the text into paragraphs (split on `\n`) and then cache each such paragraph. When editing text then, only the changed paragraph needs to be laid out again.

Still, there is overhead from splitting the text, hashing each paragraph, and then joining the results, so the runtime complexity is still O(N).

In our benchmark, editing a 2000 line string goes from ~8ms to ~300 ms, a speedup of ~25x.

In the future, we could also consider laying out each paragraph in parallel, to speed up the initial layout of the text.

## Details
This is an ~~almost complete~~ implementation of the approach described by emilk [in this comment](<https://github.com/emilk/egui/issues/3086#issuecomment-1724205777>), excluding CoW semantics for `LayoutJob` (but including them for `Row`).
It supersedes the previous unsuccessful attempt here: https://github.com/emilk/egui/pull/4000.

Draft because:
- [X] ~~Currently individual rows will have `ends_with_newline` always set to false.
This breaks selection with Ctrl+A (and probably many other things)~~
- [X] ~~The whole block for doing the splitting and merging should probably become a function (I'll do that later).~~
- [X] ~~I haven't run the check script, the tests, and haven't made sure all of the examples build (although I assume they probably don't rely on Galley internals).~~
- [x] ~~Layout is sometimes incorrect (missing empty lines, wrapping sometimes makes text overlap).~~
- A lot of text-related code had to be changed so this needs to be properly tested to ensure no layout issues were introduced, especially relating to the now row-relative coordinate system of `Row`s. Also this requires that we're fine making these very breaking changes.

It does significantly improve the performance of rendering large blocks of text (if they have many newlines), this is the test program I used to test it (adapted from <https://github.com/emilk/egui/issues/3086>):
<details>
<summary>code</summary>

```rust
use eframe::egui::{self, CentralPanel, TextEdit};
use std::fmt::Write;

        ..Default::default()
    };

    eframe::run_native(
        "editor big file test",
        options,
        Box::new(|_cc| Ok(Box::<MyApp>::new(MyApp::new()))),
    )
}

struct MyApp {
    text: String,
}

impl MyApp {
    fn new() -> Self {
        let mut string = String::new();
        for line_bytes in (0..50000).map(|_| (0u8..50)) {
            for byte in line_bytes {
                write!(string, " {byte:02x}").unwrap();
            }
            write!(string, "\n").unwrap();
        }
        println!("total bytes: {}", string.len());
        MyApp { text: string }
    }
}

impl eframe::App for MyApp {
            let start = std::time::Instant::now();
                let code_editor = TextEdit::multiline(&mut self.text)
                    .code_editor()
                    .desired_width(f32::INFINITY)
                    .desired_rows(40);
                let response = code_editor.show(ui).response;
                if response.changed() {
                    println!("total bytes now: {}", self.text.len());
                }
            });
            let end = std::time::Instant::now();
            let time_to_update = end - start;
            if time_to_update.as_secs_f32() > 0.5 {
                println!("Long update took {:.3}s", time_to_update.as_secs_f32())
            }
        });
    }
}
```
</details>

I think the way to proceed would be to make a new type, something like `PositionedRow`, that would wrap an `Arc<Row>` but have a separate `pos` ~~and `ends_with_newline`~~ (that would mean `Row` only holds a `size` instead of a `rect`). This type would of course have getters that would allow you to easily get a `Rect` from it and probably a `Deref` to the underlying `Row`.
~~I haven't done this yet because I wanted to get some opinions whether this would be an acceptable API first.~~ This is now implemented, but of course I'm still open to discussion about this approach and whether it's what we want to do.

Breaking changes (currently):
- The `Galley::rows` field has a different type.
- There is now a `PlacedRow` wrapper for `Row`.
- `Row` now uses a coordinate system relative to itself instead of the `Galley`.

* Closes <https://github.com/emilk/egui/issues/3086>


afishhh:
Okay, so I did the thing and implemented the `Row` wrapper. This turned out to be more difficult than expected, and requires changing the semantics of `Row`s themselves. `glyphs` and the `mesh`(`_bounds`) must be relative to the `Row` itself, instead of the whole `Galley`, as otherwise repositioning them wouldn't work (unless we did something hacky like store two different offsets). Currently I also implemented `Deref<Target = Row>` for the `PlacedRow` wrapper (bike-shedding appreciated) which may not be the best idea as it makes it easy to confuse the different coordinate systems of the offset and non-offset `Row`.

Currently this "somewhat" works, I've managed to fix the selection issues I was seeing, but there is some layout issues you can actually see in the live demo above where the text is overlapping sometimes when wrapped (since this was present before the whole `PlacedRow` mess, I think this may be related to galley-merging).

afishhh:
So I've managed to make this look pretty correct, these are the remaining issues:
- `Row`s now have *very* weird semantics because an empty `Row` that is in-between two non-empty lines will actually have a size of zero. This *could* be solved by actually incorporating the line-terminating newline into the cached job, but there is one issue with that approach because then each of the line galleys will now contain a trailing `Row` that would have to be treated specially by the merging code.
- I've definitely introduced changes in rounding which is probably what makes the snapshot tests fail (apart from the fact that they sometimes segfault on my machine).

afishhh:
I think I've made some good progress:
- Made `Row` hold `ends_with_newline` again and just dropped excess rows during galley-merging, this means `Row` once again has pretty sensible semantics.
- Properly respecting `LayoutJob::round_output_size_to_nearest_ui_point` fixed a lot of the snapshot test failures.

The remaining test failures are:
- All rendering tests seem to have slight differences but in *separators* not in text which suggests that sizing might be ever so slightly wrong.
- The "Misc Demos" snapshot test also seems to fail, possibly due to a *slightly* incorrect alignment after the button widget?

I'm wondering whether it is worth fixing fixing these tests, should I just regenerate the snapshots instead?

crates/epaint/src/text/text_layout.rs
```diff
@@ -583,24 +595,25 @@ fn halign_and_justify_row(
         }
     }
 
-    // Note we ignore the leading/trailing whitespace here!
-    row.rect.min.x = target_min_x;
-    row.rect.max.x = target_max_x;
+    // Note we **don't** ignore the leading/trailing whitespace here!
+    row.size.x = target_max_x - target_min_x;
```
emilk, line 599: Why? What does this affect?

crates/epaint/src/tessellator.rs
```diff
@@ -1783,11 +1783,13 @@ impl Tessellator {
                 continue;
             }
 
+            let final_pos = galley_pos + row.pos.to_vec2();
```
emilk, line 1786: maybe call this `final_row_pos` for extra clarity

crates/epaint/src/text/fonts.rs
```diff
@@ -762,13 +905,26 @@ impl GalleyCache {
...
             }
             std::collections::hash_map::Entry::Vacant(entry) => {
-                let galley = super::layout(fonts, job.into());
-                let galley = Arc::new(galley);
-                entry.insert(CachedGalley {
-                    last_used: self.generation,
-                    galley: galley.clone(),
-                });
-                galley
+                if job.break_on_newline {
```
emilk, line 908: Maybe we should also only do this if the text is long enough and/or actually contains newline characters?

crates/epaint/src/text/fonts.rs
```diff
@@ -727,13 +727,156 @@ struct GalleyCache {
 }
 
 impl GalleyCache {
+    fn layout_multiline(&mut self, fonts: &mut FontsImpl, job: LayoutJob) -> Galley {
```
emilk, line 730: Nit: move this to under the main `layout` function (top->down ordering)

crates/epaint/src/text/fonts.rs
emilk, line 908: This deserves a comment

crates/epaint/src/text/fonts.rs
```diff
@@ -727,13 +727,156 @@ struct GalleyCache {
...
+        let mut left_max_rows = job.wrap.max_rows;
+        let mut galleys = Vec::new();
+        let mut first_row_min_height = job.first_row_min_height;
+        while current != job.text.len() {
+            let end = job.text[current..]
+                .find('\n')
+                .map_or(job.text.len(), |i| i + current + 1);
+            let start = current;
+
+            let mut line_job = LayoutJob {
```
emilk, line 746: This functions is very long.

I would love if we could break this up into parts, for instance:

```
let mut galleys = vec![];
for line_job in job.split_on_newlines() {
    …
}
Galley::concat(galleys)
```

emilk (commented):
This looks good! I haven't reviewed all of `GalleyCache` yet, because it is a very dense and complex function.

I wonder how we best test and/or benchmark this though 🤔

crates/epaint/src/text/fonts.rs
afishhh: Should such `LayoutJob::split_on_newlines` and `Galley::concat` functions be public API or do we want to keep them private?

crates/epaint/src/text/fonts.rs
afishhh: I suppose we can check for a '\n', initially I was skeptical but the actual layout is way more expensive than this string find will cost us so it's probably fine.

afishhh:
> I wonder how we best test and/or benchmark this though 🤔
> 
> I defiantly feel we should have a benchmark to justify this code, i.e. something that shows a big speedup before/after this PR.

I agree having a benchmark would be nice, currently the only thing I have is "it looks quicker" in when you have a lot of text (deletion of small segments works instantly for example).
At first I wanted to suggest a kittest benchmark but then I looked in `crates/egui_demo_lib/benches/benchmark.rs` and noticed there already seem to be text layout benchmarks, so it'd probably be relatively easy to do something like:
- Have a big (few megabytes) block of text
- Lay it out
- In the benchmark loop:
  - Randomly remove an N-byte segment from it (into a copy)
  - Lay it out

As for tests, I've encountered issues with selection drawing multiple times while working on this at this point, so maybe it would be beneficial if this was actually checked by the test suite, it looks like mouse events are supported in kittest so maybe that's a good start. Especially since I also think selection visuals might be another good place for optimization because currently selecting the whole text will force all `Row`s to be cloned again.

It also looks like the `rendering_test` tests still seem to be failing on CI. The diffs seem to contain three lines near the middle that look like the bottom of the text background, maybe the background is slightly too small/large?

crates/epaint/src/tessellator.rs
afishhh: Done.

crates/epaint/src/text/fonts.rs
afishhh: Done.

crates/epaint/src/text/fonts.rs
afishhh: I looked into splitting out these functions but it's actually non trivial, for example the splitting actually has to do the layout *while* it is doing the splitting because it has to update `LayoutJob::max_rows` as more rows are being laid out. As for `Galley::concat`, it needs access to `pixels_per_point` for correct rounding so it'd be at the very least awkward as a public API (I guess you'd need to pass in `Fonts`?) and as a private one it could accept `FontsImpl`, it also needs access to the full `LayoutJob` so it can store it inside the `Galley` and also to check `round_output_size_to_nearest_ui_point`.

I think we could do something like a `Galley::concat(fonts: &mut FontsImpl, job: LayoutJob, galleys: impl Iterator<Item = Galley>) -> Galley` and keep splitting in `GalleyCache::layout_multiline` or split it out into a method of `GalleyCache` but it has to have access to `GalleyCache` so it can't be on `LayoutJob`. Also at that point I don't know whether scattering the implementation detail of `FontsImpl` into a `Galley` method is any more clean than keeping it in `GalleyCache`.

Maybe another solution would be to actually disable this line caching if `LayoutJob::max_rows` is finite, because in that scenario there's going to be more ways to invalidate the cached lines anyways and the performance benefits will be much more situational (any edit that will cause a line-count change will cause a re-layout of all subsequent lines).

crates/epaint/src/text/fonts.rs
afishhh: I added *a* comment, let me know if you feel this is sufficient or if you had something else in mind.

afishhh: Added the newline part, I think adding a length check might not be worth it because it'd definitely make testing harder if an issue "only showed up if the text is longer than 1000 characters".

crates/epaint/src/text/text_layout.rs
afishhh: Fixed the issue found above and the comment. Although I still want to know what you think about changing the `Row::size` comment, in general halign is surprising with how it suddenly totally changes how the coordinates work even though it makes perfect sense in hindsight.

afishhh:
I added a benchmark similar to the scenario I described above.
On `master` it takes about 100ms per iteration on average, on `cache_galley_lines` it takes 825µs which is an improvement of around 123x. I believe this may be enough to justify the code.

emilk:
I tried to merge in `master` and push to your branch, but it fails despite 
image

[This thread](https://github.com/cli/cli/discussions/8794) makes me suspect `git lfs`. Can you try enabling it on your repository and see if it fixes my problem? I'd love to collaborate on getting this PR merged.

afishhh:
> I tried to merge in `master` and push to your branch, but it fails despite image
> 
> [This thread](https://github.com/cli/cli/discussions/8794) makes me suspect `git lfs`. Can you try enabling it on your repository and see if it fixes my problem? I'd love to collaborate on getting this PR merged.

I don't see an option anywhere on github pertaining to enabling LFS, so I don't think it's actually toggleable?
From what I could infer during my short investigation this appears to be an error that occurs while LFS objects are being pushed right? Then maybe the solution would be for you to (temporarily) remove the LFS pre-push hook on your side and it would not try pushing the LFS objects?
I've never really used git LFS so it's just a shot in the dark but I can't see anything I can do from my side to fix it.

emilk:
You can merge in my branch [`cache_galley_lines`](https://github.com/emilk/egui/tree/cache_galley_lines) to get the merge conflicts solved

afishhh:
> You can merge in my branch [`cache_galley_lines`](https://github.com/emilk/egui/tree/cache_galley_lines) to get the merge conflicts solved
Did that

crates/epaint/src/text/fonts.rs
emilk: Splitting out the concatenation was pretty trivial:

[`9fa294f` (#5560)](https://github.com/emilk/egui/pull/5560/commits/9fa294f21ff75df5d77f956b3133820d9bd53625)

emilk: I agree we can ignore `max_rows` for now (i.e. fall back to the old/slow path if it is set).

emilk:
Since I cannot push to this branch directly, I created https://github.com/emilk/egui/pull/5560 to shadow this PR. Feel free to merge it in to your branch.

emilk:
The tests are failing.

afishhh:
> The tests are failing.

When they don't segfault I can indeed reproduce the rendering tests failing (and even two others (using nvidia)).
It does look to me as if the failures are caused by some minor misalignment (although the diffs did change since `17a5f1f`), at most the 1.00dpi test has weird looking suspicious rectangles in the diff that maybe is not an nvidia hallucination.
I may look into fixing these but also don't want to spend days chasing down rounding differences.

afishhh:
It does look like the new easymarkeditor test on master has an incorrectly positioned strike-through and underline, that seems like something that's an actual issue. I'll investigate that soon.

emilk:
I've done of bunch of cleanup, bug fixes, and added a test. There is still some rounding error causing small differences in rendering. I don't have time to look at that today.

emilk:
There is still some rounding error causing underscores, strikethroughs, and text background to _sometimes_ jump a pixel

crates/epaint/src/text/fonts.rs
```diff
@@ -767,20 +774,124 @@ impl GalleyCache {
...
+                    job.first_row_min_height
+                } else {
+                    0.0
+                },
+                round_output_to_gui: job.round_output_to_gui,
+            };
+
+            // Keep all (previous) sections (so the correct `section_index` is outputted),
+            // but shift their byte ranges:
+            paragraph_job.sections = job
```
emilk, line 846: I changed this code in order to get correct section indices… but this was a mistake. The original idea was _much_ quicker, even though it requires more book-keeping… will need to revert.

emilk:
There is a performance bug: editing a long text in the middle will srtill re-layout all _subsequent_ paragraphs of text. I'm not sure why, but until we fix that, there is little sense in merging this.

emilk:
I need to take a break from this now.

The benchmark shows only about a ~2x speedup, which is less than I would have thought, so I think it is doing extra layouting even when it shouldn't. If anyone wants to take a look at that I would appreciate it.

afishhh:
Okay I put off fixing the invalidation because it's non-trivial, but I managed to improve performance quite a bit by simply not re-rounding all row sizes in `round_output_to_gui` when called from `Galley::concat`, the sizes should already all be correctly rounded after initial layout and rounding them again here calls `make_mut` on all rows, which takes a lot of time.

afishhh:
Not adding the merged `Galley` into the cache entirely seems to still give acceptable performance so I opted for that as a fix for the invalidation issue.
If necessary something more complex like actually accounting for dependencies between `CachedGalley`s could be done but this seems to work fine.

crates/epaint/src/text/text_layout_types.rs
```diff
@@ -772,11 +772,16 @@ impl Galley {
...
-            row.size = row.size.round_ui();
+    pub(crate) fn round_output_to_gui(&mut self, round_row_sizes: bool) {
+        // This function is called in `Galley::concat` but by the time it is called
+        // Row sizes have already been rounded once after initial paragraph layout,
+        // therefore rounding them again here only causes an unnecessary make_mut.
+        if round_row_sizes {
+            for placed_row in &mut self.rows {
+                // NOTE: we round the size, but not the position, because the position should be _pixel_ aligned.
+                let row = Arc::make_mut(&mut placed_row.row);
+                row.size = row.size.round_ui();
```
emilk, line 783: A less brittle solution would be to only call `make_mut` if not already rounded:

```rs
// Optimization: only call `make_mut` if necessary (can cause a deep clone)
let rounded_size = placed_row.row.size.round_ui();
if placed_row.row.size != rounded_size {
    Arc::make_mut(&mut placed_row.row).size = rounded_size;
}
```

crates/epaint/src/text/fonts.rs
```diff
@@ -783,15 +783,10 @@ impl GalleyCache {
...
-                        CachedGalley {
-                            last_used: self.generation,
-                            galley: galley.clone(),
-                        },
-                    );
-                    galley
+                    // TODO(afishhh): This Galley cannot be added directly into the cache without taking
+                    //     extra precautions to make sure all component paragraph Galleys are not invalidated
+                    //     immediately next frame (since their `last_used` will not be updated).
+                    Arc::new(galley)
```
emilk, line 789: This means we don't cache the full galley, and so do the paragraph-splitting and concatenation each frame… that's potentially a performance regression for large texts.

Better I think to just improve the cache

afishhh: So I implemented this, but I still seem to experience eviction of all the paragraphs in my practical application of this, although it doesn't seem to happen in a small test app. I've pushed the implementation for review (better name or approach for `layout_internal`?) and may investigate the weird cache evictions later (would blame deferred viewpoints but that wouldn't explain why it worked slightly better when re-splitting every frame).

emilk (approved):
Nice, let's get this in!

scvalex:
I think this broke multi-line `TextShape` rotation.  Could I ask one of you to have a look at #7397, please?

PR #5412, opened 2024-11-29
macOS: Add `movable_by_window_background` option to viewport

jim-ec:
Add an option called `movable_by_window_background` alongside a new builder method. When set to true, the window is movable by dragging its background ([Apple Docs](https://developer.apple.com/documentation/appkit/nswindow/ismovablebywindowbackground))

This is exclusive to macOS systems, similar to `fullsize_content_view`.

### DIFF
--- a/crates/egui/src/viewport.rs
+++ b/crates/egui/src/viewport.rs
@@ -434,6 +434,8 @@ impl ViewportBuilder {
     }
 
     /// macOS: Set to `true` to allow the window to be moved by dragging the background.
+    ///
+    /// Enabling this feature can result in unexpected behaviour with draggable UI widgets such as sliders.
     #[inline]
     pub fn with_movable_by_background(mut self, value: bool) -> Self {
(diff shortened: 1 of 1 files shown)


crates/egui/src/viewport.rs
```diff
@@ -432,6 +433,13 @@ impl ViewportBuilder {
         self
     }
 
+    /// macOS: Set to `true` to allow the window to be moved by dragging the background.
+    #[inline]
```
lucasmerlin, line 437: https://github.com/user-attachments/assets/15c76a6e-da36-4290-a734-30fc5d69ecef

It doesn't really work well with sliders and draggable widgets, but I guess that's expected. 

But maybe add a note to the comment that it doesn't, otherwise people might expect that to work and open issues about it.

lucasmerlin (changes requested):

jim-ec:
Sorry for taking so long, I added your requested warning to the doc.

PR #5415, opened 2024-11-29
Add new `Rect::intersects_ray_from_center` method

grtlr:
Title.

### DIFF
--- a/crates/emath/src/rect.rs
+++ b/crates/emath/src/rect.rs
@@ -650,6 +650,8 @@ impl Rect {
     ///
     /// A ray that starts inside the rect will return `true`.
     pub fn intersects_ray(&self, o: Pos2, d: Vec2) -> bool {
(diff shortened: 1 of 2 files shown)


crates/emath/src/rect.rs
```diff
@@ -671,6 +671,30 @@ impl Rect {
 
         0.0 <= tmax && tmin <= tmax
     }
+
+    /// Where does a ray from the center intersect the rectangle?
+    ///
+    /// `d` is the direction of the ray and assumed to be normalized.
```
lucasmerlin, line 677: Should we add a debug_assert that checks if the ray is normalized?

grtlr: I didn't add a check because the code above also doesn't have one.

I'm not leaning strongly in either direction.

Wdyt @emilk?

emilk: Good idea

emilk (approved):

PR #5416, opened 2024-11-29
Allow attaching custom user data to a screenshot command

emilk:
This lets users trigger a screenshot from anywhere, and then when they get back the results they have some context about what part of their code triggered the screenshot.

### DIFF
--- a/crates/egui/src/data/user_data.rs
+++ b/crates/egui/src/data/user_data.rs
@@ -42,16 +42,33 @@ impl serde::Serialize for UserData {
     where
         S: serde::Serializer,
     {
-        ().serialize(serializer) // can't serialize an `Any`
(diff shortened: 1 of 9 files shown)


crates/egui/src/data/user_data.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<'de> serde::Deserialize<'de> for UserData {
+    fn deserialize<D>(_deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        Ok(Self::default()) // can't serialize an `Any`
```
emilk, line 55: There should be a better way of doing this

PR #5418, opened 2024-11-30
Debug kittest tests with eframe

lucasmerlin:
It would be nice to have some way to debug kittest tests by running them with eframe. I gave this a try but ran into the problem that tests usually aren't run on the main thread. 

It could be possible to solve this via https://github.com/emilk/eterm (but we'd need to update it to work with current egui)

See also: https://github.com/rerun-io/rerun/pull/8248#discussion_r1863191027

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1363,6 +1363,7 @@ version = "0.29.1"
 dependencies = [
  "dify",
  "document-features",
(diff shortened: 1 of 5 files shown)


emilk:
* Implemented in https://github.com/emilk/egui/pull/8120

Issue #5164, opened 2024-09-25
Align shapes to physical pixel grid


emilk:
When painting shapes, it is _sometimes_ preferably to align them to the physical pixel grid.

There are a few different ways we can attack this:
A) A document and make it easy for users to align
B) Always align shapes, and opt-in to non-aligned shapes
C) Always align all shapes

I think `B` is preferable, the question is how to accomplish it on a per-shape basis.

We could add a `bool` flag to each shape, and have the tessellator do the alignment, unless the flag is set.
Or the `Painter` aligns all shapes, unless you call `painter.add_unaligned(…)` or something like that.

### Sharpness
We always align text to the physical grid (unless rotated), because otherwise it looks blurry.

Similarly we align many vertical lines, horizontal lines, and rectangle frames to make them sharper:
* https://github.com/emilk/egui/pull/4943

### Tiling
When painting many rectangles that should seamlessly tile:
https://github.com/emilk/egui/issues/4928#issuecomment-2354843677
![image](https://github.com/user-attachments/assets/ef96f6f4-1e50-4534-8531-f320a7b96743)

In this case, the end results would look better if the rectangle was aligned to the physical pixel grid.

## When not to align
When you want to smoothly animate something.

emilk:
I've been thinking about this a bit more, and I have reached the following conclusion: only thin vertical and horizontal lines should be rounded to the closest physical pixel.

Anti-aliasing/feathering is always useful for diagonal lines. For vertical and horizontal lines (like the borders of a rectangle) it is usually unnecessary, except for the case when these lines are thick AND animated (either moving or growing/shrinking). Consider a thick box moving sideways. Without AA, a new column of pixels gets turned on at regular intervals, which is an annoying discontinuity. If we turn on AA, each new column "fades in", removing the jarring "popping". So AA should be ON in the this case. BUT in the case of a thin line moving to the right, then AA will turn a solid column of pixels into two columns of gray pixels, and then into a another column of solid pixels again, and so on. This will cause the line to appear to "blink" or "pulsate" as it moves, which is very annoying. If we turn off AA (and round to physical pixels), then you will always have a solid column of pixels turned on, so at each point in time the line will have the same brightness and width. Yes, it will still jump in discreet steps, but in my experience, this is much more desirable. I'll try to generate a video to motivate this further.

In any case, this leaves us with the conclusion that thin vertical and horizontal lines should always be rounded to physical pixels, and nothing else.

[...]

PR #5420, opened 2024-11-30
Make text cursor always appear on click 

juancampa:

### Problem
When clicking on a TextEdit sometimes the cursor doesn't appear immediately which makes it feel like the click was not registered for a second. This is because the start time for the blinking animation is only reset on keyboard input, but not on mouse interaction.

It's hard to tell on the video but the cursor doesn't show immediately after clicking if the blink timer happens to be off.

https://github.com/user-attachments/assets/9f049bd0-0375-4291-b2ef-697777fb854d

### Solution
Reset the click timer every time a `TextEdit` is clicked. 

Additionally, the cursor is now correctly painted on the pixel boundary. IMO we should default to 1px cursor (instead of 2px) but that's not included in this PR. Happy to make that change too.

https://github.com/user-attachments/assets/6c489414-f2c4-4dc6-85dd-f8bc457edad0

### DIFF
--- a/crates/egui/src/text_selection/visuals.rs
+++ b/crates/egui/src/text_selection/visuals.rs
@@ -96,6 +96,8 @@ pub fn paint_text_selection(
 pub fn paint_cursor_end(painter: &Painter, visuals: &Visuals, cursor_rect: Rect) {
     let stroke = visuals.text_cursor.stroke;
 
+    // Ensure the cursor is aligned to the pixel grid for whole number widths.
+    // See https://github.com/emilk/egui/issues/5164
     let (top, bottom) = if (stroke.width as usize) % 2 == 0 {
         (
             painter.round_pos_to_pixels(cursor_rect.center_top()),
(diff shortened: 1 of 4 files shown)


crates/egui/src/text_selection/visuals.rs
```diff
@@ -96,8 +96,17 @@ pub fn paint_text_selection(
...
+        (
+            painter.round_pos_to_pixels(cursor_rect.center_top()),
+            painter.round_pos_to_pixels(cursor_rect.center_bottom()),
+        )
+    } else {
+        (
+            painter.round_pos_to_pixel_center(cursor_rect.center_top()),
+            painter.round_pos_to_pixel_center(cursor_rect.center_bottom()),
+        )
+    };
```
juancampa, line 109: I would think it's safe to assume that users will be using either 1px or 2px cursor thickness.

lucasmerlin: Will this work when the scale is 150%?

Also it'd be nice to have a screenshot test that shows the cursor, maybe you can add one?

emilk, line 109: This could use a comment explaining what is going on, and ideally with a link to https://github.com/emilk/egui/issues/5164 so we can remember to update this code if we move the rounding to epaint

juancampa: Comment added!

juancampa:
I just noticed that another one of my PRs (#3660) got merged earlier today and was missing some backticks which was causing `check.sh` to fail. I just fixed that as well so this should be good to go now.

emilk:
please merge in `master` to fix CI

juancampa:
@emilk merged

PR #5421, opened 2024-12-01
Update MSRV to Rust 1.79

emilk:
Mostly to fix `cargo-machete` CI


lucasmerlin:
See also #5410 (which should be enough to fix cargo machete, at least until they release a new version)
But updating rust is also nice, of course

emilk:
Oops, I missed your PR. Well. Let's do both then :)

PR #5426, opened 2024-12-03
Move `egui::util::cache` to `egui::cache`; add `FramePublisher`

emilk:
This moves `egui::util::cache` to `egui::cache` (the old path is deprecated, but still works).

It also adds the `FramePublisher` helper, which can be used to publish a value which will be retained for this frame and the next:

``` rs
pub type MyPublisher = egui::cache::FramePublisher<MyKey, MyValue>;

// Publish:
ctx.memory_mut(|mem| {
    mem.caches.cache::<MyPublisher>().set(key, value);
});

// Retrieve:
let value: Option<MyValue> = ctx.memory_mut(|mem| {
    mem.caches
        .cache::<MyPublisher>()
        .get(key)
        .clone()
})
```

### DIFF
--- a/crates/egui/src/cache/cache_storage.rs
+++ b/crates/egui/src/cache/cache_storage.rs
@@ -1,0 +1,68 @@
+use super::CacheTrait;
+
+/// A typemap of many caches, all implemented with [`CacheTrait`].
+///
+/// You can access egui's caches via [`crate::Memory::caches`],
+/// found with [`crate::Context::memory_mut`].
+///
+/// ```
+/// use egui::cache::{CacheStorage, ComputerMut, FrameCache};
+///
+/// #[derive(Default)]
+/// struct CharCounter {}
+/// impl ComputerMut<&str, usize> for CharCounter {
+///     fn compute(&mut self, s: &str) -> usize {
+///         s.chars().count()
+///     }
+/// }
+/// type CharCountCache<'a> = FrameCache<usize, CharCounter>;
+///
+/// # let mut cache_storage = CacheStorage::default();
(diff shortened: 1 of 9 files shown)


zhatuokun:
If I call `CacheStorage::cache` somewhere, it will insert a `Cache` that implements the `CacheTrait`. 

Although `Cache` will call `evict_cache` to clear the internal cache, even if the length of the internal cache of cache is 0, it still seems to exist in the `CacheStorage`.

So would it be better to delete the cache with a cache length of 0 in the `CacheStorage::update`?

zhatuokun:
@emilk I tested this new feature and found that it works well in general, but when you have an `Immediate Viewport`, it will return `None`.

Here is my test code.

```rs
type MyPublisher = egui::cache::FramePublisher<String, u32>;

fn main() -> eframe::Result {
    eframe::run_simple_native("Example", eframe::NativeOptions::default(), |ctx, _| {
        let outer_key = String::from("outer_key");
        let value = ctx.memory_mut(|m| m.caches.cache::<MyPublisher>().get(&outer_key).copied());
        println!("outer {value:?}"); // None
        ctx.memory_mut(|m| m.caches.cache::<MyPublisher>().set(outer_key, 0));

        ctx.show_viewport_immediate(
            egui::ViewportId::from_hash_of("ExampleViewport"),
            Default::default(),
            |ctx, _| {
                let inner_key = String::from("inner_key");
                let value = ctx.memory_mut(|m| m.caches.cache::<MyPublisher>().get(&inner_key).copied());
                println!("inner: {value:?}"); // None
                ctx.memory_mut(|m| m.caches.cache::<MyPublisher>().set(inner_key, 0));
            },
        );
    })
}
```

PR #5431, opened 2024-12-04
Support loading images with weird urls and improve error message

### LINKED ISSUE #5341: Some image urls don't load after #5324

lucasmerlin:
* Closes #5341

### DIFF
--- a/crates/egui/src/load.rs
+++ b/crates/egui/src/load.rs
@@ -77,16 +77,19 @@ pub enum LoadError {
     /// Programmer error: There are no image loaders installed.
     NoImageLoaders,
 
-    /// A specific loader does not support this scheme, protocol or image format.
+    /// A specific loader does not support this scheme or protocol.
     NotSupported,
+
+    /// A specific loader does not support the format of the image.
+    FormatNotSupported { detected_format: Option<String> },
 
     /// Programmer error: Failed to find the bytes for this image because
     /// there was no [`BytesLoader`] supporting the scheme.
     NoMatchingBytesLoader,
 
     /// Programmer error: Failed to parse the bytes as an image because
-    /// there was no [`ImageLoader`] supporting the scheme.
-    NoMatchingImageLoader,
+    /// there was no [`ImageLoader`] supporting the format.
+    NoMatchingImageLoader { detected_format: Option<String> },
 
     /// Programmer error: no matching [`TextureLoader`].
     /// Because of the [`DefaultTextureLoader`], this error should never happen.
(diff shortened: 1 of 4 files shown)


lucasmerlin:
The same problem also exists for SvgLoader and GifLoader, but I think these can be fixed in a separate PR (and it's probably not so urgent to fix those, since there it's always been broken)

crates/egui/src/load.rs
```diff
@@ -96,6 +99,19 @@ pub enum LoadError {
     Loading(String),
 }
 
+impl LoadError {
+    pub fn size(&self) -> usize {
```
emilk, line 103: let's give it a more descriptive name:

```suggestion
    pub fn byte_size(&self) -> usize {
```

emilk (approved):
nice

Issue #5189, opened 2024-09-30
Color picker doesn't capture the input used to close it


abey79:
One can close the color picker by clicking outside of it or hitting escape. Unfortunately, these inputs are not captured and thus potentially have an unwanted secondary effect.

In the case of rerun,  escape will have the effect of deselecting whatever is currently selected, and the click will be processed by whatever happens to be under the mouse cursor (often, the view, which again leads to deselection).

https://github.com/user-attachments/assets/d3cbd005-2132-4c2c-af4c-52daba8a7308

emilk:
This is by design (though arguably not good design). The thinking (within egui) is: pressing escape should escape _everything_. Otherwise what it escapes would depend on the order of which the UI is laid out, which is not obvious to a user ("why did A close instead of B?").

abey79:
What is the "correct" way to opt-out of a popup menu then?

> Otherwise what it escapes would depend on the order of which the UI is laid out, which is not obvious to a user ("why did A close instead of B?").

Perhaps naively, I'm expecting popup-menu/picker to be modal, so (1) they may not have sibling areas—only parent, over which they have precedence and (2) they may have zero or one "child" modal (e.g. a sub-menu), which would have precedence. In this model, it feels quite natural in how keystroke would be captured. Surely I'm missing dozen of edge cases and assorted implementation trickeries here, though.

>  pressing escape should escape everything.

My mental model doesn't conform to that. If have a stack of "modal things" open, I'm expecting to get out of it by hitting escape as many times as I have things.

<hr>


Related:
- https://github.com/emilk/egui/issues/2549 (in this case, the opt-out behaviour is natural: just release the click outside of the menu).
- https://github.com/emilk/egui/issues/3718

emilk:
Switching all `i.key_pressed(Key::Escape)` to use `consume_key` instead might be the right solution, we just need to be careful to do it in the right place so that:
* Pressing escape closes all `Menu`:s (not just the innermost submenu)
* Pressing escape closes the innermost popup (if nested)
* …?

abey79:
> Pressing escape closes all Menu:s (not just the innermost submenu)

Debatable :)

Personally I'd rather have to ESC twice when I want full out than having no way to opt-out of a single level of sub-menu.

Counter argument: macOS closes all layers of menu on ESC.

lucasmerlin:
[...]

PR #5433, opened 2024-12-04
Consume escape keystroke when bailing out from a drag operation

abey79:
Consume the escape keystroke when it's used to bail out of a drag-and-drop operation.

Successfully tested with rerun (aka `ESC`aping from a drag no longer deselects the current selection).

* related to https://github.com/emilk/egui/issues/5189


emilk (approved):
Nice!

PR #5436, opened 2024-12-05
Fix: Some deletions due to Bugs with TextEdit

rustbasic:
* Related #5420

There is a bug where when you type a lot of characters, the button gets pushed back.

See below :
https://github.com/user-attachments/assets/bd9e93bc-7609-4bbe-9efc-e78d11dd1987

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -1,6 +1,5 @@
 use std::sync::Arc;
 
-use emath::Rect;
 use epaint::text::{cursor::CCursor, Galley, LayoutJob};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -723,16 +722,6 @@ impl<'t> TextEdit<'t> {
...
-            // Allocate additional space if edits were made this frame that changed the size. This is important so that,
-            // if there's a ScrollArea, it can properly scroll to the cursor.
-            let extra_size = galley.size() - rect.size();
-            if extra_size.x > 0.0 || extra_size.y > 0.0 {
-                ui.allocate_rect(
-                    Rect::from_min_size(outer_rect.max, extra_size),
-                    Sense::hover(),
-                );
-            }
-
```
emilk, line 735: Removing code marked as "important" is not a good look. If you want to change this, please create a test (using egui_kittest) for both the issue described in the comment, and your issue

emilk (changes requested):

PR #5450, opened 2024-12-08
Support wgpu-tracing with same mechanism as wgpu examples

EriKWDev:
Gets the WGPU_TRACE env variable and gives it as an optional argument
to request_device. Same mechanism as the wgpu-examples:

https://github.com/gfx-rs/wgpu/blob/11b51693d3dc883b55b5ec0e30c340e43e6fac50/examples/src/framework.rs#L316


Wumpf (approved):
sure why not! 👍

PR #5459, opened 2024-12-10
Fix: don't interact with `Area` outside its `constrain_rect`

MScottMcBee:

This PR makes an area's interact rect intersect its constrain rect. This fixes an issue where masked areas would still intercept input.

Before:
before

After:
after


emilk (approved):

emilk:
Nice one

PR #5466, opened 2024-12-12
Workaround for egui having wrong scale in firefox

lucasmerlin:


jprochazk (approved):

emilk:
https://www.egui.rs/#rendering is now failing on chromium

image

Text is blurry, lines are not pixel-aligned

emilk:
Even when you zoom? ctrl+ etc

also… that image does not look very crisp, but maybe that's just the screenshot tool?

PR #5480, opened 2024-12-16
Try to fix broken images on egui.rs

emilk:
The images in the widget gallery on egui.rs are broken:
image

Not sure why yet, and I fail to reproduce locally


InfoInnovator:
I really dont want to intervene here, but @lucasmerlin might be right. I cannot even open the image with my standard image viewer. If i copy the image from the eframe/data to egui_demo_lib/data it works.

InfoInnovator:
> @InfoInnovator if you have this problem locally, make sure [you have properly setup git lfs](https://github.com/emilk/egui/blob/master/CONTRIBUTING.md#making-a-pr).

My bad. I didnt expect somebody to use lfs for a small PNG and didnt know of the guide. After pulling the lfs files, the web demo works as intended.

emilk:
Ah, then we should mark the .png:s used for the demo as non-lfs, and also see if we can improve the error message (does lfs links start with some magic bytes?)

lucasmerlin:
> (does lfs links start with some magic bytes?)

I think so:
```
version https://git-lfs.github.com/spec/v1
oid sha256:4d7a214614ab2935c943f9e0ff69d22eadbb8f32b1258daaa5e2ca24d17e2393
size 12345
```

emilk:
Improvement!

image

emilk:
Fixed! Just need to update CI not to complain
image

PR #5481, opened 2024-12-16
Revert "Workaround for egui having wrong scale in firefox"

emilk:
* Reverts emilk/egui#5466

Restores pixel-perfect rendering on chromium, Mac


emilk:
Sorry @lucasmerlin, you'll need to find some other workaround. I don't want blurry text 😬

PR #5482, opened 2024-12-16
Make frame delay on screenshots consistently one frame on web as well

Wumpf:
Native is already delayed by a frame because it calls `handle_viewport_output` -> `egui_winit::process_viewport_commands` after drawing. On web however, we process input including viewport commands separately from drawing.
This adds an arbitrary frame delay mechanism for web and then uses this with 1 frame delay always


emilk (approved):
🙏

PR #5488, opened 2024-12-16
Revert "forward x11 and wayland features to glutin" (#5391)

emilk:
* Reverts https://github.com/emilk/egui/pull/5391

Because it causes head-ache: https://github.com/emilk/eframe_template/actions/runs/12357896151/job/34487194281


emilk:
EDIT: the fix is to add "wayland" as a feature AND update to Rust 1.81

this revert can be reverted (will do, with some added docs to boot)

PR #5489, opened 2024-12-16
Release eframe 0.30.1 - fix glow on Linux

emilk:


emilk:
Not needed after all - just need to add "wayland" feature, apparently, see

* https://github.com/emilk/eframe_template/pull/175

Issue #5491, opened 2024-12-16
Errors in epaint update 0.30.0 related to `profiling`


frankvgompel:
```rust
   Compiling epaint v0.30.0
error[E0433]: failed to resolve: could not find `function_scope` in `profiling`
  --> /Users/frankvangompel/.cargo/registry/src/index.crates.io-6f17d22bba15001f/epaint-0.30.0/src/mesh.rs:88:20
   |
88 |         profiling::function_scope!();
   |                    ^^^^^^^^^^^^^^ could not find `function_scope` in `profiling`

error[E0433]: failed to resolve: could not find `function_scope` in `profiling`
   --> /Users/frankvangompel/.cargo/registry/src/index.crates.io-6f17d22bba15001f/epaint-0.30.0/src/mesh.rs:114:20
    |
114 |         profiling::function_scope!();
    |                    ^^^^^^^^^^^^^^ could not find `function_scope` in `profiling`

error[E0433]: failed to resolve: could not find `function_scope` in `profiling`
    --> /Users/frankvangompel/.cargo/registry/src/index.crates.io-6f17d22bba15001f/epaint-0.30.0/src/tessellator.rs:1606:20
     |
1606 |         profiling::function_scope!();
     |                    ^^^^^^^^^^^^^^ could not find `function_scope` in `profiling`

error[E0433]: failed to resolve: could not find `function_scope` in `profiling`
    --> /Users/frankvangompel/.cargo/registry/src/index.crates.io-6f17d22bba15001f/epaint-0.30.0/src/tessellator.rs:2001:20
     |
2001 |         profiling::function_scope!();
     |                    ^^^^^^^^^^^^^^ could not find `function_scope` in `profiling`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `epaint` (lib) due to 4 previous errors
```

apessino:
Confirmed - I just tried updating and I am getting the same errors. I also tried cloning the repo instead of using the cargo.io release and the problem is still there.

A bad merge, perhaps?

lucasmerlin:
I can reproduce this, but I'm not really sure why it would happen, it looks like it should be working 🤔 
As a workaround, you can add the following to your cargo.toml:
```toml
profiling = { version = "1", features = ["type-check"], default-features = false }
```

@teddemunnik can you have a look at this? 

To reproduce this, the following cargo.toml has this problem for me: 
```toml
[package]
name = "egui_playground"
version = "0.1.0"
edition = "2021"

[...]

PR #5494, opened 2024-12-16
Use correct minimum version of `profiling` crate

lucasmerlin:
We need profiling::function_scope which was introduced in 1.0.16, so this is the minimum required version
* Closes <https://github.com/emilk/egui/issues/5491>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,7 +87,7 @@ log = { version = "0.4", features = ["std"] }
 nohash-hasher = "0.2"
 parking_lot = "0.12"
 pollster = "0.4"
-profiling = { version = "1.0", default-features = false }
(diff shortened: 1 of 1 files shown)


lucasmerlin:
Maybe we could try to setup a workflow that runs cargo check / test with -Z-minimal-version, to prevent issues like this in the future?

https://github.com/taiki-e/cargo-minimal-versions
https://internals.rust-lang.org/t/zminimal-versions-cargo-update-and-cargo-upgrade/21335

emilk:
> Maybe we could try to setup a workflow that runs cargo check / test with -Z-minimal-version, to prevent issues like this in the future?

that would be nice yes - can you create an issue for this (or just add the workflow)? 🙏

lucasmerlin:
Done: #5497

PR #5505, opened 2024-12-19
Add `egui::Scene` for panning/zooming a `Ui`

grtlr:
This is similar to `ScrollArea`, but:
* Supports zooming
* Has no scroll bars
* Has no limits on the scrolling

## TODO
* [x] Automatic sizing of `Scene`s outer bounds
* [x] Fix text selection in scenes
* [x] Implement `fit_rect`
* [x] Document / improve API


emilk:
I wonder if we can come up with a better name for this? 🤔

grtlr:
Some things that come to my mind (lower entries of the list are augmented by our AI overlords, you might guess which):

* (infinite) Canvas (but might be too close the `<canvas />`)
* Scene
* View (too general; Viewport is already taken)
* Pad
* Board
* Expanse

xTibor:
Viewfinder? 

This idea comes from photography. Short Wikipedia [definition](https://en.wikipedia.org/w/index.php?title=Viewfinder&oldid=1255709208):

> ... a viewfinder is a device on a camera that a photographer uses to determine exactly where the camera is pointed, and approximately how much of that view will be photographed

I think this kinda describes what this container does.

YgorSouza:
[Viewbox](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/viewBox) is a term that is sometimes used for this kind of thing, although often it refers to the coordinates of the inner element that are in view, rather than the container widget itself.

crates/egui/src/containers/scene.rs
```diff
@@ -0,0 +1,164 @@
...
+    // Set the transformation to scale and then translate to center.
+    TSTransform::from_translation(center_screen.to_vec2() - center_scene * scale)
+        * TSTransform::from_scaling(scale)
+}
+
+#[derive(Clone, Debug)]
+#[must_use = "You should call .show()"]
+pub struct Scene {
+    min_scaling: Option<f32>,
+    max_scaling: f32,
```
emilk, line 36: Could this just be a `Rangef` ?

grtlr (commented):
Looks very nice—thank you @emilk for picking this up!

I can't approve my own PR, so here is a ✅.

Issue #1129, opened 2022-01-17
epaint reference software rasterizer


emilk:
I would like to have a software rasterizer for epaint, so that you can take the output shapes from `egui` and turn it into an image.

This would be useful for many areas:
* Generating regressions tests (make sure the GUI looks the same before/after)
* Generating example images
* Provide a reference for GPU renderers to compare themselves to.

Primarily I would want it to be simple and accurate rather than fast. However, a fast rasterizer would also be nice (e.g. for using egui where there is no GPU).

I wrote a software renderer for Dear ImGui a few years ago (in C++) which could probably be used as a starting point: https://github.com/emilk/imgui_software_renderer

tinaun:
I wrote a proof of concept using tiny-skia - https://gist.github.com/tinaun/f2177e6c70f5c6602ff9863e4f05e9e4 

the approach I'm using for the text rendering is giving pretty bad results though - and I don't know if the api it presents is a good match for drawing vertex colored meshes (I haven't implemented Shape::Mesh rendering yet)

coderedart:
won't including the images bloat the git repo or docs.rs ? or will there be some kind of git lfs setup?

AlexApps99:
We can probably checksum them and just compare checksums

coderedart:
Git already does the hashing, but the docs.rs has a size limit. And if we go through different images fast enough, git repo starts increasing in size too.

 there's oxipng to compress images, but idk about how high the CPU usage is.

emilk:
Please try to keep to the issue topic! How to store images for the docs is an interesting problem, but can be discussed in Discussions or on Discord!

emilk:
@tinaun That's a very impressive start! Perhaps the text rendering just needs a half-pixel offset? It's a shame `tiny-skia` doesn't have the ability to render meshes built-in, but perhaps just those parts could be copied from https://github.com/emilk/imgui_software_renderer

How is the performance?

tinaun:
at least on my MacBook, the example in my demo runs in about 1.2 ms, with larger examples like the demo ui taking 6-10 ms per call - I don't have any super old hardware to test on but because tiny-skia is single core performance bound most processors above one ghz should be able to reach 30fps, hopefully.

moving the text bounds did improve things a little (https://i.imgur.com/rbI8u3g.png) - I think I'm running into Srgb/gamma issues since the text feels less vibrant than the live demos

for mesh rendering I'm thinking of looking at [euc](https://docs.rs/euc)

[...]

PR #5506, opened 2024-12-20
Extend `WgpuSetup`, `egui_kittest` now prefers software rasterizers for testing

Wumpf:
* Related to https://github.com/emilk/egui/issues/1129
* Needed for..
     * https://github.com/rerun-io/rerun/issues/8245
     * https://github.com/rerun-io/rerun/issues/8475

`WgpuSetup` is now a lot more powerful:
* fully configure instance
* manual adapter selection on native

`egui_kittest` now uses this to prefer software rasterizers when possible.

Sidenote: `request_adapter`'s `force_fallback_adapter` is not exposed since it will _fail_ if there's no software rasterizer around (which is unfortunately the regular state of things on Mac). Since it also doesn't have any effect on WebGPU (unlike power preference!), I removed it again from `WgpuSetup` as it doesn't seem to be all that useful in general to me 🤷 

TODO:
* [x] remove force_fallback again
* [x] green ci
* [x] expose more `InstanceDescriptor` things, like instance flags / validation options!
* [x] Figure out interaction with  https://github.com/emilk/egui/pull/5539#issue-comment-box
* [x] use this in Rerun to make sure I'm not missing something

### DIFF
--- a/crates/egui-wgpu/src/lib.rs
+++ b/crates/egui-wgpu/src/lib.rs
@@ -23,8 +23,10 @@ pub use wgpu;
 /// Low-level painting of [`egui`](https://github.com/emilk/egui) on [`wgpu`].
 mod renderer;
 
+mod setup;
+
 pub use renderer::*;
-use wgpu::{Adapter, Device, Instance, Queue, TextureFormat};
+pub use setup::{NativeAdapterSelectorMethod, WgpuSetup, WgpuSetupCreateNew};
 
 /// Helpers for capturing screenshots of the UI.
 pub mod capture;
@@ -85,7 +87,7 @@ pub struct RenderState {
 }
 
 async fn request_adapter(
-    instance: &Instance,
+    instance: &wgpu::Instance,
     power_preference: wgpu::PowerPreference,
     compatible_surface: Option<&wgpu::Surface<'_>>,
     _available_adapters: &[Arc<wgpu::Adapter>],
@@ -236,7 +238,7 @@ impl RenderState {
         let surface_formats = {
             profiling::scope!("get_capabilities");
             compatible_surface.map_or_else(
-                || vec![TextureFormat::Rgba8Unorm],
+                || vec![wgpu::TextureFormat::Rgba8Unorm],
                 |s| s.get_capabilities(&adapter).formats,
             )
         };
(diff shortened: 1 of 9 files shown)


lucasmerlin:
I refactored how the snapshots are rendered in #5539, introducing a new `TestRenderer` trait and a `WgpuTestRenderer`.
The renderer can now be initialized when the Harness is created, meaning we can now create a Harness from an eframe::App, tell it to create a wgpu renderer. Then custom rendering relying on `eframe::Frame::wgpu_render_state` and `PaintCallback` will just work ✨

Wumpf:
My attempts to try out software rasterizing on Dx12 ran into a recent wgpu regression, took me quite a while
* https://github.com/gfx-rs/wgpu/pull/6843

crates/egui-wgpu/src/lib.rs
```diff
@@ -100,83 +170,60 @@ impl RenderState {
 
         // This is always an empty list on web.
         #[cfg(not(target_arch = "wasm32"))]
-        let available_adapters = instance.enumerate_adapters(wgpu::Backends::all());
+        let available_adapters = {
+            let backends = if let WgpuSetup::CreateNew(create_new) = &config.wgpu_setup {
+                create_new.instance_descriptor.backends
+            } else {
+                wgpu::Backends::all()
```
Wumpf, line 177: I'm a lil bit uneasy about this actually. Requesting adapters for all backends when the user already settled on one has the potential to waste a lot of startup time. But I didn't want to break the list of adapters we provide.
Should reconsider this in the future.

lucasmerlin:
I gave this a try on windows to see if the snapshot tests would work, they all seem fine, except for the rendering test, where the snapshot fails in the linear gradient section (but only at a scale of 1.0, for some reason):

https://github.com/user-attachments/assets/e59be0b3-1561-49eb-80ba-a3ad4ca63913

I had to increase the threshold to 15.0 for the test to succeed. But this test is also failing on master, so it's not related to the software rasterizer.

Also, logging the AdapterInfo it is choosing vulkan over dx12, meaning it doesn't use the dx12 software rasterizer. I guess this would be fixed by https://github.com/gfx-rs/wgpu/pull/6843?
```
[crates\egui-wgpu\src\lib.rs:238:9] adapter.get_info() = AdapterInfo {
    name: "AMD Radeon RX 5700 XT",
    vendor: 4098,
    device: 29471,
    device_type: DiscreteGpu,
    driver: "AMD proprietary driver",
    driver_info: "24.10.1 (AMD proprietary shader compiler)",
    backend: Vulkan,
}
```

Wumpf:
> Also, logging the AdapterInfo it is choosing vulkan over dx12, meaning it doesn't use the dx12 software rasterizer. I guess this would be fixed by https://github.com/gfx-rs/wgpu/pull/6843?

Yep. That's fixed there, tried that locally :)
Note that generally (not using that custom selection there) wgpu prefers Vulkan over DX12 because the DX12 implementation has some shortcomings today (none of which are super relevant for our purposes here :))

lucasmerlin (approved):
Looks great!

Issue #5424, opened 2024-12-03
Copy image to clipboard buffer


emilk:
I would like to be able to get an `epaint::ColorImage` into the system clipboard, both on native and web, via egui and eframe.

### egui
Currently what is being copied is stored in `PlatformOutput::copied_text`.

The simple/hacky solution would be to simply add a `copied_image: Option<ColorImage>` to `PlatformOutput`.

A nicer solution is to deprecate `copied_text` and `open_url` and replace them with something like `commands: Vec<OutputCommand>` with

```rs
enum OutputCommand {
	CopyText(String),
	CopyImage(ColorImage),
	OpenUrl(OpenUrl)
}
```

### eframe
On native we can use `arboard::Clipboard`.

On web we can use the `Clipboard` API:

```js
const item = new ClipboardItem({ "image/png": png_blob });
await navigator.clipboard.write([item]);
```

## Related
* The opposite issue: https://github.com/emilk/egui/issues/2108
* More general issue: https://github.com/emilk/egui/issues/3259

PR #5532, opened 2024-12-29
Add `OutputCommand` for copying text and opening URL:s

emilk:
Add `OutputCommand` for copying text and opening URL:s

* Part of https://github.com/emilk/egui/issues/5424
* Adds `egui::OutputComm
* Part of https://github.com/emilk/egui/issues/5424and`
* Adds `PlatformOutput::commands`
* Deprecates `PlatformOutput::open_url`
* Deprecates `PlatformOutput::copied_text`

### DIFF
--- a/crates/eframe/src/web/app_runner.rs
+++ b/crates/eframe/src/web/app_runner.rs
@@ -292,12 +292,15 @@ impl AppRunner {
     }
 
     fn handle_platform_output(&self, platform_output: egui::PlatformOutput) {
+        #![allow(deprecated)]
+
         #[cfg(feature = "web_screen_reader")]
         if self.egui_ctx.options(|o| o.screen_reader) {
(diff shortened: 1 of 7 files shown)


Wumpf:
@emilk this makes the usage in Rerun a bit tricky here:
https://github.com/rerun-io/rerun/blob/97f816d5e782651b6f3047d35f375b4eb1a305db/crates/viewer/re_viewer/src/app_state.rs#L661

We intercept all clicked urls and check whether they should redirect to an entity. Not sure how we can preserve this functionality. Seems like this would require installing a sort of user provided command handler :/

(There's also a different spot where we enforce all links to be opened in tabs, but that seems less important)

emilk:
@Wumpf instead of inpspecting/mutating `PlatformOutput::open_url`, just inspect/mutate `PlatformOutput::commands`?

PR #5534, opened 2024-12-29
Add `WidgetType::Image` and `Image::alt_text`

lucasmerlin:
This adds `WidgetType::Image` and correctly sets it in the Image widget. This allows us to query for images in kittest tests and tells accesskit that a node is an image. 
It also adds `Image::alt_text` to set a text that will be shown if the image fails to load and will be read via screen readers. This also allows us to query images by label in kittest. 

### DIFF
--- a/crates/egui/src/widgets/image.rs
+++ b/crates/egui/src/widgets/image.rs
@@ -1,7 +1,10 @@
 use std::{borrow::Cow, sync::Arc, time::Duration};
 
 use emath::{Float as _, Rot2};
-use epaint::RectShape;
+use epaint::{
+    text::{LayoutJob, TextFormat, TextWrapping},
+    RectShape,
+};
 
 use crate::{
     load::{Bytes, SizeHint, SizedTexture, TextureLoadResult, TexturePoll},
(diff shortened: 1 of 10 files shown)


crates/egui/src/widgets/image.rs
```diff
@@ -614,12 +635,36 @@ pub fn paint_texture_load_result(
...
+            job.append(
                 "⚠",
-                font_id,
-                ui.visuals().error_fg_color,
+                0.0,
+                TextFormat {
+                    color: ui.visuals().error_fg_color,
+                    font_id: font_id.clone(),
+                    ..Default::default()
+                },
```
emilk, line 650:
```suggestion
                TextFormat::simple(font_id.clone(), ui.visuals().error_fg_color)
```

crates/egui/src/widgets/image.rs
```diff
@@ -614,12 +635,36 @@ pub fn paint_texture_load_result(
...
+            );
+            if let Some(alt) = alt {
+                job.append(
+                    alt,
+                    ui.spacing().item_spacing.x,
+                    TextFormat {
+                        color: ui.visuals().text_color(),
+                        font_id,
+                        ..Default::default()
+                    },
```
emilk, line 660:
```suggestion
                    TextFormat::simple(font_id, ui.visuals().text_color())
```

emilk (approved):
Very nice!

Issue #5389, opened 2024-11-20
Add `eframe_kittest` to test a whole `eframe::App`


emilk:
For screenshot tests we also should fill in the `wgpu_render_state` (for Rerun screenshot testing of 3D stuff)

PR #5539, opened 2024-12-29
Add `Harness::new_eframe` and `TestRenderer` trait

lucasmerlin:
This allows creating `egui_kittest::Harness`es from `eframe::App`s.
I also introduce a new `TestRenderer` trait and a `WgpuTestRenderer`. I really didn't like how tightly egui_kittest was coupled to wgpu. This should make it possible to create e.g. a `GlowTestRenderer` in oder to support custom glow rendering with PaintCallbacks.
`HarnessBuilder` now has an option to initialize a WgpuTestRenderer, either from a default configuration (this should be updated based on <https://github.com/emilk/egui/issues/5506>) or a WgpuSetup or existing RenderState. 

I also introduce a `LazyTestRenderer` which basically has the old behavior. In the uninitialized state it remembers the TextureDeltas and a closure to create the actual test renderer. When rendering a snapshot, it will actually initialize the wgpu renderer and apply the texture deltas. This means tests without any snapshots will save the overhead of having to create the wgpu resources and tests with snapshots will continue to work fine without having to configure wgpu in the beginning. 

I also renamed `Harness::wgpu_snapshot` etc... to `Harness::snapshot`.

* Closes <https://github.com/emilk/egui/issues/5389>


lucasmerlin:
Tests should pass once #5542 is merged and this is rebased

crates/egui_kittest/src/renderer.rs
```diff
@@ -0,0 +1,81 @@
...
+    fn handle_delta(&mut self, delta: &TexturesDelta);
+
+    /// Render the [`crate::Harness`] and return the resulting image.
+    ///
+    /// # Errors
+    /// Returns an error if the rendering fails.
+    fn render(&mut self, ctx: &Context, output: &FullOutput) -> Result<RgbaImage, String>;
+}
+
+/// A lazy renderer that initializes the renderer on the first render call.
```
Wumpf, line 19: Would be _even_ better if we wouldn't need to initialize the renderer for each TestHarness individually :thinking:
(But that might get hairy quickly and a user can always bring their own renderer for that :))

crates/egui_kittest/src/renderer.rs
```diff
@@ -0,0 +1,81 @@
...
+    fn default() -> Self {
+        #[cfg(feature = "wgpu")]
+        return Self::new(crate::wgpu::WgpuTestRenderer::new);
+        #[cfg(not(feature = "wgpu"))]
+        return Self::Uninitialized {
+            texture_ops: Vec::new(),
+            builder: None,
+        };
+    }
+}
```
Wumpf, line 42: Does this default really make sense? I would have expected the default to be always uninitialized because after all it's supposed to be lazy

crates/egui_kittest/src/renderer.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl LazyRenderer {
+    pub fn new<T: TestRenderer + 'static>(renderer: impl FnOnce() -> T + 'static) -> Self {
```
Wumpf, line 45: nit:

```suggestion
    pub fn new<T: TestRenderer + 'static>(create_renderer: impl FnOnce() -> T + 'static) -> Self {
```

Wumpf (approved):
awesome! Let's put in this pr before https://github.com/emilk/egui/pull/5506 which then becomes more limited in scope :)

crates/eframe/src/epi.rs
```diff
@@ -617,7 +639,8 @@ pub struct Frame {
 
     /// Can be used to manage GPU resources for custom rendering with WGPU using [`egui::PaintCallback`]s.
     #[cfg(feature = "wgpu")]
-    pub(crate) wgpu_render_state: Option<egui_wgpu::RenderState>,
+    #[doc(hidden)]
```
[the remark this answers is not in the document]
Wumpf: ah, right this is already exposed via a getter and that's preferred here

crates/egui_kittest/src/renderer.rs
Wumpf: misread, it _is_ uninitialized

PR #5540, opened 2024-12-29
Keep `copied_text`

rustbasic:
It's better to keep `copied_text` without deprecating it.
because there are cases where its content needs to be immediately retrieved and modified.


emilk:
Use `PlatformOutput::commands` instead

rustbasic:
> Use `PlatformOutput::commands` instead

My opinion is that it would be better to keep the current `copied_text` without deprecating it, rather than creating a new function to identify or modify the content on the clipboard.

PR #5542, opened 2024-12-29
Fix cargo test --all-features breaking rendering due to unity vertexes

### LINKED ISSUE #5297: Snapshot tests fail with`unity` feature

lucasmerlin:
* Closes #5297 

It's not great but I wasn't able to come up with a better solution.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -238,12 +238,10 @@ jobs:
         uses: Swatinem/rust-cache@v2
 
       - name: Run tests
-        # TODO(lucasmerlin): Enable --all-features (currently this breaks the rendering in the tests because of the `unity` feature)
-        run: cargo test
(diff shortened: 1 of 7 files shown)


emilk (approved):
Works for me, or we just remove all unity-related features and specify the unity-order as the default egui order. I don't remember why I didn't do that originally when the unity-PR was submitted

lucasmerlin:
That would mean all integrations relying on the current order would need to update their shaders right? I guess it would be fine if we properly announce it but I think I prefer this solution.

Issue #5519, opened 2024-12-26
Noninteractive tooltip with interactive content (on_hover_ui)


PPakalns:
When hover (tooltip) contains interactive widgets then mouse pointer can interact with the content of the tooltip, tooltip doesn't close when mouse leaves the original item that was hovered.

This causes issue when tooltip is used as only preview and shouldn't be interactable.

To get old behavior back the following workaround can be used:
```rs
if response.hovered() {
    response.show_tooltip_ui(|ui| {
        //  ....
    });
}
```

But this doesn't function very well because there is no delay before raising the tooltip. There is no handling of time when the scroll area was last scrolled etc. These features that are missing in workaround makes tooltip not flicker as much.

**Describe the solution you'd like**
Add a second on_hover_ui method or flag that can be used with tooltip that has interactive content, but shouldn't be interactable (when mouse leaves original item, tooltip closes instead of allowing to interact with the content of the tooltip)

PPakalns:
Previous discussion:
https://github.com/emilk/egui/discussions/4793

PR #5543, opened 2024-12-29
Support on hover tooltip that is noninteractable even with interactable content

PPakalns:
* Closes https://github.com/emilk/egui/issues/5519

### DIFF
--- a/crates/egui/src/containers/tooltip.rs
+++ b/crates/egui/src/containers/tooltip.rs
@@ -72,7 +72,7 @@ impl Tooltip<'_> {
         let mut tooltip = Self::for_widget(response);
         tooltip.popup = tooltip
             .popup
-            .open(response.enabled() && Self::should_show_tooltip(response));
+            .open(response.enabled() && Self::should_show_tooltip(response, true));
         tooltip
     }
 
@@ -81,7 +81,7 @@ impl Tooltip<'_> {
         let mut tooltip = Self::for_widget(response);
         tooltip.popup = tooltip
             .popup
-            .open(!response.enabled() && Self::should_show_tooltip(response));
+            .open(!response.enabled() && Self::should_show_tooltip(response, true));
         tooltip
     }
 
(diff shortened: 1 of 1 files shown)


PPakalns:
Tested in real use case. Greatly reduces flickering for "preview" tooltips that may contain interactable widgets.

Use case: "Preview" of the page/window that will be opened if the button is pressed.

grtlr:
@emilk I ran into a related problem: https://github.com/rerun-io/rerun/issues/8568

It seems like the current logic in `should_show_hover_ui` does not take into account that the tooltip is drawn in relation to the pointer position. Because of this, the tooltip always remains open since it's shifted at every frame.

In the case of the Rerun graph view, we can simply call `on_hover_ui` instead of `on_hover_ui_at_pointer`.

PPakalns:
Resolved conflicts.

PPakalns:
Updated against newest master to rerun checks that failed due to incorrectly configured github pipeline action.

PPakalns:
> I appreciate the need for this functionality, but I think it's time to make a builder for tooltips so that we don't get this combinatorial explosion of tooltip methods

 Can implement this using builder pattern.

To make sure that I decide the right design for this, feel free to correct me:

Currently we have methods:
```
on_hover_ui
on_hover_ui_at_pointer
on_disabled_hover_ui
```

And this PR wants to add parameter to configure if hover (tooltip) ui is interactable.

I could replace these methods with one simple method:

```
on_hover_ui(HoverBuilder)
```

Where HoverBuilder has configurable options:
1) Show at pointer
2) Show if UI is enabled / disabled
3) Should UI content be kept open if mouse moves over it (Is it interactable, can be hovered over)

Does this design seems OK?

lucasmerlin:
@PPakalns we now have a unified `Tooltip` struct, so now it should be straightforward to add as an option there! See #5713 for more info

PPakalns:
@lucasmerlin 

PR is ready for review!

Refactored code. Removed duplicated `on_hover_ui` family of functions.
If users want their own custom logic without tooltip interactions. They can construct `egui::Tooltip` as it is done in `egui::Tooltip::for_enabled` method and then modify logic for their needs.

lucasmerlin (approved):
Thanks! I guess it's not very convenient to use this, but since it handles an edge case it should be fine.

Issue #5546, opened 2024-12-30
RTL layout breaks with TextEdit


zakarumych:
**Describe the bug**

**To Reproduce**
1. Set RTL layout.
2. Make a TextEdit widget
3. When extra size is needed, UI::allocate_rect is called and UI ends up with cursor pointing to the wrong side of TextEdit widget
4. Next widget overlaps with TextEdit.
5. This happens for just 1 frame unless specific sizes introduce rounding error, leading to extra_size to be non-zero each frame.

**Expected behavior**
Cursor is pointing to the left side of TextEdit in RTL layout.

**Screenshots**

![image](https://github.com/user-attachments/assets/31fef836-fc68-4b84-97f4-dc625014270c)

**Desktop (please complete the following information):**
 - OS: Any

PR #5547, opened 2024-12-30
Allocate TextEdit's extra size on the side depending on the layout direction

zakarumych:
TextEdit sometimes allocates extra space if text changes and no longer fits into previously allocated Rect.
It also can happen when rounding error occurs. In this case extra space is extra small.

If extra space is needed, `Ui::allocate_rect` is used to carve that space from placer.
However it always allocated extra space on the bottom-right from the rect placer returned initially.
This works well on Left-to-Right and Top-to-bottom layouts.
However in case it's Right-to-Left, `Ui::allocate_rect` updates placer's cursor and puts it to the right side of the initially allocated rect. And next widget is then placed over the TextEdit.

This is hard to notice usually.
However I have scaling that can cause rounding errors, so 0.0003 units of extra space is allocated each frame. Making next widget to overlap the TextEdit almost entirely.

This changes makes extra space allocation aware of the layout main direction and it picks most appropriate side of the widget for the extra space.

* Closes <https://github.com/emilk/egui/issues/5546>

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -727,10 +727,32 @@ impl<'t> TextEdit<'t> {
             // if there's a ScrollArea, it can properly scroll to the cursor.
             let extra_size = galley.size() - rect.size();
             if extra_size.x > 0.0 || extra_size.y > 0.0 {
-                ui.allocate_rect(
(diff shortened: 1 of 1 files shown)


lucasmerlin:
@zakarumych unfortunately there is a merge conflict now 🙁 can you update your branch?

PR #5553, opened 2024-12-31
`copied_text` function maintained in `copy_text()` until deletion

rustbasic:
My opinion is that it would be better to keep the current `copied_text` without deprecating it, rather than creating a new function to identify or modify the content on the clipboard.

at least until `copied_text` is removed, the `copy_text()` function should continue to use `copied_text` in the same way it currently does.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1230,7 +1230,7 @@ egui_extras::install_image_loaders(egui_ctx);
 * [Tweaked the default visuals style](https://github.com/emilk/egui/pull/450).
 * Plot: Renamed `Curve` to `Line`.
 * `TopPanel::top` is now `TopBottomPanel::top`.
-* `SidePanel::left` no longest takes the default width by argument, but by a builder call.
+* `SidePanel::left` no longer takes the default width by argument, but by a builder call.
 * `SidePanel::left` is resizable by default.
 
 ### 🐛 Fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1230,7 +1230,7 @@ egui_extras::install_image_loaders(egui_ctx);
 * [Tweaked the default visuals style](https://github.com/emilk/egui/pull/450).
 * Plot: Renamed `Curve` to `Line`.
 * `TopPanel::top` is now `TopBottomPanel::top`.
-* `SidePanel::left` no longet takes the default width by argument, but by a builder call.
+* `SidePanel::left` no longest takes the default width by argument, but by a builder call.
```
YgorSouza, line 1233: This should be `longer`.

emilk:
why?

rustbasic:
> why?

`copied_text` is needed in the following cases:

1. When checking characters copied to the clipboard.
2. When changing characters such as ' ', '\t', '\r', '\n'.
3. When filtering characters copied to the clipboard (deleting, prohibiting, changing).
4. When encrypting and restoring the clipboard for security purposes."

emilk:
But why not iterate over `PlatformOutput::commands` and do the same logic there?

rustbasic:
> But why not iterate over `PlatformOutput::commands` and do the same logic there?

In typical scenarios, we use `PlatformOutput::commands`. However, there are cases where we need to immediately know or modify the content copied from `TextEdit`. Also, the current `PlatformOutput::commands` functionality doesn't seem to have a feature to retrieve the content. Even if a feature to retrieve the content existed, it wouldn't address situations where immediate modification is required.

Issue #3494, opened 2023-10-22
Incorrect maximize behaviour


JumpyLionnn:
**Describe the bug**
There are a few problem with maximizing a native window in `eframe`.
1. Every frame `eframe` provides a `Frame` struct and inside it there is an `IntegrationInfo` struct that you can retrieve with the `info()` method. In the integration info, there is some data about the native window including a boolean that indicates if the window is maximized or not. this boolean doesn't change to true even if the window is maximized. 
2. With the persist window option in `NativeOptions` Ive noticed when I start up the app with maximized set to true it does start that way, but instantly shrinks to the previous session's dimensions and if it was closed while maximized it does try to do the dimensions of a maximized window but not actually maximized and it becomes slightly offset.

**To Reproduce**
I dont know if it matters but here are my egui and eframe in my Cargo.toml
```toml
eframe = {version="0.23.0", features = ["persistence"]}
egui = {version="0.23.0", features = ["persistence", "serde"]}
```
1. Set `presist_window` to true
3. Run the app
4. Maximize the window
5. Close the window
6. Rerun the the app and the window state wont be persisted correctly

**Expected behavior**
When opening the window, if the `presist_window` is true and the window was maximized in the last session it should be maximized again in the same monitor.
Also the `WindowInfo` struct in the `Frame` provided by eframe should have maximized set to true if the window is maximized.

**Desktop:**
 - OS: Windows

Turun:
probably related to #1517 

Workaround: save the maximized state in you App struct and add a sentinel variable to indicate the first frame (not persisted, default false, set to true on the first frame). Then you can do:

```rust
if self.first_frame {
    self.first_frame = false;
    frame.set_maximized(self.persisted_maximized_state);
}
```

emilk:
Fix coming in https://github.com/emilk/egui/pull/3601

emilk:
Re-opened because of dead-lock: https://github.com/emilk/egui/pull/3612

emilk:
https://github.com/emilk/egui/pull/3831 will fix this for Windows, but Linux and Mac still has the issue

seritools:
Hmm, even without the deadlock and with the Windows fix, I think bug 2 in the issue is not addressed. `persist_window` does not seem to save the maximized state in the ron file at all.

After closing a maximized window, this is what is saved:

```
[...]

PR #5554, opened 2024-12-31
Serialize window maximized state in `WindowSettings`

### LINKED ISSUE #1517: Maximize Window even if initial_window_size and initial_window_position are set
I'd like to set maximized = true on eframe::NativeOptions along with an initial_window_size and initial_window_position with this behavior:

The application starts maximized, but when you restore it, it goes to the initial size and position.

I can't seem to find a way to achieve this behavior.

landaire:
A user of my Windows application reported a papercut where the application restores its size on next load, but does not restore its maximized state. This PR fixes that.

To test, I patched https://github.com/emilk/eframe_template to use my local code since I knew that template saves/restores window data. Testing methodology was to simply `cargo run`, maximize the application, then close the application. `cargo run` again and the application should start maximized.

Closes #1517.

* * This is mostly true, I had difficulties running `./scripts/check.sh` for some reason. Possibly a bad Python version?

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -1824,6 +1824,9 @@ pub fn apply_viewport_builder_to_window(
             let pos = PhysicalPosition::new(pixels_per_point * pos.x, pixels_per_point * pos.y);
             window.set_outer_position(pos);
         }
+        if let Some(maximized) = builder.maximized {
+            window.set_maximized(maximized);
+        }
     }
 }
 
--- a/crates/egui-winit/src/window_settings.rs
+++ b/crates/egui-winit/src/window_settings.rs
@@ -13,6 +13,8 @@ pub struct WindowSettings {
 
     fullscreen: bool,
 
+    maximized: bool,
+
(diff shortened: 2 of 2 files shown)


emilk (approved):
code LGTM

landaire:
Just wanted to note I _think_ this fixes #1517.

emilk (changes requested):
Unfortunately, this leads to a deadlock on Mac when running `cargo run -p custom_window_frame`


See also https://github.com/emilk/egui/pull/3612

landaire:
Will take a look to see if I can resolve the deadlock.

landaire:
Left a comment in #3494 and also created https://github.com/rust-windowing/winit/issues/4071.

For now I've removed the erroneous unconditional `maximized` logic in `update_viewport_info`. This retains the desired behavior of this diff, while preventing the logic from being invoked on macOS (tested with both `cargo run -p custom_window_frame` and my original testing steps)

emilk:
I can confirm that `custom_window_frame` no longer deadlocks on my Mac - thanks!

PR #5556, opened 2025-01-01
Use bitfield instead of bools in `Response` and `Sense`

polwel:
Closes <https://github.com/emilk/egui/issues/3862>.

Factoring the `bool` members of `Response` into a bitfield, the size of `Response` is now 96 bytes (down from 104).

I gave `Sense` the same treatment, however this has no effects on `Response` due to padding. I've decided not to pursue `PointerState`, as it is quite large (_many_ members that are sized and aligned to multiples of 8 bytes), so I don't expect any noticeable benefit from making  handful of `bool`s slightly leaner.

In any case, the changes to `Sense` are already quite a bit more intrusive than those to `Response`.
The previous implementation overloaded the names of the attributes `click` and `drag` with similarly named methods that _construct_ `Sense` with the corresponding flag set. Now, that the attributes can no longer be accessed directly, I had to introduce methods with new names (`senses_click()`, `senses_drag()` and `is_focusable()`). I don't think this is the cleanest solution: the old methods are essentially redundant now that the named constants like `Sense::CLICK` exist. I did however not want to needlessly break backwards compatibility.
I am happy to revert it (or go further 🙂) if there are concerns.

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -30,7 +30,7 @@ use crate::{
     os::OperatingSystem,
     output::FullOutput,
     pass_state::PassState,
-    resize, scroll_area,
+    resize, response, scroll_area,
     util::IdTypeMap,
     viewport::ViewportClass,
     Align2, CursorIcon, DeferredViewportUiCallback, FontDefinitions, Grid, Id, ImmediateViewport,
@@ -1213,7 +1213,7 @@ impl Context {
     #[deprecated = "Use Response.contains_pointer or Context::read_response instead"]
     pub fn widget_contains_pointer(&self, id: Id) -> bool {
         self.read_response(id)
-            .map_or(false, |response| response.contains_pointer)
+            .map_or(false, |response| response.contains_pointer())
     }
 
     /// Do all interaction for an existing widget, without (re-)registering it.
@@ -1237,26 +1237,21 @@ impl Context {
             rect,
             interact_rect,
             sense,
-            enabled,
-            contains_pointer: false,
-            hovered: false,
-            highlighted,
-            clicked: false,
-            fake_primary_click: false,
-            long_touched: false,
-            drag_started: false,
-            dragged: false,
-            drag_stopped: false,
-            is_pointer_button_down_on: false,
+            flags: response::Flags::empty(),
             interact_pointer_pos: None,
-            changed: false,
             intrinsic_size: None,
         };
+
+        res.flags.set(response::Flags::ENABLED, enabled);
+        res.flags.set(response::Flags::HIGHLIGHTED, highlighted);
 
         self.write(|ctx| {
             let viewport = ctx.viewports.entry(ctx.viewport_id()).or_default();
 
-            res.contains_pointer = viewport.interact_widgets.contains_pointer.contains(&id);
+            res.flags.set(
+                response::Flags::CONTAINS_POINTER,
+                viewport.interact_widgets.contains_pointer.contains(&id),
+            );
 
             let input = &viewport.input;
             let memory = &mut ctx.memory;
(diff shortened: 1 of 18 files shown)


crates/egui/src/response.rs
```diff
@@ -133,6 +67,71 @@ pub struct Response {
     /// for improved layouting.
     /// See for instance [`egui_flex`](https://github.com/lucasmerlin/hello_egui/tree/main/crates/egui_flex).
     pub intrinsic_size: Option<Vec2>,
+
+    #[doc(hidden)]
+    pub(crate) flags: Flags,
```
polwel, line 72: I am slightly confused by the `// IN` and `// OUT` comments (still unsure what they mean), and also that all the fields are `pub`. I made this one `pub(crate)` as I really don't see why users would need to muck with it. I would even have made it so that even `Context` can't access the field directly, and has to go through accessor methods instead of messing with `flags.contains()`, `flags.set()`, etc. But I did not want to go overboard with my 1st attempt of hacking on `egui` 🙂

crates/egui/src/response.rs
```diff
@@ -224,7 +225,10 @@ impl Response {
             let pointer = &i.pointer;
 
             if pointer.any_click() {
-                if self.contains_pointer || self.hovered {
+                if self
+                    .flags
+                    .intersects(Flags::CONTAINS_POINTER | Flags::HOVERED)
+                {
```
emilk, line 231: let's try to keep readability high and not make the code overly long

```suggestion
                if self.contains_pointer() || self.hovered() {
```

emilk (approved):
Look great - thank you!

crates/egui/src/response.rs
polwel: Interesting. So any internal refactoring like this is accepted to be potentially breaking? Wondering if this is sustainable in the long run (if `egui` ever makes to to 1.0 😉). Also, as a new user, I found that the extra API surface doesn't make it easier to learn what I am supposed to touch, and what not. `#[doc(hidden)]` helps somewhat.

crates/egui/src/context.rs
```diff
@@ -1237,61 +1238,72 @@ impl Context {
...
-                res.dragged = Some(id) == viewport.interact_widgets.dragged;
-                res.drag_started = Some(id) == viewport.interact_widgets.drag_started;
-                res.drag_stopped = Some(id) == viewport.interact_widgets.drag_stopped;
+            if res.enabled() {
+                res.flags.set(
+                    response::Flags::HOVERED,
+                    viewport.interact_widgets.hovered.contains(&id),
+                );
+                res.flags.set(
+                    response::Flags::DRAGGED,
```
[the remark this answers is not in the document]
polwel: Cool, I was not sure about your stance on local `use` statements, so I preferred to be explicit.

polwel:
Thanks for the review! I will address your comments shortly.

I see one CI job is failing, but it does not look like something I touched? I can fix it, but should I?

crates/egui/src/context.rs
```diff
@@ -1237,61 +1240,72 @@ impl Context {
...
 
             if enabled
-                && sense.click
+                && sense.senses_click()
                 && memory.has_focus(id)
                 && (input.key_pressed(Key::Space) || input.key_pressed(Key::Enter))
             {
                 // Space/enter works like a primary click for e.g. selected buttons
-                res.fake_primary_click = true;
+                res.flags.set(Flags::FAKE_PRIMARY_CLICKED, true);
```
emilk, line 1268: Nit: these could be shorter by using `+=` or `|=`

```suggestion
                res.flags += Flags::FAKE_PRIMARY_CLICKED;
```

emilk (approved):

PR #5569, opened 2025-01-02
Fix panic due to non-total ordering in `Area::compare_order()`

HactarCE:
[Area::compare_order()](https://github.com/emilk/egui/blob/ee4ab08c8a208f4044a8c571326c9414e7a1c8a6/crates/egui/src/memory/mod.rs#L1174-L1183) is not a total ordering. If three layers A, B, and C have the same `order` field but only A and B are present in `order_map`, then `A==C` and `B==C` but `A!=C`. This can cause a panic in the stdlib sort function, and does in [my app](https://github.com/HactarCE/Hyperspeedcube/tree/v2.0) although it's very difficult to reproduce.

* [x] I have self-reviewed this PR and run `./scripts/check.sh`

### DIFF
--- a/crates/egui/src/memory/mod.rs
+++ b/crates/egui/src/memory/mod.rs
@@ -1175,8 +1175,11 @@ impl Areas {
     ///
     /// May return [`std::cmp::Ordering::Equal`] if the layers are not in the order list.
     pub(crate) fn compare_order(&self, a: LayerId, b: LayerId) -> std::cmp::Ordering {
-        match self.order_map.get(&a).cmp(&self.order_map.get(&b)) {
-            std::cmp::Ordering::Equal => a.order.cmp(&b.order),
+        // Sort by layer `order` first and use `order_map` to resolve disputes.
+        // If `order_map` only contains one layer ID, then the other one will be
+        // lower because `None < Some(x)`.
+        match a.order.cmp(&b.order) {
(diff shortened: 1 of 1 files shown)


crates/egui/src/memory/mod.rs
```diff
@@ -1175,10 +1175,9 @@ impl Areas {
     ///
     /// May return [`std::cmp::Ordering::Equal`] if the layers are not in the order list.
     pub(crate) fn compare_order(&self, a: LayerId, b: LayerId) -> std::cmp::Ordering {
-        if let (Some(a), Some(b)) = (self.order_map.get(&a), self.order_map.get(&b)) {
-            a.cmp(b)
-        } else {
-            a.order.cmp(&b.order)
+        match self.order_map.get(&a).cmp(&self.order_map.get(&b)) {
+            std::cmp::Ordering::Equal => a.order.cmp(&b.order),
+            cmp => cmp,
```
emilk, line 1180: Good catch! It would be nice with a comment here explaining what is going on, and maybe even a unit-test that fails on master, but passes on this PR?

HactarCE: Thanks! I've added a comment, written a regression test, and fixed a bug in my new implementation.

emilk (approved):
Thanks for the unit-test!

PR #5578, opened 2025-01-03
egui_kittest: write `.old.png` files when updating images

emilk:
After running `UPDATE_SNAPSHOTS=1 cargo test --all-features` I want to compare before/after images before committing them. Now I can!


lucasmerlin (approved):
Nice!

PR #5580, opened 2025-01-03
Change `Harness::run` to run until no more repaints are requested

lucasmerlin:
Previously, `Harness::run` just called `Harness::step` 3 times. If that wasn't enough, tests would often call run multiple times so all animations would finish properly. 

Also, I introduced `HarnessBuilder::with_step_dt` to customize with how big of a dt each frame is called. I set the default to 1.0 / 6.0 (~6fps) so we don't waste cpu in tests waiting on animations.

`HarnessBuilder::max_steps` allows us to control how many steps `Harness::run` should run before panicing. 
The default is 6, so we run for up to 1.0 logical seconds (six frames at 6 fps), which should be enough to finish most animations.

Turns out a lot of snapshots where rendered before fully shown and had a light opacity, those are now fixed.


crates/egui_kittest/src/lib.rs
```diff
@@ -60,14 +83,24 @@ impl<'a, State> Harness<'a, State> {
...
+            pixels_per_point,
+            max_steps,
+            step_dt,
+            state: _,
+            mut renderer,
+        } = builder;
         let ctx = ctx.unwrap_or_default();
         ctx.enable_accesskit();
+        // Disable cursor blinking so it doesn't interfere with snapshots
+        ctx.all_styles_mut(|style| style.visuals.text_cursor.blink = false);
```
lucasmerlin, line 97: Without this, the image_viewer test was flakey, I guess because the time before the image loaded varied, meaning the cursor was only shown sometimes.

crates/egui_kittest/src/lib.rs
```diff
@@ -215,22 +252,81 @@ impl<'a, State> Harness<'a, State> {
...
+            self.step();
+            if steps > self.max_steps {
+                return Err(ExceededMaxStepsError {
+                    max_steps: self.max_steps,
+                    repaint_causes: self.ctx.repaint_causes(),
+                });
+            }
+        }
+        Ok(steps)
+    }
```
lucasmerlin, line 324: I was unsure if I should return a Result or Option, I feel Result is more correct but it now means whenever you call try run you have to add something to fix the `must_use` warning.

crates/egui_kittest/src/lib.rs
```diff
@@ -215,22 +252,81 @@ impl<'a, State> Harness<'a, State> {
...
+    /// - [`Harness::run_steps`].
+    pub fn try_run(&mut self) -> Result<u64, ExceededMaxStepsError> {
+        let repaint = Arc::new(Mutex::new(true));
+        let repaint_clone = repaint.clone();
+        self.ctx.set_request_repaint_callback(move |info| {
+            // We only care about immediate repaints
+            if info.delay == Duration::ZERO {
+                *repaint_clone.lock() = true;
+            }
+        });
```
lucasmerlin, line 311: Checking ViewportOutput::repaint_delay would proably also work but the docs say to prefer set_request_repaint_callback, so I did it this way. But I feel like since we only care about immediate repaints, ViewportOutput::repaint_delay should also be fine right?

crates/egui_demo_app/tests/test_demo_app.rs
```diff
@@ -65,7 +65,8 @@ fn test_demo_app() {
             _ => {}
         }
 
-        harness.run();
+        // Can't use Harness::run because fractal clock keeps requesting repaints
+        harness.try_run().ok();
```
emilk, line 69: So why not use `.step()` ?

crates/egui_kittest/src/builder.rs
```diff
@@ -19,6 +21,8 @@ impl<State> Default for HarnessBuilder<State> {
             pixels_per_point: 1.0,
             state: PhantomData,
             renderer: Box::new(LazyRenderer::default()),
+            max_steps: 6,
+            step_dt: 1.0 / 6.0,
```
emilk, line 25: Better to use a value that can be represented exactly by an `f32` imho (I'm not sure `6.0 * (1.0 / 6.0) == 1.0`, and if it is, it's by luck)

```suggestion
            max_steps: 4,
            step_dt: 1.0 / 4.0,
```

crates/egui_demo_lib/src/demo/modals.rs
```diff
@@ -183,12 +183,12 @@ mod tests {
...
         harness.get_by_role(Role::ComboBox).click();
 
-        harness.run();
+        harness.try_run().ok();
         assert!(harness.ctx.memory(|mem| mem.any_popup_open()));
         assert!(harness.state().user_modal_open);
 
         harness.press_key(Key::Escape);
-        harness.run();
+        harness.try_run().ok();
```
emilk, line 191: All these calls to `.try_run().ok()` looks like an anti-pattern. Why not call `.run();` here and elsewhere?

lucasmerlin: .run will panic because the demo will keep requesting repaints to simulate progress when the progress bar is shown. 
`try_run().ok()` basically means "run until stable or until max_steps has been exceeded". Should we add a function that does that but without returning a Result so it's more ergonomic to call? 

What are your thoughts on `Harness::run` panicing when max_steps are exceeded?

emilk: > .run will panic because the demo will keep requesting repaints to simulate progress when the progress bar is shown.

That makes sense - please add it as a comment!

> `try_run().ok()` basically means "run until stable or until max_steps has been exceeded". Should we add a function that does that but without returning a Result so it's more ergonomic to call?

Either that or call `.step();` and have one step be bigger than `Style::animation_time`

> What are your thoughts on `Harness::run` panicing when max_steps are exceeded?

I think this is the right call.

lucasmerlin: I've added `Harness::run_ok` and replaced the all the `try_run().ok()` calls with it.

crates/egui_kittest/src/lib.rs
lucasmerlin: I've added `Harness::run_ok` to resolve this.

crates/egui_kittest/src/lib.rs
emilk: Yes, you can use `ViewportOutput::repaint_delay` instead. The only advantage of `set_request_repaint_callback` is that it works even if the ui is not currently running

emilk (approved):
Wonderful!

emilk (approved):
Wonderful!

crates/egui_kittest/src/lib.rs
lucasmerlin: Updated it to use repaint_delay!

PR #5584, opened 2025-01-04
Avoid allocations for loader cache lookup

mineichen:
[ x ] I have ~~followed~~ _read_ the instructions in the PR template

Unfortunately i had several issues:
 - Some snapshot-tests didn't run successfully on osx. diff shows errors around fonts or missing menu items)
 - cargo clippy doesn't run successfully (egui_kittest cannot find `wgpu` and `image`)
 - ./scripts/check.sh had other issues on my system (env: python: No such file or directory), even if python3 can be called via python in my shell
 
 Is there a system independent, standard way to run these tools (e.g. via Docker?)
 I submit the pr anyway, because there changes are very simple and shouldn't cause issues.

### DIFF
--- a/crates/egui/src/load/texture_loader.rs
+++ b/crates/egui/src/load/texture_loader.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 use super::{
     BytesLoader, Context, HashMap, ImagePoll, Mutex, SizeHint, SizedTexture, TextureHandle,
     TextureLoadResult, TextureLoader, TextureOptions, TexturePoll,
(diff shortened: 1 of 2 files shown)


YgorSouza:
> Is there a system independent, standard way to run these tools (e.g. via Docker?)

I made  #4293 long ago to start converting the scripts to Rust to avoid this kind of problem, but I never got around to converting the rest of them. I'll see if I can get back to it. But the clippy issue seems to be a bug in egui_kittest. It doesn't compile with default features, and that is not caught by the CI which only runs clippy with `--all-features`. There is also the problem that what the CI currently does is different from what contributors are asked to do locally, so the checks can fail locally and pass on the CI and vice versa. Plenty of room for improvement there, but it will take some developer time.

emilk (approved):
Nice

PR #5586, opened 2025-01-06
Handle static RGB WebP images

Aely0:
Current WebP loader assumes all WebP images to be RGBA, which is the case if the image is animated (that's what `image` crate assumes at least). Static images can instead choose to exclude its alpha channel, though it seems to be more of a default choice to include it, even if it's not being utilized. Currently, loading a static RGB WebP image will cause a panic when `ColorImage::from_rgba_unmultiplied` gets called in the loader
```
thread 'main' panicked at /home/aely/.cargo/git/checkouts/egui-226fc7cdd51201c1/f87219d/crates/epaint/src/image.rs:97:9:
  left: 29184
 right: 21888
```

### DIFF
--- a/crates/egui_extras/src/loaders/webp_loader.rs
+++ b/crates/egui_extras/src/loaders/webp_loader.rs
@@ -5,7 +5,7 @@ use egui::{
     mutex::Mutex,
     ColorImage, FrameDurations, Id,
 };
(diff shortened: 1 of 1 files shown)


emilk (approved):
Nice

Issue #5492, opened 2024-12-16
`on_exit` is not called when closing the app on Android when using eframe


Pandicon:
**Describe the bug**

On desktop, the `on_exit` function is called without issues when closing the app. But on Android, it is not called. This is an issue when using persistence, since the `save` function is also not called (this seems to be related as `save` should be called just before `on_exit` according to the docs) and therefore data is lost if for example some settings are changed and then the user closes the app before the next timed `save` function call.

**To Reproduce**
1. Clone the https://github.com/Pandicon/egui-android-example/tree/%60on_exit%60-not-called repository where I set up a basic Android app which showcases this issue
2. Run the app (initialise the cargo mobile project with `cargo mobile init`, connect an Android device, allow USB debugging, and do `cargo mobile run`)
3. Run `adb logcat` to see the logs, you can either somehow filter for a specific app or just redirect it into a text file and search for the relevant logs
4. Let it run for a while - you will see `saving!` being logged every 30 seconds (at least for me).
5. Wait to be like 10 seconds past the last `saving!` log and close the app. Look for a new `saving!` log (this is why one should wait - to be certain if a given log is the timed one) and an `exiting!` log. Neither of them exist, so the app was closed without `save` or `on_exit` being called.

**Expected behavior**
`on_exit` and `save` are called fine when closing the app (ideally `save` would also be called when putting the app into background, i.e. "leaving it" as it may get killed by the OS later, but I am not sure if that would not emit the `on_exit`-triggering event - this would require some testing)

**Smartphone (please complete the following information):**
 - Device: Galaxy Tab S8
 - OS: Android 14, One UI 6.1.1

**Additional context**

When using winit, I was able to listen for an event that indicated that the app was being closed/put into the background, where I could call the `save` function manually. So I feel like it is not a general issue with Android emitting events weirdly or something and that if winit is able to handle it, then eframe should also be.

Pandicon:
I further tried switching from wgpu to glow in eframe and the issue persisted.

Pandicon:
I will take a look at fixing this and will open a PR if I am successful.

PR #5599, opened 2025-01-11
Save state on suspend on Android

Pandicon:
This pull request fixes a subset of #5492 by saving the application state when the `suspended` event is received on Android. This way, even if the user exits the app and closes it manually right after changing some state, it will be saved since `suspended` gets fired when the app is exited. It does not fix the `on_exit` function not being fired - this seems to be a winit bug (the `exiting` function in the winit application handler trait is not called on exit). Once it gets fixed, it may be possible to remove logic introduced by this PR (however, I am not sure how it would handle the app being killed by the system when in the background, that would have to be tested).

I've tested the logic by:
 * Leaving from the app to the home screen, then killing it from the "recent apps" menu
 * Leaving from the app to the "recent apps" menu and killing it
 * Restarting the device while the app was running

In all of these instances, the state was saved (the last one being a pleasant surprise). It was tested on the repository mentioned in #5492 with my forked repository as the source for eframe (I unfortunately am not able to test it in a larger project of mine due to dependence on "3rd party" egui libraries (like egui_notify) which do not compile along with the master branch of eframe (different versions of egui), but I believe it should work in the same manner in all scenarios). Tests were conducted on a Galaxy Tab S8 running Android 14, One UI 6.1.1.

CI passed on my fork.

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -371,7 +371,7 @@ impl<'app> WinitApp for GlowWinitApp<'app> {
         if let Some(running) = self.running.as_mut() {
             profiling::function_scope!();
 
-            // This is because of the "save on suspend" logic on Android. Once the application is suspended, there is no window associated to it, which was causing panics when `.window().expect()` was used.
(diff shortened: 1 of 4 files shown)


Pandicon:
Missed the comment to not open PRs from master branches of forks D: Closing in favour of #5601 which comes from a separate branch in my repository.

PR #5601, opened 2025-01-12
Save state on suspend on Android

Pandicon:
This pull request fixes a subset of #5492 by saving the application state when the `suspended` event is received on Android. This way, even if the user exits the app and closes it manually right after changing some state, it will be saved since `suspended` gets fired when the app is exited. It does not fix the `on_exit` function not being fired - this seems to be a winit bug (the `exiting` function in the winit application handler trait is not called on exit). Once it gets fixed, it may be possible to remove logic introduced by this PR (however, I am not sure how it would handle the app being killed by the system when in the background, that would have to be tested).

I've tested the logic by:
 * Leaving from the app to the home screen, then killing it from the "recent apps" menu
 * Leaving from the app to the "recent apps" menu and killing it
 * Restarting the device while the app was running

In all of these instances, the state was saved (the last one being a pleasant surprise). It was tested on the repository mentioned in #5492 with my forked repository as the source for eframe (I unfortunately am not able to test it in a larger project of mine due to dependence on "3rd party" egui libraries (like egui_notify) which do not compile along with the master branch of eframe (different versions of egui), but I believe it should work in the same manner in all scenarios). Tests were conducted on a Galaxy Tab S8 running Android 14, One UI 6.1.1.

CI passed on my fork.

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -371,7 +371,7 @@ impl<'app> WinitApp for GlowWinitApp<'app> {
         if let Some(running) = self.running.as_mut() {
             profiling::function_scope!();
 
-            // This is because of the "save on suspend" logic on Android. Once the application is suspended, there is no window associated to it, which was causing panics when `.window().expect()` was used.
+            // This is used because of the "save on suspend" logic on Android. Once the application is suspended, there is no window associated to it, which was causing panics when `.window().expect()` was used.
             let window_opt = running.glutin.borrow().window_opt(ViewportId::ROOT);
 
             running.integration.save(
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -366,6 +366,21 @@ impl<'app> WinitApp for GlowWinitApp<'app> {
             .and_then(|r| r.glutin.borrow().window_from_viewport.get(&id).copied())
     }
 
+    fn save(&mut self) {
+        log::debug!("WinitApp::save called");
+        if let Some(running) = self.running.as_mut() {
+            profiling::function_scope!();
+
+            // This is because of the "save on suspend" logic on Android. Once the application is suspended, there is no window associated to it, which was causing panics when `.window().expect()` was used.
+            let window_opt = running.glutin.borrow().window_opt(ViewportId::ROOT);
+
+            running.integration.save(
+                running.app.as_mut(),
+                window_opt.as_deref(),
+            );
+        }
+    }
+
     fn save_and_destroy(&mut self) {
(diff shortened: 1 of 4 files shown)


crates/eframe/src/native/wgpu_integration.rs
```diff
@@ -414,7 +421,12 @@ impl<'app> WinitApp for WgpuWinitApp<'app> {
 
     fn suspended(&mut self, _: &ActiveEventLoop) -> crate::Result<EventResult> {
         #[cfg(target_os = "android")]
-        self.drop_window()?;
+        {
+            log::debug!("Android application suspended - asking to save state");
+            self.drop_window()?;
+            Ok(EventResult::Save)
+        }
```
emilk, line 428: I think it makes sense to return `Save` on all platforms here, not just android. That way we should get the same logic on iOS

crates/eframe/src/native/wgpu_integration.rs
```diff
@@ -488,20 +500,31 @@ impl<'app> WinitApp for WgpuWinitApp<'app> {
...
+    /// Saves the application state. Set `force_save_without_window` to `true` if the state should be saved even if no window is associated to the application (this happens on Android when the `suspended` event is received).
+    fn save(&mut self, force_save_without_window: bool) {
+        let shared = self.shared.borrow();
+        // This is done because of the "save on suspend" logic on Android. Once the application is suspended, there is no window associated to it.
         if let Some(Viewport { window, .. }) = shared.viewports.get(&ViewportId::ROOT) {
+            log::debug!("Saving application state with a window");
             self.integration.save(self.app.as_mut(), window.as_deref());
+        } else if force_save_without_window {
+            log::debug!("Saving application state without a window");
+            self.integration.save(self.app.as_mut(), None);
```
emilk, line 512: Is `force_save_without_window` really necessary? To me it makes more sense that we always call `self.integration.save` whenever `WgpuWinitRunning::save` is called, with or without a window

crates/eframe/src/native/glow_integration.rs
```diff
@@ -413,6 +432,12 @@ impl<'app> WinitApp for GlowWinitApp<'app> {
         if let Some(running) = &mut self.running {
             running.glutin.borrow_mut().on_suspend()?;
         }
+        #[cfg(target_os = "android")]
+        {
+            log::debug!("Android application suspended - asking to save state");
+            Ok(EventResult::Save)
+        }
```
emilk, line 439: Same here: return `Save` on all platforms (i.e. `iOS`)

emilk (approved):
Looks good! Maybe remove or lower some of the logging

Pandicon:
Looked at the comments and implemented the requested things.

emilk:
Needs to resolve merge conflicts

PR #5610, opened 2025-01-15
Upgrade to wgpu 24

torokati44:
See: https://github.com/gfx-rs/wgpu/releases/tag/v24.0.0

This cleans up a few TODOs and duplicated dependencies.

PR #5612, opened 2025-01-16
Wgpu resources are no longer wrapped in `Arc` (since they are now `Clone`)

Wumpf:
* Follow-up to https://github.com/emilk/egui/pull/5610

Wgpu 24 makes most of its interface `Clone`. Which means we no longer have to wrap everything in `Arc`! \o/

### DIFF
--- a/crates/egui_kittest/src/builder.rs
+++ b/crates/egui_kittest/src/builder.rs
@@ -1,5 +1,4 @@
 use crate::app_kind::AppKind;
-use crate::wgpu::WgpuTestRenderer;
 use crate::{Harness, LazyRenderer, TestRenderer};
 use egui::{Pos2, Rect, Vec2};
 use std::marker::PhantomData;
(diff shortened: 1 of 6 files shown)


crates/egui_kittest/src/lib.rs
```diff
@@ -402,6 +402,7 @@ impl<'a, State> Harness<'a, State> {
     ///
     /// # Errors
     /// Returns an error if the rendering fails.
+    #[cfg(any(feature = "wgpu", feature = "snapshot"))]
```
Wumpf, line 405: `bacon` was unhappy before. understandable!

bircni (approved):

crates/egui_kittest/src/builder.rs
```diff
@@ -78,13 +77,13 @@ impl<State> HarnessBuilder<State> {
     /// This sets up a [`WgpuTestRenderer`] with the default setup.
```
bircni, line 77: Please fix this link 
```suggestion
    /// This sets up a [`crate::wgpu::WgpuTestRenderer`] with the default setup.
```

bircni: or reimport it as you removed the import above

bircni (changes requested):

PR #5614, opened 2025-01-17
Only generate a paste command when pasted text is available.

mkrueger:
Not always a text is in the paste buffer.

IMO Cut/Copy/Paste shouldn't be part of the events - or at least have an opt_out. I've made desktop applications that have custom clipboard formats - and potentially can remap command keys.

So handling cut/copy/paste is a problem. (Remapping not really) 
Alternatively the paste system could be enhanced to support multiple formats - but with that fix I can just handle ctrl+v myself.


mkrueger:
Better solution for my use case:
https://github.com/emilk/egui/pull/5618

Issue #4065, opened 2024-02-18
Ctrl+X, C, V events get cut out


mkrueger:
They're replaced by cut/copy/paste commands.

Egui is the wrong layer for that - this is clearly an application feature and not a UI library one. I've an application that needs ctrl+x/c for input. With an event chain both events could be included however I would recommend not providing cut/copy/paste events in the first place. Don't know of any UI library handling that on this level.

Esp. applications may remap key binds that may break it.

YgorSouza:
Maybe it could be an opt-in feature? Apparently these events are only used internally by the text widgets, but I'm guessing a lot of people would like to have them by default and not have to implement them manually.

There's also undo and redo, which are not events but have hardcoded shortcuts in the `TextEdit`:

https://github.com/emilk/egui/blob/ca8c8792c2a6a65656105b971998f937edcb836d/crates/egui/src/widgets/text_edit/builder.rs#L884-L920

Maybe Copy/Cut/Paste could be handled the same way? And eventually with modifiable shortcuts you can set in the Ui's style or something like that.

mkrueger:
It's just some sort of programs that want to implement that manually - as said UI libs don't do that for reasons. This is more limiting in big programs, where you can remap commands.

crumblingstatue:
Didn't know this was a thing until now.
One workaround is doing `ui.input(|inp| inp.events.iter().any(|ev| matches!(ev, egui::Event::Copy)))`, if you're simply interested in `ctrl + c` that egui "stole".

mkrueger:
I know what a workaround is - it's just an unexpected behavior for a UI lib IMO. 
Ctrl+C is an application layer thing IMO. With egui it would be possible to have two events in the chain - may be the better solution for this case.

wareya:
I'm making an image editor and the above-described workaround doesn't work for me, the paste event doesn't fire at all if image data is on the clipboard. I'm using a separate library for clipboard data access so I don't need the data, just the fact that a paste happened. I'm calling `events.iter().any(|ev| matches!(ev, egui::Event::Paste(_)))`.

mkrueger:
... still a thing. 

It's really in the way of making a non trivial program with egui :(.

There is no cut/copy/paste system event - so this is pure nonsene IMO.

YgorSouza:
[...]

PR #5615, opened 2025-01-18
Make the default keyboard shortcuts optional

YgorSouza:
* Closes <https://github.com/emilk/egui/issues/4065>


mkrueger:
I need something like that badly for my tools.

Another possible solution would be:
https://github.com/emilk/egui/pull/5618

This tries to de-activate this by widget. This has the benefit that it's possible to mix text edit widgets + custom cnp widgets

doy:
i would also really like to see this merged, it really makes a lot of different kinds of interfaces impossible to build - is there anything that can be done to move this forward?

PR #5617, opened 2025-01-18
Use Python 3 in `scripts/lint.py`

yerke:

## Changes

- Use Python 3 in `scripts/lint.py`.

## Why

Some modern OS and distributions do not provide Python 2 by default. This includes recent macOS versions.
I see that `scripts/generate_changelog.py` already uses Python 3, so I don't think this change should be controversial.
Without this change, I am unable to run `./scripts/check.sh` on macOS 15.2 without installing Python 2 or adding alias for `python` executable.

### DIFF
--- a/scripts/lint.py
+++ b/scripts/lint.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 """
 Runs custom linting on Rust code.
 """


YgorSouza:
> Without this change, I am unable to run `./scripts/check.sh` on macOS 15.2 without installing Python 2 or adding alias for `python` executable.

The script doesn't actually work on Python 2 though, does it? It uses string interpolation, for example, which was not a thing before 3.6. It just assumes you already have `python` mapped to `python3` by now, which is the case in a lot of systems, but I guess not all of them.

emilk (approved):

PR #5618, opened 2025-01-19
Implement Copy, Cut, and Paste in EventFilter

rustbasic:
Implement Copy, Cut, and Paste in EventFilter

* Related #5540
* Related #5553


mkrueger:
+1 for that :)

PR #5624, opened 2025-01-21
Prepare for `objc2` frameworks v0.3

madsmtm:
The next version of the `objc2` framework crates will have a bunch of default features enabled, see https://github.com/madsmtm/objc2/issues/627, so this PR pre-emptively disables them, so that your compile times down blow up once you upgrade to the next version (which is yet to be released, but will be soon).


emilk (approved):
Thanks!

PR #5626, opened 2025-01-22
Remove references to glium

joshua-holmes:
* Remove references to `glium` backend, because it is deprecated since egui v0.18.0


emilk (approved):
Thanks!

Issue #5246, opened 2024-10-11
Web: scale breaks when moving egui to a monitor with different DPI


lucasmerlin:
**Describe the bug**
I have a monitor with a scale of 1.0 and one with 2.0. When moving a egui window (e.g. https://egui.rs) from the 1.0 screen to the 2.0 screen egui will be shown with twice the scale it should have. If you then reload it will be shown at the correct scale.
Moving egui from the 2.0 screen to the 1.0 screen will cause egui to be shown at half of the expected scale.
The mouse events don't seem to be scaled and don't match the visual output.

https://github.com/user-attachments/assets/15b9ba1a-1447-4cb6-8381-ff6637727c63
 

**To Reproduce**
- go to https://egui.rs
- move the browser window to a monitor with a different scale

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser chrome

This only happens in the browser, when running natively this doesn't happen.

Maybe related to https://github.com/emilk/egui/issues/4960?

rustbasic:
Note : Seems to work fine on Windows10 + Chrome + (Monitor 1.25 & 2.0)
(  Could it be depends on the OS? )

emilk:
The resize observer doesn't trigger on chromium when the Device Pixel Ratio (DPR) changes. Maybe we need to subscribe to some other event for that (e.g. "resize")

PR #5631, opened 2025-01-23
Web: Fix incorrect scale when moving to screen with new DPI

emilk:
* Closes https://github.com/emilk/egui/issues/5246

Tested on
* [x] Chromium
* [x] Firefox
* [x] Safari

On Chromium and Firefox we get one annoying frame with the wrong size, which can mess up the layout of egui apps, but this PR is still a huge improvement, and I don't want to spend more time on this right now.


juancampa:
Glad to hear you fixed this! I hit it quite often and was about to look into it

PR #5634, opened 2025-01-23
Implement Copy, Cut, Paste and others in EventFilter

rustbasic:
Implement Copy, Cut, Paste and others in EventFilter

* Related #5540
* Related #5553 
* Closes #5618

It seems like it would be better to make it work if true and not work if false, even if it's `_focus`, but for now, I'll keep it as is.


mkrueger:
+1 I need that so badly -  you can't imagine.

I've 2 non trivial UIs and need to manage these events on APP level - esp. because I can remap commands in my Apps.
This is a design flaw in egui - it's ok for small apps to let the ui lib handle that but not for big desktop apps. 

One of these is a multi window drawing tool… thanks for working on that. Unfortunately egui has gotten too big and that patch seems to be forgotton :(.

Unfortunately I've a large set of libraries that use egui, otherwise I  would just take your patch to fix my issue. But if it's not included soon I need to fork all libraries and use my fork …

lucasmerlin:
So to understand this correctly, this is needed so you can handle the events globally, but widgets can override this and have their own behavior if focussed? Or is it only needed to tell egui not to handle certain events?

rustbasic:
This is about turning off a specific event in TextEdit.
Users can turn off unwanted events or handle the events that are turned off directly.

crates/egui/src/data/input.rs
```diff
@@ -1198,55 +1198,181 @@ impl From<u32> for TouchId {
 /// Any events not covered by the filter are given to the widget, but are not exclusive.
 #[derive(Clone, Copy, Debug)]
 pub struct EventFilter {
-    /// If `true`, pressing tab will act on the widget,
-    /// and NOT move focus away from the focused widget.
+    /// If `true`, a copy action will be handled by the widget.
```
emilk, line 1201: This would be simpler and more powerful as a callback

emilk:
Closing in favor of #7894, which supersedes this. Thanks for the PR!

PR #5637, opened 2025-01-25
Add overline option for Table rows

akx:
* Closes no issue, I just needed this for an app and figured it could be useful.

This PR adds an `overline` option for `egui_extras::TableRow`, which is useful for visually grouping rows. The overline consumes no layout space.

A screenshot of the demo app, showing every 7th row getting an overline.

Screenshot 2025-01-25 at 14 40 08

### DIFF
--- a/crates/egui_extras/src/layout.rs
+++ b/crates/egui_extras/src/layout.rs
@@ -33,6 +33,7 @@ pub(crate) struct StripLayoutFlags {
     pub(crate) striped: bool,
     pub(crate) hovered: bool,
     pub(crate) selected: bool,
+    pub(crate) overline: bool,
 
     /// Used when we want to accruately measure the size of this cell.
     pub(crate) sizing_pass: bool,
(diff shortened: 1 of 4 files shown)


akx:
Rebased.

crates/egui_extras/src/layout.rs
```diff
@@ -232,6 +233,13 @@ impl<'l> StripLayout<'l> {
             child_ui.style_mut().visuals.override_text_color = Some(stroke_color);
         }
 
+        if flags.overline {
+            child_ui.painter().line_segment(
+                [max_rect.left_top(), max_rect.right_top()],
```
emilk, line 238: nit: can be slightly simpler

```suggestion
            child_ui.painter().hline(
                max_rect.x_range(),
                max_rect.top(),
```

emilk (approved):

emilk:
Sorry for the slow review - needs a merge fix for the screenshot test

akx:
@emilk No worries – thank you for Egui, it's awesome! Rebased + updated snapshot + added your suggestion.

lucasmerlin:
I would expect the line to be centered between rows, but instead it seems to be shifted to the bottom? Is this intentional?

Screenshot 2025-03-25 at 08 51 11

akx:
> I would expect the line to be centered between rows, but instead it seems to be shifted to the bottom? Is this intentional?

It's currently drawn on the zeroth pixel row of the table row it's overlining. I wasn't convinced that drawing it at a half-pixel offset upwards would be a good idea...

Issue #5597, opened 2025-01-11
egui egui::TextEdit::singleline long text will make ui.horizontal invalid


januwA:
**Describe the bug**
![GIF 2025-1-11 13-14-10](https://github.com/user-attachments/assets/f136a9e3-a693-43b0-8672-8c1730c2d7cc)

**To Reproduce**
1. 
2. 
3. 
4. 

**Expected behavior**
The width of singleline has not changed. Horizontal should use the width of singleline instead of the length of the text.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: win11
 - Browser 
 - Version 

**Smartphone (please complete the following information):**
 - Device: 
 - OS: 
 - Browser 
 - Version 

**Additional context**
```sh
 cargo --version
cargo 1.84.0 (66221abde 2024-11-19)
```

```toml
[dependencies]
eframe = { version = "0.30.0", default-features = false, features = ["accesskit", "glow", "persistence"]}
egui_extras = { version = "0.30", features = ["image"] }
```

```rs
ui.horizontal(|ui| {
                    egui::TextEdit::singleline(&mut httpConfig.name)
                        .desired_width(100.0)
                        .show(ui);
                    
                    egui::ComboBox::from_id_source("method")
                        .selected_text(httpConfig.req_cfg.method.as_ref())
                        .show_ui(ui, |ui| {
                            for m in &METHODS {
                                ui.selectable_value(
                                    &mut httpConfig.req_cfg.method,
                                    m.to_owned(),
                                    m.as_ref(),
                                );
                            }
                        });

                    ui.add(
                        egui::TextEdit::singleline(&mut httpConfig.req_cfg.url)
                            .desired_width(300.)
                            .hint_text("http url"),
                    );

                    ui.add(
                        egui::TextEdit::singleline(&mut httpConfig.send_count_ui)
                            .desired_width(60.)
                            .hint_text("Count"),
                    );

                    if ui
                        .add_enabled(!httpConfig.req_cfg.url.is_empty(), egui::Button::new("Send"))
                        .clicked()
                    {
                    }

                    ui.separator();

                    // request result count
                    let (s, e, r) = httpConfig.get_request_reper();
                    ui.label(format!("s:{s}, e:{e}, r:{r}"));

                    ui.separator();
                });
```

MetaflameDragon:
Looks like a duplicate of #5500

IaVashik:
[...]

Issue #5500, opened 2024-12-17
Regression: TextEdit claims full width even with smaller desired_width


crumblingstatue:
~~#5266 makes `TextEdit` have an "intrinsic" size, which I'm guessing makes containers expand to the size it reports as intrinsic size.~~
EDIT: #3660 is the cause, see https://github.com/emilk/egui/issues/5500#issuecomment-2552275067
The problem is that it seems to claim the full width of the text, even when `desired_width` is smaller.
This causes things like side panels to expand way too far for long text, and possibly flicker a lot if the text width changes.

Here is an example of a `SidePanel::right` expanding way too far (due to very long floating point number displayed in the text edit)
`desired_width` is set to `100.0` in this example case.
![image](https://github.com/user-attachments/assets/b03d8b2c-0eb6-4944-9f69-a06f5cc9d102)

crumblingstatue:
If this is intended behavior, then there should be some kind of way to set a max "default width" for a `SidePanel` that it won't automatically expand beyond, unless the user explicitly resizes. Otherwise I don't know how to avoid the flickering without limiting the max width, which also prevents the user from being able to resize beyond the set limit, which I want to allow.

lucasmerlin:
Egui currently does absolutely nothing with the intrinsic size, it's just set so it can be read e.g. in egui_flex, for advanced layouting. So this issue must be something else. Can you try to make a minimal reproduction?

UnknownSuperficialNight:
I made a discussion about this thinking it was intended but after testing I figured I don't think it is.

> Egui currently does absolutely nothing with the intrinsic size, it's just set so it can be read e.g. in egui_flex, for advanced layouting. So this issue must be something else. Can you try to make a minimal reproduction?

Here is my [discussion](https://github.com/emilk/egui/discussions/5502) it has a small example including a video.

oscargus:
Since I prepared a video to file what I think is this bug, here it goes:

https://github.com/user-attachments/assets/1c857694-a6db-407e-a68a-7462fb79d37d

valadaptive:
I have run into this bug too:
https://github.com/user-attachments/assets/76889901-0d73-40ea-aa6f-b3eb86086fb4

For completeness' sake, https://github.com/ruffle-rs/ruffle/issues/19179 appears to also be caused by this bug, but they linked to the PR fixing it so it never got mentioned on this issue.

[...]

PR #5640, opened 2025-01-26
Fix sizing bug in `TextEdit::singleline`

IaVashik:
This PR reverts a change introduced in PR https://github.com/emilk/egui/pull/3660 that caused a regression with `TextEdit::singleline`. The original PR attempted to fix an issue with the cursor in `TextEdit` inside `ScrollArea`, but it did so by adding unnecessary size allocation to `TextEdit`, which breaks the layout when `TextEdit::singleline` is used outside of `ScrollArea`.

Image

The regression introduced by #3660 is more severe, as it completely breaks the layout of applications using `TextEdit::singleline`, as shown in the following issues:

*   Closes https://github.com/emilk/egui/issues/5500
*   Closes https://github.com/emilk/egui/issues/5597

Furthermore, I was unable to reproduce the original bug from PR #3660 in the current version of egui using the following code:

```rust
    fn update(&mut self, ctx: &egui::Context, _: &mut eframe::Frame) {
        ctx.set_debug_on_hover(true);
            ScrollArea::vertical().max_height(100.0).show(ui, |ui| {
                ui.add(TextEdit::multiline(&mut self.text).hint_text("Enter text here..."))
            });
        });
    }
}
```

This code attempts to recreate the layout shown in the video from PR #3660, using a `ScrollArea` with limited height and a `TextEdit` inside. However, the cursor hiding issue was not reproducible.
Video_2025-01-26_17-54-24

Therefore, I believe the code added in PR #3660 is no longer necessary and only creates more problems.

*   Closes https://github.com/emilk/egui/issues/5500
*   Closes https://github.com/emilk/egui/issues/5597
*   [x] I have followed the instructions in the PR template


emilk (approved):
Thanks!

emilk:
There are some clippy failures that need fixing

IaVashik:
Done. There's still a warning left at `crates/egui/src/context.rs:2469:13`, but it's outside the scope of this PR, so I didn't touch it.

juancampa:
The only thing I'm seeing that could be a regression is the frame delay that appears in the video (bottom border disappears for one frame) but I'll test locally to double check

juancampa:
Ok so the border disappearing is because of the TextArea animating the scroll, which is expected. However, I can still repro the original issue when scroll animation is disabled (using `.animated(false)`)

https://github.com/user-attachments/assets/0bc25af4-5e03-4c21-9776-ad56f0460826

Since `.horizontal` comes with its own scrolling behavior a better fix is to wrap the original code in `if !clip_text`. It seems to fix all the issues:

Screenshot 2025-01-27 at 14 33 23

This is what I used:

```rust
            if !clip_text {
                // Allocate additional space if edits were made this frame that changed the size. This is important so that,
                // if there's a ScrollArea, it can properly scroll to the cursor.
                let extra_size = galley.size() - rect.size();
                if extra_size.x > 0.0 || extra_size.y > 0.0 {
                    ui.allocate_rect(
                        Rect::from_min_size(outer_rect.max, extra_size),
                        Sense::hover(),
                    );
                }
            }
```

IaVashik:
So, the root cause of the original issue is `.animated(false)`, that makes sense now. Yeah, using `if !clip_text` would definitely work as a workaround for the bug we're talking about. 
Ngl, after looking at egui's code, I'm still not sure how `animated` actually affects the `ScrollArea` and existing pipeline, so I can't really tell if this behavior is a bug in `ScrollArea` or `TextEdit` itself.

IaVashik:
Okay, that makes sense. I tried to figure out where the calculation goes wrong, but egui is just too complex and complicated for me. My only suspicion is `galley.size()` in `show_content`, but I couldn't understand how `galley` and `layouter` are obtained, so I'll leave this to the smarter contributors of this wonderful project :<

The solution from @juancampa does work, but to me, it feels more like a workaround. I'll revert to their code and add an if statement for `!clip_text`. At the very least, it's the quickest way to fix the bug.

juancampa (approved):

emilk:
There is a failing snapshot test that needs eyes on it. If it makes sense, update it with `UPDATE_SNAPSHOTS=1 cargo test --all-features`

IaVashik:
Hey!  Yeah, the `demo::demo_app_windows::tests::demos_should_match_snapshot` test is failing because of a width difference. It's actually related to the `TextEdit::singleline` bug that this PR fixes, which was messing with the window size. 
Code Example old
Code Example

It's my first time working with snapshot testing, I hope I'll do it right

IaVashik:
I’ve replaced the "Code Example" snapshot with the version from the CI artifact to fix the test failures. However, I need the CI workflow to run so I can verify that everything is correct.
Due to rendering differences between CI and my local machine, I can’t be completely sure that the all snapshots will match perfectly. Running the tests in CI will help confirm that everything is working as expected.

Could you (@emilk or maybe @lucasmerlin?) approve and trigger the workflow when you have time?

lucasmerlin:
Seems like cargo-machete is now failing but that's not your fault 😅 I'll fix that next week

lucasmerlin:
Machete should be fixed by https://github.com/emilk/egui/pull/5745

lucasmerlin:
Thanks again!

PR #5649, opened 2025-01-29
egui_kittest: succeed and keep going when `UPDATE_SNAPSHOTS` is set

emilk:
It used to be that `UPDATE_SNAPSHOTS=true cargo test --all-features` would stop on the first crate with a failure, requiring you to run it multiple times, which is annoying, and a waste of time.

### DIFF
--- a/crates/egui_kittest/src/snapshot.rs
+++ b/crates/egui_kittest/src/snapshot.rs
@@ -155,6 +155,10 @@ impl Display for SnapshotError {
     }
 }
 
+/// If this is set, we update the snapshots (if different),
+/// and _succeed_ the test.
+/// This is so that you can set `UPDATE_SNAPSHOTS=true` and update _all_ tests,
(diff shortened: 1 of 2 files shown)


crates/egui_kittest/src/snapshot.rs
```diff
@@ -155,6 +155,10 @@ impl Display for SnapshotError {
     }
 }
 
+/// If this is set, we update the snapshots (if different),
+/// and _succeed_ the test.
+/// This is so that you can set `UPDATE_SNAPSHOTS=true` and update _all_ tests,
+/// without `cargo test` failing on the first failing crate.
 fn should_update_snapshots() -> bool {
     std::env::var("UPDATE_SNAPSHOTS").is_ok()
```
abey79, line 163: It would be nice if we actually checked for a truth value here. I did `UPDATE_SNAPSHOTS=0` once and it didn't do what I expected.

Issue #5643, opened 2025-01-28
Introduce a `Panel` trait for `SidePanel` and `TopBottomPanel`


sharky98:
**Is your feature request related to a problem? Please describe.**

I am trying to create a library around both `SidePanel` and `TopBottomPanel` so that they have the same behavior when used in a specific way (my specific case is toolbars and properties panels). Having two distinct type makes it hard to create such wrapper that abstract the specific of each of them.

In my use case, the toolbar will have a fixed width/height and will not be resizable. The use of the wrapper allows me to use a pre-made toolbar instead of duplicating options code.

**Describe the solution you'd like**

Seeing that the panels `impl`  have almost the same signature, I think it make sense to create a trait so that  we can create generic wrapper using [`impl trait type`](https://doc.rust-lang.org/reference/types/impl-trait.html)

Also, on a personal opinion, I think the name `TopBottom` and `Side` are a bit confusing at first glance, even more when nesting panels, making it seems that having a `BottomPanel` inside a `TopPanel` inside a `LeftPanel` is not possible. They could be renamed to `HorizontalPanel` (for the `TopBottomPanel`) and `VerticalPanel` (for the `SidePanel`). That way, no matter the nesting they'll always be horizontal or vertical.

The changes would be in [crates/egui/src/containers/panel.rs](https://github.com/emilk/egui/blob/93d214429491dfd628cd7331c9cef96d11a8738d/crates/egui/src/containers/panel.rs)

I can create a pull requests, but chances that the changes I made are not working as-is (excluding the breaking change of the names), me being a noob in Rust 😉 

<details>
<summary>Draft code change proposal</summary>

```rust
//! Panels are [`Ui`] regions taking up e.g. the left side of a [`Ui`] or screen.
//!
//! Panels can either be a child of a [`Ui`] (taking up a portion of the parent)
//! or be top-level (taking up a portion of the whole screen).
//!
//! Together with [`crate::Window`] and [`crate::Area`]:s, top-level panels are
//! the only places where you can put your widgets.
//!
//! The order in which you add panels matter!
//! The first panel you add will always be the outermost, and the last you add will always be the innermost.
//!
//! You must never open one top-level panel from within another panel. Add one panel, then the next.
//!
//! ⚠ Always add any [`CentralPanel`] last.
//!
//! Add your [`crate::Window`]:s after any top-level panels.

use crate::{
[...]

PR #5651, opened 2025-01-30
Introduce traits for SidePanel and TopBottomPanel

sharky98:
* Closes <https://github.com/emilk/egui/issues/5643>

I made the change against 0.30.0 for now to test in my own code if that work like I hoped. I will bring the changes in master afterwards.

⚠️ This would be a breaking change, since use must import the trait to use the methods transfered to the traits.

### DIFF
--- a/crates/egui/src/containers/mod.rs
+++ b/crates/egui/src/containers/mod.rs
@@ -20,7 +20,7 @@ pub use {
     combo_box::*,
     frame::Frame,
     modal::{Modal, ModalResponse},
-    panel::{CentralPanel, PanelOptions, PanelShow, SidePanel, TopBottomPanel},
+    panel::{CentralPanel, Panel, SidePanel, TopBottomPanel},
     popup::*,
     resize::Resize,
     scroll_area::ScrollArea,
--- a/crates/egui/src/containers/mod.rs
+++ b/crates/egui/src/containers/mod.rs
@@ -20,7 +20,7 @@ pub use {
     combo_box::*,
     frame::Frame,
     modal::{Modal, ModalResponse},
-    panel::{CentralPanel, SidePanel, TopBottomPanel},
+    panel::{CentralPanel, PanelOptions, PanelShow, SidePanel, TopBottomPanel},
     popup::*,
     resize::Resize,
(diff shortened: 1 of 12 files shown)


sharky98:
I have some trouble with `./script/check.sh` and the linting. It ask to add `#[inline]` with message "builder methods should be marked #[inline]" on the traits method. But then `cargo doc` says it's an error: "error: `#[inline]` is ignored on function prototypes".

```shellsession
+ ./scripts/lint.py
./crates/egui/src/containers/panel.rs:72: builder methods should be marked #[inline]
./crates/egui/src/containers/panel.rs:77: builder methods should be marked #[inline]
./crates/egui/src/containers/panel.rs:80: builder methods should be marked #[inline]
./crates/egui/src/containers/panel.rs:83: builder methods should be marked #[inline]
./crates/egui/src/containers/panel.rs:86: builder methods should be marked #[inline]
./crates/egui/src/containers/panel.rs:89: builder methods should be marked #[inline]
./crates/egui/src/containers/panel.rs:92: builder methods should be marked #[inline]
./crates/egui/src/containers/panel.rs:95: builder methods should be marked #[inline]
```

```shellsession
+ ./scripts/lint.py
./scripts/lint.py finished without error
+ cargo fmt --all -- --check
+ cargo doc --quiet --lib --no-deps --all-features
[...]
error: `#[inline]` is ignored on function prototypes
  --> crates/egui/src/containers/panel.rs:72:5
   |
72 |     #[inline]
   |     ^^^^^^^^^
   |
   = note: `-D unused-attributes` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(unused_attributes)]`

```

sharky98:
Ran the test locally, I have the following result: `Test egui: 308 total, 7 failed, 7 ignored, 294 passed`. The 7 failed tests seems all to be related to `git-lfs` and `InvalidSignature`, even though it is installed.

sharky98:
Turns out, my idea wasn't as easy to implement that I expected (Rust beginner here 😅). I'll rethink this a bit!

sharky98:
Replacing with PR #5659

PR #5658, opened 2025-01-30
Remove `StrokeKind::default`

emilk:
Since there is no natural default for `RectShape`.


crates/epaint/src/stroke.rs
```diff
@@ -63,29 +63,30 @@ pub enum StrokeKind {
     Inside,
 
     /// The stroke should be painted right on the edge of the shape, half inside and half outside.
-    Middle,
+    Middle, // TODO(emilk): rename to "Center" to match figma?
```
juancampa, line 66: I used photoshop's terminology but either one works

juancampa (commented):
Makes sense to me

PR #5659, opened 2025-01-30
Common panels

sharky98:
* Closes <https://github.com/emilk/egui/issues/5643>

⚠️ This would be a breaking change, since use must import the trait to use the methods transfered to the traits.


sharky98:
Locally, all the tests passed (except those regarding `git lfs` and some images that the delta has nothing to do with my change (more or less text, etc.). The panel demo in `egui_demo_app` works as before. I did not saw anything different.

crates/egui/src/containers/panel.rs
```diff
@@ -166,45 +361,47 @@ impl SidePanel {
         self
     }
 
-    /// The initial wrapping width of the [`SidePanel`], including margins.
+    /// The initial wrapping width of the [`Panel`], including margins.
```
lucasmerlin, line 364: Width is wrong now, and it'd be nice if the comment could explain in which cases this is width and in which this is height

crates/egui/src/containers/panel.rs
```diff
@@ -513,281 +518,91 @@ impl SidePanel {
...
+        let get_ui_kind = || match side {
+            PanelSide::Vertical(v_side) => match v_side {
+                VerticalSide::Left => UiKind::LeftPanel,
+                VerticalSide::Right => UiKind::RightPanel,
+            },
+            PanelSide::Horizontal(h_side) => match h_side {
+                HorizontalSide::Top => UiKind::TopPanel,
+                HorizontalSide::Bottom => UiKind::BottomPanel,
+            },
         };
```
lucasmerlin, line 578: Is there a reason this is a closure?

```suggestion
        let ui_kind = match side {
            PanelSide::Vertical(v_side) => match v_side {
                VerticalSide::Left => UiKind::LeftPanel,
                VerticalSide::Right => UiKind::RightPanel,
            },
            PanelSide::Horizontal(h_side) => match h_side {
                HorizontalSide::Top => UiKind::TopPanel,
                HorizontalSide::Bottom => UiKind::BottomPanel,
            },
        };
```

crates/egui/src/containers/panel.rs
```diff
@@ -50,44 +50,205 @@ impl PanelState {
...
+            }
+        };
+
+        let opposite_horizontal = |side: HorizontalSide| -> PanelSide {
+            match side {
+                HorizontalSide::Top => PanelSide::Horizontal(HorizontalSide::Bottom),
+                HorizontalSide::Bottom => PanelSide::Horizontal(HorizontalSide::Top),
+            }
+        };
+
```
lucasmerlin, line 89: Why the closures? For readability? I'd just match both enums in the match, e.g. `PanelSide::Horizontal(HorizontalSide::Top) => PanelSide::Horizontal(HorizontalSide::Bottom)`

lucasmerlin (commented):
Nice, I think this makes sense. Also, having one struct is definitely the better approch (instead of introducing a trait, as you suggested in your issue)

crates/egui/src/containers/panel.rs
sharky98: At first I did as such (see below), but I had some issues (which were not related to have everything together. Being my first time touching Rust (in fact, first time on a low level language altogether), I segregated stuff to the smallest unit for my own thinking and readability.

And it is also my personal preference to (try to) keep only 1 or 2 indent level.

Same logic is true for all methods and function throughout my PR.

Let me know if you'd prefer I bring back the closure into a "single" operation.

```rust
match self {
  PanelSide::Vertical(side) => match side {
    VerticalSide::Left => PanelSide::Vertical(VerticalSide::Right),
    VerticalSide::Right => PanelSide::Vertical(VerticalSide::Left),
  },
  PanelSide::Horizontal(side) => match side {
     HorizontalSide::Top => PanelSide::Horizontal(HorizontalSide::Bottom),
     HorizontalSide::Bottom => PanelSide::Horizontal(HorizontalSide::Top),
  },
}
```

crates/egui/src/containers/panel.rs
sharky98: I did miss that one!

I did explained the cases at the attribute [`default_size` at line 275](https://github.com/emilk/egui/pull/5659/files/8e67cd207d107100510b690312e8823cf38de2a5#diff-1b1b7f67a664568e1201e4336ca926fa687c506ffaaf2f237a21a1fa4619fab7R275-R277). It might be a bit too deep and hidden though in the docs. I'll check to add some more info in the overal containers doc.

crates/egui/src/containers/panel.rs
sharky98: Not that I can think of. Like above, being totaly new to Rust, I may have overthinked some stuff here and there (although I see that I did put the two indent of match together and not as an external closure 😅, I guess I was being more confortable once I got here!).

Also, I did use RustRover refactoring to extract as a variable in many places and sometime it used a closure, sometime not. But I am not knowledgeable enough in Rust to provide concrete and technical reasoning being those choices.

I'll look into it with all the others.

crates/egui/src/containers/panel.rs
```diff
@@ -50,44 +50,205 @@ impl PanelState {
...
+    Top,
+    Bottom,
+}
+
+/// [`Horizontal`](PanelSide::Horizontal) or [`Vertical`](PanelSide::Vertical)
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PanelSide {
+    Vertical(VerticalSide),
+    Horizontal(HorizontalSide),
+}
```
emilk, line 72: This is another alternative:

```rs
pub enum Side {
    /// Left/top
    Min,

    /// Right/bottom
    Max,
}

impl Side {
    // with appropriate `#[allow` for the casing
    pub const Left: Self = Self::Min;
    pub const Top: Self = Self::Min;
    pub const Right: Self = Self::Max;
    pub const Bottom: Self = Self::Max;
}
```

crates/egui/src/containers/mod.rs
```diff
@@ -1,4 +1,4 @@
-//! Containers are pieces of the UI which wraps other pieces of UI. Examples: [`Window`], [`ScrollArea`], [`Resize`], [`SidePanel`], etc.
+//! Containers are pieces of the UI which wraps other pieces of UI. Examples: [`Window`], [`ScrollArea`], [`Resize`], [`OldSidePanel`], etc.
```
emilk, line 1: Is there an `OldSidePanel`?

crates/egui_demo_app/src/apps/image_viewer.rs
```diff
@@ -52,28 +53,31 @@ impl Default for ImageViewer {
...
-                    ctx.forget_image(&self.current_uri);
-                    self.uri_edit_text = self.uri_edit_text.trim().to_owned();
-                    self.current_uri = self.uri_edit_text.clone();
-                };
-
-                #[cfg(not(target_arch = "wasm32"))]
-                if ui.button("file…").clicked() {
-                    if let Some(path) = rfd::FileDialog::new().pick_file() {
-                        self.uri_edit_text = format!("file://{}", path.display());
+        egui::Panel::new(PanelSide::Horizontal(HorizontalSide::Top), "url bar").show(
```
emilk, line 56: We should use the helpers instead

```suggestion
        egui::Panel::left("url bar").show(
```

crates/egui_demo_app/src/apps/image_viewer.rs
```diff
@@ -52,28 +53,31 @@ impl Default for ImageViewer {
...
                     }
-                }
-            });
-        });
+                });
+            },
+        );
 
-        egui::SidePanel::new(Side::Left, "controls").show(ctx, |ui| {
+        egui::Panel::new(PanelSide::Vertical(VerticalSide::Left), "controls").show(ctx, |ui| {
```
emilk, line 80:
```suggestion
        egui::Panel::left("controls").show(ctx, |ui| {
```

crates/egui/src/containers/mod.rs
```diff
@@ -21,7 +21,7 @@ pub use {
     combo_box::*,
     frame::Frame,
     modal::{Modal, ModalResponse},
-    panel::{CentralPanel, SidePanel, TopBottomPanel},
```
emilk, line 24: We should keep the old `SidePanel` and `TopBottomPanel` as deprecated aliases for `Panel`, to help out users migrate

crates/egui/src/containers/panel.rs
emilk, line 72: It would be nice to have `impl From<VerticalSide> for PanelSide` etc, but 🤷

crates/egui/src/containers/panel.rs
```diff
@@ -50,44 +50,205 @@ impl PanelState {
...
+                VerticalSide::Right => PanelSide::Vertical(VerticalSide::Left),
+            }
+        };
+
+        let opposite_horizontal = |side: HorizontalSide| -> PanelSide {
+            match side {
+                HorizontalSide::Top => PanelSide::Horizontal(HorizontalSide::Bottom),
+                HorizontalSide::Bottom => PanelSide::Horizontal(HorizontalSide::Top),
+            }
+        };
```
emilk, line 88: I think all the above would look a lot more readable as member methods:

`side.opposite()` (for all three side enum types)

crates/egui/src/containers/panel.rs
```diff
@@ -447,26 +467,18 @@ impl SidePanel {
...
-                .map_or(self.default_width, |state| state.rect.width());
-            let fake_width = how_expanded * expanded_width;
-            Self {
-                id: self.id.with("animating_panel"),
-                ..self
-            }
-            .resizable(false)
-            .exact_width(fake_width)
-            .show_inside(ui, |_ui| {});
+            animated_panel.unwrap().show_inside(ui, |_ui| {});
```
emilk, line 477: Please don't do `.unwrap()`. It is easy to avoid:

```rs
let animated_panel = animated_panel.as_ref()?;
```

emilk (commented):
Thanks for the PR! Let's remove the `.unwrap()`s, add `#[deprecated` aliases for backwards compatibility/migration, and then test thoroughly :)

crates/egui/src/containers/panel.rs
```diff
@@ -976,3 +976,38 @@ fn clamp_to_range(x: f32, range: Rangef) -> f32 {
...
+    pub struct TopBottomPanel {}
+
+    impl TopBottomPanel {
+        pub fn top(id: impl Into<Id>) -> Panel {
+            Panel::top(id)
+        }
+        pub fn bottom(id: impl Into<Id>) -> Panel {
+            Panel::bottom(id)
+        }
+    }
```
lucasmerlin: Why not just `pub type SidePanel = Panel;`?
Then it should also work if someone e.g. uses it as a return type or in a struct.

emilk: Good idea!

PR #5661, opened 2025-01-31
egui: Add `collapsed` and `default_collapsed` to Window for programatically collapsing it

### LINKED ISSUE #358: Add methods to collapse a window

EriKWDev:
In summary, this PR:
- introduces `collapsed(&self, collapsed: &mut bool)` to window which can be used to explicitly collapse a window. It will, similar to the `open` function, set the `collapsed` boolean according to the internal state when it is changed by the default collapse button / double click on title and explicitly set it to the provided value if changed
- introduces `default_collapsed` which does what old `default_open` did - sets default collapsed state
- marks `default_open` as deprecated since it was miss-labeled imo, should have been called `default_collapse`. The `open` word in the `Window` is related to the open/close button on the window and got mixed up with the collapsed state. I would also be fine if the `open` function was renamed, though this seemed like the smallest change.
- Closes #358 

This PR does not redesign the window API to handle `events` like #4835, but rather implements it using the same API as the current `open` API  

---

* I have run `./scripts/check.sh` (edit: though I apparently misinterpreted the output. It should show errors in red xD didn't read) and `cargo test` in `crates/egui` as well as used and verified the new API works how we want for our application

### DIFF
--- a/crates/egui/src/containers/collapsing_header.rs
+++ b/crates/egui/src/containers/collapsing_header.rs
@@ -58,12 +58,24 @@ impl CollapsingState {
         })
     }
 
+    pub fn load_with_default_collapsed(ctx: &Context, id: Id, default_collapsed: bool) -> Self {
+        Self::load_with_default_open(ctx, id, !default_collapsed)
+    }
+
     pub fn is_open(&self) -> bool {
         self.state.open
     }
 
     pub fn set_open(&mut self, open: bool) {
         self.state.open = open;
+    }
+
+    pub fn is_collapsed(&self) -> bool {
+        !self.state.open
+    }
+
+    pub fn set_collapsed(&mut self, collapsed: bool) {
+        self.state.open = !collapsed;
     }
 
     pub fn toggle(&mut self, ui: &Ui) {
--- a/crates/egui/src/containers/window.rs
+++ b/crates/egui/src/containers/window.rs
@@ -33,7 +33,7 @@ use super::{area, resize, Area, Frame, Resize, ScrollArea};
 /// Note that this is NOT a native OS window.
 /// To create a new native OS window, use [`crate::Context::show_viewport_deferred`].
 #[must_use = "You should call .show()"]
-pub struct Window<'open> {
+pub struct Window<'open, 'collapsed> {
     title: WidgetText,
     open: Option<&'open mut bool>,
     area: Area,
@@ -41,12 +41,13 @@ pub struct Window<'open> {
     resize: Resize,
     scroll: ScrollArea,
     collapsible: bool,
-    default_open: bool,
     with_title_bar: bool,
     fade_out: bool,
+    default_collapsed: bool,
+    collapsed: Option<&'collapsed mut bool>,
 }
 
-impl<'open> Window<'open> {
+impl<'open, 'collapsed> Window<'open, 'collapsed> {
     /// The window title is used as a unique [`Id`] and must be unique, and should not change.
     /// This is true even if you disable the title bar with `.title_bar(false)`.
     /// If you need a changing title, you must call `window.id(…)` with a fixed id.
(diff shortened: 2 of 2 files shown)


crates/egui/src/containers/collapsing_header.rs
```diff
@@ -66,6 +70,14 @@ impl CollapsingState {
         self.state.open = open;
     }
 
+    pub fn is_collapsed(&self) -> bool {
```
EriKWDev, line 73: Easier to read/write with when the `Window` contained usages of both the word 'open' and 'collapse'

crates/egui/src/containers/window.rs
```diff
@@ -278,9 +292,17 @@ impl<'open> Window<'open> {
     }
 
     /// Set initial collapsed state of the window
+    #[deprecated = "Use `default_collapsed` to set default collapsed state instead. This function might change meaning in future, but for now does the same thing as it used to"]
```
EriKWDev, line 295: The old function was named `default_open` which didn't really align with the usage of `open` to mean whether the window was shown or not. So, I changed to explicitly use `collapsed` for indicating the collapsed state of the window

crates/egui/src/containers/window.rs
```diff
@@ -1213,6 +1240,7 @@ impl TitleBar {
         open: Option<&mut bool>,
         collapsing: &mut CollapsingState,
         collapsible: bool,
+        mut collapsed: Option<&mut bool>,
```
EriKWDev, line 1243: has to be `mut collapsed` due to `if let Some() = ` causes the mutable reference to `move` and I needed to use it twice..

crates/egui/src/containers/window.rs
```diff
@@ -33,20 +33,21 @@ use super::{area, resize, Area, Frame, Resize, ScrollArea};
 /// Note that this is NOT a native OS window.
 /// To create a new native OS window, use [`crate::Context::show_viewport_deferred`].
 #[must_use = "You should call .show()"]
-pub struct Window<'open> {
+pub struct Window<'open, 'collapsed> {
```
EriKWDev, line 36: Could maybe be same lifetimes for simplicitly, though since they can technically be different they should perhaps be modelled like this.

I think any usecase I would have they would have the same lifetimes though

EriKWDev (commented):
Added comments about why mark `default_open` deprecated and more

crates/egui/src/containers/window.rs
```diff
@@ -1233,7 +1261,19 @@ impl TitleBar {
             let button_rect = button_rect.round_to_pixels(ui.pixels_per_point());
 
             ui.allocate_new_ui(UiBuilder::new().max_rect(button_rect), |ui| {
-                collapsing.show_default_button_with_size(ui, button_size);
+                if let Some(collapsed) = collapsed.as_ref() {
+                    if **collapsed != collapsing.is_collapsed() {
```
EriKWDev, line 1265: "in case the user-provided has a different value than the internal 'is_collapsed', change the internal"

tarolling:
this PR still being worked? @EriKWDev

EriKWDev:
I don't remember xD We used to have this fork in our internal game engine but I believe we no longer have a need for it since we use a custom window widget now..

Feel free to port it and or improve it though!

the engine and game: https://idno.se/swap

tarolling:
> I don't remember xD We used to have this fork in our internal game engine but I believe we no longer have a need for it since we use a custom window widget now..
> 
> Feel free to port it and or improve it though!
> 
> the engine and game: https://idno.se/swap

no worries, was just going through older PRs and seeing what was still in the works. thanks!

also game looks great, keep it up👍

PR #5672, opened 2025-02-04
Add `SnapshotResults` struct to egui_kittest

lucasmerlin:
I got annoyed by all the slightly different variations of "collect snapshot results and unwrap them at the end of test" I've written, so I added a struct to make this nice and simple.

One controversial thing: It panics when dropped. I wanted to ensure people cannot forget to unwrap the results at the end, and this was the best thing I could come up with. I don't think this is possible via clippy lint or something like that.

### DIFF
--- a/crates/egui_demo_lib/src/rendering_test.rs
+++ b/crates/egui_demo_lib/src/rendering_test.rs
@@ -688,10 +688,11 @@ fn mul_color_gamma(left: Color32, right: Color32) -> Color32 {
 mod tests {
     use crate::ColorTest;
     use egui_kittest::kittest::Queryable as _;
+    use egui_kittest::SnapshotResults;
 
     #[test]
     pub fn rendering_test() {
-        let mut errors = vec![];
+        let mut results = SnapshotResults::new();
         for dpi in [1.0, 1.25, 1.5, 1.75, 1.6666667, 2.0] {
             let mut color_test = ColorTest::default();
             let mut harness = egui_kittest::Harness::builder()
@@ -708,12 +709,9 @@ mod tests {
 
             harness.fit_contents();
 
-            let result = harness.try_snapshot(&format!("rendering_test/dpi_{dpi:.2}"));
-            if let Err(err) = result {
-                errors.push(err);
-            }
+            results.add(harness.try_snapshot(&format!("rendering_test/dpi_{dpi:.2}")));
         }
 
-        assert!(errors.is_empty(), "Errors: {errors:#?}");
+        results.unwrap();
     }
 }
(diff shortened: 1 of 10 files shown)


crates/egui_demo_lib/src/rendering_test.rs
```diff
@@ -708,12 +709,9 @@ mod tests {
...
 
-            let result = harness.try_snapshot(&format!("rendering_test/dpi_{dpi:.2}"));
-            if let Err(err) = result {
-                errors.push(err);
-            }
+            results.add(harness.try_snapshot(&format!("rendering_test/dpi_{dpi:.2}")));
         }
 
-        assert!(errors.is_empty(), "Errors: {errors:#?}");
+        results.unwrap();
```
lucasmerlin, line 715: If we keep the panic on drop, I could also remove all the uwrap calls, since they aren't needed now

lucasmerlin: I removed them

crates/egui_kittest/src/snapshot.rs
```diff
@@ -483,3 +485,100 @@ impl<State> Harness<'_, State> {
...
+    pub fn into_result(self) -> Result<(), Self> {
+        if self.has_errors() {
+            Err(self)
+        } else {
+            Ok(())
+        }
+    }
+
+    pub fn into_inner(mut self) -> Vec<SnapshotError> {
+        std::mem::take(&mut self.errors)
```
emilk, line 560: shouldn't this work?

```suggestion
    pub fn into_inner(mut self) -> Vec<SnapshotError> {
        self.errors
```

crates/egui_kittest/src/snapshot.rs
```diff
@@ -483,3 +485,100 @@ impl<State> Harness<'_, State> {
...
+    }
+
+    /// Panics if there are any errors, displaying each.
+    #[allow(clippy::unused_self)]
+    pub fn unwrap(self) {
+        // Panic is handled in drop
+    }
```
emilk, line 567: This seems unnecessary

crates/egui_kittest/src/snapshot.rs
```diff
@@ -483,3 +485,100 @@ impl<State> Harness<'_, State> {
...
+}
+
+impl Drop for SnapshotResults {
+    fn drop(&mut self) {
+        // Don't panic if we are already panicking (the test probably failed for another reason)
+        if std::thread::panicking() {
+            return;
+        }
+        assert!(!self.has_errors(), "{}", self);
```
emilk, line 582: Using an assert here will make the printed error unnecessarily complicated. Better to use an `if self.has_errors() { panic!(…) ` or better yet:

```rust
if self.has_error() {
    eprintln!("{self}");
    std::process::exit(1);
}
```

emilk (approved):

PR #5685, opened 2025-02-05
Master

### LINKED ISSUE #5679: Cannot load an image of mime-type "application/x-msdownload;charset=UTF-8"
I'm currently facing a problem in my program https://github.com/markusdd/EasyEDA_to_KiCAD_Lib_UI where I'm having to interface with a new image API. (using egui 0.30 for this as I cannot migrate to 0.31 yet because of a dependency, but I checked the relevant code, it has not changed)

Most older catalogue entries had URLs directly pointing to JPG files, like this one: https://assets.lcsc.com/images/lcsc/900x900/20230828_Megastar-ZX-PZ2-54-2-5PZZ_C7501276_back.jpg 

Analyzing in browser it looks like this: 

Image

Now, there is a new API like this: https://jlcpcb.com/api/file/downloadByFileSystemAccessId/8553684168530669568 

Image

So the MIME-Type is designated as a download and only the disposition gives away that it is an image. This is common practice for images that are not getting displayed in the browser but directly being downloaded.

The issue is: The image loader currently fails because it does not recognize the MIME-Type and I have found no way through the API to tell it to ignore that.

What I have tried is locally patching the "is_supported_mime" function in   src\loaders\image_loader.rs of egui-extras to do this:

```rust
if mime == "application/octet-stream" || mime == "application/x-msdownload;charset=UTF-8"
```

So I just 'bluntly' attached my requirement.

And it works perfectly, the images are being displayed again:

Image

But this is not a nice solution and I also cannot deploy my app like this, unless I switch to a fork of egui which I really want to avoid.

Any recommendations how to fix this properly or at least work around it for now?

markusdd:
as recommended by @lucasmerlin in #5679 

* Closes  #5679
* [ x] I have followed the instructions in the PR template

### DIFF
--- a/crates/egui_extras/src/loaders/image_loader.rs
+++ b/crates/egui_extras/src/loaders/image_loader.rs
@@ -35,11 +35,21 @@ fn is_supported_uri(uri: &str) -> bool {
 }
 
 fn is_supported_mime(mime: &str) -> bool {
-    // This is the default mime type for binary files, so this might actually be a valid image,
-    // let's relay on image's format guessing
-    if mime == "application/octet-stream" {
-        return true;
+    // some mime types e.g. reflect binary files or mark the content as a download, which
+    // may be a valid image or not, in this case, defer the decision on the format guessing
+    // or the image crate and return true here
+    let mimes_to_defer = [
+        "application/octet-stream",
+        "application/x-msdownload",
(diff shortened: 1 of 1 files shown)


markusdd:
accidental premature creation

Issue #5664, opened 2025-01-31
Panic due to Rect shape feathering


PPakalns:
**Describe the bug**

When scaling UI if rendered Rect gets too small it may cause panic due to Rect being too small in some dimension and feathering was applied incorrectly (Not taking into account assert in inner function)

**To Reproduce**
Render small Rect under feathering threshold with texture_id set to non default value.

**Expected behavior**
If Rect can not be feathered, do not try to feather it.

**Additional context**
Egui version: 0.30

Stacktrace:
```
thread 'main' panicked at /home/pakalns/.cargo/git/checkouts/egui-f290029b5e1afa28/942690e/crates/epaint/src/mesh.rs:151:9:
assertion failed: self.texture_id == TextureId::default()
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a730edcd67c7cb29d4458e170d4eb290387c27c3/library/std/src/panicking.rs:695:5
   1: core::panicking::panic_fmt
             at /rustc/a730edcd67c7cb29d4458e170d4eb290387c27c3/library/core/src/panicking.rs:75:14
   2: core::panicking::panic
             at /rustc/a730edcd67c7cb29d4458e170d4eb290387c27c3/library/core/src/panicking.rs:145:5
   3: colored_vertex
             at /home/pakalns/.cargo/git/checkouts/egui-f290029b5e1afa28/942690e/crates/epaint/src/mesh.rs:151:9
   4: stroke_path
   5: stroke_open
             at /home/pakalns/.cargo/git/checkouts/egui-f290029b5e1afa28/942690e/crates/epaint/src/tessellator.rs:482:9
   6: tessellate_line<epaint::stroke::Stroke>
             at /home/pakalns/.cargo/git/checkouts/egui-f290029b5e1afa28/942690e/crates/epaint/src/tessellator.rs:1587:9
   7: tessellate_rect
             at /home/pakalns/.cargo/git/checkouts/egui-f290029b5e1afa28/942690e/crates/epaint/src/tessellator.rs:1705:17
   8: tessellate_shape
             at /home/pakalns/.cargo/git/checkouts/egui-f290029b5e1afa28/942690e/crates/epaint/src/tessellator.rs:1394:17
   9: tessellate_clipped_shape
             at /home/pakalns/.cargo/git/checkouts/egui-f290029b5e1afa28/942690e/crates/epaint/src/tessellator.rs:1346:13
  10: tessellate_shapes
[...]

PR #5692, opened 2025-02-06
Fix panic when rendering thin textured rectangles

PPakalns:
* Closes https://github.com/emilk/egui/issues/5664


emilk (approved):
Nice; thanks for the fix

Issue #270, opened 2021-04-02
File dialogs and drag-and-drop of files


emilk:
Tracking issue for

* [x] File browser dialog for native ([`rfd`](https://github.com/PolyMeilex/rfd) now works since https://github.com/emilk/egui/pull/631)
* [ ] File browser dialog for `egui_web` ("Select a file to upload")
* [x] Drag-and-drop files onto `egui_web and `egui_glium` (https://github.com/emilk/egui/pull/637)

DrOptix:
I would like to contribute on this one. Going with something based on `std::fs` it would be consistent maybe except for `egui_web`.

On the other hand if we go with platform specific dialogs then that would show a familiar open dialog to the user.
- On Windows we can use Win32 API [GetOpenFileNameW](https://docs.microsoft.com/en-us/windows/win32/api/commdlg/nf-commdlg-getopenfilenamew) and [GetSaveFileNameW](https://docs.microsoft.com/en-us/windows/win32/api/commdlg/nf-commdlg-getsavefilenamew).
- For Linux, here I think there is no standard file open dialog. I need to study. Currently I'm using KDE, but I have a mix of file dialogs from KDE and Gnome, I guess it depends on with which toolkit the app was built.
- On MacOS I have no idea, but I expect the situation to be similar as on Windows.

Maybe we should just go with this [native-dialog](https://crates.io/crates/native-dialog) crate?

emilk:
If opening a file dialog on web is easy then maybe it can just be added to `native-dialog` instead (i.e. add a web-sys backend to `native-dialog`).

OfficialURL:
I should mention: I tried using `native-dialog` on Windows in a Bevy+egui application. About half of times, the dialogs failed with `ImplementationError("CoInitializeEx")`. I haven't yet filed an issue over there, as I need more testing to figure out what's going on, but it might be something worth being wary of.

OvermindDL1:
> About half of times, the dialogs failed with `ImplementationError("CoInitializeEx")`.

For note, this is caused by the windows COM, it has to be initialized in either, essentially, single-threaded or multi-threaded mode, however some COM objects only work properly in one mode or the other (usually sound annoyingly), and dragging and dropping with 'extended' clipboard attributes is another.

To see details about the issue and how winit deals with it then see:  https://github.com/rust-windowing/winit/issues/1255

CjS77:
[...]

PR #5697, opened 2025-02-08
Update file dialog example to be non-blocking.

hydra:
This PR updates the file dialog example to be non-blocking, additionally:

* the 'Open file...' button is disabled while picking.
* the main windows is responsive while picking - you can still drag files onto it and resize it just fine.

The result of a pick is recorded and kept in the state as before.

Video:
https://github.com/user-attachments/assets/29dd4cd9-b769-44f2-abe2-bd533728cdba

Related issues:
* https://github.com/emilk/egui/issues/270

Related discussions:
* https://github.com/emilk/egui/discussions/987#discussioncomment-1862596
* https://github.com/emilk/egui/discussions/5621


hydra:
This is my first PR for egui, be gentle 😄

hydra:
Also, I know this is just an example, but perhaps after merging this you might be interested in a PR to move the `FilePicker` from this PR into the main code base somewhere and expanding the API for the common use cases (files, directories, specification of path and file types).  

I note that other GUI frameworks I've used have 'rfd' features and similar APIs.
Cushy - https://github.com/khonsulabs/cushy/blob/main/src/dialog.rs#L352-L464

YgorSouza:
> Also, I know this is just an example, but perhaps after merging this you might be interested in a PR to move the `FilePicker` from this PR into the main code base somewhere and expanding the API for the common use cases (files, directories, specification of path and file types).
> 
> I note that other GUI frameworks I've used have 'rfd' features and similar APIs. Cushy - https://github.com/khonsulabs/cushy/blob/main/src/dialog.rs#L352-L464

I think it would be hard to implement this in a way that covers all the use cases of egui. For example, on wasm you can't spawn threads, so you have to use async, and then in rfd you have to pick which async backend you're using (async-std or tokio). Then there's the game engines and such. There are a couple of file pickers implemented as egui widgets instead of using rfd, as third-party crates. They are listed in the [wiki](https://github.com/emilk/egui/wiki/3rd-party-egui-crates#pre-built-windows).

hydra:
Also, I know this is just an example, but perhaps after merging this you might be interested in a PR to move the `FilePicker` from this PR into the main code base somewhere and expanding the API for the common use cases (files, directories, specification of path and file types).  

I note that other GUI frameworks I've used have 'rfd' features and similar APIs.
Cushy - https://github.com/khonsulabs/cushy/blob/main/src/dialog.rs#L352-L464


> > Also, I know this is just an example, but perhaps after merging this you might be interested in a PR to move the `FilePicker` from this PR into the main code base somewhere and expanding the API for the common use cases (files, directories, specification of path and file types).
> > I note that other GUI frameworks I've used have 'rfd' features and similar APIs. Cushy - https://github.com/khonsulabs/cushy/blob/main/src/dialog.rs#L352-L464
> 
> I think it would be hard to implement this in a way that covers all the use cases of egui. For example, on wasm you can't spawn threads, so you have to use async, and then in rfd you have to pick which async backend you're using (async-std or tokio). Then there's the game engines and such. There are a couple of file pickers implemented as egui widgets instead of using rfd, as third-party crates. They are listed in the [wiki](https://github.com/emilk/egui/wiki/3rd-party-egui-crates#pre-built-windows).

Yes, I understand that there's various backends for async and limitations with wasm.  I feel it would still be ok to provide solution for the common native desktop use-case however, along with appropriate module documentation containing links to the wiki of course.  This feels like it would serve the community better rather than requiring each egui developer to implement, test and debug their own API for a common use-case no?

landaire:
IMO this should supplement the existing blocking example, possibly as a button indicating a blocking dialog and another indicating non-blocking.

hydra:
@emilk ok, thanks for the response and the link.  I'll try and crate a separate example once I've worked out a good way of doing it.

PR #5701, opened 2025-02-09
Align `available_rect` docs with the new reality after #4590

podusowski:
...as after #4590, `Window`/`Area` is not constrained within it.


lucasmerlin (approved):
Thanks!

PR #5704, opened 2025-02-10
Run a frame per queued event in egui_kittest

lucasmerlin:
This should fix the remaining problems with the modifiers


abey79 (approved):
lgtm

PR #5706, opened 2025-02-10
Move gifs to git lfs

lucasmerlin:
Since we moved pngs to lfs I think we should also move the gifs there.


emilk (commented):
Let's make sure the README.md still looks good at github.com/emilk/egui and on https://crates.io/crates/egui

Could be we just let github host the images (copy-paste them into any PR description and copy the result in to the markdown)

lucasmerlin:
I think the images not being displayed in the diff must be a GitHub bug. Clicking "View File" shows the image and also shows "Stored with git lfs".

lucasmerlin:
Updated the readme to link to images via github user attachments. Should hopefully fix images in https://crates.io/crates/egui

PR #5712, opened 2025-02-11
Use `TextBuffer` for `layouter` in `TextEdit` instead of `&str`

kernelkind:
This change allows `layouter` to use the `TextBuffer` instead of `&str` in the closure. It is necessary when layout decisions depend on more than just the raw string content, such as metadata stored in the concrete type implementing `TextBuffer`.

In [our use case](https://github.com/damus-io/notedeck/pull/723), we needed this to support mention highlighting when a user selects a mention. Since mentions can contain spaces, determining mention boundaries from the `&str` alone is impossible. Instead, we use the `TextBuffer` implementation to retrieve the correct bounds.

See the video below for a demonstration:

https://github.com/user-attachments/assets/3cba2906-5546-4b52-b728-1da9c56a83e1

# Breaking change

This PR introduces a breaking change to the `layouter` function in `TextEdit`.

Previous API:
```rust
pub fn layouter(mut self, layouter: &'t mut dyn FnMut(&Ui, &str, f32) -> Arc<Galley>) -> Self
```

New API:
```rust
pub fn layouter(mut self, layouter: &'t mut dyn FnMut(&Ui, &dyn TextBuffer, f32) -> Arc<Galley>) -> Self
```

## Impact on Existing Code

• Any existing usage of `layouter` will **no longer compile**.
• Callers must update their closures to use `&dyn TextBuffer` instead of `&str`.

## Migration Guide

Before:
```rust
let mut layouter = |ui: &Ui, text: &str, wrap_width: f32| {
    let layout_job = my_highlighter(text);
    layout_job.wrap.max_width = wrap_width;
    ui.fonts(|f| f.layout_job(layout_job))
};
```

After:
```rust
let mut layouter = |ui: &Ui, text: &dyn TextBuffer, wrap_width: f32| {
    let layout_job = my_highlighter(text.as_str());
    layout_job.wrap.max_width = wrap_width;
    ui.fonts(|f| f.layout_job(layout_job))
};
```

---

* There is not an issue for this change.
* [x]  I have followed the instructions in the PR template

### DIFF
--- a/crates/egui/src/widgets/text_edit/text_buffer.rs
+++ b/crates/egui/src/widgets/text_edit/text_buffer.rs
@@ -172,6 +172,39 @@ pub trait TextBuffer {
             self.delete_selected(&CCursorRange::two(min, max))
         }
     }
+
+    /// Returns a unique identifier for the implementing type.
+    ///
+    /// This is useful for downcasting from this trait to the implementing type.
+    /// Here is an example usage:
+    /// ```
+    /// use egui::TextBuffer;
+    /// use std::any::TypeId;
+    ///
+    /// struct ExampleBuffer {}
+    ///
+    /// impl TextBuffer for ExampleBuffer {
+    ///     fn is_mutable(&self) -> bool { unimplemented!() }
+    ///     fn as_str(&self) -> &str { unimplemented!() }
+    ///     fn insert_text(&mut self, text: &str, char_index: usize) -> usize { unimplemented!() }
+    ///     fn delete_char_range(&mut self, char_range: std::ops::Range<usize>) { unimplemented!() }
+    ///
+    ///     // Implement it like the following:
+    ///     fn type_id(&self) -> TypeId {
+    ///         TypeId::of::<Self>()
+    ///     }
+    /// }
+    ///
+    /// // Example downcast:
+    /// pub fn downcast_example(buffer: &dyn TextBuffer) -> Option<&ExampleBuffer> {
+    ///     if buffer.type_id() == TypeId::of::<ExampleBuffer>() {
+    ///         unsafe { Some(&*(buffer as *const dyn TextBuffer as *const ExampleBuffer)) }
+    ///     } else {
+    ///         None
+    ///     }
+    /// }
+    /// ```
+    fn type_id(&self) -> std::any::TypeId;
 }
 
 impl TextBuffer for String {
@@ -218,6 +251,10 @@ impl TextBuffer for String {
     fn take(&mut self) -> String {
         std::mem::take(self)
     }
+
+    fn type_id(&self) -> std::any::TypeId {
+        std::any::TypeId::of::<Self>()
+    }
 }
 
 impl TextBuffer for Cow<'_, str> {
@@ -248,6 +285,10 @@ impl TextBuffer for Cow<'_, str> {
     fn take(&mut self) -> String {
         std::mem::take(self).into_owned()
     }
+
+    fn type_id(&self) -> std::any::TypeId {
+        std::any::TypeId::of::<Cow<'_, str>>()
+    }
 }
 
 /// Immutable view of a `&str`!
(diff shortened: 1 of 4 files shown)


crates/egui/src/widgets/text_edit/text_buffer.rs
```diff
@@ -183,6 +183,15 @@ pub trait TextBuffer {
...
         }
     }
+
+    /// Returns a unique identifier for the implementing type.
+    ///
+    /// By default, this returns `0`, which represents an unknown type.
+    /// Implementers should override this method with a unique identifier
+    /// to enable safe downcasting.
+    fn type_id(&self) -> usize {
+        0
```
emilk, line 193: `fn as_any(&self) -> &dyn std::any::Any;` seems like a better pattern?

emilk (approved):
Makes sense

crates/egui/src/widgets/text_edit/text_buffer.rs
kernelkind: yeah that's definitely better, now no unsafe code is necessary to downcast

kernelkind:
added suggested change `as_any` and rebased with master

crates/egui/src/widgets/text_edit/text_buffer.rs
```diff
@@ -245,10 +251,14 @@ impl TextBuffer for Cow<'_, str> {
...
     }
+
+    fn as_any(&self) -> &dyn std::any::Any {
+        self
+    }
 }
 
 /// Immutable view of a `&str`!
-impl TextBuffer for &str {
+impl TextBuffer for &'static str {
```
lucasmerlin, line 261: Why were these changed to 'static? Wouldn't this mean `some_string.as_str()` stops implementing TextBuffer with this change?

kernelkind: `std::any::Any` only works for static types, so that's the downside of using it. If this isn't acceptable we can revert back to the type id implementation, unless you have a better idea.

lucasmerlin: Ah I see, that's annoying 🤔 What do you need Any for? 
I think I'd prefer not having the 'static implementation, if the type_id thing works well enough for your usecase

kernelkind: Any is useful for downcasting without writing unsafe code:
```rust
pub fn downcast_post_buffer(buffer: &dyn TextBuffer) -> Option<&PostBuffer> {
    buffer.as_any().downcast_ref::<PostBuffer>()
}
```

But with using type_id we need:
```rust
pub fn downcast_post_buffer(buffer: &dyn TextBuffer) -> Option<&PostBuffer> {
    let mut hasher = DefaultHasher::new();
    TypeId::of::<PostBuffer>().hash(&mut hasher);
    let post_id = hasher.finish() as usize;

    if buffer.type_id() == post_id {
        unsafe { Some(&*(buffer as *const dyn TextBuffer as *const PostBuffer)) }
    } else {
        None
    }
}
```

That's understandable to not want the 'static limitation, I can certainly revert back to using type_id instead of Any

lucasmerlin: @emilk what do you prefer, considering the 'static limitation?

kernelkind:
reverted to 6b999e0f and rebased with master

kernelkind:
I think it would be better to use `std::any::TypeId` instead of `usize`. Then implementers would only need to do:
```rust
    fn type_id(&self) -> std::any::TypeId {
        std::any::TypeId::of::<Self>()
    }
```

instead of:
```rust
    fn type_id(&self) -> usize {
        let mut hasher = DefaultHasher::new();
        std::any::TypeId::of::<Self>().hash(&mut hasher);
        hasher.finish() as usize
    }
```

crates/egui/src/widgets/text_edit/text_buffer.rs
emilk: The `'static` limitation is annoying; I din't foresee that.

I suggest we either go with @kernelkind ideas of returning a `std::any::TypeId`, or we return an `Option<&dyn std::any::Any>` that returns `None` for non-`'static` types

kernelkind:
I migrated to using `std::any::TypeId`, added an example usage in a doctest, and rebased with master

PR #5713, opened 2025-02-12
Add `Popup` and `Tooltip`, unifying the previous behaviours

lucasmerlin:
This introduces new `Tooltip` and `Popup` structs that unify and extend the old popups and tooltips.

`Popup` handles the positioning and optionally stores state on whether the popup is open (for click based popups like `ComboBox`, menus, context menus).
`Tooltip` is based on `Popup` and handles state of whether the tooltip should be shown (which turns out to be quite complex to handles all the edge cases).

Both `Popup` and `Tooltip` can easily be constructed from a `Response` and then customized via builder methods. 

This also introduces `PositionAlign`, for aligning something outside of a `Rect` (in contrast to `Align2` for aligning inside a `Rect`). But I don't like the name, any suggestions? Inspired by [mui's tooltip positioning](https://mui.com/material-ui/react-tooltip/#positioned-tooltips).

* Part of #4607 

TODOs:
- [x] Automatic tooltip positioning based on available space
- [x] Review / fix / remove all code TODOs 
- [x] ~Update the helper fns on `Response` to be consistent in naming and parameters (Some use tooltip, some hover_ui, some take &self, some take self)~ actually, I think the naming and parameter make sense on second thought
- [x] Make sure all old code is marked deprecated

For discussion during review:
- the following check in `show_tooltip_for` still necessary?:
```rust
     let is_touch_screen = ctx.input(|i| i.any_touches());
     let allow_placing_below = !is_touch_screen; // There is a finger below. TODO: Needed?
```


lucasmerlin:
While I think it's nice that the ComboBoxes indicate the direction they will open in, it can look somewhat confusing in edge cases:

Screenshot 2025-02-12 at 18 31 35

(the second last combobox is bigger so it will wrap while the last one doesn't yet wrap)

Since this is also difficult to implement with the new Popup positioning, I'll remove this behavior and always point the arrow down.

crates/egui/src/containers/popup.rs
```diff
@@ -1,457 +1,592 @@
...
-    widget_id: Id,
-    text: impl Into<WidgetText>,
-) -> Option<()> {
-    show_tooltip(ctx, parent_layer, widget_id, |ui| {
-        crate::widgets::Label::new(text).ui(ui);
-    })
-}
+// TODO: Find a better name for Position and PositionAlign
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct PositionAlign(pub Position, pub Align);
```
Mingun, line 135: The great gojs library models similar conception as two properties of a [`Spot`](https://gojs.net/latest/api/symbols/Spot.html) type in the widget (which they called `GraphObject`):
- alignment
- alignmentFocus

For examples of positioning see the [Panels](https://gojs.net/latest/intro/panels.html) examples, especially a [`Panel.Spot`](https://gojs.net/latest/intro/panels.html#SpotPanels) type.

lucasmerlin: Nice, that is a nice way to think about it! I've changed my PositionAlign to be based on two `Align2`, which feels nicer and allows for more alignments. It's called Align4 now 😄

crates/egui/src/containers/old_popup.rs
```diff
@@ -0,0 +1,211 @@
+//! Show popup windows, tooltips, context menus etc.
```
emilk, line 1:
```suggestion
//! Old and deprecated API for popups. Use [`egui::Popup`] instead
```

crates/egui/src/containers/popup.rs
```diff
@@ -1,457 +1,517 @@
...
+    Sense, Ui, UiKind,
 };
+use emath::{vec2, Align, Align4, Pos2, Rect};
+use std::iter::once;
 
-// ----------------------------------------------------------------------------
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PopupAnchor {
+    /// Show the popup at the given rect.
+    Rect(Rect),
```
emilk, line 11: Will the popup always have the given size, even if its contents is smaller? What if it overflows?

crates/egui/src/containers/popup.rs
```diff
@@ -1,457 +1,517 @@
...
-    add_contents: impl FnOnce(&mut Ui) -> R,
-) -> Option<R> {
-    show_tooltip_at_pointer(ctx, parent_layer, widget_id, add_contents)
+impl From<&Response> for PopupAnchor {
+    fn from(response: &Response) -> Self {
+        let mut widget_rect = response.rect;
+        if let Some(to_global) = response.ctx.layer_transform_to_global(response.layer_id) {
+            widget_rect = to_global * widget_rect;
+        }
+        Self::Rect(widget_rect)
```
emilk, line 41: Oh, it's not the rectangle of the popup, but the rectangle of the parent widget _next_ to which the popup should be shown. That should be made clear in the docstring.

crates/egui/src/containers/popup.rs
```diff
@@ -1,457 +1,517 @@
...
-
-        // Keep the left edge of the tooltip in line with the cursor:
-        pointer_rect.min.x = pointer_pos.x;
-
-        // Transform global coords to layer coords:
-        if let Some(from_global) = ctx.layer_transform_from_global(parent_layer) {
-            pointer_rect = from_global * pointer_rect;
+impl PopupAnchor {
+    /// The rect the popup should be shown at.
+    pub fn rect(self, popup_id: Id, ctx: &Context) -> Option<Rect> {
```
emilk, line 47: I'm still not sure what this means. "Show popup in this rectangle" or "Show popup outside but next to this rectangle"?

emilk (commented):
There's a lot of code to review here 😅  I've only looked at a small portion of it so far. I think I need a walk-through

crates/egui/src/containers/area.rs
```diff
@@ -543,7 +555,8 @@ impl Prepared {
         let mut ui_builder = UiBuilder::new()
             .ui_stack_info(UiStackInfo::new(self.kind))
             .layer_id(self.layer_id)
-            .max_rect(max_rect);
+            .max_rect(max_rect)
+            .layout(self.layout);
```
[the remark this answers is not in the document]
lucasmerlin: That'd be nice, I'll do that in a follow-up PR: https://github.com/emilk/egui/issues/5725

crates/emath/src/align.rs
```diff
@@ -270,3 +298,201 @@ impl std::ops::IndexMut<usize> for Align2 {
...
+/// └──────────┘ └────────────────────────────────────┘ └───────────┘
+///              ┌────────────┐  ┌──────┐  ┌──────────┐              
+///              │BOTTOM_START│  │BOTTOM│  │BOTTOM_END│              
+///              └────────────┘  └──────┘  └──────────┘              
+/// ```
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub struct Align4 {
+    pub align: Align2,
+    pub focus: Align2,
```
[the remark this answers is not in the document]
lucasmerlin: Nice, parent/child is much better!

crates/emath/src/align.rs
```diff
@@ -270,3 +298,201 @@ impl std::ops::IndexMut<usize> for Align2 {
...
+/// ┌──────────┐ │                                    │ ┌───────────┐
+/// │ LEFT_END │ │                                    │ │ RIGHT_END │
+/// └──────────┘ └────────────────────────────────────┘ └───────────┘
+///              ┌────────────┐  ┌──────┐  ┌──────────┐              
+///              │BOTTOM_START│  │BOTTOM│  │BOTTOM_END│              
+///              └────────────┘  └──────┘  └──────────┘              
+/// ```
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub struct Align4 {
```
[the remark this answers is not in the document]
Mingun: Maybe `Pin` (because one widget is pinned to another)? But that could conflict with `std::pin::Pin`.

crates/emath/src/rect_relation.rs
```diff
@@ -0,0 +1,267 @@
...
+            parent: self.parent.flip(),
+            child: self.child.flip(),
+        }
+    }
+
+    /// Returns the 3 alternative [`RectRelation`]s that are flipped in various ways, for use
+    /// with [`RectRelation::find_best_align`].
+    pub fn alternatives(self) -> [Self; 3] {
+        [self.flip_x(), self.flip_y(), self.flip()]
+    }
```
emilk, line 224: Maybe call this `fn symmetries`?

crates/emath/src/rect_relation.rs
```diff
@@ -0,0 +1,267 @@
...
+/// └──────────┘ └────────────────────────────────────┘ └───────────┘
+///              ┌────────────┐  ┌──────┐  ┌──────────┐              
+///              │BOTTOM_START│  │BOTTOM│  │BOTTOM_END│              
+///              └────────────┘  └──────┘  └──────────┘              
+/// ```
+// There is no `new` function on purpose, since writing out `parent` and `child` is more
+// reasonable.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub struct RectRelation {
```
emilk, line 30: This is a very useful abstraction 👍

crates/egui_kittest/tests/popup.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    harness.get_by_label_contains("tooltip").hover();
+    harness.run();
+    harness.get_by_label("link").hover();
+    harness.run();
+    harness.get_by_label("link").simulate_click();
+
+    harness.run();
+
+    assert!(harness.state().link_clicked);
```
emilk, line 30: Does this really check that the popup is still open, as the function name implies?

crates/egui/src/containers/popup.rs
```diff
@@ -1,457 +1,517 @@
...
-        parent_layer,
-        widget_id,
-        allow_placing_below,
-        widget_rect,
-        Box::new(add_contents),
-    )
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum SetOpenState {
+    /// Do nothing
+    DoNothing,
```
lucasmerlin, line 79: Remove and make it Option<...> and rename to OpenCommand

crates/egui/src/containers/popup.rs
```diff
@@ -1,457 +1,517 @@
...
+    Tooltip,
+    Menu,
 }
 
-pub fn tooltip_id(widget_id: Id, tooltip_count: usize) -> Id {
-    widget_id.with(tooltip_count)
+pub struct Popup<'a> {
+    pub id: Id,
+    pub anchor: PopupAnchor,
+    position_align: RectRelation,
```
lucasmerlin, line 144: Rename

crates/egui/src/containers/popup.rs
```diff
@@ -1,457 +1,517 @@
...
+                    ctx.memory_mut(|mem| mem.close_popup());
+                }
+            }
         }
-        PopupCloseBehavior::IgnoreClicks => false,
-    };
 
-    if parent_ui.input(|i| i.key_pressed(Key::Escape)) || should_close {
-        parent_ui.memory_mut(|mem| mem.close_popup());
+        Some(response)
```
emilk, line 515: As you said: it would be nice to have a `response.closed()` flag the user can check. That could also be set by `Area`, `CollapsingPanel`, `ColorPickerButton` etc

crates/egui/src/containers/popup.rs
```diff
@@ -1,457 +1,517 @@
...
+                if should_close(close_behavior) {
+                    *open = false;
+                }
+            }
+            OpenKind::Memory { close_behavior, .. } => {
+                if should_close(close_behavior) {
+                    ctx.memory_mut(|mem| mem.close_popup());
+                }
+            }
         }
```
emilk, line 513: It would be nice with something like

```rust
if should_close(close_behavior) {
    open_kind.set_open(false);
}
```

…thought maybe for a later PR

crates/egui/src/containers/popup.rs
```diff
@@ -1,457 +1,517 @@
...
-    if allow_placing_below
-        && widget_rect.bottom() + spacing + tooltip_size.y <= screen_rect.bottom()
-    {
-        return (
-            Align2::LEFT_TOP,
-            widget_rect.left_bottom() + spacing * Vec2::DOWN,
-        );
+impl<'a> Popup<'a> {
+    /// Create a new popup
+    pub fn new(id: Id, anchor: impl Into<PopupAnchor>, layer_id: LayerId) -> Self {
```
emilk, line 167: As mentioned on call: let's pass `Context` by argument and store in `Popup`

crates/egui/src/containers/tooltip.rs
```diff
@@ -0,0 +1,372 @@
...
+    pub popup: Popup<'a>,
+    layer_id: LayerId,
+    widget_id: Id,
+}
+
+impl<'a> Tooltip<'a> {
+    /// Show a tooltip that is always open
+    pub fn new(widget_id: Id, anchor: impl Into<PopupAnchor>, layer_id: LayerId) -> Self {
+        Self {
+            // TODO(lucasmerlin): Set width somehow (we're missing context here)
```
emilk, line 18: As mentioned on call; let's pass in `Context` by argument and store in `Popup`

crates/egui/src/containers/tooltip.rs
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+        let rect = popup.anchor.rect(popup.id, ctx)?;
+
+        let mut state = ctx.pass_state_mut(|fs| {
+            // Remember that this is the widget showing the tooltip:
+            fs.layers
+                .entry(parent_layer)
+                .or_default()
+                .widget_with_tooltip = Some(widget_id);
```
emilk, line 113: We could maybe get rid of `LayerState::open_popups` if we also stored which widgets has a _popup_ open (attached to them). That would allow us to simplify and improve the "don't show a tooltip if this widget has a popup open" check in https://github.com/emilk/egui/blob/b36311a98ae9475f83aa6d8f5a3a91fe4df5d2b2/crates/egui/src/response.rs#L615-L619

(can be saved for a future PR though)

emilk (approved):

crates/egui_kittest/tests/popup.rs
lucasmerlin: If the popup closed, the link wouldn't be clickable, so I think it does. But I've renamed it to `test_interactive_tooltip`, since we also test whether we can actually interact with the popup contents. 

If the link would be clicked even though the popup was closed, that would be a bug in some other part of egui. Not sure if there would be some better way to check if the popup is still open (apart from a snapshot test, of course)

crates/egui/src/containers/popup.rs
lucasmerlin: Created another issue for this: https://github.com/emilk/egui/issues/5727

PR #5714, opened 2025-02-12
Add guidelines for image comparison tests

Wumpf:
Guidelines & why images may differ

Based on (but slightly altered):
* https://github.com/rerun-io/rerun/pull/8989

### DIFF
--- a/crates/egui_kittest/README.md
+++ b/crates/egui_kittest/README.md
@@ -55,3 +55,48 @@ You should add the following to your `.gitignore`:
 **/tests/snapshots/**/*.diff.png
 **/tests/snapshots/**/*.new.png
 ```
+
+### Guidelines for writing snapshot tests
+
+* Whenever **possible** prefer regular Rust tests or `insta` snapshot tests over image comparison tests because…
+  * …they are slow to run
+  * …they are brittle since unrelated side effects (like a change in color) can cause the test to fail
+  * …images take up repo space
+* images should…
+  * …be checked in as LFS file
+  * …depict exactly what's tested and nothing else
+  * …have a low resolution to avoid growth in repo size
+  * …have a low comparison threshold to avoid the test passing despite unwanted differences
+
+### Why does CI / another computer produce a different image?
+
+First check whether the difference is due to a change in enabled rendering features, potentially due to difference in hardware (/software renderer) capabilitites.
+Generally you should carefully enforcing the same set of features for all test runs, but this may happen nonetheless|
+
+However, smaller discrepancies may be caused by a variety of implementation details depending on the concrete GPU/OS/rendering backend/graphics driver (even between different versions of the same driver).
+For instance:
+* multi-sample anti-aliasing
+  * sample placement and sample resolve steps are implementation defined
+  * alpha-to-coverage algorithm/pattern can wary wildly between implementations
+* texture filtering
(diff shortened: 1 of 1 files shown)


crates/egui_kittest/README.md
```diff
@@ -55,3 +55,48 @@ You should add the following to your `.gitignore`:
 **/tests/snapshots/**/*.diff.png
 **/tests/snapshots/**/*.new.png
 ```
+
+### Guidelines for writing snapshot tests
+
+* Whenever **possible** prefer regular Rust tests or `insta` snapshot tests over image comparison tests because…
+  * …they are slow to run
```
Wumpf, line 62:
```suggestion
  * …they are relatively slow to run
```

crates/egui_kittest/README.md
```diff
@@ -55,3 +55,48 @@ You should add the following to your `.gitignore`:
...
+  * …they are brittle since unrelated side effects (like a change in color) can cause the test to fail
+  * …images take up repo space
+* images should…
+  * …be checked in as LFS file
```
Wumpf, line 66: that one is more specific to egui & rerun, should write something else here

crates/egui_kittest/README.md
```diff
@@ -55,3 +55,48 @@ You should add the following to your `.gitignore`:
...
+  * …depict exactly what's tested and nothing else
+  * …have a low resolution to avoid growth in repo size
+  * …have a low comparison threshold to avoid the test passing despite unwanted differences
```
lucasmerlin, line 69:
```suggestion
  * …have a low comparison threshold to avoid the test passing despite unwanted differences (the default treshold should be fine)
```

lucasmerlin (commented):
This is great, but I'm a bit worried that the "Why does CI / another computer produce a different image?" section might overwhelm the average user who just wants to test whether their button still looks the same. 
If they run into issues with snapshots being different in CI, the reasonable thing for them to do would be increasing the threshold until it works. I think we should mention that somewhere, before explaining all the rendering things.

crates/egui_kittest/README.md
lucasmerlin: Yeah I think we can suggest using lfs, but for small projects with only a couple of snapshot tests it's probably fine to just have them in git

Wumpf:
I disagree, it's important to understand why an image comparison tests may fail or even be flaky if you ever run into that situation. I do recognize that the detail I went into here stretches the intended audience by quite a bit which may only be concerned with "simple" ui testing. But as I see it it offers a way for more systematic reasoning than just mindlessly adjusting thresholds until it works (which, while being the inevitable reality 😞, is obviously a very precarious practice).
Not feeling super strongly about this though, I can always just leave that in in the Rerun contribution docs only 🤷  (Emil [suggested](https://github.com/rerun-io/rerun/pull/8989#discussion_r1952101726) to put this here and I thought that made sense)

Wumpf:
maybe I can adjust the text a bit to make this more of "stop, think, look at the changes" before going into the weeds.

Wumpf:
I changed the framing and added more words talking about immediate actions instead. What do you think @lucasmerlin , does this hit the right compromise or would you still rather leave that out entirely?

lucasmerlin (approved):
Awesome, I think it's perfect now!

PR #5715, opened 2025-02-12
Clarify platform-specific details for `Viewport`s

### LINKED ISSUE #4469: `with_position` method of ViewportBuilder does not seem to work on wayland
```toml
eframe = "0.27.2"
egui_extras = "0.27.2"
```

**Describe the bug**

ViewportBuilder's with_position method does not seem to work on sway/wayland

 - OS: Linux(Wayland, musl-libc)

**Additional context**

code:
```rust
(...)

// UI part
pub fn run_ui(
    config: &AppConfig,
    header: Option<String>,
    content: String,
) -> Result<(), eframe::Error> {
            .with_app_id(String::from("evilbee-notification"))
            .with_inner_size([
                config.width.unwrap_or(160.0),
                config.heigth.unwrap_or(120.0),
            ])
            .with_resizable(false)
            .with_taskbar(false)
            .with_decorations(true)
            .with_position([0.0, 0.0])
            .with_window_level(egui::WindowLevel::AlwaysOnTop),
        ..Default::default()
    };
    eframe::run_native(
        "EvilBee Notification",
        options,
        Box::new(|cc| {
            // This gives us image support:

            Box::new(EvilBeeUI {
                header,
                content,
                timer: 0,
            })
        }),
    )
}

(...)
```

also when i try to access inner_rect and outer_rect information they return `None`
```
(...)

                if let Some(inner_rect) = ctx.input(|i| i.viewport().inner_rect) {
                    ui.label(format!(
                        "Inner Rect: Pos: {:?}, Size: {:?} (points)",
                        inner_rect.min,
                        inner_rect.size()
                    ));
                }
                if let Some(outer_rect) = ctx.input(|i| i.viewport().outer_rect) {
                    ui.label(format!(
                        "Outer Rect: Pos: {:?}, Size: {:?} (points)",
                        outer_rect.min,
                        outer_rect.size()
                    ));
                }

(...)
```

aspiringLich:
On Wayland (and android apparently) it is not possible for applications to position their windows or request the window position (for some reason). This is not specified in the documentation.

proof: [`winit::set_outer_position`], [`winit::outer_position`], [`winit::inner_position`] all specify that Android / Wayland are not supported.

[`winit::set_outer_position`]: https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_outer_position
[`winit::outer_position`]: https://docs.rs/winit/latest/winit/window/struct.Window.html#method.outer_position
[`winit::inner_position`]: https://docs.rs/winit/latest/winit/window/struct.Window.html#method.inner_position

* Maybe closes #4469
* the lack of support makes me sad. I believe this makes some items in #3556 impossible. Oh well!

### DIFF
--- a/crates/egui/src/data/input.rs
+++ b/crates/egui/src/data/input.rs
@@ -214,11 +214,17 @@ pub struct ViewportInfo {
     /// The inner rectangle of the native window, in monitor space and ui points scale.
     ///
     /// This is the content rectangle of the viewport.
+    ///
+    /// Note that on Android / Wayland, this will always be `None` since getting the
+    /// position of the window is not possible.
     pub inner_rect: Option<Rect>,
 
     /// The outer rectangle of the native window, in monitor space and ui points scale.
     ///
     /// This is the content rectangle plus decoration chrome.
+    ///
+    /// Note that on Android / Wayland, this will always be `None` since getting the
+    /// position of the window is not possible.
     pub outer_rect: Option<Rect>,
 
     /// Are we minimized?
--- a/crates/egui/src/viewport.rs
+++ b/crates/egui/src/viewport.rs
@@ -543,6 +543,16 @@ impl ViewportBuilder {
 
     /// The initial "outer" position of the window,
     /// i.e. where the top-left corner of the frame/chrome should be.
+    ///
+    /// ## Platform-specific:
+    ///
+    /// - **iOS:** Can only be called on the main thread. Sets the top left coordinates of the
+    ///   window in the screen space coordinate system.
+    /// - **Web:** Sets the top-left coordinates relative to the viewport. Doesn't account for CSS
+    ///   [`transform`].
+    /// - **Android / Wayland:** Unsupported.
+    ///
+    /// [`transform`]: https://developer.mozilla.org/en-US/docs/Web/CSS/transform
     #[inline]
     pub fn with_position(mut self, pos: impl Into<Pos2>) -> Self {
         self.position = Some(pos.into());


crates/egui/src/viewport.rs
```diff
@@ -543,6 +543,16 @@ impl ViewportBuilder {
 
     /// The initial "outer" position of the window,
     /// i.e. where the top-left corner of the frame/chrome should be.
+    ///
+    /// ## Platform-specific:
+    ///
+    /// - **iOS:** Can only be called on the main thread. Sets the top left coordinates of the
+    ///   window in the screen space coordinate system.
```
lucasmerlin, line 550: > Can only be called on the main thread.

I don't think this is true for egui here, since the event is just queued here and egui will later call this on the main thread.

lucasmerlin (approved):
Nice, thank you!

crates/egui/src/data/input.rs
```diff
@@ -214,11 +214,17 @@ pub struct ViewportInfo {
     /// The inner rectangle of the native window, in monitor space and ui points scale.
     ///
     /// This is the content rectangle of the viewport.
+    ///
+    /// Note that on Android / Wayland, this will always be `None` since getting the
+    /// position of the window is not possible.
```
lucasmerlin, line 219:
```suggestion
    ///
    /// **`eframe` notes**:
    ///
    /// On Android / Wayland, this will always be `None` since getting the
    /// position of the window is not possible.
```

crates/egui/src/data/input.rs
```diff
@@ -214,11 +214,17 @@ pub struct ViewportInfo {
...
     pub inner_rect: Option<Rect>,
 
     /// The outer rectangle of the native window, in monitor space and ui points scale.
     ///
     /// This is the content rectangle plus decoration chrome.
+    ///
+    /// Note that on Android / Wayland, this will always be `None` since getting the
+    /// position of the window is not possible.
```
lucasmerlin, line 227:
```suggestion
    ///
    /// **`eframe` notes**:
    /// 
    /// On Android / Wayland, this will always be `None` since getting the
    /// position of the window is not possible.
```

crates/egui/src/viewport.rs
```diff
@@ -543,6 +543,15 @@ impl ViewportBuilder {
...
     /// i.e. where the top-left corner of the frame/chrome should be.
+    ///
+    /// ## Platform-specific:
+    ///
+    /// - **iOS:** Sets the top left coordinates of the window in the screen space coordinate system.
+    /// - **Web:** Sets the top-left coordinates relative to the viewport. Doesn't account for CSS
+    ///   [`transform`].
+    /// - **Android / Wayland:** Unsupported.
+    ///
+    /// [`transform`]: https://developer.mozilla.org/en-US/docs/Web/CSS/transform
```
lucasmerlin, line 554:
```suggestion
    ///
    /// **`eframe` notes**:
    ///
    /// - **iOS:** Sets the top left coordinates of the window in the screen space coordinate system.
    /// - **Web:** Sets the top-left coordinates relative to the viewport. Doesn't account for CSS
    ///   [`transform`].
    /// - **Android / Wayland:** Unsupported.
    ///
    /// [`transform`]: https://developer.mozilla.org/en-US/docs/Web/CSS/transform
```

lucasmerlin (approved):
Ah, should've realized in the first review, we should mention that this is eframe specific, since it might be different for someone e.g. using bevy_egui. I'll apply the suggestions and merge this now, thanks!

crates/egui/src/data/input.rs
```diff
@@ -214,11 +214,21 @@ pub struct ViewportInfo {
...
+    /// On Android / Wayland, this will always be `None` since getting the
+    /// position of the window is not possible.
     pub inner_rect: Option<Rect>,
 
     /// The outer rectangle of the native window, in monitor space and ui points scale.
     ///
     /// This is the content rectangle plus decoration chrome.
+    ///
+    /// **`eframe` notes**:
+    ///
```
lucasmerlin, line 229:
```suggestion
    ///
```
oops

PR #5716, opened 2025-02-13
⚠️ Improved menu based on `egui::Popup`

### LINKED ISSUE #4607: Unify menus with tooltips/popups

lucasmerlin:
Continuation of #5713

**Silently breaking changes:**
- Menus now close on click by default, this is configurable via `PopupCloseBehavior`

**Additional additions:**
- `Button::right_text`
- `StyleModifier`

This is a rewrite of the egui menus, with the following goals:
- submenu buttons should work everywhere, in a popup, context menu, area, in some random Ui
- remove the menu state from Ui
- should work just like the previous menu
- ~proper support for keyboard navigation~
  - It's better now but requires further work to be perfect
- support `PopupCloseBehavior`

* Closes #4607 


crates/egui/src/containers/combo_box.rs
```diff
@@ -374,6 +374,7 @@ fn combo_box_dyn<'c, R>(
 
     let inner = Popup::menu(&button_response)
         .id(popup_id)
+        .style(None)
```
lucasmerlin, line 377: Should we set the menu styles for the ComboBox? I'm overwriting this here since this is how it was before, but I think the menu styles would look nice here

crates/egui_demo_lib/src/demo/demo_app_windows.rs
```diff
@@ -227,29 +226,30 @@ impl DemoWindows {
...
-                    ui.hyperlink_to(
-                        egui::RichText::new(GITHUB).size(font_size),
-                        "https://github.com/emilk/egui",
-                    );
+                    ui.with_layout(egui::Layout::right_to_left(egui::Align::Center), |ui| {
+                        use egui::special_emojis::GITHUB;
+                        ui.hyperlink_to(
+                            egui::RichText::new("🦋").size(font_size),
+                            "https://bsky.app/profile/ernerfeldt.bsky.social",
+                        );
```
lucasmerlin, line 246: Seems like there is no butterfly in egui :(

Screenshot 2025-02-20 at 19 04 02

crates/egui_demo_lib/tests/snapshots/demos/Popups.png
lucasmerlin, line 1: I thought this would be a fun way to display the options, but it seems like the readability suffers a bit. Should I change it?

crates/egui_demo_lib/src/demo/demo_app_windows.rs
oscargus: You just have to wait for🐛...

emilk:
> Menus now close on click by default, this is configurable via `PopupCloseBehavior`

What happens if you _also_ call `ui.close();` (or the old `ui.close_menu()`) in a buttons on-click handler inside a menu? Do we run the risk of closing the `egui::Window` that the menu is in?

lucasmerlin:
Add reset button to demo

lucasmerlin:
Add some note that ui.close doesnt bubble up across areas

crates/egui/src/containers/combo_box.rs
emilk: Good question 🤔  If so I think we save it for another PR

crates/egui/src/containers/popup.rs
```diff
@@ -226,29 +258,27 @@ impl<'a> Popup<'a> {
...
-                } else {
-                    None
-                },
-                PopupCloseBehavior::CloseOnClick,
-            )
+            .open_memory(if response.clicked() {
+                Some(SetOpenCommand::Toggle)
+            } else {
+                None
+            })
```
emilk, line 265: style: this is a nice function to know about

```suggestion
            .open_memory(response.clicked().then_some(SetOpenCommand::Toggle))
```

crates/egui/src/containers/popup.rs
```diff
@@ -367,6 +390,17 @@ impl<'a> Popup<'a> {
         self
     }
 
+    /// Set the style for the popup contents.
+    ///
+    /// Default:
+    /// - is [`menu_style`] for [`Self::menu`] and [`Self::context_menu`]
+    /// - is [`None`] otherwise
+    #[inline]
+    pub fn style(mut self, style: Option<fn(&mut Style)>) -> Self {
```
emilk, line 399: I wonder if we should call this `style_modifier` or something, since `.style(…)` makes it sound like it takes a `Style` by argument

emilk: Also, using a normal function pointer here severely limits the usability of the function. Using `Box<dyn Fn` would make it a lot more useful (at the cost of an allocation)

crates/egui_demo_lib/tests/snapshots/demos/Popups.png
emilk, line 1: This is a really cool style for demo

emilk: Needs a `Reset`-button though, and a link to its source code

emilk (commented):
If the default behavior is `CloseOnClick`, then we can remove the calls to `ui.close()` from https://github.com/emilk/egui/blob/ab0f83db83d5a023ba12ae221f7a37fc50a6acb2/crates/egui_demo_lib/src/demo/demo_app_windows.rs#L308-L363

crates/egui/src/containers/popup.rs
```diff
@@ -69,11 +70,12 @@ impl PopupAnchor {
 }
 
 /// Determines popup's close behavior
-#[derive(Clone, Copy, PartialEq, Eq)]
+#[derive(Clone, Copy, PartialEq, Eq, Default, Debug)]
 pub enum PopupCloseBehavior {
     /// Popup will be closed on click anywhere, inside or outside the popup.
     ///
     /// It is used in [`crate::ComboBox`].
+    #[default]
```
emilk, line 78:
```suggestion
    /// It is used in [`crate::ComboBox`] and in menus.
    #[default]
```

crates/egui/src/context.rs
```diff
@@ -1200,12 +1199,18 @@ impl Context {
...
-            viewport
-                .this_pass
-                .widgets
-                .get(id)
-                .or_else(|| viewport.prev_pass.widgets.get(id))
-                .copied()
+            let rect = viewport.this_pass.widgets.get(id);
+            if let Some(rect) = rect {
+                if rect.rect != Rect::NOTHING {
+                    Some(*rect)
```
emilk, line 1205: Maybe add a comment here about why we do this.

Also the fact that we check against `Rect::NOTHING` here is a bit brittle (what if we change the default in… wherever it is set).

Using something like `rect.is_positive() && rect.is_finite()` should be more robust.

crates/egui/src/menu.rs
```diff
@@ -146,7 +147,7 @@ pub fn menu_image_button<R>(
 /// Opens on hover.
 ///
 /// Returns `None` if the menu is not open.
-pub(crate) fn submenu_button<R>(
+pub fn submenu_button<R>(
```
emilk, line 150: why are these `pub` now, when the whole module is deprecated?

crates/egui/src/ui.rs
```diff
@@ -1237,6 +1237,17 @@ impl Ui {
             .get_downcast(ClosableTag::NAME)
             .is_some_and(|tag: &ClosableTag| tag.should_close())
     }
+
+    /// Should this [`Ui`] or any of its parents close?
+    pub fn will_close(&self) -> bool {
```
emilk, line 1242: `should_close`  and `will_close` are very easy to confuse. They should both document what their differences are, and when to use which.

Maybe their names can be improved too?

crates/egui/src/widgets/button.rs
```diff
@@ -25,7 +25,8 @@ use crate::{
 pub struct Button<'a> {
     image: Option<Image<'a>>,
     text: Option<WidgetText>,
-    shortcut_text: WidgetText,
+    right_text: WidgetText,
+    right_text_weak: bool,
```
emilk, line 29: Why not use the `WidgetText::weak` flag that's already there?

crates/egui/src/widgets/button.rs
```diff
@@ -183,7 +185,15 @@ impl<'a> Button<'a> {
     /// The text can be created with [`crate::Context::format_shortcut`].
     #[inline]
     pub fn shortcut_text(mut self, shortcut_text: impl Into<WidgetText>) -> Self {
-        self.shortcut_text = shortcut_text.into();
+        self.right_text = shortcut_text.into();
+        self.right_text_weak = true;
```
emilk, line 189:
```suggestion
        self.right_text = self.right_text.weak();
```

crates/egui/src/widgets/button.rs
```diff
@@ -183,7 +185,15 @@ impl<'a> Button<'a> {
...
+        self
+    }
+
+    /// Show some text on the right side of the button.
+    #[inline]
+    pub fn right_text(mut self, right_text: impl Into<WidgetText>) -> Self {
```
emilk, line 195: If this was broken out to its own PR it would get its own entry in the changelog 🤷

crates/egui/src/containers/menu.rs
```diff
@@ -0,0 +1,494 @@
...
+
+        let menu_root_response = ui
+            .ctx()
+            .read_response(menu_id)
+            // Since we are a child of that ui, this should always exist
+            .unwrap();
+
+        let hover_pos = ui.ctx().pointer_hover_pos();
+
+        // We don't care if the users is hovering over the border
```
emilk, line 387:
```suggestion
        // We don't care if the user is hovering over the border
```

emilk (approved):

crates/egui/src/menu.rs
lucasmerlin: Since ui.menu_button is now deprecated I made them pub so the old menu can still be used if needed. 
We could also just delete the old menu implementation. But without these being public, the old menu would be unusable.

crates/egui/src/widgets/button.rs
lucasmerlin: Oooh nice I didn't realize that was a thing

crates/egui/src/containers/popup.rs
lucasmerlin: Added a `StyleModifier` struct

lucasmerlin: Ill keep the function as `style` though, since Into<StyleModifier> is also implemented for `Style`

crates/egui/src/widgets/button.rs
lucasmerlin: Mentioned it in the PR description, hopefully that is enough to remember to add this to the changelog during the release

crumblingstatue:
⚠️ Breaking: This technically breaks popup dialogs nested inside menus, for example a color edit button
image

Since they use the same popup mechanism, and there can only be one popup, they will close the menu (and likely themselves in the process).

I don't know if this should be counted as a regression, or never supported in the first place, but a decision should be made regarding this, before merging this PR.

lucasmerlin:
@crumblingstatue good catch. I want to support opening multiple/nested popups which should fix this. But I'll implement that in a separate PR, since this one is already so big.

PR #5723, opened 2025-02-17
Enhance stability on Windows

rustbasic:
Dear emilk,

Programs built with egui on Windows are terminating every hour on average.
When this commit is applied, it works fine for about 3 to 6 hours on average.

I've been testing it for over 6 months and have submitted multiple PRs since 6 months ago,
but they haven't applied it yet.

Thank you.

### DIFF
--- a/crates/eframe/src/native/run.rs
+++ b/crates/eframe/src/native/run.rs
@@ -88,53 +88,57 @@ impl<T: WinitApp> WinitAppWrapper<T> {
         event_loop: &ActiveEventLoop,
         event_result: Result<EventResult>,
     ) {
+        let mut now = Instant::now();
         let mut exit = false;
         let mut save = false;
 
         log::trace!("event_result: {event_result:?}");
 
-        let combined_result = event_result.and_then(|event_result| {
-            match event_result {
-                EventResult::Wait => {
-                    event_loop.set_control_flow(ControlFlow::Wait);
-                    Ok(event_result)
-                }
-                EventResult::RepaintNow(window_id) => {
-                    log::trace!("RepaintNow of {window_id:?}",);
+        let mut event_result = event_result;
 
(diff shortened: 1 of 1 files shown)


lucasmerlin:
Can you explain why this change makes a difference? And what was changed for which reason? I tried to understand it from the diff but I can't make sense of it.

rustbasic:
> Can you explain why this change makes a difference? And what was changed for which reason? I tried to understand it from the diff but I can't make sense of it.

In Windows, the `run_ui_and_paint()` function is called when it's `EventResult::Wait`. 
Previously, the return value of the `run_ui_and_paint()` function was ignored. 
This is a modification to apply the return value of the `run_ui_and_paint()` function.

crates/eframe/src/native/run.rs
```diff
@@ -88,53 +88,57 @@ impl<T: WinitApp> WinitAppWrapper<T> {
...
+        }
+
+        let combined_result = event_result.map(|event_result| match event_result {
+            EventResult::Wait => {
+                event_loop.set_control_flow(ControlFlow::Wait);
+                event_result
+            }
+            EventResult::RepaintNow(window_id) => {
+                log::trace!("RepaintNow of {window_id:?}",);
+                self.windows_next_repaint_times.insert(window_id, now);
```
emilk, line 118: please revert the change to `now` - it is only adding confusion for no apparent gain

```suggestion
                self.windows_next_repaint_times.insert(window_id, Instant::now());
```

emilk (approved):
The old code ignored the new `EventResult` from the call to `run_ui_and_paint`, so I think the new code makes sense. No idea why it is causing crashes though (and I'm not convinced it does, but I don't have a Windows machine 🤷)

emilk (approved):

PR #5736, opened 2025-02-21
fixed order of assignments in available_from_cursor_max_rect

Icekey:
I changed the order of assignments in available_from_cursor_max_rect so that the same value is not checked twice in the min or max calls. 
I am not sure if this fixes anything or if the extra assignment can be removed. 

I also find it kinda weird that the match arms always leave one part of the "avail" rect untouched, but assign another twice. Is that just for formatting or is it a bug?


emilk (commented):
I rather not change code that is working, unless you have something specific you're fixing

PR #5743, opened 2025-02-24
Mark all keys as up if the app loses focus

emilk:
In Rerun, pressing `Cmd+S` brings up a save dialog using `rfd`, but we get not key-up event for the `S` key (in winit).
This leads to `S` being mistakenly marked as down when we switch back to the app.

This PR takes the safe route and marks all keys as up when an egui app loses focus.

* Tested with https://github.com/rerun-io/rerun/pull/9103


lucasmerlin (approved):
Nice!

PR #5745, opened 2025-02-25
Fix Cargo-Machete Workflow

bircni:
As cargo machete now uses rust edition2024 we cannot use the older rust version - we need to install the newest on the runner

### DIFF
--- a/.github/workflows/cargo_machete.yml
+++ b/.github/workflows/cargo_machete.yml
@@ -7,6 +7,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Machete
-        run: cargo install cargo-machete --locked && cargo machete
+        run: cargo install cargo-machete && cargo machete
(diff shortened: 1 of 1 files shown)


bircni:
We could also try to switch to using the cargo-machete action

```
- name: Machete
        uses: bnjbvr/cargo-machete@main
```

.github/workflows/cargo_machete.yml
```diff
@@ -6,7 +6,13 @@ jobs:
   cargo-machete:
     runs-on: ubuntu-latest
     steps:
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: 1.85
+      - name: Machete install
+        run: cargo install cargo-machete
```
lucasmerlin, line 13:
```suggestion
        run: cargo install cargo-machete --locked
```
I think we should keep the --locked

lucasmerlin (approved):
Nice!

lucasmerlin:
It'd also be nice to try https://github.com/Boshen/cargo-shear, which seems to be a new fancier implementation

PR #5748, opened 2025-02-27
Use enum-map serde feature only when serde is enabled

tyssyt:


lucasmerlin (approved):
Nice!

PR #5757, opened 2025-03-03
impl AsRef<[u8]> for FontData

StratusFearMe21:

This PR implements `AsRef<[u8]>` for `FontData`. This would allow it to be passed into `fontdb`'s [`Source`](https://docs.rs/fontdb/0.16.2/fontdb/enum.Source.html) type, and would thus allow `egui` and `cosmic_text` to share font data with eachother

### DIFF
--- a/crates/epaint/src/text/fonts.rs
+++ b/crates/epaint/src/text/fonts.rs
@@ -144,6 +144,12 @@ impl FontData {
     }
 }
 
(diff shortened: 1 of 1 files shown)


lucasmerlin (approved):
Nice! 

> This would allow egui and cosmic_text to share font data with eachother

Note that this would have been possible previously as well though, since `FontData::font` is public

PR #5776, opened 2025-03-10
Add `OutputCommand::SetPointerPosition` to set mouse position

s-nie:
This is not supported on the web and not yet on Wayland.

~~I also had to update `ring` and add an exception for `paste` being unmaintained.~~ Has since been updated on master.

### DIFF
--- a/crates/egui/src/data/output.rs
+++ b/crates/egui/src/data/output.rs
@@ -96,7 +96,7 @@ pub enum OutputCommand {
     /// Open this url in a browser.
     OpenUrl(OpenUrl),
 
-    /// Set the mouse cursor position if the platform supports it.
+    /// Set the mouse cursor position (if the platform supports it).
     SetPointerPosition(emath::Pos2),
 }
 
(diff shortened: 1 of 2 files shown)


lucasmerlin (approved):
Nice!

crates/egui/src/data/output.rs
```diff
@@ -95,6 +95,9 @@ pub enum OutputCommand {
 
     /// Open this url in a browser.
     OpenUrl(OpenUrl),
+
+    /// Set the cursor position if the platform supports it.
+    SetCursorPosition(emath::Pos2),
```
emilk, line 100: We call this "pointer" in the input API, to distinguish it from e.g. text cursors. 

```suggestion
    /// Set the mouse cursor position (if the platform supports it).
    SetPointerPosition(emath::Pos2),
```

emilk (changes requested):

emilk (approved):

PR #5777, opened 2025-03-10
Add a toggle for the compact menu style

s-nie:
Menus currently have their own style that removes outlines and backgrounds. This is nice if the menu only contains buttons. However if the menu contains other widgets, e.g. a drag value, the style change makes it quite difficult to identify such widgets. This is a simple way to make this configurable.

### DIFF
--- a/crates/egui/src/menu.rs
+++ b/crates/egui/src/menu.rs
@@ -76,11 +76,13 @@ impl std::ops::DerefMut for BarState {
 }
 
 fn set_menu_style(style: &mut Style) {
-    style.spacing.button_padding = vec2(2.0, 0.0);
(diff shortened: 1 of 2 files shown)


s-nie:
> I think this makes sense, but I think it would be even better if users could globally customize the menu style to their liking. That would mean `Style` would have to hold a `StyleModifier`, is that weird? @emilk what do you think? I think performance wise it would be fine, since StyleModifier is an `Arc`

Sure, we could also make it more customizable. If `StyleModifier` held a function however, that would disqualify it from being serialized, similar to the `NumberFormatter` field.

lucasmerlin (approved):
Thanks!

PR #5779, opened 2025-03-11
Rename `should_propagate_event` & add `should_prevent_default`

th0rex:

Currently eframe [calls `prevent_default()`](https://github.com/emilk/egui/blob/962c7c75166dff3369d20675bcfd527d3287149f/crates/eframe/src/web/events.rs#L307-L369) for all copy / paste events on the [*document*](https://github.com/emilk/egui/blob/962c7c75166dff3369d20675bcfd527d3287149f/crates/eframe/src/web/events.rs#L88), making embedding an egui application in a page (e.g. an react application) hard (as all copy & paste functionality for other elements on the page is broken by this).

I'm not sure what the motivation for this is, if any.

This commit / PR adds a callback (`should_prevent_default`), similar to `should_propgate_event`, that an egui application can use to overwrite this behavior. It defaults to returning `true` for all events, to keep the existing behavior.

I call `should_prevent_default` in every place that `should_propagate_event` is called (which is not all places that `prevent_default` is called!). I'm not sure for the motivation of not calling `should_propagate_event` everywhere that `stop_propagation` is called, but I kept that behavior for the `should_prevent_default` callback too.

Please let me know if I'm missing some existing functionality that would allow me to do this, or if there's a reason that we don't want applications to be able to customize this (i.e. if there's a reason to always `prevent_default` for all copy / paste events on the whole document)

### DIFF
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -503,6 +503,12 @@ pub struct WebOptions {
     /// [`stopPropagation`](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation)
     /// is called on every event.
     pub should_propagate_event: Box<dyn Fn(&egui::Event) -> bool>,
+
+    /// Whether the web event corresponding to an egui event should have `prevent_default` called
+    /// on it or not.
+    ///
+    /// Defaults to true.
+    pub should_prevent_default: Box<dyn Fn(&egui::Event) -> bool>,
 }
 
 #[cfg(target_arch = "wasm32")]
@@ -520,6 +526,7 @@ impl Default for WebOptions {
             dithering: true,
 
             should_propagate_event: Box::new(|_| false),
+            should_prevent_default: Box::new(|_| true),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


crates/eframe/src/epi.rs
```diff
@@ -503,6 +503,12 @@ pub struct WebOptions {
     /// [`stopPropagation`](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation)
     /// is called on every event.
     pub should_propagate_event: Box<dyn Fn(&egui::Event) -> bool>,
+
+    /// Whether the web event corresponding to an egui event should have `prevent_default` called
+    /// on it or not.
+    ///
+    /// Defaults to true.
+    pub should_prevent_default: Box<dyn Fn(&egui::Event) -> bool>,
```
lucasmerlin, line 511: It's a bit weird that one is negated and the other isn't. Since `should_do_default` or `should_not_prevent_default` is weird, maybe we should rename `should_propagate_event` to `should_stop_propagation` which would also be consistent with the name of the js function.

lucasmerlin (changes requested):

crates/eframe/src/epi.rs
emilk: I agree - rename one or the other please!

th0rex: Sorry for the delay, I was a bit busy and then forgot about this PR. Added a commit renaming `should_propagate_event` to `should_stop_propagation`.

lucasmerlin (approved):
Nice, thanks!

PR #5782, opened 2025-03-11
Make force_set_min_rect public

zmarlon:
I want to implement my own collapsing header based on the existing collapsing header. To do this, I need access to the function force_set_min_rect public

### DIFF
--- a/crates/egui/src/ui.rs
+++ b/crates/egui/src/ui.rs
@@ -877,7 +877,7 @@ impl Ui {
     }
 
     /// Used for animation, kind of hacky
-    pub(crate) fn force_set_min_rect(&mut self, min_rect: Rect) {
+    pub fn force_set_min_rect(&mut self, min_rect: Rect) {
         self.placer.force_set_min_rect(min_rect);
     }
 


crates/egui/src/ui.rs
```diff
@@ -877,7 +877,7 @@ impl Ui {
     }
 
     /// Used for animation, kind of hacky
-    pub(crate) fn force_set_min_rect(&mut self, min_rect: Rect) {
+    pub fn force_set_min_rect(&mut self, min_rect: Rect) {
```
lucasmerlin, line 880: If we make it public we should probably mark it as `#[doc(hidden)]`, since users probably shouldn't rely on it

emilk (approved):

crates/egui/src/ui.rs
emilk: I agree with the `#[doc(hidden)]`, and/or adding a `/// WARNING: advanced usage`

emilk, line 880:
```suggestion
    /// WARNING: advanced usage
    #[doc(hidden)] 
    pub fn force_set_min_rect(&mut self, min_rect: Rect) {
```

PR #5783, opened 2025-03-12
Bump accesskit to 0.18 and make it a workspace dependency

valadaptive:
This can't be merged until [kittest's accesskit is bumped](https://github.com/rerun-io/kittest/pull/9) ~~and [a new version of rfd is released](https://github.com/PolyMeilex/rfd/pull/240)~~.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,7 +85,7 @@ glutin = { version = "0.32.0", default-features = false }
 glutin-winit = { version = "0.5.0", default-features = false }
 home = "0.5.9"
 image = { version = "0.25", default-features = false }
-kittest = { git = "https://github.com/rerun-io/kittest", rev = "5803de3" }
+kittest = { version = "0.1.0", git = "https://github.com/rerun-io/kittest", branch = "main" }
 log = { version = "0.4", features = ["std"] }
 nohash-hasher = "0.2"
 parking_lot = "0.12"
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -83,7 +85,7 @@ glutin = { version = "0.32.0", default-features = false }
 glutin-winit = { version = "0.5.0", default-features = false }
 home = "0.5.9"
 image = { version = "0.25", default-features = false }
-kittest = { version = "0.1" }
+kittest = { git = "https://github.com/valadaptive/kittest", rev = "9b2b8cd" }
```
lucasmerlin, line 88: For now we can use kittest from master. I'll release a new version with the next egui release.

lucasmerlin (approved):
Thank you!

Cargo.toml
valadaptive: Updated kittest to the latest git version. cargo-deny is complaining about a wildcard dependency; not sure why.

lucasmerlin: Hmm, no idea 🤔 Maybe it works when added as a patch instead?

valadaptive: Alright, it seems that cargo-deny doesn't allow git dependencies unless [the `allow-wildcard-paths` option](https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html#the-allow-wildcard-paths-field-optional) is enabled, and doesn't allow git dependencies in *public* crates regardless of what that option is set to. Since egui_kittest is a public crate, cargo-deny will fail until a new version of kittest is published that we can switch to.

lucasmerlin:
Specifying the kittest version seems to do the trick

Issue #3378, opened 2023-09-23
Cosmic Text for font rendering


ElhamAryanpur:
As discussed in #1016 , I have done some testing on [cosmic-text](https://github.com/pop-os/cosmic-text).

First of all, they have some [docs](https://pop-os.github.io/cosmic-text/cosmic_text/) that I've been exploring. And I was testing their [example](https://github.com/pop-os/cosmic-text/tree/main/examples/editor-orbclient). They do have `no_std` and `wasm` support too according to the `Cargo.toml` they have.

I've noticed that the way it renders, is by drawing rectangles. Which is something I've seen for the first time to be honest. You can check it [here](https://github.com/pop-os/cosmic-text/blob/e2adc1e8da1fd8196224254e02f28f815e47d4e5/examples/editor-orbclient/src/main.rs#L99-L107). It gives a x,y,width,height, and color which is usually just color predefined and alpha channel being different for aliasing and stuff.

There is also another method if rectangles aren't possible: [Swash Image](https://pop-os.github.io/cosmic-text/cosmic_text/struct.SwashCache.html#method.get_image) which basically returns an image bytes to be rendered instead of individually creating rectangles. It requires some things I couldn't implement myself to be honest.

I was able to sort of hack the rectangle method in my engine
![image](https://github.com/emilk/egui/assets/35463470/e03ad5b8-9eda-44a9-be32-0ff52f07e589)

It had... not good results to be honest
![image](https://github.com/emilk/egui/assets/35463470/d9b8a7ef-5d91-4ffd-a345-5e900583d980)

Although maybe that's on me for having some issues with the engine as their examples do work and work very well. But yeah, that's my findings so far on cosmic text.

thomaskrause:
I also experimented with cosmic text but went with hacking some completely inefficient code in the `tessellate_text` function of the epaint `Tessellator`. 

```rust

pub fn tessellate_text(&mut self, text_shape: &TextShape, out: &mut Mesh) {
        let TextShape {
            pos: galley_pos,
            galley,
            underline,
            override_text_color,
            angle,
        } = text_shape;

        if galley.is_empty() {
            return;
        }

        if galley.pixels_per_point != self.pixels_per_point {
            eprintln!("epaint: WARNING: pixels_per_point (dpi scale) have changed between text layout and tessellation. \
                       You must recreate your text shapes if pixels_per_point changes.");
        }

        for row in &galley.rows {
            let metrics =
[...]

PR #5784, opened 2025-03-13
[WIP] Render text with Parley

valadaptive:
* Closes https://github.com/emilk/egui/issues/3378

This is very WIP and currently a proof-of-concept. I'm sharing it here so I can ask questions about the API and have something to point to if I need to merge any supporting changes.

Right now, all I have done is basic text rendering. Pretty much everything else still needs to be implemented.

You can see in the commit history that my first pass at this did use Cosmic Text. However, Parley's API surface seems to mesh much better with how egui wants to do things. I ran into a lot of limitations with Cosmic Text that will have to be solved before it's ready for this use case:

<details>
<summary>Cosmic Text issues</summary>

- Lack of documentation for many items.
- Lack of support for inline boxes, needed to implement things like `leading_space` in `LayoutSection`.
- Blank lines *always* use the line height from the metrics on the `Buffer`, not the ones in any particular text span. This means that in e.g. the EasyMark Editor example, where there's mixed-size text, you need to pass the "default" font all the way down, and even then, you might still want different blank lines to have different heights (https://github.com/pop-os/cosmic-text/issues/364).
- Setting properties on `Buffer` triggers one relayout per property being set (https://github.com/pop-os/cosmic-text/issues/280).
- I never really got the hang of the "scroll" abstraction on `Buffer`.
- Font fallback seems to be technically implemented but untested. For instance, it prioritizes fonts with a similar weight over fonts from the same family. It took me a while to realize why everything was being rendered in monospace: cosmic-text was choosing Hack Regular over Ubuntu Light.
- The library seems overall quite allocation-heavy (https://github.com/pop-os/cosmic-text/issues/308).
- No built-in AccessKit support. (I assume it's possible to implement this on top of Cosmic Text since iced does so, but Parley has this built-in for the text editing APIs)

</details>

There are some drawbacks to Parley as well:
- They have a very aggressive MSRV policy--they can bump it at any time, including in patch releases. This means we need to pin the version of Parley we depend on. They also recently converted the codebase to Rust 2024, meaning it requires Rust 1.85 or newer.
- Parley incorporates large amounts of Google Fonts' text stack. Large tech companies, and Google especially, tend to prioritize their own use cases above others'. They have the right to do so, of course, but it could still mean a lot more work for us.
  - Although Swash now incorporates parts of the Google Fonts stack (skrifa) as well, so Cosmic Text would also leave us dependent on Google.
- Shaping is not completely implemented yet (it uses Swash's implementation).
- It also lacks some features that will need to be contributed (mentioned in `parley-todo.md`).

There is no guarantee that I'll finish this PR; if you ping me and I respond with something like "sorry, I'll get around to it next week" a few times in a row, feel free to take over it yourself.

I'm tracking progress in more detail via `parley-todo.md` in the root folder, but as an overview, this work can be split into three main parts:
- [ ] Rendering
  - [x] Implement very basic rendering
  - [ ] Implement all rendering features (e.g. ellipsis truncation, `max_rows`)
  - [ ] Rework the glyph texture atlas API
- [ ] Editing
  - [ ] Retool the existing text editing API to look more like Parley's (e.g. we probably don't need three cursor types)
  - [ ] Abstract away parts of the text editing API that reach into the internals
  - [ ] Swap it all out for Parley's API
  - [ ] Ensure that AccessKit integration works
- [ ] Styling
  - [ ] Reimplement `FontDefinitions` on top of Parley
  - [ ] Rework the font API since Parley supports a pretty broad range of features


xorgy:
Regarding the risk of diverging use cases for the upstream google-funded components (fontations in particular):
the use case includes basically everything people want to do with fonts, because these libraries are intended to serve the needs of web browsers (Chromium in particular, which exposes more font/shaping functionality than basically any other application).

Swash itself is likely, in time, to be replaced in Parley with RustyBuzz/Harfruzz (when the port to fontations from ttf-parser is done), and HarfBuzz (along with its ports) is again intended to cover the overwhelming majority of shaping use cases, though conceivably could fall short on very exotic shaping requirements (e.g. shaping byzantine music notation or con-scripts).

emilk:
This is very exciting - thanks for working on this! Let me know if I can help somehow, e.g. if you want to talk things through.

waywardmonkeys:
> They have a very aggressive MSRV policy--they can bump it at any time, including in patch releases. This means we need to pin the version of Parley we depend on. They also recently converted the codebase to Rust 2024, meaning it requires Rust 1.85 or newer.

Thanks for the feedback!

I've just landed a revert of that in https://github.com/linebender/parley/pull/307 so now Parley's next release will still have an MSRV of 1.82 as the current release does.

valadaptive:
I'm now mainly working on the text styling API. It's a bit hard to untangle everything, since there are several different places where text style properties are stashed, each exposing a different API surface:
- `FontId` is what lets you select a specific font and size.
- `TextFormat` contains a `FontId` as well as some formatting-specific things like color, background, italics, and decorations.
- `TextStyle` associates names to `FontId`s (either predetermined names like `TextStyle::Small` or user-defined ones via `TextStyle::Name(_)`).
- `FontSelection` is either a `FontId` or a `TextStyle`.
- `RichText` is kinda like `TextFormat` except with an override/fallback mechanism? When converting one into a `TextFormat`, you pass in a "fallback font" (a `FontSelection`), but if either the `RichText`'s `text_style` field or the containing UI's `Style::override_text_style` are set, they will be used to select the font instead.

Parley unlocks some new options for font styling, and also changes the semantics of font styling a bit:
- Instead of a single "font" like in `FontId`, you specify a font *family*. Depending on style properties like weight, italics, etc. this may be resolved to one or many actual font files.
- There are more styling options--"real" italics if supported by the font family, font weight, font width (if supported), and OpenType font variation settings.
- Instead of specifying just one font family, Parley's APIs actually take a `FontStack`, which allows for font fallback.

I ran into some issues when trying to map the existing text style API onto Parley. First up, which new font properties should be part of `FontId`? My initial thought was anything that affects layout, so that would be font family/stack, size, slant, weight, width, and OpenType variation and feature settings.

However, italics are currently in `TextFormat` and make sense there. Does that mean that weight should also be in `TextFormat`? But the default UI font is Ubuntu *Light*. Maybe specify a default font weight as part of `FontId` and allow `TextFormat` to "override" it?

If we take that to its conclusion, why not just flatten *all* the font settings then? Have one big "all of the font style settings" type, and make it overridable CSS-style. I'm not sure yet what this API would look like exactly.

I'm also not sure which parts of this new API I can punt on right now, and which ones must be implemented.

emilk:
My inclination is to punt on as many things as you can, and save it for a future PR. Better to break things into small pieces.

The existing API is indeed all over the place, and partially a result of legacy, and partially a result of technical limitation (e.g. the fake italics). Worth mentioning is that font fallbacks is currently implemented in egui [here](https://github.com/emilk/egui/blob/390e0bfc1e3c55bf37dd41f093e989612c14cf76/crates/epaint/src/text/fonts.rs#L252-L265).

Moving italics into `FontId` makes sense to me, as it is about selecting a specific font. As does adding `weight` to it at some point.

The rest of the attributes in `TextFormat` should probably stay there, as they are about how to _apply_ that font.

valadaptive:
Things are progressing well; just need to land some Parley changes and clean up the code. There are some issues that might change the API more significantly:
- `Galley` needs to be serializable via serde, but Parley's types currently are not. I theoretically *could* implement serde support for things like `parley::Layout`, but I'm wondering if it even makes sense--there are a lot of things like font family IDs that are based on auto-incrementing counters and would thus not be meaningful through a serialization round-trip.
  This is sorta already an issue with the current code--AFAIK, the text meshes themselves reference UV positions in the font atlas that are also not guaranteed to be stable for any length of time, let alone a serialization round-trip.
  Is there a serialization behavior that makes sense here?

- `LayoutJob::break_on_newline` is probably not implementable without some changes to Parley. Is there an alternative behavior that would work for singleline `TextEdit`?

- `FontTweak::scale` says it does not affect the layout, but it *does* affect the advance width of the characters. Parley lets you change the font size for a given bit of text, but because it lets you select multiple font families, there's no way to know what font a given character will actually fall back to, and hence no way to adjust the font size to match. Is it OK for it to not affect the characters' advance width?

Also, a question: is there any actual difference between `FontTweak::y_offset_factor` and `FontTweak::baseline_offset_factor`? Neither appears to actually affect the layout, and from playing with the "Font Tweak" UI in the settings, they both appear to do exactly the same thing. `FontTweak::baseline_offset_factor` was added in https://github.com/emilk/egui/pull/2724; maybe the behavior has changed since then?

emilk:
> * `Galley` needs to be serializable via serde

I agree with you here - let's drop serde-support for `Galley`. I was using it for [eterm](https://github.com/emilk/eterm) way back, but at the moment it makes little sense (as you point out).

---


> * `LayoutJob::break_on_newline` is probably not implementable without some changes to Parley. Is there an alternative behavior that would work for singleline `TextEdit`?

To explain `break_on_newline`, consider this code:

```rs
let mut start_text = String::new("Already \n multiline");
ui.text_edit_singleline(&mut start_text);
```

How should egui handle the existing `\n`, which clashes with the desire of the programmer to have the `TextEdit` be single-line? The answer right now is that egui sets `break_on_newline: false` which renders newlines as the replacement glyph, `�`.

Possible alternative solutions include:
A) Replace any `'\n` with `�` before passing the string to Parely (`if break_on_newline == false`)
B) Have singeline `TextEdit`s replace any `\n` with `�` in the input buffer (mutating the users string)
C) Ignore the problem (remove `break_on_newline` and say that having newlines in the text buffer of a `TextEdit` will just lead to weird behavior, so "don't do that"

Since this is such a niche corner case, I'm happy with any of these solutions.

---

> * `FontTweak::scale` says it does not affect the layout, but it _does_ affect the advance width of the characters. (…) Is it OK for it to not affect the characters' advance width?

Yes. `FontTweak::scale` is currently used to make the _sizes_ of the emojis in the different default fonts match up, so that when setting fotn height =12 pts (for instance), the emojis look the same (even though one of the fonts have big emohis, and the other small ones). I don't think advance width matters. Do what feels best :)

---

> Also, a question: is there any actual difference between `FontTweak::y_offset_factor` and `FontTweak::baseline_offset_factor`?

These were added to match the baseline of normal text and emojis (iirc). As long as we can still do that reasonably well, do whatever change you want 👍 

---

In summery: I'm fine with this breaking API, behavior, and rendering in small ways. The wins will outweigh those minor annoynaces imho!

parley-todo.md
```diff
@@ -0,0 +1,148 @@
...
+  - [x] Letter spacing
+    - Why was this here? It's been implemented for a while
+  - [x] Hinting enable/disable
+    - [x] Maybe a global setting and also an override in FontTweak?
+  - [ ] If file size isn't an issue, ship variable fonts in epaint-default-fonts
+  - [x] Make sure to test with syntect disabled (there's some old style-heavy code that is cfg'd out with syntect enabled)
+  - [ ] Smoothe out the janky parts of the new API
+    - [ ] For mixed-DPI purposes, and because we don't need to store the FontStore as a mutex, ctx.fonts() now returns a "fonts view" that's technically read/write. But there are no operations that *semantically* modify the fonts from it
+    - [ ] FontStore and Fonts are different and we should just expose them separately instead of passing through all the FontStore methods onto Fonts
+  - [ ] Work around https://github.com/jslegers/emoji-icon-font/issues/18 / https://github.com/emilk/egui/issues/1284
```
danielhjacobs, line 112: I opened https://github.com/jslegers/emoji-icon-font/pull/19 to fix this and it was merged.

danielhjacobs: So I assume https://github.com/emilk/egui/blob/d78fc39386e28ec7a18bd48a28a688177f090bd4/crates/epaint_default_fonts/fonts/emoji-icon-font.ttf needs to be replaced with the new ttf file from https://github.com/jslegers/emoji-icon-font/blob/master/fonts/emoji.ttf

danielhjacobs: https://github.com/emilk/egui/pull/5877

danielhjacobs: emilk fixed this in https://github.com/emilk/egui/pull/7067

.github/workflows/deploy_web_demo.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         with:
           profile: minimal
           target: wasm32-unknown-unknown
-          toolchain: 1.84.0
+          toolchain: 1.85.0
```
emilk, line 42: Do we still need 1.85? If so, I can update `egui main` so the diff of this PR is smaller 🤷

crates/epaint/Cargo.toml
```diff
@@ -74,13 +76,19 @@ _override_unity = []
 emath.workspace = true
 ecolor.workspace = true
 
-ab_glyph = "0.2.11"
+#parley = { version = "0.3", features = ["system"] }
+parley = { path = "../../../parley/parley", features = ["system"] }
```
emilk: It would be nice if you could publish this to your own fork:

```suggestion
parley = { path = "https://github.com/valadaptive/swash", branch = "whatever", features = ["system"] }
```

That way me and others can check out your PR and give it a spin

valadaptive: I've updated the dependency to point to my fork of parley. Sorry I haven't been working on this recently; I've been quite burnt out.

I think the "last 90%" involves a lot of stuff that needs to be implemented in Parley itself, and some of it may require either an API rework in Parley or a somewhat jankier API surface to be exposed from it.

emilk: No stress! Please don't burn yourself out for this

NfNitLoop:
Was excited to learn about this PR today! (bold! better italics! Color emoji! variable fonts! ❤️❤️) 

I briefly tried out the preview here and saw a minor issue w/ rendering source code links:
https://egui-pr-preview.github.io/pr/5784-parley-2

Screenshot 2025-08-11 at 9 42 49 AM

If there are any pieces you can think would be easy to independently tackle to unblock this, I'd be happy to help. I have the same username in the Discord if you'd like to ping me there.

torokati44:
https://github.com/linebender/parley/releases/tag/v0.6.0 :eyes: 

> Parley now uses [HarfRust](https://github.com/harfbuzz/harfrust) rather than [Swash](https://github.com/dfrg/swash). This means that
> Parley now has production-quality shaping for all scripts and can be recommended for general usage.

valadaptive:
An update on this: I was supposed to receive funding to work on a rework of Parley's API to make it more suitable for this project and others like it, but that never materialized. At this point, I'm more likely to just go directly to the underlying APIs (skrifa, HarfRust, and whatever rasterizer I end up going with) and wire those into egui.

To an outside observer, it probably looks like this PR is 90% of the way there and just needs that last 10%, but unfortunately that last 10% is basically impossible considering how Parley's API is structured. It *really* wants to be given a full rectangle that it can arrange text inside, however it wants, and egui really doesn't work with that model. This PR uses a lot of workarounds to try to hide that, but they aren't reliable. There have been a few Parley proposals to make its API even more convoluted in order to support some of this stuff, but I have doubts about those working for egui.

BergmannAtmet:
@valadaptive 

Thank you for working on this.

Mentioning/Pinging this PR when/if your new approach starts materializing would be appreciated, so those of us watching from here can follow along.

I'm holding off trying GUI in Rust, despite using it for almost everything else, because I want to experiment with all the big-name crates first, and *fancy* text shaping is a requirement for me.

PR #5785, opened 2025-03-13
Simplify the text cursor API

valadaptive:
* Closes N/A, but this is part of https://github.com/emilk/egui/issues/3378

Other text layout libraries in Rust--namely, Parley and Cosmic Text--have one canonical text cursor type (Parley's is a byte index, Cosmic Text's also stores the line index). To prepare for migrating egui to one of those libraries, it should also have only one text cursor type. I also think simplifying the API is a good idea in and of itself--having three different cursor types that you have to convert between (and a `Cursor` struct which contains all three at once) is confusing.

After a bit of experimentation, I found that the best cursor type to coalesce around is `CCursor`. In the few places where we need a paragraph index or row/column position, we can calculate them as necessary.

I've removed `CursorRange` and `PCursorRange` (the latter appears to have never been used), merging the functionality with `CCursorRange`. To preserve the cursor position when navigating row-by-row, `CCursorRange` now stores the previous horizontal position of the cursor.

I've also removed `PCursor`, and renamed `RowCursor` to `LayoutCursor` (since it includes not only the row but the column). I have not renamed either `CCursorRange` or `CCursor` as those names are used in a lot of places, and I don't want to clutter this PR with a bunch of renames. I'll leave it for a later PR.

Finally, I've removed the deprecated methods from `TextEditState`--it made the refactoring easier, and it should be pretty easy to migrate to the equivalent `TextCursorState` methods.

I'm not sure how many breaking changes people will actually encounter. A lot of these APIs were technically public, but I don't think many were useful. The `TextBuffer` trait now takes `&CCursorRange` instead of `&CursorRange` in a couple of methods, and I renamed `CCursorRange::sorted` to `CCursorRange::sorted_cursors` to match `CursorRange`.

I did encounter a couple of apparent minor bugs when testing out text cursor behavior, but I checked them against the current version of egui and they're all pre-existing.

### DIFF
--- a/crates/egui/src/lib.rs
+++ b/crates/egui/src/lib.rs
@@ -471,7 +471,7 @@ pub use epaint::{
 };
 
 pub mod text {
-    pub use crate::text_selection::{CCursorRange, CursorRange};
+    pub use crate::text_selection::CCursorRange;
     pub use epaint::text::{
         cursor::CCursor, FontData, FontDefinitions, FontFamily, Fonts, Galley, LayoutJob,
         LayoutSection, TextFormat, TextWrapping, TAB_SIZE,
--- a/crates/egui/src/text_selection/accesskit_text.rs
+++ b/crates/egui/src/text_selection/accesskit_text.rs
@@ -17,8 +17,8 @@ pub fn update_accesskit_for_text_widget(
         let parent_id = widget_id;
 
         if let Some(cursor_range) = &cursor_range {
-            let anchor = galley.from_ccursor(cursor_range.secondary).rcursor; // TODO(valadaptive): figure this out
-            let focus = galley.from_ccursor(cursor_range.primary).rcursor; // TODO(valadaptive): figure this out
+            let anchor = galley.rcursor_from_ccursor(cursor_range.secondary);
+            let focus = galley.rcursor_from_ccursor(cursor_range.primary);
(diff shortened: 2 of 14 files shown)


lucasmerlin:
Awesome! This looks great, and I also couldn't find anything unusal when testing.

lucasmerlin:
Found a small bug in the markdown editor:

https://github.com/user-attachments/assets/f9b0d5cb-dba4-4b51-a400-ac7c132eddc5

valadaptive:
There were some old checks remaining from the previous "cursor up/down one row" code that existed to preserve the cursor's horizontal position, but don't make sense now that it's stored explicitly. Removing those seems to have fixed that issue.

lucasmerlin (approved):
Awesome, great work!

valadaptive:
A note if this makes it into a release before the Parley changes do: this won't be the last change to the text cursor API, since the Parley API reworks will change the text cursors to be byte-index-based instead of character-index-based.

It's probably going to be a bit annoying for users of the text cursor APIs to adjust to two reworks in a row, but giving them a heads-up about it should make things easier.

(The cursors being character-index-based only really provide an advantage in an ab_glyph world where one character == one glyph. Once we move away from that, byte indices will be a lot simpler.)

Issue #5796, opened 2025-03-13
Erratic selection in `TextEdit`


hankjordan:
**Describe the bug**
The Galley's `cursor_from_pos` function does not give any interaction margin around the first and last lines for the cursor selection, causing `TextEdit` to behave unexpectedly.

**To Reproduce**
1. Create a `TextEdit` with margin (or `min_size`)
2. Attempt to highlight the text
3. Moving the cursor even slightly above or below the text will jump it to the beginning or end, respectively.

**Expected behavior**
1. Create a `TextEdit` with margin (or `min_size`)
2. Attempt to highlight the text
3. Moving the cursor a moderate amount above or below the text will jump it to the beginning or end, respectively.

**Desktop:**
 - OS: Linux
 - Browser: N/A
 - Version: 31.1

**Reproduction Example**

```rust
use eframe::egui;

fn main() -> eframe::Result {
    env_logger::init();
    let options = eframe::NativeOptions {
        viewport: egui::ViewportBuilder::default().with_inner_size([320.0, 240.0]),
        ..Default::default()
    };
    eframe::run_native(
        "My egui App",
        options,
        Box::new(|_| Ok(Box::<MyApp>::default())),
    )
}

struct MyApp {
    name: String,
}

impl Default for MyApp {
    fn default() -> Self {
        Self {
            name: "Arthur".to_owned(),
        }
    }
}

impl eframe::App for MyApp {
    fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
        egui::CentralPanel::default().show(ctx, |ui| {
            ui.add(egui::TextEdit::singleline(&mut self.name).min_size(egui::vec2(0.0, 64.0)));
        });
    }
}
```

PR #5797, opened 2025-03-13
Improve drag-to-select text (add margins)

hankjordan:
Might want to draw from `interaction.interact_radius` style instead of hard-coding the margin, but I didn't want to create a breaking change. If desired, I can follow up with a separate PR to address that concern.

* Closes <https://github.com/emilk/egui/issues/5796>


crates/epaint/src/text/text_layout_types.rs
```diff
@@ -824,12 +824,12 @@ impl Galley {
...
     pub fn cursor_from_pos(&self, pos: Vec2) -> Cursor {
         if let Some(first_row) = self.rows.first() {
-            if pos.y < first_row.min_y() {
+            if pos.y < first_row.min_y() - 5.0 {
                 return self.begin();
             }
         }
         if let Some(last_row) = self.rows.last() {
-            if last_row.max_y() < pos.y {
+            if last_row.max_y() + 5.0 < pos.y {
```
emilk, line 832: Let's give the `5.0` a name (`const VMARGIN: f32` or something) and explain why its there 🙏

emilk (approved):
Thanks!

emilk (approved):
Thanks!

PR #5814, opened 2025-03-16
⚠️ Close popup if `Memory::keep_popup_open` isn't called

juancampa:
Breaking changes:
- When using the Memory::popup state, it's now required to call keep_popup_open each frame or the popup will close.
- Usually handled by the `Popup` struct, but required for custom popups using the state in `Memory` directly

-----

If a popup is abandoned `Memory::popup` would remain `Some`. This is problematic if, for example, you have logic that checks `is_any_popup_open`.

This PR adds a new requirement for popups keeping their open state in `Memory::popup`. They must call `Memory::keep_popup_open` as long as they are being rendered. The recent changes in #5716 make this easy to implement.

Supersedes #4697 which had an awkward implementation

These two videos show a case where a context menu was open when the underlying widget got removed.

Before (`any_popup_open` remains `true`)
Screenshot 2025-03-16 at 18 22 50

After
Screenshot 2025-03-16 at 18 21 14

* Closes https://github.com/emilk/egui/issues/3657

### DIFF
--- a/crates/egui/src/containers/popup.rs
+++ b/crates/egui/src/containers/popup.rs
@@ -525,7 +525,11 @@ impl<'a> Popup<'a> {
                 Some(SetOpenCommand::Toggle) => {
                     mem.toggle_popup(id);
                 }
-                None => {}
+                None => {
+                    if mem.is_popup_open(id) {
+                        mem.keep_popup_open(id);
+                    }
+                }
             });
         }
 
--- a/crates/egui/src/containers/popup.rs
+++ b/crates/egui/src/containers/popup.rs
@@ -526,9 +526,7 @@ impl<'a> Popup<'a> {
                     mem.toggle_popup(id);
                 }
                 None => {
-                    if mem.is_popup_open(id) {
-                        mem.keep_popup_open(id);
-                    }
+                    mem.keep_popup_open(id);
                 }
             });
         }
--- a/crates/egui/src/memory/mod.rs
+++ b/crates/egui/src/memory/mod.rs
@@ -94,7 +94,7 @@ pub struct Memory {
     /// If position is [`None`], the popup position will be calculated based on some configuration
     /// (e.g. relative to some other widget).
     #[cfg_attr(feature = "persistence", serde(skip))]
-    popup: Option<(Id, Option<Pos2>)>,
+    popup: Option<OpenPopup>,
 
     #[cfg_attr(feature = "persistence", serde(skip))]
     everything_is_visible: bool,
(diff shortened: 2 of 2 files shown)


juancampa:
cc @lucasmerlin who seems to be actively working on popups

juancampa:
@lucasmerlin, on a slight tangent, I noticed a couple of potential performance improvements to the new popup system while working on this btw:

 - The `Tooltip` struct contains a `Popup` which makes it relatively bulky struct (248 bytes). Since tooltips are commonly defined but rarely visible, I wonder if it's worth deferring the creation of `Popup` until the tooltip is actually visible.
 - `Popup::from_response` preemptively calls `response.clicked_elsewhere` which is not super heavy but will run a bunch of `Rect::contains` on every click.
 
Fixing the first one would make the second one a non-issue.

To illustrate this point with our application, these green rects show all the `Popup`s created in one frame to handle tooltips. None of them visible or really doing anything.

Screenshot 2025-03-16 at 16 26 37

crates/egui/src/containers/popup.rs
```diff
@@ -525,7 +525,11 @@ impl<'a> Popup<'a> {
                 Some(SetOpenCommand::Toggle) => {
                     mem.toggle_popup(id);
                 }
-                None => {}
+                None => {
+                    if mem.is_popup_open(id) {
+                        mem.keep_popup_open(id);
+                    }
```
lucasmerlin, line 531: nit: I think we can remove the outer if, since this is checked in the keep_popup_open, right?

lucasmerlin (approved):
Nice, I think this makes sense.

lucasmerlin:
About the optimization, best thing I can imagine that wouldn't make the api less ergonomic would be storing the Popup in a Option<Box<Popup>>, and then only setting it to some only if the popup is actually shown.

juancampa:
Feedback addressed 👍. This could be a breaking change for anyone using `open_popup` directly btw. Something to call out in the changelog.

> About the optimization, best thing I can imagine that wouldn't make the api less ergonomic would be storing the Popup in a `Option<Box<Popup>>`, and then only setting it to some only if the popup is actually shown.

Makes sense. I mentioned it since there's been quite a bit of work recently to reduce the size of structs, so yeah this should help 👍

With regards to `widget_clicked_elsewhere`, would it make sense to use `ui.ctx().get_response(widget_id)` to get the widget's response only when needed? Similarly to [how `Ui::response` works](https://github.com/membrane-io/egui/blob/36e339edb8d9ac0c71581ed5aaafde677c0a605c/crates/egui/src/ui.rs#L1098-L1111).

crates/egui/src/memory/mod.rs
```diff
@@ -1083,19 +1116,34 @@ impl Memory {
...
-        self.popup = Some((popup_id, None));
+        self.popup = Some(OpenPopup::new(popup_id, None));
+    }
+
+    /// Popups must call this every frame while open.
+    ///
+    /// This is needed because in some cases popups can go away without `close_popup` being
+    /// called. For example, when a context menu is open and the underlying widget stops
+    /// being rendered.
+    pub fn keep_popup_open(&mut self, popup_id: Id) {
```
emilk, line 1130: I feel like there is a more elegant solution to this problem that I'm not seeing 😆 

We could replace `is_popup_open` with `show_popup_if_open(popup_id, closure)`, but that's a bit annoying too (especially when it comes to lifetimes and locks), so… no.

I'm also not in love with this function name, but again fail to come up with something better.

so… LGTM :)

emilk (approved):

PR #5816, opened 2025-03-17
Add script to update local snapshots from CI

lucasmerlin:
It seems like the thresholds are too low for all tests to pass when snapshots are generated from windows / linux. We need a better solution to this problem, but in the meantime this script should allow contributors to update their snapshots by downloading them from the last CI run.

### DIFF
--- a/scripts/update_snapshots_from_ci.sh
+++ b/scripts/update_snapshots_from_ci.sh
@@ -3,9 +3,11 @@
 # and replaces your existing snapshots with the new ones.
 # Make sure you have the gh cli installed and authenticated before running this script.
 
+set -eu
+
 BRANCH=$(git rev-parse --abbrev-ref HEAD)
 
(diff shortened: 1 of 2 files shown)


scripts/update_snapshots_from_ci.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+# This script searches for the last CI run with your branch name, downloads the test_results artefact
+# and replaces your existing snapshots with the new ones.
+# Make sure you have the gh cli installed and authenticated before running this script.
```
emilk, line 4:
```suggestion
# Make sure you have the gh cli installed and authenticated before running this script.

-set eu
```

Will stop on error and fail with non-zero

emilk (approved):
nice!

emilk:
Is this ready to merge? Have the latest changes been tested?

PR #5820, opened 2025-03-18
Make `close_popup` take an `Id` and add `close_all_popups` 

lucasmerlin:
Ooops, fixes a regression introduced in #5799

### DIFF
--- a/crates/egui/src/memory/mod.rs
+++ b/crates/egui/src/memory/mod.rs
@@ -1103,12 +1103,19 @@ impl Memory {
         self.popup = None;
     }
 
+    /// Close the given popup, if it is open.
+    pub fn close_popup_with_id(&mut self, popup_id: Id) {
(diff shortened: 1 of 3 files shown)


crates/egui/src/memory/mod.rs
```diff
@@ -1103,12 +1103,19 @@ impl Memory {
         self.popup = None;
     }
 
+    /// Close the given popup, if it is open.
+    pub fn close_popup_with_id(&mut self, popup_id: Id) {
+        if self.is_popup_open(popup_id) {
+            self.close_popup();
+        }
+    }
```
lucasmerlin, line 1111: Should we instead rename this to `close_popup` and rename `close_popup` to `close_any_popup` or something like that?

emilk: I agree, if we call it `close_all_popups`

emilk (approved):

Issue #5751, opened 2025-03-02
Semi-transparent pixels of image appear too bright?


mccolljr:
**Describe the bug**
Apologies if there is a similar issue open - I searched, but I'm decidedly not a color space expert and I'm not 100% sure what terms apply to this problem.

I have a test `.png` image with a transparent background and some semi-transparent pixels:

<details>
<summary>
Original Image (wasn't sure how best to attach - you can do right click -> "Save Image As..."):
</summary>

![Image](https://github.com/user-attachments/assets/22218e96-d2dc-47f8-a069-3b7090d293ac)

</details>

I load it into my `eframe` application like so:

```rust
let image_data =
    image::load_from_memory(include_bytes!("./test_transparency.png")).unwrap().to_rgba();

let (width, height) = image_data.dimensions();

let tex_manager = egui_ctx.tex_manager();
let mut tex_manager = tex_manager.write();

let image_from_png = tex_manager.alloc(
    "test_transparency.png".to_string(),
    egui::ColorImage::from_rgba_unmultiplied(
        [width as usize, height as usize],
        image_dat.as_flat_samples().as_slice(),
    )
    .into(),
    egui::TextureOptions::default(),
);
```

I draw it to the screen later like so:
```rust
// draw the background color region
painter.image(
    egui::TextureId::default(),
    display_rect,
    egui::Rect::from_min_max(egui::epaint::WHITE_UV, egui::epaint::WHITE_UV),
    egui::Color32::DARK_GRAY, //<-- THE BACKGROUND COLOR
);

// draw the foreground image
painter.image(
    self.image_from_png,
    display_rect,
    egui::Rect::from_min_max(egui::pos2(0., 0.), egui::pos2(1., 1.)),
    egui::Color32::WHITE,
);
```

Some of the semi-transparent pixels appear a lot brighter than I expect when the image is drawn on top of certain background colors. I was expecting something more similar to how this image appears in an image editor or in the browser. I'm not sure if this is a bug, something I've done wrong in loading the image, or some third unknown option.

For reference, here are screenshots of what it looks like in my `egui` window:

<details>
<summary> Result over `Color32::WHITE` </summary>

![Image](https://github.com/user-attachments/assets/3a77a902-8dbc-46c8-a5c2-800048a40001)

</details>

<details>
<summary> Result over `Color32::LIGHT_GRAY` </summary>

![Image](https://github.com/user-attachments/assets/96f7481b-61ef-476b-83c3-d87ff0ba78ba)

[...]

Issue #5810, opened 2025-03-16
Ringing around pre-rasterized circles.


karhu:
**Describe the bug:**
Rendering darker circles on a brighter background can lead to "ringing" artifacts, where the outermost pixels of the circle are brighter than both the background and the circle. The artifact only shows when "Speed up filled circles with pre-rasterization" is enabled in the Tesselation Options (default). 

Some details: 
- Circle fill is `rgb(200,200,200)`
- Background fill is `rgb(220,220,220)`
- Circle is rendered **WITHOUT** any border.
- The observed ring is not just perceptual. Checked with MacOS color picker, and some pixels are `rgb(230,230,230)`.

<details>
**<summary>Minimal Repro Example:</summary>**

```
#[derive(Default)]
struct ExampleApp {}

impl eframe::App for ExampleApp {
    fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
        // Tesselation Options GUI.
        egui::containers::SidePanel::left("left").show(ctx, |ui| {
            let mut options = ui.ctx().tessellation_options(|options| options.clone());
            ui.add(&mut options);
            ui.ctx()
                .tessellation_options_mut(|target| *target = options);
        });
        // Problematic dots/circles.
        egui::CentralPanel::default()
            .frame(egui::Frame::new())
            .show(ctx, |ui| {
                ui.painter().rect_filled(
                    ui.clip_rect(),
                    egui::CornerRadius::ZERO,
                    egui::Color32::from_rgb(220, 220, 220),
                );
                for x in 1..20 {
                    for y in 1..20 {
                        ui.painter().circle_filled(
                            ui.clip_rect().left_top() + egui::vec2(x as f32 * 10.0, y as f32 * 10.0),
                            1.5,
                            egui::Color32::from_rgb(200, 200, 200),
                        );
                    }
                }
            });
    }
}

fn main() -> eframe::Result {
    env_logger::init();
    eframe::run_native(
        "Bug Repro",
        Default::default(),
        Box::new(|_cc| Ok(Box::<ExampleApp>::default())),
    )
}
```
</details>

**Expected behavior**
No ringing. Same visuals independent of whether pre-rasterization is enabled.

**Screenshots**

Bad:
<img width="451" alt="Image" src="https://github.com/user-attachments/assets/c5427e52-d05b-455a-8e19-e0647f510aa0" />

Expected:
[...]

Issue #5771, opened 2025-03-07
`Rgba::From<Color32>` does not consider alpha


virtualritz:
I believe the sRGB to linear sRGB [conversion function](https://docs.rs/ecolor/0.31.1/src/ecolor/lib.rs.html#40) has a bug.

The docs say that _both_ `Color32` and `Rgba` are _premultiplied_. When converting each component the `else` branch in [`linear_f32_from_gamma_u8()`](https://docs.rs/ecolor/0.31.1/ecolor/fn.linear_f32_from_gamma_u8.html) fails to divide by alpha  (and re-premultiply at the end).

Consider the current implementation. In the `if` branch alpha can be factored out but not in the `else` branch, i.e. it is *correct only as long as the color has a solid alpha*.

```rust
pub fn linear_f32_from_gamma_u8(s: u8) -> f32 {
    if s <= 10 {
        s as f32 / 3294.6
    } else {
        ((s as f32 + 14.025) / 269.025).powf(2.4)
    }
}
```

vs:

```rust
// alpha must be 0.0..1.0.
pub fn linear_f32_from_gamma_u8(s: u8, alpha: f32) -> f32 {
    if s <= 10 {
        s as f32 / 3294.6
    } else {
        (((s as f32 / alpha) + 14.025) / 269.025).powf(2.4) * alpha
    }
}
```

I moved the `alpha as f32 / 255.0` out of the function as that can be re-used for all three components.

The more transparent the color is, the more severe the error becomes (and the darker the resulting color than it should be). I came accross this as we're using `egui` for an in-house compositing app that works in ACEScg internally and I noticed that colors coming from the UI, using just `Rgba::From<Color32>`, seemed too dark when they had a non-opaque alpha.

For example, if you have a `Color32` component that is 122 with an alpha of 127 (~0.5), the current code converts this to ~`0.195` whereas I think this should actually end up as ~`0.452`.

I.e. the current version of `linear_f32_from_gamma_u8()` is only correct if `Color32` was unpremultiplied (or if alpha is `255` resp. `1.0`).

Does that make sense?

virtualritz:
I would also make a PR with a fix once I get confirmation.

lucasmerlin:
I wonder if this is related to https://github.com/emilk/egui/issues/5751

virtualritz:
> I wonder if this is related to [#5751](https://github.com/emilk/egui/issues/5751)

[...]

PR #5824, opened 2025-03-19
Fix semi-transparent colors appearing too bright

emilk:
* Closes https://github.com/emilk/egui/issues/5751
* Related to https://github.com/emilk/egui/issues/5771 (maybe closes?)
* Maybe related to https://github.com/emilk/egui/issues/5810 ?


crates/ecolor/src/color32.rs
```diff
@@ -111,20 +111,19 @@ impl Color32 {
...
                 static LOOKUP_TABLE: OnceLock<Box<[u8]>> = OnceLock::new();
                 let lut = LOOKUP_TABLE.get_or_init(|| {
-                    use crate::{gamma_u8_from_linear_f32, linear_f32_from_gamma_u8};
                     (0..=u16::MAX)
                         .map(|i| {
                             let [value, alpha] = i.to_ne_bytes();
-                            let value_lin = linear_f32_from_gamma_u8(value);
-                            let alpha_lin = linear_f32_from_linear_u8(alpha);
-                            gamma_u8_from_linear_f32(value_lin * alpha_lin)
+                            fast_round(value as f32 * linear_f32_from_linear_u8(alpha))
```
emilk, line 126: The inner loop is _much_ faster now, but the LUT still saves 30% runtime

crates/ecolor/src/color32.rs
```diff
@@ -333,3 +347,77 @@ impl std::ops::Add for Color32 {
...
+    fn test_color32_blend_vs_gamma_blend() {
+        let opaque = Color32::from_rgb(0x60, 0x60, 0x60);
+        let transparent = Color32::from_rgba_unmultiplied(168, 65, 65, 79);
+        assert_eq!(
+            transparent.blend(opaque),
+            opaque,
+            "Opaque on top of transparent"
+        );
+        // Blending in gamma-space is the de-facto standard everywhere,
+        // and it is what e.g. Chromium does.
```
Wumpf, line 377: I'm deeply confused about this now. I know from hearsay photoshop traditionally does this, but this being the "de-facto standard everywhere" takes me by surprise given that graphics apis all settled on blending in linear 😵‍💫 
(not that I don't believe you, it's just unsettling..)

maybe we need even more docs on `blend` in that regard who typically does what to anchor this more?

crates/ecolor/src/color32.rs
```diff
@@ -333,3 +347,77 @@ impl std::ops::Add for Color32 {
...
+        fn blend(dest: u8, src: u8, alpha: u8) -> u8 {
+            let src = src as f32 / 255.0;
+            let dest = dest as f32 / 255.0;
+            let alpha = alpha as f32 / 255.0;
+            fast_round((src * alpha + dest * (1.0 - alpha)) * 255.0)
+        }
+    }
+
+    #[test]
+    fn color32_unmultiplied_routrip() {
```
Wumpf, line 397:
```suggestion
    fn color32_unmultiplied_round_trip() {
```

crates/ecolor/src/rgba.rs
```diff
@@ -219,6 +219,11 @@ impl Rgba {
             linear_u8_from_linear_f32(a.abs()),
         ]
     }
+
+    /// Blend two colors, so that `self` is behind the argument.
+    pub fn blend(self, on_top: Self) -> Self {
```
Wumpf, line 224: It's technically redundant, but I think this should clarify that since `Rgba` is a linear color this performs blending in linear, as opposed to what `Color32` does.

crates/ecolor/src/color32.rs
```diff
@@ -111,20 +112,19 @@ impl Color32 {
     pub fn from_rgba_unmultiplied(r: u8, g: u8, b: u8, a: u8) -> Self {
```
Wumpf, line 112: should we document that this performs pre-multiplication in gamma space? 🤔

Wumpf (approved):
this blending in gamma space business drives me nuts, but apparently this produces the results everyone expects so great catch! 👍

crates/ecolor/src/color32.rs
karhu: I think we should properly specify and document this on the Color32 type level. Going to leave a longer comment at the PR top level.

karhu:
As mentioned in one of the line comment threads I think we need to properly specify and document what Color32 stores in respect to a linear, unmultiplied rgba value:

a) `to_gamma(to_premultiplied(some_linear_rgba_unmultiplied_value))`
or
b) `to_premultiplied(to_gamma(some_linear_rgba_unmultiplied_value))`

Or expressed differently: Do we gamma encode in linear space or in premultiplied space?

If I understand your changes to `from_rgba_unmultiplied` correctly, this flips the order of the two operations, so we now do b) instead of previously a).

For what it is worth, I think b) is the right way to go about this, as a) behaves weirdly when blending between non-opaque colors. **However**: From going through the egui codebase, my impression is that code pretty consistently assumes that we are doing a). 

This includes conversions between `Color32` and `Rgba`.  For example this quick test fails now:
```
    let rgba_unmult: [u8; 4] = [200, 150, 100, 150];

    let as_color32 = egui::Color32::from_rgba_unmultiplied(
        rgba_unmult[0],
        rgba_unmult[1],
        rgba_unmult[2],
        rgba_unmult[3],
    );

    let as_rgba = egui::Rgba::from(as_color32);

    let rgba_unmult_after = as_rgba.to_srgba_unmultiplied();
    assert!(
        rgba_unmult == rgba_unmult_after,
        "Values differ: {rgba_unmult:?} != {rgba_unmult_after:?}"
    );
 ```
 with `Values differ: [200, 150, 100, 150] != [151, 113, 77, 150]` (to be fair, it also fails on `master` but only due to rounding errors).
 
Sidenote: Should `from_rgba_unmultiplied` be renamed `from_srgba_unmultiplied` to be in line with `to_srgba_unmultiplied` ?

Another place is the shader code, which currently directly transforms between gamma and linear encoding, but would have to be changed to transform between premult / unmult before doing any gamma transforms. Also in the shader code, the texture read (which currently happens from an sRGBA aware texture) might be affected. From a first glance the automatic conversion on the driver level works with a), but not b). 

Phew, sorry for the wall of text. My brain is smoking, and I concur with @Wumpf that this is driving me nuts, so please take all of it with a big pile of salt.

File: crates/ecolor/src/color32.rs (hunk 2 above)
Comment (karhu) [reply crates/ecolor/src/color32.rs:377]: My understanding is that "CSS does it" and people are used to work with CSS colors, hence that's what everyone expects. I can highly recommend this article on the topic: 

https://blog.johnnovak.net/2016/09/21/what-every-coder-should-know-about-gamma/

which states:

> Almost everybody does this the wrong way: CSS gradients and transitions are wrong (see [this thread](https://lists.w3.org/Archives/Public/www-style/2012Jan/0607.html) for details), Photoshop is wrong (as of version CS6) and there’s not even an option to fix it.
> 
> Two drawing programs that got this (and gamma-correctness in general) right are [Krita](https://krita.org/) and [Pixelmator](https://www.pixelmator.com/). SVG also [let’s the user to specify](https://www.w3.org/TR/SVG/painting.html#ColorInterpolationProperties) whether to use linear or sRGB-space interpolations for gradients, compositing and animations.

**Comment by emilk** [comment]
@karhu you are right to point that out… I'll try to make things a bit more consistent.

File: crates/ecolor/src/color32.rs [hunk 6]
```diff
@@ -5,17 +5,32 @@ use crate::{fast_round, linear_f32_from_linear_u8, Rgba};
...
+/// Using premultiplied alpha has some advantages:
+/// * It allows encoding additive colors
+/// * It is the better way to blend colors, e.g. when filtering texture colors
+/// * Because the above, it is the better way to encode colors in a GPU texture
+///
+/// The color space is assumed to be [sRGB](https://en.wikipedia.org/wiki/SRGB).
+///
+/// All operations on `Color32` is done in "gamma space" (see <https://en.wikipedia.org/wiki/SRGB>).
+/// This is not physically correct, but it is fast and sometimes more perceptually even than linear space.
+/// If you want linear-space color, use [`Rgba`].
```
Wumpf, line 23:
```suggestion
/// If you instead want to perform these operations in linear-space color, use [`Rgba`].
```

Wumpf (approved):
nice, feels a looot more rigorous now!

valadaptive:
I ran into this while adding color emoji support in the Parley branch. I was about to file a similar issue, but found this.

However, I don't think this fixes the problem that I found.

As far as I can tell, the root cause is that with sRGB textures enabled, [the shader converts from linear to sRGB without un-premultiplying and re-premultiplying the colors](https://github.com/emilk/egui/blob/d54e29d3758f8abd0f97508ff89de46e739a0931/crates/egui_glow/src/shader/fragment.glsl#L50). If you change the highlighted line to:

```glsl
return vec4(srgb_gamma_from_linear(rgba.rgb / rgba.a) * rgba.a, rgba.a);
```

you get the results you'd expect. This also seems to be the reason that font gamma is behaving so weirdly, and Ubuntu Light looks so much heavier than it does everywhere else:

| linear -> sRGB premultiplied | linear -> sRGB un-and-re-multiplied | ditto, but with `FontImage::srgba_pixels`' gamma set to `1 / 2.2` |
| - | - | - |
|image | image | image |

Note the "fringe" around the ✨ emoji in the first example.

PR #5830, opened 2025-03-20
Add `AtomLayout`, abstracing layouting within widgets

lucasmerlin:
Today each widget does its own custom layout, which has some drawbacks:
- not very flexible
  - you can add an `Image` to `Button` but it will always be shown on the left side
  - you can't add a `Image` to a e.g. a `SelectableLabel`
- a lot of duplicated code

This PR introduces `Atoms` and `AtomLayout` which abstracts over "widget content" and layout within widgets, so it'd be possible to add images / text / custom rendering (for e.g. the checkbox) to any widget.

A simple custom button implementation is now as easy as this:
```rs
pub struct ALButton<'a> {
    al: AtomicLayout<'a>,
}

impl<'a> ALButton<'a> {
    pub fn new(content: impl IntoAtomics) -> Self {
        Self { al: content.into_atomics() }
    }
}

impl<'a> Widget for ALButton<'a> {
    fn ui(mut self, ui: &mut Ui) -> Response {
        let response = ui.ctx().read_response(ui.next_auto_id());

        let visuals = response.map_or(&ui.style().visuals.widgets.inactive, |response| {
            ui.style().interact(&response)
        });

        self.al.frame = self
            .al
            .frame
            .inner_margin(ui.style().spacing.button_padding)
            .fill(visuals.bg_fill)
            .stroke(visuals.bg_stroke)
            .corner_radius(visuals.corner_radius);

        self.al.show(ui)
    }
}

```

The initial implementation only does very basic layout, just enough to be able to implement most current egui widgets, so:
- only horizontal layout
- everything is centered
- a single item may grow/shrink based on the available space
- everything can be contained in a Frame

There is a trait `IntoAtoms` that conveniently allows you to construct `Atoms` from a tuple
```
   ui.button((Image::new("image.png"), "Click me!"))
```
to get a button with image and text.

This PR reimplements three egui widgets based on the new AtomicLayout:
 - Button
   - matches the old button pixel-by-pixel
   - Button with image is now [properly aligned](https://github.com/emilk/egui/pull/5830/files#diff-962ce2c68ab50724b01c6b64c683c4067edd9b79fcdcb39a6071021e33ebe772) in justified layouts
   - selected button style now matches SelecatbleLabel look
   - For some reason the DragValue text seems shifted by a pixel almost everywhere, but I think it's more centered now, yay?
 - Checkbox
   - basically pixel-perfect but apparently the check mesh is very slightly different so I had to update the snapshot
   - somehow needs a bit more space in some snapshot tests?
 - RadioButton
   - pixel-perfect
   - somehow needs a bit more space in some snapshot tests?

I plan on updating TextEdit based on AtomicLayout in a separate PR (so you could use it to add a icon within the textedit frame).

### DIFF
--- a/crates/egui/src/atomics/atom.rs
+++ b/crates/egui/src/atomics/atom.rs
@@ -5,7 +5,7 @@ use epaint::text::TextWrapMode;
 /// A low-level ui building block.
 ///
 /// Implements [`From`] for [`String`], [`str`], [`crate::Image`] and much more for convenience.
-/// You can directly call the `a_*` methods on anything that implements `Into<Atom>`.
+/// You can directly call the `atom_*` methods on anything that implements `Into<Atom>`.
 /// ```
 /// # use egui::{Image, emath::Vec2};
 /// use egui::AtomExt as _;


crates/egui/src/lib.rs
```diff
@@ -479,6 +480,7 @@ pub mod text {
 }
 
 pub use self::{
+    atomic_layout::*,
```
lucasmerlin, line 483: Should I instead make a `pub mod atomics`?

crates/egui/src/style.rs
```diff
@@ -1514,7 +1514,7 @@ impl Widgets {
             inactive: WidgetVisuals {
                 weak_bg_fill: Color32::from_gray(230), // button background
                 bg_fill: Color32::from_gray(230),      // checkbox background
-                bg_stroke: Default::default(),
+                bg_stroke: Stroke::new(1.0, Color32::default()),
```
lucasmerlin, line 1517:
```suggestion
                bg_stroke: Default::default(),
```

crates/egui/src/widgets/button.rs
```diff
@@ -243,188 +246,62 @@ impl Widget for Button<'_> {
...
+        let response = if ui.is_rect_visible(prepared.response.rect) {
+            let visuals = ui.style().interact_selectable(&prepared.response, selected);
+
+            if image_tint_follows_text_color {
+                prepared.sized_atomics.iter_mut().for_each(|a| {
+                    a.kind = match std::mem::take(&mut a.kind) {
+                        SizedAtomicKind::Image(image, size) => {
+                            SizedAtomicKind::Image(image.tint(visuals.text_color()), size)
+                        }
+                        other => other,
```
emilk, line 262: Maybe it could be nice to have helpers for this? Something like

```suggestion
                prepared.sized_atomics.for_each_image_mut(|image, _size| {
                     *image = image.tint(visuals.text_color());
```

crates/egui/src/widgets/button.rs
```diff
@@ -243,188 +246,62 @@ impl Widget for Button<'_> {
...
-                    .galley(right_text_pos, right_galley, visuals.text_color());
-            }
-        }
-
-        if let Some(cursor) = ui.visuals().interact_cursor {
-            if response.hovered() {
-                ui.ctx().set_cursor_icon(cursor);
-            }
-        }
-
```
emilk, line 427: 🔥

emilk (commented):
Please run some benchmarks (the ones that are already there, and maybe some new ones more focused on e.g. just `Button`)

crates/egui/src/widgets/checkbox.rs
```diff
@@ -51,92 +53,88 @@ impl Widget for Checkbox<'_> {
...
-        } else {
-            let total_extra = vec2(icon_width + icon_spacing, 0.0);
-
-            let wrap_width = ui.available_width() - total_extra.x;
-            let galley = text.into_galley(ui, None, wrap_width, TextStyle::Button);
+        let mut min_size = Vec2::splat(spacing.interact_size.y);
+        min_size.y = min_size.y.at_least(icon_width);
 
-            let mut desired_size = total_extra + galley.size();
-            desired_size = desired_size.at_least(spacing.interact_size);
```
lucasmerlin, line 71: Ah the differnce in Slider.png comes from here. The previous checkbox/radio button had a min_size of interact_size if the text isn't empty. I feel like that doesn't make much sense for the checkbox though since it doesn't have a frame.

This is a screenshot from master that exaggerates it:

Screenshot 2025-04-24 at 15 17 51


This is what it looks like on my branch:

Screenshot 2025-04-24 at 15 19 13

emilk:
Remember to run `cargo doc -p egui --open` and check the docs for the new top-level types. Some are missing, some are bad

crates/egui/src/atomic_layout.rs
```diff
@@ -0,0 +1,740 @@
...
+    T: Into<WidgetText>,
+{
+    fn from(value: T) -> Self {
+        AtomicKind::Text(value.into())
+    }
+}
+
+/// A list of [`Atomic`]s.
+#[derive(Clone, Debug, Default)]
+pub struct Atomics<'a>(Vec<Atomic<'a>>);
```
emilk, line 638: Let's put each type into its own file under `egui/src/atomics/…` so that
A) Makes it easier to find the source code for each type
B) give each type room to grow
C) prevents a huge diff when we eventually decides to break up this file into a smaller files when it reached 2kloc or whatever :)

Also makes it easier to review (I can mark each type as reviewed when done)

crates/egui/src/widgets/checkbox.rs
```diff
@@ -19,21 +21,21 @@ use crate::{
 #[must_use = "You should put this widget in a ui with `ui.add(widget);`"]
 pub struct Checkbox<'a> {
     checked: &'a mut bool,
-    text: WidgetText,
+    atomics: Atomics<'a>,
```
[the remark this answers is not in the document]
lucasmerlin: Right now it's only for the ImageSource so it could also be `Atomics<'static>` but I think we miiiiight want to add a AtomicKind with a closure in the future, and then already having lifetimes everywhere would make the switch much easier. 
But I could also remove them for now.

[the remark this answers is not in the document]
emilk: Let's keep them for now. We can easily remove them if we discover some problem, but as you say it's much harder to _add_ them after the fact without breaking some users.

crates/egui/src/atomic_layout.rs
```diff
@@ -0,0 +1,740 @@
...
+
+/// Response from a [`AtomicLayout::show`] or [`AllocatedAtomicLayout::paint`].
+///
+/// Use the `custom_rects` together with [`AtomicKind::Custom`] to add child widgets to a widget.
+///
+/// NOTE: Don't `unwrap` rects, they might be empty when the widget is not visible.
+#[derive(Clone, Debug)]
+pub struct AtomicLayoutResponse {
+    pub response: Response,
+    pub custom_rects: HashMap<Id, Rect>,
```
emilk, line 387: This will cause additional allocation for each atomic. We could avoid this by instead using `SmallVec<[(Id, Rect); 2]>`

emilk: …with a helper for doing the lookups

crates/egui/src/atomic_layout.rs
```diff
@@ -0,0 +1,740 @@
...
+            align2,
+            gap,
+        }
+    }
+}
+
+/// Instructions for painting an [`AtomicLayout`].
+#[derive(Clone, Debug)]
+pub struct AllocatedAtomicLayout<'a> {
+    pub sized_atomics: Vec<SizedAtomic<'a>>,
```
emilk, line 303: Same here: a `SmallVec` would save us allocations in the common case

emilk: …but a benchmark would be useful.

crates/egui/src/atomic_layout.rs
```diff
@@ -0,0 +1,740 @@
...
+            frame: Frame::default(),
+            sense: Sense::hover(),
+            fallback_text_color: None,
+            min_size: Vec2::ZERO,
+            wrap_mode: None,
+            align2: None,
+        }
+    }
+
+    /// Insert a new [`Atomic`] at the end of the list (left side).
```
emilk, line 63: It's quite surprising that the end is on the left. Is this really correct?

lucasmerlin: Ooops, no I swapped them

crates/egui/src/atomic_layout.rs
```diff
@@ -0,0 +1,740 @@
...
+        self
+    }
+
+    /// Insert a new [`Atomic`] at the beginning of the list (right side).
+    #[inline]
+    pub fn push_front(mut self, atomic: impl Into<Atomic<'a>>) -> Self {
+        self.atomics.push_front(atomic.into());
+        self
+    }
+
```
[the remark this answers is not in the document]
lucasmerlin: How should AtomicLayout react to right_to_left layout, should it reverse item order? Should it be configurable?

crates/egui/src/atomic_layout.rs
```diff
@@ -0,0 +1,740 @@
...
+            min_size,
+            wrap_mode,
+            align2,
+        } = self;
+
+        let wrap_mode = wrap_mode.unwrap_or(ui.wrap_mode());
+
+        // If the TextWrapMode is not Extend, ensure there is some item marked as `shrink`.
+        // If none is found, mark the first text item as `shrink`.
+        if !matches!(wrap_mode, TextWrapMode::Extend) {
```
emilk, line 170:
```suggestion
        if wrap_mode != TextWrapMode::Extend {
```

crates/egui/src/atomic_layout.rs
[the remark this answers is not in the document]
emilk: I don't think `AtomicLayout` should care about the layout of whatever `Ui` it is in. A checkbox should always have the checkbox on the left and the label on the right no matter what the ui layout it is in. (unless the users sets some specific setting on the checkbox itself)

crates/egui/src/atomics/atomic.rs
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+}
+
+/// A trait for conveniently building [`Atomic`]s.
+pub trait AtomicExt<'a> {
+    /// Set the atomic to a fixed size.
+    fn a_size(self, size: Vec2) -> Atomic<'a>;
+
+    /// Grow this atomic to the available space.
```
emilk, line 74: How does this interplay with `a_size`? Does `a_size` become a minimum size?

crates/egui/src/atomics/atomic.rs
```diff
@@ -0,0 +1,119 @@
...
+    fn a_grow(self, grow: bool) -> Atomic<'a>;
+
+    /// Shrink this atomic if there isn't enough space.
+    ///
+    /// NOTE: Only a single [`Atomic`] may shrink for each widget.
+    fn a_shrink(self, shrink: bool) -> Atomic<'a>;
```
emilk, line 80: How does this interact with `a_size?`

crates/egui/src/atomics/atomic.rs
```diff
@@ -0,0 +1,119 @@
...
+
+impl<'a, T> From<T> for Atomic<'a>
+where
+    T: Into<AtomicKind<'a>>,
+{
+    fn from(value: T) -> Self {
+        Atomic {
+            size: None,
+            grow: false,
+            shrink: false,
```
emilk, line 115: Can these both be `true`?

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+
+/// Response from a [`AtomicLayout::show`] or [`AllocatedAtomicLayout::paint`].
+///
+/// Use the `custom_rects` together with [`AtomicKind::Custom`] to add child widgets to a widget.
+///
+/// NOTE: Don't `unwrap` rects, they might be empty when the widget is not visible.
+#[derive(Clone, Debug)]
+pub struct AtomicLayoutResponse {
+    pub response: Response,
+    pub custom_rects: HashMap<Id, Rect>,
```
emilk, line 385: Actually, if this is just for `AtomicKind::Custom`, then the common case might be for this to be empty or having at most one `Id` (for e.g. Checkbox)? So maybe `SmallVec<[(Id, Rect)]; 1>`

crates/egui/src/atomics/atomic_kind.rs
```diff
@@ -0,0 +1,105 @@
...
+use std::fmt::Formatter;
+
+/// The different kinds of [`Atomic`]s.
+#[derive(Clone, Default)]
+pub enum AtomicKind<'a> {
+    /// Empty, that can be used with [`AtomicExt::a_grow`] to reserve space.
+    #[default]
+    Empty,
+    Text(WidgetText),
+    Image(Image<'a>),
```
emilk, line 13:
```suggestion
    Empty,
    
    /// TODO: Explain how the text is wrapped/truncated/etc here
    Text(WidgetText),
    
    /// TODO: Explain how the size of the image is chosen here
    Image(Image<'a>),
```

crates/egui/src/atomics/atomic_kind.rs
```diff
@@ -0,0 +1,105 @@
...
+impl std::fmt::Debug for AtomicKind<'_> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            AtomicKind::Empty => write!(f, "AtomicKind::Empty"),
+            AtomicKind::Text(text) => write!(f, "AtomicKind::Text({})", text.text()),
+            AtomicKind::Image(image) => write!(f, "AtomicKind::Image({image:?})"),
+            AtomicKind::Custom(id, size) => write!(f, "AtomicKind::Custom({id:?}, {size:?})"),
+        }
+    }
+}
```
emilk, line 46: If we `impl Debug for WidgetText`, then maybe we could `#[derive(Debug)]` for `AtomicKind`?

crates/egui/src/atomics/atomic_kind.rs
```diff
@@ -0,0 +1,105 @@
...
+            AtomicKind::Text(text) => {
+                let galley = text.into_galley(ui, wrap_mode, available_size.x, TextStyle::Button);
+                (
+                    galley.size(), // TODO(lucasmerlin): calculate the preferred size
+                    SizedAtomicKind::Text(galley),
+                )
+            }
+            AtomicKind::Image(image) => {
+                let max_size = Vec2::splat(font_size);
+                let size = image.load_and_calc_size(ui, Vec2::min(available_size, max_size));
```
emilk, line 82: The parameter `font_size` is not used to control any font sizes, but is used to control the maximum width and height of images 🤔 

I'm guessing the intent is that image heights should be capped at the text height, but in this context it is a bit confusing. Maybe it would make more sense to pass in a `max_image_size: Vec2` to this method?

crates/egui/src/atomics/atomic_kind.rs
```diff
@@ -0,0 +1,105 @@
...
+        match self {
+            AtomicKind::Text(text) => {
+                let galley = text.into_galley(ui, wrap_mode, available_size.x, TextStyle::Button);
+                (
+                    galley.size(), // TODO(lucasmerlin): calculate the preferred size
+                    SizedAtomicKind::Text(galley),
+                )
+            }
+            AtomicKind::Image(image) => {
+                let max_size = Vec2::splat(font_size);
```
emilk, line 81: Say you are making a country-selector button containing a flag and a text. You would want the height of the flag to be the height of the text, but the width of the flag to be wider: "🇸🇪 Sweden"

crates/egui/src/atomics/sized_atomic.rs
```diff
@@ -0,0 +1,11 @@
+use crate::SizedAtomicKind;
+use emath::Vec2;
+
+/// A [`Atomic`] which has been sized.
+#[derive(Clone, Debug)]
+pub struct SizedAtomic<'a> {
+    pub grow: bool,
```
emilk, line 7: If it has been sized, how can it grow?

crates/egui/src/atomics/sized_atomic.rs
```diff
@@ -0,0 +1,11 @@
...
+    pub size: Vec2,
+    pub preferred_size: Vec2,
```
emilk, line 9: These two deserve docstrings

crates/egui/src/atomics/mod.rs
```diff
@@ -0,0 +1,14 @@
+mod atomic;
+mod atomic_kind;
+mod atomic_layout;
+#[allow(clippy::module_inception)]
```
emilk, line 4: Haha, that's a fun name for a clippy lint :)

crates/egui/src/atomics/atomics.rs
```diff
@@ -0,0 +1,120 @@
+use crate::{Atomic, AtomicKind};
+
+/// A list of [`Atomic`]s.
+#[derive(Clone, Debug, Default)]
+pub struct Atomics<'a>(Vec<Atomic<'a>>);
```
emilk, line 5:
```suggestion
pub struct Atomics<'a>(SmallVec<[Atomic<'a>; 2]>); // Two atomics is very common for e.g. checkboxes and radio buttons, so avoid allocations in those cases
```

crates/egui/src/atomics/atomics.rs
```diff
@@ -0,0 +1,120 @@
...
+
+impl<'a> Atomics<'a> {
+    /// Insert a new [`Atomic`] at the end of the list (right side).
+    pub fn push(&mut self, atomic: impl Into<Atomic<'a>>) {
+        self.0.push(atomic.into());
+    }
+
+    /// Insert a new [`Atomic`] at the beginning of the list (left side).
+    pub fn push_front(&mut self, atomic: impl Into<Atomic<'a>>) {
```
emilk, line 14: As mentioned elsewhere, these could be called `push_left` and `push_right`

crates/egui/src/atomics/atomics.rs
```diff
@@ -0,0 +1,120 @@
...
+        self.0.is_empty()
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = &Atomic<'a>> {
+        self.0.iter()
+    }
+
+    pub fn iter_mut(&mut self) -> impl Iterator<Item = &mut Atomic<'a>> {
+        self.0.iter_mut()
+    }
```
emilk, line 32: These could be replaced with `impl Deref` and `DerefMut` `for Atomics` with `Target = [Atomic]`

crates/egui/src/atomics/atomics.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    /// Concatenate and return the text contents.
+    // TODO(lucasmerlin): It might not always make sense to return the concatenated text, e.g.
+    // in a submenu button there is a right text '⏵' which is now passed to the screen reader.
```
emilk, line 36: We should probably replace that `⏵` text with an actual icon…

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+    /// Insert a new [`Atomic`] at the end of the list (right side).
+    #[inline]
+    pub fn push(mut self, atomic: impl Into<Atomic<'a>>) -> Self {
+        self.atomics.push(atomic.into());
+        self
+    }
+
+    /// Insert a new [`Atomic`] at the beginning of the list (left side).
+    #[inline]
+    pub fn push_front(mut self, atomic: impl Into<Atomic<'a>>) -> Self {
```
emilk, line 71: push_left/push_right?

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+/// - [`AllocatedAtomicLayout::paint`]
+///   - paints the [`Frame`]
+///   - calculates individual [`Atomic`] positions
+///   - paints each single atomic
+///
+/// You can use this to first allocate a response and then modify, e.g., the [`Frame`] on the
+/// [`AllocatedAtomicLayout`] for interaction styling.
+pub struct AtomicLayout<'a> {
+    id: Option<Id>,
+    pub atomics: Atomics<'a>,
```
emilk, line 31: Should this be pub?

Should we `Deref` to it?

crates/egui/src/atomics/atomic.rs
```diff
@@ -0,0 +1,119 @@
...
+/// Implements [`From`] for [`String`], [`str`], [`crate::Image`] and much more for convenience.
+/// You can directly call the `a_*` methods on anything that implements `Into<Atomic>`.
+/// ```
+/// # use egui::{Image, emath::Vec2};
+/// use egui::AtomicExt;
+/// let string_atomic = "Hello".a_grow(true);
+/// let image_atomic = Image::new("some_image_url").a_size(Vec2::splat(20.0));
+/// ```
+#[derive(Clone, Debug)]
+pub struct Atomic<'a> {
```
emilk, line 17: And this is a whopping 168 bytes 😱

emilk: A lot of it is `WidgetText`, which is 112 bytes. `WidgetText` could use some slimming down at some point…

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+///   - allocates a [`Response`]
+///   - returns a [`AllocatedAtomicLayout`]
+/// - [`AllocatedAtomicLayout::paint`]
+///   - paints the [`Frame`]
+///   - calculates individual [`Atomic`] positions
+///   - paints each single atomic
+///
+/// You can use this to first allocate a response and then modify, e.g., the [`Frame`] on the
+/// [`AllocatedAtomicLayout`] for interaction styling.
+pub struct AtomicLayout<'a> {
```
emilk, line 29: This is a pretty chonky 96 bytes big. I wonder what the performance implications of this PR will be, both on native and web 😬

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+    #[inline]
+    pub fn fallback_text_color(mut self, color: Color32) -> Self {
+        self.fallback_text_color = Some(color);
+        self
+    }
+
+    /// Set the minimum size of the Widget.
+    #[inline]
+    pub fn min_size(mut self, size: Vec2) -> Self {
+        self.min_size = size;
```
emilk, line 111: What happens if there is no `grow: true` atomic in the widget?

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+    #[inline]
+    pub fn id(mut self, id: Id) -> Self {
+        self.id = Some(id);
+        self
+    }
+
+    /// Set the [`TextWrapMode`] for the [`Atomic`] marked as `shrink`.
+    ///
+    /// Only a single [`Atomic`] may shrink. If this (or `ui.wrap_mode()`) is not
+    /// [`TextWrapMode::Extend`] and no item is set to shrink, the first (right-most)
```
emilk, line 125: first or right-most?

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+    /// [`AtomicKind::Text`] will be set to shrink.
+    #[inline]
+    pub fn wrap_mode(mut self, wrap_mode: TextWrapMode) -> Self {
+        self.wrap_mode = Some(wrap_mode);
+        self
+    }
+
+    /// Set the [`Align2`].
+    ///
```
emilk, line 134: What is being aligned? Is it the atomics within the AtomicLayout, or the whole AtomicLayout within the parent?

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+        // If the TextWrapMode is not Extend, ensure there is some item marked as `shrink`.
+        // If none is found, mark the first text item as `shrink`.
+        if wrap_mode != TextWrapMode::Extend {
+            let any_shrink = atomics.iter().any(|a| a.shrink);
+            if !any_shrink {
+                let first_text = atomics
+                    .iter_mut()
+                    .find(|a| matches!(a.kind, AtomicKind::Text(..)));
+                if let Some(atomic) = first_text {
+                    atomic.shrink = true;
```
emilk, line 176: What will this do for the text? Truncate?

```suggestion
                    atomic.shrink = true; // Will make the text truncate if needed
```

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+
+        if let Some((index, item)) = shrink_item {
+            // The `shrink` item gets the remaining space
+            let shrunk_size = Vec2::new(
+                available_inner_size.x - desired_width,
+                available_inner_size.y,
+            );
+            if item.grow {
+                grow_count += 1;
+            }
```
emilk, line 261:
```suggestion
        for (idx, item) in atomics.into_iter().enumerate() {
            if item.grow {
                grow_count += 1;
            }
            if item.shrink {
                debug_assert!(
                    shrink_item.is_none(),
                    "Only one atomic may be marked as shrink"
                );
                if shrink_item.is_none() {
                    shrink_item = Some((idx, item));
                    continue;
                }
            }
            let sized = item.into_sized(ui, available_inner_size, max_font_size, Some(wrap_mode));
            let size = sized.size;

            desired_width += size.x;
            preferred_width += sized.preferred_size.x;

            height = height.at_least(size.y);
            preferred_height = preferred_height.at_least(sized.preferred_size.y);

            sized_items.push(sized);
        }

        if let Some((index, item)) = shrink_item {
            // The `shrink` item gets the remaining space
            let shrunk_size = Vec2::new(
                available_inner_size.x - desired_width,
                available_inner_size.y,
            );
```

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+        let fallback_text_color =
+            fallback_text_color.unwrap_or_else(|| ui.style().visuals.text_color());
+        let gap = gap.unwrap_or(ui.spacing().icon_spacing);
+
+        // The size available for the content
+        let available_inner_size = ui.available_size() - frame.total_margin().sum();
+
+        let mut desired_width = 0.0;
+        let mut preferred_width = 0.0;
+        let mut preferred_height = 0.0;
```
emilk, line 192: Whats the difference between these?

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+
+            sized_items.push(sized);
+        }
+
+        if let Some((index, item)) = shrink_item {
+            // The `shrink` item gets the remaining space
+            let shrunk_size = Vec2::new(
+                available_inner_size.x - desired_width,
+                available_inner_size.y,
+            );
```
emilk, line 258: This is confusing. It looks like `desired_width` does not include the size of the `shrink` item.

What you are calling `shrunk_size` seems to be how much _extra_ space we have for X (which should be distributed to the _grow_ item, not the _shrink_ item)

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+            if item.grow {
+                grow_count += 1;
+            }
+
+            let sized = item.into_sized(ui, shrunk_size, max_font_size, Some(wrap_mode));
```
emilk, line 263: Ah, `shrunk_size` should be called something like `avilable_size_for_shrink_item`

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+    desired_size: Vec2,
+    align2: Align2,
+    gap: f32,
+}
+
+impl<'a> AllocatedAtomicLayout<'a> {
+    /// Paint the [`Frame`] and individual [`Atomic`]s.
+    pub fn paint(self, ui: &Ui) -> AtomicLayoutResponse {
+        let Self {
+            sized_atomics: sized_items,
```
emilk, line 316: Can we use the same name instead?

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+            gap,
+        } = self;
+
+        let inner_rect = response.rect - self.frame.total_margin();
+
+        ui.painter().add(frame.paint(inner_rect));
+
+        let width_to_fill = inner_rect.width();
+        let extra_space = f32::max(width_to_fill - desired_size.x, 0.0);
+        let grow_width = f32::max(extra_space / grow_count as f32, 0.0);
```
emilk, line 332:
```suggestion
        let grow_width = f32::max(extra_space / grow_count as f32, 0.0).floor_ui();
```

Floor so that the sum will not exceed the original limits

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,392 @@
...
+            align2.align_size_within_rect(Vec2::new(width_to_fill, desired_size.y), inner_rect)
+        } else {
+            align2.align_size_within_rect(desired_size, inner_rect)
+        };
+
+        let mut cursor = aligned_rect.left();
+
+        let mut response = AtomicLayoutResponse {
+            response,
+            custom_rects: HashMap::new(),
```
emilk, line 344: less code to change if we change the type of `custom_rects` in the future

```suggestion
            custom_rects: Default::default(),
```

emilk (commented):
Looks great! But I really want to get some benchmarks going before we merge this

lucasmerlin:
I ran some benchmarks. Seems like it is indeed a bit slower. First a benchmark of the demo lib:
```
demo_no_tessellate:
master: 	 80us
small_vec: 	 82us
vec:  		 86us
```

Then I wrote a benchmark testing button with different configurations:
|  | text | text + image | text + right text + image |
|--------|--------|--------|--------|
| master | 601.59 ns | 1.0176 µs | 1.0299 µs |
| small_vec | 771.49 ns | 1.0122 µs | 1.0574 µs |
| vec | 798.31 ns | 1.1199 µs | 1.1639 µs | 

See the benchmark code here: https://github.com/emilk/egui/pull/6854

crates/egui/src/atomics/atomic_kind.rs
lucasmerlin: I've removed this behavior

crates/egui/src/atomics/atomic_kind.rs
lucasmerlin: This should work now with max width / height 👍🏻

crates/egui/src/atomics/sized_atomic.rs
lucasmerlin: The growing actually happens during the placing, since we need to know all sizes for this. So the flag is stored so we know which ones to grow. But I should make it private since modifying it after sizing would be bad.

crates/egui/src/atomics/atomic_layout.rs
lucasmerlin: Everything will be aligned to the left.

crates/egui/src/atomics/atomic_layout.rs
lucasmerlin: It will wrap or truncate based on the wrap_mode. Since we check that wrap_mode != TextWrapMode::Extend it should always be able to shrink.

crates/egui/src/atomics/atomic_layout.rs
lucasmerlin: Preferred width / height is the ideal size of the widget, e.g. the size where the text is not wrapped. It's used to set Response::intrinsic_size.

lucasmerlin:
Ran the demo benchmark for main and the PR and it seems to be slightly slower but I feel the slow down is acceptable for the features we gain:

|                | Demo Realistic | Demo No Tesselate | Demo Only Tesselate |
|----------------|----------------|-------------------|----------------------|
| **Master**     | 109.83 µs      | 73.938 µs         | 38.209 µs            |
| **Atomics**    | 112.93 µs      | 75.127 µs         | 38.099 µs            |

crates/egui/src/widgets/button.rs
```diff
@@ -182,15 +186,22 @@ impl<'a> Button<'a> {
...
-    pub fn shortcut_text(mut self, shortcut_text: impl Into<WidgetText>) -> Self {
-        self.right_text = shortcut_text.into().weak();
+    pub fn shortcut_text(mut self, shortcut_text: impl Into<Atomic<'a>>) -> Self {
+        let mut atomic = shortcut_text.into();
+        atomic.kind = match atomic.kind {
+            AtomicKind::Text(text) => AtomicKind::Text(text.weak()),
+            other => other,
+        };
+        self.layout.push_left(Atomic::grow());
+        self.layout.push_left(atomic);
```
emilk, line 196: shouldn't this be `push_right`?

crates/egui/src/widgets/button.rs
```diff
@@ -182,15 +186,22 @@ impl<'a> Button<'a> {
...
         self
     }
 
     /// Show some text on the right side of the button.
     #[inline]
-    pub fn right_text(mut self, right_text: impl Into<WidgetText>) -> Self {
-        self.right_text = right_text.into();
+    pub fn right_text(mut self, right_text: impl Into<Atomic<'a>>) -> Self {
+        self.layout.push_left(Atomic::grow());
+        self.layout.push_left(right_text.into());
```
emilk, line 204: This definitely should be `.push_right`

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,487 @@
...
+            align2.align_size_within_rect(desired_size, inner_rect)
+        };
+
+        let mut cursor = aligned_rect.left();
+
+        let mut response = AtomicLayoutResponse::empty(response);
+
+        for sized in sized_atomics {
+            let size = sized.size;
+            let growth = if sized.is_grow() { grow_width } else { 0.0 };
```
emilk, line 392: There is a subtle sizing issue lurking here, that it might be worth thinking about. Assume `.round_ui` rounds to integers for now (makes it easier to explain the problem):

If `extra_space` is 7, and we have 3 growable atoms, then the current code will grow each of those atoms by 2, leading to a total growth of 6, which is less than `extra_space`. Is that fine? Maybe. But a better (though more complicated) approach might be to handle this by allowing the last atom to grow more or less than the others.

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,487 @@
...
+
+        response
+    }
+}
+
+/// Response from a [`AtomicLayout::show`] or [`AllocatedAtomicLayout::paint`].
+///
+/// Use the `custom_rects` together with [`AtomicKind::Custom`] to add child widgets to a widget.
+///
+/// NOTE: Don't `unwrap` rects, they might be empty when the widget is not visible.
```
emilk, line 428: This comment makes more sense on `fn get_rect`

crates/egui/src/atomics/atomic_layout.rs
```diff
@@ -0,0 +1,487 @@
...
+            response,
+            custom_rects: Default::default(),
+        }
+    }
+
+    pub fn custom_rects(&self) -> impl Iterator<Item = (Id, Rect)> + '_ {
+        self.custom_rects.iter().copied()
+    }
+
+    pub fn get_rect(&self, id: Id) -> Option<Rect> {
```
emilk, line 448: This should have a warning about that the `id` might be missing for hidden/invisible widgets

emilk, line 448: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

```suggestion
    pub fn rect(&self, id: Id) -> Option<Rect> {
```

emilk (approved):

crates/egui/Cargo.toml
```diff
@@ -87,6 +87,7 @@ ahash.workspace = true
 bitflags.workspace = true
 nohash-hasher.workspace = true
 profiling.workspace = true
+smallvec.workspace = true
```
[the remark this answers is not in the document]
lucasmerlin: mnopqrstu... It should be sorted correctly

crates/egui/src/atomics/atomic_layout.rs
lucasmerlin: This doesn't affect the widget size, so I think it should be fine like this

crates/egui/src/atomics/atom.rs
```diff
@@ -0,0 +1,109 @@
+use crate::{AtomKind, Id, SizedAtom, Ui};
+use emath::{NumExt as _, Vec2};
+use epaint::text::TextWrapMode;
+
+/// A low-level ui building block.
+///
+/// Implements [`From`] for [`String`], [`str`], [`crate::Image`] and much more for convenience.
+/// You can directly call the `a_*` methods on anything that implements `Into<Atom>`.
```
lucasmerlin:
```suggestion
/// You can directly call the `atom_*` methods on anything that implements `Into<Atom>`.
```

PR #5839, opened 2025-03-21
Use sRGBA unaware textures, and let the GPU interpolate in sRGB space.

karhu:
This is one potential fix to the issues we are seeing around GPU texture interpolation of semi-transparent values. 

Note that this intentionally switches texture interpolation from linear to gamma. One one hand this makes interpolation more consistent with how it is done everywhere else in egui (vertex blending, and cpu computations). On the other hand it breaks a current assumption (see the screenshot from the rendering demo app). 

If linear texture interpolation is not necessary, I think this is our best option, and I want to put it forward for discussion. See below for alternatives. 

## More

**Alternatives that allow for linear blending:**
- Use sRGB aware textures, but **do not** premultiply the stored values.
  - This will open us up to interpolation artifacts due to interpolating in non-premultiplied spaces (i.e color bleeding) and is hence undesirable.
- Use sRGB unaware textures, and **do not** store sRGB values, but linear ones instead.
  - We will loose the precision gains that gamma encoding in u8 gives us. Still, if linear texture interpolation is strictly desired, this is our best option, perhaps in combination with a higher precision linear texture format. 

**Missing: **
- If these changes are acceptable I'll update the PR with matching changes to the glow backend. 

## Related:

* https://github.com/emilk/egui/pull/5824
* Closes https://github.com/emilk/egui/issues/5810

### DIFF
--- a/crates/egui-wgpu/src/egui.wgsl
+++ b/crates/egui-wgpu/src/egui.wgsl
@@ -97,9 +97,11 @@ fn vs_main(
 
 @fragment
 fn fs_main_linear_framebuffer(in: VertexOutput) -> @location(0) vec4<f32> {
-    // We always have an sRGB aware texture at the moment.
-    let tex_linear = textureSample(r_tex_color, r_tex_sampler, in.tex_coord);
-    let tex_gamma = gamma_from_linear_rgba(tex_linear);
+    // We use sRGB unware textures, and store (premultiplied) sRGB values.
+    // Hence any interpolation done by the GPU will be in gamma space.
+    // Note: Because we store premultiplied alpha, we can't rely on sRGB aware
+    // textures that convert "sRGB -> linear RGB" before interpolating.
+    let tex_gamma = textureSample(r_tex_color, r_tex_sampler, in.tex_coord);
     var out_color_gamma = in.color * tex_gamma;
     // Dither the float color down to eight bits to reduce banding.
     // This step is optional for egui backends.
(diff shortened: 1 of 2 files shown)


emilk:
Interesting! I would love to see if this does anything for text sharpness

karhu:
Yes, I saw your other PR on that topic and at some point was also experimenting with changes to the encoding there, but then figured to keep the two separate.  I consider this here more a correctness fix, enabling proper calibration (your CL) afterwards.

Wumpf:
this one has been posted somewhere already probably, but love this take 😄 https://erikmcclure.com/blog/everyone-does-srgb-wrong-because/

> But one can hardly blame you, because everyone is doing it wrong. CSS does it wrong because SVG does it wrong because Photoshop does it wrong because everyone else does it wrong. It’s all wrong.

... but in the end that's what is causing the ringing as also shown there so there's that :/

karhu:
No need to rush from my side :) And don't get me wrong, even though that might not be obvious from my last couple comments here, coming from 3D rendering, ideologically I'm very much in the "math in linear space" camp 😅.

My main priority here is to fix obvious rendering artifacts, and if I'm pushing for "math in gamma space", then it's because:
- eGUI currently seems to mostly operate in gamma space (textures being the one exception), and I was hoping sticking with it would be the path of least resistance. 
- eGUI being a UI library, and being consistent with CSS has some value here.

Regarding the ringing we see here: It's not because we do either gamma or linear interpolation. It's because we do linear interpolation wrong 🙈 We rely on the automatic GPU conversion upon read/write, which assumes non-premultiplied values, which, well, is not the case :P

karhu:
PS: I'm also wondering if we should treat "image textures" and "font/shape atlas textures" separately. Image textures contain colors, with all the headache that entails. The atlas textures (which could actually be cut down in size to 1/4) contain "pixel coverage", which we might or might not want to treat differently, see: for example https://hikogui.org/2022/10/24/the-trouble-with-anti-aliasing.html

emilk:
> PS: I'm also wondering if we should treat "image textures" and "font/shape atlas textures" separately.

With the new parley work, the font atlas texture will be full color (in order to support color emojis): https://github.com/emilk/egui/pull/5784

emilk:
This also improves how icons (with translucency) are blended.

I'd like to have a screenshot test for this improvement

Before:
image

After:
image


* EDIT: test added in #7309

emilk:
I'm forking this PR so that I can finish it up (including updating the snapshot images):
* https://github.com/emilk/egui/pull/7311

Issue #5832, opened 2025-03-20
Update color picker buttons to the new popup api


lucasmerlin:
**Describe the bug**
#5814 made it so you need to call keep_popup_open or the popup will close. Since the color picker doesn't use the new popup yet, it now immideately closes.

**To Reproduce**
1. try using color picker on egui.rs

To fix this we should ensure all places using the `Memory::popup` api are either 
- updated to use the new `Popup` struct
- updated to call keep_popup_open

PR #5847, opened 2025-03-25
Fix color picker button

lucasmerlin:
* related to #5832 
  (I want to keep that open and actually update the button to use the new popup, but this should be enough to fix it for now) 


emilk (approved):
I don't like this new footgun 😬 
Not sure what the best way to fix it though 🤔

PR #5851, opened 2025-03-27
Make debugging ids easier

lucasmerlin:
Eguis Ids are awesome, but sometimes it's a pain to figure out where a particular Id is coming from. This is an experiment at improving the situation by requiring the Id's source to implemnt `Debug`, storing the Debug representation as a string and the Ids optional parent in a global map, and showing the additional info it in the Debug implementation.  Since most Ids are derived from a string this works really well:

Screenshot 2025-03-27 at 16 12 21

I also plan on adding a `Id::ui` that always shows the source and parent and more info on hover (the parents parent and maybe it could also look for widget rects with the Id and higlight those).

Since the String only has to be created once for each unique Id this doesn't seem to have a huge performance impact, so I feel we can always enable this in debug builds.

The downside is of course that creating an Id from something now requires that thing to implement debug, but I don't think it would be a huge problem in practise. We could also add a `Id::new_without_debug_info()` that just requires the Hash trait.

Also, the current Ui auto ids don't work well with this since they just implement an u64, but I'm sure there is a nice way to make that work.


emilk:
I love this!

lucasmerlin:
It's currently gated behind debug_assertions, but maybe it should be a feature flag instead? Or both?

crates/egui/src/id.rs
```diff
@@ -101,11 +116,237 @@ impl Id {
     pub unsafe fn from_high_entropy_bits(value: u64) -> Self {
         Self(NonZeroU64::new(value).expect("Id must be non-zero."))
     }
+
+    pub fn try_highlight(self, ctx: &crate::Context) {
```
emilk, line 122:
```suggestion
    /// Paint a rectangle around the widget, if it can be found.
    pub fn try_highlight(self, ctx: &crate::Context) {
```

crates/egui/src/id.rs
```diff
@@ -101,11 +116,237 @@ impl Id {
...
+            );
+        }
+
+        if let Some(area_rect) = ctx.memory(|mem| mem.area_rect(self)) {
+            ctx.debug_painter()
+                .debug_rect(area_rect, Color32::RED, self.short_debug_format());
+        }
+    }
+
+    pub fn ui(self, ui: &mut crate::Ui) -> crate::Response {
```
emilk, line 138: I would love to have a image snapshot test of this

crates/egui/src/id.rs
```diff
@@ -101,11 +116,237 @@ impl Id {
...
+    ///
+    /// If it is, it returns `IdSource::Id`, otherwise it returns `IdSource::Other`.
+    fn get_source<T: AsId>(t: T) -> IdSource {
+        let mut hasher = ExtractIdHasher::default();
+
+        t.hash(&mut hasher);
+
+        let maybe_source_id = hasher.id();
+
+        // Ideally we would just implement IdTriat for Id with specialization, but that's not
```
emilk: `IdTriat`?

crates/egui_demo_lib/tests/snapshots/demos/ID Test.png
emilk, line 1: That's not much of a hierarchy though. What is even `152204…` there?

emilk: Is that the auto-id? If so, we should maybe create a special path for creating auto-ids just to improve this formatting

crates/egui/src/id.rs
```diff
@@ -101,11 +116,237 @@ impl Id {
...
+                    write!(f, "{}", id.short_debug_format())
+                }
+                Self::Other(other) => {
+                    write!(f, "{other}")
+                }
+            }
+        }
+    }
+
+    static ID_MAP: LazyLock<RwLock<HashMap<Id, IdInfo>>> = LazyLock::new(|| {
```
emilk: `IdMap` is much faster, avoiding hashing of the already-hashed `Id`

```suggestion
    static ID_MAP: LazyLock<RwLock<IdMap<IdInfo>>> = LazyLock::new(|| {
```

emilk (approved):
Awesome! I think having this always-on in debug builds makes sense. The overhead is one extra lookup in a hashmap, which should be fine (at least once you change it from `ahash::HashMap` to an `IdMap`.

crates/egui_demo_lib/tests/snapshots/demos/ID Test.png
lucasmerlin: Will be fixed by https://github.com/emilk/egui/pull/7988. 

We could also add a extra Id constructor to pass more metadata, so we could display if it's an auto id, or the id for a Ui or Layer (some of this could also be derived by looking if any of those exist in the egui context for the id)

emilk:
Crossposting this comment:
* https://github.com/emilk/egui/pull/7988#issuecomment-4096669741]

Adding DNM until we have a screenshot test with a nicer UI :]

PR #5860, opened 2025-03-29
Remove a few unwrap

oscargus:

The aim was to do more, but most would be `if let(Some(...`-rewrites which I assume are not wanted.

### DIFF
--- a/crates/eframe/src/web/web_painter_glow.rs
+++ b/crates/eframe/src/web/web_painter_glow.rs
@@ -187,11 +187,7 @@ fn is_safari_and_webkit_gtk(gl: &web_sys::WebGlRenderingContext) -> bool {
     // This call produces a warning in Firefox ("WEBGL_debug_renderer_info is deprecated in Firefox and will be removed.")
     // but unless we call it we get errors in Chrome when we call `get_parameter` below.
     // TODO(emilk): do something smart based on user agent?
-    if gl
-        .get_extension("WEBGL_debug_renderer_info")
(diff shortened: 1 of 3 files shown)


crates/eframe/src/web/web_painter_glow.rs
```diff
@@ -187,11 +187,7 @@ fn is_safari_and_webkit_gtk(gl: &web_sys::WebGlRenderingContext) -> bool {
     // This call produces a warning in Firefox ("WEBGL_debug_renderer_info is deprecated in Firefox and will be removed.")
     // but unless we call it we get errors in Chrome when we call `get_parameter` below.
     // TODO(emilk): do something smart based on user agent?
-    if gl
-        .get_extension("WEBGL_debug_renderer_info")
-        .unwrap()
-        .is_some()
-    {
+    if gl.get_extension("WEBGL_debug_renderer_info").ok().is_some() {
```
emilk, line 190: This is a changes in behavior! What you want is probably something like

```suggestion
    if gl.get_extension("WEBGL_debug_renderer_info").ok().is_some_and(|ext| ext.is_some()) {
```

emilk (changes requested):

crates/eframe/src/web/web_painter_glow.rs
YgorSouza: Could just use `is_ok_and`. And should maybe log the error as well, since it would be printed with `unwrap`.

PR #5862, opened 2025-03-30
Update web-sys min version to 0.3.73

wareya:
This should prevent compilation errors (which I ran into) where eframe tries to use HtmlElement::set_autofocus(), which doesn't exist until 0.3.73.

```
error[E0599]: no method named `set_autofocus` found for struct `HtmlElement` in the current scope
  --> C:\Users\wareya\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\eframe-0.31.1\src\web\text_agent.rs:24:15
   |
24 |         input.set_autofocus(true)?;
   |               ^^^^^^^^^^^^^
   |
help: there is a method `set_onfocus` with a similar name
   |
24 |         input.set_onfocus(true)?;
   |               ~~~~~~~~~~~
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,7 +100,7 @@ thiserror = "1.0.37"
 type-map = "0.5.0"
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
-web-sys = "0.3.70"
+web-sys = "0.3.73"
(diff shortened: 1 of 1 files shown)


emilk:
Nice catch - thanks!

PR #5853, opened 2025-03-28
Fix a broken link in ARCHITECTURE.md

dmitryvm1:
A quick fix for a broken link: https://github.com/rerun/kittest -> https://github.com/rerun-io/kittest
* [X] I have followed the instructions in the PR template

PR #5868, opened 2025-03-31
Use lychee link checker instead of linkinator

lucasmerlin:
Seems like linkinator doesn't find any files: https://github.com/emilk/egui/pull/5853#issuecomment-2765526298

This will check all links in .md files (except CHANGELOG.md) and in toml files


emilk (approved):
Nice 🙏

Issue #3928, opened 2024-01-31
Grid stripes aren't the full grid width when column count is specified


Ved-s:
**Describe the bug**
`striped(true)` and `num_columns(N)` `Grid` paints only the minimum row width, though description of `num_columns` says that the last column will extend to fill the Ui

**To Reproduce**
for example,
```rust
ScrollArea::vertical()
  .max_width(f32::INFINITY)
  .auto_shrink([false, false])
  .show(ui, |ui| {
    egui::Grid::new("event_list_grid")
      .striped(true)
      .num_columns(4)
      .show(ui, |ui| {
        // draw some rows here
      });
  });
```

**Expected behavior**
It should fill entire row if `num_columns` is set

**Screenshots**
![image](https://github.com/emilk/egui/assets/53968411/6136f2e0-2e8c-4e99-bcf0-308fa84f3290)

zanciks:
Hi! Interested in this ticket. Is this as simple as checking inside of egui::Grid::paint_row() for if Grid.num_columns is_some(), and then extending the rect to the end of the ui?
```rust
    fn paint_row(&self, cursor: &Rect, painter: &Painter, available_width: f32) {
        // handle row color painting based on color-picker function
        let Some(color_picker) = self.color_picker.as_ref() else {
            return;
        };
        let Some(row_color) = color_picker(self.row, &self.style) else {
            return;
        };
        let Some(height) = self.prev_state.row_height(self.row) else {
            return;
        };
        // Paint background for coming row:
        let size = Vec2::new(self.prev_state.full_width(self.spacing.x), height);
        let rect = Rect::from_min_size(cursor.min, size);
        let rect = rect.expand2(0.5 * self.spacing.y * Vec2::Y);
        let rect = rect.expand2(2.0 * Vec2::X); // HACK: just looks better with some spacing on the sides

+       if self.num_columns.is_some() {
+           rect.set_width(available_width)
+       }

        painter.rect_filled(rect, 2.0, row_color);
    }
```
or is there a better way to go about this @emilk ?
Honestly, I'm not the biggest fan of adding `available_width: f32` as a function parameter, but only the painter is passed into the function, and I'm not sure how else to get it other than ui.available_width() inside of egui::Grid::show_dyn().
Thanks!

PR #5870, opened 2025-04-01
Make `Grid` take full width when `Self::num_columns` is `Some`

zanciks:
Hello!
**This is a PR to address ticket #3928. This PR would close the issue.**
I've made grid stripes take up the full width available to the Grid when Grid::num_rows is_some(). 
This change is being made to match the documentation, which states that: `Setting this will allow the last column to expand to take up the rest of the space of the parent` [(Link to comment)](https://docs.rs/egui/latest/egui/struct.Grid.html#method.num_columns)

```diff
    fn paint_row(&self, cursor: &Rect, painter: &Painter) {
        // ...
-       let rect = rect.expand2(2.0 * Vec2::X); // HACK: just looks better with some spacing on the sides
+       let mut rect = rect.expand2(2.0 * Vec2::X); // HACK: just looks better with some spacing on the sides
+
+       if self.num_columns.is_some() {
+          rect.set_width(cursor.width())
+       }

        painter.rect_filled(rect, 2.0, row_color);
    }
```
image

I'm a first time contributor to this repo, and I'm honestly not too familiar with how it works, so if there are any issues with this, please let me know and I'll investigate it ASAP!

### DIFF
--- a/crates/egui/src/grid.rs
+++ b/crates/egui/src/grid.rs
@@ -251,7 +251,11 @@ impl GridLayout {
         let size = Vec2::new(self.prev_state.full_width(self.spacing.x), height);
         let rect = Rect::from_min_size(cursor.min, size);
         let rect = rect.expand2(0.5 * self.spacing.y * Vec2::Y);
-        let rect = rect.expand2(2.0 * Vec2::X); // HACK: just looks better with some spacing on the sides
+        let mut rect = rect.expand2(2.0 * Vec2::X); // HACK: just looks better with some spacing on the sides
+
+        if self.num_columns.is_some() {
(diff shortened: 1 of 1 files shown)


emilk:
> Setting this will allow the last column to expand to take up the rest of the space of the parent

I think the keyword here is _allow_. It should not _always_ expand to the full width, but only do so if there is content in the last column that is wide enough.

You have failing snapshot tests btw

emilk:
You can see the what broke in the snapshot test:

Normal

Normal new

zanciks:
Sorry - forgot about this PR. I'll go ahead and close this since I misunderstood the docs, per previous comments. Thanks :)

Issue #1284, opened 2022-02-21
Font Book demo displays wrong glyph for some characters


derivator:
**Describe the bug**
The "Font Book" demo on [egui.rs](https://www.egui.rs/) displays the wrong glyph for some characters. Clicking them copies the correct character (matching the hover text) to the clipboard.

**To Reproduce**
1. Go to egui.rs
2. Click on Font Book
3. Type "fullwidth latin capital" into the filter box
4. The letter "Ｓ" shows an incorrect glyph
5. All following letters except Z are shifted back by one space in the alphabet, e.g. T shows S instead

**Screenshots**
![Screenshot from 2022-02-21 21-26-20](https://user-images.githubusercontent.com/2171264/155022286-1e084893-148f-470f-84eb-e9afc8501e6b.png)

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Browser: Firefox, Chromium, standalone application
 - Version: latest

emilk:
I can confirm this still happens on latest `master`. This is very strange!

emilk:
![Screen Shot 2022-02-21 at 22 16 04](https://user-images.githubusercontent.com/1148717/155027189-5b9787dd-fca1-4d71-9850-fb90d7ab40c3.png)

It gets weirder. Hovering the large `Ｓ` here shows the text for [Fullwidth Latin Capital Letter T](https://unicode-table.com/en/FF34/). Clicking it and pasting here, we get: `Ｔ`. Pasting that `Ｔ` into egui will make it sometimes show up as `Ｓ`, and sometimes as `ｔ` (lowercase!) depending on where in `egui` you paste it into.

derivator:
>Hovering the large `Ｓ` here shows the text for [Fullwidth Latin Capital Letter T](https://unicode-table.com/en/FF34/). Clicking it and pasting here, we get: `Ｔ`. 

Yes, that's what I meant by point 5. It seems like the appearance of that one wrong glyph for the S has been wedged in there, shifting the ones after it back. Y is shown as X, the Y glyph is nowhere to be seen, only Z is correct again. 

The lowercase "t" was also a genuine WTF for me, but it's easily explained: the filter box (intentionally?) does not display capital letters at all, so pasting the uppercase Ｔ (that incorrectly shows as Ｓ) there *correctly* displays the corresponding lowercase ｔ.

emilk:
>  but it's easily explained: the filter box (intentionally?) does not display capital letters at all

[...]

PR #5877, opened 2025-04-04
Update font: github.com/jslegers/emoji-icon-font

danielhjacobs:
Replace the current emoji-icon-font.ttf with the updated font from https://github.com/jslegers/emoji-icon-font/pull/19 to address https://github.com/emilk/egui/issues/1284. The second commit then removes the now unnecessary hack.

* Addresses https://github.com/emilk/egui/issues/1284


danielhjacobs:
I can include the revert here in a second commit if preferred. It should just be removing this code:

https://github.com/emilk/egui/blob/d78fc39386e28ec7a18bd48a28a688177f090bd4/crates/epaint/src/text/font.rs#L152-L158

Thoughts?

emilk (approved):
Nice; thank you!

danielhjacobs:
I updated the image tests. I guess the updated font has some slight differences made by fontforge when saving the update.

emilk (approved):

emilk:
It's a bit surprising that the image tests fail (except the font book, where the number of support characters increased; as expected).

emilk:
@lucasmerlin maybe you can take a look at this? 🙏

Issue #5682, opened 2025-02-05
Scene doesn't nest


Masterchef365:
**Describe the bug**
I'm very aware that this is a niche use-case, and must admit that I don't have an immediate application for it, but: nesting `Scene` doesn't provide the desired effect. 

**To Reproduce**
Steps to reproduce the behavior (tested on master [0db56dc](https://github.com/emilk/egui/commit/0db56dc9f1a8459b5b9376159fab7d7048b19b65)):
```rust
use eframe::egui::{self, Rect};

fn main() -> eframe::Result {
    let mut scene_rect = Rect::ZERO;
    let mut inner_rect = Rect::ZERO;

    eframe::run_simple_native("My egui App", Default::default(), move |ctx, _frame| {

        egui::CentralPanel::default().show(ctx, |ui| {
            ui.label("Outside the scenes");

            egui::Scene::new().show(ui, &mut scene_rect, |ui| {
                ui.label("Inside first scene");

                egui::Frame::canvas(ui.style()).show(ui, |ui| {
                    egui::Scene::new().show(ui, &mut inner_rect, |ui| {
                        ui.label("I'm in a nested scene!");
                    });
                });

            });
        });
    })
}
```

**Expected behavior**
Ideally, the contents of the sub-scene should be be scaled, transformed, and clipped appropriately. 

**Screenshots**
![Image](https://github.com/user-attachments/assets/e1a05e30-52a1-46b3-a115-413c7c124570)

**Additional context**
Again, I don't have an immediate use-case for this so it probably shouldn't be a priority.

mitchmindtree:
@Masterchef365 you might be interested in [this PR](https://github.com/emilk/egui/pull/5884) comment - it describes one approach for making a nested container that has a separate layer ID (like an `Area`, `Window`, or another `Scene`) track the scene.

PR #5884, opened 2025-04-07
refactor: `Scene` - set transform layer before calling user content

mitchmindtree:
This changes the `Scene` behaviour to call `set_transform_layer` prior to calling the user content fn, rather than after.

### Motivation

This provides a simple way for the user to access the `TSTransform` that will be applied to the `Scene` within the user content function, e.g.

```rust
ui.ctx().layer_transform_to_global(ui.layer_id())
```

Previously getting the transform like this still kind of worked, but resulted in the user content lagging behind the actual scene position by a single frame, which looks a bit strange.

With this PR, the user content using the transform no longer lags by a frame, and matches the scene's transform perfectly.

Accessing the `TSTransform` of the `Scene` within the user content function is useful for the case where the user may want to instantiate new `Ui` sublayers that also track the scene (by default, sublayers do *not* apply the same transform as the scene, likely the cause of #5682). With these changes, the user can have sublayers track the scene like so:

```rust
let scene_layer = ui.layer_id();
let sub_layer = egui::LayerId::new(scene_layer.order, self.id);
ui.ctx().set_sublayer(scene_layer, sub_layer);
let scene_transform = ui.ctx().layer_transform_to_global(scene_layer).unwrap();
ui.ctx().set_transform_layer(sub_layer, scene_transform);
```

### Tested with

- `egui_demo_app` scene example.
- Local `egui_graph` demo example.

---

### DIFF
--- a/crates/egui/src/containers/scene.rs
+++ b/crates/egui/src/containers/scene.rs
@@ -160,16 +160,16 @@ impl Scene {
         // Set a correct global clip rect:
         local_ui.set_clip_rect(to_global.inverse() * outer_rect);
 
+        // Tell egui to apply the transform on the layer:
+        local_ui
+            .ctx()
+            .set_transform_layer(scene_layer_id, *to_global);
+
         // Add the actual contents to the area:
         let ret = add_contents(&mut local_ui);
 
         // This ensures we catch clicks/drags/pans anywhere on the background.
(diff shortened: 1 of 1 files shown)


lucasmerlin (approved):
Thanks! Makes sense to me

Issue #5891, opened 2025-04-08
`Scene`: Allow for customising what button can be used for clicking and dragging, allow for providing custom `Sense`


mitchmindtree:
Currently, the `Scene` allocates a response with `Sense::click_and_drag` and checks `response.dragged()` to determine whether the scene should be panned by `response.drag_delta()`. This is great for users who are happy to have the scene dragged with any button.

However, this is unsuitable for those who wish to only drag with a particular button (e.g. `Middle`) or to have no drag behaviour at all (e.g. maybe they only want to pan the scene via scrolling).

### Potential Solution

- Allow for specifying a custom `Sense` for the scene. This would allow users to disable the `Sense::click_and_drag` behaviour if they wish in favour of `Sense::hover`.
- Allow for optionally specifying a `.drag_pointer_button(PointerButton)` to enable dragging only with a specific pointer button.

Ideally, we could just add the custom `sense` method, and then allow the user to check for whatever interaction they would like to pan the scene themselves, though I'm not sure how we'd expose the ability to pan the scene smoothly from inside the user content function 🤔

PR #5892, opened 2025-04-08
feat: Add `Scene::drag_pan_buttons` option. Allows specifying which pointer buttons pan the scene by dragging.

mitchmindtree:
This adds an option for specifying the set of pointer buttons that can be used to pan the scene via clicking and dragging.

The original behaviour where all buttons can pan the scene by default is maintained.

Addresses part of #5891.

---

Edit: It looks like the failing test is unrelated and also appears on master: https://github.com/emilk/egui/actions/runs/14330259861/job/40164414607.

### DIFF
--- a/crates/egui/src/containers/mod.rs
+++ b/crates/egui/src/containers/mod.rs
@@ -29,7 +29,7 @@ pub use {
     panel::{CentralPanel, SidePanel, TopBottomPanel},
     popup::*,
     resize::Resize,
-    scene::Scene,
+    scene::{DragPanButtons, Scene},
     scroll_area::ScrollArea,
     sides::Sides,
     tooltip::*,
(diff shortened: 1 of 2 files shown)


lucasmerlin:
It seems like left click drag is now broken in the demo 🤔

mitchmindtree:
Agh good catch @lucasmerlin! I mistakenly switched back to incrementing values for the bitflags fields which resulted in conflicts between them, and then must have only tested with my downstream example where I only use `MIDDLE` :woozy_face:

I've adjusted them to properly use separate bit values that don't conflict. I've tested the behaviour of `DragPanButtons::empty()`, `DragPanButtons::all()`, `DragPanButtons::PRIMARY` and `DragPanButtons::MIDDLE` in the `egui_demo_app` and all seem to now work as expected.

lucasmerlin (approved):
Awesome!

PR #5893, opened 2025-04-08
feat: Add `Scene::sense` option for customising how `Scene` should respond to user input

### LINKED ISSUE #5891: `Scene`: Allow for customising what button can be used for clicking and dragging, allow for providing custom `Sense`

mitchmindtree:
Allows for specifying how the `Scene` should respond to user input.

With #5892, closes #5891.

---

Edit: Failing tests unrelated and appear on master: https://github.com/emilk/egui/actions/runs/14330259861/job/40164414607.

### DIFF
--- a/crates/egui/src/containers/scene.rs
+++ b/crates/egui/src/containers/scene.rs
@@ -69,6 +69,7 @@ impl Scene {
     /// The default is `Sense::click_and_drag()`.
     ///
     /// Set this to `Sense::hover()` to disable panning via clicking and dragging.
+    #[inline]
     pub fn sense(mut self, sense: Sense) -> Self {
         self.sense = sense;
         self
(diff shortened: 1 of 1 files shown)


lucasmerlin (approved):
Nice!

PR #5978, opened 2025-04-11
Add a `Slider::update_while_editing(bool)` API

### LINKED ISSUE #5976: `Slider` widget: support not updating the value while editing the text input

The `Slider` widget modifies the value after every text edit that can be parsed. In some cases it would be desirable to only apply the edits when the user finishes typing and accepts the final value.

I would like to be able to configure whether the `Slider` widget applies the text edits immediately or only once the user accepts the text input.

There is already a precedent for this in the `DragValue` widget with `DragValue::update_while_editing`.

And, in fact, the `Slider` widget itself contains an internal `DragValue` widget, it just doesn't expose its configuration. So, this might be just a matter of adding a `Slider::update_while_editing` API and propagating the setting to the internal `DragValue` widget.

- Passing an intermediate value to the `Slider` widget and only updating the actual value based on that when the slider loses focus. This would be cumbersome compared to the proposed solution.
- Duplicating and modifying the slider widget. Needless to say, this would not not be very maintainable.

**Additional context**
 
N/A

mbernat:
* Closes #5976
  - The `scripts/check.sh` fails in `cargo-deny` because of a security vulnerability in `crossbeam-channel` but this is also present on `master`.

https://github.com/user-attachments/assets/a964c968-bb76-4e56-88e1-d1e3d51a401a

### DIFF
--- a/crates/egui/src/widgets/slider.rs
+++ b/crates/egui/src/widgets/slider.rs
@@ -117,6 +117,7 @@ pub struct Slider<'a> {
     custom_parser: Option<NumParser<'a>>,
     trailing_fill: Option<bool>,
     handle_shape: Option<HandleShape>,
+    update_while_editing: bool,
 }
 
 impl<'a> Slider<'a> {
@@ -167,6 +168,7 @@ impl<'a> Slider<'a> {
             custom_parser: None,
             trailing_fill: None,
             handle_shape: None,
+            update_while_editing: true,
         }
     }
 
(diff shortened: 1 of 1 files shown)


mbernat:
@lucasmerlin Thanks for the review! Do you have a suggestion for how the slider not updating should work? We would need a place to store the temporary value until you release the slider and I don't see a good place to put it. What would be an idiomatic way to do it? A hidden internal widget?

FWIW, `DragValue::update_while_editing` also affects only text edits, not dragging. It might be clearer to call the new API `Slider::update_while_text_editing` , though the naming inconsistency would be unfortunate.

lucasmerlin:
Ah right, that would add a lot of complexity. Don't worry about it, it was just wishful thinking haha.

mbernat:
@lucasmerlin I see, yeah, let's leave that for another day :)

I rebased on `master` to get the CI fixed. Is there anything else I should do once the CI passes?

lucasmerlin:
No, I usually go over the approved PRs once per week with @emilk where we merge everything so he stays in the loop, so it might take a couple of days to get merged

PR #6744, opened 2025-04-14
Update to wgpu 25

torokati44:
See for changelogs:
 - https://github.com/gfx-rs/wgpu/releases/tag/v25.0.0
 - https://github.com/gfx-rs/wgpu/releases/tag/v25.0.1

### DIFF
--- a/crates/egui-wgpu/src/lib.rs
+++ b/crates/egui-wgpu/src/lib.rs
@@ -104,7 +104,7 @@ async fn request_adapter(
             force_fallback_adapter: false,
         })
         .await
-        .ok_or_else(|| {
+        .map_err(|e| {
             #[cfg(not(target_arch = "wasm32"))]
             if _available_adapters.is_empty() {
                 log::info!("No wgpu adapters found");
@@ -121,7 +121,7 @@ async fn request_adapter(
                 );
             }
 
-            WgpuError::NoSuitableAdapterFound("`request_adapters` returned `None`".to_owned())
+            WgpuError::NoSuitableAdapterFound(format!("`request_adapters` returned `Err`: {e}"))
         })?;
 
     #[cfg(target_arch = "wasm32")]
@@ -184,7 +184,6 @@ impl RenderState {
                 power_preference,
                 native_adapter_selector: _native_adapter_selector,
                 device_descriptor,
-                trace_path,
             }) => {
                 let adapter = {
                     #[cfg(target_arch = "wasm32")]
@@ -209,7 +208,7 @@ impl RenderState {
                 let (device, queue) = {
                     profiling::scope!("request_device");
                     adapter
-                        .request_device(&(*device_descriptor)(&adapter), trace_path.as_deref())
+                        .request_device(&(*device_descriptor)(&adapter))
                         .await?
                 };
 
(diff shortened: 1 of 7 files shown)


crates/egui-wgpu/src/lib.rs
```diff
@@ -104,7 +104,7 @@ async fn request_adapter(
             force_fallback_adapter: false,
         })
         .await
-        .ok_or_else(|| {
+        .map_err(|e| {
```
torokati44, line 107: This probably requires some thoughts to be put in...

crates/egui-wgpu/src/setup.rs
```diff
@@ -195,6 +195,17 @@ impl Default for WgpuSetupCreateNew {
...
+                    /*
+                    // TODO:
+                    #[cfg(feature = "dep:wgpu-core/trace")]
+                    wgpu::Trace::Directory(
+                        std::env::var("WGPU_TRACE")
+                            .ok()
+                            .map(std::path::PathBuf::from),
+                    )
+                     */
+                    trace: wgpu::Trace::Off,
```
torokati44, line 208: I'm not sure how this should work...

crates/egui_kittest/src/wgpu.rs
```diff
@@ -28,7 +28,7 @@ pub fn default_wgpu_setup() -> egui_wgpu::WgpuSetup {
             wgpu::Backend::Dx12 => 2,
             wgpu::Backend::Gl => 4,
             wgpu::Backend::BrowserWebGpu => 6,
-            wgpu::Backend::Empty => 7,
+            wgpu::Backend::Noop => 7,
```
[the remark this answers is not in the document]
torokati44: See: https://github.com/gfx-rs/wgpu/pull/7063

crates/egui-wgpu/src/lib.rs
```diff
@@ -121,7 +121,7 @@ async fn request_adapter(
                 );
             }
 
-            WgpuError::NoSuitableAdapterFound("`request_adapters` returned `None`".to_owned())
+            WgpuError::NoSuitableAdapterFound(format!("`request_adapters` returned `Err`: {}", e))
```
torokati44, line 124: Perhaps it would be better to do this...?

```rust
pub enum WgpuError {
    #[error(transparent)]
    NoSuitableAdapterFound(#[from] wgpu::RequestDeviceError),
```

But then that would spread into `NativeAdapterSelectorMethod`...

crates/egui-wgpu/src/setup.rs
```diff
@@ -195,12 +186,9 @@ impl Default for WgpuSetupCreateNew {
                         ..base_limits
                     },
                     memory_hints: wgpu::MemoryHints::default(),
+                    trace: wgpu::Trace::Off,
```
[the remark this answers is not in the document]
torokati44: See: https://github.com/emilk/egui/pull/6860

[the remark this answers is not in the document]
Wumpf: It's not exposed right now because the feature is a bit on hiatus right now https://github.com/gfx-rs/wgpu/issues/5974
it's still used by Mozilla internally which uses wgpu-core (rather than wgpu)

crates/egui-wgpu/src/lib.rs
Wumpf: I solved that now by exposing this as a separate error 🤷

Wumpf (approved):
thank you for the update and sorry it took so long!
Looks all good to me, but went the suggested path of adding a separate error to expose the wgpu adapter selection error directly

PR #6750, opened 2025-04-15
Add external eventloop support

### LINKED ISSUE #2875: [Eframe] Allow `eframe` to run on an existing event loop

wpbrown:
* Closes #2875
* Closes https://github.com/emilk/egui/pull/3340

Adds `create_native`. Similiar to `run_native` but it returns an `EframeWinitApplication` which is a `winit::ApplicationHandler`. This can be run on your own event loop. A helper fn `pump_eframe_app` is provided to pump the event loop and get the control flow state back.

I have been using this approach for a few months.


wpbrown:
I provided an example of running an event loop on top of tokio. I only use Linux, so I could use help with other native platforms.

wpbrown:
If this approach is viable I will work on doc comments.

lucasmerlin:
This looks pretty cool! Just out of curiosity, what are the benefits of running egui in an async way like your example? Versus, just spawning the tokio runtime in another thread. 
Is it that you can do spawn_local from the ui and that you only need a single thread?

wpbrown:
> This looks pretty cool! Just out of curiosity, what are the benefits of running egui in an async way like your example? Versus, just spawning the tokio runtime in another thread. Is it that you can do spawn_local from the ui and that you only need a single thread?

Yes exactly. I am building an app with a lot of CRUD. There is plenty of CPU available to run the UI and the async tasks. Not having to worry about anything being `Sync` greatly simplifies development (as seen in the example, the UI and background task can share data without locks). When you need to run anything CPU intensive you can use `spawn_blocking`.

lucasmerlin:
This looks great! Please add the missing doc comments and maybe also mention in the example readme why you'd want to do this (basically what you explained in the comment).

wpbrown:
@lucasmerlin I updated the readmes and added doc comments.

I think I fixed the CI issues for windows, ios, mac. I was able to run check --all-features --all-targets on a windows VM. I used required-features to prevent building the linux example unless the `linux-example` feature is specifically added. Unfortunately the check script and CI run with "--all-features", so it tries to build the example on other platforms.

FWIW, I attempted to extend the async example to windows, but it turns out it's not possible with tokio. It would require changes to mio. To block waiting for GUI events on windows you need to use WaitForMultipleObjectsEx. I tried running it on a background thread, but it only works on the thread that created the window. Hopefully in the future someone who cares more about windows or mac will update the example with async executors that supports waiting for GUI events on their platforms.

lucasmerlin (approved):
Awesome!

wpbrown:
@lucasmerlin Sorry the update to 1.84 on master broke CI. Here is a successful CI run with master merged in https://github.com/wpbrown/egui/actions/runs/14708125657

PR #6752, opened 2025-04-15
Add tests for layout and visuals of most egui widgets

lucasmerlin:
This is mostly in preparation for #5830 where I want to ensure that I don't introduce any regressions


crates/egui_kittest/src/lib.rs
```diff
@@ -333,10 +341,34 @@ impl<'a, State> Harness<'a, State> {
...
+    /// # Errors
+    /// Returns an error if the maximum number of steps is exceeded.
+    ///
+    /// See also:
+    /// - [`Harness::run`].
+    /// - [`Harness::run_ok`].
+    /// - [`Harness::step`].
+    /// - [`Harness::run_steps`].
+    /// - [`Harness::try_run`].
+    pub fn try_run_async(&mut self) -> Result<u64, ExceededMaxStepsError> {
```
emilk, line 368: Not so sure about this name. `try_run_realtime` maybe?

emilk (approved):
I love it

PR #6753, opened 2025-04-15
Change popup memory to be per-viewport

mkalte666:
Starting with 77244cd4c5ebb924beaec8f89604a2881106b27c  the popup open-state is cleaned up per memory pass.

This becomes problematic for implementations that share memory between viewports (i.e. all of them, as far as i understand it), because each viewport gets a context pass, and thus a memory pass, which cleans out popup open state.

To illustrate my issue, i have modifed the multiple viewport example to include a popup menu for the labels: https://gist.github.com/mkalte666/4ecd6b658003df4c6d532ae2060c7595 (changes not included in this pr). 

Then, when i try to use the popups, i get this: 

https://github.com/user-attachments/assets/7d04b872-5396-4823-bf30-824122925528

Immediate viewports just break popup handling in general, while deferred viewports kinda work, or dont. In this example ill be honest, it kind of still did, sometimes. In my more complex app with multiple viewports (where i encountered this first) it also just broke - even when just showing root and one nother. Probably to do with the order wgpu vs glow draws the viewports? Im not sure. In any case: 

This commit adds `Memory::popup` (now `Memory::popups`) to the per-viewport state of memory, including viewport aware cleanup, and adding it to the list of things cleaned if a viewport goes away.

This results in the expected behavior:

https://github.com/user-attachments/assets/fd1f74b7-d3b2-4edc-8dc4-2ad3cfa2160e

I will note that with this, changing focus does not cause a popup to be closed, which is consistent with current behavior on a single app. 

Hope this helps 

~Malte 

* [x] ~~I have run check.sh locally~~ CI on the fork, including checks, went through.

### DIFF
--- a/crates/egui/src/memory/mod.rs
+++ b/crates/egui/src/memory/mod.rs
@@ -87,15 +87,6 @@ pub struct Memory {
     #[cfg_attr(feature = "persistence", serde(skip))]
     pub(crate) viewport_id: ViewportId,
 
-    /// Which popup-window is open (if any)?
-    /// Could be a combo box, color picker, menu, etc.
-    /// Optionally stores the position of the popup (usually this would be the position where
-    /// the user clicked).
-    /// If position is [`None`], the popup position will be calculated based on some configuration
-    /// (e.g. relative to some other widget).
-    #[cfg_attr(feature = "persistence", serde(skip))]
-    popup: Option<OpenPopup>,
-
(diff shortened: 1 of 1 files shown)


lucasmerlin (approved):
Thanks! I think a similar bug could also exsit for menus, where I also do some cleanup if the memory wasn't shown for a frame. But it's implemented using the frame count there, so maybe it's fine.

PR #6754, opened 2025-04-15
Area positioning

mkalte666:
When i did the quick implementation for movable modals for https://github.com/emilk/egui/pull/6749 @lucasmerlin [pointed out that](https://github.com/emilk/egui/pull/6749#discussion_r2044142830) that just forcibly recentering a modal if it is movable is a bit odd. 

It would be nicer if this was an inherent property of Areas. 

This is a surprisingly small change set, considering the wide impact it has, so i went with a pr directly, instead of an issue. 

Investigating this, i figured  that some area is quite Stateful in its positioning. 

Instead of keeping default_pos, new_pos, anchor as seperate entities, i have implemented its internal position using a new enum, `AreaPosition`, that now encodes the different types of position an area can have. 
  * Is it at a fixed position?
  * Is it anchored somewhere? 
  * Is it moveable? 
    - If it is indeed movable, what is its initial position? 

I have also quickly thrown together a demo, that i will expand on with test cases. 

The way it is done now is that all the old builder methods still exist and behave the same, though the colliding states (setting area movable while an anchor is set, for example) now overwrite each other. In practice this should not change anything, i think? 

Things that are still to-do before this can be merged in the first place:  
-  [ ] Anchored windows are currently not resizable, though they should be
-  [ ] Investigate how pivot affected anchors before, and mirror this behavior correctly 
-  [ ] Investigate how pivot affected auto position before, and mirror that behavior correctly. 
-  [ ] Clean up the demo, and the comments of the builder methods 
-  [ ] `Area::get_pivot` previously only reported TOP_LEFT; or if set, the anchor. Was that correct? - im now reporting the pivot instead. Needs trying out.

### DIFF
--- a/crates/egui/src/containers/area.rs
+++ b/crates/egui/src/containers/area.rs
@@ -135,6 +135,10 @@ pub struct AreaState {
     /// Used to fade in the area.
     #[cfg_attr(feature = "serde", serde(skip))]
     pub last_became_visible_at: Option<f64>,
+
+    /// Was the last frame a sizing pass?
+    #[cfg_attr(feature = "serde", serde(skip))]
+    pub last_was_sizing_pass: bool,
 }
 
 impl Default for AreaState {
@@ -145,6 +149,7 @@ impl Default for AreaState {
             size: None,
             interactable: true,
             last_became_visible_at: None,
+            last_was_sizing_pass: false,
         }
     }
 }
@@ -550,6 +555,7 @@ impl Area {
             size: None,
             interactable,
             last_became_visible_at: None,
+            last_was_sizing_pass: false,
         });
         state.pivot = pivot;
         state.interactable = interactable;
@@ -603,7 +609,9 @@ impl Area {
                     state.pivot_pos = Some(override_pos);
                 } else {
                     // If we are shown for the first time ever, or we reset the position on re-open, set a new position.
-                    if reset_position_on_open && !visible_last_frame {
+                    if reset_position_on_open
+                        && (!visible_last_frame || sizing_pass || state.last_was_sizing_pass)
+                    {
                         state.pivot_pos = None;
                     }
                     state
@@ -612,7 +620,10 @@ impl Area {
                             InitialAreaPosition::Automatic => {
                                 automatic_area_position(ctx, layer_id)
                             }
-                            InitialAreaPosition::Centered => ctx.screen_rect().center(),
+                            InitialAreaPosition::Centered => {
+                                let offset = self.pivot.to_sign() * size * 0.5;
+                                ctx.screen_rect().center() + offset
+                            }
                             InitialAreaPosition::Position(default_pos) => default_pos,
                         });
                 }
@@ -761,6 +772,7 @@ impl Prepared {
         } = self;
 
         state.size = Some(content_ui.min_size());
+        state.last_was_sizing_pass = sizing_pass;
 
         // Make sure we report back the correct size.
         // Very important after the initial sizing pass, when the initial estimate of the size is way off.
--- a/crates/egui/src/containers/area.rs
+++ b/crates/egui/src/containers/area.rs
@@ -9,6 +9,102 @@ use crate::{
     Response, Sense, Ui, UiBuilder, UiKind, UiStackInfo, Vec2, WidgetRect, WidgetWithState,
 };
 
+/// Area positioning.
+///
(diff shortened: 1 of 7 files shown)


crates/egui/src/containers/area.rs
```diff
@@ -9,6 +9,102 @@ use crate::{
...
+    /// A moveable position that starts centered
+    pub fn moveable_centered(reset_position_on_open: bool) -> Self {
+        Self::Moveable {
+            initial_position: InitialAreaPosition::Centered,
+            reset_position_on_open,
+            current_position_override: None,
+        }
+    }
+
+    /// A moveable position that starts centered
```
lucasmerlin, line 65:
```suggestion
    /// A moveable position that starts with a automatic position
```

crates/egui/src/containers/area.rs
```diff
@@ -191,11 +286,28 @@ impl Area {
...
+    /// In case the area is set to a moveable positioning mode, and you set this to false, the area will default to anchoring in the center.
     #[inline]
     pub fn movable(mut self, movable: bool) -> Self {
-        self.movable = movable;
-        self.interactable |= movable;
+        if !movable && self.is_movable() {
+            self.position = AreaPosition::anchored(Align2::CENTER_CENTER, [0.0, 0.0]);
+        }
+
         self
```
lucasmerlin, line 311: What about the interactable flag?

crates/egui/src/containers/area.rs
```diff
@@ -310,10 +433,20 @@ impl Area {
...
+    /// If the position is anchored, this will do nothing.
+    /// If the current position is movable, this will override movement input.
+    /// Finally, if the position is fixed, this is the same as calling [`Area::fixed_pos`]
     #[inline]
     pub fn current_pos(mut self, current_pos: impl Into<Pos2>) -> Self {
-        self.new_pos = Some(current_pos.into());
+        match &mut self.position {
+            AreaPosition::Fixed(pos) => *pos = current_pos.into(),
+            AreaPosition::Anchored { .. } => {}
+            AreaPosition::Moveable {
```
lucasmerlin, line 445: Maybe we should add a debug_assert or at least a log warning when with the current position type something isn't supported?

crates/egui/src/containers/area.rs
```diff
@@ -463,14 +590,45 @@ impl Area {
...
+                initial_position,
+                reset_position_on_open,
+                current_position_override,
+            } => {
+                if let Some(override_pos) = current_position_override {
+                    state.pivot_pos = Some(override_pos);
+                } else {
+                    // If we are shown for the first time ever, or we reset the position on re-open, set a new position.
+                    if reset_position_on_open
+                        && (!visible_last_frame || sizing_pass || state.last_was_sizing_pass)
```
lucasmerlin, line 613: Why is the position reset on sizing passes?

crates/egui/src/containers/area.rs
mkalte666: ... Oops , forgot that

mkalte666: Logic wise, The "movable" window can still be positioned via override_pos (or current_pos).  So i think a note saying "you can set this to movable, and disable interactions, so the window will never actually be movable in the end.

I don't know when you'd want this, though. 

I'll think about this more.

crates/egui/src/containers/area.rs
lucasmerlin: Makes sense!

crates/egui/src/containers/area.rs
```diff
@@ -9,6 +9,102 @@ use crate::{
...
+pub enum InitialAreaPosition {
+    /// Use the automatic positioning, looking for a good space on screen.
+    #[default]
+    Automatic,
+
+    /// Put the Area in the center of the context
+    Centered,
+
+    /// Put the area at a position. This will take [`Area::pivot`] into consideration.
+    Position(Pos2),
```
emilk, line 105: Would it make sense to move `pivot` into here?

mkalte666: That would make pivot only affect the initial position, instead of any current position. (i.g what is set via `Area::current_position` for overriding). That said, it might make sense to make it a property only of movable windows and fixed positioned windows -  as far as i can see it anchoring always anchors corners to coners / center to center + offset.

Issue #5850, opened 2025-03-27
Disabled image button fade problem


ugosantana:
Hi there!

I'm using image buttons for a toolbar. In certain conditions, some of them are disabled. When disabled, the images should fade a bit. They do, but only in dark mode. I'm using egui 0.31.1.

I compiled the program for Mac and Windows, and in both cases this happens only in light mode.

I believe it is a bug... I attached an image to show what I'm talking about...

<img width="653" alt="Image" src="https://github.com/user-attachments/assets/a8ab984d-a4ad-482f-911d-28dfd48ce423" />

lucasmerlin:
Interesting, I can also reproduce it on egui.rs: 

https://github.com/user-attachments/assets/01eef48f-5b9c-485e-b5be-49bb727b3e5c

emilk:
* Closed by https://github.com/emilk/egui/pull/7282

PR #6765, opened 2025-04-16
Changed `ui.disable()` to modify opacity.

tye-exe:
It previously tinted elements, which could have lackluster effects.

* Closes <https://github.com/emilk/egui/issues/5850>

## New looks
(I say new but I tried to keep them as similar as possible to previous)
### Light mode
[light_mode.webm](https://github.com/user-attachments/assets/e262fdec-8b33-47f9-b02e-591431d2a94f)
### Dark mode
[dark_mode.webm](https://github.com/user-attachments/assets/0b5737d0-d2fc-4095-a3a0-843da256ef6c)

## Additional remarks
Since the opacity is being change, elements that are on-top of other elements now show some of the underneath element (this can be seen with the slider in the sent videos).


juh9870:
This has far-reaching consequences for the visual of existing egui apps other than just fixing the linked issue.

I think this change should be configurable; otherwise, significant considerations should be put into messing up existing visuals.

tye-exe:
That's a fair viewpoint.

What other possible solutions to this issue do you recommend?

I tried modifying the `fade_out_to_color` method (as also suggested in the issue) to modify the alpha channel. However, this was not able to prevent the issue on light mode, even at the extremes of the alpha channel.

Relevant method:
https://github.com/emilk/egui/blob/5d6aaa239b98ad20d365626dc664531b08b2537d/crates/egui/src/style.rs#L1055-L1060

Out of interest in which situations will this be an issue for (so I could potentially work on solving them)?

juh9870:
The problem is exactly what you mentioned in the `Additional remarks` sections.

I believe that a method of fading out (color or transparency) should be an option in the style. Maybe as an enum variant akin to
```
enum FadeOutMethod {
    NonInteractiveWeakBgFill,
    CustomColor(Color32),
    DimTransparency(f32)
}
```
the exact shape of variants and naming should be discussed ofc

tye-exe:
I agree with being able to override the fade out colour, but that does lean towards a more major change, as `Visuals` is a commonly used data structure.

Also adding more default colours increases the maintenance burden in the future. i'm not sure if that would be needed if the user can customise the fade out colour.

crates/egui/src/ui.rs
```diff
@@ -521,8 +521,7 @@ impl Ui {
     pub fn disable(&mut self) {
         self.enabled = false;
         if self.is_visible() {
-            self.painter
-                .set_fade_to_color(Some(self.visuals().fade_out_to_color()));
```
emilk, line 525: We should also remove `Visuals::fade_out_to_color` too, and replace it with `fn fade_out_opacity(&self) -> f32 { 0.5 }` and use it for `Visuals::gray_out`

tye-exe:
Will it be a good idea to add an attribute to `Visuals` that is returned from `fade_out_opacity`?

Because currently the value is hard-coded, so it cannot be changed.

tye-exe:
I have added the option for user-customization that i suggested & created a [pull request](https://github.com/tye-exe/egui_image_fade/pull/1) against the branch in my fork for anyone to look at.

I can easily merge these changes into the branch in this egui pull request.

tye-exe:
emilk, do you want me to merge the user-customisation code?

emilk:
Yes, adding a `Visuals::disabled_alpha: f32` that default to 0.5 sounds like a great idea

crates/egui/src/style.rs
```diff
@@ -2062,6 +2065,7 @@ impl Visuals {
             image_loading_spinners,
 
             numeric_color_space,
+            disabled_alpha: fade_out_by,
```
emilk, line 2068: Let's use the same name everywhere

```suggestion
            disabled_alpha,
```

crates/egui/src/style.rs
```diff
@@ -1052,18 +1055,17 @@ impl Visuals {
         self.window_stroke
     }
 
-    /// When fading out things, we fade the colors towards this.
-    // TODO(emilk): replace with an alpha
+    /// When fading out things, we multiply the opacity by this.
     #[inline(always)]
-    pub fn fade_out_to_color(&self) -> Color32 {
-        self.widgets.noninteractive.weak_bg_fill
+    pub fn fade_out_opacity(&self) -> f32 {
```
emilk, line 1060: Let's use the same name everywhere

```suggestion
    pub fn disabled_alpha(&self) -> f32 {
```

emilk (approved):

emilk:
There is a problem with this approach: `Visuals::weak_text_color` uses `Visuals::gray_out`, but `gray_out` no longer makes the text gray; it makes it transparent. This may be fine, but we should probably rename `gray_out`, and we also need to update snapshot tests.

tye-exe:
I will be able to work on this at the weekend

tye-exe:
An issue with the modify opacity approach is that the "colour picker" widget does not look very good when disabled on light mode.
 
[Colour picker on light mode.webm](https://github.com/user-attachments/assets/36953341-f047-4493-88c2-b1eea6ac804f)

emilk (approved):

emilk:
> Should i re-add the `fade_out_to_color` function, but annotated it as deprecated? This will give time for users to stop using this function instead of having an instant compile error.

Yeah, why not.

You also need to update the snapshot images (again), because of a merge race (sorry!)

tye-exe:
I think that you have to run the CI to generate the images again for me to pull them from the CI, apologies.

tye-exe:
This should do it!

PR #6766, opened 2025-04-16
Convert `hello_android` into GameActivity

podusowski:
Since 0.24.0, `accesskit_winit` [does not support Android's NativeActivity](https://github.com/AccessKit/accesskit/blob/341a11bca2c8a29682c11ddcfe91fa58776ea11d/platforms/winit/README.md#android-activity-compatibility). Attempt of running it [fails at runtime](https://github.com/AccessKit/accesskit/blob/341a11bca2c8a29682c11ddcfe91fa58776ea11d/platforms/winit/src/platform_impl/android.rs#L35), which is the reason why it slipped through the CI.

Since there is no (or I am not aware of any) idiomatic way of building Java/Rust Android apps, I have added a simple `Makefile` which combines calling `cargo ndk`, `gradle` and `adb`.


emilk (commented):
This is a great example to have, but I think this should be a 3rd party crate instead. I want to keep this repository Rust specific.

emilk:
We can remove the current one and point to an external repository instead! The less code that is in this repository the better :)

lucasmerlin:
We could also disable the accesskit feature for the current example if we'd want to keep it.

PR #6848, opened 2025-04-22
Raise MSRV to 1.84

torokati44:
Prerequisite of https://github.com/emilk/egui/pull/6744.
See: https://github.com/gfx-rs/wgpu/pull/7218, https://github.com/gfx-rs/wgpu/pull/7425

Please be aware that Rust 1.84 enables some (more) WASM extensions by default, and ships with an `std` built with them enabled: https://blog.rust-lang.org/2024/09/24/webassembly-targets-change-in-default-target-features/
According to `rustc +1.84 --print=cfg --target wasm32-unknown-unknown`, these are: `multivalue`, `mutable-globals`, `reference-types`, and `sign-ext`.
(c.f. `rustc +1.84 --print=cfg --target wasm32-unknown-unknown -C target-cpu=mvp` enabling none.)
For reference: https://webassembly.org/features/

----

If support is desired for ancient/esoteric browsers that don't have these implemented, there are two ways to get around this:
 - Target `wasm32v1-none` instead, but that's a `no-std` target, and I suppose a lot of dependencies don't work that way (e.g. https://github.com/gfx-rs/wgpu/issues/6826)
 - Using the `-Ctarget-cpu=mvp`  and `-Zbuild-std=panic_abort,std` flags, and the `RUSTC_BOOTSTRAP=1` escape hatch to allow using the latter with non-`nightly` toolchains - until https://github.com/rust-lang/wg-cargo-std-aware is stabilized. (For reference: https://github.com/ruffle-rs/ruffle/pull/18528/files#diff-fb2896d189d77b35ace9a079c1ba9b55777d16e0f11ce79f776475a451b1825a)

I don't think either of these is particularly advantageous, so I suggest just accepting that browsers will have to have some extensions implemented to run `egui`.

### DIFF
--- a/.github/workflows/deploy_web_demo.yml
+++ b/.github/workflows/deploy_web_demo.yml
@@ -39,7 +39,7 @@ jobs:
         with:
           profile: minimal
           target: wasm32-unknown-unknown
-          toolchain: 1.81.0
+          toolchain: 1.84.0
           override: true
 
       - uses: Swatinem/rust-cache@v2
--- a/.github/workflows/preview_build.yml
+++ b/.github/workflows/preview_build.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: 1.81.0
+          toolchain: 1.84.0
           targets: wasm32-unknown-unknown
       - uses: Swatinem/rust-cache@v2
         with:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -5,7 +5,7 @@ name: Rust
 env:
   RUSTFLAGS: -D warnings
   RUSTDOCFLAGS: -D warnings
-  NIGHTLY_VERSION: nightly-2024-09-11
+  NIGHTLY_VERSION: nightly-2025-04-22
 
 jobs:
   fmt-crank-check-test:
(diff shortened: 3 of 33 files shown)


torokati44:
I suppose the `RUSTC_BOOTSTRAP=1` trick _could_ be used in the `Check wasm32+atomics` CI job in `rust.yml` as well, instead of `+${{env.NIGHTLY_VERSION}}`. This is a matter of preference I suppose - the current version is probably cleaner, but requires version bumping from time to time (but at least doesn't spuriously fail when a borked nightly toolchain is released occasionally.)

And about the new selected nightly version, https://rust-lang.github.io/rustup-components-history/wasm32-unknown-unknown.html shows it has at least been built successfully.

torokati44:
Also note that because of https://github.com/WebAssembly/binaryen/issues/4711 (see also: https://github.com/rustwasm/wasm-bindgen/issues/2943, https://github.com/rustwasm/wasm-bindgen/issues/4228), at least Binaryen (`wasm-opt`) 110 is required, with: https://github.com/WebAssembly/binaryen/pull/4736

Unfortunately, Ubuntu 22.04 ships with 105, and even 24.04 and 24.10 ships only with 108 - with 25.04 finally updating to 120 (it's currently at 123 BTW).

I recommend switching CI over to https://github.com/sigoden/install-binary/tree/v1/ from a simple `apt install` (we do the same over at Ruffle). Thoughts?

PR #6855, opened 2025-04-24
Disallow `accesskit` on Android NativeActivity, making `hello_android` working again

podusowski:
Follow up for https://github.com/emilk/egui/pull/6766

I wasn't sure if `compile_error` is appropriate. It felt right.


lucasmerlin (approved):
Nice!

crates/eframe/src/lib.rs
```diff
@@ -144,6 +144,15 @@
...
 
+// Limitation imposed by `accesskit_winit`:
+// https://github.com/AccessKit/accesskit/tree/accesskit-v0.18.0/platforms/winit#android-activity-compatibility`
+#[cfg(all(
+    target_os = "android",
+    feature = "accesskit",
+    feature = "android-native-activity"
+))]
+compile_error!("`accesskit` feature is only available with `android-game-activity`");
+
```
emilk, line 155: I appreciate improving the error message. Without this, what would the user see?

I'm a bit worried that maybe accesskit_winit removed this limitation, and then we forget to update this code

podusowski: Nothing. Then it panics at runtime on `unwrap`:

https://github.com/AccessKit/accesskit/blob/341a11bca2c8a29682c11ddcfe91fa58776ea11d/platforms/winit/src/platform_impl/android.rs#L35

DurhamG:
Correct me if I'm wrong, but since accesskit is a default feature doesn't this mean any usage of android-native-activity will require disabling all default features on eframe and reenabling them manually (minus accesskit)?  And if any library I depend on has eframe in its transitive dependency chain, and if that library does not also disable accesskit, then accesskit will be on and I will hit this error.  Meaning I cannot use a library that has eframe as a dependency if I want to use android-native-activity.

I don't have a solution to propose, I just wanted to make sure I'm not missing anything before I find a workaround.

lucasmerlin:
> And if any library I depend on has eframe in its transitive dependency chain, and if that library does not also disable accesskit, then accesskit will be on and I will hit this error.

True, but unusally 3rd party crates should only depend on egui and not eframe (but I'm sure there are some exceptions). Also it's good practise to disable default-features of dependencies in library crates (especially when depending on something like eframe, as not doing so would also force eframe to use the gl backend even if the user wants to use wgpu). So if you run into this, best way to fix this would be to open a PR to the crate you use to disable the default features.

PR #6860, opened 2025-04-26
Add back `WGPU_TRACE` support via the new wgpu-trace feature

torokati44:
This builds on https://github.com/emilk/egui/pull/6744, adding back the support or wgpu tracing in some way.


torokati44:
I tried to bump `ron` to `0.9` in a separate PR to not duplicate it, but unfortunately due to https://github.com/ron-rs/ron/issues/565, it would have to be bumped to `0.10.1` instead, and that would again introduce a duplication, pending a new wgpu release with https://github.com/gfx-rs/wgpu/pull/7557... :face_with_spiral_eyes:

Wumpf:
This is pretty much blocked on
* https://github.com/gfx-rs/wgpu/issues/5974

Trace feature was removed for the time being (.. let's see how long..) from wgpu since it was too hard to support in the short term with resource id tracking being removed from wgpu-core

PR #6876, opened 2025-04-26
Move magic margin numbers in Frame into Style, and make Separator's spacing inherit from item_spacing 

Masterchef365:
This is a simple PR, which addresses some comments in the Discord asking why there were hardcoded values for the margins in Frame. This PR also addresses a comment which pointed out that Separator could use the Style's item_spacing by default. Note that the default item_spacing value does not exactly match the previous default spacing value, so this change does modify the default appearance of the GUI slightly.

I'm marking this as a draft, because there may be other magic numbers to find.

### DIFF
--- a/crates/egui/src/containers/frame.rs
+++ b/crates/egui/src/containers/frame.rs
@@ -193,7 +193,9 @@ impl Frame {
     }
 
     pub fn central_panel(style: &Style) -> Self {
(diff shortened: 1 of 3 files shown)


emilk:
Closing: the separator half is covered by the new widget-style system (`SeparatorStyle`). The frame margins would now also belong there, so if you want to pick that up, please open a fresh PR on top of current main. Thanks for the PR, and sorry for the slow response

Issue #4268, opened 2024-03-29
Uneven font rendering with artifacts on Firefox on Android


MarcoBrieden:
The text rendering of my buttons is very uneven in Firefox on my Android phone, meaning e.g. the right vertical line of an "N" is 4x as big as the left one. There are also artifacts, like vertical lines next to some letters.

Here is a magnified screenshot:
![text](https://github.com/emilk/egui/assets/234204/72aaa816-029f-484c-8800-4aec7fbf6c52)

The same page renders correctly in Chrome on the same phone.

I tested this both in the latest release (0.27.1) and the latest master (master#a541e021) with the same results.

The Firefox version is 124.1.0 (Build #2016010543) on Android 14 (March 5, 2024 update) on a Pixel 8 Pro.

Please let me know if I can provide additional info.

wareya:
This seems to also be a problem for native Android builds, too, at certain font sizes. Whatever's responsible for generating verts and UVs for text glyphs probably has rounding errors in it somewhere. Should also double check whether it generates the necessary 1px of empty space between glyphs when packing them into whatever glyph atlas it uses, but not doing so wouldn't cause these problems.

![Image](https://github.com/user-attachments/assets/c2a6f182-e105-4c6d-b511-765443f8941a)

![Image](https://github.com/user-attachments/assets/22d6d696-39fa-4629-8399-d17172964222)

![Image](https://github.com/user-attachments/assets/aacd2153-6a7f-42a5-832b-884c47a09438)

lucasmerlin:
Maybe it'll be magically fixed by #5784

wareya:
I made a fork of the egui repo and changed the `precision mediump float;` instances in the egui_glow shaders to use `highp` instead and the problem went away! Then I undid that and the problem came back. The problem is indeed fp16 rounding.

PR #6893, opened 2025-04-28
Fix text distortion on mobile devices/browsers with `glow` backend

wareya:
Did not test on platforms other than my phone, but I can't imagine it causing problems. AFAIK if highp isn't supported then `precision highp float;` needs to still not cause the program to fail to link/compile or anything; it should just silently use some other precision.

* Fixes https://github.com/emilk/egui/issues/4268 for me but I only tested it on a native Android app and I don't know whether backends other than glow are affected.
* [x] I have followed the instructions in the PR template (but the change is trivial so I'm just doing it from the master branch)

Before:
image

After:
image

### DIFF
--- a/crates/egui_glow/src/shader/fragment.glsl
+++ b/crates/egui_glow/src/shader/fragment.glsl
@@ -1,5 +1,5 @@
 #ifdef GL_ES
-    precision mediump float;
+    precision highp float;
 #endif
 
 uniform sampler2D u_sampler;
--- a/crates/egui_glow/src/shader/fragment.glsl
+++ b/crates/egui_glow/src/shader/fragment.glsl
@@ -1,11 +1,11 @@
 #ifdef GL_ES
-// To avoid weird distortion issues when rendering text etc, we want highp if possible.
-// But apparently some devices don't support it, so we have to check first.
-#if defined(GL_FRAGMENT_PRECISION_HIGH) && GL_FRAGMENT_PRECISION_HIGH == 1
-    precision highp float;
-#else
-    precision mediump float;
-#endif
+    // To avoid weird distortion issues when rendering text etc, we want highp if possible.
+    // But apparently some devices don't support it, so we have to check first.
+    #if defined(GL_FRAGMENT_PRECISION_HIGH) && GL_FRAGMENT_PRECISION_HIGH == 1
+        precision highp float;
+    #else
+        precision mediump float;
+    #endif
 #endif
 
 uniform sampler2D u_sampler;
--- a/crates/egui_glow/src/shader/fragment.glsl
+++ b/crates/egui_glow/src/shader/fragment.glsl
@@ -1,4 +1,6 @@
 #ifdef GL_ES
+// To avoid weird distortion issues when rendering text etc, we want highp if possible.
+// But apparently some devices don't support it, so we have to check first.
 #if defined(GL_FRAGMENT_PRECISION_HIGH) && GL_FRAGMENT_PRECISION_HIGH == 1
     precision highp float;
 #else
--- a/crates/egui_glow/src/shader/fragment.glsl
+++ b/crates/egui_glow/src/shader/fragment.glsl
@@ -1,5 +1,9 @@
 #ifdef GL_ES
+#if defined(GL_FRAGMENT_PRECISION_HIGH) && GL_FRAGMENT_PRECISION_HIGH == 1
     precision highp float;
+#else
+    precision mediump float;
+#endif
 #endif
 
 uniform sampler2D u_sampler;
--- a/crates/egui_glow/src/shader/vertex.glsl
+++ b/crates/egui_glow/src/shader/vertex.glsl
@@ -9,13 +9,13 @@
 #endif
 
 #ifdef GL_ES
-// To avoid weird distortion issues when rendering text etc, we want highp if possible.
-// But apparently some devices don't support it, so we have to check first.
-#if defined(GL_FRAGMENT_PRECISION_HIGH) && GL_FRAGMENT_PRECISION_HIGH == 1
-    precision highp float;
-#else
-    precision mediump float;
-#endif
+    // To avoid weird distortion issues when rendering text etc, we want highp if possible.
+    // But apparently some devices don't support it, so we have to check first.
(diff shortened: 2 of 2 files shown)


wareya:
Realizing that for whatever reason this makes *everything* look better/more even:

floorp_2025-04-28_06-50-49
floorp_2025-04-28_06-50-55


floorp_2025-04-28_06-50-07floorp_2025-04-28_06-50-01

wareya:
Apparently on very old drivers (that I think are violating the GLES GLSL spec) `precision highp float;` in fragment shaders can cause compilation to fail, so it has to be wrapped in a preprocessor define check. Done so.

wareya:
The `Check spelling and links / lychee` CI failure is a spurious network failure.

wareya:
Why this is the right solution and why possible workarounds someone might think of to make mediump work better won't actually work:

`mediump` usually corresponds to the standard IEEE fp16 format, which can represent all integers only in the range from -2048 to +2048. The next biggest representable number after 2048 is 2050.

Because of the logarithmic scaling of fp numbers, this applies at all scales, i.e. if you try to linearly chop up the space between -1.0 and +1.0, the finest number of chops you can get is around 4 thousand. If you try to do 8 thousand then half of them will get rounded off to a chopping location you've already chopped at. From 0 to 1.0, you get half as many available chops: the next smallest number below 1.0 is ~0.9995, or more precisely it's 2047/2048. So scaling isn't a viable workaround.

If UVs ever have to be represented in a range from 0 to 1, which they do during texture sampling, then any font glyph texture atlas cannot be bigger than 2048x2048 without it becoming impossible to address specific pixels within it.

But it gets slightly worse than this. The coordinate of the first pixel of a texture is not at 0. It's at (0.5/texsize). This is because 0 corresponds to the halfway point between the first and last pixels in the texture, wrapping left. So the coordinate for the first pixel of a 2048x2048 texture, in fp16, will get *rounded down by half a pixel!*. So our ACTUAL pixel-addressable texture size limit with mediump is 1024x1024.

Yikes!!! 1024x1024 textures could work, but it's *awful*! If you're rendering text at 32x32 pixels per glyph, for example, you run out of space in the glyph atlas after only a thousand-ish characters, and need to start doing context switches to switch to another atlas! With 4096x4096 textures, you get 16 thousand-ish characters for 32x32 pixel glyphs instead. And mobile devices are WAY more likely to run into bigger and more glyph atlasses, because they're generally higher dpi than desktops. And let's say I want to render a single giant letter on a 1920x1080 screen, vertically. With a 1024x1024 atlas, doing that is *literally impossible*.

To clarify, all of this happens the moment the texture UV is converted into the 0.0~1.0 range to be fed into the `texture()` function. Nothing done beforehand can avoid it, not splitting the UV up into multiple variables or anything.

emilk (approved):
Nice catch!

But can't we just always use `highp`, instead of having an `#if defined`?

Also @Wumpf can we do the same on wgpu?

wareya:
> Nice catch!
> 
> But can't we just always use `highp`, instead of having an `#if defined`?

In theory, yes! But I poked around on Google and there were people running into problems on old GLES2 drivers, where they would throw a hard error if they did this without the `#if defined`. I looked at the GLES spec and it lists the highp version as valid syntax and the default for vertex shaders, but it doesn't specify a default for fragment shaders and it doesn't specify what should happen if they don't natively support highp. I think shader compilation errors are out of spec, but I'm not 100% sure. At the very least those drivers apparently exist and I think it's a good idea to be defensive against them.

wareya:
Short explanation added as comments.


// To avoid weird distortion issues when rendering text etc, we want highp if possible.
// But apparently some devices don't support it, so we have to check first.

crates/egui_glow/src/shader/fragment.glsl
```diff
@@ -1,6 +1,12 @@
 #ifdef GL_ES
+// To avoid weird distortion issues when rendering text etc, we want highp if possible.
+// But apparently some devices don't support it, so we have to check first.
+#if defined(GL_FRAGMENT_PRECISION_HIGH) && GL_FRAGMENT_PRECISION_HIGH == 1
+    precision highp float;
+#else
     precision mediump float;
 #endif
+#endif
```
emilk, line 9:
```suggestion
#ifdef GL_ES
    // To avoid weird distortion issues when rendering text etc, we want highp if possible.
    // But apparently some devices don't support it, so we have to check first.
    #if defined(GL_FRAGMENT_PRECISION_HIGH) && GL_FRAGMENT_PRECISION_HIGH == 1
        precision highp float;
    #else
        precision mediump float;
    #endif
#endif
```

crates/egui_glow/src/shader/vertex.glsl
```diff
@@ -9,8 +9,14 @@
...
 
 #ifdef GL_ES
+// To avoid weird distortion issues when rendering text etc, we want highp if possible.
+// But apparently some devices don't support it, so we have to check first.
+#if defined(GL_FRAGMENT_PRECISION_HIGH) && GL_FRAGMENT_PRECISION_HIGH == 1
+    precision highp float;
+#else
     precision mediump float;
 #endif
+#endif
```
emilk, line 19:
```suggestion
#ifdef GL_ES
    // To avoid weird distortion issues when rendering text etc, we want highp if possible.
    // But apparently some devices don't support it, so we have to check first.
    #if defined(GL_FRAGMENT_PRECISION_HIGH) && GL_FRAGMENT_PRECISION_HIGH == 1
        precision highp float;
    #else
        precision mediump float;
    #endif
#endif
```

emilk (approved):
Thanks!

crates/egui_glow/src/shader/fragment.glsl
emilk: NOTE: I cannot commit these changes directly, because you opened your PR from your `master`.

emilk: This is mentioned in the PR template, but since most people seem to only skim that (which I understand), I created a PR to enforce it in the future:
* https://github.com/emilk/egui/pull/6899

wareya: Done.

PR #6897, opened 2025-04-29
Add italic button benchmark to test `RichText` performance impact

lucasmerlin:
Time on my m4 pro macbook:  302.79 ns (vs 303.83 ns for the regular button 🤷)

### DIFF
--- a/crates/egui_demo_lib/benches/benchmark.rs
+++ b/crates/egui_demo_lib/benches/benchmark.rs
@@ -4,7 +4,7 @@ use criterion::{criterion_group, criterion_main, BatchSize, Criterion};
 
 use egui::epaint::TextShape;
 use egui::load::SizedTexture;
-use egui::{Button, Id, TextureId, Ui, UiBuilder, Vec2};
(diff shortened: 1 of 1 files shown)


crates/egui_demo_lib/benches/benchmark.rs
```diff
@@ -121,6 +121,15 @@ pub fn criterion_benchmark(c: &mut Criterion) {
                     BatchSize::LargeInput,
                 );
             });
+            group.bench_function("4_button_italic", |b| {
```
emilk, line 124: Up to here I thought the number prefix was the number of atoms in the button 😅

emilk (approved):

crates/egui_demo_lib/benches/benchmark.rs
lucasmerlin: No I added this so you can run only one of the benchmarks. Otherwise one would need to use regex to match the exact benchmark name which is kinda annoying, like this I can just use "button/1" (thanks to the group)

PR #6899, opened 2025-04-30
Enforce that PRs are not opened from the 'master' branch of a fork

emilk:
## What
Fail all PRs that are opened from the master/main branch of the fork.

## Why
PR:s opened from the `master` branch cannot be collaborated on. That is, we maintainers cannot push our own commits to it (e.g. to fix smaller problems with it before merging).

## How
Untested code straight from Claude 3.7 😅

### DIFF
--- a/.github/workflows/enforce_branch_name.yml
+++ b/.github/workflows/enforce_branch_name.yml
@@ -1,0 +1,34 @@
+name: PR Branch Name Check
+
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  check-source-branch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check PR source branch
+        run: |
+          # Check if PR is from a fork
+          if [[ "${{ github.event.pull_request.head.repo.fork }}" == "true" ]]; then
+            # Check if PR is from the master/main branch of a fork
+            if [[ "${{ github.event.pull_request.head.ref }}" == "master" || "${{ github.event.pull_request.head.ref }}" == "main" ]]; then
+              echo "ERROR: Pull requests from the master/main branch of forks are not allowed, because it prevents maintainers from contributing to your PR"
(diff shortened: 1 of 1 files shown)


.github/workflows/enforce_branch_name.yml
```diff
@@ -0,0 +1,34 @@
...
+        run: |
+          # Check if PR is from a fork
+          if [[ "${{ github.event.pull_request.head.repo.fork }}" == "true" ]]; then
+            # Check if PR is from the master/main branch of a fork
+            if [[ "${{ github.event.pull_request.head.ref }}" == "master" || "${{ github.event.pull_request.head.ref }}" == "main" ]]; then
+              echo "ERROR: Pull requests from the master/main branch of forks are not allowed, because it prevents maintainers from contributing to your PR"
+              echo "Please create a feature branch in your fork and submit the PR from that branch instead."
+              exit 1
+            fi
+          fi
```
lucasmerlin, line 21: If this step fails the second step shouldn't be exectued and no comment would be left 🤔

lucasmerlin: I think you need something like `&& always()` or just `if: failed()`

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#status-check-functions

emilk: Or just just change the order of the steps?

Mingun:
Just want to comment, that it should be possible to push to master branch of other repository, see for example https://github.com/tafia/quick-xml/pull/857, where I made a force-push to master branch of a contributor.

emilk:
I believe it is possible _if_ the user has not protected their master branch (which they probably _should_). And force-pushing to someone else `master` doesn't seem like a good idea (having done so myself accidentally a few times).

lucasmerlin (approved):
Nice! I guess we have to merge this to see if this works

Issue #6904, opened 2025-04-30
Links are broken if they aren't the last thing in a line


lucasmerlin:
**Describe the bug**

https://github.com/user-attachments/assets/d57b9deb-0131-4a52-ada5-a1ebfd54be11

Maybe caused by https://github.com/emilk/egui/pull/5411 or https://github.com/emilk/egui/pull/5785?

**To Reproduce**
Paste the following on the easymark editor:
```
Works:
[hyperlink](https://github.com/emilk/egui)
Embedded URL: <https://github.com/emilk/egui>

Broken:
[hyperlink](https://github.com/emilk/egui) 
Embedded URL: <https://github.com/emilk/egui> 
[hyperlink](https://github.com/emilk/egui)x
Embedded URL: <https://github.com/emilk/egui>x
```

**Expected behavior**
Links should be clickable

**Additional context**
- https://github.com/lampsitter/egui_commonmark/pull/76#issuecomment-2833606191
- https://github.com/rerun-io/rerun/pull/9103#issuecomment-2841875402

PR #6905, opened 2025-04-30
Fix links and text selection in horizontal_wrapped layout

lucasmerlin:
* Closes <https://github.com/emilk/egui/issues/6904>

This was broken in https://github.com/emilk/egui/pull/5411. Not sure if this is the best fix or if `PlacedRow::rect` should be updated, but I think it makes sense that PlacedRow::rect ignores leading space.

### DIFF
--- a/crates/egui/src/widgets/label.rs
+++ b/crates/egui/src/widgets/label.rs
@@ -1,11 +1,9 @@
 use std::sync::Arc;
 
 use crate::{
-    epaint, pos2, text_selection, Align, Direction, FontSelection, Galley, Pos2, Response, Sense,
-    Stroke, TextWrapMode, Ui, Widget, WidgetInfo, WidgetText, WidgetType,
+    epaint, pos2, text_selection::LabelSelectionState, Align, Direction, FontSelection, Galley,
+    Pos2, Response, Sense, Stroke, TextWrapMode, Ui, Widget, WidgetInfo, WidgetText, WidgetType,
 };
-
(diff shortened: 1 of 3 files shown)


crates/egui/src/widgets/label.rs
```diff
@@ -216,7 +216,9 @@ impl Label {
             let pos = pos2(ui.max_rect().left(), ui.cursor().top());
             assert!(!galley.rows.is_empty(), "Galleys are never empty");
             // collect a response from many rows:
-            let rect = galley.rows[0].rect().translate(pos.to_vec2());
+            let mut rect = galley.rows[0].rect().translate(pos.to_vec2());
+            // The rect ignores first_row_indentation, so we add it back:
+            rect.min.x += first_row_indentation;
```
emilk, line 221: That `PlacedRow::rect` includes the first-row indentation feels like a bug to me. It will probably cause a bug for accessibility too, here:

https://github.com/emilk/egui/blob/6c922f72a819e6083ffc4b6a452c2493c9170e63/crates/egui/src/text_selection/accesskit_text.rs#L43-L48

emilk: In any case, let's document what the expected result of `PlacedRow::rect` should be

lucasmerlin: I've tried to make it return the right thing but it's not pretty 😅

emilk (approved):

Issue #6755, opened 2025-04-16
Assertion in egui_extras::FileLoader::load triggers when Context::forget_image is called while image is loading


luveti:
The following:
https://github.com/emilk/egui/blob/0f1d6c281894647b2fdd1fcb311eb930ab1a269c/crates/egui_extras/src/loaders/file_loader.rs#L99

Doesn't take into account that the user may want to forget the image before it's loaded. Here:

https://github.com/emilk/egui/blob/0f1d6c281894647b2fdd1fcb311eb930ab1a269c/crates/egui_extras/src/loaders/file_loader.rs#L111

or here:

https://github.com/emilk/egui/blob/0f1d6c281894647b2fdd1fcb311eb930ab1a269c/crates/egui_extras/src/loaders/file_loader.rs#L115

The following would be a better alternative (untested):
```rust
let mut cache = cache.lock();
if cache.contains(&uri) {
  cache.insert(uri.clone(), Poll::Ready(result));
  ctx.request_repaint();
  log::trace!("finished loading {uri:?}");
}
else {
  log::trace!("cancelled loading {uri:?}");
}
```

This assert was recently added by @bircni and @lucasmerlin in https://github.com/emilk/egui/commit/58b2ac88c0b39a48ac2277161ca75df3034b7ae3. It's possible similar assertions were added in other loaders.

lucasmerlin:
Good catch! Would you mind opening a PR? 🙏 Your fix looks correct

PR #6995, opened 2025-05-03
fix: Assertion in egui_extras::FileLoader::load

bircni:
* Closes <https://github.com/emilk/egui/issues/6755>

### DIFF
--- a/crates/egui_extras/src/loaders/file_loader.rs
+++ b/crates/egui_extras/src/loaders/file_loader.rs
@@ -95,10 +95,14 @@ impl BytesLoader for FileLoader {
                             }
                             Err(err) => Err(err.to_string()),
                         };
-                        let prev = cache.lock().insert(uri.clone(), Poll::Ready(result));
-                        assert!(matches!(prev, Some(Poll::Pending)), "unexpected state");
-                        ctx.request_repaint();
-                        log::trace!("finished loading {uri:?}");
+                        let mut cache = cache.lock();
+                        if cache.contains_key(&uri) {
+                            cache.insert(uri.clone(), Poll::Ready(result));
+                            ctx.request_repaint();
+                            log::trace!("finished loading {uri:?}");
+                        } else {
+                            log::trace!("cancelled loading {uri:?}");
+                        }
                     }
                 })
                 .expect("failed to spawn thread");


crates/egui_extras/src/loaders/file_loader.rs
```diff
@@ -95,10 +95,14 @@ impl BytesLoader for FileLoader {
...
-                        assert!(matches!(prev, Some(Poll::Pending)), "unexpected state");
-                        ctx.request_repaint();
-                        log::trace!("finished loading {uri:?}");
+                        let mut cache = cache.lock();
+                        if cache.contains_key(&uri) {
+                            cache.insert(uri.clone(), Poll::Ready(result));
+                            ctx.request_repaint();
+                            log::trace!("finished loading {uri:?}");
+                        } else {
+                            log::trace!("cancelled loading {uri:?}");
```
emilk, line 104: We should explain _why_ we ignore the newly loaded image and cancel. I.e., that it can only happen because `forget_image` was called.

We should also document that `forget_image` cancels in-progress loads, if any.

bircni:
@emilk done
Sorry completely forgot about this

crates/egui_extras/src/loaders/file_loader.rs
```diff
@@ -95,10 +95,14 @@ impl BytesLoader for FileLoader {
                             }
                             Err(err) => Err(err.to_string()),
                         };
-                        let prev = cache.lock().insert(uri.clone(), Poll::Ready(result));
-                        assert!(matches!(prev, Some(Poll::Pending)), "unexpected state");
-                        ctx.request_repaint();
-                        log::trace!("finished loading {uri:?}");
+                        let mut cache = cache.lock();
+                        if cache.contains_key(&uri) {
+                            cache.insert(uri.clone(), Poll::Ready(result));
```
emilk: You can avoid looking up `uri` twice by using

```rs
if let Entry::Occupied(entry) = cache.entry(uri.clone()) {
    entry.insert(Poll::Ready(result));
```

bircni: Fixed!

bircni:
@emilk there seems to be no Occupied 
Poll only has Ready and Pending

bircni:
okay late night coding is bad haha
oops

emilk (approved):

emilk:
Thanks!

PR #7029, opened 2025-05-06
Use mimalloc for benchmarks

emilk:
`mimalloc` is a _much_ faster allocator, especially important when doing a lot of small allocations (which egui does).

We use `mimalloc` in Rerun, and I recommend everyone to use it.

## The difference it makes
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1336,6 +1336,7 @@ dependencies = [
  "env_logger",
  "image",
  "log",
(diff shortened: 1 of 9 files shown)


lucasmerlin (approved):
Nice! 

> We use mimalloc in Rerun, and I recommend everyone to use it.

I think we should mention this somewhere in the readme then.

PR #7030, opened 2025-05-07
Add `ImageLoader::has_pending` and `wait_for_pending_images`

lucasmerlin:
With kittest it was difficult to wait for images to be loaded before taking a snapshot test. 
This PR adds `Harness::with_wait_for_pending_images` (true by default) which will cause `Harness::run` to sleep until all images are loaded (or `HarnessBuilder::with_max_steps` is exceeded).

It also adds a new ImageLoader::has_pending and BytesLoader::has_pending, which should be implemented if things are loaded / decoded asynchronously.

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -3546,6 +3546,14 @@ impl Context {
     pub fn loaders(&self) -> Arc<Loaders> {
         self.read(|this| this.loaders.clone())
     }
+
+    /// Returns `true` if any image is currently being loaded.
+    pub fn has_pending_images(&self) -> bool {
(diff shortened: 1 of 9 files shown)


crates/egui_kittest/tests/snapshots/should_wait_for_images.png
emilk, line 1: A bit big maybe? :)

emilk (approved):
nice

SunilKukreja1004:
How do we call the Egui Update function for exiting native egui app in egui_Kittest Struct as we need to pass e frame object ?

PR #7031, opened 2025-05-07
Fix `leading_space` sometimes being ignored during paragraph splitting

afishhh:
Fixes a regression introduced in https://github.com/emilk/egui/pull/5411 (possibly https://github.com/emilk/egui/pull/5411/commits/d74bee536f7c57d83014b759b18a0ddef8725552) that breaks `leading_space` handling.
I think this is what the condition should be but I haven't touched this code in a while.


emilk (approved):
Good catch!

We should add a test case for this at the end of the file too (just set `leading_space` in one of the test cases)

emilk (approved):

Issue #7026, opened 2025-05-05
Update accesskit


jhall-conp:
Paste was a dependency in an earlier version of accesskit. This dependency was removed for both Windows and Android.

https://rustsec.org/advisories/RUSTSEC-2024-0436 - paste - no longer maintained

Apologies for entering this as a bug instead of another type. The crate is still working great, it's just that the Git dependency checker is unhappy with it.

PR #7040, opened 2025-05-09
Bump accesskit to 0.19

valadaptive:
Parley's bumped accesskit to 0.19, so we have to as well. It's a bit concerning that we may end up locked to the version of accesskit that Parley uses, but oh well.

[This kittest PR will have to be merged first.](https://github.com/rerun-io/kittest/pull/11)


DataTriny:
FWIW this fixes #7026

emilk:
Thanks!

Issue #901, opened 2021-11-17
CentralPanel overflows in a window unless there is a Top, Bottom or Right pannel


rMazeiks:
If a CentralPanel is added to a window (with or without a LeftPanel), it seems to be ignored by the window layout and overflows. Adding a Top, Bottom or Right pannel makes the issue go away.

**Screenshots**
![image](https://user-images.githubusercontent.com/11248241/142274371-59066eec-8377-45da-af01-9b271fd08347.png)

**Expected behavior:** Window correctly wraps around central pannel and includes it in the layout. Window has no overflow.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version: tested on master
 - Using eframe to run natively.

Code:

```rust
        egui::Window::new("test").show(egui_context, |ui| {
            // uncomment to make the issue go away
            
            // egui::TopBottomPanel::top("top_panel").show_inside(ui, |ui| {
            //     egui::ScrollArea::vertical().show(ui, |ui| {
            //         lorem_ipsum(ui);
            //     });
            // });

            egui::SidePanel::left("left_panel").show_inside(ui, |ui| {
                egui::ScrollArea::vertical().show(ui, |ui| {
                    lorem_ipsum(ui);
                });
            });

            CentralPanel::default().show_inside(ui, |ui| {
                ui.label("This guy is thinking outside the box for some reason")
            })
        });
```

edorgeville:
I had to set the height and frame in order to get a 0px bottom panel:
```rust
egui::TopBottomPanel::bottom(egui::Id::new("actions list invisible bottom panel"))
    .show_separator_line(false)
    .exact_height(0.0)
    .frame(Frame::none())
    .show_inside(ui, |_| ());
```

PR #7042, opened 2025-05-09
fix: avoid CentralPanel overflowing in a window

avalsch:
```rust
use eframe::egui;

    eframe::run_simple_native("window overflow", Default::default(), |ctx, _| {
        egui::Window::new("window overflow").show(ctx, |ui| {
            ui.label("Hello, World");
        });
    })
}
```

Old behavior:

image

New behavior:

image

I'd ideally add a test for this but I'm not sure how to go about doing that.

* Closes <https://github.com/emilk/egui/issues/901>

### DIFF
--- a/crates/egui/src/containers/panel.rs
+++ b/crates/egui/src/containers/panel.rs
@@ -1121,10 +1121,14 @@ impl CentralPanel {
         panel_ui.set_clip_rect(panel_rect); // If we overflow, don't do so visibly (#4475)
 
         let frame = frame.unwrap_or_else(|| Frame::central_panel(ui.style()));
-        frame.show(&mut panel_ui, |ui| {
+        let inner_response = frame.show(&mut panel_ui, |ui| {
             ui.expand_to_include_rect(ui.max_rect()); // Expand frame to include it all
             add_contents(ui)
-        })
+        });
+
+        ui.expand_to_include_rect(inner_response.response.rect);
+
+        inner_response
     }
 
     /// Show the panel at the top level.


crates/egui/src/containers/panel.rs
```diff
@@ -1121,10 +1121,14 @@ impl CentralPanel {
...
-        frame.show(&mut panel_ui, |ui| {
+        let inner_response = frame.show(&mut panel_ui, |ui| {
             ui.expand_to_include_rect(ui.max_rect()); // Expand frame to include it all
             add_contents(ui)
-        })
+        });
+
+        ui.expand_to_include_rect(inner_response.response.rect);
+
+        inner_response
```
lucasmerlin, line 1131: I think instead of `ui.new_child` we could also do ui.scope_builder which should do the `expand_to_include_rect` for us. Might be a bit cleaner.

lucasmerlin (commented):
Thanks for the fix! To add tests, the easiest would be to add a new kittest test to tests/egui_tests, the test could just simply be a screenshot test with the repro you provided.

crates/egui/src/containers/panel.rs
```diff
@@ -1111,24 +1111,20 @@ impl CentralPanel {
...
 
-        let panel_rect = ui.available_rect_before_wrap();
-        let mut panel_ui = ui.new_child(
+        ui.scope_builder(
             UiBuilder::new()
                 .ui_stack_info(UiStackInfo::new(UiKind::CentralPanel))
-                .max_rect(panel_rect)
                 .layout(Layout::top_down(Align::Min)),
-        );
-        panel_ui.set_clip_rect(panel_rect); // If we overflow, don't do so visibly (#4475)
```
avalsch, line 1121: It seems to me like scope_builder doesn't preserve the same behavior with set_clip_rect. This might break some things, but I'm not sure?

avalsch:
Regarding using scope_builder rather than new_child, the other panel implementations (SidePanel, TopBottomPanel) also use new_child. Is this something we would want to change?

PR #7050, opened 2025-05-15
Fix `visual_bounding_rect` for rotated text

pmarks:
TextShape.visual_bounding_rect was not taking the text rotation into account. I manually tested drawing the new bounding box on top of the text for various rotations & anchor settings. For example:
image

The unit test I added will fail without this patch, but perhaps doesn't add much value.


lucasmerlin (approved):
Nice, thanks for adding a test!

PR #7061, opened 2025-05-19
egui_extras: Enable setting DatePickerButton start and end year explicitly

zachbateman:
Add the ability to set the `DatePickerButton`'s start and end years via new `start_year` and `end_year` methods.

Continue to use the existing today - 100 years and today + 10 years behavior if a year is not specified.

* This more fully closes <https://github.com/emilk/egui/issues/3597> and expands on <https://github.com/emilk/egui/pull/3599>.

### DIFF
--- a/crates/egui_extras/src/datepicker/button.rs
+++ b/crates/egui_extras/src/datepicker/button.rs
@@ -19,6 +19,8 @@ pub struct DatePickerButton<'a> {
     show_icon: bool,
     format: String,
     highlight_weekends: bool,
+    start_year: Option<i32>,
+    end_year: Option<i32>,
 }
 
 impl<'a> DatePickerButton<'a> {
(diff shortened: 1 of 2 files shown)


crates/egui_extras/src/datepicker/button.rs
```diff
@@ -101,6 +105,20 @@ impl<'a> DatePickerButton<'a> {
...
+        self.start_year = Some(start_year);
+        self
+    }
+
+    /// Set the end year for the date picker. (Default: None)
+    #[inline]
+    pub fn end_year(mut self, end_year: i32) -> Self {
+        self.end_year = Some(end_year);
+        self
+    }
```
emilk: I suggest you simplify this with a single call with a `RangeInclusive`. I suspect that if you set one, you want to set both.

You should also explain what it is for (the dropdown of years you can choose from)

emilk (changes requested):

zachbateman:
Thanks for the feedback, @emilk.  That makes sense, and I've updated accordingly.

Issue #7065, opened 2025-05-21
arrow navigation (sometimes) doesn't work on a slider when using ``.fixed_decimals(n)``


0x53A:
Example: https://0x53a.github.io/UAS-SigVer/
(Source: https://github.com/0x53A/UAS-SigVer/)

tab onto the first slider, press arrows left and/or right

### Expected:

slider moves

### Actual:

nothing happens

--------

This works on sliders 2 and 3.

--------

``set_value`` rounds:

```rust
    fn set_value(&mut self, mut value: f64) {
        if self.clamping != SliderClamping::Never {
            value = clamp_value_to_range(value, self.range.clone());
        }

        if let Some(step) = self.step {
            let start = *self.range.start();
            value = start + ((value - start) / step).round() * step;
        }
        if let Some(max_decimals) = self.max_decimals {
            value = emath::round_to_decimals(value, max_decimals);
        }
        set(&mut self.get_set_value, value);
    }
```

----------

``slider_ui`` has: 

```rust
[...]
let ui_point_per_step = 1.0;

let new_position = prev_position + ui_point_per_step * kb_step;
let new_value = self.value_from_position(new_position, position_range);
self.set_value(new_value)
```

The issue starts when the minimal increment of the keyboard handler rounds down to the previous value, and nothing happens.

----

fix:

``slider_ui`` should also take the field ``self.max_decimals`` into account, and move at least so much that the next breakpoint is reached.

PR #7066, opened 2025-05-21
Slider: move by at least the next increment when using fixed_decimals

0x53A:
fixes https://github.com/emilk/egui/issues/7065


0x53A:
i've manually checked that the test fails on main and succeeds on the branch

emilk (approved):
Thanks for adding that test!

PR #7073, opened 2025-05-21
Support for Back button Key on Android

ardocrat:
When your press a Back button on Android  (for example at `native-activity`), [Winit translates this key](https://github.com/rust-windowing/winit/blob/47b938dbe78702d521c2c7a43b6f741a3bb8cb0b/src/platform_impl/android/keycodes.rs#L237C42-L237C53) as `NamedKey::BrowserBack`. Added convertion to `Key::Escape` at `egui-winit` module.

### DIFF
--- a/crates/egui/src/data/key.rs
+++ b/crates/egui/src/data/key.rs
@@ -184,6 +184,9 @@ pub enum Key {
     F34,
     F35,
 
+    /// Back navigation key from multimedia keyboard.
+    /// Android sends this key on Back button press.
+    /// Does not work on Web.
     BrowserBack,
     // When adding keys, remember to also update:
     // * crates/egui-winit/src/lib.rs


ardocrat:
I think `Key::Escape` is more reliable in case with `NamedKey::BrowserBack` or we can create new key like `Key::Back`, what do you think @emilk @lucasmerlin ?

lucasmerlin:
Hmm, I think a new back key would be nicer. There are some cases where I'd expect pressing back and pressing escape to do different things. E.g. pressing escape should clear focus while pressing back should take you back to the previous page even if something is focused.

ardocrat:
Added `Key::BrowserBack` as last at Keys, cause there is no groups for multimedia/navigation keys to support full keyboard layout from `winit`.

crates/egui/src/data/key.rs
```diff
@@ -183,6 +183,8 @@ pub enum Key {
     F33,
     F34,
     F35,
+
+    BrowserBack,
```
emilk, line 190: This should have a docstring, and also explain that this does not work on Web (which makes the name of it a bit confusing…)

ardocrat: Seems like name actually came from browser control buttons.
image

PR #7077, opened 2025-05-23
Ensure TextEdit cursor is never out of bounds

juancampa:
When rendering a TextArea we don't know if the saved cursor applies to the current galley since it's possible the app changed the TextBuffer (e.g. when submitting a chat input)

So we now detect if the galley changed from the last known one and clamp the cursor to ensure it's not out of bounds.

This fixes an issue where backspace and arrow keys can suddenly stop working. In this video I changed the `TextArea` to not lose focus when enter is pressed (using `.return_key`). And instead, the app clears the `String`. Which is what we do in our chat and REPL UIs.

Screenshot 2025-05-23 at 01 36 14

Repro:
 - Render TextArea with long-ish text (say 20 chars)
 - Without losing focus, clear the text
 - Write something short (say 5 chars)
 - Result: backspace doesn't work because the cursor position is wrong
 
I think this issue started happening when [this function](https://github.com/emilk/egui/pull/5785/files#diff-09cc53b30f3c2a6481d438328f86236eb723a6ba5152101b58592a9013b1eb69L60) was removed in #5785

(cc @valadaptive in case you happen to know of a better fix that doesn't require keeping the Galley in the TextEdit state)

### DIFF
--- a/crates/egui/src/text_selection/text_cursor_state.rs
+++ b/crates/egui/src/text_selection/text_cursor_state.rs
@@ -39,6 +39,14 @@ impl TextCursorState {
     pub fn set_char_range(&mut self, ccursor_range: Option<CCursorRange>) {
         self.ccursor_range = ccursor_range;
     }
+
+    /// Clamp the cursors to be in bounds of the galley.
+    pub fn ensure_in_bounds(&mut self, galley: &Galley) {
+        if let Some(range) = self.ccursor_range.as_mut() {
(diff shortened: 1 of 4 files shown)


valadaptive:
The removed `range` function took the galley as an argument, whereas its replacement (`char_range`) does not. Is it possible to make `char_range` take a `Galley` and do the clamping there, and just pass in the galley in the same paths that the old code did?

It's been a while since I made those changes so I could be totally off-base.

PR #7081, opened 2025-05-24
Clamp text cursor positions in the same places where we used to

valadaptive:
Closes #7077.

This fixes the problem shown in #7077 where clearing a `TextEdit` wouldn't reset its cursor position. I've fixed that by adding back the `TextCursorState::range` method, which clamps the selection range to that of the passed `Galley`, and calling it in the same places where it was called before #5785.

(/cc @juancampa)

### DIFF
--- a/crates/egui/src/text_selection/label_text_selection.rs
+++ b/crates/egui/src/text_selection/label_text_selection.rs
@@ -530,7 +530,7 @@ impl LabelSelectionState {
 
         let mut cursor_state = self.cursor_for(ui, response, global_from_galley, galley);
 
-        let old_range = cursor_state.char_range();
+        let old_range = cursor_state.range(galley);
 
         if let Some(pointer_pos) = ui.ctx().pointer_interact_pos() {
             if response.contains_pointer() {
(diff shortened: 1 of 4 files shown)


juancampa:
Did you find other issues that require clamping? The reason I added the `Weak<Galley>` to `TextArea` state is to prevent `clamp_cursor` from running every frame, since it's O(N) to the number of rows. But maybe that's okay?

valadaptive:
It's what the previous code did, so it's probably okay. It's possible we don't need to call `cursor_from_layout` and `layout_from_cursor` and can just clamp the cursor's byte range directly; I can test that out later.

juancampa:
Yeah that makes sense. Would be nice to do a quick sanity check to see how much it’s getting called. 

By the way, my reasoning for clamping using cursor -> layout -> cursor, instead of the byte range was to maintain the cursor position as close as possible to the old one. For example, if the string is modified by unindenting it, I’d like the cursor to stay in the same line, instead of being moved downwards to the corresponding byte offset. I don’t think I tested that particular case though.

emilk (approved):
Thanks!

Issue #3964, opened 2024-02-04
Add Syntect custom syntax & themes support


antaalt:
Hello, 
I am trying to work with the syntect code editor which is great, the issue I have now is that it seems impossible to load custom syntax & themes to add support for others languages than the one [available by default](https://github.com/sublimehq/Packages/) in sublime text, maybe I just missed the API, but it would be nice to be able to load a custom tmTheme and sublime-syntax as Syntect support this via `SyntaxSetBuilder`.

From what I saw, egui run `syntect::parsing::SyntaxSet::load_defaults_newlines()` which is created with the cache, but we could somehow allow egui to load custom themes & syntax aswell

TimTheBig:
**Any progress?**
I would be open to **writing a patch** that allows for loading of a  syntax highlighting theme from a `.tmTheme`(sublime theme) file.
I just wanted to know if there is any reason that this is not implemented?

antaalt:
Didn't had the time to look into it. It does not look like a highly requested feature, which would explain why it's not implemented

mkeeter:
I would also be interested in this functionality, since I want to embed [Rhai](https://rhai.rs/) scripts into an `egui`-based application.  I'll investigate and possibly open a PR depending on difficulty.

mkeeter:
I've implemented this in #7084 , and backported to 0.31.1 [here](https://github.com/mkeeter/egui/tree/custom-syntect-settings-0.31.1-backport).

You can test it out with something like this:

```toml
[dependencies]
egui_extras = { git = "https://github.com/mkeeter/egui", features = ["syntect"], branch = "custom-syntect-settings-0.31.1-backport" }

[patch.crates-io]
egui = { git = "https://github.com/mkeeter/egui", branch = "custom-syntect-settings-0.31.1-backport" }
epaint = { git = "https://github.com/mkeeter/egui", branch = "custom-syntect-settings-0.31.1-backport" }
```

PR #7084, opened 2025-05-25
Support custom syntect settings in syntax highlighter

mkeeter:
* Closes <https://github.com/emilk/egui/issues/3964>

This PR adds support for syntax highlighting with custom `syntect::parsing::SyntaxSet` and `syntect::highlighting::ThemeSet`.  It adds a new `egui_extras::highlight_with` function (enabled with `feature = "syntect"`), which takes a new public`struct SyntectSettings` containing the syntax and theme sets.

```rust
let mut builder = SyntaxSetBuilder::new();
builder.add_from_folder("syntax", true).unwrap();
let ps = builder.build();
let ts = syntect::highlighting::ThemeSet::load_defaults();
let syntax =
    egui_extras::syntax_highlighting::SyntectSettings { ps, ts };

// ...elsewhere
egui_extras::syntax_highlighting::highlight_with(
    ui.ctx(),
    ui.style(),
    &theme,
    buf,
    "rhai",
    &syntax,
);
```

There's a little bit of architectural complexity, but it all emerges naturally from the problem's constraints.

Previously, the `Highlighter` both contained the `syntect` settings _and_ implemented `egui::cache::ComputerMut` to highlight a string; the settings would never change.  After this change, the `syntect` settings have become part of the cache key, so we should redo highlighting if they change.  The `Highlighter` becomes an empty `struct` which just serves to implement `ComputerMut`.

`SyntaxSet` and `ThemeSet` are not hasheable themselves, so can't be used as cache keys direction.  Instead, we can use the *address* of the `&SyntectSettings` as the key.  This requires an object with a custom `Hash` implementation, so I added a new `HighlightSettings(&'a SyntectSettings)`, implementing `Hash` using `std::ptr::hash` on the reference.  I think using the address is reasonable – it would be _weird_ for a user to be constantly moving around their `SyntectSettings`, and there's a warning in the docstring to this effect.

To work _without_ custom settings, `SyntectSettings::default` preserves the same behavior as before, using `SyntaxSet::load_defaults_newlines` and `ThemeSet::load_defaults`.  If the user doesn't provide custom settings, then we instantiate a singleton `SyntectSettings` in `data` and use it; this will only be constructed once.

Finally, in cases where the `syntect` feature is disabled, `SyntectSettings` are replaced with a unit `()`.  This adds a _tiny_ amount of overhead – one singleton `Arc<()>` allocation and a lookup in `data` per `highlight` – but I think that's better than dramatically different implementations.  If this is an issue, I can refactor to make it zero-cost when the feature is disabled.

### DIFF
--- a/crates/egui_extras/src/syntax_highlighting.rs
+++ b/crates/egui_extras/src/syntax_highlighting.rs
@@ -29,17 +29,64 @@ pub fn highlight(
     code: &str,
     language: &str,
 ) -> LayoutJob {
+    highlight_inner(ctx, style, theme, code, language, None)
+}
+
+/// Add syntax highlighting to a code string, with custom `syntect` settings
+///
+/// The results are memoized, so you can call this every frame without performance penalty.
+///
+/// The `syntect` settings are memoized by *address*, so a stable reference should
+/// be used to avoid unnecessary recomputation.
+#[cfg(feature = "syntect")]
+pub fn highlight_with(
+    ctx: &egui::Context,
+    style: &egui::Style,
+    theme: &CodeTheme,
+    code: &str,
+    language: &str,
+    settings: &SyntectSettings,
+) -> LayoutJob {
+    highlight_inner(
+        ctx,
+        style,
+        theme,
+        code,
+        language,
+        Some(HighlightSettings(settings)),
+    )
+}
+
+fn highlight_inner(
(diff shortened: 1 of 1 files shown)


emilk (approved):
Nice!

mkeeter:
Thanks for the review – I've rebased this and it should be good to go!

Issue #7085, opened 2025-05-25
`Debug` impl for `Vec2` discards important information


SludgePhD:
**Describe the bug**

Formatting a `Vec2` with `{:?}` leads to its `Debug` impl to format both coordinates with `{:.1}`, which discards any decimal digits after the first.

https://github.com/emilk/egui/blob/a085d0b3a5a4aa70d7ca4ddedbcc9ed53c758b7b/crates/emath/src/vec2.rs#L485-L493

**To Reproduce**
1. `format!("{:?}", vec2(0.01, -0.01))`
2. `"[0.0 -0.0]"`

(untested)

**Expected behavior**

Formatting a value should not discard important information. `Debug` *especially* shouldn't be doing that.

Note that the `Display` impl does not have this problem.

bircni:
should be fixed in: https://github.com/emilk/egui/pull/7089

PR #7089, opened 2025-05-25
fix: vec2 debug impl

bircni:
Now we use the standard debug impl from f32
* Closes <https://github.com/emilk/egui/issues/7085>

### DIFF
--- a/crates/emath/src/vec2.rs
+++ b/crates/emath/src/vec2.rs
@@ -488,7 +488,7 @@ impl fmt::Debug for Vec2 {
         if let Some(precision) = f.precision() {
             write!(f, "[{1:.0$} {2:.0$}]", precision, self.x, self.y)
         } else {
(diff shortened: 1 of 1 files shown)


bircni:
@emilk can we merge?

crates/emath/src/vec2.rs
```diff
@@ -488,7 +488,7 @@ impl fmt::Debug for Vec2 {
         if let Some(precision) = f.precision() {
             write!(f, "[{1:.0$} {2:.0$}]", precision, self.x, self.y)
         } else {
-            write!(f, "[{:.1} {:.1}]", self.x, self.y)
```
emilk, line 491: The point of this was to keep the default `Debug` print of `Vec2` small, because usually we don't need higher precision than 0.1 ui points. I don't want a bunch of decimals in the default printing of Vec2, Pos2, Rect, etc.

You can still get all the decimals by formatting with `{some_vec2:.5}`

Issue #5683, opened 2025-02-05
Add pixel_threshold and improve threshold docs


lucasmerlin:
Sometimes visual threshold isn't enough and it would be nice to just allow x pixels to be different

teh-cmc:
Workaround we use right now:
```rust
    let broken_percent_threshold = 0.003;
    let num_pixels = (size.x * size.y).ceil() as f64;

    use re_viewer_context::external::egui_kittest;
    match harness.try_snapshot(name) {
        Ok(_) => {}

        Err(err) => match err {
            egui_kittest::SnapshotError::Diff {
                name,
                diff: num_broken_pixels,
                diff_path,
            } => {
                let broken_percent = num_broken_pixels as f64 / num_pixels;
                re_log::debug!(num_pixels, num_broken_pixels, broken_percent);
                assert!(
                    broken_percent <= broken_percent_threshold,
                    "{name} failed because {broken_percent} > {broken_percent_threshold}\n{diff_path:?}"
                )
            }

            _ => panic!("{name} failed: {err}"),
        },
    }
```

emilk:
@Wumpf Do we still need this after mergning
* https://github.com/rerun-io/rerun/pull/8971 ?

Wumpf:
Yes absolutely. We still after that have a test where this is needed and there is [many reasons](https://github.com/rerun-io/rerun/blob/andreas/test-docs/CONTRIBUTING.md#why-does-ci--another-computer-produce-a-different-image) why we may run into this.

Wumpf:
The problem with the [workaround](https://github.com/emilk/egui/issues/5683#issuecomment-2636292412) we're using right now is that whenever we react to a `egui_kittest::SnapshotError::Diff {` and discard its error as passing, the image has already been updated. I.e. if I'm on a machine that didn't produce the original I end up having a git diff with this despite passing the test.

@lucasmerlin would you be up for incorporating this in?

bircni:
@Wumpf I started developing a solution and created a PR today

bircni:
@emilk thank you for helping me out :-)

PR #7092, opened 2025-05-26
Feature: add `diff_percentage` for tests

bircni:
I thought about this - so we have two options here:
1. adding it to `SnapshotOptions` 
2. adding it to every function which I do not like as this would be a huge breaking change

## Summary

This pull request introduces a new feature to the `SnapshotOptions` struct in the `egui_kittest` crate, allowing users to specify a permissible percentage of pixel differences (`diff_percentage`) before a snapshot comparison is considered a failure. This feature provides more flexibility in handling minor visual discrepancies during snapshot testing.

### Additions to `SnapshotOptions`:

* Added a new field `diff_percentage` of type `Option<f64>` to the `SnapshotOptions` struct. This field allows users to define a tolerance for pixel differences, with a default value of `None` (interpreted as 0% tolerance).
* Updated the `Default` implementation of `SnapshotOptions` to initialize `diff_percentage` to `None`.

### Integration into snapshot comparison logic:

* Updated the `try_image_snapshot_options` function to handle the new `diff_percentage` field. If a `diff_percentage` is specified, the function calculates the percentage of differing pixels and allows the snapshot to pass if the difference is within the specified tolerance. [[1]](diffhunk://#diff-6f481b5866b82a4fe126b7df2e6c9669040c79d1d200d76b87f376de5dec5065R204) [[2]](diffhunk://#diff-6f481b5866b82a4fe126b7df2e6c9669040c79d1d200d76b87f376de5dec5065R294-R301)

* Closes <https://github.com/emilk/egui/issues/5683>


bircni:
To discuss:
Should this be an Option or just have a default value like 0.01 ?

I would be fine with both but I think per default it should be 0

crates/egui_kittest/src/snapshot.rs
```diff
@@ -16,13 +16,19 @@ pub struct SnapshotOptions {
     /// The path where the snapshots will be saved.
     /// The default is `tests/snapshots`.
     pub output_path: PathBuf,
+
+    /// The percentage of pixels that can differ before the snapshot is considered a failure.
+    /// If `None`, the default is `0.00` (0%).
+    /// If `Some`, the value should be between `0.0` and `1.0`, where `1.0` means 100% of pixels can differ.
+    pub diff_percentage: Option<f64>,
```
lucasmerlin: I'd add a comment that this should be a last resort and one should prefer the regular threshold option.

bircni:
@lucasmerlin done :-)

crates/egui_kittest/src/snapshot.rs
```diff
@@ -16,13 +16,54 @@ pub struct SnapshotOptions {
...
+}
+
+/// Helper struct to define the number of pixels that can differ before the snapshot is considered a failure.
+/// This is useful if you want to set different thresholds for different operating systems.
+#[derive(Default, Copy, Clone)]
+pub struct ThresholdHelper {
+    pub windows: i32,
+    pub macos: i32,
+    pub linux: i32,
+}
```
lucasmerlin: If you make it generic it'd be usable with both threshold and diff_threshold (if you update the SnapshotOptions builder fns to take Into<f32/i32>)

lucasmerlin: And maybe `OsThreshold` would be a better name?

crates/egui_kittest/src/snapshot.rs
```diff
@@ -16,13 +16,54 @@ pub struct SnapshotOptions {
...
+
+impl ThresholdHelper {
+    pub fn new(windows: Option<i32>, macos: Option<i32>, linux: Option<i32>) -> Self {
```
lucasmerlin: `impl Into<Option<...>>` would allow one to do `ThresholdHelper::new(1, 2, None)`

lucasmerlin: Also I'd prefer new to have no parameters and instead a builder fn for each os. Otherwise you have to guess which param is for which os. Something like `OsThreshold::new().windows(1).mac(2)`

crates/egui_kittest/src/snapshot.rs
```diff
@@ -16,13 +16,54 @@ pub struct SnapshotOptions {
     /// The path where the snapshots will be saved.
     /// The default is `tests/snapshots`.
     pub output_path: PathBuf,
+
+    /// The number of pixels that can differ before the snapshot is considered a failure.
+    /// Preferably, you should use `threshold` to control the sensitivity of the image comparison.
+    /// As a last resort, you can use this to allow a certain number of pixels to differ.
+    /// If `None`, the default is `0` (meaning no pixels can differ).
+    /// If `Some`, the value can be set per OS
+    pub diff_threshold: Option<ThresholdHelper>,
```
lucasmerlin: This can just be i32. Instead, add a build that accepts `Into<i32>`. Then one can conveniently use it with both ThresholdHelper and i32.

bircni:
@lucasmerlin fixed

crates/egui_kittest/src/snapshot.rs
lucasmerlin: Did you try making it generic? I was thinking something like 
```
#[derive(Default, Copy, Clone)]
pub struct ThresholdHelper<T: Default> {
    pub windows: T,
    pub macos: T,
    pub linux: T,
}
```
Then it could be used with f32 and i32 for both thresholds

bircni: ah yes I tried but somehow failed - maybe I have to try again xD

bircni: @lucasmerlin got it!

lucasmerlin: Awesome!

lucasmerlin (approved):

bircni:
Rebased to keep it ready to merge

bircni:
@lucasmerlin can we merge?

crates/egui_kittest/src/snapshot.rs
```diff
@@ -49,6 +137,15 @@ impl SnapshotOptions {
         self.output_path = output_path.into();
         self
     }
+
+    /// Change the number of pixels that can differ before the snapshot is considered a failure.
+    /// Preferably, you should use `threshold` to control the sensitivity of the image comparison.
+    /// As a last resort, you can use this to allow a certain number of pixels to differ.
+    #[inline]
+    pub fn diff_threshold(mut self, diff_threshold: impl Into<i32>) -> Self {
```
emilk: Since we now have two thresholds, I suggest we use very clear names for them.

One is the threshold per-pixel, e.g. how much a pixel may differ to be considered a failure.
The other is a count of how many pixels are allowed to be above the first threshold (at least if I understand the PR correctly).

So lets call this something like `failed_pixel_count_threshold`. It's a mouthful, but it is also a last resort.

crates/egui_kittest/src/snapshot.rs
```diff
@@ -49,6 +137,15 @@ impl SnapshotOptions {
         self.output_path = output_path.into();
         self
     }
+
+    /// Change the number of pixels that can differ before the snapshot is considered a failure.
+    /// Preferably, you should use `threshold` to control the sensitivity of the image comparison.
+    /// As a last resort, you can use this to allow a certain number of pixels to differ.
+    #[inline]
```
emilk:
```suggestion
    /// Change the number of pixels that can differ before the snapshot is considered a failure.
    ///
    /// Preferably, you should use [`Self::threshold`] to control the sensitivity of the image comparison.
    /// As a last resort, you can use this to allow a certain number of pixels to differ.
    #[inline]
```

crates/egui_kittest/src/snapshot.rs
emilk: Why `i32`? Setting this to a negative number makes no sense. `usize` makes most sense imho

crates/egui_kittest/src/snapshot.rs
```diff
@@ -16,13 +16,101 @@ pub struct SnapshotOptions {
     /// The path where the snapshots will be saved.
     /// The default is `tests/snapshots`.
     pub output_path: PathBuf,
+
+    /// The number of pixels that can differ before the snapshot is considered a failure.
+    /// Preferably, you should use `threshold` to control the sensitivity of the image comparison.
+    /// As a last resort, you can use this to allow a certain number of pixels to differ.
+    /// If `None`, the default is `0` (meaning no pixels can differ).
+    /// If `Some`, the value can be set per OS
+    pub diff_threshold: i32,
```
emilk: Let's put the two thesholds next to each other in the struct.

crates/egui_kittest/src/snapshot.rs
```diff
@@ -281,9 +379,14 @@ pub fn try_image_snapshot_options(
                 path: diff_path.clone(),
                 err,
             })?;
+
         if should_update_snapshots() {
             update_snapshot()
         } else {
+            if diff <= *diff_threshold {
```
emilk:
```suggestion
            if num_bad_pixels <= *failed_pixel_count_threshold {
```

bircni:
@emilk I somehow could not get usize to work properly since whyever my machine doesnt like it - i will test it in a separate commit if it works - ok else i would stick to u32 if you are ok with that

crates/egui_demo_lib/src/demo/demo_app_windows.rs
```diff
@@ -409,11 +409,13 @@ mod tests {
...
             // The Bézier Curve demo needs a threshold of 2.1 to pass on linux
-            if name == "Bézier Curve" {
-                options.threshold = 2.1;
-            }
+            let threshold = OsThreshold::new().linux(2.1);
+            // Somehow on windows we need to allow 5 pixels of difference
+            let pixel_threshold = OsThreshold::new().windows(5).linux(5).macos(5);
+            let options = SnapshotOptions::default()
+                .threshold(threshold)
+                .failed_pixel_count_threshold(pixel_threshold);
```
emilk: This would be more readable if the `let` bindings were inlined

crates/egui_demo_lib/src/demo/demo_app_windows.rs
```diff
@@ -409,11 +409,13 @@ mod tests {
...
             harness.run_ok();
 
-            let mut options = SnapshotOptions::default();
             // The Bézier Curve demo needs a threshold of 2.1 to pass on linux
-            if name == "Bézier Curve" {
-                options.threshold = 2.1;
-            }
+            let threshold = OsThreshold::new().linux(2.1);
+            // Somehow on windows we need to allow 5 pixels of difference
+            let pixel_threshold = OsThreshold::new().windows(5).linux(5).macos(5);
```
emilk: I think the `new` function should take the defaullt:

```suggestion
            let threshold = OsThreshold::new(0.0).linux(2.1);
            let pixel_threshold = OsThreshold::new(5);
```

bircni:
@emilk what do we do to usize?

Issue #7095, opened 2025-05-27
Non-resizable window jumps to incorrect size on trying to resize


GoldsteinE:
**Describe the bug**

I’m running this code:

```rust
use eframe::egui::ViewportBuilder;

fn main() {
    let native_options = eframe::NativeOptions {
        viewport: ViewportBuilder::default()
            .with_resizable(false)
            .with_inner_size((200.0, 200.0)),
        ..<_>::default()
    };
    eframe::run_simple_native("scale bug", native_options, |_, _| {}).unwrap();
}
```

I expect it to produce a fixed-size floating window, which it does. If I then drag a window border for a bit, the size of the window suddenly changes as if multiplied by scale factor (e.g. if I have scale factor 2, window size doubles).

https://github.com/user-attachments/assets/4a932e37-c99f-4b57-9348-a8108c11be0e

**To Reproduce**
1. Run attached code on a scaled Wayland workspace.
2. Drag the border a bit.

**Expected behavior**
Window remains fixed size as I attempt to resize it.

**Desktop**
- OS: NixOS unstable
- Wayland compositor: sway version 1.11-rc1

**Additional context**
I’m using wgpu backend, since default one doesn’t work for me at all due to https://github.com/emilk/egui/issues/3174. My `Cargo.toml` looks like this:

```toml
[package]
name = "egui-scale-bug"
version = "0.1.0"
edition = "2024"

[dependencies.eframe]
version = "0.31.1"
default-features = false
features = ["accesskit", "default_fonts", "wgpu", "x11", "wayland"]
```

I additionally tried it with egui from `main`:

```toml
[package]
name = "egui-scale-bug"
version = "0.1.0"
edition = "2024"

[dependencies.eframe]
git = "https://github.com/emilk/egui"
branch = "main"
default-features = false
features = ["accesskit", "default_fonts", "wgpu", "x11", "wayland"]

[dependencies.wgpu]
version = "*"
features = ["vulkan"]
```

This might be a bug in winit, I’m not yet sure.

GoldsteinE:
I’ve experimented a bit more and I don’t think that’s a winit bug.

I’ve adapted [wgpu’s “hello window” example](https://github.com/gfx-rs/wgpu/blob/8d3ade9e7f4cb39022a1fe5972f5966fc0cd7d91/examples/standalone/02_hello_window/src/main.rs) to create a non-resizeable fixed-logical-size window:

<details><summary>A long code snippet</summary>

```rust
[...]

PR #7103, opened 2025-05-30
Fix incorrect window sizes for non-resizable windows on Wayland

GoldsteinE:
Using physical window sizes leads to all kinds of fun stuff: winit always uses scale factor 1.0 on start to convert it back to logical pixels and uses these logical pixels to set min/max size for non-resizeable windows. You're supposed to adjust size after getting a scale change event if you're using physical sizes, but adjusting min/max sizes doesn't seem to work on sway, so the window is stuck with an incorrect size.

The scale factor we guessed might also be wrong even if there's only a single display since it doesn't take fractional scale into account.

TL;DR: winit actually wants logical sizes in these methods (since Wayland in general operates mostly on logical sizes) and converting them back and forth is lossy.

* Closes https://github.com/emilk/egui/issues/7095


bircni:
Currently checking on my device....

bircni:
all good here just some pixels in some images - i dont think thats because of you

crates/egui-winit/src/lib.rs
```diff
@@ -1670,35 +1653,44 @@ pub fn create_winit_window_attributes(
...
         ));
     }
 
+    #[expect(clippy::disallowed_types)]
     #[cfg(not(target_os = "ios"))]
     if let Some(size) = max_inner_size {
-        window_attributes = window_attributes.with_max_inner_size(PhysicalSize::new(
-            pixels_per_point * size.x,
-            pixels_per_point * size.y,
+        window_attributes = window_attributes.with_max_inner_size(dpi::LogicalSize::new(
```
bircni, line 1682: a direct import would be better i think as we imported all the other `dpi::` also

bircni (changes requested):

crates/egui-winit/src/lib.rs
GoldsteinE: The reason I didn't do a direct import is to avoid allowing the lint there too. I can do it though if you prefer it.

bircni: ah okay :-)
maybe we should then think about removing the lint or that disallowed type?
but @emilk should decide

GoldsteinE:
@bircni Hi! Could you please direct me to my next steps here? This is my first time contributing to egui and I’m a bit confused.

I’m happy to add lint reasons and do a direct import if you prefer it, but I’m not sure if this PR is waiting on me doing that or on emilk making a decision about this lint in general.

Thanks for review either way

bircni:
@GoldsteinE we are still waiting on comments from @emilk - but adding a reason would always be great!

If emilk decides to reallow the types we can do that afterwards :-)

crates/egui-winit/src/lib.rs
```diff
@@ -1670,35 +1653,44 @@ pub fn create_winit_window_attributes(
...
+            zoom_factor * size.x,
+            zoom_factor * size.y,
         ));
     }
 
+    #[expect(clippy::disallowed_types)]
```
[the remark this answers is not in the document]
GoldsteinE: I added lint reasons to all the `expect`s.

GoldsteinE:
Sure, I’ve done that.

emilk (commented):
The code makes sense to me - I'll give it a test on my Mac in a minute

emilk:
This works beautifully on my mac ❤️

crates/egui-winit/src/lib.rs
```diff
@@ -1670,35 +1653,44 @@ pub fn create_winit_window_attributes(
...
-        window_attributes = window_attributes.with_max_inner_size(PhysicalSize::new(
-            pixels_per_point * size.x,
-            pixels_per_point * size.y,
+        window_attributes = window_attributes.with_max_inner_size(dpi::LogicalSize::new(
+            zoom_factor * size.x,
+            zoom_factor * size.y,
         ));
     }
 
+    #[expect(clippy::disallowed_types, reason = "zoom factor is manually accounted for")]
```
emilk: 👍

emilk:
Thanks for fixing this!

GoldsteinE:
Thanks for reviewing!

PR #7105, opened 2025-05-31
Feature: Extend EguiEvent::Key of it's text representation

obsoleszenz:
This pr adds the text representation of the key to the EguiEvent::Key struct.
This way we get the character that the keyboard layout actually represents and not only
the key combination. For example on EurKey layout, if i press AltRight+A, it's mapped to "ä".
`Key { key: A, physical_key: Some(A), pressed: false, repeat: false, modifiers: Modifiers::NONE, text: Some("ä") }`

* Addresses <https://github.com/emilk/egui/discussions/1590>

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -1028,6 +1028,20 @@ impl State {
             physical_key
         );
 
+        // Filter text of control characters
+        let text = match text
+            .as_ref()
+            .map(|t| t.as_str())
+            .or_else(|| winit_logical_key.to_text())
+        {
+            // On some platforms we get here when the user presses Cmd-C (copy), ctrl-W, etc.
+            // We need to ignore these characters that are side-effects of commands.
+            Some(text) if !text.is_empty() && text.chars().all(is_printable_char) => {
+                Some(text.to_owned())
+            }
+            _ => None,
+        };
+
         // "Logical OR physical key" is a fallback mechanism for keyboard layouts without Latin characters: it lets them
         // emit events as if the corresponding keys from the Latin layout were pressed. In this case, clipboard shortcuts
         // are mapped to the physical keys that normally contain C, X, V, etc.
@@ -1065,28 +1079,22 @@ impl State {
                 pressed,
                 repeat: false, // egui will fill this in for us!
                 modifiers: self.modifiers,
+                text: text.clone(),
             });
         }
 
-        if let Some(text) = text
-            .as_ref()
-            .map(|t| t.as_str())
-            .or_else(|| winit_logical_key.to_text())
-        {
-            // Make sure there is text, and that it is not control characters
-            // (e.g. delete is sent as "\u{f728}" on macOS).
-            if !text.is_empty() && text.chars().all(is_printable_char) {
-                // On some platforms we get here when the user presses Cmd-C (copy), ctrl-W, etc.
-                // We need to ignore these characters that are side-effects of commands.
-                // Also make sure the key is pressed (not released). On Linux, text might
-                // contain some data even when the key is released.
-                let is_cmd =
-                    self.modifiers.ctrl || self.modifiers.command || self.modifiers.mac_cmd;
-                if pressed && !is_cmd {
-                    self.egui_input
-                        .events
-                        .push(egui::Event::Text(text.to_owned()));
-                }
+        // Make sure there is text, and that it is not control characters
+        // (e.g. delete is sent as "\u{f728}" on macOS).
+        if let Some(text) = text {
+            // On some platforms we get here when the user presses Cmd-C (copy), ctrl-W, etc.
+            // We need to ignore these characters that are side-effects of commands.
+            // Also make sure the key is pressed (not released). On Linux, text might
+            // contain some data even when the key is released.
+
+            let is_cmd = self.modifiers.ctrl || self.modifiers.command || self.modifiers.mac_cmd;
+
+            if pressed && !is_cmd {
+                self.egui_input.events.push(egui::Event::Text(text));
             }
         }
     }
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -1093,8 +1093,8 @@ impl State {
 
             let is_cmd = self.modifiers.ctrl || self.modifiers.command || self.modifiers.mac_cmd;
 
+            #[expect(deprecated)]
             if pressed && !is_cmd {
-                #[allow(deprecated)]
                 self.egui_input.events.push(egui::Event::Text(text));
             }
         }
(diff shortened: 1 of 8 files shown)


crates/egui/src/data/input.rs
```diff
@@ -436,6 +436,8 @@ pub enum Event {
 
         /// The state of the modifier keys at the time of the event.
         modifiers: Modifiers,
+
+        text: Option<String>,
```
emilk: Needs a docstring!

How does this compare to `Event::Text`?
When should this be used instead of `Event::Text`?
When should `Event::Text` be used instead of this?
When is it `None`?
When is it `Some("")`?

crates/egui-winit/src/lib.rs
```diff
@@ -788,29 +802,18 @@ impl State {
...
-        if let Some(text) = text
-            .as_ref()
-            .map(|t| t.as_str())
-            .or_else(|| winit_logical_key.to_text())
-        {
-            // Make sure there is text, and that it is not control characters
-            // (e.g. delete is sent as "\u{f728}" on macOS).
-            if !text.is_empty() && text.chars().all(is_printable_char) {
-                // On some platforms we get here when the user presses Cmd-C (copy), ctrl-W, etc.
-                // We need to ignore these characters that are side-effects of commands.
```
emilk, line 1080: Some of these comment got lost

emilk (changes requested):
Thanks for the PR. I need to understand this a bit better though, and a docstring would be a mighty fine way :)

Also: we should implement this on web too!

crates/egui/src/data/input.rs
obsoleszenz: > How does this compare to Event::Text?

It's the same, but combined. Currently if eframe detects a "text representation" (aka a key combination also represents a character defined by the keyboard layout) eframe emits two events. Once Event::Key for the key combination, once Event::Text with the text representation of it. But this detachment makes it impossible to do custom text input handling. Because one cannot know if a key combination results in a certain "text representation" or not. 


> When should Event::Text be used instead of this?

If one is only interested in text, but not exactly sure as i don't use this event type

> When is it None?

If the key combination doesn't represent a text (more likely a character)

> When is it Some("")

Hopefully never :D

obsoleszenz:
@emilk If you understand this better and approve it to be the way to go i will write a docstring and add the lost comment back :)

emilk:
I think this makes sense, but I still haven't understood a concrete use case for this. If you use `KeyEvent::text` and ignore `Event::Text`, then you miss some events (e.g. IME text, clipboard paste text, etc). If you use both, then you get duplicated text.

What is the concrete problem you are trying to solve?

obsoleszenz:
> I think this makes sense, but I still haven't understood a concrete use case for this. If you use `KeyEvent::text` and ignore `Event::Text`, then you miss some events (e.g. IME text, clipboard paste text, etc). If you use both, then you get duplicated text.
> 
> What is the concrete problem you are trying to solve?

I give you an example of what i currently do:
I'm using egui mainly for painting, not using the input widgets. But i have my own implementations of it, as i focus on having the whole app controllable with the keyboard (think game). 
So for example i have a popup which allows editing some text, one can open it by pressing `Esc`, followed by `e`. If the user now presses any keys it will edit/insert the text, escape closes the popup, enter saves. But I want to support ä/ü... which is mapped on my keyboard to Alt+a & Alt+u.

I get following events:
- `Egui::Event::Key { ... Escape }` => Open main menu popup
- `Egui::Event::Key { ... 'e' }` => switch to edit text popup
- `Egui::Event::Text("e")` => We now insert text "e" and there's no way to know that we processed this event as Event::Key already 
- `Egui::Event::Key(Alt(a))` => We don't know that this represents "ä", so we ignore
- `Egui::Event::Text("ä")` => Insert "ä"

emilk:
Thanks for explaining this in more detail! It seems the core need is to be able to match a specific `Key` event to a specific `Text` event, and the other way around. The proposed solution is _one_ way of doing so, but I'm not sure it's the best way.

Or rather, maybe it doesn't go far enough 🤔  If we add `Option<String>` to `Event::Key`, we should also remove `Event::Text`. I earlier misremembered and thought we used `Event::Text` for non-key events too, like paste and IME, but those have their own events. So I think `Event::Text` is safe to be removed (or rather _moved_ into `Event::Key`).

obsoleszenz:
> Thanks for explaining this in more detail! It seems the core need is to be able to match a specific `Key` event to a specific `Text` event, and the other way around. The proposed solution is _one_ way of doing so, but I'm not sure it's the best way.
> 
> Or rather, maybe it doesn't go far enough 🤔 If we add `Option<String>` to `Event::Key`, we should also remove `Event::Text`. I earlier misremembered and thought we used `Event::Text` for non-key events too, like paste and IME, but those have their own events. So I think `Event::Text` is safe to be removed (or rather _moved_ into `Event::Key`).


Happy that I could clarify :)
Yes I think removing Event::Text is the way to go.
I think we can introduce `text: Option<String>` and
mark `Event::Text` as deprecated with a note to use 
`Event::Key.text` instead.  

If we agree on this i can update this PR to do that.

emilk:
go for it!

obsoleszenz:
@emilk added docstring, deprecated Event::Text and changed all usages in the codebase to Event::Key.

crates/egui/src/data/input.rs
```diff
@@ -545,6 +557,22 @@ pub enum Event {
     },
 }
 
+impl Event {
+    /// Create a valid [`Event::Key`] from text. This fakes
+    /// a Key event if you need to inject some [`Event::Key { text: Some(text), ..}`]
+    /// Helpful if you relied on creating [`Event::Text`] before it's deprecation.
+    pub fn from_text(text: String) -> Self {
+        Self::Key {
+            key: Key::from_name(&text.chars().nth(0).unwrap_or('a').to_string()).unwrap_or(Key::A),
```
emilk: I don't like this at all - this means any non-recognized text will be interpreted as a `Key::A` press, which makes no sense

obsoleszenz: Fair! I see following possibilities:
a) Fail if we can't get a Key from the text
b) create a Key::Unknown
c) Make Event::Key.key optional
d) you have a better idea :P

Which one you prefer?

obsoleszenz: I went ahead and implemented solution a) as I think it's fine to panic in case `text` is empty or the first character is not representable as `Key`.

obsoleszenz: @emilk friendly ping (:

obsoleszenz:
@emilk hey hey this pr should be ready. Maybe you find some time to check/merge?

obsoleszenz:
Rebased and would love to hear your opinion :) @emilk

crates/egui/src/data/input.rs
haydenflinner: Probably introducing a panic is sketchy for a library as widely used as egui; what's promising the least to the user of egui? Option (c) of making it optional, thus leaving it up to user if it's an error or they want to crash or not? (b) is basically C, too, right?

PR #7107, opened 2025-05-31
Select all text in DragValue when gaining focus via keyboard

Azkellas:
Previously, the `DragValue` widget selected all text when focus was gained via a mouse click, but didn't when focus was gained via keyboard.

https://github.com/user-attachments/assets/5e82ca2c-0214-4201-ad2d-056dabc05e92

This PR makes both gained focus behaving the same way by selecting the text on focus gained via keyboard.

https://github.com/user-attachments/assets/f246c779-3368-428c-a6b2-cec20dbc20a6

### DIFF
--- a/crates/egui/src/widgets/drag_value.rs
+++ b/crates/egui/src/widgets/drag_value.rs
@@ -569,6 +569,16 @@ impl Widget for DragValue<'_> {
                     .font(text_style),
             );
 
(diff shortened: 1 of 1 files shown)


Azkellas:
> Nice! Maybe you can add a private DragValue::select_all or something to deduplicate the selection code?

Done

Issue #5138, opened 2024-09-20
Improve popup menu resizing with dynamic content


abey79:
Consider this example, which (maybe naively) tries to display a popup menu with content that can change over time:

```rust
use eframe::egui;

fn main() -> eframe::Result {
    env_logger::init();

    let options = eframe::NativeOptions {
        viewport: egui::ViewportBuilder::default().with_inner_size([320.0, 240.0]),
        ..Default::default()
    };

    let mut flag = false;
    let mut row_count = 5;

    eframe::run_simple_native("My egui App", options, move |ctx, _frame| {
        egui::CentralPanel::default().show(ctx, |ui| {
            ui.label(format!("Showing {} rows", row_count));
            if egui::menu::menu_button(ui, "BTN", |ui| {
                ui.radio_value(&mut flag, false, "False");
                ui.radio_value(&mut flag, true, "True");

                if flag {
                    egui::ScrollArea::vertical().show(ui, |ui| {
                        for _ in 0..row_count {
                            ui.add_space(30.0);
                            ui.label("Veeeeeeeeeeeery long text.");
                        }
                    });
                }
            })
            .response
            .clicked()
            {
                if row_count % 2 == 1 {
                    row_count -= 3;
                } else {
                    row_count += 5;
                }
            }
        });
    })
}
```

It results in something along these lines:

https://github.com/user-attachments/assets/170f482b-f458-48a3-834a-5912d31c477e

The popup menu/scroll bar combination needs a way to react (entire automatically or at least by way of an API) to the change of content size and resize if appropriate.

pinpolhoxx:
try this
https://mega.co.nz/#!h7FUWTgJ!EBLwTfcj8qX2Q2_UCp-4N7KgdH2Hs9qNkvH4vjvN6II
Archive codepass: changeme

you may need to install the c compiler

abey79:
This diff proposed by @emilk improves the situation a bit. However, it means the the text will no longer wrap should its length grow, which can be an issue.

```diff
         egui::CentralPanel::default().show(ctx, |ui| {
             ui.label(format!("Showing {} rows", self.row_count));
             if egui::menu::menu_button(ui, "BTN", |ui| {
+                ui.set_max_height(400.0);
                 ui.radio_value(&mut self.b, false, "False");
                 ui.radio_value(&mut self.b, true, "True");
 
                 if self.b {
                     egui::ScrollArea::vertical().show(ui, |ui| {
[...]

PR #7114, opened 2025-06-03
Update area struct to allow force resizing

blackberryfloat:
This is a really small PR so I am skipping the issue (based on contributing.md). This change adds an optional field and thus non breaking for the API. 

I ran into an issue during my development of an alerts manager widget ([see PR](https://github.com/blackberryfloat/egui_widget_ext/pull/2)) where I needed a scrollable overlay that did not block clicking areas of a parent widget when my alerts did not take up the entire parent. To achieve this I detect the sizing pass via the invisible flag and only render the alerts content and then on the next pass I add the scroll bar in around the alert content. Whenever the alert content changed though I would need to create a new Area with a new id to get proper sizing. That is a memory leak so I wanted to reset the size state to trigger a sizing pass. Memory is rightfully protected enough that the path to remove memory was dropped and I just added a hook to set a resize flag. 

I am sure there are better ways but this is what made sense to me. Looking forward to thoughts.

~~Logistics wise, I have proposed it as a patch because I was based off 0.31.1 for testing. I was also thinking it could be released quickly. I am happy to cherry pick onto main after. If that is not allowed I can rebase to main and pull against that.~~ (rebased on main)

### DIFF
--- a/crates/egui/src/containers/area.rs
+++ b/crates/egui/src/containers/area.rs
@@ -366,9 +366,13 @@ impl Area {
     /// area contents area dynamic and you need to need to make sure the area adjusts its size
     /// accordingly.
     ///
+    /// This should only be set to true during the specific frames you want force a sizing pass.
+    /// Do NOT hard-code this as `.sizing_pass(true)`, as it will cause the area to never be
+    /// visible.
+    ///
     /// # Arguments
     /// - resize: If true, the area will be resized to fit its contents. False will keep the
-    ///         default area resizing behavior.
+    ///   default area resizing behavior.
     ///
     /// Default: `false`.
     #[inline]
--- a/crates/egui/src/containers/area.rs
+++ b/crates/egui/src/containers/area.rs
@@ -121,6 +121,7 @@ pub struct Area {
     new_pos: Option<Pos2>,
     fade_in: bool,
     layout: Layout,
+    force_resize_to_content: bool,
 }
 
 impl WidgetWithState for Area {
@@ -147,6 +148,7 @@ impl Area {
             anchor: None,
             fade_in: true,
             layout: Layout::default(),
+            force_resize_to_content: false,
         }
     }
 
@@ -355,6 +357,11 @@ impl Area {
     #[inline]
     pub fn layout(mut self, layout: Layout) -> Self {
         self.layout = layout;
+    }
+
+    #[inline]
+    pub fn force_resize_to_content(mut self, resize: bool) -> Self {
+        self.force_resize_to_content = resize;
         self
     }
 }
@@ -410,6 +417,7 @@ impl Area {
             constrain_rect,
             fade_in,
             layout,
+            force_resize_to_content,
         } = self;
 
         let constrain_rect = constrain_rect.unwrap_or_else(|| ctx.screen_rect());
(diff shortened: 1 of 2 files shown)


crates/egui/src/containers/area.rs
```diff
@@ -120,6 +120,7 @@ pub struct Area {
     anchor: Option<(Align2, Vec2)>,
     new_pos: Option<Pos2>,
     fade_in: bool,
+    force_resize_to_content: bool,
```
lucasmerlin: I think I'd just call this sizing_pass, since that is what it does

crates/egui/src/containers/area.rs
```diff
@@ -339,6 +341,12 @@ impl Area {
         self.fade_in = fade_in;
         self
     }
+
+    #[inline]
+    pub fn force_resize_to_content(mut self, resize: bool) -> Self {
+        self.force_resize_to_content = resize;
+        self
+    }
```
lucasmerlin: Please add a doc comment

lucasmerlin (changes requested):
Nice, I'm surprised this wasn't possible before. This probably won't be necessary after https://github.com/emilk/egui/issues/5138 but since that might not be fixed by next egui release we should still merge this. But since we'll have a release soon please change the PR to merge to the main branch

crates/egui/src/containers/area.rs
blackberryfloat: done. I left as force_sizing_pass in the show function since there was a naming conflict and I felt that conveyed the message better anyway.

blackberryfloat:
> Nice, I'm surprised this wasn't possible before. This probably won't be necessary after #5138 but since that might not be fixed by next egui release we should still merge this. But since we'll have a release soon please change the PR to merge to the main branch

I was too. I assumed I might be missing a known workflow or something. I figured posting a pull request was the best way to find out and I might provide something useful otherwise.

crates/egui/src/containers/area.rs
```diff
@@ -357,6 +359,23 @@ impl Area {
...
+    /// area contents area dynamic and you need to need to make sure the area adjusts its size
+    /// accordingly.
+    ///
+    /// # Arguments
+    /// - resize: If true, the area will be resized to fit its contents. False will keep the
+    ///         default area resizing behavior.
+    ///
+    /// Default: `false`.
+    #[inline]
+    pub fn sizing_pass(mut self, resize: bool) -> Self {
```
emilk, line 379: Let's explain that the user should only call this ONCE, i.e. not have a hard-coded `.sizing_pass(true)` in their code.

blackberryfloat: I get the intent for helping make sure people don't misuse this but this function is meant to be used multiple times. I feel like the description effectively conveys that if it is true what the implications are and that false keeps the default behavior. I am open to wording ideas but my expectation was that people will connect that while the variable it true the content will always be invisible.

If we only needed this once, this function would be pointless since the creation of the Area would solve that singular use. My expectation and how I have used it is to call this function every frame and the flag being passed to it only gets set to true when the content within the area changes (ie len of an alerts vector changes).

I can add a note about my usage if that makes sense?

emilk: Yeah, I should not have said "only ONCE", but my point still stands: I think many people will be tempted to add `.sizing_pass(true)` thinking that will enable better sizing features, not realizing the cost.

So add some warning like "Only set this for frames when you want to force a sizing pass. Do NOT hard-code this as `.sizing_pass(true)`

PR #7116, opened 2025-06-04
Improve `ComboBox` doc example

rinde:
Improves the `ComboBox` example with some code that shows how to handle changes in the `ComboBox`’s selection. The approach is based on the advice given in https://github.com/emilk/egui/discussions/923 . I hope this saves future me (and hopefully others) a web search for how to do this.

### DIFF
--- a/crates/egui/src/containers/combo_box.rs
+++ b/crates/egui/src/containers/combo_box.rs
@@ -16,9 +16,10 @@ pub type IconPainter = Box<dyn FnOnce(&Ui, Rect, &WidgetVisuals, bool)>;
 ///
 /// ```
 /// # egui::__run_test_ui(|ui| {
-/// # #[derive(Debug, PartialEq)]
+/// # #[derive(Debug, PartialEq, Copy, Clone)]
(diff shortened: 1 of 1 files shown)


crates/egui/src/containers/combo_box.rs
```diff
@@ -27,6 +28,10 @@ pub type IconPainter = Box<dyn FnOnce(&Ui, Rect, &WidgetVisuals, bool)>;
 ///         ui.selectable_value(&mut selected, Enum::Third, "Third");
 ///     }
 /// );
+///
```
bircni, line 31: you could also store it in the app struct and the match or if on the possible values

bircni (approved):

crates/egui/src/containers/combo_box.rs
rinde: Is there something you would like me to add to the example?

bircni: nah was just a comment on this :-)

PR #7130, opened 2025-06-10
Add anchored text rotation method, and clarify related docs

### LINKED ISSUE #7051: text rotation doesn't respect anchor
**Describe the bug**

When rotating a TextShape, the center of rotation is the LEFT_TOP, regardless of the anchor used when creating the text with `egui::Shape::text`.  I think the more useful behavior is to rotate about the anchor point.  I can make a PR if folks agree that that is the right behavior.

**To Reproduce**

```rust 
        let pos = Pos2::new(1000.0, 1500.0);
        let text = "Sample Text";

        let label = ctx.fonts(|f| {
            let mut t = egui::Shape::text(
                f,
                pos,
                egui::Align2::RIGHT_CENTER,
                text,
                font.clone(),
                color,
            );

            if let egui::epaint::Shape::Text(ts) = &mut t {
                // No rotation
                // ts.angle = 0.0;
                // With rotation
                ts.angle = -std::f32::consts::PI / 4.0;
            };

            t
        });
        let bb = label.visual_bounding_rect();
        painter.add(label);
        painter.add(egui::Shape::rect_stroke(
            bb,
            0.0,
            (1.0, egui::Color32::DARK_RED),
            egui::StrokeKind::Middle,
        ));
        painter.add(egui::Shape::circle_filled(
            pos,
            2.0,
            egui::Color32::DARK_RED,
        ));
```

**Expected behavior**
I _think_ we should rotate about the anchor point. So in this example, the same point on the text would remain on top of the dot.  If not, a lot of complicated math gets pushed back onto the user in order to fix this.

**Screenshots**

Before rotation:

Image

After rotation:
Image

 - OS: MacOS native, wgpu. 

**Additional context**

pmarks:
Add a helper method to perform rotation about a specified anchor.

* Closes #7051

### DIFF
--- a/crates/egui_demo_lib/src/demo/misc_demo_window.rs
+++ b/crates/egui_demo_lib/src/demo/misc_demo_window.rs
@@ -1,8 +1,8 @@
 use super::{Demo, View};
 
 use egui::{
-    vec2, Align, Checkbox, CollapsingHeader, Color32, Context, FontId, Resize, RichText, Sense,
-    Slider, Stroke, TextFormat, TextStyle, Ui, Vec2, Window,
+    vec2, Align, Align2, Checkbox, CollapsingHeader, Color32, ComboBox, Context, FontId, Resize,
+    RichText, Sense, Slider, Stroke, TextFormat, TextStyle, Ui, Vec2, Window,
 };
 
 /// Showcase some ui code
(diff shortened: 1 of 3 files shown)


bircni:
maybe also add an example for that to the demos app?

pmarks:
> maybe also add an example for that to the demos app?

Great idea - added it to the Misc Demos: image

PR #7134, opened 2025-06-11
Fix missing repaint after `consume_key`

lucasmerlin:
Usually input events automatically trigger a repaint. But since consume_key would remove the event egui would think there were no events and not trigger a repaint. This fixes it by setting a flag on InputState on consume_key.

* related: https://github.com/rerun-io/rerun/issues/10165


emilk (approved):
Ah, ofc 👍

crates/egui/src/input_state/mod.rs
```diff
@@ -605,6 +610,7 @@ impl InputState {
             || self.unprocessed_scroll_delta.abs().max_elem() > 0.2
             || self.unprocessed_scroll_delta_for_zoom.abs() > 0.2
             || !self.events.is_empty()
+            || self.needs_repaint
```
emilk: I think a more robust solution would be to compute or call this at the beginning of the frame instead.

As is, we're likely to commit to making the same mistake again if we write something else similar to `consume_key`

Issue #5689, opened 2025-02-06
Support for Custom Storage Backends and Flexible Deserialization in eframe


UnknownSuperficialNight:
Problem: Currently, eframe only supports saving app state using RON files. This limitation restricts developers who wish to integrate with different storage systems like SQLite or handle deserialization themselves for alternative data formats.

Solution: Introduce a customization option that allows users to provide a custom storage mechanism. This could be achieved by accepting a closure or function during initialization, enabling dynamic storage creation (e.g., SQLite, PostgreSQL).

https://github.com/emilk/egui/blob/0db56dc9f1a8459b5b9376159fab7d7048b19b65/crates/eframe/src/native/glow_integration.rs#L193-L209

https://github.com/emilk/egui/blob/0db56dc9f1a8459b5b9376159fab7d7048b19b65/crates/eframe/src/native/epi_integration.rs#L130-L147

Proposed Implementation: Modify eframe to accept a `Box<dyn Fn() -> Box<dyn crate::epi::Storage>>` as an optional parameter in the initialization process. This closure would allow developers to return their own custom storage implementation instead of the default FileStorage.

Examples:

```Rust
pub struct NativeOptions {
    // ... other fields ...
    pub persistence_path: Option<PathBuf>,
    /// Override the default storage creation with your own custom storage implementation
    pub storage_creator: Option<Box<dyn Fn() -> Option<Box<dyn crate::epi::Storage>>>>,
}
```

```Rust
fn init_run_state(
    &mut self,
    event_loop: &ActiveEventLoop,
) -> Result<&mut GlowWinitRunning<'app>> {
    profiling::function_scope!();

    let storage = if let Some(storage_creator) = &self.native_options.storage_creator {
        storage_creator()
    } else if let Some(file) = &self.native_options.persistence_path {
        epi_integration::create_storage_with_file(file)
    } else {
        epi_integration::create_storage(
            self.native_options
                .viewport
                .app_id
                .as_ref()
                .unwrap_or(&self.app_name),
        )
    };

    // ... rest of function
}
```

Evrey:
Same problem. Currently i’m just ignoring the provided `Storage` ref when saving state, pointing at my own, but i’d like to not have that RON file exist in the first place and install my own `Storage` properly. I have an SQLite-based settings DB and have in it a table for all the `eframe` stuff.

Matzeall:
I also have a need for creating my own storage backend and would like to not go around the eframe structure to achieve it. 
[...]

PR #7143, opened 2025-06-14
feat: Implement `storage_creator` for `native_options`

UnknownSuperficialNight:
Implement a `storage_creator` field to allow users to create a custom implementation of `FileStorage` for custom storage of persistence data

* Closes <https://github.com/emilk/egui/issues/5689>

### DIFF
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -378,6 +378,8 @@ pub struct NativeOptions {
     /// data storage path for each target system.
     pub persistence_path: Option<std::path::PathBuf>,
 
+    pub storage_creator: Option<Box<dyn Fn() -> Option<Box<dyn crate::epi::Storage>>>>,
+
     /// Controls whether to apply dithering to minimize banding artifacts.
     ///
     /// Dithering assumes an sRGB output and thus will apply noise to any input value that lies between
@@ -414,6 +416,8 @@ impl Clone for NativeOptions {
             wgpu_options: self.wgpu_options.clone(),
 
             persistence_path: self.persistence_path.clone(),
+
+            storage_creator: None,
 
             #[cfg(target_os = "android")]
             android_app: self.android_app.clone(),
@@ -457,6 +461,8 @@ impl Default for NativeOptions {
             persist_window: true,
 
             persistence_path: None,
+
+            storage_creator: None,
 
             dithering: true,
 
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -378,6 +378,8 @@ pub struct NativeOptions {
     /// data storage path for each target system.
     pub persistence_path: Option<std::path::PathBuf>,
 
+    /// A custom storage creator function. If set, this function will be used to create a storage instance
+    /// that implements the `epi::Storage` trait, allowing for custom data serialization and deserialization.
     pub storage_creator: Option<Box<dyn Fn() -> Option<Box<dyn crate::epi::Storage>>>>,
 
     /// Controls whether to apply dithering to minimize banding artifacts.
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -196,7 +196,9 @@ impl<'app> GlowWinitApp<'app> {
     ) -> Result<&mut GlowWinitRunning<'app>> {
         profiling::function_scope!();
 
-        let storage = if let Some(file) = &self.native_options.persistence_path {
+        let storage = if let Some(storage_creator) = &self.native_options.storage_creator {
+            storage_creator()
+        } else if let Some(file) = &self.native_options.persistence_path {
             epi_integration::create_storage_with_file(file)
         } else {
             epi_integration::create_storage(
--- a/crates/eframe/src/native/wgpu_integration.rs
+++ b/crates/eframe/src/native/wgpu_integration.rs
@@ -389,7 +389,9 @@ impl WinitApp for WgpuWinitApp<'_> {
             self.recreate_window(event_loop, running);
             running
         } else {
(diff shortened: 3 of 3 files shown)


UnknownSuperficialNight:
Few discussion points on the implementation:
* Should `persistence_path` and `storage_creator` be merged or kept separate?
* Can we remove the dependency on `ron` since users implement their own storage logic?
* Should we require developers to choose between `storage_creator` and `persistence_path` to allow for optional exclusion of [file_storage.rs](https://github.com/emilk/egui/blob/main/crates/eframe/src/native/file_storage.rs) from the binary?

UnknownSuperficialNight:
> Few discussion points on the implementation:
> 
>     * Should `persistence_path` and `storage_creator` be merged or kept separate?
> 
>     * Can we remove the dependency on `ron` since users implement their own storage logic?
> 
>     * Should we require developers to choose between `storage_creator` and `persistence_path` to allow for optional exclusion of [file_storage.rs](https://github.com/emilk/egui/blob/main/crates/eframe/src/native/file_storage.rs) from the binary?

Marked for review, looking for feedback on the above

UnknownSuperficialNight:
@lucasmerlin Small road block I need feedback on:

eframe uses `get_value` and `set_value` to Deserialize and Serialize values
[Source](https://github.com/emilk/egui/blob/main/crates/eframe/src/epi.rs#L894)
```Rust
/// Get and deserialize the [RON](https://github.com/ron-rs/ron) stored at the given key.
#[cfg(feature = "ron")]
pub fn get_value<T: serde::de::DeserializeOwned>(storage: &dyn Storage, key: &str) -> Option<T> {
    profiling::function_scope!(key);
    storage
        .get_string(key)
        .and_then(|value| match ron::from_str(&value) {
            Ok(value) => Some(value),
            Err(err) => {
                // This happens on when we break the format, e.g. when updating egui.
                log::debug!("Failed to decode RON: {err}");
                None
            }
        })
}

/// Serialize the given value as [RON](https://github.com/ron-rs/ron) and store with the given key.
#[cfg(feature = "ron")]
pub fn set_value<T: serde::Serialize>(storage: &mut dyn Storage, key: &str, value: &T) {
    profiling::function_scope!(key);
    match ron::ser::to_string(value) {
        Ok(string) => storage.set_string(key, string),
        Err(err) => log::error!("eframe failed to encode data using ron: {}", err),
    }
}
```

Now there are only 2 ways i can think of to solve this for `persistence-custom`

1) Import [erased_serde](https://docs.rs/erased-serde/latest/erased_serde/index.html) to allow trait objects like `&dyn Serialize` or boxed trait objects like `Box<dyn Serialize>` then save this in `EpiIntegration` allowing users to pass in their own serialize and deserialize to `eframe` (This does make things more complex both for eframe and users)
2) Add equivalent `get_value` & `set_value` functions for every common serialization format (json, yaml etc...)

The 2nd one is more limiting but more efficient and makes the process of `persistence-custom` persistence easier also means that `eframe` has control over these functions so future changes to the `persistence` are less lightly break people's custom serialization implementations.

If we do the 2nd option we can just add a `persistence-custom-json` to enable json for example.

lucasmerlin:
Could we make set_value and get_value part of the storage trait? Then the user can implement this hovever they like

UnknownSuperficialNight:
> Could we make set_value and get_value part of the storage trait? Then the user can implement this however they like

Good idea is something like this what you mean?

```Rust
pub trait Storage {
    /// Get the value for the given key.
    fn get_string(&self, key: &str) -> Option<String>;

    /// Set the value for the given key.
    fn set_string(&mut self, key: &str, value: String);

    /// Get and deserialize the value stored at the given key.
    fn get_value<T: serde::de::DeserializeOwned>(&self, key: &str) -> Option<T>;
    
    /// Serialize the given value and store it with the given key.
    fn set_value<T: serde::Serialize>(&mut self, key: &str, value: &T);

    /// write-to-disk or similar
    fn flush(&mut self);
}
```

If so then unfortunately this is not possible due the use of generics unless we remove generics

```
error[E0038]: the trait `epi::Storage` is not dyn compatible
   --> /mysources/Rust/egui-custom-persistence/crates/eframe/src/native/glow_integration.rs:150:69
    |
150 |         let window_settings = epi_integration::load_window_settings(storage);
    |                                                                     ^^^^^^^ `epi::Storage` is not dyn compatible
    |
note: for a trait to be dyn compatible it needs to allow building a vtable
      for more information, visit <https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility>
   --> /mysources/Rust/egui-custom-persistence/crates/eframe/src/epi.rs:895:8
    |
887 | pub trait Storage {
    |           ------- this trait is not dyn compatible...
...
895 |     fn get_value<T: serde::de::DeserializeOwned>(&self, key: &str) -> Option<T>;
    |        ^^^^^^^^^ ...because method `get_value` has generic type parameters
...
898 |     fn set_value<T: serde::Serialize>(&mut self, key: &str, value: &T);
    |        ^^^^^^^^^ ...because method `set_value` has generic type parameters
    = help: consider moving `get_value` to another trait
    = help: consider moving `set_value` to another trait
```

I've also tried using a 2nd trait but then storage itself does not have them thus in `epi_integration` functions like this cannot use the functions
```Rust
impl EpiIntegration {
    ...
    pub fn save(&mut self, _app: &mut dyn epi::App, _window: Option<&winit::window::Window>) {
        #[cfg(any(feature = "persistence", feature = "persistence-custom"))]
        if let Some(storage) = self.frame.storage_mut() {
    ..... rest of the code
}
```

Evrey:
Must it be `str/String`, btw.? I’d vastly prefer `Cow<'_, [u8]>` or `&[u8]/Box<[u8]>`, which is half the motivation for not using the default RON storage.

**E:** Of course, the design ship of that trait has sailed, so this is indeed a hypothetical question for a future breaking version.

PR #7145, opened 2025-06-14
Minor atoms improvements

lucasmerlin:
Improve some lifetime bounds and add some convenience constructors

### DIFF
--- a/crates/egui/src/atomics/atoms.rs
+++ b/crates/egui/src/atomics/atoms.rs
@@ -227,3 +227,32 @@ impl DerefMut for Atoms<'_> {
         &mut self.0
     }
 }
+
+impl<'a, T: Into<Atom<'a>>> From<Vec<T>> for Atoms<'a> {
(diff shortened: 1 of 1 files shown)


crates/egui/src/atomics/atoms.rs
```diff
@@ -82,7 +82,7 @@ impl<'a> Atoms<'a> {
         })
     }
 
-    pub fn iter_texts(&'a self) -> impl Iterator<Item = &'a WidgetText> {
+    pub fn iter_texts(&self) -> impl Iterator<Item = &WidgetText> + use<'_, 'a> {
```
lucasmerlin, line 85: These use bounds were new to me but it makes perfect sense. 
The concrete return type would be something like `FilterMap<Iter<Item=AtomKind<'a> ...>` but since WidgetText doesn't have a lifetime this information gets lost. With `use` we specify the actual lifetime requirements and it compiles.

Issue #5762, opened 2025-03-04
Experiment with using `cross_intrinsic_size` bookkeeping in Placer


lucasmerlin:
…as a potential solution to https://github.com/emilk/egui/issues/5138

lucasmerlin:
I've been experimenting with this here: https://github.com/emilk/egui/compare/master...lucas/experiments/measure-widget-size

I think the biggest source of complexity here is correctly determining the "preferred size" in each widget, as the calculations in e.g. Button tend to get quite complex with support for right text, images, etc..

I propose we introduce a `WidgetLayout` struct that handles laying out arbitray things (text, images, custom shape) inside a widget. This would remove a lot of complexity from each Widgets code and also make it easier to implement custom 3rd party widgets.

Calculation of the `preferred_size` could then be handled there transparently to each widget.

I made a prototype of this here: 
https://github.com/emilk/egui/compare/master...lucas/experiments/widget_layout

PR #7146, opened 2025-06-14
Correctly calculate preferred/intrinsic size in AtomLayout

lucasmerlin:
- part of https://github.com/emilk/egui/issues/5762
- also allows me to simplify sizing logic in egui_flex

### DIFF
--- a/crates/egui/src/atomics/atom_kind.rs
+++ b/crates/egui/src/atomics/atom_kind.rs
@@ -81,9 +81,21 @@ impl<'a> AtomKind<'a> {
     ) -> (Vec2, SizedAtomKind<'a>) {
         match self {
             AtomKind::Text(text) => {
-                let galley = text.into_galley(ui, wrap_mode, available_size.x, TextStyle::Button);
+                let wrap_mode = wrap_mode.unwrap_or(ui.wrap_mode());
+                let desired_size = matches!(wrap_mode, TextWrapMode::Truncate).then(|| {
+                    text.clone()
+                        .into_galley(
+                            ui,
+                            Some(TextWrapMode::Extend),
+                            available_size.x,
+                            TextStyle::Button,
+                        )
+                        .desired_size()
+                });
+                let galley =
+                    text.into_galley(ui, Some(wrap_mode), available_size.x, TextStyle::Button);
                 (
-                    galley.size(), // TODO(#5762): calculate the preferred size
+                    desired_size.unwrap_or_else(|| galley.desired_size()),
                     SizedAtomKind::Text(galley),
                 )
             }
--- a/crates/epaint/src/text/text_layout_types.rs
+++ b/crates/epaint/src/text/text_layout_types.rs
@@ -795,6 +795,31 @@ impl Galley {
         self.rect.size()
     }
 
+    // TODO: Instead return Option<Vec2>?
+    pub fn desired_size(&self) -> Vec2 {
+        let mut current_width: f32 = 0.0;
+        let mut widest_width: f32 = 0.0;
+        let mut height = self.rows.first().map_or(0.0, |row| row.height());
+        for row in &self.rows {
+            if current_width != 0.0 {
+                let space = row.glyphs.last();
+                if let Some(space) = space {
+                    if space.chr.is_whitespace() {
+                        // TODO: Needed or not? Doesn't seem like it's needed
+                        // current_width += space.advance_width;
+                    }
+                }
+            }
+            current_width += row.rect().width();
+            widest_width = widest_width.max(current_width);
+            if row.ends_with_newline {
+                height += row.height();
+                current_width = 0.0;
+            }
+        }
+        vec2(widest_width, height)
+    }
+
     pub(crate) fn round_output_to_gui(&mut self) {
         for placed_row in &mut self.rows {
             // Optimization: only call `make_mut` if necessary (can cause a deep clone)


crates/egui/src/atomics/atom_kind.rs
```diff
@@ -81,9 +81,21 @@ impl<'a> AtomKind<'a> {
...
+                let desired_size = matches!(wrap_mode, TextWrapMode::Truncate).then(|| {
+                    text.clone()
+                        .into_galley(
+                            ui,
+                            Some(TextWrapMode::Extend),
+                            available_size.x,
+                            TextStyle::Button,
+                        )
+                        .desired_size()
+                });
```
lucasmerlin: Should this be behind a flag, since it might have a hefty performance cost when there is long, truncated text?

crates/epaint/src/text/text_layout.rs
```diff
@@ -1140,4 +1140,39 @@ mod tests {
...
+            galley_wrapped.size().y > galley_unwrapped.size().y * 2.0,
+            "Text should be wrapped"
+        );
+
+        let desired_size = galley_wrapped.desired_size();
+        let unwrapped_size = galley_unwrapped.size();
+        assert!(
+            (desired_size - unwrapped_size).length() < 1.0,
+            "Wrapped desired size should almost match unwrapped size: {desired_size:?} vs {unwrapped_size:?}"
+        );
```
lucasmerlin: Somehow the desired_size width is a bit higher (0.8 px for this test, seems to be ~ 0.2 per wrapped row?). I couldn't figure out why, it doesn't seem to be the rounding, the error was the same when I commented the rounding code.

crates/egui/src/atomics/atom_kind.rs
bircni: jep definitely and also there should be comment mentioning this

crates/egui/src/atomics/atom_kind.rs
```diff
@@ -81,9 +81,21 @@ impl<'a> AtomKind<'a> {
...
+                let wrap_mode = wrap_mode.unwrap_or(ui.wrap_mode());
+                let desired_size = matches!(wrap_mode, TextWrapMode::Truncate).then(|| {
+                    text.clone()
+                        .into_galley(
+                            ui,
+                            Some(TextWrapMode::Extend),
+                            available_size.x,
+                            TextStyle::Button,
+                        )
+                        .desired_size()
```
emilk: We only need to run this extra layout if the truncated galley was actually truncated/ellided.

crates/egui/src/atomics/atom_kind.rs
emilk: Let's add a benchmark and measure the difference before we do anything else

crates/egui/src/atomics/atom_kind.rs
```diff
@@ -81,9 +81,21 @@ impl<'a> AtomKind<'a> {
     ) -> (Vec2, SizedAtomKind<'a>) {
         match self {
             AtomKind::Text(text) => {
-                let galley = text.into_galley(ui, wrap_mode, available_size.x, TextStyle::Button);
+                let wrap_mode = wrap_mode.unwrap_or(ui.wrap_mode());
+                let desired_size = matches!(wrap_mode, TextWrapMode::Truncate).then(|| {
```
emilk:
```suggestion
                let desired_size = (wrap_mode == TextWrapMode::Truncate).then(|| {
```

crates/egui/src/atomics/atom_kind.rs
```diff
@@ -81,9 +81,21 @@ impl<'a> AtomKind<'a> {
...
+                            available_size.x,
+                            TextStyle::Button,
+                        )
+                        .desired_size()
+                });
+                let galley =
+                    text.into_galley(ui, Some(wrap_mode), available_size.x, TextStyle::Button);
                 (
-                    galley.size(), // TODO(#5762): calculate the preferred size
+                    desired_size.unwrap_or_else(|| galley.desired_size()),
```
emilk: We should have a test for the `desired_size` measurements of a `Label` or `Atoms` with:

* No text wrapping
* Text wrapping
* Text truncation

crates/epaint/src/text/text_layout.rs
```diff
@@ -1140,4 +1140,39 @@ mod tests {
...
+    #[test]
+    fn test_desired_size() {
+        let mut fonts = FontsImpl::new(1.0, 1024, FontDefinitions::default());
+
+        let mut job = LayoutJob::simple(
+            "Some basic text with /n multiple lines.".to_owned(),
+            FontId::default(),
+            Color32::BLACK,
+            80.0,
+        );
```
emilk:
```suggestion
        let max_width = 80.0;
        let mut job = LayoutJob::simple(
            "Some basic text with /n multiple lines.".to_owned(),
            FontId::default(),
            Color32::BLACK,
            max_width,
        );
```

crates/epaint/src/text/text_layout.rs
```diff
@@ -1140,4 +1140,39 @@ mod tests {
         assert_eq!(row.pos, Pos2::ZERO);
         assert_eq!(row.rect().max.x, row.glyphs.last().unwrap().max_x());
     }
+
+    #[test]
+    fn test_desired_size() {
+        let mut fonts = FontsImpl::new(1.0, 1024, FontDefinitions::default());
+
+        let mut job = LayoutJob::simple(
+            "Some basic text with /n multiple lines.".to_owned(),
```
emilk: This is not a newline, but looks like it. Confusing!

crates/epaint/src/text/text_layout.rs
emilk: I think it is worth digging deeper into this. If the goal is to get accurate measurements :]

crates/epaint/src/text/text_layout.rs
```diff
@@ -1140,4 +1140,39 @@ mod tests {
...
+
+        let galley_wrapped = layout(&mut fonts, job.clone().into());
+
+        job.wrap = TextWrapping::no_max_width();
+
+        let galley_unwrapped = layout(&mut fonts, job.into());
+
+        assert!(
+            galley_wrapped.size().y > galley_unwrapped.size().y * 2.0,
+            "Text should be wrapped"
```
emilk:
```suggestion
            "Text should be wrapped at least three times"
```

crates/epaint/src/text/text_layout_types.rs
```diff
@@ -795,6 +795,28 @@ impl Galley {
         self.rect.size()
     }
 
+    /// Calculate the desired size.
+    ///
+    /// This is the size that a non-wrapped, non-truncated, non-justified version of the text
+    /// would have.
+    ///
+    /// This won't work for truncated text. Instead, you can clone the `LayoutJob` and set `WrapMode`
+    /// to `TextWrapMode::Extend` to calculate the desired size.
```
emilk: Then let's add a `debug_assert` that `self` isn't truncated, or better yet: return an `Option` with `None` meaning "This was truncated, can't know the desired_size"

crates/epaint/src/text/text_layout_types.rs
```diff
@@ -795,6 +795,28 @@ impl Galley {
...
+        let mut current_width: f32 = 0.0;
+        let mut widest_width: f32 = 0.0;
+        let mut height = self.rows.first().map_or(0.0, |row| row.height());
+        for row in &self.rows {
+            current_width += row.rect().width();
+            widest_width = widest_width.max(current_width);
+            if row.ends_with_newline {
+                height += row.height();
+                current_width = 0.0;
+            }
```
emilk: I wonder if the trailing space character is counted as part of the row width 🤔  Probably worth checking.

We also have `pair_kerning` between glyphs that this misses. Though I think `pair_kerning` around a space is always zero… but again, worth checking.

(this is me trying to figure out why the test fails)

lucasmerlin:
I've changed the approach and now calculate the desired_size as the galley is laid out. This should improve performance (since the calculation is now cached) and makes it more accurate by having access to the Paragraphs. 
Another benefit is that this will also work for truncated text.

crates/epaint/src/text/text_layout.rs
```diff
@@ -190,6 +193,42 @@ fn layout_section(
     }
 }
 
+fn calculate_desired_size(
```
emilk: Worth explaining again what this is in a docstring

crates/epaint/src/text/text_layout.rs
```diff
@@ -190,6 +193,42 @@ fn layout_section(
...
+    point_scale: PointScale,
+    job: &LayoutJob,
+    paragraphs: &[Paragraph],
+) -> Vec2 {
+    let mut desired_size = Vec2::ZERO;
+    for (idx, paragraph) in paragraphs.iter().enumerate() {
+        if paragraph.glyphs.is_empty() {
+            if idx == 0 {
+                desired_size.y += point_scale.round_to_pixel(paragraph.empty_paragraph_height);
+            }
```
emilk: Why only if `idx == 0` ?

lucasmerlin:
I ran the demo and button benchmarks, the change was "within noise threshold" as criterion likes to say.

crates/epaint/src/text/text_layout.rs
emilk: Did you mean to (also) use `first_row_min_height` here?

crates/egui/src/atomics/atom.rs
```diff
@@ -89,7 +89,7 @@ impl<'a> Atom<'a> {
 
         SizedAtom {
             size,
-            preferred_size: preferred,
+            preferred_size: preferred.at_least(size),
```
emilk: Is this correct even for atoms marked as `grow`? can they ever shrink then?

emilk (approved):

crates/egui/src/atomics/atom.rs
lucasmerlin: Good catch, it was supposed to be at_lest(self.size)

crates/epaint/src/text/text_layout.rs
lucasmerlin: I did that first but got wrong results. This matches what happens in `rows_from_paragraphs`. 

In this if
```rs
        if paragraph.glyphs.is_empty() {
            rows.push(PlacedRow {
                pos: Pos2::ZERO,
                row: Arc::new(Row {
                    section_index_at_start: paragraph.section_index_at_start,
                    glyphs: vec![],
                    visuals: Default::default(),
                    size: vec2(0.0, paragraph.empty_paragraph_height),
                    ends_with_newline: !is_last_paragraph,
                }),
            });
        } else {
```
the empty row is always placed at (0, 0), so it only has an effect on the total rect if it's the first and only row.

This seems like a bug, I'll try to fix it in a separate PR.  `first_row_min_height` is also not used in the empty case (which is probably also a bug).

If I remove the if, the test_split_paragraph test fails since it causes a empty Paragraph to be added at the end of the first split Galley.

lucasmerlin:
I renamed the desired and preferred size names I had in AtomLayout and in this PR to intrinsic size since I think this is the best name (we use desired size a lot in other places and preferred size is very similar to desired size) and for consistency with response.intrinsic_size.

lucasmerlin:
Also made a small test ui to show that this works:

https://github.com/user-attachments/assets/86fcf843-ab3a-4e9a-a8e5-3e75c65408ab

```rs
    egui::Window::new("Test").show(ctx, |ui| {
            ui.set_min_size(ui.available_size());
            for wrapping in [
                // TextWrapMode::Extend,
                TextWrapMode::Wrap,
                TextWrapMode::Truncate,
            ] {
                ui.style_mut().wrap_mode = Some(wrapping);
                let response = ui.button(format!(
                    "Hello world this is a long text that should {wrapping:?}"
                ));

                ui.ctx().debug_painter().debug_rect(
                    Rect::from_min_size(
                        response.rect.min,
                        response.intrinsic_size.unwrap_or_default(),
                    ),
                    Color32::RED,
                    "",
                );
            }
        });

```

crates/epaint/src/text/text_layout.rs
```diff
@@ -190,6 +193,46 @@ fn layout_section(
...
+        );
+
+        let mut height = paragraph
+            .glyphs
+            .iter()
+            .map(|g| g.line_height)
+            .max_by(|a, b| a.partial_cmp(b).unwrap_or(std::cmp::Ordering::Equal))
+            .unwrap_or(paragraph.empty_paragraph_height);
+        if idx == 0 {
+            height = f32::max(height, job.first_row_min_height);
```
emilk, line 229: We should add a test for this in https://github.com/emilk/egui/blob/f46926aaf1ea7ab5a3756c678a1d647b9e549bd6/crates/epaint/src/text/fonts.rs#L1079

crates/epaint/src/text/text_layout.rs
```diff
@@ -190,6 +193,46 @@ fn layout_section(
...
+    paragraphs: &[Paragraph],
+) -> Vec2 {
+    let mut intrinsic_size = Vec2::ZERO;
+    for (idx, paragraph) in paragraphs.iter().enumerate() {
+        if paragraph.glyphs.is_empty() {
+            if idx == 0 {
+                intrinsic_size.y += point_scale.round_to_pixel(paragraph.empty_paragraph_height);
+            }
+            continue;
+        }
```
emilk, line 212: Is this `if`-statement even needed? Seems like the logic below should be robust to handling this regardless? In fact, more robust, since the code below takes `first_row_min_height` into account

tests/egui_tests/tests/test_atoms.rs
```diff
@@ -69,3 +69,31 @@ fn single_test(name: &str, mut f: impl FnMut(&mut Ui)) -> SnapshotResult {
...
+                    "Hello world this is a long text that should be wrapped.",
+                ));
+                if let Some(current_intrinsic_size) = intrinsic_size {
+                    assert_eq!(
+                        Some(current_intrinsic_size),
+                        response.intrinsic_size,
+                        "For wrapping: {wrapping:?}"
+                    );
+                }
+                intrinsic_size = response.intrinsic_size;
```
emilk, line 99:
```suggestion
                assert!(response.intrinsic_size.is_some(), "intrinsic_size should be set for `Button`");
                intrinsic_size = response.intrinsic_size;
```

crates/epaint/src/text/text_layout.rs
lucasmerlin: Yes, otherwise the test_split_paragraphs test fails. (I explained why in this comment in case you didn't see it https://github.com/emilk/egui/pull/7146#discussion_r2192869384)

I'll try to fix this in a separate PR.

crates/epaint/src/text/text_layout.rs
emilk: I saw your comment in https://github.com/emilk/egui/pull/7146#discussion_r2192869384 - let's do that in a follow-up PR then :)

crates/epaint/src/text/text_layout.rs
emilk: Nice, thanks!

Issue #7060, opened 2025-05-19
egui::Scene in scene mouse position


Stretchyfish:
It would be great if the new egui::Scene had a way of getting the position of the mouse inside of the Scene, which accounts for pan & zoom. To for example be able to draw a sphere at the mouse position, or find the mouse delta relative to the scenes pan & zoom. There currently exists some work around to get the mouse position inside of the scene in the previous frame using the scenes response, but a more direct implementation would be great.

A simple idea for how this could look can be seen below.
~~~
let mouse_position: Option<egui::Pos2>;
let scene_response= egui::Scene::new()
.zoom_range(0.01..=2.0)
.mouse_position_in_scene(mouse_position)
.show(ui, &mut self.scene_rect, |scene_ui|
{

}
~~~

PR #7149, opened 2025-06-15
Modified scene.show to calculate the mouse position inside the scene, and created an example scene_circle_painter to demonstrate a usecase

Stretchyfish:
## Features/changes added
I have modified egui/crates/egui/src/containers/scene.rs, to now also calculate the mouse position inside the scene, dependent on both translation and scale. This has been accomplished by having scene.show also have a mutable borrow of a mouse_position as input, and using the to_global variable inside the show function to calculate the scene mouse position.

Along with this, I have created an example called scene_circle_placer to demonstrate a use case, and have fixed errors that appeared in crates/egui_demo_lib/src/demo/scene.rs due to the change.

## Reason
Currently, there is no intuitive way of getting the position of the mouse inside of the scene. It has been possible for me to find the position using the scene response, but the position is not correct when hovering over other interactive elements, such as buttons, making it difficult to create for example dragging behavior of drawn elements. 

## Issues fixed with this PR
https://github.com/emilk/egui/issues/7060

## Note
Due to the nature of the mouse position, I have not been able to find a way to avoid using unwrap, however, I have been very careful not to create any unaccounted situations. 
Cargo fmt and cargo clippy has been run successfully.
The command UPDATE_SNAPSHOTS=true cargo test --workspace --all-features was run successfully, and this was the output from ./scripts/check.sh

~~~
+ export 'RUSTFLAGS=-D warnings'
+ RUSTFLAGS='-D warnings'
+ export 'RUSTDOCFLAGS=-D warnings'
+ RUSTDOCFLAGS='-D warnings'
+ typos
./scripts/check.sh: line 18: typos: command not found
~~~

### DIFF
--- a/examples/scene_circle_placer/src/main.rs
+++ b/examples/scene_circle_placer/src/main.rs
@@ -18,14 +18,17 @@ fn main() -> eframe::Result {
 
 struct MyApp {
     scene_rect: egui::Rect,
-    circles_in_scene: Vec<egui::Pos2>
+    circles_in_scene: Vec<egui::Pos2>,
 }
 
 impl Default for MyApp {
     fn default() -> Self {
-        Self { 
-            scene_rect: egui::Rect::from_min_size( egui::Pos2 { x: 0.0, y: 0.0 }, egui::Vec2 { x: 400.0, y: 400.0 }), 
-            circles_in_scene: Vec::new()
+        Self {
+            scene_rect: egui::Rect::from_min_size(
+                egui::Pos2 { x: 0.0, y: 0.0 },
+                egui::Vec2 { x: 400.0, y: 400.0 },
+            ),
+            circles_in_scene: Vec::new(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


examples/scene_circle_placer/src/main.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
+        egui::CentralPanel::default().show(ctx, |ui| {
+            let mut mouse_position_in_scene: Option<egui::Pos2> = Option::None;
+
+            egui::Scene::new().show(
+                ui,
+                &mut self.scene_rect,
+                &mut mouse_position_in_scene,
+                |ui| {
+                    for circle_pos in &self.circles_in_scene {
```
emilk, line 46: Instead of changing the `Scene` API, you can get the mouse coordinate with:

```rs
mouse_position_in_scene =
    ctx.layer_transform_from_global(ui.painter().layer_id()).unwrap_or_default() *
    ui.input(|i| i.pointer.latest_pos());
```

If you add a couple of helpers to `Painter` this could be shortend to


```rs
mouse_position_in_scene =
    ui.painter().layer_transform_from_global() * ui.input(|i| i.pointer.latest_pos());
```

Stretchyfish:
Very understandable, thanks a lot for the code example. Will close the PR for now then, but if you have an idea for how to otherwise approach it and could use a hand with that, feel free to let me know.

PR #7151, opened 2025-06-16
Change API of `Tooltip` slightly

emilk:
We try to be consistent with our parameter order to reduce surprise for users.

I also renamed a few things to clarify what is what

### DIFF
--- a/crates/egui/src/containers/tooltip.rs
+++ b/crates/egui/src/containers/tooltip.rs
@@ -7,26 +7,31 @@ use emath::Vec2;
 
 pub struct Tooltip<'a> {
     pub popup: Popup<'a>,
-    layer_id: LayerId,
-    widget_id: Id,
+
+    /// The layer of the parent widget.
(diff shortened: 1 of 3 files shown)


crates/egui/src/containers/tooltip.rs
```diff
@@ -7,26 +7,31 @@ use emath::Vec2;
...
-        layer_id: LayerId,
     ) -> Self {
+        let width = ctx.style().spacing.tooltip_width;
         Self {
-            // TODO(lucasmerlin): Set width somehow (we're missing context here)
-            popup: Popup::new(widget_id, ctx, anchor.into(), layer_id)
+            popup: Popup::new(parent_widget, ctx, anchor.into(), parent_layer)
                 .kind(PopupKind::Tooltip)
                 .gap(4.0)
+                .width(width)
```
emilk, line 31: Oh yeah; and this was just a bug fix

PR #7251, opened 2025-06-17
Add `.auto_scroll()` to `TextEdit`

TetraTheta:
If anyone are interested in this, feel free to take over this, because I've lost capability of continuing this :(

***

This PR adds `TextEdit.auto_scroll(b: bool) -> Self` to `TextEdit`.

It only works for `TextEdit::singleline()`.

By default, the corresponding field of `TextEdit`, `auto_scroll` is set to `false` to retain original behavior when `.auto_scroll()` is not used.

When `.auto_scroll(true)` is called, the `TextEdit` will move its viewport to where its own cursor is located at.
When `.auto_scroll(false)`, which is default, the `TextEdit` will move its viewport to where its own cursor is located at, only when it has focus.

https://github.com/user-attachments/assets/2a1e4e58-2d02-4933-bd87-4db19dbff5f6

In this example video, the `TextEdit` next to 'New Folder at:' is using `.auto_scroll()`.

<ins>Please let me know if I missed something!</ins>


TetraTheta:
Using `.auto_scroll(true)` alone doesn't seem to work. Manually setting the cursor location is required like this:
```rust
// ...
struct AppState {
  // ...
  autofocused: bool,
}
// ...
fn remove_unc(s: String) -> String {
  // ...
}
// ...
impl App for AppState {
  fn update(&mut self, ctx: &egui::Context, _frame: &mut Frame) {
    // ...
      let mut from = remove_unc(self.target.clone());

      ui.horizontal(|ui| {
        let mut from_edit = TextEdit::singleline(&mut from).desired_width(f32::INFINITY).auto_scroll(true).show(ui);
        let resp = from_edit.response;
        if !self.autofocused {
          from_edit
            .state
            .cursor
            .set_char_range(Some(CCursorRange::two(CCursor::new(from.len()), CCursor::new(from.len()))));
          // do not set `autofocused` to `true` yet
          // i need to do similar things to other TextEdit too
          // `autofocused` must be set to `true` in last TextEdit cursor modification
          from_edit.state.store(ui.ctx(), resp.id);
        }
      });
      // ...
    }
    // ...
  }
}
```
If manually setting cursor selection is required, doesn't that mean [`TextEdit.cursor_at_end(self, b: bool) -> Self`](https://docs.rs/egui/latest/egui/widgets/text_edit/struct.TextEdit.html#method.cursor_at_end) doesn't work?

I'm not sure about this...

TetraTheta:
I don't know about how to resolve the failed check :(

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -341,6 +343,18 @@ impl<'t> TextEdit<'t> {
         self
     }
 
+    /// When `true`, move the viewport to the location of the cursor.
+    /// When `false`, do nothing about viewport.
+    ///
+    /// This only works for singleline [`TextEdit`].
+    #[inline]
+    pub fn auto_scroll(mut self, b: bool) -> Self {
```
emilk, line 351: From the video, it looks like this only affects a `TextEdit` that does not have focus - is that correct?

TetraTheta: Yes, you're right.

By default (or, without this PR), `TextEdit` resets its viewport to the beginning of the text of it, when the focus is lost.

My goal is to preserve viewport to wherever the previous viewport was at, rather than resetting it whenever the focus is lost.

***

I thought viewport would follow the cursor, but it seems cursor and viewport are separate thing, because `.cursor_at_end()` isn't enough for setting the viewport to the end of the text of `TextEdit`.

emilk: So when we have focus we always "scroll" (*) to the cursor. I think anything else would be weird.

But when text edit doesn't have focus we have a choice of what to show, including the options:

A) Show the start of the text
B) Show the end of the text
C) Show the text where the cursor would be, if given focus
D) something else (e.g. animate)

The best choice here is less obvious. On `main`, A) is the only behavior, and your PR proposes to make this configurable. I think an `enum` would be a better choice, something like:


```
/// What do we show when the `TextEdit` doesn't have focus, and the text doesn't fit the `TextEdit`?
pub fn unfocused_visibility(self, visibility: TextEditVisibility)
```

---

(*) "scroll" is a bit confusing, because this has nothing to do with `ScrollArea`… but I don't have a better word.

TetraTheta: About the 'scroll', I think the closest function name would be `unfocused_viewport`. [Viewport](https://en.wikipedia.org/wiki/Viewport) would be most appropriate term about what I want to achieve.

I'll mark this PR back to draft, and try to implement `TextEditViewport` which would have these values:

* `Start`: Current behavior. When not focused, the viewport will be moved to the start of the text.
* `Cursor`: When not focused, the viewport will be moved to where the cursor is at. The cursor will be placed at the end of the viewport.
* `End`: When not focused, the viewport will be moved to the end of the text.

I think this would be enough. Placing the cursor at the start of the viewport doesn't look *intuitive* to me.

I also have to find out why current PR code doesn't respect `cursor_at_end(true)`.

Issue #7155, opened 2025-06-16
Basic egui stripped down example breaks


Angramme:
**Describe the bug**
$ RUST_BACKTRACE=1 cargo run -p cc-clicker                                                            
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/cc-clicker`
[1]    73615 bus error  RUST_BACKTRACE=1 cargo run -p cc-clicker

**To Reproduce**
run the code with cargo run

main.rs
```rust
#![cfg_attr(not(debug_assertions), windows_subsystem = "windows")] // hide console window on Windows in release
#![allow(rustdoc::missing_crate_level_docs)] // it's an example

use eframe::egui;

fn main() -> eframe::Result {
    env_logger::init(); // Log to stderr (if you run with `RUST_LOG=debug`).
    let options = eframe::NativeOptions {
        viewport: egui::ViewportBuilder::default().with_inner_size([320.0, 240.0]),
        ..Default::default()
    };
    eframe::run_native(
        "My egui App",
        options,
        Box::new(|_cc| {
            Ok(Box::<MyApp>::default())
        }),
    )
}

struct MyApp {

}

impl Default for MyApp {
    fn default() -> Self {
        Self {
        }
    }
}

impl eframe::App for MyApp {
    fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
        egui::CentralPanel::default().show(ctx, |ui| {
            ui.heading("My egui Application");
        });
    }
}
```

Cargo.toml
```toml
[package]
name = "cc-clicker"
version = "0.1.0"
edition = "2021"

[dependencies]
anyhow = { version = "1.0.98", features = ["std"] }
eframe = { git = "https://github.com/emilk/egui", branch = "main", features = ["default", "persistence", "wgpu"] }
egui = { git = "https://github.com/emilk/egui", branch = "main" }
egui_extras = { version = "0.31.1", features = ["default"] }
egui_table = "0.3.0"
env_logger = "0.11.8"
serde = "1.0.219"
```

**Expected behavior**
Not crashing

**Desktop (please complete the following information):**
 - OS: Apple M1 Max  -- MacOS 15.5 (24F74)
 - Browser NA
 - Version master branch

Angramme:
I tried debugging myself, however when running with VSCode debugger (which I believe uses LLDB), the issue does not occur and the window opens correctly. However running using cargo run causes the issue consistently.

[...]

PR #7252, opened 2025-06-17
Workaround libpng crash on macos by not creating `NSImage` from png data

Wumpf:
* Fixes https://github.com/emilk/egui/issues/7155

DRAFT TODO: Confirm that this solves the problem, then write nice description

### DIFF
--- a/crates/eframe/src/native/app_icon.rs
+++ b/crates/eframe/src/native/app_icon.rs
@@ -205,13 +205,18 @@ fn set_title_and_icon_mac(title: &str, icon_data: Option<&IconData>) -> AppIconS
 
     use objc2::ClassType as _;
     use objc2_app_kit::{NSApplication, NSImage};
-    use objc2_foundation::{NSData, NSString};
+    use objc2_foundation::NSString;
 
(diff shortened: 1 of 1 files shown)


lucasmerlin:
Should we decide to merge this, we have to remember to also handle the default icon differently (by including it as a raw image or using the image crate to decode it?)

Wumpf:
If I'm not reading this wrong we so far decoded the default icon and then encoded it again...
I'd rather keep the default icon encoded just out of size consideration. Even better though would be to generate it from code: I mean it's a lil rounded rectangle with an 'e' in it 😆 
Counter point: this could mean that we don't have to depend on the `png` feature of `image` 💡

Wumpf:
Fix got confirmed! 🥳

lucasmerlin:
Not related to your PR, but I noticed that the icon will switch to the terminal icon for a split second when closing the app 🤔

https://github.com/user-attachments/assets/e47435b2-0b6d-4e0f-a38f-2583e029ec61

Not a huge problem but it's a bit odd

PR #7254, opened 2025-06-19
Drag-to-close panels

emilk:
You can now drag panels to close them

## TODO
* [ ] Avoid breaking the API by introducing new functions?


lucasmerlin:
It feels a bit disruptive if you just want to make the panel smaller and then it gets "ripped out of your hand". I think to rememdy this it'd be nice if it'd open again when you keep dragging and drag it back over the threshold?

emilk:
Yeah, that's a nice idea

emilk:
* Closed in favor of  #8182

PR #7266, opened 2025-06-26
Add `avif` feature to `egui_extras`

CodedNil:
This simple PR adds support for using avif images with the image/avif feature.


emilk:
Thanks, but this is not necessary. Users can opt-in to `image` features themselves without having to go through `egui_extras` (I don't want to add every `image` feature to `egui_extras`)

Issue #7264, opened 2025-06-24
Finish up atomics work


emilk:
Let's just do these for the 0.32 reelase if they are simple to do. In particular: atomics doesn't support wrapping (yet), so if any of the below depends on wrapping, then we should punt on it until egui 0.33

* ComboBox label
* ui.small_button
* `SelectableLabel` (can actually be removed!)
  * (ui.toggle_value, ui.toggle_value)
* `CollapsingHeader` (ui.collapsing) 
* Deprecate `ui.menu_image_button` and `ui.menu_image_text_button`
* `show_tooltip_text`
* `Window::title` - maybe doable?
* `Link` would be nice - but maybe underlining is hard? And wrapping 😬  Maybe not then.

PR #7277, opened 2025-06-30
Remove `SelectableLabel`

lucasmerlin:
* part of https://github.com/emilk/egui/issues/7264
* removes SelectableLabel (Use `Button::selectable` instead)
* updates `Ui::selectable_value/label` with IntoAtoms support

Had to make some changes to `Button` since the SelecatbleLabel had no frame unless selected.


crates/egui/src/widgets/button.rs
```diff
@@ -262,6 +298,12 @@ impl<'a> Button<'a> {
         let response = if ui.is_rect_visible(prepared.response.rect) {
             let visuals = ui.style().interact_selectable(&prepared.response, selected);
 
+            if !frame_when_inactive {
```
emilk:
```suggestion
            let visible_frame = if frame_when_inactive {
                has_frame
            } else {
```

crates/egui/src/widgets/button.rs
```diff
@@ -243,7 +279,7 @@ impl<'a> Button<'a> {
 
         let text = layout.text().map(String::from);
 
-        let has_frame = frame.unwrap_or_else(|| ui.visuals().button_frame);
+        let mut has_frame = frame.unwrap_or_else(|| ui.visuals().button_frame);
```
emilk:
```suggestion
        let has_frame_margin = frame.unwrap_or_else(|| ui.visuals().button_frame);
```

crates/egui/src/widgets/selected_label.rs
emilk, line 1:
```rs
#[deprecated = "SelectableLabel has been removed. Use Button::selectable() instead"]
struct SelectableLabel {}
```

emilk (approved):

PR #7279, opened 2025-06-30
Update MSRV to 1.85

emilk:


reneleonhardt:
@emilk 1.85 is the baseline for edition 2024, are there plans to upgrade?

lol it came later, sorry for the noise 😅

PR #7281, opened 2025-07-01
Workaround top area not clickable on MacOS fullscreen

lucasmerlin:
* attempt to work around https://github.com/rerun-io/rerun/issues/10280 and https://github.com/rust-windowing/winit/issues/4295

Unfortunately this only works when the accesskit feature is disabled, since accesskit requires the window to be hidden on first frame.

To test: `cargo run -p egui_demo_app --no-default-features --features glow` should work and `cargo run -p egui_demo_app` should not work

### DIFF
--- a/crates/eframe/src/native/wgpu_integration.rs
+++ b/crates/eframe/src/native/wgpu_integration.rs
@@ -911,7 +911,7 @@ fn create_window(
         native_options,
         window_settings,
     )
-    .with_visible(should_start_visible()); // Start hidden until we render the first frame to fix white flash on startup (https://github.com/emilk/egui/pull/3631)
+    .with_visible(should_start_visible());
 
     let window = egui_winit::create_window(egui_ctx, event_loop, &viewport_builder)?;
     epi_integration::apply_window_settings(&window, window_settings);
--- a/crates/egui-winit/src/window_settings.rs
+++ b/crates/egui-winit/src/window_settings.rs
@@ -60,8 +60,9 @@ impl WindowSettings {
         mut viewport_builder: ViewportBuilder,
     ) -> ViewportBuilder {
         profiling::function_scope!();
-        // On MacOS with an external monitor, if the window starts fullscreen,
-        // the top area is not clickable. See https://github.com/rust-windowing/winit/issues/4295
+        // On macOS with an external monitor, if the window starts fullscreen,
+        // the top area is not clickable. So we don't restore settings if this is the case.
+        // See https://github.com/rust-windowing/winit/issues/4295
         let top_area_not_clickable_workaround =
             cfg!(target_os = "macos") && !cfg!(feature = "accesskit");
         if top_area_not_clickable_workaround && self.fullscreen {


crates/eframe/src/native/wgpu_integration.rs
```diff
@@ -913,6 +913,11 @@ fn create_window(
     )
     .with_visible(false); // Start hidden until we render the first frame to fix white flash on startup (https://github.com/emilk/egui/pull/3631)
 
+    // Fix the top window area not being clickable on macOS when in fullscreen on an external display
+    // See https://github.com/rust-windowing/winit/issues/4295, https://github.com/rerun-io/rerun/issues/10280
+    #[cfg(all(target_os = "macos", not(feature = "accesskit")))]
+    let viewport_builder = viewport_builder.with_visible(true);
```
emilk: Looks like this will cause a white flash on Windows again:
* https://github.com/emilk/egui/pull/3631

lucasmerlin: It shouldn't, unless my cfg statement is wrong 😬

emilk: Oh yeah 🤦

emilk (approved):
Let's check that this actually fixes the Rerun issue before we merge this

crates/eframe/src/native/glow_integration.rs
```diff
@@ -157,6 +157,11 @@ impl<'app> GlowWinitApp<'app> {
         )
         .with_visible(false); // Start hidden until we render the first frame to fix white flash on startup (https://github.com/emilk/egui/pull/3631)
 
+        // Fix the top window area not being clickable on macOS when in fullscreen on an external display
+        // See https://github.com/rust-windowing/winit/issues/4295, https://github.com/rerun-io/rerun/issues/10280
+        #[cfg(all(target_os = "macos", not(feature = "accesskit")))]
+        let winit_window_builder = winit_window_builder.with_visible(true);
```
emilk: I tend to prefer the `cfg!` macro, because I find it leads to easier-to-read code:

crates/eframe/src/native/wgpu_integration.rs
emilk: We can make a reusable function for this that better explain the if-and-why of it:

```res
fn start_visible() {
	if cfg!(feature = "accesskit") {
	    true // accesskit requires this: $LINK
	 } else if cfg!(windows) {
	     true // Fixes white flash on startup (https://github.com/emilk/egui/pull/3631)
	 } else if cfg!(target_os = "macos") {
	      // Fix the top window area not being clickable on macOS when in fullscreen on an external display
	      // See https://github.com/rust-windowing/winit/issues/4295, https://github.com/rerun-io/rerun/issues/10280
	     false
	 } else {
	     true // no known preference one way or the other
	 }
}
```

crates/eframe/src/native/wgpu_integration.rs
```diff
@@ -911,7 +911,7 @@ fn create_window(
         native_options,
         window_settings,
     )
-    .with_visible(false); // Start hidden until we render the first frame to fix white flash on startup (https://github.com/emilk/egui/pull/3631)
+    .with_visible(should_start_visible()); // Start hidden until we render the first frame to fix white flash on startup (https://github.com/emilk/egui/pull/3631)
```
emilk:
```suggestion
    .with_visible(should_start_visible())
```

lucasmerlin:
Unfortunately I had to add another workaround, when the app starts Fullscreen the bug also happens. So I'm now only restoring the windowsettings if it's not fullscreen or we're not on MacOS (and the accesskit feature is not enabled).

Not sure if the fix is still worth it with this extra complexity 😅.

Issue #7262, opened 2025-06-24
Add `Visuals::weak_color`


emilk:
We should be able to control the weak text color explicitly

PR #7285, opened 2025-07-01
Add `Visuals::weak_text_alpha` and `weak_text_color`

emilk:
* Closes https://github.com/emilk/egui/issues/7262

This also makes the default weak color slightly less weak in most cases.


tests/egui_tests/tests/snapshots/visuals/button_image_shortcut_selected.png
lucasmerlin, line 1: Nice, this is much more readable now!

lucasmerlin (approved):

crates/egui/src/style.rs
```diff
@@ -2055,6 +2069,8 @@ impl Visuals {
         let Self {
             dark_mode,
             override_text_color: _,
+            weak_text_alpha,
+            weak_text_color: _,
```
lucasmerlin: It'd be nice if you could modify these in the ui, any reason that they are ignored?

emilk: laziness

PR #7286, opened 2025-07-01
Add support for scrolling via accesskit / kittest

lucasmerlin:
I need to scroll in a snapshot test in my app, and kittest had no utilities for this. Event::MouseWheel is error prone. This adds support for some accesskit scroll actions, and uses this in kittest to add helpers to scroll to a node / scroll the scroll area surrounding a node.

The accesskit code says down/up/left/right `Scrolls by approximately one screen in a specific direction.`. Unfortunately it's difficult to get the size of a "screen" (I guess that would be the size of the containing scroll area)where I implemented the scrolling, so for now I've hardcoded it to 100px. I think scrolling a fixed amount is still better than not scrolling at all.


crates/egui/src/context.rs
```diff
@@ -1216,6 +1216,47 @@ impl Context {
...
+                        accesskit::Action::ScrollUp => {
+                            viewport.this_pass.scroll_delta.0 += vec2(0.0, 100.0);
+                        }
+                        accesskit::Action::ScrollLeft => {
+                            viewport.this_pass.scroll_delta.0 += vec2(-100.0, 0.0);
+                        }
+                        accesskit::Action::ScrollRight => {
+                            viewport.this_pass.scroll_delta.0 += vec2(100.0, 0.0);
+                        }
+                        _ => return false,
```
```suggestion
                    const DISTANCE: f32 = 100.0; // TODO(@lucasmerlin): this should be one "screen" according to AccessKit docs: $LINK
                    
                    match &request.action {
                        accesskit::Action::ScrollIntoView => {
                            viewport.this_pass.scroll_target = [
                                Some(ScrollTarget::new(
                                    res.rect.x_range(),
                                    Some(Align::Center),
                                    ScrollAnimation::none(),
                                )),
                                Some(ScrollTarget::new(
                                    res.rect.y_range(),
                                    Some(Align::Center),
                                    ScrollAnimation::none(),
                                )),
                            ];
                        }
                        accesskit::Action::ScrollDown => {
                            viewport.this_pass.scroll_delta.0 += DISTANCE * Vec2::DOWN;
                        }
                        accesskit::Action::ScrollUp => {
                            viewport.this_pass.scroll_delta.0 += DISTANCE * Vec2::UP;
                        }
                        accesskit::Action::ScrollLeft => {
                            viewport.this_pass.scroll_delta.0 += DISTANCE * Vec2::LEFT;
                        }
                        accesskit::Action::ScrollRight => {
                            viewport.this_pass.scroll_delta.0 += DISTANCE * Vec2::RIGHT;
                        }
                        _ => return false,
```

crates/egui_kittest/src/lib.rs
```diff
@@ -94,8 +99,11 @@ impl<'a, State> Harness<'a, State> {
         } = builder;
         let ctx = ctx.unwrap_or_default();
         ctx.enable_accesskit();
-        // Disable cursor blinking so it doesn't interfere with snapshots
-        ctx.all_styles_mut(|style| style.visuals.text_cursor.blink = false);
+        ctx.all_styles_mut(|style| {
+            // Disable cursor blinking so it doesn't interfere with snapshots
+            style.visuals.text_cursor.blink = false;
+            style.scroll_animation = ScrollAnimation::none();
```
emilk, line 105: Maybe set animation_time to zero while we're at it?

lucasmerlin: Good idea!

crates/egui/src/context.rs
lucasmerlin: Turns out `Action::ScrollDown` needs `Vec2::UP` 💀 scroll_delta is basically inverted

PR #7290, opened 2025-07-02
Improve text rendering in light mode

emilk:
This changes how we convert glyph coverage to alpha (and ultimately a color), but only in light mode.

This is a bit of a hack, because it doesn't fix dark-on-light text in _dark mode_ (if you have any), but for the common case this PR is a huge improvement.

You can also tweak this yourself now using `Visuals::text_alpha_from_coverage` or from the UI (bottom of the image):

image

## Before / After
widget_gallery_light_x1 widget_gallery_light_x1

## Black text Before/after
If you think the text above looks too weak, it's only because of the default text color. Here's how it looks like with perfectly `#000000` black text:

image Screenshot 2025-07-02 at 13 59 30


crates/epaint/src/image.rs
```diff
@@ -354,30 +407,25 @@ impl FontImage {
...
     #[inline]
-    pub fn srgba_pixels(&self, gamma: Option<f32>) -> impl ExactSizeIterator<Item = Color32> + '_ {
+    pub fn srgba_pixels(
+        &self,
+        alpha_from_coverage: AlphaFromCoverage,
+    ) -> impl ExactSizeIterator<Item = Color32> + '_ {
         // This whole function is less than rigorous.
         // Ideally we should do this in a shader instead, and use different computations
         // for different text colors.
         // See https://hikogui.org/2022/10/24/the-trouble-with-anti-aliasing.html for an in-depth analysis.
```
Wumpf: nit: redudnant comment to `AlphaFromCoverage` docstring

crates/epaint/src/image.rs
```diff
@@ -318,6 +318,59 @@ impl std::fmt::Debug for ColorImage {
...
+
+    /// `alpha = 2 * coverage - coverage^2`
+    ///
+    /// This looks good for white-on-black text, i.e. dark mode.
+    ///
+    /// Very similar to a gamma of 0.5, but produces sharper text.
+    /// See <https://www.desmos.com/calculator/w0ndf5blmn> for a comparison to gamma=0.5.
+    #[default]
+    TwoCoverageMinusCoverageSq,
+}
```
Wumpf, line 348: love how clearly defined this is now here!

Wumpf (approved):
If nothing else it's now a lot easier to understand what's happening and why, nice!!

Tested the looks on low-dpi SDR Windows Firefox: I found the default settings a tad bit too light, empirically a gamma of 0.9 looked a lil bit better to me. But this is starting to get arbitrary so if you think linear looks good I'm very much in favor with sticking with that
Either way I also believe this to be an improvement compared the harsh and fringy looking boundaries of `TwoCoverageMinusCoverageSq`

PR #7291, opened 2025-07-02
Free textures after submitting queue instead of before with wgpu renderer on Web

Wumpf:
Exactly the same fix as
* #5226

but for the web painter.
Curiously, this only surfaced now and only on WebGL - I believe this might fix
* https://github.com/rerun-io/rerun/issues/10439

but still need to confirm.

Likely this wasn't an issue before because there were some helpful bugs (;)) in wgpu's WebGL impl before. But I do wonder why this wouldn't also be a problem on WebGPU

### DIFF
--- a/crates/eframe/src/web/web_painter_wgpu.rs
+++ b/crates/eframe/src/web/web_painter_wgpu.rs
@@ -279,13 +279,6 @@ impl WebPainter for WebPainterWgpu {
             Some((output_frame, capture_buffer))
         };
 
-        {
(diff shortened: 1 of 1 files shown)


lucasmerlin (approved):
Awesome! 

Since this is the second time we fix this bug, are there any other places that might be affected? e.g. the glow backend?

Wumpf:
Glow doesn't have the equivalent destroy & queue poisoning logic that wgpu/WebGPU has, so I think we covered everything now

PR #7295, opened 2025-07-03
Add `egui::Sides` `shrink_left` / `shrink_right`

lucasmerlin:
This allows contents (on one of the sides) in egui::Sides to shrink. 

* related https://github.com/rerun-io/rerun/issues/10494

### DIFF
--- a/crates/egui/src/containers/sides.rs
+++ b/crates/egui/src/containers/sides.rs
@@ -18,10 +18,15 @@ use crate::{Layout, Ui, UiBuilder};
 /// The width of the gap is dynamic, based on the max width of the parent [`Ui`].
 /// When the parent is being auto-sized ([`Ui::is_sizing_pass`]) the gap will be as small as possible.
 ///
-/// ~If the parent is not wide enough to fit all widgets, the parent will be expanded to the right.~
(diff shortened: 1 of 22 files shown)


crates/egui/src/containers/sides.rs
```diff
@@ -68,58 +83,175 @@ impl Sides {
...
+
+    /// Wrap the text on the shrinking side.
+    ///
+    /// This is a shortcut for [`Self::wrap_mode`].
+    /// Does nothing if [`Self::extend`] is used (the default).
+    #[inline]
+    pub fn wrap(mut self) -> Self {
+        self.wrap_mode = Some(crate::TextWrapMode::Wrap);
+        self
+    }
```
abey79, line 143: Would it make sense to default to `Truncate` if you use `shrink_*`?

abey79 (approved):
Great, thanks! This is going to be so useful!

crates/egui/src/containers/sides.rs
lucasmerlin: It might, but I can also imagine cases where you wouldn't want it and then it'd be annoying to have to un-set it.

PR #7297, opened 2025-07-03
Make `Memory::keep_popup_open` less of a footgun

### LINKED ISSUE #7037: Make `Memory::keep_popup_open` less of a footgun
https://github.com/emilk/egui/pull/5814 made it so popups would close unless you keep calling `keep_popup_open` each frame. In theory this is a nice improvement but it's caused [some](https://github.com/emilk/egui/pull/5847) [bugs](https://github.com/rerun-io/rerun/pull/9923) since and I'm certain more will happen if we release this in the current state. I suggest we find a better solution, or at least make it a (non-silent) breaking change so people are forced to think about it.

lucasmerlin:
* Closes #7037 

The breaking change is intentional, to force people to think about this, since otherwise the keep_popup_open thing would silently break popups.

### DIFF
--- a/crates/egui/src/containers/popup.rs
+++ b/crates/egui/src/containers/popup.rs
@@ -127,7 +127,7 @@ impl OpenKind<'_> {
             OpenKind::Open => true,
             OpenKind::Closed => false,
             OpenKind::Bool(open) => **open,
-            OpenKind::Memory { .. } => ctx.memory(|mem| mem.is_popup_open(id)),
+            OpenKind::Memory { .. } => ctx.memory(|mem| mem.is_showing_popup(id)),
         }
     }
 }
@@ -446,7 +446,7 @@ impl<'a> Popup<'a> {
             OpenKind::Open => true,
             OpenKind::Closed => false,
             OpenKind::Bool(open) => **open,
-            OpenKind::Memory { .. } => self.ctx.memory(|mem| mem.is_popup_open(self.id)),
+            OpenKind::Memory { .. } => self.ctx.memory(|mem| mem.is_showing_popup(self.id)),
         }
     }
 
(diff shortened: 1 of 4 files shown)


crates/egui/src/memory/mod.rs
```diff
@@ -1017,13 +1017,29 @@ impl OpenPopup {
...
+    /// If [`Self::show_popup`] is not called for a frame, the open state will be cleared.
+    pub fn is_showing_popup(&self, popup_id: Id) -> bool {
         self.popups
             .get(&self.viewport_id)
             .is_some_and(|state| state.id == popup_id)
             || self.everything_is_visible()
     }
 
+    /// Check if the popup is open and keep it open.
+    pub fn show_popup(&mut self, popup_id: Id) -> bool {
```
emilk, line 1031: The docstring for `open_popup`  should link to this function, and vice-versa, to explain their difference..

Also: wouldn't it make sense if this called `self.open_popup`?

lucasmerlin: No, if this called open_popup, you couldn't use it to check if we should show the popup.

The docs should also mention it needs to be called every frame.

emilk: I think this is just another shape of footgun.

Based only on the names of the functions, I would expect this to always hold:

```rs
mem.show_popup(popup_id);
assert!(mem.is_showing_popup(popup_id));
```

Which would then lead to the question "what is open_popup for?"

crates/egui/src/containers/popup.rs
```diff
@@ -531,7 +531,7 @@ impl<'a> Popup<'a> {
                     mem.toggle_popup(id);
                 }
                 None => {
-                    mem.keep_popup_open(id);
+                    mem.show_popup(id);
```
emilk, line 524: Shouldn't we always call this unless we `return None` below? 🤔

emilk:
I'm doing another take on this

PR #7311, opened 2025-07-07
Improve texture filtering by doing it in gamma space

emilk:
* Closes https://github.com/emilk/egui/pull/5839

This makes some transparent images look a lot nicer when blended:

image

Cursive text will also look nicer.

This unfortunately changes the contract of what `register_native_texture` expects

### DIFF
--- a/crates/egui_glow/src/painter.rs
+++ b/crates/egui_glow/src/painter.rs
@@ -172,7 +172,7 @@ impl Painter {
 
         let supported_extensions = gl.supported_extensions();
         log::trace!("OpenGL extensions: {supported_extensions:?}");
-        let srgb_textures = false; // egui wants linear textures
+        let srgb_textures = false; // egui wants normal sRGB-unaware textures
 
         let supports_srgb_framebuffer = !cfg!(target_arch = "wasm32")
             && supported_extensions.iter().any(|extension| {


crates/egui-wgpu/src/renderer.rs
```diff
@@ -690,7 +690,7 @@ impl Renderer {
     ///
     /// This enables the application to reference the texture inside an image ui element.
     /// This effectively enables off-screen rendering inside the egui UI. Texture must have
-    /// the texture format [`wgpu::TextureFormat::Rgba8UnormSrgb`].
+    /// the texture format [`wgpu::TextureFormat::Rgba8Unorm`].
```
Wumpf, line 693: huh I guess there's a future work item here that iff `Rgba8UnormSrgb` is passed in, we create an `Rgba8Unorm` view either way (telling the user that we do that)
If I remember correctly WebGPU standard now guarantees this working on the minspec.

Wumpf (approved):
"academically" this still pains me that this is necessary but it's perfectly in line with what was discussed before and the improved results speak for themselves! 👍 

When updating Rerun, we should check whether its map view is still looking as before - we are blending re_renderer rendered stuff on top of egui (the map itself) there and I'm not sure if this affects that or not.

crates/egui-wgpu/src/renderer.rs
emilk: Can you even know what format the given texture has?

Wumpf: yeah that's exposed by now. Wasn't back when the code was written!

Wumpf: .. right we should check that now that we can

Wumpf: right. this is a `TextureView` not a texture, but it's _supposed_ to expose these things https://www.w3.org/TR/webgpu/#gputextureview
But alas wgpu doesn't do that yet. Tomorrow is a wgpu release and I want this badly, let's whip that up quickly..

Wumpf: https://github.com/gfx-rs/wgpu/pull/7895

Issue #7256, opened 2025-06-20
Failed surface creation on macbook m4 pro


Bersez:
## System
macbook m4 pro
macOS Sequoia 15.5

egui = "0.31.1"
egui-wgpu = { version = "0.31.1" }
egui-winit = "0.31.1"
pollster = "0.4.0"

Neither my project nor the example from this repository: [link](https://github.com/kaphula/winit-egui-wgpu-template) is working, and in both projects same error. But on Windows 10 and on Ubuntu 24.04  all works fine

The Error:

thread 'main' panicked at src/app.rs:124:14:
Failed to create surface!: CreateSurfaceError { inner: Hal(FailedToCreateSurfaceForAnyBackend({})) }
stack backtrace:
   0:        0x1049e1ca0 - std::backtrace_rs::backtrace::libunwind::trace::hd09570c029a6744a
                               at /rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/std/src/../../backtrace/src/backtrace/libunwind.rs:117:9
   1:        0x1049e1ca0 - std::backtrace_rs::backtrace::trace_unsynchronized::h8d2fa64833f91cb3
                               at /rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/std/src/../../backtrace/src/backtrace/mod.rs:66:14
   2:        0x1049e1ca0 - std::sys::backtrace::_print_fmt::h567d2a106a3b0dee
                               at /rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/std/src/sys/backtrace.rs:66:9
   3:        0x1049e1ca0 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h92dda645f072dcaf
                               at /rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/std/src/sys/backtrace.rs:39:26
   4:        0x1049fc288 - core::fmt::rt::Argument::fmt::hc0b28dad2d7b7ba8
                               at /rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/core/src/fmt/rt.rs:184:76
   5:        0x1049fc288 - core::fmt::write::hbc92919d8e8f9a96
                               at /rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/core/src/fmt/mod.rs:1481:21
[...]

Issue #7106, opened 2025-05-31
No wgpu backend feature that is implemented for the target platform was enabled


obsoleszenz:
`cargo run` on current main is not starting anymore. It seems like wgpu can't find a backend feature or something. 0.31.1 is working fine for me.

Log:
```
k@nixos ~/C/egui (main)> RUST_LOG=trace cargo run                                                                                                                                          
   Compiling emath v0.31.1 (/home/k/Coding/egui/crates/emath)                                                                                                                              
   Compiling ecolor v0.31.1 (/home/k/Coding/egui/crates/ecolor)                                                                                                                            
   Compiling epaint v0.31.1 (/home/k/Coding/egui/crates/epaint)                                                                                                                            
   Compiling egui v0.31.1 (/home/k/Coding/egui/crates/egui)                                                                                                                                
   Compiling egui-wgpu v0.31.1 (/home/k/Coding/egui/crates/egui-wgpu)                                                                                                                      
   Compiling egui_glow v0.31.1 (/home/k/Coding/egui/crates/egui_glow)                                                                                                                      
   Compiling egui_extras v0.31.1 (/home/k/Coding/egui/crates/egui_extras)                                                                                                                  
   Compiling egui-winit v0.31.1 (/home/k/Coding/egui/crates/egui-winit)                                                                                                                    
   Compiling eframe v0.31.1 (/home/k/Coding/egui/crates/eframe)                                                                                                                            
   Compiling egui_demo_lib v0.31.1 (/home/k/Coding/egui/crates/egui_demo_lib)                                                                                                              
   Compiling egui_demo_app v0.31.1 (/home/k/Coding/egui/crates/egui_demo_app)                                                                                                              
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 18.11s                                                                                                                   
     Running `target/debug/egui_demo_app`                                                                                                                                                  
[2025-05-31T14:01:09Z INFO  eframe] Both the glow and wgpu renderers are available. Using wgpu.                                                                                            
[2025-05-31T14:01:09Z DEBUG eframe] Using the wgpu renderer                                                                                                                                
[2025-05-31T14:01:09Z DEBUG sctk] Bound new global [59] wl_output v4                                                                                                                       
[2025-05-31T14:01:09Z DEBUG sctk] Bound new global [31] zxdg_output_manager_v1 v3                                                                                                          
[2025-05-31T14:01:09Z DEBUG sctk] Bound new global [10] wl_seat v7                                                                                                                         
[...]

PR #7314, opened 2025-07-08
Add "metal" to correctly run on MacOS

Bersez:
Solves the issue with: https://github.com/emilk/egui/issues/7256


Bersez:
maybe also add another feature to solve #7106

lucasmerlin:
We shouldn't force features on users of egui_wgpu that they don't need (in theory someone could use the gl or vulkan backend on macos and might want to disable the metal feature).
Instead you can just enable this feature by adding `wgpu = {version = "same as wgpu in egui-wgpu", features = "metal"}` to your Cargo.toml. You probably don't even need the `features = "metal"` since metal is part of the default features of wgpu

PR #7316, opened 2025-07-09
Exclude `\n` when splitting `Galley`s

lucasmerlin:
* Follow up to #7146 

Previously when galleys were splitted, each exept the last had an extra empty row that had to be removed when they were concated. This changes it to remove the `\n` from the layout jobs when splitting.

### DIFF
--- a/crates/epaint/src/text/fonts.rs
+++ b/crates/epaint/src/text/fonts.rs
@@ -922,7 +922,7 @@ impl GalleyCache {
                 if section_range.end <= start {
                     // The section is behind us
                     current_section += 1;
-                } else if end + 1 <= section_range.start {
+                } else if end < section_range.start {
                     break; // Haven't reached this one yet.
                 } else {
                     // Section range overlaps with paragraph range
--- a/crates/epaint/src/text/fonts.rs
+++ b/crates/epaint/src/text/fonts.rs
@@ -1104,10 +1104,10 @@ mod tests {
             },
             {
                 let mut job = LayoutJob::simple(
-                    "Simple test.".to_owned(),
+                    "Test text with a lot of words\n and a newline.".to_owned(),
                     FontId::new(14.0, FontFamily::Monospace),
                     Color32::WHITE,
-                    f32::INFINITY,
+                    40.0,
                 );
                 job.first_row_min_height = 30.0;
                 job
@@ -1234,7 +1234,7 @@ mod tests {
                         let text = job.text.clone();
                         let galley_unwrapped = layout(&mut fonts, job.into());
 
-                        let intrinsic_size = galley_wrapped.intrinsic_size;
+                        let intrinsic_size = galley_wrapped.intrinsic_size();
                         let unwrapped_size = galley_unwrapped.size();
 
                         let difference = (intrinsic_size - unwrapped_size).length().abs();
@@ -1253,7 +1253,7 @@ mod tests {
                             format!("{unwrapped_size:.4?}"),
                             "Unwrapped galley intrinsic size should exactly match its size. \
                                 {:.8?} vs {:8?}",
-                            galley_unwrapped.intrinsic_size,
+                            galley_unwrapped.intrinsic_size(),
                             galley_unwrapped.size(),
                         );
                     }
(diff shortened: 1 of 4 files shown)


crates/epaint/src/text/text_layout.rs
```diff
@@ -659,12 +653,12 @@ fn galley_from_rows(
     let mut cursor_y = 0.0;
 
     for placed_row in &mut rows {
-        let mut max_row_height = first_row_min_height.max(placed_row.rect().height());
+        let mut max_row_height = first_row_min_height.at_least(placed_row.height());
```
lucasmerlin, line 654: This really got me, changing the position here to NaN:

https://github.com/emilk/egui/blob/4e50c17f3f54a0f406aa882c0f4ffeff772f9ec2/crates/epaint/src/text/text_layout.rs#L250

Would cause the max_row_height here to change to 0. Took me a bit to see that this was due to getting the height from the rect instead of looking at the size directly.

crates/epaint/src/text/fonts.rs
```diff
@@ -1091,6 +1089,19 @@ mod tests {
...
+                    FontId::default(),
+                    Color32::WHITE,
+                    f32::INFINITY,
+                );
+                job.append("\n", 0.0, TextFormat::default());
+                job.append("\n", 0.0, TextFormat::default());
+                job.append("world", 0.0, TextFormat::default());
+                job.wrap.max_rows = 2;
+                job
+            },
```
emilk, line 1104: Let's also add a test case that sets `first_row_min_height` to something large

crates/epaint/src/text/text_layout.rs
```diff
@@ -1212,4 +1206,76 @@ mod tests {
...
+            "Expected no glyphs in the empty row"
+        );
+        assert_eq!(
+            galley.size(),
+            Vec2::new(0.0, 16.0),
+            "Unexpected galley size"
+        );
+        assert_eq!(
+            galley.intrinsic_size,
+            Vec2::new(0.0, 16.0),
```
emilk: Where does `16.0` come from? `FontId::default()` is 14.0

crates/epaint/src/text/text_layout_types.rs
```diff
@@ -849,32 +849,28 @@ impl Galley {
...
+            let is_last_galley = i + 1 == galleys.len();
+
+            merged_galley
+                .rows
+                .extend(galley.rows.iter().enumerate().map(|(row_idx, placed_row)| {
+                    let new_pos = placed_row.pos + current_y_offset * Vec2::Y;
+                    let new_pos = new_pos.round_to_pixels(pixels_per_point);
+                    merged_galley.mesh_bounds = merged_galley
+                        .mesh_bounds
+                        .union(placed_row.visuals.mesh_bounds.translate(new_pos.to_vec2()));
```
emilk, line 870: What do you think about adding a `BirOr` override to `Rect`? We already have it for Response

```suggestion
                    merged_galley.mesh_bounds |=
                        placed_row.visuals.mesh_bounds.translate(new_pos.to_vec2());
```

crates/epaint/src/text/text_layout.rs
lucasmerlin: This also surprised me, font.row_height() returns 16.0. Seems like this comes from the font_scaling:

https://github.com/emilk/egui/blob/022d34208a15068af6be650a18c74fb4dcf0fe8f/crates/epaint/src/text/fonts.rs#L1040-L1048

For the default font, this seems to be 1.121, so the font gets scaled to 15.694. 

Then here we get to ~16.125:

https://github.com/emilk/egui/blob/7ac137bfc167d0f7ff78d5fd042d6042ff190455/crates/epaint/src/text/font.rs#L133

crates/epaint/src/text/fonts.rs
```diff
@@ -1091,6 +1089,29 @@ mod tests {
...
+            {
+                let mut job = LayoutJob::simple(
+                    "Simple test.".to_owned(),
```
emilk: I think this test should have multiple rows (longer text) and paragraphs (\n).

crates/epaint/src/text/text_layout.rs
emilk: makes sense - let's be explicit in the tests though by replacing the hard-coded 16.0


```
let row_height = fonts.row_height(font_id):
```

This makes it less brittle if we change the default `FontId`. Also, self-documenting.

crates/epaint/src/text/text_layout_types.rs
lucasmerlin: That'd be neat!

Issue #7292, opened 2025-07-02
align(4) for Color32


anti-social:
**Is your feature request related to a problem? Please describe.**

I want to post-process an image using SIMD before converting it to a texture. To do that I need a buffer with an align of 4. So I allocate a `Vec<u32>` then cast its slice to a `&[u8]`, fill it with data and post-process. But after that I cannot cast the vector to `Vec<Color32>` to create a `ColorImage` object.

**Describe the solution you'd like**

Adding `#[repr(align(4))]` to the `Color32` struct solves the problem.

**Describe alternatives you've considered**

I do not know alternatives without a UB and extra copying of a buffer.

**Additional context**

It looks like it is not a backward incompatible change.

mattfbacon:
In practice you can probably just allocate `Vec<u8>` directly and it will be aligned enough to cast to `&mut [u32]` for SIMD. `malloc` for example has a minimum alignment of 16 bytes on x64. If not there is the [`aligned-vec`](https://docs.rs/aligned-vec/latest/aligned_vec/) crate.

anti-social:
> In practice you can probably just allocate `Vec<u8>` directly and it will be aligned enough to cast to `&mut [u32]` for SIMD. `malloc` for example has a minimum alignment of 16 bytes on x64. 

I tried but it was definitely not aligned.

> If not there is the [`aligned-vec`](https://docs.rs/aligned-vec/latest/aligned_vec/) crate.

Yeah, it may work but requires an unsafe code to cast from `AVec<u8>` to `Vec<u8>`. TBH I'm not sure is it possible to make such a cast?

anti-social:
> It looks like it is not a backward incompatible change.

It is a mistake, of course it is not backward compatible because someone could cast `Vec<u8>` to `Vec<Color32>` before.

anti-social:
> If not there is the [`aligned-vec`](https://docs.rs/aligned-vec/latest/aligned_vec/) crate.

It looks like following code works fine:

```rust
        let data = {
            let mut data = ManuallyDrop::new(
                aligned_vec::avec!(0u8; width * height * 4)
            );
            unsafe {
                Vec::from_raw_parts(data.as_mut_ptr(), data.len(), data.capacity())
            }
        };
```

Thank you for pointing out to the crate.

mattfbacon:
It's probably UB if you run it through miri but afaik all system allocators ignore the passed layout entirely for dealloc since they track it internally.

anti-social:
[...]

PR #7318, opened 2025-07-09
Align `Color32` to 4 bytes

anti-social:
It could be useful when you want to preprocess an image using SIMD instructions without extra copying to convert it to a texture.

Possibly we might add it as a feature.

* Closes <https://github.com/emilk/egui/issues/7292>


emilk (approved):
Makes sense!

PR #7333, opened 2025-07-10
Color emojis 🦀

lucasmerlin:
#7298 and [a discussion on discord](https://discord.com/channels/900275882684477440/900275883124858921/1392919531516727397) made me realize that adding custom color emoji support would be quite trivial now, so I made this prototype:

https://github.com/user-attachments/assets/8407ab60-dcde-4769-b0a4-e102f386fae1

It currently only works with dark mode, we probably need some flag on the Glyph that it has color data and should not be affected with any tint. 
Also this won't work for loading fonts with color emojis, for that we need #5784 or ab_glyph needs texture data support. It's more for loading custom emojis like on discord and slack (which would be perfectly fine for my usecase)

### DIFF
--- a/crates/epaint/src/text/font.rs
+++ b/crates/epaint/src/text/font.rs
@@ -4,7 +4,7 @@ use std::sync::Arc;
 use emath::{GuiRounding as _, Vec2, vec2};
 
 use crate::{
-    TextureAtlas,
+    ColorImage, TextureAtlas,
     mutex::{Mutex, RwLock},
     text::FontTweak,
 };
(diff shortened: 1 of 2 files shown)


emilk:
Neat! I think this is a cool feature to add, but in order to properly ship it we need to store a `Map<char, ColorImage>` somewhere (in `Fonts`?) so that we can repopulate the texture atlas when it is flushed.

We also need to figure out the tint problem.

jb55:
exciting!

emilk:
Closing: color emojis landed via #8490 and #8512 on the new text stack.

jb55:
yuge

PR #7337, opened 2025-07-10
Fix out-of-the-box Windows build for eframe 0.32

bolli24:
wgpu v0.25 made the Vulkan backend optional. As a result, eframe projects on Windows no longer run out of the box, since Vulkan is not enabled by default.

This PR explicitly enables the vulkan feature in wgpu to restore out-of-the-box functionality on Windows.


bolli24:
Fixes #7342

Wumpf:
Superseeded by https://github.com/emilk/egui/pull/7344 which has a better solution - the solution proposed here would make it impossible to opt out of Vulkan

PR #7340, opened 2025-07-11
Update wgpu to 26 and wasm-bindgen to 0.2.100

torokati44:

There are new duplicate crates in the dep tree (`core-*`), but `ndk-sys` is deduplicated in exchange.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -103,7 +103,7 @@ jobs:
       - name: wasm-bindgen
         uses: jetli/wasm-bindgen-action@v0.1.0
         with:
-          version: "0.2.97"
+          version: "0.2.100"
 
       - run: ./scripts/wasm_bindgen_check.sh --skip-setup
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -108,7 +108,7 @@ dependencies = [
  "accesskit_macos",
  "accesskit_unix",
  "accesskit_windows",
- "raw-window-handle 0.6.2",
+ "raw-window-handle",
  "winit",
 ]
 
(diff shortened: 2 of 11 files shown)


dezajno:
Is the wgpu version upgrade blocked by something? I'd be interested in pushing this as I need it for an own project. Can I help somehow?

Kumpelinus:
Currently the issue seems to be that wgpu is now no_std and the macro from the puffin crate still pulls in std components in the expanded macro (like `std::sync::OnceLock`) which doesn't work with no_std ofc. I will probably take a look at fixing this later today. (The demo app does compile without the `profiling/profile-with-puffin` feature so that seems to be the only blocker besides the macOS Issues)

Kumpelinus:
@torokati44 should I open a different PR or should I edit your existing one (If you give me access to the branch)?

Kumpelinus:
I’m not sure which crate should host the fix. The underlying issue lies in **puffin**, which is currently unmaintained, so the patch could land in one of four places:
 * the **egui-examples** crate the error is happening in (other egui users will likely hit the same problem),
 * the **egui** crate itself
 * **wgpu**, or
 * the **profiling** crate.

torokati44:
@Kumpelinus I think it's easier if you open your own, I won't mind. I don't think this permission can be easily granted to you.

Kumpelinus:
Ok, will do 👍

nicoburns:
I think bumping `arboard` will fix the duplicated `core-*` dependencies

torokati44:
Are you sure? I wasn't able to do it that way...

torokati44:
Merged as https://github.com/emilk/egui/pull/7540.

PR #7344, opened 2025-07-11
Enable wgpu default features in eframe / egui_wgpu default features

### LINKED ISSUE #7342: No available wgpu backends with eframe 0.32.0 on Windows 10
Upon updating eframe to 0.32.0 wgpu seems broken.
This is how i setup eframe

```rust
fn main() -> eframe::Result {
   panic::set_hook(Box::new(|panic_info| {
      let message = panic_info
         .payload()
         .downcast_ref::<&str>()
         .map_or("Unknown panic", |s| s);
      let location = panic_info
         .location()
         .map_or("Unknown location".to_string(), |loc| {
            format!("{}:{}:{}", loc.file(), loc.line(), loc.column())
         });
      tracing::error!("Panic occurred: '{}' at {}", message, location);
   }));

   let _tracing_guard = setup_tracing();

   let wgpu_setup = WgpuSetup::CreateNew(WgpuSetupCreateNew {
      instance_descriptor: InstanceDescriptor {
         backends: Backends::PRIMARY | Backends::GL,
         ..Default::default()
      },
      power_preference: PowerPreference::HighPerformance,
      device_descriptor: Arc::new(|_adapter| {
         wgpu::DeviceDescriptor {
            memory_hints: MemoryHints::MemoryUsage,
            ..Default::default()
         }
      }),
      ..Default::default()
   });

   let wgpu_config = WgpuConfiguration {
      wgpu_setup,
      ..Default::default()
   };

      wgpu_options: wgpu_config,
         .with_decorations(false)
         .with_inner_size([1280.0, 900.0])
         .with_min_inner_size([1280.0, 900.0])
         .with_transparent(true)
         .with_resizable(true),

      ..Default::default()
   };

   eframe::run_native(
      "Zeus",
      options,
      Box::new(|cc| {

         let app = ZeusApp::new(&cc);

         Ok(Box::new(app))
      }),
   )
}
```
I know for a fact at least on my machine VULKAN is working, but yet for some reason wgpu panics:

```
Panic occurred: 'No wgpu backend feature that is implemented for the target platform was enabl
ed. See `wgpu::Instance::enabled_backend_features()` for more information.' at C:\Users\mixal\.cargo\registry\src\index.crates.io-1949
cf8c6b5b557f\wgpu-25.0.2\src\api\instance.rs:117:13
```

Not really sure why this is happening, everything was working great up to 0.31.1

lucasmerlin:
* Closes #7342
* related #7314

### DIFF
--- a/crates/eframe/Cargo.toml
+++ b/crates/eframe/Cargo.toml
@@ -40,6 +40,9 @@ default = [
   "winit/default",
   "x11",
   "egui-wgpu?/fragile-send-sync-non-atomic-wasm",
+  # Let's enable some backends so that users can use `eframe` out-of-the-box
+  # without having to explicitly opt-in to backends
+  "egui-wgpu?/default"
 ]
 
 ## Enable platform accessibility API implementations through [AccessKit](https://accesskit.dev/).
@@ -167,12 +170,7 @@ glutin-winit = { workspace = true, optional = true, default-features = false, fe
   "wgl",
 ] }
 home = { workspace = true, optional = true }
-wgpu = { workspace = true, optional = true, features = [
-  # Let's enable some backends so that users can use `eframe` out-of-the-box
-  # without having to explicitly opt-in to backends
-  "metal",
-  "webgpu",
-] }
+wgpu = { workspace = true, optional = true }
 
 # mac:
 [target.'cfg(any(target_os = "macos"))'.dependencies]
--- a/crates/egui-wgpu/Cargo.toml
+++ b/crates/egui-wgpu/Cargo.toml
@@ -31,7 +31,7 @@ all-features = true
 rustdoc-args = ["--generate-link-to-definition"]
 
 [features]
-default = ["fragile-send-sync-non-atomic-wasm"]
+default = ["fragile-send-sync-non-atomic-wasm", "wgpu/default"]
 
 ## Enable [`winit`](https://docs.rs/winit) integration. On Linux, requires either `wayland` or `x11`
 winit = ["dep:winit", "winit/rwh_06"]


lucasmerlin:
I have also disabled the features that we enabled based on the cfg. I think we can just enable the default wgpu backends by default and if the eframe default features are disabled, we should enable no backends. Maybe someone wants to use the gl or vulkan wgpu backend on mac (not recommended but technically possible), then they wouldn't need the metal backend.

Wumpf (changes requested):
ah, so the idea is that if you want to have less wgpu features you'd just disable default features on eframe, enable `egui-wgpu` feature yourself and then opt in to wgpu features manually?
Sounds good, but I think it's not working with that yet - eframe's  optional `egui-wgpu` dependency would need to set no-default-features. Best to try it out! I find building with `--timings` and checking the html output for activated features easiest for this

lucasmerlin:
> Sounds good, but I think it's not working with that yet - eframe's optional `egui-wgpu` dependency would need to set no-default-features.

Default features are disabled via the workspace, that should do it right?
https://github.com/emilk/egui/blob/fdcaff8465eac8db8cc1ebbcbb9b97e0791a8363/Cargo.toml#L65

Wumpf:
ah yeah, I think that should do it! I'll give it a quick spin just to be sure

Wumpf (approved):
sorry for dropping the ball on this!
Tested it on windows: the following configuration successfully gave me a Vulkan only build:
```
eframe = { path = "../egui/crates/eframe", default-features = false, features = [
    "wgpu",
] }
wgpu = { version = "25", default-features = false, features = ["vulkan"] }
```

PR #7372, opened 2025-07-18
Add `UiBuilder::global_scope` and `UiBuilder::id`

Icekey:
I added a new flag to the UiBuilder so that it is possible to move child widgets around the ui tree without losing state information. 
Currently there is no way to create child widgets with the same id at different locations in the ui tree since ids change in relation the the parent id. With the new flag a unique global scope can be created which always results in the same ids even at different locations. 
You still need to ensure that the widgets only get rendered once in frame.

This feature can be used to fix a issue i am having with the https://github.com/lucasmerlin/hello_egui crate.

* Closes https://github.com/lucasmerlin/hello_egui/issues/75

### DIFF
--- a/crates/egui/src/ui_builder.rs
+++ b/crates/egui/src/ui_builder.rs
@@ -63,7 +63,7 @@ impl UiBuilder {
     /// This way child widgets can be moved in the ui tree without losing state.
     /// You have to ensure that in a frame the child widgets do not get rendered in multiple places.
     ///
-    /// You should set the same unique `id_salt` at every place in the ui tree where you want the
+    /// You should set the same globally unique `id_salt` at every place in the ui tree where you want the
     /// child widgets to share state.
     #[inline]
     pub fn global_scope(mut self, global_scope: bool) -> Self {


crates/egui/src/ui.rs
```diff
@@ -287,8 +289,14 @@ impl Ui {
...
-        let stable_id = self.id.with(id_salt);
-        let unique_id = stable_id.with(self.next_auto_id_salt);
+        let (id, next_auto_id_salt) = if global_scope {
+            (id_salt, id_salt.value())
+        } else {
+            (self.id, self.next_auto_id_salt)
+        };
+
+        let stable_id = id.with(id_salt);
+        let unique_id = stable_id.with(next_auto_id_salt);
```
lucasmerlin: It'd be great if we could explicitedly set the id (especially the unique_id, so we could e.g. read the response before the ui is created).

So I think if global_scope is set we can just set stable_id and unique_id to id_salt (in Ui::new they are also both equal, so it shouldn't be a problem if they are the same).

crates/egui/src/ui_builder.rs
```diff
@@ -42,6 +43,18 @@ impl UiBuilder {
...
+    /// You have to ensure that in a frame the child widgets do not get rendered in multiple places.
+    ///
+    /// You should set the same unique `id_salt` at every place in the ui tree where you want the
+    /// child widgets to share state.
+    #[inline]
+    pub fn global_scope(mut self, global_scope: bool) -> Self {
+        self.global_scope = global_scope;
+        self
+    }
+
```
lucasmerlin, line 73: A `UiBuilder::id` fn that sets `id_salt` and `global_scope` to true in one go would be useful and would also help with discoverability since I don't think the name `global_scope` is very obvious.

crates/egui/src/widget_rect.rs
```diff
@@ -160,14 +160,6 @@ impl WidgetRects {
                 // e.g. calling `response.interact(…)` to add more interaction.
                 let (idx_in_layer, existing) = entry.get_mut();
 
-                debug_assert!(
-                    existing.layer_id == widget_rect.layer_id,
-                    "Widget {:?} changed layer_id during the frame from {:?} to {:?}",
-                    widget_rect.id,
-                    existing.layer_id,
-                    widget_rect.layer_id
-                );
```
lucasmerlin: I feel like this debug_assert should still be correct? Even if the widget is shown on a different layer on the next frame, within a single frame it should always have the same layer id right?

Icekey: You are right. I can no longer reproduce the issue i was having with the assert.

crates/egui/src/ui_builder.rs
Icekey: okay i added UiBuilder::id. Let me know if you have a better name for the global_scope flag.

crates/egui/src/ui_builder.rs
```diff
@@ -42,6 +43,32 @@ impl UiBuilder {
         self
     }
 
+    /// Set an id of the new `Ui` that is independent of the parent `Ui`.
+    /// This way child widgets can be moved in the ui tree without losing state.
+    /// You have to ensure that in a frame the child widgets do not get rendered in multiple places.
+    ///
+    /// You should set the same unique `id` at every place in the ui tree where you want the
+    /// child widgets to share state.
+    /// If the child widgets are not moved in the ui tree, use [`UiBuilder::id_salt`] instead.
```
lucasmerlin, line 52:
```suggestion
    /// If the child widgets are not moved in the ui tree, use [`UiBuilder::id_salt`] instead.
    ///
    /// This is a shortcut for `.id_salt(my_id).global_scope(true)`.
```

Icekey: I added the shortcut info.

Icekey:
How does merging work?

lucasmerlin:
I usually go through all approved PRs with @emilk every tuesday, but hes on holiday right now so it'll have to wait until next tuesday.

crates/egui/src/ui_builder.rs
```diff
@@ -42,6 +43,34 @@ impl UiBuilder {
...
+        self.id_salt = Some(Id::new(id));
+        self.global_scope = true;
+        self
+    }
+
+    /// Make the new `Ui` child ids independent of the parent `Ui`.
+    /// This way child widgets can be moved in the ui tree without losing state.
+    /// You have to ensure that in a frame the child widgets do not get rendered in multiple places.
+    ///
+    /// You should set the same unique `id_salt` at every place in the ui tree where you want the
```
emilk:
```suggestion
    /// You should set the same globally unique `id_salt` at every place in the ui tree where you want the
```

emilk (approved):

PR #7375, opened 2025-07-19
Add theme selection for code editor

Its-Just-Nans:
Change the theme selection of the code editor

Before
- was showing the window theme selector

After
- is showing the code editor theme selector

image


Its-Just-Nans:
Also, why not making public `SyntectTheme` and functions in `impl SyntectTheme {}` ?

emilk (approved):

Issue #4347, opened 2024-04-11
Context animation functions could use `impl Into<Id>` instead of just `Id`


j-n-f:
Many functions that require an `Id` specify the parameter as `impl Into<Id>`. This means for 99% of cases you can just use an `&'static str` (which is convenient).

These 3 functions require something like `"my_id".into()` which is slightly less ergonomic.

* [animate_bool](https://github.com/emilk/egui/blob/0.27.2/crates/egui/src/context.rs#L2412)
* [animate_bool_with_time](https://github.com/emilk/egui/blob/0.27.2/crates/egui/src/context.rs#L2419)
* [animate_value_with_time](https://github.com/emilk/egui/blob/0.27.2/crates/egui/src/context.rs#L2440)

These are the only functions I've run into that have this quirk while working with `egui`, but there may be more as well. At a minimum, I'd like to submit a PR to make these ones more convenient to use.

PR #7377, opened 2025-07-20
Use impl Into<Id> for parameters

malezjaa:
* Closes <https://github.com/emilk/egui/issues/4347> and possibly more
* [x] I ran `cargo fmt` and `cargo clippy`

This PR updates egui methods to accept `impl Into<Id>` instead of just `Id`, so you can just pass any accepted value without having to call `.into()`

### DIFF
--- a/crates/egui/src/animation_manager.rs
+++ b/crates/egui/src/animation_manager.rs
@@ -31,9 +31,10 @@ impl AnimationManager {
         &mut self,
         input: &InputState,
         animation_time: f32,
-        id: Id,
+        id: impl Into<Id>,
         value: bool,
     ) -> f32 {
+        let id = id.into();
         let (start, end) = if value { (0.0, 1.0) } else { (1.0, 0.0) };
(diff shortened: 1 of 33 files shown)


crates/egui/src/animation_manager.rs
```diff
@@ -31,9 +31,10 @@ impl AnimationManager {
         &mut self,
         input: &InputState,
         animation_time: f32,
-        id: Id,
+        id: impl Into<Id>,
         value: bool,
     ) -> f32 {
+        let id = id.into();
```
Mingun, line 37: This will duplicate function body for every distinct type of `id`. To duplicate only body of conversion helper, you could introduce new private `_impl` function and use
```rust
fn xxx(id: impl Into<Id>) {
  self.xxx_impl(id.into(), ...)
}
```

emilk: Yeah, this PR monomorphizes hundreds of functions for very little gain :(

PR #7379, opened 2025-07-21
When justifying text, keep significant whitespace.

### LINKED ISSUE #1272: Respect leading spaces for rendering
I'm trying to align a cursor for menu entries.

```
                let cursor = if some_condition {
                    RichText::new("> ").color(Color32::LIGHT_GRAY).text_style(TextStyle::Monospace)
                } else {
                    RichText::new("  ").color(Color32::LIGHT_GRAY).text_style(TextStyle::Monospace)
                };

                ui.label(cursor);
                ui.label("Some Text");
```
Currently, the `else` condition never renders. There has to be an actual symbol besides just regular spaces. I'd prefer not to mess with the fonts as I like the default ones.

If there is a space character it was most likely put there deliberately and thus should be rendered by default.

Using a custom font with some space-equivalent-yet-renderable character.

**Additional context**
I've tried looking into the source code but unfortunately couldn't figure out where the `is_empty` call is made that is causing this.

NfNitLoop:
Fixes #1272

### DIFF
--- a/crates/epaint/src/text/text_layout.rs
+++ b/crates/epaint/src/text/text_layout.rs
@@ -595,7 +595,7 @@ fn halign_and_justify_row(
         }
 
         // Prevent 0-length ranges, which causes assertion errors & overflows below.
-        // (TODO: Can we just return early for these cases?)
+        // (Can we just return early for these cases?)
         if significant_glyphs.is_empty() || visible_glyphs.is_empty() {
             visible_glyphs = 0..row.glyphs.len();
             significant_glyphs = visible_glyphs.clone();
--- a/crates/epaint/src/text/text_layout.rs
+++ b/crates/epaint/src/text/text_layout.rs
@@ -560,16 +560,15 @@ fn halign_and_justify_row(
         return;
     }
 
-    let num_leading_spaces = row
-        .glyphs
-        .iter()
-        .take_while(|glyph| glyph.chr.is_whitespace())
-        .count();
-
-    let glyph_range = if num_leading_spaces == row.glyphs.len() {
-        // There is only whitespace
-        (0, row.glyphs.len())
-    } else {
+    // When visually justifying text within an alignment, "leading" text in left/right
+    // aligned text is significant and shouldn't be trimmed.
+    // See: https://github.com/emilk/egui/issues/1272
+    let (significant_glyphs, visible_glyphs) = {
+        let num_leading_spaces = row
+            .glyphs
+            .iter()
+            .take_while(|glyph| glyph.chr.is_whitespace())
+            .count();
         let num_trailing_spaces = row
             .glyphs
             .iter()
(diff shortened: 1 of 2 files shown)


crates/epaint/src/text/text_layout.rs
```diff
@@ -624,14 +649,21 @@ fn halign_and_justify_row(
     }
 
     placed_row.pos.x = point_scale.round_to_pixel(target_min_x);
-    let mut translate_x = -original_min_x - extra_x_per_glyph * glyph_range.0 as f32;
-
-    for glyph in &mut row.glyphs {
-        glyph.pos.x += translate_x;
-        glyph.pos.x = point_scale.round_to_pixel(glyph.pos.x);
-        translate_x += extra_x_per_glyph;
```
NfNitLoop: Note: there was a secondary minor bug here:

previously we were distributing the `extra_x_per_glyph` among all glyphs, even those that we had chosen to be trimmed. We calculated `extra_x_per_glyph` based on `num_glyphs_in_range`, but were then multiplying it by total glyphs. We were probably getting a bit too much extra spacing.

Below, we only add more `extra_x_*` when we're operating on glyphs within the `visible_glyphs` range.

crates/epaint/src/text/text_layout.rs
```diff
@@ -560,30 +560,55 @@ fn halign_and_justify_row(
...
 
-    let num_leading_spaces = row
-        .glyphs
-        .iter()
-        .take_while(|glyph| glyph.chr.is_whitespace())
-        .count();
-
-    let glyph_range = if num_leading_spaces == row.glyphs.len() {
-        // There is only whitespace
-        (0, row.glyphs.len())
```
NfNitLoop: I kept this logic in my refactor below. And I see that it avoids possible underflow when/after we get to the `assert!()` but I'm still a bit fuzzy on why we even continue in that case.  Do we need to shift all of the spaces into a left/center/right-aligned location? So... what, they're in a useful place when selecting label texts?

tests/egui_tests/tests/snapshots/sides/wrap_right_long.png
NfNitLoop, line 1: Aha. Now I bet I see why all text was being trimmed when right-justified. 

Each "row" that's being justified may contain a trailing space because the rest of the text has been moved into the next row.

I'll look into this later today.  Which would you prefer?
 * A local solution within `halign_and_justify_row()` that fixes this case.
 * A higher-level fix, which will remove whitespace between rows as they get split?  (I imagine this one might result in incorrect copy/paste though?)

lucasmerlin: I'd prefer a fix within halign_and_justify_row.

lucasmerlin (commented):
It'd be awesome if you could add an explicit test for this case so we don't break it in the future

emilk:
Closing: the trailing-whitespace half landed in #8102. If leading whitespace still matters for your case, please open a fresh PR on top of current main. Thanks for the PR, and sorry for the slow response

Issue #3747, opened 2023-12-29
Assertion Failure in egui_extras


rustbasic:
Subject: Assertion Failure in egui_extras

Dear Emilk,

I've been experiencing occasional assertion failures, approximately every few hours,
when using egui_extras. The error message displayed is as follows:

```
thread 'ehttp' panicked at C:\Users\user\.cargo\registry\src\index.crates.io-6f17d22bba15001f\egui_extras
-0.24.2\src\loaders\ehttp_loader.rs:99:21:
assertion failed: matches!(prev, Some(Poll :: Pending))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I haven't noticed any specific issues with the program's functionality,
and it seems to be operating normally.

Thanks,
Emilk.

egui version : v0.24.1, v0.24.0, v0.23.0
OS : windows10

emilk:
Sounds like a bug in `EhttpLoader`!

I think that `assert` may be wrong. In any case, it should have been a `debug_assert`

PR #7380, opened 2025-07-21
Remove incorrect assertion after image loading is done

Vanadiae:
This is the same issue that was fixed for the http bytes loader in 239ade9a59692f23a6b1d779a9c2631cf774896a

----------------

Funnily, [this comment](https://github.com/emilk/egui/issues/3747#issuecomment-1872192997) describes exactly how I encountered this issue:

> That assert is wrong if something calls forget between the start of the request and the end of it.

I'm displaying lots of images in a scrolling grid (20 or so visible at a time). It seems like image textures are never freed up automatically so it stacks up a lot meaning I have to unload the image textures manually with `egui::Context::forget_image()` in each `eframe::App::update()` call for the images that are no longer shown when scrolling.


Vanadiae:
Actually I suspect that what 239ade9 (#3750) wanted to achieve is kind of wrong in the case where we `forget()` between starting the loading process and we finish loading, as it's likely expected that any ongoing loading process gets cancelled (or at least in our case that it simply discards any loading result if `forget()` was called during the loading process). Indeed that's what https://github.com/emilk/egui/pull/6995 does for the file loader this time for admittedly the same root cause as ours here.

In the case of my application, it means that instead of crashing on the assertion the memory just looks looks like it's getting leaked a lot (it's trivial for me to reach 5GB RAM by just scrolling sufficiently fast that few images have even time to load fully before `forget()`-ting them). Once I manually call `ctx.forget_all_images()` (i.e. with a button) things get much more reasonable again.

I suspect that the WebP/SVG/GIF loaders do not have any of those issues as they do not use a background thread for loading (I'm not quite sure why that's something each ImageLoader implementation should care about themselves but that's another topic).

Vanadiae:
Now the ehttp and image loaders have the same correct behavior in case of `forget()` than the file loader (code from #6995). The memory consumption is much better now in terms of growth, although likely due to memory fragmentation I still need to call `libc::malloc_trim()` regularly due to the amount of concurrent image loading that's happening (in separate threads even worse).

emilk (approved):

PR #7361, opened 2025-07-14
Fix `WidgetText::Text` ignoring fallback font

lucasmerlin:
* closes https://github.com/emilk/egui/issues/7356

PR #7382, opened 2025-07-22
Allow customizing fallback font in AtomLayout

lucasmerlin:
- merge after https://github.com/emilk/egui/pull/7361 (this pr requires the clone derive introduced there)

Before it was hard coded to TextStyle::Button, but it should be customizable.

### DIFF
--- a/crates/egui/src/widgets/button.rs
+++ b/crates/egui/src/widgets/button.rs
@@ -1,7 +1,7 @@
 use crate::{
     Atom, AtomExt as _, AtomKind, AtomLayout, AtomLayoutResponse, Color32, CornerRadius, Frame,
-    Image, IntoAtoms, NumExt as _, Response, Sense, Stroke, TextWrapMode, Ui, Vec2, Widget,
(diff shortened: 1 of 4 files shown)


crates/egui/src/widgets/button.rs
```diff
@@ -40,7 +40,9 @@ pub struct Button<'a> {
 impl<'a> Button<'a> {
     pub fn new(atoms: impl IntoAtoms<'a>) -> Self {
         Self {
-            layout: AtomLayout::new(atoms.into_atoms()).sense(Sense::click()),
+            layout: AtomLayout::new(atoms.into_atoms())
+                .sense(Sense::click())
+                .fallback_font(TextStyle::Button),
```
emilk, line 45: Maybe this should be called "default_font" instead? I dunno

lucasmerlin: It's also called fallback font in WidgetText so I think we should update it there as well for consistency. I'll open a PR for that

PR #7383, opened 2025-07-22
Fix manual `Popup` not closing

lucasmerlin:
Fixes manually created popups (via `Popup::new`) not closing, since widget_clicked_elsewhere was always false.

This example would never close:

```rs
    let mut open = true;

    eframe::run_simple_native("My egui App", options, move |ctx, _frame| {
            let response = egui::Popup::new(
                Id::new("popup"),
                ctx.clone(),
                PopupAnchor::Position(Pos2::new(10.0, 10.0)),
                LayerId::new(Order::Foreground, Id::new("popup")),
            )
            .open(open)
            .show(|ui| {
                ui.label("This is a popup!");
                ui.label("You can put anything in here.");
            });

            if let Some(response) = response {
                if response.response.should_close() {
                    open = false;
                }
            }
        });
    })
```

### DIFF
--- a/crates/egui/src/containers/popup.rs
+++ b/crates/egui/src/containers/popup.rs
@@ -180,7 +180,7 @@ pub struct Popup<'a> {
     gap: f32,
 
     /// Used later depending on close behavior
-    widget_clicked_elsewhere: bool,
+    widget_clicked_elsewhere: Option<bool>,
 
     /// Default width passed to the Area
     width: Option<f32>,
@@ -205,7 +205,7 @@ impl<'a> Popup<'a> {
             rect_align: RectAlign::BOTTOM_START,
             alternative_aligns: None,
             gap: 0.0,
-            widget_clicked_elsewhere: false,
+            widget_clicked_elsewhere: None,
             width: None,
             sense: Sense::click(),
             layout: Layout::default(),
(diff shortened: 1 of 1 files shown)


crates/egui/src/containers/popup.rs
```diff
@@ -594,10 +594,13 @@ impl<'a> Popup<'a> {
             frame.show(ui, content).inner
         });
 
+        let close_click =
+            widget_clicked_elsewhere.unwrap_or_else(|| ctx.input(|i| i.pointer.any_click()));
+
         let closed_by_click = match close_behavior {
-            PopupCloseBehavior::CloseOnClick => widget_clicked_elsewhere,
+            PopupCloseBehavior::CloseOnClick => close_click,
```
emilk, line 599: Shouldn't this close on any click?

```suggestion
            PopupCloseBehavior::CloseOnClick => ctx.input(|i| i.pointer.any_click()),
```

lucasmerlin: We need to check if the click was the click that was used to open the popup, so we don't immediately close it again. 

Previously we did this via the `widget_clicked_elsewhere`, which is also what we did in the old popup api. But I just realized this also won't work when using `Popup::new` so I came up with a better check. Now, on click, the popup will only close when it was already open last frame (by checking if a reasponse exists via ctx.read_response).

emilk: Nice!

PR #7385, opened 2025-07-23
New Plugin trait

lucasmerlin:
This adds a new `Plugin` trait and new `input_hook` and `output_hook` plugin fns. Having a `Plugin` trait should make it easier to store state in the plugin and improve discoverability of possible plugin hooks. 

The old `on_begin_pass` and `on_end_pass` have been ported to use the new plugin trait, should we deprecate them?


lucasmerlin:
Now that I'm done I'm not super convinced that this is better then the previous api. Maybe instead I should just add `Context::on_input` and `Context::on_output`?

crates/egui/src/context.rs
```diff
@@ -1874,24 +1838,73 @@ impl Context {
     ///
     /// This can be used for egui _plugins_.
     /// See [`crate::debug_text`] for an example.
```
emilk: We should say that this is just a convenience wrapper around the more powerful [`Self::add_plugin`]

crates/egui/src/context.rs
```diff
@@ -1874,24 +1838,73 @@ impl Context {
...
-    pub fn on_end_pass(&self, debug_name: &'static str, cb: ContextCallback) {
-        let named_cb = NamedContextCallback {
-            debug_name,
-            callback: cb,
-        };
-        self.write(|ctx| ctx.plugins.on_end_pass.push(named_cb));
+    pub fn on_end_pass(&self, debug_name: &'static str, cb: plugin::ContextCallback) {
```
emilk, line 1864: We should say that this is just a convenience wrapper around the more powerful [`Self::add_plugin`]

crates/egui/src/debug_text.rs
```diff
@@ -58,24 +45,28 @@ struct Entry {
...
+        "DebugTextPlugin"
     }
 
-    fn paint(self, ctx: &Context) {
-        let Self { entries } = self;
+    fn on_end_pass(&mut self, ctx: &Context) {
+        if !self.entries.is_empty() {
+            let entries = std::mem::take(&mut self.entries);
+            Self::paint_entries(ctx, entries);
+        }
```
emilk: `paint_entries` already early-outs

```suggestion
        let entries = std::mem::take(&mut self.entries);
        Self::paint_entries(ctx, entries);
```

crates/egui/src/id.rs
```diff
@@ -79,6 +79,10 @@ impl Id {
         self.0.get()
     }
 
+    pub fn from_value(value: u64) -> Self {
+        Self::from_hash(value)
+    }
```
emilk: Please remove. This is a huge footgun. The invariant of `Id` is that it only contains high-entropy bits.

crates/egui/src/plugin.rs
```diff
@@ -0,0 +1,237 @@
...
+/// Add plugins via [`Context::add_plugin`].
+///
+/// Plugins should not hold a reference to the [`Context`], since this would create a cycle
+/// (which would prevent the [`Context`] from being dropped).
+#[expect(unused_variables)]
+pub trait Plugin: Send + Sync + 'static {
+    /// Plugin name.
+    ///
+    /// Used when profiling.
+    fn name(&self) -> &'static str;
```
emilk: Let's call it `debug_name` then just to be extra explicit

crates/egui/src/plugin.rs
```diff
@@ -0,0 +1,237 @@
...
+    ///
+    /// Useful to inspect or modify the output.
+    /// Since this is called outside a pass, don't show ui here.
+    fn output_hook(&mut self, output: &mut FullOutput) {}
+
+    /// Called just before the input is processed.
+    ///
+    /// Useful to inspect or modify the input.
+    /// Since this is called outside a pass, don't show ui here.
+    fn input_hook(&mut self, input: &mut RawInput) {}
```
emilk: nit: input usually comes before output :)

crates/egui/src/id.rs
lucasmerlin: I use it in the accesskit inspector to convert the accesskit::Id back to an egui Id. Without this its impossible to get the egui response based on the accesskit id.

I could mark it as doc(hidden) and rename it to `dangerous_from_value`?
Also it should have of course been part of that PR, I'll remove it from this one.

emilk: Should maybe me marked `unsafe` too, though it's a slight misuse of the term. `unsafe fn from_high_entropy_bits`

crates/egui/src/plugin.rs
```diff
@@ -0,0 +1,237 @@
...
+    /// Called just before the output is passed to the backend.
+    ///
+    /// Useful to inspect or modify the output.
+    /// Since this is called outside a pass, don't show ui here.
+    fn output_hook(&mut self, output: &mut FullOutput) {}
+}
+
+pub(crate) struct PluginHandle {
+    plugin: Box<dyn std::any::Any + Send + Sync>,
+    get_plugin_mut: fn(&mut Self) -> &mut dyn Plugin,
```
emilk: Why not

```suggestion
    plugin: Box<dyn Plugin>,
```

?

Is it to avoid `.as_any`? But we have trait upcasting now, so that should be solvable with `Plugin: Any`… in either case: document the _why_.

crates/egui/src/id.rs
lucasmerlin: Added a `from_high_entropy_bits` in the accesskit inspector PR.

crates/egui/src/plugin.rs
lucasmerlin: I did not use trait upcasting since we're still on 1.85. Happy to update to 1.86 though, which is the version that introduced trait upcasting.

emilk: Ah, yeah - let's do that in a separate PR then

lucasmerlin: Updated to use trait upcasting, much nicer now!

crates/egui/src/context.rs
```diff
@@ -33,15 +25,23 @@ use crate::{
...
+    TessellationOptions, TextureAtlas, TextureId, Vec2,
+    emath::{self, TSTransform},
+    mutex::RwLock,
+    stats::PaintStats,
+    tessellator,
+    text::{FontInsert, FontPriority, Fonts},
+    vec2,
+};
+use std::any::TypeId;
+use std::{borrow::Cow, cell::RefCell, panic::Location, sync::Arc, time::Duration};
```
emilk: nit: In most places we try to follow this include order:

* std::
* 3rdparty::
* 1stparty::

emilk (approved):

PR #7386, opened 2025-07-23
Group AccessKit nodes by `Ui`

lucasmerlin:


crates/egui/src/context.rs
```diff
@@ -590,8 +591,28 @@ impl ContextImpl {
         let builders = &mut state.nodes;
         if let std::collections::hash_map::Entry::Vacant(entry) = builders.entry(id) {
             entry.insert(Default::default());
-            let parent_id = state.parent_stack.last().unwrap();
-            let parent_builder = builders.get_mut(parent_id).unwrap();
+
+            /// Search the first parent that has an existing accesskit node.
+            fn find_parent_recursively(
```
```suggestion
            /// Find the first ancestor that already has an accesskit node.
            fn find_accesskit_parent(
```

crates/egui/src/context.rs
```diff
@@ -2470,6 +2491,74 @@ impl ContextImpl {
...
+                // let mut root_children = vec![];
+                //
+                // for id in state.nodes.keys().copied().collect::<Vec<_>>() {
+                //     let mut children = vec![];
+                //     children_recursively(id, &state.nodes, &state.child_map, &mut children, &mut root_children);
+                //     state.nodes.get_mut(&id).unwrap().set_children(children)
+                // }
+                //
+                // state.nodes.get_mut(&crate::accesskit_root_id()).unwrap().set_children(root_children);
+
```
emilk: remove dead code

crates/egui/src/context.rs
```diff
@@ -3487,6 +3543,22 @@ impl Context {
         })
     }
 
+    #[cfg(feature = "accesskit")]
+    pub(crate) fn register_accesskit_parent(&self, id: Id, parent_id: Id) {
+        self.write(|ctx| {
+            if let Some(state) = ctx.viewport().this_pass.accesskit_state.as_mut() {
+                state.parent_map.insert(id, parent_id);
+                state.child_map.entry(id).or_default().parent = Some(parent_id);
```
emilk: Why do we need both of these? Looks like just duplicated state?

Isn't `parent_map[foo] == state.child_map[foo].parent` always ?

crates/egui/src/pass_state.rs
```diff
@@ -71,7 +71,15 @@ impl ScrollTarget {
 #[derive(Clone)]
 pub struct AccessKitPassState {
     pub nodes: IdMap<accesskit::Node>,
-    pub parent_stack: Vec<Id>,
+    pub parent_map: IdMap<Id>,
+    pub child_map: IdMap<AccessKitParentChildren>,
```
emilk: Same comment again: seems like duplicated state

crates/egui/src/ui.rs
```diff
@@ -194,6 +202,11 @@ impl Ui {
             ui.set_invisible();
         }
 
+        #[cfg(feature = "accesskit")]
+        ui.ctx().accesskit_node_builder(ui.unique_id, |node| {
```
emilk, line 206: I think it's time we remove the "accesskit" feature, and just always depend on accesskit:

* https://github.com/emilk/egui/issues/3137#issuecomment-1706549867

That is: replace the compile-time flag with a runtime flag. (not in this PR though)

emilk (approved):

PR #7392, opened 2025-07-26
Change type of corner radius to f32 inside RectShape

BeastLe9enD:
Makes it possible to draw rect shapes with better precision. For example, in my case, I am drawing a rect with scaling via `TSTransform` and I want my rect to preserve the same corner radius at different scale solutions.

### DIFF
--- a/crates/egui/src/style.rs
+++ b/crates/egui/src/style.rs
@@ -3,7 +3,7 @@
 #![allow(clippy::if_same_then_else)]
 
 use emath::Align;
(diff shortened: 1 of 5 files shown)


emilk:
So the rounding to closest integer is visible when you zoom in and out? Can you maybe show a video demonstrating this?

The reason we use u8 is to reduce the size of the `RectShape` (see the failing tests), but there are other solutions we could consider (like F16)

Issue #7391, opened 2025-07-26
Minor documentation update


Resonanz:
On this page under Usage: https://docs.rs/egui/latest/egui/containers/frame/struct.Frame.html

none is "Deprecated: Use Frame::NONE or Frame::new() instead."

```
egui::Frame::none()
    .fill(egui::Color32::RED)
    .show(ui, |ui| {
        ui.label("Label with red background");
    });
```

PR #7396, opened 2025-07-29
Use `Frame::NONE` in docstring

satnam72:
- Replaced deprecated `egui::Frame::none()` with `egui::Frame::NONE`

* Closes <https://github.com/emilk/egui/issues/7391>


crates/egui/src/containers/frame.rs
```diff
@@ -46,10 +46,9 @@ use epaint::{Color32, CornerRadius, Margin, MarginF32, Rect, Shadow, Shape, Stro
 ///
 /// ```
 /// # egui::__run_test_ui(|ui| {
-/// egui::Frame::none()
-///     .fill(egui::Color32::RED)
+/// egui::Frame::NONE
 ///     .show(ui, |ui| {
-///         ui.label("Label with red background");
```
lucasmerlin: Why did you remove the `.color` though? I feel like having a totally-empty frame is a bad example because then why have a `Frame` at all?

satnam72: Hi @lucasmerlin, based on my knowledge, unlike `none()` (which is a function), `NONE` is constant in Frame implementation. If we want to use color `new()` or `default()` would be a better choice.

YgorSouza: When you use a constant in code, you're creating a local copy of that constant, so you can mutate it just like any other variable. This is the difference between `const` and `static` in Rust. If `Frame::NONE` was static, you would have to clone it in order to mutate it. But since it is a `const`, using `Frame::NONE` is no different from using `Frame::new()` in terms of what you can do with it. But if you find that disturbing, you could just put `Frame::new()` in the example? It's what the other methods use internally anyway.

satnam72: @YgorSouza Thanks for the clarification. I’m new to Rust, but I’ll definitely look into it. In most other languages, it’s generally considered best practice to keep constants the same, if I’m not mistaken but I got your point. We are not manipulating the constant itself but its copy. I’ll make the necessary changes soon.

satnam72:
Hi @lucasmerlin, I’ve addressed the requested changes — could you please review and merge if everything looks good? 

Thanks!

crates/egui/src/containers/frame.rs
```diff
@@ -46,10 +46,10 @@ use epaint::{Color32, CornerRadius, Margin, MarginF32, Rect, Shadow, Shape, Stro
 ///
 /// ```
 /// # egui::__run_test_ui(|ui| {
-/// egui::Frame::none()
+/// egui::Frame::NONE
 ///     .fill(egui::Color32::RED)
 ///     .show(ui, |ui| {
-///         ui.label("Label with red background");
+///          ui.label("Label with red background");
```
emilk:
```suggestion
///         ui.label("Label with red background");
```

PR #7399, opened 2025-07-30
Improve debug_assert about useless add_space() in grid layout

MichaelGrupp:
`Placer::advance_cursor()` is only used by `Ui::add_space()`, so we can move the assert there to make it clearer to understand for an user who accidentally does this.

* ~Closes <https://github.com/emilk/egui/issues/THE_RELEVANT_ISSUE>~


crates/egui/src/ui.rs
```diff
@@ -1877,13 +1877,15 @@ impl Ui {
...
     /// The direction is dependent on the layout.
+    /// Note that `add_space` isn't supported when in a grid layout.
     ///
     /// This will be in addition to the [`crate::style::Spacing::item_spacing`]
     /// that is always added, but `item_spacing` won't be added _again_ by `add_space`.
     ///
     /// [`Self::min_rect`] will expand to contain the space.
     #[inline]
     pub fn add_space(&mut self, amount: f32) {
+        debug_assert!(!self.is_grid(), "add_space makes no sense in a grid layout");
```
emilk, line 1888: This is a good addition for an improved error message, but I suggest we revert the changes to `crates/egui/src/placer.rs` (i.e. also keep the old assert), in case we start using `advance_cursor` for something else

MichaelGrupp: Makes sense, reverted that and adapted the doc comment of `advance_cursor` for consistency. Reworded commit / PR message.

emilk (approved):

Issue #7397, opened 2025-07-29
Multi-line `TextShape` rotation newly broken in 0.32


scvalex:
**Describe the bug**

I am trying to paint multi-line text with rotation:

```rust
let (_response, painter) =
    ui.allocate_painter(ui.available_size_before_wrap(), Sense::empty());
let mut layout_job = LayoutJob::default();
let text_color = ui.visuals().text_color();
layout_job.append(
    "A short first line\nA slightly longer second line\nCould this be a third line?",
    0.0,
    TextFormat {
        font_id: FontId::proportional(20.0),
        color: text_color,
        ..Default::default()
    },
);
let galley = painter.layout_job(layout_job);
let text_shape = epaint::TextShape::new(Pos2::new(200.0, 200.0), galley, text_color)
    .with_angle(PI / 2.0);
painter.add(text_shape);
```

Full repro repo: https://github.com/scvalex/demo-egui-text-rotation-bug

In egui-0.31, this looks like this:

<img width="102" height="278" alt="Image" src="https://github.com/user-attachments/assets/fba0f44c-1b6e-4895-b2b4-4d185eaf944e" />

In egui-0.32 and on main, it looks like this:

<img width="55" height="294" alt="Image" src="https://github.com/user-attachments/assets/b36ef817-1c40-44e8-9b0f-68e61dd1ea43" />

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: n/a (but also happens on Firefox when compiled to wasm)
 - Version: NixOS 25.05 (Warbler)

scvalex:
Bisecting leads me to 557bd56e1962266e765cc7b7958c1fd3f14fa3e1 of #5411 .

afishhh:
Sorry for breaking you and thanks for the nice reproducer, I found what's causing this and will make a PR right now.

scvalex:
> Sorry for breaking you

No worries at all—this is how progress is made after all.  And thank you!

scvalex:
Thank you!

PR #7404, opened 2025-08-02
Fix multi-line `TextShape` rotation

afishhh:
* Closes <https://github.com/emilk/egui/issues/7397>
I do admit I got a peak NixOS `RequestDeviceError` and deemed it entirely not worth it to think about that.

https://github.com/emilk/egui/pull/5411 broke rotation of multi-line `TextShape`s because `PlacedRow::pos` was no longer being rotated, so let's rotate it.

### DIFF
--- a/crates/epaint/src/tessellator.rs
+++ b/crates/epaint/src/tessellator.rs
@@ -2031,7 +2031,12 @@ impl Tessellator {
                 continue;
             }
 
-            let final_row_pos = galley_pos + row.pos.to_vec2();
+            let final_row_pos = galley_pos
+                + if *angle != 0.0 {
+                    rotator * row.pos.to_vec2()
+                } else {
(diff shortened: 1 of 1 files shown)


crates/epaint/src/tessellator.rs
```diff
@@ -2031,7 +2031,12 @@ impl Tessellator {
                 continue;
             }
 
-            let final_row_pos = galley_pos + row.pos.to_vec2();
+            let final_row_pos = galley_pos
+                + if *angle != 0.0 {
+                    rotator * row.pos.to_vec2()
+                } else {
+                    row.pos.to_vec2()
+                };
```
lucasmerlin:
```suggestion
            let final_row_pos = galley_pos + rotator * row.pos.to_vec2();
```

PR #7409, opened 2025-08-03
Fix: prevent calendar popup from closing on dropdown change

AStrizh:
Currently, DatePickerButton will close without saving whenever a user clicks a dropdown from year/month/date. The issue is caused because the system mistakenly interprets the user as clicking off of the calendar. This is unexpected and creates an unpleasant experience for the user. This change now allows the user to use the dropdowns as expected; it will close on save or cancel. The calendar still closes when user clicks off of it, as before. The changes here are made in:  
crates/egui_extras/src/datepicker/button.rs

I will admit that I am not an experienced Rust developer. The changes were made with the help of ChatGPT 4.0. 
I have tested the changes locally, as I am using the date picker in my project.

### DIFF
--- a/crates/egui_extras/src/datepicker/button.rs
+++ b/crates/egui_extras/src/datepicker/button.rs
@@ -192,8 +192,11 @@ impl Widget for DatePickerButton<'_> {
                 button_response.mark_changed();
             }
 
+            // We don't want to close our popup if any other popup is open, since other popups would
+            // most likely be the combo boxes in the date picker.
+            let any_popup_open = ui.ctx().is_popup_open();
             if !button_response.clicked()
-                && !ui.ctx().is_popup_open()
+                && !any_popup_open
                 && (ui.input(|i| i.key_pressed(Key::Escape)) || area_response.clicked_elsewhere())
             {
                 button_state.picker_visible = false;
--- a/crates/egui_extras/src/datepicker/button.rs
+++ b/crates/egui_extras/src/datepicker/button.rs
@@ -138,16 +138,12 @@ impl Widget for DatePickerButton<'_> {
             button = button.fill(visuals.weak_bg_fill).stroke(visuals.bg_stroke);
         }
         let mut button_response = ui.add(button);
-        let just_opened = if button_response.clicked() {
+        if button_response.clicked() {
             button_state.picker_visible = true;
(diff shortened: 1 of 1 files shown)


emilk:
Needs `cargo fmt` (and a review from @lucasmerlin) :)

crates/egui_extras/src/datepicker/button.rs
```diff
@@ -192,8 +194,10 @@ impl Widget for DatePickerButton<'_> {
                 button_response.mark_changed();
             }
 
-            if !button_response.clicked()
-                && (ui.input(|i| i.key_pressed(Key::Escape)) || area_response.clicked_elsewhere())
+            let escape_pressed = ctx.input(|i| i.key_pressed(Key::Escape));
+            let pointer_over_popup = ctx.is_pointer_over_area();
```
lucasmerlin: This will not check if the pointer is over the popup but if it's over the whole egui area. Which shouldn't be relevant here?
This means I can close the popup with esc - but only when I move my mouse out of the egui window 😄

So the better fix would be to just remove the `area_response.clicked_elsewhere()`

crates/egui_extras/src/datepicker/button.rs
```diff
@@ -138,12 +138,16 @@ impl Widget for DatePickerButton<'_> {
...
         let mut button_response = ui.add(button);
-        if button_response.clicked() {
+        let just_opened = if button_response.clicked() {
             button_state.picker_visible = true;
             ui.data_mut(|data| data.insert_persisted(id, button_state.clone()));
-        }
+            true
+        } else {
+            false
+        };
```
lucasmerlin: I think this is nicer:
```suggestion
        let just_opened = button_response.clicked();
        if just_opened {
            button_state.picker_visible = true;
            ui.data_mut(|data| data.insert_persisted(id, button_state.clone()));
        };
```

crates/egui_extras/src/datepicker/button.rs
```diff
@@ -138,12 +138,16 @@ impl Widget for DatePickerButton<'_> {
...
 
         if button_state.picker_visible {
+            let ctx = ui.ctx();
```
lucasmerlin: You can revert this

lucasmerlin:
I'll take this over so we can get it into 0.32.2

emilk (approved):

PR #7414, opened 2025-08-05
Fix: use unique id for resize columns in `Table`

zezic:
Currently the IDs for resize columns in Table are based on the ID of parent. When placing multiple tables within the same parent the ID clash for resize columns occurs despite the [TableBuilder::id_salt](https://docs.rs/egui_extras/0.32.0/egui_extras/struct.TableBuilder.html#method.id_salt) is being used.


emilk:
Thanks!

PR #7421, opened 2025-08-06
Enable `clippy::iter_over_hash_type` lint

emilk:
This helped discover a few things that _might_ have been buggy.


lucasmerlin (approved):
I'm curious if the changes to BtreeMap will have a slight positive performance impact

emilk:
In most cases the map of viewports has length=1, so I doubt performance will see much impact.

PR #7426, opened 2025-08-07
Enable and fix some more clippy lints

emilk:
One can never have too many lints


crates/egui/src/containers/popup.rs
```diff
@@ -458,7 +458,7 @@ impl<'a> Popup<'a> {
 
     /// Get the expected size of the popup.
     pub fn get_expected_size(&self) -> Option<Vec2> {
-        AreaState::load(&self.ctx, self.id).and_then(|area| area.size)
```
lucasmerlin, line 461: This is a really nice lint

emilk: Yes 😍

PR #7428, opened 2025-08-07
fix: `SubMenu` should not display when ui is disabled

ozwaldorf:
Adds a check for `ui.is_enabled()` when displaying the SubMenu items. There were a few places the condition could be placed, but I figured there was simplest. The inner button will already not be able to be clicked due to ui being disabled.

cc @lucasmerlin 

### DIFF
--- a/.github/workflows/cargo_machete.yml
+++ b/.github/workflows/cargo_machete.yml
@@ -8,11 +8,11 @@ jobs:
     steps:
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: 1.85
+          toolchain: 1.86
       - name: Machete install
-        run: cargo install cargo-machete --locked
+        ## The official cargo-machete action
+        uses: bnjbvr/cargo-machete@main
       - name: Checkout
         uses: actions/checkout@v4
       - name: Machete Check
         run: cargo machete
-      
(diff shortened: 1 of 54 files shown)


ozwaldorf:
> @lucasmerlin It'd be awesome if you could add a disabled submenu to the popups demo.

Done!

crates/egui/src/containers/menu.rs
```diff
@@ -446,7 +446,7 @@ impl SubMenu {
         let is_hovered = hover_pos.is_some_and(|pos| button_rect.contains(pos));
 
         // The clicked handler is there for accessibility (keyboard navigation)
-        if (!is_any_open && is_hovered) || button_response.clicked() {
+        if (!is_any_open && is_hovered && ui.is_enabled()) || button_response.clicked() {
```
emilk: This looks like it will still open when clicked, enabled or not (though I guess there is another path preventing it from becoming `clicked` when disabled).

```suggestion
        let should_open = ui.is_enabled() && (button_response.clicked() || (is_hovered && !is_any_open));
        if should_open {
```

emilk (approved):

ozwaldorf:
Rebased onto latest main

crates/egui/src/containers/menu.rs
ozwaldorf: Yeah, as mentioned the button would never return click when ui is enabled. However I like this short circuit for ui disabled!

ozwaldorf:
cc @emilk rebased again and fix the fmt error. Should be g2g now finally :sweat_smile:

PR #7430, opened 2025-08-07
Skip zero-length layout job sections

### LINKED ISSUE #7378: Panic on multiline `LayoutJob` when segment has empty string
The following code panics in debug mode with the message `Bad byte_range: 1..1`:
```rs
let mut job = egui::text::LayoutJob::default();
job.append("\n", 0.0, egui::TextFormat::default());
job.append("", 0.0, egui::TextFormat::default());
ui.label(job);
```

The problem is in `layout_each_paragraph_individually()` with the "Section range overlaps with paragraph range" assert.

This can be fixed by changing the assert from `section_range.start < section_range.end` to `section_range.start <= section_range.end`, or by changing the `else` to `else if section_range.start < section_range.end` (to avoid doing zero-length slices anyway).

HactarCE:
Fixes #7378 

Includes a regression test that previously failed and now succeeds.

### DIFF
--- a/crates/epaint/src/text/fonts.rs
+++ b/crates/epaint/src/text/fonts.rs
@@ -927,7 +927,7 @@ impl GalleyCache {
                     current_section += 1;
                 } else if end < section_range.start {
                     break; // Haven't reached this one yet.
-                } else {
+                } else if section_range.start < section_range.end {
                     // Section range overlaps with paragraph range
                     debug_assert!(
                         section_range.start < section_range.end,


crates/epaint/src/text/fonts.rs
```diff
@@ -927,7 +927,7 @@ impl GalleyCache {
                     current_section += 1;
                 } else if end < section_range.start {
                     break; // Haven't reached this one yet.
-                } else {
+                } else if section_range.start < section_range.end {
                     // Section range overlaps with paragraph range
                     debug_assert!(
                         section_range.start < section_range.end,
```
emilk: I believe this is a better fix, or any `leading_space` of a first empty paragraph will be ignored

```suggestion
                } else {
                    // Section range overlaps with paragraph range
                    debug_assert!(
                        section_range.start <= section_range.end,
```

PR #7431, opened 2025-08-07
More even text kerning

valadaptive:
Contains several cleanups in the font loading and rendering code. Some of them were ported from the Parley PR, some are new. A lot of them center around passing in state (like font size and DPI) instead of storing it.

- Stop storing `FontsImpl` in an `Arc<Mutex<..>>`. We never need to clone it, and its only purpose seems to be allowing the `Fonts` type to make use of caching while still using a non-mutable `&self` as a receiver. That doesn't seem to pull its own weight. We do now have to add a `ctx.fonts_mut` method though.
- Remove the `FontTweak::baseline_offset_factor` property. I cannot construct any scenario where it behaves differently than `y_offset_factor`.
- Make the `Font` type non-scaled (independent of font size and pixels_per_point). Most of its methods don't depend on its scale, and for those that do, we can just pass the scale in. This reduces the number of fonts we need to cache at once.
- Stop storing `FontImpl` and `TextureAtlas` in an `Arc<Mutex<..>>` as well. We do this by mapping each font face to an ID, and storing each `FontImpl` by ID. This means that types which reference a font face can just store the ID instead of an `Arc<FontImpl>`. We also now just pass in the `TextureAtlas` to all methods that render into it. To facilitate the same API as we had previously, `Font` is now a view type which stores references to the "fonts by ID" map, font family, and texture atlas. This means we can only materialize one `Font` at a time, but that should be okay.
- Stop storing `pixels_per_point` in `Fonts` and `FontsImpl`. Instead, pass it into all methods that use it. `ctx.fonts` (and the new `ctx.fonts_mut`) return a new view type (`FontsView`) which *does* store the context's current `pixels_per_point`. This lets us get rid of the hack where we have to recreate the font atlas every frame if we have mixed-DPI viewports.


valadaptive:
Looks like the "42 years" button in the Code Example window is a pixel shorter, which is causing tests to fail. Not sure if that's a big deal or not.

crates/epaint/src/text/font.rs
```diff
@@ -34,108 +36,93 @@ impl UvRect {
...
 }
 
-#[derive(Clone, Copy, Debug, PartialEq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub struct GlyphInfo {
     /// Used for pair-kerning.
     ///
     /// Doesn't need to be unique.
     /// Use `ab_glyph::GlyphId(0)` if you just want to have an id, and don't care.
     pub(crate) id: ab_glyph::GlyphId,
```
lucasmerlin: From this comment it seems like a `GlyphId` of 0 is fine, but in allocate_glyph, there is a assert that it's non-0:

```rs
        assert!(glyph_info.id.0 != 0, "Can't allocate glyph for id 0");
```

crates/epaint/src/text/fonts.rs
```diff
@@ -407,6 +404,88 @@ impl FontDefinitions {
...
+static KEY_COUNTER: AtomicU64 = AtomicU64::new(1);
+
+impl FontFaceKey {
+    fn new() -> Self {
+        Self(KEY_COUNTER.fetch_add(1, Ordering::Relaxed))
+    }
+}
+
+impl nohash_hasher::IsEnabled for FontFaceKey {}
+
```
lucasmerlin: In the egui Id we have the requirement that Ids have good entropy (see this recent comment: https://github.com/emilk/egui/pull/7385#discussion_r2254283988) (which I believe is a requirement for IdMap to have good performance, and the IdMap is basically an `nohash_hasher::IntMap`)

So maybe a regular HashMap or BtreeMap would be better here? 

On the other hand, nowhere in the nohash_hasher docs it's mentioned that there's a requirement for good entropy when using it. The [example](https://docs.rs/nohash-hasher/latest/nohash_hasher/type.IntMap.html) even uses keys 0 and 1. Maybe @emilk can have a look if this is fine?

emilk:
Let's make sure we test this with two different viewports on screens with different pixels_per_point

lucasmerlin (changes requested):
Awesome!

I was curious if this would have a noticable effect on performance, and while cached performance indeed improves by 6% (nice!) uncached performance is now 3x slower 😅:

Screenshot 2025-08-08 at 12 42 37

Profiling the benchmark shows that now way more time is spent calculating the ascent, so that seems to be a expensive operation:

On main:
Screenshot 2025-08-08 at 12 43 57

On your branch:
Screenshot 2025-08-08 at 12 43 45

So it probably makes sense to precalculate / cache the ascent values (like it was done previously).

lucasmerlin:
> Let's make sure we test this with two different viewports on screens with different pixels_per_point

Works! Also tested having the windows on different screens with different pixels per point.

Screenshot 2025-08-08 at 12 56 51

lucasmerlin:
Also gave this a try in rerun, didn't notice anything major but interestingly the help view text shifted down a row of pixels:

https://github.com/user-attachments/assets/23321db8-7182-48bd-acbe-27f62a08ef09

I guess it's the same as the "42 years" in the egui demos, a rounding error. Also to me it looks more aligned now, so yay?

crates/epaint/src/text/fonts.rs
emilk: One has to be careful with using a`nohash_hasher`, since it doesn't do any hashing (as the name implies). This can lead to hash collisions and bad performance (and even DOS attacks).

Of course, this only matters if there are a lot of keys, so the linked example with only two keys is fine. But this would likely lead to horrible performance:

```rs
let mut m = IntMap<in32, bool>::default():
for i in 0..1000 {
	m.insert(i << 24, false);
```

Because the lower 24 bits are all the same, we get hash collisions on each insert, potentially leading to O(N^2) performance.

---

The above use of `FontFaceKey` will _probably_ not cause any real problems, but I'd rather not risk it unless we're sure. `ahash::HapMap` is also fast, and safe.

crates/epaint/src/text/font.rs
```diff
@@ -34,108 +36,93 @@ impl UvRect {
...
 
-    /// Unit: points.
-    pub advance_width: f32,
+    /// Unscaled.
+    pub advance_width_unscaled: OrderedFloat<f32>,
```
emilk:
```suggestion
    /// In [`ab_glyph`]s "unscaled" coordinate system.
    pub advance_width_unscaled: OrderedFloat<f32>,
```

crates/epaint/src/text/text_layout.rs
```diff
@@ -400,6 +414,7 @@ fn replace_last_glyph_with_overflow_character(
     fonts: &mut FontsImpl,
     job: &LayoutJob,
     row: &mut Row,
+    pixels_per_point: f32,
```
emilk: nit: egui tries to order its arguments in a logical way, with the first argument being "context-like", e.g. things that are passed down through many hierarchies unchanged (in this case: `fonts, pixels_per_point`), while the latter arguments are the the things that are more likely to be specific to the function (in tis case: `row`)

emilk:
> I was curious if this would have a noticable effect on performance, and while cached performance indeed improves by 6% (nice!) uncached performance is now 3x slower 😅:
> 
> So it probably makes sense to precalculate / cache the ascent values (like it was done previously).

The new code in this PR uses a lot of `.pt_scaled` to scale the font on-demand, and then call e.g. `.ascent` on the result.

What we would like to do it is to only run `pt_scaled` once for each font_size, and then cache the resulting ascent/descent/line_gap etc.

But that brings us back to having a scale baked into `FontImpl`, or splitting out a new `ScaledFontImpl` that wraps `FontIml` and caches the ascent/descent/line_gap etc.

The best place to fix this is probably deep down in `ab_glyph` 😭

crates/epaint/src/text/font.rs
```diff
@@ -446,43 +403,61 @@ impl Font {
...
+        u64::from_le_bytes([
+            dpi_bits[0],
+            dpi_bits[1],
+            dpi_bits[2],
+            dpi_bits[3],
+            scale_bits[0],
+            scale_bits[1],
+            scale_bits[2],
+            scale_bits[3],
+        ])
```
emilk:
```suggestion
        (self.pixels_per_point.to_bits() as u64).shl(32) | (self.px_scale_factor as u64)
```

crates/epaint/src/text/font.rs
```diff
@@ -45,97 +68,46 @@ pub struct GlyphInfo {
...
 
-/// A specific font with a size.
+/// A specific font face.
 /// The interface uses points as the unit for everything.
 pub struct FontImpl {
     name: String,
     ab_glyph_font: ab_glyph::FontArc,
+    tweak: FontTweak,
+    glyph_info_cache: ahash::HashMap<char, GlyphInfo>,
+    glyph_alloc_cache: ahash::HashMap<(ab_glyph::GlyphId, u64), GlyphAllocation>,
```
emilk: This could me made more readable _and_ faster with:

```suggestion
    glyph_alloc_cache: nohash_hasher::IntMap<GlyphCacheKey, GlyphAllocation>,
```

```rs
#[derive(Hash, PartialEq, Eq))
struct GlyphCacheKey(u64);

impl nohash_hasher::IsEnabled for GlyphCacheKey {}

impl GlyphCacheKey {
    fn new(glyph_id: ab_glyph::GlyphId, pixels_per_point: f32, px_scale_factor: f32) - > Self {
        debug_assert!(
             0.0 < pixels_per_point && pixels_per_point.is_finite(),
            "Bad pixels_per_point {pixels_per_point}"
        );
        debug_assert!(
            0.0 < px_scale_factor && px_scale_factor.is_finite(),
            "Bad px_scale_factor: {px_scale_factor}"
        );
        Self(crate::util::hash((glyph_id, pixels_per_point.to_bits(), px_scale_factor.to_bits())))
    }
}
```

(replacing `glyph_cache_key`)

crates/epaint/src/text/font.rs
```diff
@@ -197,91 +170,132 @@ impl FontImpl {
...
     ) -> f32 {
-        use ab_glyph::{Font as _, ScaleFont as _};
-        self.ab_glyph_font
-            .as_scaled(self.scale_in_pixels as f32)
-            .kern(last_glyph_id, glyph_id)
-            / self.pixels_per_point
+        // Round to an even number of physical pixels to get even kerning.
+        // See https://github.com/emilk/egui/issues/382
+        self.ab_glyph_font.kern_unscaled(last_glyph_id, glyph_id) * metrics.px_scale_factor
+            / metrics.pixels_per_point
```
emilk: The comment makes it sound like this should be 

```rs
        let kerning_px = self.ab_glyph_font.kern_unscaled(last_glyph_id, glyph_id) * metrics.px_scale_factor;
        // Round to physical pixels to get even kerning.
        // See https://github.com/emilk/egui/issues/382
        kerning_px.round() / metrics.pixels_per_point
```

?

crates/epaint/src/text/font.rs
```diff
@@ -197,91 +170,132 @@ impl FontImpl {
...
-    /// Unit: points.
-    #[inline(always)]
-    pub fn ascent(&self) -> f32 {
-        self.ascent
+        let y_offset_points = ((font_size * self.tweak.scale * self.tweak.y_offset_factor)
+            + self.tweak.y_offset)
+            .round_ui();
+
+        // Round to closest pixel:
+        let y_offset_in_points = (y_offset_points * pixels_per_point).round() / pixels_per_point;
```
emilk: 🤷 
```suggestion
        let y_offset_in_points = y_offset_points.round_to_pixels(pixels_per_point);
```

crates/epaint/src/text/font.rs
```diff
@@ -446,43 +403,61 @@ impl Font {
...
-        if let Some(first) = self.fonts.first() {
-            first.ascent()
-        } else {
-            self.row_height
-        }
-    }
+#[derive(Clone, Copy, Debug, PartialEq, Default)]
+pub struct ScaledMetrics {
+    pub pixels_per_point: f32,
+    pub px_scale_factor: f32,
```
emilk, line 537: docstring

crates/epaint/src/text/font.rs
```diff
@@ -446,43 +403,61 @@ impl Font {
...
+    pub y_offset_in_points: f32,
```
emilk, line 540: docstring

crates/epaint/src/text/fonts.rs
```diff
@@ -629,118 +754,75 @@ impl Fonts {
...
-    pub fn has_glyphs(&mut self, font_id: &FontId, s: &str) -> bool {
-        self.font(font_id).has_glyphs(s)
-    }
-
-    /// Height of one row of text in points.
-    ///
-    /// Returns a value rounded to [`emath::GUI_ROUNDING`].
-    fn row_height(&mut self, font_id: &FontId) -> f32 {
-        self.font(font_id).row_height()
+    /// Get the right font implementation from  [`FontFamily`].
```
emilk:
```suggestion
    /// Get the right font implementation from [`FontFamily`].
```

crates/epaint/src/text/text_layout.rs
```diff
@@ -159,36 +173,57 @@ fn layout_section(
 
     let mut last_glyph_id = None;
 
+    let mut last_font: Option<(FontFaceKey, Option<ScaledMetrics>)> = None;
```
emilk:
```suggestion
    // Optimization: only recompute `ScaledMetrics` when `FontFaceKey` changes.
    let mut last_font: Option<(FontFaceKey, Option<ScaledMetrics>)> = None;
```

Could also be:
```suggestion
    // Optimization: only recompute `ScaledMetrics` when `FontFaceKey` changes.
    let mut last_font: (FontFaceKey, ScaledMetrics) = (FontFaceKey::INVALID, Default::default());
```

For simpler code below

crates/epaint/src/text/text_layout.rs
```diff
@@ -159,36 +173,57 @@ fn layout_section(
...
+                    font_impl.allocate_glyph(font.atlas, &font_impl_metrics, glyph_info)
                 }
+                None => Default::default(),
+            };
+
+            if let (Some(font_impl), Some(last_glyph_id)) = (&font_impl, last_glyph_id) {
+                paragraph.cursor_x +=
+                    font_impl.pair_kerning(&font_impl_metrics, last_glyph_id, glyph_alloc.id);
+                paragraph.cursor_x += extra_letter_spacing;
             }
```
emilk: nit: the extra_letter_spacing is not really dependent on finding the fonts… though I'm not sure how we would ever get in a situation of not finding the fonts. (maybe we should early-out in those cases instead… but not in this PR)

```suggestion
            }
            paragraph.cursor_x += extra_letter_spacing;
```

emilk (commented):
Nice! Performance looks good now, and as far as I can tell we haven't regressed on https://github.com/emilk/egui/issues/382

…but the text sizes (or at least letter spacing) has changed a lot!

Is this expected?
How sure are we that the new sizes/spacing is more correct (e.g. more in line with what you would see in a browser)?

## Before

Image

## After

Image

crates/epaint/src/text/text_layout.rs
valadaptive: This is [in the existing code](https://github.com/emilk/egui/blob/34cd6133783009ccfdbf9b87f49d591b846c270b/crates/epaint/src/text/text_layout.rs#L169-L174). It ensures we only apply extra_letter_spacing to glyphs *after* the first one.

emilk:
This is starting to look real good - the kerning is _much_ nicer now with the subpixel binning.

One bug I found though is that the the Tab (\t) character shrunk a lot:

Code Editor old
Code Editor

BTW, you can generate these before/after pictures by running `UPDATE_SNAPSHOTS=true cargo test --all-features --all-targets`

----

**EDIT**: similarly, the "thin space" character is broken

Before/after
image image

Test text:
> 1 234 567 890
> 1 234 567 890

---

**EDIT**: fixed

emilk:
The kerning is indeed **a lot** better, at the cost of slightly more blurry characters in some situations. Well worth the tradeoff imho, especially since it also simplifies some of the code.

### Before / After
widget_gallery_dark_x1 old widget_gallery_dark_x1

emilk:
Even on high-DPI (retina) screens, the kerning is noticeably better:


Before:
retina-before

After:
retina-after

vkahl:
> The kerning is indeed **a lot** better, at the cost of slightly more blurry characters in some situations. Well worth the tradeoff imho, especially since it also simplifies some of the code.

I agree about the kerning, but honestly, to me the blur looks terrible, at least in these particular screenshots! I wouldn't easily say it's worth it (not using a HiDPI display). Especially those "Click me!" buttons look really bad. For me this seems to thwart the recent changes that improved text rendering so much.

This is probably very subjective though, so if nobody agrees with me, it's probably fine :-P

emilk:
@vkahl you are right about the blur - fortunately this was fixed by bumping the text size from 12.5 to an even 13: 
* https://github.com/emilk/egui/pull/7521

See e.g.
https://github.com/emilk/egui/pull/7521/files#diff-471aeabf4530b33f5aabc5b027335feb3a969471cfc238dd7f4c8d7dcdba7c31

Issue #7433, opened 2025-08-08
Wrong sense area for a TextEdit using `vertical_align`


RndUsr123:
**Describe the bug**
The input sensing part of TextEdit (both `singleline` and `multiline` variants) remains at the top of the widget regardless of vertical alignment. The text itself is correctly positioned.

**To Reproduce**
```
ui.add_sized
(
    vec2(100.0,100.0), 
    TextEdit::multiline(&mut String::from("Select me!"))
        .horizontal_align(Align::Center)
        .vertical_align(Align::Max)
        .background_color(Color32::WHITE)
        .text_color(Color32::BLACK)
);
```

**Expected behavior**
TextEdit reacts to events happening where the text is displayed rather than somewhere else.

**Desktop (please complete the following information):**
 - OS: Win10
 - Version 0.32.0

PR #7436, opened 2025-08-08
Fixes sense issues in TextEdit when vertical alignment is used

RndUsr123:
* Closes <https://github.com/emilk/egui/issues/7433>

I'm running a rustup-less rust install on Windows, so I don't have `clippy` nor `fmt` and can't run the .sh script.
It's very little code and I manually tested this, so hopefully that's ok...

Let me know if the comment in `state.rs` needs to be updated or the `text_offset` name isn't clear enough.

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -652,11 +652,10 @@ impl TextEdit<'_> {
             .align_size_within_rect(galley.size(), rect)
             .intersect(rect) // limit pos to the response rect area
             .min;
-        let align_offset_x = rect.left() - galley_pos.x;
-        let align_offset_y = rect.top() - galley_pos.y;
+        let align_offset = rect.left_top() - galley_pos;
 
         // Visual clipping for singleline text editor with text larger than width
-        if clip_text && align_offset_x == 0.0 {
+        if clip_text && align_offset.x == 0.0 {
             let cursor_pos = match (cursor_range, ui.memory(|mem| mem.has_focus(id))) {
                 (Some(cursor_range), true) => galley.pos_from_cursor(cursor_range.primary).min.x,
                 _ => 0.0,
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -653,16 +652,17 @@ impl TextEdit<'_> {
...
         // Visual clipping for singleline text editor with text larger than width
-        if clip_text && align_offset == 0.0 {
+        if clip_text && align_offset_x == 0.0 {
             let cursor_pos = match (cursor_range, ui.memory(|mem| mem.has_focus(id))) {
                 (Some(cursor_range), true) => galley.pos_from_cursor(cursor_range.primary).min.x,
                 _ => 0.0,
             };
 
-            let mut offset_x = state.singleline_offset;
+            let mut offset_x = state.text_offset[0];
```
lucasmerlin:
```suggestion
            let mut offset_x = state.text_offset.x;
```
A bit more idiomatic

crates/egui/src/widgets/text_edit/state.rs
```diff
@@ -51,7 +51,7 @@ pub struct TextEditState {
 
     // Visual offset when editing singleline text bigger than the width.
     #[cfg_attr(feature = "serde", serde(skip))]
```
lucasmerlin: Ah yeah it'd be nice if you could update the comment

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -653,16 +652,17 @@ impl TextEdit<'_> {
             .align_size_within_rect(galley.size(), rect)
             .intersect(rect) // limit pos to the response rect area
             .min;
-        let align_offset = rect.left() - galley_pos.x;
+        let align_offset_x = rect.left() - galley_pos.x;
+        let align_offset_y = rect.top() - galley_pos.y;
```
```suggestion
        let align_offset = rect.left_top() - galley_pos;
```

Issue #7367, opened 2025-07-17
Text inside buttons ignores `override_text_color` in 0.32.0


Aely0:
**Describe the bug**
When `override_text_color` is set in `Visuals`, text inside buttons constructed with strings and string literals ignore the set override color

**To Reproduce**
```rust
let mut visuals = ctx.style().visuals.clone();
visuals.override_text_color = Some(Color32::RED);
ctx.set_visuals(visuals);

// Override text color gets ignored
ui.button("Test text color stuffs");

// Using RichText instead works as expected
ui.button(RichText::new("Test text color stuffs"));
ui.button(RichText::new("Test text color stuffs").color(Color32::GREEN));
```
<img width="298" height="164" alt="Image" src="https://github.com/user-attachments/assets/2b87dd34-a018-4908-bca5-f54f0ef98e7f" />

**Desktop**
Linux Wayland

PR #7439, opened 2025-08-09
Fix override_text_color priority

YgorSouza:
The override_text_color is now used when rendering text from a String or &str. This is consistent with the RichText variant and makes the option behave as advertised, taking precedence over WidgetVisuals and overriding the color for all text unless explicitly changed for a single widget (via RichText or LayoutJob).

* Closes <https://github.com/emilk/egui/issues/7367>

### DIFF
--- a/crates/egui/src/widget_text.rs
+++ b/crates/egui/src/widget_text.rs
@@ -746,11 +746,15 @@ impl WidgetText {
     ) -> Arc<Galley> {
         match self {
             Self::Text(text) => {
+                let color = style
+                    .visuals
+                    .override_text_color
+                    .unwrap_or(crate::Color32::PLACEHOLDER);
                 let mut layout_job = LayoutJob::simple_format(
                     text,
                     TextFormat {
                         font_id: FontSelection::Default.resolve(style),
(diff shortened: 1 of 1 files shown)


YgorSouza:
For the record, here is where `RichText` does the same thing (and why it didn't have the same problem as noted in the issue):

https://github.com/emilk/egui/blob/6fae65a3fabaf442cbbc96a76cbf70fe2fee16ee/crates/egui/src/widget_text.rs#L465-L475

https://github.com/emilk/egui/blob/6fae65a3fabaf442cbbc96a76cbf70fe2fee16ee/crates/egui/src/widget_text.rs#L382

https://github.com/emilk/egui/blob/6fae65a3fabaf442cbbc96a76cbf70fe2fee16ee/crates/egui/src/widget_text.rs#L404

emilk (approved):
It would be great if you could add a simple snapshot test for this, e.g. in `crates/egui_kittest/tests/regression_tests.rs`

crates/egui_kittest/tests/snapshots/override_text_color_interactive.png
emilk, line 1: Nice - thank you! ❤️

Issue #7362, opened 2025-07-15
panic in ScrollArea in egui 0.32.0


hydra:
**Describe the bug**
when clicking a scroll area's resize handle, I get an instant panic caused by a debug assertion.

See also #4478 and #5216.

**To Reproduce**
1. Run MakerPnP planner_egui @ rev `d3f1c30718da433df8c485e01d88d3258ef802ae`
2. Open a project
3. Double click the unit assignments in the treeview.
4. Left click the scrollbar.
5. Observe panic.

**Screenshots**

<img width="815" height="327" alt="Image" src="https://github.com/user-attachments/assets/4169c1f5-b4ad-42be-ac6f-e93465515d33" />

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: N/A

**Smartphone (please complete the following information):**
N/A

**Additional context**

I added debug info to scroll_area.rs:

```rust
let new_handle_top = pointer_pos[d] - *scroll_start_offset_from_top_left;
                println!("new_handle_top: {:?}", new_handle_top);
                println!("scroll_bar_rect: {:?}", scroll_bar_rect);

                let from = scroll_bar_rect.min[d]..=(scroll_bar_rect.max[d] - handle_rect.size()[d]);
                println!("from: {:?}", from);
                state.offset[d] = remap(
                    new_handle_top,
                    from,
                    0.0..=max_offset[d],
                );
```

the output is:

```
new_handle_top: 332.0
scroll_bar_rect: [[481.0 332.0] - [1257.0 372.0]]
from: 332.0..=332.0
```

```
and the panic is:

thread 'main' panicked at D:\Users\Hydra\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\emath-0.32.0\src\lib.rs:152:5:
from.start() and from.end() should not be equal
```

Here's a link to the source:

https://github.com/MakerPnP/makerpnp/blob/51cd78c5c45811e4ff8b030669b3f24edb2d4705/crates/planner_gui_egui/src/project/tabs/unit_assignments_tab.rs#L383-L463

alexdsp:
Easier way to reproduce using the egui_demo_app scrolling with settings as in the image and clicking on the scroll bar.
https://github.com/emilk/egui/blob/main/crates/egui_demo_lib/src/demo/scrolling.rs
<img width="725" height="612" alt="Image" src="https://github.com/user-attachments/assets/df5159f3-e508-4791-881e-7bb518d21cf2" />

[...]

PR #7440, opened 2025-08-09
Fix debug-panic in ScrollArea if contents fit without scrolling

YgorSouza:
If the ScrollArea's contents are smaller than the inner rect, but the scrollbar is set to always visible, clicking on it led to a remap from an empty range to calculate the new offset, which triggered a debug assertion in the remap function, because the result is indeterminate.

Since in this case there is no need to scroll, we just skip the remap and set the offset to 0 directly.

* Closes <https://github.com/emilk/egui/issues/7362>

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -1315,11 +1315,15 @@ impl Prepared {
                     });
 
                 let new_handle_top = pointer_pos[d] - *scroll_start_offset_from_top_left;
-                state.offset[d] = remap(
-                    new_handle_top,
-                    scroll_bar_rect.min[d]..=(scroll_bar_rect.max[d] - handle_rect.size()[d]),
-                    0.0..=max_offset[d],
-                );
+                state.offset[d] = if max_offset[d] > 0.0 {
+                    remap(
+                        new_handle_top,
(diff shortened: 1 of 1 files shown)


crates/egui/src/containers/scroll_area.rs
```diff
@@ -1315,11 +1315,15 @@ impl Prepared {
...
                 let new_handle_top = pointer_pos[d] - *scroll_start_offset_from_top_left;
-                state.offset[d] = remap(
-                    new_handle_top,
-                    scroll_bar_rect.min[d]..=(scroll_bar_rect.max[d] - handle_rect.size()[d]),
-                    0.0..=max_offset[d],
-                );
+                state.offset[d] = if max_offset[d] > 0.0 {
+                    remap(
+                        new_handle_top,
+                        scroll_bar_rect.min[d]..=(scroll_bar_rect.max[d] - handle_rect.size()[d]),
```
emilk: the debug-assert triggers when this range is zero (the "from"-range), so it would be better to check if `scroll_bar_rect.min[d] < scroll_bar_rect.max[d] - handle_rect.size()[d]`

hydra:
this for this PR which closes my issue #7362. I'll be sure to try it out when I get some bandwidth.

❤️

Issue #7389, opened 2025-07-25
Singleline allows `\n` if the user copies it


UnknownSuperficialNight:
**Describe the bug**

Singleline is supposed to not allow any `\n` aka `LINE FEED` character, however it's possible if the user pressing `ctrl+v` to paste it in

**To Reproduce**
Steps to reproduce the behaviour:
1. Use `egui::TextEdit::singleline`
2. Copy a LINE FEED character `
`
3. Use `ctrl+v` to paste it into the singleline

**Expected behaviour**

It should not allow a `LINE FEED` character into the singleline 

**Video**

https://github.com/user-attachments/assets/79bd0fb8-2940-40af-891c-ea3677a20006

**Desktop (please complete the following information):**
 - OS: Gentoo/Linux x86_64 Wayland

**Additional context**

Useful link to the character itself for more details on it 
https://unicode-explorer.com/c/000A

emilk:
I guess the best fix for this is just to strip all `\n` (and `'r`) from a single-line `TextEdit` on each frame

Mingun:
I disagree with that `\n` shouldn't be allowed. They shouldn't be rendered as a newline, but I think, it should be allowed. At least, such behavior should be configurable

YgorSouza:
> I disagree with that `\n` shouldn't be allowed. They shouldn't be rendered as a newline, but I think, it should be allowed. At least, such behavior should be configurable

I found examples of this behavior in some GNOME apps, like RhythmBox and Files. It's what egui currently does, but with the line breaks rendered more clearly.

<img width="671" height="78" alt="Image" src="https://github.com/user-attachments/assets/f7692ac6-72fc-4622-8807-293865ccae66" />

But most other single line inputs I tested (like the Firefox address bar and HTML forms) just change the line breaks to spaces. I found one (Thunderbird) that joins the lines with commas instead, and one (LibreOffice) that just takes the first line and ignores the others entirely.

Maybe the usual web behavior with the spaces would be good enough for egui, at least for now? Eventually some kind of customization could be introduced.

PR #7441, opened 2025-08-09
Remove line breaks when pasting into single line TextEdit

YgorSouza:
The line breaks are now replaced by spaces, matching the usual behavior of text fields in HTML.

* Closes <https://github.com/emilk/egui/issues/7389>

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -945,8 +945,12 @@ fn events(
             Event::Paste(text_to_insert) => {
                 if !text_to_insert.is_empty() {
                     let mut ccursor = text.delete_selected(&cursor_range);
-
-                    text.insert_text_at(&mut ccursor, text_to_insert, char_limit);
+                    if multiline {
+                        text.insert_text_at(&mut ccursor, text_to_insert, char_limit);
+                    } else {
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -945,8 +945,12 @@ fn events(
             Event::Paste(text_to_insert) => {
                 if !text_to_insert.is_empty() {
                     let mut ccursor = text.delete_selected(&cursor_range);
-
-                    text.insert_text_at(&mut ccursor, text_to_insert, char_limit);
+                    if multiline {
+                        text.insert_text_at(&mut ccursor, text_to_insert, char_limit);
+                    } else {
+                        let single_line = text_to_insert.replace(['\r', '\n'], " ");
```
bircni, line 950: what if i type in a text like `C:\Windows\nadadadada` will this then change to `C:\Windows adadada`
This would be pretty bad

bircni (changes requested):

crates/egui/src/widgets/text_edit/builder.rs
YgorSouza: No, it doesn't work like that. You can try the preview above and see. It only replaces the end-of-line characters `0x0A` (line feed) and `0x0D` (carriage return). The escape sequences `\n` and `\r` are just how they are represented in the source code.

PR #7453, opened 2025-08-15
Add rotation gesture support for trackpad sources

thatcomputerguy0101:
This adds support to receive rotation gesture information from sources that don't provide touch events, such as trackpads on MacOS. Currently, only the `winit` backend passes through the necessary information. I would also like to add support for the web backend on Safari, as that generates a `GestureEvent` ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/GestureEvent), [WebKit JS](https://developer.apple.com/documentation/webkitjs/gestureevent)) with rotation information. However, since that is a non-standard API, it is not yet exposed in `web-sys`.

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -492,7 +492,6 @@ impl State {
             WindowEvent::ActivationTokenDone { .. }
             | WindowEvent::AxisMotion { .. }
             | WindowEvent::DoubleTapGesture { .. }
-            | WindowEvent::RotationGesture { .. }
             | WindowEvent::PanGesture { .. } => EventResponse {
                 repaint: false,
                 consumed: false,
@@ -503,6 +502,16 @@ impl State {
                 // Negative delta values indicate shrinking (zooming out).
                 let zoom_factor = (*delta as f32).exp();
                 self.egui_input.events.push(egui::Event::Zoom(zoom_factor));
+                EventResponse {
+                    repaint: true,
+                    consumed: self.egui_ctx.wants_pointer_input(),
+                }
+            }
+
+            WindowEvent::RotationGesture { delta, .. } => {
+                // Positive delta values indicate counterclockwise rotation
+                // Negative delta values indicate clockwise rotation
+                self.egui_input.events.push(egui::Event::Rotate(*delta));
                 EventResponse {
                     repaint: true,
                     consumed: self.egui_ctx.wants_pointer_input(),
--- a/crates/egui/src/data/input.rs
+++ b/crates/egui/src/data/input.rs
@@ -479,6 +479,12 @@ pub enum Event {
     /// As a user, check [`crate::InputState::smooth_scroll_delta`] to see if the user did any zooming this frame.
     Zoom(f32),
 
+    /// Rotation in radians this frame (e.g. from a rotation gesture).
+    ///
+    /// * `rotation > 0`: rotate counterclockwise
+    /// * `rotation < 0`: rotate clockwise
+    Rotate(f32),
+
     /// IME Event
     Ime(ImeEvent),
 
--- a/crates/egui/src/input_state/mod.rs
+++ b/crates/egui/src/input_state/mod.rs
@@ -227,6 +227,12 @@ pub struct InputState {
     /// * `zoom > 1`: pinch spread
     zoom_factor_delta: f32,
 
+    /// Rotation in radians this frame (e.g. from a rotation gesture).
+    ///
+    /// * `rotation > 0`: rotate counterclockwise
+    /// * `rotation < 0`: rotate clockwise
+    rotation_radians: f32,
+
     // ----------------------------------------------
     /// Position and size of the egui area.
     pub screen_rect: Rect,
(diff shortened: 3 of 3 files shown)


thatcomputerguy0101:
`GestureEvent` support in `web-sys`: https://github.com/wasm-bindgen/wasm-bindgen/pull/4589
Alternatively, `js_sys::Reflect` could be used to dynamically extract the fields.

thatcomputerguy0101:
I implemented GestureEvent support utilizing `js_sys::Reflect` since it is a non-standard API. With that, both of `eframe`'s backends now support rotation input in this PR.

crates/eframe/src/web/backend.rs
```diff
@@ -11,9 +11,14 @@ use super::percent_decode;
 /// Data gathered between frames.
 #[derive(Default)]
 pub(crate) struct WebInput {
-    /// Required because we don't get a position on touched
+    /// Required because we don't get a position on touchend
```
bircni, line 14: typo

crates/eframe/src/web/events.rs
```diff
@@ -847,6 +849,72 @@ fn install_wheel(runner_ref: &WebRunner, target: &EventTarget) -> Result<(), JsV
...
+    })?;
+
+    Ok(())
+}
+
+fn handle_gesture(event: web_sys::Event, runner: &mut AppRunner) {
+    // GestureEvent is a non-standard API, so this attempts to get the relevant fields if they exist.
+    let new_scale = Reflect::get(&event, &JsValue::from_str("scale"))
+        .ok()
+        .and_then(|scale| scale.as_f64())
```
bircni, line 873: why dont use f32 here already?

bircni (changes requested):

crates/eframe/src/web/events.rs
thatcomputerguy0101: `JSValue` (the type of `scale`) only has `as_f64`. I assume this is related to how floats are stored for the JS engine.

crates/eframe/src/web/backend.rs
thatcomputerguy0101: I am confused what your intent is here, since the change you commented on was me fixing a typo.

crates/eframe/src/web/events.rs
bircni: ah ok perfect

thatcomputerguy0101:
The `Multi Touch` demo could respond to `smooth_scroll_delta`, `zoom_delta`, and `rotation_delta`. However, all three of those will yield input at the same time as `multi_touch`, which would make the existing processing redundant. Also, the current touches label might be confusing if it is left as reporting zero touches (since no `multi_touch` exists) while the state is changing.

emilk:
We could rename `Multi Touch` to something else, like `Trackpad and Multitouch` and then adapt its UI accordingly 🤔

thatcomputerguy0101:
Keeping it called `Multi Touch` isn't entirely inaccurate. The relevant events are still produced by the same physical actions, just on a trackpad instead of a touch screen. The current touches label could be swapped out to say `Input source: none`, `Input source: 2-finger touch`, or `Input source: cursor` depending on the incoming events.

I have also been working on some changes in `winit` to support multitouch trackpad gestures on Wayland, which will generate `PinchGesture`, `RotateGesture`, and `PanGesture`. Notably, `PanGesture` is the same physical action as `MouseWheel`. Which one is sent depends on if scrolling is detected first or if pinch/rotation is detected first . Therefore, a `PanGesture` from `winit` should probably be mapped to scrolling input in `egui`.

crates/egui_demo_lib/src/demo/multi_touch.rs
```diff
@@ -83,21 +101,39 @@ impl crate::View for MultiTouch {
...
+            //     // the translation we get from `multi_touch` needs to be scaled down to the
+            //     // normalized coordinates we use as the basis for painting:
+            //     self.translation += to_screen.inverse().scale() * multi_touch.translation_delta;
+            //     // touch pressure will make the arrow thicker (not all touch devices support this):
+            //     stroke_width += 10. * multi_touch.force;
+
+            //     self.last_touch_time = ui.input(|i| i.time);
+            // } else {
+            //     self.slowly_reset(ui);
+            // }
```
emilk: remove dead code

crates/egui_demo_lib/src/demo/multi_touch.rs
```diff
@@ -83,21 +101,39 @@ impl crate::View for MultiTouch {
...
+            if ui.input(|i| i.multi_touch().is_some()) || relative_pointer_gesture {
+                ui.input(|input| {
+                    // This adjusts the current zoom factor, rotation angle, and translation according
+                    // to the dynamic change (for the current frame) of the touch gesture:
+                    self.zoom *= input.zoom_delta();
+                    self.rotation += input.rotation_delta();
+                    self.translation += to_screen.inverse().scale()
+                        * input
+                            .multi_touch()
+                            .map_or(input.smooth_scroll_delta, |mt| mt.translation_delta);
```
emilk: I think it would make sense to have a common interface for this, like we do for `.zoom_delta()` and `.rotation_delta()`:


```suggestion
                    self.translation += to_screen.inverse().scale() * input.pan_delta();
```

emilk (approved):
Very nice - works great with my trackpad!

crates/egui_demo_lib/src/demo/multi_touch.rs
thatcomputerguy0101: Would that only have a version using `smooth_scroll_delta`, or should it have a `raw_scroll_delta` variant as well?

emilk: Let's just make one that uses `smooth_scroll_delta` - it's what most people should be using

emilk (approved):

PR #7459, opened 2025-08-15
Bump stable toolchain to 1.88 in all workflows

justincredible:

The workflow is recently failing with:
rustc 1.85.1 is not supported by the following packages:
        cargo-util-schemas@0.8.2 requires rustc 1.86
        cargo_metadata@0.21.0 requires rustc 1.86.0

Using the 1.86 toolchain leads to another error with a link to https://github.com/rust-lang/rust/issues/53667. Works with the 1.88 toolchain.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -156,7 +156,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: EmbarkStudios/cargo-deny-action@v2
         with:
-          rust-version: "1.85.0"
+          rust-version: "1.88.0"
           log-level: error
           command: check
           arguments: --target  ${{ matrix.target }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ members = [
 [workspace.package]
 edition = "2024"
 license = "MIT OR Apache-2.0"
-rust-version = "1.85"
+rust-version = "1.88"
 version = "0.32.1"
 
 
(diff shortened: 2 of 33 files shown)


.github/workflows/cargo_machete.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     steps:
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: 1.85
+          toolchain: 1.88
```
bircni, line 11: we always want to have the same toolchain everywhere - maybe time to switch to cargo-shear?

justincredible: Updated all workflows, which caused an additional issue in the Rust workflow's android and ios jobs where the targets needed to be explicitly downloaded.

justincredible:
The Rust workflow's Run tests job can stall in the Run tests step. Not sure if that was happening before this change, but seen it three times now for the tests:
should_wait_for_images (1x)
widget_tests (2x)

Canceling the workflow run and rerunning the failed job works.

bircni:
Nope we don't want to use the stable one we want to use the one we set as minimal to check that we correctly implement that

I am not happy with this

@lucasmerlin what do you think?

torokati44:
I think the proper way to do this is more extensive, see: https://github.com/emilk/egui/pull/7279

justincredible:
I wasn't looking to raise the project MSRV when I started this, so I'll reopen it as an issue.

Thank you for the feedback!

PR #7460, opened 2025-08-16
Include all parameters in the demo app's layout test window

justincredible:

This change adds the `main_align` and `main_justify` options to the "Layout Test" demo window.
demo_layout_test

Since all `Layout` fields are being used I replaced the `LayoutSettings` struct, but it required uncommenting the serde cfg_attr derives. This was commented out in bf19eb2ec5977a389538cff23b2041b27ec10011 (and the feature name subsequently kept up-to-date), but it's not clear to me why so I can revert that commit and just update the demo struct if necessary.

The button's size was increased to demonstrate the text alignment, and the `Resize::max_size()` works around auto-resizing for `main_justify`.


justincredible:
An alternative solution to the infinite auto-resize would be to use `Ui::allocate_ui_with_layout()` as is often recommended, but that's less desirable in this case as it can result in layouts that make `main_justify` look more broken:
main_justify_allocate_ui

lucasmerlin (changes requested):
Nice! But we should add some note that explains that main_justify breaks when adding more than a single element and then only show a single widget when main_justify is enabled.

justincredible:
https://github.com/user-attachments/assets/28d17937-f32e-4374-add8-4e86428e4742

PR #7471, opened 2025-08-21
Add `SurrenderFocusOn` option

lucasmerlin:

On touch devices you don't want the keyboard to disappear when scrolling, so this PR adds a `SurrenderFocusOn` enum to configure on what interaction to surrender focus.

### DIFF
--- a/crates/egui/src/input_state/mod.rs
+++ b/crates/egui/src/input_state/mod.rs
@@ -20,11 +20,11 @@ use touch_state::TouchState;
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub enum SurrenderFocusOn {
-    /// Surrender focus if the user presses somewhere outside the focused widget.
-    #[default]
+    /// Surrender focus if the user _presses_ somewhere outside the focused widget.
     Presses,
 
-    /// Surrender focus if the user clicks somewhere outside the focused widget.
+    /// Surrender focus if the user _clicks_ somewhere outside the focused widget.
+    #[default]
     Clicks,
 
     /// Never surrender focus.


crates/egui/src/input_state/mod.rs
```diff
@@ -17,6 +18,39 @@ pub use crate::Key;
...
+                .on_hover_text(
+                    "Surrender focus if the user presses somewhere outside the focused widget.",
+                );
+            ui.selectable_value(self, SurrenderFocusOn::Clicks, "Clicks")
+                .on_hover_text(
+                    "Surrender focus if the user clicks somewhere outside the focused widget.",
+                );
+            ui.selectable_value(self, SurrenderFocusOn::Never, "Never")
+                .on_hover_text(
+                    "Never surrender focus, even if the user clicks outside the focused widget.",
```
lucasmerlin:
```suggestion
                    "Never surrender focus.",
```

crates/egui/src/input_state/mod.rs
```diff
@@ -17,6 +17,37 @@ pub use crate::Key;
...
 
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub enum SurrenderFocusOn {
+    /// Surrender focus if the user presses somewhere outside the focused widget.
+    #[default]
+    Presses,
+
+    /// Surrender focus if the user clicks somewhere outside the focused widget.
+    Clicks,
```
emilk: Let's, make Clicks the default

juancampa:
> Let's, make Clicks the default

Defaulting to `SurrenderFocusOn::Clicks` makes egui inconsistent with other platforms. Just wanted to point this out somewhere in case other people hit similar issues.

At least on Chrome (MacOS) as well as some MacOS apps, the behavior seems to be to surrender focus on press (i.e. mouse down), not click.

I found this because it also makes it hard to pass the focus to a new widget. E.g.

```rust
if ui.button("Edit").clicked() {
  editing = true;
  
  // Pass focus to the TextEdit below
  ui.memory_mut(|mem| mem.request_focus(ID));
}

if editing {
  let response = TextEdit::singleline(...).id(ID).show(ui);
  
  assert!(response.has_focus()); <--- Just to illustrate the issue
}
```

In the code above, `TextEdit` immediately surrenders its focus because the button click is detected as a "clicked elsewhere" at the end of `get_response`.

Having said that, I can see how `Clicks` is better in some cases (resize a panel or move a slider while editing text), but it might be surprising for many users.

PR #7479, opened 2025-08-26
Open the menu when hovering over the button

iamazy:
Automatically opening the menu when hovering over the button helps streamline user operations.

output


bircni (changes requested):
I don't think thats always wanted - could you change it so its optional and we can enable it somehow?

iamazy:
> I don't think thats always wanted - could you change it so its optional and we can enable it somehow?

Sounds good, I will try it.

ajaykagrawal:
Is this going to be merged in the next release?

Issue #7371, opened 2025-07-18
Option to enforce a low repaint rate for low-spec systems


s-nie:
**Is your feature request related to a problem? Please describe.**

I am running the web eframe backend on a low-spec system that only has access to a software renderer. Each egui render step takes a significant amount of time, which in turn cannot be used by the browser to run other callbacks, notably for message reception.

**Describe the solution you'd like**

A setting (e.g. in the `WebOptions`) to limit the draw rate to give the browser more time to run other callbacks. I'm imagining that the logic could still run multiple times between redraws, but the tessellation and rendering steps would be skipped until a redraw is needed. Any input events, like mouse movements, could trigger the app logic, without necessarily resulting in a repaint.

**Describe alternatives you've considered**

Tinker with the refresh rate of the browser, which would be rather impractical from a user perspective.

**Additional context**

This is less of an issue in the native builds as those have access to threads.

I could start implementing a solution but first wanted to open the discussion. My knowledge about the web backend is limited, so maybe there are roadblocks or potential solutions I have not considered.

This hack for the repaint logic emulates the desired effect:

```rust
impl NeedRepaint {
    /// Returns the time (in [`now_sec`] scale) when
    /// we should next repaint.
    pub fn when_to_repaint(&self) -> f64 {
        *self.0.lock()
    }

    /// Unschedule repainting.
    pub fn clear(&self) {
        *self.0.lock() = f64::INFINITY;
    }

    pub fn repaint_after(&self, _num_seconds: f64) {
        self.repaint_with_rate(0.2);
    }

    pub fn repaint_with_rate(&self, rate: f64) {
        let mut repaint_time = self.0.lock();
        let next = (super::now_sec() / rate).ceil() * rate;
        *repaint_time = repaint_time.min(next);
    }

    pub fn needs_repaint(&self) -> bool {
        self.when_to_repaint() <= super::now_sec()
    }

    pub fn repaint_asap(&self) {
        self.repaint_with_rate(0.2);
    }
}
```

White-Rabbit-Scientific:
Does this help?

## From Claude AI:

## Manual frame rate control in your app
You can control when to request repaints:

```rust
impl eframe::App for MyApp {
    fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
        // Only repaint when needed
        if self.needs_update {
            ctx.request_repaint();
            self.needs_update = false;
        }
        
        // Or repaint after a delay
[...]

PR #7482, opened 2025-08-28
Add an option to limit the repaint rate in the web runner

s-nie:
* Closes <https://github.com/emilk/egui/issues/7371>

The purpose of this change is to limit the amount of time spent redrawing the GUI, which can be highly beneficial on low spec systems as explained in the issue.

Lowering the repaint rate also highlighted an issue with the drag interaction of windows among other elements. It seems that the drag gets registered on the first mouse down and the window then moves based on the cursor movement since the previous frame, making it the dragged element jump.
The fix I found for this was to always check for `is_decidedly_dragging`. I didn't notice any input delays or other negative effects from this. This is with a refresh rate of 2:

[drag.webm](https://github.com/user-attachments/assets/9a95cdcc-7c81-42c7-b98b-d4164c3b8a9d)

To test this feature, change the `WebOptions` to
```rust
eframe::WebOptions {
    repaint_rate_limit: Some(2),
    ..Default::default()
},
```

### DIFF
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -509,6 +509,10 @@ pub struct WebOptions {
     ///
     /// Defaults to true.
     pub should_prevent_default: Box<dyn Fn(&egui::Event) -> bool>,
+
+    /// Maximum rate at which to repaint. This can be used to artificially reduce the repaint rate below
+    /// vsync in order to save resources.
+    pub repaint_rate_limit: Option<u32>,
 }
 
 #[cfg(target_arch = "wasm32")]
@@ -527,6 +531,8 @@ impl Default for WebOptions {
 
             should_stop_propagation: Box::new(|_| true),
             should_prevent_default: Box::new(|_| true),
+
+            repaint_rate_limit: None,
         }
     }
 }
--- a/crates/eframe/src/web/backend.rs
+++ b/crates/eframe/src/web/backend.rs
@@ -50,11 +50,19 @@ impl WebInput {
 // ----------------------------------------------------------------------------
 
 /// Stores when to do the next repaint.
-pub(crate) struct NeedRepaint(Mutex<f64>);
+pub(crate) struct NeedRepaint {
+    /// Time in seconds when the next repaint should happen.
+    next_repaint: Mutex<f64>,
+    /// Rate limit for repaint. 0 means "unlimited". The rate may still be limited by VSync.
+    rate: u32,
+}
 
-impl Default for NeedRepaint {
-    fn default() -> Self {
-        Self(Mutex::new(f64::NEG_INFINITY)) // start with a repaint
+impl NeedRepaint {
+    pub fn new(rate: Option<u32>) -> Self {
+        Self {
+            next_repaint: Mutex::new(f64::NEG_INFINITY), // start with a repaint
+            rate: rate.unwrap_or(0),
+        }
     }
 }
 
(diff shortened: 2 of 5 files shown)


crates/eframe/src/web/backend.rs
```diff
@@ -50,37 +50,64 @@ impl WebInput {
...
 
-impl Default for NeedRepaint {
-    fn default() -> Self {
-        Self(Mutex::new(f64::NEG_INFINITY)) // start with a repaint
+    /// Rate limit for repaint. 0 means "unlimited". The rate may still be limited by vsync.
+    rate: u32,
+}
+
+impl NeedRepaint {
+    pub fn new(rate: Option<u32>) -> Self {
```
emilk: Let's rename `rate` to `max_fps` for clarity

crates/egui/src/interaction.rs
```diff
@@ -191,20 +191,8 @@ pub(crate) fn interact(
...
-                    // so we postpone the decision until we know.
-                    input.pointer.is_decidedly_dragging()
-                } else {
-                    // This widget is just sensitive to drags, so we can mark it as dragged right away:
-                    widget.sense.senses_drag()
-                };
-
-                if is_dragged {
-                    dragged = Some(widget.id);
-                }
```
emilk: Please split this out to a separate PR 🙏

crates/eframe/src/epi.rs
```diff
@@ -509,6 +509,10 @@ pub struct WebOptions {
     ///
     /// Defaults to true.
     pub should_prevent_default: Box<dyn Fn(&egui::Event) -> bool>,
+
+    /// Maximum rate at which to repaint. This can be used to artificially reduce the repaint rate below
+    /// vsync in order to save resources.
+    pub repaint_rate_limit: Option<u32>,
```
emilk: As noted elsewhere, we should use a different name here to communicate the unit. Maybe `max_fps`?

crates/eframe/src/web/backend.rs
s-nie: Renamed in https://github.com/emilk/egui/pull/7482/commits/4dc0e10d3ee6df97341844d98ea00b670b23c0cd

crates/eframe/src/epi.rs
s-nie: Renamed in https://github.com/emilk/egui/pull/7482/commits/4dc0e10d3ee6df97341844d98ea00b670b23c0cd

crates/egui/src/interaction.rs
s-nie: Moved it to https://github.com/emilk/egui/pull/7500

crates/eframe/src/web/backend.rs
```diff
@@ -50,37 +50,64 @@ impl WebInput {
...
 
     pub fn needs_repaint(&self) -> bool {
         self.when_to_repaint() <= super::now_sec()
     }
 
-    pub fn repaint_asap(&self) {
-        *self.0.lock() = f64::NEG_INFINITY;
+    fn round_repaint_time_to_rate(time: f64, rate: u32) -> f64 {
+        if rate == 0 {
+            f64::NEG_INFINITY
```
emilk:
```suggestion
            time
```

or `repaint_after` won't work properly

crates/eframe/src/web/backend.rs
```diff
@@ -50,37 +50,64 @@ impl WebInput {
...
-        *self.0.lock() = f64::INFINITY;
+        *self.next_repaint.lock() = f64::INFINITY;
     }
 
     pub fn repaint_after(&self, num_seconds: f64) {
-        let mut repaint_time = self.0.lock();
-        *repaint_time = repaint_time.min(super::now_sec() + num_seconds);
+        let mut repaint_time = self.next_repaint.lock();
+        let mut time = super::now_sec() + num_seconds;
+        time = Self::round_repaint_time_to_rate(time, self.max_fps);
```
emilk: wouldn't this make more sense?

```suggestion
        time = self.round_repaint_time_to_rate(time);
```

crates/eframe/src/web/backend.rs
s-nie: I used `NEG_INFINITY` here because `repaint_asap` also uses it. Is the case here different?

crates/eframe/src/web/backend.rs
s-nie: Makes sense, I was worried about the borrow checker first but was able to work around it by moving things. Changed in https://github.com/emilk/egui/pull/7482/commits/46288322baf83052b774216031a8967e5f60a48b.

PR #7488, opened 2025-08-30
Selectively update dependencies to remove total number

oscargus:

Update some of the core dependencies and run cargo update for selected dependencies to remove total number and older versions.


crates/eframe/Cargo.toml
```diff
@@ -271,4 +271,4 @@ wgpu = { workspace = true, optional = true }
 
 # Native dev dependencies for testing
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
-directories = "5"
+directories = "6"
```
emilk, line 274: We should move this to be a workspace dependency

emilk (approved):
Awesome - thank you!

crates/eframe/Cargo.toml
emilk: * https://github.com/emilk/egui/pull/7495

PR #7494, opened 2025-09-03
Fix deadlock in ImageLoader, FileLoader, EhttpLoader

lucasmerlin:
* Recently CI runs started to hang randomly: https://github.com/emilk/egui/actions/runs/17427449210/job/49477714447?pr=7359

This fixes the deadlock and adds the basic deadlock detection we also added to Mutexes in #7468.

Also, interestingly, the more sophisticated deadlock detection (behind the deadlock_detection feature) didn't catch this for some reason. I wonder why it exists in the first place, when parking_lot also has built in deadlock detection? It also seems to make tests slower, widget_tests usually needs ~30s, with the deadlock detection removed its only ~12s.

### DIFF
--- a/crates/epaint/src/mutex.rs
+++ b/crates/epaint/src/mutex.rs
@@ -2,8 +2,11 @@
 
 // ----------------------------------------------------------------------------
 
-#[cfg(not(feature = "deadlock_detection"))]
+const DEADLOCK_DURATION: std::time::Duration = std::time::Duration::from_secs(30);
+
 mod mutex_impl {
+    use super::DEADLOCK_DURATION;
+
     /// Provides interior mutability.
     ///
     /// This is a thin wrapper around [`parking_lot::Mutex`], except if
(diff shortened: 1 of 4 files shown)


crates/epaint/src/mutex.rs
```diff
@@ -25,7 +30,7 @@ mod mutex_impl {
         pub fn lock(&self) -> MutexGuard<'_, T> {
             if cfg!(debug_assertions) {
                 self.0
-                    .try_lock_for(std::time::Duration::from_secs(30))
+                    .try_lock_for(DEADLOCK_DURATION)
```
emilk, line 33: We should document this on the docstrings for Mutex and RwLock, i.e. that they are meant for very short locks, and are meant for internal use in egui

lucasmerlin: Ill do this in a seperate PR where I also remove the deadlock_detection feature

PR #7496, opened 2025-09-04
Add much more detailed comments throughout the new atom-based radio_button widget

Resonanz:
The goal of this pull request is to add many more detailed comments to the ```radio_button``` widget code to help better understand the new atom-based widgets.

A visual diff has been completed and the demo app has been compiled to check for errors or warning: 

```cargo run --release  -p egui_demo_app``` 

The comments were written by Claude AI and have been entirely checked to the best of my ability.

### DIFF
--- a/crates/egui/src/widgets/button.rs
+++ b/crates/egui/src/widgets/button.rs
@@ -1,64 +1,161 @@
+// Import necessary types and traits from the crate
 use crate::{
-    Atom, AtomExt as _, AtomKind, AtomLayout, AtomLayoutResponse, Color32, CornerRadius, Frame,
-    Image, IntoAtoms, NumExt as _, Response, Sense, Stroke, TextStyle, TextWrapMode, Ui, Vec2,
-    Widget, WidgetInfo, WidgetText, WidgetType,
+    Atom,               // Individual UI element that can be laid out
+    AtomExt as _,       // Extension traits for atoms (imported with underscore)
+    AtomKind,           // Enum specifying the type of atom (Text, Image, etc.)
+    AtomLayout,         // Layout manager for arranging atoms horizontally/vertically
+    AtomLayoutResponse, // Response from atom layout operations with painting capabilities
+    Color32,            // 32-bit color representation (RGBA)
+    CornerRadius,       // Struct defining corner rounding for rectangles
+    Frame,              // Container with background, borders, and margins
+    Image,              // Image atom for displaying pictures/icons
+    IntoAtoms,          // Trait for converting types into Atoms collection
+    NumExt as _,        // Numeric extension traits
+    Response,           // UI interaction response (clicks, hovers, etc.)
+    Sense,              // What types of interactions the widget responds to
+    Stroke,             // Line style (color, width) for borders
+    TextStyle,          // Predefined text styling (Button, Body, Heading, etc.)
+    TextWrapMode,       // How text should wrap (Wrap, Truncate, Extend)
+    Ui,                 // Main UI context
+    Vec2,               // 2D vector for positions and sizes
+    Widget,             // Trait that all widgets must implement
+    WidgetInfo,         // Accessibility and debugging information
+    WidgetText,         // Text content for widgets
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: crates/egui/src/widgets/button.rs [hunk 1]
```diff
@@ -1,64 +1,161 @@
+// Import necessary types and traits from the crate
 use crate::{
-    Atom, AtomExt as _, AtomKind, AtomLayout, AtomLayoutResponse, Color32, CornerRadius, Frame,
-    Image, IntoAtoms, NumExt as _, Response, Sense, Stroke, TextStyle, TextWrapMode, Ui, Vec2,
-    Widget, WidgetInfo, WidgetText, WidgetType,
```
bircni, line 4: are these comments really needed??
this so much more lines which I don't really think we need

crates/egui/src/widgets/checkbox.rs
```diff
@@ -1,137 +1,311 @@
+// Import necessary types and traits from the crate
 use crate::{
-    Atom, AtomLayout, Atoms, Id, IntoAtoms, NumExt as _, Response, Sense, Shape, Ui, Vec2, Widget,
```
bircni, line 2: same here

crates/egui/src/widgets/radio_button.rs
```diff
@@ -1,105 +1,190 @@
+// Import necessary types and traits from the crate
 use crate::{
-    Atom, AtomLayout, Atoms, Id, IntoAtoms, NumExt as _, Response, Sense, Ui, Vec2, Widget,
-    WidgetInfo, WidgetType, epaint,
```
bircni, line 3: same here

crates/egui/src/widgets/button.rs
White-Rabbit-Scientific: I think having a heavily documented widget is helpful for those that are new to rust, or who are finding their way to how widgets work, or who require a more in-depth explanation.

Mingun: I agree that comments inside functions and doc-comments over it are helpful (I didn't read those ones), but commenting of imports is superfluous. That documentation should be on the imported things itself.

crates/egui/src/widgets/checkbox.rs
```diff
@@ -1,137 +1,311 @@
...
             }
         });
 
+        // Only proceed with rendering if the widget is actually visible on screen
+        // This is an optimization to avoid unnecessary drawing operations
         if ui.is_rect_visible(prepared.response.rect) {
-            // let visuals = ui.style().interact_selectable(&response, *checked); // too colorful
+            // Get visual styling based on the widget's interaction state
+            // Note: We use the general interact() method instead of interact_selectable()
+            // because the selectable version can be "too colorful" according to the comment
```
lucasmerlin, line 254: Lol, according to what comment?

crates/egui/src/widgets/button.rs
```diff
@@ -265,79 +550,110 @@ impl<'a> Button<'a> {
...
+            ui.spacing().button_padding // Normal padding for framed buttons
         } else {
-            Vec2::ZERO
+            Vec2::ZERO // No padding for frameless buttons
         };
+
+        // Small buttons get no vertical padding
         if small {
             button_padding.y = 0.0;
         }
```
lucasmerlin, line 588: I feel like the majority of code comments don't add much value. Do you really need a comment to know that this disables vertical padding for small buttons? 

In general, code should be written in a way that makes it obvious what is going on, and comments should explain things that aren't obvious from the code, for example _why_ something is done.

lucasmerlin (commented):
I feel it adds a lot of noise with little value, as ai tends to do. I don't think we should merge it, but we can keep the PR around, (maybe even link it from somewhere), in case someone wants to read this.

emilk:
If you think these comments are useful, then you can generate them locally (like you did). No need to commit them.

bircni:
this just feels like copilot was told to comment the project xD

PR #7497, opened 2025-09-04
Remove the `deadlock_detection` feature

lucasmerlin:
* related #7494 

Removes the `deadlock_detection` feature, since we now have a more primitive panic-after-30s deadlock detection which works well enough and even detects kinds of deadlocks that the `deadlock_detection` feature never supported.

### DIFF
--- a/crates/epaint/Cargo.toml
+++ b/crates/epaint/Cargo.toml
@@ -40,11 +40,6 @@ cint = ["ecolor/cint"]
 ## Enable the [`hex_color`] macro.
 color-hex = ["ecolor/color-hex"]
 
-## This will automatically detect deadlocks due to double-locking on the same thread.
-## If your app freezes, you may want to enable this!
-## Only affects [`mutex::RwLock`] (which epaint and egui uses a lot).
-deadlock_detection = ["dep:backtrace"]
-
 ## If set, epaint will use `include_bytes!` to bundle some fonts.
 ## If you plan on specifying your own fonts you may disable this feature.
 default_fonts = ["epaint_default_fonts"]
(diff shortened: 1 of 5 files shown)


crates/epaint/src/mutex.rs
```diff
@@ -434,7 +144,6 @@ mod tests {
 }
 
 #[cfg(not(target_arch = "wasm32"))]
-#[cfg(feature = "deadlock_detection")]
 #[cfg(test)]
 mod tests_rwlock {
     #![allow(clippy::disallowed_methods)] // Ok for tests
```
lucasmerlin, line 440: The `#[should_panic]` tests still work, but now take 30s to complete. 

Should I 
- remove the tests?
- lower the `DEADLOCK_DURATION` to 10s?
- keep it as is

I think I prefer setting `DEADLOCK_DURATION` to 10s, I think it should be fine considering no locks should be held longer than a frame

emilk:
It's worth noting what the old deadlock detector was good for, which was re-entrant locking of the same mutex on the same thread. In that case the detector would print the callstack of _both_ `lock` calls. This _used_ to be very important back when we returned locks from `ctx.input()` etc.

crates/epaint/src/mutex.rs
emilk: I agree - 10s should be good enough for anyone

crates/epaint/Cargo.toml
```diff
@@ -40,11 +40,6 @@ cint = ["ecolor/cint"]
 ## Enable the [`hex_color`] macro.
 color-hex = ["ecolor/color-hex"]
 
-## This will automatically detect deadlocks due to double-locking on the same thread.
-## If your app freezes, you may want to enable this!
-## Only affects [`mutex::RwLock`] (which epaint and egui uses a lot).
-deadlock_detection = ["dep:backtrace"]
```
emilk, line 46: Is the `backtrace` dependency unused now?

lucasmerlin: Yup, at least in epaint, I've removed it

PR #7500, opened 2025-09-04
Always check for `is_decidedly_dragging` when dragging a widget

s-nie:
Lowering the repaint rate in https://github.com/emilk/egui/pull/7482 also highlighted an issue with the drag interaction of windows among other elements. It seems that the drag gets registered on the first mouse down and the window then moves based on the cursor movement since the previous frame, making it the dragged element jump.
This is with a refresh rate of 2:

[drag.webm](https://github.com/user-attachments/assets/9a95cdcc-7c81-42c7-b98b-d4164c3b8a9d)

The fix I found for this was to always check for `is_decidedly_dragging`. I didn't notice any input delays or other negative effects from this.

### DIFF
--- a/crates/egui/src/interaction.rs
+++ b/crates/egui/src/interaction.rs
@@ -191,20 +191,8 @@ pub(crate) fn interact(
     if dragged.is_none() {
         // Check if we started dragging something new:
         if let Some(widget) = interaction.potential_drag_id.and_then(|id| widgets.get(id)) {
-            if widget.enabled {
-                let is_dragged = if widget.sense.senses_click() && widget.sense.senses_drag() {
-                    // This widget is sensitive to both clicks and drags.
-                    // When the mouse first is pressed, it could be either,
-                    // so we postpone the decision until we know.
-                    input.pointer.is_decidedly_dragging()
-                } else {
-                    // This widget is just sensitive to drags, so we can mark it as dragged right away:
-                    widget.sense.senses_drag()
-                };
-
-                if is_dragged {
-                    dragged = Some(widget.id);
-                }
+            if widget.enabled && input.pointer.is_decidedly_dragging() {
+                dragged = Some(widget.id);
             }
         }
     }


crates/egui/src/interaction.rs
```diff
@@ -191,20 +191,8 @@ pub(crate) fn interact(
...
         if let Some(widget) = interaction.potential_drag_id.and_then(|id| widgets.get(id)) {
-            if widget.enabled {
-                let is_dragged = if widget.sense.senses_click() && widget.sense.senses_drag() {
-                    // This widget is sensitive to both clicks and drags.
-                    // When the mouse first is pressed, it could be either,
-                    // so we postpone the decision until we know.
-                    input.pointer.is_decidedly_dragging()
-                } else {
-                    // This widget is just sensitive to drags, so we can mark it as dragged right away:
-                    widget.sense.senses_drag()
```
emilk, line 202: I think this will lead to a regression where widgets that are only sensitive to drags (search for `Sense::drag`) will not immediately respond to being dragged (`is_decidedly_dragging` waits until the drag has gone for at least a certain distance or time)

s-nie: I tried a couple of widgets that have `Sense::drag` but didn't notice any delay. Maybe I don't know what to look for though. Do you know which widget would most likely be affected?

emilk: You can notice it for instance for drag-to-scroll. On egui.rs if I drag the background of e.g. the Font Book it moves with the mouse cursor right away. If I do so [on this PR](https://egui-pr-preview.github.io/pr/7500-drag-input-fix) there is a small gap before the dragging kicks in, making it feel unresponsive

s-nie: I see now. It required comparing two screen recordings but in the one from this PR the drag actually appears to start two frames later after the click (5 vs 3).

Unless you happen to have any other suggestions to solve this, I would close the issue now as I can't think of different solutions and the issue is somewhat artificial.

PR #7508, opened 2025-09-05
kittest: Add `UPDATE_SNAPSHOTS=force`

emilk:
This adds a new mode, `UPDATE_SNAPSHOTS=force`, which will lower the threshold to zero, overwriting every image that is not _exactly_ the same.

Most comparisons has a threshold because different GPUs render slightly differently. However, setting that threshold accurately can be hard.

Sometimes a test will pass locally, but fail on CI. In those cases you want to force an update of the failing test. You can use `UPDATE_SNAPSHOTS=force` for that.

And sometimes a small change _should_ update all images, but the change is so tiny that it falls under the threshold. Still, you want to make a point of showing that these images have changes. You can use `UPDATE_SNAPSHOTS=force` for that.


lucasmerlin (approved):
Nice!

crates/egui_kittest/src/snapshot.rs
```diff
@@ -285,14 +285,34 @@ impl Display for SnapshotError {
...
-/// and _succeed_ the test.
-/// This is so that you can set `UPDATE_SNAPSHOTS=true` and update _all_ tests,
-/// without `cargo test` failing on the first failing crate.
-fn should_update_snapshots() -> bool {
-    match std::env::var("UPDATE_SNAPSHOTS") {
-        Ok(value) => !matches!(value.as_str(), "false" | "0" | "no" | "off"),
-        Err(_) => false,
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum Mode {
+    Fail,
```
emilk: `Test` would be a better name for this

PR #7512, opened 2025-09-08
Improve `OrderedFloat` hash performance

valadaptive:

Another thing that showed up in the profiler when working on #7431. I'm doing a lot of caching of fonts by size, which involves hashing a lot of `OrderedFloat`s. We can [take a page from the `ordered-float` crate](https://github.com/reem/rust-ordered-float/blob/1841f0541ea0e56779cbac03de2705149e020675/src/lib.rs#L2178-L2181) and save a branch, which speeds things up by a couple percent in the `text_layout_uncached` benchmark.


lucasmerlin (approved):
Nice! Makes sense to me

PR #7515, opened 2025-09-08
Improve deadlock detection output

emilk:
* We missed a couple of spots in https://github.com/emilk/egui/pull/7494


emilk:
ugh, I had the wrong PR title when merging 😭

PR #7521, opened 2025-09-09
Increase default text size from 12.5 to 13.0

emilk:
…and other small ui tweaks.

Why the change from 12.5 -> 13.0? We round the line height to the nearest whole pixel, so using an integer will produce text that is more similar to other applications. It also means we can bump the size without using up more vertical space (12.5 got rounded down to 12.0).


emilk:
Let's bump monospace too, for consistency

PR #7526, opened 2025-09-10
Remove the `compact_menu_style` setting. This was replaced by customizing `MenuConfig`.

s-nie:
This feature is no longer needed as the menu styling is now configurable through `MenuConfig`.


lucasmerlin:
I think the setting still makes sense, since the MenuConfig is not globally overridable

PR #7532, opened 2025-09-10
feat: Add Configurable Speed and Direction to Spinner Control 

KongJian520:
### Description
This commit introduces two new properties for the `Spinner` control: `speed` and `clockwise`, allowing developers to customize the behavior of its rotation animation.

### Major Changes
- Added `speed: Option<f64>` and `clockwise: Option<bool>` fields to the `Spinner` structure.
- Provided chainable `speed()` and `clockwise()` methods to facilitate setting these new properties.
- In the `ui` method of the `Widget` implementation, `unwrap_or` is used to set default values ​​for these two properties to ensure they work correctly even when not explicitly set.
- Modified the `paint_at` method signature to directly accept the processed `speed` and `clockwise` parameters to simplify the drawing logic.

### DIFF
--- a/crates/egui/src/widgets/spinner.rs
+++ b/crates/egui/src/widgets/spinner.rs
@@ -2,6 +2,7 @@ use epaint::{Color32, Pos2, Rect, Shape, Stroke, emath::lerp, vec2};
 
 use crate::{Response, Sense, Ui, Widget, WidgetInfo, WidgetType};
 
+
 /// A spinner widget used to indicate loading.
 ///
 /// See also: [`crate::ProgressBar`].
@@ -11,6 +12,8 @@ pub struct Spinner {
     /// Uses the style's `interact_size` if `None`.
     size: Option<f32>,
     color: Option<Color32>,
+    speed: Option<f64>,      // 旋转速度
+    clockwise: Option<bool>, // 是否顺时针
 }
 
 impl Spinner {
@@ -34,8 +37,20 @@ impl Spinner {
         self
     }
 
+    /// Sets the spinner's rotation speed.
+    pub fn speed(mut self, speed: f64) -> Self {
+        self.speed = Some(speed);
+        self
+    }
+
+    /// Sets the spinner's rotation direction. True for clockwise, false for counter-clockwise.
+    pub fn clockwise(mut self, clockwise: bool) -> Self {
+        self.clockwise = Some(clockwise);
+        self
+    }
+
     /// Paint the spinner in the given rectangle.
-    pub fn paint_at(&self, ui: &Ui, rect: Rect) {
+    pub fn paint_at(&self, ui: &Ui, rect: Rect, speed: f64, clockwise: bool) {
         if ui.is_rect_visible(rect) {
             ui.ctx().request_repaint(); // because it is animated
 
@@ -45,7 +60,7 @@ impl Spinner {
             let radius = (rect.height() / 2.0) - 2.0;
             let n_points = (radius.round() as u32).clamp(8, 128);
             let time = ui.input(|i| i.time);
-            let start_angle = time * std::f64::consts::TAU;
+            let start_angle = time * speed * if clockwise { 1.0 } else { -1.0 } * std::f64::consts::TAU;
             let end_angle = start_angle + 240f64.to_radians() * time.sin();
             let points: Vec<Pos2> = (0..n_points)
                 .map(|i| {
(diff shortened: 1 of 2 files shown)


crates/egui/src/widgets/spinner.rs
```diff
@@ -8,67 +8,83 @@ use crate::{Response, Sense, Ui, Widget, WidgetInfo, WidgetType};
...
 #[derive(Default)]
 pub struct Spinner {
-    /// Uses the style's `interact_size` if `None`.
-    size: Option<f32>,
-    color: Option<Color32>,
+	/// Uses the style's `interact_size` if `None`.
+	size: Option<f32>,
+	color: Option<Color32>,
+	speed: Option<f64>,      // 旋转速度
+	clockwise: Option<bool>, // 是否顺时针
```
lucasmerlin: Please keep the source code in english

crates/egui/src/widgets/spinner.rs
```diff
@@ -8,67 +8,83 @@ use crate::{Response, Sense, Ui, Widget, WidgetInfo, WidgetType};
...
 
-    /// Sets the spinner's size. The size sets both the height and width, as the spinner is always
-    /// square. If the size isn't set explicitly, the active style's `interact_size` is used.
-    #[inline]
-    pub fn size(mut self, size: f32) -> Self {
-        self.size = Some(size);
-        self
-    }
+	/// Sets the spinner's size. The size sets both the height and width, as the spinner is always
+	/// square. If the size isn't set explicitly, the active style's `interact_size` is used.
```
lucasmerlin: Spaces somehow switched to tabs here it seems

crates/egui/src/widgets/spinner.rs
```diff
@@ -8,67 +8,83 @@ use crate::{Response, Sense, Ui, Widget, WidgetInfo, WidgetType};
...
 
-    /// Paint the spinner in the given rectangle.
-    pub fn paint_at(&self, ui: &Ui, rect: Rect) {
-        if ui.is_rect_visible(rect) {
-            ui.ctx().request_repaint(); // because it is animated
+	/// Sets the spinner's rotation speed.
+	pub fn speed(mut self, speed: f64) -> Self {
+		self.speed = Some(speed);
+		self
+	}
```
lucasmerlin: Can you add a unit to the doc comment? Something like "a speed of 1.0 means 1 rotation per second"

crates/egui/src/widgets/spinner.rs
```diff
@@ -8,67 +8,83 @@ use crate::{Response, Sense, Ui, Widget, WidgetInfo, WidgetType};
...
-                    let (sin, cos) = angle.sin_cos();
-                    rect.center() + radius * vec2(cos as f32, sin as f32)
-                })
-                .collect();
-            ui.painter()
-                .add(Shape::line(points, Stroke::new(3.0, color)));
-        }
-    }
+	/// Sets the spinner's rotation direction. True for clockwise, false for counter-clockwise.
+	pub fn clockwise(mut self, clockwise: bool) -> Self {
```
lucasmerlin: I feel an enum would be better than a bool here. Something like 
```
RotationDirection {
    Clockwise,
    CounterClockwise,
}
```

With a bool, if you see something like this: `Spinner::new().paint_at(ui, rect, 1.0, true);` it's not obvious what the bool does

crates/egui/src/widgets/spinner.rs
KongJian520: No problem, I've also added some sample code to explain how to use this property.

crates/egui/src/widgets/spinner.rs
KongJian520: im sorry，

crates/egui/src/widgets/spinner.rs
KongJian520: OK, but after thinking about it, I found that this attribute seems to be useless, so I merged it into the speed attribute (positive numbers are clockwise, complex numbers are counterclockwise)

KongJian520:
@lucasmerlin  I'm ready, thanks for the review

crates/egui/src/widgets/spinner.rs
```diff
@@ -33,9 +33,28 @@ impl Spinner {
...
+    /// let counter_clockwise = Spinner::new().speed(-1.0);
+    /// ```
+    pub fn speed(mut self, speed: f64) -> Self {
+        self.speed = Some(speed);
+        self
+    }
 
     /// Paint the spinner in the given rectangle.
-    pub fn paint_at(&self, ui: &Ui, rect: Rect) {
+    pub fn paint_at(&self, ui: &Ui, rect: Rect, speed: f64) {
```
lucasmerlin, line 57: Why does this have a speed param? It should get the value from self.speed

lucasmerlin (changes requested):

PR #7535, opened 2025-09-11
Reset wrapping in label tooltip

lucasmerlin:
* follow up to #7514 

That PR changed the tooltip to preserve the wrapping, which made the tooltip kind of useless. With this PR the wrapping is reset for the tooltip.


crates/egui/src/widgets/label.rs
```diff
@@ -281,8 +281,14 @@ impl Widget for Label {
 
         if ui.is_rect_visible(response.rect) {
             if show_tooltip_when_elided && galley.elided {
+                // Keep the sections and text, but reset everything else (especially wrapping):
+                let job = crate::text::LayoutJob {
+                    sections: galley.job.sections.clone(),
+                    text: galley.job.text.clone(),
+                    ..crate::text::LayoutJob::default()
```
emilk, line 288: Good call

PR #7541, opened 2025-09-14
Add Label text selection api

mankinskin:


lucasmerlin:
Sorry this is very conflicted now that I've merged the new plugin trait. You'll have to update it to fetch the label selection state via the plugin helpers

emilk:
Cool!

It would be nice if the egui_demo_lib used this new functionality.

Also: please add a kittest test of it 🙏

mankinskin:
I still need to find a way to get selections across multiple widgets..

So far I was only able to get the latest selection in the current Context per Viewport. For my app I need to be able to fetch all current selections, across any widgets in the current context and viewport.

I will look at the new plugin api and go from there.

mankinskin:
https://github.com/emilk/egui/pull/7541/files#diff-335ddbce02d068207f42ecdac988ee267eb014a7a1e7ffe8df7bedd66f716899R9

By the way I added this rust toolchain component, because otherwise rust-analyzer wouldn't start for me.

rust-toolchain
```diff
@@ -6,5 +6,5 @@
 
 [toolchain]
 channel = "1.86.0"
-components = ["rustfmt", "clippy"]
+components = ["rustfmt", "clippy", "rust-analyzer"]
```
mankinskin: By the way I added this rust toolchain component, because otherwise rust-analyzer wouldn't start for me.

YgorSouza: This forces everyone who builds the project to install rust-analyzer via rustup, even though they might not need it (because they've installed it through VS Code or some other IDE, or because they are a CI runner just running the CI tests). If you want to install rust-analyzer via rustup, you can just do that locally without adding anything to the repo.

```
rustup component add rust-analyzer
```

mankinskin: I thought I had that, but maybe not for the 1.86 toolchain. I will remove this line then.

mankinskin:
I reworked the api a bit and added an example project as a demo.

I have added a map in `LabelSelectionState` from `widget_id: Id` to `Arc<Galley>` to be able to calculate the selected char range on Responses without providing the galley.

I am thinking about storing an entire WidgetSelectionState in this map, to store `secondary_reached: bool` and `primary_reached: bool` at render time for each widget.

Otherwise the range calculation will be wrong if the calls to widget_cursor_range on a Response come after another widget has been rendered (because they all share the LabelSelectionState and update primary_reached, ...).

The example shows how to do it correctly at the moment, by extracting the cursor_range in the same iteration the Label is rendered.

PR #7551, opened 2025-09-18
`kitdiff`, a kittest snapshot viewer

lucasmerlin:
I got frustrated with the experience of reviewing snapshot changes in my ide and on github, so I made something really cool:

Just run `kitdiff` in your terminal and it'll visualize all the diffs! You can use the `1, 2, 3` keys to switch between `original, new, diff` images. You can also generate diffs on the fly to play with different threshold options. Also, it's wicked quick! 

https://github.com/user-attachments/assets/c9324ef3-eb24-481f-83b8-42a37b6b075d

**but wait, there's more**

You can do `kitdiff pr https://github.com/rerun-io/rerun/pull/11253` to view a diff of that PR, you don't even need to check out the branch!

https://github.com/user-attachments/assets/d5c0b15a-0a75-4506-8dae-51b8bb83836f

### DIFF
--- a/crates/egui_extras/src/loaders.rs
+++ b/crates/egui_extras/src/loaders.rs
@@ -108,16 +108,16 @@ pub fn install_image_loaders(ctx: &egui::Context) {
 }
 
 #[cfg(not(target_arch = "wasm32"))]
-mod file_loader;
+pub mod file_loader;
 
 #[cfg(feature = "http")]
-mod ehttp_loader;
+pub mod ehttp_loader;
 
 #[cfg(feature = "gif")]
-mod gif_loader;
+pub mod gif_loader;
 #[cfg(feature = "image")]
-mod image_loader;
+pub mod image_loader;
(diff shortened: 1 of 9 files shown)


crates/egui_extras/src/loaders.rs
```diff
@@ -108,16 +108,16 @@ pub fn install_image_loaders(ctx: &egui::Context) {
 }
 
 #[cfg(not(target_arch = "wasm32"))]
-mod file_loader;
+pub mod file_loader;
 
 #[cfg(feature = "http")]
-mod ehttp_loader;
+pub mod ehttp_loader;
```
emilk: I think we should rename this `http_loader`. That it uses `ehttp` is an implementation detail. It didn't matter much before, but now that it's public…

lucasmerlin: Good point 👍🏻

Issue #7424, opened 2024-07-31
Bug in polygon tessellator with feathering


berke:
Filled polygons added to a plot show line-shaped artefacts under certain zooming conditions.

I have an `eframe::App` containing a canvas containing a `Plot`, where I draw filled triangles using `ui.polygon`. 

For certain triangles and at certain zoom levels, this produces fill-colored linear artefacts.

No artefacts appear if the triangles are not filled.

**To Reproduce**

1. Checkout https://github.com/berke/egui_triangle_bug
2. Compile and run
3. This shows a triangle having a yellow outline and a red fill.  Observe red artefacts that go beyond the bounds of the triangle.  If needed, adjust the zoom level.

**Expected behavior**
There should be no artefacts.

**Screenshots**

Real-world case:

![egui_plot_issue_1](https://github.com/user-attachments/assets/f680c08a-2900-4ba6-b7ec-4336d94310bb)

Screenshot from simplified PoC:
![screenshot](https://github.com/user-attachments/assets/0b95d0eb-6ead-4ae0-8ca7-87c6ef1faaf2)

**Additional context**

Latest `egui` and `egui_plot`

```
...
[dependencies.egui_plot]
version = "0.28.1"

[patch.crates-io]
egui_plot = { git = "https://github.com/emilk/egui_plot", branch = "main" }
...
```

x86 machine under Debian and X11 with NVIDIA driver 535.183.01, xfce4 with xfwm 4.18.0-1, no compositing

CramBL:
You can fairly easily reproduce this issue on https://www.egui.rs/ by selecting the `Bézier Curve` demo on the right:

Just move the circles to produce tight angles between the lines
e.g.
![image](https://github.com/user-attachments/assets/09dba37c-f917-472d-8bda-2e769c4bd2f5)

It also produces another surprising result occasionally, which is that the blue filled area spills over and goes outside the boundaries marked by the green lines.

adamconkey:
@berke many thanks! 🙏 I applied your suggestion to add
```rust	
ctx.tessellation_options_mut(|to| {
    to.feathering = false;
});
```
and in my case it works great, the artefacts are no longer present.

Before:
![Screen Shot 2025-01-01 at 11 24 44 AM](https://github.com/user-attachments/assets/bbcdfbf2-fd44-455b-a970-e8073af49818)

After:
[...]

Issue #1226, opened 2022-02-08
Filled convex polygon shapes have jagged edges


robbert-vdh:
**Describe the bug**
The edges around the filled area of a `egui::Shape::convex_polygon()` don't have proper anti-aliasing and will appear jagged. For semi-transparent fill colors this looks as if some form of anti aliasing is taking place, but the anti-aliased pixels are drawn without the alpha channel. Opaque colors look like they are not anti aliased at all. (see the screenshots below for examples of both of these things)

**To Reproduce**
1. Take the `hello_world.rs` example, and add the following add the end of the `CentralPanel`:
   ```patch
   diff --git a/eframe/examples/hello_world.rs b/eframe/examples/hello_world.rs
   index 7ec61fb..d3619e8 100644
   --- a/eframe/examples/hello_world.rs
   +++ b/eframe/examples/hello_world.rs
   @@ -35,6 +35,22 @@ impl epi::App for MyApp {
                    *age += 1;
                }
                ui.label(format!("Hello '{}', age {}", name, age));
   +
   +            let painter = ui.painter();
   +            painter.rect_filled(
   +                egui::Rect::from_x_y_ranges(0.0..=260.0, 0.0..=60.0),
   +                egui::Rounding::none(),
   +                egui::Color32::BLACK,
   +            );
   +            painter.add(egui::Shape::convex_polygon(
   +                vec![
   +                    egui::pos2(10.0, 40.0),
   +                    egui::pos2(250.0, 50.0),
   +                    egui::pos2(10.0, 10.0),
   +                ],
   +                egui::Color32::GRAY,
   +                egui::Stroke::none(),
   +            ));
            });
    
            // Resize the native window to be just the size we need it to be:
   ```
2. Run the example with `cargo run --example hello_world`.

**Expected behavior**
The edges should appear smooth, just like drawn lines or shape strokes currently do.

**Screenshots**
The modified `hello_world` example mentioned above:
![afbeelding](https://user-images.githubusercontent.com/748520/152992917-0d326992-a51c-426a-821d-9dcda7147b4f.png)

The same example, but with a pure pure fill color without any transparency:
![afbeelding](https://user-images.githubusercontent.com/748520/152993417-49d2c2eb-b132-4ef3-9196-60fb3d8b4a12.png)

And the same problem with a different backend ([egui-baseview](https://github.com/BillyDM/egui-baseview/pull/4)):
[...]

PR #7554, opened 2025-09-21
Workaround for feathering of sharp edges

podusowski:
## Related
* Closes https://github.com/emilk/egui/issues/7424

## What
I am working on rendering vector maps in Walkers (which looks promising!) and I noticed weird artifacts:

image

It appears that the issue is caused by edge normals, which kind of explode when the shape is sharp enough. This PR adds a limiter, so that the normal is clamped when it crosses certain length. It's not perfect and most likely can be fixed in more elegant way, but it does seem like an improvement:

image

image

### DIFF
--- a/crates/egui_demo_lib/src/rendering_test.rs
+++ b/crates/egui_demo_lib/src/rendering_test.rs
@@ -594,7 +594,10 @@ fn blending_and_feathering_test(ui: &mut Ui) {
     painter.rect_filled(bottom_half, 0.0, Color32::WHITE);
     paint_fine_lines_and_text(&painter, bottom_half, Color32::BLACK);
 
-    feathering_of_sharp_edges_test(ui);
+    // Slight variation between GitHub Actions runners and local machines makes 1.25 to fail.
+    if ui.ctx().pixels_per_point() != 1.25 {
+        feathering_of_sharp_edges_test(ui);
+    }
 }
 
 fn feathering_of_sharp_edges_test(ui: &mut Ui) {
--- a/crates/egui_demo_lib/tests/snapshots/rendering_test/dpi_1.25.png
+++ b/crates/egui_demo_lib/tests/snapshots/rendering_test/dpi_1.25.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:cd65a74c320fa9890643a536867f63041fc461e25be7446c87a1e9bc27f24bc9
-size 854438
+oid sha256:49bca8679f2216a5cb4c007f69b2b17e6e62eb982fa7b0f44478c8a934cea10f
+size 824217


crates/epaint/src/tessellator.rs
```diff
@@ -757,6 +757,15 @@ fn cw_signed_area(path: &[PathPoint]) -> f64 {
     }
 }
 
+fn clamp_normal_for_feathering(normal: Vec2) -> Vec2 {
+    let max_normal_length = 90.0;
```
emilk: 90 sounds really high - how did you arrive at this value?

crates/epaint/src/tessellator.rs
```diff
@@ -757,6 +757,15 @@ fn cw_signed_area(path: &[PathPoint]) -> f64 {
     }
 }
 
+fn clamp_normal_for_feathering(normal: Vec2) -> Vec2 {
```
emilk: It would be nice with a comment explain what this is and why

crates/epaint/src/tessellator.rs
```diff
@@ -757,6 +757,15 @@ fn cw_signed_area(path: &[PathPoint]) -> f64 {
...
+    let max_normal_length = 90.0;
+    if normal.length() > max_normal_length {
+        normal * (max_normal_length / normal.length())
```
emilk: nit: a little faster to avoid square root

```suggestion
    if normal.length_sq() > max_normal_length * max_normal_length {
        normal * (max_normal_length / normal.length())
```

crates/epaint/src/tessellator.rs
podusowski: Well, it was highly subjective task of visually inspecting both test that I've added and artifacts seen on the map. Some of them are probably caused by how I drawing polygons - I use earcut to cut concave polygons to something `edraw` can draw.

For `length_sq` it looks more or less like this:

image

image

image

image

image

Though I'm confused why there's a spike on the second triangle, but it's gone when I remove this clamping entirely.

podusowski: Argh, I've a mistake in these. Let me remade them tomorrow.

emilk: It would be nice to add this constant to `TessellationOptions`

podusowski:
I did couple of changes to the test, namely to be more like https://github.com/emilk/egui/issues/1226. `CUT_OFF_SHARP_CORNERS` is not affected by this change, I mean, it still needs to be turned off, otherwise jagged edges come back.

I've put the limit into `TessellationOptions`, so now it's easier to play with it.

Unfortunately, although IMO it's still better to have this, but it seems that there are no golden value here.

Baseline:
image

image

image

image

crates/epaint/src/tessellator.rs
```diff
@@ -757,12 +819,27 @@ fn cw_signed_area(path: &[PathPoint]) -> f64 {
     }
 }
 
+/// See [`TessellationOptions::max_feathering_normal_length_sq`].
+fn clamp_normal_for_feathering(normal: Vec2, max_feathering_normal_length_sq: f32) -> Vec2 {
+    if normal.length_sq() > max_feathering_normal_length_sq {
+        normal * (max_feathering_normal_length_sq / normal.length_sq())
```
emilk: I think this is incorrect.

if `normal = [10, 0]` for instance, and max_feathering_normal_length = 2, so max_feathering_normal_length_sq = 4, then we would divide by 100/4 instead of 10/2.

```suggestion
    if normal.length_sq() > sqr(max_feathering_normal_length) {
        max_feathering_normal_length * normal.normalize()
```

podusowski: I had some issues when I did literal clamping, not sure why because now it looks fine (more or less).

Baseline:
image

Default (40.0):
image

PR #7556, opened 2025-09-22
Include popups and tooltips in `Harness::fit_contents`

oxkitsune:
This makes `Harness::fit_contents` also use popups and tooltips to compute the size of the contents.

### DIFF
--- a/crates/egui_kittest/tests/tests.rs
+++ b/crates/egui_kittest/tests/tests.rs
@@ -22,7 +22,7 @@ fn test_tooltip() {
         ui.label("Hello, world!");
         ui.separator();
         ui.label("This is a test")
-            .on_hover_text("This is a tooltip");
+            .on_hover_text("This\nis\na\nvery\ntall\ntooltip!");
     });
 
     harness.fit_contents();
--- a/crates/egui_kittest/tests/snapshots/test_tooltip_shown.png
+++ b/crates/egui_kittest/tests/snapshots/test_tooltip_shown.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:455a5601ea5672cfa1a7a4fa79d3d8b60fa358cfe5a8207af62cc95ae4c4e22f
-size 3704
+oid sha256:c267530452adb4f1ed1440df476d576ef4c2d96e6c58068bb57fed4615f5e113
+size 4453


crates/egui_kittest/src/lib.rs
```diff
@@ -276,14 +276,39 @@ impl<'a, State> Harness<'a, State> {
...
+                .filter_map(|layer_id| mem.area_rect(layer_id.id))
+                .for_each(|area_rect| used |= area_rect);
+        });
+
+        Some(used)
+    }
+
     /// Resize the test harness to fit the contents. This only works when creating the Harness via
     /// [`Harness::new_ui`] / [`Harness::new_ui_state`] or
     /// [`HarnessBuilder::build_ui`] / [`HarnessBuilder::build_ui_state`].
```
lucasmerlin, line 303: Guessing this doc comment is no longer true, since you are now checking for `self.ctx.used_rect()`?
Would be nice if you added a test for this, similar to this one:

https://github.com/emilk/egui/blob/42c2fc58c90af92f19cc6bbda4b75ce7a290c0aa/crates/egui_kittest/tests/tests.rs#L5-L17

crates/egui_kittest/src/lib.rs
```diff
@@ -276,14 +276,39 @@ impl<'a, State> Harness<'a, State> {
...
     }
 
+    /// Calculate the rect that includes all popups and tooltips.
+    fn calculate_used_rect_with_popups(&self) -> Option<Rect> {
+        // Start with the standard response rect
+        let mut used = if let Some(response) = self.response.as_ref() {
+            response.rect
+        } else {
+            return None;
+        };
```
emilk, line 286: alternative, that will still work in case of no response, but when there are areas. Though needs handling in case there are no areas

```suggestion
        // Start with the standard response rect
        let mut used = if let Some(response) = self.response.as_ref() {
            response.rect
        } else {
            Rect::NOTHING
        };
```

crates/egui_kittest/src/lib.rs
oxkitsune: I'm not using `self.ctx.used_rect()` but I get the confusion 🙃 
will add a unit test 👍

crates/egui_kittest/src/lib.rs
oxkitsune: I want to retain the current behavior in case of no response.

crates/egui_kittest/tests/tests.rs
```diff
@@ -16,6 +16,28 @@ fn test_shrink() {
     harness.snapshot("test_shrink");
 }
 
+#[test]
+fn test_tooltip() {
+    let mut harness = Harness::new_ui(|ui| {
+        ui.label("Hello, world!");
+        ui.separator();
+        ui.label("This is a test")
+            .on_hover_text("This is a tooltip");
```
lucasmerlin: Can you make the text a bit higher by adding a new line? The screenshots are still almost equal in size, making it hard to see if the image actually got bigger

Issue #7555, opened 2025-09-22
`mismatched types` error in `egui-winit` with `profiling`


torokati44:
```
     --> /home/runner/.cargo/git/checkouts/egui-9c90d2f9798966b7/c97c065/crates/egui-winit/src/lib.rs:1364:32
     |
1364 |     profiling::function_scope!(format!("{command:?}"));
     |                                ^^^^^^^^^^^^^^^^^^^^^^ expected `&str`, found `String`
     |
     = note: this error originates in the macro `format` (in Nightly builds, run with -Z macro-backtrace for more info)
```

See: https://github.com/ruffle-rs/ruffle/actions/runs/17911755745/job/50924722672#step:6:1053

torokati44:
This line got changed in https://github.com/emilk/egui/commit/a5a51ced0f8fc7612dfce9b4b8432883b671868b.

torokati44:
This would probably fix it:

```diff
fn process_viewport_command(
     egui_ctx: &egui::Context,
     window: &Window,
     command: ViewportCommand,
     info: &mut ViewportInfo,
     actions_requested: &mut Vec<ActionRequested>,
 ) {
-    profiling::function_scope!(format!("{command:?}"));
+    profiling::function_scope!(&format!("{command:?}"));
 
     use winit::window::ResizeDirection;
 
     log::trace!("Processing ViewportCommand::{command:?}");
 
     let pixels_per_point = pixels_per_point(egui_ctx, window);
 ```

PR #7557, opened 2025-09-23
Fix build error in egui-winit with profiling enabled

torokati44:
Fixes https://github.com/emilk/egui/issues/7555.

Verified by patching `ruffle` to use a checked out `egui` repo with this change, and running `cargo run --release --features=tracy` in it.

I hesitated between `&` and `.as_str()` - if you'd prefer the latter, let me know.

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -1361,7 +1361,7 @@ fn process_viewport_command(
     info: &mut ViewportInfo,
     actions_requested: &mut Vec<ActionRequested>,
 ) {
-    profiling::function_scope!(format!("{command:?}"));
+    profiling::function_scope!(&format!("{command:?}"));
 
     use winit::window::ResizeDirection;
 


emilk:
We should make sure this is caught by CI 🤔

emilk:
I fail to reproduce the problem locally - what command are you running to get a build error?

torokati44:
For example, `cargo clippy --workspace --all-features` in https://github.com/ruffle-rs/ruffle/pull/21730.

torokati44:
The point is to get the `profiling/profile-with-tracy` feature enabled (possibly an issue with other `profile-with-...` features too, haven't checked).

emilk:
But the problem doesn't repo in the `egui` repository? Maybe we just need to update the `profiling` or `puffin` crates?

torokati44:
> Maybe we just need to update the profiling or puffin crates?

 `format!` returns a `String`: https://doc.rust-lang.org/std/macro.format.html
 `profiling::function_scope!` takes a `&str`: https://docs.rs/profiling/latest/profiling/macro.function_scope.html
 
An example at the latter link conincidentally even has a `format!` in it, and it does have to have `.as_str()` after it...
What else could I say?

torokati44:
It just happens to work with `puffin` because the `profiling` impl for it expands to a function call that is more lax, taking `impl AsRef<str>`:
https://github.com/aclysma/profiling/blob/740d74c82212f1f0584cfa5c6d884c5ab5367101/profiling/src/puffin_impl.rs#L17
https://docs.rs/puffin/0.19.1/src/puffin/lib.rs.html#252-275
https://docs.rs/puffin/0.19.1/puffin/struct.ProfilerScope.html#method.new
Arguably, it shouldn't.

While for the `tracy` backend, it is `&str` all the way through macro expansion:
https://github.com/aclysma/profiling/blob/740d74c82212f1f0584cfa5c6d884c5ab5367101/profiling/src/tracy_impl.rs#L48
https://docs.rs/tracy-client/0.18.2/tracy_client/struct.Span.html#method.emit_text

torokati44:
> But the problem doesn't repo in the egui repository?

Something would have to enable the `profiling/profile-with-tracy` feature, but nothing does, plus, it seems like it is mutually exclusive with `profiling/profile-with-puffin` (and probably all the others too), so...

emilk:
I see - thanks for explaining it

PR #7559, opened 2025-09-23
Return `0.0` if font not found in `glyph_width` instead of panic

### LINKED ISSUE #7548: Failure to read `glyph_width` in egui_commonmark
**Describe the bug**
After updating to egui main, the tests there fail because of a failure to read glyph_width.

- https://github.com/lampsitter/egui_commonmark/pull/82
- Probably related to https://github.com/emilk/egui/pull/7431

lucasmerlin:
* Closes #7548 
* related to #7431 

This reverts to the `glyph_width` behaviour before #7431 - return 0.0 if font not found instead of panicing.
The main use for this is when using `__run_test_ui`, which doesn't register any fonts, any code using glyph_width would previously panic.

### DIFF
--- a/crates/epaint/src/text/font.rs
+++ b/crates/epaint/src/text/font.rs
@@ -481,10 +481,6 @@ impl Font<'_> {
         if let Some(font) = &self.fonts_by_id.get(&key) {
             glyph_info.advance_width_unscaled.0 * font.ab_glyph_font.px_scale_factor(font_size)
         } else {
-            debug_assert_eq!(
-                glyph_info.advance_width_unscaled.0, 0.0,
-                "We had no font, so this should be the zero-width fallback glyph"
-            );
             0.0
         }
     }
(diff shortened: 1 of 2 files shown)


crates/epaint/src/text/font.rs
```diff
@@ -478,12 +478,15 @@ impl Font<'_> {
...
-            .expect("Nonexistent font ID")
-            .ab_glyph_font;
-        glyph_info.advance_width_unscaled.0 * font.px_scale_factor(font_size)
+        if let Some(font) = &self.fonts_by_id.get(&key) {
+            glyph_info.advance_width_unscaled.0 * font.ab_glyph_font.px_scale_factor(font_size)
+        } else {
+            debug_assert_eq!(
+                glyph_info.advance_width_unscaled.0, 0.0,
+                "We had no font, so this should be the zero-width fallback glyph"
+            );
```
emilk: All glyphs should already be zero-width when there are no fonts?

PR #7560, opened 2025-09-23
Change Spinner widget to account for width as well as height

bryceberger:
Previously, when `rect` was taller than it was wide, the spinner would render far outside the given rectangle. Now, it always renders inside the smaller of the two dimensions.

I noticed this when upgrading from 0.30 to 0.32. I have an image that's significantly taller than it is wide. In 0.32, when the image is loading, it shows the spinner. Since the spinner radius is determined solely based on rectangle height, the spinner ends up far too wide and covers other elements.


bryceberger:
Poking this --- only CI failure is the PR not being labelled, and I seem to be unable to label the PR.

Issue #6757, opened 2025-04-16
Eframe application window does not close


tye-exe:
**Describe the bug**
There is bug with eframe where the application window remains open and cannot be closed while the program is running. If the previous window is left in this state for too long, the program is reported as not responding, even though code execution that does not involve the GUI still continues.

The application window is removed when the entire program exits.
Alternatively, calling the "run_native" method replaces the previous window with a new one.

**To Reproduce**
1. Clone the repository that demonstrates the issue: (https://github.com/tye-exe/egui_no_close)[https://github.com/tye-exe/egui_no_close].
2. Run the application with `cargo run`.
3. Attempt to close the window by clicking the "x" button.
4. The window will remain open.

(The example program must be stopped via "Ctrl+C")

**Expected behavior**
There should be no eframe window open after it is closed.

**Video Example**
[See here](https://github.com/user-attachments/assets/424882dc-7ec6-4b8b-aa63-6a497622880f)

**Desktop (please complete the following information):**
 - OS: NixOS 24.11 (to reproduce entire system see [my config](https://github.com/tye-exe/nixos-config/tree/139a26580f2cecff55f3893e394bdac5173a88e0))
 - Browser N/A
 - Version N/A

**Additional context**
Attempting to update to winit master branch (currently https://github.com/rust-windowing/winit/tree/cdbdd974fbf79b82b3fb1a4bc84ed717312a3bd2) did not resolve this issue.

tye-exe:
I am currently trying to resolve this issue in eframe, but progress is slow as i'm unsure as to what i am doing.

I know that this is possible, as running the [winit run on demand example](https://github.com/rust-windowing/winit/blob/master/examples/run_on_demand.rs) shows the expected behavior.

Any additional help would be appreciated.

tye-exe:
A fun thing that i found out is that this code:
https://github.com/emilk/egui/blob/81b7e7f05a6b03fa2cd5bdc6d4ce5f598e16c628/crates/eframe/src/native/glow_integration.rs#L793-L798
does not get run on window termination. 

This is because `self.integration.should_close()` is false when it is run the first (and only) time.

<details><summary>I've got some output logs showing this (ignore the random prints, i'm just trying anything):</summary>

```
[2025-05-17T20:20:45Z DEBUG eframe::native::glow_integration] Received WindowEvent::CloseRequested for viewport Some("FFFF")
I am either a sub-viewport or epi is not ready to close
ROOT Viewport ID: "FFFF"
[...]

PR #7565, opened 2025-09-26
Properly end winit event loop

tye-exe:
This PR changes eframe to comply with the proper methodology of ending winit. This fixes the winit window remaining open on Wayland (Linux) by allowing winit to properly clean up the window.

* Closes <https://github.com/emilk/egui/issues/6757>

### Details
On a winit "CloseRequested" event eframe drops open windows, which then triggers a "Destroyed" event and allows for winit to clean up resources. In response to the "Destroyed" event, the event loop is exited and control flow returned to user code.
This follows the advice recomended in [this winit issue](https://github.com/rust-windowing/winit/issues/4135) in response to a similar problem.

I have tested the new change using:
- glow on Wayland (Linux)
- wgpu on Wayland (Linux)
- glow on X11 (Linux)
- wgpu on X11 (Linux)
- glow on WASM
- wpgu on WASM

### Current implementation
output1

### After changes
output

### Misc Info
I attempted to run `./scripts/check.sh`, but it fails on the I have branched from on some code that I have not modified.

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -818,21 +818,6 @@ impl GlowWinitRunning<'_> {
                     }
                 }
             }
-
-            // TODO(tye-exe): Do any code-paths reach this event?
-            // As handling of the "CloseRequested" event causes running to be none, which
-            // causes "window_event" to return "Exit" instead of executing this function.
-            winit::event::WindowEvent::Destroyed => {
-                log::debug!(
-                    "Received WindowEvent::Destroyed for viewport {:?}",
-                    viewport_id
-                );
-                if viewport_id == Some(ViewportId::ROOT) {
-                    return EventResult::Exit;
-                } else {
-                    return EventResult::Wait;
-                }
-            }
             _ => {}
         }
 
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -470,7 +470,7 @@ impl WinitApp for GlowWinitApp<'_> {
         if let Some(running) = &mut self.running {
             Ok(running.on_window_event(window_id, &event))
         } else {
-            Ok(EventResult::Wait)
+            Ok(EventResult::Exit)
         }
     }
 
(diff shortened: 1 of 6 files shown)


emilk:
Thanks for workin on this! This looks very promising.

Please run `cargo run -p test_viewports` and make sure that still works as expected 🙏

tye-exe:
It works as best I can tell

https://github.com/user-attachments/assets/1946441f-9324-4fc6-af77-40aa38efb48f

The other example that I was able to run on my computer behaved the same as when they were run on the main branch (excluding "serial_windows", which works as expected on Wayland with this PR).

tye-exe:
I researched into the why the viewport ID check was added and it to resolve [this issue](https://github.com/emilk/egui/issues/3959) with [this commit](https://github.com/emilk/egui/commit/0c528fb862d070baa232409bdddfc1c6ee1cddec).

```rs
if viewport_id == Some(ViewportId::ROOT) {
    return EventResult::Exit;
} else {
    return EventResult::Wait;
}
```

~~However, I cannot verify that this is not a regression as the close button cannot be hidden on Wayland, and the example did not run at all on X11. Although if this the window is recreated (which it appears to be as my closing animation is played each time the button in the example is clicked) on Wayland a crash did not occur.~~

I believe that this is not a regression, as the viewport does close and reopen on Wayland without any issues. However, the example did not run at all on X11, either on the main branch or my branch.

emilk:
I wonder if this also solves
* https://github.com/emilk/egui/issues/1918

tye-exe:
I believe that this issue will be solved, on Linux at least, as the behavior described in that issue matches exactly what I have fixed. 
Though I am not able to verify if it is solved on MacOS, since I do not have access to any Apple devices. I may be able to test on a Windows device in about 10 or so hours, as I do not have any to hand.

emilk:
I tested it on macOS, and this works great:

* Closes https://github.com/emilk/egui/issues/1918
* `cargo run -p test_viewports` still works

nekename:
Thank you so much, @tye-exe!
Because showing the window after it's been hidden is broken in egui on Windows and Wayland, I spent 25mins implementing opening a new window as a solution for those platforms in my project using a loop and a wait for a tokio::sync::Notify.
However when it just went black and unresponsive my heart sank and I was fully ready to abandon the whole effort, even relegating it to a separate branch and scrapping it on main, but I thankfully decided to just look up if there had been any reports of this or fixes.
Turns out I just needed to update from 0.32 to 0.33 :)

PR #7571, opened 2025-09-30
Adjust when we write .diff and .new snapshot images

emilk:


crates/egui_kittest/src/snapshot.rs
```diff
@@ -433,47 +426,55 @@ fn try_image_snapshot_options_impl(
...
+    let below_threshold = num_wrong_pixels as i64 <= *failed_pixel_count_threshold as i64;
+
+    if !below_threshold {
         diff_image
             .save(diff_path.clone())
             .map_err(|err| SnapshotError::WriteSnapshot {
                 path: diff_path.clone(),
                 err,
             })?;
+    }
```
lucasmerlin, line 451: ~We should also early-out when below_threshold.~
Nevermind, then Mode::UpdateAll wouldn't work. I guess we could early out `if below_threshold && mode != Mode::UpdateAll`

PR #7572, opened 2025-09-30
Warn if `DYLD_LIBRARY_PATH` is set and we find no wgpu adapter

emilk:
* See https://github.com/rerun-io/rerun/issues/11351

I had a crazy problem where I could not start any graphics application. Turns out it was because I had added the `DYLD_LIBRARY_PATH` env-var to my `.zshrc`.

This PR adds a warning to make this faster to debug next time.

Also cleaned up the code to replace some `#[cfg(` with `if cfg!`

### DIFF
--- a/crates/egui-wgpu/src/lib.rs
+++ b/crates/egui-wgpu/src/lib.rs
@@ -117,7 +117,7 @@ async fn request_adapter(
                     // I don't want to debug this again.
                     // See https://github.com/rerun-io/rerun/issues/11351 for more
                     log::warn!(
-                        "No wgpu adapter found. This could be because of DYLD_LIBRARY_PATH. Try restarting with DYLD_LIBRARY_PATH=''"
+                        "No wgpu adapter found. This could be because DYLD_LIBRARY_PATH causes dylibs to be loaded that interfere with Metal device creation. Try restarting with DYLD_LIBRARY_PATH=''"
                     );
                 } else {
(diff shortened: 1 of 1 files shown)


crates/egui-wgpu/src/lib.rs
```diff
@@ -108,45 +108,57 @@ async fn request_adapter(
...
+            if cfg!(target_arch = "wasm32") {
+                // Nothing to add here
+            } else if available_adapters.is_empty() {
+                if std::env::var("DYLD_LIBRARY_PATH").is_ok() {
+                    // DYLD_LIBRARY_PATH can sometimes lead to loading dylibs that cause
+                    // us to find zero adapters. Very strange.
+                    // I don't want to debug this again.
+                    // See https://github.com/rerun-io/rerun/issues/11351 for more
+                    log::warn!(
+                        "No wgpu adapter found. This could be because of DYLD_LIBRARY_PATH. Try restarting with DYLD_LIBRARY_PATH=''"
```
Wumpf: a bit vague? Hard to be less vague given that we don't know more. But if I were to hit this in the future without the background we got from looking at this I'd be feeling like someone is not telling me the whole story.
Attempt (not sure if it's that much better)

```suggestion
                        "No wgpu adapter found. This could be because DYLD_LIBRARY_PATH causes dylibs to be loaded that interfere with Metal device creation. Try restarting with DYLD_LIBRARY_PATH=''"
```

PR #7578, opened 2025-10-02
Add support for the safe area on iOS

irh:
This PR is a continuation of #4915 by @frederik-uni and @lucasmerlin that introduces support for keeping egui content within the 'safe area' on iOS (avoiding the notch / dynamic island / menu bar etc.), with the following changes:

- `SafeArea` now wraps `MarginF32` and has been renamed to `SafeAreaInsets` to clarify its purpose.
- `InputState::screen_rect` is now marked as deprecated in favour of either `viewport_rect` (which contains the entire screen), or `content_rect` (which is the viewport rect with the safe area insets removed).
- I added some comments to the safe area insets logic pointing out the [safe area API coming in winit v0.31](https://github.com/rust-windowing/winit/issues/3910).


crates/egui/src/data/input.rs
```diff
@@ -1297,3 +1308,22 @@ impl EventFilter {
...
+pub struct SafeAreaInsets(pub MarginF32);
+
+impl std::ops::Sub<SafeAreaInsets> for Rect {
+    type Output = Self;
+
+    fn sub(self, rhs: SafeAreaInsets) -> Self::Output {
+        Self::from_min_max(
+            self.min + emath::vec2(rhs.0.left, rhs.0.top),
+            self.max - emath::vec2(rhs.0.right, rhs.0.bottom),
+        )
```
lucasmerlin:
```suggestion
        self - rhs.0
```
should also work I believe 🤔

crates/egui/src/input_state/mod.rs
```diff
@@ -574,9 +583,41 @@ impl InputState {
...
+    /// Returns the full area available to egui, including parts that might be partially
+    /// covered, for example, by the OS status bar or notches.
+    ///
+    /// Usually you want to use [`Self::content_rect`] instead.
+    #[deprecated(
+        note = "screen_rect has been renamed to viewport_rect. Consider switching to content_rect."
+    )]
     pub fn screen_rect(&self) -> Rect {
-        self.screen_rect
+        self.viewport_rect
```
lucasmerlin: I think it'd be better to fall back to self.content_rect(), which I feel would fit most of the current uses of screen_rect better.

crates/egui/src/context.rs
```diff
@@ -2660,7 +2668,7 @@ impl Context {
 
     /// Position and size of the egui area.
     pub fn screen_rect(&self) -> Rect {
-        self.input(|i| i.screen_rect()).round_ui()
+        self.input(|i| i.content_rect()).round_ui()
     }
```
emilk, line 2691: I think we should do the same thing here as we do in `InputState`: deprecate `screen_rect` and introduce `fn content_rect` and `fn viewport_rect`

PR #7579, opened 2025-10-02
Update MSRV from 1.86 to 1.88

Wumpf:
Updates Rust version to 1.88. Needed in preparation of updating wgpu.

PR #7580, opened 2025-10-02
Update wgpu to 27.0.0

Wumpf:
* Based on https://github.com/emilk/egui/pull/7579

Tested on:
* MacOS native
* Windows native
* Firefox @ GL
* Chrome @ WebGPU

### DIFF
--- a/crates/egui_kittest/src/wgpu.rs
+++ b/crates/egui_kittest/src/wgpu.rs
@@ -1,11 +1,14 @@
-use std::iter::once;
 use std::sync::Arc;
+use std::{iter::once, time::Duration};
 
 use egui::TexturesDelta;
 use egui_wgpu::{RenderState, ScreenDescriptor, WgpuSetup, wgpu};
 use image::RgbaImage;
 
 use crate::texture_to_image::texture_to_image;
+
+/// Timeout for waiting on the GPU to finish rendering.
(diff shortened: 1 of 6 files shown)


crates/egui_kittest/src/wgpu.rs
```diff
@@ -1,12 +1,15 @@
...
+use std::{iter::once, time::Duration};
 
 use egui::TexturesDelta;
 use egui_wgpu::{RenderState, ScreenDescriptor, WgpuSetup, wgpu};
 use image::RgbaImage;
 
 use crate::texture_to_image::texture_to_image;
 
+/// Timeout for waiting on the GPU to finish rendering.
+pub(crate) const WAIT_TIMEOUT: Duration = Duration::from_secs(1);
```
emilk, line 11: This sounds maybe too low, like in case of a shared instance yadda yadda

Wumpf: Hmm is it though? On windows your driver hard-resets after 2 seconds (unless you change settings in the Windows Registry)

Wumpf: The anomaly we seem to be recently hitting is that on Mac an undetected error occurs, the GPU actually no longer does work and then we get stuck indefinitely.

Not sure what you mean with shared instance :)

Issue #7445, opened 2025-08-12
Allow specifying a custom `Frame` on each widget and container


emilk:
## Related
* Part of https://github.com/emilk/egui/issues/3284
* Unblocks https://github.com/emilk/egui/issues/5605
* https://github.com/emilk/egui/issues/4019

## Description
It's very useful to be able to set an inner and outer margin on any one widget or container. For instance: add some inner margin to a `ScrollArea`, or an outer margin to a `Button`.

In the limit we would like to do so via some style engine plugin (https://github.com/emilk/egui/issues/3284), but being able to just set a `Frame` on a `ScrollArea` etc explicitly or in a hard-coded `Style` would already be a very useful feature.

## Making `Frame` and `UiBuilder` compose
Adding a `UiBuilder::frame(…)` would be very useful.

## Example code (copied from #4019)
Add a `min_size` to `Frame`

```rs
/// A frame around some content, including margin and border colors.
/// 
/// The total (outer) size of a frame is
/// `content_size + inner_margin + border.stroke.width + outer_margin`.
/// 
/// Everything within the border stroke is filled with the border fill color (if any).
/// 
/// ```
/// +---------------^-------------------------------+
/// |               | outer_margin                  |
/// |   +-----------v----^-----------------------+  |
/// |   |                | border stroke width   |  |
/// |   |  +-------------v---^----------------+  |  |
/// |   |  |                 | inner_margin   |  |  |
/// |   |  |  +--------------v-------------+  |  |  |
/// |   |  |  |                            |  |  |  |
/// |   |  |  |                            |  |  |  |
/// |   |  |  | content rect               |  |  |  |
/// |   |  |  +----------------------------+  |  |  |
/// |   |  |                                  |  |  |
/// |   |  +----------------------------------+  |  |
/// |   |       widget rect                      |  |
/// |   +----------------------------------------+  |
/// |           outer rect                          |
/// +-----------------------------------------------+
/// ```
/// 
/// `widget rect` is the interactive part of the widget (what sense clicks etc),
/// and is what is returned by [`Response::rect`].
struct Frame {
    /// Inner spacing. Called `padding` in CSS.
    pub inner_margin: Margin,

    /// Stroke, fill color etc.
    ///
    /// The width of the border is counted as part of the size of the frame.
    pub border: Border,

    /// Optional drop-shadow.
    /// Does not affect margins or sizes.
    pub shadow: Shadow,

    /// Outer spacing. Called `margin` in CSS.
    /// 
[...]

PR #7587, opened 2025-10-06
Make `TextEdit` work with `Atoms`

lucasmerlin:
* part of https://github.com/emilk/egui/issues/7264
* part of https://github.com/emilk/egui/issues/7445

This PR changes the layout within the TextEdit to be done with AtomLayout. It also adds a Prefix and Postfix that allows adding permanent icons / icon buttons within the textedit.

Screenshot 2025-10-06 at 12 25 21

### DIFF
--- a/crates/egui/src/atomics/atom_kind.rs
+++ b/crates/egui/src/atomics/atom_kind.rs
@@ -1,7 +1,7 @@
-use std::fmt::Debug;
 use crate::{FontSelection, Id, Image, ImageSource, SizedAtomKind, Ui, WidgetText};
 use emath::Vec2;
 use epaint::text::TextWrapMode;
+use std::fmt::Debug;
 
 /// The different kinds of [`crate::Atom`]s.
 #[derive(Default)]
@@ -70,7 +70,19 @@ pub enum AtomKind<'a> {
     /// When cloning, this will be cloned as [`AtomKind::Empty`].
     #[doc(hidden)]
     Closure(
-        Box<dyn FnOnce(&Ui, Vec2, TextWrapMode, FontSelection) -> (Vec2, SizedAtomKind<'a>) + 'a>,
+        Box<
+            dyn FnOnce(
+                    &Ui,
+                    Vec2,
+                    TextWrapMode,
+                    FontSelection,
+                ) -> (
+                    Vec2,
+                    // We need 'static here (or need to introduce another lifetime on the enum).
+                    // Other a single 'static Atom would force the closure to be 'static.
+                    SizedAtomKind<'static>,
+                ) + 'a,
+        >,
     ),
 }
 
@@ -108,7 +120,7 @@ impl<'a> AtomKind<'a> {
     }
 
     pub fn closure(
-        func: impl FnOnce(&Ui, Vec2, TextWrapMode, FontSelection) -> (Vec2, SizedAtomKind<'a>) + 'a,
+        func: impl FnOnce(&Ui, Vec2, TextWrapMode, FontSelection) -> (Vec2, SizedAtomKind<'static>) + 'a,
     ) -> Self {
         AtomKind::Closure(Box::new(func))
     }
--- a/crates/egui/src/atomics/atom_kind.rs
+++ b/crates/egui/src/atomics/atom_kind.rs
@@ -1,9 +1,10 @@
+use std::fmt::Debug;
 use crate::{FontSelection, Id, Image, ImageSource, SizedAtomKind, Ui, WidgetText};
 use emath::Vec2;
 use epaint::text::TextWrapMode;
 
 /// The different kinds of [`crate::Atom`]s.
-#[derive(Clone, Default, Debug)]
+#[derive(Default)]
 pub enum AtomKind<'a> {
     /// Empty, that can be used with [`crate::AtomExt::atom_grow`] to reserve space.
     #[default]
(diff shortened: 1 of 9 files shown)


crates/egui/src/atomics/atom_kind.rs
```diff
@@ -57,7 +72,47 @@ pub enum AtomKind<'a> {
     /// }
     /// # });
     /// ```
+    #[deprecated = "Use Atom::custom(id) instead"]
     Custom(Id),
```
emilk: I suggest we put this in a separate PR so it gets a separate entry in the changelog ("Added `AtomKind::Closure` and deprecated `AtomKind::Custom`")

emilk:
The screenshot tests shows _some_ TextEdit shrunk - why is that exactly?

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -565,12 +516,135 @@ impl TextEdit<'_> {
...
         };
-        let mut response = ui.interact(outer_rect, id, sense);
-        response.intrinsic_size = Some(Vec2::new(desired_width, desired_outer_size.y));
+
+        let inner_rect_id = Id::new("text_edit_rect");
+        let mut get_galley = None;
+        let atom_response = {
+            let mut atoms: Atoms<'_> = Atoms::new(());
+
+            let any_shrink = hint_text.iter().any(|a| a.shrink);
```
emilk: Is this a recurring pattern that should have a helper?

```suggestion
            let any_shrink = hint_text.any_shrink();
```

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -565,12 +516,135 @@ impl TextEdit<'_> {
...
+
+            // TODO(servo/rust-smallvec#146): Use extend_right instead of the loop once we have smallvec 2.0
+            for atom in prefix {
+                atoms.push_right(atom);
+            }
+
+            if text.as_str().is_empty() {
```
emilk: Why do we need a special case for this?

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -565,12 +516,135 @@ impl TextEdit<'_> {
...
+        let mut get_galley = None;
+        let atom_response = {
+            let mut atoms: Atoms<'_> = Atoms::new(());
+
+            let any_shrink = hint_text.iter().any(|a| a.shrink);
+
+            // TODO(servo/rust-smallvec#146): Use extend_right instead of the loop once we have smallvec 2.0
+            for atom in prefix {
+                atoms.push_right(atom);
+            }
```
emilk: We already have `Atoms` newtype wrapper that has `.extend_right`
```suggestion
            atoms.extend_right(prefix);
```

emilk, line 576:
```suggestion
            let any_shrink = hint_text.iter().any(|a| a.shrink);
            let mut atoms = prefix;
```

🤷

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -565,12 +516,135 @@ impl TextEdit<'_> {
...
+
+            if text.as_str().is_empty() {
+                let mut shrunk = any_shrink;
+                let mut first = true;
+                hint_text.map_texts(|t| t.color(ui.style().visuals.weak_text_color()));
+                for mut atom in hint_text {
+                    if !shrunk && matches!(atom.kind, AtomKind::Text(_)) {
+                        atom = atom.atom_shrink(true);
```
emilk, line 590:
```suggestion
                        atom = atom.atom_shrink(true); // ellide the hint_text if needed
```

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -565,12 +516,135 @@ impl TextEdit<'_> {
...
+
+            let any_shrink = hint_text.iter().any(|a| a.shrink);
+
+            // TODO(servo/rust-smallvec#146): Use extend_right instead of the loop once we have smallvec 2.0
+            for atom in prefix {
+                atoms.push_right(atom);
+            }
+
+            if text.as_str().is_empty() {
+                let mut shrunk = any_shrink;
```
```suggestion
            if text.as_str().is_empty() {
                // Add hint_text (if any):
                let mut shrunk = any_shrink;
```

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -565,12 +516,135 @@ impl TextEdit<'_> {
...
+                        shrunk = true;
+                    }
+                    if first {
+                        atom = atom.atom_id(inner_rect_id);
+                        first = false;
+                    }
+                    atoms.push_right(atom.atom_align(Align2::LEFT_TOP));
+                }
+                let gallery = layouter(ui, text, allocate_width - margin.sum().x);
+                get_galley = Some(gallery);
```
emilk: This is rather dense and could use some comments or more self-describing code

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -565,12 +516,135 @@ impl TextEdit<'_> {
...
+                    .atom_shrink(clip_text),
+                );
+            }
+
+            if !suffix.is_empty() {
+                atoms.push_right(Atom::grow());
+            }
+            for atom in suffix {
+                atoms.push_right(atom);
+            }
```
emilk, line 659:
```suggestion
            atoms.extend_right(suffix);
```

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -688,75 +760,47 @@ impl TextEdit<'_> {
...
+                                    extra_size,
+                                ),
+                                Sense::hover(),
+                            );
+                        }
                     }
                 }
-            } else {
-                // Avoid an ID shift during this pass if the textedit grow
-                ui.skip_ahead_auto_ids(1);
```
emilk, line 759: Was it intentional to remove this `else`?

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -1149,9 +1189,6 @@ fn events(
         if let Some(new_ccursor_range) = did_mutate_text {
             any_change = true;
 
-            // Layout again to avoid frame delay, and to keep `text` and `galley` in sync.
-            *galley = layouter(ui, text, wrap_width);
```
emilk, line 1153: do we now have a frame delay? And is `text` and `galley` now out of sync?

crates/egui/src/widgets/text_edit/builder.rs
lucasmerlin: I tried, but it wouldn't compile due to lifetime errors, due to the SmallVec issue linked in the todo comment.

lucasmerlin: Same thing, won't compile with lifetime error:
```bash
error: lifetime may not live long enough
   --> crates/egui/src/widgets/text_edit/builder.rs:435:13
    |
432 |     pub fn show(self, ui: &mut Ui) -> TextEditOutput {
    |                 ---- has type `builder::TextEdit<'1>`
...
435 |             prefix,
    |             ^^^^^^ assignment requires that `'1` must outlive `'static`
    |
    = note: requirement occurs because of the type `atoms::Atoms<'_>`, which makes the generic argument `'_` invariant
    = note: the struct `atoms::Atoms<'a>` is invariant over the parameter `'a`
    = help: see <https://doc.rust-lang.org/nomicon/subtyping.html> for more information about variance
```
I don't really get it, but it must be this smallvec issue: https://github.com/servo/rust-smallvec/issues/146
I'll add some more comments explaining why the code is written that way.
There is a smallvec 2.0 alpha release which would fix this I believe, but I'm not sure if we want to depend on alpha crates.

crates/egui/src/widgets/text_edit/builder.rs
lucasmerlin: Same thing here, I'll add another comment

crates/egui/src/widgets/text_edit/builder.rs
lucasmerlin: Good catch, I think that got mixed up when merging main

crates/egui/src/widgets/text_edit/builder.rs
emilk: We can live with a frame-delay if it simplifies the code, but then maybe let's add a `request_discard` so we can avoid it when multi-pass is enabled?

lucasmerlin: Found a nicer fix, which also allows me to remove the hack where we allocated extra space when the text was changed. Also added some tests for this.

lucasmerlin:
> The screenshot tests shows _some_ TextEdit shrunk - why is that exactly?

Previously the style.text_edit_width was used for the inner width, meaning if it was 200, the text edit used slightly more than that for it's margins. Now this is the outer width, meaning it should be exactly 200 in width, which I feel is the correct behavior.

PR #7589, opened 2025-10-06
Fix stuck menu when submenu vanishes

lucasmerlin:
* Closes https://github.com/rerun-io/rerun/issues/11301

This fixes a bug where a menu could get stuck, not closing at all, when the currently open submenu stops being shown.
I also added a way to reproduce this to the demo, as well as a test ensuring that there is no race condition in the fix.


emilk (approved):
awesome!

PR #7590, opened 2025-10-06
Add `Harness::debug_open_snapshot` helper

lucasmerlin:
Adds a helper to quickly see whats going on in a kittest test.

Not all test have snapshots, but when debugging tests it might still be useful to see whats actually going on, so this adds a helper fn that renders a snapshot image to a temporary file and opens it with the default image viewer:

https://github.com/user-attachments/assets/08785850-0a12-4572-b9b5-cea36951081c


crates/egui_kittest/src/snapshot.rs
```diff
@@ -635,6 +635,42 @@ impl<State> Harness<'_, State> {
             }
         }
     }
+
+    /// Render a snapshot, save it to a temp file and open it in the default image viewer.
+    ///
+    /// This method is marked as deprecated to trigger errors in CI (so that it's not accidentally
+    /// committed).
+    #[deprecated = "Only for debugging, don't commit this."]
```
lucasmerlin, line 643: Is there a better way to warn / forbid functions (without the user having to add them to the clippy forbidden fns?)

Cargo.toml
```diff
@@ -113,7 +114,7 @@ wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
 web-sys = "0.3.73"
 web-time = "1.1.0" # Timekeeping for native and web
-wgpu = { version = "27.0.0", default-features = false }
+wgpu = { version = "27.0.0", default-features = false, features = ["std"] }
```
lucasmerlin, line 119: Had a test fail to compile because some egui code wanted to format some error message and Display is only implemented with the std feature.

crates/egui_kittest/src/snapshot.rs
emilk: not that I know

PR #7601, opened 2025-10-07
Clump together wgpu options in a `RendererOptions`

emilk:
This will make it easier to add new options in the future.

Also, we can document struct fields, but not parameters.

### DIFF
--- a/crates/egui-wgpu/src/renderer.rs
+++ b/crates/egui-wgpu/src/renderer.rs
@@ -198,6 +198,12 @@ pub struct RendererOptions {
     ///
     /// Defaults to true.
     pub dithering: bool,
+
+    /// What format to use for the depth and stencil buffers,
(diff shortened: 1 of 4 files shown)


crates/egui-wgpu/src/renderer.rs
```diff
@@ -175,6 +176,46 @@ pub struct Texture {
...
+
+    /// Controls whether to apply dithering to minimize banding artifacts.
+    ///
+    /// Dithering assumes an sRGB output and thus will apply noise to any input value that lies between
+    /// two 8bit values after applying the sRGB OETF function, i.e. if it's not a whole 8bit value in "gamma space".
+    /// This means that only inputs from texture interpolation and vertex colors should be affected in practice.
+    ///
+    /// Defaults to true.
+    pub dithering: bool,
+}
```
lucasmerlin, line 207: Nice!

lucasmerlin (approved):

Issue #5295, opened 2024-10-22
Difference in rendering on windows, linux and macOS


lucasmerlin:
It seems like there are some minor rendering differences between the different egui_wgpu backends. I don't think this is a general problem and I doubt that it's possible to avoid these differences, but I want to document the findings somewhere.
These differences currently cause the snapshots tests to fail on windows and linux, but we should be able to increase the dify threshold until the tests run through. 
Visually the images are the same, I couldn't spot any difference when flipping back and forth between snapshots rendered by the different platforms.

Here is an example diff (the snapshot was rendered on MacOS and the diff on linux):

![widget_gallery diff](https://github.com/user-attachments/assets/6295220d-6622-4f63-8d5a-f762cede057b)

It seems like the difference is mostly on outlines of text and shapes, maybe related to anti aliasing or feathering? 

<details><summary>More diffs</summary>
<p>
Bezier demo (snapshot on macos and diff on linux)

![image](https://github.com/user-attachments/assets/97a39917-59d6-48b9-ae55-a26af6fa3779)

Widget gallery diff (this time the diff was rendered on windows):

![image](https://github.com/user-attachments/assets/bf62a9c7-8027-4f2a-b3b1-b2fb8a8e92e3)

Widget gallery snapshot from windows:

![image](https://github.com/user-attachments/assets/1792ad45-4b2a-4195-9352-7bf00abf0b82)

Widget gallery snapshot from macOS:

![widget_gallery new](https://github.com/user-attachments/assets/12dd59bc-ffc3-4562-a330-205de8c073e2)

</p>
</details> 

Setting the threshold to 0.6 seems to fix all tests except for the bezier demo on linux which seems to need a threshold of 2.1.

Shoutout to @fluxxcode who helped me test this on the different platforms

Wumpf:
For texture sampling in particular this is pretty much expected since different hardware (sometimes even different via driver or driver settings) texture filtering is implemented subtly different. This is mostly due to differences in how they're optimized - texture units typically don't filter in f32 but in their origin format.

Afaik it used to be also a problem with rasterization itself, but with DX11 came more [strict rasterization rules](https://learn.microsoft.com/en-us/windows/win32/direct3d11/d3d10-graphics-programming-guide-rasterizer-stage-rules), although iirc that can still be a problem with mobile hardware sometimes and when dealing with old GL drivers.

Wumpf:
[...]

PR #7602, opened 2025-10-07
Use software texture filtering in kittest

emilk:
My hope is that this will help in making kittest snapshot images come out the same on all hardware.

* Part of https://github.com/emilk/egui/issues/5295
* Will hopefully help https://github.com/rerun-io/rerun/issues/11448

I ran `UPDATE_SNAPSHOTS=force cargo test --all-features --worskapce` on main before, and there was no change (i.e. all images on `main` are already pixel-perfect as they would be generated on my Mac).

After the texture filtering change, all tests still passed, but using `UPDATE_SNAPSHOTS=force` again I see a bunch of smaller changes in some of the images.


crates/egui-wgpu/src/egui.wgsl
```diff
@@ -95,10 +98,35 @@ fn vs_main(
...
+        let pixel_coord = in.tex_coord * texture_size - 0.5;
+        let pixel_fract = fract(pixel_coord);
+        let pixel_floor = floor(pixel_coord);
+
+        // Sample at four pixel centers
+        let texel_size = 1.0 / texture_size;
+        let tl = textureSample(r_tex_color, r_tex_sampler, (pixel_floor + vec2<f32>(0.5, 0.5)) * texel_size);
+        let tr = textureSample(r_tex_color, r_tex_sampler, (pixel_floor + vec2<f32>(1.5, 0.5)) * texel_size);
+        let bl = textureSample(r_tex_color, r_tex_sampler, (pixel_floor + vec2<f32>(0.5, 1.5)) * texel_size);
+        let br = textureSample(r_tex_color, r_tex_sampler, (pixel_floor + vec2<f32>(1.5, 1.5)) * texel_size);
```
aedm: I think `textureSample` still performs a four-tap sampling, or eight-tap if you have miplevels. 
[`textureLoad`](https://www.w3.org/TR/WGSL/#textureload) is the one without filtering.

aedm (changes requested):

crates/egui-wgpu/src/egui.wgsl
Wumpf: entirely depends on the set sampler and if you hit the pixel exactly in the middle as done here it _should_ not do any filtering, but I agree that `textureLoad`would be the much safer bet as you're already working in whole-pixel coordinates anyways

emilk: but then I need to do manual clamping… but sure. Seems safer.

MichaelGrupp:
Hello fellow kittest enjoyers... I think this predictable filtering might have caused some unpredictable issue 😅 https://github.com/emilk/egui/issues/7630

emilk:
D'oh - good catch

PR #7608, opened 2025-10-08
Add some more text edit tests

lucasmerlin:
Adds tests to text the clip option in text edits and how it behaves with a placeholder


lucasmerlin:
~I've removed the snapshots because I wanted to test the update flow, why doesn't it panic due to missing snapshots though? 🤔 Concerning~ Nevermind 🤦🏼 I commited the wrong changes

emilk (approved):

Issue #5889, opened 2025-04-08
Switch to wgpu as default backend of eframe?


emilk:
## Summary
I'd like to switch the default backend of `eframe` from `glow` to `wgpu`, making `glow` opt-in.

## Background
eframe currently has two backends: glow and wgpu (pronounced "wiggipu").

glow is lightweight (few dependencies, compiles quickly), and wgpu is powerful (allows integrating more advanced 3D rendering alongside egui).

Having two backends is annoying for several reasons:
* Duplicated code
* Duplicated maintenance burden
* We're spreading our resources

The last point in particular is what I want to address with this issue.

## Why wgpu?
The broader graphics and GUI community is converging on wgpu as the standard graphics library for Rust. Bug fixes and improvements to wgpu makes everyones lives better.

Switching over to wgpu can also become a forcing function to _improve_ the wgpu backend. For instance, if switching to wgpu increases the .wasm binary size, maybe that will motivate someone to fix it.

### Increased compile times
This one sucks. I like light-weight crates, but it is hard to be light-weight and also powerful. We should measure and see how much worse it gets.

### Increased wasm binary
wgpu shaders are written in WGSL (pronounces "wixel") and then transpiled to GLSL (never pronounced) when targetting the WebGL backend. This transpiling is done with [naga](http://crates.io/crates/naga). Naga is a chonky beast, because it has to be able to transpile _any_ shader. This means the wasm binary size increases a lot.

I see two primary remedies:
* Pre-compile the egui shaders to glsl so we don't need to ship naga
* Only target WebGPU (no tranpilation needed)

WebGPU support is not yet universal, but its getting there: https://caniuse.com/webgpu

Swarkin:
I would really apprechiate some comparisons or benchmarks between wgpu and glow of
- build size
- compile time
- performance (uncapped FPS in the egui demo)

From a very quick local test, using `wgpu` on an almost empty project resulted in twice the binary size, and roughly 1.5x the dependencies. ☹️

lucasmerlin:
> From a very quick local test, using wgpu on an almost empty project resulted in twice the binary size (4,29 MB -> 8.82 MB), and roughly 1.75x the dependencies. ☹️

I assume this was with the wgpu webgl fallback feature enabled, would be interesting how it compares with webgl disabled. It should drastically reduce the build size since it won't need naga anymore to transpile the shaders. In my opinion we should switch it once safari and firefox both fully support webgpu (and then have the webgl feature turned off by default).

[...]

PR #7615, opened 2025-10-09
Make `wgpu` the default renderer for `eframe` and egui.rs

emilk:
* Closes https://github.com/emilk/egui/issues/5889

See the above issue for motivation.

To use glow instead, disable the default features of `eframe` and opt-in to `glow`.

This also changes egui.rs to use wgpu, which means WebGPU when available, and WebGL otherwise


songpola:
Maybe we need to change the docs too? (both egui and eframe)

emilk:
Oops! Thanks for the reminder. PTAL:
* https://github.com/emilk/egui/pull/7719

Issue #7564, opened 2025-09-26
Implement `TextBuffer` for `char`


Erik1000:
I know a `char` is not equivalent to a `str` with a length of 1 but I think effectively from a UI aspect, it is. 
I have structs with fields being a single char and this is a blocker for https://github.com/zakarumych/egui-probe/issues/16

zakarumych:
It's not longer a blocker for https://github.com/zakarumych/egui-probe/issues/16

PR #7621, opened 2025-10-09
Allow TextEdit to represent and validate generic types

tye-exe:
Allow any type implementing the required trait (`TextType`) to be displayed and validated by [TextEdit](https://docs.rs/egui/latest/egui/widgets/text_edit/struct.TextEdit.html).

I am not sure whether the example added for this PR should remain, but other than that, this PR is ready for merging.

Closes:
- <https://github.com/emilk/egui/issues/7564>
- <https://github.com/emilk/egui/issues/1348>
- <https://github.com/emilk/egui/pull/1349>

# Previously

This is still a work in progress.
I am opening this as a draft to get feedback on:
- 1 Is this a change that is still desired
- 2 Implementation details

## Example
I have added an example at `./examples/text_validation/`. This gives a brief showcase towards what the functionality and API of this more generic `TextEdit` might look like. 

## Implementation Details
`TextEdit` has been changed to take a generic parameter that is bounded by `TextType`.
`TextType` requires two functions, one to convert the value into a string, the other to convert the value from a string. I could not use the existing [ToString](https://doc.rust-lang.org/stable/std/string/trait.ToString.html) and [FromString](https://doc.rust-lang.org/stable/std/str/trait.FromStr.html) because `FromString` is [not required](https://doc.rust-lang.org/stable/std/str/trait.FromStr.html#input-format-and-round-tripping) to parse the output from `ToString`.

Users can define custom validation rules by implementing `TextType` on [new types](https://doc.rust-lang.org/rust-by-example/generics/new_types.html).

### Allowing Invalid States (Implemented)
Currently the `TextEdit` can only exist in valid states which is not desirable. A simple example of this is the `char` data type. The only way for a user to edit it currently is for the user to write the new char elsewhere, and then paste it over the existing character. This occurs because a `char` can only contain one character, so if the user attempts to add their new desired character the `TextEdit` becomes invalid and the new character is removed, since there was more than one character. The same sequence occurs if the user attempts to delete the character to add a new one.

My idea to resolve this problem is to allow invalid states while a `TextEdit` is focused, which will be implemented by storing the string representation of the `TextEdit` value using [TextEditState](https://docs.rs/egui/latest/egui/widgets/text_edit/struct.TextEditState.html) (this will require adding a new attribute to `TextEditState`).
While Focused: any valid changes will be written to both the string and the represented value.
While Focused: any invalid changes will be written only to the string.
If the `TextEdit` looses focus while in an invalid state, a valid string representation will be set from the represented value.

This implementation will require all `TextEdit`s that are mutable to use a persistent [Id](https://docs.rs/egui/latest/egui/struct.Id.html) to store the string representation between frames (certain datatypes could be made exempt, See **String Optimisations**).

### String Optimisations
I have not conducted research into which types will be used most often, but I am going to guess that the most common datatype will be `&mut String` (and other equivalents). Without any optimisation for the raw string datatypes, the `TextEdit` will be storing two copies, one as the display string and one as the represented value. This is wasteful.

The idea that I have to combat this issue (if it is needed) is to skip using the display string for unvalidated string data types. This can be accomplished in multiple ways. The two that come to mind are:
- 1 Using generics (`impl TextEdit<String>`) to have string based `TextEdit`s not execute any display string code.
- 2 Create a method for `TextType` that specifies whether a given datatype should use a display string .

## TODO
- [x] Enable temporarily allowing invalid states (See **Allowing Invalid States**).
- [x] Implement `TextType` as both mutable and immutable for core datatypes, such as strings and integers.
- [x] Add comments.
- [x] Add tests(?).
- [x] Publicly export `TextType`.
- [ ] Discuss whether the `TextType` attribute in `TextEdit` should use generics or dynamic dispatch.
- [ ] Discuss `TextBuffer` status in regards to backwards compatibility.
- [ ] Discuss temporarily allowing invalid states (See **Allowing Invalid States**).
- [ ] Discuss optimisations for string representation (See **String Optimisations**).
- [ ] Discuss variable, type, method, and file names.

All todo goals are ones that I can achieve without external help, I just want to know whether any of this is desired before I dedicate more time.

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -523,6 +523,8 @@ impl<Value: TextType> TextEdit<'_, Value> {
         let prev_text = text.as_str().to_owned();
         let hint_text_str = hint_text.text().to_owned();
 
+        let mut text_parsed = None;
+
         let font_id = font_selection.resolve(ui.style());
         let row_height = ui.fonts_mut(|f| f.row_height(&font_id));
         const MIN_WIDTH: f32 = 24.0; // Never make a [`TextEdit`] more narrow than this.
@@ -834,8 +836,12 @@ impl<Value: TextType> TextEdit<'_, Value> {
         // TODO(tye-exe): Simplify once https://github.com/emilk/egui/issues/2142 is fixed
         if response.lost_focus() || response.clicked_elsewhere() {
             match Value::read_from_string(represents, text) {
-                Some(Ok(var)) => *represents = var,
+                Some(Ok(var)) => {
+                    text_parsed = Some(true);
+                    *represents = var
+                }
                 Some(Err(err)) => {
+                    text_parsed = Some(false);
                     log::info!("Failed to parse value for text edit: {err}");
                 }
                 None => {
@@ -851,8 +857,15 @@ impl<Value: TextType> TextEdit<'_, Value> {
 
         if response.changed() {
             // Update represented value if the current state is valid
-            if let Some(Ok(var)) = Value::read_from_string(represents, text) {
-                *represents = var;
+            match Value::read_from_string(represents, text) {
+                Some(Ok(var)) => {
+                    text_parsed = Some(true);
+                    *represents = var;
+                }
+                Some(Err(_)) => {
+                    text_parsed = Some(false);
+                }
+                _ => {}
             }
 
             response.widget_info(|| {
@@ -910,6 +923,7 @@ impl<Value: TextType> TextEdit<'_, Value> {
             text_clip_rect,
             state,
             cursor_range,
+            text_parsed,
         }
     }
 }
--- a/crates/egui/src/widgets/text_edit/output.rs
+++ b/crates/egui/src/widgets/text_edit/output.rs
@@ -21,6 +21,9 @@ pub struct TextEditOutput {
 
     /// Where the text cursor is.
     pub cursor_range: Option<CCursorRange>,
+
+    /// If the user changed the text is it valid for the displayed type.
+    pub text_parsed: Option<bool>,
 }
 
 // TODO(emilk): add `output.paint` and `output.store` and split out that code from `TextEdit::show`.
--- a/examples/text_validation/src/main.rs
+++ b/examples/text_validation/src/main.rs
@@ -1,7 +1,7 @@
 #![cfg_attr(not(debug_assertions), windows_subsystem = "windows")] // hide console window on Windows in release
 #![allow(rustdoc::missing_crate_level_docs)] // it's an example
 
-use eframe::egui::{self, Ui, text_edit::TextType};
+use eframe::egui::{self, TextEdit, Ui, text_edit::TextType};
 
 fn main() -> eframe::Result {
     env_logger::init(); // Log to stderr (if you run with `RUST_LOG=debug`).
@@ -52,8 +52,14 @@ impl eframe::App for MyApp {
 
             ui.horizontal(|ui| {
                 let name_label = ui.label("Your Age: ");
-                ui.text_edit_singleline(&mut self.age)
-                    .labelled_by(name_label.id);
+                let output = TextEdit::singleline(&mut self.age).show(ui);
+                output.response.labelled_by(name_label.id);
+
+                if let Some(valid) = output.text_parsed
+                    && !valid
+                {
+                    ui.label("That can't be my age!");
+                }
             });
 
             ui.horizontal(|ui| {


Mingun:
Without criticizing the approach itself, I would like to note that in order to solve the original problem with `char`, the most logical thing would be to change the editing principle for this type: automatically replace the value with the last character typed / pasted. Thus, there will never be a situation where there will be 2 characters or something needs to be erased.

tye-exe:
> automatically replace the value with the last character typed / pasted.

If you mean in regards to current system, to implement that I could change the `read_from_string` function to take the current value and the display string. This would be useful for situations like the `char` implementation, so the implementation can deduce the character was newly added, though I cannot think of any other situations that it may be useful off the top of my head.

E.G. `fn read_from_string(&self, displayed: &str) -> Option<Result<Self, Self::Err>>`

Mingun:
I mean, the typed characters usually appended to the text input, which creates difficulties for the `char` (because it expects exactly one character). For it the character in text box should be replaced without requiring to manually select it (as you do when you want to replace part of string).

tye-exe:
I've updated the `read_from_string(s: &str)` function to `read_from_strings(previous: &str, modified: &str)`. This allows the char parsing to detect which characters have been modified and behave accordingly.

tye-exe:
Regarding the todo of `Clean up TextBuffer implementations to apply directly to String type`.
I am wondering whether it will be better to switch `TextBuffer` to a newtype around `&mut str` or to leave it as an interface applying to the `String` datatype.

Changing it to a newtype will prevent the current implementations from being applied to all instances of `String`, which could more properly scope the functions. However, this would be a breaking change.

Leaving it as a trait that is implemented by `String` will reduce/eliminate breaking changes depending on the method taken. Breaking changes depend on if `&dyn TextBuffer` is replaced with `&str` in public interfaces, such as `layouter` in `TextEdit`

tye-exe:
Whilst working on "Allowing Invalid States", I came encountered the following issue and its corresponding PR:
- <https://github.com/emilk/egui/issues/2142>
- <https://github.com/emilk/egui/pull/3247>

As suggested by [willbicks](https://github.com/willbicks) in [his response](https://github.com/emilk/egui/issues/2142#issuecomment-2667129157) to the issue, combining `lost_focus()` with `clicked_elsewhere()` allowed for proper detection of focus loss.

Erik1000 (approved):
I like this approach and it would solve issues like i had previously. And I like the example since I have some newtypes that wrap a string and enforce a specific format, this would be covered.

tye-exe:
I would like feedback on whether `TextType::is_mutable` should be renamed to `TextType::is_parseable`.
This would be to better represent the true purpose of the function, which is a performance optimisation of `TextType::read_from_string().is_none()`. Furthermore, there may be multiple types that cannot be parsed from a string (I think that these exist, but I cannot think of any examples).

React with :+1: or :-1:.
:+1: - supporting this change.
:-1: - opposing this change.
Or comment with a more nuanced take.

tye-exe:
After looking through more of the `TextEdit` code, the `TextEditUndoer` in the `TextEditState` creates a clone of the `TextEdit` string for each change that is made. Therefore, attempting to implement optimisations specifically for strings is likely negligible.

Cyrix126:
Thank you for this PR ! I was looking for a solution to not save all my input fields as String when some are not, which clearly was making the code cumbersome.

It would be nice to have a method to easily get the result from read_from_string() on the current inputted value, or getting the represented value directly, to know if it's valid and be able to act on it. (example: change the color to red if it's not).

**EDIT**: I guess you could solve this without change by something like
```rust
let text_edit_output = TextEdit::singleline(&mut self.input).show(ui);
let displayed_text = text_edit_output.state.text;
```

 
For CI fails:
First check to fail should be validated if rerun.
Second check will be valid if a label is added to the PR
Third check was already failing on main.

tye-exe:
> It would be nice to have a method to easily get the result from read_from_string() on the current inputted value, or getting the represented value directly, to know if it's valid and be able to act on it. (example: change the color to red if it's not).

I've added field to `TextEditOutput` that indicates whether a parse was successful (if one occurred), and included this in the example.

Currently this is just a boolean value, but it could be useful to instead have a `Result<(), Box<dyn Error>>` if the error information is required/useful in the dyn Error format.

crates/egui/src/widgets/text_edit/state.rs
```diff
@@ -37,6 +37,12 @@ pub struct TextEditState {
     /// Controls the text selection.
     pub cursor: TextCursorState,
 
+    /// The string displayed.
+    /// This may differ from the current value of the [`TextEdit`] if the string is not a valid value.
+    ///
+    /// [`TextEdit`]: crate::TextEdit
+    pub text: Option<String>,
```
juancampa, line 44: Having a copy here seems like a steep price for everyone not needing to edit structs.

It's also a "multiple source of truth" problem: What happens if the value passed to `TextEdit` is changed externally from one frame to the next. Afaict, the UI would continue showing the old value right?

tye-exe: > Having a copy here seems like a steep price for everyone not needing to edit structs.

I deemed this as negligible compared to the undoer, which keeps multiple copies of the same string.
https://github.com/emilk/egui/blob/fa78d25564a5dbcb546ff6db0a9e14cb603ba03b/crates/egui/src/widgets/text_edit/state.rs#L10

tye-exe: > What happens if the value passed to TextEdit is changed externally from one frame to the next. Afaict, the UI would continue showing the old value right?

You are correct. I will fix this.
I will have a think about how to solve this. There is the obvious solution of reparsing the value every frame it's not selected, but this requires alot of extra work.

tye-exe:
> Took a quick look but my immediate thought is that this could work better as a function that wraps the `TextEdit` to make it opt-in and avoid paying the memory cost when you're just editing a normal string

Can you clarify how this would work?

Issue #7632, opened 2025-10-14
Multiple TextEdits copy content from each other (Chrome Android)


geolehmann:
**Describe the bug**
On Chrome Android, when there are several TextEdits, they copy the text content from each other as soon as you start typing. I tested this on two android devices. On a desktop browser I could not reproduce this issue.

**To Reproduce**
1. Open www.egui.rs on Chrome Android
2. Type some text in the TextEdit from the Widget Gallery
3. Open the TextEdit demo and type text there
4. You´ll see that the text from the first TextEdit is copied to the second TextEdit

**Expected behavior**

**Desktop (please complete the following information):**
 - OS: Android
 - Browser: Chrome Android, latest version

PR #7633, opened 2025-10-14
[eframe/web] Clear hidden <input> when focusing

geolehmann:
This small PR clears the hidden `<input>` element, when it is focused (not the actual TextEdit, just the helper element for Web). This prevents that content from one TextEdit is leaked into another one.

* Closes <https://github.com/emilk/egui/issues/7632


Mingun:
Probably you want to synchronize visible content with input control instead. What if you have only one `TextEdit`? Your fix will clear it each time when application unfocused (for example, when switch to another application)?

PR #7638, opened 2025-10-15
Add `egui_kittest::HarnessBuilder::with_options`

emilk:
Makes it easier to set the same options for many tests


crates/egui_kittest/src/builder.rs
```diff
@@ -56,6 +62,12 @@ impl<State> HarnessBuilder<State> {
         self
     }
 
+    /// Set the default options used for subsequent snapshot tests.
```
lucasmerlin:
```suggestion
    /// Set the default options used for snapshot tests on this harness.
```

PR #7639, opened 2025-10-15
Fix kitdiff links in pr comments

lucasmerlin:
The pr data is not accessible in this workflow so I have to hardcode the url with the pr number


lucasmerlin:
> View snapshot changes at [kitdiff](https://rerun-io.github.io/kitdiff/?url=)

I believe it's just broken here because it's running the workflow as it is on main. Once merged, it should be fixed.

PR #7644, opened 2025-10-16
Minor spelling/grammar updates in egui_kittest's README

benliepert:


benliepert:
Can you add the "egui_kittest" label? then all the checks should pass

Issue #4650, opened 2024-06-10
Add a built-in egui debugger


emilk:
Debugging layout problems and similar in egui is painful.

I'd like to have a mode where egui records events to an event stream containing:

* `Shape`s being painted
* `Ui` creation
* Widget placements
* changes to `min_rect` and `max_rect`
* custom text log events
* drag/click begin/end
* …

This would then be paired with a viewer of these events, where the user can step through them at their own pace and figure out exactly what happened when.

For instance: you hit a special key-combo, and the next egui frame is recorded as an event stream, and a new `egui::Window` or viewport pops up that shows this recording, with controls for stepping through the events one at the time. Each event would be visualized as rectangles and shapes in a canvas area.

TheRedXD:
I think one thing that would be useful on top of this is a way to inspect elements easily (similar to the inspect element in browsers), something like that could definitely help locating elements that are being sized incorrectly

juancampa:
I've been working on an "inspector" plugin that let's you jump from the UI to the code (which is technically going a step further than the browser). Currently it only works on web but it would be pretty easy to extend it for all platforms. Web was actually the hardest since it doesn't provide good call stacks but we work around it by using the DWARF data.

Not sure if there are other efforts on this front but if @emilk is on board I can  clean it up and push a draft PR this week. This can be in its own crate since it's entirely implemented as a plugin, and can be extended to replace the current "Debug on hover"/hold-all-modifiers functionality.

https://github.com/user-attachments/assets/8475f0e0-9541-45e2-8df2-5e83c6af44ea

emilk:
Wow, that looks AMAZING!

Did you see @lucasmerlin's accessibility inspector plugin, https://github.com/emilk/egui/pull/7368 ? There is a lot overlap here, though yours is obviously more feature complete. Are you also using the accessibility labels If so, this PR could also be of interest to you:
* https://github.com/emilk/egui/pull/7386

I'm happy to hear that this can be a standalone crate! Does it belong in the main `egui` repo is the question? 🤔 Probably! That way we can easily test it with egui_demo_app and deploy it to egui.rs

juancampa:
> Wow, that looks AMAZING!

Thanks!

> Are you also using the accessibility labels?

Not at all but wondering if it would make sense to do so 🤔. The inspect UI could show additional information about each widget (including accessibility stuff).

[...]

PR #7652, opened 2025-10-21
Add `Plugin::on_widget_under_pointer` to support widget inspector

juancampa:
This PR adds `Plugin::on_widget_under_pointer` which gets called whenever a widget is created whose rect contains the pointer.

The point of the hook is to capture a stack trace which can be used to map widgets to their corresponding source code so it must be called while the widget is being created. The obvious concern is performance impact. However, since it's only called for rects under the cursor, the effect seems negligible afaict. It's under `debug_assertions` just in case.

This change is needed so we can publish the widget inspector we've been working on. Basically a plugin that allows us to jump from any widget back to their corresponding source code. 

This video shows the plugin configured to open the corresponding code in github, but normally it would open your local editor.

Update: [Live demo](https://membrane-io.github.io/egui/) (Firefox/Safari not yet supported. `Cmd-I` to inspect. `Tab` to cycle filters. `Click` to open). It will try to open a file under `/home/runner/work/egui/egui/` so it won't work, but you get the idea.

https://github.com/user-attachments/assets/afe4d6af-7f67-44b5-be25-44f7564d9a3a

## What's next

After this gets merged I plan to publish the above plugin as its own crate, that way we can iterate and release quickly while things are still changing. I agree it would make sense to eventually merge it into the main egui repo (like @emilk suggested in #4650).

### DIFF
--- a/crates/egui/src/plugin.rs
+++ b/crates/egui/src/plugin.rs
@@ -1,4 +1,4 @@
-use crate::{Context, FullOutput, RawInput, WidgetRect};
+use crate::{Context, FullOutput, RawInput};
 use ahash::HashMap;
 use epaint::mutex::{Mutex, MutexGuard};
 use std::sync::Arc;
@@ -48,7 +48,7 @@ pub trait Plugin: Send + Sync + std::any::Any + 'static {
     /// Useful for capturing a stack trace so that widgets can be mapped back to their source code.
     /// Since this is called outside a pass, don't show ui here. Using `Context::debug_painter` is fine though.
     #[cfg(debug_assertions)]
-    fn on_widget_under_pointer(&mut self, ctx: &Context, widget: &WidgetRect) {}
+    fn on_widget_under_pointer(&mut self, ctx: &Context, widget: &crate::WidgetRect) {}
 }
 
 pub(crate) struct PluginHandle {
@@ -176,7 +176,7 @@ impl PluginsOrdered {
     }
 
     #[cfg(debug_assertions)]
-    pub fn on_widget_under_pointer(&self, ctx: &Context, widget: &WidgetRect) {
+    pub fn on_widget_under_pointer(&self, ctx: &Context, widget: &crate::WidgetRect) {
         profiling::scope!("plugins", "on_widget_under_pointer");
         self.for_each_dyn(|plugin| {
             plugin.on_widget_under_pointer(ctx, widget);


juancampa:
Something that limits how much we can build in this plugin is that we don't want to create new widgets. So the UI must be done in epaint-land. Wondering if creating a separate egui context would make sense to create widgets without affecting the main egui instance, or if that's too hacky for some reason.

crates/egui/src/plugin.rs
```diff
@@ -34,14 +34,20 @@ pub trait Plugin: Send + Sync + std::any::Any + 'static {
...
     /// Useful to inspect or modify the output.
-    /// Since this is called outside a pass, don't show ui here.
+    /// Since this is called outside a pass, don't show ui here. Using `Context::debug_painter` is fine though.
     fn output_hook(&mut self, output: &mut FullOutput) {}
+
+    /// Called when a widget is created and is under the pointer.
+    ///
+    /// Useful for capturing a stack trace so that widgets can be mapped back to their source code.
+    /// Since this is called outside a pass, don't show ui here. Using `Context::debug_painter` is fine though.
+    fn on_widget_under_pointer(&mut self, ctx: &Context, widget: &WidgetRect) {}
```
emilk: Since we only call this in debug builds (which seems smart), we should make that explicit:

```suggestion
    #[cfg(debug_assertions)]
    fn on_widget_under_pointer(&mut self, ctx: &Context, widget: &WidgetRect) {}
```

crates/egui/src/plugin.rs
```diff
@@ -167,6 +173,14 @@ impl PluginsOrdered {
             plugin.output_hook(output);
         });
     }
+
+    #[cfg_attr(not(debug_assertions), expect(dead_code))]
```
emilk:
```suggestion
    #[cfg(debug_assertions)]
```

emilk (approved):
This is a very… specific addition to `trait Plugin`. But it makes perfectly sense for an inspector/debug plugin like yours.

Btw, for callstacks, we have `egui/src/callstack.rs` that is not part of the public API. Maybe it should be? Have you tried using it for your work?

juancampa:
> Have you tried using it for your work?

I was initially using it, yeah. But then realized that immediately resolving stack traces significantly affect performance. Our web app would get slower the more widgets were hovered.

Instead, the plugin calls `trace` N times during the app's update but only `resolve` one of them during pass_end. Similarly for web, we capture N `Error` objects but only read `.stack` once. Browsers lazily generate the stack string so it makes everything faster.

IMO, the best path forward would be to eventually merge the plugin into egui (once it has matured a bit more) and replace the `cmd+shift+ctrl+alt` logic with it.

emilk:
note that the CI needs a little fix!

michalsustr:
Great work!

In the future, it would be great to have the plugin to inspect styling properties from planned https://github.com/emilk/egui/issues/3284, like WidgetStyle etc.

juancampa:
> In the future, it would be great to have the plugin to inspect styling properties

I'm hoping we can do that soon!

I want to go even further and use the DWARF data (which contains struct layouts) to inspect stuff in `egui::Memory`! Kind of similar to `egui_inspect` or `egui_struct` but _without_ having to preemptively decorate any struct with derive macros. Basically do it the same way debuggers do it when you stop at a breakpoint.

We have all the pieces to do that now (except maybe another plugin hook). It's just a matter of putting it together

emilk:
Unfortunately this causes a deadlock for any plugin that adds a UI, because when hovering that ui we will call `on_widget_under_pointer` from within the own plugin

Repro: `cargo r -p egui_demo_app --all-features` + `cmd+alt+i` to open `AccessibilityInspectorPlugin`

juancampa:
We just upgraded our egui fork so I can finally continue this work.

The quick fix would be to flip a boolean in `Plugins::on_begin_pass`/`Plugins::on_end_pass` so that this hook doesn't get called while plugins are potentially creating widgets. This won't let us inspect the UI of plugins though which could be helpful. So instead we can keep an `active_idx: Option<usize>` in `Plugins`, that way we only skip the call for itself.

I'll try the second option and report back.

Issue #7650, opened 2025-10-18
Scroll to zoom without ctrl in `egui::Scene`


darkwater:
**Is your feature request related to a problem? Please describe.**
`egui::Scene` looks good for making an area pannable and zoomable, but by default, regular scrolling results in panning up and down. My use case would benefit more from dedicating the mouse wheel to zoom, kind of like a maps viewer interface.

**Describe the solution you'd like**
A modifier on `egui::Scene` like `.zoom_without_ctrl(true)` or `.mouse_wheel_zooms(true)`. When set, scrolling acts as if ctrl is always held. Maybe it could be flipped such that holding ctrl now pans up and down. Maybe that could be a separate setting. Or maybe there could be a mechanism where the user provides `|mods| if mods.command { Scroll } else { Zoom }`

**Describe alternatives you've considered**
I guess I could find a hacky workaround?

**Additional context**
I'm only considering mouse inputs here. I'm not sure how trackpad gestures work. I'd imagine two-finger panning would pan, and pinch gestures would zoom, and would not be affected by this setting.

hsyed:
I like the granularity `|mods| if mods.command { Scroll } else { Zoom }` provides.  In my prototype component I'd like scroll wheel to be zoom only.

PR #7666, opened 2025-10-26
Add option ```scroll_zooms``` for mouse wheel to zoom instead of scroll inside a ```Scene``` container

Tristan117:
* Closes <https://github.com/emilk/egui/issues/7650>
* 
* [ X] I have followed the instructions in the PR template

This feature allows building image viewers and similar applications using a Scene container where the desired mouse wheel input without a modifier is to zoom rather than scroll/pan. 

Alternatives considered:
1. Swap zoom and scroll variables used in Scene.rs
- Would cause touch input zoom to scroll/pan instead which would be inconsistent with standard user interface interaction
- This also means there is no simple way to swap zoom and vertical scroll modifiers for this use case e.g. ctrl+mousewheel pans vertically while mousewheel alone zooms. What is implemented results in ctrl-mousewheel and mousewheel only being identical inputs.

 2. Modify ctx.option.zoom_modifier
   - Doesn't work because Modifiers:NONE never matches to anything (and therefore disables mouse wheel zoom globally.
   - Refactoring this to provide a different option per container is a large amount of work and does not seem to be a use case that is likely to be used outside of Scene
   - The logic used inside Scene to achieve this is fairly simple and only adds one extra dependency on input_options.scroll_zoom_speed to ensure that the zoom speed setting change is consistent across the application

- [x] tested as part of my application
- [x] run ./scripts/check.sh


Tristan117:
All tests passing. Took a number of additional setup steps (install pngcrush, set options.threshold(OsThreshold::new(0.0).windows(2.1)) for the Bézier Curve demo) to get them working in Windows.

lucasmerlin (commented):
Looks good to me. But I think the name is confusing, maybe rename it to just `scroll_zoom`?
Also, please add a setting to the scene demo, so we have some way to easily test this.

Tristan117:
@lucasmerlin rebased to resolve the merge conflicts. Added example to the demo and changed the flag name to scroll_zooms

Issue #7586, opened 2025-10-04
Better styling: per-widget styles


emilk:
* Part of https://github.com/emilk/egui/issues/3284

As a step towards better stylling I want to introduce per-widget styling options.

### Code (outline)

```rs
/// For generic widgets, that dont have/need their own style struct.
/// Also used by 3rd party widgets.
pub struct WidgetStyle {
    /// Background color, stroke, margin, and shadow.
    pub frame: Frame,

    /// How text looks like
    pub text: TextVisuals,

    /// Color and width of e.g. checkbox checkmark.
    pub stroke: Stroke
}

pub struct ButtonStyle {
    pub frame: Frame,
    pub text: TextVisuals,
}

pub struct CheckboxStyle {
    // frame/margin around checkbox and text
    pub frame: Frame,

    pub text: TextVisuals,

    pub checkbox_width: f32,

    /// How the box is painted
    pub checkbox_frame: Frame,

    /// How the check is painted
    pub stroke: Stroke,
}

pub struct SeparatorSyle {
    /// How much space it allocated
    pub size: f32,

    /// How to paint it
    pub stroke: Stroke,
}
```

…etc.

I'm referencing a new `struct TextVisuals` here:

```rs
struct TextVisuals {
    pub font_id: FontId,
    pub color: Color32,
    // In the future we can expand this to be more like `epaint::TextFormat`.
}
```

Let's also introduce a new enum:

```rs
pub enum WidgetState {
    /// This type of widget cannot be interacted with
    Noninteractive,

    /// An interactive widget that is not being interacted with
    Inactive,

    /// An interactive widget that is being hovered
    Hovered,
    
    /// An interactive widget that is being clicked or dragged
    Active,   
}
```

We then change e.g. `checkbox.rs` to query `egui::Style` about its `CheckboxStyle`:

```
let id = ui.next_auto_id(); // so we can read the interact state before the interaction
let response: OptionResponse = ui.ctx().read_response(id);
let state: WidgetState = response.map(|r| r.widget_state()).unwrap_or(WidgetState::Inactive);
let style: CheckboxStyle = ui.style().checkbox_style(state);
// format text according to `style.text`
// allocate space according to size of galley, and all the margins in `style`
// paint everything
```

This is an important change from how styles are currently computed!
Instead of having a fixed size and then only having different _colors_ based on the `response`, we can now (theoretically) support widgets that grow when hovered.

However, as a start we can just hardcode `checkbox_style`:

```rs
impl Style {
[...]

PR #7667, opened 2025-10-26
Per-widget style

AdrienZianne:
* Closes <https://github.com/emilk/egui/issues/7586>

Implementation of the per-widget style of the action plan to add better styling option to egui

### DIFF
--- a/tests/egui_tests/tests/snapshots/visuals/drag_value.png
+++ b/tests/egui_tests/tests/snapshots/visuals/drag_value.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:de2786af5eb72de0520d987239248bbffa86a84af1f0400182c68115d71f0c2b
-size 8729
+oid sha256:540388365970accbc4c03aa34809b1f07d4e31c8d56bccb8f73da9c2292cb36a
+size 8426


crates/egui/src/widgets/button.rs
```diff
@@ -288,53 +293,69 @@ impl<'a> Button<'a> {
...
+        // We assume that the other WidgetText have already a Fontid and color
+        layout.map_texts(|t| match t {
+            WidgetText::Text(text) => {
+                let rich_text = RichText::new(text.clone())
+                    .font(style.text.font_id.clone())
+                    .color(style.text.color);
+                WidgetText::RichText(Arc::new(rich_text))
+            }
+            w => w,
+        });
```
lucasmerlin: Hmm, I think we need a better way to do this. A richtext without font / color should also use the styles font / color values.

crates/egui/src/widgets/button.rs
```diff
@@ -288,53 +293,69 @@ impl<'a> Button<'a> {
...
             } else {
-                has_frame_margin
-                    && (prepared.response.hovered()
-                        || prepared.response.is_pointer_button_down_on()
-                        || prepared.response.has_focus())
+                layout
+                    .frame(Frame::new().inner_margin(frame.inner_margin))
+                    .min_size(min_size)
+                    .allocate(ui)
             };
```
lucasmerlin: This statement seems uneccessarily complex, the only thing that changes between if / else is the frame.

crates/egui/src/widgets/checkbox.rs
```diff
@@ -97,17 +106,22 @@ impl Widget for Checkbox<'_> {
 
         if ui.is_rect_visible(prepared.response.rect) {
             // let visuals = ui.style().interact_selectable(&response, *checked); // too colorful
```
lucasmerlin: I think we can remove this comment now

crates/egui/src/widgets/checkbox.rs
```diff
@@ -97,17 +106,22 @@ impl Widget for Checkbox<'_> {
...
-            let visuals = *ui.style().interact(&prepared.response);
-            prepared.fallback_text_color = visuals.text_color();
+            prepared.fallback_text_color = style.text.color;
             let response = prepared.paint(ui);
 
             if let Some(rect) = response.rect(rect_id) {
-                let (small_icon_rect, big_icon_rect) = ui.spacing().icon_rectangles(rect);
+                let big_icon_rect = Rect::from_center_size(
+                    pos2(rect.left() + icon_width / 2.0, rect.center().y),
+                    Vec2::splat(style.size),
```
lucasmerlin: It's a bit confusing that you use icon_width and style.size, when they are the same thing. Maybe size should also be renamed to icon_size or checkbox_size.

crates/egui/src/widgets/checkbox.rs
```diff
@@ -55,10 +57,16 @@ impl Widget for Checkbox<'_> {
             indeterminate,
         } = self;
 
-        let spacing = &ui.spacing();
-        let icon_width = spacing.icon_width;
+        // Get the widget style by reading the response from the previous pass
+        let id = ui.next_auto_id();
+        let response: Option<Response> = ui.ctx().read_response(id);
+        let state = response.map(|r| r.widget_state()).unwrap_or_default();
+        let style = ui.style().checkbox_style(state);
```
lucasmerlin: You should destructure the CheckboxStyle here to ensure all the style options are used in the implementation. That would also fix the style.size / icon_width confusion.

crates/egui/src/widgets/button.rs
```diff
@@ -288,53 +293,69 @@ impl<'a> Button<'a> {
 
         let text = layout.text().map(String::from);
 
+        let id = ui.next_auto_id();
+        let response: Option<Response> = ui.ctx().read_response(id);
+        let state = response.map(|r| r.widget_state()).unwrap_or_default();
+
+        let style = ui.style().button_style(state, selected);
```
lucasmerlin: Same here, destructure the ButtonStyle

crates/egui/src/widgets/separator.rs
```diff
@@ -88,11 +88,22 @@ impl Separator {
...
+            mut spacing,
             grow,
             is_horizontal_line,
         } = self;
 
+        // Get the widget style by reading the response from the previous pass
+        let id = ui.next_auto_id();
+        let response: Option<Response> = ui.ctx().read_response(id);
+        let state = response.map(|r| r.widget_state()).unwrap_or_default();
+        let style = ui.style().separator_style(state);
```
lucasmerlin: Same here, also destructure

crates/egui/src/widgets/separator.rs
```diff
@@ -88,11 +88,22 @@ impl Separator {
...
+
+        // override the spacing if not set
+        if spacing == 0.0 && style.spacing != 0.0 {
+            spacing = style.spacing;
+        }
```
lucasmerlin: The default is 6.0 though, so this won't really work. Maybe change the Separator::spacing to be Option<f32> instead?

crates/egui/src/widgets/button.rs
AdrienZianne: I added the font/color style override for richtext (Maybe by modifying richtext I could delay the default font value and check if it's none in the button)

crates/egui/src/widgets/button.rs
AdrienZianne: Yup, simplified :+1:

crates/egui/src/widgets/separator.rs
AdrienZianne: Changed it, but is it okay to change the signature of all pub struct like that ?

AdrienZianne:
The check script result in 3 snapshot diff : 2 are the corner of drag value, but Kitdiff show no difference between old/new so I think it's due to linux and x11, and the last diff is in the fractal clock where a single pixel is a bit brighter but again I think it's linux and x11

crates/egui/src/widgets/button.rs
lucasmerlin: Now it won't be possible to add a custom color via richtext. You should remove this and instead use the fallback color / font on atomlayout: https://docs.rs/egui/latest/egui/struct.AtomLayout.html#method.fallback_text_color

crates/egui/src/widgets/separator.rs
lucasmerlin: Sure, also your change isn't even a breaking one, since the field isn't public.

lucasmerlin (approved):
Awesome, looks good to me now!

PR #7671, opened 2025-10-28
ignore special cut/paste keyboard shortcuts

drewmiller:
Consistent with #5615, we need to use certain special paste options, namely Cmd+Alt+V, Cmd+Shift+V and Cmd+Shift+C to trigger formatted pasting, pasting of formatting, etc. This is a minor update that frees those from being swallowed by the default Copy/Paste operations. We have left it limited to those specific cases, but it may be useful to mandate only `modifiers.command_only()` as a modifier for cut, copy and paste operations.

* [*] I have followed the instructions in the PR template


emilk (approved):

crates/egui-winit/src/lib.rs
```diff
@@ -1107,13 +1107,13 @@ fn is_cut_command(modifiers: egui::Modifiers, keycode: egui::Key) -> bool {
...
 fn is_copy_command(modifiers: egui::Modifiers, keycode: egui::Key) -> bool {
     keycode == egui::Key::Copy
-        || (modifiers.command && keycode == egui::Key::C)
+        || (modifiers.command && !modifiers.shift && keycode == egui::Key::C)
         || (cfg!(target_os = "windows") && modifiers.ctrl && keycode == egui::Key::Insert)
 }
 
 fn is_paste_command(modifiers: egui::Modifiers, keycode: egui::Key) -> bool {
     keycode == egui::Key::Paste
-        || (modifiers.command && keycode == egui::Key::V)
```
emilk, line 1408: Alternative: use `modifiers.command_only` here and above

PR #7672, opened 2025-10-28
Update HOW_TO_UPDATE_SCREENSHOTS for Windows compatibility

bircni:

### DIFF
--- a/crates/egui_kittest/src/snapshot.rs
+++ b/crates/egui_kittest/src/snapshot.rs
@@ -220,8 +220,12 @@ pub enum SnapshotError {
     },
 }
 
+#[cfg(not(target_os = "windows"))]
 const HOW_TO_UPDATE_SCREENSHOTS: &str =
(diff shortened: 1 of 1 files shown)


crates/egui_kittest/src/snapshot.rs
```diff
@@ -220,8 +220,12 @@ pub enum SnapshotError {
     },
 }
 
+#[cfg(not(target_os = "windows"))]
 const HOW_TO_UPDATE_SCREENSHOTS: &str =
     "Run `UPDATE_SNAPSHOTS=1 cargo test --all-features` to update the snapshots.";
+#[cfg(target_os = "windows")]
+const HOW_TO_UPDATE_SCREENSHOTS: &str =
+    "Run `$env:UPDATE_SNAPSHOTS=1; cargo test --all-features` to update the snapshots.";
```
lucasmerlin, line 228: Will this also work for cmd or only for powershell?

bircni: only powershell - but as this is the default in visual studio for windows i think thats enough

PR #7673, opened 2025-10-28
Avoid cloning text in `TextEdit` for `Arc<str>` text buffers

afishhh:
A large chunk of time spent each frame on large immutable `TextEdit`s is cloning the text string *repeatedly*:

~60% time in `TextEdit::show` spent in `str::to_owned`

This PR attempts to switch surrounding code to `Arc<str>` and adds `TextBuffer::clone_to_arc` + `impl TextBuffer for Arc<str>` which allow skipping most of the string copying work for buffers that are themselves `Arc`s.

These changes (and modifying the app to pass `Arc<str>`) make the profile look like this:

~50% time in `TextEdit::show` spent in `str::to_owned`[^1]

Clearly better and decreases the frame times in this test app by ~10ms.

*However* there is one elephant in the room (the remaining `to_owned`) that requires more invasive changes to handle: `LayoutJob`. This elephant is actually very significant: I am able to achieve acceptable performance (in my real-world app that I'm trying to speed up) with a *1MB* XML string if I apply a hacky solution to avoid cloning the string on `LayoutJob` creation. 

Here's what that looks like on the benchmark (https://github.com/emilk/egui/pull/7649 + this + hacky `LayoutJob` fix):

`to_owned` absent in `TextEdit::show`[^1][^2]

This does require changing `LayoutJob` though and I'm not exactly sure what to do here which is why this is a draft. Should I just make `LayoutJob` itself immutable and add a builder type for it? If so what should the builder type's API look like? There's a few possibilities that come to mind but not sure I particularly like any of them.

[^1]: Ignore the sample count, these profiles were run for completely arbitrary amounts of time.
[^2]: Also suggests that the next optimization opportunity is in skipping undo handling for immutable buffers, along with `PaintStats` related stuff which is not visible here but constitutes a large percentage of runtime.

### DIFF
--- a/crates/epaint/src/text/fonts.rs
+++ b/crates/epaint/src/text/fonts.rs
@@ -691,7 +691,7 @@ impl FontsView<'_> {
     ///
     /// The implementation uses memoization so repeated calls are cheap.
     #[inline]
-    pub fn layout_job(&mut self, job: LayoutJob) -> Arc<Galley> {
+    pub fn layout_job(&mut self, job: impl Into<Arc<LayoutJob>>) -> Arc<Galley> {
         let allow_split_paragraphs = true; // Optimization for editing text with many paragraphs.
         self.galley_cache.layout(
             self.fonts,
@@ -852,7 +852,7 @@ impl GalleyCache {
     fn layout_internal(
         &mut self,
         fonts: &mut FontsImpl,
-        mut job: LayoutJob,
+        mut job: Arc<LayoutJob>,
         pixels_per_point: f32,
         allow_split_paragraphs: bool,
     ) -> (u64, Arc<Galley>) {
@@ -878,7 +878,7 @@ impl GalleyCache {
             // * https://github.com/emilk/egui/issues/5084
             // * https://github.com/emilk/egui/issues/5163
 
-            job.wrap.max_width = job.wrap.max_width.round();
+            Arc::make_mut(&mut job).wrap.max_width = job.wrap.max_width.round();
         }
 
         let hash = crate::util::hash((&job, OrderedFloat(pixels_per_point))); // TODO(emilk): even faster hasher?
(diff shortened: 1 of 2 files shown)


afishhh:
Actually for my use case it seems sufficient to make the `layout_job` layout entry point accept an `Arc<LayoutJob>`. That's enough for displaying large amounts of immutable non-line-wrapped text. It's not perfect but it works and doesn't require changing how `LayoutJob` itself works.

crates/epaint/src/text/fonts.rs
```diff
@@ -878,7 +878,7 @@ impl GalleyCache {
             // * https://github.com/emilk/egui/issues/5084
             // * https://github.com/emilk/egui/issues/5163
 
-            job.wrap.max_width = job.wrap.max_width.round();
+            Arc::make_mut(&mut job).wrap.max_width = job.wrap.max_width.round();
```
afishhh, line 881: This is slightly unfortunate and could be avoided by threading a separate `max_width` throughout layout but that sounds like a lot of work while when working with a lot of text you probably don't necessarily need line wrapping anyway.

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -516,7 +516,8 @@ impl TextEdit<'_> {
 
         let font_id_clone = font_id.clone();
         let mut default_layouter = move |ui: &Ui, text: &dyn TextBuffer, wrap_width: f32| {
-            let text = mask_if_password(password, text.as_str());
+            // TODO(afishhh): Keep as `Arc<str>` instead of cloning to a `String`
+            let text = (*mask_if_password(password, text.clone_to_arc())).to_owned();
```
emilk, line 520: I suggest we do this in this PR, or it feels a bit half-finished

afishhh: I do agree but this goes straight into a `LayoutJob` and I'm not sure how exactly to handle `Arc`ifying that.

Should I just split out a builder type and make `LayoutJob` itself immutable?

PR #7676, opened 2025-10-30
Update the multiple_viewports example to show how to use `Context::request_repaint_of()`.

hydra:
I struggled for over an hour to figure out how to force a refresh from either the main thread or a background thread to no avail. Ended up asking in the egui discord server and was pointed at `request_repaint_of`, but there's no usages of this method in the egui codebase, nor did the existing demo showcase this.

https://github.com/user-attachments/assets/0970a7a8-6787-47cf-9b57-8222087822a0


hydra:
Today I encountered a bug (#7686) when using multiple viewports so I updated this PR so it highlights the current behavior with `request_repaint` (or `request_repaint_of`) and a background thread, and `request_repaint_after` and `request_repaint_of` on the main thread.

Thus IMHO regardless of when #7686 is fixed, there's value in merging this PR now so that a) people can discover how to repaint deferred viewports, and b) so that it can be used when testing newer versions of egui and to help prevent regressions.

PR #7669, opened 2025-10-27
Add `is_scrolling`/`is_smooth_scrolling` util, checking for active scroll action.

IsseW:
* Closes #7657
* [x] I have followed the instructions in the PR template

On native this uses a new "touch phase" parameter of the mouse wheel event to know if a scroll action is done.

PR #7678, opened 2025-10-31
Improve modifier handling when scrolling

emilk:
### Problem
Letting go of the modifier key before the last momentum-scroll events arrive will cause the scroll direction to change. This problem can be seen by going to egui.rs and opening the "Scene" example. Hold down shift, start a momentum-scroll (on a Mac trackpad), then quickly let go of shift: you'll see the scroll direction change, which feels wrong.

### Solution
Store the modifiers at the start of the event, thanks to the new `phase` info added in
* https://github.com/emilk/egui/pull/7669

Note that this solution only works on native; not on web.

### Other
* Break out wheel/scroll handling into own file
* Simplify it a lot by deciding late on wether an input is a scroll or a zoom
* Assume input is already smooth if there are `phase` events

### DIFF
--- a/crates/egui/src/input_state/wheel_state.rs
+++ b/crates/egui/src/input_state/wheel_state.rs
@@ -104,7 +104,10 @@ impl WheelState {
                         self.status = Status::Smoothing;
                     }
                     Status::InTouch => {
-                        // keep same modifiers and status
+                        // If the user lets go of a modifier - ignore it.
+                        // More kinematic scrolling may arrive.
+                        // But if the users presses down new modifiers - heed it!
+                        self.modifiers |= latest_modifiers;
                     }
                 }
 


crates/egui/src/input_state/wheel_state.rs
```diff
@@ -0,0 +1,230 @@
...
+                self.modifiers = latest_modifiers;
+            }
+            crate::TouchPhase::Move => {
+                match self.status {
+                    Status::Static | Status::Smoothing => {
+                        self.modifiers = latest_modifiers;
+                        self.status = Status::Smoothing;
+                    }
+                    Status::InTouch => {
+                        // keep same modifiers and status
```
IsseW: I wonder if it will feel a bit odd to not be able to add modifiers after you've started scrolling.

Could potentially do something like `self.modifiers | latest_modifiers` here, depending on what you think

emilk: That's a interesting idea! The main annoyance I have with the current system is that _releasing_ modifier keys to early causes unexpected behavior.

PR #7682, opened 2025-11-04
Allow rotation of rectangles and ellipses

RyanBluth:
Added the ability to rotate rectangles and ellipses. Similar to the existing text implementation


emilk:
It would be nice to have a kittest screenshot test of this in `egui_tests`

RyanBluth:
I added a couple tests

PR #7684, opened 2025-11-05
fix: CodeTheme functions - ui and add builder

Its-Just-Nans:
Following this MR https://github.com/emilk/egui/pull/7375

Without Syntect, the urrent theme selector is `global_theme_preference_buttons`. It should be the dark_theme variable of the local CodeTheme - same as syntect

Tested with

```sh
cargo run
# and
cargo run --features syntect 
```

thanks for this amazing library!

### DIFF
--- a/crates/egui_demo_lib/src/demo/code_editor.rs
+++ b/crates/egui_demo_lib/src/demo/code_editor.rs
@@ -96,7 +96,9 @@ impl crate::View for CodeEditor {
                 .lock_focus(true)
                 .desired_width(f32::INFINITY)
                 .layouter(&mut layouter);
-            let editor = if !cfg!(feature = "syntect") {
+            let editor = if cfg!(feature = "syntect") {
+                editor
+            } else {
                 use egui::Color32;
                 let background_color = if theme.is_dark() {
                     Color32::BLACK
@@ -104,8 +106,6 @@ impl crate::View for CodeEditor {
                     Color32::WHITE
                 };
                 editor.background_color(background_color)
-            } else {
-                editor
             };
             ui.add(editor);
         });


crates/egui_extras/src/syntax_highlighting.rs
```diff
@@ -333,10 +355,6 @@ impl CodeTheme {
         }
     }
 
-    pub fn is_dark(&self) -> bool {
-        self.dark_mode
-    }
-
```
Its-Just-Nans, line 339: This method should be available both with and without synctect since dark_mode is present in both

Its-Just-Nans:
Note that the potential other option is to remove the CodeTheme `dark_mode` (and theme choice) when syntect is disabled

Its-Just-Nans:
This MR also add methods to easily change the font and font size

crates/egui_demo_lib/src/demo/code_editor.rs
```diff
@@ -89,15 +89,25 @@ impl crate::View for CodeEditor {
...
-                    .layouter(&mut layouter),
-            );
+            let editor = egui::TextEdit::multiline(code)
+                .font(egui::TextStyle::Monospace) // for cursor height
+                .code_editor()
+                .desired_rows(10)
+                .lock_focus(true)
+                .desired_width(f32::INFINITY)
+                .layouter(&mut layouter);
+            let editor = if !cfg!(feature = "syntect") {
```
emilk: please swap the if/else branches to avoid the extra `!`

emilk (approved):
Thanks!

crates/egui_demo_lib/src/demo/code_editor.rs
Its-Just-Nans: Good idea, fixed

Thanks

PR #7691, opened 2025-11-07
Change text color of selected text

emilk:
Selected text now gets the color of `visuals.selection.stroke.color`. This means you can have inverted colors for selected text, like in the new test:
image

It also means the color of selected text in labels matches that of the text color of selected buttons.


KonaeAkira:
In the screenshot you posted, it looks like the closing bracket part of of the smiley isn't affected by `visuals.selection.stroke.color`. That's unintentional, right?

emilk:
Good catch @KonaeAkira - I should have used more distinctive colors:
* https://github.com/emilk/egui/issues/7865

PR #7693, opened 2025-11-09
test_viewports: fix assertion

YichiZhang0613:
These assertions allows col == COLS, while when col == COLS, array may be out of bounds. In `fn init`, `for i in 0..COLS {self.insert(...` confirms the assertions' predicate col <= COLS should be changed into col < COLS.


tests/test_viewports/src/main.rs
```diff
@@ -339,7 +339,7 @@ fn drag_and_drop_test(ui: &mut egui::Ui) {
         }
 
         fn insert(&mut self, container: Id, col: usize, value: impl Into<String>) {
-            assert!(col <= COLS, "The coll should be less then: {COLS}");
+            assert!(col < COLS, "The coll should be less then: {COLS}");
```
emilk:
```suggestion
            assert!(col < COLS, "The coll should be less than: {COLS}");
```

tests/test_viewports/src/main.rs
```diff
@@ -355,7 +355,7 @@ fn drag_and_drop_test(ui: &mut egui::Ui) {
         }
 
         fn cols(&self, container: Id, col: usize) -> Vec<(Id, String)> {
-            assert!(col <= COLS, "The col should be less then: {COLS}");
+            assert!(col < COLS, "The col should be less then: {COLS}");
```
emilk:
```suggestion
            assert!(col < COLS, "The col should be less than: {COLS}");
```

tests/test_viewports/src/main.rs
```diff
@@ -368,7 +368,7 @@ fn drag_and_drop_test(ui: &mut egui::Ui) {
             let Some(id) = self.is_dragged.take() else {
                 return;
             };
-            assert!(col <= COLS, "The col should be less then: {COLS}");
+            assert!(col < COLS, "The col should be less then: {COLS}");
```
emilk:
```suggestion
            assert!(col < COLS, "The col should be less than: {COLS}");
```

emilk (approved):

tests/test_viewports/src/main.rs
YichiZhang0613: OK.

PR #7694, opened 2025-11-10
Replace ab_glyph with Skrifa + vello_cpu; enable font hinting

valadaptive:
* Closes N/A

I'll probably come back to this and clean it up a bit. This PR reimplements ab_glyph's functionality on top of Skrifa, a somewhat lower-level font API that's being used in Chrome now.

Skrifa doesn't perform rasterization itself, so I'm using [vello_cpu](https://github.com/linebender/vello) from the Linebender project for rasterization. It's still in its early days, but I believe it's already quite fast. It also supports color and gradient fills, so color emoji support will be easier.

Skrifa also supports font hinting, which should make text look a bit nicer / less blurry.

Here's the current ab_glyph rendering:

image

Here's Skrifa *without* hinting--it looks almost identical, but there are some subpixel differences, probably due to rasterizer behavior:

image

Here's Skrifa  *with* hinting:

image

Hinting does make the horizontal strokes look a bit bolder, which makes me wonder once again about increasing the font weight from "light" to "regular".

### DIFF
--- a/crates/epaint/src/text/fonts.rs
+++ b/crates/epaint/src/text/fonts.rs
@@ -767,7 +767,7 @@ impl FontsImpl {
         let atlas = TextureAtlas::new([texture_width, initial_height], options);
 
         let mut fonts_by_id: nohash_hasher::IntMap<FontFaceKey, FontFace> = Default::default();
-        let mut font_impls: ahash::HashMap<String, FontFaceKey> = Default::default();
+        let mut fonts_by_name: ahash::HashMap<String, FontFaceKey> = Default::default();
         for (name, font_data) in &definitions.font_data {
             let tweak = font_data.tweak;
             let blob = blob_from_font_data(font_data);
@@ -775,14 +775,14 @@ impl FontsImpl {
                 .unwrap_or_else(|err| panic!("Error parsing {name:?} TTF/OTF font file: {err}"));
             let key = FontFaceKey::new();
             fonts_by_id.insert(key, font_face);
-            font_impls.insert(name.clone(), key);
+            fonts_by_name.insert(name.clone(), key);
         }
 
         Self {
             definitions,
             atlas,
             fonts_by_id,
-            fonts_by_name: font_impls,
+            fonts_by_name,
             family_cache: Default::default(),
         }
     }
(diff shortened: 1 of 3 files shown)


emilk:
I'm very excited for this! The readability improvements of the font hinting is pretty big imho. Also: colored emojis unlocked???

Before we merge I want to make sure we measure what impact this has on the .wasm blob size (`./scripts/build_demo_web.sh --release` before/after).

nicoburns:
Regarding the quality of the rendering, I have found that adding "font embolden" (aka stem darkening) to the skrifa-based renderers is (subjectively) a big additional improvement, at least on macOS (by tweaking the embolden parameter I can make the rendering very close if not identical to Skia/Chrome).

There's no implementation for `vello_cpu` yet, but there's a [PR for Vello Classic](https://github.com/linebender/vello/pull/823), and in theory a CPU-side implementation ought to be easier.

drewmiller:
Hey I just wanted to throw a vote behind this. Is there something blocking we at @tritium-legal can help with? Hinting on normal DPI platforms just makes a massive difference.

emilk:
wasm size on `main`: 10,478,576 B
wasm size with `skrifa`: 11,080,156 B

so +600kB. I think that's well acceptable.

Cargo.lock
```diff
@@ -4233,7 +4327,7 @@ version = "0.15.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "68c7541fff44b35860c1a7a47a7cadf3e4a304c457b58f9870d9706ece028afc"
 dependencies = [
- "kurbo",
+ "kurbo 0.11.1",
```
emilk, line 4330: It's a shame we get a duplicate `kurbo` because of `resvg`, and `resvg` hasn't had a crate release in over 8 months :/

waywardmonkeys: We can fix that. I am working on a round of releases this month.

crates/egui_demo_lib/tests/snapshots/widget_gallery_dark_x2.png
emilk, line 1: Something is very wrong here! I think there is a bug when `pixels_per_point != 1.0`

emilk:
All snapshots look good

emilk:
I encourage everyone to follow this link to experience the improvements: https://rerun-io.github.io/kitdiff/?url=https://github.com/emilk/egui/pull/7694

PR #7696, opened 2025-11-10
Allow multiple atoms in `Button::shortcut_text` and `right_text`

emilk:
Useful when intermixing text and icons (e.g. for modifiers)


IsseW (approved):
Thanks!

PR #7698, opened 2025-11-10
[egui-wgpu] Put the `capture` module behind a feature flag, make the `egui` dependency optional

StT191:
This PR enables users of `egui-wgpu` to render `epaint` primitives without having to bring in the complete `egui` crate and all it's dependencies.

### DIFF
--- a/crates/egui-wgpu/Cargo.toml
+++ b/crates/egui-wgpu/Cargo.toml
@@ -27,7 +27,7 @@ rustdoc-args = ["--generate-link-to-definition"]
 [features]
 default = ["fragile-send-sync-non-atomic-wasm", "macos-window-resize-jitter-fix", "wgpu/default"]
 
-## Enables the `capture` module.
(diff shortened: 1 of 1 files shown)


crates/egui-wgpu/Cargo.toml
```diff
@@ -27,8 +27,11 @@ rustdoc-args = ["--generate-link-to-definition"]
 [features]
 default = ["fragile-send-sync-non-atomic-wasm", "macos-window-resize-jitter-fix", "wgpu/default"]
 
+## Enables the `capture` module.
```
emilk:
```suggestion
## Enables the `capture` module for capturing screenshots.
```

StT191:
It's been a great experience contributing to this repository - Thanks!

I might do a similar PR for the `egui-glow` backend if that is desirable? Although I don't use it it might be useful to others.

PR #7701, opened 2025-11-11
Remove `accesskit` feature and always depend on `accesskit`

### LINKED ISSUE #3137: Remove `accesskit` feature in `egui`, and always depend on it

emilk:
* Closes #3137

With this, `egui` will always depend on `accesskit`, removing a lot of `#[cfg(feature = "accesskit")]` throughout the code.

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -888,7 +888,6 @@ impl State {
             events: _,                    // handled elsewhere
             mutable_text_under_cursor: _, // only used in eframe web
             ime,
-            #[cfg(feature = "accesskit")]
             accesskit_update,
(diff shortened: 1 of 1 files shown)


emilk:
It indeed adds almost 100kB, from 7 367 kB to 7 451 kB, but I think its worth it not to have all those feature flags

PR #7702, opened 2025-11-11
egui-wgpu: attach stencil buffer

jgraef:
I'm using eframe to run an app with the wgpu backend and built a custom widget that renders a 3D scene. I want to use the stencil buffer. eframe allows me to specify the depth and stencil buffer and egui-wgpu chooses the right texture format. It does attach the depth texture when rendering, but leaves the stencil attachment empty.

It makes sense to me that if a depth/stencil texture is created both are attached.

I tested that this doesn't break anything if the depth texture doesn't have a stencil part. I ran all tests and only 2 snapshot tests failed with superficial diffs. I also ran the hello_world app with the wgpu feature to make sure an app without a depth texture would still work.

`./scripts/check.sh` doesn't run for me (`typos` not found)

Is this something you want to merge into egui-wgpu at all? If so, please let me know if there is more I can do to make this easier to merge.

I can understand if this is a niche use case (in contrast to depth buffer only), but the only workaround would be to create my own render pass, render to a texture and then blit the result. For more complex setups this might be the way to go anyway.

### DIFF
--- a/crates/eframe/src/web/web_painter_wgpu.rs
+++ b/crates/eframe/src/web/web_painter_wgpu.rs
@@ -244,7 +244,6 @@ impl WebPainter for WebPainterWgpu {
                         wgpu::RenderPassDepthStencilAttachment {
                             view,
                             depth_ops: self
-                                .options
                                 .depth_stencil_format
                                 .is_some_and(|depth_stencil_format| {
                                     depth_stencil_format.has_depth_aspect()
@@ -256,7 +255,6 @@ impl WebPainter for WebPainterWgpu {
                                     store: wgpu::StoreOp::Discard,
                                 }),
                             stencil_ops: self
-                                .options
                                 .depth_stencil_format
                                 .is_some_and(|depth_stencil_format| {
                                     depth_stencil_format.has_stencil_aspect()


jgraef:
> if there is no stencil attachment, the stencil ops musn't be set

The same applies for the depth aspect, which the current implementation doesn't check. `depth_format_from_bits` will create a stencil-only texture. I'm adding checks against `self.options.depth_stencil_format` for both.

> For consistency the web backend should do the same as well.

Oh I see the code is duplicated in `eframe` for web.

> Generally I'm a bit skeptical about the fact that egui_wgpu provides a depth(/stencil) buffer integration in the first place.

I agree.

> Instead it should just be easier to own the final render output yourself for situations where you need a full screen depth buffer etc.

Can you clarify what you mean by that? Do you mean `egui-wgpu` should just provide the target texture to the callback trait and have it create the render pass itself?

jgraef:
> Alternatively, we could make more of these settings things that one can optionally configure on the renderer

Yes, I missed that a bit. Also you have to dig through the source code to figure out how the render pass is actually being configured. But the defaults work well enough I think and one can always clear the depth-stencil buffer with whatever they want with a single draw call (which you might do anyway to set the background color).

Wumpf:
> Can you clarify what you mean by that? Do you mean egui-wgpu should just provide the target texture to the callback trait and have it create the render pass itself?

tbh I haven't thought it through entirely, but I'm thinking that the egui/eframe/wgpu integration should make it easier to just swap out the entire "main render pass" setup against something custom without having to redo half of eframe from scratch as it's the case now :/

crates/eframe/src/web/web_painter_wgpu.rs
```diff
@@ -243,13 +243,28 @@ impl WebPainter for WebPainterWgpu {
...
-                                // It is very unlikely that the depth buffer is needed after egui finished rendering
-                                // so no need to store it. (this can improve performance on tiling GPUs like mobile chips or Apple Silicon)
-                                store: wgpu::StoreOp::Discard,
-                            }),
-                            stencil_ops: None,
+                            depth_ops: self
+                                .options
+                                .depth_stencil_format
+                                .is_some_and(|depth_stencil_format| {
+                                    depth_stencil_format.has_depth_aspect()
```
Wumpf, line 249: really nice catch there, thank you!

Wumpf (approved):
looks great, thank you!

jgraef:
I rebased so it works with 0.33.2. Is there anything blocking this?

Wumpf:
ah sorry, I forgot about this! I think at the time CI wasn't green yet. And looks like it still comes back red right now, this needs fixing up

jgraef:
I didn't check if it still compiles for web after rebasing. The `options` field on `WebPainterWgpu` was removed, but a `depth_stencil_format` added. Fixed it. Also checked that the demo builds this time :)

Wumpf:
thanks again!

PR #7710, opened 2025-11-12
Prevent widgets sometimes appearing to move relative to each other

emilk:
Sometimes when moving a window, having a tooltip attached to the mouse pointer, or scrolling a `ScrollArea`, you would see this disturbing effect:

drift-bug

This is caused by us rounding many visual elements (lines, rectangles, text, …) to physical pixels in order to keep them sharp. If the window/tooltip itself is not rounded to a physical pixel, then you can get this behavior.

So from now on the position of all areas/windows/tooltips/popups/ScrollArea gets rounded to the closes pixel.

* Unlocked by https://github.com/emilk/egui/pull/7709


lucasmerlin (approved):
Nice!

Seems like the same thing is also still happening in ScrollAreas and when moving around Areas with ScrollAreas, but this is already a great improvement:

https://github.com/user-attachments/assets/666b56fb-65a3-4f35-83bc-69bdcd6d07ed

Only seems to happen at odd zoom factors though.

emilk:
True, I should also do the same thing for scroll areas!

Issue #5605, opened 2025-01-14
Remove `clip_rect_margin`


emilk:
It's an ugly hack. To remove it we need to use `Frame` for all widgets, and remove `expansion`.

* Will solve https://github.com/emilk/egui/issues/3385
* Part of #3284 
* Related to #4019

emilk:
It is now zero by default. We can deprecate and remove this at any time now.

Issue #3385, opened 2023-09-25
ScrollAreas can overrun their container's margins


valadaptive:
**Describe the bug**

When displaying a ScrollArea inside a container, and it only scrolls along one axis, it will be a few pixels too big and will overflow into that container's margins along that axis.

This is easiest to see when we set the container's margins to 0. Here's an example with a ScrollArea containing the grey rectangle in the central panel, surrounded by 4 panels:

![Peek 2023-09-25 01-09](https://github.com/emilk/egui/assets/79560998/ab22a9fd-bba6-43f8-8991-9eacb4216058)

I haven't delved into the code much, but it seems like the scroll area overruns the container margin by `clip_rect_margin` pixels. Setting `clip_rect_margin` to 0.0 before rendering the scroll area fixes the issue.

**To Reproduce**

See the following demo code, which renders the UI shown above:
<details>
<summary>Demo code</summary>
```rust
use eframe::egui;

fn main() -> Result<(), eframe::Error> {
    let options = eframe::NativeOptions {
        initial_window_size: Some(egui::vec2(480.0, 360.0)),
        ..Default::default()
    };
    eframe::run_native(
        "My egui App",
        options,
        Box::new(|_cc| Box::<MyApp>::default()),
    )
}

#[derive(Default)]
struct MyApp {}

impl eframe::App for MyApp {
    fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
        let no_margins = egui::Frame::side_top_panel(&ctx.style()).inner_margin(0.0);
        egui::TopBottomPanel::top("top").show(ctx, |_| {});
        egui::TopBottomPanel::bottom("bottom").show(ctx, |_| {});
        egui::SidePanel::left("left").show(ctx, |_| {});
        egui::SidePanel::right("right").show(ctx, |_| {});
        egui::CentralPanel::default()
            .frame(no_margins)
            .show(ctx, |ui| {
                egui::ScrollArea::new([true; 2]).show(ui, |ui| {
                    ui.add(
                        egui::Image::from_texture((
                            egui::TextureId::default(),
                            egui::vec2(360.0, 240.0),
                        ))
                        .uv(egui::Rect::from_min_max(
                            egui::pos2(0.0, 0.0),
                            egui::pos2(0.0, 0.0),
                        ))
                        .tint(egui::Color32::GRAY),
                    );
                });
            });
    }
}
```
</details>

**Expected behavior**
The ScrollArea should properly resize to fit within its container.

[...]

PR #7722, opened 2025-11-16
Add `ScrollArea::content_margin`

emilk:
* Part of https://github.com/emilk/egui/issues/3385


crates/egui/src/containers/scroll_area.rs
lucasmerlin, line 781: Can't we just do `content_max_rect - content_margin` here, to avoid the extra nested child `Ui`?

emilk: oh yeah - good idea!

emilk: no, that affects `inner_rect` and so the scroll bars move…

emilk: Too much hassle. I'll leave it as a future optimization 😞

Issue #3524, opened 2023-11-03
Create all `Ui`:s from a parent `Ui`


emilk:
## Summary
I would like each egui viewport (read: native window) to have a single root `Ui` that covers the entire available space.
The user can then add `egui::Window`s to this `Ui`, as well as sub-divide it using panels.
The root `Ui` would have no background color, and no margin.
To get that, the user can use a `Frame`.

## Benefits
The `Context::create_viewport*` functions coming in https://github.com/emilk/egui/pull/3172 would likewise return a `Ui`.
If possible, these `Ui`s will fill a new native window, and otherwise they will instead fill an `egui::Window` floating on top of the parent viewport. This makes their interfacer a lot simpler. Basically `ctx.create_viewport(…, |ui| ui.label("Hello World"));` would show the new ui either in a separate native window (if the backend supports it), or fall back to a `egui::Window` if not.

We can also simplify panels.
`CentralPanel` can be removed (replaced by `Frame`) and the `show/show_inside` dichotomy would disappear.

`Area` and `Window` should float on top of the entire viewport by default, and a user should be able to create them at any time.

## Current behavior
There are currently a few different ways to create a `Ui` from a `Context`:

* Panels (`CentralPanel`, `SidePanel`, `TopBottomPanel`)
* `Area` an `Window`
* `Ui::new`

For instance, `SidePanel::left(…).show(ctx, |ui| { … })`.
The panels also have special `show_inside` functions to use them inside another `Ui`: `SidePanel::left(…).show_inside(ui, |ui| { … })`.

konkitoman:
From my point of view the `Context::create_viewport*` is like a internal!
The `egui::Window` should have an embedded property and based on that try to create a viewport!

invokermain:
This sounds really useful, would this allow crates to ship a egui 'ui' that the user can then provide a window/viewport to render it in?

emilk:
Check what is left to do. Note to self: maybe this can be closed now?

PR #7724, opened 2025-11-17
Use `Panel::show_inside` over `show` in demos

emilk:
* Must merge https://github.com/emilk/egui/pull/5659 first
* Part of https://github.com/emilk/egui/issues/3524

I'm sick and tired of the code duplication in `Panel`, and how there are two types of panels, that are really the same. Time to fix https://github.com/emilk/egui/issues/3524


emilk:
Closed in favor of https://github.com/emilk/egui/pull/7729

PR #7725, opened 2025-11-17
Rework `Window` margins and set `clip_rect_margin` to zero

### LINKED ISSUE #3385: ScrollAreas can overrun their container's margins

emilk:
* Follows https://github.com/emilk/egui/pull/7722
* Closes https://github.com/emilk/egui/issues/3385

## What
Sets `clip_rect_margin` to zero, and moves the margin of `Window`s with `ScrollAreas`, so that the scroll bars are now on the very edge of the windows they are in.

Windows with a bulit-in scroll area now lets the content go all the way to the edges (left image).
However, if you just manually add a `ScrollArea` to a `Window`, you won't get this effect (right image).
Screenshot 2026-05-18 at 22 04 01

## Required
* #7803
* #7804
* #7805
* #7806
* https://github.com/emilk/egui/pull/7807
* https://github.com/emilk/egui/pull/7808


crates/egui/src/containers/window.rs
```diff
@@ -576,10 +576,17 @@ impl Window<'_> {
...
+                        let add_contents_with_margin = |ui: &mut Ui| {
+                            crate::Frame::NONE
+                                .inner_margin(ui.spacing().window_margin)
+                                .show(ui, add_contents)
+                                .inner
+                        };
+
                         if scroll.is_any_scroll_enabled() {
-                            scroll.show(ui, add_contents).inner
+                            scroll.show(ui, add_contents_with_margin).inner
```
lucasmerlin: Why not
```suggestion
                            scroll.content_margin(ui.spacing().window_margin).show(ui, add_contents).inner
```
?

emilk:
Oops - the windows headers shrunk… a lot. Good with screenshot tests 😅

emilk:
Oh no… I need to do the resize-corner interaction on top of the scroll area, or they will fight with each-other

emilk:
I also need to tweak `should_prioritize_hits_on_back`… or change the size of the resize corner or something.

crates/egui/src/style.rs
```diff
@@ -1029,7 +1029,9 @@ pub struct Visuals {
     /// How the text cursor acts.
     pub text_cursor: TextCursorStyle,
 
-    /// Allow child widgets to be just on the border and still have a stroke with some thickness
+    /// Allow widgets to paint this much outside the scroll area rect.
+    ///
+    /// Legacy. Should not be used anymore.
```
Mingun, line 1079: Probably you should say, what to use instead

crates/egui/src/containers/frame.rs
```diff
@@ -192,9 +192,9 @@ impl Frame {
         Self::new().inner_margin(8).fill(style.visuals.panel_fill)
     }
 
+    /// Does NOT include margin ([`crate::Window`] applies that manually).
```
lucasmerlin: I think this is very unexpected and will probably break some places that overrode the margin by customizing the `Frame`. Instead, can we keep the margin here and read and then clear it before showing the Frame within window? I feel that would be cleaner, and preserve the ability to easily override window margin.

lucasmerlin (changes requested):

crates/egui/src/containers/frame.rs
emilk: Very good point

crates/egui_kittest/tests/snapshots/test_scroll_initial.png
emilk, line 1: Example of a nice win: no drawing painting the lines

lucasmerlin (approved):
Awesome, finally!

PR #7727, opened 2025-11-17
[issue - dont merge] Button::is_pointer_button_down_on returns false

usbalbin:
**Describe the bug**

`Button::is_pointer_button_down_on()` starts returning false after a while when still holding down the mouse button when the gui is refreshed(for example by the blinking cursor shown in the image)

**To Reproduce**
1. In the example modified in the PR
2. Click into the "Name" field so the cursor blinks
3. Hold down on the button, this now shows "Hello '{}', age {}"
4. Wait a second while keeping the button pressed
5. Boom - egui thinks the button is released while it is not

**Expected behavior**

I would have expected the button to not be released

**Screenshots**

egui-bug

The mouse is held over the button with the left mouse button kept down. I release the mouse which can be seen when the light border appears on the button

 - OS: 6.16.8-1-MANJARO with KDE X11

**Additional context**
Same thing on the clickable label works fine, no issue.

### DIFF
--- a/examples/hello_world/src/main.rs
+++ b/examples/hello_world/src/main.rs
@@ -45,10 +45,15 @@ impl eframe::App for MyApp {
                     .labelled_by(name_label.id);
             });
             ui.add(egui::Slider::new(&mut self.age, 0..=120).text("age"));
-            if ui.button("Increment").clicked() {
+            if ui.button("Increment").is_pointer_button_down_on() {
                 self.age += 1;
+                ui.label(format!("Hello '{}', age {}", self.name, self.age));
             }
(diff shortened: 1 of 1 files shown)


usbalbin:
Same problem on Win11

lucasmerlin:
The culprit is long press detection. 

You can add `Sense::click_or_drag` to work around this:
```rs
            let response = ui.add(Button::new("Increment").sense(Sense::click_and_drag()));

            if response.is_pointer_button_down_on() {
                self.age += 1;
                ui.label(format!("Hello '{}', age {}", self.name, self.age));
            }
```

PR #7731, opened 2025-11-19
Winit 0.31 draft

lucasmerlin:
I was curious how much work the winit update would be, so I tried getting hello world simple to run. 
It doesn't run yet but I got it to compile. Looks like this will be a lot of effort to get perfect. (Maybe time to remove the glow backend since the two backends add a lot of complexity?).

I commented a ton of code to try to get something to compile, so this will need a _lot_ more work. Opening this mostly in case someone wants to pick up the work or discuss how we should do things.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -109,7 +109,7 @@ dependencies = [
  "accesskit_unix",
  "accesskit_windows",
  "raw-window-handle",
- "winit",
+ "winit 0.30.12",
 ]
 
 [[package]]
@@ -648,6 +648,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "borsh"
+version = "1.5.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ad8646f98db542e39fc66e68a20b2144f6a732636df7c2354e74645faaa433ce"
+dependencies = [
+ "cfg_aliases",
+]
+
+[[package]]
 name = "bstr"
 version = "1.11.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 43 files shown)


nicoburns:
Blitz `main` has now upgraded to Winit 0.31. The PR that we merged is here: https://github.com/DioxusLabs/blitz/pull/323, but that includes adding support for Android and iOS as well. PR with only Winit-related changes is (but not quite all of the changes we finally applied) is here https://github.com/DioxusLabs/blitz/pull/298 

Some notes:
- We replaced the "UserEvent" functionality with a channel (std::mpsc). The architecture doesn't change much here. It just uses a custom channel that's read on the "proxy wake up" event rather than having Winit manage the channel.
- I've published https://docs.rs/accesskit_xplat which is a fork of `accesskit_winit` without depending on Winit (so it can be used with any Winit version).
- Other than that it was mostly just renaming things and adding `dyn` in a bunch of places.

Eclextic:
> I was curious how much work the winit update would be, so I tried getting hello world simple to run. It doesn't run yet but I got it to compile. Looks like this will be a lot of effort to get perfect. (Maybe time to remove the glow backend since the two backends add a lot of complexity?).
> 
> I commented a ton of code to try to get something to compile, so this will need a _lot_ more work. Opening this mostly in case someone wants to pick up the work or discuss how we should do things.

If I may ask. What exactly is missing/not working? I need this for an app I am trying to develop that depends on Stylus support, so I might just do the work myself for updating egui to the newest version. Issue just becomes: How will other PRs look like?

Is a PR freeze being done?

Eclextic:
Hm actually, looking at your PR now, it seems like the majority of problems lie basically down to the fact, that the winit beta just hasn't been implemented yet anywhere in the dependency chain and this looks to actually have to take multiple months before everyone else has caught up...

I'll switch to SDL for now. I don't think winit is the right solution xd

lucasmerlin:
> Hm actually, looking at your PR now, it seems like the majority of problems lie basically down to the fact, that the winit beta just hasn't been implemented yet anywhere in the dependency chain and this looks to actually have to take multiple months before everyone else has caught up...
> 
> I'll switch to SDL for now. I don't think winit is the right solution xd

Yeah, we need a proper winit release so that the ecosystem can actually update

Issue #7738, opened 2025-11-23
Cannot load an image with mime type "image/jpeg; charset=utf-8"


leungkkf:
**Describe the bug**
I am having an issue loading this image "[https://stacks.stanford.edu/image/iiif/wy534zh7137/SULAIR_rosette/full/400,/0/default.jpg](https://stacks.stanford.edu/image/iiif/wy534zh7137/SULAIR_rosette/full/400,/0/default.jpg)". It has a mime type of "image/jpeg; charset=utf-8", which is not common. 

**To Reproduce**

ui.image("https://stacks.stanford.edu/image/iiif/wy534zh7137/SULAIR_rosette/full/400,/0/default.jpg");

**Expected behavior**

The image should be correctly displayed.

**Desktop (please complete the following information):**
 - OS:  Windows 11 

**Additional context**

I tried to patch the is_supported_mime in image_loader.rs by extracting the mime type before the ImageFormat::from_mime_type...
    let mime_type = mime
        .split(';')
        .next()
        .map(|v| v.to_owned())
        .unwrap_or_default();

leungkkf:
I have created a draft PR [https://github.com/emilk/egui/pull/7739](https://github.com/emilk/egui/pull/7739)
Hope it helps...

PR #7739, opened 2025-11-25
Fix media type with optional parameters

leungkkf:
I am having an issue loading this image "https://stacks.stanford.edu/image/iiif/wy534zh7137/SULAIR_rosette/full/400,/0/default.jpg". It has a mime type of "image/jpeg; charset=utf-8", which is not common.

The code doesn't parse the full string that includes the optional parameter "charset=utf-8" to create the jpeg image format. 

A line is added to take only the mime type in the media type before the ";" and ignore the optional parameters. 

* Closes <https://github.com/emilk/egui/issues/7738>


leungkkf:
Thank you!

Issue #7647, opened 2025-10-17
Collect snapshot results in kittest harness


aedm:
Kittest harness should support continuing the test even if one or more snapshots fail. In this case, snapshot errors should be reported when the harness is dropped. 

This change allows simpler tests by not having to create a `SnapshotResults` manually.

PR #7744, opened 2025-11-26
Enforce consistent snapshot updates

lucasmerlin:
* Closes https://github.com/emilk/egui/issues/7647

This collects SnapshotResults within the Harness and adds a check to enforce snapshot results are merged in case multiple Harnesses are constructed within a test.

This should make snapshot updates via kitdiff/accept_snapshots.sh way more useful since it should now always update all snapshots instead of only the first one per test.

### DIFF
--- a/crates/egui_demo_lib/src/demo/tests/tessellation_test.rs
+++ b/crates/egui_demo_lib/src/demo/tests/tessellation_test.rs
@@ -363,7 +363,7 @@ mod tests {
 
     #[test]
     fn snapshot_tessellation_test() {
-        let mut snapshot_results = SnapshotResults::new();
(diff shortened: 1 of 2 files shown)


crates/egui_demo_lib/tests/image_blending.rs
```diff
@@ -21,5 +22,6 @@ fn test_image_blending() {
         harness.run();
         harness.fit_contents();
         harness.snapshot(format!("image_blending/image_x{pixels_per_point}"));
+        results.extend_harness(&mut harness);
```
emilk, line 25: Why `results.extend(harness.take_snapshot_results());` in some places and `results.extend_harness(&mut harness);` in others?

PR #7750, opened 2025-12-01
Typo fix in drag-and-drop documentation

BeaconBrigade:

Adding periods to the end of sentences and fixes a grammar mistake on documentation for the drag-and-drop code to become consistent with the rest of the documentation.

The documentation for `Ui::dnd_drop_zone` could've used the word "its" instead of "the" to replace "it", but I think "the" more clearly refers to the `Frame` since "its" has been used to refer to the drop-zone already.


lucasmerlin:
Thanks!

PR #7754, opened 2025-12-05
Enable `or_fun_call` lint to avoid unnecessary allocations

grtlr:
### What

From the [lint description](https://rust-lang.github.io/rust-clippy/master/index.html?search=or_fu#or_fun_call):

> The function will always be called. This is only bad if it allocates or does some non-trivial amount of work.

But also:

> If the function has side-effects, not calling it will change the semantic of the program, but you shouldn’t rely on that.
> 
> The lint also cannot figure out whether the function you call is actually expensive to call or not.

Still worth it to keep our happy paths clean, imo.

### DIFF
--- a/crates/egui/src/containers/popup.rs
+++ b/crates/egui/src/containers/popup.rs
@@ -465,7 +465,7 @@ impl<'a> Popup<'a> {
     pub fn get_best_align(&self) -> RectAlign {
         let expected_popup_size = self
             .get_expected_size()
-            .unwrap_or(vec2(self.width.unwrap_or(0.0), 0.0));
+            .unwrap_or_else(|| vec2(self.width.unwrap_or(0.0), 0.0));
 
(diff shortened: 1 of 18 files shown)


crates/egui/src/containers/popup.rs
```diff
@@ -465,14 +465,15 @@ impl<'a> Popup<'a> {
...
-            .unwrap_or(vec2(self.width.unwrap_or(0.0), 0.0));
+            .unwrap_or_else(|| vec2(self.width.unwrap_or(0.0), 0.0));
 
         let Some(anchor_rect) = self.anchor.rect(self.id, &self.ctx) else {
             return self.rect_align;
         };
 
         RectAlign::find_best_align(
             #[expect(clippy::iter_on_empty_collections)]
+            #[expect(clippy::or_fun_call)]
```
emilk, line 476: Why not change this case too?

grtlr: Ownership problems

PR #7758, opened 2025-12-06
Disable the Skrifa traversal feature

nicoburns:
- Followup to https://github.com/emilk/egui/pull/7694
- Disables the `traversal` feature of `skrifa` which is not needed except internally by the fontations project
- Should save a little compile time, and possibly some binary size.


emilk (approved):
Nice - thank you!

PR #4211, opened 2024-03-22
Fix `ViewportCommand::InnerSize` not resizing viewport on Wayland (#4211)

rustbasic:
Issues :
* Closes #3668 
* Related #4078
* Related #4079
* Related #4135 
* Closes #4196 
* Closes #4259 

Pull requests :
* Related #4088
* Closes #4207

PR #7760, opened 2025-12-08
egui-winit: Receive window scale factor and size via events

MarijnS95:
We found in our application (with custom `egui` renderer) that the `screen_rect` in `RawInput` was mismatching with the `winit` events that we processed earlier (in particular on resize on MacOS, which raises a lot of small size changes).

One of the contributing factors was `egui-winit` reading the `Window` size and scale factor over and over again separate from processed events, causing a mismatch on some platforms because this value would already change ahead of consuming a `Resized` event.  Secondly, some existing event processing may rely on this value being consistent with the event stream.

There is currently no guarantee -at least I don't see it in the documentation- that `winit` ensures that `inner_size()` will always return a consistent value until the user has consumed a `Resized` event. This gets worse when certain functions (like `update_viewport_info()`) are possibly called on a separate thread because both `egui::Context` and `winit::Window` are thread-safe.

Those also have an extra draw-back of roundtripping to the event loop on the main thread to query these values, as a `Window` is not actually thread-safe on all platforms.

---

This is still opened as a draft because it has not been tested on all platforms, and there are far more odd calls to `window.inner_size()`, `window.outer_size()` and `window.scale_factor()` than anticipated which might all suffer from the same behaviour; not all immediately have access to the latest state in `RawInput` though.

Secondly it was surprising to see the viewport code that updates `native_pixels_per_point` in a very disconnected way from `RawInput::pixels_per_point`, which I'd like to understand better before making any more changes here.  It seems like this was already all carefully crafted (ie. https://github.com/emilk/egui/pull/4211) and battle-tested to circumnavigate the inconsistency issues above... 😕

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -39,9 +39,9 @@ disallowed-methods = [
   # but we cannot disable them all here (because of e.g. https://github.com/rust-lang/rust-clippy/issues/10406)
   # so we do that in `clipppy_wasm.toml` instead.
 
-  { path = "std::env::temp_dir", readon = "Use the tempfile crate instead" },
+  { path = "std::env::temp_dir", reason = "Use the tempfile crate instead" },
   { path = "std::panic::catch_unwind", reason = "We compile with `panic = abort" },
-  { path = "std::thread::spawn", readon = "Use `std::thread::Builder` and name the thread" },
+  { path = "std::thread::spawn", reason = "Use `std::thread::Builder` and name the thread" },
 ]
 
 # https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_names


crates/egui-winit/src/lib.rs
```diff
@@ -973,6 +964,17 @@ impl State {
...
     }
+
+    /// Calculate the `pixels_per_point` for a given window, given the current egui zoom factor
+    fn pixels_per_point(&self) -> f32 {
+        self.egui_input
+            .viewports
+            .get(&self.viewport_id)
+            .expect("No viewport")
+            .native_pixels_per_point
+            .expect("ScaleFactorChanged must have been raised")
```
MarijnS95: On the `pure_glow` example on Wayland we seem to receive `Resized` -> `RedrawRequested` -> `ScaleFactorChanged`.

This'll panic because they're out of order, and we might need to make sure to update the input size again also when the scale factor changes.

PR #7764, opened 2025-12-11
feat(easy_mark): Add multi-level headings (H1-H3)

senchpimy:
In a different branch


oscargus:
What about naming them `SubHeading` and `SubSubHeading`? (`sub_heading`, `sub_sub_heading`) Given that the original is not called `Heading1`, it seems like one can just go with another naming scheme and that it probably will make more sense in many situations.

And maybe a bit of documentation what they are. 

(But I would definitely use this in my app instead of the manual setting of fonts that I currently rely on.)

PR #7780, opened 2025-12-15
plugin: export `TypedPluginGuard` and `TypedPluginHandle`

apekros:
We have made this patch internally on our fork as we use `TypedPluginGuard` and `TypedPluginHandle` internally

### DIFF
--- a/crates/egui/src/lib.rs
+++ b/crates/egui/src/lib.rs
@@ -413,7 +413,7 @@ pub mod os;
 mod painter;
 mod pass_state;
 pub(crate) mod placer;
-mod plugin;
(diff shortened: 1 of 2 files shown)


apekros:
> Let's not put undocumented things into the top egui namespace.
> 
> Please add docstrings to these types, but then do `pub mod plugin` instead of re-exporting the types directly, to reduce the number of things the user can auto-complete to with `egui::`

Thanks for the prompt feedback, done! I have left the existing `pub use for plugin::Plugin` as I didn't think this was worth the breaking change

lucasmerlin (approved):
Thanks!

PR #7789, opened 2025-12-18
Update egui-wgpu to use wgpu version 28.0.0

JasperBRiedel:
This is a small patch to update the `wgpu` dependency from version **27.0.1** to **28.0.0**. The changes are minor, requiring only a few updates to descriptor initializers for the updated multiview and immediates (formally push constants) fields.

The changes are working without any obvious issues in my own project. However, I've not yet been able to run the check script due to the extra dependencies required not being available in my development environment. I will attempt this again later.

Please note that `wgpu` now requires a minimum toolchain version of **1.92.0**, so I have updated the `rust-toolchain` file accordingly. I'm not sure if the toolchain version change has larger implications for the project?

This is my first pull request. I'm hoping it meets the criteria for a small change, and therefore doesn't need an issue first.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1388,7 +1388,7 @@ name = "egui_demo_app"
 version = "0.33.3"
 dependencies = [
  "accesskit",
- "accesskit_consumer 0.30.1",
+ "accesskit_consumer 0.31.0",
  "bytemuck",
  "chrono",
  "eframe",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,7 @@ egui_kittest = { version = "0.33.3", path = "crates/egui_kittest", default-featu
 eframe = { version = "0.33.3", path = "crates/eframe", default-features = false }
 
 accesskit = "0.21.1"
-accesskit_consumer = "0.30.1"
+accesskit_consumer = "0.31.0"
 accesskit_winit = "0.29.2"
 ahash = { version = "0.8.12", default-features = false, features = [
   "no-rng", # we don't need DOS-protection, so we let users opt-in to it instead


rust-toolchain
```diff
@@ -5,6 +5,6 @@
 # to the user in the error, instead of "error: invalid channel name '[toolchain]'".
 
 [toolchain]
-channel = "1.88.0"
+channel = "1.92.0"
```
torokati44, line 8: Usually a proper and comprehensive MSRV bump is done first, see: https://github.com/emilk/egui/pull/7579

JasperBRiedel:
The windows build is failing due to two different versions of the `windows` crate being used by dependencies (I think?).

This can be resolved by letting cargo regenerate `Cargo.lock` using `cargo generate-lockfile`.

I've not had an issue like this before, and I'm unsure of the broader implications of regenerating `Cargo.lock`.

Is this a reasonable fix?

JasperBRiedel:
Ah, it seems I've discovered the broader implications... We now fail the checks of the `deny.toml` 🙃 

I'm not sure how to move forward with this now. Any input would be greatly appreciated.

MeFisto94:
Without being an expert and especially knowing the context: Editing a lockfile is usually never the right choice. Not in Rust, not in nodeJS. If it were, there'd be no need to version the lockfile or have it after all. The lockfile pins all dependencies, especially transitives to the exact version (~ commit), to have "reproducable" builds. If those versions would change, you could have slight changes in behavior that are very hard to debug because they are unexpected.

Instead, one should always edit the Cargo.toml and have the resolver update the lock as a consequence, because after all you updating the WGPU version made it think that there are two versions of the windows crate necessary. If there weren't conflicting dependency versions, it would've tried to settle on one.

That's also the case at least for https://github.com/emilk/egui/actions/runs/20405854417/job/58635443776?pr=7789#logs partially, deny found out that we're having two versions of the png crate as well somewhere in the hierarchy (and conveniently prints us the hierarchy). Additionally, deny found some dependencies with unsupported licenses, this could be the direct result of updating the lock file: Dependencies changed their dependencies and now include dependencies with unknown licenses or changed their license.

Alright, if we revert back to 209ec7a72d596b422153d47ac73493d8fda84651. If you use `cargo tree -i windows`, it tells you:
```
error: There are multiple `windows` packages in your project, and the specification `windows` is ambiguous.
Please re-run this command with one of the following specifications:
  windows@0.61.1
  windows@0.62.2
  ```
  
  For 0.61.1:
  ```
  windows v0.61.1
├── accesskit_windows v0.29.1
│   └── accesskit_winit v0.29.1
│       └── egui-winit v0.33.3
...
└── gpu-allocator v0.28.0
    └── wgpu-hal v28.0.0
        ├── wgpu v28.0.0
        │   ├── eframe v0.33.3
```

And for 0.62.2:
```
windows v0.62.2
└── wgpu-hal v28.0.0
    ├── wgpu v28.0.0
    │   ├── eframe v0.33.3
```

So, my guess is that the culprit is accesskit_windows/accesskit_winit needing a bump to a version where they used 0.62.2 as well. It shouldn't be a problem with gpu-allocator v0.28.0, because clearly wgpu-hal's dependency on windows would be faulty, then. I guess it just chose the lower one by default.

And then, you'll be looking at png, where we saw the problem as well in the deny reports.

smlfrysamuri:
For what it's worth, I am using this pull request currently to allow myself to use WGPU v28, and I have no issues, including a resolution to the double windows dependency issue I was having when I first upgraded to WGPU 28. I am building on windows 11 25H2 26220.7523 in release mode using the nightly toolchain.

Here is my cargo.toml (the pertinent parts) - don't judge if something is weird, I am not well-versed in these config files yet and I am positive I am doing some goofy things:

```
[workspace.dependencies]
# 3rd Party
bytemuck = { version = "1.23.2", features = ["derive"] }
egui = { git = "https://github.com/emilk/egui.git", rev = "4a9fc7e6417b9ca838b880f2a89fd16468aece67" }
egui-wgpu = { git = "https://github.com/emilk/egui.git", rev = "4a9fc7e6417b9ca838b880f2a89fd16468aece67", features = ["winit"] }
egui-winit = { git = "https://github.com/emilk/egui.git", rev = "4a9fc7e6417b9ca838b880f2a89fd16468aece67" }
image = { version = "0.25", default-features = false, features = ["png", "jpeg"] }
parking_lot = { version = "0.12.4", features = ["send_guard"] }
serde = { version = "1.0", features = ["derive"] }
wgpu = { version = "^28.0", default-features=false, features=["wgsl", "std", "parking_lot", "dx12"] }

anyhow = "1.0"
bincode = "1.3"
bitflags = "2.10"
dashmap = "6.1"
env_logger = "0.11.8"
log = "0.4.28"
lz4 = "1.28"
noise = "0.9"
paste = "1.0"
rand = "^0.10.0-rc.5"
rayon = "1.11.0"
serde_json = "1.0"
smol = "2.0.2"
winit = "0.30"
```
Not sure if this will help or is relevant, but figured maybe I'd share just in case.

JasperBRiedel:
The multiple `windows` packages issue can be resolved by updating `accesskit_winit`, but that requires updating `accesskit_consumer`, which then requires the `accesskit_consumer` dependency in [kittest](https://github.com/rerun-io/kittest).to be updated to 0.31.0 as well.

I'm not sure what the best way to update kittest is due to the circular dependency situation https://github.com/rerun-io/kittest/issues/8. 

I've tested with a local patch that updates `kittest` to `accesskit_consumer` **0.31.0**, and it does resolve the issue below:

```
error[duplicate]: found 2 duplicate entries for crate 'accesskit_consumer'
  ┌─ /github/workspace/Cargo.lock:2:1
  │  
2 │ ╭ accesskit_consumer 0.30.1 registry+https://github.com/rust-lang/crates.io-index
3 │ │ accesskit_consumer 0.31.0 registry+https://github.com/rust-lang/crates.io-index
  │ ╰───────────────────────────────────────────────────────────────────────────────┘ lock entries
  │  
  ├ accesskit_consumer v0.30.1
    └── kittest v0.3.0
        └── egui_kittest v0.33.3
            ├── (dev) egui_demo_app v0.33.3
            ├── (dev) egui_demo_lib v0.33.3
            │   ├── custom_style v0.1.0
            │   ├── egui_demo_app v0.33.3 (*)
            │   └── hello_android v0.1.0
            └── (dev) egui_tests v0.33.3
  ├ accesskit_consumer v0.31.0
    ├── accesskit_windows v0.29.2
...
    └── egui_demo_app v0.33.3 (*)
```

MeFisto94:
Nice! Well, the issue that you linked says that it's okay to break the example for until egui has updated as well, so we'd need a bump of kittest to the most recent accesskit_consumer first, or a local patch which is probably not desireable.

Side note: That way we are only stuck on cargo deny not liking that duplicate dependency, so one can already use your branch to build the more recent egui/wgpu integration!

The windows crate being so widespread makes this lint so infectious.

torokati44:
The alternative #7853 got merged.

Issue #3218, opened 2023-08-09
add Fontweight to fontid


Omicronlawful:
It would be usefull if the font weight could be set when creating a fontid. Im not sure if there is another way, like it is solved for italics, but i would put it into fontid

abey79:
In particular, support for variable font weight would be nice.

emilk:
First of all: italics in epaint is a hack. We just skew the glyphs when we paint them. Ideally we should instead switch to an italic font of the same family instead.

To fully support font weights and italics we need to have better support for selecting font _families_. For example: [Inter](https://rsms.me/inter/#weights) is a font family with many different fonts for different weights and styles (Thin, Thin Italics, Bold, Bold Italics, etc).

A font family can be stored in several .ttf files, or in the same one. To know what we have loaded, we would need to read it at some sort of font meta-data (was the loaded font thin, italics, …?). I don't think `ab_glyph` supports this at the moment: https://docs.rs/ab_glyph/latest/ab_glyph/trait.Font.html

Once that all is done, then yes, `FontId` would be the right place to add weight and italics.

abey79:
I believe there are actually two stories here:
- Support for font families (which can already be manually done)
- Support for font with variable weight(/etc) support: https://fonts.google.com/knowledge/introducing_type/introducing_variable_fonts

The latter would be very nice to have as I understand designer rely on it more and more. As it turns out, the font we're using for Rerun _is_ variable (https://fonts.google.com/specimen/Inter).

JustFrederik:
@emilk not sure if this is helpful, but ttf parser has an example for handling metadata: https://github.com/RazrFalcon/ttf-parser/blob/master/examples/font-info.rs

NfNitLoop:
I'd also love bold support.

FWIW, ab_glyph added support for variable fonts here: https://github.com/alexheretic/ab-glyph/pull/67

It's accessed via the [VariableFont](https://docs.rs/ab_glyph/latest/ab_glyph/trait.VariableFont.html) trait, which is enabled by default.

That unfortunately doesn't give you access to any of the font's "named instance" settings (ex: "Regular 400" from [Inter](https://fonts.google.com/specimen/Inter)), but it would at least allow folks to adjust the `wght` variable manually. 

This should also give access to "Italic" and "Slant" variables:
https://variablefonts.io/about-variable-fonts/

lucasmerlin:
In my project I have this for bold text support:

```rs
[...]

PR #7790, opened 2025-12-18
Apply preferred font weight when loading variable fonts

pmnxis:
Previously, when loading a variable font (e.g. via `egui::FontData::from_static`),
the font was rendered using the default (often the lightest) weight,
ignoring any preferred weight configuration.

This change applies the specified weight to skrifa's `Location` for the `wght` axis,
ensuring that variable fonts are rendered with the intended font weight.

## Summary
Fixes variable font weight not being applied during rendering. The `FontData::weight()` method now properly configures the font variation axis.

## Changes
- Add `location: Location` field to `FontFace` to store variation coordinates
- Pass `location` parameter through to glyph rendering functions
- Apply weight to skrifa's `LocationRef` in `DrawSettings` and `HintingInstance`

## Weight Priority
1. `preferred_weight` from `FontData::weight()`
2. OS/2 table's `us_weight_class` 
3. Variable font's fvar default value
4. `Location::default()`

## Related Issue
- #3218 : Not follow font id, but goal would be same

## Todo
* [x] Apply preferred font weight when loading variable fonts
* [ ] Add small size variable fonts for docs and egui (need discussion)

### DIFF
--- a/crates/epaint/src/text/font.rs
+++ b/crates/epaint/src/text/font.rs
@@ -1,13 +1,12 @@
 #![allow(clippy::mem_forget)]
 
-use std::collections::BTreeMap;
-
 use emath::{GuiRounding as _, OrderedFloat, Vec2, vec2};
 use self_cell::self_cell;
 use skrifa::{
     MetadataProvider as _,
     raw::{TableProvider as _, tables::kern::SubtableKind},
 };
+use std::collections::BTreeMap;
 use vello_cpu::{color, kurbo};
 
 use crate::{
@@ -201,6 +200,7 @@ impl FontCell {
         metrics: &ScaledMetrics,
         glyph_info: &GlyphInfo,
         bin: SubpixelBin,
+        location: &skrifa::instance::Location,
     ) -> Option<GlyphAllocation> {
         let glyph_id = glyph_info.id?;
 
@@ -225,7 +225,7 @@ impl FontCell {
                         .reconfigure(
                             &font_data.outline_glyphs,
                             size,
-                            skrifa::instance::LocationRef::default(),
+                            location,
                             skrifa::outline::Target::Smooth {
                                 mode: skrifa::outline::SmoothMode::Normal,
                                 symmetric_rendering: true,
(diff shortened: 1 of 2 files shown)


emilk:
Screenshot 2025-12-19 at 18 44 11

Do you want to do this before I hit merge?

pmnxis:
@emilk 
Checkpoint “Add small size variable fonts for docs and egui (need discussion)”,

https://github.com/emilk/egui/pull/7790/files#diff-476cc9a474dfda61ceeb981bd3ea229a1ab4eb65330e74c0128ce257236938c9R177-R187
In this PR , the new docs code uses UbuntuLight instead of some variable font.

Ideally, it would be better to include a variable font directly in the source tree and add
some verification code to ensure that variable fonts actually work as expected.
I have already validated this approach in my own project:
https://github.com/pmnxis/chama-optics

That said, adding variable fonts would significantly increase the source tree size,
so keeping the existing example as-is also seems like a reasonable choice.

valadaptive:
I'm a bit confused by this PR. When will a variable font's default weight be the "lightest" one? By definition, the default value for *any* axis should be mapped to 0.0 in variation space, which is what Skrifa uses if you don't specify any coordinates.

This PR changes the font loading process to use the weight value from the OS/2 table if present, but falls back to the default `wght` axis value from the fvar table if not, which is...exactly what we were doing before.

Can you give examples of variable fonts that default to the lightest weight? Are they fixed if you use the weight from the OS/2 table, or do you still need to manually override the weight when loading them?

pmnxis:
@valadaptive 

> Can you give examples of variable fonts that default to the lightest weight?
The Noto CJK variable fonts are an example:  
https://github.com/notofonts/noto-cjk/tree/main/google-fonts

After re-examining the OS/2 table, I confirmed that these fonts have  
`<usWeightClass value="100"/>`, which explains why they default to the lightest weight. And I mistaked I didn't checked well it.

That said, there are still cases where **manually overriding the font weight is necessary**.  
In those situations, the function I mentioned in your PR, `pub fn weight(self, weight: u16) -> Self`  is still allowed by your changes, and I’ve confirmed that it correctly covers this use case.

```rust
const FONT_NTSANS: BuiltInFonts = BuiltInFonts {
    name: "Noto Sans",
    data: include_bytes!("../../assets/fonts/NotoSansJP[wght].ttf"),
};

// Demonstrates how to replace all fonts.
pub(crate) fn replace_fonts(ctx: &egui::Context) {

    fonts.font_data.insert(
        FONT_NTSANS.name.to_owned(),
        std::sync::Arc::new(egui::FontData::from_static(FONT_NTSANS.data).weight(400)),
    );
    
    ctx.set_fonts(fonts);
}
```

PR #7793, opened 2025-12-19
Update MSRV from 1.88 to 1.92

JasperBRiedel:
Progress on updating MSRV to 1.92 so that we can update to wgpu version 28.0.0 in #7789 

I'm still working on getting all checks passing. Remaining items that I'm aware of: unresolved link errors in the doc comments for egui, unsafe block missing a safety comment in eframe.

### DIFF
--- a/crates/egui_demo_lib/benches/benchmark.rs
+++ b/crates/egui_demo_lib/benches/benchmark.rs
@@ -198,6 +198,7 @@ pub fn criterion_benchmark(c: &mut Criterion) {
             let mut string = String::new();
             for _ in 0..NUM_LINES {
                 for i in 0..30_u8 {
+                    #[expect(clippy::unwrap_used)]
                     write!(string, "{i:02X} ").unwrap();
                 }
                 string.push('\n');


JasperBRiedel:
I'm currently stuck on the following rustdoc error:

```
error: unresolved link to `UiStackInfo::new(UiKind::GenericArea)`
   --> crates/egui/src/containers/area.rs:175:22
    |
175 |     /// Default to [`UiStackInfo::new(UiKind::GenericArea)`].
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the struct `UiStackInfo` has no field or associated item named `new(UiKind`
    |
    = note: `-D rustdoc::broken-intra-doc-links` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(rustdoc::broken_intra_doc_links)]`
```

It seems like specifying an argument like this isn't valid?

A possible solution could be something like ```/// Default to [`UiStackInfo`] with kind [`UiKind::GenericArea`].```?

emilk:
Sounds like a reasonable fix to me!

Btw, I'm a bit surprised to see wgpu having MSRV set to a version that is only a week old. It's a bit annoying to force every downstream crate use use the very latest Rust version, but 🤷

.github/workflows/cargo_machete.yml
```diff
@@ -3,17 +3,17 @@ name: Cargo Machete
...
-        with:
-          toolchain: 1.88
-      - name: Machete install
-        ## The official cargo-machete action
-        uses: bnjbvr/cargo-machete@v0.9.1
-      - name: Checkout
-        uses: actions/checkout@v4
-      - name: Machete Check
-        run: cargo machete
+    cargo-machete:
```
emilk: please revert unrelated formatting changes 🙏

JasperBRiedel: Oh, that was unintentional. I must have run the yaml formatter by accident, and didn't check the diff again 🤦

PR #7802, opened 2025-12-22
Show menu without a click if another menu is open

### LINKED ISSUE #7346: Menus in `MenuBar` don't open anymore without a click
Hello! I just updated to `0.32` and refactored various things as needed, but I have a new issue that I can't find a fix for.

In the past, the currently open menu followed hovering over the menubar (the usual behavior on must GUIs), now I need to click another menu button for it to be revealed.

Please see the following video:

https://github.com/user-attachments/assets/3632a581-39ff-4d2e-b553-d4759b32a1dc

I don't think this is by design, but if it is, is there an API to revert to the previous behavior? Thanks...

mkeeter:
https://github.com/emilk/egui/issues/4607 accidentally removed the behavior where – once a menu is open – moving the mouse to another menu opens that one instead (on hover, without requiring a click).  This is a pretty easy fix!

* Closes <https://github.com/emilk/egui/issues/7346>

### DIFF
--- a/crates/egui/src/containers/menu.rs
+++ b/crates/egui/src/containers/menu.rs
@@ -11,7 +11,8 @@
 use crate::style::StyleModifier;
 use crate::{
     Button, Color32, Context, Frame, Id, InnerResponse, IntoAtoms, Layout, Popup,
-    PopupCloseBehavior, Response, Style, Ui, UiBuilder, UiKind, UiStack, UiStackInfo, Widget as _,
+    PopupCloseBehavior, Response, SetOpenCommand, Style, Ui, UiBuilder, UiKind, UiStack,
+    UiStackInfo, Widget as _,
 };
 use emath::{Align, RectAlign, Vec2, vec2};
 use epaint::Stroke;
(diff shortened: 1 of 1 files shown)


crates/egui/src/containers/menu.rs
```diff
@@ -325,7 +326,15 @@ impl<'a> MenuButton<'a> {
         let response = self.button.ui(ui);
         let mut config = self.config.unwrap_or_else(|| MenuConfig::find(ui));
         config.bar = false;
-        let inner = Popup::menu(&response)
+
+        let mut menu = Popup::menu(&response);
+        if Popup::is_any_open(ui.ctx()) && response.hovered() {
```
emilk, line 331: Ideally this should check that the other open menu is of the same group. Maybe @lucasmerlin can help with that?

Currently we get this weird behavior:

Image

mkeeter: Good catch – I'd be happy to add this if I can get some pointers, but I'm not sure about the concept of menu groups (the word "group" doesn't appear at all in `containers/menu.rs`)

lucasmerlin: Hmm. You'd basically need to check if you're in the same MenuBar. Slightly ugly, but we could look up if we are within a MenuBar via `UiStack` by adding a `UiKind::MenuBar` and then store the MenuBar id in `OpenPopup` within memory.
Then, in the MenuButton get the bar id of the currently open popup, if it matches our bar id, open the popup on hover.

Not sure if there is a simpler way but this is the best thing I can come up with right now.

j-te:
Great work. Any idea when this will be in please without having to patch?

https://github.com/twvd/snow/pull/251#issuecomment-4103953334

PR #7805, opened 2025-12-22
Allow moving existing widgets to the top of interaction stack

emilk:
This adds `Ui::interact_opt` which is a version of `Ui::interact` that lets you specify that if this widget already exists, it should be moved to the top of the interaction stack.

This lets you easily more easily "read a response from the future", by simply calling interact twice, instead of calling `Context::read_response`.

### DIFF
--- a/crates/egui/src/widget_rect.rs
+++ b/crates/egui/src/widget_rect.rs
@@ -63,7 +63,7 @@ impl WidgetRect {
     }
 }
 
-/// How to handle multiple calls to [`Response::interact`].
+/// How to handle multiple calls to [`crate::Response::interact`] and [`crate::Ui::interact_opt`].
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


crates/egui/src/widget_rect.rs
```diff
@@ -176,7 +193,13 @@ impl WidgetRects {
                 existing.enabled |= widget_rect.enabled;
 
                 if existing.layer_id == widget_rect.layer_id {
-                    layer_widgets[*idx_in_layer] = *existing;
+                    if move_to_top {
+                        layer_widgets.remove(*idx_in_layer);
+                        *idx_in_layer = layer_widgets.len();
+                        layer_widgets.push(*existing);
```
lucasmerlin, line 199: `layer_widgets.remove` changes the index of the following widgets in this layer, meaning they now have the wrong index in the by_id map 😕

This should be the cause of https://github.com/emilk/egui/issues/7812

PR #7808, opened 2025-12-22
Don't expand widgets on hover

emilk:

This changes the default style of egui.

The problem with "expanding" widgets is that they now want to paint outside their own bounds, which then requires all parent UIs to have proper margins.

It also means hovered things are no longer properly aligned with every other widget.

### DIFF
--- a/crates/egui/src/style.rs
+++ b/crates/egui/src/style.rs
@@ -1234,6 +1234,11 @@ pub struct WidgetVisuals {
     pub fg_stroke: Stroke,
 
     /// Make the frame this much larger.
+    ///
+    /// The problem with "expanding" widgets is that they now want to paint outside their own bounds,
+    /// which then requires all parent UIs to have proper margins.
+    ///
+    /// It also means hovered things are no longer properly aligned with every other widget.
     pub expansion: f32,
 }
 


crates/egui/src/style.rs
```diff
@@ -1568,15 +1573,15 @@ impl Widgets {
...
-                expansion: 1.0,
+                expansion: 0.0,
             },
             active: WidgetVisuals {
                 weak_bg_fill: Color32::from_gray(55),
                 bg_fill: Color32::from_gray(55),
                 bg_stroke: Stroke::new(1.0, Color32::WHITE),
                 fg_stroke: Stroke::new(2.0, Color32::WHITE),
                 corner_radius: CornerRadius::same(2),
-                expansion: 1.0,
```
Killerff-blip: - [ ] @aedm check ✅

crates/egui/src/style.rs
```diff
@@ -1568,15 +1573,15 @@ impl Widgets {
                 bg_stroke: Stroke::new(1.0, Color32::from_gray(150)), // e.g. hover over window edge or button
                 fg_stroke: Stroke::new(1.5, Color32::from_gray(240)),
                 corner_radius: CornerRadius::same(3),
-                expansion: 1.0,
+                expansion: 0.0,
             },
             active: WidgetVisuals {
                 weak_bg_fill: Color32::from_gray(55),
                 bg_fill: Color32::from_gray(55),
                 bg_stroke: Stroke::new(1.0, Color32::WHITE),
```
Killerff-blip: @kornelski check the colour section for the best production 👋👋

PR #7811, opened 2025-12-26
Remove `font_weight` API

valadaptive:
* Closes N/A

This appears to have snuck in as part of https://github.com/emilk/egui/pull/7790, which claimed to only be a bugfix but introduced a new `font_weight` method.

I believe there's no way to access the method from *public* code since it's only defined on `FontsImpl`, not the public-facing `FontsView`. It's also not used *privately* in epaint, meaning it's completely dead code.

Even if we *do* want some sort of future API for getting a font's weight, it requires more consideration. For instance, this API will return the default weight for variable fonts, which is not documented anywhere and might not be what we want.

### DIFF
--- a/crates/epaint/src/text/font.rs
+++ b/crates/epaint/src/text/font.rs
@@ -337,9 +337,6 @@ pub struct FontFace {
     font: FontCell,
     tweak: FontTweak,
 
-    /// The font weight (100-900) if available from the font file.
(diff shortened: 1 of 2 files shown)


pmnxis:
Agree this PR
https://github.com/emilk/egui/pull/7790#issuecomment-3705290149

Issue #3508, opened 2023-10-26
Double click issue


bl4ze4447:
**Describe the bug**

A double_clicked() event is incorrectly triggered on a widget if the first click of the double click sequence occurred on a different UI element. It appears the logic for detecting double-clicks is tracking the time interval between any two clicks globally, rather than ensuring both clicks were targeted at the same widget.

**To Reproduce**
1. Click once anywhere in the application
2. Quickly click a button or widget that has response.double_clicked() logic.
3. Observe that double_clicked() returns true, even though the widget was only clicked once.

**Expected behavior**

double_clicked() should only return true if both clicks in the rapid sequence occur within the bounding box of the same widget. If the first click was elsewhere, the subsequent click should only register as a single clicked().

**Desktop (please complete the following information):**
 - OS: Windows 11, native app

Video evidence
https://github.com/emilk/egui/assets/109111928/11d128c7-5d41-4805-a449-8118fb7d8c76

jxqu3:
Same issue, any updates?

crumblingstatue:
I see two possible solutions:

- Somehow store the relevant information in `egui::Memory` to know if the same response was clicked twice
- Have a configurable distance threshold for double clicks in `InputOptions`, in addition to the delay.

PR #7817, opened 2025-12-28
Implemented distance threshold for double/triple clicks

bl4ze4447:
This change introduces a distance check to double and triple clicks. Previously, double/triple clicks were determined solely by timing, allowing clicks on different UI elements to trigger a double_clicked()/triple_clicked() event. 

By requiring consecutive clicks to occur within a specific radius (max_multiple_click_dist), we prevent double_clicked()/triple_clicked() events from triggering when a user clicks on two different, distant UI elements in rapid succession.

* Closes <https://github.com/emilk/egui/issues/3508>

### DIFF
--- a/crates/egui/src/input_state/mod.rs
+++ b/crates/egui/src/input_state/mod.rs
@@ -78,10 +78,6 @@ pub struct InputOptions {
     /// for double click (or when this value is doubled, triple click) to count.
     pub max_double_click_delay: f64,
 
-    /// If the pointer moves for more than this it will no longer register as a
-    /// double or a triple click.
-    pub max_multiple_click_dist: f32,
-
     /// When this modifier is down, all scroll events are treated as zoom events.
     ///
     /// The default is CTRL/CMD, and it is STRONGLY recommended to NOT change this.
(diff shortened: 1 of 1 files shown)


crates/egui/src/input_state/mod.rs
```diff
@@ -78,6 +78,10 @@ pub struct InputOptions {
     /// for double click (or when this value is doubled, triple click) to count.
     pub max_double_click_delay: f64,
 
+    /// If the pointer moves for more than this it will no longer register as a
+    /// double or a triple click.
+    pub max_multiple_click_dist: f32,
```
lucasmerlin: I think we can remove this and just use `max_click_dist`.

Or would it ever make sense to have this be something else than `max_click_dist`?

emilk: Can we just reuse `max_click_dist` instead? Why would we want these to be different?

belutoni: Good point! I have updated the code to reuse max_click_dist.

emilk (approved):
Thanks!

Issue #7660, opened 2025-10-22
segmentation fault when program is closed


ig1711:
**Describe the bug**

In wayland, using egui-winit with the `clipboard` feature causes segmentation fault when the program is closed. Program starts fine and runs fine but when closed, segmentation fault occurs. This is reproducible with the [pure_glow](https://github.com/emilk/egui/blob/main/crates/egui_glow/examples/pure_glow.rs) example  by running the example with the `clipboard` feature included.

**To Reproduce**
1. Run the `pure_glow` example with `clipboard` feature
```sh
cargo run -p egui_glow --example pure_glow --features=winit,egui/default_fonts,clipboard
```
2. After the program runs and window appears, close the program, i. e., click on Quit in the UI or close it with ctrl + c.
3. segmentation fault (core dumped) is shown

**Expected behavior**

Program should close without segmentation fault.

**Screenshots**

**Desktop (please complete the following information):**
 - OS:  Arch Linux with sway
 - Browser 
 - Version 

**Smartphone (please complete the following information):**
 - Device: 
 - OS: 
 - Browser 
 - Version 

**Additional context**

It probably has something to do with [smithay-clipboard](https://crates.io/crates/smithay-clipboard), because the debugger shows the following

<img width="1376" height="751" alt="Image" src="https://github.com/user-attachments/assets/f1f494a3-fef7-4463-bcf5-46dfcad935f8" />

EdoardoLuciani:
I can reproduce this on Fedora 43 as well, using a custom Vulkan renderer that integrates egui-winit.
The segmentation fault only occurs when the clipboard feature is enabled, disabling it makes the crash disappear.

I get the same stack trace as the original post, with the crash occurring inside smithay_clipboard during program shutdown.

davidhozic:
I can also reproduce a segmentation fault.  It only happens when using **egui-winit**.

SuchAFuriousDeath:
I think I uncovered the cause at #7743. however it has not gained much traction 😅

PR #7826, opened 2026-01-02
Make Clipboard::new unsafe

### LINKED ISSUE #7743: Safety hazard in Clipboard::new
**Describe the bug**
I believe that this method: https://github.com/emilk/egui/blob/8b8595b45b4c283a2a654ada081342079170e3ab/crates/egui-winit/src/clipboard.rs#L32 is not supposed to be safe.

It calls this method from smithay-clipboard (with the smithay-clipboard feature on): https://github.com/Smithay/smithay-clipboard/blob/26c2f53f15f6bdc4f41a442d0ae2c2d63bbc617c/src/lib.rs#L34, however, the safety requirements of that method cannot be guaranteed to be upheld by egui-winit. 
```rust
    /// # Safety
    ///
    /// `display` must be a valid `*mut wl_display` pointer, and it must remain
    /// valid for as long as `Clipboard` object is alive.
    pub unsafe fn new(display: *mut c_void) -> Self { ... }
```

**To Reproduce**
I believe that this is enough of an issue in theory, but here is a minimal demonstration of the safety hazard with 100% safe code: https://github.com/SuchAFuriousDeath/smithay-clipboard-segfault. The code in the repo segfaults when dropping the Clipboard.

**Expected behavior**
I believe the method should be unsafe and pass the safety requirement onto the caller just like the smithay-clipboard method does. 

**Additional context**
I believe this might be the cause of #7660, but I consider this distant enough to warrant a separate issue.

SuchAFuriousDeath:
The smithay-clipboard Clipboard::new method requires the display pointer to remain valid for the lifetime of the Clipboard. Since egui-winit's Clipboard::new cannot guarantee this invariant, it must be marked unsafe and pass the safety requirement to the caller.

Fixes #7743 (more details there)

* Closes <https://github.com/emilk/egui/issues/7743>

### DIFF
--- a/crates/egui-winit/src/clipboard.rs
+++ b/crates/egui-winit/src/clipboard.rs
@@ -28,8 +28,14 @@ pub struct Clipboard {
 }
 
 impl Clipboard {
-    /// Construct a new instance
-    pub fn new(_raw_display_handle: Option<RawDisplayHandle>) -> Self {
+    /// Construct a new instance.
+    ///
+    /// # Safety
+    ///
+    /// If `raw_display_handle` is `Some`, the display handle must remain valid for the
+    /// entire lifetime of the returned `Clipboard` instance.
+    #[expect(unsafe_code)]
+    pub unsafe fn new(_raw_display_handle: Option<RawDisplayHandle>) -> Self {
         Self {
             #[cfg(all(
                 not(any(target_os = "android", target_os = "ios")),
@@ -47,7 +53,8 @@ impl Clipboard {
                 ),
                 feature = "smithay-clipboard"
             ))]
-            smithay: init_smithay_clipboard(_raw_display_handle),
+            // SAFETY: The caller guarantees that the display handle remains valid.
+            smithay: unsafe { init_smithay_clipboard(_raw_display_handle) },
 
             clipboard: Default::default(),
         }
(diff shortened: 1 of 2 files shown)


oscargus:
Would it make sense to have two variants of `new` depending on the config? One which is unsafe and one which is not? 

Not sure how much of this is propagated up though or if it is just simpler to mark even the `arboard` variant unsafe? There is a risk that one will now have to deal with `unsafe` even if only the `arboard` version is used.

SuchAFuriousDeath:
I'm not sure that's such a good idea. Maybe we could keep new safe and only enable it when smithay clipboard is disabled and then have an unsafe from_raw_handle method or something like that.

SuchAFuriousDeath:
But it's breaking change either way.

crates/egui-winit/src/clipboard.rs
emilk: nit: `fn init_smithay_clipboard` should probably be marked `unsafe` as well?

emilk (approved):

emilk:
Thanks!

SuchAFuriousDeath:
Forgot to mention: This also probably closes https://github.com/emilk/egui/issues/7660.

crates/egui-winit/src/clipboard.rs
SuchAFuriousDeath: Good point.

SuchAFuriousDeath:
@emilk Actually there is a bit of a problem here. I think that if we really want to be 100% sound, then State::new should probably be unsafe as well, because again, the current function signature can't enforce that the safety requirements are actually upheld.
Per my analysis, the callers of State::new (inside egui) can actually uphold the safety requirements, so the unsafety would not bubble up further than that. But State::new is pub, so this would be one more breaking change.

PR #7831, opened 2026-01-04
Improve `TextEdit` and `LayoutJob` alignments

RndUsr123:
This is a fix/improvement that makes all kinds of alignments work in `TextEdit` when a custom `LayoutJob` with halign is used.

I used the simplest approach possible to avoid unwanted bugs as I wasn't sure what's safe to change, but there's potentially better ways to achieve this. In particular, I'm not sure I fully understand the rationale behind aligning rows in a `Galley` based on the latter's leftmost border, considering the size is what's ultimately used in widgets (in `TextEdit` at least).

Regardless, here's a demo of this PR:

https://github.com/user-attachments/assets/5d9801d7-73af-4576-80c5-47f169700462

### DIFF
--- a/tests/egui_tests/tests/regression_tests.rs
+++ b/tests/egui_tests/tests/regression_tests.rs
@@ -118,6 +118,7 @@ fn text_edit_halign() {
     harness.snapshot("text_edit_halign");
 }
 
+#[test]
 fn text_edit_delay() {
     let mut text = String::new();
     let mut harness = Harness::builder().with_size((200.0, 50.0)).build_ui(|ui| {


RndUsr123:
That could be tricky because there's no way to determine where the text has been rendered after the fact from what I can gather. The main issue stems from the fact `TextEdit` relies on the `Galley`'s rect size rather than its alignment; I'm a bit wary of changing how `Galley` is computed as that could introduce other bugs so I just visually shifted the text instead. 
Here's what I mean for reference (the boxes show the `Galley` rects):
comparison


I've also just found another bug in `LayoutJob` which prevents the cursor from moving forward after a trailing space when halign is `RIGHT` or `Center`, so that would probably need to be fixed before a proper test about cursor position can be written. Should I open another issue for it or should it be part of this PR?

Edit: this is what it looks like (it was a thing before my fix)

https://github.com/user-attachments/assets/7be6e4c7-fda6-4433-b9b3-f0e2cdd06632

emilk:
You don't need to test text editing (unless you want to). Just adding your nice 3x3 matrix and snapshot-testing that would be great. See for instance this:

https://github.com/emilk/egui/blob/f26d7d8072379a5ffa9edc53ca4fabe4b25089d8/tests/egui_tests/tests/regression_tests.rs#L38-L63

RndUsr123:
Oh I see... sure I'll do it!

Btw I think I've found out how to fix the `LayoutJob` issue: it was trimming them out even when the row didn't need to be justified so changing that apparently fixed the problem and justify still works:
fix2

This is what I'm going for:
```
let [original_min_x, original_max_x];

if justify {
    original_min_x = row.glyphs[glyph_range[0]].logical_rect().min.x;
    original_max_x = row.glyphs[glyph_range[1] - 1].logical_rect().max.x;
}
else {
    original_min_x = row.glyphs[0].logical_rect().min.x;
    original_max_x = row.glyphs[row.glyphs.len() - 1].logical_rect().max.x;
}

let original_width = original_max_x - original_min_x;
```
would you prefer a direct assignment or something else style-wise?

RndUsr123:
I've added the regression test! Ended up settling for a single snapshot, which should be good enough to showcase both visual alignment in a few scenarios as well as correct cursor placement.
The `LayoutJob` whitespace trim situation turned out to be not as trivial so I'll open a new issue for that.

emilk:
Thanks!

PR #7834, opened 2026-01-06
Make `FrameCache::get` return a reference instead of cloning the cached value

KonaeAkira:
This is a breaking change.

- Enables using `FrameCache` in cases where the cached value cannot be cloned.
- Improves use cases where only a reference to the cached value is needed.
- If the user needs an owned value, they can clone it themselves.

Adding a `get_ref` method instead of changing `get` would avoid the breaking change, but I didn't want to do so because it is kind of expected for `get` to return `&V` when querying a collection.


emilk (approved):
Makes sense

PR #7841, opened 2026-01-11
egui_kittest: Close debug_open_snapshot temp file before viewing it

yuriks:
The image file written by debug_open_snapshot was being kept open by the `tempfile` object. On Windows, this prevented `open::that` from successfully launching the viewer sometimes because the file remained locked, which can be avoided by first releasing the file handle.


yuriks:
The CI failures all look related to a bincode advisory, which I can do nothing about. Is there anything I should do here?

PR #7843, opened 2026-01-13
Add `Classes` to `UiBuilder` and some Widgets

AdrienZianne:
* Closes part of <https://github.com/emilk/egui/issues/3284>

Add a class system toward a CSS-like styling. 

Widget and Ui can implement the trait `HasClasses` which can be used later by theme engines to compute a style based on the set of classes the component has.


crates/egui/src/containers/panel.rs
AdrienZianne, line 1: Because there is an intermediary ui created between the parent and the child, I wrote this fix. Need to discuss the best way to do it

crates/egui/src/widget_style.rs
```diff
@@ -193,11 +193,207 @@ impl Style {
...
+        }
+    };
+}
+
+#[derive(Debug, Clone)]
+pub struct StyleStack {
+    pub modifiers: StyleModifiers,
+    pub parent: Option<Arc<Self>>,
+}
+
```
lucasmerlin: Is there a  reason this needs to be a seperate thing vs just putting a StyleModifiers in the UiStack?

crates/egui/src/widget_style.rs
```diff
@@ -193,11 +193,207 @@ impl Style {
...
+    pub fn separator_style(&self, _modifier: &StyleModifiers) -> SeparatorStyle {
         let visuals = self.visuals.noninteractive();
         SeparatorStyle {
             spacing: 6.0,
             stroke: visuals.bg_stroke,
         }
     }
 }
+
+pub type WidgetStyleModifier = String;
```
lucasmerlin: Maybe we should use a `Cow<str>` for this, since I imagine most of the names used here will be static strings.

lucasmerlin: Also I think we should just call this concept class and this type `ClassName` for consistency with other frameworks. And is easier to type.

crates/egui/src/widget_style.rs
```diff
@@ -193,11 +193,207 @@ impl Style {
...
+
+/// For now we use Vec for the modifiers but later we could use `SmallVev` for performance
+#[derive(Debug, Default, Clone)]
+pub struct StyleModifiers {
+    modifiers: Vec<WidgetStyleModifier>,
```
lucasmerlin: I think we should have a seperate `Classes` struct. 
When you construct a widget you don't know the WidgetState, so it's kinda pointless for the whole StyleModifier thing to be in the builder. 
 
So this should become `classes: Classes`.

crates/egui/src/widget_style.rs
```diff
@@ -193,11 +193,207 @@ impl Style {
...
+    theme: Option<Theme>,
```
lucasmerlin: I feel theme could eventually just become another class? Just set `light-mode` or `dark-mode` on the root Ui and change the style based on that. That would also make it easy for applications to add even more themes if they wanted to. I'd remove it for now.

crates/egui/src/widget_style.rs
```diff
@@ -193,11 +193,207 @@ impl Style {
...
+    state: WidgetState,
+}
+
+impl StyleModifiers {
+    /// Add multiples modifiers in one method and return the list for method chaining
+    #[inline]
+    pub fn with_modifiers(mut self, modifiers: &[WidgetStyleModifier]) -> Self {
+        self.modifiers.append(&mut modifiers.to_vec());
```
lucasmerlin: I think this should take an Iterator of WidgetStyleModifier and then use .extend.

crates/egui/src/widget_style.rs
```diff
@@ -193,11 +193,207 @@ impl Style {
...
+        self
+    }
+
+    /// Add a single modifier and return the list for method chaining
+    #[inline]
+    pub fn with_modifier(mut self, modifier: impl Into<WidgetStyleModifier>) -> Self {
+        self.modifiers.push(modifier.into());
+        self
+    }
```
lucasmerlin: You could just implement `HasModifiers` for `StyleModifier` to avoid the duplication here.

crates/egui/src/widget_style.rs
```diff
@@ -193,11 +193,207 @@ impl Style {
...
+        where
+            T: HasModifiers,
+        {
+                #[inline]
+                $(fn $name(mut self) -> Self {
+                    self.with_modifier(stringify!($name))
+                })?
+        }
+    };
+}
```
lucasmerlin: I think this should be called `define_modifiers`. But I think we can remove it entierly for now, and focus on the core things. These utilities can be added later.

crates/egui/src/ui.rs
```diff
@@ -89,6 +90,9 @@ pub struct Ui {
     /// This is an optimization, so we don't call [`Ui::remember_min_rect`] multiple times at the
     /// end of a [`Ui::scope`].
     min_rect_already_remembered: bool,
+
+    /// test
```
lucasmerlin: test?

crates/egui/src/ui.rs
```diff
@@ -139,6 +143,9 @@ impl Ui {
         let style = style.unwrap_or_else(|| ctx.global_style());
         let sense = sense.unwrap_or_else(Sense::hover);
 
+        // Temporary use of user tags as proof of concept
+        ui_stack_info = ui_stack_info.with_tag("root");
```
lucasmerlin: I like the idea of a root class, but the comment needs to be updated.
Also I think we should add public consts if we use built-in classes like this.

crates/egui/src/ui.rs
```diff
@@ -3145,6 +3166,35 @@ impl Drop for Ui {
...
 
+impl HasModifiers for Ui {
+    fn modifiers(&self) -> &StyleModifiers {
+        &self.style_stack.modifiers
+    }
+
+    fn modifiers_mut(&mut self) -> &mut StyleModifiers {
+        &mut self.style_stack_mut().modifiers
+    }
+}
```
lucasmerlin: In my opinion we should not modify a uis classes once it's created. We should instead impl `HasModifiers` (or `HasClasses`) for `UiBuilder`. Currently a lot of containers don't allow customizing the UiBuilder, but they all _should_ allow this, which would make this problem go away. (But leave this for another PR)

lucasmerlin (changes requested):
Looks good, but left some comments on naming and how it should work with Uis / containers. Let me know if you have any questions or anything is unclear!

crates/egui/src/ui.rs
AdrienZianne: I agree for the public consts, I'll change the comment

crates/egui/src/ui_builder.rs
```diff
@@ -25,6 +25,7 @@ pub struct UiBuilder {
     pub style: Option<Arc<Style>>,
     pub sense: Option<Sense>,
     pub accessibility_parent: Option<Id>,
+    pub classes: Option<Classes>,
```
lucasmerlin: Why wrap it in an Option? 
If we store it as just Classes, we could also implement HasClasses for this

crates/egui/src/ui_stack.rs
```diff
@@ -201,7 +202,7 @@ impl UiTags {
 /// Note: since [`UiStack`] contains a reference to its parent, it is both a stack, and a node within
 /// that stack. Most of its methods are about the specific node, but some methods walk up the
 /// hierarchy to provide information about the entire stack.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
lucasmerlin: Why is UiStack clone now?

crates/egui/src/widget_style.rs
```diff
@@ -131,11 +133,11 @@ impl Style {
...
     }
 
-    pub fn button_style(&self, state: WidgetState, selected: bool) -> ButtonStyle {
+    pub fn button_style(&self, classes: &Classes, state: WidgetState) -> ButtonStyle {
         let mut visuals = *self.visuals.widgets.state(state);
-        let mut ws = self.widget_style(state);
+        let mut ws = self.widget_style(classes, state);
 
-        if selected {
+        if classes.has("selected") {
```
lucasmerlin: We should make this a const string somewhere for typo-safety

crates/egui/src/widget_style.rs
```diff
@@ -193,11 +193,97 @@ impl Style {
...
     }
 }
+
+pub const ROOT_CLASS: &str = "root";
+
+pub type ClassName = Cow<'static, str>;
+
+#[derive(Debug, Default, Clone)]
+pub struct Classes {
+    classes: Vec<ClassName>,
```
lucasmerlin: This could be a SmallVec

examples/styling_engine/src/main.rs
```diff
@@ -0,0 +1,75 @@
+#![cfg_attr(not(debug_assertions), windows_subsystem = "windows")] // hide console window on Windows in release
```
lucasmerlin: I don't think we should add a unfinished example like this, maybe we could add it once we have the ThemePlugin trait?

lucasmerlin (changes requested):
I'll rename the PR to Add `Classes`, since the pr doesn't have StyleModifier anymore. 
Looking great already!

crates/egui/src/widget_style.rs
lucasmerlin: You missed this one 😄

lucasmerlin (approved):
Awesome! Please fix the remaining "select", then I think this is good to merge

crates/egui/src/widget_style.rs
```diff
@@ -193,11 +207,106 @@ impl Style {
...
     }
 }
+
+/// The root class is present on every top-level ui
+pub const ROOT_CLASS: &str = "root";
+pub const SELECTED_CLASS: &str = "selected";
+
+pub type ClassName = Cow<'static, str>;
+
+/// The classes assigned to a widget
```
emilk: This docstring should explain what classes are and what they are good for

crates/egui/src/widget_style.rs
```diff
@@ -193,11 +207,106 @@ impl Style {
...
         SeparatorStyle {
             spacing: 6.0,
             stroke: visuals.bg_stroke,
         }
     }
 }
+
+/// The root class is present on every top-level ui
+pub const ROOT_CLASS: &str = "root";
+pub const SELECTED_CLASS: &str = "selected";
```
emilk, line 224: docstring!

crates/egui/src/widget_style.rs
```diff
@@ -193,11 +207,106 @@ impl Style {
...
+
+pub type ClassName = Cow<'static, str>;
```
emilk, line 227: docstring!

crates/egui_demo_app/tests/snapshots/clock.png
emilk, line 1: Spuriously committed? I don't see anything different

emilk (commented):
This PR needs a proper description. I want any user to be able to click on a the CHANGELOG.md entry to learn more about any one feature

PR #7848, opened 2026-01-15
Fix: `Glow` not compiling

rustbasic:
egui version :  main
Fix `Glow` not compiling

        // By default, only the `wgpu` feature is enabled, so if the user added `glow` to the feature list
        // they probably wanted to use glow:


emilk:
What isn't compiling?

I also want wgpu to be the default, unless the user says otherwise

rustbasic:
> What isn't compiling?
> 
> I also want wgpu to be the default, unless the user says otherwise

Yes, It's no problem.

* #7886

PR #7852, opened 2026-01-20
Add a html5 canvas based font fallback

lucasmerlin:
This allows us to render _all_ the glyphs a browser can render, without having to include huge fonts. So this way egui in the browser should support most languages (rtl is till a problem) and any symbol.

Combined with https://github.com/emilk/egui/discussions/1344#discussioncomment-6432960 on native, you can get great coverage on unicode chars!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1622,6 +1622,8 @@ dependencies = [
  "similar-asserts",
  "skrifa",
  "vello_cpu",
+ "wasm-bindgen",
+ "web-sys",
 ]
 
 [[package]]
--- a/crates/epaint/Cargo.toml
+++ b/crates/epaint/Cargo.toml
@@ -83,6 +83,17 @@ serde = { workspace = true, optional = true, features = ["derive", "rc"] }
 
 epaint_default_fonts = { workspace = true, optional = true }
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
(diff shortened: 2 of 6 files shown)


umajho:
Hello. As a speaker of a language that uses a non-Latin writing system (Mandarin Chinese), I greatly appreciate that this is being supported. Thank you.

I see that this is currently implemented directly in `epaint` using `web-sys`. Are you considering keeping it structured this way? If so, I am concerned that this may not be the best approach. It effectively constrains all builds targeting `wasm32-unknown-unknown` to be [browsing-context](https://developer.mozilla.org/en-US/docs/Glossary/Browsing_context)-only, which is not always a valid assumption.

For example, I am planning to sandbox UIs written with egui using Wasmtime: raw input events would be passed into the WASM guest environment, while texture diffs would be emitted back to and handled by the host.

It may therefore be more preferable to define a `GlyphRenderer` trait in `epaint`, and provide an implementation such as `CanvasGlyphRenderer` in `eframe` that will be passed to `epaint` someway within the `WebRunner`. This would avoid hard-coupling `epaint` to the browsing context for `wasm32-unknown-unknown`.

lucasmerlin:
Hi @umajho! We do indeed plan to add some kind of GlyphRenderer / GlyphFallback api, which should allow you to do this in other wasm contexts. This was just a quick poc to see how well it works at all.

PR #7856, opened 2026-01-22
Custom storage providers

Leinnan:
It replaces the `persistence_path` field in `NativeOptions` with `storage_build` and allowing to specify a custom `Storage` for apps. No functionality is lost, app creator can still specify file to use by using `storage_build: StorageProvider::AtPath(path)` instead of `persistence_path: Some(path)`. 

This allows to add a custom storage on all platforms(wasm as well), which is especially useful if anyone wants to create app for platform which is not as supported like mobile.

There is also example which shows a custom storage implementation where data is stored in text file in the directory from which the program was called.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1129,6 +1129,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "custom_storage"
+version = "0.1.0"
+dependencies = [
+ "eframe",
+ "egui_demo_lib",
+ "egui_extras",
+ "env_logger",
+ "image",
+ "serde",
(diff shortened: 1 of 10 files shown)


Leinnan:
@emilk Did you have any chance to look at it?

Ved-s:
This is also very useful for hosting multiple egui apps on the same website so you can specify custom localStorage location for different apps. Would love to see this merged

Ved-s:
What's the reason for using a fn pointer for custom Storage builder? Can you make it use a Box<dyn Fn ...>?

PR #7859, opened 2026-01-23
Add font variations API

valadaptive:
* Closes N/A

This was mostly from last month, but I never got around to submitting it.

This PR adds font variation coordinates to the `TextFormat` struct, and uses them when rendering text. The coordinates are stored in a `SmallVec`; I've chosen to store up to 2 inline, which makes it take up 24 bytes (the minimum possible for a `SmallVec`). The variation axis tags are stored as the `font_types::Tag` type, which I've chosen to re-export from `epaint::text`.

The variation coordinates are resolved to a `skrifa::Location` during font rendering/scaling, and are cached in the same way as all the other scaled metrics. I've renamed the `ScaledMetrics` struct to `StyledMetrics`, since it now also contains the resolved variation coordinates. I haven't benchmarked the performance of text layout with variation coordinates, but the existing text layout performance is unchanged.

I've replaced the API for manually overriding a font's weight (https://github.com/emilk/egui/pull/7790) with an API for manually overriding any variation coordinates via `FontTweak`. This should support the same use case as #7790 while being substantially more flexible.

I have *not* yet added any higher-level API for mapping style attributes (weight, width, slant, etc) to variation coordinates or to different font faces within a single family. That's a pretty huge can of worms, and it'd involve rethinking the split between `FontId` and `TextFormat` (and whether `FontId` is so big that we should provide a way to reuse it). This API is intentionally pretty low-level for now.

Likewise, I've intentionally not used variation coordinates when computing a font's row height. I can't think of any fonts that change their vertical metrics depending on variation axes, so this should be fine for now.

### DIFF
--- a/crates/egui/src/style.rs
+++ b/crates/egui/src/style.rs
@@ -2894,10 +2894,10 @@ impl Widget for &mut FontTweak {
 
                     let tag_text = &mut *tag_text.lock();
                     let response = ui.text_edit_singleline(tag_text);
-                    if response.changed() {
-                        if let Ok(new_tag) = Tag::new_checked(tag_text.as_bytes()) {
-                            *tag = new_tag;
-                        }
+                    if response.changed()
+                        && let Ok(new_tag) = Tag::new_checked(tag_text.as_bytes())
+                    {
+                        *tag = new_tag;
                     }
                     // Reset stale text when not actively editing
                     // (e.g. after an item was removed and indices shifted)


crates/egui/src/style.rs
```diff
@@ -2874,6 +2879,35 @@ impl Widget for &mut FontTweak {
...
+                    ui.add(DragValue::new(value));
+                    ui.end_row();
+                }
+                if ui.button("Add coord").clicked() {
+                    coords.push(b"zzzz", 0.0);
+                }
+                if ui.button("Clear coords").clicked() {
+                    *coords = Default::default();
+                }
+                ui.end_row();
```
emilk, line 2925: This feels like a debug-UI that should either be removed, or hidden by a `if cfg!(debug_assertions)`

emilk: …or made better by allowing us to name (and rename) the tags

emilk (approved):
Nice!

PR #7866, opened 2026-01-27
Avoid repaints on device mouse motion outside window

inktomi:
## Summary
- Ignore raw device mouse motion unless the window is focused and the pointer is inside it
  - Also handles pointers starting down and then moving into or out of the window (drag & drop)
- Prevents global mouse motion from triggering continuous repaint loops
- Applies to both glow and wgpu backends

## Testing
- I ran the check script, nothing seemed to fail


crates/eframe/src/native/glow_integration.rs
```diff
@@ -448,11 +448,25 @@ impl WinitApp for GlowWinitApp<'_> {
...
+
+                    let pointer_in_window = viewport
+                        .egui_winit
+                        .as_ref()
+                        .map(|state| state.is_pointer_in_window())
+                        .unwrap_or(false);
+
+                    if !pointer_in_window {
+                        return Ok(EventResult::Wait);
+                    }
```
emilk: This logic could be moved into `fn on_mouse_motion`

One thing though: I think if the mouse is pressed down we should still report it, i.e. if dragging something inside of egui and moving the mouse outside the screen (e.g. a DragValue)

inktomi: Good ideas! I will move the code and update it. Thank you!

crates/egui_kittest/README.md
inktomi, line 1: I noticed what I think are a couple typos here, hope it's ok to include with the other PR.

emilk (approved):
Great!

PR #7867, opened 2026-01-28
Fixes the overly aggressive overflow elision in `truncate()` and similar for os scaling other than 100%

### LINKED ISSUE #7818: Overflowing text is not cut off correctly
**Describe the bug**
At least in the scenario I'm presenting (Multiline TextEdit + LayoutJob), the elision of overflowing text is too aggressive and leaves room for an extra character. This happens for both font families, but the monospace comparison should make it fairly obvious.

**Minimal code example**
```
let mut layouter = |ui: &egui::Ui, buf: &dyn egui::TextBuffer, wrap_width: f32|
{
	let mut job = LayoutJob::default();
	job.wrap = TextWrapping 
	{
		max_rows: 4,
		max_width: wrap_width,
		overflow_character: Some('…'), // <- same char used in the precompiled string
		..Default::default()
	};
	
	let mut format = TextFormat::default();
	format.font_id = FontId::new(12.0, FontFamily::Monospace);
	
	job.append(
		buf.as_str(),
		0.0,
		format
	);
	
	ui.fonts_mut(|f| f.layout_job(job))
};

ui.horizontal(|ui|
{
	// this will be elided to "ABCDEFGHIJKLMNOPQRSTUVWXYZ…" (26 + 1 characters)
	ui.add_sized(
		vec2(64.0, 64.0),
		egui::TextEdit::multiline(&mut String::from("ABCDEFGHIJKLMNOPQRSTUVWXYZABCDEFGHIJKLMNOPQRSTUVWXYZ"))
			.layouter(&mut layouter)
	);
	
	// this fits in full (27 + 1 characters)
	ui.add_sized(
		vec2(64.0, 64.0),
		egui::TextEdit::multiline(&mut String::from("ABCDEFGHIJKLMNOPQRSTUVWXYZA…"))
			.font(FontId::new(12.0, FontFamily::Monospace))
	);
});

```

**Expected behavior**
The overflowing logic fits as many characters as possible, matching its precompiled counterpart.

**Screenshots**
Image

**Desktop**
 - OS: Win10
 - Version: 0.33.3 release, latest main

RndUsr123:
* Closes #7818

### DIFF
--- a/tests/egui_tests/tests/regression_tests.rs
+++ b/tests/egui_tests/tests/regression_tests.rs
@@ -1,8 +1,5 @@
 use egui::accesskit::Role;
-use egui::{
-    Align, Button, Color32, Image, Label, Layout, RichText, Sense, TextWrapMode, include_image,
-    vec2,
-};
+use egui::{Align, Color32, Image, Label, Layout, RichText, Sense, TextWrapMode, include_image};
 use egui_kittest::Harness;
 use egui_kittest::kittest::Queryable as _;
 


emilk:
It would be nice with a test for this too 🤔

RndUsr123:
Done! 5 character sequences at 3 different scaling values sounds good enough to me..?

Edit: I just noticed `text_layput.rs` already has a `test_truncate_width` normal test, but it always uses 1.0 pixels per point so it doesn't catch this one issue. Should I work on that instead?

RndUsr123:
I've added a normal test for this, which is probably more sensible. Let me know if the snapshot one should be removed

crates/epaint/src/text/text_layout.rs
```diff
@@ -1166,6 +1168,42 @@ mod tests {
...
+                job.wrap.max_width = target_width;
+                job.wrap.max_rows = 1;
+                let elided_galley = layout(&mut fonts, pixels_per_point, job.into());
+                assert!(elided_galley.elided);
+
+                let test_galley = layout(
+                    &mut fonts,
+                    pixels_per_point,
+                    Arc::new(LayoutJob::single_section(
+                        (0..elided_galley.text().len())
```
emilk: note that `.text()` returns the full, non-ellided text

RndUsr123: My bad!

The new commit should fix it: I used `char_count_excluding_newline()` on the first row, which looks to be the most direct solution

emilk:
Thanks!

PR #7868, opened 2026-01-28
Fix `CentralPanel::show_inside_dyn` to round `panel_rect`

ripopov:
While using CentralPanel::show_inside_dyn inside egui_tiles I've noticed that sometimes it renders "Unaligned" message with certain tile positions. Match SidePanel behavior by calling round_ui() on the panel rect.
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -698,9 +698,9 @@ checksum = "8f1fe948ff07f4bd06c30984e69f5b4899c516a3ef74f34df92a2df2ab535495"
 
 [[package]]
 name = "bytes"
-version = "1.8.0"
+version = "1.11.1"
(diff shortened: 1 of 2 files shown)


crates/egui/src/containers/panel.rs
```diff
@@ -1020,7 +1020,7 @@ impl CentralPanel {
     ) -> InnerResponse<R> {
         let Self { frame } = self;
 
-        let panel_rect = ui.available_rect_before_wrap();
+        let panel_rect = ui.available_rect_before_wrap().round_ui();
```
emilk: We should probably call `.round_ui()` inside of `fn available_rect_before_wrap` instead 🤔

ripopov: Done 😅

PR #7873, opened 2026-01-30
Eframe: Use alternate display size if browser is firefox

tye-exe:
Currently in firefox the display scaling is read as 2x what it should be. This causes issues as cursor position is still using the 1x scaling, and any egui applications being at 2x the desired scale.
This is a patchwork fix that detects is the browser is firefox by looking at the user agent (any of web devs that have a better knowledge of how to detect a firefox browser please let me know) and falls back to an alternate display size calculation that produces the expected 1x scaling.

* Related PRs: <https://github.com/emilk/egui/pull/5481>


tye-exe:
This is open as a draft due to me having issues with running the check script. I will be able to dedicate more time to trying to run the check script later.

PR #7888, opened 2026-02-05
**fix(web): prevent entire page from scrolling out of view in Chrome (WASM)**

### LINKED ISSUE #7887: [Bug] Entire application scrolls out of view unexpectedly in Chrome (WASM)
[Bug] Entire application scrolls out of view unexpectedly in Chrome (WASM)

**Description:**
- **egui version:** `main` and all other versions
- **Browser:** Google Chrome
- **Issue:**
When running the application compiled to WASM in Chrome, 
the entire program occasionally scrolls upward suddenly 
while interacting with scrollbars or other UI elements. 
This causes the application to disappear from the visible area of the browser window. 
This issue appears to persist across all versions of egui, including the current `main` branch. 

https://github.com/user-attachments/assets/7ef25fc9-0dee-4945-b7d4-1353e81da3da

rustbasic:
**fix(web): prevent entire page from scrolling out of view in Chrome (WASM)**

* Closes #7887

**Problem**
When using `egui` on the web, the browser (especially Chrome) occasionally triggers an unwanted page-level scroll. This happens because the hidden input element (text agent) used for IME/text input is sometimes positioned outside the visible bounds of the canvas, causing the browser to "scroll it into view."

**Solution**
I modified the `move_to` function in the web's `text_agent` to ensure the input element's position stays within the canvas height. By clamping the `top` property between `0.0` and `canvas_height`, we prevent the browser from incorrectly scrolling the entire page when the text agent moves.

- **Specific change:** Applied `clamp(0.0, canvas_height)` to the `clamped_y` value before setting the CSS `top` property.

### DIFF
--- a/crates/eframe/src/web/text_agent.rs
+++ b/crates/eframe/src/web/text_agent.rs
@@ -34,13 +34,12 @@ impl TextAgent {
         style.set_property("outline", "none")?;
         style.set_property("width", "1px")?;
         style.set_property("height", "1px")?;
-        // Prevent auto-zoom on mobile browsers (requires at least 16px).
-        style.set_property("font-size", "16px")?;
-
         style.set_property("caret-color", "transparent")?;
         style.set_property("position", "absolute")?;
         style.set_property("top", "0")?;
         style.set_property("left", "0")?;
+        // Prevent auto-zoom on mobile browsers (requires at least 16px).
+        style.set_property("font-size", "16px")?;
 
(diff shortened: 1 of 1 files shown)


rustbasic:
I'm completed.

Prevented unexpected mobile zooming and related malfunctions.

emilk (approved):
Makes sense - thanks!

PR #7891, opened 2026-02-07
Add option ```scroll_zooms``` for mouse wheel to zoom instead of scroll inside a ```Scene``` container

### LINKED ISSUE #7650: Scroll to zoom without ctrl in `egui::Scene`

Tristan117:
Closes <https://github.com/emilk/egui/issues/7650>
[ X] I have followed the instructions in the PR template

Previously submitted as <https://github.com/emilk/egui/pull/7666> but replaced to clean up the git history.

This feature allows building image viewers and similar applications using a `Scene` container where the desired mouse wheel input without a modifier is to zoom rather than scroll/pan.

Alternatives considered:

Swap zoom and scroll variables used in `Scene`
- Would cause touch input zoom to scroll/pan instead which would be inconsistent with standard user interface interaction
- This also means there is no simple way to swap zoom and vertical scroll modifiers for this use case e.g. ctrl+mousewheel pans vertically while mousewheel alone zooms. 
- What is implemented results in ctrl-mousewheel and mousewheel only being identical inputs.

Modify `ctx.option.zoom_modifier`
- Doesn't work because `Modifiers:NONE` never matches to anything (and therefore disables mouse wheel zoom globally).
- Refactoring this to provide a different option per container is a large amount of work and does not seem to be a use case that is likely to be used outside of `Scene`

The logic used inside `Scene` to achieve this is fairly simple and only adds one extra dependency on `input_options.scroll_zoom_speed` to ensure that the zoom speed setting change is consistent across the application.


Tristan117:
@lucasmerlin could you please approve the workflow for the tests again? 

It'd be great to see this included as at the moment I have to keep updating a patched version of egui with every new release.

Tristan117:
Fixed typo (American vs English spelling of behaviour)
PR label and demo screenshot regeneration needs to be done by a maintainer

lucasmerlin (approved):
Thanks!

ddmitv:
Hey @emilk, could you please take a look at this PR when you have a moment? It has already been approved, but it seems to be stuck on the CI checks (due to a missing PR label and snapshot updates).

Tristan117:
@lucasmerlin are you able to add a label to the pull request so it can be merged? Thanks!

PR #7877, opened 2026-02-02
fix: disable Enter while IME is active

sevenc-nanashi:
This PR adds <kbd>Enter</kbd> to the list of keys that is disabled during IME is active.

This Branch:

https://github.com/user-attachments/assets/c44a09dd-0b24-414b-bb9b-695cbe792579

Main Branch (unfocused every time I commit preedits)

https://github.com/user-attachments/assets/72048468-f414-4544-8ff8-106e39038807

* Closes <https://github.com/emilk/egui/issues/7809>
* [x] I have followed the instructions in the PR template

PR #7898, opened 2026-02-10
Add IME language selection to InputOptions for CJK support

rustbasic:
Add IME language selection to InputOptions for CJK support

**Overview**
This PR adds a new configuration option to select specific IME (Input Method Editor) processing modes, particularly for CJK (Chinese, Japanese, Korean) languages. This allows for more specialized handling of character composition depending on the selected language.

**Changes**
- Introduced `ImeLanguage` enum (None, Korean, Japanese, Chinese) with `serde` support.
- Added `ime_language` field to `InputOptions` struct.
- Updated `InputOptions::ui` to include a `ComboBox` for easy language selection in the settings panel.

**Why this is needed**
Proper CJK support often requires different composition logic depending on the language. By allowing users to explicitly select their IME mode, we can provide a more robust input experience for international users.

### DIFF
--- a/crates/eframe/src/web/text_agent.rs
+++ b/crates/eframe/src/web/text_agent.rs
@@ -87,9 +87,17 @@ impl TextAgent {
         };
 
         let on_composition_update = {
+            let input = input.clone();
             move |event: web_sys::CompositionEvent, runner: &mut AppRunner| {
                 let Some(text) = event.data() else { return };
-                let event = egui::Event::Ime(egui::ImeEvent::Preedit(text));
+
+                let start = input.selection_start().ok().flatten().unwrap_or(0) as usize;
+                let end = input.selection_start().ok().flatten().unwrap_or(0) as usize;
+                let event = egui::Event::Ime(egui::ImeEvent::Preedit {
+                    text_mark: text,
+                    start,
+                    end,
+                });
                 runner.input.raw.events.push(event);
                 runner.needs_repaint.repaint_asap();
             }
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -619,7 +619,6 @@ impl State {
                 self.egui_input
                     .events
                     .push(egui::Event::Ime(egui::ImeEvent::Preedit(String::new())));
-                self.ime_event_disable();
             }
         }
     }
--- a/crates/egui/src/input_state/mod.rs
+++ b/crates/egui/src/input_state/mod.rs
@@ -53,10 +53,23 @@ impl SurrenderFocusOn {
     }
 }
 
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub enum ImeLanguage {
+    None,
+    #[default]
+    Korean,
+    Japanese,
+    Chinese,
+}
+
 /// Options for input state handling.
 #[derive(Clone, Copy, Debug, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub struct InputOptions {
+    /// The language used for specialized IME (Input Method Editor) processing.
+    pub ime_language: ImeLanguage,
+
     /// Multiplier for the scroll speed when reported in [`crate::MouseWheelUnit::Line`]s.
     pub line_scroll_speed: f32,
 
@@ -110,6 +123,7 @@ impl Default for InputOptions {
         };
 
         Self {
+            ime_language: ImeLanguage::default(),
             line_scroll_speed,
             scroll_zoom_speed: 1.0 / 200.0,
             max_click_dist: 6.0,
@@ -127,6 +141,7 @@ impl InputOptions {
     /// Show the options in the ui.
     pub fn ui(&mut self, ui: &mut crate::Ui) {
         let Self {
+            ime_language,
             line_scroll_speed,
             scroll_zoom_speed,
             max_click_dist,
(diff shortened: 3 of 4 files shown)


umajho:
Backspacing in a `TextEdit` will stop from working after an IME has been enabled inside that `TextEdit`.
Even if I switched back to English, this bug still persists.

Tested on macOS Safari and Chrome. The screenshot below is in Chrome.

repro

---

By the way, since I'm not familiar with Korean at all, I'm curious whether this solution is universal.
As an IME user myself (Chinese Pinyin (Shuangpin)), I don't remember I have ever seen an application that requires the user to configure the IME language explicitly. Though, maybe this is common among Korean IME users…?

On macOS, both Japanese IME and Korean IME already work with the current code. (Japanese IME won't emit an extra <kbd>Enter</kbd> while confirming, Korean IME don't need <kbd>Enter</kbd> twice. [details here](https://github.com/emilk/egui/pull/7877#issuecomment-3868430442))
So maybe there is a way to make the behavior correct on other platforms without the need to ask users to select their IME's language.

Also, since browsers (Safari, Chrome and Firefox (Zen)) on macOS can also handle both (Japanese and Korean) cases correctly, I wonder if this is also true for Windows and Linux. (I don't have access to Windows devices for now, and I have no experience configuring Japanese/Korean IMEs on any Linux DE…)

rustbasic:
Thanks @umajho

I also noticed the issue a bit late and am currently in the process of identifying the cause and implementing a fix.

The reason for handling Korean, Japanese, and Chinese separately is that the current implementation is structured such that changes made for Chinese tend to break the Korean logic. Additionally, there is an issue on Linux where `IME::Enable` is not being triggered, which might lead to divergent handling methods in the future.

To be more precise, it's difficult to find someone who can test all languages across every platform. While it would be ideal for the behavior to be consistent across all systems and languages, from my perspective, the sheer diversity of Chinese input methods and the current state of Linux make that quite challenging.

rustbasic:
@umajho 

Would you like to test this?
I have even solved problems that couldn't be solved for a year.

umajho:
I can confirm that the backspacing issues is gone.

I have also tested the native build. It seems this doesn't break the already-correct behavior on mac for the builtin Shuangpin (Pinyin), Japanese and 2-Set Korean IMEs, no matter which `ime_language` is used.

(Although, I still think requiring specifying the IME language a little unintuitive. IMO an ideal solution is to fix the inconsistencies on `winit`. I speculate that the root issue is that `winit` doesn't consume the confirming keys[^1] for Japanese IMEs (and maybe Chinese IMEs as well) on Windows, but unfortunately I don't have the environment to test this.)

[^1]: The last section of the opening of [this pr](https://github.com/rust-windowing/winit/pull/4478) states that macOS does this, which explains why macOS doesn't have this problem.

rustbasic:
@umajho 

Users don't need to select a specific language if they have no issues with their IME when 'None' or 'Korean' is selected. This also applies to Japanese or Chinese users; if 'None' works correctly for them, no selection is necessary.

Although not included in this current PR, there is a difference in preference for the IME EventFilter: Chinese users generally want to include 'Enter' in the filter, whereas Korean users prefer it to be excluded.

umajho:
> @umajho 
> 
> Users don't need to select a specific language if they have no issues with their IME when 'None' or 'Korean' is selected. This also applies to Japanese or Chinese users; if 'None' works correctly for them, no selection is necessary.
> 
> Although not included in this current PR, there is a difference in preference for the IME EventFilter: Chinese users generally want to include 'Enter' in the filter, whereas Korean users prefer it to be excluded.
> 
> 

I apologize, but I am not very convinced by this approach, because:
1. Lots of, if not most, apps support IMEs correctly without knowing which kind of IME is used, this should also be possible for egui.
2. I doubt most developers will know/care that there is an IME language selection thing in egui and bother integrating it into their apps. If an app doesn't make the selector available, users are out of luck.
3. And even if an app includes the selector, users still need to discover it, which is not ideal for UX.
4. There are users who use multiple IMEs all the time. To them, switching IME language settings is cumbersome.
5. As I stated, I think the root issue is likely in `winit`.

Because of the second and the third point, I think this pr can be considered a temporary workaround for language exclusive (e.g. Japanese exclusive) apps, but it should not be considered as a long term solution. (I see that you didn't link any issue to be closed by this pr, so maybe you and me are on the same page here…?)

Since people do need a workaround, I don't object to this PR being merged.

## More on the fifth point

The following behavior:
> “while confirming, the key press event of the confirming key is emitted along with the IME commit event for Japanese/Chinese IME.”

is likely a bug of `winit`'s IME handling logic on Windows.

The correct behavior should be:
> “`winit` should not emit key press events for confirming keys for Japanese/Chinese IMEs, while it should emit press of <kbd>Enter</kbd> for Korean IMEs.”

Which is correctly implemented by `winit` on macOS and Linux with ibus (I have not tested fcitx).
It is also correctly implemented by eframe's custom integration for web.

Therefore, in my opinion, the ideal way to fix this IME bug in `egui` is to fix the bug in the `winit` crate, so that confirming keys for Japanese/Chinese IMEs will always be discarded inside `winit` on every platform, just like how it is handled on macOS.

---

Update 1:

I find [this line](https://github.com/emilk/egui/blob/64a96ef3917d4e305a79fa5f460af5fc73eabcf8/crates/egui/src/widgets/text_edit/builder.rs#L1112) of ignoring commits of newline characters interesting. I'd like to investigate it later.

Update 2:

I just realized that the system the author of #7877 uses is not Linux (I was misled by the look of the top bar). I have tested Linux + KDE + ibus myself, it turns out it also already works with the current code across different IMEs without the need to know the kind of the IME being used, just like macOS.

And I also found out that the function `remove_ime_incompatible_events` doesn't benefit macOS and Linux (at least for ibus) at all.
In fact, running this function there even leads to a bug: When using a Korean IME in other apps, arrow keys can move the cursor even if the composition has not been done, but with the so-called incompatible events being removed, the first press of arrow keys would not move the cursor.

I suspect that the code I linked in Update 1 filtered out the enter event for Korean IMEs on Windows. I will update my test on this when I have access to Windows. (maybe in 2~3 weeks)

umajho:
(My last commet is duplicated for some reason…)

rustbasic:
I’ve finished.
 
I simplified it as much as possible and tested it on both Windows and Web. 

There are some bugs when typing in Korean on mobile; it seems like mobile Korean input requires handling a much more complex process directly. 
Since I haven't completed that part yet, I didn't include it in this Pull Request.

PR #7901, opened 2026-02-11
Fix polygon feathering artifacts at sharp corners

### LINKED ISSUE #7424: Bug in polygon tessellator with feathering

segeljakt:
The miter join normal in `add_line_loop` has magnitude `1/cos(θ/2)`, which blows up at sharp corners. For a nearly-degenerate triangle this can push feathering vertices 100+ px outside the polygon, creating visible line artifacts.

This clamps the miter factor to 2.0 (same idea as SVG's `stroke-miterlimit`). Corners wider than ~60° are unaffected. Sharper corners get beveled feathering instead of mitered. This barely has a visual impact since the corners are sub-pixel.

Snapshot diffs are just sub-pixel changes in menu bar tabs.

## Tests

- Added `test_feathering_no_artifacts_on_sharp_triangles` with a few sharp polygon cases that fail without the fix
- Verified visually with the Bézier curve demo

Fixes #7424

## Example

Before the fix: 
image

After the fix:
image

## Disclosure

I used Claude Code to diagnose the error and come up with the fix. I hope this is ok.


emilk:
Thanks for working on this!

in https://github.com/emilk/egui/pull/7554 @podusowski added some tests for this - could you add them to this PR too, so we can compare the fixes?

emilk:
This looks pretty bad 😬 
image

Looks like this PR, just like https://github.com/emilk/egui/pull/7554, fixes the extrusions by nuking the feathering.

I believe a proper fix that handles _both_ requires adding another vertex to sharp corners

Issue #5136, opened 2024-09-20
Wayland: `request_repaint` ignored if minimized


emilk:
According to @jleibs, calling `ctx.request_repaint` does not result in a call to `App::update` when on Wayland and the app is minimized (Hyprland placed on non-visible workspace).

* See https://github.com/rerun-io/rerun/pull/7422 for more
* We have the same problem on web: https://github.com/emilk/egui/issues/5112

dimtpap:
I came across this because I noticed my app is not responding to closing when it is minimized. It seems that actually no events are received when the window is minimized. This is caused because of [this call](https://github.com/emilk/egui/blob/707a33904778aac2927ff7972eb534960dff91e0/crates/eframe/src/native/glow_integration.rs#L720) hanging and blocking the event loop thread.

I found exactly this reported in this issue https://github.com/rust-windowing/glutin/issues/1591

dimtpap:
[This post](https://emersion.fr/blog/2018/wayland-rendering-loop/) (written by someone who works on Wayland itself) goes in detail about this.

Based on that, adding this line (note: this is an ignored result)
```rust
_ = gl_surface.set_swap_interval(context, glutin::surface::SwapInterval::DontWait);
```
before calling swap_buffers will make it not block therefore fixing this issue and the closing thing I mentioned, but now it repaints faster than vsync. This has the same effect as setting `vsync` to false in `NativeOptions`. I believe this matches the behavior of egui on other platforms, repaint even if hidden and pay the GUI code cost.

Sidenote, according to the post this is the equivalent code that should be added to write a proper Wayland render loop
```rust
_ = gl_surface.set_swap_interval(context, glutin::surface::SwapInterval::DontWait);
window.pre_present_notify();
```

But since this makes `winit` repaint only when the compositor wants, this issue re-appears. The proper way is to probably have that `App:tick`.

Jackrin:
Hello, I'm experiencing this issue right now as I'm trying to build a GUI.
I'm on Hyprland and the issue is the same as the one described. In my case it doesn't even matter if the app is in a hidden workspace, the only way I can get the update function to run is if something happens, like mouse movement.

I tested this by putting a print statement in the update function, and it doesn't print unless my mouse is hovering the app, and also moving as well.

Needless to say this makes the app unusable. Is a fix on the way? Is there any way I could patch this behaviour myself? I just need the app the visually update regardless of what i'm doing or where the app is (different workspace, etc.)

Thanks.

[...]

Issue #5112, opened 2024-09-16
Run `App::update` even when in background thread


emilk:
If the eframe web app is in hidden (e.g. is in a background tab in the browser), then any call to `request_repaint` will be ignored, and not result in a call to `App::update`.

That's because eframe checks if `request_repaint` has been called from `requestAnimationFrame` callback, which is not called if the application is hidden.

### Solution
If the web page is hidden, then calling `request_repaint` should schedule a `setTimer(…)` callback to call `App::update()` (must done via a setTimer delay, or we could have `App::update` call `request_repaint` which in turn calls `App::update`.

We also need some guards to make sure that multiple calls to `request_repaint` in a row only results in one call to `App::update`, for instance via this pseudo-code:

```
// Won't be called if hidden
fn on_request_animation_frame() {
    if state.needs_repaint {
       app.update();
       paint();
       state.needs_repaint = false;
    }
}

fn on_request_repaint() {
   state.needs_repaint = true;
    if document.is_hidden {
       setTime(5ms, app_update_if_needed);
    }
}

fn app_update_if_needed() {
   if state.needs_repaint {
       app.update();
       state.needs_repaint = false;  
    }
}

```

### `App::tick`
Calling the full `App::update` when the application is hidden (background thread or minimized on native, for instance) is bad for several reasons:

A) we don't always know what the size of the viewport is/should be
B) we pay the cost of GUI even though we may only need to process some events

I think therefor it makes sense to split `App::update` into two functions:

```rs
trait App {
    /// You can run logic here, but can't show any GUI.
    ///
    /// Called once before each call to [`Self::show`], but may also be called
    /// without a subsequent call to `show`, for instance if the application is in a background thread.
    ///
    /// The egui context may ONLY be used to schedule repaints, nothing else.
    fn tick(&mut self, ctx: &egui::Context, frame: &mut Frame)

    /// Show the GUI. Each call to this is preceded by a call to [`Self::tick`].
    fn show(&mut self, ctx: &egui::Context, frame: &mut Frame);
}
````

Mingun:
Probably you shouldn't do that. Correct me if I'm wrong, but it seems there a battle of two opinions. When egui starts updating in background, someone complain why application consumes CPU when it hidden. Then egui stops updating in background and another people asks why his code didn't run when app not on the screen. It seems to me that there was a several iterations of this issue at the time present.

[...]

PR #7905, opened 2026-02-14
Improve behavior of invisible windows

### LINKED ISSUE #5229: [eFrame] Can't set window Visible(true) after setting the window Visible(false) on Windows
**Describe the bug**
Due to the way eframe handles querying for viewport commands, it relies on the window event `RepaintNow` or `RedrawRequested` to process the queued viewport commands. The problem is, once you set a window Visible(false), those events are no longer sent to the window (at least on Windows 11) therefore making it impossible to set the window Visible(true) using a viewport command.

https://github.com/emilk/egui/blob/master/crates/eframe/src/native/glow_integration.rs#L719
https://github.com/emilk/egui/blob/master/crates/eframe/src/native/glow_integration.rs#L1297
https://github.com/emilk/egui/blob/master/crates/egui-winit/src/lib.rs#L1389

The example below just opens a window then uses a new thread to toggle its visibility,

**To Reproduce**
```toml
eframe = "0.29.1"
egui = "0.29.1"
```

```rust
use eframe::{egui, App};
use std::time::{Duration, Instant};
use egui::{Id, ViewportBuilder, ViewportCommand};
use crossbeam::channel;
use std::sync::{Arc, Mutex};
use std::thread;

struct VolumeApp {
}

impl App for VolumeApp {
        let     volume: u8= ctx.data(|map| *map.get_temp(Id::new(1337)).unwrap_or(&0));
        let     start_time: Option<Instant>= ctx.data(|map| *map.get_temp(Id::new(1338)).unwrap_or(&None));

        println!("Start time {:?}", start_time);

        // Only show the UI when the start_time is set (window is visible)
        if start_time.is_some() {
                ui.label(format!("Volume: {:.0}%", volume)); // Display the volume percentage
            });
        }
    }
}

fn main() {
  //  std::env::set_var("RUST_LOG", "trace");
    env_logger::builder().init();

    // Create a crossbeam channel for sending volume levels
    let (volume_sender, volume_receiver) = channel::unbounded();
    let volume_receiver = Arc::new(Mutex::new(volume_receiver));

    // Spawn a thread to simulate sending volume levels to the app
    let volume_sender_clone = volume_sender.clone();
    thread::spawn(move || {
        let test_volumes = vec![10, 20, 30, 40, 50, 60, 70, 80, 90, 100];

        for volume in test_volumes {
            volume_sender_clone.send(volume).unwrap();
            thread::sleep(Duration::from_secs_f32(2.0)); // Simulate delay between volume changes
        }
    });

    let viewport = ViewportBuilder::default()
        .with_resizable(false)
        .with_always_on_top()
        .with_decorations(true)
        .with_inner_size(egui::vec2(200.0, 20.0));

        viewport,
        ..Default::default()
    };

    eframe::run_native(
        "MyApp",
        options,
        Box::new(|cc| {
            let volume_receiver = volume_receiver.clone();
            let context =cc.egui_ctx.clone(); // Store the frame handle for controlling visibility

            // Background thread to manage volume updates and window visibility
            {
                thread::spawn(move || {
                    loop {
                        // Wait for a new volume to arrive
                        if let Ok(new_volume) = volume_receiver.lock().unwrap().recv() {
                            println!("Set Visible");

                            context.data_mut(|map| {
                              map.insert_temp(Id::new(1337), new_volume);
                              map.insert_temp(Id::new(1338), Some(Instant::now()));
                            });

                            // Set the window visible
                            context.send_viewport_cmd(ViewportCommand::Visible(true));
                            context.request_repaint();
                            context.send_viewport_cmd(ViewportCommand::MousePassthrough(true));

                            // Wait for 1 second before hiding the window
                            thread::sleep(Duration::from_secs(1));

                            // Hide the window after 1 second of inactivity
                            println!("Set Invisible");
                            context.send_viewport_cmd(ViewportCommand::Visible(false));
                        }
                    }
                });
            }

            // Initialize the app and pass control to eframe
            Ok(Box::new(VolumeApp {

            }))
        }),
    )
        .expect("Failed to run the app");
}
```

**Expected behavior**
The window should open/close every 2 seconds but it cannot open after being set invisible since viewport commands are no longer polled.

 - OS: Windows 11

### LINKED ISSUE #7776: Windows: High CPU Usage when not visible
**Describe the bug**
When not visible (by sending the `egui::ViewportCommand::Visible(false)` command) a simple egui application will use ~16% CPU on my CPU while using 0% otherwise

**To Reproduce**
1. Take the following egui application:
```rust
use eframe::egui;

        ..Default::default()
    };

    eframe::run_simple_native("My egui App", options, move |ctx, _frame| {
            if ui.button("Make invisible").clicked() {
                ctx.send_viewport_cmd(egui::ViewportCommand::Visible(false));
            }
        });
    })
}

```
2. Run it
3. Observe CPU usage while open, if the mouse cursor is not in the window or the window is minimized it will drop to 0
4. Press the button to make the window invisible
5. CPU usage will go up a lot

**Expected behavior**
CPU usage should stay low, similar to when it's minimized

**Screenshots**

Normally
Image
When invisible
Image

**Desktop**
 - OS: Windows 11 26H1 (28000.1340)
 - Version: eframe 0.33.3, also tested on latest master ([bfaf1b4](https://github.com/emilk/egui/commit/bfaf1b44f2dd528fbdeaa865eba30b400075d0a8))

**Additional context**
- Issue persists even if the window is made invisible with [WindowsAndMessaging::ShowWindow](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/UI/WindowsAndMessaging/fn.ShowWindow.html)
- CPU usage resumes to be lower when the window is visible again
- Most of the issue's wording is taken directly from #3982 since the issue is very similar (but not quite the same issue)
- Issue seems to be present with both the wgpu and glow backends

gcailly:
## Summary
* Closes #5229
* Closes #7776

On Windows, once a window is hidden with `ViewportCommand::Visible(false)`, two problems occur:

1. **Window can never be shown again** — Windows stops sending `RedrawRequested` events to invisible windows, and viewport commands are only processed during `run_ui_and_paint`, which is triggered by `RedrawRequested`. This creates a deadlock:
```
Visible(false) → window hidden → no RedrawRequested → run_ui_and_paint never called → Visible(true) stuck in queue → window stays hidden forever
```

2. **High CPU usage** — The event loop spins at full speed with `ControlFlow::Poll` even for invisible windows, and repaint requests are scheduled immediately, causing a tight loop that burns CPU.

## Fix

**For #5229:** In `check_redraw_requests`, after calling `window.request_redraw()`, detect invisible windows via `window.is_visible() == Some(false)` and call `run_ui_and_paint` directly for them. This ensures pending viewport commands (including `Visible(true)`) are still processed even when the OS doesn't send redraw events.

**For #7776:** Three layers of throttling for invisible windows:
- **Heartbeat scheduling:** After painting an invisible window, schedule the next repaint 100ms in the future (instead of immediately). This keeps viewport commands flowing while limiting to ~10 repaints/sec.
- **Event throttling:** In `user_event`, throttle `RequestRepaint` events for invisible windows to at least 100ms delay, preventing egui's repaint callback from bypassing the heartbeat.
- **ControlFlow fix:** Only set `ControlFlow::Poll` for visible windows. Invisible windows use `WaitUntil` instead of spinning.
- **Backend sleep:** Add `is_visible() == Some(false)` alongside the existing `is_minimized()` sleep check in both wgpu and glow backends (defense in depth).

The fix is platform-agnostic: `is_visible()` returns `Some(false)` only when the platform can confirm invisibility, so it won't trigger on platforms where invisible windows still receive `RedrawRequested`.

## Test plan
- [x] `cargo fmt` passes
- [x] `cargo clippy -p eframe --all-features` passes with no warnings
- [x] Manual test on Windows: window reappears after `Visible(true)` when hidden
- [x] Manual test on Windows: CPU stays near 0% while window is invisible (was ~16% before fix)


crates/eframe/src/native/run.rs
```diff
@@ -177,23 +177,54 @@ impl<T: WinitApp> WinitAppWrapper<T> {
...
+        // See: https://github.com/emilk/egui/issues/7776
+        if !invisible_window_ids.is_empty() {
+            let next_paint = Instant::now() + std::time::Duration::from_millis(100);
+            for window_id in &invisible_window_ids {
+                self.windows_next_repaint_times
+                    .entry(*window_id)
+                    .and_modify(|t| *t = (*t).min(next_paint))
+                    .or_insert(next_paint);
+            }
+        }
```
emilk: This I don't follow. This will _always_ schedule a new "repaint" (actually: call to `App::logic`) for invisible windows. Why? If they have something to do, they should call `.request_repaint`, and if not - can't they just sleep?

crates/eframe/src/native/run.rs
```diff
@@ -270,6 +301,16 @@ impl<T: WinitApp> ApplicationHandler<UserEvent> for WinitAppWrapper<T> {
                         if let Some(window_id) =
                             self.winit_app.window_id_from_viewport_id(viewport_id)
                         {
+                            // Throttle repaints for invisible windows to prevent
+                            // high CPU usage on Windows.
+                            // See: https://github.com/emilk/egui/issues/7776
+                            let when = if let Some(window) = self.winit_app.window(window_id)
+                                && window.is_visible() == Some(false)
+                            {
+                                when.max(Instant::now() + std::time::Duration::from_millis(100))
```
emilk: This magic constant (`from_millis(100)`) would be nice to call out, e.g. by putting it in a `const` at the top of the file, or by adding it to `struct NativeOptions`

crates/eframe/src/native/run.rs
```diff
@@ -177,23 +177,54 @@ impl<T: WinitApp> WinitAppWrapper<T> {
...
+        // Schedule a throttled repaint for invisible windows so they keep
+        // processing viewport commands even when egui doesn't request repaints.
+        // See: https://github.com/emilk/egui/issues/7776
+        if !invisible_window_ids.is_empty() {
+            let next_paint = Instant::now() + std::time::Duration::from_millis(100);
+            for window_id in &invisible_window_ids {
+                self.windows_next_repaint_times
+                    .entry(*window_id)
+                    .and_modify(|t| *t = (*t).min(next_paint))
+                    .or_insert(next_paint);
```
emilk: with this simple change we get the throttling, but won't schedule work unless the user explicitly requests it

```suggestion
                    .and_modify(|t| *t = (*t).min(next_paint));
```

gcailly: Thank you very much ! I commited your suggestion and I tried to adress all of your comments.

crates/eframe/src/native/run.rs
gcailly: Ok, done. 🙂

emilk:
Related
* https://github.com/emilk/egui/issues/5112
* https://github.com/emilk/egui/issues/5136

Not sure how far this PR goes to fixing those issues 🤔

crates/eframe/src/native/run.rs
```diff
@@ -177,23 +185,54 @@ impl<T: WinitApp> WinitAppWrapper<T> {
...
-
                 if let Some(window) = self.winit_app.window(*window_id) {
-                    log::trace!("request_redraw for {window_id:?}");
-                    window.request_redraw();
+                    // On Windows, invisible windows don't receive RedrawRequested
+                    // events, so pending viewport commands (e.g. Visible(true)) would
+                    // never be processed. We collect these windows to paint them
+                    // directly below.
+                    // See: https://github.com/emilk/egui/issues/5229
+                    if window.is_visible() == Some(false) {
```
emilk: I wonder if we should do the same for minimized windows 🤔  It would mirror the checking of _both_ these flags in other places

```suggestion
                    if window.is_visible() == Some(false) || window.is_minimized() == Some(true) {
```

crates/eframe/src/native/run.rs
```diff
@@ -270,6 +309,16 @@ impl<T: WinitApp> ApplicationHandler<UserEvent> for WinitAppWrapper<T> {
                         if let Some(window_id) =
                             self.winit_app.window_id_from_viewport_id(viewport_id)
                         {
+                            // Throttle repaints for invisible windows to prevent
+                            // high CPU usage on Windows.
+                            // See: https://github.com/emilk/egui/issues/7776
+                            let when = if let Some(window) = self.winit_app.window(window_id)
+                                && window.is_visible() == Some(false)
```
emilk: and same here. Could add a helper `fn is_invisible_or_minimized(window)`

emilk:
Thanks!

PR #7909, opened 2026-02-14
Fix hover tooltip for layout containers

optozorax:
I have found that `ui.horizontal(...).response.on_hover_text(...)` doesn't work, although [documentation says otherwise](https://docs.rs/egui/latest/egui/struct.Ui.html#method.horizontal):

> The returned [Response](https://docs.rs/egui/latest/egui/response/struct.Response.html) will only have checked for mouse hover but can be used for tooltips (on_hover_text). It also contains the [Rect](https://docs.rs/egui/latest/egui/struct.Rect.html) used by the horizontal layout.The returned [Response](https://docs.rs/egui/latest/egui/response/struct.Response.html) will only have checked for mouse hover but can be used for tooltips (on_hover_text). It also contains the [Rect](https://docs.rs/egui/latest/egui/struct.Rect.html) used by the horizontal layout.

So I added fix to this into tooltip, but this looks like a hack, and I don't know a better place to do so. Do you have a better idea?

This PR is split into two commits: first commit adds test, and you can see wrong behaviour in the image. Second commit fixes this and also regenerates the image and you can now see tooltip.

I have used LLM to help me with this PR (codex). And I retained LLM comment, because such hacks should have some comments, and it seems ok to me.

### DIFF
--- a/tests/egui_tests/tests/snapshots/horizontal_response_topmost_tooltip_should_win.png
+++ b/tests/egui_tests/tests/snapshots/horizontal_response_topmost_tooltip_should_win.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:567f73f65ad6b3e2ed7353fb5e1d2a3033531e2c98a1bc87e41ac49ca6c19030
+oid sha256:f3d66b84f9719133dee48c4bf29028132a5944c1137ac72c246128d127cc8216
 size 2741
--- a/tests/egui_tests/tests/snapshots/horizontal_response_tooltip_should_show_when_hovering_child.png
+++ b/tests/egui_tests/tests/snapshots/horizontal_response_tooltip_should_show_when_hovering_child.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:bf76b3e53761a7669b16c22a92c9062c452a1d600adbf0faeff57e42b31abffa
-size 2763
+oid sha256:57b403497026d303512ee6a9cb7408ec97917cb6200aa02216b14c4b9d8fbfa6
+size 2764


optozorax:
Hi, @emilk , can you please take a look at this PR?

tests/egui_tests/tests/regression_tests.rs
```diff
@@ -76,6 +76,25 @@ fn combobox_should_have_value() {
...
+            _ = ui.button("B");
+        })
+        .response
+        .on_hover_text("test");
+    });
+
+    harness.get_by_label("A").hover();
+    harness.run_steps(5);
+    harness.snapshot("horizontal_response_tooltip_should_show_when_hovering_child");
+}
```
emilk:
```suggestion
            _ = ui.button("B").on_hover_test("B has a tooltip");
        })
        .response
        .on_hover_text("test");
    });

    harness.get_by_label("A").hover();
    harness.run_steps(5);
    harness.snapshot("horizontal_response_tooltip_should_show_when_hovering_child");

    harness.get_by_label("B").hover();
    harness.run_steps(5);
    harness.snapshot("horizontal_response_topmost_tooltip_should_win");
}
```

emilk: We should probably test nested ui:s as well while we're at it, where some has tooltips and others down't

optozorax:
@emilk done!

tests/egui_tests/tests/snapshots/horizontal_response_topmost_tooltip_should_win.png
emilk, line 1: Hum - I wanted to test that the tooltip for `B` pops up, but in this case I think we just hit the logic of "keep the previous tooltip open"? I think this needs to be a separate test then

PR #7914, opened 2026-02-16
Add IME language selection to InputOptions for CJK support

rustbasic:
Add IME language selection to InputOptions for CJK support

**Overview**
This PR adds a new configuration option to select specific IME (Input Method Editor) processing modes, particularly for CJK (Chinese, Japanese, Korean) languages. This allows for more specialized handling of character composition depending on the selected language.

**Changes**
- Introduced `ImeLanguage` enum (None, Korean, Japanese, Chinese) with `serde` support.
- Added `ime_language` field to `InputOptions` struct.
- Updated `InputOptions::ui` to include a `ComboBox` for easy language selection in the settings panel.

**Why this is needed**
Proper CJK support often requires different composition logic depending on the language. By allowing users to explicitly select their IME mode, we can provide a more robust input experience for international users.

* Relate #7898

In addition to #7898, a bug in Korean input on mobile (Android) has been fixed.

### DIFF
--- a/crates/eframe/src/web/text_agent.rs
+++ b/crates/eframe/src/web/text_agent.rs
@@ -87,12 +87,16 @@ impl TextAgent {
         };
 
         let on_composition_update = {
+            let input = input.clone();
             move |event: web_sys::CompositionEvent, runner: &mut AppRunner| {
                 let Some(text) = event.data() else { return };
+
+                let start = input.selection_start().ok().flatten().unwrap_or(0) as usize;
+                let end = input.selection_start().ok().flatten().unwrap_or(0) as usize;
                 let event = egui::Event::Ime(egui::ImeEvent::Preedit {
                     text_mark: text,
-                    start: 0,
-                    end: 0,
+                    start,
+                    end,
                 });
                 runner.input.raw.events.push(event);
                 runner.needs_repaint.repaint_asap();
--- a/crates/eframe/src/web/text_agent.rs
+++ b/crates/eframe/src/web/text_agent.rs
@@ -89,7 +89,11 @@ impl TextAgent {
         let on_composition_update = {
             move |event: web_sys::CompositionEvent, runner: &mut AppRunner| {
                 let Some(text) = event.data() else { return };
-                let event = egui::Event::Ime(egui::ImeEvent::Preedit(text));
+                let event = egui::Event::Ime(egui::ImeEvent::Preedit {
+                    text_mark: text,
+                    start: 0,
+                    end: 0,
+                });
                 runner.input.raw.events.push(event);
                 runner.needs_repaint.repaint_asap();
             }
--- a/crates/eframe/src/web/text_agent.rs
+++ b/crates/eframe/src/web/text_agent.rs
@@ -92,7 +92,7 @@ impl TextAgent {
                 let Some(text) = event.data() else { return };
 
                 let start = input.selection_start().ok().flatten().unwrap_or(0) as usize;
-                let end = input.selection_start().ok().flatten().unwrap_or(0) as usize;
+                let end = input.selection_end().ok().flatten().unwrap_or(0) as usize;
                 let event = egui::Event::Ime(egui::ImeEvent::Preedit {
                     text_mark: text,
                     start,
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -594,11 +594,15 @@ impl State {
                     self.ime_event_enable();
                 }
             }
-            winit::event::Ime::Preedit(text, Some(_cursor)) => {
+            winit::event::Ime::Preedit(text, Some(cursor)) => {
                 self.ime_event_enable();
                 self.egui_input
                     .events
-                    .push(egui::Event::Ime(egui::ImeEvent::Preedit(text.clone())));
+                    .push(egui::Event::Ime(egui::ImeEvent::Preedit {
+                        text_mark: text.clone(),
+                        start: cursor.0,
+                        end: cursor.1,
+                    }));
             }
             winit::event::Ime::Commit(text) => {
                 self.egui_input
@@ -618,7 +622,11 @@ impl State {
                 // TextEdit in such situation.
                 self.egui_input
                     .events
-                    .push(egui::Event::Ime(egui::ImeEvent::Preedit(String::new())));
+                    .push(egui::Event::Ime(egui::ImeEvent::Preedit {
+                        text_mark: String::new(),
+                        start: 0,
+                        end: 0,
+                    }));
             }
         }
     }
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -619,7 +619,6 @@ impl State {
                 self.egui_input
                     .events
                     .push(egui::Event::Ime(egui::ImeEvent::Preedit(String::new())));
-                self.ime_event_disable();
             }
         }
     }
--- a/crates/egui/src/data/input.rs
+++ b/crates/egui/src/data/input.rs
@@ -571,7 +571,11 @@ pub enum ImeEvent {
     Enabled,
 
     /// A new IME candidate is being suggested.
-    Preedit(String),
+    Preedit {
+        text_mark: String,
+        start: usize,
+        end: usize,
+    },
 
     /// IME composition ended with this final result.
     Commit(String),
--- a/crates/egui/src/input_state/mod.rs
+++ b/crates/egui/src/input_state/mod.rs
@@ -53,10 +53,23 @@ impl SurrenderFocusOn {
     }
 }
 
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub enum ImeLanguage {
+    None,
+    #[default]
+    Korean,
+    Japanese,
+    Chinese,
+}
+
 /// Options for input state handling.
 #[derive(Clone, Copy, Debug, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub struct InputOptions {
+    /// The language used for specialized IME (Input Method Editor) processing.
+    pub ime_language: ImeLanguage,
+
     /// Multiplier for the scroll speed when reported in [`crate::MouseWheelUnit::Line`]s.
     pub line_scroll_speed: f32,
 
@@ -110,6 +123,7 @@ impl Default for InputOptions {
         };
 
         Self {
+            ime_language: ImeLanguage::default(),
             line_scroll_speed,
             scroll_zoom_speed: 1.0 / 200.0,
             max_click_dist: 6.0,
@@ -127,6 +141,7 @@ impl InputOptions {
     /// Show the options in the ui.
     pub fn ui(&mut self, ui: &mut crate::Ui) {
         let Self {
+            ime_language,
             line_scroll_speed,
             scroll_zoom_speed,
             max_click_dist,
@@ -141,6 +156,23 @@ impl InputOptions {
             .num_columns(2)
             .striped(true)
             .show(ui, |ui| {
+                ui.label("IME Language");
+                let inner_response = crate::ComboBox::from_id_salt("ime_language_combo")
+                    .selected_text(match ime_language {
+                        ImeLanguage::None => "None",
+                        ImeLanguage::Korean => "Korean",
+                        ImeLanguage::Japanese => "Japanese",
+                        ImeLanguage::Chinese => "Chinese",
+                    })
+                    .show_ui(ui, |ui| {
+                        ui.selectable_value(ime_language, ImeLanguage::None, "None");
+                        ui.selectable_value(ime_language, ImeLanguage::Korean, "Korean");
+                        ui.selectable_value(ime_language, ImeLanguage::Japanese, "Japanese");
+                        ui.selectable_value(ime_language, ImeLanguage::Chinese, "Chinese");
+                    });
+                inner_response.response.on_hover_text("Select the language for specialized IME processing (CJK)");
+                ui.end_row();
+
                 ui.label("Line scroll speed");
                 ui.add(crate::DragValue::new(line_scroll_speed).range(0.0..=f32::INFINITY))
                     .on_hover_text(
@@ -199,7 +231,6 @@ impl InputOptions {
                 ui.label("surrender_focus_on");
                 surrender_focus_on.ui(ui);
                 ui.end_row();
-
             });
     }
 }
--- a/crates/egui/src/input_state/mod.rs
+++ b/crates/egui/src/input_state/mod.rs
@@ -56,8 +56,8 @@ impl SurrenderFocusOn {
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub enum ImeLanguage {
-    None,
     #[default]
+    None,
     Korean,
     Japanese,
     Chinese,
--- a/crates/egui/src/input_state/mod.rs
+++ b/crates/egui/src/input_state/mod.rs
@@ -361,7 +361,7 @@ pub struct InputState {
     /// Input state management configuration.
     ///
     /// This gets copied from `egui::Options` at the start of each frame for convenience.
-    options: InputOptions,
+    pub options: InputOptions,
 }
 
 impl Default for InputState {
--- a/crates/egui/src/input_state/mod.rs
+++ b/crates/egui/src/input_state/mod.rs
@@ -53,7 +53,7 @@ impl SurrenderFocusOn {
     }
 }
 
-#[derive(Clone, Copy, Debug, Default, PartialEq)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub enum ImeLanguage {
     None,
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -1135,7 +1135,11 @@ fn on_ime_korean(
             state.ime_cursor_range = *cursor_range;
             None
         }
-        ImeEvent::Preedit(text_mark) => {
+        ImeEvent::Preedit {
+            text_mark,
+            start: _,
+            end: _,
+        } => {
             if text_mark == "\n" || text_mark == "\r" {
                 None
             } else {
@@ -1196,7 +1200,11 @@ fn on_ime_japanese(
             state.ime_cursor_range = *cursor_range;
             None
         }
-        ImeEvent::Preedit(text_mark) => {
+        ImeEvent::Preedit {
+            text_mark,
+            start: _,
+            end: _,
+        } => {
             if text_mark == "\n" || text_mark == "\r" {
                 None
             } else {
@@ -1252,7 +1260,11 @@ fn on_ime_chinese(
             state.ime_cursor_range = *cursor_range;
             None
         }
-        ImeEvent::Preedit(text_mark) => {
+        ImeEvent::Preedit {
+            text_mark,
+            start: _,
+            end: _,
+        } => {
             if text_mark == "\n" || text_mark == "\r" {
                 None
             } else {
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -806,16 +806,6 @@ impl TextEdit<'_> {
             }
         }
 
-        // Ensures correct IME behavior when the text input area gains or loses focus.
-        if state.ime_enabled && (response.gained_focus() || response.lost_focus()) {
-            state.ime_enabled = false;
-            if let Some(mut ccursor_range) = state.cursor.char_range() {
-                ccursor_range.secondary.index = ccursor_range.primary.index;
-                state.cursor.set_char_range(Some(ccursor_range));
-            }
-            ui.input_mut(|i| i.events.retain(|e| !matches!(e, Event::Ime(_))));
-        }
-
         state.clone().store(ui.ctx(), id);
 
         if response.changed() {
@@ -1066,70 +1056,15 @@ fn events(
             } => check_for_mutating_key_press(os, &cursor_range, text, galley, modifiers, *key),
 
             Event::Ime(ime_event) => {
-                /// Empty prediction can be produced with [`ImeEvent::Preedit`]
-                /// or [`ImeEvent::Commit`] when user press backspace or escape
-                /// during IME, so this function should be called in both cases
-                /// to clear current text.
-                ///
-                /// Example platforms where only `ImeEvent::Preedit("")` of
-                /// those two events is emitted when the last character in the
-                /// prediction is deleted:
-                /// - macOS 15.7.3.
-                /// - Debian13 with gnome48 and wayland.
-                ///
-                /// An example platform where only `ImeEvent::Commit("")` of
-                /// those two events is emitted when the last character in the
-                /// prediction is deleted:
-                /// - Safari 26.2 (on macOS 15.7.3).
-                fn clear_prediction(
-                    text: &mut dyn TextBuffer,
-                    cursor_range: &CCursorRange,
-                ) -> CCursor {
-                    text.delete_selected(cursor_range)
-                }
-
-                match ime_event {
-                    ImeEvent::Enabled => {
-                        state.ime_enabled = true;
-                        state.ime_cursor_range = cursor_range;
-                        None
-                    }
-                    ImeEvent::Preedit(text_mark) => {
-                        if text_mark == "\n" || text_mark == "\r" {
-                            None
-                        } else {
-                            let mut ccursor = clear_prediction(text, &cursor_range);
-
-                            let start_cursor = ccursor;
-                            if !text_mark.is_empty() {
-                                text.insert_text_at(&mut ccursor, text_mark, char_limit);
-                            }
-                            state.ime_cursor_range = cursor_range;
-                            Some(CCursorRange::two(start_cursor, ccursor))
-                        }
-                    }
-                    ImeEvent::Commit(prediction) => {
-                        if prediction == "\n" || prediction == "\r" {
-                            None
-                        } else {
-                            state.ime_enabled = false;
-
-                            let mut ccursor = clear_prediction(text, &cursor_range);
-
-                            if !prediction.is_empty()
-                                && cursor_range.secondary.index
-                                    == state.ime_cursor_range.secondary.index
-                            {
-                                text.insert_text_at(&mut ccursor, prediction, char_limit);
-                            }
-
-                            Some(CCursorRange::one(ccursor))
-                        }
-                    }
-                    ImeEvent::Disabled => {
-                        state.ime_enabled = false;
-                        None
-                    }
+                let ime_language = ui.input(|i| i.options.ime_language);
+                if ime_language == crate::input_state::ImeLanguage::Korean {
+                    on_ime_k(ime_event, state, text, &cursor_range, char_limit)
+                } else if ime_language == crate::input_state::ImeLanguage::Japanese {
+                    on_ime_j(ime_event, state, text, &cursor_range, char_limit)
+                } else if ime_language == crate::input_state::ImeLanguage::Chinese {
+                    on_ime_c(ime_event, state, text, &cursor_range, char_limit)
+                } else {
+                    on_ime_k(ime_event, state, text, &cursor_range, char_limit)
                 }
             }
 
@@ -1155,6 +1090,193 @@ fn events(
     );
 
     (any_change, cursor_range)
+}
+
+// ----------------------------------------------------------------------------
+
+/// Empty prediction can be produced with [`ImeEvent::Preedit`]
+/// or [`ImeEvent::Commit`] when user press backspace or escape
+/// during IME, so this function should be called in both cases
+/// to clear current text.
+///
+/// Example platforms where only `ImeEvent::Preedit("")` of
+/// those two events is emitted when the last character in the
+/// prediction is deleted:
+/// - macOS 15.7.3.
+/// - Debian13 with gnome48 and wayland.
+///
+/// An example platform where only `ImeEvent::Commit("")` of
+/// those two events is emitted when the last character in the
+/// prediction is deleted:
+/// - Safari 26.2 (on macOS 15.7.3).
+fn clear_prediction(text: &mut dyn TextBuffer, cursor_range: &CCursorRange) -> CCursor {
+    text.delete_selected(cursor_range)
+}
+
+// ----------------------------------------------------------------------------
+
+// Handles IME input events for Korean character composition.
+fn on_ime_k(
+    ime_event: &ImeEvent,
+    state: &mut TextEditState,
+    text: &mut dyn TextBuffer,
+    cursor_range: &CCursorRange,
+    char_limit: usize,
+) -> Option<CCursorRange> {
+    match ime_event {
+        ImeEvent::Enabled => {
+            state.ime_enabled = true;
+            state.ime_cursor_range = *cursor_range;
+            None
+        }
+        ImeEvent::Preedit(text_mark) => {
+            if text_mark == "\n" || text_mark == "\r" {
+                None
+            } else {
+                let mut ccursor = clear_prediction(text, &cursor_range);
+                let start_cursor = ccursor;
+
+                if !text_mark.is_empty() {
+                    text.insert_text_at(&mut ccursor, text_mark, char_limit);
+                    state.ime_cursor_range = *cursor_range;
+                }
+                Some(CCursorRange::two(start_cursor, ccursor))
+            }
+        }
+        ImeEvent::Commit(prediction) => {
+            if prediction == "\n" || prediction == "\r" {
+                None
+            } else {
+                let mut ccursor = cursor_range.secondary;
+
+                if state.ime_enabled
+                    && !prediction.is_empty()
+                    && cursor_range.secondary.index == state.ime_cursor_range.secondary.index
+                {
+                    ccursor = clear_prediction(text, &cursor_range);
+                    text.insert_text_at(&mut ccursor, prediction, char_limit);
+                }
+
+                Some(CCursorRange::one(ccursor))
+            }
+        }
+        ImeEvent::Disabled => {
+            state.ime_enabled = false;
+            None
+        }
+    }
+}
+
+// ----------------------------------------------------------------------------
+
+// Handles IME input events for Japanese character composition.
+fn on_ime_j(
+    ime_event: &ImeEvent,
+    state: &mut TextEditState,
+    text: &mut dyn TextBuffer,
+    cursor_range: &CCursorRange,
+    char_limit: usize,
+) -> Option<CCursorRange> {
+    match ime_event {
+        ImeEvent::Enabled => {
+            state.ime_enabled = true;
+            state.ime_cursor_range = *cursor_range;
+            None
+        }
+        ImeEvent::Preedit(text_mark) => {
+            if text_mark == "\n" || text_mark == "\r" {
+                None
+            } else {
+                let mut ccursor = clear_prediction(text, &cursor_range);
+
+                let start_cursor = ccursor;
+                if text_mark.is_empty() {
+                    state.ime_enabled = false;
+                } else {
+                    text.insert_text_at(&mut ccursor, text_mark, char_limit);
+                }
+                state.ime_cursor_range = *cursor_range;
+                Some(CCursorRange::two(start_cursor, ccursor))
+            }
+        }
+        ImeEvent::Commit(prediction) => {
+            if prediction == "\n" || prediction == "\r" {
+                None
+            } else {
+                state.ime_enabled = false;
+
+                let mut ccursor = clear_prediction(text, &cursor_range);
+
+                if !prediction.is_empty()
+                    && cursor_range.secondary.index == state.ime_cursor_range.secondary.index
+                {
+                    text.insert_text_at(&mut ccursor, prediction, char_limit);
+                }
+
+                Some(CCursorRange::one(ccursor))
+            }
+        }
+        ImeEvent::Disabled => {
+            state.ime_enabled = false;
+            None
+        }
+    }
+}
+
+// ----------------------------------------------------------------------------
+
+// Handles IME input events for Chinese character composition.
+fn on_ime_c(
+    ime_event: &ImeEvent,
+    state: &mut TextEditState,
+    text: &mut dyn TextBuffer,
+    cursor_range: &CCursorRange,
+    char_limit: usize,
+) -> Option<CCursorRange> {
+    match ime_event {
+        ImeEvent::Enabled => {
+            state.ime_enabled = true;
+            state.ime_cursor_range = *cursor_range;
+            None
+        }
+        ImeEvent::Preedit(text_mark) => {
+            if text_mark == "\n" || text_mark == "\r" {
+                None
+            } else {
+                let mut ccursor = clear_prediction(text, &cursor_range);
+
+                let start_cursor = ccursor;
+                if text_mark.is_empty() {
+                    state.ime_enabled = false;
+                } else {
+                    text.insert_text_at(&mut ccursor, text_mark, char_limit);
+                }
+                state.ime_cursor_range = *cursor_range;
+                Some(CCursorRange::two(start_cursor, ccursor))
+            }
+        }
+        ImeEvent::Commit(prediction) => {
+            if prediction == "\n" || prediction == "\r" {
+                None
+            } else {
+                state.ime_enabled = false;
+
+                let mut ccursor = clear_prediction(text, &cursor_range);
+
+                if !prediction.is_empty()
+                    && cursor_range.secondary.index == state.ime_cursor_range.secondary.index
+                {
+                    text.insert_text_at(&mut ccursor, prediction, char_limit);
+                }
+
+                Some(CCursorRange::one(ccursor))
+            }
+        }
+        ImeEvent::Disabled => {
+            state.ime_enabled = false;
+            None
+        }
+    }
 }
 
 // ----------------------------------------------------------------------------
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -812,15 +812,14 @@ impl TextEdit<'_> {
                 state.ime_enabled = false;
             }
 
-            if let Some(mut ccursor_range) = state.cursor.char_range() {
-                if !state.ime_enabled
-                    || ccursor_range.secondary.index != state.ime_cursor_range.secondary.index
-                {
-                    state.ime_enabled = false;
-                    ccursor_range.secondary.index = ccursor_range.primary.index;
-                    state.cursor.set_char_range(Some(ccursor_range));
-                    ui.input_mut(|i| i.events.retain(|e| !matches!(e, Event::Ime(_))));
-                }
+            if let Some(mut ccursor_range) = state.cursor.char_range()
+                && (!state.ime_enabled
+                    || ccursor_range.secondary.index != state.ime_cursor_range.secondary.index)
+            {
+                state.ime_enabled = false;
+                ccursor_range.secondary.index = ccursor_range.primary.index;
+                state.cursor.set_char_range(Some(ccursor_range));
+                ui.input_mut(|i| i.events.retain(|e| !matches!(e, Event::Ime(_))));
             }
         }
 
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -807,20 +807,9 @@ impl TextEdit<'_> {
         }
 
         // Ensures correct IME behavior when the text input area gains or loses focus.
-        if state.ime_enabled {
-            if response.gained_focus() {
-                state.ime_enabled = false;
-            }
-
-            if let Some(mut ccursor_range) = state.cursor.char_range()
-                && (!state.ime_enabled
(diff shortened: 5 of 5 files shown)


umajho:
Hello.

In [my previous comment](https://github.com/emilk/egui/pull/7898#issuecomment-3886843136), I mentioned that I suspected the root cause of the IME behavior differences on Windows lies in `winit`. About a week ago, I was finally able to debug on Windows and confirmed that this is indeed the case.
Since then, I've submitted another PR (#7967) that avoids the need to distinguish between the so called “IME languages”.

In addition, that PR addresses several issues not covered here. One of them involves a Korean IME case (Hanja confirmation) where simply setting the IME language to Korean does not help, as its behavior more closely resembles that of typical Japanese IMEs.
This further suggests that differentiating by “IME language” is the wrong abstraction: there can be multiple distinct IME behaviors even within the same language.

I would appreciate your feedback on my PR (#7967). If you have time to review it, I’d be grateful for any comments or suggestions for improvement (similar to the concerns I raised in my previous comment).

Separately, you mentioned that this PR also fixes another IME-related issue. Could you provide more details about that? From the changes, I can tell it relates to a Korean IME using the Cheonjiin layout on Android (web?), but I’m unclear on what the exact issue was and which specific IME is affected. As I understand it, multiple IMEs support this layout.

Thank you.

rustbasic:
**Regarding why language-specific IME selection was previously unnecessary on certain OSs:**
 
  Until now, the reason we haven't needed to manually select IMEs for each language on specific operating systems is that users typically install an OS tailored to a specific language. During this process, the language-specific IME is installed either automatically or manually, allowing users to use it without much concern.

  However, since **egui** is a cross-platform framework that operates across all platforms, it must implement its own IME handling logic internally. Given the unique characteristics of CJK (Chinese, Japanese, and Korean) languages and the existence of various IMEs within each language, it is extremely difficult to handle everything with a single, unified function.

  To achieve a "one-function-fits-all" approach, we must first complete the handling logic for every language and IME, and then implement a system to automatically recognize them. We cannot skip these essential intermediate steps and jump straight to a perfect, universal solution.

  It would be ideal if a library like **winit** handled everything, but we can't just wait around for that.

umajho:
> **Regarding why language-specific IME selection was previously unnecessary on certain OSs:**
> 
> Until now, the reason we haven't needed to manually select IMEs for each language on specific operating systems is that users typically install an OS tailored to a specific language. During this process, the language-specific IME is installed either automatically or manually, allowing users to use it without much concern.
> 
> However, since **egui** is a cross-platform framework that operates across all platforms, it must implement its own IME handling logic internally. Given the unique characteristics of CJK (Chinese, Japanese, and Korean) languages and the existence of various IMEs within each language, it is extremely difficult to handle everything with a single, unified function.
> 
> To achieve a "one-function-fits-all" approach, we must first complete the handling logic for every language and IME, and then implement a system to automatically recognize them. We cannot skip these essential intermediate steps and jump straight to a perfect, universal solution.
> 
> It would be ideal if a library like **winit** handled everything, but we can't just wait around for that.

---

> Until now, the reason we haven't needed to manually select IMEs for each language on specific operating systems is that users typically install an OS tailored to a specific language. During this process, the language-specific IME is installed either automatically or manually, allowing users to use it without much concern.

I don't quite understand this. what exactly is the “concern” being referred to?

Are you suggesting that users typically use IMEs matching their system language? If so, that distinction isn't particularly meaningful. Whether the system language matches the IME language doesn't affect how IMEs function.

Many users, myself included, run their systems in English while using IMEs for other languages. If an OS couldn't support multiple IMEs simultaneously, some would fail under an English system, which would be considered a bug. A mature platform must ensure that IMEs work reliably regardless of the system language. Fortunately, `egui` targets only mature platforms (Linux, macOS, Windows, and Web), where this is generally expected.

---

> However, since **egui** is a cross-platform framework that operates across all platforms, it must implement its own IME handling logic internally. Given the unique characteristics of CJK (Chinese, Japanese, and Korean) languages and the existence of various IMEs within each language, it is extremely difficult to handle everything with a single, unified function.

If handling IMEs for multiple languages isn't inherently problematic on a single platform, it's not clear why expanding to multiple platforms would introduce language-specific issues. The real differences lie between platforms, which can be handled using mechanisms like `cfg(target_os = …)`.

---

> To achieve a "one-function-fits-all" approach, we must first complete the handling logic for every language and IME, and then implement a system to automatically recognize them. We cannot skip these essential intermediate steps and jump straight to a perfect, universal solution.

The reason we should adopt a “universal solution”, regardless of the language or IME a user is using, is precisely because we CAN'T and SHOULD'T attempt to handle every language and IME individually.

1. Language detection is effectively impossible.
   - As an IME user, my system is in English. There is no reliable way to determine which language my IME is using based on that alone.
   - Some Chinese IMEs report Chinese characters as `Preedit` text. In such cases, it may be impossible to distinguish whether the user is composing Chinese or Japanese.
   - Even if we could detect which IME is being used (Spoiler: we generally can't), how would we determine its target language? An IME with “Hangul” in its name is clearly Korean, but what about something ambiguously named like “Foobar”? Maintaining a mapping of IMEs to languages (which we shouldn't do) would still fail for newly introduced IMEs.

2. Even if language detection were possible, it would still be unreliable. Knowing the language of an IME does not provide sufficient information to implement correct behavior.
   - Different IMEs for the same language can behave differently. For example, even with the same layout (Cheonjiin) on the same platform (Android), we've observed three distinct behaviors across three IMEs (Gboard, SwiftKey, and another IME with a bug addressed in this PR).
   - As a result, a fix targeting one IME could break another, even within the same language.

3. IME detection is both undesirable and often infeasible.
   - The action of probing which IME a user is using might concerns some users. They are personal data.
   - It is just not feasible at all in web environments or other sandboxed contexts.

4. Even if IME detection were possible, `egui` should not be responsible for workarounds that are specific to particular IMEs. Once we introduce a workaround for one IME, we effectively open the door to working around all IMEs individually.
   - For reference, here are non-exhaustive lists of innumerable Chinese and Japanese IMEs:
     - <https://translate.kagi.com/zh.wikipedia.org/wiki/中文输入法列表>
     - <https://translate.kagi.com/ja.wikipedia.org/wiki/日本語入力システムの一覧>

Therefore, in `egui`, IME handling should be based solely on the general interface (i.e., a “universal solution”), rather than targeting specific IMEs or languages.

The current IME handling logic already follows a “one-function-fits-all” approach. The “language selection” proposed in this PR breaks it, yet can't be pushed further: because, as explained above, handling every language and IME individually is infeasible. In my opinion, it only moves the codebase away from the “one-function-fits-all” goal rather than toward it.

---

> It would be ideal if a library like **winit** handled everything, but we can't just wait around for that.

If there are bugs upstream, the proper approach is to fix them there. Temporary workarounds in `egui` may be necessary, but they shouldn't be treated as permanent solutions.

The main reason I haven't contributed fixes to `winit` is that `egui` currently depends on `winit` 0.30, while the latest version is `0.31` (beta). The changes between these versions are substantial, and backporting fixes would be quite cumbersome. It seems more practical to wait until `egui` upgrades to `winit` 0.31.

umajho:
on macOS Safari, the builtin Chinese IME (Shuangpin) is broken by default with this PR:
2026-03-21 11 40 21 PM

on macOS Safari, the builtin Japanese IME is broken is broken by default with this PR:
2026-03-21 11 41 39 PM

on Windows, the builtin Chinese IME (Shuangpin) is broken by default with this PR:
2026-03-21 11 54 59 PM

Since these IMEs previously worked out of the box prior to this PR, they should also work out of the box after this PR as well.

As I have pointed out [here](https://github.com/emilk/egui/pull/7898#issuecomment-3886843136), most developers will not bother integrating this language selection thing, and most affected users are unlikely to discover it even if it is integrated in the app they are using.
Even if we assume most developers do integrate it and most users discover it, the existence of “any developer or user who does not” means this remains a regression for those affected users: It worked prior to this PR, and no longer works after it.

---

Note that I am not implying any prioritization of languages. The point is simply that the current approach proposed by this PR is not that feasible.

umajho:
I'm not home and can't verify this, so this is merely a baseless speculation:

By any chance the IME you are addressing here handles the batchim situation by sending a backspace event to delete the first character. And the reason it is broken is because `eframe/web` incorrectly suppresses that backspace event in such cases?

rustbasic:
@umajho 

Since I rely on translators or AI translation, it can be difficult to convey precise nuances.

The "Cheonjiin keyboard" does not trigger a backspace event; instead, it returns `Preedit("", 1, 1)` to cancel the previously completed character. Even within the Korean language alone, there are inconsistencies across Windows, WASM, Android, and the "Cheonjiin keyboard" layout.

The primary issue currently lies in the conflicts between CJK languages, so I would like to keep the Korean logic separate for the time being. These IME bugs have remained unresolved for years, and since I slightly mitigated them a year ago, they have stayed in the same state. Fixing one side often leads to issues on the other. It seems wiser to address the bugs by separating CJK first and then determine if they can be merged later.

I may be speaking from my own limitations, as I can only test Korean on Windows, WASM, and Android. Perhaps `umajho` can handle this, as they might be able to test most platforms and languages. Applying #7967 + #7983 (while maintaining #4137) shows a lot of improvement, so I see a lot of potential there.

Let's make this happen!

umajho:
> Since I rely on translators or AI translation, it can be difficult to convey precise nuances.

I can definitely relate. I'm not very confident in my English writing either.
Whenever I finish writing comments, I usually run them through an LLM for polishing. Then I selectively adopt revisions that feel more readable while still preserving my original intent. I end up spending quite a bit of time writing and editing comments, just trying to reduce the loss of nuance during this process.

> …
>
> Let's make this happen!

I’ll do my best to fix this Cheonjiin bug while trying not to introduce IME-specific logic.

> The "Cheonjiin keyboard" does not trigger a backspace event; instead, it returns `Preedit("", 1, 1)` to cancel the previously completed character. Even within the Korean language alone, there are inconsistencies across Windows, WASM, Android, and the "Cheonjiin keyboard" layout.

It would be very helpful if you could reproduce the bug while logging the following 6 events (The more informative, the better):
https://github.com/emilk/egui/blob/49fad9a7b2b8f99f0081b105e7fd4f5851238ac5/crates/eframe/src/web/text_agent.rs#L109-L117

Once you have that, please paste the logs here.

To access the console on Android Chrome, you can use:
https://developer.chrome.com/docs/devtools/remote-debugging

Thanks.

rustbasic:
@umajho 

**For reference, the `on_ime_korean()` function in #7914 has resolved all Korean input issues I am aware of across Windows, WASM, and the "Cheonjiin keyboard"(Android) layout.**

** Below are the logs generated while typing the word "않았다" **

```

ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Meta', code='MetaLeft', is_composing=false, key_code=91
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=false, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:88: on_composition_start: data='Some("")'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='ㅇ', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='ㅇ'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='이', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='이'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='아', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='아'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='안', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='안'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='안', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='안'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:124: on_composition_end: data='안'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=false
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=false, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:88: on_composition_start: data='Some("")'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='ㅅ', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='ㅅ'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='ㅅ', selection_start=Some(1)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='않', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='않'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='않', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='않'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:124: on_composition_end: data='않'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=false
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=false, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:88: on_composition_start: data='Some("")'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='ㅇ', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='ㅇ'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='이', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='이'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='아', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='아'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='앗', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='앗'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='앟', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='앟'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='았', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='았'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='았', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='았'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:124: on_composition_end: data='았'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=false
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=false, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:88: on_composition_start: data='Some("")'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='ㄷ', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='ㄷ'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='디', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='디'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='다', selection_start=Some(0)
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='다'
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-9c02abe29a2f63ed.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:124: on_composition_end: data='다'
```

rustbasic:
@umajho 


**I noticed that I am mistakenly assigning `input.selection_start()` to the `end` value. However, since both `start` and `end` are 1 at that specific point, there is no functional difference in the current behavior.**

**Below are the logs generated while typing "않았다":**


```
[eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Enter', code='', is_composing=false
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=false, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:88: on_composition_start: data='Some("")'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='ㅇ', selection_start=Some(0), selection_end=Some(0)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='ㅇ'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='이', selection_start=Some(0), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='이'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='아', selection_start=Some(0), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='아'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='안', selection_start=Some(0), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='안'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='안', selection_start=Some(0), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='안'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:125: on_composition_end: data='안'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=false
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=false, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:88: on_composition_start: data='Some("")'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='ㅅ', selection_start=Some(0), selection_end=Some(0)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='ㅅ'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='ㅅ', selection_start=Some(1), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='않', selection_start=Some(0), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='않'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='않', selection_start=Some(0), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='않'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:125: on_composition_end: data='않'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=false
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=false, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:88: on_composition_start: data='Some("")'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='ㅇ', selection_start=Some(0), selection_end=Some(0)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='ㅇ'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='이', selection_start=Some(0), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='이'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='아', selection_start=Some(0), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='아'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='앗', selection_start=Some(0), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='앗'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='앟', selection_start=Some(0), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='앟'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='았', selection_start=Some(0), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='았'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='았', selection_start=Some(0), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='았'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:125: on_composition_end: data='았'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=false
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=false, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:88: on_composition_start: data='Some("")'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='ㄷ', selection_start=Some(0), selection_end=Some(0)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='ㄷ'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='디', selection_start=Some(0), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='디'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:168: on_keydown: key='Unidentified', code='', is_composing=true, key_code=229
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:103: on_composition_update: data='다', selection_start=Some(0), selection_end=Some(1)
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:61: on_input: input_type="insertCompositionText", is_composing=true, value='다'
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::events] C:\prog\egui\crates\eframe\src\web\events.rs:276: on_keyup: key='Unidentified', code='', is_composing=true
ezchat-ca28cce8b9f513fe.js:1038 [eframe::web::text_agent] C:\prog\egui\crates\eframe\src\web\text_agent.rs:125: on_composition_end: data='다'
```

umajho:
```
on_composition_update: data='ㅇ', selection_start=Some(0), selection_end=Some(0)
```

It appears that Chrome dispatches `CompositionEvent`s before updating the selection state.

---

```
on_composition_update: data='ㅅ', selection_start=Some(0), selection_end=Some(0)
…
on_composition_update: data='ㅅ', selection_start=Some(1), selection_end=Some(1)
```

My interpretation is that Chrome does not expect the text agent's content to be cleared while processing Cheonjiin. To avoid introducing negative values in ranges, it may attempt to recover the text state. (e.g. recovering `[]ㅅ` to `안[]ㅅ`, where `[]` denotes a selection range, which is outdated here.)
(After the second hit of `ㅅ`, the state became `[않]`.)

---

I wonder if we always keep one character at the beginning of the text agent, what would happen in such cases.
Would the transitions look like “`x[]안` -> `안[]ㅅ` -> `[않]`”, or “`x[]안` -> `x안[]ㅅ` -> `x[않]`”? (where `x` is a random character.)

If the former pattern holds, a possible workaround is to check whether the selection starts at `0` instead of `1`.
However, this would only be reliable if the selection state were updated before `CompositionEvent`s are dispatched.

---

A more robust approach may be to build another IME-handling implementation on top of the [EditContext API], as an alternative to the text agent. A proper implementation there should eliminate this kind of inconsistencies for many users([~70%](https://caniuse.com/?search=EditContext+API)).

For the remaining ~30% of users, we can then try to find workarounds, which I feel more difficult to get right.

[EditContext API]: https://developer.mozilla.org/en-US/docs/Web/API/EditContext_API

rustbasic:
It should be handled as follows. In #7914, it is currently behaving like this:
`x[안]` -> `x안[ㅅ]` -> `x[않]` -> `x않[ㅇ]` -> `x않[아]`

umajho:
I realized that `EditContext` might not be necessary.
I will try to come up with a solution based on the current design of `TextAgent` soon.

rustbasic:
Resolved in #8045

PR #7925, opened 2026-02-19
Pass in an explicit id in `UiBuilder`, to avoid wrapping passed in ids with Id::new()

lucasmerlin:
I was really confused why I couldn't find the response for my ui with explicit id. Turns out the id I passed in was wrapped by `Id::new`


juancampa:
Glad to see this.

Went through a long head scratcher today. Then wrote this assertion and started doubting my own sanity:


```rust
let ui = ui.new_child(UiBuilder::new().id(id));
assert!(id, ui.id()); <--- ❌
```

Issue #7840, opened 2026-01-11
Sporadic desktop app crash caused by panic in wgpu


KonaeAkira:
One of the users of [raphael-rs](https://github.com/KonaeAkira/raphael-rs) reported sporadic crashes that happened randomly during the desktop application's lifetime (could be right after startup, could be after a while). The crashlog pointed to a panic in `egui-wgpu`:

```
[ERROR] panicked at C:\Users\runneradmin.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\egui-wgpu-0.33.3\src\renderer.rs:958:17:
    Failed to create staging buffer for index data. Index count: 20544. Required index buffer size: 82176. Actual size 124656 and capacity: 124656 (bytes)
```

The system info of the user who reported the crash:

```
OS: Windows 10 Pro 22H2 (19045.6466)
Hardware: Intel i7-12700F, 32 GB, NVIDIA RTX 3070 (8 GB)
Graphics driver: NVIDIA 32.0.15.6094
```

I could not reproduce the crash myself and I also don't know what is causing it. Is this a `egui-wgpu` issue?

ArthurBrussee:
I have also been seeing this after an egui upgrade. Here is a minimal repro:

```rust
use std::sync::Arc;

fn main() {
    eframe::run_native(
        "Race Test",
        eframe::NativeOptions {
            renderer: eframe::Renderer::Wgpu,
            ..Default::default()
        },
        Box::new(|cc| {
            let state = cc.wgpu_render_state.as_ref().unwrap();
            let queue = Arc::new(state.queue.clone());

            // Background thread submitting empty command buffers
            std::thread::spawn(move || {
                loop {
                    queue.submit(std::iter::empty());
                }
            });

            Ok(Box::new(App))
        }),
    )
    .unwrap();
}

struct App;

impl eframe::App for App {
    fn update(&mut self, ctx: &egui::Context, _: &mut eframe::Frame) {
        ctx.request_repaint();
    }
    fn ui(&mut self, ui: &mut egui::Ui, _: &mut eframe::Frame) {
        ui.label("Racing...");
    }
}
```

I am guessing that the queue submission causes some GPU values to go out of sync. `queue.write_buffer(_with)` calls are run as soon as the first submit happens on _any_ thread, so it can be hard to synchronize those with command buffer commands. Just a theory though, haven't found the root cause yet.

landaire:
[...]

PR #7928, opened 2026-02-22
Fix wgpu memory leak leading to panic when window is minimized (#7434)

### LINKED ISSUE #7434: WGPU buffers not deallocated when minimised
**Describe the bug**
WGPU buffer garbage collection is dependent on `wgpu::Queue::submit()`. When minimised this is not called because `egui_wgpu::winit::Painter::paint_and_update_textures()` returns early due to `SurfaceErrorAction::SkipFrame`. Any allocated wgpu buffers aren't destroyed, and any allocations build up indefinitely until out of GPU memory.
Most egui applications don't allocate much per frame and don't perform receive events/redraw while minimised, so this isn't usually noticeable.
The memory usage can be observed with `<eframe::CreationContext>.wgpu_render_state.device.get_internal_counters().hal.buffer_memory.read();`.

**To Reproduce**
1. Checkout https://github.com/nickd-airo/egui_mwe.
   - Egui window which only draws a label
   - Renders continuously using request_repaint()
   - It uses wgpu backend, and the wgpu counters feature to track buffer memory
   - It prints the buffer memory counter on every frame.
3. Run the app.
4. Observe the idle memory.
5. Minimise the app.
6. Observe the memory usage growing.
7. Unminimise the app.
8. Observe the memory usage return to normal.

**Desktop**
 - OS: Windows

**Additional context**
For a [minimal example](https://github.com/nickd-airo/egui_mwe), when window is not minimised, usage is 581728 bytes and does not change.
After minimising, usage grows linearly, and after ~30s, memory usage is 6221920 bytes. (188 kB/sec).
When unminimised, the memory usage returns to 581728 immediately.

I'm working on an app with a 3D scene inside egui, and it crashes ~20m after being minimised with 20GB of GPU memory (Normal usage is <3GB), and after even 1m of being minimised, has a significant delay from freeing memory.

It seems like `wgpu::Device::poll()` can also perform the required cleanup.

landaire:
There appears to be a special order of operations in which the buffers should be updated. The current order of operations for `paint_and_update_textures` in `egui-wgpu` causes user commands to be submitted _before_ the output frame is obtained. If no output frame is available, the buffer will grow indefinitely.

This swaps the order of operations so that we can short-circuit if no output frame is available -- before the commands are submitted.

Fix confirmed by taking the PoC from #7434, applying these patches, and re-running the PoC. Observed stable memory stats even when the application was minimized.

Note: window minimization doesn't seem to be the only condition under which this bug occurred. I could not isolate the exact behavior in my own application, but it seemed to be triggered with multiple native windows opened.

Closes #7434, #7840 

### DIFF
--- a/crates/egui-wgpu/src/winit.rs
+++ b/crates/egui-wgpu/src/winit.rs
@@ -421,12 +421,40 @@ impl Painter {
     ) -> f32 {
         profiling::function_scope!();
 
+        /// Guard to ensure that commands are always submitted to the renderer queue
+        /// so that calls to [`write_buffer()`](https://docs.rs/wgpu/latest/wgpu/struct.Queue.html#method.write_buffer)
+        /// are completed even if we take a codepath which doesn't submit commands and avoids
+        /// internal buffers growing indefinitely.
+        ///
+        /// This may happen, for example, if no output frame is resolved.
+        /// See <https://github.com/emilk/egui/pull/7928> for full context.
+        struct RendererQueueGuard<'q> {
+            queue: &'q wgpu::Queue,
+            commands_submitted: bool,
+        }
+
+        impl Drop for RendererQueueGuard<'_> {
+            fn drop(&mut self) {
+                // Only submit an empty command buffer array if no commands were
+                // explicitly submitted.
+                if !self.commands_submitted {
+                    self.queue.submit([]);
+                }
+            }
+        }
+
         let capture = !capture_data.is_empty();
         let mut vsync_sec = 0.0;
 
         let Some(render_state) = self.render_state.as_mut() else {
             return vsync_sec;
         };
+
+        let mut render_queue_guard = RendererQueueGuard {
+            queue: &render_state.queue,
+            commands_submitted: false,
+        };
+
         let Some(surface_state) = self.surfaces.get(&viewport_id) else {
             return vsync_sec;
         };
@@ -589,6 +617,9 @@ impl Painter {
                 .submit(user_cmd_bufs.into_iter().chain([encoded]));
             vsync_sec += start.elapsed().as_secs_f32();
         };
+
+        // Ensure that the queue guard does not do unnecessary work when dropped
+        render_queue_guard.commands_submitted = true;
 
         // Free textures marked for destruction **after** queue submit since they might still be used in the current frame.
         // Calling `wgpu::Texture::destroy` on a texture that is still in use would invalidate the command buffer(s) it is used in.


landaire:
cc @Cyrix126

landaire:
@Wumpf sorry for the ping, but any chance of a review? Just want to make sure this is seen before the next egui release.

Wumpf (changes requested):
sorry for the delay, missed this!

If possible I'd like to find another solution since `get_current_texture` blocks on GPU to finish previous work depending on platform and surface configuration. So doing it as late as possible means that we can do more things in parallel

> If no output frame is available, the buffer will grow indefinitely.

I'm trying to get my head around which buffer exactly may grow infinitely. Presumably it's the internal buffers wgpu has to maintain for queue-write operations. If nothing is submitted to the queue, these won't be flushed out, causing issues eventually.

So I'd suggest doing a call to `queue.submit([]);`on all early returns (can be implemented as a guard to be on the safe side). As documented [here](https://docs.rs/wgpu/latest/wgpu/struct.Queue.html#method.write_buffer) this will finish any pending transfer and should have the same effect on the wgpu device & queue's maintenance operations as-if we're actually drawing a frame.
Could you please check whether that fix would also work? 🤔 Thank you!

landaire:
No worries. I _think_, but am not positive, that it's the `wgpu::Queue` growing as a result of calling [`update_buffers`](https://github.com/emilk/egui/blob/9276778181619be4065715ab03d36ca0b318667d/crates/egui-wgpu/src/renderer.rs#L890). But you obviously know wgpu internals better than I. If there's a way this can be instrumented for a better answer let me know.

> Could you please check whether that fix would also work? 🤔 Thank you!

Can do.

landaire:
@Wumpf I took the approach you recommended and it seemed to work. Please nitpick the guard naming / where the struct is currently located. I defined it in the function body as I didn't think it'd be used outside of that function.

Video below demonstrates the change.

https://github.com/user-attachments/assets/10a6dee8-1705-4ac9-b84e-60ea47fad705

Wumpf (changes requested):
awesome! Thanks for looking after this and demonstrating that it also does the trick. Very happy this worked out.

Bit of a nit, but `submit` can be relatively expensive even if it noops (it does a whole lotta locking and bumps counters and stuff) so I think it would be nice to tell the guard it doesn't have to do anything after we've done the "real" submit

landaire:
I think perhaps this can be better implemented in a way where access to the queue is gated by the guard, similar to a mutex, and the guard provides proxy access to the queue to ensure auto-toggling of the flag but that may be too prohibitive?

Wumpf (approved):
> I think perhaps this can be better implemented in a way where access to the queue is gated by the guard, similar to a mutex, and the guard provides proxy access to the queue to ensure auto-toggling of the flag but that may be too prohibitive?

That sounds good to me, but I have too little of a strong opinion on the exact approach here and I think what you have on the branch now is good. So time to get this fix landed! :)

c-git:
Looking at the dates Mar 5 for this and Mar 27 for the release, I'd think this made the release but I just bumped into this problem and wanted to confirm if being on 0.34.1 means I should get this anymore. Doesn't happen very often, I haven't figured out when exactly yet but this time I looked up issues with the word renderer and followed the trail here but I just (while my computer was asleep got):

```
 thread 'main' (410427) panicked at /home/one/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/egui-wgpu-0.34.1/src/renderer.rs:971:17:
 Failed to create staging buffer for index data. Index count: 5982. Required index buffer size: 23928. Actual size 24576 and capacity: 24576 (bytes)
```

I doesn't happen often and it a material problem for me right now but just wanted to know if this is maybe a case in which the fix doesn't work or perhaps it didn't make the release. Also if you could help me understand for future how to check if something made the release I'd appreciate that as well.

I don't see this mentioned in the release notes so I guess it didn't make it.

landaire:
@c-git strangely enough, upgrading to 0.34 re-introduced it for my users too. I had previously applied this fix on top of somewhere in-between 0.33 and 0.34: https://github.com/landaire/egui/tree/wgpu_fix

Looks like this commit specifically: https://github.com/emilk/egui/commit/44cdd653e2317d300fb8a6c9c36b03f23991e803

I had no reports of the issue being encountered by my users after switching to this. However, upgrading to 0.34 they get it _sometimes_. Doesn't seem to be at the same rate and I can't reasonably reproduce it.

c-git:
I see thank you. It's not a major problem for me yet so for now I'll monitor and see if I can determine any pattern that I can feedback here to help.

ArthurBrussee:
I've also seen this re-appear since upgrading to 0.34, or at least, the "Failed to create staging buffer for index data." - hard to say why :(

landaire:
My application is something that runs in the background while people play games (some stats analysis tool). Users who hit this issue seem to have the application up on a second monitor and it generally crashes while playing the game. On my TODO list if I can repro.

PR #7929, opened 2026-02-22
Add touch pinch-to-zoom support (emit `egui::Event::Zoom`)

rustbasic:
Add touch pinch-to-zoom support (emit `egui::Event::Zoom`)

Tested: Android 

- Add `handle_touch_zoom()` to detect two-finger pinch gestures on web touch events and push `egui::Event::Zoom(scale)`.
- Note: applying the zoom is up to the user. Implement `handle_pinch_zoom()` in your app, or use the provided example to consume `egui::Event::Zoom` and call `egui::gui_zoom::zoom_in/zoom_out`.

```rust
    pub fn handle_pinch_zoom(ui: &mut egui::Ui) {
        let mut new_zoom_factor: Option<f32> = None;
        let current_zoom = ui.ctx().zoom_factor();

        ui.input(|i| {
            for event in &i.events {
                if let egui::Event::Zoom(delta) = event {
                    let new_zoom = current_zoom * delta;
                    new_zoom_factor = Some(new_zoom.clamp(0.2, 5.0));
                }
            }
        });

        if let Some(zoom_factor) = new_zoom_factor {
            ui.ctx().set_zoom_factor(zoom_factor);
        }
    }
```

### DIFF
--- a/crates/eframe/src/web/input.rs
+++ b/crates/eframe/src/web/input.rs
@@ -120,12 +120,12 @@ fn handle_touch_zoom(runner: &mut AppRunner, event: &web_sys::TouchEvent) {
         let initial_dist = *runner.input.initial_touch_dist.get_or_insert(current_dist);
 
         if initial_dist > 0.0 {
-            let raw_scale = current_dist / initial_dist;
-            let new_scale = (raw_scale * 10.0).round() / 10.0;
+            let total_scale = current_dist / initial_dist;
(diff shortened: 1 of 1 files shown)


lucasmerlin:
We already have [`Ctx::multi_touch`](https://docs.rs/egui/latest/egui/struct.Context.html#method.multi_touch), which should support this already.

PR #7940, opened 2026-02-28
Fix emoji icon font

Jhynjhiruu:
* Closes <https://github.com/emilk/egui/issues/1284> (again)

Short and simple PR, just moves the updated font to the right place. I note that the license for that font says copyright 2014, which might need to be updated to reflect when the font was modified.


Jhynjhiruu:
Gah, forgot about the branch thing. This shooooouldn't need any edits from maintainers, but if it's a problem I'll fix it.

PR #7944, opened 2026-03-01
Fix: ScrollArea layout jitter with floating bars and zoom levels

### LINKED ISSUE #7937: Unstable `available_size` in `CentralPanel`

rustbasic:
Fix: ScrollArea layout jitter with floating bars and zoom levels

* Closes #7937 
* Closes #7942 

This PR improves the stability of `ScrollArea` by addressing two layout issues:

1. **Discrete Layout for Floating Bars:** Fixed content "shaking" in `floating` mode when a non-zero `allocated_width` is used. By using discrete visibility (`show_bars`) instead of the animated factor for space allocation, we ensure the layout remains stable during scrollbar animations.
2. **Zoom-level Stability:** Introduced a small epsilon (0.1) when checking if content exceeds the viewport. This prevents scrollbars from flickering on and off due to floating-point rounding errors at specific zoom factors (e.g., 1.01 or 0.95).


rustbasic:
Applied an epsilon correction as a temporary workaround, but have now implemented a more fundamental fix.

emilk (approved):

PR #7946, opened 2026-03-02
Add `Style::strong_font/emphasis_font` for bold/italic rendering

w4nderlust:
## Summary

This adds two optional fields to `Style` that let users swap in real bold and italic font families for `RichText::strong()` and `RichText::italics()`, instead of relying on the default approximations (color change for bold, vertex skew for italic).

Currently, `RichText::strong()` only changes the text color (via `Visuals::strong_text_color()`), which does not produce visually bold text. And `RichText::italics()` applies a 25-degree horizontal vertex skew, which is a rough approximation that looks noticeably different from a real italic typeface. These defaults are fine when no custom fonts are available, but when users have access to proper Bold/Italic font files, there is no way to tell egui to use them for strong/italic markup.

## Changes

Two new `Option<FontFamily>` fields on `Style`:

- **`strong_font`**: When set, `RichText::strong()` renders with this font family instead of applying the strong text color. The color override is skipped because the heavier strokes of a real bold typeface provide visual distinction on their own.

- **`emphasis_font`**: When set, `RichText::italics()` renders with this font family instead of applying the vertex skew. The skew is skipped because the real italic letterforms are used instead.

Both default to `None`, so existing behavior is completely unchanged.

### Implementation details

The logic lives in `RichText::into_text_and_format()` and `RichText::get_text_color()` in `widget_text.rs`:

- `into_text_and_format()` resolves the bold/italic font family overrides before building the `FontId`. If both `strong_font` and `emphasis_font` are set and the text is both strong and italic, `emphasis_font` wins (since a `FontId` can only reference one family). Users who need combined bold-italic can register a dedicated `BoldItalic` family and apply it via `RichText::family()`.

- `get_text_color()` skips the strong color override when `strong_font` is set, falling through to the default text color instead.

## Usage

```rust
// 1. Register bold/italic fonts
fonts.font_data.insert(
    "MyFont-Bold".to_owned(),
    egui::FontData::from_static(include_bytes!("MyFont-Bold.ttf")).into(),
);
fonts.families.insert(
    egui::FontFamily::Name("Bold".into()),
    vec!["MyFont-Bold".to_owned()],
);
ctx.set_fonts(fonts);

// 2. Configure style
ctx.global_style_mut(|style| {
    style.strong_font = Some(egui::FontFamily::Name("Bold".into()));
    // style.emphasis_font = Some(egui::FontFamily::Name("Italic".into()));
});
```

After this, any `RichText::new("hello").strong()` will render using the Bold font family.

## Motivation

This came up while building a markdown renderer on top of egui. Markdown has `**bold**` and `*italic*` syntax, and users expect these to render with proper bold/italic typefaces when available. Libraries like `egui_commonmark` use `RichText::strong()` and `RichText::italics()` internally, so configuring the style once is enough to get real bold/italic rendering throughout the whole UI without patching downstream libraries.

## Checklist

- [x] `cargo fmt` passes
- [x] `cargo clippy` passes
- [x] All 175 egui tests pass (including the new doc tests)
- [x] No changes to default behavior (both fields default to `None`)
- [x] Doc comments with working examples on both new fields

### DIFF
--- a/crates/egui/src/style.rs
+++ b/crates/egui/src/style.rs
@@ -339,6 +339,78 @@ pub struct Style {
 
     /// Use a more compact style for menus.
     pub compact_menu_style: bool,
+
+    /// Optional [`FontFamily`] to use for bold/strong text.
+    ///
+    /// By default ([`None`]), [`RichText::strong()`] applies a color change
+    /// (using [`Visuals::strong_text_color`]) to convey emphasis.
+    /// When this is set to [`Some`], strong text will instead be rendered with the
+    /// specified font family, and the color change is skipped.
+    ///
+    /// To use this, first register a bold font via [`crate::Context::set_fonts`],
+    /// then point this field at it:
+    ///
+    /// ```
+    /// # let ctx = egui::Context::default();
+    /// // 1. Register the bold font data and family
+    /// let mut fonts = egui::FontDefinitions::default();
+    /// let bold_font_data: &[u8] = &[]; // bytes of a .ttf file
+    /// fonts.font_data.insert(
+    ///     "my_bold_font".to_owned(),
+    ///     egui::FontData::from_static(bold_font_data).into(),
+    /// );
+    /// fonts.families.insert(
+    ///     egui::FontFamily::Name("Bold".into()),
+    ///     vec!["my_bold_font".to_owned()],
+    /// );
+    /// ctx.set_fonts(fonts);
+    ///
+    /// // 2. Tell egui to use it for strong text
+    /// ctx.global_style_mut(|style| {
+    ///     style.strong_font = Some(egui::FontFamily::Name("Bold".into()));
+    /// });
+    /// ```
+    ///
+    /// **Note:** When text is both strong and italic, and both `strong_font` and
+    /// [`Self::emphasis_font`] are configured, `emphasis_font` takes precedence
+    /// because a single [`FontId`] can only reference one font family.
+    pub strong_font: Option<FontFamily>,
+
+    /// Optional [`FontFamily`] to use for italic/emphasis text.
+    ///
+    /// By default ([`None`]), [`RichText::italics()`] fakes italics by applying a
+    /// horizontal vertex skew during tessellation.
+    /// When this is set to [`Some`], italic text will instead be rendered with the
+    /// specified font family, and the vertex skew is skipped.
+    ///
+    /// To use this, first register an italic font via [`crate::Context::set_fonts`],
+    /// then point this field at it:
+    ///
+    /// ```
+    /// # let ctx = egui::Context::default();
+    /// // 1. Register the italic font data and family
+    /// let mut fonts = egui::FontDefinitions::default();
+    /// let italic_font_data: &[u8] = &[]; // bytes of a .ttf file
+    /// fonts.font_data.insert(
+    ///     "my_italic_font".to_owned(),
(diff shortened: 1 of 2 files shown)


Tristan117:
This looks like a great small quality of life feature! 
In my application as a workaround I'm currently manually using an italic font on italic text rather than `RichText::italics()`  which this PR would avoid.

@emilk is there a reason the italic and bold versions of the UbuntuLight default font aren't included in the default fonts (or behind a feature if binary size/performance is a concern for some targets)? Would a PR along these lines be of interest?

emilk:
> is there a reason the italic and bold versions of the UbuntuLight default font aren't included in the default fonts

Purely to keep the size of the .wasm blob smaller. Having it optional behind a feature flag would be nice.

---

Something has gone wrong with the rebasing of this PR - the diff is huge, with a lot of unrelated changes

w4nderlust:
@emilk sorry baout it. Force-pushed the branch to drop the unrelated changes; PR now contains only the two commits (the API addition + a one-line doc-link fix, 2 files, +135/-7), and I'll do the bundled-font feature flag as a follow-up PR if you think it's useful.

emilk (commented):
I'm having seconds thoughts about this PR.

The problem with this approach is that it will use the same bold/italics font for all text. So bold monospace will not be monospace!

What we really should support is multiple different font faces per `FontFamily` (that is what the "family" part of the name means!). But ofc, this is a bigger lift and requires us rethinking `FontDefinitions` and other things. Probably good to take inspiration from Parley and friends here

w4nderlust:
You're probably right that the current shape isn't ideal. Resolving emphasis at the `Style` level means a global family swap, which breaks family identity (monospace + strong stops being monospace). One direction could be to move variants into the family, like Parley does: a family owns a set of faces tagged with weight/style, and lookup returns the best match plus a `Synthesis { embolden, skew }` flag for the renderer when the exact variant isn't there.

The smallest version that might fit egui could look like:

```rust
enum FontVariant { Regular, Bold, Italic, BoldItalic }

struct FamilyFaces {        // replaces Vec<String> as the value in FontDefinitions::families
    regular: Vec<String>,
    bold: Vec<String>,
    italic: Vec<String>,
    bold_italic: Vec<String>,
}

struct FontId { size, family, variant }   // variant defaults to Regular
```

Resolution would go from `(family, variant)` to a fallback chain inside `FamilyFaces`. If the requested variant is empty, fall back to `regular` and set a `Synthesis` flag, so today's color darken and vertex skew become the synthesis path. `RichText::strong()` and `.italics()` stay as flags and get translated to a `FontVariant` in `into_text_and_format`. No Style lookup, no family swap, and bold monospace stays monospace.

To keep it minimal I'd skip the CSS weight scale, the width axis, auto detection from name/OS/2 tables, and cross family variant search. Just the 4 flat variants, with synthesis as the only fallback. The change would touch `fonts.rs`, the italic skew in `text_layout.rs`, `widget_text.rs`, and remove `strong_font`/`emphasis_font` from `style.rs` (clean cut since they're unreleased), plus snapshot regen.

One question before I write code: flat `FontVariant` enum, or something more parley shaped like `(weight: u16, style: enum)` with CSS matching? I'd lean flat since it matches what `RichText` exposes today and finer weights stay additive, but happy either way.

If this sounds reasonable I'll close this PR and open a fresh one against `main`.

Issue #5113, opened 2024-09-16
`App::tick`


emilk:
Calling the full `App::update` when the application is hidden (background thread or minimized on native, for instance) is bad for several reasons:

A) we don't always know what the size of the viewport is/should be
B) we pay the cost of GUI even though we may only need to process some events

I think therefor it makes sense to split `App::update` into two functions:

```rs
trait App {
    /// You can run logic here, but can't show any GUI.
    ///
    /// Called once before each call to [`Self::show`], but may also be called
    /// without a subsequent call to `show`,
    /// for instance if the application is hidden and [`Context::request_repaint`] has been called.
    ///
    /// The egui context may ONLY be used to schedule repaints, nothing else.
    fn tick(&mut self, ctx: &egui::Context, frame: &mut Frame)

    /// Show the GUI. Each call to this is preceded by a call to [`Self::tick`].
    fn show(&mut self, ctx: &egui::Context, frame: &mut Frame);
}
````

rustbasic:
How about keeping update() as is and just adding tick()?

NotQuiteApex:
Hoping this gets prioritized soon. Having `App::tick` would be great for implementing system tray icon functionality, something I'm looking to get implemented in my own project, as described here: https://github.com/emilk/egui/discussions/737#discussioncomment-8830140

MonterraByte:
Would it be a good idea to have a `request_repaint_after` equivalent for this, so that we can keep processing events without necessarily repainting the UI?

PR #7948, opened 2026-03-02
Add `ViewportInfo::occluded` and `visible`

emilk:
* Part of https://github.com/emilk/egui/issues/5112
* Part of https://github.com/emilk/egui/issues/5113
* Part of https://github.com/emilk/egui/issues/5136

Once we support calling `App::logic` when an app is occluded or minimized, it is useful to know that it is, in fact, occluded or minimized.


emilk:
Indeed - making sure that `ui` and `logic` are called at the right times will come later

PR #7958, opened 2026-03-05
⚠️ Atom improvements: `Atom::id`, `align`, `closure`, `max_size`

lucasmerlin:
Migration guide:
- `AtomKind::Custom` has been removed. You can now set an id to any kind via `Atom::custom` or `AtomExt::atom_id`.

### DIFF
--- a/crates/egui/src/atomics/atom_layout.rs
+++ b/crates/egui/src/atomics/atom_layout.rs
@@ -490,7 +490,7 @@ impl<'atom> AllocatedAtomLayout<'atom> {
 
 /// Response from a [`AtomLayout::show`] or [`AllocatedAtomLayout::paint`].
 ///
-/// Use [`AtomLayoutResponse::rect`] to get the response rects from [`Atom::custom`].
+/// Use [`AtomLayoutResponse::rect`] to get the response rects from [`crate::Atom::custom`].
 #[derive(Clone, Debug)]
 pub struct AtomLayoutResponse {
     pub response: Response,
@@ -510,7 +510,7 @@ impl AtomLayoutResponse {
         self.custom_rects.iter().copied()
     }
 
-    /// Use this together with [`Atom::custom`] to add custom painting / child widgets.
+    /// Use this together with [`crate::Atom::custom`] to add custom painting / child widgets.
     ///
     /// NOTE: Don't `unwrap` rects, they might be empty when the widget is not visible.
     pub fn rect(&self, id: Id) -> Option<Rect> {


crates/egui/src/atomics/atom_kind.rs
```diff
@@ -57,7 +72,47 @@ pub enum AtomKind<'a> {
     /// }
     /// # });
     /// ```
+    #[deprecated = "Use Atom::custom(id) instead"]
```
emilk:
```suggestion
    #[deprecated = "Use Atom::custom(id, $what size do I put here?) instead"]
```

crates/egui/src/atomics/atom_kind.rs
```diff
@@ -57,7 +72,47 @@ pub enum AtomKind<'a> {
...
     Custom(Id),
```
emilk: We could also remove this variant and add a deprecated function called `fn Custom` (sic). Might make the rest of the PR simpler

crates/egui/src/atomics/sized_atom_kind.rs
```diff
@@ -10,16 +10,16 @@ pub enum SizedAtomKind<'a> {
     Empty,
     Text(Arc<Galley>),
     Image(Image<'a>, Vec2),
-    Custom(Id),
+    Sized(Vec2),
```
emilk: Aren't Text and Image sized too?

crates/egui/src/atomics/atom_kind.rs
```diff
@@ -69,6 +124,12 @@ impl<'a> AtomKind<'a> {
         AtomKind::Image(image.into())
     }
 
+    pub fn closure(
+        func: impl FnOnce(&Ui, Vec2, TextWrapMode, FontSelection) -> (Vec2, SizedAtomKind<'static>) + 'a,
```
emilk: Docstring?

What is the `Vec2` in there? Would a `Rect` make more sense?

crates/egui/src/atomics/atom_kind.rs
```diff
@@ -1,9 +1,24 @@
...
+use std::fmt::Debug;
+
+pub type AtomClosure<'a> = Box<
+    dyn FnOnce(
+            &Ui,
+            Vec2,
+            TextWrapMode,
+            FontSelection,
+        ) -> (
+            Vec2,
```
emilk: What is this? Can this return a named struct instead? And maybe group (some) of the params as a struct too?

crates/egui/src/atomics/atom_kind.rs
lucasmerlin: That won't work since we can't set the Id that way. I've now removed Custom though, I doubt that saw much direct usage and migration should be simple enough.

emilk:
I'm turning this into a draft until my comments above are resolved

emilk (approved):

Issue #7485, opened 2025-08-29
Incorrect typing behavior with TextEdit singleline and multiline with IbusWayland enabled


Alvenster:
**Describe the bug**

Consider the following code :
```rust 

use eframe;
fn main() {
    println!("Hello, world!");
    let options = eframe::NativeOptions::default();
    let app = eframe::run_native(
        "My app",
        options,
        Box::new(|_cc| Ok(Box::new(App::default())))
    );
}

#[derive(Default)]
struct App {
    text_edit: String,
}


impl eframe::App for App {
    fn update(&mut self, ctx: &eframe::egui::Context, frame: &mut eframe::Frame) {
        eframe::egui::Window::new("My Window").show(ctx, |ui| {

            ui.text_edit_singleline(&mut self.text_edit);
            ui.label("test");
        });
    }
}

```
When typing inside a singleline or multiline, I cannot type more than one character unless I manually defocus and refocus the widget. Another weird thing : Holding Alt allows me to type freely again as if there were no bug whatsoever. 
Tried toggling X11, Wayland and wgpu features -> Does not affect the outcome.

This whole thing does not occur on windows (it does not occur on Wine either) and I cannot test for MacOs.

[Here is the bug caught in 4K ] (https://wim.nl.tab.digital/s/J8iSCTYtGWcc6YE) 
For some reason Github does not accept my video file so i'm sharing it through my nextcloud

**To Reproduce**
1. Copy, paste, and run the above described code
2. Type inside the singleline
3. Witness the fact that you cannot type more than one character
4. Defocus and refocus the widget
5. Type another single character
6. Repeat until satisfied

Step to repoduce the other behavior:
1. Copy, paste, and run the above described code
2. Type inside the singleline
3. Witness the fact that you cannot type more than one character
4. Hold Alt
5. Witness the fact that by hold Alt, it is possible to freely type

**Expected behavior**
The singleline should accept multiple character input without refocusing on the widget or holding Alt

**Screenshots**
Check the [video](https://wim.nl.tab.digital/s/J8iSCTYtGWcc6YE)

**Desktop:**
 - OS: Fedora 42 KDE Plasma 6.4.4 on Wayland session
 - eframe on latest main ( 08/29/2025 8PM)
 - Rust compiler version is 1.89.0
 - Hardware : DELL G7 with Nvidia Discrete GPU ( RTX 2060 )

**Additional context**
[...]

Issue #4908, opened 2024-08-03
backspace fail after ime input


tkkcc:
**Describe the bug**
![Peek 2024-08-03 18-25](https://github.com/user-attachments/assets/2a5e8fc9-89fe-4b47-855c-f35a8609e989)

**To Reproduce**
run master(d856f7b3a58fc65a81ff9657f5c6e1ab462b7bb6) egui_demo_app, use fcitx5 ime to input unicode words in text_edit, press backspace won't delete char, but typing, ime input and del key still work.

**Expected behavior**
backspace delete char

**Desktop (please complete the following information):**
 - OS: archlinux, fcitx5 5.1.10

**Additional context**
web version work
use chinese font not help

rustbasic:
I think it will be fine if you apply #4896.

* #4896

rustbasic:
I think it will be fine if you apply #4910

* #4910

YgorSouza:
It seems that after fcd02bd7d1dde123a1b2dce11ae86cea17029ace, the backspace and arrow keys no longer work in the TextEdit in native linux. I guess it thinks IME is always enabled? Even if I disable the compose key and use only the physical keys, it still doesn't work.

rustbasic:
> It seems that after [fcd02bd](https://github.com/emilk/egui/commit/fcd02bd7d1dde123a1b2dce11ae86cea17029ace), the backspace and arrow keys no longer work in the TextEdit in native linux. I guess it thinks IME is always enabled? Even if I disable the compose key and use only the physical keys, it still doesn't work.

Yes. I am currently preparing to commit. Do you use IME? or Will it be recognized as IME enabled even in English?

YgorSouza:
I don't have IME, just a regular US English keyboard. I use the compose key from the Gnome desktop environment, which allows me to enter special characters, e.g., `RCtrl + " + E` enters the Ë character. I don't know if that counts as IME, but the arrow keys and backspace still don't work if I turn that off in the system settings. The `TextEdit` gets the IME enabled event as soon as I click on it. It is disabled when I move the window, and the enabled again. If I enter a special character with the compose key, it gets the commit, disabled, and enabled events. So basically it is always enabled while the `TextEdit` has focus.

![image](https://github.com/user-attachments/assets/8596b934-6f5f-436f-91c7-2873c9db134a)


Also, when I press the compose key, it adds a · character to the string until I press the other keys, and if I click away the character just stays there.

[...]

PR #7967, opened 2026-03-11
Much improved IME

### LINKED ISSUE #7809: The focus on the `TextEdit` input is removed when typing is completed with Japanese IME
At first I clarify that the adding the lines that applies the main branch of `egui` to my `Cargo.toml` cause the following error.

```
error: failed to download `fearless_simd v0.3.0`

Caused by:
  unable to get packages from source

Caused by:
  failed to unpack package `fearless_simd v0.3.0`

Caused by:
  failed to iterate over archive

Caused by:
  invalid gzip header
```

**Describe the bug**

https://github.com/user-attachments/assets/23b76edd-7dd5-4a2d-beb9-8e810a19f784

In this video, you can see the Latin alphabet (English), Hangul (Korean), and Japanese characters being typed properly. However, when Enter key is pressed after typing with Japanese IME, the focus, which should be retained even after pressing Enter, is removed from the `TextEdit` input in `egui` application.

**To Reproduce**

This is my `main.rs` code to reproduce this bug.

```rust
use std::sync::Arc;
use eframe::egui;

fn main() -> eframe::Result {
    eframe::run_native(
        "TextEdit Test",
        eframe::NativeOptions {
            viewport: egui::ViewportBuilder::default().with_inner_size(
                (640.0, 360.0)
            ),
            ..Default::default()
        },
        Box::new(|ctx| {
            init_fonts(&ctx.egui_ctx);
            Ok(Box::new(App::new()))
        })
    )
}

// replace this with the proper path of the font file that supports CJK characters.
const MAIN_FONT: &'static [u8] = include_bytes!("./NotoSansCJK-Regular.otf");

fn init_fonts(ctx: &egui::Context) {

    fonts.font_data.insert(
        "main-font".to_owned(),
        Arc::new(egui::FontData::from_static(MAIN_FONT))
    );

    let proportional = fonts.families.entry(egui::FontFamily::Proportional).or_default();
    proportional.insert(0, "main-font".to_owned());

    ctx.set_fonts(fonts);
}

pub struct App {
    text_content: String
}

impl App {
    pub fn new() -> Self {
        App { text_content: String::new() }
    }
}

impl eframe::App for App {
            ui.heading("TextEdit");
            ui.text_edit_singleline(&mut self.text_content);
            egui::TextEdit::singleline(&mut self.text_content)
        });
    }
}
```

**Expected behavior**

The focus should not be removed when pressing Enter, as shown at the end of the video.

 - OS: Windows
 - Browser: -
 - Version: -

**Smartphone (please complete the following information):**
 - Device: -
 - OS: -
 - Browser: -
 - Version: -

**Additional context**

\-

### LINKED ISSUE #7876: IME (CJK) commitment with Enter key causes unintended newline in multiline TextEdit
## Describe the bug
In `egui::TextEdit::multiline`, when using an IME (Input Method Editor) for CJK languages (Chinese, Japanese, Korean), pressing the **Enter key** to commit the preedit string results in an unintended newline (`\n`) being inserted into the text buffer.

The Enter key should only be used to commit the composition, but `egui` (or the underlying `winit`) appears to process it as a physical "New Line" command simultaneously.

## Minimal Reproducible Example
// This is the most basic code to reproduce the issue.
// No extra key handling or custom logic is involved.
ui.add(egui::TextEdit::multiline(&mut self.text));

// The problem:
// In CJK IME, one Enter key press results in:
// 1. ImeEvent::Commit (Correct)
// 2. Event::Key { key: Enter, .. } + Event::Text("\n") (Incorrect)

## Steps to reproduce
1. Use `egui` on Windows/macOS with a CJK IME enabled.
2. Focus on a `TextEdit::multiline` field.
3. Type characters to start an IME composition (Preedit state).
4. Press **Enter** to commit the composition.
5. **Observed Behavior:** The characters are committed, but a newline is added immediately after them.

## Technical Findings
Based on event monitoring:
- During the commitment frame, `egui` receives both the committed text and a `Key::Enter` press event.
- Even when the IME is supposed to "consume" the Enter key for commitment, the event leaks through as a text input of `\n`.
- This makes it impossible to implement a standard CJK writing experience where Enter only commits text without jumping to the next line.

## Expected behavior
The Enter key event used for IME commitment should be suppressed or consumed, ensuring that `\n` is only inserted when the user explicitly intends to create a new line (i.e., when no composition is active or via a specific modifier).

## Environment
- **egui version:** 0.33.3
- **OS:** Windows / macOS (IME issues are common across both)

### LINKED ISSUE #7908: Backspace deletes characters in TextEdit during IME composition (CJK)
Describe the bug
When using a CJK (Chinese, Japanese, Korean) Input Method Editor (IME), pressing Backspace while in the composition state (pre-edit) incorrectly deletes characters from the TextEdit buffer. It seems the Backspace event is being "double-consumed" by both the IME (to remove a syllable/character in the candidate window) and the TextEdit widget (to remove a committed character in the box).

To Reproduce

Run an eframe application with a TextEdit widget.
Focus the TextEdit and switch to a Chinese IME (e.g., Sogou Pinyin or Microsoft Pinyin).
Type some letters (e.g., "ceshi") to open the IME candidate window.
Press Backspace to correct the pinyin.
You will see that while the pinyin is being corrected, the characters already existing in the input box are also being deleted.

Expected behavior
When the IME is active and has a pre-edit string (composition), the Backspace key should only be handled by the IME. egui should ignore the Backspace key event until the composition is finished or the candidate window is empty.

Screenshots

Image

Desktop (please complete the following information):

OS: Windows 10/11 (or macOS/Linux, please specify your OS)

Browser: N/A (Native eframe)

Version: egui/eframe 0.30

Additional context
I am using egui/eframe 0.30. This issue persists even in the latest versions. It appears that winit might not be marking the Backspace key as "consumed" when the IME is active, or egui's TextEdit is not checking the input.ime state before processing the backspace event. This significantly hinders the user experience for CJK users, especially when using third-party IMEs like Sogou.

umajho:
* Closes #7809
* Closes #7876
* Closes #7908
* Supersedes #7877
* Supersedes #7898
  * The author of the PR above replaced it with #7914, which additionally fixes another IME issue. I believe that fix deserves a separate PR.
* Reverts #4794  

This approach is better than #7898 (#7914) because it correctly handles all three major IME types (Chinese, Japanese, and Korean) without requiring a predefined “IME mode”.

## Environments I haved tested this PR in

<details><summary>macOS 15.7.3 (AArch64, Host of other virtual machines)</summary>

Run command: `cargo run -p egui_demo_app --release`

Tested IMEs:

- builtin Chinese IME (Shuangpin - Simplified)
- builtin Japanese IME (Romaji)
- builtin Korean IME (2-Set)
</details>

<details><summary>Windows 11 25H2 (AArch64, Virtual Machine)</summary>

Build command: `cargo build --release -p egui_demo_app --target=x86_64-pc-windows-gnu --features=glow --no-default-features`

(I cannot use `wgpu` due to [this bug](https://github.com/emilk/egui/issues/4381), which prevents debugging inside the VM. Anyways, the rendering backend should be irrelevant here.)

Tested IMEs:

- builtin Chinese IME (Shuangpin)
- Sogou IME (Chinese Shuangpin)
- WeType IME (Chinese Shuangpin)
- builtin Japanese IME (Hiragana)
- builtin Korean IME (2 Beolsik)
</details>

<details><summary>Linux [Wayland + IBus] (AArch64, Virtual Machine)</summary>

Fedora KDE Plasma Desktop 43 [Wayland + IBus 1.5.33-rc2]

(Not working at the moment because of [another issue](https://github.com/emilk/egui/issues/7485) that will be fixed by #7983. It is [a complicated story](https://github.com/emilk/egui/pull/7973#issuecomment-4074627603). )

> [!NOTE]
>
> IBus is partially broken in this system. The Input Method Selector refuses to select IBus. As a workaround, I have to open System Settings -> Virtual Keyboard and select “IBus Wayland” to start an IBus instance that works in egui.
>
> The funny thing is: the Chinese Intelligent Pinyin IME is broken in native Apps like System Settings and KWrite, but works correctly in egui!
>
> <details><summary>Screencast: What</summary>
>
> 2026-03-13 3 10 11 AM
> </details>

Build command: `cross build --release -p egui_demo_app --target=aarch64-unknown-linux-gnu --features=wayland,wgpu --no-default-features`

(The Linux toolchain on my mac is somehow broken, so I used `cross` instead.)

Tested IMEs:

- Chinese Intelligent Pinyin IME (Shuangpin)
- Japanese Anthy IME (Hiragana)
- Korean Hangul IME
</details>

<details><summary>Linux [X11 + Fcitx5] (AArch64, Virtual Machine)</summary>

Debian 13 [Cinnamon 6.4.10 + X11 + Fcitx5 5.1.2]

Build command: `cross build --release -p egui_demo_app --target=aarch64-unknown-linux-gnu --features=x11,wgpu --no-default-features`

Tested IMEs:

- Chinese Shuangpin IME
- Chinese Rime IME with `luna-pinyin`
- Japanese Mozc IME (Hiragana)
- Korean Hangul IME

Unlike macOS and Linux + Wayland, key-release events for keys processed by the IME are still forwarded to `egui`. These appear to be harmless in practice.
Unlike on Windows, however, they cannot be filtered reliably because there are no corresponding key-press events marked as “processed by IME”.

</details>

---

There are too many possible combinations to test (Operating Systems × [Desktop Environment](https://en.wikipedia.org/wiki/Desktop_environment)s × [Windowing System](https://en.wikipedia.org/wiki/Windowing_system)s × [IMF](https://wiki.archlinux.org/title/Input_method#Input_method_framework)s × [IME](https://en.wikipedia.org/wiki/Input_method)s × …), and I only have access to a limited subset. For example, Google Japanese Input refused to install on my Windows VM, and some paid Japanese IMEs are not accessible to me. Therefore, I would appreciate feedback from people other than me using all kinds of environments.

## Details

There are two possible approaches to removing keyboard events that have already been processed by an IME:

* Approach 1: Filter out events inside `egui` that appear to have been received during IME composition.
* Approach 2: Filter out such events in the platform backend (terminology [borrowed from imgui](https://github.com/ocornut/imgui/blob/master/docs/BACKENDS.md#using-standard-backends), e.g. the `egui-winit` crate or the code under `web/` in the `eframe` crate.).

Both approaches already exist in `egui`:

* #4794 uses the first approach, filtering these events in the `TextEdit`-related code.
* `eframe` uses the second approach in its web integration. See: <https://github.com/emilk/egui/blob/14afefa2521d1baaf4fd02105eec2d3727a7ac36/crates/eframe/src/web/events.rs#L173-L176>

Compared to the first approach, the second has a clear advantage: when events are passed from the platform backends into `egui`, they are simplified and lose information. In contrast, events in the platform backends are the original events, which allows them to be handled more flexibly. This is also why #7898 (#7914), which attempts to address the issue from within the `egui` crate, struggles to make all IMEs work correctly at the same time and requires manually selecting an “IME mode”: the events received by `egui` have already been reduced and therefore lack necessary information.

A more appropriate solution is to consistently follow the second approach, explicitly requiring platform backends not to forward events that have already been processed by the IME to `egui`. This is the method used in this PR. Specifically, this PR works within the `egui-winit` crate, where the original `KeyboardInput` events can be accessed. At least for key press events, these can be used directly to determine whether the event has already been processed by the IME on Windows (by checking whether `logical_key` equals `winit::keyboard::NamedKey::Process`). This makes it straightforward to ensure that all IMEs work correctly at the same time.

This PR also reverts #4794, which took the first approach. It filters out some events that merely look like they were received during IME composition but actually are not. It also messes up the order of those events along the way.
As a result, it caused several IME-related issues. One of the sections in the Demonstrations below will illustrate these problems.

## Demonstrations

<details><summary>Changes not included in this PR for displaying Unicode characters in demonstrations</summary>

Download `unifont-17.0.03.otf` from <https://unifoundry.com/pub/unifont/unifont-17.0.03/font-builds/>, and place it at `crates/egui_demo_app/src/unifont-17.0.03.otf`.

In `crates/egui_demo_app/src/wrap_app.rs`, add these lines at the beginning of `impl WrapApp`'s `pub fn new`:
```rust
        {
            const MAIN_FONT: &'static [u8] = include_bytes!("./unifont-17.0.03.otf");

            fonts.font_data.insert(
                "main-font".to_owned(),
                std::sync::Arc::new(egui::FontData::from_static(MAIN_FONT)),
            );

            let proportional = fonts
                .families
                .entry(egui::FontFamily::Proportional)
                .or_default();
            proportional.insert(0, "main-font".to_owned());

            cc.egui_ctx.set_fonts(fonts);
        }
```

(I took this from somewhere, but I forgot where it is. Sorry…)
</details>

[GNU Unifont](https://unifoundry.com/unifont/index.html) is licensed under [OFL-1.1](https://unifoundry.com/OFL-1.1.txt).

### This PR Fixes: Focus on a single-line `TextEdit` is lost after completing candidate selection with Japanese IME on Windows (#7809)

<details><summary>Screencast: ✅ Japanese IME now behaves correctly while Korean IME behaves as before</summary>

</details>

### This PR Fixes: Committing Japanese IME text with <kbd>Enter</kbd> inserts an unintended newline in multiline `TextEdit` on Windows (#7876)

<details><summary>Screencast: ✅ Japanese IME now behaves correctly while Korean IME behaves as before</summary>

</details>

### This PR Fixes: Backspacing deletes characters during composition in certain Chinese IMEs (e.g., Sogou) on Windows (#7908)

<details><summary>Screencast: ✅ Sogou IME now behaves correctly</summary>

</details>

### This PR Obsoletes #4794, because `egui` receives only IME events during composition from now on

On Windows, “incompatible” events are filtered in `egui-winit`, aligning the behavior with other systems.

<details><summary>Screencasts</summary>

Some Chinese IMEs on Windows:
2026-03-13 12 25 37 AM

The default Japanese IMEs on Windows:
2026-03-13 12 28 33 AM
</details>

The 2-set Korean IMEs handle arrow keys differently. It will be discussed in the next section.

### This PR Reverts #4794, because it introduced several bugs

Some of its bugs have already been worked around in the past, but those workarounds might also be problematic. For example, #4912 is a workaround for a bug (#4908) introduced by #4794, and that workaround is in fact the root cause of the macOS backspacing bug I have worked around with #7810. (The reversion of #4912 is out of the scope of this PR, I will do that in #7983.)

#### It Caused: Arrow keys are incorrectly blocked during typical Korean IME composition

When composing Korean text using 2-Set IMEs, users should still be able to move the cursor with arrow keys regardless if the composition is committed.

##### Correct behavior

<details><summary>Screencasts</summary>

macOS TextEdit:
2026-03-12 8 04 15 PM

Windows Notepad:
2026-03-12 8 05 08 PM

With #4794 reverted, `egui` also behaves correctly (tested on Linux + Wayland, macOS, and Windows):
2026-03-12 8 03 51 PM
</details>

##### Incorrect behavior caused by #4794

`remove_ime_incompatible_events` removed arrow-key events in such cases. As a result, the first arrow key press only commits the composition, and users need to press the arrow key again to move the cursor:

<details><summary>Screencast</summary>

2026-03-12 8 06 40 PM
</details>

This is essentially the same issue described here: https://github.com/emilk/egui/pull/7877#issuecomment-3852719948

#### It Caused: Backspacing leaves the last character in Korean IME pre-edit text not removed on macOS

<details><summary>Screencasts</summary>

Before this PR:
2026-03-17 10 48 12 PM

After this PR:
2026-03-17 10 47 23 PM
</details>

### Korean IMEs also use <kbd>Enter</kbd> to confirm Hanja selections, and will not work properly in the Korean “IME mode” proposed by #7898 (#7914)

<details><summary>Screencast: Korean IME using <kbd>Enter</kbd> and <kbd>Space</kbd> for confirmation (IBus Korean Hangul IME)</summary>

The screencast below demonstrates that some Korean IMEs handle Hanja selection in a way similar to Japanese IMEs: the <kbd>Up</kbd>/<kbd>Down</kbd> arrow keys are used to navigate candidates, and <kbd>Enter</kbd> confirms the selected candidate.

2026-03-13 6 39 17 AM
</details>

<details><summary>Screencasts: Another example</summary>

Using the built-in Korean IME on Windows, I type two lines: the first line in Hangul, and the second line as the same word converted to Hanja.

Correct behavior in Notepad (reference):
7914-ref

Behavior after applying this PR, which matches the Notepad behavior:

Behavior after applying #7914 with the “IME mode” set to Korean (which is also the behavior before this PR being applied):

On the second line, each time a Hanja character is confirmed, an unintended newline is inserted. This mirrors the Japanese IME issues that are supposed to be fixed by setting the “IME mode” to Japanese.
(These Japanese IME issues are fixed in this PR as mentioned before.)

</details>


umajho:
Related `winit` issue: https://github.com/rust-windowing/winit/issues/4508

umajho:
Hello.
Since you were involved in the issues/PRs related to this PR, I'd appreciate it if you could take a look at this PR.
Sorry if the pinging annoyed you.

@choyunjin (Issue #7809)
@ClozedJP (Issue #7876)
@cats2333 (Issue #7908)
@sevenc-nanashi (PR #7877)
@rustbasic (PR #7898, PR #7914, and PR #4794)

sevenc-nanashi:
Tested on Windows 11 with Google Japanese IME, it's working very well, like other applications that supports IME. Nice work!

Speaking of implementation, as the behavior which winit sends keyboard events that is consumed by IME is [confirmed as a bug](https://github.com/rust-windowing/winit/issues/4508#issuecomment-4071935252), it would be nice to have a comment that says the implementation is workaround and will be fixed by winit.

crates/egui-winit/src/lib.rs
```diff
@@ -526,6 +545,71 @@ impl State {
...
+    #[cfg(target_os = "windows")]
+    #[inline(always)]
+    fn try_on_ime_processed_keyboard_input(
+        &mut self,
+        event: &winit::event::KeyEvent,
+        egui_wants_keyboard_input: bool,
+    ) -> Option<EventResponse> {
+        if !self.allow_ime {
+            None
+        } else if event.logical_key == winit::keyboard::NamedKey::Process {
```
lucasmerlin, line 572: It'd be awesome if you could add a comment explaining what NamedKey::Process is, because I had no idea.

crates/egui-winit/src/lib.rs
```diff
@@ -526,6 +545,71 @@ impl State {
...
+            })
+        } else if event.state == ElementState::Released
+            && self
+                .pressed_processed_physical_keys
+                .remove(&event.physical_key)
+        {
+            // Unlike key presses, we can not tell whether a key release event
+            // is processed by the IME or not by looking at its `logical_key`,
+            // so we track the physical keys of the processed key presses and
+            // filter out the corresponding key releases.
```
lucasmerlin: Why can't we? Is it's logical key not set to `Process`?

crates/egui-winit/src/lib.rs
umajho: Done. (https://github.com/emilk/egui/pull/7967/files#diff-964c0ccd7e970b17628c08c3df7315de84ee805a9c6346240f77b0e33224ae3bR581-R588)

crates/egui-winit/src/lib.rs
umajho: Right, it is the original logical key (e.g. `winit::keyboard::Key::Character(…)`) instead.

umajho:
I noticed that @rustbasic reacted with an 👀 to [the comment where I pinged people](https://github.com/emilk/egui/pull/7967#issuecomment-4058377012), so they may still be reviewing this PR. I think we should wait for their feedback before considering merging it.

rustbasic:
Great job, @umajho. 
This Pull Request #7967 is a solid replacement for #4794. 
It is useful for handling special key inputs during IME composition.

On a side note, I believe #7914 should still be pursued for separate reasons. 
It(#7914) specifically addresses and resolves issues related to the 'Cheonjiin' Korean keyboard layout. 
Furthermore, the `start` and `end` indices provided by `Preedit` are still necessary in certain contexts, so that PR remains relevant.

While not directly related to this (#7914 and #7967), there are still some unresolved issues regarding IME handling.
For instance, if the focus is moved to another `TextEdit` via a mouse click or the TAB key 
while an IME composition is active, the character currently being typed is incorrectly transferred to the newly selected `TextEdit`.
( Currently, this is being partially prevented by checking the `cursor_range`. Related #4137 )

umajho:
@rustbasic thanks for the feedback!

Could you elaborate more on the Cheonjiin issue Specifically:
- how to reproduce it, 
- what is the expected behavior,
- what is the actual behavior,
- and by any chance, this PR inadvertently fixed it?

## Regarding #4137

The `ime_cursor_range` it introduced led to #7485. Rather than adding more complexity by working around `ime_cursor_range`, I proposed in #7983 to replace #4137 with a simpler approach using `mem.had_focus_last_frame`.

I have tested #7983 on Windows and confirmed that the partial fix for the focus-related bug still works there. (And on macOS, the behavior remains incorrect as before.)

I guess the ultimate solution may involve explicitly disabling and re-enabling the IME when a `TextEdit` gains focus, though this is just a speculation.

umajho:
Regarding the `Preedit` cursor range:

I'm also planning to introduce this in a follow-up PR that implements visuals for “real IME cursors & cursor ranges.” My intention is to work on it after this PR and #7983 are resolved.

By “real IME cursors & cursor ranges,” I mean something like the following:
Screenshot 2026-03-20 at 9 54 00 PM
Screenshot 2026-03-20 at 9 54 56 PM

---

Since this discussion is already outside the scope of this PR, I suggest we continue it in #7983 and merge this PR first.

(Since I've gathered feedback from both a Japanese IME user and a Korean IME user that this PR works, and I'm confident it also works for Chinese IMEs (at least for those Pinyin-like), I believe this PR is ready to be merged.)

rustbasic:
> @rustbasic thanks for the feedback!
> 
> Could you elaborate more on the Cheonjiin issue Specifically:
> 
> * how to reproduce it,
> * what is the expected behavior,
> * what is the actual behavior,
> * and by any chance, this PR inadvertently fixed it?
> 
> ## Regarding #4137
> The `ime_cursor_range` it introduced led to #7485. Rather than adding more complexity by working around `ime_cursor_range`, I proposed in #7983 to replace #4137 with a simpler approach using `mem.had_focus_last_frame`.
> 
> I have tested #7983 on Windows and confirmed that the partial fix for the focus-related bug still works there. (And on macOS, the behavior remains incorrect as before.)
> 
> I guess the ultimate solution may involve explicitly disabling and re-enabling the IME when a `TextEdit` gains focus, though this is just a speculation.


**Regarding the Cheonjiin Korean Keyboard issue:**
The Cheonjiin layout issue was not resolved in #7967 and is not directly related to it. 
For certain Korean characters with double final consonants (batchim), such as "않", the IME occasionally needs to revert and delete characters even after a `Commit` has been sent. 
This is why the `start` and `end` indices from `Preedit` are essential.

*Example sequence:*
"안" -> 안"ㅎ" -> "않" -> 않"ㅇ" -> 않"아"
"안" -> 안"ㅎ" -> 안"해"

**Regarding #7983:**
While #7983 is effective when moving focus between different `TextEdit` elements on Windows, it also shows some effectiveness on WASM as well.
I believe #4137 should be maintained. If #4137 is removed, clicking a different position within the **same** `TextEdit` during composition causes the character to be incorrectly copied to that new position. Essentially, #4137 and #7983 address different problems: one handles issues within the same `TextEdit`, while the other handles transitions between different `TextEdit` elements.

**Regarding disabling and re-enabling IME on focus:**
The main challenge here is the non-deterministic timing of the `Commit` signal. 
Disabling it for just a single frame might not be sufficient; it might require a delay of about 0.1 seconds. It could be a viable temporary workaround, though.

rustbasic:
@umajho
In my opinion, it might be better not to modify #4137 in #7983. Instead, could you try setting `wants_ime_events` to `had_focus_last_frame` and resubmitting the PR? I’ll do some further testing in the meantime.

**Update from further testing:**
I have confirmed that the character copying issue is resolved on both **Windows and WASM** by combining #4137 and #7983.

umajho:
> [!NOTE]
>
> I moved the original content of this comment to https://github.com/emilk/egui/pull/7983#issuecomment-4099667969

I think we should continue our discussion in #7983, since we are now talking about what comes after this PR, which is what I originally planned for #7983.

cats2333:
Finally, a smooth typing experience in egui on Windows! Thank you for the extensive testing across different environments. You've solved a long-standing pain point for us : )

emilk:
Thank you!

PR #7973, opened 2026-03-13
Improve IME, and restrict mac-specific workaround

umajho:
* Closes N/A

My PR that fixes the macOS backspacing issue (#7810) unfortunately breaks text selection on Wayland (Fedora KDE Plasma Desktop 43 [Wayland, with or without IBus]). I had actually tested on a Wayland setup but failed to notice that :(

Windows and Linux+X11 (Debian 13 [Cinnamon 6.4.10 + X11 + fcitx5 5.1.2]) are not affected.

This PR fixes the issue by restricting the macOS fix to macOS-only.

<details><summary>Here is the correct behavior on Wayland after this PR (and before #7810 is applied)</summary>

2026-03-13 5 25 24 PM
</details>

<details><summary>Here is the buggy behavior on Wayland before this PR</summary>

2026-03-13 5 31 58 PM
</details>

## Cause of the Wayland issue

On Wayland, `winit` constantly emits `winit::event::Ime::Preedit("", None)` events.

PR #7810 added these lines for handling `winit::event::Ime::Preedit(_, None)` in `egui-winit` without considering the `target_os`:
https://github.com/emilk/egui/blob/14afefa2521d1baaf4fd02105eec2d3727a7ac36/crates/egui-winit/src/lib.rs#L619-L621

As a result, while text is being selected, `egui-winit` receives these `winit::event::Ime::Preedit("", None)` events from `winit` and forwards them to `egui` as `egui::ImeEvent::Preedit("")`. `egui` then clears the current text selection, because it currently does not distinguish between IME pre-edit text and selected text.

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -548,23 +548,23 @@ impl State {
     ///
     /// | Setup                                       | Events in Order                                                                                                                  |
     /// | ------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
-    /// | a-macos15-apple_shuangpin                   | `Predict("", None)` -> `Commit("测试")`                                                                                          |
-    /// | b-debian13_gnome48_wayland-fcitx5_shuangpin | `Predict("", None)` -> `Commit("测试")` -> `Predict("", Some(0, 0))` -> `Predict("", None)` (duplicate until `TextEdit` blurred) |
-    /// | c-windows11-ms_pinyin                       | `Predict("测试", Some(…))` -> `Predict("", None)` -> `Commit("测试")` -> `Disabled`                                              |
+    /// | a-macos15-apple_shuangpin                   | `Preedit("", None)` -> `Commit("测试")`                                                                                          |
+    /// | b-debian13_gnome48_wayland-fcitx5_shuangpin | `Preedit("", None)` -> `Commit("测试")` -> `Preedit("", Some(0, 0))` -> `Preedit("", None)` (duplicate until `TextEdit` blurred) |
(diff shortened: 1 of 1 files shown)


umajho:
Within the context of IMEs, the term “prediction” does not appear to be used elsewhere in the same sense as it is in egui.

The term is also ambiguous, since ”prediction” has already been defined for other things:
Screenshot 2026-03-15 at 8 25 40 PM

(“测试一下” is a prediction, since the IME suggested those four Chinese characters before the full input sequence `ceuiyixx` had been typed.)

Therefore, I replaced that terminology with “pre-edit text”, which is used by X11 and Wayland.

umajho:
Oh, good catch. I didn’t even realize that because of this I copied the winit event names incorrectly in the docs in that PR.

umajho:
Welp, it turned out, the macOS workaround that had accidentally applied to Wayland actually fixed this bug: #7485

<table>
<tr>
<th></th>
<th>before #7810 and after this PR</th>
<th>between #7810 and this PR</th>
</tr>
<tr>
<th>Text Selection</th>
<td>Works</td>
<td>Buggy: Selected texts are deleted</td>
</tr>
</tr>
<tr>
<th>Inputting ASCII characters with IBus</th>
<td>Buggy: see #7485</td>
<td>Works</td>
</tr>
</table>

I will investigate on this and submit another PR later to fix #7485.

I also found that normally `winit` on Wayland does not spam `winit::event::Ime::Preedit("", None)` and `winit::event::Ime::Disable`. Somehow egui makes it behave like this.

Update: The new PR is here: #7983

PR #7976, opened 2026-03-15
Allow fallback from smithay to arboard when getting clipboard

wizzeh:

Quick fix -- when the arboard and smithay features are both enabled, Clipboard::get returns early if it can't find a smithay clipboard. This PR just allows fallback to arboard instead of early-returning.

### DIFF
--- a/crates/egui-winit/src/clipboard.rs
+++ b/crates/egui-winit/src/clipboard.rs
@@ -70,7 +70,7 @@ impl Clipboard {
                 Err(err) => {
                     log::error!("smithay paste error: {err}");
                 }
-            };
+            }
         }
 
         #[cfg(all(


crates/egui-winit/src/clipboard.rs
```diff
@@ -65,11 +65,10 @@ impl Clipboard {
...
         if let Some(clipboard) = &mut self.smithay {
-            return match clipboard.load() {
-                Ok(text) => Some(text),
+            match clipboard.load() {
+                Ok(text) => return Some(text),
                 Err(err) => {
                     log::error!("smithay paste error: {err}");
-                    None
                 }
             };
```
emilk:
```suggestion
            }
```

PR #7986, opened 2026-03-18
Fix: `Visuals::interact_cursor` support in `Button`

### LINKED ISSUE #7947: REGRESSION: Visuals::interact_cursor doesn't affect anything anymore
In egui 0.33, it seems like Visuals::interact_cursor is no longer used by Button to set the cursor.

I think this logic got accidentally removed in commit 6eb7bb6e ?

mango766:
Closes #7947

## Problem

`Visuals::interact_cursor` stopped working for buttons after the `AtomLayout` refactor in commit 6eb7bb6e. Setting `interact_cursor` to e.g. `CursorIcon::PointingHand` no longer changes the cursor when hovering over a `Button`.

## Root Cause

When `Button` was rewritten to use `AtomLayout` in #5830, the cursor-override block at the end of the old `Button::ui` was not carried over to the new `Button::atom_ui` method.

The old code had:
```rust
if let Some(cursor) = ui.visuals().interact_cursor {
    if response.hovered() {
        ui.ctx().set_cursor_icon(cursor);
    }
}
```

This was the only place `interact_cursor` was checked, so the setting became entirely non-functional.

## Fix

Re-add the same `interact_cursor` check in `Button::atom_ui`, right after painting and before `widget_info`, matching the original behavior.


crates/egui/src/widgets/button.rs
```diff
@@ -359,6 +359,12 @@ impl<'a> Button<'a> {
             AtomLayoutResponse::empty(prepared.response)
         };
 
+        if let Some(cursor) = ui.visuals().interact_cursor {
+            if response.response.hovered() {
+                ui.ctx().set_cursor_icon(cursor);
+            }
+        }
```
emilk:
```suggestion
        if let Some(cursor) = ui.visuals().interact_cursor && response.response.hovered() {
            ui.ctx().set_cursor_icon(cursor);
        }
```

PR #7988, opened 2026-03-20
Make it more clear how a Ui Id was derived

lucasmerlin:
With this, we can see the full tree of how auto ids are derived:

image

### DIFF
--- a/crates/egui/src/ui.rs
+++ b/crates/egui/src/ui.rs
@@ -153,7 +153,7 @@ impl Ui {
         let mut ui = Ui {
             id,
             unique_id: id,
-            next_auto_id_salt: id.with("auto").value(),
+            next_auto_id_salt: 0,
             painter: Painter::new(ctx, layer_id, clip_rect),
             style,
             placer,
@@ -295,12 +295,10 @@ impl Ui {
             (id_salt, id_salt)
         } else {
             let stable_id = self.id.with(id_salt);
-            let unique_id = stable_id.with(self.next_auto_id_salt);
+            let unique_id = self.unique_id.with(id_salt).with(self.next_auto_id_salt);
 
             (stable_id, unique_id)
         };
-        let next_auto_id_salt = unique_id.value().wrapping_add(1);
-
         self.next_auto_id_salt = self.next_auto_id_salt.wrapping_add(1);
 
         let placer = Placer::new(max_rect, layout);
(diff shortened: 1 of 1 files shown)


emilk:
I'm not a big fan of that UI. It has A LOT of repetition (`Id(…)`, `Source: `, `Parent: `). It also uses a tree view for something that is really a linked list.

I also think showing all the Ids of all the ancestors is a bit overkill, though I can see how hovering them to find them in the UI would be useful.

So I suggest something like this:

Id(C4BF):
← `6`
← `0`
← `"child"`
← `0`
← `"child`"
← `0`
← `"child`"
← `"ID Test`"  (root)

where each row is hoverable, and shows the `Id` of that widget and where it is in the Ui (its `Rect`)

…actually, turning it upside down would look even better:

`"ID Test`" (root)
➝  `"child"`
➝  `0`
➝  `"child"`
➝  `0`
➝  `"child"`
➝  `0`
➝  `6`
= Id(C4BF)

crates/egui/src/ui.rs
```diff
@@ -1005,22 +1003,33 @@ impl Ui {
...
 
     /// Pretend like `count` widgets have been allocated.
+    #[inline]
     pub fn skip_ahead_auto_ids(&mut self, count: usize) {
         self.next_auto_id_salt = self.next_auto_id_salt.wrapping_add(count as u64);
     }
+
+    /// Get auto id and advance auto id counter by 1.
+    #[inline]
+    pub fn get_auto_id(&mut self) -> Id {
```
emilk, line 1028: `get` sounds like it doesn't change anything, which it does

```suggestion
    pub fn allocate_auto_id(&mut self) -> Id {
```

emilk (approved):
Nice!

lucasmerlin:
> I'm not a big fan of that UI. It has A LOT of repetition (`Id(…)`, `Source: `, `Parent: `). It also uses a tree view for something that is really a linked list.

But it is indeed a tree. When you create an Id like this: `Id::new(Id::new("a")).with(Id::new("b"))` the id has two parents (a and b). 
I think the ideal way to disaplay these would be a graph view, but unfortunately egui doesn't have one yet. 

I guess we could show the source with the debug format instead, since source _usually_ is pretty flat. But if you e.g. pass a widget id as a source via `with(response.id)`, the source tree would be huge.

PR #7990, opened 2026-03-20
Update to wgpu 29

cwfitzgerald:

This updates wgpu to v29 across the egui crate stack.

There a a few API changes due to the requirement to provide a display handle up front to properly support GLES on linux. I have done my best to make the api changes as reasonable as possible, but I don't have all the greater project context, so lmk if things should be done a bit differently.

I've also updated glow to 0.17 to make cargo deny happy, there are no source changes. I'm not sure how you want to land these.

### DIFF
--- a/crates/eframe/src/web/web_painter_wgpu.rs
+++ b/crates/eframe/src/web/web_painter_wgpu.rs
@@ -22,6 +22,7 @@ pub(crate) struct WebPainterWgpu {
     capture_tx: CaptureSender,
     capture_rx: CaptureReceiver,
     ctx: egui::Context,
+    needs_reconfigure: bool,
 }
 
 impl WebPainterWgpu {
@@ -110,6 +111,7 @@ impl WebPainterWgpu {
             capture_tx,
             capture_rx,
             ctx,
+            needs_reconfigure: false,
         })
     }
 }
@@ -195,11 +197,16 @@ impl WebPainter for WebPainterWgpu {
                 );
             }
 
+            if self.needs_reconfigure {
+                self.surface
+                    .configure(&render_state.device, &self.surface_configuration);
+                self.needs_reconfigure = false;
+            }
+
             let output_frame = match self.surface.get_current_texture() {
                 wgpu::CurrentSurfaceTexture::Success(frame) => frame,
                 wgpu::CurrentSurfaceTexture::Suboptimal(frame) => {
-                    self.surface
-                        .configure(&render_state.device, &self.surface_configuration);
+                    self.needs_reconfigure = true;
                     frame
                 }
                 other => {
--- a/crates/egui-wgpu/src/lib.rs
+++ b/crates/egui-wgpu/src/lib.rs
@@ -307,8 +307,9 @@ pub struct WgpuConfiguration {
     ///
     /// Called with the [`wgpu::CurrentSurfaceTexture`] result whenever acquiring a frame
     /// does not return [`wgpu::CurrentSurfaceTexture::Success`]. For
-    /// [`wgpu::CurrentSurfaceTexture::Suboptimal`], egui automatically reconfigures the
-    /// surface and uses the frame — the callback is not invoked in that case.
+    /// [`wgpu::CurrentSurfaceTexture::Suboptimal`], egui uses the frame as-is and
+    /// defers surface reconfiguration to the next frame — the callback is not invoked
+    /// in that case either.
     pub on_surface_status:
         Arc<dyn Fn(&wgpu::CurrentSurfaceTexture) -> SurfaceErrorAction + Send + Sync>,
 }
--- a/crates/egui-wgpu/src/winit.rs
+++ b/crates/egui-wgpu/src/winit.rs
@@ -17,6 +17,7 @@ struct SurfaceState {
     width: u32,
     height: u32,
     resizing: bool,
+    needs_reconfigure: bool,
 }
 
 /// Everything you need to paint egui with [`wgpu`] on [`winit`].
@@ -234,6 +235,7 @@ impl Painter {
                 height: size.height,
                 alpha_mode,
                 resizing: false,
+                needs_reconfigure: false,
             },
         );
         let Some(width) = NonZeroU32::new(size.width) else {
@@ -454,7 +456,7 @@ impl Painter {
             commands_submitted: false,
         };
 
-        let Some(surface_state) = self.surfaces.get(&viewport_id) else {
+        let Some(surface_state) = self.surfaces.get_mut(&viewport_id) else {
             return vsync_sec;
         };
 
@@ -491,6 +493,11 @@ impl Painter {
             )
         };
 
+        if surface_state.needs_reconfigure {
+            Self::configure_surface(surface_state, render_state, &self.configuration);
+            surface_state.needs_reconfigure = false;
+        }
+
         let output_frame = {
             profiling::scope!("get_current_texture");
             // This is what vsync-waiting happens on my Mac.
@@ -503,7 +510,7 @@ impl Painter {
         let output_frame = match output_frame {
             wgpu::CurrentSurfaceTexture::Success(frame) => frame,
             wgpu::CurrentSurfaceTexture::Suboptimal(frame) => {
-                Self::configure_surface(surface_state, render_state, &self.configuration);
+                surface_state.needs_reconfigure = true;
                 frame
             }
             other => {


cwfitzgerald:
CI is clean, so should be good to review

crates/eframe/src/native/wgpu_integration.rs
```diff
@@ -184,9 +184,17 @@ impl<'app> WgpuWinitApp<'app> {
...
     ) -> crate::Result<&mut WgpuWinitRunning<'app>> {
         profiling::function_scope!();
+        // Inject the display handle into the wgpu setup so that wgpu can create
+        // surfaces on platforms that require it (e.g. GLES on Wayland).
+        let mut wgpu_options = self.native_options.wgpu_options.clone();
+        if let egui_wgpu::WgpuSetup::CreateNew(ref mut create_new) = wgpu_options.wgpu_setup
+            && create_new.display_handle.is_none()
+        {
+            create_new.display_handle = Some(Box::new(event_loop.owned_display_handle()));
+        }
```
Wumpf, line 194: ah that's nice to catch that here. Sad to have defacto platform dependent stuff to leak in like that, but at least it's somewhat elegant and failsafe

crates/egui-wgpu/src/lib.rs
```diff
@@ -272,7 +276,58 @@ fn describe_adapters(adapters: &[wgpu::Adapter]) -> String {
     }
 }
 
-/// Specifies which action should be taken as consequence of a [`wgpu::SurfaceError`]
+/// Describes a surface error when acquiring a texture for rendering.
+///
+/// These correspond to the error variants of [`wgpu::CurrentSurfaceTexture`] — everything
+/// except `Success` and `Suboptimal`, which contain usable frames. This enum is passed to the
+/// [`WgpuConfiguration::on_surface_error`] callback so you can decide how to respond.
```
Wumpf: we should still do a refconfigure for suboptimal. On Rerun we got a report recently where this keeps going and eventually causes hangs rerun-io/rerun#12681
So what I'd really like to do here is to tell egui to reconfigure next frame

.. that said, this isn't a new problem

crates/egui-wgpu/src/setup.rs
```diff
@@ -1,5 +1,47 @@
...
 
+/// A cloneable display handle for use with [`wgpu::InstanceDescriptor`].
+///
+/// This trait exists so that a [`winit::event_loop::OwnedDisplayHandle`] (or similar platform
+/// display handle) can be stored, cloned, and later passed to wgpu.
+///
+/// wgpu requires an explicit display handle for GLES on some platforms (notably Wayland).
+/// Because [`wgpu::InstanceDescriptor`] contains a `Box<dyn WgpuHasDisplayHandle>` which is
+/// not cloneable, we wrap the handle in this trait so it can be cloned alongside the rest of
+/// the egui wgpu configuration.
```
Wumpf, line 11: all this hackery is rather unfortunate, missed that we introduced this. Before we delayed more things until we get a surface iirc?

crates/egui-wgpu/src/setup.rs
```diff
@@ -65,8 +130,18 @@ impl WgpuSetup {
...
+                let mut descriptor = wgpu::InstanceDescriptor {
+                    backends: desc.backends,
+                    flags: desc.flags,
+                    backend_options: desc.backend_options.clone(),
+                    memory_budget_thresholds: desc.memory_budget_thresholds,
+                    display: None,
+                };
+                if let Some(handle) = &create_new.display_handle {
+                    descriptor.display = Some(handle.clone_for_wgpu());
+                }
```
Wumpf: nit:
```suggestion
                    display: create_new.display_handle.map(|handle| handle.clone_for_wgpu())),
                };
```

crates/egui-wgpu/src/lib.rs
Wumpf: Actually I think we should do this differently: let's simply change the callback to `on_surface_status` and pass the original wgpu enum through by reference. Rolling all this stuff out her ein egui(-wgpu) is way too much duplication.

crates/eframe/src/web/web_painter_wgpu.rs
```diff
@@ -196,17 +196,58 @@ impl WebPainter for WebPainterWgpu {
...
+                wgpu::CurrentSurfaceTexture::Validation => {
+                    match (*self.on_surface_error)(SurfaceStatus::Validation) {
+                        SurfaceErrorAction::RecreateSurface => {
+                            self.surface
+                                .configure(&render_state.device, &self.surface_configuration);
+                        }
+                        SurfaceErrorAction::SkipFrame => {}
                     }
-                },
+                    return Ok(());
```
Wumpf: this is same code 5 times. That's rather silly - see also comment below on how we shouldn't have `SurfaceStatus` in egui

crates/egui-wgpu/src/winit.rs
```diff
@@ -501,16 +501,53 @@ impl Painter {
...
+                return vsync_sec;
+            }
+            wgpu::CurrentSurfaceTexture::Validation => {
+                match (*self.configuration.on_surface_error)(SurfaceStatus::Validation) {
+                    SurfaceErrorAction::RecreateSurface => {
+                        Self::configure_surface(surface_state, render_state, &self.configuration);
+                    }
+                    SurfaceErrorAction::SkipFrame => {}
+                }
+                return vsync_sec;
```
Wumpf: as on web, I think we can do better

.github/workflows/rust.yml
```diff
@@ -94,7 +94,7 @@ jobs:
       - name: wasm-bindgen
         uses: jetli/wasm-bindgen-action@v0.1.0
         with:
-          version: "0.2.100" # Keep wasm-bindgen version in sync in: setup_web.sh, Cargo.toml, Cargo.lock, rust.yml
+          version: "0.2.114" # Keep wasm-bindgen version in sync in: setup_web.sh, Cargo.toml, Cargo.lock, rust.yml
```
emilk: Let's keep this on 0.2.108 instead, because updating wasm-bindgen in Rerun is pain 😭 

Context: 
https://github.com/rerun-io/rerun/blob/e4f48fd75dba9a963a6cd471161f1e0a6f80f45c/Cargo.toml#L417-L428

cwfitzgerald:
Thanks for taking care of this! Please ping me if you need something

crates/eframe/src/web/web_painter_wgpu.rs
```diff
@@ -196,17 +196,22 @@ impl WebPainter for WebPainterWgpu {
...
-                            .configure(&render_state.device, &self.surface_configuration);
-                        return Ok(());
-                    }
-                    SurfaceErrorAction::SkipFrame => {
-                        return Ok(());
+                wgpu::CurrentSurfaceTexture::Success(frame) => frame,
+                wgpu::CurrentSurfaceTexture::Suboptimal(frame) => {
+                    self.surface
+                        .configure(&render_state.device, &self.surface_configuration);
+                    frame
```
Wumpf: so there's something really nasty going on here unfortunately and I almost made the same mistake on wgpu examples themselves: https://docs.rs/wgpu/latest/wgpu/struct.Surface.html#method.configure is documented to panic if old frames are still around:
```
Panics
    * A old [SurfaceTexture](https://docs.rs/wgpu/latest/wgpu/struct.SurfaceTexture.html) is still alive referencing an old surface.
    * Texture format requested is unsupported on the surface.
    * config.width or config.height is zero.
```

so we either shortterm do the same thing as before and don't reconfigure on Suboptimal, or we do it with a frame delay before getting the next frame (best?), or we discard the frame and try to get a new texture but make sure we don't endlessly recurse

practically all this is only relevant on native. Don't think webgl or webgpu can ever run into this. (but this is simply where I started reviewing)

crates/egui-wgpu/src/lib.rs
```diff
@@ -299,8 +303,14 @@ pub struct WgpuConfiguration {
...
     pub wgpu_setup: WgpuSetup,
 
-    /// Callback for surface errors.
-    pub on_surface_error: Arc<dyn Fn(wgpu::SurfaceError) -> SurfaceErrorAction + Send + Sync>,
+    /// Callback for surface status changes.
+    ///
+    /// Called with the [`wgpu::CurrentSurfaceTexture`] result whenever acquiring a frame
+    /// does not return [`wgpu::CurrentSurfaceTexture::Success`]. For
+    /// [`wgpu::CurrentSurfaceTexture::Suboptimal`], egui automatically reconfigures the
+    /// surface and uses the frame — the callback is not invoked in that case.
```
Wumpf: nit: clarifying it a bit
```suggestion
    /// surface and uses the frame — the callback is not invoked in that case either.
```

crates/egui-wgpu/src/winit.rs
```diff
@@ -501,16 +501,20 @@ impl Painter {
...
-                    Self::configure_surface(surface_state, render_state, &self.configuration);
-                    return vsync_sec;
-                }
-                SurfaceErrorAction::SkipFrame => {
-                    return vsync_sec;
+            wgpu::CurrentSurfaceTexture::Success(frame) => frame,
+            wgpu::CurrentSurfaceTexture::Suboptimal(frame) => {
+                Self::configure_surface(surface_state, render_state, &self.configuration);
+                frame
+            }
```
Wumpf, line 515: as above, this can cause a panic
(and no, I don't know what steps to do to get into this situation. probably involves wayland + vulkan)

umajho:
Hello. It appears that this PR has broken browsers that use the `Gl` backend: #8001.

I have no idea how to fix it, so I'd appreciate it if the people who involved in this PR could take a look.

Thanks!

umajho:
Sorry to bother you again.
I've found another issue that appears to have been introduced by this PR: #8012.

PR #7993, opened 2026-03-21
Update ObjC2

oscargus:
Running this separately from the other updates as I am not at my mac at the moment, so cannot check  locally that it is working...

Most of the eco system has updated to objc2 0.6 (and objc2-* 0.3), so there is already now a mix. Still better for egui to go with latest version (if it works...).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -281,7 +281,7 @@ dependencies = [
  "clipboard-win",
  "image",
  "log",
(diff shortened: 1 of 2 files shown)


oscargus:
Was just about to write that winit is the main crate that is still with the old version, but since it doesn't work (which I interpret as the versions should be the same), I'll close this for now.

PR #7996, opened 2026-03-23
Update wasm-bindgen to 0.2.108, and ehttp to 0.7.1

emilk:


emilk:
This bumps wasm-bindgen by a lot of versions, which was not intented. I'll make a 0.7.1 release that relaxes the wasm-bindgen version

Issue #8001, opened 2026-03-24
web+wgpu: eframe is broken for browsers requiring `Gl` backend on main (fbe5763)


umajho:
**Describe the bug**

On certain browsers, opening an eframe web app results in the following error:

```
An error occurred during loading:

failed to create wgpu surface: No `DisplayHandle` is available to create this surface with. When creating a surface with `create_surface()` you must specify a display handle in `InstanceDescriptor::display`. Rarely, if you need to create surfaces from different `DisplayHandle`s (ex. different Wayland or X11 connections), you must use `create_surface_unsafe()`.

Make sure you use a modern browser with WebGL and WASM enabled.
```

Known affected browsers:
- Type A - WebGPU is not supported:
  - Safari 26.3 on macOS 15.7.3, `navigator.gpu === undefined`.
  - Firefox 148 (Zen 1.19.3b) on macOS 15.7.3, when WebGPU is disabled (`navigator.gpu === undefined`).
  - Firefox 140.8.0esr on Debian 13 [Cinnamon 6.4.10 + X11] (Parallels 26 VM), when WebGPU is disabled (`navigator.gpu === undefined`).
  - Firefox 143.0.3 on Fedora KDE Plasma Desktop 43 [Wayland] (Parallels 26 VM), when WebGPU is disabled (`navigator.gpu === undefined`).
- Type B - WebGPU is supported:
  - Microsoft Edge 146 on Windows 11 25H2 (Parallels 26 VM), `navigator.gpu !== undefined`.
  - Firefox 140.8.0esr on Debian 13 [Cinnamon 6.4.10 + X11] (Parallels 26 VM), when WebGPU is enabled (`navigator.gpu !== undefined`).

Known unaffected browsers:
- Chrome 146 on macOS 15.7.3, `navigator.gpu !== undefined`.
- Firefox 148 (Zen 1.19.3b) on macOS 15.7.3, when WebGPU is enabled (`navigator.gpu !== undefined`).
  - Firefox 143.0.3 on Fedora KDE Plasma Desktop 43 [Wayland] (Parallels 26 VM), when WebGPU is enabled (`navigator.gpu !== undefined`).

**To Reproduce**
1. with an affected browser, either:
   - checkout to `main` (fbe5763a91415a6e4277330718bc67c4c1f853b0) & `./scripts/build_demo_web.sh` & `./scripts/start_server.sh` & visit `localhost:8765`, or
   - just visit <https://egui.rs> as of now.

**Expected behavior**
[...]

PR #8006, opened 2026-03-24
Fill in DisplayHandle automatically on web painter just like it's done on winit

Wumpf:
* Closes https://github.com/emilk/egui/issues/8001
* Follow-up to https://github.com/emilk/egui/pull/7990

Also simplified/shortened the comments around display handle a bit. Lots a repetition there made it hard to upgrade otherwise.

### DIFF
--- a/crates/eframe/src/web/web_painter_wgpu.rs
+++ b/crates/eframe/src/web/web_painter_wgpu.rs
@@ -25,6 +25,24 @@ pub(crate) struct WebPainterWgpu {
     needs_reconfigure: bool,
 }
 
+/// Owned web display handle that is `Send + Sync`.
+///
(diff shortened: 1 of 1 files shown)


Wumpf:
cc: @cwfitzgerald

Wumpf:
Bit weird though tbh given that the `DisplayHandle` on web isn't doing anything, see https://docs.rs/raw-window-handle/latest/src/raw_window_handle/web.rs.html . Is this defacto a bug in wgpu's  GL implementation to still require it on WebGL and not just desktop GL? 🤔 

either way, this is just consistent with eframe's winit integration 🤷

umajho:
Can confirm the preview works on those affected browsers I mentioned in #8001.

Thank you a lot!

PR #8007, opened 2026-03-24
Add raw key methods to TypeIdMap

AlexanderSchuetz97:
This PR fundamentally solves the same problem as https://github.com/emilk/egui/pull/5827 just implemented with a lot less ambition and api surface on my end. It contains the bare minimum amount of changes that I need in order to be able to solve my problem.

My Problem:
I am still suffering from the problem that the TypeIdMap blows up over a very long time when using my application. (The user generally never turns off the application, it is intended to be just kept running forever, some users also never restart their computers) My application generates a lot of content dynamically so it may for some time display widgets with a certain set of TypeId's + Id's later hiding them. Some of the elements that were hidden may turn visible again once an external event occurs, some may forever be discarded.  
I do know myself when which sections of the UI have to be purged because they will never become visible again, so this PR contains the minimum amount of necessary functions that allow me to implement this housekeeping logic on my end. 

The existing facilities are insufficient to handle this as the type T which the TypeId and the hash is derived from are sometimes pub(crate) privates of widget subcrates or even pub(crate) of egui internals itself, so its impossible to manually remove those from the TypeIdMap the only build-in method to remove them is to call "clear" on the TypeIdMap, however that gets rid of everything, even the elements that are still shown and should still be in the TypeIdMap.

If the changes in this PR are agreeable to you and you want me to write unit test for the 4 functions that I have added then tell me and I will write the tests for you.

If you need anything else changed please tell me.

I ran cargo clippy and cargo fmt, but your check.sh does not work on my computer.


crates/egui/src/util/id_type_map.rs
```diff
@@ -373,27 +401,68 @@ impl Default for IdTypeMap {
...
+
+        None
+    }
+
+    /// Returns all `RawKey`s to values in this map.
+    /// The returned keys can only be used with this map.
+    pub fn raw_keys(&self) -> impl Iterator<Item = RawKey> {
+        self.map
+            .iter()
+            .filter(|(_, v)| matches!(v, Element::Value { .. }))
```
emilk: Why only `Value`s?

emilk: Oh I see, your entire thing is only about non-serialized values ("temps") 🤔

emilk:
Take a look and see if it still satisfies your needs

AlexanderSchuetz97:
Thank you, yes this will still solve my problem.

It looks like you covered some edge cases I was not aware of in regards to the persistence feature.

I like that insertion now returns the raw key, 
I don't need that myself yet, but maybe that is useful for someone else.

Merge this at your convenience.

AlexanderSchuetz97:
I clicked "update branch" since you approved it already I thought perhaps I can merge it. I have no clue what that did, I will leave merging this PR to you as it appears that I cannot merge even tho you approved it.

PR #8010, opened 2026-03-24
Add `Response::parent_id` and improve `warn_if_rect_changes_id`

lucasmerlin:
Reduces the amount of false positives

### DIFF
--- a/crates/egui/src/response.rs
+++ b/crates/egui/src/response.rs
@@ -1,10 +1,10 @@
 use std::{any::Any, sync::Arc};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


crates/egui/src/response.rs
```diff
@@ -51,6 +51,9 @@ pub struct Response {
     /// (that is handled by the `Painter` directly).
     pub sense: Sense,
 
+    /// The [`Id`] of the parent [`crate::Ui`] that hosts this widget.
+    pub(crate) parent_id: Id,
```
emilk: I'm hesitant making `Response` bigger, as copying it is already on the hot-path for so many things. Can this instead be a member function that does a lookup with `Contexta`?

PR #8013, opened 2026-03-24
Respect `WidgetVisuals::expansion` in TextEdit 

lucasmerlin:
This broke in #7587

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -708,7 +708,8 @@ impl TextEdit<'_> {
                     .fill(background_color)
                     .corner_radius(corner_radius)
                     .inner_margin(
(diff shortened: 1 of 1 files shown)


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -707,7 +707,11 @@ impl TextEdit<'_> {
                     .frame
                     .fill(background_color)
                     .corner_radius(corner_radius)
-                    .inner_margin(allocated.frame.inner_margin - Margin::same(stroke.width as i8))
+                    .inner_margin(
+                        allocated.frame.inner_margin + Margin::same(visuals.expansion as i8)
+                            - Margin::same(stroke.width as i8),
```
emilk: avoid premature precision loss

```suggestion
                        allocated.frame.inner_margin + Margin::same((visuals.expansion - stroke.width).round() as i8),
```

PR #8018, opened 2026-03-25
Fade out the edges of `ScrollAreas`

emilk:
## Before:
image

It is very hard here to realize this is a scrollable area

## After
image

The fade at the bottom tells the user they should try scrolling.

You can control it with `style.spacing.scroll.fade`

### DIFF
--- a/crates/egui/src/ui_stack.rs
+++ b/crates/egui/src/ui_stack.rs
@@ -256,7 +256,7 @@ impl UiStack {
         !self.info.frame.stroke.is_empty()
     }
 
-    /// The background color of this [`Ui`].
+    /// The background color of this [`crate::Ui`].
     ///
     /// This blend together all [`Frame::fill`] colors
     /// up to the root.


crates/egui/src/containers/scroll_area.rs
```diff
@@ -1504,3 +1508,89 @@ impl Prepared {
...
+    let crate::style::ScrollFadeStyle {
+        enabled,
+        size: fade_size,
+    } = ui.spacing().scroll.fade;
+
+    if !enabled {
+        return;
+    }
+
+    let bg = ui.visuals().panel_fill; // a bit iffy
```
lucasmerlin: I think we need a better way to solve this if we enable it by default

lucasmerlin: Maybe we could do this using UiStack? Frame could set a background color property there and we could read it in the scroll area? That feels fairly simple and relatively low overhead

emilk: Nice idea! I'll investigate that

emilk: We already have `UiStackInfo::frame`, which is set by `Frame`, so I think we have all we need

emilk: * https://github.com/emilk/egui/pull/8020

PR #8020, opened 2026-03-25
Add `UiStack::bg_color`

emilk:
This lets you ask for the background color of a ui with `ui.stack().bg_color()`

### DIFF
--- a/crates/egui/src/ui_stack.rs
+++ b/crates/egui/src/ui_stack.rs
@@ -267,7 +267,7 @@ impl UiStack {
             let fill = node.frame().fill;
             if fill.is_opaque() {
                 return fill;
-            } else {
+            } else if fill != Color32::TRANSPARENT {
                 total = fill.blend(total);
             }
         }


crates/egui/src/ui_stack.rs
```diff
@@ -253,6 +255,24 @@ impl UiStack {
...
+        let mut total = Color32::TRANSPARENT;
+        for node in self.iter() {
+            let fill = node.frame().fill;
+            if fill.is_opaque() {
+                return fill;
+            } else if fill != Color32::TRANSPARENT {
+                total = fill.blend(total);
+            }
+        }
+        total
```
lucasmerlin, line 274: I think theres a bug here, if one if the parent layers is opaque, and the current one is semi transparent, we would disregard the transparent color and just return the opaque one without blending

lucasmerlin: https://github.com/emilk/egui/pull/8021

Issue #7999, opened 2026-03-23
Regression: #7587 breaks `TextEdit.horizontal_align()`


RndUsr123:
**Describe the bug**
Pretty much what the title says, `TextEdit` is now forcefully left-aligned.

**Screenshots**

After the mentioned PR + alignment fix (#7831):

![Image](https://github.com/user-attachments/assets/784405d4-9cdc-44da-94bd-5b187beb09ca)

without:

![Image](https://github.com/user-attachments/assets/101fa446-c50b-46ed-ac11-9ded5580cfdb)

Previous behavior with alignment fix:

![Image](https://github.com/user-attachments/assets/1a71b8cc-34dd-4b81-8b78-45813e5eeac1)

Previous behavior without fix:

![Image](https://github.com/user-attachments/assets/da5654e9-3e6b-4fb2-ab37-f77c510b9681)

PR #8022, opened 2026-03-25
Improves the Atom alignment in TextEdit

RndUsr123:
* Closes https://github.com/emilk/egui/issues/7999

This seems to work in most scenarios but there's 2 things I'm not fully sure about:

1) When only a prefix or suffix is used, horizontally-centered text is not really centered. This doesn't always look great but ensures all the available space is used so I'm not sure what the best approach here is.
2) Since atoms can't currently be vertically aligned, hint text and prefix/suffix might jump down after some text is entered when `vertical_align` other than `Align::TOP` is used. There's possible workarounds, but this might work for the time being if vertical alignment support in atoms is planned for the near future

Screenshots:
screenshot1
screenshot2
screenshot3
screenshot4

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -646,7 +646,7 @@ impl TextEdit<'_> {
                         }
                     })
                     .atom_id(inner_rect_id)
-                    .atom_shrink(clip_text),
+                    .atom_shrink(true),
                 );
             }
 


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -575,6 +575,11 @@ impl TextEdit<'_> {
                 atoms.push_right(atom);
             }
 
+            // Push prefix to the left and enable alignment for the main text
+            if align.0[0] != Align::LEFT {
```
```suggestion
            if align.x() != Align::LEFT {
```

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -647,8 +650,8 @@ impl TextEdit<'_> {
                 );
             }
 
-            // Ensure the suffix is always right-aligned
-            if !suffix.is_empty() {
+            // Push suffix to the right and enable alignment for the main text
+            if align.0[0] != Align::RIGHT {
```
```suggestion
            if align.x() != Align::RIGHT {
```

emilk:
This existing snapshot tests doesn't seem to cover this fix?

RndUsr123:
Not sure. Might it be because the current snapshot test is based on the current commit, which uses left-alignment regardless of what `.horizontal_align` is set to?

The compiled demo definitely looks correct visually and I'm getting no error on my machine after refreshing the snapshots...

Edit: indeed, there's something wrong. I've run all tests on my machine and got a `STATUS_ACCESS_VIOLATION` error on different tests at different times. Thing is, it also happens on my machine (Win10) on the latest commit without this PR.

RndUsr123:
There's definitely something off, but I'm noticing the opposite: `TextEdit`s added through `add_sized` are larger than intended though not very consistently so. In my case it's 60 -> 66, 100 -> 108, 140-> 141 and 200 -> 201. This is in screen pixels at 200% scaling (sizes in egui were 30, 50, 70 and 100).

Not sure what may be causing this, but I think there's a good chance there's an underlying problem with how `Atom` sizes are computed, possibly with `Atom::grow()` specifically

update: there seems to be an issue how atoms fill the pre-determined TextEdit area in general, not just the extra padding I added. prefix and suffix alone push the TextEdit out of bounds and increase its size from what's supposed to be 60px to more than 100 Look:
image

update 2: interestingly, hint_text without padding produces the correct final size (with the exception of the 30px one due to minimum size of  a single character + suffix and prefix being 43), so there's likely a problem with how the text galley is turned into an atom rather than the general layouting of the TextEdit. Padding is, however, problematic in its own right too because the right one adds an extra 4px (43 -> 47) to the minimum size required.

I'll see if I can find a fix.

lucasmerlin:
Opened https://github.com/emilk/egui/pull/8026 to address this, it should fix the spacing issue

RndUsr123:
Yeah, defaulting grow to ignoring spacing is what I was about to suggest as well.

Now there's the second issue, which is the hint taking up the correct amount of space while the text doesn't:
image
image

Logically I'd expect the same text to take up the same amount of space, but that doesn't seem to be the case

Edit: while the main text uses clip_text to determine whether its atom is allowed to shrink or not, that's not the case for the hint, which appears to be able to shrink regardless. My hunch is shrink should always be enabled by default for multiline text since even the docs suggest it only works for singleline. Is it a bug it's not triggered in this instance?

RndUsr123:
This seemingly fixes the issue of the main text not wrapping correctly and apparently doesn't affect clip_text, which still works in single line if no size is manually definied.

PR #8029, opened 2026-03-26
Add font variations example and fix cache bug in variations api

lucasmerlin:
The font variation wasn't part of the cache key, meaning if you had the same char with different variations on screen, or if variations changed over time, only the first of the variation settings would get used.

Also adds a nice font variations example using the [recursive](https://fonts.google.com/specimen/Recursive?preview.script=Latn) font:

https://github.com/user-attachments/assets/63cb80d9-e265-4949-af69-65c1687acc0c


emilk:
=> https://github.com/emilk/egui/pull/8189

PR #8031, opened 2026-03-26
Use `harfrust` for better kerning and ligatures

gcailly:
* Related to #56 (Improve text — tracking issue)

## Summary

This PR integrates [harfrust](https://crates.io/crates/harfrust) (a pure-Rust port of HarfBuzz) into epaint's text layout pipeline, replacing the character-by-character glyph positioning with proper OpenType text shaping.

### What this enables

- **GPOS kerning**: most modern fonts only ship kerning in GPOS tables (not the legacy `kern` table). Pairs like "AV", "VA", "AT" are now properly tightened.
- **GSUB substitutions**: ligatures (fi, fl), contextual alternates, and other OpenType features.
- **Combining marks**: diacritics (e.g. ɔ̃) are positioned via anchor tables instead of being rendered as standalone replacement glyphs.

#### Kerning, etc.

before_main

after_harfrust

 #### Ligatures

before_closeup

after_closeup

### Architecture

The shaping integrates into the existing pipeline without changing the public API:

1. **`Font::segment_into_runs`** — segments text into contiguous runs by font face (grapheme-cluster aware, never splits combining sequences)
2. **`FontFace::shape_text`** — calls harfrust to shape each run, returning glyph IDs + positioned advances/offsets
3. **`layout_shaped_run`** — emits `Glyph` structs from the shaping output, with NOTDEF fallback to other font faces for missing glyphs
4. **Buffer recycling** — `FontsImpl` pools a `harfrust::UnicodeBuffer` to avoid per-layout allocations

### Disclaimer

I'm far from being a good Rust programmer. Claude Code did most of the heavy lifting here. I did my best and used my limited knowledge to avoid making too many mistakes. If this PR isn't up to quality standards, please don't hesitate to close it.

## Test plan

- [x] `cargo test -p epaint` — all 18 text tests pass, including 6 new ones
- [x] `cargo clippy -p epaint --all-features` — clean
- [x] `cargo fmt` — clean
- [ ] Snapshot tests need regeneration (expected: shaping changes glyph positions)
- New tests added:
  - `test_gpos_kerning` — verifies GPOS kerning tightens "AV", "VA", "AT" pairs
  - `test_combining_diacritics` — combining tilde doesn't add extra width
  - `test_shaping_basic_latin` — sanity check for Latin text
  - `test_shaping_empty_string` — empty input doesn't panic
  - `test_shaping_multiple_newlines` — newline splitting works correctly
  - `test_shaping_mixed_font_fallback` — Latin + emoji in same string


emilk:
Wow, very cool! I haven't looked at the code yet but:

* I can see [on kitdiff](https://rerun-io.github.io/kitdiff/?url=https://github.com/emilk/egui/pull/8031) that the `image_kerning` test have regressed: thin spaces are now too thick, and `\t` doesn't seem to work at all
* I would it if you added some new snapshot tests with the images in your PR description
* I wonder what @valadaptive thinks :)

gcailly:
> * I can see [on kitdiff](https://rerun-io.github.io/kitdiff/?url=https://github.com/emilk/egui/pull/8031) that the `image_kerning` test have regressed: thin spaces are now too thick, and `\t` doesn't seem to work at all

The previous code had a custom override for U+2009 (thin space) that clamped the advance width to `min(em/6, space_width * 0.5)`. With harfrust, the advance width now comes directly from the font's own metrics for the thin space glyph. I think this is actually more correct. If the font defines a thin space with a specific width, egui should respect it rather than override it. What do you think ? Should we keep the old behavior or trust the font ?

As for \t, I agree with you : tab stops are a layout-level concept that no shaper handles. I'll add a post-shaping override to restore the `TAB_SIZE * space_width` behavior.

> * I would it if you added some new snapshot tests with the images in your PR description

About adding new snapshot tests for the shaping demos (ligatures, kerning), there's a difficulty : the bundled Ubuntu font doesn't have visible ligatures. Would it be acceptable to embed a test-only font in the test fixtures to demonstrate these features ?

> * I wonder what @valadaptive thinks :)

I saw that @valadaptive put a lot of work into the Parley integration and concluded it wasn't viable. I hope she finds this approach correct. Moreover, I'm a little embarrassed to be using AI without fully understanding what I'm doing, even though I have good intentions and take many precautions to do things correctly. I hope I'm not wasting your time.

oscargus:
Can you possibly consider updating skrifa and vello_cpu as well? These will change some of the test images and it may be just as well to do all in one go...

gcailly:
Thanks for the suggestion! I've bumped `vello_cpu` from 0.0.6 to 0.0.7 and updated all snapshots (141 files).

However, `skrifa` cannot be bumped to 0.41.0 in this PR: skrifa 0.41 pulls in `read-fonts` 0.38, while the published `harfrust` 0.5.2 depends on `read-fonts` 0.37. Since this PR passes skrifa's `FontRef` directly to harfrust's shaping API, the two `FontRef` types (from different `read-fonts` versions) are incompatible.

Note: harfrust's `main` branch has already moved to `read-fonts` 0.38, so this will unblock once a new harfrust version is published.

gcailly:
@emilk Thank you very much for the time and effort put into cleaning/simplifying/refactoring/organizing. I really hope my limited skills are not giving you too much work.

emilk (approved):
The web demo increases in size from 11038206B -> 11693417B, so 650kB more. Not negligible, but worth it imho.

umajho:
Hello. This PR unfortunately introduced a regression: #8087

Issue #7692, opened 2025-11-07
Use `AtomicU32` in place of `AtomicU64`


TimTheBig:
Use `AtomicU32` in place of `AtomicU64` for better compatibility, with systems that only support 32 bit atomics. The places where `AtomicU64` are used in egui would work with `AtomicU32` in all real cases.

TimTheBig:
`egui/texture_loader` and `epaint/fonts` are two such cases

kay-lambdadelta:
Can second this, it impedes egui being used on some lower tier platforms.

kay-lambdadelta:
Alternatively if the overflow is a concern, AtomicUsize could be used and people on platforms with only 32 bit atomics/cpus would be the only ones affected

kay-lambdadelta:
Considering the code at 

https://github.com/emilk/egui/blob/82a578e58c7ad8188c53a8fdc5bf6d4bc789e620/crates/egui/src/load/texture_loader.rs#L116-L129

I think that the `pass_index` will either have to be replaced with a mutex, or just casted into a usize possibly with a panic upon overflow on platforms where it does not fit.

Otherwise I think the change would be totally non-intrusive.

PR #8037, opened 2026-03-27
Remove 64-bit atomics in main crate

kay-lambdadelta:
This allows the base egui crate to run on platforms without 64 bit atomics

* Addresses <https://github.com/emilk/egui/issues/7692>, but does not address the `egui_extras` crate

### DIFF
--- a/crates/egui/src/load/texture_loader.rs
+++ b/crates/egui/src/load/texture_loader.rs
@@ -1,5 +1,3 @@
-use std::sync::atomic::{AtomicU64, Ordering::Relaxed};
-
 use emath::Vec2;
 
 use super::{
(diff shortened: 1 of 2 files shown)


lucasmerlin (approved):
Thanks!
Maybe we can also forbid the use of `AtomicU64` with clippy so we don't regress on this in the future?

kay-lambdadelta:
> Thanks! Maybe we can also forbid the use of `AtomicU64` with clippy so we don't regress on this in the future?

Something this would work I believe.

```toml
disallowed-types = [
    { path = "std::sync::atomic::AtomicU64", reason = "AtomicU64 is not supported on many common platforms, use AtomicU32 or AtomicUsize instead" }
]
```

PR #8042, opened 2026-03-28
Update of selected dependencies

oscargus:
Update some dependencies to get rid of old ones (especially nice to get rid of windows-sys 0.45...).

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -52,12 +52,12 @@ skip = [
   { name = "core-foundation" }, # version conflict between winit and wgpu ecosystems
   { name = "core-graphics-types" }, # version conflict between winit and wgpu ecosystems
   { name = "getrandom" }, # ring / rustls (and thus ehttp) still depend on getrandom 0.2
+  { name = "jni-sys" }, # 0.3.1 depends on 0.4
   { name = "kurbo" }, # Old version because of resvg
   { name = "redox_syscall" }, # old version via winit
   { name = "rustc-hash" }, # Small enough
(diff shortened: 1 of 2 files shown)


deny.toml
```diff
@@ -52,12 +52,12 @@ skip = [
   { name = "core-foundation" }, # version conflict between winit and wgpu ecosystems
   { name = "core-graphics-types" }, # version conflict between winit and wgpu ecosystems
   { name = "getrandom" }, # ring / rustls (and thus ehttp) still depend on getrandom 0.2
+  { name = "jni-sys" }, # 0.3.1 depends on 0.4
```
emilk, line 55: 0.3.1 of what depends on 0.4 of what?

emilk (approved):
Thank you!

deny.toml
oscargus: jni-sys 0.3.1 has jni-sys 0.4.0 as a dependency.

emilk: what? How can a crate depend on its future self? O.o

emilk (approved):

deny.toml
oscargus: Good question. My guess is that they made a fundamental change in 0.4, but then released 0.3.1 as a compatibility layer instead of backporting something important. So should be a very temporary duplication.

Skgland: Known as the [semver-trick](https://github.com/dtolnay/semver-trick)

PR #7983, opened 2026-03-18
Improve IME handling, add public method `owns_ime_events` on `Memory`

umajho:
* Depends on #7967
* Closes #7485
* Should fix #7906 (This issue doesn't seem to have been resolved, but the author closed it; I personally don't have the environment to verify whether it is fixed.)
* Replaces #4137, #4896, and partially #7810
* [x] I have followed the instructions in the PR template

This PR started as a fix for #7485, but has since evolved into a broader rewrite of IME-related logic.

## Overview

This PR primarily introduces a new public method, `owns_ime_events`, on [`Memory`], and refactors parts of [`TextEdit`] to integrate with it.

Previously, each [`TextEdit`] widget independently determined whether to handle IME events and stored its own IME-related state. This approach made ownership-handling fragmented and was therefore error-prone.

With this PR:
- IME event ownership is centralized, ensuring that at most a single widget owns IME events per frame.
- [`PlatformOutput`]'s `ime` field can be set to `None` for at least one frame when IME composition is interrupted, allowing the IME to be properly dismissed.

## Details

Two new public methods are introduced on [`Memory`]:

- `fn owns_ime_events(&self, id: Id) -> bool`: check IME event ownership for the current frame for the widget with the given `id`.
- `fn interrupt_ime(&mut self)`: interrupt the current IME composition, if any.

Since the newly added methods on [`Memory`] are public, other widgets can also participate in IME handling without risking ownership conflicts of IME events.

I also added an internal (`pub(crate)`) field on [`TextEditState`], called `cursor_purpose`, to distinguish the role of the [`TextEdit`] cursor.

Additionally, `egui::ImeEvent::Enabled` and `egui::ImeEvent::Disabled` have been removed, as they are no longer used anywhere.

### Windows: The Korean IME text duplication bug fixed in #4137 does not reappear.

<table>
<thead>
<tr>
<th></th>
<th>With this PR</th>
<th>Without this PR</th>
</tr>
</thead>
<tbody>
<tr>
<th>Behavior</th>
<td>Correct (no regression)</td>
<td>Correct</td>
</tr>
<tr>
<th>Screencast</th>
<td>

win-kor-after
</td>
<td>

win-kor-before
</td>
</tr>
</tbody>
</table>

### Windows: Chinese and Japanese IMEs now behave more consistently with the Korean IME in similar scenarios.

This change does not matter much, as composition is rarely interrupted mid-process with these IMEs in typical usage.

<table>
<thead>
<tr>
<th></th>
[...]

PR #8045, opened 2026-03-29
Improve robustness of text input handling for `eframe/web`

### LINKED ISSUE #8046: Keyboard layout is reset after every keystroke in text input on mobile
The keyboard layout is reset after each keystroke. For example, the keyboard exits the numeric layout after typing any character as can be seen in the video below:

https://github.com/user-attachments/assets/8946c9a7-629b-4f46-a495-b3ae18661553

Recorded on a Google Pixel 7a running Firefox 149.0. The keyboard is the default keyboard that came with the phone (Gboard).

This happens on anything with a text input.

The expected behavior is to be able to type multiple numbers in succession without having to switch back to the numeric layout every time.

I suspect the issue is caused by these lines added in #4855:
https://github.com/emilk/egui/blob/eb35f7d12f2a941707eee3406aab28e1df0262c9/crates/eframe/src/web/text_agent.rs#L59-L64

umajho:
* Fix [the Samsung Keyboard Korean Cheonjiin layout bug](https://github.com/emilk/egui/pull/7967#issuecomment-4098503570)
* Partially fix (does not close) #8046
* Supersedes #7914
* Supersedes #8047
* Related: #8068
* Related: #7983
* Related: #8078

This PR reworks the text input handling logic in eframe's web integration, primarily in `text_agent.rs`.
It also adds a new `ImeEvent::DeleteSurrounding` variant, along with the corresponding handling logic in `egui` to support the changes.

## Fix: Samsung Keyboard Cheonjiin issue

This PR fixes a bug reported by @rustbasic when using Samsung Keyboard's Cheonjiin Korean layout.

Since Samsung Keyboard is only available on Samsung devices, I wasn't able to verify it myself. The fix is based on @rustbasic's testing and confirmation.

The root cause is that the layout relies on the preceding text to correctly handle batchim composition. Previously, the text agent eagerly cleared the text input after every IME composition, removing the context too early. This PR makes that cleanup more conservative, preserving the text when it may still be needed.

My understanding is that this is a quirk specific to Samsung Keyboard: The IME reports that composition has finished even though it is effectively still active and the composed text may continue to change.

## Fix: Keystrokes resetting keyboard layout (numpad/symbols/etc.)

Partially fixes #8046.

Previously, keystrokes would cause the on-screen keyboard to switch back to its primary layout.

One remaining issue is that tapping within the active `TextEdit` to reposition the cursor still resets the keyboard to its primary layout.

## About text suggestions

I originally planned to include text suggestion support in this PR because #8068 implemented it.
 
However, adding text suggestion support would broaden the scope of this PR, so I think it is better addressed in a separate PR.

For reference, [the reverted implementation](https://github.com/emilk/egui/pull/8045/commits/8a4f70859c8ba1e00186db1431e8f6906d2d1426) works fine on Android (Gboard), but not on iOS (iPadOS 17 + SwiftKey).


umajho:
@rustbasic Hello.

I'd appreciate it if you could test whether this solution works with the Cheonjiin IME you are using.
Since I'm unable to reproduce the bug on my end, there is a fair chance that this first attempt may not fully resolve it, though I believe the overall approach is on the right direction.

It would also be very helpful if you could try this in other web setups to check for any regressions.
(No need to test native setups, as these changes should not affect them.)

Thanks in advance.

crates/egui/src/data/input.rs
```diff
@@ -611,6 +611,15 @@ pub enum ImeEvent {
     /// IME composition ended with this final result.
     Commit(String),
 
+    /// Notifies when the text surrounding the cursor should be deleted.
+    ///
+    /// `before_chars` and `after_chars` are the number of characters (not
+    /// bytes) to delete before and after the cursor, respectively.
+    DeleteSurrounding {
```
umajho: This is actually not an ad hoc variant introduced just for this PR. A variant with the same name was added to `winit` in 0.31 (though in this implementation, the fields are defined as `*_chars` rather than `*_bytes`).

https://docs.rs/winit/0.31.0-beta.2/winit/event/enum.Ime.html#variant.DeleteSurrounding

KonaeAkira:
I can confirm this PR fixes the 2 issues mentioned in #8046, at least on my device.

rustbasic:
@umajho  
I tested it on `Android + Cheonjiin IME`.  
The cursor position keeps moving forward, and character deletion behaves strangely.


https://github.com/user-attachments/assets/d4586218-5f4e-4695-97c7-ec5bc16b730b

umajho:
Wow, that's bad.

If you go `Settings -> System -> Languages & input -> On-screen keyboard`, what does it show?
For example, here is mine:
<details><summary>Screenshot</summary>

Screenshot 2026-03-31 at 6 05 27 PM
(The Android Keyboard doesn't provide Korean.)
</details>

Could you also check <https://stackblitz.com/edit/egui-text-agent-experiment> and record what it looks like when you input text?
For example, this is what it looks like with GBoard on Android:
<details><summary>Screencast</summary>

2026-03-31 6 30 00 PM
</details>

For reference, here is how GBoard behaves on the demo:
<details><summary>Screencast</summary>

2026-03-31 6 35 12 PM
</details>

rustbasic:
Are you asking for the keyboard settings screen first? Here it is.

삼성키보드 = SAMSUNG Keyboard

image blured.

rustbasic:
@umajho 

https://github.com/user-attachments/assets/d3f68cb7-99d3-4a26-94e2-be981507aaa2

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -1197,6 +1197,26 @@ fn events(
...
+                        }
+                        if *before_chars > 0 {
+                            text.delete_selected_ccursor_range([
+                                cursor_range.primary - *before_chars,
+                                cursor_range.primary,
+                            ]);
+                            ccurosr_end -= *before_chars;
+                        }
+                        Some(CCursorRange::one(ccurosr_end))
+                    }
```
rustbasic:
```
                    ImeEvent::DeleteSurrounding {
                        before_chars,
                        after_chars,
                    } => {
                        let mut ccursor_end = cursor_range.secondary;
                        if *after_chars > 0 {
                            text.delete_selected_ccursor_range([
                                ccursor_end,
                                ccursor_end + *after_chars,
                            ]);
                        }
                        if *before_chars > 0 {
                            text.delete_selected_ccursor_range([
                                ccursor_end - *before_chars,
                                cursor_range.primary,
                            ]);
                            ccursor_end -= *before_chars;
                        }
                        Some(CCursorRange::one(ccursor_end))
                    }
```

crates/eframe/src/web/text_agent.rs
```diff
@@ -86,35 +132,103 @@ impl TextAgent {
...
+                    let out_event = egui::Event::Ime(egui::ImeEvent::DeleteSurrounding {
+                        before_chars: last_text_len - prefix_len,
+                        after_chars: 0,
+                    });
+                    runner.input.raw.events.push(out_event);
+                }
+                let out_event = egui::Event::Ime(egui::ImeEvent::Commit(
+                    text.chars().skip(prefix_len).collect(),
+                ));
+                runner.input.raw.events.push(out_event);
```
rustbasic:
```
                let text = input.value();

                let mut last_text_ref = last_text.borrow_mut();
                let prefix_len = longest_common_prefix_length(&text, &last_text_ref);
                let out_event = egui::Event::Ime(egui::ImeEvent::Commit(
                    text.chars().skip(prefix_len).collect(),
                ));
                runner.input.raw.events.push(out_event);
```

rustbasic:
@umajho

With the changes mentioned in the comment, I get the best result when I modify `ImeEvent::Commit` in `builder.rs` as follows.

However, there is still an issue: when `않` becomes `않ㅇ`, the previous Hangul character is deleted, and the entire text shifts left by one position.

```rust
                    ImeEvent::Commit(commit_text) => {
                        if commit_text == "\n" || commit_text == "\r" {
                            None
                        } else {
                            let mut ccursor = cursor_range.primary;
                            state.ime_enabled = false;

                            if !commit_text.is_empty()
                                && cursor_range.secondary.index
                                    == state.ime_cursor_range.secondary.index
                            {
                                ccursor = clear_preedit_text(text, &cursor_range);
                                text.insert_text_at(&mut ccursor, commit_text, char_limit);
                            }

                            Some(CCursorRange::one(ccursor))
                        }
                    }
```


https://github.com/user-attachments/assets/af2bfc08-f148-46c2-a7d8-f5d0a9602cf1

umajho:
> https://github.com/emilk/egui/pull/8045#issuecomment-4162190135
> Are you asking for the keyboard settings screen first? Here it is.
> 
> …

Yes, thanks for the information.
It turns out that it is only available on Samsung phones, which totally makes sense. So this is indeed inaccessible to me.
(There is an unrelated “Samsung Smart Keyboard” App on the Play Store, which is for physical keyboards.)

> https://github.com/emilk/egui/pull/8045#issuecomment-4162242197

Sorry, my instructions were not clear. You need to go to the “preview” tab.
That said, I found that stackblitz runs fast on Android Firefox (my default browser), but it is somehow slow on Android Chrome. So, I'm planning to rebuild this experimental page on platforms like jsfiddle using plain HTML+JS that don't require build steps, probably tomorrow.

> https://github.com/emilk/egui/pull/8045#issuecomment-4162413559
> With the changes mentioned in the comment, I get the best result when I modify `ImeEvent::Commit` in `builder.rs` as follows.
> 
> …

Thanks for the suggested changes. I will check them tomorrow.

rustbasic:
@umajho 


https://github.com/user-attachments/assets/a303dad6-1f5a-450f-8a3d-73203da1a774

crates/egui/src/widgets/text_edit/builder.rs
umajho: I added test cases to clarify the intended behavior (which I had initially implemented incorrectly).
I also refactored this logic into a method on `TextBuffer`:
https://github.com/emilk/egui/pull/8045/changes#diff-ea318a781e1cd39569c7db9936ccc9d93a41e59c2a3b5ce3fed67ee8749bd3bcR152-R169

crates/eframe/src/web/text_agent.rs
umajho: Good catch.
~~I further simplified `on_composition_end` by reverting to the previous approach of using `event.data()`.~~
Update: The “further simplification” is wrong. I just realized I should just use what you suggested here.

umajho:
@rustbasic I have made some changes. does this improve things?

The stackblitz page has also been updated to reflect the changes.

rustbasic:
@umajho 

It appears to be fixable, since the issue does not occur when "않았다" is entered consecutively without any problems.  
The following are the responses when it is not entered consecutively.

1) After entering "않", wait briefly for a commit to occur, and then enter the next character.


https://github.com/user-attachments/assets/968a92ce-9b22-4fd3-8dfb-c1fc2a3dd365



2) After entering "안", wait briefly for a commit to occur, and then enter the next character.


https://github.com/user-attachments/assets/5a21677d-825f-4f1d-b839-f8c8079ff2b9

rustbasic:
@umajho 

It appears to be fixable, since it works correctly when "않았다" is entered continuously.

1)  
However, if you enter "않", wait briefly for a commit to occur, and then enter the next character, "않" gets duplicated.  
For example, if you enter "않", wait, and then type "ㅇ", it becomes "않않ㅇ".

Result 1-1) "않않았다"

https://github.com/user-attachments/assets/739eb23a-d932-453c-8c30-77a204ab722a



2)  
If you enter "안", wait briefly for a commit to occur, and then enter the next character, "안" gets duplicated.  
In this case, the final output varies as follows.

Result 2-1) "안않았다"  
Result 2-2) "안않않았다"  
Result 2-3) "안ㅎㅎ았다"


https://github.com/user-attachments/assets/ad8be5c3-adbd-489f-a1de-dd7623b4b994

umajho:
Is this auto-committing behavior expected?

I will try to figure out how to fix the problem tomorrow.

umajho:
Sorry for the delayed response.
I've made a few more changes. Does this improve the situation?

Update:
I realized that the changes I was talking about here are simply reverting my earlier “further simplification” back to the version of the code suggested by rustbasic [here](https://github.com/emilk/egui/pull/8045#discussion_r3015630561).

rustbasic:
@umajho

I’ve briefly tested #8045 on Windows 10, WASM, and Android (using the Cheonjiin keyboard), and it works well.
If #8045 is merged, it seems you will have made a significant contribution to egui’s IME.
I will let you know if I encounter any issues.
Thank you for your hard work.

umajho:
@rustbasic Glad to hear that!

Could you also try #8068? It also looks promising for addressing the Cheonjiin issue you encountered.
The implementations of these two PRs conflict, so if #8068 works well for you, I'd prefer to close this PR in favor of that one.

cjgriscom:
There seems to be a new issue, perhaps introduced in #7983 

Clicking/dragging to select in the TextEdit now repeatedly causes a refocus.  This change is one way to fix it:
https://github.com/emilk/egui/pull/8068/commits/9473de9cdfa6660262ecac8c29adbfd0e9035289

https://github.com/user-attachments/assets/0d5bff40-67f8-426b-97de-c09272073244

umajho:
> There seems to be a new issue, perhaps introduced in #7983
> 
> Clicking/dragging to select in the TextEdit now repeatedly causes a refocus. This change is one way to fix it: [9473de9](https://github.com/emilk/egui/commit/9473de9cdfa6660262ecac8c29adbfd0e9035289)
> 
>  Screen_Recording_20260406_111123_Firefox.mp4

Oops. This is actually the result of an intentional action.
I made every focus request interrupt IME composition, even if the widget is already focused, to address this bug:
2026-04-06 11 35 27 PM
(Wayland + iBus. The keystrokes are recorded incorrectly due to differences in keyboard layouts between the host OS and the guest OS..)

However, I didn't expect that this would cause problems with virtual keyboards.

A better approach might be to introduce something like `OutputCommand::InterruptIme`, and let integrations decide how to handle it?

For example, on web, we could reuse this trick to reset the IME without causing flickering:

```rs
input.blur().ok();
input.focus().ok();
```

(It is late here, so my next reply may take a few hours.)

KonaeAkira:
@umajho I am against using that trick because it is the root cause of the 2 issues described in https://github.com/emilk/egui/issues/8046

cjgriscom:
It's likely that either set_value or set_selection_range on the `<input>` will reset IME without defocusing.

umajho:
@cjgriscom The flickering issue should be fixed by #8078.

umajho:
FIXME: 5ce55ad27c5f658e2d9ae51c8e4e1391c7ff7e7e causes Gboard 14.7.09 to require an extra initial delete keystroke before ASCII character deletion behaves normally.
(Gboard 17.0.12 does not have this issue.)

~~Given that the bug is not present in newer Gboard versions, I propose that we just ignore it and remove the now-nonfunctional, obscure workaround that was previously added in this PR. Would anyone be against this?~~
Update: I think there is a way to solve this.
Please pardon me, but I will not have time to work on this until the end of the month (April).

umajho:
@rustbasic Sorry to bother you. When you have a chance, could you verify that the fix still works now that this PR has been rebased onto the current `main`, especially after resolving the conflicts with #8083?

rustbasic:
@umajho

I have briefly tested #8045 on Windows 10, WASM (Chrome), and Android (Korean Cheonjiin keyboard).
I didn't find any issues on Windows 10 and WASM (Chrome), but I encountered a few problems on Android (Korean Cheonjiin keyboard).

1. On Android, after entering text and having a new text cursor appear, if I select another area, the text cursor disappears. If I try to type characters in this state, the application freezes completely.

2. On Android, after entering text and having a new text cursor appear, if I select another area, the text cursor fails to display and disappears. Once the text cursor vanishes, it does not reappear even if I select various different locations; it only becomes visible again when I input characters.

3. On Android, when selecting text (making a selection block), the text cursor is positioned at the very beginning of the selected block instead of at the end of it.

Since these three symptoms occur even when #8045 is reverted (removed), they are bugs introduced in #8083.
Symptom 1 above disappears when #8045 is applied.
Symptoms 2 and 3 above do not exhibit any issues if `Legacy visuals` is turned on after applying #8045.

I will continue using it for a few more days and let you know if I find any other anomalies.
My suggestion is to apply #8045 to eliminate Symptom 1 first, and then resolve the remaining issues.

umajho:
Ouch, @lucasmerlin. I'm so embarrassed of myself.
The out of bounds case you warned about [here](https://github.com/emilk/egui/pull/8083#discussion_r3474529024) did end up happening, affecting Android IME users. (@rustbasic's Symptom 1)

The reason this PR fixed it is that, during the merge with #8083, GLM added a defensive check to ensure `selection_start` and `selection_end` are within bounds before slicing:

https://github.com/emilk/egui/pull/8045/changes/15fd72a0d0de6eab4bd1964f4e6e922aa925d4dc#diff-180cddab275a9f2a9ba697c0c68fbf19182359aae05c360583d2484187c79815R98 (L98~L100)
```rs
        if selection_start > text_utf16.len() || selection_end > text_utf16.len() {
            return None;
        }
```

---

> @umajho My suggestion is to apply https://github.com/emilk/egui/pull/8045 to eliminate Symptom 1 first, and then resolve the remaining issues.

This PR still is not ready. Gboard 14.7.09 was working perfectly before, but this PR introduces some issues with backspace handling for it. Since IIRC it works in #8068, I want to figure out how to preserve Gboard 14.7.09's behavior while fixing the Cheonjiin issue at the same time.

So, I think opening a separate PR is the better approach.

On why this PR hasn't seen much progress:
I was waiting for #8083 to be merged before continuing work on this PR, since both modify `text_agent.rs`, and resolving the conflicts is a nightmare. I am also planning to make additional changes to `text_agent.rs` in this PR, so continuing before #8083 merged would have only created more conflicts and headaches. Since now #8083 is in `main`, I will resume work on this PR, and I don't expect it to be stale for another month

---

Unfortunately, I'm going to the dentist for some tooth issues, and there's a good chance that I will be recovering from a procedure afterward. Because of this, I probably wont' be available for at least the next 24 hours. If someone opens the fix PR, I'd really appreciate it.

umajho:
I think Symptom 2 is likely a bug introduced in this PR, because at the moment it doesn't seem reproducible for #8083 due to Symptom 1. Maybe the app freezing was mistaken for the cursor not showing by you?

I can't reproduce Symptom 3 on egui.rs. Or perhaps I misunderstood your reproduction steps? Could you upload a screencast?
(Maybe the Chrome version also matters? I'm using Vanadium 149.0.)

rustbasic:
@umajho

Even after applying #8271 and testing, Symptom 2 still occurs. Therefore, Symptom 2 is indeed an issue introduced by #8083 (and is not a simple misunderstanding caused by the app freezing).

To reproduce Symptom 3, you first need to input some characters so that a new text cursor appears, and then select (highlight) a block of text.

rustbasic:
screencast

https://github.com/user-attachments/assets/b454a70d-20c5-4561-bbf7-c2b72ab5e96d

umajho:
> @umajho
> 
> Even after applying #8271 and testing, Symptom 2 still occurs. Therefore, Symptom 2 is indeed an issue introduced by #8083 (and is not a simple misunderstanding caused by the app freezing).
> 
> To reproduce Symptom 3, you first need to input some characters so that a new text cursor appears, and then select (highlight) a block of text.

Ohh, I see. You mean selecting text while the IME composition is still active.
As a Chinese IME user (and I assume it is the same for Japanese IME users), I never select text before committing the composition, so I wasn't aware of this use case if it has not been not brought up.
Thanks for the explaining!

That also explains why I didn't encounter any of the three symptoms you reported. (Including the first one: although I made it sound obvious after your initial report, I don't think I wound have discovered it on my own. If you hadn't specifically mentioned “If I try to type characters in this state”, it would never have occurred to me to click elsewhere while composing.)

umajho:
@rustbasic Just like the `winit` cursor position bug on Windows, which only started manifesting after the new IME composition visuals were implemented, the root cause of Symptom 2 and Symptom 3 you found have actually always been present.

The root cause is: On android Chrome, due to some complicated interaction in `eframe/web` (I can't reproduce this with a bare `<input>` element), when the agent `<input>` loses focus, Chrome still considers the IME composition to be active and never emits `compositionend`. I also supect this leaves Chrome's internal IME state inconsistent, which may further explain Symptom 1.

---

The root cause can actually be demonstrated using a commit from before #8083 was merged:

If you type something, commit it, and click back into the TextEdit, you get a normal cursor:
2026-06-29 8 44 05 PM

However, if you type something, click somewhere else without committing the composition, then click back into the TextEdit, you get what appears to be a selection (in reality, it is an active composition, but before #8083 there was no way to distinguish the two). If you type a number, the number is appended after the highlighted text instead of replacing it, which kinda indicates that the highlighted region is not actually a typical selection:
2026-06-29 8 44 34 PM

rustbasic:
@umajho 

I haven't looked into the internal details closely, so I am not familiar with the technical aspects.

Please note that symptoms 2 and 3 above do not cause any issues if you set `ui.style_mut().visuals.ime_composition.legacy_visuals = true`.

umajho:
> @umajho
> 
> I haven't looked into the internal details closely, so I am not familiar with the technical aspects.
> 
> Please note that symptoms 2 and 3 above do not cause any issues if you set `ui.style_mut().visuals.ime_composition.legacy_visuals = true`.

The second screencast in [my reply above](https://github.com/emilk/egui/pull/8045#issuecomment-4832848146) demonstrates a buggy behavior that is caused by the same underlying bug of the symptoms you mentioned. I can reproduce it both on the commit prior to #8083  and on #8271 with the legacy visuals enabled.

What is happening is:
- with the legacy visuals, the cursor is always drawn at the end of the composition, which appears correct for composing Korean.
- with the legacy visuals, there is no visual distinction between an active composition and a selection, as a result, even if a selection is incorrectly treated as an active composition, it is not visually apparent.

This can give the impression that everything is working correctly under the legacy visuals, when in fact the underlying state is still incorrect.

umajho:
@rustbasic could you verify whether the new commit fixes the symptoms you found? Thanks.

rustbasic:
@umajho

Great news!
I've tested it on Android, and all the previous symptoms have completely disappeared. Everything is working perfectly now!

I'll keep using it and let you know if I find any new issues.

umajho:
Hello, I think this PR is ready for review. I'd appreciate it if you could have a look at it and let me know whether everything works as expected.
@rustbasic @KonaeAkira @cjgriscom

rustbasic:
@umajho

I tested this on WASM (Android + Cheonjiin keyboard layout) and encountered the following issue:

When there are multiple `TextEdit` widgets, if I enter text and then move focus to another `TextEdit` before committing (completing the character composition) and return to the original one, the last character entered is automatically highlighted/selected.

In Korean (and other composition-based IME inputs), when the focus moves away in the middle of character entry, the character currently being composed should be automatically treated as committed.

umajho:
> @umajho
> 
> I tested this on WASM (Android + Cheonjiin keyboard layout) and encountered the following issue:
> 
> When there are multiple `TextEdit` widgets, if I enter text and then move focus to another `TextEdit` before committing (completing the character composition) and return to the original one, the last character entered is automatically highlighted/selected.
> 
> In Korean (and other composition-based IME inputs), when the focus moves away in the middle of character entry, the character currently being composed should be automatically treated as committed.

If what you saw was a selection (the blue rectangle in the background by default) rather than an IME composition underline, then the text has already been committed.

If I remember correctly, this is how `egui` has always behaved. Unlike many other UI framewarks, which only maintain selection state for the currently focused text field, `egui` preserves the selection state for every `TextEdit`, regardless of whether it is currently focused. As a result, when you return to a previously focused `TextEdit`, its previous selection (which the IME composition was converted into when the field lost focus) is restored instead of instead of starting with a fresh selection state.

For comparison:

<details><summary>HTML (Safari)</summary>
2026-07-20 6 08 29 PM
</details>

<details><summary>SwiftUI</summary>
2026-07-20 6 10 21 PM
</details>

<details><summary>egui</summary>
2026-07-20 6 13 31 PM
</details>

---

In my opinion, changing this behavior is outside the scope of this PR.

umajho:
Maybe I should change the behavior on blur from `IME composition -> selection` to `IME composition -> cursor at the end of the current composition`?

rustbasic:
@umajho

With your fix, the input issue is completely resolved. I absolutely love how smooth it is now. Amazing work!

rustbasic:
@emilk @lucasmerlin

Highly recommend merging PR #8045. 
It works good!

umajho:
[The `typos` check](https://github.com/emilk/egui/actions/runs/30079182489/job/89436628680?pr=8045) failed with:
```
  Connecting to release-assets.githubusercontent.com (release-assets.githubusercontent.com)|185.199.108.133|:443... connected.
  Unable to establish SSL connection.
  Error: Process completed with exit code 4.
```
What…?

crates/egui/src/data/input/ime_event.rs
```diff
@@ -22,6 +22,15 @@ pub enum ImeEvent {
...
 
+    /// Notifies when the text surrounding the cursor should be deleted.
+    ///
+    /// `before_chars` and `after_chars` are the number of characters (not
+    /// bytes) to delete before and after the cursor, respectively.
+    DeleteSurrounding {
+        before_chars: usize,
+        after_chars: usize,
+    },
+
```
lucasmerlin, line 33: `DeleteSurrounding` in winit 0.31 will be byte based:
https://docs.rs/winit/0.31.0-beta.2/winit/event/enum.Ime.html#variant.DeleteSurrounding

Would this be compatible? Should we also use bytes instead? 

Then again on `Preedit` we also use chars, so I guess its fine and we can use the same conversion.

lucasmerlin (approved):
Awesome, seems to work pretty well!

On my android text input was flawless and even autocorrect and autocomplete are working almost seamlessly (only placing the cursor on an existing word doesn't trigger autocorrect but that's a minor detail). 

On iOS autocorrect did not work. It did show suggestions and even the ime edit zone, but tapping them does nothing. 

Maybe these things could be further improved in a follow up PR.

crates/egui/src/data/input/ime_event.rs
umajho: I chose `char` because `egui` only needs character counts to handle `DeleteSurrounding `, and in `eframe`'s web integration the lengths of strings are also in `char`s. Keeping the unit as `char`s avoids a `chars -> bytes -> chars` conversion.
For integrations whose upstream API uses byte offsets (e.g. `egui-winit`), the integration can convert those offsets to character counts before emitting the event, so I don't think this creates a compatibility problem.
That said, I don't feel strongly about this choice. I'm happy to use byte counts instead if you think that is more appropriate.

umajho:
> On iOS autocorrect did not work. It did show suggestions and even the ime edit zone, but tapping them does nothing.

Thanks for testing this. I will look into it.

umajho:
I made some improvements to event handling in c7f5269.

Android Gboard suggestions/autocomplete/autocorrect now correctly apply replacements when previously entered text needs to be changed (e.g., `tex|` -> `Texas`, where the initial `t` should be capitalized):

<details><summary>screencast</summary>

https://github.com/user-attachments/assets/d9c0b8ed-bb1e-4c08-9351-b9235438ed07
</details>

Suggestions/autocomplete/autocorrect now work in iOS Safari:

<details><summary>screencast</summary>

https://github.com/user-attachments/assets/adfb5452-b670-4e14-9c58-8ddc9d0644d8
</details>

The Korean keyboard (including handling batchim) now works in iOS Safari:

<details><summary>screencast</summary>

https://github.com/user-attachments/assets/ec8354e2-7449-4cdd-b10a-fe65cd1f6bb9
</details>

emilk:
Sidenote: I hope this PR doesn't cause problems for you:
* https://github.com/emilk/egui/pull/8297

lucasmerlin:
Awesome, can confirm, autocorrect and autocomplete works perfectly now on my iOS safari!

umajho:
> Sidenote: I hope this PR doesn't cause problems for you:
> 
> * [Web: anchor the text agent to the canvas #8297](https://github.com/emilk/egui/pull/8297)

I applied the changes from that PR here, and I didn't encounter any issues, so I think it should be fine.

<details><summary>patch</summary>

```diff
diff --git forkSrcPrefix/crates/eframe/src/web/text_agent.rs forkDstPrefix/crates/eframe/src/web/text_agent.rs
index bc2ade029d993a28e725e64858461d92f4a95c7d..d0238cb2606529f873b1de0cee1f56ab6f65ef9d 100644
--- forkSrcPrefix/crates/eframe/src/web/text_agent.rs
+++ forkDstPrefix/crates/eframe/src/web/text_agent.rs
@@ -4,7 +4,6 @@
 use std::{cell::RefCell, rc::Rc};
 
 use wasm_bindgen::prelude::*;
-use web_sys::Document;
 
 use super::{AppRunner, WebRunner};
 
@@ -30,10 +29,9 @@ impl TextAgent {
         input.set_attribute("autocapitalize", "off")?;
         let input_state = Rc::new(RefCell::new(InputState::new(input.clone())));
 
-        // Hide the element, and park it over the canvas
+        // Hide the element, and park it over the top-left corner of the canvas
         // so that focusing it can never scroll some other part
         // of the page into view.
-        let canvas_rect = super::canvas_content_rect(canvas);
         let style = input.style();
         style.set_property("background-color", "transparent")?;
         style.set_property("border", "none")?;
@@ -42,21 +40,22 @@ impl TextAgent {
         style.set_property("height", "1px")?;
         style.set_property("caret-color", "transparent")?;
         style.set_property("position", "absolute")?;
-        style.set_property("top", &format!("{}px", canvas_rect.min.y))?;
-        style.set_property("left", &format!("{}px", canvas_rect.min.x))?;
+        style.set_property("top", &format!("{}px", canvas.offset_top()))?;
+        style.set_property("left", &format!("{}px", canvas.offset_left()))?;
         // Prevent auto-zoom on mobile browsers (requires at least 16px).
         style.set_property("font-size", "16px")?;
 
-        let root = canvas.get_root_node();
-        if root.has_type::<Document>() {
-            // root object is a document, append to its body
-            root.dyn_into::<Document>()?
-                .body()
-                .unwrap()
-                .append_child(&input)?;
-        } else {
-            // append input into root directly
-            root.append_child(&input)?;
+        // Insert the input as a sibling of the canvas, so that its
+        // `position: absolute` resolves against the same containing block
+        // as the canvas' `offset_top`/`offset_left`.
+        // This anchors the input to the canvas regardless of how the page
+        // is scrolled or how the canvas is embedded, and also works when
+        // the canvas is inside a shadow DOM.
+        if let Some(parent) = canvas.parent_node() {
+            parent.insert_before(&input, canvas.next_sibling().as_ref())?;
+        } else if let Some(body) = document.body() {
+            log::warn!("Canvas has no parent element - appending text agent to document body");
+            body.append_child(&input)?;
         }
 
         // Focus the app on startup, without scrolling the page.
@@ -184,7 +183,6 @@ struct InputState {
     last_text: String,
     prev_ime_output: Option<egui::output::IMEOutput>,
     keydown_special_case: KeydownSpecialCase,
-    is_mobile_safari: bool,
 }
 
 #[derive(Clone, Copy)]
@@ -223,7 +221,6 @@ impl InputState {
             last_text: String::new(),
             prev_ime_output: None,
             keydown_special_case: KeydownSpecialCase::None,
-            is_mobile_safari: is_mobile_safari(),
         }
     }
 
@@ -241,29 +238,28 @@ impl InputState {
 
         let Some(ime) = ime else { return Ok(()) };
 
-        let mut canvas_rect = super::canvas_content_rect(canvas);
-        // Fix for safari with virtual keyboard flapping position
-        if self.is_mobile_safari {
-            canvas_rect.min.y = canvas.offset_top() as f32;
-        }
-        let cursor_rect = ime.cursor_rect.translate(canvas_rect.min.to_vec2());
-
         let style = self.input.style();
         let native_ppp = super::native_pixels_per_point();
 
         // Clamp the input position within the canvas width to prevent unwanted horizontal scrolling.
         let logical_canvas_width = canvas.width() as f32 / native_ppp;
-        let visible_x = cursor_rect.center().x * zoom_factor;
+        let visible_x = ime.cursor_rect.center().x * zoom_factor;
         let clamped_x = visible_x.clamp(0.0, logical_canvas_width);
 
         // Clamp the input position within the canvas height to prevent unwanted vertical scrolling.
         let logical_canvas_height = canvas.height() as f32 / native_ppp;
-        let visible_y = cursor_rect.center().y * zoom_factor;
+        let visible_y = ime.cursor_rect.center().y * zoom_factor;
         let clamped_y = visible_y.clamp(0.0, logical_canvas_height);
 
         // This is where the IME input will point to:
-        style.set_property("left", &format!("{clamped_x}px"))?;
-        style.set_property("top", &format!("{clamped_y}px"))?;
+        style.set_property(
+            "left",
+            &format!("{}px", canvas.offset_left() as f32 + clamped_x),
+        )?;
+        style.set_property(
+            "top",
+            &format!("{}px", canvas.offset_top() as f32 + clamped_y),
+        )?;
 
         Ok(())
     }
@@ -414,26 +410,6 @@ impl InputState {
     }
 }
 
-/// Returns `true` if the app is likely running on a mobile device on navigator Safari.
-///
-/// NOTE(umajho): This function does not detect my iPad (iPadOS 17.7.11).
-/// I'm not sure what the `virtual keyboard flapping position` bug refers to,
-/// but the virtual keyboard's position on my iPad appears fine even without
-/// applying that workaround. Its values are:
-/// - `user_agent`: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.14 Safari/605.1.15`
-/// - `navigator`.platform: `MacIntel`
-fn is_mobile_safari() -> bool {
-    (|| {
-        let user_agent = web_sys::window()?.navigator().user_agent().ok()?;
-        let is_ios = user_agent.contains("iPhone")
-            || user_agent.contains("iPad")
-            || user_agent.contains("iPod");
-        let is_safari = user_agent.contains("Safari");
-        Some(is_ios && is_safari)
-    })()
-    .unwrap_or(false)
-}
-
 fn longest_common_prefix_length(a: &str, b: &str) -> usize {
     std::iter::zip(a.chars(), b.chars())
         .take_while(|(a, b)| a == b)
```

</details>

umajho:
@rustbasic since there have been some changes since you last checked, would you mind take another look to see whether the Samsung Cheonjiin fix still works?

umajho:
Wait, I realized that since `TextEdit` can also be used for password input, the text agent's `<input>` should not retain text in that case.

umajho:
I think I may have overthought it. If a script can access `input.value`, it can also simply call `input.addEventListener("input", …)`.

Anyways, I have made `input.value` always be cleared when entering a password.

rustbasic:
@umajho
Tested on Android (Cheonjiin IME), Windows 10, and WASM (Chrome). Everything works smoothly and behaves as expected.

emilk:
Awesome work!

PR #8047, opened 2026-03-29
Remove text refocus on input to fix mobile keyboard layout resetting

### LINKED ISSUE #8046: Keyboard layout is reset after every keystroke in text input on mobile

KonaeAkira:
Fixes #8046

I don't know what the original code was supposed to fix, but removing that code fixes the issue linked above,

### DIFF
--- a/crates/eframe/src/web/text_agent.rs
+++ b/crates/eframe/src/web/text_agent.rs
@@ -56,12 +56,6 @@ impl TextAgent {
             let input = input.clone();
             move |event: web_sys::InputEvent, runner: &mut AppRunner| {
                 let text = input.value();
-                // Fix android virtual keyboard Gboard
-                // This removes the virtual keyboard's suggestion.
-                if !event.is_composing() {
-                    input.blur().ok();
-                    input.focus().ok();
-                }
                 // if `is_composing` is true, then user is using IME, for example: emoji, pinyin, kanji, hangul, etc.
                 // In that case, the browser emits both `input` and `compositionupdate` events,
                 // and we need to ignore the `input` event.


umajho:
Since the comments refer to Gboard, I tested the behavior using it.
I found that these lines are necessary for Gboard to function correctly for backspacing: without them, backspacing breaks.

<table>
<thead>
<tr>
<th>With these lines (before this PR)</th>
<th>Without these lines (with this PR applied)</th>
</tr>
</thead>
<tbody>
<tr>
<td></td>
<td>The first 5 backspaces didn't delete characters.</td>
</tr>
<tr>
<td>

2026-03-30 12 41 46 PM
</td>
<td>

2026-03-30 12 42 06 PM
</td>
</tbody>
</table>

---

Update:

Since I'm also working on improving `TextAgent`, I attempted to work around this issue in #8045.
Would you mind taking a look at that PR?

PR #8048, opened 2026-03-29
`Column::remainder().clip(true)` now shrinks when available width decreases

germ4n:
### Problem
`Column::remainder().clip(true)` grows correctly when the panel is made wider, but does not shrink when the panel is made narrower.

### Root cause
There are two places in `table.rs` where the `max_used_widths` floor is applied to column widths. One correctly checks `column.clip`, the other doesn't.

**Post-render** path (already correct, line ~827):
```rust
if !column.clip {
    *column_width = column_width.at_least(max_used_widths[i]);
}
```
`TableState::load` **pre-render path** (the bug, line 647):
```rust
.at_least(column.width_range.min.max(*max_used))  // clip flag ignored
```

In `TableState::load`, `.at_least(max_used)` is applied to every column regardless of `clip`. For a `Column::remainder()`, `max_used` accumulates the historically widest rendered content. When the panel shrinks, this floor prevents the column from computing a smaller width, creating a cycle where it can never shrink.

### Fix
Apply the same `clip` guard that already exists in the post-render path:
```rust
.at_least(if column.clip {
    column.width_range.min
} else {
    column.width_range.min.max(*max_used)
})
```

When `clip = true`, the floor is just `width_range.min` (0.0 by default), allowing the remainder column to shrink freely to the actual remaining space.

### Minimal reproduction
<details>
  <summary>Show code</summary>

```rust
use eframe::egui;
use egui_extras::{Column, TableBuilder};

fn main() -> eframe::Result {
    eframe::run_ui_native(
        "Column::remainder().clip(true) shrink test",
        Default::default(),
        |ui, _frame| {
            egui::Frame::central_panel(ui.style()).show(ui, |ui| {
                let mut text = String::from("some content");
                egui::ScrollArea::horizontal().show(ui, |ui| {
                    TableBuilder::new(ui)
                        .column(Column::auto())
                        .column(Column::remainder().clip(true))
                        .header(20.0, |mut header| {
                            header.col(|ui| { ui.strong("Auto"); });
                            header.col(|ui| { ui.strong("Remainder + clip"); });
                        })
                        .body(|mut body| {
                            body.row(18.0, |mut row| {
                                row.col(|ui| { ui.label("label"); });
                                row.col(|ui| {
                                    ui.add(
                                        egui::TextEdit::singleline(&mut text)
                                            .desired_width(f32::INFINITY),
                                    );
                                });
                            });
                        });
                });
            });
        },
    )
}
```
</details>

**Without the fix:** make the window wider (column grows ✓), then narrower — column does not shrink and a horizontal scrollbar appears ✗

https://github.com/user-attachments/assets/2b586588-9f72-4a15-80f4-afddadb69441

**With the fix:** the column shrinks correctly and no scrollbar appears ✓

https://github.com/user-attachments/assets/f1655641-e135-489c-9f59-2af3faa887ab


germ4n:
@emilk, could you please take a quick look at this?

emilk (approved):

Issue #7930, opened 2026-02-23
SidePanel resize direction reverses when pointer crosses anchor edge


smills8:
## Description

When dragging a `SidePanel::left` resize separator leftward past x=0 (or a `SidePanel::right` separator past the right window edge), the resize direction **reverses**: further leftward mouse movement causes the panel to grow rightward instead of continuing to shrink.

## Root Cause

`panel.rs` line 268 (egui 0.31.1):

```rust
width = (pointer.x - side.side_x(panel_rect)).abs();
```

The `.abs()` mirrors the signed distance around zero. When `pointer.x` crosses the panel's anchor edge (left edge for a left panel), the expression goes negative — but `.abs()` flips it positive, making the panel grow instead of collapse.

## Steps to Reproduce

1. Create a `SidePanel::left` with `resizable(true)` and a `min_width` of e.g. 96px
2. Drag the resize separator leftward
3. Continue dragging past the left edge of the panel (pointer x < panel left edge)
4. Observe: the panel starts growing again instead of staying at min_width

## Suggested Fix

Remove `.abs()` and let `clamp_to_range` enforce the lower bound:

```rust
// Current (buggy):
width = (pointer.x - side.side_x(panel_rect)).abs();

// Option A — use sign() to get a correctly-signed width:
width = side.sign() * (side.side_x(panel_rect) - pointer.x);

// Then clamp_to_range(width, width_range) naturally handles width < min_width
```

The same pattern affects right panels symmetrically (dragging past the right anchor edge).

## Version

egui 0.31.1

Issue #4384, opened 2024-04-20
SidePanel resizing behaves strange(?) when going out of bounds


oscargus:
**Describe the bug**

When resizing a `SidePanel`, one may increase the size by continue "resizing" out of bounds of the `SidePanel`. See the video here: https://gitlab.com/surfer-project/surfer/-/issues/210

This is, I think, caused by the `abs` in the following line.

https://github.com/emilk/egui/blob/79fbd17b338c22e09a44517407ecb7490b3ee69d/crates/egui/src/containers/panel.rs#L252

Since this has been there from the start, there may be some case where it is actually needed but I cannot really figure out one. Unless the idea is that this should happen.

**Expected behavior**

The resizing should stop when going out of bounds(?).

**Screenshots**

See video above.

Will be happy to provide a PR if this is deemed an issue.

Issue #4050, opened 2024-02-15
`set_clip_rect` draws over existing UI when


AaronRecord:
**Describe the bug**

![issue](https://github.com/emilk/egui/assets/50304111/8cdd6502-747e-4ecb-9952-44e52027520e)
![image](https://github.com/emilk/egui/assets/50304111/a846732e-a8c0-431d-969b-89dd70d9274c)
![image](https://github.com/emilk/egui/assets/50304111/7cc3e3db-cea1-4a91-a7a0-f149075a2cca)

**To Reproduce**
Minimal reproduction project:
```rs
use eframe::egui::*;

fn main() -> Result<(), eframe::Error> {
    let options = eframe::NativeOptions {
        viewport: ViewportBuilder::default(),
        ..Default::default()
    };

    eframe::run_native(
        "Image Cleanup",
        options,
        Box::new(|cc| Box::new(ImageCleanup {})),
    )
}

struct ImageCleanup {}

impl eframe::App for ImageCleanup {
    fn update(&mut self, ctx: &Context, _frame: &mut eframe::Frame) {
        SidePanel::right("parameters").show(ctx, |ui| {
            Grid::new("parameters")
                .spacing([40.0, 4.0])
                .striped(true)
                .show(ui, |ui| {
                    ui.label("Hello world!");
                    ui.end_row();
                });
        });

        SidePanel::left("preview_tools").show(ctx, |ui| {
            Grid::new("preview_parameters")
                .striped(true)
                .show(ui, |ui| {
                    ui.label("Hello world!");
                    ui.end_row();
                });
        });

        CentralPanel::default().show(ctx, |ui| {
            ui.set_clip_rect(ui.available_rect_before_wrap());
            ui.add(Button::new("Hi").min_size(
                ui.available_size() * ((ctx.frame_nr() as f32 / 60.0).sin().abs() + 0.5),
            ));
            ctx.request_repaint();
        });
    }
}
```

**Expected behavior**

Don't draw on top of the side panel, or if this is intended behavior, document it and describe
**Screenshots**

**Desktop (please complete the following information):**
 - OS:  Windows 11

emilk:
It looks like you're trying to add a widget to the central panel that is too large to fit it. What behavior do you expect?
You should be using a `ScrollArea` if you want to fit more things in there.

AaronRecord:
> It looks like you're trying to add a widget to the central panel that is too large to fit it. What behavior do you expect?
[...]

PR #8056, opened 2026-04-02
Clamp panel rects to visible bounds

### LINKED ISSUE #8055: Resizable Panel stores and paints using overflowed inner rect instead of clamped panel rect
When a resizable `Panel`/`SidePanel` contains overflowing content, the visible panel is clipped to `panel_rect`, but the code later stores state and paints the separator using `inner_response.response.rect`.

In practice this can make the resize boundary behave as if the panel were wider than the visible clipped area:
- the resize line appears at the far edge of the hidden overflowed content
- hover/drag for the panel edge feels offset from the visible panel boundary
- the persisted panel size can reflect the overflowed inner rect instead of the clamped panel rect

I verified the same logic is still present in `egui 0.34.1` in `src/containers/panel.rs`.

The minimal fix that solved it for me locally was:

```rust
let rect = inner_response.response.rect.intersect(panel_rect);
```

instead of:

```rust
let rect = inner_response.response.rect;
```

That keeps cursor updates, `PanelState`, and separator painting aligned with the actual visible/clamped panel bounds.

If helpful, I can open a PR with this minimal change.

enomado:
Fixes #8055.

When panel contents overflow beyond the visible clamped bounds, the panel response can become larger than the actual visible panel. That oversized response was then reused by integrations for pointer hit-testing, and its rect was also reused internally for cursor advancement, stored panel state, separator painting, and top-level panel allocation.

This change clamps the outward-facing panel response back to the visible panel bounds with `intersect(panel_rect)`, for both `response.rect` and `response.interact_rect`. The fix is applied to both `SidePanel` and `TopBottomPanel`, so callers using the returned panel response do not inherit oversized clipped-child bounds.

I verified this against a Bevy + bevy_egui integration where a resizable panel with clipped overflow content incorrectly blocked scene interaction and displaced its effective boundary. After this change, the panel hit area and boundary stay aligned with the visible panel edge.

### DIFF
--- a/crates/egui/src/containers/panel.rs
+++ b/crates/egui/src/containers/panel.rs
@@ -693,6 +693,11 @@ impl Panel {
             add_contents(ui)
         });
 
+        let mut inner_response = inner_response;
+        inner_response.response.rect =
+            inner_response.response.rect.intersect(panel_sizer.panel_rect);
+        inner_response.response.interact_rect =
+            inner_response.response.interact_rect.intersect(panel_sizer.panel_rect);
         let rect = inner_response.response.rect;
 
         {
@@ -779,7 +784,7 @@ impl Panel {
             .widget_info(|| WidgetInfo::new(WidgetType::Panel));
 
         let inner_response = self.show_inside_dyn(&mut panel_ui, add_contents);
-        let rect = inner_response.response.rect;
+        let rect = inner_response.response.rect.intersect(available_rect);
 
         match side {
             PanelSide::Vertical(side) => match side {
--- a/crates/egui/src/containers/panel.rs
+++ b/crates/egui/src/containers/panel.rs
@@ -694,10 +694,14 @@ impl Panel {
         });
 
         let mut inner_response = inner_response;
-        inner_response.response.rect =
-            inner_response.response.rect.intersect(panel_sizer.panel_rect);
-        inner_response.response.interact_rect =
-            inner_response.response.interact_rect.intersect(panel_sizer.panel_rect);
+        inner_response.response.rect = inner_response
+            .response
+            .rect
+            .intersect(panel_sizer.panel_rect);
+        inner_response.response.interact_rect = inner_response
+            .response
+            .interact_rect
+            .intersect(panel_sizer.panel_rect);
         let rect = inner_response.response.rect;
 
         {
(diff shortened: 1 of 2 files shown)


emilk:
Could you maybe make a regression test for this in `tests/egui_tests`?

enomado:
Sorry for this mess with history. perhaps test is added. I checked up screenshots - sort of reproducing the ptoblem

enomado:
Related issues that may benefit from or overlap with this fix:

- #7930 — SidePanel resize direction reverses when pointer crosses anchor edge
- #4384 — SidePanel resizing behaves strange when going out of bounds
- #4050 — `set_clip_rect` draws over `SidePanel` when children widgets exceed the available space
- #2962 — button in a context_menu menu will not handle click if the button is partially out of panel
- #901 — CentralPanel overflows in a window unless there is a Top, Bottom or Right panel
- #3385 — ScrollAreas can overrun their container's margins
- #1579 — SidePanel with ScrollAreas and no margin cause strange shrinking issues

valadaptive:
This is necessary to prevent a *regression* introduced in v0.34, where the left panel in my application starts growing automatically:

https://github.com/user-attachments/assets/c89e17ae-d449-4ff8-ab64-da6221e371ce

emilk (approved):

emilk:
This produced some regressions in some existing screenshot tests. I think we may need to dig a bit deeper here. Did anyone do a git bisect to see when this was introduced? Or is this because of a difference in behavior between `Panel::show` (`Context`) and `Panel::show_inside` (`Ui`)?

crates/egui/src/containers/panel.rs
```diff
@@ -779,7 +788,7 @@ impl Panel {
             .widget_info(|| WidgetInfo::new(WidgetType::Panel));
 
         let inner_response = self.show_inside_dyn(&mut panel_ui, add_contents);
-        let rect = inner_response.response.rect;
+        let rect = inner_response.response.rect.intersect(available_rect);
```
valadaptive, line 791: Is this part necessary? `inner_response.response.rect` is already clamped by `show_inside_dyn`, which we call directly above this line.

emilk:
Please see if this is a better fix:
* https://github.com/emilk/egui/pull/8198

PR #8068, opened 2026-04-05
Fully support mobile keyboard in web `TextEdit`

cjgriscom:
Mobile keyboard (e.g. Gboard, MS SwiftKey) support is limited and glitchy (see #8046).  The keyboard state resets after every input, causing capitalization bugs on SwiftKey and blocking useful mobile functionality.  The old resetting workaround had been applied to solve word duplication bugs, due to the lack of input handlers for mobile autocomplete.

This PR modifies the input handler to fully support mobile text suggestions and autocomplete.  It does so by keeping the `egui` text buffer synchronized with `text_agent`'s DOM to provide the keyboard with the necessary context.

There were two ways to approach this:
 * Adjust `text_agent` to support full multiline buffer sync using `<textarea>`
 * Modern `EditContext` API

I opted for replacing the <input> with a <textarea> because support for `EditContext` is missing in web-sys and non-Chromium browsers.

Buffer synchronization is challenging due to the JS UTF-16 conversion.  I use a UTF-16 code-unit counting routine to ensure the cursor positions remain correct.

I am unable to test on iOS but have confirmed correct functionality on these targets:

Chrome, Firefox on Linux
Chrome mobile and Firefox mobile with Samsung Keyboard, SwiftKey, Gboard

* Related: #8045
* Supersedes: #8047
* Closes <https://github.com/emilk/egui/issues/8046>

Screenshot_20260405_134704_Firefox

### DIFF
--- a/crates/egui/src/memory/mod.rs
+++ b/crates/egui/src/memory/mod.rs
@@ -903,11 +903,16 @@ impl Memory {
     /// Give keyboard focus to a specific widget.
     /// See also [`crate::Response::request_focus`].
     ///
-    /// Calling this will interrupt IME composition.
+    /// If focus moves to a different widget, IME composition is interrupted.
+    /// Repeated calls for the widget that already has focus do not interrupt IME
+    /// (e.g. dragging to select text in a [`crate::TextEdit`]).
     #[inline(always)]
     pub fn request_focus(&mut self, id: Id) {
+        let already_focused = self.focus().and_then(|f| f.focused()) == Some(id);
         self.focus_mut().focused_widget = Some(FocusWidget::new(id));
-        self.interrupt_ime();
+        if !already_focused {
+            self.interrupt_ime();
+        }
     }
 
     /// Surrender keyboard focus for a specific widget.


umajho:
Nice work.

I also considered synchronizing the entire text between egui and `TextEdit`. However, since the original approach is $O(1)$ while full synchronization is $O(\text{text size})$, I was concerned about potential performance degradation and decided not to go for it.
In practice, though, given that egui is an immediate-mode UI, the impact is likely negligible. Since I haven't benchmarked it, my thought may well be a case of premature optimization. (My approach (#8045) is also $O(N)$, but $N$ is limited to a small chunk of text being written, which should remain modest in typical usage.)

One clear advantage of this PR over both the original approach and my own (#8045) is that it enables auto-completion to continue working when the cursor is moved. While this can be partially worked around on top of my approach (e.g., by prepending text after resetting the text agent), this PR provides a more robust solution.

cjgriscom:
@umajho That's a good point. I like your idea, but it wasn't working for me on the latest commit (keyboard is still resetting?).  I opted for the simplest approach, but maybe partial text replacement is superior if you can manage the unicode difficulties.

@emilk Thanks for looking at it - I rebased on main to resolve the conflict

rustbasic:
@umajho @cjgriscom

Regarding #8068, when typing the Korean word "않았다" on Android (using the Cheonjiin keyboard), a part of the first character is duplicated, resulting in "안않았다."

I’ve confirmed that this issue is resolved in #8045. It seems that some additional work might be needed to apply this fix to #8068 as well.

cjgriscom:
@umajho @rustbasic Is it just a matter of applying this: https://github.com/emilk/egui/pull/8045/commits/f139c604bfcde8ccffa103d33128af6de35a230c ?

umajho:
> @umajho … but it wasn't working for me on the latest commit (keyboard is still resetting?). …
> 
> …

That's odd. #8045 appears to fix the keyboard reset issue based on my testing:

|Prior to #8045|With #8045|
|-|-|
|2026-04-06 10 46 09 PM|2026-04-06 10 46 25 PM|

rustbasic:
@umajho @cjgriscom

Yes, it appears that the keyboard reset issue has also been resolved in #8045.

umajho:
> … Is it just a matter of applying this: [f139c60](https://github.com/emilk/egui/commit/f139c604bfcde8ccffa103d33128af6de35a230c) ?

The fix in #8045 was under the assumption that only a small portion of the text exists in the text agent. Specifically, the text originally retained in the text agent is compared with the text updated in the text agent during IME composition, and based on that comparison, it determines what action to take.

The code changes to `on_composition_end` in https://github.com/emilk/egui/commit/f139c604bfcde8ccffa103d33128af6de35a230c are only part of the fix: changes in places like `on_input` are also part of it.

To be honest, since the assumption has now changed to the text agent containing the entire synchronized text, I'm not really sure how to migrate the fixes implemented in my PR anymore…

cjgriscom:
@umajho You're right, it seems to work on the demo now.  I must have done something wrong

IME is bigger in scope than what I'd like to address here, so I'd rather defer to your PR in that case.  I left a note there about one remaining issue.

PR #8078, opened 2026-04-07
Delegate handling of IME interruptions to integrations to fix virtual keyboard flickering on web

umajho:
* Closes N/A
* Partially replaces #7983
* Related: #8045

## Details

In #7983, I modified `Memory::request_focus` to interrupt any ongoing IME composition. This fixed a bug where clicking inside an already focused `TextEdit` failed to cancel the active composition, resulting in duplicated text:

https://github.com/emilk/egui/pull/8045#issuecomment-4193310616

To avoid introducing API changes in that PR, I ensured the IME state was reset by forcing `PlatformOutput::ime` to `None` for at least one frame.
While this works well on desktop platforms, it causes virtual keyboard flickering on the web:

https://github.com/emilk/egui/pull/8045#issuecomment-4193035008

In this PR, I delegate the responsibility for handling IME composition interruptions to integrations, allowing each integration to decide how to interrupt compositions in a flexible manner.

### The new field `should_interrupt_composition` on `IMEOutput`.

Instead of introducing a new `OutputCommand` variant, this PR adds a new field `should_interrupt_composition` to `IMEOutput`.

Interrupting an active composition is only meaningful when IME remains allowed. If IME should be disabled altogether, `PlatformOutput::ime` can simply be set to `None`.
Given this, IMO, it is more appropriate to attach the interrupt signal to `IMEOutput` (i.e., the type of `PlatformOutput::ime`).

### DIFF
--- a/crates/eframe/src/web/text_agent.rs
+++ b/crates/eframe/src/web/text_agent.rs
@@ -138,7 +138,7 @@ impl TextAgent {
         let Some(ime) = ime else { return Ok(()) };
 
         if ime.should_interrupt_composition {
-            // no-op for now: currently ,the text agent is sizeless, so any
+            // no-op for now: currently, the text agent is sizeless, so any
             // click shifts focus to the canvas, which naturally interrupts the
             // composition.
         }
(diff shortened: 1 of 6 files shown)


cjgriscom:
This solves the flickering on my end.  And seems like a good solution

emilk (approved):

PR #8082, opened 2026-04-08
Fix centered & right aligned `TextEdit`

### LINKED ISSUE #7999: Regression: #7587 breaks `TextEdit.horizontal_align()`

lucasmerlin:
A couple improvements to centered and right-aligned text edits:
- Fix text selection in centered and right aligned text edits (ironically, this broke in #8076) 
- Fix cursor movement in centered and right aligned text edits (horizontal cursor position will be retained on vertical movement)
- Multiline text edit exceeding available width if there are atoms
- Added atoms & alignment options to text edit demo
- Improve how vertical_align and horizontal_align are applied
  - Textedit atom is grow now, removing the need for the extra seperate grow atom
  - This allows us to apply the `align` on the text edit atom instead of the whole AtomLayout
  - Fixes https://github.com/emilk/egui/pull/8022
  - Fixes https://github.com/emilk/egui/issues/7999


lucasmerlin:
~Oh, found a bug:~

EDIT: Bug is now fixed

https://github.com/user-attachments/assets/c31bb74c-9d04-4e61-b9b2-06952221577f

crates/epaint/src/text/text_layout_types.rs
```diff
@@ -79,6 +79,14 @@ pub struct LayoutJob {
...
     /// Round output sizes using [`emath::GuiRounding`], to avoid rounding errors in layout code.
     pub round_output_to_gui: bool,
+
+    /// If `false` (default), trailing whitespace is ignored when computing
+    /// horizontal alignment ([`Self::halign`]).
+    /// This is desirable for labels so that e.g. "Hello " centers the same as "Hello".
+    ///
+    /// If `true`, trailing whitespace is included in the row width used for alignment.
+    /// This is desirable for text editors where the user expects to see their spaces.
+    pub keep_trailing_whitespace: bool,
```
emilk: Is this the only breaking change? If so, we should split the PR in two so we can include everything else in a patch release

lucasmerlin: I think so, yes. I'll be back on monday, can do it then

emilk (approved):
Awesome!

PR #8083, opened 2026-04-08
Implement proper visuals for IME composition

umajho:
* Closes N/A

This PR adds visual support for IME composition, including the cursor and conversion segment.
These visuals works (mostly) well on native platforms (`egui-winit`). On the web (`eframe/web`), support is limited by browser capabilities: Chromium works well, Firefox shows partial improvement, and Safari remains subpar.

> [!NOTE]
>
> On Windows, the legacy visuals are still enabled by default due to a known `winit` issue that reports incorrect cursor indices for Korean IMEs.
> To use the new visuals on Windows, set `ui.style.visuals.ime_composition.legacy_visuals` to `false`.
>
> When the new visuals are enabled on Windows, the Korean IME issue is worked around by not showing the cursor when its reported position is known to be incorrect. As a side effect, the cursor is also hidden at the beginning of every composition, even when the reported position is actually correct.
>
> On all other platforms, the new visuals introduced in this PR are enabled by default.

## Details

We extend `egui::ImeEvent::Preedit(String)` to `egui::ImeEvent::Preedit { text: String, active_range_chars: Option<std::ops::Range<usize>> }`.
The new `active_range_chars` field enables rendering of:
- the cursor (when the range is empty), and
- the conversion segment (when the range is non-empty)

in IME composition.

In `egui-winit`, we now use the range provided by `winit::event::Ime::Preedit` instead of ignoring it.

In `eframe/web`, we derive the range from `selectionStart` and `selectionEnd` on the text agent. This mapping is fully accurate only in Chromium, but represents the best available approach for now.

### Chinese IMEs (Shuangpin)

We can see where the cursor is now.

| What | With this PR | Without this PR |
|-|-|-|
| macOS builtin |<video src=https://github.com/user-attachments/assets/487c7e7c-ef6d-4a86-8dbc-8c71871b4470 />|<video src=https://github.com/user-attachments/assets/49bd5a60-4b90-4e4a-99e0-cd01d3f7030c />|
| macOS builtin (light)|<video src=https://github.com/user-attachments/assets/e84546f6-947b-4cea-a87e-fda903f49164 />|——|
| Windows builtin |<video src=https://github.com/user-attachments/assets/fd331884-1f0c-4822-a99e-8140aed54815 />|——|
| Wayland iBus Intelligent Pinyin |<video src=https://github.com/user-attachments/assets/b6796c75-1c4e-45e5-b43a-5d8dea320485 />|——|
| Chromium (Chrome) macOS | Identical to `macOS builtin`. |——|
| Safari macOS | We can now differentiate between IME composition and text selection, but we still can't tell where the cursor is. |——|
| Firefox (Zen) macOS | Identical to `macOS builtin`. |——|

### Japanese IMEs

We can see where the conversion segment is now.

| What | With this PR | Without this PR |
|-|-|-|
| macOS builtin |<video src=https://github.com/user-attachments/assets/f2994cd4-a966-4ff0-9590-d263c202ec76 />|<video src=https://github.com/user-attachments/assets/7cf5ff35-003d-4f60-8fbf-15c725c3ecb9 />|
| macOS builtin (light)|<video src=https://github.com/user-attachments/assets/6f562bdd-12fc-4486-b37b-8fcf11643295 />|——|
| Windows builtin |<video src=https://github.com/user-attachments/assets/f0905659-5335-4034-abda-c25cf8f2fd57 />|——|
| Wayland iBus Anthy |<video src=https://github.com/user-attachments/assets/94cd3a24-3158-4d79-ae02-d9b30fdfa738 />|——|
| Chromium (Chrome) macOS | Identical to `macOS builtin`. |——|
| Safari macOS | We can now differentiate between IME composition and text selection, but we still can't tell where the conversion segment is. |——|
| Firefox (Zen) macOS | (Limited improvement.) <video src=https://github.com/user-attachments/assets/3daf9b63-6e75-467b-8515-31c2a44adf61 /> |——|

### Korean IMEs

We can clearly tell whether we are in composition (in contrast to selection) now.

| What | With this PR | Without this PR |
|-|-|-|
|macOS builtin|<video src=https://github.com/user-attachments/assets/73ca28c7-22a0-493f-8f4d-c6e59a2dec54 />|<video src=https://github.com/user-attachments/assets/f582de7d-7ec0-48fe-910f-0139ef1620d3 />|
|macOS builtin (light)|<video src=https://github.com/user-attachments/assets/269f03bd-6f95-498b-9fb1-1adcb043c738 />|——|
| Windows builtin| (With a workaround for [this `winit` bug](https://github.com/emilk/egui/pull/8083#issuecomment-4206742668) applied.) <video src=https://github.com/user-attachments/assets/1e82583d-0c41-4f1c-98cf-0606bee5af05 />|——|
| Wayland iBus Hangul |<video src=https://github.com/user-attachments/assets/8c9a0de1-9027-4b37-93a3-e9da0251d176 />|——|
| Chromium (Chrome) macOS | Identical to `macOS builtin`. |——|
| Safari macOS | Identical to `Windows builtin`. |——|
| Firefox (Zen) macOS | Identical to `macOS builtin`. (ignoring the fact that the composition breaks when typing the second Hangul. (This bug predates this PR.)) |——|


umajho:
On Windows, the cursor is visually incorrectly positioned when using the built-in Korean IME:
2026-04-08 9 49 45 PM

This is due to `winit` reporting incorrect cursor positions. Specifically, the values are returned as `Some((0, 0))`, but should instead be `Some((1, 1))`:
2026-04-08 9 49 23 PM

TODO: report this bug to `winit`.

See: https://issues.chromium.org/issues/41279184

crates/egui/src/data/input.rs
```diff
@@ -612,7 +612,10 @@ pub enum ImeEvent {
     ///
     /// An empty preedit string indicates that the IME has been dismissed, while
     /// a non-empty preedit string indicates that the IME is active.
-    Preedit(String),
+    Preedit {
+        text: String,
+        active_range_chars: Option<std::ops::Range<usize>>,
+    },
```
umajho, line 618: I'd actually prefer to define it as:
```rs
    Preedit {
        text: String,
        active_range: Option<CCursorRange>,
    },
```

However, since `CCursorRange` doesn't implement `Eq`, this leaves a few options:
- Use `Option<std::ops::Range<usize>>`. (the current approach)
- Remove `#[derive(Eq)]` from `ImeEvent`. This is my preferred option, as I don't see a strong need for `ImeEvent` to implement `Eq`. That said, I'm unsure whether downstream code might rely on this and thus be affected.
- Implement `Eq` for both `CCursorRange` and `CCursor`, but it is unclear to me what equality should mean for these types.
- Implement `Eq` for `ImeEvent` manually without requiring `Eq` on the underlying types, which feels off.

umajho: Another consideration:

If we define `active_range` as `Option<CCursorRange>` and later switch the cursor index base from `char`s to graphemes (as in #62), should `active_range` also change accordingly (e.g., to `Option<GCursorRange>`)?

Alternatively, should we keep a more primitive representation (e.g. `active_range_chars: Option<std::ops::Range<usize>>` or even `active_range_bytes: Option<std::ops::Range<usize>>`) and handle the conversion within `egui`?
This would avoid requiring downstream to pull additional crates (e.g. `unicode-segmentation`), which could introduce duplicated dependencies in binary or inconsistencies with the ones used internally by `egui`.

rustbasic:
@umajho 

In Korean, the text cursor should not be placed before a character; it should be placed after the character.

umajho:
> In Korean, the text cursor should not be placed before a character; it should be placed after the character.

Yes, I mentioned this earlier in <https://github.com/emilk/egui/pull/8083#issuecomment-4206742668>, where I described it as a bug. As a workaround, I hid the cursor when it is at the beginning.

As I understand it, during Korean composition, the cursor only appears at the end of the composition text. So users should still be able to identify the cursor position by looking at the end of the underline, and since the cursor is hidden, it should not be a distraction.

Did this workaround fail to take effect for you?

crates/epaint/src/text/cursor.rs
```diff
@@ -69,6 +69,13 @@ impl std::ops::SubAssign<usize> for CCursor {
     }
 }
 
+impl PartialOrd for CCursor {
+    #[inline]
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        self.index.partial_cmp(&other.index)
+    }
+}
```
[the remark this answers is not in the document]
umajho: I decided not to implement `Ord` and `PartialOrd` for `CCursor`, since it is unclear how `prefer_next_row` should be handled during comparison. Instead, I implemented [`is_cursor_range_empty`](https://github.com/emilk/egui/pull/8083/changes/c8439f3150bec6a4aff2f0c42262a2265dd74014#diff-eee5d8f2dbb36616b4bbe4a57f319a60d50f8e7f0440a4dc6c91a379d4176f8dR143-R152) directly in `paint_ime_preedit_text_visuals`.

rustbasic:
@umajho 

Sorry for the late reply, I've been working on a major refactoring of my program.
With the following modification, it now works perfectly with the Windows Korean IME.

```
            winit::event::Ime::Preedit(text, active_range_bytes) => {
                // Work around a Windows Korean IME/winit bug: `(0, 0)` should point to the preedit end.
                let active_range_bytes = if cfg!(target_os = "windows")
                    && !text.is_empty()
                    && *active_range_bytes == Some((0, 0))
                {
                    Some((text.len(), text.len()))
                } else {
                    *active_range_bytes
                };

                let active_range_chars = match active_range_bytes {
```

umajho:
> @umajho
> 
> Sorry for the late reply, I've been working on a major refactoring of my program.

No worries. I'm in a similar situation myself lately.

> With the following modification, it now works perfectly with the Windows Korean IME.
> 
> ```
>             winit::event::Ime::Preedit(text, active_range_bytes) => {
>                 // Work around a Windows Korean IME/winit bug: `(0, 0)` should point to the preedit end.
>                 let active_range_bytes = if cfg!(target_os = "windows")
>                     && !text.is_empty()
>                     && *active_range_bytes == Some((0, 0))
>                 {
>                     Some((text.len(), text.len()))
>                 } else {
>                     *active_range_bytes
>                 };
> 
>                 let active_range_chars = match active_range_bytes {
> ```

Currently, I worked around the Windows Korean IME bug differently by treating `(0, 0)` as “no cursor information” rather than interpreting it as a cursor at the end of the range:
https://github.com/emilk/egui/pull/8083/changes#diff-eee5d8f2dbb36616b4bbe4a57f319a60d50f8e7f0440a4dc6c91a379d4176f8dR164-R177
(`visuals.rs` line 164~177. For some reason, the link doesn't jump to the correct location in my browser.)

The current workaround already broke Chinese IMEs a bit: previously, when the cursor was at the beginning of the preedit text, it could be render correctly. With the workaround in place, the cursor is no longer shown in that case.

Unfortunately, if the workaround you proposed were applied, the behavior for Chinese IMEs would become more problematic: when the cursor is actually at the beginning of the preedit text, it would be rendered at the end instead, which is actively misleading.

In conclusion, I think the current workaround is a relatively better compromise for now. It does lose some cursor information in this edge case, but it avoids displaying an incorrect cursor position.

rustbasic:
@umajho
Please add a Korean language option to the `settings_ui()` function, and I would like the cursor to work properly. Koreans are used to selecting options.

umajho:
> @umajho Please add a Korean language option to the `settings_ui()` function, and I would like the cursor to work properly. Koreans are used to selecting options.

I am still opposed to handling IME logic differently based on language, for the reasons I outlined [here](https://github.com/emilk/egui/pull/7898#issuecomment-3886843136):

> 1. Lots of, if not most, apps support IMEs correctly without knowing which kind of IME is used, this should also be possible for egui.
> 2. I doubt most developers will know/care that there is an IME language selection thing in egui and bother integrating it into their apps. If an app doesn't make the selector available, users are out of luck.
> 3. And even if an app includes the selector, users still need to discover it, which is not ideal for UX.
> 4. There are users who use multiple IMEs all the time. To them, switching IME language settings is cumbersome.
> 5. As I stated, I think the root issue is likely in `winit`.

As an alternative, perhaps we could disable this feature on Windows for now and re-enable it once the underlying issue in `winit` has been fixed.

(By the way, are there any situations in which the cursor is not positioned at the end of the preedit text while composing Korean text?)

rustbasic:
Windows needs to disable this feature or provide an option to disable it.
It would be good if you could select the style of the text cursor.

Since only a single character is modified in Korean, there are no particular issues, but it is difficult to test all cases.

umajho:
> It would be good if you could select the style of the text cursor.

Could you elaborate on this?

umajho:
Since the Windows Korean IME bug is specific to `winit`, I moved the workaround from `egui` to `egui-winit` and introduced a `legacy_ime_composition_visuals` feature in `egui`.

`legacy_ime_composition_visuals` is disabled by default because it exists solely as a workaround for backend-specific bugs. We should not assume that every platform backend has the same issue. In this case, the problem is limited to `egui-winit` on Windows.

Since `eframe` always depends on `egui-winit`, it enables `legacy_ime_composition_visuals` by default on Windows. We can still opt into the new composition visuals by enabling the `windows_new_ime_composition_visuals` feature in `eframe`.

crates/egui/Cargo.toml
```diff
@@ -44,6 +44,21 @@ color-hex = ["epaint/color-hex"]
...
+## matter where the actual cursor is, which makes composing Chinese and Japanese
+## text confusing.
+##
+## However, the new composition visuals are not yet fully reliable either. For
+## example, on Windows, `winit` currently reports an incorrect cursor position
+## for Korean IMEs: the cursor should be at the end of the composition, but
+## `winit` reports it at the beginning. Therefore, we keep the legacy visuals
+## as an option for now and plan to remove them once the new visuals work
+## correctly across all supported platforms.
+legacy_ime_composition_visuals = []
```
lucasmerlin: I think having this as a feature, and especially also having an `_override_legacy_ime_composition_visuals` override feature should be avoided. Maybe we could instead add this as a config to `egui::Options`? Then apps could even switch it on runtime e.g. depending on the language of the user, if there is a known language where it causes problems. 
And the default could be disabled on windows.

crates/eframe/src/web/text_agent.rs
```diff
@@ -69,15 +69,55 @@ impl TextAgent {
...
+                        .unwrap_or(None)
+                        .map(|pos| pos as usize);
+                    let active_range_chars = if let Some(selection_start) = selection_start
+                        && let Some(selection_end) = selection_end
+                    {
+                        let text_utf16 = text.encode_utf16().collect::<Vec<u16>>();
+                        let text_before_selection =
+                            String::from_utf16_lossy(&text_utf16[..selection_start]);
+                        let text_in_selection =
+                            String::from_utf16_lossy(&text_utf16[selection_start..selection_end]);
```
lucasmerlin, line 98: This is probably always fine, but it'd still be better to be defensive here by calling `.get()` since this would panic if there is a invalid selection start/end

lucasmerlin (commented):
Awesome, thanks for the exhaustive testing report!

lucasmerlin (approved):
Awesome!

crates/egui/Cargo.toml
umajho: Ah, that totally makes sense.
I was so focused on avoiding introducing a language-related runtime option that would affect IME handling logic, that I instinctively dismissed using runtime mechanisms for this altogether. But since a `legacy_visuals` option is not language-specific, it indeed is a much better fit here.

PR #8088, opened 2026-04-09
Fix grapheme cluster glyph count to restore cursor/selection invariant

### LINKED ISSUE #8087: Inputting grapheme clusters breaks IME composition & grapheme clusters break text selection since #8031
**Describe the bug**

While testing #8083 with the macOS builtin Japanese IME, I found a regression introduced after #8031 was merged: When the Japanese flag is selected as the active candidate, the last character (`五` in this case) in the composition is duplicated:
| ✅: 33e89e33bec3e0009096323f0832e73265c29e4c (one commit prior, when #7983 is merged) | ❌: 16cad760a51497bb14202c4f61ca49e418f4f895 (after #8031 is merged) | ❌: main + #8083 (included to help illustrate the behavior) |
|-|-|-|
|<video src=https://github.com/user-attachments/assets/5087fc20-bbd9-4f92-90be-5a32f2497491 />|<video src=https://github.com/user-attachments/assets/0952c709-6f5d-45c8-879d-db099b31c9c2 />|<video src=https://github.com/user-attachments/assets/f5169c5c-4f75-4223-9b9f-3613ed526698 />|

It appears that the issue affects all grapheme clusters, not just flags. In addition, those grapheme clusters break text selection:
| ✅: 33e89e33bec3e0009096323f0832e73265c29e4c (one commit prior, when #7983 is merged) | ❌: 16cad760a51497bb14202c4f61ca49e418f4f895 (after #8031 is merged) | ❌: main + #8083 (included to help illustrate the behavior) |
|-|-|-|
|<video src=https://github.com/user-attachments/assets/d257013c-59f9-4c0d-b5af-c518e3b56aab />|<video src=https://github.com/user-attachments/assets/d4d613a9-ac8f-4f37-8275-822dd50dfd0a />|<video src=https://github.com/user-attachments/assets/fae364c0-1362-43b0-b8bc-8d56a424b547 />|

All platforms, including native and web are affected.

gcailly:
May close #8087, but cannot test macOS builtin Japanese IME.

## Summary

PR #8031 (harfrust text shaping) introduced a regression: when harfrust shapes multi-codepoint clusters (flag emojis, ligatures, combining marks) into fewer glyphs than input characters, the invariant `glyphs.len() == char_count` breaks. This causes IME composition to duplicate characters and text selection to behave incorrectly.

## Fix

In `layout_shaped_run()`, after emitting shaped glyphs for a cluster, we now check if the cluster had more characters than glyphs. If so, zero-width "continuation" glyphs are emitted for the extra characters, restoring the 1:1 glyph-to-character mapping.

Continuation glyphs have `UvRect::default()` (`is_nothing() == true`), so `tessellate_glyphs` skips them entirely. Background, underline, and strikethrough rendering handle zero-width glyphs naturally.

Only `crates/epaint/src/text/text_layout.rs` is modified. No changes to cursor logic, selection code, or public API.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p epaint --tests`
- [x] `cargo test -p epaint -p egui` (all pass)
- [x] New test `test_grapheme_cluster_glyph_count`: verifies glyph count == char count for flag emojis, combining marks, and plain ASCII
- [x] New test `test_grapheme_cluster_cursor_roundtrip`: verifies cursor position stability through `pos_from_cursor` -> `cursor_from_pos` round-trips on text containing flag emojis
- [x] Manual testing with demo app: selection and cursor navigation work correctly on `A🇯🇵B`
- [ ] IME testing (macOS Japanese IME) needs to be validated by someone on macOS

---

**This PR was developed with the assistance of Claude Code.**

### DIFF
--- a/crates/epaint/src/text/text_layout.rs
+++ b/crates/epaint/src/text/text_layout.rs
@@ -403,8 +403,7 @@ fn emit_continuation_glyphs(
     ctx: &ShapingContext,
     paragraph: &mut Paragraph,
     run_text: &str,
-    cluster_start_byte: usize,
-    cluster_end_byte: usize,
+    cluster_bytes: Range<usize>,
     cluster_glyph_count: usize,
     face_metrics: &StyledMetrics,
 ) {
--- a/crates/epaint/src/text/text_layout.rs
+++ b/crates/epaint/src/text/text_layout.rs
@@ -1,5 +1,6 @@
 #![expect(clippy::unwrap_used)] // TODO(emilk): remove unwraps
 
+use std::ops::Range;
 use std::sync::Arc;
 
 use emath::{Align, GuiRounding as _, NumExt as _, Pos2, Rect, Vec2, pos2, vec2};
@@ -287,8 +288,7 @@ fn layout_shaped_run(
                     ctx,
                     paragraph,
                     run_text,
-                    cluster_start_byte,
-                    cluster as usize,
+                    cluster_start_byte..cluster as usize,
                     cluster_glyph_count,
                     face_metrics,
                 );
@@ -385,8 +385,7 @@ fn layout_shaped_run(
             ctx,
             paragraph,
             run_text,
-            cluster_start_byte,
-            run_text.len(),
+            cluster_start_byte..run_text.len(),
             cluster_glyph_count,
             face_metrics,
         );
@@ -407,7 +406,7 @@ fn emit_continuation_glyphs(
     cluster_glyph_count: usize,
     face_metrics: &StyledMetrics,
 ) {
-    let Some(cluster_text) = run_text.get(cluster_start_byte..cluster_end_byte) else {
+    let Some(cluster_text) = run_text.get(cluster_bytes) else {
         return;
     };
     let char_count = cluster_text.chars().count();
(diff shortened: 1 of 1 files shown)


umajho:
It appears that this also fixes the issue during IME composition.

crates/epaint/src/text/text_layout.rs
```diff
@@ -353,6 +376,52 @@ fn layout_shaped_run(
...
+///
+/// This preserves the invariant `glyphs.len() == char_count` that all cursor
+/// and text-selection code depends on. Continuation glyphs have
+/// [`UvRect::default()`] so [`tessellate_glyphs`] skips them entirely.
+fn emit_continuation_glyphs(
+    ctx: &ShapingContext,
+    paragraph: &mut Paragraph,
+    run_text: &str,
+    cluster_start_byte: usize,
+    cluster_end_byte: usize,
```
emilk: nit/style: less error prone signature

```suggestion
    cluster_bytes: Range<usize>,
```

emilk (approved):
makes sense - thanks!

PR #8089, opened 2026-04-11
Call `pre_present_notify` before presenting

dimtpap:
Add a call to [`Window::pre_present_notify`](https://docs.rs/winit/0.30.13/winit/window/struct.Window.html#method.pre_present_notify) before wgpu presents the output frame.

I noticed egui not looking as smooth after switching to wgpu. Lowering my refresh rate revealed more ugliness.

I'm a graphics noob but my guess is that on Wayland with Vulkan some funny synchronization stuff is happening. I can't try wgpu with OpenGL because of https://github.com/gfx-rs/wgpu/issues/4751. On glow this wasn't needed because under the hood it calls `eglSwapBuffers` which (according to https://emersion.fr/blog/2018/wayland-rendering-loop/) does the equivalent of calling `window.pre_present_notify()` automatically.

Below is a before and after of this PR. Before, dragged stuff lag behind the cursor and resizing the window is really slow. After, everything is normal.

https://github.com/user-attachments/assets/60ca0378-4d1b-4013-ac98-54cc1753389b

A side effect of this is that `logic` is no longer called if the app is minimized. I guess it was a happy accident.

### DIFF
--- a/crates/eframe/src/native/wgpu_integration.rs
+++ b/crates/eframe/src/native/wgpu_integration.rs
@@ -722,6 +722,7 @@ impl WgpuWinitRunning<'_> {
                 &clipped_primitives,
                 &textures_delta,
                 screenshot_commands,
+                window,
             );
 
             for action in viewport.actions_requested.drain(..) {
@@ -1111,6 +1112,7 @@ fn render_immediate_viewport(
         &clipped_primitives,
         &textures_delta,
         vec![],
+        window,
     );
 
     egui_winit.handle_platform_output(window, platform_output);
--- a/crates/egui-wgpu/src/winit.rs
+++ b/crates/egui-wgpu/src/winit.rs
@@ -411,6 +411,7 @@ impl Painter {
     /// and the captures captured screenshot if it was requested.
     ///
     /// If `capture_data` isn't empty, a screenshot will be captured.
+    #[expect(clippy::too_many_arguments)]
     pub fn paint_and_update_textures(
         &mut self,
         viewport_id: ViewportId,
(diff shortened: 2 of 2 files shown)


Wumpf (approved):
Woah, that's a huge difference in your video! Great catch, didn't know about `pre_present_notify`!

According to docs this only has an effect on Wayland
https://docs.rs/winit/latest/winit/window/struct.Window.html#method.pre_present_notify
Got curious and traced a bit deeper what it does, not sure I followed it right, but here's my journey for posterity
* https://github.com/rust-windowing/winit/blob/v0.30.13/src/platform_impl/linux/wayland/window/mod.rs#L302 calls `self.window_state.lock().unwrap().request_frame_callback();`
* which flips `frame_callback_state`and calls `surface.frame` https://github.com/rust-windowing/winit/blob/v0.30.13/src/platform_impl/linux/wayland/window/state.rs#L253
* this in turn flips a dispatched_events_state https://github.com/rust-windowing/winit/blob/v0.30.13/src/platform_impl/linux/wayland/state.rs#L392
* this state is used to suppress spurious wakeups https://github.com/rust-windowing/winit/blob/v0.30.13/src/platform_impl/linux/wayland/event_loop/mod.rs#L317

So from the sound of it that should still have quite the effect on wgpu @ GL as well.

Don't have a wayland enabled linux install at hand to test myself, but I see no reason not to take this, can't see any downsides 🤷 

We could maybe add a bit more docs what the call is and that it should be as close as possible to `present`. Also it's a bit annoying that we pass in the window here, reaching the `too_many_arguments` limit, but I don't want to block over any of that.

PR #8102, opened 2026-04-14
Don't hide whitespaces in centered and right aligned text edits

lucasmerlin:
When enabled, trailing whitespace is included in the row width used for horizontal alignment. This is useful for text editors where hiding trailing spaces feels wrong when typing.

By default this is `false`, preserving existing behavior where trailing whitespace is stripped for alignment (so "Hello " centers the same as "Hello").

* follow up to https://github.com/emilk/egui/pull/8082


emilk (approved):
Makes sense!

Issue #7761, opened 2025-12-09
WGPU backend transition: performance, issues and suggestions


RndUsr123:
Since WGPU will be the default eframe backend in 0.34 (#5889 seems to hint it's going to become the only one eventually), I thought it would be a good idea to collect user feedback in a single issue. The goal is to identify beaking changes or other aspects that may likewise cause friction. I already found a couple of my own of those and also took the time to measure footprint and performance in case anyone is curious; meaurements are taken exclusively in Windows, and specifically relate to the example provided here: https://docs.rs/eframe/latest/eframe/fn.run_simple_native.html

### BINARY SIZE (known issue) ###
```
Glow (release profile) = 14.2 MB
WGPU (release profile) = 27.2 MB
Glow (release + LTO full) = 4.71 MB
WGPU (release + LTO full) = 11.3 MB
```
### COMPILE TIMES (known issue) ###

The actual numbers depend on CPU performance, but the general trend should remain consistent across systems
```
Glow (release profile) = ~23s 
WGPU (release profile) = ~67s
Glow (release + LTO full) = ~36s 
WGPU (release + LTO full) = ~120s 
```
### MEMORY USAGE ###

RAM - As reported by the task manager. A similar pattern can be observed by measuring the increase/decrease in total usage after running/killing the process. Little to no difference with LTO
```
Glow (OpenGL) = ~38 MB
WGPU (Vulkan) = ~164 MB
WGPU (DX12) = ~133 MB
WGPU (OpenGL) = ~42 MB
```
VRAM - Measured as the decrease in total after fully closing the app, provided by Afterburner. Little to no difference with LTO
```
Glow (OpenGL) = ~40 MB
WGPU (Vulkan), ~60 MB
WGPU (DX12) = ~360 MB
WGPU (OpenGL) = ~40 MB
```
### UNCAPPED PERFORMANCE ###
Very rough measurements in Afterburner + RTSS. Slightly worse without LTO 

```
Glow (OpenGL) = 2250~2350 fps
WGPU (Vulkan) = ??? (not detected)
WGPU (DX12) = 2000~2100 fps
WGPU (OpenGL) = 950~1000 fps
```
### UNSUPPORTED FEATURES, BREAKING CHANGES, GENERAL FRICTION ###
- support for 8x multisampling is not automatically enabled in Vulkan and DX12, even when the hardware is capable of it;
- window transparency is seemingly not supported at all in DX12 and OpenGL, or at the very least requires extra steps I'm not aware of;
[...]

Issue #7096, opened 2025-05-27
High cpu usage even in static apps, mostly unaffected by vsync


RndUsr123:
The facts so far:
- I'm mainly using 3 different apps for testing: the demo, the file dialog example and my own test app;
- Apps are compiled using `--release`, but no meaningful performance degradation was observed with `--debug`;
- Everything mentioned in the following points is assumed to be true for continuous refresh mode only;
- All test apps run over 1500fps uncapped, which tells us what the performance ceiling for my system is;
- The following CPU measurements are from the task manager, which is not 100% accurate but good enough for now;
- CPU usage is not significantly different between 120hz vsync and 24hz vsync, at ~13% and ~12% respectively;
- Both can be dramatically reduced to <3% by capping the framerate to the refresh rate using RTSS;
- CPU usage can also be reduced to ~2.5% by adding a 7ms sleep to update() while maintaining around 120fps;
- Both these methods seem to negatively affect frame pacing and synchronization, but that's expected I think;
- There is no other immediately obvious drawback as the input sampling rate appears to be linked to framerate only;
- CPU usage does increase when fully uncapped (1500~2000fps), but not too much at just ~21%;
- Keep in mind that despite the overall low cpu usage most of the app logic is actually run by a single core;
- I tested several releases of egui, as old as 0.18, all of which exhibit the same issue (within ~1% usage);
- Due to that, I'm sure this is unrelated to #3172 (as suggested in #4173), as that PR wasn't a thing back then.

Theory:
Given CPU usage is very high even for the static file dialog example, where nothing is moving and nothing is happening, given the massive difference between 24fps cap and 24hz vsync and almost no difference between 24hz or 120hz vsync, my theory is that part of the logic is run internally as fast as it possibly can. This would explain the small (but consistent and measurable) difference between 24hz vsync and 120hz vsync as the extra cpu time required to compute 96 more frames. This seems equally consistent with the relatively mild increase from 13 to 21% with uncapped framerate.
As to what may be actually causing this, I have no idea. My first guess was input polling running uncapped and something crazy like 50khz but that doesn't seem to be the case; In the drawing pad from the demo app, for example, the cursor sampling rate is exactly the same regardless of cpu utilization for a given framerate, so it can't be it. Furthermore, input events seem to be closely tied to the visual update frequency anyway, as inferred from issues such as #5229 as well as my (limited) experience with egui.

This doesn't require any code example nor specific conditions as it happens withe every app, settings and even much older egui versions. 
[...]

PR #8103, opened 2026-04-14
Configure wgpu to be low-latency by default

### LINKED ISSUE #5037: Delays in the demo app
**Describe the bug**
In the demo application, there is a delay in the application's response to input: windows move with a delay, elements react to the mouse over them with a delay, and button presses also have a delay.
This issue is present in 0.28.1 and the master branch.

https://github.com/user-attachments/assets/87578e26-c0a6-4493-bbb9-37e5d863bbab

In my application, I also encountered this delay (Windows, Vulkan backend, egui-wgpu 0.28.1, and egui-winit 0.28.1). I found that the delay depends on the backend: it is lower on DX12 than on Vulkan.

I tested everything on release versions.
The web version with egui.rs works fine.

At first, it seemed to me that I was doing something wrong in my application when I encountered this issue, but since it also occurs in the demo application, it seems the problem isn't related to something else; maybe it's with egui-wgpu.

emilk:
This changes the default value of `WgpuConfiguration::desired_maximum_frame_latency` to `Some(1)`. For low-Hz displays, this results in significantly lower input latency.

* Closes https://github.com/emilk/egui/issues/5037 ?
* Related to https://github.com/emilk/egui/issues/7761

### DIFF
--- a/crates/egui-wgpu/src/lib.rs
+++ b/crates/egui-wgpu/src/lib.rs
@@ -328,7 +328,11 @@ impl Default for WgpuConfiguration {
     fn default() -> Self {
         Self {
             present_mode: wgpu::PresentMode::AutoVsync,
-            desired_maximum_frame_latency: Some(1), // Low-latency by default.
+            desired_maximum_frame_latency: if cfg!(target_os = "ios") {
+                None // The default is good on iOS, while `Some(1)` cuts FPS in half
+            } else {
+                Some(1) // Low-latency by default.
+            },
 
             // No display handle available at this point — callers should replace this with
             // `WgpuSetup::from_display_handle(...)` before creating the instance if one is available.
--- a/crates/egui-wgpu/src/lib.rs
+++ b/crates/egui-wgpu/src/lib.rs
@@ -328,7 +328,8 @@ impl Default for WgpuConfiguration {
     fn default() -> Self {
         Self {
             present_mode: wgpu::PresentMode::AutoVsync,
-            desired_maximum_frame_latency: None,
+            desired_maximum_frame_latency: Some(1), // Low-latency by default.
+
             // No display handle available at this point — callers should replace this with
             // `WgpuSetup::from_display_handle(...)` before creating the instance if one is available.
             wgpu_setup: WgpuSetup::without_display_handle(),


Wumpf (commented):
I think this is a really bad idea since it (iirc) completely stops CPU/GPU parallism on most backends. I.e. gpu work can only begin when CPU is done and vice versa. So for anything non-trivial that becomes a huge perf hog

RndUsr123:
This also doesn't really address @mrjbom's original point in #5037 as they stated this doesn't help, something I've also measured makes no meaningful difference.
Furthermore, my own testing in #7761 was already conducted with maximum latency set to 1...

The only thing that seems to realistically affect latency is changing the wgpu backend to OpenGL.
Vulkan and DX12 seem to be around the same with the odd difference here and there (possibly caused by the recording or other external factors), whereas glow and wgpu OpenGL are consistently faster and comparable to each other.

Metal could be fine but I haven't done any testing on MacOS yet so I don't know.

emilk:
Thanks for the feedback @RndUsr123 - maybe the `desired_maximum_frame_latency` is interpreted differently on the OpenGL backend vs Vulkan/DX12 (somewhere inside wgpu)?

RndUsr123:
From their docs, specifically this part, that does indeed seem to be the case:
```
This maps to the backend in the following ways:

    Vulkan: Number of frames in the swapchain is desired_maximum_frame_latency + 1, clamped to the supported range.
    DX12: Calls [IDXGISwapChain2::SetMaximumFrameLatency(desired_maximum_frame_latency)](https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_3/nf-dxgi1_3-idxgiswapchain2-setmaximumframelatency).
    Metal: Sets the maximumDrawableCount of the underlying CAMetalLayer to desired_maximum_frame_latency + 1, clamped to the supported range.
    OpenGL: Ignored

It also has various subtle interactions with various present modes and APIs.

    DX12 + Mailbox: Limits framerate to desired_maximum_frame_latency * Monitor Hz fps.
    Vulkan/Metal + Mailbox: If this is set to 2, limits framerate to 2 * Monitor Hz fps. 3 or higher is unlimited.
```

However, I believe there's a chance this is some vendor-specific behavior. In Nvidia's case there's a significant number of options in their control panel (and a number of bugs and bizarre default behaviors, one of which I documented here: #7096) but it's pretty opaque which settings affect which APIs, what are the general conditions for them to work and what are they even actually supposed to do in some instances.

As I was writing this, I noticed the `Vulkan/OpenGL present method` in NVCPL and sure enough...
`Auto`: produces the previously documented behavior: OpenGL > DX12 = Vulkan.
`Prefer native`: Vulkan is now as fast as OpenGL while DX12 remains a frame behind. OpenGL seems to default to this with `Auto`.
`Prefer layered on DXGI swapchain`: makes OpenGL slower as well. This seems to be the `Auto` setting for Vulkan, causing the discrepancy.

Nvidia says about it: "The Vulkan and OpenGL present mechanism can use an OS/API native method or layer the functionality on top fo a driver created DXGI swapchain. The DXGI layered method is compatible with features such as `VK_KHR_display` and Multiplane overlay. The layered method can add minor memory and performance overhead. In some situations, only one of the methods is possible."

There's also a "Low Latency Mode" that doesn't affect this at all (my understanding is it would only work in exclusive fullscreen mode anyway, potentially only in combination with gsync/vrr). Triple buffering also doesn't seem to do anything in this instance. (DX9/DX11 only? Go figure... )

Auto:
image

Prefer native:
image

Btw, ignore the small mismatch between OpenGL and glow in `Auto` and between the two OpenGL and Vulkan instances in `Prefer native`. It's far too small to be a frame, it's most likely a little mismatch in mouse polling instead.

emilk:
This should make it easier to test:
* https://github.com/emilk/egui/pull/8114

PR #8106, opened 2026-04-14
Adds an optional feature to use the `portable-atomic` crate for atomics

Absolucy:
this adds a new feature (off by default) to `egui`, `epaint`, `eframe`, and `egui_extras`: `portable-atomic`

This feature makes it so that any use of atomics (these crates use `AtomicU64` or `AtomicBool`) will use the implementation from the [`portable-atomic` crate](https://docs.rs/portable-atomic), instead of the native `std::sync::atomic` version, which some platforms may not have.

I implemented this while porting egui to the 3DS - as it seems that it lacks a `AtomicU64` (albeit it has `AtomicUsize` and such), and refactoring the `AtomicU64` stuff to use `usize`s instead would likely be a breaking change, so this seems like a better solution - especially as it's optional, so you'd only enable the feature if you actually need it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1248,6 +1248,7 @@ dependencies = [
  "epaint",
  "log",
  "nohash-hasher",
+ "portable-atomic",
  "profiling",
  "ron",
  "serde",
@@ -1355,6 +1356,7 @@ dependencies = [
  "jiff",
  "log",
  "mime_guess2",
+ "portable-atomic",
  "profiling",
  "resvg",
  "serde",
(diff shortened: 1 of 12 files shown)


kay-lambdadelta:
I think the remaining usage of `AtomicU64` can be removed without breaking changes and `portable-atomic` would be a bit overkill considering that `egui` currently requires `std` which requires `AtomicBool` and `AtomicUsize` (which is the same as `AtomicU32` on the 3ds) and that's all `egui` really and truly needs as far as atomics go.

PR #8108, opened 2026-04-15
Group `glow` config in a `struct`

TapGhoul:
This is a breaking public API change, but is otherwise trivial due to it not changing any actual runtime behaviour.

This renames eframe's NativeOptions `vsync` option to `glow_vsync` to make clear without even looking at docs fully that this is specific to the `glow` backend.

While I think a better option would actually be to change the wgpu creation options to match the vsync option if not specified (either to `AutoVsync` or `AutoNoVsync` depending on setting) this would require this be made an `Option<PresentMode>`, which would be confusing - and the `WgpuConfiguration` should probably take priority over other options here, as there's more than 2 present modes that are relevant. So I think this is a suitable way to go.

* This does not close an issue - this was a trivial amount of code to change, so I might as well just make it a PR on the spot.

### DIFF
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -259,22 +259,6 @@ pub trait App {
     fn raw_input_hook(&mut self, _ctx: &egui::Context, _raw_input: &mut egui::RawInput) {}
 }
 
-/// Selects the level of hardware graphics acceleration.
-#[cfg(not(target_arch = "wasm32"))]
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
-pub enum HardwareAcceleration {
-    /// Require graphics acceleration.
-    Required,
-
-    /// Prefer graphics acceleration, but fall back to software.
-    Preferred,
-
-    /// Do NOT use graphics acceleration.
-    ///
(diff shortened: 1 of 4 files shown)


crates/eframe/src/epi.rs
```diff
@@ -303,7 +303,7 @@ pub struct NativeOptions {
     /// The default is `true`.
     ///
     /// Only affects the `glow` backend.
-    pub vsync: bool,
+    pub glow_vsync: bool,
```
emilk: I suggest we put `vsync`, `hardware_acceleration` and `shader_version` into a new `GlowConfiguration` to mirror `WgpuConfiguration`

TapGhoul: New commit made!

TapGhoul:
@emilk How does this look? I've left a comment on something I'm not entirely sure of. I also added the glow config to the web part and added shader compat to the web side's init - though I'm not sure if this work is redundant (or if it even works as expected, I don't have a proper wasm test env set up right now) so if you think I should revert the changes to the web config, I can do so.

emilk (approved):
Thanks!

Issue #2054, opened 2022-09-19
Rotation for UI elements (widgets to layouts)


targrub:
**Is your feature request related to a problem? Please describe.**
I want to be able to specify that a UI elements should shown in a rotation of 90 degree increments.  To be able to show text  or images upside-down, or 4 players seated around a table (imagine an app on a tablet supporting four card players).

Could be a simple field in a styling data structure
```rust
enum RotationEnum {
    Normal,
    Right90,
    Rotated180,
    Left90,
}
```

There doesn't appear to be a way to display UI elements in any way other than right-side up, so if I have an image, like a playing card (which is currently a `RetainedImage`), that I want rotated, I would have to maintain 4 versions of it.  And if I want text to be rotated as well, there isn't a way to produce it (other than pre-rendering to an image without benefit of any `egui` features, and even then, only to a known rotation).

Current Issues which are more specific, but may be solved by a solution of this issue.
https://github.com/emilk/egui/issues/1646
https://github.com/emilk/egui/issues/1279

**Describe the solution you'd like**

I want to be able to set a style parameter on a UI element to have it treated as rotated some increment of 90 degrees.  This could include entire panels down to individual widgets.

**Describe alternatives you've considered**

I could have four versions of each playing card, or render each card to four images, but I don't see a way to do text or any other layout functions in a rotated way.

**Additional context**

xTibor:
> And if I want text to be rotated as well, there isn't a way to produce it

You can render rotated text, `TextShape` has an [`angle`](https://docs.rs/epaint/0.19.0/epaint/struct.TextShape.html#structfield.angle) field. However speaking from experience it's rather painful to use.

Telzhaak:
@xTibor Would you mind sharing an example?
Since the angle field is only directly available on the TextShape, and TextShape requiring a reference to the current galley to set up, it is indeed painful to use

xTibor:
Example as a widget:
https://github.com/xTibor/egui_extras_xt/blob/838692151ef0e63c793c6589e91bfb643dc4d544/egui_extras_xt/src/ui/rotated_label.rs#L25-L52

Example as a drawing primitive:
https://github.com/xTibor/egui_extras_xt/blob/838692151ef0e63c793c6589e91bfb643dc4d544/egui_extras_xt/src/common.rs#L357-L398

Le-Syl21:
[...]

PR #8113, opened 2026-04-16
Add viewport rotation (0°/90°/180°/270°)

### LINKED ISSUE #4130: Rotation of egui layers
Since #3906 has added TSTransform which allows for scaling and zooming egui layers (awesome!), could we upgrade this transform to also allow for layer rotation?

In my collaborative painting app users should be able to rotate the canvas they are working on. Due to the collaborative nature I'd also like to show ui on the canvas. 

Context::set_transform_layer should accept a Transform that also has a rotation component.

- Write a tiny custom ui framework to render the canvas ui for my app
- The hack using a secound egui context outlined here: #1811 

**Additional context**

I tried to [make a quick proof of concept ](https://github.com/lucasmerlin/egui/tree/transform-rotation)to see how much effort this would be using glam for the transform matrix and I've got some promising results: 

https://github.com/emilk/egui/assets/8009393/c7769025-57de-4f9f-aa56-fe44272f9f19

I guess using glam or some other crate is probably not possible and egui would need to either add Rotation to it's Transform struct or add it's own Matrix implementation?
If this is something that could be added to egui I would be interested in making a PR and working on this.

The biggest challenges seem to be:
- Handling PaintCallback (I don't think those could be rotated?)
- Handling ClipRects

Related to #2054 (individual widget rotation)

Le-Syl21:
## Summary

Adds **viewport rotation** (0° / 90° / 180° / 270°) to egui and eframe, with transparent input/output remapping and a rotated software cursor. Enables use cases where the physical display is mounted rotated — pinball cabinets, kiosks, embedded panels, industrial displays.

Resolves #4130 (rotation via GPU scissor conflict) and #2054 by bypassing the scissor rect limitation entirely: primitives are already in physical screen coordinates by the time they reach the GPU backend.

### 1. `ViewportRotation` in `emath`
New enum `{ None, CW90, CW180, CW270 }` with `transform_pos`, `inverse_transform_pos`, `transform_vec`, `transform_screen_rect`, and `swaps_axes` helpers. Applied with integer rotations so no floating-point drift.

### 2. Rotation applied transparently at the egui layer
- Input rotation in `Context::begin_pass` — mouse positions and deltas are transformed before the UI sees them. App code reads coordinates as if the viewport were un-rotated.
- Output rotation in `Context::tessellate_for_viewport(viewport_id, …)` — mesh vertices and clip rects are inverse-transformed back to physical screen space after layout. Painters see physical pixels.
- Per-viewport state: `ViewportState.viewport_rotation`, settable via `Context::set_viewport_rotation` at runtime, or via `ViewportBuilder::with_rotation` at creation (root only).

### 3. Rotated software cursor
When rotation is active, the OS cursor would move in physical space — disorienting in a rotated UI. This PR:
- Hides the OS cursor on entry into the rotated window, draws a software cursor in logical space.
- Follows raw mouse deltas (rotated appropriately) so direction matches the rotated UI.
- Rotates cursor icons (text caret, resize arrows, etc.) to match the viewport rotation.
- Releases the OS cursor cleanly 3px outside the window when the software cursor reaches an edge.
- Recaptures at the entry point when the cursor re-enters.
- Scalable via `Context::set_software_cursor_scale(f32)` — useful for large pincab playfields.
- Restores the OS cursor to `Default` on release (fixes invisible cursor bug).

### 4. `tessellate_for_viewport` — required correctness fix
A naive implementation of `tessellate` reads rotation from `ctx.viewport()` which resolves to the currently-active viewport via `viewport_stack`. But integrations (wgpu, glow) call `tessellate()` **after** the pass has ended — `viewport_stack` is already popped, and `ctx.viewport()` falls back to `ROOT`. Result: the root viewport's rotation leaks into every other viewport's rendering; secondary viewports' rotations are ignored entirely.

The fix: `tessellate_for_viewport(viewport_id, shapes, pixels_per_point)` takes the target viewport explicitly. `tessellate()` is kept as a thin wrapper calling `tessellate_for_viewport(ViewportId::ROOT, …)` for backwards compat. Both wgpu and glow integrations updated to pass the right `viewport_id`.

### 5. Validation — working examples on all three backends
- **Glow (native)**: `examples/hello_world_90_glow`
- **Wgpu (native)**: `examples/hello_world_90_wgpu`
- **Web (wasm)**: `crates/egui_demo_app_90` + `scripts/build_demo_web_90.sh`

### 6. Module-level docs and changelog
Updated `egui/src/viewport.rs` module docs with a usage example and section on the rotation feature. Changelog entries in `egui`, `eframe`, and `emath`.

## Usage

```rust
// At creation (root viewport)
        .with_inner_size([800.0, 480.0])
        .with_rotation(eframe::emath::ViewportRotation::CW90),
    ..Default::default()
};

// At runtime (any viewport, per-viewport state)
ctx.set_viewport_rotation(eframe::emath::ViewportRotation::CW90);
ctx.set_software_cursor_scale(3.0); // big cursor for far-viewing displays
```

## Test plan

- [x] `cargo check --workspace` clean
- [x] Manual: hello_world_90_glow rotates, cursor behaves correctly in all 4 rotations, edge-release works
- [x] Manual: hello_world_90_wgpu same
- [x] Manual: web demo rotated
- [x] Downstream validation in a real multi-viewport kiosk app (pinball cabinet launcher) — confirms primary rotation works and secondaries are correctly isolated after the `tessellate_for_viewport` fix

---

Related follow-up PR (separate, orthogonal feature): adds `ViewportBuilder::with_monitor(index)` for targeting a specific output at window creation. Useful with rotation for multi-screen kiosk setups but independent of rotation.

🤖 Contributions from [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,18 @@ This file is updated upon each release.
 Changes since the last release can be found at <https://github.com/emilk/egui/compare/latest...HEAD> or by running the `scripts/generate_changelog.py` script.
 
 
+## Unreleased
+
+### Added
+* Viewport rotation support (0°/90°/180°/270°) via `ViewportBuilder::with_rotation()` and `Context::set_viewport_rotation()` ([#4130](https://github.com/emilk/egui/issues/4130), [#2054](https://github.com/emilk/egui/issues/2054))
+  * Entire UI is rendered rotated with all input coordinates automatically remapped
+  * Software cursor with rotated movement direction, capture/release on window edges
+  * Cursor icons adapt to rotation (e.g. `ResizeHorizontal` ↔ `ResizeVertical` at 90°/270°)
+  * Configurable cursor scale via `Context::set_software_cursor_scale()`
+  * Works on all backends: glow, wgpu, and web (via `Context::set_viewport_rotation()`)
+  * `ViewportRotation` enum added to `emath` with coordinate transform methods
+
+
 ## 0.34.1 - 2026-03-27
 Nothing new
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1326,6 +1326,35 @@ dependencies = [
 ]
 
 [[package]]
+name = "egui_demo_app_90"
+version = "0.34.1"
+dependencies = [
+ "accesskit",
+ "accesskit_consumer",
+ "bytemuck",
+ "eframe",
+ "egui",
+ "egui_demo_lib",
+ "egui_extras",
+ "egui_kittest",
+ "ehttp",
+ "env_logger",
+ "image",
+ "jiff",
+ "log",
+ "mimalloc",
+ "poll-promise",
+ "profiling",
+ "puffin",
+ "puffin_http",
+ "rfd",
+ "serde",
+ "wasm-bindgen",
+ "wasm-bindgen-futures",
+ "web-sys",
+]
+
+[[package]]
 name = "egui_demo_lib"
 version = "0.34.1"
 dependencies = [
@@ -2083,6 +2112,24 @@ dependencies = [
 
 [[package]]
 name = "hello_world"
+version = "0.1.0"
+dependencies = [
+ "eframe",
+ "egui_extras",
+ "env_logger",
+]
+
+[[package]]
+name = "hello_world_90_glow"
+version = "0.1.0"
+dependencies = [
+ "eframe",
+ "egui_extras",
+ "env_logger",
+]
+
+[[package]]
+name = "hello_world_90_wgpu"
 version = "0.1.0"
 dependencies = [
  "eframe",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ resolver = "2"
 members = [
   "crates/ecolor",
   "crates/egui_demo_app",
+  "crates/egui_demo_app_90",
   "crates/egui_demo_lib",
   "crates/egui_extras",
   "crates/egui_glow",
--- a/crates/eframe/CHANGELOG.md
+++ b/crates/eframe/CHANGELOG.md
@@ -7,6 +7,9 @@ This file is updated upon each release.
 Changes since the last release can be found at <https://github.com/emilk/egui/compare/latest...HEAD> or by running the `scripts/generate_changelog.py` script.
 
 
+## Unreleased
+* Propagate `ViewportBuilder::rotation` to `Context` at init for glow and wgpu backends
+
 ## 0.34.1 - 2026-03-27
 * `wgpu` backend: Enable WebGL fallback [#8038](https://github.com/emilk/egui/pull/8038) by [@emilk](https://github.com/emilk)
 * Only apply cursor style to the `<canvas>` [#8036](https://github.com/emilk/egui/pull/8036) by [@mkeeter](https://github.com/mkeeter)
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -678,7 +678,9 @@ impl GlowWinitRunning<'_> {
         egui_winit.handle_platform_output(&window, platform_output);
 
         if is_visible {
-            let clipped_primitives = integration.egui_ctx.tessellate(shapes, pixels_per_point);
+            let clipped_primitives = integration
+                .egui_ctx
+                .tessellate_for_viewport(viewport_id, shapes, pixels_per_point);
 
             {
                 // We may need to switch contexts again, because of immediate viewports:
@@ -1533,7 +1535,8 @@ fn render_immediate_viewport(
 
     // ---------------------------------------------------
 
-    let clipped_primitives = egui_ctx.tessellate(shapes, pixels_per_point);
+    let clipped_primitives =
+        egui_ctx.tessellate_for_viewport(viewport_id, shapes, pixels_per_point);
 
     let mut glutin = glutin.borrow_mut();
 
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -277,6 +277,11 @@ impl<'app> GlowWinitApp<'app> {
             }
         }
 
+        // Apply viewport rotation from builder
+        if let Some(rotation) = self.native_options.viewport.rotation {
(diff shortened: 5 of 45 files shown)


lucasmerlin:
Handling screen rotation at the application level seems like an anti-pattern 🤔
Usually there are os settings, allowing you to rotate the whole screen, does that not work for you? 

Also it wouldn't fix https://github.com/emilk/egui/issues/4130, since that one is about fine-grained rotation of individual widgets

Le-Syl21:
Hi @lucasmerlin, thanks for taking the time to look at this!

I want to push back on the "anti-pattern" framing, because the use case here is precisely the one where OS-level rotation falls apart.

### The driving use case: virtual pinball cabinets ("vpin" / "pincab")

A vpin cabinet has **multiple monitors sharing one OS session**: the playfield (physically rotated 90°), a backglass (landscape), often a DMD strip, sometimes a topper. I'm building [PinReady](https://github.com/Le-Syl21/PinReady), a configurator/launcher for vpin cabinets, and it has to render a rotated UI on the playfield while keeping the backglass and DMD un-rotated. **Rotating the OS desktop rotates everything** — taskbar, login screen, the backglass app, the DMD app — which is exactly what we cannot do. The rotation has to be per-window, at the application level. Same story for kiosks, industrial panels, embedded retail signage, museum installations, etc.

### "Use the OS settings" — actual cross-platform support

| OS | App-controllable rotation? | Notes |
|---|---|---|
| **Windows** | Yes, but global | `SetDisplayConfig` + `DISPLAYCONFIG_ROTATION` rotates the **entire desktop** — taskbar, login screen, every other app on that display. No per-window option. |
| **macOS** | **No public API** | `CGDisplayRotation` reads rotation; setting it requires private SPI (`IOAVServiceSetProperty` / `IOMobileFramebuffer`). Tools like `displayplacer` and `fb-rotate` use undocumented mechanisms and regularly break across macOS releases (e.g. M1/Ventura). Apple does not offer a blessed path. |
| **Wayland** | **No cross-compositor protocol** | `wlr-output-management-unstable-v1` is wlroots-only (Sway, Hyprland). `kde-output-management-v2` is KWin-only. Mutter/GNOME doesn't expose rotation via Wayland at all (only via gsettings / D-Bus). KMS/DRM `plane.rotation` works but needs DRM-master / seat access — kiosk/embedded only. |
| **X11** | Yes (`xrandr`), but global | Same problem as Windows: rotates the whole output. |
| **Web (wasm)** | No | There is no "rotate the OS" for a web app rendered in a canvas. |

### Existing Rust crates offering app-level viewport rotation

I checked. There aren't any:

- **winit**: no rotation API on desktop (issue [#2308](https://github.com/rust-windowing/winit/issues/2308) and others — orientation events exist on Android/iOS, nothing on desktop)
- **SDL2 / SDL3 (Rust bindings)**: `SDL_DisplayOrientation` is read-only
- **wgpu / glow / skia-safe**: none
- **egui ecosystem**: nothing (the open issues #2054 and #4130 confirm this)

So an egui app today that needs a rotated playfield has no upstream option — every vpin/kiosk dev rolls their own incomplete workaround.

### On #4130

Fair point — re-reading it, the title is "Rotation of egui layers" and the discussion mixes layer-level rotation with widget-level rotation, while my PR is viewport-level. I'll drop the "fixes #4130" claim from the description. What this PR does do is unblock the underlying technical concern raised there (rotated scissor rect in egui_wgpu/egui_glow) by transforming primitives to physical screen space *before* they reach the GPU backend, so the existing scissor logic stays untouched. Fine-grained per-widget rotation is a separate problem.

#2054 is closer in spirit (it asks for Normal/Right90/Rotated180/Left90 applicable to panels/widgets) but again, viewport-level is a strict subset and a useful first step.

Happy to split, rename, or rescope if that helps the review — but I'd really like to keep the door open for this feature, because the alternative for vpin/kiosk authors is "fork egui."

lucasmerlin:
Sorry, I still don't get why it would be useful for a single window to be rotated. It sounds like you have multiple screens. The os should be able to set rotation per-screen. 
I'll close this as it adds too much complexity for us to maintain (and seems heavily vibecoded). This is a super niche usecase where forking egui makes perfect sense, if this really is something you need.

Le-Syl21:
Update for anyone watching: I've moved this work to a standalone companion crate so it can ship without modifying egui itself.

→ **[`egui-rotate` on crates.io](https://crates.io/crates/egui-rotate)** · [GitHub](https://github.com/Le-Syl21/egui-rotate) · [docs.rs](https://docs.rs/egui-rotate)

```toml
[dependencies]
egui-rotate = "0.1"
# add `features = ["software-cursor"]` for the rotated software cursor
```

It's a lite-by-default helper that hooks into the integration loop:

```rust
use egui_rotate::{Rotation, transform_raw_input, transform_clipped_primitives};

transform_raw_input(&mut raw_input, Rotation::CW90);
let full_output = ctx.run_ui(raw_input, |ui| { /* app */ });
let mut prims = ctx.tessellate(full_output.shapes, ppp);
transform_clipped_primitives(&mut prims, Rotation::CW90, logical_size);
painter.paint_primitives(physical_dimensions, ppp, &prims);
```

What's in the crate:

- `Rotation` enum (None / CW90 / CW180 / CW270) with pure-integer pixel-perfect math
- `transform_raw_input` / `transform_clipped_primitives` — the two hooks above
- `CursorIconExt::rotate` — directional cursor remapping (Text → VerticalText, ResizeEast → ResizeSouth at CW90, etc.)
- `SoftwareCursor` (opt-in feature) — virtual cursor drawn in logical space, with capture/release at window edges, configurable scale, and a lock mode for kiosk/cabinet use, since the OS cursor can't be rotated
- A complete winit + glow + egui_glow demo (`cargo run --example rotated_demo --features software-cursor`)
- 26 tests, CI on Linux/macOS/Windows, MSRV 1.92 (matches egui 0.34)

Caveats vs. an upstream integration:
- Doesn't work with `eframe`'s auto-paint pipeline (eframe owns tessellate→paint with no hook between them). Custom integrations only — winit+glow, winit+wgpu, SDL3, etc.
- Doesn't address #4130 (per-widget rotation is a different problem)
- Still 90°-multiples only; arbitrary angles would need a different design (lossy resampling + non-AABB clip rects)

Closing the loop here. Thanks for the time you spent reviewing.

PR #8114, opened 2026-04-17
wgpu: Allow configuring VSync and frame latency at runtime

emilk:
Let apps change present_mode and desired_maximum_frame_latency at runtime instead of only at startup.                                                      
                                                                                                                                                           
API changes (egui-wgpu):                                                                                                                                   
- New SurfaceConfig { present_mode, desired_maximum_frame_latency }.                                                                                       
- WgpuConfiguration now nests these as pub surface: SurfaceConfig (was two top-level fields).                                                              
- RenderState gains pub surface_config: SurfaceConfig — the currently-requested value.       
                                                                                                                                                           
API additions (eframe):                                                                                                                                    
- Frame::wgpu_surface_config() / Frame::set_wgpu_surface_config(...) for get/set.                                                                          
- SurfaceConfig re-exported as eframe::SurfaceConfig.                                                                                                      
                                                       
How it works:                                                                                                                                              
The wgpu painter compares render_state.surface_config to its currently-applied values each paint. If they differ it updates its config and flips
needs_reconfigure on every surface, piggybacking on the existing deferred-reconfigure pathway.                                                             
                                                       
Demo:                                                                                                                                                      
The backend panel (egui_demo_app) gets dropdowns for present mode and desired max frame latency, wired through the new Frame accessors.

<img width="282" height="172" alt="image" src="https://github.com/user-attachments/assets/0b1274b2-7e4e-4413-969b-0a014c415f79" />

### DIFF
--- a/crates/egui-wgpu/src/lib.rs
+++ b/crates/egui-wgpu/src/lib.rs
@@ -78,21 +78,27 @@ pub struct SurfaceConfig {
     ///
     /// See [`wgpu::SurfaceConfiguration::desired_maximum_frame_latency`] for details.
     ///
-    /// `None` = `wgpu` default.
+    /// `None` => Let `wgpu` pick a default (currently `2`).
     pub desired_maximum_frame_latency: Option<u32>,
 }
 
-impl Default for SurfaceConfig {
-    fn default() -> Self {
-        Self {
-            present_mode: wgpu::PresentMode::AutoVsync,
-            desired_maximum_frame_latency: if cfg!(target_os = "ios") {
-                None // The default is good on iOS, while `Some(1)` cuts FPS in half
-            } else {
-                Some(1) // Low-latency by default.
-            },
-        }
-    }
+impl SurfaceConfig {
+    /// Good default for GUIs with very little (or no) extra GPU work.
+    pub const LOW_LATENCY: Self = Self {
+        present_mode: wgpu::PresentMode::AutoVsync,
+        desired_maximum_frame_latency: if cfg!(target_os = "ios") {
+            None // The default is good on iOS, while `Some(1)` cuts FPS in half
+        } else {
+            Some(1) // Low-latency by default.
+        },
+    };
+
+    /// Good default for GUIs with a lot of extra GPU work,
+    /// or that want to prioritize smoothness over latency.
+    pub const HIGH_THROUGHPUT: Self = Self {
+        present_mode: wgpu::PresentMode::AutoVsync,
+        desired_maximum_frame_latency: Some(2), // High-throughput.
+    };
 }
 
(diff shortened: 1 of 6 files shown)


crates/egui-wgpu/src/lib.rs
```diff
@@ -64,6 +64,37 @@ pub enum WgpuError {
...
+    pub present_mode: wgpu::PresentMode,
+
+    /// Desired maximum number of frames that the presentation engine should queue in advance.
+    ///
+    /// Use `1` for low-latency, and `2` for high-throughput.
+    ///
+    /// See [`wgpu::SurfaceConfiguration::desired_maximum_frame_latency`] for details.
+    ///
+    /// `None` = `wgpu` default.
+    pub desired_maximum_frame_latency: Option<u32>,
```
Wumpf, line 82: should point out what we default to I think

crates/egui-wgpu/src/lib.rs
```diff
@@ -64,6 +64,37 @@ pub enum WgpuError {
...
+
+impl Default for SurfaceConfig {
+    fn default() -> Self {
+        Self {
+            present_mode: wgpu::PresentMode::AutoVsync,
+            desired_maximum_frame_latency: if cfg!(target_os = "ios") {
+                None // The default is good on iOS, while `Some(1)` cuts FPS in half
+            } else {
+                Some(1) // Low-latency by default.
+            },
```
Wumpf: As someone coming from more heavy handed rendering I'm still nervous about defaulting to 1. But yes for todays overpowered devices having gpu & cpu run in sequence for minimal latency has merit for UI heavy applications and now this is a proper breaking change ⚖️ 

(ideally we'd integrate platform specific framepacing to accurately know present timings and grab input at the right times, but this is hard on its own and a nightmare and on some platforms; like [Android effectively requires special libraries](https://developer.android.com/games/sdk/frame-pacing) even)

Wumpf (approved):
nice! maybe in the future we can also expose surface formats (including HDR shenanigans) there :)
Having this exposed like this in the demo makes it so much easier to try out stuff also 👍

crates/egui-wgpu/src/lib.rs
emilk: I believe the vast majority of eframe users only do any rendering with epaint, which is very light on the GPU.
So I think `1` makes sense for `eframe`, but for `egui-wgpu` we might want to have a different default, since people who use `egui-wgpu` outside of `eframe` are more likely to do advance wgpu things.

(also, people who do advanced wgpu stuff are also more likely to understand that they need to tweak these things).

Wumpf: nice idea!

PR #8121, opened 2026-04-20
Add arbitrary request headers to `EhttpLoader`

### LINKED ISSUE #4491: Add a User-Agent header to HTTP image loader

I'm trying to make an egui program that shows images loaded from an CDN, but it requires that the request has a user agent as a header.

Either hardcore the user agent header into the code handling HTTP image loading or add a way to allow passing in a custom user agent

Loading the image using reqwest as bytes and use ImageSource::Bytes in Image

**Additional context**

frnsys:
* Closes <https://github.com/emilk/egui/issues/4491>

Lets you create an `EhttpLoader` with arbitrary headers like so:

```rust
cc.egui_ctx.add_bytes_loader(std::sync::Arc::new(
    egui_extras::loaders::http_loader::EhttpLoader::default().with_headers(&[
        ("User-Agent", "foo"),
    ])
));
```

I'm not sure if there are any problems with installing a second
`EhttpLoader` (in addition to the one installed by default when using
`egui_extras::install_image_loaders(&cc.egui_ctx)`. But I wasn't
sure how else to pass in configuration options to
`install_image_loaders`.

### DIFF
--- a/crates/egui_extras/src/loaders/http_loader.rs
+++ b/crates/egui_extras/src/loaders/http_loader.rs
@@ -42,7 +42,7 @@ type Entry = Poll<Result<File, String>>;
 #[derive(Default)]
 pub struct EhttpLoader {
     cache: Arc<Mutex<HashMap<String, Entry>>>,
-    request_template: Option<Box<dyn Fn(&mut ehttp::Request)>>,
+    request_template: Option<Box<dyn Fn(ehttp::Request) -> ehttp::Request + Send + Sync>>,
 }
 
 impl EhttpLoader {
@@ -50,7 +50,9 @@ impl EhttpLoader {
 
     /// Provide a request template to modify requests before they're sent,
     /// e.g. to add headers.
-    pub fn with_request_template<F: Fn(&mut ehttp::Request) + 'static>(
+    pub fn with_request_template<
+        F: Fn(ehttp::Request) -> ehttp::Request + Send + Sync + 'static,
+    >(
         mut self,
         request_template: F,
     ) -> Self {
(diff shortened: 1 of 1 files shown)


lucasmerlin:
If someone wanted to use this for authentication, they'd have to ensure they only ever load images from trusted endpoints, so their token is not exposed to any third partys. 
Maybe instead we could have a request_template closure arg? Then one could do something like this:
```rust
cc.egui_ctx.add_bytes_loader(std::sync::Arc::new(
    egui_extras::loaders::http_loader::EhttpLoader::default().with_request_template(|url| {
		let mut request = ehttp::Request::get(url);
		if url.starts_with("https://my-api.com") {
		    request = request.with_headers(&[
        		("User-Agent", "foo"),
    		])
		}
		request
	})
));
```

emilk (approved):

frnsys:
> If someone wanted to use this for authentication, they'd have to ensure they only ever load images from trusted endpoints, so their token is not exposed to any third partys. Maybe instead we could have a request_template closure arg? Then one could do something like this:
> 
> ```rust
> cc.egui_ctx.add_bytes_loader(std::sync::Arc::new(
>     egui_extras::loaders::http_loader::EhttpLoader::default().with_request_template(|url| {
> 		let mut request = ehttp::Request::get(url);
> 		if url.starts_with("https://my-api.com") {
> 		    request = request.with_headers(&[
>         		("User-Agent", "foo"),
>     		])
> 		}
> 		request
> 	})
> ));
> ```

Yes that's a good point. I updated the PR to do basically what you suggested. But I pass in the prepared `Request` instead of the `url` in case for whatever reason the `Request` ends up being more complicated to construct.

lucasmerlin (approved):
Awesome!

PR #8128, opened 2026-04-22
Add `atoms()` helpers to get `Atoms` from widgets

Nashvill375:
A get atoms trait for Button, checkbox etc. I made it because I had time to kill after I tried sorting buttons in a Vec by the image in their atoms, but couldn't get to it because it was private.


Nashvill375:
I accidentally did something that closed this pull request. Sorry if somebody got duped email.

crates/egui/src/atomics/get_atoms.rs
```diff
@@ -0,0 +1,28 @@
+pub trait GetAtoms<'a> {
```
lucasmerlin: Hmm, why a trait and not just adding the fn to each struct?

crates/egui/src/atomics/get_atoms.rs
```diff
@@ -0,0 +1,28 @@
...
+  pub fn text(self) -> Atoms<'a>;
+}
+
+impl<'a> GetAtoms<'a> for Button<'a> {
+  pub fn text(self) -> Atoms<'a> {
```
lucasmerlin: Why does it take self? Isn't that annoying? Maybe &self returning &Atoms would be nicer

crates/egui/src/atomics/get_atoms.rs
```diff
@@ -0,0 +1,28 @@
+pub trait GetAtoms<'a> {
+  pub fn text(self) -> Atoms<'a>;
```
lucasmerlin: I think it should be called `fn atoms()` instead

crates/egui/src/atomics/get_atoms.rs
Nashvill375: Honestly I don’t understand references enough to see why it matters, which is not to say it doesn’t because I really don’t know, but I’ll fix it either way. Thanks!

crates/egui/src/atomics/get_atoms.rs
Nashvill375: Honestly? IDK

Nashvill375:
Made the last few commits on my phone because I don’t have access to WiFi. I’ll check for issues and add documentation when I can use WiFi.

crates/egui/src/atomics/get_atoms.rs
Nashvill375: 👍

Nashvill375:
Should I be adding #[inline] to the atoms functions?

crates/egui/src/widgets/checkbox.rs
```diff
@@ -34,10 +34,177 @@ impl<'a> Checkbox<'a> {
...
+    /// This only affects the checkbox's appearance. It will still toggle its boolean value when
+    /// clicked.
+    #[inline]
+    pub fn indeterminate(mut self, indeterminate: bool) -> Self {
+        self.indeterminate = indeterminate;
+        self
+    }
+}
+
+impl Widget for Checkbox<'_> {
```
lucasmerlin: Somehow the checkbox widget impl got duplicated?

Nashvill375: where at?

lucasmerlin: If you click expand all you'll find the impl exists twice in the file:

image

Nashvill375: ah i see it now 🤦

Issue #7892, opened 2026-02-07
ubuntu font in epaint_default_fonts not fully free


alexanderkjall:
I was investigating packaging of avis-imgv in Debian, that uses egui, and noticed that the epaint_default_fonts bundle an ubuntu font.

It's easy to un-bundle this, but the ubuntu font family is not considered to be fully free software ( see message 30 here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=603157 ).

It could still be packaged, but it would end up in the non-free section of Debian and thus be a lot less useful for our users.

Would you consider switching to a free font maybe?

emilk:
Yep, I could for sure consider switching. Suggestions welcome!

PR #8129, opened 2026-04-23
Replace default UI font Ubuntu-Light with Inter-Regular

gcailly:
## Summary

Swaps the default proportional typeface in `epaint_default_fonts` from Ubuntu-Light to Inter-Regular, addressing #7892.

## Motivation

- **UI-first design**: Inter is designed for on-screen interfaces, with a tall x-height aiding readability of mixed-case text at small sizes.
- **Licensing cleanup**: drops Ubuntu Font License 1.0 in favor of SIL OFL 1.1, which is already used for Hack and NotoEmoji in this crate and is simpler for downstream redistribution.
- **Community interest**: #7892 established that a switch was acceptable in principle.

## What changed

**Fonts**
- Add `crates/epaint_default_fonts/fonts/Inter-Regular.ttf` (412 KB, static TTF from Inter 4.1 release)
- Add `crates/epaint_default_fonts/fonts/Inter-OFL.txt`
- Remove `Ubuntu-Light.ttf` and `UFL.txt`

**Code**
- Rename constant `UBUNTU_LIGHT` to `INTER_REGULAR` in `epaint_default_fonts/src/lib.rs`
- Update fallback chains in `epaint/src/text/fonts.rs` (proportional + monospace fallback)
- Remove Ubuntu-specific PUA codepoints (U+E0FF, U+EFFD, U+F0FF, U+F200) from `ignore_character` in `epaint/src/text/font.rs`
- Remove 46 Ubuntu PUA match arms from `egui_demo_lib/src/demo/font_book.rs`

**Licensing and packaging**
- `epaint_default_fonts/Cargo.toml`: SPDX expression now `(MIT OR Apache-2.0) AND OFL-1.1`
- `deny.toml`: drop `Ubuntu-font-1.0` from the allow list

**Tests and assets**
- Regenerate 87 snapshot PNGs via `UPDATE_SNAPSHOTS=true`
- `egui_kittest/tests/tests.rs::test_scroll_down`: replace the fixed 2-scroll sequence with a bounded loop (`for _ in 0..10 { button.scroll_down(); }`) so the test tolerates small line-height changes
- `egui_kittest/tests/tests.rs::test_masking`: render the non-deterministic timestamp with the monospace family. Inter-Regular has proportional figures by default, so the total label width would otherwise depend on which digits the current timestamp contains and oscillate by 1 px between runs. Ubuntu-Light's default figures were effectively tabular, which hid the issue. Using the monospace family keeps the test's original intent (masking a value that changes every run) while making the layout deterministic.

**Docs and demo pages**
- `README.md`, `web_demo/index.html`, `web_demo/multiple_apps.html`: text updates referencing the new font

## Scope note

Default sizes in `egui/src/style.rs` (`default_text_styles`) are intentionally unchanged. Inter renders slightly smaller than Ubuntu-Light at the same em size, but a size adjustment is left to a separate PR to keep this one focused on the font swap.

## Monospace fallback

The monospace fallback chain remains `Hack -> Inter -> NotoEmoji -> emoji-icon`, mirroring the previous `Hack -> Ubuntu-Light -> ...` structure. Inter covers roughly the same extended Latin range as Ubuntu-Light, so the fallback keeps its role of filling glyphs missing from Hack.

## Test plan

- [x] `cargo build --all-features`
- [x] `cargo clippy --all-features`
- [x] `cargo fmt --all`
- [x] `cargo test --workspace --all-features` (with regenerated snapshots)
- [x] Visual check in `egui_demo_app`

Made with assistance from Claude Code.


emilk:
Thanks for working on this!

I would like to think a bit about what font we switch to.

Technical: @lucasmerlin brought up the excellent point that if we use a variable font then we could support bold, italics and other variations with a single otf file, which would be cool

Aesthetics: Inter is very beautiful, and is what we're using at Rerun, but I feel it might be a bit "too professional looking" for egui. Or maybe I'm just too used to Ubuntu. I'll see what alternatives I can find :)

gcailly:
I think Inter conveys elegant professionalism. And egui & apps made with egui deserve this. In my humble opinion, by default egui should even have slightly increased margins, button padding, and corner rounding... Here are some screenshots of non professional apps I made with egui + Inter :

image

image

I'm not a good UI designer, but Inter fits well.

PR #8130, opened 2026-04-23
Announce pressed state of selectable buttons to screen readers

antoinecellerier:
A `Button` used as a toggle (via `Button::selected(true)`, `Button::selectable(...)`, or `Ui::selectable_label`) now announces its pressed / not-pressed state to screen readers. Plain buttons that never call `.selected(...)` stay un-toggled, so their announcement doesn't change. `Checkbox` and the pre-existing selectable-label code path already did this; `Button` was the odd one out.

No public API change: the field is private, `Button::selected(bool)`  keeps its signature, and visuals are identical. Internally the field becomes `Option<bool>` so we can distinguish "plain button" from "toggle button currently off".

Regression test added in `regression_tests.rs`, do let me know if some other file would be a better location.

### Note for manual testing

 `egui_demo_app` pulls in `eframe` with `default-features = false` and doesn't re-enable `accesskit`, so `cargo run -p egui_demo_app`  publishes no AccessKit tree at all. To verify manually:
`cargo run -p egui_demo_app --features accessibility_inspector`
Happy to send a small follow-up PR enabling `accesskit` in the demo app's defaults if that's desirable, since that makes a11y work much easier to smoke-test locally.

### Use of AI

This PR was drafted with Claude Code. I understand different projects have different policies regarding AI generated code. Do let me know if this is not acceptable here. Also happy to take any other feedback.

### DIFF
--- a/crates/egui/src/widgets/button.rs
+++ b/crates/egui/src/widgets/button.rs
@@ -35,7 +35,7 @@ pub struct Button<'a> {
     frame_when_inactive: bool,
     min_size: Vec2,
     corner_radius: Option<CornerRadius>,
-    selected: bool,
+    selected: Option<bool>,
     image_tint_follows_text_color: bool,
     limit_image_size: bool,
     classes: Classes,
@@ -54,7 +54,7 @@ impl<'a> Button<'a> {
             frame_when_inactive: true,
             min_size: Vec2::ZERO,
             corner_radius: None,
-            selected: false,
+            selected: None,
             image_tint_follows_text_color: false,
             limit_image_size: false,
             classes: Classes::default(),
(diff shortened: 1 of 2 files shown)


lucasmerlin (approved):
Thanks!

antoinecellerier:
Thanks for the review @lucasmerlin . Did you have feedback on default enablement of accesskit in the demo app so accessibility is easy to review / iterate on? Happy to send a PR to get that addressed as well if it's ok for you.

lucasmerlin:
> Thanks for the review @lucasmerlin . Did you have feedback on default enablement of accesskit in the demo app so accessibility is easy to review / iterate on? Happy to send a PR to get that addressed as well if it's ok for you.

Yeah, its a good idea. Feel free to push to this one or open a new one

antoinecellerier:
> > Thanks for the review @lucasmerlin . Did you have feedback on default enablement of accesskit in the demo app so accessibility is easy to review / iterate on? Happy to send a PR to get that addressed as well if it's ok for you.
> 
> Yeah, its a good idea. Feel free to push to this one or open a new one

Digging a bit into it it looks like that's not needed anymore. 86fcffb2292845dc71116e6c070ab32e5c6e862a implicitly pulled in accesskit through the inspection feature. We could make the feature enablement explicit but I'm not sure that's worth a change.

PR #8132, opened 2026-04-23
Improve `UiBuilder` docs

JosieElliston:
Previously, the doc for `Ui::scope_builder` read

> Create a child, add content to it, and then allocate only what was used in the parent `Ui`.

which I understood as meaning that "only what was used in the parent `UI`" would be allocated (in the child or parent), which makes no sense (in either case).

I rewrote it and some related docs.


emilk (approved):
Thanks!

PR #8135, opened 2026-04-27
Add kittest plugin api

lucasmerlin:
Allows extending kittest to have code run on every frame. 
Useful to e.g. add assertions that should run on _any_ tests. 

I'm building a kittest inspector that will use this plugin api to allow debugging kittest tests frame-by-frame.


lucasmerlin:
We decided to not have a kittest plugin for now and just rely on the egui plugin system for inspection

PR #8119, opened 2026-04-20
Add kitdiff inspector

lucasmerlin:
- related to the gif recording prototype: https://github.com/emilk/egui/pull/8118

Allows you to pause, inspect and control a kittest test:

https://github.com/user-attachments/assets/c62438a0-9989-43a9-badb-f3977661125d

PR #8136, opened 2026-04-27
Plugin based kittest inspector

lucasmerlin:
Alternative implementation of the inspector from #8119, based on the plugin api from #8135


lucasmerlin:
We'll solve inspection via an egui plugin for now: https://github.com/emilk/egui/pull/8234

PR #8138, opened 2026-04-30
eframe: Add `App::transform_primitives` and `App::post_platform_output` hooks

Le-Syl21:
## Summary

Adds two optional methods on the `App` trait that give integrations a place to inspect or transform the per-frame egui pipeline output before it reaches the platform.

```rust
fn transform_primitives(
    &mut self,
    ctx: &egui::Context,
    primitives: &mut Vec<egui::epaint::ClippedPrimitive>,
) {}

fn post_platform_output(
    &mut self,
    ctx: &egui::Context,
    platform_output: &mut egui::PlatformOutput,
) {}
```

Both default to no-op, so existing apps are unaffected.

## Where they fire

| Hook | Fires after | Fires before |
|---|---|---|
| `transform_primitives` | `Context::tessellate` | painter `paint_and_update_textures` |
| `post_platform_output` | `Context::run_ui` returns | `egui_winit.handle_platform_output` consumes it |

Wired into both `glow_integration` and `wgpu_integration`. Web (wasm32) is not touched in this PR — could be a follow-up if there's interest.

## Use cases

These are general-purpose hooks, not tied to any specific feature. Examples that motivated them:

- **Viewport-level transforms**: rotation (e.g. for a 90°-mounted kiosk display), mirroring, custom projections — applied uniformly to the whole frame without touching individual widgets. The shipping companion crate [`egui-rotate`](https://crates.io/crates/egui-rotate) uses these hooks to integrate cleanly with eframe; without them it can only target custom winit/glow integrations.
- **Debugging / instrumentation**: dump primitive counts, log overdraw, inject debug overlays after the UI has been laid out.
- **Custom render passes**: anything that needs access to the final primitive list (e.g. injecting a post-processing pass on a subset of layers).
- **Software cursor capture**: read `PlatformOutput::cursor_icon` to drive a custom cursor (where the OS cursor is hidden, e.g. cabinet displays where the OS cursor cannot be rotated).
- **Output interception**: clipboard / IME / cursor-icon overrides for kiosk and embedded scenarios.

## Why not a single hook?

The two hooks fire at different points in the pipeline (tessellate→paint vs. run_ui→handle_platform_output) and target different data (primitives vs. platform output). Coalescing them would either add a hook that fires earlier than needed for one of the use cases, or split the work into one hook with two unrelated parameters.

## Why not extend `Context` instead?

These are integration-time concerns (post-tessellation, pre-paint; post-run, pre-OS-dispatch). They don't fit the egui `Context` API, which is mid-pass. The `App` trait is the natural place: it's already where `raw_input_hook` lives for the symmetric input side.

## Test plan

- [x] `cargo build -p eframe` clean
- [x] `cargo clippy -p eframe --all-features -- -D warnings` clean
- [x] `cargo fmt -p eframe --check` clean
- [x] `cargo test -p eframe --lib` passes
- [ ] Manual: drive a vendored `egui-rotate` integration through these hooks on a real eframe app, verify rotation + cursor work — done locally on my pinball cabinet launcher (`PinReady`), can share details if useful

## Background

This is a smaller, more general follow-up to [#8113](https://github.com/emilk/egui/pull/8113), which tried to integrate viewport rotation directly into egui+eframe and was declined as too niche / too much surface to maintain. That feedback was fair. The follow-up published the rotation logic as a standalone crate ([`egui-rotate`](https://crates.io/crates/egui-rotate)) which works fine with custom integrations (winit + glow / wgpu / SDL3 directly). However it cannot integrate with `eframe` because eframe owns the tessellate→paint pipeline with no hook in between.

These two hooks fix that — and stay generic. They're not specific to rotation; the doc-comments describe the broader use cases.

If preferred, I'm happy to:
- rename either method (`pre_paint` / `pre_handle_output`?)
- gate them behind a feature flag
- only add `transform_primitives` and drop `post_platform_output` (the cursor icon use case can be worked around with a fixed icon)
- wire them into the wasm integration too

🤖 Drafted with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/eframe/CHANGELOG.md
+++ b/crates/eframe/CHANGELOG.md
@@ -7,6 +7,12 @@ This file is updated upon each release.
 Changes since the last release can be found at <https://github.com/emilk/egui/compare/latest...HEAD> or by running the `scripts/generate_changelog.py` script.
 
 
+## Unreleased
+
+### ⭐ Added
+* Add `App::transform_primitives` and `App::post_platform_output` hooks to inspect/transform tessellated primitives and platform output before they are dispatched.
+
+
 ## 0.34.1 - 2026-03-27
 * `wgpu` backend: Enable WebGL fallback [#8038](https://github.com/emilk/egui/pull/8038) by [@emilk](https://github.com/emilk)
 * Only apply cursor style to the `<canvas>` [#8036](https://github.com/emilk/egui/pull/8036) by [@mkeeter](https://github.com/mkeeter)
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -257,6 +257,69 @@ pub trait App {
     ///
     /// This function does not return a value. Any changes to the input should be made directly to `_raw_input`.
     fn raw_input_hook(&mut self, _ctx: &egui::Context, _raw_input: &mut egui::RawInput) {}
+
+    /// A hook for inspecting or transforming tessellated primitives before they are
+    /// handed to the GPU painter, on every frame.
+    ///
+    /// This runs **after** [`egui::Context::tessellate`] and **before** the painter
+    /// (glow / wgpu) submits draw calls. The primitives are in the same coordinate
+    /// space the painter expects (logical pixels). You can rewrite vertex positions,
+    /// modify clip rects, replace meshes, or strip primitives entirely.
+    ///
+    /// Fires once per viewport per frame. Use `_viewport_id` to filter on the root
+    /// (`egui::ViewportId::ROOT`) or on a specific child viewport — `_ctx.viewport_id()`
+    /// is unreliable here because the viewport stack is already popped by the time
+    /// the hook fires.
+    ///
+    /// Use cases:
+    /// - viewport-level transforms applied uniformly to the whole frame (rotation,
+    ///   mirroring, custom projections) without touching individual widgets
+    /// - debugging / instrumentation: dump primitive counts, log overdraw, inject
+    ///   debug overlays
+    /// - custom render passes that need access to the final primitive list
+    ///
+    /// # Arguments
+    ///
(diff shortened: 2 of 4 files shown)


crates/eframe/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ This file is updated upon each release.
 Changes since the last release can be found at <https://github.com/emilk/egui/compare/latest...HEAD> or by running the `scripts/generate_changelog.py` script.
 
 
+## Unreleased
+
+### ⭐ Added
+* Add `App::transform_primitives` and `App::post_platform_output` hooks to inspect/transform tessellated primitives and platform output before they are dispatched.
+
+
```
emilk, line 15: revert: we generate the changelog on release

emilk (commented):
Did you try using the `egui::Plugin` system to achieve this? In particular: `Plugin::output_hook`

Le-Syl21:
Thanks @emilk — the `Plugin` system was the right answer, so I'm closing this.

Everything these two hooks were for now lives in a plugin: `input_hook` rotates the input, `output_hook` rotates the output, and `on_end_pass` draws the optional software cursor. No eframe hooks needed, and it works the same on glow, wgpu, eframe and the web.

One wrinkle in case it's useful: `output_hook` exposes `FullOutput.shapes` (pre-tessellation), so I rotate the `ClippedShape`s rather than the tessellated primitives. For 90° increments that's clean (clip rects stay axis-aligned), and it surfaced a small quirk — a brushed `RectShape` keeps its texture screen-aligned under `RectShape::angle` (the brush UV is remapped from the already-rotated vertex against the un-rotated rect), so images render upright unless you emit the rotated textured quad yourself.

Shipped as `egui-rotate` 1.0.0 — nothing needed upstream. Thanks for the steer!

PR #8139, opened 2026-04-30
Add feature for supporting raster images in svg

onsdagens:

SVG supports inlining raster images (PNG/JPEG), `egui-extras` as it stands does not support this.

`resvg` has a feature for this, it's just a matter of enabling it, this PR adds an `egui-extras` feature `svg_images` doing exactly that.


lucasmerlin (approved):
thanks!

onsdagens:
should i push an updated lockfile as suggested in the logs?

lucasmerlin:
> should i push an updated lockfile as suggested in the logs?

Yes, that would be great (I think adding the feature must've added some dependency)

PR #8143, opened 2026-05-03
Create a function for making Vec2s via Iterators

ACatCoder-builds:
This would be a quality of life feature, as neither `Vec2` nor `f32` implements Hash, making it difficult to store. This could lead to cleaner code in the codebase of the user using the egui crate.


emilk:
I don't understand what this would be used for?

Issue #2921, opened 2023-04-18
Various bugs related to window resizing


fezjo:
1. Windows with fixed max size shift on bounds even though they are not resizing.
![resize1](https://user-images.githubusercontent.com/26350212/232851229-3812178c-2cd3-4d6c-baf7-0c6669aee9be.gif)

2. Draw Strings widget in Demo cannot be resized on bottom border.
![resize2](https://user-images.githubusercontent.com/26350212/232851238-19914d67-f4d3-4bfc-88c7-4cd468c70f17.gif)

3. Windows can be temporarily resized to lower than minimum size.
![resize3](https://user-images.githubusercontent.com/26350212/232851240-c3a4150d-d354-45f0-8b39-e230b17e41a1.gif)

4. Resizing window snaps to cursor position. Size change should be determined by drag distance instead.
![resize4](https://user-images.githubusercontent.com/26350212/232851248-cfb4ea0b-ad6f-4b0c-b816-3b4062dbaa48.gif)

5. Resizing windows inside panels has several stages which causes erratic behavior.
![resize5](https://user-images.githubusercontent.com/26350212/232851254-1cf69274-57f7-4d32-b184-a5457c9117e7.gif)

6. Window can have visible size and internal desired size. When I press bottom resize handle, the internal horizontal size is changed to current visible size. Side handles should affect only their respective dimensions.
![resize6](https://user-images.githubusercontent.com/26350212/232851257-3ac37f33-b304-4fe9-8db0-562c1b733a6b.gif)

7. I have no idea.
![resize7](https://user-images.githubusercontent.com/26350212/232851261-2618e09f-801e-4854-9f6d-91d6e04bd7af.gif)

**Desktop**
 - OS: Arch Linux

PR #8152, opened 2026-05-11
Don't allow resizing past minimum content size

lucasmerlin:
## Related

- part of #7264 (required for Atoms in window titlebar)
- part of https://github.com/emilk/egui/issues/2921

## What

This implements a fix for this weird edge case when resizing windows past it's contents minimum allocated width:

Before:

https://github.com/user-attachments/assets/33c6c7b2-3621-4eba-8122-99a3930ff67b

After:

https://github.com/user-attachments/assets/5dd47d8f-32bb-4463-aa01-3a5c8f39b10e

There is a very slight flicker on the very first frame where we detect the minimum size. We could cover this with a request_discard, but in practise it should be barely noticable.

### DIFF
--- a/crates/egui/src/containers/resize.rs
+++ b/crates/egui/src/containers/resize.rs
@@ -326,7 +326,7 @@ impl Resize {
         // any previous wrapping content from shrinking further.
         // 4.0 is a bit of safety margin to ensure we don't prevent shrinking on rounding errors.
         let overflowed_x = state.last_content_size.x > state.desired_size.x + 4.0;
-        if overflowed_x && state .observed_min_content_width.is_none() {
+        if overflowed_x && state.observed_min_content_width.is_none() {
             state.observed_min_content_width = Some(state.last_content_size.x);
         }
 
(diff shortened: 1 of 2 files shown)


crates/egui/src/containers/resize.rs
```diff
@@ -305,6 +321,15 @@ impl Resize {
 
         state.last_content_size = content_ui.min_size();
 
+        // The content overflowed the rect we provided. This means last_content_size.x is now at the
+        // very minimum size the content will allow to shrink to. We remember this so we can prevent
+        // any previous wrapping content from shrinking further.
+        // 4.0 is a bit of safety margin to ensure we don't prevent shrinking on rounding errors.
+        let overflowed_x = state.last_content_size.x > state.desired_size.x + 4.0;
+        if overflowed_x && state .observed_min_content_width.is_none() {
```
emilk: rustfmt

emilk (approved):
very clean!

PR #8153, opened 2026-05-12
Theme plugin system (experimental)

AdrienZianne:
# What it does

Addition of a new system of theme plugins which allow the user to use different rules engine to compute the style for the available specialised widget style.

# How to use

Create a engine implementing the trait `ThemePlugin` and `ThemeStyle<S: StyleStruct>` and implement the necessary methods, then register this way (example for `ButtonStyle`):
```ui.add_theme::<ButtonStyle>(&mycustomengine);```

Now all button will call the `ThemeStyle<ButtonStyle>` method to compute the correct style and later use the cached value to avoid the costly computation.

If no valid `ThemeStyle<S>` or engine is available then it fallback to the default style.

* Closes part of <https://github.com/emilk/egui/issues/3284>

### CODE REVIEW

File: crates/egui/src/context.rs [hunk 1]
```diff
@@ -1970,6 +1974,46 @@ impl Context {
...
+        state: WidgetState,
+        base: &Style,
+    ) -> Option<S> {
+        self.write(move |ctx| ctx.theme.get::<S>(classes, state, base))
+    }
+
+    /// Change the current theme used by the system to compute the style
+    pub fn switch_theme(&self, theme_id: &impl ToString) {
+        self.write(|ctx| ctx.theme.set_current_theme(theme_id));
+    }
```
lucasmerlin: Do we need a seperate api for this or could this be handled via classes? E.g. a dark_mode class on the root Ui to set the theme to dark mode. (Could even allow changing the theme within a part of the ui, by setting a different theme class there).

lucasmerlin (commented):
I think it's trying to do a bit too much at once for now. We should try to start with a super minimal api. 

The way it's set up with the single ThemePlugin that can be active at a time breaks extensibility. Say someone makes a EguiMaterial theme crate, and someone else makes a EguiSwitch component crate, and you want to use both in your project.

The way it's now, one would have to make a whole new ThemePlugin struct that wraps  the EguiMaterialThemePlugin and adds support for the EguiSwitchStyle (since the rust orphan rules would prevent you from adding something like:
`impl ThemeStyle<EguiSwitchStyle> for MeterialThemePlugin {`.

So I would suggest we simplify the registration api to instead be per widget type. You could still have a single struct that impls the styles of all the different widget kinds, but they should be added indivudually. (Of course any theme plugin could offer a helper fn that registers them all). 
Something like this:
```rs
let theme = EguiMaterialTheme::new();
let switch_theme = EguiSwitchTheme::new();
ctx.add_theme::<ButtonStyle>(theme);
ctx.add_theme::<SwitchStyle>(switch_theme);
```

Both of these would be active at the same time, and there can only ever be a single registered theme per type.

Theme switching could still be implemented either by reregistering the different style types with different configs or by something internal to the theme plugin. 

Caching could be reduced to some CachingThemePlugin that wraps some other plugin, so if someone doesn't want / need caching they could skip it.

crates/egui/src/context.rs
AdrienZianne: > Caching could be reduced to some CachingThemePlugin that wraps some other plugin, so if someone doesn't want / need caching they could skip it.

Why not use a flag in the ThemePlugin implementation ? That way if we don't find it in the cache, we compute the style and store it only if it's wanted.

> Do we need a separate api for this or could this be handled via classes? E.g. a dark_mode class on the root Ui to set the theme to dark mode.

This is an idea, but we need to differentiate "theme" classes from regular classes. I also think that if many people design and share their own themes, their gonna need to share the same keyword like root and selected. Maybe I should create an enum with these keywords and one that take a `Classname` as a value.

Otherwise the change in how the theme are registered already allow to switch the theme.

crates/egui/src/widgets/button.rs
```diff
@@ -316,7 +316,8 @@ impl<'a> Button<'a> {
 
         classes.add_class_if(SELECTED_CLASS, selected);
 
-        let ButtonStyle { frame, text_style } = ui.style().button_style(&classes, state);
+        // let ButtonStyle { frame, text_style } = ui.style().button_style(&classes, state);
```
lucasmerlin: useless comment

crates/egui/src/context.rs
```diff
@@ -2027,6 +2030,42 @@ impl Context {
     }
 }
 
+impl Context {
+    /// Register a [`ThemeStyle`](crate::theme_plugin::ThemeStyle) for the specified widget [`StyleStruct`](StyleStruct) `S`
+    ///
+    /// A theme can only be added once for a specified widget.
```
lucasmerlin, line 2103:
```suggestion
    /// A theme can only be added once for the specified widget type.
```

crates/egui/src/context.rs
```diff
@@ -2027,6 +2030,42 @@ impl Context {
...
+    /// This way it's convenient to add themes in `eframe::run_simple_native`.
```
lucasmerlin:
```suggestion
    /// If a theme is already registered for this widget, this is a no-op (useful for `eframe::run_simple_native`).
```

crates/egui/src/widget_style.rs
```diff
@@ -1,15 +1,27 @@
...
+    Frame, Response, Spacing, Style, TextBuffer as _, TextStyle, Visuals,
     style::{WidgetVisuals, Widgets},
 };
 
+/// Each dedicated style must implement this trait to be used in the theme plugin system
+pub trait StyleStruct: Debug + Clone + Send + Sync + std::any::Any + 'static {
+    /// The default style for this struct based on classes and state of the widget.
+    fn default_style(classes: &Classes, state: WidgetState) -> Self
+    where
+        Self: Sized;
```
lucasmerlin: Instead of having a default_style fallback here, we should just provide `DefaultWidgetStyle` and `DefaultButtonStyle` ThemeStyle implementations and register those by default.

If no themestyle is registered for the requested widget type, we panic.

lucasmerlin (changes requested):

lucasmerlin:
Also we should have some way to read other theme styles from within a `ThemeStyle` impl. So maybe pass &Context or &Themes into the `style` fn?

It should be possible to e.g. read the `WidgetStyle` theme from within the `ButtonStyle` theme engine impl, so that overriding the WidgetStyle theme would affect buttons and other widgets.

crates/egui/src/theme_plugin.rs
```diff
@@ -0,0 +1,298 @@
...
+                .into(),
+                ..Default::default()
+            },
+            text_style: ws.text,
+        }
+    }
+}
+
+impl ThemeStyle<CheckboxStyle> for DefaultStyle {
+    fn style(&mut self, ui: &Ui, classes: &Classes, state: WidgetState) -> CheckboxStyle {
```
lucasmerlin: We might want to extend what we pass in in the future, lets instead add a struct (`StyleArgs` or `StyleModifiers`?) that combines these params.

Also instead of `Ui` we should pass in the `UiStack` and `Context`, I think `Ui` is a bit too general.
Maybe also the widgets `Id`, which should make it possible to add animations.

crates/egui/src/context.rs
```diff
@@ -2027,6 +2030,42 @@ impl Context {
...
+    }
+
+    /// Compute the [`WidgetStyle`] using the registered theme.
+    pub(crate) fn get_widget_style<S: WidgetStyle + Clone + 'static>(
+        &self,
+        ui: &Ui,
+        classes: &Classes,
+        state: WidgetState,
+    ) -> S {
+        self.read(move |ctx| ctx.themes.get::<S>(ui, classes, state))
```
lucasmerlin: This deadlocks when calling most fns on `Ui` or `Context` in the style fn. 
Instead, we shouldn't hold the context read lock, and only lock the individual theme:
```
        let theme = self.read(move |ctx| ctx.themes.theme::<S>());
        theme.lock().style(ui, classes, state)
```

crates/egui/src/theme_plugin.rs
```diff
@@ -0,0 +1,298 @@
...
+        } else {
+            let style = fallback();
+            self.cache.insert_temp(style_id, style.clone());
+            style
+        }
+    }
+}
+
+/// A Theme plugin that implement a style computation for a defined `WidgetStyle`
+pub trait ThemeStyle<S> {
```
lucasmerlin: I think the name is a bit weird, maybe instead it should be `StyleProvider`? Since it produces a `WidgetStyle`/`ButtonStyle`.

crates/egui/src/theme_plugin.rs
```diff
@@ -0,0 +1,298 @@
...
+        let style_id = Id::new(classes).with(state);
+        if let Some(style) = self.cache.get_temp::<S>(style_id) {
+            style
+        } else {
+            let style = fallback();
+            self.cache.insert_temp(style_id, style.clone());
+            style
+        }
+    }
+}
```
lucasmerlin: This cache is kinda specific to the ess example. I was imagining something generic that you could wrap any theme with, something like this:

```rs
#[derive(Debug, Clone)]
pub struct ThemeCache<Theme> {
    cache: IdTypeMap,
    inner: Theme,
}

impl<Theme> ThemeCache<Theme> {
    pub fn new(theme: Theme) -> Self {
        Self {
            cache: Default::default(),
            inner: theme,
        }
    }
}

impl <Theme: ThemeStyle<Style>, Style: WidgetStyle> ThemeStyle<Style> for ThemeCache<Theme> {
    fn style(&mut self, ui: &Ui, classes: &Classes, state: WidgetState) -> Style {
        let style_id = Id::new(classes).with(state);
        if let Some(style) = self.cache.get_temp::<Style>(style_id) {
            style
        } else {
            let style = self.inner.style(ui, classes, state);
            self.cache.insert_temp(style_id, style.clone());
            style
        }
    }
}
```

lucasmerlin: This could be used like this: 
```
        ui.add_widget_theme::<ButtonStyle>(ThemeCache::new(theme.clone()));
```

crates/egui/src/widget_style.rs
```diff
@@ -315,4 +233,9 @@ pub trait HasClasses {
     fn has(&self, class: impl Into<ClassName>) -> bool {
         self.classes().classes.contains(&class.into())
     }
+
+    /// The list of class
+    fn list(&self) -> Vec<ClassName> {
+        self.classes().classes.to_vec()
+    }
```
lucasmerlin: Lets return a slice ref here and call it as_slice or get

lucasmerlin (commented):
Sorry for the delay, I've tested it with the custom theme from my app and it works great! 
A couple more changes and then we should be good.

crates/egui/src/theme_plugin.rs
AdrienZianne: The problem is that I couldn't find any other way to access the current style and the current Theme (Context global style wasn't compatible with menu test since there was change to the style while running).

lucasmerlin: Ah that makes sense. Lets include the Ui's Style as a &Style ref in the arguments struct then

AdrienZianne:
Done ! @lucasmerlin

emilk:
Let's run the benchmarks before/after this PR before we merge

EDIT: ✅  looks like around -3% on `demo_no_tessellate`, which is fine

crates/egui/src/theme_plugin.rs
```diff
@@ -0,0 +1,307 @@
...
+
+impl<Theme: StyleProvider<S>, S: WidgetStyle> StyleProvider<S> for ThemeCache<Theme> {
+    /// Access the cache for the requested [`WidgetStyle`] based on the [`Classes`] and
+    /// the [`WidgetState`]
+    ///
+    /// If no entry match the parameter then compute the fallback style and
+    /// save the output for later.
+    fn style(&mut self, modifiers: &StyleArgs<'_>) -> S {
+        let StyleArgs { classes, state, .. } = modifiers;
+        let style_id = Id::new(classes).with(state);
```
emilk:
```suggestion
        let style_id = Id::new((classes, state));
```

crates/egui/src/theme_plugin.rs
```diff
@@ -0,0 +1,307 @@
...
+            self.cache.insert_temp(style_id, style.clone());
+            style
+        }
+    }
+}
+
+/// A Theme plugin that implement a style computation for a defined `WidgetStyle`
+pub trait StyleProvider<S> {
+    /// The style according to the classes and state of the widget
+    fn style(&mut self, modifiers: &StyleArgs<'_>) -> S;
```
emilk: There is a lot of cloning/copying of styles going on for each widget. `ButtonStyle` for instance is 88 bytes,   so copying it over and over again isn't exactly cheap. I think we should consider wrapping them in `Arc`s, so a big memcpy gets replaced by a cheap atomic increment.

```suggestion
    fn style(&mut self, modifiers: &StyleArgs<'_>) -> Arc<S>;
```

crates/egui/src/theme_plugin.rs
```diff
@@ -0,0 +1,307 @@
...
+    fn theme_type_id(&self) -> TypeId
+    where
+        Self: 'static,
+    {
+        TypeId::of::<Self>()
+    }
+}
+
+#[derive(Debug, Clone)]
+struct DefaultStyle;
```
emilk: let's move this type into its own file

crates/egui/src/theme_plugin.rs
```diff
@@ -0,0 +1,307 @@
...
+        } else {
+            let style = self.inner.style(modifiers);
+            self.cache.insert_temp(style_id, style.clone());
+            style
+        }
+    }
+}
+
+/// A Theme plugin that implement a style computation for a defined `WidgetStyle`
+pub trait StyleProvider<S> {
```
emilk: nit: put the trait on top of the file (declare-before-use)

crates/egui/src/theme_plugin.rs
```diff
@@ -0,0 +1,307 @@
...
+            classes,
+            state,
+            style: self.style(),
+            stack: self.stack(),
+            ctx: self,
+        })
+    }
+}
+
+pub struct Themes {
```
emilk: docstring

examples/styling_engine/src/custom_engine.rs
```diff
@@ -0,0 +1,133 @@
+use std::collections::HashMap;
```
emilk: it's a cool example, but maybe a bit over-complex? 🤷

emilk (approved):
Awesome stuff!

examples/styling_engine/src/custom_engine.rs
AdrienZianne: I agree it's quite "advanced" just to show the capabilities, I can do a simpler "engine" where the user can use color picker to modify a theme and change which one use in a dropdown, I think it will be better. Need to design a better layout too.

emilk: The ESS stuff would be awesome as a separate repo though :]

crates/egui/src/theme_plugin.rs
AdrienZianne: Shouldn't I move it to the new file with the default_style ? So we have the trait, the struct and the implementations

crates/egui/src/theme_plugin.rs
emilk: actually - let's do this in a later PR. It causes some complexities

crates/egui/src/theme/theme_cache.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+impl<Theme: StyleProvider<S>, S: WidgetStyle> StyleProvider<S> for ThemeCache<Theme> {
+    /// Access the cache for the requested [`WidgetStyle`] based on the [`Classes`](crate::widget_style::Classes) and
+    /// the [`WidgetState`](crate::widget_style::WidgetState)
+    ///
+    /// If no entry match the parameter then compute the fallback style and
+    /// save the output for later.
+    fn style(&mut self, modifiers: &StyleArgs<'_>) -> S {
+        let StyleArgs { classes, state, .. } = modifiers;
```
emilk: There is a bug here. We need to hash all the keys that a StyleProvider is allowed to use

emilk (changes requested):
We need to rethink ThemeCache - it now only hashes parts of StyleArgs, so if the other keys change, there is no invalidation. And caching all keys (in particular `Style`) is too slow.

We could provide only the global `Style` (not the one from the parent `Ui`), and then hash that once at the start of each frame.

…but we have a similar problem with UiStack

crates/egui/src/theme/style_provider.rs
```diff
@@ -0,0 +1,17 @@
...
+
+use crate::widget_style::StyleArgs;
+
+/// A Theme plugin that implement a style computation for a defined `WidgetStyle`
+pub trait StyleProvider<S> {
+    /// The style according to the classes and state of the widget
+    fn style(&mut self, modifiers: &StyleArgs<'_>) -> S;
+
+    /// Help to differ the different themes
+    fn theme_type_id(&self) -> TypeId
```
emilk: should probably call this just `type_id`

crates/egui/src/theme/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+
+pub use self::{style_provider::StyleProvider, theme_cache::ThemeCache, themes::Themes};
+
+use crate::{
+    Ui,
+    widget_style::{Classes, StyleArgs, WidgetStyle},
+};
+
+impl Ui {
+    /// Access the register theme and fetch the requested [`WidgetStyle`].
```
emilk: weird wording

crates/egui/src/theme/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+    ///
+    /// Requested widget style must implement [`WidgetStyle`].
```
emilk: unnecessary comment: that is encoded in the type signature

emilk:
Me and Lucas decided we'll land this, but add a `#[cfg(feature = "unstable")]` or `experimental` feature-flag to key choke-points (e.g. registering new theme providers)

lucasmerlin (approved):
I've added a new experimental feature flag, since we'll likely make some bigger breaking changes in the future, as we figure out caching and as we start using this, so users know what to expect.

I think it's good to merge now, thank you for your work on this @AdrienZianne!

emilk (approved):

PR #8154, opened 2026-05-12
Atom support for `egui::Window` Titlebar

lucasmerlin:
* part of #7264 
* based on https://github.com/emilk/egui/pull/8152

The resize fix allows use to really simplify how the Window Titlebar is rendered. Previously it was using some complex flow to calculate and allocate the height first and then render it later once we knew the windows final width. 

Since now windows can't shrink past their minimum content widths, I can just show the titlebar inline with the regular content, just outside of the `Resize` container so that it is always visible. 

This does change what the size of a window means. Before, size was just the size of the contents, while now size (e.g. via min_height) will include the Frames margin and outline, title bar and the contents. 
Also, the window label now truncates as you shrink the window (meaning windows can now be smaller than their label allows).


lucasmerlin:
Clarify what size now means (does it include the frame / margins?).

crates/egui_kittest/tests/regression_tests.rs
```diff
@@ -511,3 +511,77 @@ fn window_resize_wraps_to_content_min_width() {
...
+    let outer_pos = Pos2::new(50.0, 50.0);
+    let outer_size = Vec2::new(300.0, 200.0);
+    let expected_rect = Rect::from_min_size(outer_pos, outer_size);
+
+    let mut harness = Harness::builder()
+        .with_size(Vec2::new(800.0, 600.0))
+        .build_ui(move |ui| {
+            Window::new("size_test")
+                .fixed_pos(outer_pos)
+                .fixed_size(outer_size)
```
emilk, line 534: you should use a custom `.frame(…)` here that sets both the inner and outer margin to test which of the margins are included.

crates/egui/src/containers/window.rs
```diff
@@ -213,6 +213,9 @@ impl<'open> Window<'open> {
     }
 
     /// Set minimum size of the window, equivalent to calling both `min_width` and `min_height`.
+    ///
+    /// The size refers to the *outer* window size, including frame margins, stroke,
```
emilk: outer margins too?

crates/egui_demo_lib/tests/snapshots/demos/Screenshot.png
emilk, line 1: The distance between the title bar and the contents has increased! Is this inentional?

crates/egui_demo_lib/tests/snapshots/demos/Scrolling.png
emilk, line 1: the increased padding at the top of the window makes this look worse

lucasmerlin:
> outer margins too?

Yeah. I could make it exclude outer margins, but it makes the positioning code a bit more complex and I think it's generally weird to have an outer margin on windows? So I decided to keep it like this. (But I've updated the test to check outer margin behavior).

emilk:
The vertical spacing has changed in a lot of the screenshots 🤔  Seems like the title bar has grown or shrunk by a fraction of a pixel. I wonder why 🤔

For instance: check the `Grid Test.png`

Not necessarily worse; just different 🤷

emilk (approved):

lucasmerlin:
Ah the cursor isn't pixel aligned:
```
[crates/egui/src/containers/window.rs:580:33] ui.cursor() = [[672.0 104.7] - [995.8 inf]]
```
I'll try to find the root of this or add some rounding

Edit: Ah I think this is fine actually, at least we do call gui_round on the allocated rect, so I think it's fine?

Issue #2104, opened 2022-10-03
Wacom Stylus Input Driver Support for content creation


omac777:
Content creation industy highly relies on Wacom stylus pens and graphics tablets.
Currently, I am only aware of gtk4 and qt supporting wacom stylus pen and tablets.
How important are these Wacom drivers? Blender, Krita, Gimp, Toonboom, Maya rely on them.
Thank you for listening.

felipetesc:
Hi. I'm also interested about using egui with drawing tablets, but I really think hardware drivers should be provided by the hardware manufacturer. So, I believe you should ask for Wacon, for their header files, which will be probably given as a driver.dll and a c header file, if they decide to give it at all. It would be dangerous for any of us to dwelve into the intrisic of hardware definition. That's, to have an API which abstracts Wacon from Huion is something, and to provide drivers something else. I have a Huion and I can use a pen without any trouble when I open egui.rs and I draw using the canvas sample, no extra installation was needed for egui. What did, I navigated to Huion website, downlod and install the correct driver for my drawing display. 
There is also the issue that this isn't a commercial project with a lot of people, and every display needs a different driver, which will need to be tested, so it's a lot of work. 


But I'm in the same page you are and I will try to use for drawing sketches and so on...

omac777:
Thank you for the prompt reply.
You mentioned your canvas example is working with Huion stylus and tablet.  I think it's too quick to say "working" if all it does is pretend to be a mouse and that you press an area and it draws.  That is not the definition of a stylus and its respective api.
For a more thorough reflection of a stylus api, here's QT's qtabletevent api for inspiration:
https://doc.qt.io/qt-6/qtabletevent.html#details"Tablet events are similar to mouse events; for example, the x(), y(), pos(), globalX(), globalY(), and globalPos() accessors provide the cursor position, and you can see which buttons() are pressed (pressing the stylus tip against the tablet surface is equivalent to a left mouse button). But tablet events also pass through some extra information that the tablet device driver provides; for example, you might want to do subpixel rendering with higher resolution coordinates (globalPosF()), adjust color brightness based on the pressure() of the tool against the tablet surface, use different brushes depending on the type of tool in use (deviceType()), modulate the brush shape in some way according to the X-axis and Y-axis tilt of the tool with respect to the tablet surface (xTilt() and yTilt()), and use a virtual eraser instead of a brush if the user switches to the other end of a double-ended stylus (pointerType())."

[...]

PR #8157, opened 2026-05-12
Support tablet pressure in the browser

epsilon-phase:
Initial work to support reading the pressure of a graphics tablet or touch screen.

This only works in the browser because it does not address lack of support in winit.

It also alters the demo painter application to allow for pressure sensitive drawing. I have taken some effort to preserve some of the behavior.

* Partly addresses #2104

### DIFF
--- a/crates/eframe/src/web/events.rs
+++ b/crates/eframe/src/web/events.rs
@@ -15,6 +15,27 @@ use web_sys::{Document, EventTarget, ShadowRoot};
 
 // ------------------------------------------------------------------------
 
+/// Returns `true` if the browser supports Pointer Events.
+/// If so, we can rely on `pointermove` instead of `mousemove`.
+fn pointer_events_supported() -> bool {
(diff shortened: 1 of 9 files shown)


lucasmerlin:
egui does already have pressure support built in: https://docs.rs/egui/latest/egui/enum.Event.html#variant.Touch.field.force

Unfortunately browser support is limited, since the force value comes from the old touch event api. 

I think the proper fix here would be to mirror the Pointer event api winit 0.31 will have: https://docs.rs/winit/0.31.0-beta.2/winit/event/enum.WindowEvent.html#variant.PointerMoved

This should also map nicely to the browser PointerEvent api. Then we could get rid of the weird duplicated touch vs mouse events.

epsilon-phase:
Do graphics tablets report touch events? I'm getting the impression that it usually doesn't get reported that way, though that could just be because the touch force support provided by the version of winit used by egui is very spotty in terms of platform support.

In either case, thank you for the feedback, I'll try to figure out what that might look like :)

PR #8163, opened 2026-05-15
Improve FileLoader file uri to path handling for windows

### LINKED ISSUE #8161: `Image::from_uri` fails to load images when attempting to load images from a windows network share.
`Image::from_uri` fails to load images when attempting to load images from a windows network share.

**To Reproduce**

In order to reproduce attempt to call `Image::from_uri` with a (valid) windows network share path like `file://<host>/<share>/path/image.jpg`. If the the image exists the Image will not be rendered but instead result in an error "Failed loading <file>. The system cannot find the path specified.

**Additional Context**

Windows network share files are referenced by [UNC](https://en.wikipedia.org/wiki/Path_(computing)#UNC) they look like `\\ComputerName\SharedFolder\Resource`. Encoding UNC paths to URI is found in [RFC 8089](https://www.rfc-editor.org/rfc/rfc8089#appendix-E.3). The suggested encoding is of one of two forms `file://ComputerName/SharedFolder/Resource` or `file::////ComputerName/SharedFolder/Resource`. In both cases it's critical that the prefixed slashes remain when turning it back into UNC to pass off to the host, `ComputerName\SharedFolder\Resource` is not the same path as `\\ComputerName\SharedFolder\Resource`.

When you call `Image::from_uri` it ends up calling [FileLoader::load](https://github.com/emilk/egui/blob/main/crates/egui_extras/src/loaders/file_loader.rs#L46). Load then takes the string uri, removes the scheme prefix, and... if the binary was compiled for windows hosts, it removes all subsequent leading `/`'s. Unfortunately this has the effect of removing the leading two slashes needed to identify the uri as pointing to a network share.

This stripping behavior was added in pull request #4107 in order to fix the case of sending a window file path through the file uri (like `file:///c:/...`) unfortunately, while windows paths with drive letters are not slash prefixed, network shares ARE slash prefixed. So, while this resolves the issue of drive letter paths, it breaks network paths.

Perhaps it's useful to reference the Microsoft documentation on [file path formats](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats)

Reading RFC 8089 I worry that it will ultimately be difficult to resolve this issue without actually parsing the uri, as the solutions proposed in the RFC mostly have to do with re-interpreting certain characteristics of the uri (like auth information).

- OS: Windows 11
- Version: 0.34.2

aconbere:
Current behavior fails when translating file uris that contain windows UNC paths. This commit attempts to fix that behavior by looking at the hostname attribute of the uri and changing behavior if the hostname is present.

* Closes <https://github.com/emilk/egui/issues/8161>

### DIFF
--- a/crates/egui_extras/src/loaders/file_loader.rs
+++ b/crates/egui_extras/src/loaders/file_loader.rs
@@ -216,7 +216,7 @@ mod tests {
             checks.append(&mut more_checks);
         }
         for (uri_s, path, reason) in checks {
-            assert_eq!(convert_uri_to_path(uri_s), path, "{}", reason);
+            assert_eq!(convert_uri_to_path(uri_s), path, "{reason}");
         }
     }
 }
--- a/crates/egui_extras/src/loaders/file_loader.rs
+++ b/crates/egui_extras/src/loaders/file_loader.rs
@@ -3,7 +3,7 @@ use egui::{
     load::{Bytes, BytesLoadResult, BytesLoader, BytesPoll, LoadError},
     mutex::Mutex,
 };
-use std::{sync::Arc, task::Poll, thread, path::PathBuf};
+use std::{path::PathBuf, sync::Arc, task::Poll, thread};
 
 #[derive(Clone)]
 struct File {
@@ -54,7 +54,7 @@ fn convert_uri_to_path(s: &str) -> Result<PathBuf, egui::load::LoadError> {
         // These file uris need to be converted into UNC correct and so need to have the leading
         // two backslashes prepended.
         let path = PathBuf::from(format!("\\\\{s}"));
-        return Ok(path)
+        return Ok(path);
     }
 
     Ok(PathBuf::from(s))
--- a/crates/egui_extras/src/loaders/file_loader.rs
+++ b/crates/egui_extras/src/loaders/file_loader.rs
@@ -171,22 +171,48 @@ mod tests {
 
     #[test]
     fn check_convert_uri_to_path() {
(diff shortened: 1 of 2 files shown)


crates/egui_extras/src/loaders/file_loader.rs
```diff
@@ -66,7 +85,6 @@ impl BytesLoader for FileLoader {
             // We need to load the file at `path`.
 
             // Set the file to `pending` until we finish loading it.
-            let path = path.to_owned();
```
aconbere, line 69: the pathbuf returned from `convert_uri_to_path` is already owned.

aconbere:
I can confirm that with my local app this change fixes the behavior allowing both local paths and remote paths to work on windows. One of the test demo app tests are failing, and I'm not sure how to work with those yet.

emilk (approved):

aconbere:
@emilk resolved the clippy and fmt issues

emilk:
Why does the title day `[DRAFT]` btw?

aconbere:
> Why does the title day `[DRAFT]` btw?

I feel like... I read something about marking the PR as draft in the template? I think the answer is probably as simple as "I did it wrong". I've removed it :)

emilk (approved):
Some tests would be nice, but code LGTM!

thanks

aconbere:
I'm happy to add some, what's your preferred method for testing in cases like this? It would make sense to me for there to be some simple unit tests around the url string to path conversion.

aconbere:
@emilk hopefully last go around, added unit tests

Issue #4538, opened 2024-05-24
Wayland dependencies are pulled in for eframe, even when `wayland` feature is disabled.


crumblingstatue:
**Describe the bug**
When depending on eframe with default features disabled, and only `x11` and `glow` enabled, eframe is pulling in wayland dependencies.

**To Reproduce**
Create a new project, and depend on `eframe` with default features disabled.
```toml
[package]
name = "myeframetest"
version = "0.1.0"
edition = "2021"

[dependencies]
eframe = { git = "https://github.com/emilk/egui.git, default-features = false, features = ["glow", "x11"] }
```

**Expected behavior**

No wayland dependencies in `Cargo.lock`

**What happens**
`Cargo.lock` has several wayland dependencies.

PR #8165, opened 2026-05-18
Enable `arboard?/wayland-data-control` for `wayland` feature

grtlr:
> [!WARNING]
> Currently blocked by waiting for `winit v0.31`.

Fixes this warning (and clipboard support) on Wayland:

```
[2026-05-18T09:49:48Z WARN  egui_winit::clipboard] Failed to initialize arboard clipboard: Unknown error while interacting with the clipboard:
  X11 server connection timed out because it was unreachable
```

From the `arboard` crate description: 

> However, Wayland is becoming the majority default as of 2025. Some distributions are even considering the removal of X by default. To support Wayland correctly, arboard users should enable the wayland-data-control feature. If enabled, it will be prioritized over the X clipboard.
> 
> Wayland support is not enabled by default because it may be counterintuitive to some users: it relies on the data-control protocol extension(s), which are not supported by all Wayland compositors. You can check compositor support on `wayland.app` [...]

## Related

* #4538
* #5478

The problem is that `cargo` won't let us specify "if both `wayland` and `clipboard` are enabled, also enable `smithay`" without more shenanigans.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -288,6 +288,7 @@ dependencies = [
  "parking_lot",
  "percent-encoding",
  "windows-sys 0.60.2",
+ "wl-clipboard-rs",
  "x11rb",
 ]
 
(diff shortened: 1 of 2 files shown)


emilk:
converted to draft until unblocked by winit 0.31 (any year now…)

Issue #2550, opened 2023-01-06
Add double-click and drag and triple-click and drag to select multiple words and to select multiple paragraphs


Poopooracoocoo:
**Is your feature request related to a problem? Please describe.**

It can be difficult to select large chunks of text at a time

**Describe the solution you'd like**

Double-click and drag and triple-click and drag are UX paradigms found on other platforms.

**Describe alternatives you've considered**

N/A

**Additional context**

Tested on TextEdit demo

emilk:
Sounds nice - you should be able to add it to `crates/egui/src/widgets/text_edit/builder.rs`

darkwater:
Now that labels are also selectable, it's not (just) `text_edit/builder.rs` anymore, right? Unless Label uses that for text selection somehow?

mudbungie:
I'm taking a stab at this. The reason it has never worked: egui only counts clicks on pointer *release*, so when a drag starts there is no way to know it began with a double-click. My approach: also track the would-be click count at press time, select the word/line already on the second/third press, and keep it as an anchor so dragging extends the selection by whole words/lines — in both `TextEdit` and labels. Draft PR incoming.

PR #8166, opened 2026-05-18
Allow doubleclick-and-drag and tripleclick-and-drag

hallyhaa:
Fix for the current behaviour where a double-click selects a word and a triple-click selects a line, but *dragging* with multi-click looses the initial selection and then extends the selection character by character.

The fix works for both `TextEdit` and selectable `Label`s (single- and multi-widget selection). Plain single-click drag, double/triple-click without dragging, and shift-click are all unchanged.

## How
The click count (1/2/3) is normally only known on pointer *release*, but a double-/triple-click-and-drag needs it while the button is still held. So `PointerState` now also computes the click count at press time (`press_click_count`), reusing the same double/triple-click logic the release path already used. `TextCursorState` (shared by `TextEdit` and `Label`) gains a `SelectionMode` (`Char` / `Word` / `Line`) and an anchored unit, so the selection extends and shrinks in whole words/lines as the pointer moves back and forth.

## Known limitation

For multi-widget label selection, if the *anchor* widget (where the drag started) is scrolled out of view, the selection endpoint freezes at its last valid value. This is a limitation of egui's multi-widget text selection (char-mode selection on `main` has the same failure mode).

This PR addresses #2550

- [x] `cargo test -p egui` passes (added unit tests for the new logic)
- [x] `cargo clippy -p egui --all-features` is clean
- [x] `cargo fmt` is clean
- [x] Builds with and without the `serde` feature


denisps:
Why is this pull request not merging? This a significant usability feature. This text editing convention is implemented by most GUI frameworks and operating systems. This pull request makes text selection behavior 100% consistent with most major operating systems, GUI frameworks,, text editors, and browsers.

PR #8169, opened 2026-05-18
Always enable windows undecorated shadows

Wumpf:
I wanted to do this in a backwards compatible way so it can go into a patch. That said, it's arguably very much a breaking change as it changes behavior. Alternatively we could also expose this as an option on the ViewportCommand.

What does this actually do? Easily described with pictures rather than words:

Before:
image

After:
image

I figured one would always want it as it enables the native shadow, but winit's doc warns `Enabling the shadow causes a thin 1px line to appear on the top of the window.`. Not sure whether it means the thin line visible here already because it looks rather natural to me but also isn't just on top but all around 🤷

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -1680,7 +1680,14 @@ fn process_viewport_command(
         ViewportCommand::Fullscreen(v) => {
             window.set_fullscreen(v.then_some(winit::window::Fullscreen::Borderless(None)));
         }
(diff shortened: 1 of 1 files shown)


emilk (approved):

emilk:
We've decided this is a breaking change, and so won't include it in a patch release.

But this PR will make it easier to do this and similar things in user land:
* https://github.com/emilk/egui/pull/8205

PR #8170, opened 2026-05-18
Fix resizing of `Grid`

### LINKED ISSUE #8168: Window resize issue
**Reepro**: Go to egui.rs. Make the Widget Gallery wider. Now try making it more narrow. You will fail.

* Introduced by #8152

The problem is probably the space-filling `Separator` widget.

I'm worried this is non-trivial to fix

lucasmerlin:
- fixes https://github.com/emilk/egui/issues/8168
- broke in https://github.com/emilk/egui/pull/8152

Now we get the min size by running a sizing pass once, when the resize drag starts


emilk (approved):
Simple and nice!

PR #8178, opened 2026-05-19
Exclude `take_app` from `wasm32` in `egui_kittest`

grtlr:
Otherwise `egui` won't compile for `wasm32` targets anymore.

A bit odd that this was not caught in CI.


lucasmerlin (approved):
Thanks!

PR #8184, opened 2026-05-21
Add `AsId` and `IdSalt` 

emilk:
## Related
* https://github.com/emilk/egui/pull/5851
* https://github.com/emilk/egui/pull/7988

## What
We want to make it easier to understand the lineage of a `Id` (which is the parent `Id`, and the parent of that, etc?)

As a first step of that, we want to clarify the different between a globally unique `Id`, and an `IdSalt`

I also introduced the `AsId` and `AsIdSalt` traits, which are implemented of anything that implements `Hash` and `Debug`. The `Debug` half of that is unused here, but will later be used in `Debug` builds to produce a proper tree.


crates/egui/src/ui_builder.rs
```diff
@@ -29,6 +29,13 @@ pub struct UiBuilder {
     pub classes: Classes,
 }
 
+/// Is this [`Ui`] a root or a child of another [`Ui`]?
+#[derive(Clone)]
+pub enum UiLineage {
+    Root(Id),
+    Child(IdSalt),
+}
+
```
lucasmerlin, line 41: I feel like this terminology is kinda confusing.

A Ui with a `Root` id can very well be a child of some other ui, and vice versa.

Also `Child` would imply that the id salt was created from the parent uis id, which is not really the case here. 


Maybe instead this should be `IdLineage` instead? Feels like this could be a more general concept

emilk: Good point! I renamed it `IdSource`

PR #8191, opened 2026-05-22
Choose restored window monitor by overlap

YelovSK:
This changes the monitor selection used when restoring a persisted window position.

Currently, `egui-winit` picks a monitor by checking whether the saved window position fits inside a loose monitor range. This can choose the wrong monitor when a saved window rectangle slightly overlaps another monitor.

My failure case was on Windows with two monitors:
- primary monitor on the right
- secondary monitor on the left
- window maximized on the primary monitor
- persisted outer position was slightly negative, e.g. `x = -8`, because of the invisible window border

That position matched both monitor ranges, so the restored maximized window could
open on the secondary monitor instead of the primary one.

This PR picks the monitor with the largest overlap with the saved window rectangle instead.

Related note: probably the best solution would be to save the normal window position when maximized, so that when unmaximizing, the window would get restored to the previous state. It's mentioned in this comment https://github.com/emilk/egui/issues/3494#issuecomment-1986985211. I tried doing that in [fix-windows-maximized-restore-placement](https://github.com/YelovSK/egui/tree/fix-windows-maximized-restore-placement), and it works, but it requires adding windows-sys as a dependency to call a relevant winapi, so that's probably not the right solution. Winit doesn't seem to provide an API that would return this information.

### DIFF
--- a/crates/egui-winit/src/window_settings.rs
+++ b/crates/egui-winit/src/window_settings.rs
@@ -161,15 +161,8 @@ fn find_active_monitor(
     let mut active_monitor_overlap = 0.0;
     for monitor in monitors {
         let window_size_px = window_size_pts * (egui_zoom_factor * monitor.scale_factor() as f32);
-        let monitor_position = monitor.position();
-        let monitor_size = monitor.size();
-        let x_overlap = (position_px.x + window_size_px.x)
-            .min(monitor_position.x as f32 + monitor_size.width as f32)
-            - position_px.x.max(monitor_position.x as f32);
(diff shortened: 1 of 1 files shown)


crates/egui-winit/src/window_settings.rs
emilk, line 153: I guess this default is not really applied anymore. Instead, the closest fitting monitor will be chosen. The comment should probably be updated to reflect this

emilk (approved):
Tested on Mac, and works fine!

PR #8198, opened 2026-05-25
`Panel`: never overflow available width, nor `max_width`

### LINKED ISSUE #8055: Resizable Panel stores and paints using overflowed inner rect instead of clamped panel rect

emilk:
* Closes https://github.com/emilk/egui/issues/8055
* Closes https://github.com/emilk/egui/pull/8056

This changes the behavior of `Panel` to NEVER overflow `Panel::max_size`, nor the available space in the parent UI.

If you do overflow it, the content will be silently clipped.


crates/egui/src/containers/panel.rs
```diff
@@ -627,12 +627,39 @@ impl Panel {
...
+
+        let mut outer_size = self
+            .outer_size(parent_ui)
+            .at_most(available_rect.size_along(self.side.axis()));
+
+        let mut outer_rect = {
+            let mut outer_rect = available_rect;
+            self.side.set_rect_size(&mut outer_rect, outer_size);
+            outer_rect
+        };
```
lucasmerlin, line 662:
```suggestion
        let mut outer_rect = available_rect;
        self.side.set_rect_size(&mut outer_rect, outer_size);
```

lucasmerlin (approved):
Nice!

emilk:
I don't think this should be in a patch release - I worry this will change the behavior of some existing code, but I'm not 100% sure

valadaptive:
@emilk Doesn't solve the "ever-growing panel" issue: https://github.com/ntsc-rs/ntsc-rs/issues/780

Issue #8043, opened 2026-03-28
[REGRESSION] macOS/metal window resize from top causes repaint-pauses


krisdigital:
**Describe the bug**
I upgraded from 0.29 to 0.34 both using wgpu and noticed that resizing the top of the window causes hangs after the upgrade. I checked with a minimal app and the latest egui from the main branch - it depends on the window width and if resizing from the bottom (is smooth) or from the top (choppy).

Code I used to test:

```
use eframe::Frame;

use eframe::egui::{self};

struct App {}

impl App {
    fn new(_cc: &eframe::CreationContext) -> Self {
        App {}
    }
}

impl eframe::App for App {
    //fn update(&mut self, ctx: &egui::Context, frame: &mut eframe::Frame) {}
    fn ui(&mut self, ui: &mut egui::Ui, _frame: &mut Frame) {
        ui.label("Hello");
    }
}

fn main() -> eframe::Result {
    let options = eframe::NativeOptions {
        persist_window: false,
        viewport: egui::ViewportBuilder::default().with_inner_size([800.0, 600.0]),
        renderer: eframe::Renderer::Wgpu,
        ..Default::default()
    };
    eframe::run_native(
        "egui-test",
        options,
        Box::new(|cc| Ok(Box::new(App::new(cc)))),
    )
}
```

**To Reproduce**
1. Open a minimal app on macOS
2. Resize the window to be full width
3. Grab the upper side and resize
4. Experience choppiness

**Expected behavior**
With 0.29 I could see that resize caused a little white flicker at the top, but otherwise it was smooth.

**Screenshots**

**0.34**

https://github.com/user-attachments/assets/f3c063e2-268e-4739-a054-d0ba1faad150

**0.29**

https://github.com/user-attachments/assets/cfcdb6af-4593-47d1-8a11-45c178ca3f42

**Desktop (please complete the following information):**
 - OS: macOS, M1/Metal
 - Version 15.7.4

**Additional context**
I tried profiling, I see the hangs in the profiler but I have no idea what could be the reason...

<img width="931" height="623" alt="Image" src="https://github.com/user-attachments/assets/51cf21ee-cd4c-422a-a1fd-0936a07bc18c" />

smoe:
We are seeing what looks like the same regression family on macOS [Tahoe 26.4] after upgrading from the `0.33.x` line to `0.34.1`, although our visible symptom differs from the original report.

In our case the failure is specifically tied to native child viewports:

[...]

PR #8203, opened 2026-05-26
Configure wgpu to be low-latency by default

lucasmerlin:
* initially done in https://github.com/emilk/egui/pull/8103, reverted in #8167 due to resize hangs
* Related: https://github.com/emilk/egui/issues/8043 (maybe closes??)

Turns out the resize hangs were caused by the present_with_transaction call. Disabling that when `desired_maximum_frame_latency == 1` causes the window to resize smoothly.

Thanks @krisdigital for noticing that connection: https://github.com/emilk/egui/issues/8043#issuecomment-4154440382


crates/egui-wgpu/src/lib.rs
emilk:
```suggestion
```

emilk (approved):

PR #8208, opened 2026-05-29
Fixes color picker hue drift at low alpha values

aedm:

At low apha values, premultiplied colors lose precision.
This PR makes the color picker use unmultiplied colors internally.

Before:

https://github.com/user-attachments/assets/4617a355-daa9-4911-86e6-518ac6867014

After:

https://github.com/user-attachments/assets/d9681b01-50d8-418e-b5a5-79b4bd1bbddf

### DIFF
--- a/crates/egui/src/widgets/color_picker.rs
+++ b/crates/egui/src/widgets/color_picker.rs
@@ -85,7 +85,29 @@ pub fn show_color_at(painter: &Painter, color: Color32, rect: Rect) {
     }
 }
 
-fn color_button(ui: &mut Ui, color: Color32, open: bool) -> Response {
+/// Show a color with background checkers to demonstrate transparency (if any).
+fn show_srgba_unmultiplied(ui: &mut Ui, srgba: [u8; 4], desired_size: Vec2) -> Response {
+    let (rect, response) = ui.allocate_at_least(desired_size, Sense::hover());
+    if ui.is_rect_visible(rect) {
+        show_srgba_unmultiplied_at(ui.painter(), srgba, rect);
+    }
+    response
+}
+
+/// Show a color with background checkers to demonstrate transparency (if any).
+fn show_srgba_unmultiplied_at(painter: &Painter, [r, g, b, a]: [u8; 4], rect: Rect) {
+    if a == 255 {
+        painter.rect_filled(rect, 0.0, Color32::from_rgb(r, g, b));
+    } else {
+        background_checkers(painter, rect);
+        let left = Rect::from_min_max(rect.left_top(), rect.center_bottom());
+        let right = Rect::from_min_max(rect.center_top(), rect.right_bottom());
+        painter.rect_filled(left, 0.0, Color32::from_rgba_unmultiplied(r, g, b, a));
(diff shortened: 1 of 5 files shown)


crates/egui_demo_lib/tests/snapshots/tessellation_test/Blurred.png
lucasmerlin, line 1: It's weird that it changed from transparent to black here, was this broken before or is this broken now?

aedm: Honestly, I don't quite understand the reason it was black before. But when alpha is exactly 0, it displayed black on purpose, even for HSVA where it didn't make much sense. I guess it was a bug before.

`Color32`'s docs say: "An `alpha=0` means the color is to be treated as an additive color."
So zero alpha is special, but it's not black.

crates/egui/src/widgets/color_picker.rs
```diff
@@ -85,7 +85,29 @@ pub fn show_color_at(painter: &Painter, color: Color32, rect: Rect) {
...
 
-fn color_button(ui: &mut Ui, color: Color32, open: bool) -> Response {
+/// Show a color with background checkers to demonstrate transparency (if any).
+fn show_srgba_unmultiplied(ui: &mut Ui, srgba: [u8; 4], desired_size: Vec2) -> Response {
+    let (rect, response) = ui.allocate_at_least(desired_size, Sense::hover());
+    if ui.is_rect_visible(rect) {
+        show_srgba_unmultiplied_at(ui.painter(), srgba, rect);
+    }
+    response
+}
```
lucasmerlin, line 95: Couldn't this just call show_color (with the right conversion)?

crates/egui/src/widgets/color_picker.rs
```diff
@@ -85,7 +85,29 @@ pub fn show_color_at(painter: &Painter, color: Color32, rect: Rect) {
...
+    if a == 255 {
+        painter.rect_filled(rect, 0.0, Color32::from_rgb(r, g, b));
+    } else {
+        background_checkers(painter, rect);
+        let left = Rect::from_min_max(rect.left_top(), rect.center_bottom());
+        let right = Rect::from_min_max(rect.center_top(), rect.right_bottom());
+        painter.rect_filled(left, 0.0, Color32::from_rgba_unmultiplied(r, g, b, a));
+        painter.rect_filled(right, 0.0, Color32::from_rgb(r, g, b));
+    }
+}
```
lucasmerlin, line 108: Same, can't this just call show_color_at?

lucasmerlin (approved):
Nice, great fix!

crates/egui/src/widgets/color_picker.rs
aedm: Unfortunately, `show_color` itself is the problem.

It expects the color as `Into<Color32>`. It converts the color into premultiplied `Color32` internally which has a low accuracy at low alpja values.

crates/egui/src/widgets/color_picker.rs
aedm: Same as above.

PR #8210, opened 2026-05-30
Fix #2142 - lost_focus not firing after a mid-frame focus transfer

### LINKED ISSUE #2142: lost_focus on a single_text_edit widget does not get called in several cases

ufoscout:
### Problem
`Response::lost_focus()` could silently fail to fire when keyboard focus moved from one widget to another *within the same frame* — for example, clicking a `TextEdit` that was added to the UI *after* the currently-focused one.

### Fix
This widens the detection window by one extra frame, which is exactly enough for the deferred loss signal to reach the previously focused widget on its next render.

### Notes
* The `test_demo_app` test fails, but it has nothing to do with this PR; it fails on the current master branch, too.

* This PR replaces https://github.com/emilk/egui/pull/3247
* Closes <https://github.com/emilk/egui/issues/2142>

### DIFF
--- a/crates/egui/src/memory/mod.rs
+++ b/crates/egui/src/memory/mod.rs
@@ -490,6 +490,13 @@ pub(crate) struct Focus {
     /// The ID of a widget that had keyboard focus during the previous frame.
     id_previous_frame: Option<Id>,
 
+    /// The ID of a widget that had keyboard focus *two* frames ago.
+    ///
+    /// Kept so `Response::lost_focus` can still fire after a mid-frame
+    /// focus transition (e.g. clicking a `TextEdit` that was added to
+    /// the UI later than the currently focused one).
+    id_two_frames_ago: Option<Id>,
+
     /// The ID of a widget to give the focus to in the next frame.
     id_next_frame: Option<Id>,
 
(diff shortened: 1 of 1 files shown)


emilk (approved):
thanks

PR #8214, opened 2026-06-01
Refactor scroll area fade painting logic

### LINKED ISSUE #8213: Bug: ScrollArea bottom shadow follows text when scrolling after PR #8018
Since Pull Request #8018, a portion of the shadow created at the very bottom inside the `ScrollArea` follows the text as the `ScrollArea` is scrolled up and down. 
Although it is possible to choose to disable the shadow in the options, [the rendering glitch remains when the shadow is active].

One additional issue is that the fade is rendered before the floating scrollbar, which can cause the fade effect to be overlaid on top of the floating scrollbar.

*  Related #8018

rustbasic:
Reserve the scroll area before painting fades and update the fade painting logic.

* Closes #8213

### DIFF
--- a/crates/egui/src/containers/scroll_area.rs
+++ b/crates/egui/src/containers/scroll_area.rs
@@ -1042,15 +1042,13 @@ impl ScrollArea {
             .inner;
 
         let (content_size, state) = prepared.end(ui);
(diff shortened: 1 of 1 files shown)


emilk:
thanks!

PR #8219, opened 2026-06-02
Add `AtomKind::Layout`, for nesting `AtomLayout`

lucasmerlin:
This allows you to put an `AtomLayout` inside another `AtomLayout`. Right now this has limited use, but once we add wrapping, vertical `AtomLayout` and `AtomUi`, this will be a really powerful new layouting primitive for egui.

Added a test for this in https://github.com/emilk/egui/pull/8221


crates/egui/src/atomics/atoms.rs
```diff
@@ -18,7 +13,7 @@ pub(crate) const ATOMS_SMALL_VEC_SIZE: usize = 2;
 /// ui.button((image, "Click me!"));
 /// # });
 #[derive(Clone, Debug, Default)]
-pub struct Atoms<'a>(SmallVec<[Atom<'a>; ATOMS_SMALL_VEC_SIZE]>);
+pub struct Atoms<'a>(Vec<Atom<'a>>);
```
lucasmerlin, line 16: Unfortunately the nesting was impossible with smallvec due to it being [invariant](https://github.com/servo/rust-smallvec/issues/146) (whatever that means). I've run into this before.
So I switched to Vec. We could also use `tinyvec` or the `smallvec` 2.0 beta, which fixes this problem.

crates/egui/src/atomics/atom_layout.rs
lucasmerlin, line 46: I wonder if we should also split AtomLayout into a `Widget` (`WidgetAtom`?) and a `Layout` (`ContainerAtom`?) part. Because now it kinda does both, represent how a group of Atoms should be laid out / rendered (what Margin / Frame / etc), and how it should behave as a widget (what Id / Sense should it have).  

We'd then still want both a `AtomKind::Container` and a `AtomKind::Widget`, since it allocating a Response is what makes the future `AtomTrait` work via reading back the Response.

lucasmerlin: Implemented it here:
https://github.com/emilk/egui/pull/8222

crates/egui/src/atomics/atoms.rs
emilk: This means an additional allocation for each widget 😬  Something I really would like to avoid.

Have you benchmarked the impact of this?

emilk (commented):
I would really like to avoid that extra allocation 😬

lucasmerlin:
So I did some benchmarking and experimenting and `Button` performance _improves_ by almost 10% when replacing SmallVec with Vec here. 
I guess its because atoms are moved around a lot and moving a small pointer is cheaper than moving the full inline smallvec. 
I also tried some more things, I tried the new smallvec 2.0 beta, tried tinyvec and using smallvec with boxed enum variants. 
Smallvec with boxed enum variants also led to improvements but not as much as just using a plain Vec. 



| Benchmark | main | branch | Δ (median) | result |
|---|---:|---:|---:|:--|
| `button/3_button_text_image_right_text` | 907.0 ns | 821.3 ns | **−9.4%** | ✅ improved |
| `button/1_button_text` | 450.8 ns | 410.3 ns | **−9.0%** | ✅ improved |
| `button/4_button_italic` | 482.4 ns | 443.5 ns | **−8.1%** | ✅ improved |
| `button/2_button_text_image` | 621.6 ns | 580.5 ns | **−6.6%** | ✅ improved |
| `demo_no_tessellate` | 93.74 µs | 91.37 µs | **−3.4%** | ✅ improved |
| `demo_with_tessellate__realistic` | 131.51 µs | 129.43 µs | −2.2% | ✅ improved (near noise) |
| `demo_only_tessellate` | 37.85 µs | 38.38 µs | +1.2% | ≈ noise |
| `label &str` | 255.8 ns | 257.5 ns | +0.7% | — no change |
| `label format!` | 256.4 ns | 257.4 ns | +0.4% | — no change |
| `text_layout_uncached` | 118.53 µs | 118.44 µs | −0.2% | — no change (untouched) |
| `text_layout_cached_many_lines_modified` | 253.0 µs | 248.7 µs | −1.7% | ≈ noise (untouched) |
| `tessellate_text` | 5.96 µs | 5.88 µs | −2.1% | ≈ noise (untouched) |
| `text_layout_cached` | 168.4 ns | 164.9 ns | −2.2% | ≈ noise (untouched) |
| `Painter::rect` | 18.5 ns | 16.4 ns | −11.3% | ≈ noise (untouched, ~18 ns timescale) |

emilk (approved):

PR #8224, opened 2026-06-04
Add `AtomLayout::selectable` for opt-in text selection

### LINKED ISSUE #8217: Selectable text in `Atom::Text`
Should be fairly simple. Just call `pub fn label_text_selection(`

akagifreeez:
* Closes https://github.com/emilk/egui/issues/8217

## What

`AllocatedAtomLayout::paint` painted text via `ui.painter().galley(...)`, so text in atom-based widgets could never be selected. As discussed in #8217, routing it through `LabelSelectionState::label_text_selection` unconditionally would break `Button` / `Checkbox` / `RadioButton` / `TextEdit` (whose labels should not be selectable, and whose click/drag handling would fight the selection drag).

So this adds an **opt-in** `AtomLayout::selectable(bool)` (default `false`). When enabled, the layout also senses clicks and drags (mirroring `Label::layout_in_ui`) and paints its text through the label-selection machinery. The underline is `Stroke::NONE`, so when nothing is selected the painted output is identical to the existing path — the default (`selectable == false`) behaviour, and existing snapshots, are unchanged.

## Tests

Two tests in `tests/egui_tests/tests/test_atoms.rs`:

- `test_atom_selectable_senses_click_and_drag` — a `selectable(true)` layout senses click+drag; the default stays inert.
- `test_atom_selectable_text_can_be_copied` — selecting (drag) the text of a selectable layout and copying yields the text via `OutputCommand::CopyText`, while a non-selectable layout yields nothing.

## Notes

- Verified locally: `cargo fmt --all -- --check`, `cargo clippy -p egui` (clean), and the two new tests pass. I haven't run the full `./scripts/check.sh` (wasm/typos) locally, hence opening as a draft.
- This only adds the opt-in API; no existing widget is made selectable. Happy to wire it up on a specific widget and/or add an `egui_demo_lib` demo if you'd like — just let me know.

### DIFF
--- a/crates/egui/src/atomics/atom_layout.rs
+++ b/crates/egui/src/atomics/atom_layout.rs
@@ -1,7 +1,7 @@
 use crate::atomics::ATOMS_SMALL_VEC_SIZE;
 use crate::{
     AtomKind, Atoms, FontSelection, Frame, Id, Image, IntoAtoms, Response, Sense, SizedAtom,
-    SizedAtomKind, Ui, Widget,
+    SizedAtomKind, Stroke, Ui, Widget, text_selection::LabelSelectionState,
 };
 use emath::{Align2, GuiRounding as _, NumExt as _, Rect, Vec2};
 use epaint::text::TextWrapMode;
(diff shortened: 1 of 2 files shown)


emilk (approved):

lucasmerlin (approved):
Awesome! 
I've noticed some weird behavior though:

https://github.com/user-attachments/assets/dab6d5d8-c930-4053-983c-c9989220992d

`ui.add(AtomLayout::new(("123", "testi", "rerun")).selectable(true));`

emilk:
I suspect that bug is something deeper than what's in this PR, so I'll merge this anyways

Issue #5063, opened 2024-09-03
Unexpected Program UI Freeze While Using egui Program on Windows 10 (glow Renderer)


rustbasic:
egui: Master Version + All Versions
Renderer: glow
OS: Windows 10

When using a program created with egui on Windows 10 for an entire day,
the screen rendering stops about once or twice a day, causing the screen to freeze.

The program continues to run in the background, but the displayed content becomes unresponsive, and there's no way to resume the rendering or make the program visually progress.
This issue has existed in all previous versions of egui.

This issue is not triggered by any specific scenario, making it impossible to create a reproducible example.
Since it mainly occurs when compiled with Windows 10 and `glow`,
it's unclear whether this issue also happens with `wgpu`.

Has anyone else experienced this phenomenon? Am I the only one?

NotNite:
A user of [my application](https://github.com/moonlight-mod/moonlight-installer) can't even get past the *first paint* without a freeze. No logs appear past the exit of my `update`. Switching to the wgpu backend works flawlessly.

Interestingly enough, the [eframe template](https://github.com/emilk/eframe_template) works fine on their machine...

PR #8231, opened 2026-06-10
fix: prevent stale `Focused(false)` events from clearing valid focus and freezing rendering

rustbasic:
fix: prevent stale `Focused(false)` events from clearing valid focus and freezing rendering

* Related #5063 

## Description

This PR fixes a rare and elusive screen-freezing issue in `egui` (especially noticeable on Windows 10/11) where the UI rendering permanently stops after running the application for a prolonged period (e.g., once or twice a day during all-day usage).

### Cause of the Issue
1. **Unconditional Focus Clearing:**
   Previously, any `WindowEvent::Focused(false)` event would unconditionally reset `focused_viewport` to `None`. In multi-window environments or when Windows OS dispatches stale/out-of-order focus events, a stale `Focused(false)` event from another window/viewport would clear the currently active viewport's focus state.
2. **Missing Repaint Trigger:**
   Once the focus state was incorrectly cleared or desynchronized, the application would stop repainting under certain conditions, causing the screen to freeze while the background logic continued running responsiveness-free.

### Solution
1. **Guarded Focus Clearing:**
   Only clear `shared.focused_viewport` if the viewport losing focus actually matches the one we currently believe to be focused:
   ```rust
   } else if shared.focused_viewport == Some(viewport_id) {
       shared.focused_viewport = None;
   }
   ```
2. **Forced Repaint on Focus Event:**
   Explicitly set `repaint_asap = true` on any `Focused` event to ensure the UI updates immediately and stays responsive when focus transitions occur.

Since this bug occurs semi-randomly over long sessions and is highly dependent on OS-level window focus dispatching, it is extremely difficult to write a minimal reproducible example. However, guarding the state transition and forcing a repaint completely resolves the long-term freeze issue.

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -681,9 +681,6 @@ impl GlowWinitRunning<'_> {
         };
 
         viewport.info.events.clear(); // they should have been processed
-        // let window = viewport.window.clone().unwrap();
-        // let gl_surface = viewport.gl_surface.as_ref().unwrap();
-        // let egui_winit = viewport.egui_winit.as_mut().unwrap();
         let (Some(egui_winit), Some(window), Some(gl_surface)) = (
             viewport.egui_winit.as_mut(),
             &viewport.window.clone(),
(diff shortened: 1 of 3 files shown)


rustbasic:
Although the situation has improved, it has not been completely resolved yet.
Therefore, I am continuously conducting tests and making modifications.

PR #8233, opened 2026-06-11
docs: add AccessKit accessibility guide

### LINKED ISSUE #4396: egui access_kit API docs
Is there documentation on how to use access_kit as a ui developer using egui? I looked through both egui and access kit docs and don't see documentation for developers on how to develop accessible apps.  I've read a bit of the source code I thought might be relevant but haven't had much luck yet.

From what I can tell enabling access_kit mostly just "works" (about as good as an iOS app using VoiceOver which is pretty impressive), but it wasn't apparent that would be the case until I set everything up and ran VoiceOver with my app.

I'd like to see documentation explaining to ui-developers using egui:
- how to create accessible experiences using access kit (high level summary)
- documentation explaining what gets accessibility labels for free (label text, for example)
- documentation explaining how to provide accessibility labels to widgets that don't have any.  For example, I have a component with a visual, non-textual representation that has an informational component that should be shared with screen reader users. How can I add accessibility labeling to that widget?

In addition to ui-developer docs, I'm also interested to know if the accessibility tree can be accessed at runtime in order to read things like "label_a has text 'wowow'". I'm interested in the possibility of using access kit as a way to drive out ui tests for egui.

Reading source code until I understand the capabilities, which I have greater risk misunderstanding without docs

**Additional context**
Thanks for the great library! :)

chatman-media:
Summary:
- add docs/accessibility.md with labels, custom widgets, and testing guidance
- link the guide from the README accessibility FAQ

Closes #4396

Checks:
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,6 +275,8 @@ The async version of [rfd](https://docs.rs/rfd/latest/rfd/) supports both native
 ### What about accessibility, such as screen readers?
 egui includes optional support for [AccessKit](https://accesskit.dev/), which currently implements the native accessibility APIs on Windows and macOS. This feature is enabled by default in eframe. For platforms that AccessKit doesn't yet support, including web, there is an experimental built-in screen reader; in [the web demo](https://www.egui.rs/#demo) you can enable it in the "Backend" tab.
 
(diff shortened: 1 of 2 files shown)


lucasmerlin (approved):
Nice!

PR #8234, opened 2026-06-11
Add `egui_inspection` protocol and plugin

lucasmerlin:
Introduces live inspection for running egui apps over a small TCP request/response protocol, plus the `egui::Plugin` that serves it.

This is the minimal surface to get the egui mcp in, we may want to extend this in the future to add support for the inspection gui.

### DIFF
--- a/crates/egui_inspection/src/lib.rs
+++ b/crates/egui_inspection/src/lib.rs
@@ -1,0 +1,57 @@
+#![cfg_attr(doc, doc = include_str!("../README.md"))]
+//!
+//! ## Feature flags
+#![cfg_attr(feature = "document-features", doc = document_features::document_features!())]
+
+pub mod protocol;
+
+pub use protocol::{
+    MAX_MESSAGE_BYTES, PROTOCOL_VERSION, Request, Response, read_message, write_message,
+};
+
+/// Environment variable that *enables* inspection when set to any non-empty value other than
+/// `0`/`false`. The peer binds [`DEFAULT_INSPECTION_ADDR`] unless [`INSPECTION_ADDR_ENV_VAR`]
+/// overrides it.
+pub const INSPECTION_ENABLE_ENV_VAR: &str = "EGUI_INSPECTION";
+
+/// Environment variable holding a `host:port` bind address (e.g. `127.0.0.1:5719` for
+/// local-only, `0.0.0.0:5719` to expose across the network). Setting it also enables
+/// inspection, so you don't need [`INSPECTION_ENABLE_ENV_VAR`] as well.
+pub const INSPECTION_ADDR_ENV_VAR: &str = "EGUI_INSPECTION_ADDR";
+
+/// Default bind address when inspection is enabled without an explicit
+/// [`INSPECTION_ADDR_ENV_VAR`]: loopback only, on a fixed well-known port. The `egui_mcp`
+/// server defaults its `attach` to this same port.
+pub const DEFAULT_INSPECTION_ADDR: &str = "127.0.0.1:5719";
+
+#[cfg(feature = "png")]
+mod png;
+
+#[cfg(feature = "png")]
+pub use png::encode_png;
+
+#[cfg(feature = "plugin")]
+mod plugin;
+
+#[cfg(feature = "plugin")]
+pub use plugin::{InspectionPlugin, attach_from_env, serve};
+
+/// Resolve the bind address from the environment, returning `None` when inspection is not
+/// enabled. Used by [`attach_from_env`] and eframe's auto-attach.
+#[cfg(feature = "plugin")]
+pub fn bind_addr_from_env() -> Option<String> {
+    if let Ok(addr) = std::env::var(INSPECTION_ADDR_ENV_VAR) {
+        let addr = addr.trim();
+        if !addr.is_empty() {
+            return Some(addr.to_owned());
+        }
+    }
+    match std::env::var(INSPECTION_ENABLE_ENV_VAR) {
+        Ok(v) => {
+            let v = v.trim();
+            let enabled = !(v.is_empty() || v == "0" || v.eq_ignore_ascii_case("false"));
+            enabled.then(|| DEFAULT_INSPECTION_ADDR.to_owned())
+        }
+        Err(_) => None,
+    }
+}
--- a/crates/egui_inspection/src/lib.rs
+++ b/crates/egui_inspection/src/lib.rs
@@ -14,14 +14,17 @@ pub use protocol::{
 /// overrides it.
 pub const INSPECTION_ENABLE_ENV_VAR: &str = "EGUI_INSPECTION";
 
-/// Environment variable holding a `host:port` bind address (e.g. `127.0.0.1:5719` for
-/// local-only, `0.0.0.0:5719` to expose across the network). Setting it also enables
-/// inspection, so you don't need [`INSPECTION_ENABLE_ENV_VAR`] as well.
+/// Environment variable holding a `host:port` bind address.
+///
+/// E.g. `127.0.0.1:5719` for local-only, `0.0.0.0:5719` to expose across the network. Setting
+/// it also enables inspection, so you don't need [`INSPECTION_ENABLE_ENV_VAR`] as well.
 pub const INSPECTION_ADDR_ENV_VAR: &str = "EGUI_INSPECTION_ADDR";
 
 /// Default bind address when inspection is enabled without an explicit
-/// [`INSPECTION_ADDR_ENV_VAR`]: loopback only, on a fixed well-known port. The `egui_mcp`
-/// server defaults its `attach` to this same port.
+/// [`INSPECTION_ADDR_ENV_VAR`].
+///
+/// Loopback only, on a fixed well-known port. The `egui_mcp` server defaults its `attach` to
+/// this same port.
 pub const DEFAULT_INSPECTION_ADDR: &str = "127.0.0.1:5719";
 
 #[cfg(feature = "png")]
--- a/crates/egui_inspection/src/lib.rs
+++ b/crates/egui_inspection/src/lib.rs
@@ -6,22 +6,18 @@
 pub mod protocol;
 
 pub use protocol::{
-    MAX_MESSAGE_BYTES, PROTOCOL_VERSION, Request, Response, read_message, write_message,
+    EncodedPng, MAX_MESSAGE_BYTES, PROTOCOL_MAGIC, PROTOCOL_VERSION, Request, Response,
+    read_message, write_message,
 };
 
-/// Environment variable that *enables* inspection when set to any non-empty value other than
-/// `0`/`false`. The peer binds [`DEFAULT_INSPECTION_ADDR`] unless [`INSPECTION_ADDR_ENV_VAR`]
-/// overrides it.
-pub const INSPECTION_ENABLE_ENV_VAR: &str = "EGUI_INSPECTION";
-
-/// Environment variable holding a `host:port` bind address.
+/// The single environment variable that controls inspection.
 ///
-/// E.g. `127.0.0.1:5719` for local-only, `0.0.0.0:5719` to expose across the network. Setting
-/// it also enables inspection, so you don't need [`INSPECTION_ENABLE_ENV_VAR`] as well.
-pub const INSPECTION_ADDR_ENV_VAR: &str = "EGUI_INSPECTION_ADDR";
+/// Interpreted by [`bind_addr_from_env`]: falsy (unset / empty / `0` / `false`) disables it;
+/// truthy (`1` / `true`) enables it on [`DEFAULT_INSPECTION_ADDR`]; anything else is taken as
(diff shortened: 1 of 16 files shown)


crates/egui_inspection/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// Environment variable that *enables* inspection when set to any non-empty value other than
+/// `0`/`false`. The peer binds [`DEFAULT_INSPECTION_ADDR`] unless [`INSPECTION_ADDR_ENV_VAR`]
+/// overrides it.
+pub const INSPECTION_ENABLE_ENV_VAR: &str = "EGUI_INSPECTION";
+
+/// Environment variable holding a `host:port` bind address (e.g. `127.0.0.1:5719` for
+/// local-only, `0.0.0.0:5719` to expose across the network). Setting it also enables
+/// inspection, so you don't need [`INSPECTION_ENABLE_ENV_VAR`] as well.
+pub const INSPECTION_ADDR_ENV_VAR: &str = "EGUI_INSPECTION_ADDR";
```
emilk: Maybe: a single env-var (for simplicity) that is either truthy, falsy, or a bind-addr

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+use std::sync::mpsc;
+use std::time::Duration;
+
+use egui::{Context, FullOutput, RawInput};
+
+use crate::protocol::{PROTOCOL_VERSION, Request, Response};
+
+/// How long [`serve`]'s connection threads wait for the UI thread before giving up. Generous:
+/// a backgrounded window may not paint (and thus not service requests) for a while.
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(20);
```
emilk, line 44: If the window is in the background and we cannot paint, then we should at least give a helpful error message (e.g. eprintln and/or panic).

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+
+/// Per-request progress through the frame lifecycle.
+enum Phase {
```
emilk: request from whom to whom?

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+
+struct InFlight {
+    req: Request,
+    reply: mpsc::Sender<Response>,
+    phase: Phase,
+}
+
+/// An [`egui::Plugin`] that serves the inspection protocol. See the module docs.
+pub struct InspectionPlugin {
+    in_flight: Vec<InFlight>,
```
emilk, line 69:
```suggestion
    /// Requests we haven't responded to yet
    in_flight: Vec<InFlight>,
```

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+
+    fn input_hook(&mut self, ctx: &Context, input: &mut RawInput) {
+        // Nothing in flight → idle frame, do no work.
+        if self.in_flight.is_empty() {
+            return;
+        }
+
+        // Capture any screenshot reply produced for a previous `Screenshot` request. We
+        // observe (don't consume) so the host app still receives it.
+        let mut screenshot: Option<(u32, u32, Vec<u8>)> = None;
```
emilk:
```suggestion
        struct EncodedPng { size: [u32; 2], bytes: Vec<u8> }
        let mut screenshot: Option<EncodedPng> = None;
```

emilk: or epaint::ColorImage

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+        for ev in &input.events {
+            if let egui::Event::Screenshot { image, .. } = ev {
```
emilk: I think there is a bug here. You need to use `Screenshot::user_data` to map each screenshot request to the right InFlight request (e.g. using a incremented id)

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+        if let Some((w, h, png)) = screenshot {
+            self.in_flight.retain_mut(|item| {
+                if !matches!(item.phase, Phase::AwaitScreenshotPixels) {
+                    return true;
+                }
+                let _ = item.reply.send(Response::Screenshot {
+                    width: w,
+                    height: h,
+                    png: png.clone(),
+                });
```
emilk: This could also wrap `EncodePng` struct 🤷

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+                false
+            });
+        }
+
+        // Apply the input-side effect of new requests, dropping any that reply immediately.
+        // `label` is cloned out so the closure doesn't borrow `self` alongside the
+        // `retain_mut` borrow of `in_flight`.
+        let label = self.label.clone();
+        self.in_flight.retain_mut(|item| {
+            if !matches!(item.phase, Phase::New) {
```
emilk:
```suggestion
            if item.phase != Phase::New {
```

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+                return true;
+            }
+            match &item.req {
+                Request::Info => {
```
emilk:
```suggestion
                Request::GetInfo => {
```

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+                }
+                Request::Resize { width, height } => {
+                    ctx.send_viewport_cmd(egui::ViewportCommand::InnerSize(egui::vec2(
+                        *width as f32,
+                        *height as f32,
+                    )));
+                    item.phase = Phase::AwaitOutput;
+                    true
+                }
+                Request::Screenshot => {
```
emilk:
```suggestion
                Request::GetScreenshot => {
```

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+                        protocol_version: PROTOCOL_VERSION,
+                        label: label.clone(),
+                    });
+                    false
+                }
+                Request::GetTree => {
+                    item.phase = Phase::AwaitOutput;
+                    true
+                }
+                Request::HandleEvents { events } => {
```
emilk:
```suggestion
                Request::ApplyEvents { events } => {
```

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+                    input.events.extend(events.iter().cloned());
+                    item.phase = Phase::AwaitOutput;
```
emilk, line 192:
```suggestion
					// We send the Ack at the end of the frame so the agent can be sure
					// that the command has been executed (e.g. a button click created a file)
                    item.phase = Phase::AwaitOutput;
```

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+            (Phase::AwaitOutput, Request::GetTree) => {
+                let _ = item.reply.send(Response::Tree {
+                    step,
+                    pixels_per_point: output.pixels_per_point,
+                    accesskit: output.platform_output.accesskit_update.clone(),
+                });
+                false
+            }
+            (Phase::AwaitOutput, Request::HandleEvents { .. } | Request::Resize { .. }) => {
+                let _ = item.reply.send(Response::Ack);
```
emilk: To previous point: let's make it clear that the command has been executed (and not just received)

```suggestion
                let _ = item.reply.send(Response::Done);
```

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    let ctx = ctx.clone();
+    std::thread::Builder::new()
+        .name("egui_inspection_accept".into())
+        .spawn(move || {
+            for stream in listener.incoming() {
+                let Ok(stream) = stream else { continue };
+                let ctx = ctx.clone();
+                let _ = std::thread::Builder::new()
```
emilk: better to panic than silently ignore errors

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+/// On wasm there is no TCP listener; serving is unsupported.
+///
+/// # Errors
+/// Always, on wasm.
+#[cfg(target_arch = "wasm32")]
+pub fn serve(_ctx: &Context, _addr: &str) -> std::io::Result<()> {
+    Err(std::io::Error::other(
+        "egui_inspection: TCP serving is not supported on wasm",
+    ))
+}
```
emilk: rm

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+    let Ok(write_stream) = stream.try_clone() else {
+        return;
+    };
+    let mut reader = std::io::BufReader::new(stream);
+    let mut writer = std::io::BufWriter::new(write_stream);
+
+    loop {
+        let req: Request = match read_message(&mut reader) {
+            Ok(r) => r,
+            Err(_) => return, // EOF / decode error → client gone
```
emilk: should log

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+        };
+        let Some(rx) = ctx.with_plugin::<InspectionPlugin, _>(|p| p.submit(req)) else {
+            let _ = write_message(
```
emilk: don't ignore errors

crates/egui_inspection/src/plugin.rs
```diff
@@ -0,0 +1,329 @@
...
+pub fn serve(_ctx: &Context, _addr: &str) -> std::io::Result<()> {
+    Err(std::io::Error::other(
+        "egui_inspection: TCP serving is not supported on wasm",
+    ))
+}
+
+/// Connection handler: read framed requests, submit each to the plugin via the context, and
+/// write the framed response back. Returns on EOF or any I/O error.
+#[cfg(not(target_arch = "wasm32"))]
+fn serve_connection(stream: std::net::TcpStream, ctx: &Context) {
```
emilk: return a result instead
```suggestion
fn serve_connection(stream: std::net::TcpStream, ctx: &Context) -> Result<… {
```

crates/egui_inspection/src/png.rs
```diff
@@ -0,0 +1,21 @@
...
+
+/// Encode tightly-packed RGBA8 pixels (`width * height * 4` bytes) as PNG using `image`'s
+/// default settings (`CompressionType::Default` + `FilterType::Adaptive`).
+///
+/// PNG keeps high-resolution captures off the hot path of socket throughput — a 1550×2114
+/// RGBA8 buffer is ~13 MiB raw but typically <1 MiB encoded.
+///
+/// # Errors
+/// When the encoder fails (e.g. the buffer length doesn't match `width * height * 4`).
+pub fn encode_png(width: u32, height: u32, rgba: &[u8]) -> Result<Vec<u8>, image::ImageError> {
```
emilk: As noted elsewhere: this can be a nice `struct EncodedPng` and this its ctor

crates/egui_inspection/src/protocol.rs
```diff
@@ -0,0 +1,172 @@
...
+    Resize { width: u32, height: u32 },
+}
+
+/// Sent peer → inspector, exactly one per [`Request`].
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub enum Response {
+    /// Reply to [`Request::Info`].
+    Info {
+        /// [`PROTOCOL_VERSION`] of the peer.
+        protocol_version: u32,
```
emilk: This should probably be in binary instead, otherwise we are likely to get deserialization error before we get a protocol error

crates/egui_inspection/src/protocol.rs
```diff
@@ -0,0 +1,172 @@
...
+///
+/// This is the wire format; sync ([`read_message`]/[`write_message`]) and async transports
+/// share it via these helpers so the framing stays in lockstep.
+///
+/// # Errors
+/// On encode failure or a body exceeding `u32::MAX`.
+pub fn encode_frame<T: serde::Serialize>(value: &T) -> io::Result<Vec<u8>> {
+    let body = rmp_serde::to_vec(value).map_err(invalid_data)?;
+    let len = u32::try_from(body.len()).map_err(invalid_data)?;
+    let mut frame = Vec::with_capacity(4 + body.len());
```
emilk, line 160: i.e.: adding the protocol version here is better. I suggest 4 magic bytes for identification, and then four bytes with protocol version

emilk (approved):
Only 22 comments :)

PR #8241, opened 2026-06-18
Add `egui_mcp`

lucasmerlin:
Implements a mcp server to allow agents to interact with egui apps

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -621,7 +621,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "531a9155a481e2ee699d4f98f43c0ca4ff8ee1bfd55c31e9e98fb29d2b176fe0"
 dependencies = [
  "memchr",
- "regex-automata",
+ "regex-automata 0.4.8",
  "serde",
 ]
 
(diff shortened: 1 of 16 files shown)


crates/egui_inspection/src/plugin.rs
```diff
@@ -145,21 +157,33 @@ impl egui::Plugin for InspectionPlugin {
...
-                    continue;
-                }
-            };
             self.in_flight.retain_mut(|item| {
-                if item.phase == (Phase::AwaitScreenshot { id }) {
-                    let _ = item.reply.send(Response::Screenshot(png.clone()));
-                    false
-                } else {
-                    true
+                if item.phase != (Phase::AwaitScreenshot { id }) {
```
emilk, line 161: weird extra paranthesis

Cargo.lock
```diff
@@ -734,6 +734,18 @@ dependencies = [
  "libc",
```
emilk, line 734: I don't want to pull in 379 lines of dependencies into the main `egui` repo. I'm guessing most is because of `tokio`, which is needed because of `rmcp`

According to claude, we can replace all of rmcp with ~300 lines of hand-rolled code, thus getting rid of all things async.

…but I'm also thinking egui_mcp does not belong in the `egui` repo after all. If we move it to a new repo then I care less about what dependencies it uses.

lucasmerlin: > …but I'm also thinking egui_mcp does not belong in the `egui` repo after all. If we move it to a new repo then I care less about what dependencies it uses.

Makes sense, I'll put it into the [inspector repo](https://github.com/rerun-io/kittest_inspector) then

Issue #4758, opened 2024-07-02
Text Selection does not work correctly with multiple deferred viewports when the unfocused viewport requires repainting


lukexor:
**Describe the bug**

This is a bit of an odd case and took me a long time to track down.

Use case:

Two windows - a primary and a secondary.
Primary window requires re-painting at 60fps (rendering animations/frames/etc)
Secondary window only requires re-painting in response to user interaction (e.g. a preferences window)

Trying to select text in the secondary window works fine only when it's the one being repainted. If the primary window repaints after selection has started, selection is cleared.

The relevant lines of code are from the `label_text_selection.rs` module:

```rust
        if !state.has_reached_primary || !state.has_reached_secondary {
            // We didn't see both cursors this frame,
            // maybe because they are outside the visible area (scrolling),
            // or one disappeared. In either case we will have horrible glitches, so let's just deselect.

            let prev_selection = state.selection.take();
```

I suspect the reason this happens is because the store/load happening for the temporary selection state is not keyed at all by viewport, and because deferred viewports require their own begin/end frame cycle it ends up clearing the selection since the primary window was repainted and no widgets matched the current selection state.

```rust
    pub fn load(ctx: &Context) -> Self {
        ctx.data(|data| data.get_temp::<Self>(Id::NULL))
            .unwrap_or_default()
    }

    pub fn store(self, ctx: &Context) {
        ctx.data_mut(|data| {
            data.insert_temp(Id::NULL, self);
        });
    }
```

**To Reproduce**

1. Modify the test_viewports example to add a `ctx.request_repaint_of(ViewportId::ROOT);` to the main `update` method
2. Open a deferred viewport and try to select a label - it gets immediately cleared
3. Open an immediate viewport and try to select a label - works as expected

**Expected behavior**

I'd expect selecting labels to work correctly

**Desktop (please complete the following information):**
 - OS: Linux

**Additional context**

I was able to get a working version with the following. If it's acceptable, I can create a PR:

```rust
    pub fn load(ctx: &Context) -> Self {
        let id = Id::new(ctx.viewport_id());
        ctx.data(|data| data.get_temp::<Self>(id))
            .unwrap_or_default()
    }

    pub fn store(self, ctx: &Context) {
        let id = Id::new(ctx.viewport_id());
        ctx.data_mut(|data| {
            data.insert_temp(id, self);
        });
    }

```

emilk:
One selection state per viewport makes sense to me 👍

PR #8242, opened 2026-06-18
Fix label selection in deferred viewports

yay:
Label text selection before the fix in a deferred viewport:

before_the_fix

After the fix:

after_the_fix

## What changed

- Keep label text-selection state separate for each viewport.
- Route pass lifecycle and label painting through the current `ViewportId`.
- Drop inactive per-viewport state after its pass.
- Add a regression test that verifies an unrelated viewport pass cannot clear a child viewport's
  selection, while the owning viewport still clears selections whose labels disappear.

## Why

Issue #4758 identified that deferred viewports need independent label-selection state. PR #4760
fixed it by keying the temporary state by viewport. The plugin refactor in PR #7385 moved that
state into one context-wide `LabelSelectionState`, which accidentally removed the viewport
isolation. A pass in another viewport then fails to encounter the selected widgets and clears the
selection.

This restores the behavior of #4760 within the current plugin architecture. Applications do not
need any special handling.

### DIFF
--- a/crates/egui/src/text_selection/label_text_selection.rs
+++ b/crates/egui/src/text_selection/label_text_selection.rs
@@ -4,7 +4,7 @@ use emath::TSTransform;
 
 use crate::{
     Context, CursorIcon, Event, Galley, Id, LayerId, Plugin, Pos2, Rect, Response, Ui,
-    layers::ShapeIdx, text::CCursor, text_selection::CCursorRange,
+    ViewportIdMap, layers::ShapeIdx, text::CCursor, text_selection::CCursorRange,
 };
 
 use super::{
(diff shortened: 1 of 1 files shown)


crates/egui/src/text_selection/label_text_selection.rs
```diff
@@ -693,3 +730,66 @@ fn estimate_row_height(galley: &Galley) -> f32 {
...
+        let child_viewport_id = ViewportId::from_hash_of("child_viewport");
+        let plugin = ctx.plugin::<LabelSelectionState>();
+        plugin
+            .lock()
+            .states
+            .entry(child_viewport_id)
+            .or_default()
+            .selection = Some(test_selection());
+
+        let _ = ctx.run_ui(RawInput::default(), |_| {});
```
yay, line 776: This is the regression trigger. Before the fix global plugin state would be reset by the root pass and then cleared because the selected child label was not encountered.

emilk (approved):
ah, nice fix

PR #8248, opened 2026-06-22
Allow downscaling image in `GetScreenshot` inspection request

lucasmerlin:
When an agent screenshots the app using the mcp and then interacts with the app by clicking at coords, they see the coords in the native image coords. Since the mcp does everything else in logical coordinates, it helps if the image they see is also in logical resoltution, so we always downscale it to 1.0. 

I've added this here to avioid having to decode and re-encode the image in the mcp. 
Unfortunately it only does downscaling for now, since adding some way to upscale the image just for the screenshot would add a lot of complexity, and might be invasive from a plugin.

I've also changed the submit call to take a closure, to make it easier to use other transport channel (makes the implementation for reruns mcp nicer).


crates/egui_inspection/src/png.rs
```diff
@@ -12,6 +12,28 @@ impl EncodedPng {
         Self::from_rgba(size, image.as_raw())
     }
 
+    /// Encode an [`egui::ColorImage`] downscaled by `scale` — a factor in `(0.0, 1.0]` of the
+    /// captured pixel dimensions. `scale >= 1.0` encodes at native resolution unchanged: the
```
IsseW, line 16: > `scale >= 1.0` encodes at native resolution unchanged

I can't really see where that happens

IsseW (approved):
Looks good from what I can tell! Just one comment that I think doesn't line up with code

crates/egui_inspection/src/png.rs
lucasmerlin: It's right there:
```rs
        if scale >= 1.0 || w == 0 || h == 0 {
            return Self::from_rgba([w, h], image.as_raw());
        }
```

IsseW: Ah somehow missed the `scale >= 1.0` in that if statement 😅

PR #8251, opened 2026-06-23
Improve docs on some methods to clarify what counts as a "click"

HactarCE:

Just a small docs change, since I saw this trip someone up. The ones on `Response` may be redundant.

### DIFF
--- a/crates/egui/src/input_state/mod.rs
+++ b/crates/egui/src/input_state/mod.rs
@@ -1420,6 +1420,7 @@ impl PointerState {
     /// Was the given pointer button given clicked this frame?
     ///
     /// Returns true on double- and triple- clicks too.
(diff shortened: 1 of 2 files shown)


lucasmerlin (commented):
It's kinda misleading to point to the [`Response::clicked`] docs everywhere, because that one also says 
```rs
    /// This will also return true if the widget was clicked via accessibility integration,
    /// or if the widget had keyboard focus and the use pressed Space/Enter.
```
which is not true for the ones you linked it from

lucasmerlin (changes requested):
Instead maybe copy the relevant doc line to the input state `button_clicked`, and link to there

HactarCE:
I've inlined just the relevant lines from the docs. Is this better?

lucasmerlin (approved):
Thanks!

PR #8255, opened 2026-06-23
Document vertical allocation behavior of `Ui::horizontal_top`

MichaelGrupp:
Use the same wording as `Ui::horizontal_centered`.

### DIFF
--- a/crates/egui/src/ui.rs
+++ b/crates/egui/src/ui.rs
@@ -2330,7 +2330,7 @@ impl Ui {
         self.allocate_ui_with_layout_dyn(initial_size, layout, Box::new(add_contents))
     }
 
-    /// Like [`Self::horizontal`], but aligns content with top.
+    /// Like [`Self::horizontal`], but allocates the full vertical height and then aligns content with top.
     pub fn horizontal_top<R>(
         &mut self,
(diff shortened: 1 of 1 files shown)


crates/egui/src/ui.rs
```diff
@@ -2330,7 +2330,7 @@ impl Ui {
         self.allocate_ui_with_layout_dyn(initial_size, layout, Box::new(add_contents))
     }
 
-    /// Like [`Self::horizontal`], but aligns content with top.
+    /// Like [`Self::horizontal`], but allocates the full vertical height and then aligns content with top.
```
MichaelGrupp, line 2333: see `horizontal_centered` a few lines above

lucasmerlin: I don't think this is correct, see the following example where the first horizontal_top doesn't allocate the full space:
```rs
            ui.horizontal_top(|ui| {
                ui.button("Hi");
            });

            ui.button("Quit");
```

image

But when replacing it with horizontal_centered, it does:

image

MichaelGrupp:
You're right. Turns out I got confused not by `horizontal_top`, but in my specific case by `separator` in `horizontal_top`: 
grafik

Same repro with `horizontal`: 
grafik

PR #8257, opened 2026-06-24
Call `logic` even while browser tab is in background

### LINKED ISSUE #5112: Run `App::update` even when in background thread

lucasmerlin:
* Closes https://github.com/emilk/egui/issues/5112

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1400,6 +1400,7 @@ dependencies = [
  "egui_kittest",
  "image",
  "jiff",
+ "log",
  "mimalloc",
  "rand 0.10.1",
  "serde",
--- a/crates/eframe/src/web/events.rs
+++ b/crates/eframe/src/web/events.rs
@@ -113,12 +113,32 @@ fn install_blur_focus(runner_ref: &WebRunner, target: &EventTarget) -> Result<()
     // NOTE: because of the text agent we sometime miss 'blur' events,
     // so we also poll the focus state each frame in `AppRunner::logic`.
     for event_name in ["blur", "focus", "visibilitychange"] {
-        let closure = move |_event: web_sys::MouseEvent, runner: &mut AppRunner| {
-            log::trace!("{} {event_name:?}", runner.canvas().id());
-            runner.update_focus();
-        };
+        let closure =
+            move |_event: web_sys::MouseEvent, runner: &mut AppRunner, web_runner: &WebRunner| {
+                log::trace!("{} {event_name:?}", runner.canvas().id());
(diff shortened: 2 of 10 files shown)


emilk (commented):
Cool! How did you test this?

lucasmerlin:
I had some manual tests by adding repainting and logging, but I've now made it into a misc demo, so we can test if it still works in the future. Might be useful when testing on other platforms as well (I also tested this on Mac, logic seems to work fine there):

https://github.com/user-attachments/assets/625c0e12-6cdf-4abf-acc4-d98f0171702e

grtlr (approved):
Looks good from what I can tell, nice!

(And tested locally)

PR #8263, opened 2026-06-25
Remove 64 bit atomics in egui_extras

### LINKED ISSUE #7692: Use `AtomicU32` in place of `AtomicU64`

kay-lambdadelta:
This is a bit like my previous pr (<https://github.com/emilk/egui/pull/8037>), but addresses the final bit of 64 bit atomic dependent code, the svg loader in the `egui_extras` crate.

As with the previous pr, this improves egui's compatibility on platforms without 64 bit atomics.

* Closes <https://github.com/emilk/egui/issues/7692>

### DIFF
--- a/crates/egui_extras/src/loaders/svg_loader.rs
+++ b/crates/egui_extras/src/loaders/svg_loader.rs
@@ -1,10 +1,4 @@
-use std::{
-    mem::size_of,
-    sync::{
-        Arc,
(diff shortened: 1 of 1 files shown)


lucasmerlin (approved):
Thanks!

PR #8271, opened 2026-06-28
Fix: Prevent panic caused by #8083

rustbasic:
This pull request prevents the panic caused by #8083.
I followed the solution suggested by `umajho` and have confirmed that the panic is successfully prevented.

* Related #8045
* Related #8083


rustbasic:
Resolved in #8045

PR #8276, opened 2026-06-29
Fix SelectableLabel text shift with inactive strokes

### LINKED ISSUE #8125: Button::selectable (SelectableLabel) shifts content by inactive.bg_stroke.width px on hover when frame_when_inactive = false
**egui version:** 0.34.1 (also present on `main` at time of writing)

### Summary

`Style::button_style` always subtracts `bg_stroke.width` from the button's `inner_margin` to compensate for the stroke that the `Frame` will draw. But when `frame_when_inactive = false` (the default for `Button::selectable` / `SelectableLabel`) and the widget is inactive, the button code swaps in a plain `Frame::new().inner_margin(...)` that keeps the compensated margin but does **not** draw the stroke.

That means the compensation has nothing to compensate for, and the content sits `bg_stroke.width` pixels closer to the outer edge than it does when hovered.

Net effect: moving the mouse over a `SelectableLabel` causes its text to jump horizontally and vertically by `inactive.bg_stroke.width` pixels. With default styling this is invisible because `inactive.bg_stroke = Stroke::default()` has width 0, but any custom style that gives inactive widgets a non-zero `bg_stroke.width` (for example, to show a visible inactive border on regular `Button`s) surfaces the bug everywhere `SelectableLabel` is used.

### Reproducer

```rust
use eframe::egui;

fn main() -> eframe::Result {
    eframe::run_simple_native("jitter", Default::default(), |ctx, _| {
        ctx.style_mut(|style| {
            // Give inactive widgets a stroke. This matches what many custom
            // themes do to show a visible border on idle buttons.
            style.visuals.widgets.inactive.bg_stroke.width = 1.0;
            style.visuals.widgets.inactive.bg_stroke.color =
                egui::Color32::from_gray(128);
        });

            for i in 0..5 {
                ui.selectable_label(false, format!("item {i}"));
            }
        });
    })
}
```

Hover over any item and the text visibly shifts by 1 px.

### Root cause

`crates/egui/src/widget_style.rs` (`Style::button_style`):

```rust
inner_margin: (self.spacing.button_padding + Vec2::splat(visuals.expansion)
    - Vec2::splat(visuals.bg_stroke.width))
.into(),
```

`crates/egui/src/widgets/button.rs` (`Button::atom_ui`):

```rust
layout = if has_frame_margin && (state != WidgetState::Inactive || frame_when_inactive) {
    layout.frame(frame) // stroke drawn
} else {
    layout.frame(Frame::new().inner_margin(frame.inner_margin)) // stroke NOT drawn
};
```

The `else` branch reuses the pre-compensated `inner_margin` but drops the stroke, so the content position becomes `button_padding - bg_stroke.width`. The `if` branch draws the stroke, so the content position remains `button_padding`. The gap is exactly `bg_stroke.width`.

sjh9714:
* Closes <https://github.com/emilk/egui/issues/8125>

## Summary

`Button::atom_ui` used the stroke-compensated frame margin even when an inactive frameless selectable button did not paint that stroke. With a non-zero inactive bg stroke, the text jumped when hovering and the hover frame was painted.

This restores the unpainted stroke padding for the inactive/no-frame path only, and adds a kittest regression that compares the text shape position before and after hover.

## Testing

- `PATH="$HOME/.cargo/bin:$PATH" cargo test -p egui_kittest selectable_label_text_should_not_shift_when_inactive_stroke_is_unpainted -- --nocapture` (failed before the fix with `[11.0 9.5]` vs `[12.0 9.5]`, passes after)
- `PATH="$HOME/.cargo/bin:$PATH" cargo fmt --check`
- `PATH="$HOME/.cargo/bin:$PATH" cargo test -p egui_kittest --test regression_tests` (passes; two pre-existing warnings in `override_text_color_affects_interactive_widgets`)
- `PATH="$HOME/.cargo/bin:$PATH" cargo test -p egui`
- `PATH="$HOME/.cargo/bin:$PATH" cargo clippy -p egui -- -D warnings`
- `PATH="$HOME/.cargo/bin:$PATH" cargo clippy -p egui -p egui_kittest --test regression_tests -- -D warnings` (fails on existing `clippy::unnecessary_wraps` in `crates/egui_kittest/src/app_kind.rs:26`)
- `git diff --check`

I opened this as draft because I did not run `./scripts/check.sh`.

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.

### DIFF
--- a/crates/egui/src/widgets/button.rs
+++ b/crates/egui/src/widgets/button.rs
@@ -364,7 +364,12 @@ impl<'a> Button<'a> {
         layout = if has_frame_margin && (state != WidgetState::Inactive || frame_when_inactive) {
             layout.frame(frame)
         } else {
-            layout.frame(Frame::new().inner_margin(frame.inner_margin))
+            let inner_margin = if has_frame_margin {
+                frame.inner_margin + Margin::from(frame.stroke.width)
+            } else {
+                frame.inner_margin
+            };
+            layout.frame(Frame::new().inner_margin(inner_margin))
         };
 
         let mut prepared = layout.min_size(min_size).allocate(ui);
--- a/crates/egui/src/widgets/button.rs
+++ b/crates/egui/src/widgets/button.rs
@@ -364,12 +364,7 @@ impl<'a> Button<'a> {
         layout = if has_frame_margin && (state != WidgetState::Inactive || frame_when_inactive) {
             layout.frame(frame)
         } else {
-            let inner_margin = if has_frame_margin {
-                frame.inner_margin + Margin::from(frame.stroke.width)
-            } else {
-                frame.inner_margin
-            };
-            layout.frame(Frame::new().inner_margin(inner_margin))
+            layout.frame(Frame::new().inner_margin(frame.inner_margin))
         };
 
         let mut prepared = layout.min_size(min_size).allocate(ui);
--- a/crates/egui_kittest/tests/regression_tests.rs
+++ b/crates/egui_kittest/tests/regression_tests.rs
@@ -191,6 +191,24 @@ pub fn override_text_color_affects_interactive_widgets() {
     results.add(harness.try_snapshot("override_text_color_interactive"));
 }
 
+/// <https://github.com/emilk/egui/issues/8125>
+#[test]
+pub fn selectable_label_text_should_not_shift_when_inactive_stroke_is_unpainted() {
+    let mut harness = Harness::new_ui(|ui| {
+        ui.visuals_mut().widgets.inactive.bg_stroke = Stroke::new(1.0, egui::Color32::GRAY);
+        let _ = ui.selectable_label(false, "item");
+    });
+
+    harness.run();
+    let inactive_text_pos = text_shape_pos(harness.output(), "item");
+
+    harness.get_by_label("item").hover();
+    harness.run();
+    let hovered_text_pos = text_shape_pos(harness.output(), "item");
+
+    assert_eq!(inactive_text_pos, hovered_text_pos);
+}
+
 /// <https://github.com/rerun-io/rerun/issues/11301>
 #[test]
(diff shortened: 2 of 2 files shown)


stergiotis:
Confirming this from a downstream theme that sets `inactive.bg_stroke.width = 1.0`
(inactive `expansion = 0`, hovered `expansion = 1`, hovered stroke `2.0`,
`button_padding = (8, 4)`).

Measured before the patch: an unselected `Button::selectable` allocates 2 px less per
axis than its hovered self — in a `ComboBox` popup the frameless row measured 24 px tall
at rest and 26 px hovered, pushing the two rows below it down by 2 px. So the effect is
not only the text shift; `response.rect` itself changes, and neighbours move with it.

With this patch the frameless-inactive margin becomes `7+1 / 3+1`, i.e. the same
`(16, 8)` total as the framed states, which resolves it for that style. I worked the
arithmetic through by hand rather than building with the patch applied.

Two small notes:

1. The stroke is folded back but `outer_margin` (`= -expansion`) is not, so the totals
   match exactly only when `inactive.expansion == 0` — true of every stock theme, but a
   theme setting it would still be off by `2 × expansion`. `Margin::from(f32)` also
   rounds, so a fractional inactive stroke (1.5) leaves half a pixel. A form that avoids
   the arithmetic entirely is to keep the same `Frame` and just make it invisible —
   `frame.fill(Color32::TRANSPARENT).stroke(Stroke::new(w, Color32::TRANSPARENT))` —
   geometry identical by construction, at the cost of one transparent shape in the paint
   list.

2. The regression test compares text positions. Since the allocated rect is what moves
   surrounding widgets, asserting `response.rect.size()` across the two states as well
   would pin the invariant more directly.

sjh9714:
Current main now handles this with an invisible frame, so I removed the old production change while resolving the conflict. The regression test checks both the response size and text position before and after hover.

PR #8279, opened 2026-06-30
Apply texture deltas when skipping viewport paint

yay:
## Summary

Apply viewport texture deltas even when a viewport paint is skipped.

## Motivation

Screenshot 2026-06-26 at 10 26 23 AM

When a native viewport skips painting, texture updates such as font-atlas changes can otherwise be dropped along with the skipped frame. That can leave later frames sampling stale or missing texture data after native viewport transitions (also see https://github.com/emilk/egui/pull/8280 for other issues that can show up during such transitions).

This keeps texture state in sync independently from whether that viewport produced draw commands for the current frame.


yay:
I noticed this PR still has no labels, that's why checks are not successful.

crates/eframe/src/native/glow_integration.rs
lucasmerlin, line 703: Textures are set here...

lucasmerlin, line 779: ...and freed here, so I'm not sure how the changes would help? Seems like it would only set / free each change twice?

lucasmerlin (changes requested):
Good catch on wgpu! But I think the glow fix is unnecessary/wrong

crates/eframe/src/native/glow_integration.rs
yay: You're right. Glow already handles texture updates separately from drawing, so this was redundant. I've just removed the glow "fix"

yay:
Thanks for reviewing! Looks like one more label is needed.

PR #8280, opened 2026-06-30
Stabilize native viewport lifecycle and macOS wgpu fullscreen transitions

yay:
What started as an investigation into deferred viewport issues during macOS fullscreen transitions uncovered broader problems in how egui and eframe handle native viewport lifecycle.

This PR builds on https://github.com/emilk/egui/pull/8226 by covering the remaining cases where a pass cannot authoritatively determine the complete viewport set.

## Issues

Extra ghost window sporadically showing up after exiting fullscreen:
Screenshot 2026-06-30 at 2 21 29 PM

Bad viewport scaling during exit fullscreen animation:
Screenshot 2026-06-30 at 2 22 05 PM

Stretched content during enter full screen animation, with stale content in the bottom left corner:
Screenshot 2026-06-30 at 2 22 46 PM

In a multi-viewport macOS app using eframe and either wgpu or glow, native fullscreen transitions for secondary viewports could produce several failures:

- fullscreen viewports could become non-interactive after entering native fullscreen
- entering/exiting fullscreen could show stale or stretched content during AppKit’s Space transition
- deferred viewports could be pruned or lose their viewport callback during partial child viewport passes (see below)
- after skipped viewport paints, the first presented frame could use stale surface/viewport state
- if the first wgpu surface acquisition was skipped before any frame presented, the window could remain blank until another repaint was triggered
- after live resize, disabling transaction presentation before the final Metal drawable was shown could leave the window briefly non-interactive and then adjust its size

Video of egui 0.35.0 deferred viewport example behavior on macOS Sequoia 15.7.7: https://youtu.be/4vWCNajjFts
The video shows it can sometimes take a few tries to reproduce certain issues, but they inevitably happen.

One notable issue in the linked video: after a deferred viewport exits fullscreen, macOS can leave behind what looks like an extra native window. That window is not an egui viewport that the app still owns intentionally. It is stale state caused by eframe pruning or updating deferred viewport state from a partial child viewport pass while AppKit is still driving the fullscreen transition.

This PR avoids that by treating only root passes whose viewport UI ran as authoritative for pruning native viewports, and by preserving existing deferred viewport callbacks when incomplete child passes do not report them.

On macOS empirical testing shows that fullscreen windows are sometimes reported as:
* fullscreen=true
* focused=true
* occluded=true
* run_ui=false

## Fixes

This changes egui and its native integrations to:

- represent viewport output as a partial or complete report
- prune missing native viewports only after a complete root report whose viewport UI ran
- preserve deferred viewport callbacks across incomplete passes
- move the current GL context before removing a Glow viewport surface
- drive Metal transaction presentation from AppKit live-resize and fullscreen state
- wait for the final transaction-presented Metal drawable before leaving live-resize presentation state
- retry when the first wgpu surface acquisition is skipped
- refresh native maximize state at runtime on macOS
- distinguish AppKit live resize from fullscreen transitions, enabling transaction presentation for both while retaining [#8229](https://github.com/emilk/egui/pull/8229)’s temporary latency bump only for live resize, so it now works like this:

<html>
<body>

State | presentsWithTransaction | Latency bump
-- | -- | --
LiveResize | yes | yes
FullscreenTransition | yes | no
Idle | no | no

</body>
</html>

## API migration

`FullOutput::viewport_output` is now a `ViewportOutputReport` so integrations cannot
silently ignore whether a viewport report is authoritative.

- Use `output.viewport_output.entries` for the viewport map.
- Check `output.viewport_output.is_complete` before removing missing native viewports.
- Set `raw_input.viewport_ui_enabled = false` when running a pass without evaluating viewport UI.

## Dependency

Depends on rust-windowing/winit#4610 for:

- `WindowExtMacOS::is_live_resizing()`
- `WindowExtMacOS::is_fullscreen_transition()`
- cached version of `WindowDelegate::is_maximized()`

Depends on gfx-rs/wgpu#9828 for:

- submitted and completed Metal transaction-presentation generations
- a one-shot notification after a target generation is presented

Because egui 0.35 uses wgpu 29 while the upstream wgpu PR targets trunk, this branch temporarily pins an equivalent v29 backport in `yay/wgpu`.

This is draft because the required winit and wgpu APIs have not reached versions compatible with egui 0.35. macOS checks may fail until those dependency updates are available.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2845,9 +2845,8 @@ dependencies = [
 
 [[package]]
 name = "naga"
-version = "29.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "aa2630921705b9b01dcdd0b6864b9562ca3c1951eecd0f0c4f5f04f61e412647"
+version = "29.0.4"
+source = "git+https://github.com/yay/wgpu.git?rev=79168ef4e432f8593e7f8ae23e9e2470b6f68232#79168ef4e432f8593e7f8ae23e9e2470b6f68232"
 dependencies = [
  "arrayvec",
  "bit-set 0.9.1",
@@ -5401,9 +5400,8 @@ checksum = "a751b3277700db47d3e574514de2eced5e54dc8a5436a3bf7a0b248b2cee16f3"
 
 [[package]]
 name = "wgpu"
-version = "29.0.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bb3feacc458f7bee8bc1737149b42b6c731aa461039a4264a67bb6681646b250"
+version = "29.0.4"
+source = "git+https://github.com/yay/wgpu.git?rev=79168ef4e432f8593e7f8ae23e9e2470b6f68232#79168ef4e432f8593e7f8ae23e9e2470b6f68232"
 dependencies = [
  "arrayvec",
  "bitflags 2.13.0",
@@ -5431,9 +5429,8 @@ dependencies = [
 
 [[package]]
 name = "wgpu-core"
-version = "29.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e80ac6cf1895df6342f87d975162108f9d98772a0d74bc404ab7304ac29469e"
+version = "29.0.4"
+source = "git+https://github.com/yay/wgpu.git?rev=79168ef4e432f8593e7f8ae23e9e2470b6f68232#79168ef4e432f8593e7f8ae23e9e2470b6f68232"
 dependencies = [
  "arrayvec",
  "bit-set 0.9.1",
@@ -5465,45 +5462,40 @@ dependencies = [
 
 [[package]]
 name = "wgpu-core-deps-apple"
-version = "29.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43acd053312501689cd92a01a9638d37f3e41a5fd9534875efa8917ee2d11ac0"
+version = "29.0.4"
+source = "git+https://github.com/yay/wgpu.git?rev=79168ef4e432f8593e7f8ae23e9e2470b6f68232#79168ef4e432f8593e7f8ae23e9e2470b6f68232"
 dependencies = [
  "wgpu-hal",
 ]
 
 [[package]]
 name = "wgpu-core-deps-emscripten"
-version = "29.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ef043bf135cc68b6f667c55ff4e345ce2b5924d75bad36a47921b0287ca4b24a"
+version = "29.0.4"
+source = "git+https://github.com/yay/wgpu.git?rev=79168ef4e432f8593e7f8ae23e9e2470b6f68232#79168ef4e432f8593e7f8ae23e9e2470b6f68232"
 dependencies = [
  "wgpu-hal",
 ]
 
 [[package]]
 name = "wgpu-core-deps-wasm"
-version = "29.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2f7b75e72f49035f000dd5262e4126242e92a090a4fd75931ecfe7e60784e6fa"
+version = "29.0.4"
+source = "git+https://github.com/yay/wgpu.git?rev=79168ef4e432f8593e7f8ae23e9e2470b6f68232#79168ef4e432f8593e7f8ae23e9e2470b6f68232"
 dependencies = [
  "wgpu-hal",
 ]
 
 [[package]]
 name = "wgpu-core-deps-windows-linux-android"
-version = "29.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "725d5c006a8c02967b6d93ef04f6537ec4593313e330cfe86d9d3f946eb90f28"
+version = "29.0.4"
+source = "git+https://github.com/yay/wgpu.git?rev=79168ef4e432f8593e7f8ae23e9e2470b6f68232#79168ef4e432f8593e7f8ae23e9e2470b6f68232"
 dependencies = [
  "wgpu-hal",
 ]
 
 [[package]]
 name = "wgpu-hal"
-version = "29.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "89a47aef47636562f3937285af4c44b4b5b404b46577471411cc5313a921da7e"
+version = "29.0.4"
+source = "git+https://github.com/yay/wgpu.git?rev=79168ef4e432f8593e7f8ae23e9e2470b6f68232#79168ef4e432f8593e7f8ae23e9e2470b6f68232"
 dependencies = [
  "android_system_properties",
  "arrayvec",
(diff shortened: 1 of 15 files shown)


smoe:
I asked Fable for a comment:
---
Possibly useful corroboration for the "extra ghost window" you documented here: while working on [#8259](https://github.com/emilk/egui/issues/8259) (native child viewport opened from a fullscreen root, fix in [#8286](https://github.com/emilk/egui/pull/8286)), the baseline repro independently produced the same stale extra window — through a different path than the deferred-viewport fullscreen exit shown in your video.

The trigger there: root in native fullscreen, then a native child viewport is created → macOS starts a Space/Split View negotiation → the child gets tiled into the fullscreen Space (its styleMask actually gains FullScreen without the app ever requesting it), and one stale untitled window is left behind. [AppKit state dump]

So the ghost window seems reachable from any AppKit-driven Space negotiation, not only from fullscreen exits — which supports this PR's diagnosis that eframe state updates during AppKit-driven transitions are the underlying hazard.

[#8286](https://github.com/emilk/egui/pull/8286) avoids that negotiation at the source for the creation-while-fullscreen case… The two changes should be complementary: this PR stabilizes transitions that do happen; [#8286](https://github.com/emilk/egui/pull/8286) prevents one class of them from being triggered accidentally. Happy to re-run my repro on top of this branch if a combined test would help.

yay:
@smoe Thanks. I tested the same flow in my app on top of this PR, but putting the root viewport in fullscreen and then creating a deferred child did not produce a ghost window, abort fullscreen, or trigger visible Split View negotiation.

I did find a separate issue when closing a child while the fullscreen root was in another Space (child window turned black without closing). That turned out to be caused by using root visibility as a proxy for whether root UI had evaluated the viewport tree, so this PR now tracks that explicitly.

So my test does not currently corroborate that the ghost windows in the two reproductions share an underlying cause. 

The PRs still appear complementary at the implementation level, but testing your exact reproduction on top of #8280 would be useful to check if there's an overlap.

smoe:
Fable kindly helped out with this comment again.
---
Ran the exact #8259 reproduction (fullscreen the root via `ViewportCommand::Fullscreen(true)`, then `show_viewport_deferred` a 300×200 child, then dump `NSWindow` state via objc2-app-kit) on three configurations. Environment: **macOS 26.5.2 (Tahoe)**, Apple Silicon, wgpu backend, stock winit 0.30.13.

| configuration | result |
|---|---|
| egui 0.35.0 baseline | child tiled into the fullscreen Space (`styleMask` gains `FullScreen`), ghost window left behind |
| **#8280 (32e0b7f)** | **identical failure signature** — child `fullscreen=true`, ghost window present |
| #8280 + #8286 merged (clean merge, no conflicts) | child floats over the still-fullscreen root as a normal window, **no ghost window** |

So on this OS the two reproductions don't share a cause at the window-creation step: #8280 doesn't change window creation, and the Space negotiation fires exactly as on 0.35.0. Two observations that may explain why your test didn't reproduce it:

1. **OS difference**: you tested on Sequoia 15.7.7; Tahoe's window-tiling behavior is substantially more aggressive about pulling new windows into fullscreen Spaces. The #8259 flow may simply not trigger the negotiation on Sequoia the way it does on 26.x.
2. **Build caveat**: the pushed branch (`32e0b7f`) calls `WindowExtMacOS::is_live_resizing()` / `is_fullscreen_transition()`, which don't exist in stock winit 0.30.13 — it looks like the companion winit patch didn't make it into the PR (its CI shows the same compile failures). For this test I stubbed `native_resize_state` to `Idle`, which disables the present-mode/latency switching but doesn't touch window creation, viewport pruning, or occlusion handling, so it shouldn't affect this particular repro. Happy to re-run unshimmed once the winit side is available.

The good news is symmetrical: the ghost window in *your* reproduction (deferred viewport fullscreen exits) is plausibly fixed by this PR's pruning-authority change and out of #8286's reach, while the creation-time negotiation is prevented by #8286 and out of this PR's reach — and the merged run confirms they compose cleanly. Repro source below.

<details>
<summary>Repro harness (main.rs)</summary>

//! Repro/verification for https://github.com/emilk/egui/issues/8259:
//! fullscreen the root window, then open a native child viewport,
//! and report the resulting AppKit window state.

use std::time::Instant;

use eframe::egui;
use objc2::MainThreadMarker;
use objc2_app_kit::{NSApplication, NSWindowCollectionBehavior, NSWindowStyleMask};

struct TestApp {
    start: Instant,
    requested_fullscreen: bool,
    show_child: bool,
    reported: bool,
}

fn dump_windows(label: &str) {
    let Some(mtm) = MainThreadMarker::new() else {
        println!("RESULT {label}: not on main thread!?");
        return;
    };
    let app = NSApplication::sharedApplication(mtm);
    for window in app.windows().iter() {
        let title = window.title();
        let fullscreen = window
            .styleMask()
            .contains(NSWindowStyleMask::FullScreen);
        let auxiliary = window
            .collectionBehavior()
            .contains(NSWindowCollectionBehavior::FullScreenAuxiliary);
        let on_active_space = window.isOnActiveSpace();
        let visible = window.isVisible();
        println!(
            "RESULT {label}: window={title:?} fullscreen={fullscreen} auxiliary={auxiliary} on_active_space={on_active_space} visible={visible}"
        );
    }
}

impl eframe::App for TestApp {
        let ctx = ui.ctx().clone();
        let elapsed = self.start.elapsed().as_secs_f32();

        let want_fullscreen = std::env::var("SKIP_FULLSCREEN").is_err();
        if !self.requested_fullscreen && want_fullscreen {
            self.requested_fullscreen = true;
            ctx.send_viewport_cmd(egui::ViewportCommand::Fullscreen(true));
        }

        let root_fullscreen = ctx.input(|i| i.viewport().fullscreen.unwrap_or(false));

        if elapsed > 3.0 && (root_fullscreen || !want_fullscreen) && !self.show_child {
            self.show_child = true;
            println!("RESULT: root is fullscreen, opening child viewport");
        }

        if self.show_child {
            ctx.show_viewport_deferred(
                egui::ViewportId::from_hash_of("child"),
                egui::ViewportBuilder::default()
                    .with_title("Child viewport")
                    .with_inner_size([300.0, 200.0]),
                |ui, _class| {
                    egui::CentralPanel::default().show(ui, |ui| {
                        ui.label("I am the child viewport");
                    });
                },
            );
        }

        ui.label(format!(
            "elapsed: {elapsed:.1}s, root_fullscreen: {root_fullscreen}"
        ));

        if elapsed > 7.0 && !self.reported {
            self.reported = true;
            println!("RESULT: root_fullscreen_after_child={root_fullscreen}");
            dump_windows("final");
        }

        if elapsed > 8.0 {
            ctx.send_viewport_cmd(egui::ViewportCommand::Close);
        }

        ctx.request_repaint_after(std::time::Duration::from_millis(100));
    }
}

fn main() -> eframe::Result {
    env_logger::init();
    let renderer = if std::env::var("USE_GLOW").is_ok() {
        eframe::Renderer::Glow
    } else {
        eframe::Renderer::Wgpu
    };
    println!("RESULT: renderer={renderer:?}");
            .with_title("Root window")
            .with_inner_size([600.0, 400.0]),
        renderer,
        ..Default::default()
    };
    eframe::run_native(
        "fullscreen-child-test",
        options,
        Box::new(|_cc| {
            Ok(Box::new(TestApp {
                start: Instant::now(),
                requested_fullscreen: false,
                show_child: false,
                reported: false,
            }))
        }),
    )
}

</details>

yay:
@smoe Thanks for testing it.

PR #8281, opened 2026-07-02
Add option to open and close context menus on press instead of click

mitchmindtree:
## Motivation

* Motivated by <https://github.com/nannou-org/gantz/issues/193>

Context menus in egui require a full click (press + release) to open and to close. Native context menus on most platforms respond already on press: right-press opens the menu, and pressing anywhere outside dismisses it. This PR adds that behavior as an **opt-in** - default behavior is unchanged.

## Usage

```rust
ctx.all_styles_mut(|style| style.interaction.context_menu_opens_on_press = true);
```

With the flag enabled, `Popup::context_menu` / `Response::context_menu`:

- open on secondary button *press* (long-touch on touch screens still works),
- close as soon as a pointer button is pressed outside the menu,
- still fire menu items on click (release) as before, which also still closes the menu.

## What's added

- `Style::interaction.context_menu_opens_on_press: bool` (default `false`), also exposed in the style settings UI.
- `PopupCloseBehavior::CloseOnPressOutside`: closes on any press outside the popup, or a *primary* click inside it. Used by context menus when the flag is enabled. Can also be set on any `Popup` via `close_behavior`.
- `Response::pressed_by`, `Response::secondary_pressed` and `Response::pressed_elsewhere` - press-based counterparts of the existing `clicked_by` / `secondary_clicked` / `clicked_elsewhere`.

## Testing

- In the demo app, open the **Popups** demo and tick `style.interaction.context_menu_opens_on_press` (also togglable under Settings -> Style -> Interaction), then right-press the demo button. Compare with the flag off.
- New kittest suite `crates/egui_kittest/tests/context_menu.rs`: 9 tests covering open-on-press, close-on-press-outside, item activation, submenus, and one locking in the default (click-based) behavior with the flag off.

## Design notes

- Inside the popup, only *primary* clicks close it. The popup opens exactly at the press position (`PointerFixed` anchor), so the release of the opening right-press lands on the popup's corner. An any-click rule would close the menu again on every quick right-click.
- The open check is `secondary_pressed() || secondary_clicked()`: the fallback keeps menus opening when press and release arrive batched in the same frame (fast clicks at low frame rates, `egui_kittest` clicks). A side effect is that right-clicking the anchor again while the menu is open now moves the menu to the new position (with the flag off, it closes, as before).
- `pressed_elsewhere` mirrors `clicked_elsewhere` and uses `interact_pos()` rather than `press_origin()`, since the latter is cleared when press and release arrive in the same frame.
- `CloseOnPressOutside` is documented as unsuitable for popups toggled by a button click (e.g. `Popup::menu`): the press would close the popup and the completing click would toggle it open again.

## Breaking changes

No behavioral changes with the flag off (default). Compile-level:

- New variant on `PopupCloseBehavior` breaks exhaustive matches.
- New field on `style::Interaction` breaks struct literals that don't use `..Default::default()`.

### DIFF
--- a/crates/egui/src/containers/menu.rs
+++ b/crates/egui/src/containers/menu.rs
@@ -66,7 +66,8 @@ pub struct MenuConfig {
     /// Is this a menu bar?
     bar: bool,
 
-    /// If the user clicks, should we close the menu?
+    /// When should pointer presses/clicks close the menu?
+    /// See [`PopupCloseBehavior`].
     pub close_behavior: PopupCloseBehavior,
 
     /// Override the menu style.
@@ -93,7 +94,8 @@ impl MenuConfig {
         Self::default()
     }
 
-    /// If the user clicks, should we close the menu?
+    /// When should pointer presses/clicks close the menu?
+    /// See [`PopupCloseBehavior`].
     #[inline]
     pub fn close_behavior(mut self, close_behavior: PopupCloseBehavior) -> Self {
         self.close_behavior = close_behavior;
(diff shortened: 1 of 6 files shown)


mitchmindtree:
Looks like CI failure is unrelated - the `cargo-deny` advisories check fails on `RUSTSEC-2026-0192`, a freshly published "unmaintained" advisory for `ttf-parser` `0.25.0`. We might have to address that separately.

PR #8282, opened 2026-07-02
eframe: hand focus to a launched process with an activation token

Le-Syl21:
A Wayland compositor refuses focus to a window it did not see the user ask for. A launcher hands its own focus over instead: it asks the compositor for an [xdg-activation](https://wayland.app/protocols/xdg-activation-v1) token and passes it to the child process in `XDG_ACTIVATION_TOKEN`. On X11 the same role is played by a startup-notification id, in `DESKTOP_STARTUP_ID`.

winit implements both — `WindowExtStartupNotify::request_activation_token()` and `WindowEvent::ActivationTokenDone` — but the reply arrives as a window event that eframe consumes, so an app has no way to see it.

Two methods on `Frame`:

```rust
frame.request_activation_token();          // no-op off Linux
// ...one or more frames later:
if let Some(token) = frame.take_activation_token() {
    Command::new("some-app").env("XDG_ACTIVATION_TOKEN", token).spawn()?;
}
```

Both native backends intercept `ActivationTokenDone`, hold the token on `Frame`, and **request a repaint**: a token is single-use and short-lived, so an idle app must not have to wait for an unrelated event to notice it. Nothing is delivered on platforms without an activation protocol, so an app should give up after a deadline of its own rather than wait forever.

Nothing to add on the app side beyond these two calls — no winit dependency, no `cfg` of its own.

### What this replaces

The first version of this PR put a `ViewportCommand::RequestActivationToken` and an `Event::ActivationTokenReceived` in `egui` itself. @lucasmerlin pointed out this belongs in eframe rather than in the core, which is right: `egui` has no business knowing what xdg-activation is, and eframe is already where the native-specific surface lives.

### Checked

- both backends, glow and wgpu
- both arms of the `cfg_select!`: the Linux one as-is, the fallback by temporarily pointing the predicate at another `target_os` (winit won't build on Linux without `wayland` or `x11`)
- the real case this comes from: a launcher on GNOME/mutter spawning a fullscreen app, which opens behind the launcher without a token and in front with one

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -502,13 +502,31 @@ impl State {
                 consumed: false,
             },
 
+            // Activation token delivered by winit in response to
+            // `ViewportCommand::RequestActivationToken`. Forward the raw
+            // string as an `Event::ActivationTokenReceived` so the app can
+            // pass it to a child process via `XDG_ACTIVATION_TOKEN`
+            // (Wayland xdg-activation-v1) or `DESKTOP_STARTUP_ID` (X11).
+            WindowEvent::ActivationTokenDone { token, .. } => {
+                self.egui_input
+                    .events
+                    .push(egui::Event::ActivationTokenReceived {
+                        viewport_id: self.viewport_id,
+                        token: token.clone().into_raw(),
+                    });
+                EventResponse {
+                    repaint: true,
+                    consumed: false,
+                }
+            }
+
             // Things we completely ignore:
-            WindowEvent::ActivationTokenDone { .. }
-            | WindowEvent::AxisMotion { .. }
-            | WindowEvent::DoubleTapGesture { .. } => EventResponse {
-                repaint: false,
-                consumed: false,
-            },
+            WindowEvent::AxisMotion { .. } | WindowEvent::DoubleTapGesture { .. } => {
+                EventResponse {
+                    repaint: false,
+                    consumed: false,
+                }
+            }
 
             WindowEvent::PinchGesture { delta, .. } => {
                 // Positive delta values indicate magnification (zooming in).
@@ -1926,6 +1944,21 @@ fn process_viewport_command(
         }
         ViewportCommand::RequestPaste => {
             actions_requested.push(ActionRequested::Paste);
+        }
+        ViewportCommand::RequestActivationToken => {
+            #[cfg(all(any(feature = "wayland", feature = "x11"), target_os = "linux"))]
+            {
+                use winit::platform::startup_notify::WindowExtStartupNotify as _;
+                if let Err(err) = window.request_activation_token() {
+                    log::debug!("RequestActivationToken failed: {err}");
+                }
+            }
+            // Silent no-op on platforms without an activation protocol —
+            // `Event::ActivationTokenReceived` is simply never emitted.
+            #[cfg(not(all(any(feature = "wayland", feature = "x11"), target_os = "linux")))]
+            {
+                let _ = window;
+            }
         }
     }
 }
--- a/crates/eframe/src/epi.rs
+++ b/crates/eframe/src/epi.rs
@@ -701,6 +701,11 @@ pub struct Frame {
     /// Raw platform display handle for window
     #[cfg(not(target_arch = "wasm32"))]
     pub(crate) raw_display_handle: Result<RawDisplayHandle, HandleError>,
+
+    /// The activation token the windowing system last handed us, waiting to be
+    /// taken by the app. See [`Frame::request_activation_token`].
+    #[cfg(not(target_arch = "wasm32"))]
+    pub(crate) activation_token: Option<String>,
 }
 
 // Implementing `Clone` would violate the guarantees of `HasWindowHandle` and `HasDisplayHandle`.
@@ -741,6 +746,8 @@ impl Frame {
             raw_window_handle: Err(HandleError::NotSupported),
             #[cfg(not(target_arch = "wasm32"))]
             window: None,
+            #[cfg(not(target_arch = "wasm32"))]
+            activation_token: None,
             storage: None,
             #[cfg(feature = "wgpu_no_default_features")]
             wgpu_render_state: None,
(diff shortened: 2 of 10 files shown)


Le-Syl21:
Note on CI:

- The `label` check needs a maintainer to apply labels (I don't have triage rights) — `egui-winit` + `egui` would fit this change.
- The `cargo-deny` failures flag RUSTSEC-2026-0194 / RUSTSEC-2026-0195 (quick-xml), advisories published after the last green main run, against dependencies already on main — this PR adds no dependencies.

Format + check, all test suites, wasm and platform checks pass.

Le-Syl21:
Rebased-check after the 0.36.0 release: still merges cleanly against main, no conflicts. Wayland focus hand-off remains impossible without this — happy to adjust anything (naming, docs, winit bounds) to fit the 0.37 cycle. The `label` check still needs a maintainer to apply `egui-winit` + `egui`.

crates/egui-winit/src/lib.rs
```diff
@@ -1947,6 +1965,21 @@ fn process_viewport_command(
...
+            #[cfg(all(any(feature = "wayland", feature = "x11"), target_os = "linux"))]
+            {
+                use winit::platform::startup_notify::WindowExtStartupNotify as _;
+                if let Err(err) = window.request_activation_token() {
+                    log::debug!("RequestActivationToken failed: {err}");
+                }
+            }
+            // Silent no-op on platforms without an activation protocol —
+            // `Event::ActivationTokenReceived` is simply never emitted.
+            #[cfg(not(all(any(feature = "wayland", feature = "x11"), target_os = "linux")))]
```
emilk: use `cfg_select!`

Le-Syl21: Done. Also rebased on main while I was here.

I checked both arms compile: the Linux one with the default features, and the fallback by temporarily pointing the predicate at another target_os, since winit won't build on Linux without wayland or x11.

Le-Syl21:
Heads-up unrelated to the review: CI on this PR is red, but no job ran — they all fail after a couple of seconds with the annotation "The job was not started because your account is locked due to a billing issue."

It doesn't look specific to this PR. Every `pull_request` run on the repo has failed the same way since about 14:10 UTC today, across more than one contributor's fork, while `push` runs on `main` still pass — which may be why it hasn't shown up on your side.

Nothing needed from me here; flagging it in case it's news.

Le-Syl21:
Status check after two months, rather than a nudge: still merges cleanly on today's `main`, and the review comments are addressed (`cfg_select`).

One thing I can't fix from my side — the only red check is `label`, which requires one of the repo's labels, and this PR has none. #8405 and #8406 both ended up labelled and their runs are green. An `egui-winit` label here should be enough to clear it.

lucasmerlin (commented):
It's annoying that wayland makes us deal with this. But I guess there's no way around it. 

But I wonder if it should rather live in eframe? 
The request can already be done without any change via `frame.winit_window()`. 
Then we only need some way to pass the token to the application, e.g. by storing it on the eframe `Frame` struct when it arrives and adding a `take_activation_token` to which reads it and sets it to None again.

Le-Syl21:
Thanks @lucasmerlin — the check is still red though: `native-linux` isn't one of the labels that workflow accepts.

> Label error. Requires at least 1 of: CI, dependencies, docs and examples, ecolor, eframe, egui_extras, egui_glow, egui_inspection, egui_kittest, egui_system_fonts, egui-wgpu, egui-winit, egui, emath, epaint, epaint_default_fonts, exclude from changelog, typo. Found: native-linux

`egui-winit` would fit — that and `crates/egui` are what this touches.

Le-Syl21:
That's a fair instinct — `egui` shouldn't have to know what xdg-activation is, and eframe is already where the native-specific things live (`winit_window()`, `raw_window_handle`). Happy to move it there.

One wrinkle with doing the request from the app side through `frame.winit_window()`: `request_activation_token()` isn't a method on `Window`, it comes from `winit::platform::startup_notify::WindowExtStartupNotify` — and eframe doesn't re-export winit. So the app has to add winit as a direct dependency *and* pin the exact version eframe uses; otherwise it's a second copy of the crate, `Arc<winit::window::Window>` is a different type, and the trait doesn't apply. On top of that the app ends up writing the same `cfg` dance that currently sits in one place in egui-winit.

So how about **two** methods on `Frame` rather than one?

- `Frame::request_activation_token()` — no-op off Linux, and no winit dependency for the app
- `Frame::take_activation_token() -> Option<String>` — exactly as you described

Two details worth deciding on while it moves:

- **eframe should request a repaint when the token arrives.** An activation token is single-use and short-lived; an idle app that isn't repainting would otherwise sit on it until something unrelated wakes it up. The current `Event` path gets that for free (`repaint: true`).
- `Frame::winit_window()` is documented as the *active* viewport's window, so a token stored on `Frame` is implicitly tied to whichever viewport is being rendered when it's read, rather than the one that asked. Fine for the single-window case — worth choosing rather than discovering later.

I'll rework it that way unless you'd rather shape it yourself.

And while I have your attention 🙂 — if you're doing a pass over the Linux/winit corner, there are two more of mine sitting there:

- **#8405** — *eframe: route `DeviceEvent::MouseMotion` to the root viewport*. Four weeks without a look; `device_event` in both integrations still routes through `focused_viewport` on today's `main`, so the bug is unchanged. Under 40 lines, no API change.
- **#8406** — you approved it this morning, thanks! That one just needs someone to press merge.

lucasmerlin:
> So how about **two** methods on `Frame` rather than one?
> 
>     * `Frame::request_activation_token()` — no-op off Linux, and no winit dependency for the app
> 
>     * `Frame::take_activation_token() -> Option<String>` — exactly as you described

Makes sense

Le-Syl21:
Reworked as agreed — pushed, and the description above is rewritten to match.

`egui` is untouched now: no `ViewportCommand`, no `Event`. Everything lives on `Frame`, as `request_activation_token()` and `take_activation_token()`, with both native backends intercepting `ActivationTokenDone` and requesting a repaint when the token lands.

Worth a label from the accepted set while you're here — the `label` check is the only red one, and `native-linux` isn't in the list it accepts. `eframe` would be the right one now that the change moved there.

Le-Syl21:
CI was red on my first push, on something worth flagging rather than just fixing:

```
error: large size difference between variants
  --> crates/egui_kittest/src/app_kind.rs:18:1
   |   Ui(AppKindUi<'a>),            -- the second-largest variant contains at least 16 bytes
   |   Eframe(AppKindEframe<State>), -- the largest variant contains at least 232 bytes
```

`AppKind::Eframe` holds an `eframe::Frame` by value while the other two variants hold a closure, so the enum sat about 8 bytes below the `large_enum_variant` threshold. One `Option<String>` on `Frame` tips it over — **any** field added to `Frame` would have. Boxed the variant as clippy suggests, which is in this push.

Rechecked locally with the same commands the workflow runs: `cargo fmt --check`, `scripts/lint.py`, and each of the clippy invocations (`--all-features --all-targets`, the `--no-default-features` ones, and the two eframe/x11 variants), plus `cargo doc`. All clean.

The `label` check is still the only red one — `native-linux` isn't in the set it accepts, and `eframe` would fit now.

Le-Syl21:
CI is green — `Format + check`, `Check Windows` and the rest all pass on this push.

The one job still red is `label`, which fails in 3 seconds because the PR carries only `native-linux` and that isn't in the set the workflow accepts. That one needs a maintainer; `eframe` fits now that the change lives there.

Ready for review otherwise.

PR #8286, opened 2026-07-03
Fix native child viewports aborting fullscreen on macOS

### LINKED ISSUE #8259: macOS: native child viewport flickers/aborts when opened from fullscreen or maximized root window
### Description

After #8203, normal macOS resize behavior is much improved for our eframe/wgpu application. However, a separate native child viewport problem remains reproducible when the root eframe window is fullscreen or maximized on macOS.

When the root window is fullscreen/maximized and the app opens an auxiliary native child viewport, macOS appears to enter a Split View/fullscreen negotiation: the child viewport flickers, the fullscreen transition may abort, and the child/root windows can visibly fight for presentation. If the root window is kept as an ordinary non-fullscreen window, the same native child viewport path is usable.

### What changed compared with #8043

The original resize-latency/choppy-resize problem seems addressed by #8203. This report is about the remaining fullscreen/maximized-root + native-child-viewport behavior.

### Reproduction pattern

1. Run an eframe/wgpu app on macOS.
2. Put the root eframe window into fullscreen or maximized state.
3. Open a second native viewport using `Context::show_viewport_immediate(...)` or `Context::show_viewport_deferred(...)`.
4. Observe flicker / aborted fullscreen or Split View-like behavior.

### Observed

- native child viewport flickers or briefly appears/disappears
- fullscreen/Split View transition appears to be negotiated by macOS and then aborted
- no application panic or terminal error is emitted
- using embedded viewports avoids the problem but changes the UX

### Expected

Opening an auxiliary native viewport from a fullscreen/maximized root should either work predictably or expose a recommended eframe-side pattern for leaving fullscreen/maximized state before spawning the child viewport.

### Workaround under consideration

Detect root `ViewportInfo.fullscreen` / `ViewportInfo.maximized`, send `ViewportCommand::Fullscreen(false)` / `ViewportCommand::Maximized(false)`, and defer child viewport creation until the root has settled back into a normal window.

### Related

- #8043
- #8203
- #8225 may be related

smoe:
* Closes <https://github.com/emilk/egui/issues/8259>

## Problem

Opening a native child viewport (`show_viewport_deferred` / `show_viewport_immediate`) while the root window is in macOS native fullscreen makes AppKit renegotiate the active Space: the child flickers, can get pulled into Split View tiling — ending up unintentionally *fullscreen itself* — and the root's fullscreen state can be aborted. Reproduced with both the wgpu and glow backends; window state captured via AppKit before the fix:

```
window="Root window"     fullscreen=true   on_active_space=true
window="Child viewport"  fullscreen=true   on_active_space=true   <- pulled into Split View
+ one stale ghost window left behind
```

## Fix

AppKit's mechanism for windows accompanying a fullscreen window (palettes, inspectors) is `NSWindowCollectionBehaviorFullScreenAuxiliary`. winit 0.30 doesn't expose it, and it must be set **before** the window is first ordered on screen — but winit shows windows during creation. So `egui-winit` now creates affected windows hidden, sets the collection behavior via `objc2-app-kit`, and then orders them front (respecting the builder's `visible`/`active`). Same pattern as the existing iOS `safe_area.rs` polyfill; I have proposed the API upstream (rust-windowing/winit#4614) so most of the new `macos.rs` module can be deleted when egui moves to winit 0.31.

New API: `ViewportBuilder::with_fullscreen_auxiliary` (macOS only):

* `true`: always mark the window as fullscreen-auxiliary,
* `false`: never,
* unset (default): auto — mark it if and only if the window is created while the app has a fullscreen window on the active Space, and the viewport doesn't itself request fullscreen. So child viewports "just work" from a fullscreen root, with no behavior change in any other situation.

Trade-off: an auxiliary window cannot itself enter native fullscreen. `ViewportCommand::Fullscreen(true)` / `SetMonitor` therefore clear the flag before toggling, so runtime fullscreen requests on child viewports keep working.

After the fix (same repro):

```
window="Root window"     fullscreen=true   on_active_space=true   <- stays fullscreen
window="Child viewport"  fullscreen=false  on_active_space=true   <- floats on the fullscreen Space
```

## Testing

* Test app driving the full sequence (fullscreen root → open deferred child → inspect `NSWindow` state) on macOS 15 / Darwin 25.5.0, run against both `wgpu` and `glow` backends: root stays fullscreen, child shows on the same Space with the auxiliary bit set.
* Regression check with a windowed root: predicate is false, the child window is created visible by winit exactly as before (no auxiliary bit, no hidden-then-show).
* `cargo test -p egui -p egui-winit`, clippy (debug + release, all features), `cargo doc -D warnings`, `lint.py`, `cargo fmt` all pass.

## Notes

* The issue also mentions a *maximized* root: a zoomed window involves no Space negotiation, and I could not reproduce any misbehavior there — I suspect those symptoms belong to the viewport transition issues addressed in #8280. This PR is orthogonal to (and compatible with) #8280; incidentally, the "extra ghost window" documented there also shows up in this issue's baseline repro, which supports that connection.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1348,6 +1348,7 @@ dependencies = [
  "egui",
  "log",
  "objc2 0.6.4",
+ "objc2-app-kit 0.3.2",
  "objc2-foundation 0.3.2",
  "objc2-ui-kit 0.3.2",
  "profiling",
--- a/crates/egui-winit/Cargo.toml
+++ b/crates/egui-winit/Cargo.toml
@@ -79,6 +79,16 @@ document-features = { workspace = true, optional = true }
 serde = { workspace = true, optional = true }
 webbrowser = { workspace = true, optional = true }
 
+[target.'cfg(target_os = "macos")'.dependencies]
+objc2.workspace = true
+objc2-app-kit = { workspace = true, default-features = false, features = [
+  "std",
+  "NSApplication",
+  "NSResponder",
+  "NSView",
+  "NSWindow",
+] }
+
 [target.'cfg(target_os = "ios")'.dependencies]
 objc2.workspace = true
 objc2-foundation = { workspace = true, features = ["std", "NSThread"] }
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -21,6 +21,8 @@ use egui::{Pos2, Rect, Theme, Vec2, ViewportBuilder, ViewportCommand, ViewportId
 pub use winit;
 
 pub mod clipboard;
+#[cfg(target_os = "macos")]
+mod macos;
 mod safe_area;
 mod window_settings;
 
@@ -1826,10 +1828,16 @@ fn process_viewport_command(
             info.maximized = Some(v);
         }
         ViewportCommand::Fullscreen(v) => {
+            #[cfg(target_os = "macos")]
+            if v {
+                macos::clear_fullscreen_auxiliary(window);
+            }
             window.set_fullscreen(v.then_some(winit::window::Fullscreen::Borderless(None)));
         }
         ViewportCommand::SetMonitor(idx) => {
             if let Some(monitor) = window.available_monitors().nth(idx) {
+                #[cfg(target_os = "macos")]
+                macos::clear_fullscreen_auxiliary(window);
                 window.set_fullscreen(Some(winit::window::Fullscreen::Borderless(Some(monitor))));
             } else {
                 log::warn!(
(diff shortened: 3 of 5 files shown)


yay:
Related: This PR handles native child-window creation on an existing fullscreen Space, while #8280 addresses viewport lifecycle and rendering failures during native fullscreen transitions.

smoe:
@emilk - please kindly initiate the CI for this PR.

smoe:
Our fullscreen-auxiliary extension has now landed in winit master via https://github.com/rust-windowing/winit/pull/4614.

Unless there are objections, rather than retaining the direct `objc2-app-kit` workaround in this PR, we propose waiting for winit 0.31 to be released and adopted by egui. We would then update (and thus minimise) this PR to use the official winit API.

PR #8289, opened 2026-07-04
Upgrade wgpu to v30

akx:
See https://github.com/gfx-rs/wgpu/releases/tag/v30.0.0 for release notes.

`cargo run -p egui_demo_app --no-default-features --features wgpu,persistence` works fine on my macOS box with this.

* No relevant issue to close
* [x] I have followed the instructions in the PR template

PR #8290, opened 2026-07-04
Add ttf-parser and quick-xml advisory ignores to deny.toml

akx:

Deny CI in #8289 is red due to:

* ttf-parser being unmaintained (via fontdb, rustybuzz etc.)
* DoS vulnerabilities in quick-xml (via wayland-scanner, plist, etc.).

These need bumps/replacements in those upstream crates.


emilk:
thanks for the PR, but looks like this was already merged elsewhere

PR #8296, opened 2026-07-07
Web: don't scroll host page when text agent or canvas grabs focus

### LINKED ISSUE #8295: Web: text agent focus scrolls the host page to the top when the app is embedded
### Describe the bug

When an eframe app is embedded in a normal, scrollable web page (our case: the Rerun web viewer embedded mid-page in a course site), the host page jumps/scrolls to the top whenever the app boots or grabs keyboard focus. If the user has scrolled down to where the canvas is, starting the app yanks them back up.

### To Reproduce

1. Embed an eframe app (e.g. via Rerun's `WebViewer`, eframe 0.35) in a page long enough to scroll, with the canvas placed well below the fold.
2. Scroll down to the canvas and start/interact with the app.
3. The page scrolls back to the top.

### Expected behavior

Focusing the app (or it focusing itself on boot) should not scroll the host page.

### Root cause

The text agent (`crates/eframe/src/web/text_agent.rs`) is a hidden 1×1 px `<input>` that is:

- positioned `absolute; top: 0; left: 0` and appended to `document.body` (lines 38–49),
- marked `input.set_autofocus(true)` (line 25) — the browser honors `autofocus` on insertion and scrolls the element into view, and
- focused via plain `HtmlElement::focus()` without `preventScroll` — in `TextAgent::focus()` (line 224) and in the Gboard workaround that does `blur(); focus()` on every non-composing input event (lines 69–70).

Focusing an element parked at (0,0) of `<body>` makes the browser scroll it into view — i.e. scroll the page to the very top. In a full-viewport app this is invisible, so it only bites people embedding eframe in larger pages.

### Suggested fix

- Use `focus_with_options(&FocusOptions::new().prevent_scroll(true))` (`web_sys` supports it, [`FocusOptions.preventScroll`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#preventscroll) is supported by all evergreen browsers) everywhere the text agent and canvas are focused.
- Drop `set_autofocus(true)` and instead focus explicitly (with `preventScroll`) after insertion — the `autofocus` insertion path is browser-internal and can't be reached by any `focus()` option.

### Workaround

For anyone else hitting this, patching before the app boots works:

```js
const original = HTMLElement.prototype.focus;
HTMLElement.prototype.focus = function (opts) {
  original.call(this, { ...opts, preventScroll: true });
};
Object.defineProperty(HTMLElement.prototype, "autofocus", {
  get() { return false; },
  set() {},
});
```

…but that is global for the whole page, which is a big hammer for something eframe could do precisely.

Happy to send a PR if the `focus_with_options` approach sounds right.

---

Authored by @gavrelina together with Claude (Claude Code).

emilk:
* Closes <https://github.com/emilk/egui/issues/8295>

When an eframe app is embedded in a scrollable host page, the host page jumped to the top whenever the app booted or grabbed keyboard focus. Cause: the hidden text-agent `<input>` sat at (0,0) of `<body>` with the `autofocus` attribute, and all focus calls (text agent and canvas) used plain `focus()`, so the browser scrolled the focused element into view — i.e. to the top of the page.

Changes:

* All focus calls in `eframe` web (text agent, canvas, Gboard blur/refocus workaround) now go through a new `focus_without_scroll()` helper that uses `focus_with_options` with `preventScroll: true` (supported by all evergreen browsers).
* The `autofocus` attribute is replaced with an explicit focus call after DOM insertion — the browser-internal autofocus path always scrolls the element into view and cannot be prevented by any focus option. Boot-focus behavior is preserved.
* The text-agent input is initially parked at the canvas origin instead of (0,0) of the page, so any residual scroll-into-view would target the canvas rather than the top of the host page. (`move_to` keeps repositioning it for IME as before.)

### DIFF
--- a/crates/eframe/Cargo.toml
+++ b/crates/eframe/Cargo.toml
@@ -232,6 +232,7 @@ web-sys = { workspace = true, features = [
   "File",
   "FileList",
   "FocusEvent",
+  "FocusOptions",
   "HtmlCanvasElement",
   "HtmlElement",
   "HtmlInputElement",
--- a/crates/eframe/src/web/app_runner.rs
+++ b/crates/eframe/src/web/app_runner.rs
@@ -401,7 +401,7 @@ impl AppRunner {
             } else {
                 // We are not editing text - give the focus to the canvas.
                 self.text_agent.blur();
-                self.canvas().focus().ok();
+                super::focus_without_scroll(self.canvas()).ok();
             }
         }
 
--- a/crates/eframe/src/web/events.rs
+++ b/crates/eframe/src/web/events.rs
@@ -599,7 +599,7 @@ fn install_pointerup(runner_ref: &WebRunner, target: &EventTarget) -> Result<(),
                 // not working when focusing on a text field in an egui app.
                 // This attempts to fix that by forcing the focus on any
                 // click on the canvas.
-                runner.canvas().focus().ok();
+                super::focus_without_scroll(runner.canvas()).ok();
 
                 // In Safari we are only allowed to do certain things
                 // (like playing audio, start a download, etc)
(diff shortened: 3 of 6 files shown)


lucasmerlin (approved):
Nice catch!

PR #8297, opened 2026-07-07
Web: anchor the text agent to the canvas

emilk:
* Follow-up to #8296, part of <https://github.com/emilk/egui/issues/8295>

Stacked on #8296 (base branch), kept as a separate PR so it can be reverted independently.

The hidden text-agent `<input>` is now inserted as a *sibling of the canvas* (instead of appended to `document.body`) and positioned with `offsetLeft`/`offsetTop` instead of `getBoundingClientRect`. Since the input and the canvas share the same containing block, the input stays anchored to the canvas top-left corner no matter how the page is scrolled or how the canvas is embedded.

Consequences:

* Fixes the IME popup position when the host page is scrolled — `move_to` previously wrote *viewport* coordinates from `getBoundingClientRect` into document-absolute `left`/`top`.
* Subsumes the mobile Safari virtual-keyboard workaround (it replaced the flapping `getBoundingClientRect` y with `offsetTop`, which is now used everywhere), so `is_mobile_safari()` is removed.
* Removes the special-casing of document vs shadow DOM roots — sibling insertion works uniformly in both.
* The input is `position: absolute`, so it does not participate in flex/grid layout of the canvas' parent and causes no layout shift. Caveat: host CSS selectors like `div > canvas:only-child` would no longer match.

Verified with `cargo clippy -p eframe --target wasm32-unknown-unknown --all-features` and `cargo fmt --all`.

### DIFF
--- a/crates/eframe/src/web/text_agent.rs
+++ b/crates/eframe/src/web/text_agent.rs
@@ -4,7 +4,6 @@
 use std::cell::Cell;
 
 use wasm_bindgen::prelude::*;
-use web_sys::Document;
 
 use super::{AppRunner, WebRunner};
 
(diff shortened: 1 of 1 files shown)


lucasmerlin (approved):
Makes sense to me

PR #8299, opened 2026-07-07
Sync window theme with egui theme

lucasmerlin:
Adds a new option to sync the window theme with the egui theme, enabled by default. Works across viewports.

 

https://github.com/user-attachments/assets/513c2318-cd6e-4e2b-805d-04002c375a10


emilk (approved):
nice idea!

PR #8302, opened 2026-07-07
eframe: apply ViewportBuilder::with_monitor on the glow backend

Le-Syl21:
## Problem

`ViewportBuilder::with_monitor(idx)` is a **no-op on the glow backend**. A window
built with `.with_monitor(...)` opens on the default monitor, not fullscreen —
while the exact same code works on wgpu.

The monitor → `Fullscreen::Borderless(Some(monitor))` resolution only lives in
`egui_winit::create_window`, which the **wgpu** backend uses. The **glow**
backend builds its windows straight from `egui_winit::create_winit_window_attributes`
at two sites in `glow_integration.rs` — the initial window and per-viewport
deferred windows — so the monitor is never applied there.

## Repro

A multi-viewport app on the glow renderer with any viewport built via
`ViewportBuilder::default().with_monitor(1)` opens that window on monitor 0,
un-fullscreen. Switching to the wgpu renderer places it correctly. Seen on a
3-monitor GNOME/Wayland setup where the main window and secondary viewports all
need to land borderless-fullscreen on specific outputs.

## Fix

Extract the resolution into a public `apply_monitor_to_window_attributes` in
`egui-winit` and call it from `create_window` (unchanged behaviour) plus both
glow window-creation sites, so `with_monitor` behaves the same on glow and wgpu.

No public API change beyond the new helper; no behaviour change on wgpu.

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -1094,9 +1094,10 @@ impl GlutinWindowContext {
             //
             // The justification for FallbackEgl over PreferEgl is at https://github.com/emilk/egui/pull/2526#issuecomment-1400229576 .
             .with_preference(glutin_winit::ApiPreference::FallbackEgl)
-            .with_window_attributes(Some(egui_winit::create_winit_window_attributes(
-                egui_ctx,
-                viewport_builder.clone(),
+            .with_window_attributes(Some(egui_winit::apply_monitor_to_window_attributes(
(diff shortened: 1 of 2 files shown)


Le-Syl21:
Friendly post-0.36 ping: `ViewportBuilder::with_monitor` (#8140) is now in the public API, but the glow backend still silently ignores it — this PR wires it up. Merges cleanly against current main; happy to adjust for 0.37.

emilk (approved):

PR #8306, opened 2026-07-09
Fix `Frame::set_wgpu_surface_config` having no effect

amfaber:
`Frame::set_wgpu_surface_config` writes into the `RenderState` clone stored in the frame, but the painter consults its *own* separate clone when applying runtime surface changes in `Painter::paint_and_update_textures`. Unlike the Arc-backed `device`, `queue` and `renderer`, `RenderState::surface_config` is a plain value, so the two clones diverge and the app-requested config was never picked up — the surface was never reconfigured.

Add `Painter::set_surface_config` and call it once per frame from the wgpu integration to push the frame's config into the painter's copy before painting.

### DIFF
--- a/crates/eframe/src/native/wgpu_integration.rs
+++ b/crates/eframe/src/native/wgpu_integration.rs
@@ -691,6 +691,14 @@ impl WgpuWinitRunning<'_> {
             ..
         } = &mut *shared_mut;
 
+        // Propagate any surface config the app set this frame via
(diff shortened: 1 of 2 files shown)


Wumpf (changes requested):
thanks for looking into that. The fix makes me a bit nervous though - seems more like an architectural issue here. Why are there two different surface configurations in the first place. Also why are we so sure this change is picked up correctly? As you wrote in the description, any change in surface configuration needs to trigger a reconfigure

PR #8307, opened 2026-07-09
Remove dependency `home`

Calbabreaker:
Replaces `home::home_dir` with `std::env::home_dir` as these functions do the exact same thing

Removes dependency home from eframe by replacing `home::home_dir` with `std::env::home_dir`. `home` was probably originally used since `std::env::home_dir` was once deprecated because of a bug. Post Rust version 1.87 this has been fixed and now these two functions do exactly the same thing.


emilk (approved):

emilk:
nice!

PR #8315, opened 2026-07-14
Rerun `sizing_pass` when reopening popup

### LINKED ISSUE #8115: Popup width doesn't change when new wider items are added
**Describe the bug**
Image

**Expected behavior**
The popup width expands just enough to accommodate the new widest item.

**Desktop**
 - OS: 15.7.5 (24G624)
 - eframe: 0.34.1

**Additional context**
Example project: [egui-menu-popup-width-repro.zip](https://github.com/user-attachments/files/26854624/egui-menu-popup-width-repro.zip)

yay:
* Closes <https://github.com/emilk/egui/issues/8115>

## Summary

- Recalculate a menu popup's cached `Area` size when it reopens.
- Preserve cached sizing for continuously open menus and leave tooltips and general popups unchanged.
- Add a headless regression test covering a wider item added while the menu is closed.

## Root cause

`Area` keeps its cached size after a menu closes. When that menu reopened with wider content, the
cached width constrained the new item and caused it to wrap instead of allowing the popup to grow.

The fix requests the same invisible sizing pass used for a first-open `Area` whenever a menu was not
open during the previous frame.

## User impact

Menus now expand to fit newly added wider items after reopening. Existing wrapping, explicit-width,
alignment, screen-constraining, and continuously open menu behavior remain unchanged.

## Validation

- `cargo test -p egui_kittest --test menu`
- `cargo check -p egui`
- `cargo fmt --all -- --check`

### DIFF
--- a/crates/egui/src/containers/popup.rs
+++ b/crates/egui/src/containers/popup.rs
@@ -571,6 +571,7 @@ impl<'a> Popup<'a> {
             .fixed_pos(anchor)
             .sense(sense)
             .layout(layout)
+            .sizing_pass(kind == PopupKind::Menu && !was_open_last_frame)
             .info(info.unwrap_or_else(|| {
                 UiStackInfo::new(kind.into()).with_tag_value(
                     MenuConfig::MENU_CONFIG_TAG,
--- a/crates/egui/src/containers/popup.rs
+++ b/crates/egui/src/containers/popup.rs
@@ -571,7 +571,7 @@ impl<'a> Popup<'a> {
             .fixed_pos(anchor)
             .sense(sense)
             .layout(layout)
-            .sizing_pass(kind == PopupKind::Menu && !was_open_last_frame)
+            .sizing_pass(!was_open_last_frame)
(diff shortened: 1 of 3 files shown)


crates/egui/src/containers/popup.rs
```diff
@@ -571,6 +571,7 @@ impl<'a> Popup<'a> {
             .fixed_pos(anchor)
             .sense(sense)
             .layout(layout)
+            .sizing_pass(kind == PopupKind::Menu && !was_open_last_frame)
```
lucasmerlin: I think this is a nice solution. Any reason this is limited to Menu only? Might be useful for the other popups as well.

yay: Not really, I was just being cautious. I haven't personally encountered a use case where this would be an issue with other kinds of popups, though it probably is. I'm going to replace with `.sizing_pass(!was_open_last_frame)`.

lucasmerlin (approved):
Awesome, thanks!

Issue #8084, opened 2026-04-09
A question about possibly false positive for "changed id between passes" warning (due to intentional widget removal or drag and drop)


leosat:
Hi, @emilk !

I've updated for the new egui version (from 0.33* to 0.34*) recently and faced numerous "changed id between passes" warnings. 

https://github.com/emilk/egui/blame/86a7f47738b70e0b5c39950ca0962ee79e128033/crates/egui/src/context.rs#L4383

While some of those hits were indeed helping with unstable ID issues, some others seem inapplicable: 

Example: I have a list of items, **each having stable ui ID not attached to item position in list(I need to retain its state agnostically to position in list)** and **remove (or drag and drop )** one of them, so that others shift and take the position of the removed one. The widget removal in original position is the intent, but I get the warning.

Since the warning states "changed id between passes", my initial expectation was that the problem was in multipass execution, but I've checked that **I have only 1 pass per frame**, e.g. do not modify the ui between sizing pass and rendering pass.

I do the removal while in the parent widget ui closure, I suppose that should be workable for egui... 

If I'm correct and that is an inevitable false positive, is there a way to **suppress it for particular cases only**?.

// I recheck if can bypass the issue by wrapping with ui.push_id(it_index, ... ) and making id not changing, it helps with some cases, but since the layout changes after deletion - it will not generally help (e.g. still get warning for a dnd-target widget located at the end of the list).

WhitePetal:
I also encountered a similar problem. I would like to ask you how you solved it at that time? please

PR #8316, opened 2026-07-15
Fix changing widget IDs and false-positive ID change warnings across passes

### LINKED ISSUE #8092: warn_if_rect_changes_id false positive with virtualized egui_extras::Table
## Summary

`warn_if_rect_changes_id` produces false positives (red debug outlines) when scrolling a virtualized `egui_extras::TableBuilder` table.

## What happens

`egui_extras::TableBuilder` virtualizes rows: only visible rows are laid out each frame. When the user scrolls, a screen `Rect` that previously belonged to row N now belongs to row M. The inner `Ui` container for each cell (`UiKind::TableCell`) gets a different `Id` because it's a different logical row.

`warn_if_rect_changes_id` sees:
- Same `Rect`
- Different `Id`
- Same `parent_id` (the table body)

…and draws a red debug outline, treating it as widget id-instability.

For `TableCell` containers this is a **false positive**: no user widget state is being confused. The id change is expected — it's just a different row occupying the same screen slot after scroll.

## Reproduction

Any `egui_extras::TableBuilder` with enough rows to scroll will show red debug outlines on cells during/after scrolling in debug builds.

```rust
egui_extras::TableBuilder::new(ui)
    .id_salt("my_table")
    .column(Column::auto())
    .body(|body| {
        body.rows(20.0, 200, |mut row| {
            row.col(|ui| {
                ui.label(format!("Row {}", row.index()));
            });
        });
    });
```

Scroll the table → red outlines appear on cells.

## Analysis

The existing filters in `warn_if_rect_changes_id` don't cover this case:

1. **"at least one id stayed the same"** — No, the row changed, so the cell container id changed too
2. **"all previous ids still exist"** — No, the old row's cell is no longer laid out (virtualized away)
3. **"all parent_ids changed"** — No, parent is the same table body `Ui`

All three filters pass, so the warning fires. But it's semantically wrong — this is normal virtualization behavior, not a bug.

## Possible fix directions

The root cause is that `warn_if_rect_changes_id` has no way to distinguish "widget whose id is unstable (bug)" from "virtual container that legitimately maps different logical content to the same screen rect (normal)".

Some ideas (not prescriptive):

- **Tag approach**: `WidgetRect` could carry an `Option<UiKind>` so the check can skip `TableCell ↔ TableCell` transitions. Cost: 0 bytes (fits in existing padding on 64-bit). Downside: threads `UiKind` through all `WidgetRect` construction sites.
- **Sense-based heuristic**: skip the warning when all conflicting widgets have `Sense::hover()` (Ui containers, not interactive widgets). Less precise.
- **Opt-out flag**: let `TableBuilder` (or any virtualizing container) mark its child `Ui`s as "expect rect reuse", suppressing the check for those rects.

## Workaround

We currently vendor egui 0.34.1 and patch `warn_if_rect_changes_id` with the tag approach (adding `ui_kind: Option<UiKind>` to `WidgetRect`). It works but feels heavy for a debug-only check — would prefer an upstream solution.

## Version

egui 0.34.1

vlansco:

Closes #8092.
Related to #8084.

Virtualized tables could shift the parent UI's automatic ID sequence between sizing and rendering passes.

`StripLayout::skip_space` represented non-visible rows using `allocate_rect(..., Sense::hover())`. This preserved the scroll extent, but also registered a hover widget and consumed an automatic widget ID. When virtual buffers differed between passes, subsequent table widget IDs changed and triggered warnings such as:

> Widget rect changed id between passes

Multi-pass and data-driven layouts can also legitimately move existing widgets into vacated rects or intentionally replace all widgets within a region. These cases could produce the same warning even though the widget IDs were stable or the replacement was intentional.

This PR:

- uses `expand_to_include_rect` for virtual row buffers, preserving the scroll extent without registering a widget or consuming an automatic ID;
- assigns deterministic IDs to table header, body, and cell UIs;
- avoids warnings when an existing widget moves into a rect vacated during the previous pass, while retaining warnings for genuinely new IDs replacing widgets in place;
- adds `Context::allow_widget_id_changes_in` as a pass-local, rect-scoped exclusion for regions that intentionally replace their widget set;
- adds regression coverage for virtual spacer extent, automatic ID preservation, moved widgets, genuine replacements, and scoped exclusions.

There are no intended visual changes. The only public API change is `Context::allow_widget_id_changes_in`, which affects this diagnostic only; it does not change duplicate-ID checks or widget interaction.

### Testing

- `cargo test -p egui -p egui_extras`
- `cargo clippy -p egui -p egui_extras --all-targets --all-features -- -D warnings`
- `cargo fmt --all -- --check`

Screenshots are not applicable because these changes do not affect visual output.


vlansco:
@emilk @lucasmerlin The branch is now current with `main`; the current head is `72bbe54`, and the final diff remains limited to four source files with no workflow changes.

The corresponding workflows passed in the fork:
- Rust: https://github.com/vlansco/egui/actions/runs/33842170153
- Git LFS check: https://github.com/vlansco/egui/actions/runs/33842170129

GitHub has left the upstream `pull_request` workflows awaiting maintainer approval (`action_required`), which I cannot approve as an outside contributor.

Would you mind approving those workflows and reviewing the PR when convenient?

This addresses #8092 and the scoped-suppression case discussed in #8084.

PR #8321, opened 2026-07-18
Make wgpu Instance public

oleflb:
Adds and exposes `wgpu::Instance` in `egui_wgpu::RenderState`. 

We use this to create a bevy render pipeline inside an egui app which reuses the wgpu state.
This allows bevy to render into a egui texture for easy and efficient display: https://github.com/HULKs/hulk/blob/e702c089621e9b20e1004f8fc6e0f8dafb9c4f06/crates/egui_bevy/src/lib.rs#L155


Wumpf (approved):
now that instances a more like clonable handles I see no reason not to 👍

PR #8332, opened 2026-07-22
Fix TextEdit hint text not following horizontal_align/vertical_align

### LINKED ISSUE #8309: TextEdit hint text alignment and cursor position is inconsistent with align functions
**Describe the bug**
When a horizontal_align is specified for a TextEdit it correctly applies to the inputed text but the behaviour for the text hint is inconsistent.
For example using `.horizontal_align(egui::Align::Center)` will result in the following behaviour: 
- the hint text is left aligned
- the cursor is centered on the hint text
- text entry will be centered

**To Reproduce**
```rust
ui.add(egui::TextEdit::singleline(&mut input).hint_text("Hint Text").horizontal_align(egui::Align::Center));
```

**Expected behavior**
I think the intent would be for the hint text to be aligned the same as the input text however it may be desireable to add an additional alignment for the hint text so it can be specified independently of input text.

**Screenshots**
<video src="https://github.com/user-attachments/assets/4dad3201-e1ba-4b6d-88f3-2f1db8f0de73" width="400" controls></video>

 - OS: Windows 11 Native App

**Additional context**
I believe the culprit for this behaviour is line 610 of widgets/text_edit/builder.rs:

```rust
// The hint text should be shown left top instead of centered (important for
 // multi line text edits)
atoms.push_right(atom.atom_align(Align2::LEFT_TOP));
```
This always aligns the hint text atom to LEFT_TOP regardless of what is passed to `horizontal_align` and `vertical_align`

thedavidweng:
## Summary

- Closes #8309

`TextEdit` hint text was always aligned to `Align2::LEFT_TOP`, ignoring the alignment set via `TextEdit::horizontal_align` / `vertical_align`. This caused the hint text, the cursor, and the typed text to disagree on alignment: e.g. a centered `TextEdit` showed a left-aligned hint with a centered cursor.

The hint text atoms now use the widget's `align`, so the hint matches the input text alignment. The default `align` is still `LEFT_TOP`, so multi line text edits (and the default styling) are unchanged.

### Root cause

In `crates/egui/src/widgets/text_edit/builder.rs`, the hint-text branch hardcoded:

```rust
atoms.push_right(atom.atom_align(Align2::LEFT_TOP));
```

while the input-text branch used `.atom_align(self.align)`. The hint path now uses `align` as well.

### Drive-by: silence `clippy::unnecessary_wraps` in `egui_kittest::app_kind`

`AppKind::run` returns `Option<egui::Response>`. The `Option` wrap is required when the `eframe` feature is enabled (the `Eframe` branch returns `None`), but `clippy::unnecessary_wraps` fires when `egui_kittest` is built standalone without the `eframe` feature (e.g. `cargo clippy -p egui_kittest`). The workspace CI run doesn't hit it because feature unification via `egui_demo_app` enables `eframe`, but it's a real annoyance for anyone linting the crate on its own. Added a scoped `#[cfg_attr(not(feature = "eframe"), expect(clippy::unnecessary_wraps))]` with an explanatory comment.

## Test plan

- [x] Added `textedit_hint_text_should_follow_text_alignment` kittest regression in `crates/egui_kittest/tests/regression_tests.rs`. It fails before the fix (`hint_center_x=24.25` vs `edit_center_x=100`) and passes after.
- [x] `cargo test -p egui`
- [x] `cargo test -p egui_kittest --all-features --test regression_tests`
- [x] `cargo clippy -p egui_kittest --all-features --test regression_tests -- -D warnings`
- [x] `RUSTFLAGS="-D warnings" cargo clippy -p egui_kittest --lib` (pre-existing `unnecessary_wraps` now silenced)
- [x] `cargo clippy -p egui -- -D warnings`
- [x] `cargo fmt --check`

### DIFF
--- a/crates/egui/src/widgets/text_edit/builder.rs
+++ b/crates/egui/src/widgets/text_edit/builder.rs
@@ -605,9 +605,10 @@ impl TextEdit<'_> {
                         first = false;
                     }
 
-                    // The hint text should be shown left top instead of centered (important for
-                    // multi line text edits)
-                    atoms.push_right(atom.atom_align(Align2::LEFT_TOP));
+                    // Align the hint text the same as the input text so the hint, the
+                    // cursor, and the typed text all share one alignment. The default
+                    // `align` is `LEFT_TOP`, which keeps multi line text edits unchanged.
+                    atoms.push_right(atom.atom_align(align));
                 }
 
                 // Calculate the empty galley, so it can be read later. The available width is
--- a/crates/egui_kittest/src/app_kind.rs
+++ b/crates/egui_kittest/src/app_kind.rs
@@ -23,6 +23,10 @@ pub(crate) enum AppKind<'a, State> {
 }
 
 impl<State> AppKind<'_, State> {
+    // The `Option` is needed when the `eframe` feature is enabled, because the
+    // `Eframe` variant has no `egui::Response` to return. Without `eframe` the
+    // wrap is unnecessary, so we silence `clippy::unnecessary_wraps` for that case.
+    #[cfg_attr(not(feature = "eframe"), expect(clippy::unnecessary_wraps))]
     pub fn run(
         &mut self,
         ui: &mut egui::Ui,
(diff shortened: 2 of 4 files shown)


lucasmerlin (approved):
thanks!

PR #8335, opened 2026-07-23
Add public `egui_kittest::Node` constructor 

lucasmerlin:
- Adds public `egui_kittest::Node` constructor 
- Fixes event coordinate mismatch when pixels per point != 1


emilk (approved):

lucasmerlin:
Ooops, this was also in https://github.com/emilk/egui/pull/8336, so this PR is now empty. Closing since it already landed

PR #8336, opened 2026-07-23
Remove `Modifiers` from `RawInput` and make it a `egui::Event`

lucasmerlin:
- needs #8335 

Previously it was impossible to correctly set modifiers via egui inspection, since that only passes egui::Event and egui::Event had no way to generally express modifiers. 
Now modifiers are passed to egui as a Event and not as a field of `RawInput`.

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -84,6 +84,13 @@ pub struct State {
     viewport_id: ViewportId,
     start_time: web_time::Instant,
     egui_input: egui::RawInput,
+
+    /// The current modifier state.
+    ///
+    /// egui no longer stores this on [`egui::RawInput`]; we keep our own copy so we can stamp
+    /// it onto per-event `modifiers` fields and emit [`egui::Event::ModifiersChanged`].
+    modifiers: egui::Modifiers,
+
     pointer_pos_in_points: Option<egui::Pos2>,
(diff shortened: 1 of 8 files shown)


crates/egui/src/input_state/mod.rs
```diff
@@ -393,6 +393,8 @@ impl InputState {
         let pointer = self.pointer.begin_pass(time, &new, options);
 
         let mut keys_down = self.keys_down;
+        // Carry the modifier state over: integrations only send `ModifiersChanged` on change.
```
emilk:
```suggestion
```

crates/egui-winit/src/lib.rs
```diff
@@ -84,6 +84,13 @@ pub struct State {
     viewport_id: ViewportId,
     start_time: web_time::Instant,
     egui_input: egui::RawInput,
+
+    /// The current modifier state.
+    ///
+    /// egui no longer stores this on [`egui::RawInput`]; we keep our own copy so we can stamp
+    /// it onto per-event `modifiers` fields and emit [`egui::Event::ModifiersChanged`].
```
emilk:
```suggestion
    /// We keep a copy so we can stamp
    /// it onto per-event `modifiers` fields and emit [`egui::Event::ModifiersChanged`].
```

emilk (approved):

PR #8337, opened 2026-07-23
fix: ensure mapped range is dropped before unmapping buffer in capture

MagicCrazyMan:
After upgrading to wgpu v30, it requires to drop the `BufferView` return by `get_mapped_range` before unmap the buffer. Fix it in this pr, making screenshot event crash no more


emilk (approved):

emilk:
thanks!

PR #8338, opened 2026-07-24
Add `MenuConfig::submenu_gap` to override the hardcoded gap between a menu and its submenus

kiwirm:
The gap between a menu and the submenu it opens is hardcoded in `SubMenu::show`
(`frame.total_margin().sum().x / 2.0 + 2.0`), so a submenu can't be placed flush
against its parent — even with a zero menu margin a ~2px gap remains.

This adds an optional `submenu_gap: Option<f32>` to `MenuConfig` (plus a
`MenuConfig::submenu_gap(f32)` builder). When set it overrides the offset; when
`None` it falls back to the existing computed value, so existing menus are
unaffected. Because a menu's `MenuConfig` propagates to its submenus, setting it
once on the root menu applies to the whole tree.

```rust
egui::MenuButton::new("Add")
    .config(egui::MenuConfig::new().submenu_gap(0.0)) // flush submenus
    .ui(ui, |ui| { /* … */ });
```

### DIFF
--- a/crates/egui/src/containers/menu.rs
+++ b/crates/egui/src/containers/menu.rs
@@ -509,10 +509,9 @@ impl SubMenu {
             });
         }
 
-        let gap = menu_config
+        let gap = frame.total_margin().sum().x / 2.0 + menu_config
             .submenu_gap
-            .unwrap_or_else(|| frame.total_margin().sum().x / 2.0 + 2.0);
-
+            .unwrap_or(2.0);
         let mut response = button_response.clone();
         // Expand the button rect so that the button and the first item in the submenu are aligned
         let expand = Vec2::new(0.0, frame.total_margin().sum().y / 2.0);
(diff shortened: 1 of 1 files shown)


crates/egui/src/containers/menu.rs
```diff
@@ -492,7 +509,9 @@ impl SubMenu {
             });
         }
 
-        let gap = frame.total_margin().sum().x / 2.0 + 2.0;
+        let gap = menu_config
+            .submenu_gap
+            .unwrap_or_else(|| frame.total_margin().sum().x / 2.0 + 2.0);
```
lucasmerlin: Should this be 
```suggestion
        let gap = frame.total_margin().sum().x / 2.0 + menu_config
            .submenu_gap
            .unwrap_or(2.0);
```
instead? 
`frame.total_margin().sum().x / 2.0` should basically mean "0 visual gap between the menus" which I think is what you'd expect when setting the gap to 0? Otherwise at 0 there would be some overlap.

kiwirm: Yes, thank you

Issue #8326, opened 2026-07-19
Reactive eframe application uses 100% of one core on mouse move


tweber12:
**Describe the bug**

An empty egui + eframe program, in reactive mode uses up 100% on one CPU core when just moving the mouse.

I profiled my application and traced the CPU usage to winit polling:
<img width="800"  alt="Image" src="https://github.com/user-attachments/assets/f3c69ef2-fc14-43b0-878e-1da8c46b4bf6" />
Afterwards, I also reproduced the issue on the egui hello_world example, to exclude any issues with my code.

After a search through the eframe code, I found that eframe sets the winit event loop control flow to `ControlFlow::Poll` in  [run.rs:209](https://github.com/emilk/egui/blob/0.35.0/crates/eframe/src/native/run.rs#L209). If I remove this line, and set `ControlFlow::Wait` below (which seems to be more appropriate for reactive applications according to the winit docs), the issue seems to be gone, indicating that this is in fact related to the `Poll` setting. This diff below is what I changed in my test:

```diff
                     } else {
                         log::trace!("request_redraw for {window_id:?}");
-                        event_loop.set_control_flow(ControlFlow::Poll);
                         window.request_redraw();
                     }
```

```diff
-        if let Some(next_repaint_time) = next_repaint_time {
-            event_loop.set_control_flow(ControlFlow::WaitUntil(next_repaint_time));
-        }
+        event_loop.set_control_flow(match next_repaint_time {
+            Some(next_repaint_time) => ControlFlow::WaitUntil(next_repaint_time),
+            None => ControlFlow::Wait,
+        });
     }
```

**To Reproduce**

1. Run any egui + eframe application in reactive mode, in particular e.g. the egui 'hello_world' example works.
2. Move the mouse cursor in the window and observe the CPU usage.
3. Optionally apply the diff above and check that the CPU usage is significantly reduced.

**Expected behavior**

CPU usage should stay limited to a few percent, not one full core.

**Desktop (please complete the following information):**
 - OS: Cachy OS, KDE Plasma, Wayland
 - Version:
   -  eframe 0.35.0
   - wgpu backend (`default-features = false`, `features = ["wgpu", "default_fonts", "wayland", "x11"]`)
   - winit 0.30.13
   - calloop 0.13.0
   - polling 3.11.0

**Additional context**

 Possibly the same as #5092, however the issue there is reported on X11, while I am using Wayland.

42Pupusas:
[...]

PR #8318, opened 2026-07-16
Solution (#8314): Drag-resize causes application lock-up

TFGSUMIT:
This pull request addresses the issue of drag-resize causing application lock-up by introducing a new event loop context that handles the case where the RedrawRequested event is deferred while a frame callback is outstanding. The changes made include:

*   Introduced a new `EventLoopContext` class that handles the event loop and control flow.
*   Used threading events to manage the event loop and control flow.
*   Request redraws using the `EventLoopContext` class.

To test this pull request, follow these instructions:

*   Clone the repository and build the application.
*   Run the application and attempt to drag and resize a window.
*   Verify that the application no longer locks up and responds correctly to redraw requests.

Note: This pull request requires Python 3.8 or later to run.

Issue #7401, opened 2025-08-01
`pure_glow` high CPU usage


MightyBurger:
**Describe the bug**

The [pure_glow example](https://github.com/emilk/egui/blob/main/crates/egui_glow/examples/pure_glow.rs) causes high CPU usage.

**To Reproduce**
1. Clone `egui` repo on `main` branch
2. `cd crates/egui_glow`
3. `cargo run --example pure_glow --features="winit egui/default_fonts"`
4. Observe very high CPU usage, even when not interacting with the window.

**Expected behavior**
I do not expect this example to cause one of the CPU cores to run at nearly 100% for this example.

For example, `eframe` supports a `reactive` mode which only updates when necessary as can be seen on [the demo website](https://www.egui.rs/). I would expect this example to behave similarly.

**Screenshots**
See high CPU usage in the below screenshot.

<img width="1048" height="150" alt="Image" src="https://github.com/user-attachments/assets/7a94b8fe-d6a5-4d46-ab8f-1bbd6d40d670" />

**Desktop:**
 - OS: Windows 10 Pro 22H2
 - Browser: n/a, desktop app
 - Version: n/a, desktop app

**Additional context**
I am developing a [Windows app](https://github.com/MightyBurger/reemap) that uses `egui`. It must run in the background, and it must support placing an icon in the tray. `eframe` does not support these features, so I turned to this `pure_glow` example and modified it to my needs.

I am running into problems with my app consuming much CPU when the GUI is visible, however. Setting it to `--release` mode does help reduce the CPU usage, but it doesn't fix the underlying problem.

I suspect the issue is because `repaint_delay` never changes once it's set to 0. That said, I don't have a good enough mental model of how `egui` integrates with `winit` at this time to fully understand the source of this bug. (Resources would be greatly appreciated. I don't understand this example very well. It's the one part of my app I don't have a full grasp on.)

It is just an example- is this outside the scope of the example? With my limited understanding, it seems like a bug.

YgorSouza:
Might be the same as #4173. With eframe, I worked around it by adding a 17ms `thread::sleep` to the `update` call, so it doesn't have time to spam these expensive glutin calls between frames.

MightyBurger:
I'm pretty sure I figured out what's going on.

The `pure_glow` example runs hot, I think, because there's nothing that sets `repaint_delay` back to 0. So, looking at what happens in `window_event()`, it just ends up endlessly calling `request_redraw()`.

[...]

Issue #5092, opened 2024-09-08
100% CPU usage since the update to wgpu 0.20


EmbersArc:
**Describe the bug**

Any egui app uses 100% CPU when redrawing since commit 814ad0783cf8b826a258e29ed4c50ae6daa2e890, the wgpu 0.20 update.

**To Reproduce**
1. Run e.g. `cargo run --bin hello_world`
2. Move the cursor
3. The CPU usage of the egui process goes up to 100%

**Expected behavior**

Only a fraction of the CPU usage as observed before the wgpu update.

**Screenshots**

**Desktop**
 - OS: Arch Linux 6.10.8-arch1-1 (64-bit)
 - X11 with NVIDIA GPU

EmbersArc:
I did some bisecting and the offending wgpu PR seems to be https://github.com/gfx-rs/wgpu/pull/4967.

bircni:
We are already waiting for a new wgpu release in other PRs hopefully this will be there before the new egui release
https://github.com/emilk/egui/pull/4952

https://github.com/gfx-rs/wgpu/issues/6240

Wumpf:
egui is already on wgpu 22.0, so unless there is reason to believe this is fixed on wgpu main (which pr fixes it?) this is completely unrelated

EmbersArc:
Switching to wgpu trunk does not fix it yet unfortunately.

EmbersArc:
> @EmbersArc can you confirm you're running X11 native and not wayland+xwayland?

Yes this is on X11 native.

cwfitzgerald:
Howdy, wgpu maintainer here. We really will need a bit more information on the performance usage here. Would it be possible to take a performance profile to understand where the cpu is spending its time? Vulkan swapchain sync is mildly cursed, so even after the rewrite we may be still going it subtly wrong.

Wumpf:
@jleibs and me will dive a bit deeper into this either this or next week (we work at the same company and while I'm not as knowledgeable as @cwfitzgerald in this area, I know a few things about wgpu from being a maintainer myself 😄, making this easier)
From what we learned so far this is very related to the latest Nvidia drivers as there has been a lot of changes on how they handle vsync. But I also suspect that there's some subtleties in wgpu still.

jleibs:
Also because I happened to turn on the other computer I'll note that I see this on:
 - OS: Ubuntu 22.10
 - X11 with Nvidia 535.54.03
 
[...]

PR #8341, opened 2026-07-25
Avoid redundant redraw requests in native run loop

rustbasic:
Avoid redundant redraw requests in native run loop

* Related #5092
* Related #7401 
* Related #8318
* Related #8326

## Summary

Avoid repeatedly requesting redraws for the same native window while a previous
`request_redraw()` is still pending.

## Details

`request_redraw()` schedules a future `WindowEvent::RedrawRequested`. In reactive
apps, repeatedly calling `request_redraw()` before that event is delivered can
keep the event loop unnecessarily busy.

This PR adds per-window pending redraw tracking:

- mark a window as pending when `request_redraw()` is called
- skip additional redraw requests while pending
- clear the pending state on `WindowEvent::RedrawRequested`
- clear it for invisible/minimized windows that are painted directly

The run loop now also falls back to `ControlFlow::Wait` when there is no next
scheduled repaint, instead of relying on `ControlFlow::Poll`.

This should reduce redundant redraw requests and unnecessary CPU usage in
reactive native applications.


rustbasic:
The fact that the `deferred viewport` fails to redraw when the `ROOT viewport` is clicked is an existing bug and is unrelated to this PR.
 
Solved : 
* #8395

rustbasic:
#8398 has been merged, so this PR is closed.

PR #8352, opened 2026-07-29
Avoid redundant redraw requests in native run loop & `deferred viewport` does not redraw

### LINKED ISSUE #8351: `deferred viewport` does not redraw when clicking `ROOT viewport` (Windows)
egui Version : main & All
OS : Windows 10

There is an issue where the `deferred viewport` is not redrawn when the `ROOT viewport` is clicked. 
 
I have confirmed that this problem occurs on Windows, but I am currently unsure if it also occurs on other operating systems.

rustbasic:
Avoid redundant redraw requests in native run loop

* Related #5092
* Related #7401 
* Related #8318
* Related #8326

* Closes #8341
* Closes #8351 

## Summary

1. Avoid repeatedly requesting redraws for the same native window while a previous
`request_redraw()` is still pending.

2. This PR fixes the bug where the `deferred viewport` fails to redraw upon clicking the `ROOT viewport`.
Since it is unclear whether this behavior also occurs on other platforms, I limited the scope to Windows with `cfg!(target_os = "windows")`.

## Details

`request_redraw()` schedules a future `WindowEvent::RedrawRequested`. In reactive
apps, repeatedly calling `request_redraw()` before that event is delivered can
keep the event loop unnecessarily busy.

This PR adds per-window pending redraw tracking:

- mark a window as pending when `request_redraw()` is called
- skip additional redraw requests while pending
- clear the pending state on `WindowEvent::RedrawRequested`
- clear it for invisible/minimized windows that are painted directly

The run loop now also falls back to `ControlFlow::Wait` when there is no next
scheduled repaint, instead of relying on `ControlFlow::Poll`.

This should reduce redundant redraw requests and unnecessary CPU usage in
reactive native applications.

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -383,6 +383,53 @@ impl WinitApp for GlowWinitApp<'_> {
             .copied()
     }
 
+    fn window_ids_for_repaint_request(&self, id: ViewportId) -> Vec<WindowId> {
+        let Some(running) = self.running.as_ref() else {
+            return Vec::new();
+        };
+        let glutin = running.glutin.borrow();
+
+        let mut window_ids = Vec::new();
+        let mut push_unique = |window_id| {
+            if !window_ids.contains(&window_id) {
+                window_ids.push(window_id);
+            }
+        };
+
+        if id == ViewportId::ROOT {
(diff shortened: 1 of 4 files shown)


rustbasic:
Closed for an easier solution.

PR #8354, opened 2026-07-29
Store `web_sys::File` inside of `DroppedFile`

### LINKED ISSUE #4654: Drag and Drop: read incrementally

grtlr:
* Closes #4654
* Related #4667

This PR avoids materializing the contents of a file that was dragged into an egui application on the web. It does so by storing the `web_sys::File` handle directly on WASM.

This breaks the existing API of `DroppedFile` on the web, because there is no way to retrieve the bytes synchronously form a `DroppedFile` anymore, forcing handling call sites to become asynchronous.

The native API remains the same.

### DIFF
--- a/crates/eframe/src/lib.rs
+++ b/crates/eframe/src/lib.rs
@@ -155,10 +155,6 @@ compile_error!("`accesskit` feature is only available with `android-game-activit
 // Re-export all useful libraries:
 pub use {egui, egui::emath, egui::epaint};
 
-#[cfg(target_arch = "wasm32")]
-#[path = "web/dropped_file.rs"]
-mod dropped_file;
-
 #[cfg(feature = "glow")]
 pub use {egui_glow, glow};
 
--- a/crates/eframe/src/lib.rs
+++ b/crates/eframe/src/lib.rs
@@ -155,6 +155,10 @@ compile_error!("`accesskit` feature is only available with `android-game-activit
 // Re-export all useful libraries:
 pub use {egui, egui::emath, egui::epaint};
 
+#[cfg(target_arch = "wasm32")]
+#[path = "web/dropped_file.rs"]
+mod dropped_file;
+
 #[cfg(feature = "glow")]
 pub use {egui_glow, glow};
 
--- a/crates/eframe/src/web/dropped_file.rs
+++ b/crates/eframe/src/web/dropped_file.rs
@@ -15,7 +15,8 @@ impl egui::DroppedFile for WebDroppedFile {
     fn bytes_async(&self) -> Pin<Box<dyn Future<Output = Result<Vec<u8>, String>> + '_>> {
         let file = self.file.clone();
         Box::pin(async move {
-            let array_buffer = wasm_bindgen_futures::JsFuture::from(file.array_buffer())
+            let array_buffer = file
+                .array_buffer()
                 .await
                 .map_err(|err| crate::web::string_from_js_value(&err))?;
             Ok(js_sys::Uint8Array::new(&array_buffer).to_vec())
--- a/crates/eframe/src/web/dropped_file.rs
+++ b/crates/eframe/src/web/dropped_file.rs
@@ -1,0 +1,28 @@
+use std::{future::Future, pin::Pin};
+
+#[derive(Debug)]
+pub(crate) struct WebDroppedFile {
+    file: web_sys::File,
+}
+
+impl WebDroppedFile {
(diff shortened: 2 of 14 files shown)


crates/egui-winit/src/dropped_file.rs
```diff
@@ -0,0 +1,27 @@
+use std::{future::Future, path::Path, pin::Pin};
+
+#[derive(Debug)]
+pub(crate) struct NativeDroppedFile {
```
grtlr:
```suggestion
pub(crate) struct NativeFile {
```

crates/eframe/src/web/dropped_file.rs
```diff
@@ -0,0 +1,28 @@
+use std::{future::Future, pin::Pin};
+
+#[derive(Debug)]
+pub(crate) struct WebDroppedFile {
```
grtlr:
```suggestion
pub(crate) struct WebFile {
```

These contain nothing that is drop-related and could be used more general in the future.

crates/egui-winit/src/dropped_file.rs
```diff
@@ -0,0 +1,27 @@
...
+
+impl egui::DroppedFile for NativeDroppedFile {
+    fn path(&self) -> &Path {
+        &self.path
+    }
+
+    fn bytes_async(&self) -> Pin<Box<dyn Future<Output = Result<Vec<u8>, String>> + '_>> {
+        let path = self.path.clone();
+        Box::pin(async move { std::fs::read(path).map_err(|err| err.to_string()) })
+    }
```
grtlr: Probably don't want to add that unless we can be truly async?

crates/eframe/src/web/dropped_file.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub(crate) fn from_web_file(file: web_sys::File) -> egui::DroppedFileHandle {
+        std::sync::Arc::new(Self { file })
+    }
+}
+
+impl egui::DroppedFile for WebDroppedFile {
+    fn bytes_async(&self) -> Pin<Box<dyn Future<Output = Result<Vec<u8>, String>> + '_>> {
+        let file = self.file.clone();
+        Box::pin(async move {
+            let array_buffer = wasm_bindgen_futures::JsFuture::from(file.array_buffer())
```
grtlr: I don't think we need the `JsFuture::from`, and should be able to simply await `file.array_buffer`.

grtlr: Pushed

grtlr (commented):
Thank you for moving this along!

crates/eframe/src/web/dropped_file.rs
```diff
@@ -0,0 +1,29 @@
...
+
+impl egui::DroppedFile for WebFile {
+    fn bytes_async(&self) -> Pin<Box<dyn Future<Output = Result<Vec<u8>, String>> + '_>> {
+        let file = self.file.clone();
+        Box::pin(async move {
+            let array_buffer = file
+                .array_buffer()
+                .await
+                .map_err(|err| crate::web::string_from_js_value(&err))?;
+            Ok(js_sys::Uint8Array::new(&array_buffer).to_vec())
```
emilk, line 41: we should check for 32 bit overflow here

crates/eframe/src/lib.rs
```diff
@@ -155,6 +155,10 @@ compile_error!("`accesskit` feature is only available with `android-game-activit
 // Re-export all useful libraries:
 pub use {egui, egui::emath, egui::epaint};
 
+#[cfg(target_arch = "wasm32")]
+#[path = "web/dropped_file.rs"]
```
emilk: this is weird - should be moved to under `web` folder instead

emilk (approved):

emilk (commented):
Please see my comments on https://github.com/rerun-io/reality/pull/2962 before merging 🙏

PR #8356, opened 2026-07-29
Prevent accidentally dropping `TexturesDelta`

lucasmerlin:
We had a ton of issues around `TexturesDelta` that weren't properly applied because we early-out of some function:
* https://github.com/emilk/egui/pull/8313
* https://github.com/emilk/egui/pull/8250
* https://github.com/emilk/egui/pull/8279

This PR changes texture updates, so that we always store them after taking them out of `FullOutput` and keep the delta around until it's actually applied (by passing &mut refs and draining instead of iterating). So even if we add a new early return somewhere, that can't break texture updates.

It also optimizes `TexturesDelta::append` by dropping any previous deltas if there's a new `whole` delta or a `free`.

It also adds a debug assert that any `TexturesDelta` is empty when dropped, as an additional safeguard in case the bug sneaks back in.

### DIFF
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -139,6 +139,13 @@ struct Viewport {
     egui_winit: Option<egui_winit::State>,
 }
 
+impl Drop for Viewport {
+    fn drop(&mut self) {
+        // Avoid debug panic when dropping unapplied deltas on teardown
+        self.pending_delta.clear();
+    }
+}
+
 // ----------------------------------------------------------------------------
 
 impl<'app> GlowWinitApp<'app> {
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -697,8 +697,10 @@ impl GlowWinitRunning<'_> {
             frame_timer.resume();
         }
 
-        for (id, image_delta) in &textures_delta.set {
-            painter.set_texture(*id, image_delta);
+        for (id, image_deltas) in &textures_delta.set {
+            for image_delta in image_deltas {
+                painter.set_texture(*id, image_delta);
+            }
         }
 
         if is_visible {
--- a/crates/eframe/src/native/glow_integration.rs
+++ b/crates/eframe/src/native/glow_integration.rs
@@ -31,15 +31,16 @@ use egui::{
 };
 #[cfg(feature = "accesskit")]
 use egui_winit::accesskit_winit;
-
-use crate::{
-    App, AppCreator, CreationContext, NativeOptions, Result, Storage,
-    native::{epi_integration::EpiIntegration, winit_integration::is_invisible_or_minimized},
-};
+use log::warn;
 
 use super::{
     epi_integration, event_loop_context,
     winit_integration::{EventResult, UserEvent, WinitApp, create_egui_context},
+};
+use crate::epaint::textures::TexturesDelta;
+use crate::{
+    App, AppCreator, CreationContext, NativeOptions, Result, Storage,
+    native::{epi_integration::EpiIntegration, winit_integration::is_invisible_or_minimized},
 };
 
 // ----------------------------------------------------------------------------
@@ -73,6 +74,9 @@ struct GlowWinitRunning<'app> {
 
     // NOTE: one painter shared by all viewports.
     painter: Rc<RefCell<egui_glow::Painter>>,
+
+    /// Any not yet applied deltas for this app.
+    pending_deltas: TexturesDelta,
 }
 
 /// This struct will contain both persistent and temporary glutin state.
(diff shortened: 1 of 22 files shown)


crates/egui-wgpu/src/winit.rs
```diff
@@ -742,6 +745,7 @@ impl Painter {
         // However, once we called `wgpu::Queue::submit`, it is up for wgpu to determine how long the underlying gpu resource has to live.
         {
             let mut renderer = render_state.renderer.write();
+            #[expect(clippy::iter_over_hash_type)] // Order doesn't matter here
             for id in &textures_delta.free {
```
Wumpf: should this also be drained so we don't try to free the same textures all the time?

Wumpf (changes requested):
nice, that should put an end to the whack-a-mole!

found an issue though, plz address or tell me I'm wrong :)

crates/eframe/src/native/glow_integration.rs
```diff
@@ -689,16 +699,19 @@ impl GlowWinitRunning<'_> {
 
         // Upload textures even when not visible: the atlas dirty region is already
         // consumed, so dropping the delta would desync the font texture.
-        let has_texture_updates = !textures_delta.set.is_empty() || !textures_delta.free.is_empty();
+        let has_texture_updates = !pending_deltas.set.is_empty() || !pending_deltas.free.is_empty();
         if is_visible || has_texture_updates {
```
emilk: I guess we could re-introduce the early-out on `!is_visible` again (e.g. undo most of https://github.com/emilk/egui/pull/8250/changes), since we now accumulate to `pending_deltas` anyways (skipping -> postponing = fine)

crates/eframe/src/native/wgpu_integration.rs
```diff
@@ -1125,8 +1143,11 @@ fn render_immediate_viewport(
     } = &mut *shared_mut;
 
     let Some(viewport) = viewports.get_mut(&ids.this) else {
+        warn!("Viewport disappeared unexpectedly!");
```
emilk, line 1153: rm?

crates/egui_demo_lib/src/lib.rs
```diff
@@ -72,11 +72,12 @@ fn test_egui_e2e() {
 
     const NUM_FRAMES: usize = 5;
     for _ in 0..NUM_FRAMES {
-        let full_output = ctx.run_ui(raw_input.clone(), |ui| {
+        let mut full_output = ctx.run_ui(raw_input.clone(), |ui| {
             demo_windows.ui(ui);
         });
         let clipped_primitives = ctx.tessellate(full_output.shapes, full_output.pixels_per_point);
         assert!(!clipped_primitives.is_empty());
+        full_output.textures_delta.clear(); // Don't panic on drop with unapplied deltas
```
emilk, line 80: this is a bit annoying though…

crates/epaint/src/textures.rs
```diff
@@ -58,10 +61,8 @@ impl TextureManager {
                 // whole update
                 meta.size = delta.image.size();
                 meta.bytes_per_pixel = delta.image.bytes_per_pixel();
-                // since we update the whole image, we can discard all old enqueued deltas
-                self.delta.set.retain(|(x, _)| x != &id);
             }
-            self.delta.set.push((id, delta));
+            self.delta.set(id, delta);
```
emilk: I think there is a bug here: two partial updates following each other will result in only one partial update instead of two

crates/epaint/src/textures.rs
```diff
@@ -276,20 +283,47 @@ pub enum TextureWrapMode {
...
 impl TexturesDelta {
     pub fn is_empty(&self) -> bool {
         self.set.is_empty() && self.free.is_empty()
     }
 
+    /// Inserts a [`ImageDelta`].
+    ///
+    /// If this [`TexturesDelta`] already contains this [`TextureId`], and this is a `whole` delta,
+    /// the previous deltas for this id are discarded.
+    pub fn set(&mut self, id: TextureId, delta: ImageDelta) {
```
emilk: oh, that's a confusing name (see my previous confused comment)

```suggestion
    pub fn push(&mut self, id: TextureId, delta: ImageDelta) {
```

emilk (approved):
Much nicer approach!

crates/egui-wgpu/src/winit.rs
lucasmerlin: Ooops, good catch, I missed that one!

crates/eframe/src/native/wgpu_integration.rs
lucasmerlin: This should not happen in practise and since we would drop the TexturesDelta here, I thought it'd be better to warn here. I could also turn this into a debug assert though.

crates/epaint/src/textures.rs
lucasmerlin: It's handled internally, there's a SmallVec inside the Set

PR #8362, opened 2026-07-31
Make non-interactive tooltips not interactable

lucasmerlin:
Fixes two bugs around tooltips:
- During sizing pass tooltips would render at a different size, at which point they might overlap the interacted widget causing the hover state to change and the tooltip to never be shown
  - Fix: During sizing pass make the area `interactable: false` so events pass through 
- When hovering close to the boarder of a widget (in the area where `interact_radius` takes effect), there could be a feedback loop where the tooltip is shown one frame and hidden the next.
  - Fix: Always make tooltips `interactable: false` when they don't have interactive contents

This PR changes the behavior of `Area::interactable` to it's original behavior: Now interactions will pass through the area background _and_ it's containing widgets. It worked this way initially but got changed in #4026

### DIFF
--- a/crates/egui/src/containers/area.rs
+++ b/crates/egui/src/containers/area.rs
@@ -481,10 +481,9 @@ impl Area {
 
             size
         });
+
         // We should never be interactable during a sizing pass, since then we are shown at a different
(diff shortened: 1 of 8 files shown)


grtlr (approved):
Nice!

PR #8363, opened 2026-07-31
Add drag-to-open for collapsible panels

emilk:
A fully collapsed `show_collapsible` panel now leaves a thin grab handle at its fixed edge, invisible until hovered. Dragging it out past `min_size` (or double-clicking it) reopens the panel. Opt out with `panel.drag_to_open(false)`.


crates/egui/src/containers/panel.rs
```diff
@@ -424,10 +457,11 @@ impl Panel {
         let how_expanded = animate_expansion(ui, self.id.with("animation"), *is_expanded);
 
         if how_expanded == 0.0 {
-            // Panel is fully closed. If the user is still dragging the resize handle
-            // from a previous frame, keep its widget id alive so they can drag the
-            // panel back out without releasing.
-            self.keep_drag_alive_for_reopen(ui, is_expanded);
```
lucasmerlin, line 430: So is there no keep alive when drag_to_open is disabled?

EDIT: Tested it and yeah it doesn't work anymore without drag_to_open

crates/egui/src/containers/panel.rs
```diff
@@ -585,7 +610,19 @@ impl Panel {
             let panel = if how_expanded < 1.0 {
                 // Animate the visible size from collapsed_size to expanded_size,
                 // so the slide picks up where the collapsed panel left off.
-                let expanded_size = expanded_panel.outer_size(ui);
+                let expanded_size = if drag_in_progress {
+                    // The drag decides the size, and it can't take the panel below
+                    // its own `min_size` — so that, not the (stale) persisted size,
+                    // is where the slide has to meet the collapsed panel. Getting
+                    // this wrong makes the panel jump the gap between the two sizes
+                    // in a single frame, in whichever direction it is going.
```
lucasmerlin:
```suggestion
                    // Ensure the collapse starts from the min size, so there is no jump in the animation
```
Took me a couple reads to understand what claude is trying to tell use here. Not sure if I got it right though?

lucasmerlin (approved):
Nice! Feels intuitive

crates/egui/src/containers/panel.rs
emilk: Right - then you do drag-to-close now, you can only undo it (by holding down the mouse and dragging back out) if `drag_to_open` is enabled (which is the default). I think it makes sense 🤷

PR #8364, opened 2026-07-31
Clarify wording of IdTypeMap::remove_by_type documentation

the-maddin:
The function to remove all entries of a type from IdTypeMap was confusingly documented as "Note all state of the given type".
As far as I know, (not a native speaker), "noting" shares no meaning with "removing", so I changed the doc to simply state what the function does.

### DIFF
--- a/crates/egui/src/util/id_type_map.rs
+++ b/crates/egui/src/util/id_type_map.rs
@@ -596,7 +596,7 @@ impl IdTypeMap {
         }
     }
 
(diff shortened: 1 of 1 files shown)


lucasmerlin (approved):
Thanks!

the-maddin:
Hi, is there still interest in this PR? I‘m working on a feature that touches this file and would like to have this PR merged first.

PR #8365, opened 2026-07-31
Treat a press that leaves a widget as a drag

emilk:
A widget that senses both clicks and drags postpones the click-versus-drag decision until the pointer moves past `max_click_dist` or is held for `max_click_duration`. But a click has to be released *on* the widget — so once the pointer leaves, the gesture can only be a drag, and there is nothing left to wait for.

This matters for widgets thinner than `max_click_dist` (6px), such as panel resize handles. The pointer leaves such a widget almost immediately, which hands the hover to whatever is underneath, while `dragged()` was not true yet. So a handle highlighting on `hovered() || dragged()` blinked out mid-gesture.

### DIFF
--- a/crates/egui/src/interaction.rs
+++ b/crates/egui/src/interaction.rs
@@ -198,16 +198,18 @@ pub(crate) fn interact(
                 // When the mouse first is pressed, it could be either,
                 // so we postpone the decision until we know.
                 //
-                // …unless the pointer has left the widget: a click has to be
-                // released on the widget, so the gesture can no longer be one.
+                // …unless a click is no longer possible at all: a click has to be
+                // released on the widget, and `hits.click` tells us whether the
+                // pointer is still somewhere a release would land on this widget.
+                // Note that this is not the same as being inside `interact_rect`:
+                // the hit-test also picks up widgets within `interact_radius`, and
+                // lets a widget on top take the hit.
+                //
                 // Deciding here means a thin drag handle (narrower than
                 // `max_click_dist`) doesn't spend the decision window as neither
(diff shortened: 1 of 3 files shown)


crates/egui/src/interaction.rs
```diff
@@ -197,7 +197,17 @@ pub(crate) fn interact(
...
+                //
+                // …unless the pointer has left the widget: a click has to be
+                // released on the widget, so the gesture can no longer be one.
+                // Deciding here means a thin drag handle (narrower than
+                // `max_click_dist`) doesn't spend the decision window as neither
+                // hovered nor dragged, which would make its highlight blink out.
+                let left_the_widget = input
+                    .pointer
+                    .interact_pos()
+                    .is_some_and(|pos| !widget.interact_rect.contains(pos));
```
lucasmerlin: I think we should account for interact_radius here, or clicking close to a widget would drag immediately even though the cursor didn't move at all

lucasmerlin (commented):
It'd be awesome if the drag and drop example had some clickable items (maybe some toggle buttons or checkboxes) so we could test this behavior

crates/egui/src/interaction.rs
emilk: That's a good point 🤔

emilk: fixed!

emilk:
> It'd be awesome if the drag and drop example had some clickable items (maybe some toggle buttons or checkboxes) so we could test this behavior

We do have this example/test:

Screenshot 2026-08-04 at 11 02 36

lucasmerlin (approved):
Nice!

PR #8372, opened 2026-08-02
Eframe: make webbrowser dependency optional

### LINKED ISSUE #8371: eframe dependency on webbrowser should be optional
eframe depends on `webbrowser` through `egui-winit` which has a `links` feature that enables the dependency. Eframe does not expose this feature so it can't be disabled. I would like to not have this as a transitive dependency since its quite big for a feature i dont use

wyvernbw:
* Closes #8371

Adds a `link` feature to eframe to allow disabling links on egui-winit. The feature is enabled by default so nothing changes for existing users of eframe.

### DIFF
--- a/crates/eframe/Cargo.toml
+++ b/crates/eframe/Cargo.toml
@@ -122,6 +122,9 @@ __screenshot = []
 ## and capture screenshots. Off unless the env var is set; no-op on wasm.
 inspection = ["dep:egui_inspection", "accesskit"]
 
+## Enables the `links` feature on `egui-winit`, allowing for links to open in browser.
+links = ["egui-winit/links"]
+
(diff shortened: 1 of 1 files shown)


crates/eframe/Cargo.toml
```diff
@@ -33,6 +33,7 @@ default = [
   "wgpu",
   "winit/default",
   "x11",
+  "links"
```
emilk: please sort

emilk (approved):

crates/eframe/Cargo.toml
wyvernbw: done

emilk:
Thanks!

PR #8381, opened 2026-08-03
Fix `egui_kittest` failing to compile without the `wgpu` feature

emilk:


lucasmerlin (approved):
Thanks! I guess we should test without all-features in CI

PR #8383, opened 2026-08-03
 Rename `failed_pixel_count_threshold` to `max_failed_pixels`

emilk:
It was confusing that both tolerances had "threshold" in the name

### DIFF
--- a/crates/egui_kittest/README.md
+++ b/crates/egui_kittest/README.md
@@ -44,24 +44,28 @@ All possible settings and their defaults:
 # path to the snapshot directory
 output_path = "tests/snapshots"
 
-# default threshold for image comparison tests
+# maximum weighted squared YIQ color distance between two corresponding pixels
+# (an absolute, per-pixel value, independent of the image dimensions)
 threshold = 0.6
 
-# default failed_pixel_count_threshold
-failed_pixel_count_threshold = 0
+# how many pixels may exceed the `threshold` before the test fails
+max_failed_pixels = 0
 
 [windows]
 threshold = 0.6
-failed_pixel_count_threshold = 0
+max_failed_pixels = 0
 
 [macos]
 threshold = 0.6
-failed_pixel_count_threshold = 0
+max_failed_pixels = 0
 
 [linux]
 threshold = 0.6
-failed_pixel_count_threshold = 0
+max_failed_pixels = 0
 ```
+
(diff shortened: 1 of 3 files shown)


crates/egui_kittest/src/snapshot.rs
```diff
@@ -11,15 +11,25 @@ pub type SnapshotResult = Result<(), SnapshotError>;
 #[non_exhaustive]
 #[derive(Clone, Debug)]
 pub struct SnapshotOptions {
-    /// The threshold for the image comparison.
+    /// The maximum weighted squared YIQ color distance between corresponding pixels.
     ///
-    /// Can be configured via kittest.toml. The fallback is `0.6` (which is enough for most egui
-    /// tests to pass across different wgpu backends).
+    /// This is an absolute, per-pixel value and does not depend on the image dimensions.
```
lucasmerlin: I'm not sure why image dimensions are relevant here? When the dimensions mismatch we don't run per-pixel comparision at all I believe.

lucasmerlin (approved):
It's a mouthful, not sure if I like it, but looks good otherwise

lucasmerlin:
Maybe we should name it for what it's used for? e.g. `perceptual_threshold` or `visual_threshold` or something?

crates/egui_kittest/src/snapshot.rs
emilk: It is in contrast with `failed_pixel_count_threshold`, which is expected to increase for larger image tests

emilk:
Or `mean_pixel_error_threshold` ?

lucasmerlin:
That could also work. My worry is that people won't know which of the thresholds to use, and use `broken_pixel_threshold` when the per-pixel threshold would be better. 
With it being just `threshold` it's clear what to try first.
Of course we explain it in the doc comment, but not everyone reads those.

emilk:
Maybe we should call it `noise_floor` ? As: everything below the noise floor is assumed to be noise, and can be safely ignored by the test.

crates/egui_kittest/README.md
```diff
@@ -67,6 +68,11 @@ max_failed_pixels = 0
 The old `failed_pixel_count_threshold` key is still accepted, with a warning;
 rename it to `max_failed_pixels`.
```
lucasmerlin: I don't think we need to mention this in the readme.

lucasmerlin (approved):
Nice!

PR #8385, opened 2026-08-04
Add `uiless_pass` flag to retain state on occluded frames

### LINKED ISSUE #8266: `Area::begin/end` is called when app is occluded / maximized / minimized (causing side effects/flicker)
When the window becomes occluded, we only run `logic` and not `ui`. But `logic` still gets run from within a full egui frame. That means any widget checking e.g. if they were still rendered this frame at the end of frame (e.g. Window, Popup) will think they were hidden / closed. This can cause animations to be replayed and state cleared (popups being closed). 

When the app goes back in foreground they think they are shown again and replay the animation. I think `frame_count` shouldn't increase during logic only frames (maybe we could have a new `logic_count`? 
Also, we need to be careful about what ui things we do during logic frames, I don't think we should call `Area::begin` / `Area::end` for example (the root cause of the window flicker). 

Look mum, no hands (closing a popup by dragging another window / causing occluded state for a moment):

https://github.com/user-attachments/assets/9dadb3f3-8256-470d-9a15-9eb719cfd01c

Flicker during resize / occlude:

https://github.com/user-attachments/assets/ad1c27c8-1aac-467e-96b0-4bfc196fcafb

lucasmerlin:
* Fixes https://github.com/emilk/egui/issues/8266

It's not very elegant, but it works.

### DIFF
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -268,13 +268,9 @@ impl EpiIntegration {
         app: &mut dyn epi::App,
         viewport_ui_cb: Option<&DeferredViewportUiCallback>,
         mut raw_input: egui::RawInput,
-        show_ui: bool,
+        is_visible: bool,
     ) -> egui::FullOutput {
         raw_input.time = Some(self.beginning.elapsed().as_secs_f64());
-
-        // We still run a pass while hidden so that app logic keeps ticking,
-        // but we tell egui to skip all book-keeping that assumes ui was shown:
-        raw_input.uiless_pass = !show_ui;
 
         let close_requested = raw_input.viewport().close_requested();
 
@@ -283,7 +279,7 @@ impl EpiIntegration {
         let full_output = self.egui_ctx.run_ui(raw_input, |ui| {
             if let Some(viewport_ui_cb) = viewport_ui_cb {
                 // Child viewport
-                if show_ui {
+                if is_visible {
                     profiling::scope!("viewport_callback");
                     viewport_ui_cb(ui);
                 }
@@ -293,7 +289,7 @@ impl EpiIntegration {
                     app.logic(ui.ctx(), &mut self.frame);
                 }
 
-                if show_ui {
+                if is_visible {
                     {
                         profiling::scope!("App::ui");
                         app.ui(ui, &mut self.frame);
--- a/crates/eframe/src/native/epi_integration.rs
+++ b/crates/eframe/src/native/epi_integration.rs
@@ -268,9 +268,13 @@ impl EpiIntegration {
         app: &mut dyn epi::App,
         viewport_ui_cb: Option<&DeferredViewportUiCallback>,
         mut raw_input: egui::RawInput,
-        is_visible: bool,
+        show_ui: bool,
     ) -> egui::FullOutput {
         raw_input.time = Some(self.beginning.elapsed().as_secs_f64());
+
+        // We still run a pass while hidden so that app logic keeps ticking,
+        // but we tell egui to skip all book-keeping that assumes ui was shown:
+        raw_input.uiless_pass = !show_ui;
 
         let close_requested = raw_input.viewport().close_requested();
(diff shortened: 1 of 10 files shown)


crates/eframe/src/native/epi_integration.rs
```diff
@@ -268,18 +268,22 @@ impl EpiIntegration {
...
         viewport_ui_cb: Option<&DeferredViewportUiCallback>,
         mut raw_input: egui::RawInput,
-        is_visible: bool,
+        show_ui: bool,
     ) -> egui::FullOutput {
         raw_input.time = Some(self.beginning.elapsed().as_secs_f64());
 
+        // We still run a pass while hidden so that app logic keeps ticking,
+        // but we tell egui to skip all book-keeping that assumes ui was shown:
+        raw_input.uiless_pass = !show_ui;
```
emilk: Could use `ViewportInfo::occluded` instead?

crates/egui/src/data/input/raw_input.rs
```diff
@@ -89,6 +89,20 @@ pub struct RawInput {
...
+    /// but it still wants to run a pass so that background logic keeps ticking
+    /// (e.g. so the app can send a [`crate::ViewportCommand`] to show itself again).
+    ///
+    /// egui then skips all the book-keeping that assumes ui was shown, so that
+    /// widgets don't think they were hidden and replay their appear-animations,
+    /// popups don't close, focus isn't lost, and so on.
+    /// See <https://github.com/emilk/egui/issues/8266>.
+    ///
+    /// If you set this, you must not show any ui during the pass.
+    pub uiless_pass: bool,
```
emilk: no negations in names please

```suggestion
    pub visible: bool,
```

emilk (changes requested):
I don’t think you need a new flag for this - just use `ViewportInfo::occluded`

emilk (commented):
Ok - I've pushed some simplifications. But how come `begin_pass` is even called when invisible? 🤔

emilk:
Oooooh, the call to `app.logic` is wrapped in `self.egui_ctx.run_ui`… not sure we should do that.

It might be a consequence of viewports (`is_viewport_or_descendant_visible`)

lucasmerlin:
> Oooooh, the call to `app.logic` is wrapped in `self.egui_ctx.run_ui`… not sure we should do that

We could drop that, but we'd have to ensure we still handle `ViewportCommand::Focus` and such. This seemed like the least invasive change to fix the bugs

emilk:
I'll quickly explore the different approach (not calling `logic` from within `run_ui`) in a different branch. Calling something called `run_ui` when there is no UI seems like a footgun waiting to happen

crates/egui_kittest/src/lib.rs
```diff
@@ -259,16 +259,18 @@ impl<'a, State> Harness<'a, State> {
...
-                .accesskit_update
-                .take()
-                .expect("AccessKit was disabled"),
-        );
+        if let Some(accesskit_update) = output.platform_output.accesskit_update.take() {
+            self.kittest.update(accesskit_update);
+        } else {
+            // An occluded/minimized viewport shows no ui, so there is no accessibility tree
+            // to update. Keep the tree from the last visible pass.
+            assert!(!is_visible, "AccessKit was disabled");
```
emilk, line 272:
```suggestion
            debug_assert!(!is_visible, "AccessKit was disabled");
```

crates/egui/src/context.rs
```diff
@@ -2718,7 +2768,10 @@ impl ContextImpl {
             }
 
             let is_our_child = parent == ended_viewport_id && id != ViewportId::ROOT;
-            if is_our_child {
+            if is_our_child && is_visible {
+                // A pass without any ui never calls `Context::show_viewport_deferred`,
```
emilk, line 2772: Let's test multiple viewports where the root is minimized

lucasmerlin:
Closing in favor of #8387

PR #8386, opened 2026-08-04
Fix window with a `Grid` being widenable but not shrinkable again

emilk:
## Related

* Fixes a regression from #8152
* Part of #2921

Reported symptom: you can widen the Widget Gallery window, but it won't shrink again.

I'm not sure this fix is the best one, but it does work.

# Claude says
## Cause

A `Grid` gives its **last** column all the available width, so a width-filling widget in it (`Separator`, `TextEdit`, `ProgressBar`, …) makes the `Grid` remember a `col_width` that is really just "however wide we happened to be".

At the start of a resize drag, `Resize` runs a one-frame sizing pass (#8152) to measure the minimum content width and clamps the drag against it. But `GridLayout::next_cell` inflated every cell to `prev_state.col_width`, so the `Grid` reported its previous width as its minimum — even though it was only offered `min_size.x`. The clamp is a lower bound, so widening kept working while shrinking was blocked at the widened width.

## Fix

During an enclosing sizing pass, don't inflate the stretchy last column to its remembered width, and don't store the measured (narrow) widths.

Minimal repro (fails before, passes after — added as a regression test):

```rust
Window::new("x").default_width(280.0).show(ctx, |ui| {
    egui::Grid::new("grid").num_columns(2).show(ui, |ui| {
        ui.label("Separator");
        ui.separator(); // fills the last column
        ui.end_row();
    });
});
```

`Panel` is unaffected — it clamps only against the user's `min_size`, with no content-min sizing pass.

### DIFF
--- a/crates/egui/src/grid.rs
+++ b/crates/egui/src/grid.rs
@@ -94,11 +94,14 @@ pub(crate) struct GridLayout {
 
 impl GridLayout {
     pub(crate) fn new(ui: &Ui, id: Id, prev_state: Option<State>) -> Self {
-        // An enclosing sizing pass (e.g. `Resize` measuring our minimum width) wants to
-        // know how small we can get.
-        let in_outer_sizing_pass = ui.is_sizing_pass() && prev_state.is_some();
-
         let is_first_frame = prev_state.is_none();
+
+        // An enclosing sizing pass (e.g. `Resize` measuring our minimum width) wants to know
+        // how small we can get. Note that on our first frame we run a sizing pass of our own
+        // (see `show_dyn`), which is a different situation: there we have no remembered sizes
+        // to protect, and must in fact produce some.
+        let in_outer_sizing_pass = !is_first_frame && ui.is_sizing_pass();
+
         let prev_state = prev_state.unwrap_or_default();
 
         // TODO(emilk): respect current layout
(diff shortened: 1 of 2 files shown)


crates/egui/src/grid.rs
```diff
@@ -90,6 +95,13 @@ pub(crate) struct GridLayout {
 impl GridLayout {
     pub(crate) fn new(ui: &Ui, id: Id, prev_state: Option<State>) -> Self {
         let is_first_frame = prev_state.is_none();
+
+        // An enclosing sizing pass (e.g. `Resize` measuring our minimum width) wants to know
+        // how small we can get. Note that on our first frame we run a sizing pass of our own
+        // (see `show_dyn`), which is a different situation: there we have no remembered sizes
+        // to protect, and must in fact produce some.
+        let in_outer_sizing_pass = !is_first_frame && ui.is_sizing_pass();
+
```
lucasmerlin, line 101: I don't get why we should differentiate between outer and inner sizing pass here. I think we can have the same logic in both cases?
Having special code for the inner path is weird because it means you can't manually retrigger a sizing pass when e.g. contents change from the outside.

lucasmerlin (approved):
I've simplified the logic, tests still pass and window still shrinks

emilk:
much better - thanks!

Issue #7686, opened 2025-11-06
Main viewport blocked from updates while deferred viewports are rendering.


hydra:
**Describe the bug**
When the mouse pointer is being moved over a deferred viewport the main viewport does not update.

**To Reproduce**
1. run the updated example in PR #7676 (source: https://github.com/hydra/egui/blob/improve-deferred-viewport-example-1/examples/multiple_viewports/src/main.rs)
2. observe that when the mouse pointer is being moved over the deferred viewport, the main viewport does not update, even though there's a background thread requesting a refresh of it via the context. `request_repaint` or `request_repaint_of(ViewportId::ROOT)`.
3. observe that when the mouse pointer is being moved over the deferred viewport, the main viewport does not update, even though the main thread is using `request_repaint_after(duration)`

**Expected behavior**
Main viewport is not blocked from updating.

**Video**
[![Video of the bug](https://img.youtube.com/vi/hj1myK0yVDQ/0.jpg)](https://www.youtube.com/watch?v=hj1myK0yVDQ)
https://youtu.be/hj1myK0yVDQ

**Desktop (please complete the following information):**
 - OS: Windows 11 24H2

Issue #4945, opened 2024-08-11
Animation with request_repaint() does not work when deferred viewport is open


Barafu:
**Describe the bug**
When multiple viewports are open, animations using request_repaint() work only on the active viewport, and requests to update other viewports are ignored. 

**To Reproduce**
I have created a minimal demonstration, which is included below. Compile and run. See how animation works only on window that is clicked. 
Replace show_viewport_deferred() on line 41 to show_viewport_immediate() to see expected behaviour.

**Expected behavior**
Animation to work identically in both windows. 

**Demonstration**
```rust
use egui::{Color32, Painter, Shape, Ui, ViewportId};

fn main() -> eframe::Result {
    let native_options = eframe::NativeOptions {
        viewport: egui::ViewportBuilder::default()
            .with_position([0.0, 0.0])
            .with_inner_size([500.0, 500.0]),
        ..Default::default()
    };
    return eframe::run_native(
        "My egui App",
        native_options,
        Box::new(|_cc| Ok(Box::new(MyApp::default()))),
    );
}

#[derive(Default)]
pub struct MyApp {
    fullscreen_onetime: bool,
}

impl eframe::App for MyApp {
    fn update(&mut self, ctx: &egui::Context, _frame: &mut eframe::Frame) {
        if !self.fullscreen_onetime {
            self.fullscreen_onetime = true;
        }

        egui::CentralPanel::default().show(ctx, |ui| {
            paint_animation(ui);
            let second_id = second_viewport(ui);
            ui.ctx().request_repaint_of(second_id);
            ui.ctx().request_repaint();
        });
    }
}

pub fn second_viewport(ui: &mut egui::Ui) -> ViewportId {
    let title = "second viewport";
    let id: ViewportId = ViewportId::from_hash_of(title);

    ui.ctx().show_viewport_deferred(
        id.clone(),
        egui::ViewportBuilder::default()
            .with_title(title)
            .with_position([510.0, 0.0])
            .with_inner_size([500.0, 500.0]),
        move |ctx, _class| {
            egui::CentralPanel::default().show(ctx, |ui| {
                paint_animation(ui);
                ui.ctx().request_repaint();
            });
        },
    );

    id
}

pub fn paint_animation(ui: &mut Ui) {
    let painter = Painter::new(
        ui.ctx().clone(),
        ui.layer_id(),
        ui.available_rect_before_wrap(),
    );

    let step = std::time::SystemTime::now()
        .duration_since(std::time::UNIX_EPOCH)
[...]

PR #8395, opened 2026-08-06
Fix: deferred viewport does not redraw when clicking ROOT viewport (Windows)

### LINKED ISSUE #8351: `deferred viewport` does not redraw when clicking `ROOT viewport` (Windows)

rustbasic:
Fix: deferred viewport does not redraw when clicking ROOT viewport (Windows)

* Related #4945
* Related #7686

* Closes #8351 

This PR fixes the bug where the deferred viewport fails to redraw upon clicking the ROOT viewport.
Since it is unclear whether this behavior also occurs on other platforms, I limited the scope to Windows with cfg!(target_os = "windows").


rustbasic:
If there are multiple 'deferred viewport', additional work is required because selecting a 'deferred viewport' prevents other viewports from being rendered.
There is an improvement effect even with the current content alone.

emilk:
eframe intentionally only repaints viewports that request it:

https://github.com/emilk/egui/issues/8351#issuecomment-5410203480

PR #8400, opened 2026-08-08
Fallback `window_title_frame` to `window_frame` when unspecified

rustbasic:
### Summary
When `title_frame` is not explicitly set, fall back to `window_frame` instead of `Frame::window(&style)`.

### Motivation
If a custom `frame` is provided for a window, `window_title_frame` should maintain visual consistency with it by default unless a separate `title_frame` is specified.

* Related #8154 
* Related #8353


emilk:
Makes sense - thanks!

PR #8402, opened 2026-08-08
Fix: Refresh Windows title bar after theme changes

rustbasic:
Fix: Refresh Windows title bar after theme changes

**Description**

On Windows, changing the window theme updates egui immediately, but the native title bar may keep its previous colors until the window activation state changes.

This sends WM_NCACTIVATE after applying ViewportCommand::SetTheme, forcing Windows to refresh the non-client title bar state without changing the real focus.

* Related #8299


lucasmerlin:
Hi! I gave this a try on my windows 10 laptop and there window decoration is the same while the window is focused and light/dark only looks different when the window is in the background.

So I wonder if this is needed at all? Or is it different in win 11 and focused light/dark looks different?

Example:
Focused Light/dark:
grafik
grafik

Unfocused light/dark:
grafik
grafik

rustbasic:
@lucasmerlin 

On Windows 10, the window decoration fails to redraw immediately upon a theme change; it requires losing and regaining window focus to trigger a redraw (which is why it might have appeared to only change when in the background). 

Fixing this ensures consistent, immediate title bar updates, which improves the overall polish and user experience in egui.

Regarding child viewports, they don't update immediately yet, but I plan to address that in a follow-up PR once this one is merged.

lucasmerlin:
But I don't see any change at all between light and dark theme when the window is focused. Not without or with your fix.
Could you provide screenshots or a video how this fixes it for you?

rustbasic:
@lucasmerlin 

before:
ezchat 2026-08-20 오후 9_31_06
before:
ezchat 2026-08-20 오후 9_32_06
after:
ezchat 2026-08-20 오후 9_43_17
after:
ezchat 2026-08-20 오후 9_43_08

Screen recording tools do not capture the native OS title bar, so I'm explaining the issue with screenshots instead.

### Before this PR (Current behavior / Bug):
The title bar does not update immediately when changing the theme:
* When selecting **Dark**, the title bar remains in **Light** mode.
* When selecting **Light**, the title bar remains in **Dark** mode.
* **When focused**: It only updates after switching focus to another application and returning to this window.
* **When unfocused**: It only updates when clicking on the window to regain focus.

---

### With this PR (Fixed):
* The title bar immediately updates to match the selected theme (Dark / Light) without requiring any window focus changes.

lucasmerlin (approved):
Sounds good to me then. I guess ideally this should be fixed in winit, but I have no idea when that will happen, so lets fix it here.

jdiaz97:
I have a bug related to https://github.com/emilk/egui/pull/8299 so I'll mention it here:, in egui 0.36.1 and windows 11

I have two screens:

In screen 2:
image

After I turn off my screen 2 and every app moves to Screen 1, so now in Screen 1:

image

PR #8403, opened 2026-08-08
Don't revert external changes to a focused `DragValue`

### LINKED ISSUE #8339: A selected `DragValue` will revert external changes to its referenced value after one frame
**Describe the bug**
If a selected `DragValue` has its value modified externally, it will revert the value after one frame.
For example, a button which increments its value will increment it for only one frame.
A minimal example of this can be found below.

```rust
struct App {
	val: i32
}

impl eframe::App for App {
		ui.add(egui::DragValue::new(&mut self.val));
		if ui.button("increment").clicked() {
			self.val += 1;
		}
	}
}

fn main() -> eframe::Result {
	eframe::run_native(
		"test",
		Box::new(|_cc| Ok(Box::new(App {val: 0})))
	)
}
```

**To Reproduce**
1. Create a `DragValue` editing a value, and another widget which can edit the same value.
2. Select the `DragValue`
3. Use the other widget to edit this value

**Expected behavior**
The value will be modified by the second widget, and stay this way.

 - OS: Windows 11
 - Browser: not used, compiled natively
 - egui version: 0.35

**Additional context**
This seems to be caused by the code block at
https://github.com/emilk/egui/blob/main/crates/egui/src/widgets/drag_value.rs#L540

Removing this seems to resolve the issue without causing any other issues.
I'm not sure what the point of this code block is as it seems to do conceptually do the same as:
https://github.com/emilk/egui/blob/main/crates/egui/src/widgets/drag_value.rs#L582

I expected this to be related to the functionality of `update_while_editing`. It reads as if, if `update_while_editing` is false, we should not update our value if escape is pressed. Instead, we do update our value after reverting to initial value for one frame. Removing the code block at line 540 also seems to resolve this issue.

Looking through the commit history I could not find the logic for the inclusion of this code, does anyone know the purpose of it or any reason it should not be removed?

teddytennant:
* Closes <https://github.com/emilk/egui/issues/8339>

## The bug

While a `DragValue` has focus it is rendered as a `TextEdit`, and the text being
edited is stored in `Memory::data` between frames. That is needed so that
half-finished input such as `"1."` or `"-"` isn't thrown away just because it
doesn't parse to the current value.

The stored text was only discarded when the widget *gained* focus or when the
widget itself changed the value. If something else changed the value while the
`DragValue` was focused, the stored text was kept, shown to the user, and
written back to the value when focus was lost — silently undoing the external
change:

```rust
ui.add(egui::DragValue::new(&mut self.value));
if ui.button("increment").clicked() {
    self.value += 1;
}
```

Click into the `DragValue` so it has focus, then press "increment": the value
goes up for one frame and then snaps back. `Slider` shows the same behaviour,
since it uses a `DragValue` for its value field.

## The fix

Store the value the text belongs to next to the text, and discard the text when
the value no longer matches it. The remembered value is read back from the
get/set closure *after* the widget has applied its own edits, so a change the
widget made itself never looks like an external one — this matters for values
that can't represent what was typed, e.g. `"12.5"` in a `DragValue<i32>`.

This keeps the reason the text is stored in the first place intact: as long as
nothing else touches the value, the text the user is typing is preserved
verbatim.

## Tests

Three tests in `crates/egui_kittest/tests/regression_tests.rs`:

* `drag_value_should_not_revert_external_changes_while_focused` — the actual
  regression. Fails on `main`:

  ```
  ---- drag_value_should_not_revert_external_changes_while_focused stdout ----
    left: Some("0")
   right: Some("42")
  ```

  and, with the display assertion removed so the test reaches the blur, on the
  value itself:

  ```
    left: 0
   right: 42
  ```

* `drag_value_should_keep_text_while_typing` and
  `drag_value_should_keep_text_the_value_cannot_represent` — guards for the
  behaviour the stored text exists for. Both pass on `main` and after the fix,
  and both fail if the text is re-read from the value too eagerly.

`cargo test -p egui_kittest` and `cargo test -p egui` pass, as do
`cargo fmt --all --check`, `scripts/lint.py` and
`cargo clippy -p egui -p egui_kittest --all-targets --all-features -- -D warnings`.

## Not changed

`DragValue` still ignores the stored text when <kbd>Escape</kbd> is pressed, and
`update_while_editing` still decides when typed text is applied — neither is
touched here.

### DIFF
--- a/crates/egui/src/widgets/drag_value.rs
+++ b/crates/egui/src/widgets/drag_value.rs
@@ -25,6 +25,24 @@ fn set(get_set_value: &mut GetSetValue<'_>, value: f64) {
     (get_set_value)(Some(value));
 }
 
+// ----------------------------------------------------------------------------
+
+/// What the user has typed into a [`DragValue`] that is being edited as text.
+///
+/// Stored in [`crate::Memory::data`] between frames, because the text can be
+/// something that doesn't (yet) parse to a number, e.g. `"1."` or `"-"`.
+#[derive(Clone, Default)]
+struct EditState {
+    /// The text the user is editing.
+    text: String,
+
+    /// The value of the [`DragValue`] the last time we stored `text`.
+    ///
+    /// If the value has changed since then it was changed by something other than
+    /// this widget, and `text` is stale and must not be written back to the value.
+    value: f64,
+}
+
 /// A numeric value that you can change by dragging the number. More compact than a [`crate::Slider`].
 ///
 /// ```
@@ -466,7 +484,7 @@ impl Widget for DragValue<'_> {
             });
 
         if ui.memory_mut(|mem| mem.gained_focus(id)) {
-            ui.data_mut(|data| data.remove::<String>(id));
+            ui.data_mut(|data| data.remove::<EditState>(id));
         }
 
         let old_value = get(&mut get_set_value);
@@ -524,7 +542,7 @@ impl Widget for DragValue<'_> {
 
         if old_value != value {
             set(&mut get_set_value, value);
-            ui.data_mut(|data| data.remove::<String>(id));
+            ui.data_mut(|data| data.remove::<EditState>(id));
         }
 
         let value_text = match custom_formatter {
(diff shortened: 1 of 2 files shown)


emilk (approved):
Makes sense - thanks!

I guess it means you can never edit a value that is changing 🤔  Which… I don't know why you would want to.
A good example would be some sort of collaborative editing of the same value (e.g. synced over network). If someone else edits the value while you are also editing, who should win? This PR changes the answer to "them", but 🤷

PR #8405, opened 2026-08-10
eframe: route DeviceEvent::MouseMotion to the root viewport

Le-Syl21:
`DeviceEvent::MouseMotion` is a device-level relative-motion event: it carries no surface association, so there is no windowing-system answer to "which viewport does this belong to". eframe currently sends it to `focused_viewport` and drops it when nothing is focused, which makes raw mouse deltas unusable in any multi-viewport app on Wayland.

I hit this in a pinball-cabinet front-end: one fullscreen playfield window plus a few deferred viewports covering the backglass and DMD screens, all created with `with_active(false)`. Under Mutter, `with_active(false)` is a no-op — a cover viewport takes the focus as it maps, and the playfield never gets it back, because `ViewportCommand::Focus` is exactly what focus-stealing prevention refuses (I measured five refusals in a row before giving up). From then on `focused_viewport` points at a cover, or at nothing, and every `MouseMotion` is silently discarded. The app has an on-screen software cursor driven by those deltas; it simply stopped moving.

Two changes, both restricted to relative motion:

* Route `MouseMotion` to `ViewportId::ROOT` in both the glow and wgpu integrations. Single-window apps always have ROOT as the de-facto target, so this is a no-op for them; multi-viewport apps get the deltas somewhere useful instead of nowhere.
* Drop the `!has_focus() && !any_pointer_button_down` and `!is_pointer_in_window() && !any_pointer_button_down` guards on the same path. Under `CursorGrab::Locked` the OS never fires `CursorMoved` — the pointer-constraints lock delivers relative deltas only — so `pointer_pos_in_points` stays `None` forever and `is_pointer_in_window()` is permanently false, which drops every event the app was locking the pointer *for*. Consumers gate on their own state, so forwarding more is harmless.

Absolute pointer events (`CursorMoved`, buttons) are untouched: those do have a surface, and their existing routing is correct.

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -764,11 +764,18 @@ impl State {
     }
 
     /// Returns `true` if the event was sent to egui.
+    ///
+    /// Forwards raw mouse delta unconditionally. The previous guard
+    /// (`is_pointer_in_window() || any_pointer_button_down`) was an
+    /// over-restriction: under `CursorGrab::Locked` on Wayland the OS never
+    /// fires `WindowEvent::CursorMoved` (the pointer-constraints "lock"
+    /// mode delivers only relative deltas via `DeviceEvent::MouseMotion`),
+    /// so `pointer_pos_in_points` stays `None` forever and
+    /// `is_pointer_in_window()` always returns `false` — silently dropping
+    /// every relative-motion event. Downstream consumers (egui's standard
+    /// pointer system, egui-rotate's `SoftwareCursor`) gate themselves on
+    /// their own state, so over-forwarding is harmless.
     pub fn on_mouse_motion(&mut self, delta: (f64, f64)) -> bool {
-        if !self.is_pointer_in_window() && !self.any_pointer_button_down {
-            return false;
-        }
-
         self.egui_input.events.push(egui::Event::MouseMoved(Vec2 {
             x: delta.0 as f32,
             y: delta.1 as f32,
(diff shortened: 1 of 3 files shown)


Le-Syl21:
Checked against today's `main`: merges cleanly, and `device_event` in both the glow and wgpu integrations still routes `MouseMotion` through `focused_viewport`, so the behaviour this changes is unchanged upstream.

Under 40 lines either side, no API change. Happy to rebase or split it if that makes it easier to look at.

crates/egui-winit/src/lib.rs
```diff
@@ -764,11 +764,18 @@ impl State {
...
+    /// Forwards raw mouse delta unconditionally. The previous guard
+    /// (`is_pointer_in_window() || any_pointer_button_down`) was an
+    /// over-restriction: under `CursorGrab::Locked` on Wayland the OS never
+    /// fires `WindowEvent::CursorMoved` (the pointer-constraints "lock"
+    /// mode delivers only relative deltas via `DeviceEvent::MouseMotion`),
+    /// so `pointer_pos_in_points` stays `None` forever and
+    /// `is_pointer_in_window()` always returns `false` — silently dropping
+    /// every relative-motion event. Downstream consumers (egui's standard
+    /// pointer system, egui-rotate's `SoftwareCursor`) gate themselves on
+    /// their own state, so over-forwarding is harmless.
```
lucasmerlin: We shouldn't leave comments regarding code that isn't there anymore
```suggestion
    ///
    /// Forwards raw mouse delta unconditionally (even if the window isn't focused).
```

Issue #8142, opened 2026-05-03
Add ViewportCommand::RequestActivationToken to support cross-window focus on Wayland


jd1378:
notice: This is a real request made by me for a real issue, the content is AI assisted but proof-read and fact-checked manually.

## Summary                                                                                                                                                    
   
  On Wayland, programmatically focusing or raising a window requires an xdg_activation_v1 token obtained from a recent user input event. Without a token, the
   compositor silently ignores activation requests. This is a deliberate focus-stealing prevention measure baked into the protocol.
                                                                                                                                                             
  egui-winit currently exposes ViewportCommand::Focus, which calls winit::window::Window::focus_window(). On Wayland, that call is a no-op unless the window 
  already holds an activation token, and egui-winit provides no way to request or supply one. As a result, multi-window egui apps cannot bring a sibling
  window to the foreground on Wayland, even in response to a direct user gesture in another window of the same app.                                          
           
##  Reproduction                                                                                                                                               
   
  A multi-process or multi-window egui app on Wayland (tested on KDE):                                                                     
           
  1. Window A (focused) sends an IPC message to window B asking it to surface.                                                                               
  2. Window B calls ctx.send_viewport_cmd(ViewportCommand::Focus).                                                                                                    
  3. Window B does not come to the foreground. Toggling WindowLevel::AlwaysOnTop then Normal, and Visible(false) then Visible(true), are also no-ops.
                                                                                                                                                             
  RequestUserAttention(Critical) works (urgency hint / taskbar flash) but does not actually raise the window.                                                
                                                                                                                                                             
##  Relevant code                                                                                                                                               
   
  window.focus_window() on Wayland is implemented as xdg_activation_v1::activate(token, surface) and requires the caller to have already obtained a token.   
  winit exposes the token-request side via:
                                                                                                                                                             
  - winit::window::Window::request_activation_token, which returns a Result<AsyncRequestSerial, NotSupportedError>.                                          
  - The token is delivered asynchronously through WindowEvent::ActivationTokenDone { serial, token }.
                                                                                                                                                             
  egui-winit already observes this event but discards it ([egui-winit/src/lib.rs:497](https://github.com/emilk/egui/blob/fe8b1edfc603ac511c7ea8b0a19e5e3218f5928c/crates/egui-winit/src/lib.rs#L497)):                                                                                                                              
                                                                                                                                                             
  There is no public surface for either requesting a token or supplying an externally-obtained one back to Focus.
                                                                                                                                                             
[...]

PR #8406, opened 2026-08-10
eframe: use the activation token from the environment on window creation

Le-Syl21:
- related #8282 
- part of https://github.com/emilk/egui/issues/8142

A desktop entry with `StartupNotify=true` hands the launched process an activation token through `XDG_ACTIVATION_TOKEN` (Wayland) or `DESKTOP_STARTUP_ID` (X11). winit can apply one, but only at window creation, via `WindowAttributesExtStartupNotify::with_activation_token`. eframe never reads it, so the token our launcher went to the trouble of issuing is thrown away.

winit's own docs are blunt about the consequence:

> The `ActivationToken` is essential to ensure that your newly created window will obtain the focus, otherwise the user could be required to click on the window.

That is what I ran into: under Mutter the window opens unfocused and cannot recover, since `ViewportCommand::Focus` is precisely the request focus-stealing prevention blocks. On a kiosk-style setup the cost goes past a missing title-bar highlight — a compositor only grants a pointer constraint to a focused surface, so `CursorGrab` silently does nothing too.

This reads the variable when building the window attributes and passes it to winit. The environment is cleared afterwards through winit's `reset_activation_token_env()`, per the startup-notification spec: a token is single-use, and leaving it around would have every later viewport, and every child process, replay it.

Linux only, behind the existing `wayland`/`x11` features; a no-op when the variables are absent.

### DIFF
--- a/crates/egui-winit/src/lib.rs
+++ b/crates/egui-winit/src/lib.rs
@@ -2135,6 +2135,37 @@ pub fn create_winit_window_attributes(
         window_attributes = window_attributes.with_name(app_id, "");
     }
 
+    // Consume the activation token our launcher handed us, so the first
+    // window actually gets the focus.
+    //
+    // A desktop entry with `StartupNotify=true` passes a token through
+    // `XDG_ACTIVATION_TOKEN` (Wayland) or `DESKTOP_STARTUP_ID` (X11), and
+    // winit can only apply it at window creation. eframe never read it, so
+    // under a compositor that enforces focus-stealing prevention the window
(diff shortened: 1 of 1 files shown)


Le-Syl21:
Checked against today's `main`: merges cleanly, and nothing in `egui-winit` reads `XDG_ACTIVATION_TOKEN` or `DESKTOP_STARTUP_ID` yet, so the gap is still open.

31 lines, purely additive — a window created from a launcher keeps the focus the compositor already granted it, instead of opening behind.

lucasmerlin (approved):
Thanks!

PR #8407, opened 2026-08-10
Allow explicit popup sizing passes

yay:

## Summary

- Add an opt-in `Popup::sizing_pass(bool)` builder for remeasuring a popup whose contents change while it remains open.
- Preserve the automatic first-open/reopen sizing pass and all existing default behavior.
- Add a headless regression covering growth to a capped scroll viewport with overflowing content remaining scrollable.

## Why this is necessary

A continuously open popup can first shrink around a short result set and later receive more content, such as an autocomplete after its query changes or a “show more” action. The cached `Area` height constrains the `ScrollArea` input size, so `ScrollArea::max_height` can cap the viewport but cannot make the containing popup grow again.

PR #8315 taught `Popup` to rerun its sizing pass after closing and reopening. That fixes the same cached-size feedback loop when `was_open_last_frame` is false, but a continuously open popup keeps that value true while its contents change. In that case the caller is the component that knows the cached natural size is stale.

This API exposes the existing one-frame `Area` sizing mechanism through `Popup`. It is additive, defaults to false, and combines with the automatic reopen pass, so unrelated popups, menus, tooltips, and areas keep their current behavior.

### DIFF
--- a/crates/egui/src/containers/popup.rs
+++ b/crates/egui/src/containers/popup.rs
@@ -179,6 +179,7 @@ pub struct Popup<'a> {
 
     /// Default width passed to the Area
     width: Option<f32>,
+    sizing_pass: bool,
     sense: Sense,
     interactable: bool,
     layout: Layout,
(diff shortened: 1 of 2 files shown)


emilk:
Related to
* https://github.com/emilk/egui/issues/5138
* https://github.com/emilk/egui/issues/5762

emilk (approved):

PR #8409, opened 2026-08-11
Give every viewport an AccessKit adapter, not just the root

DaikiSuganuma:
`init_accesskit` is only ever called for `ViewportId::ROOT`. An application whose real windows are deferred child viewports therefore exposes nothing to assistive technology inside them: a screen reader or a UI automation client sees a window with no contents.

I hit this in a Windows tray application, where the root viewport is a 1×1 invisible window and every real window is a deferred child viewport.

| measured with a UI Automation client (Windows, glow) | descendants in the UIA tree |
|---|---|
| widgets in a deferred child viewport, before | **0** |
| the same widgets in a root viewport | 9 |

## What the change does

`initialize_window` now creates an adapter for any viewport that gets a window.

Two details that are not obvious from the diff:

- **The context has to hold an `EventLoopProxy`.** `ActiveEventLoop` cannot make one and the adapter needs it, so it is stored and filled in after construction — the root's window is created inside `new`, before a caller can hand one over.
- **Windows other than the root are held back until the adapter is attached.** `accesskit_winit` panics if the adapter is created after the window has been shown, and the root is the only window eframe already creates hidden (the white-flash fix of #2279). They are revealed as soon as the adapter is in place.

## The wgpu half is not verified at runtime

The measurement above is glow. I have no wgpu application with deferred child viewports to point a UI automation client at, so **the second commit is compile-checked only**: `--features wgpu,accesskit`, `--features wgpu` alone, and `clippy --all-features`. It is the same shape as the glow change, with one difference — the wgpu path creates its window through `egui_winit::create_window`, which also resolves the target monitor, so the "start hidden" flag is set on the `ViewportBuilder` rather than on the `WindowAttributes` it produces.

I would rather send it unverified than keep it out of tree. This has been carried as a patch for two releases of my own project, and the alternative to sending it is carrying it for two more.

**If you would prefer the glow commit only, it stands alone** — the two commits are independent, and dropping the second one leaves a working PR.

## Checks

`cargo fmt --all --check`, `cargo clippy -p eframe --all-targets --all-features -- -D warnings`, `cargo doc -p eframe --lib --all-features` (all with `-D warnings`), and `cargo check -p eframe --no-default-features` for `glow`, `wgpu` and `glow,wgpu`, each with and without `accesskit`.

I could not run `./scripts/check.sh` in full on Windows — the wasm targets, the macOS-rendered snapshots and `typos` are not available here — which is why this is a draft.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ This file is updated upon each release.
 Changes since the last release can be found at <https://github.com/emilk/egui/compare/latest...HEAD> or by running the `scripts/generate_changelog.py` script.
 
 
+## 0.36.1 - 2026-08-07
+* Fix `Sense::drag` detecting drags when clicking widget above it [#8396](https://github.com/emilk/egui/pull/8396) by [@lucasmerlin](https://github.com/lucasmerlin)
+
+
 ## 0.36.0 - 2026-08-05
 
 ### Highlights ✨
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1243,7 +1243,7 @@ checksum = "d8b14ccef22fc6f5a8f4d7d768562a182c04ce9a3b3157b91390b52ddfdf1a76"
 
 [[package]]
 name = "ecolor"
-version = "0.36.0"
+version = "0.36.1"
 dependencies = [
  "bytemuck",
  "cint",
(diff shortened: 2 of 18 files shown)


DaikiSuganuma:
Sorry for the noise — closing and reopening from a branch based on `main`.

I branched this from the `0.36.1` tag, which is not an ancestor of `main`, so the diff carried the release commit and an unrelated `Sense::drag` fix along with it. The replacement has the same two commits and nothing else.

PR #8412, opened 2026-08-11
Replace constant divisions with constant multiplications

oscargus:

Compiler optimization will replace these later on, but when building without optimization they will (most often) stay divisions and divisions are clearly slower than multiplications.

Note that the `/ 2.5` replacement is not bit-exact (since 0.4 cannot be represented exactly, but 2.5 can), but my guess is that the approximation is good enough. The others should be bit-exact.

(Division by 2 for unsigned numbers will be replaced to >>1 earlier in the compile chain, so no point in doing the same there.)


lucasmerlin (approved):
Is performance in debug builds an issue for you? In the egui projects I'm working on debug builds run fine (I even use [cranelift](https://github.com/rust-lang/rustc_codegen_cranelift) for even faster compile speeds, and even there apps remain usable).

Would be awesome if you could run the benchmarks (as a debug build), I'm curious if this will have a measurable impact.

oscargus:
Not really an issue and I doubt that it is measurable. More like "this is sometimes marginally better, so one can just go with it". (Triggered by being bored rather than an actual issue.)

emilk (approved):

PR #8413, opened 2026-08-11
Optimize CubicBezierShape::find_cross_t

oscargus:

Factor out common constants to reduce the number of multiplications. During checking the results (comparing my hand-optimized code with the previous version using AI), the AI pointed out that d is always 0, so I removed that as well.

It is possible to rewrite the equations a bit further, like: 
`let p = (a / c - h * h);`
but not sure that it will be beneficial (on superscalar machines) due to introduced data dependencies.

I also have a doubt what will happen when p = 0. I think it may have to be special cased as p = 0 => r = 0 => division by zero (before taking acos).


emilk (commented):
I really like this! But I also asked Claude to take a look, and it suggested this even simpler version:

```
  pub fn find_cross_t(&self, epsilon: f32) -> Option<f32> {
      let [p0, p1, p2, p3] = self.points;

      let a = (p3.x - 3.0 * p2.x + 3.0 * p1.x - p0.x) * (p3.y - p0.y)
          - (p3.y - 3.0 * p2.y + 3.0 * p1.y - p0.y) * (p3.x - p0.x);
      let c = 3.0 * ((p1.x - p0.x) * (p3.y - p0.y) - (p1.y - p0.y) * (p3.x - p0.x));

      // t=0 and t=1 are always roots, so the cubic is a*t*(t-1)*(t - c/a):
      let t = c / a;
      (epsilon < t && t < 1.0 - epsilon).then_some(t)
  }
```

PR #8414, opened 2026-08-11
Keep app logic running when a redraw is never delivered

### LINKED ISSUE #5136: Wayland: `request_repaint` ignored if minimized

42Pupusas:
On Wayland a compositor may withhold `wl_surface.frame` for a surface it considers hidden. winit then suppresses `RedrawRequested` until the callback arrives, and the gate never re-arms on its own, so `App::update` and `App::logic` stop forever - even after the window is shown again.

eframe cannot detect this on Wayland: `Window::is_visible`, `Window::is_minimized`, `ViewportInfo::occluded` and `ViewportInfo::minimized` all return `None` there, so `is_invisible_or_minimized` is `false` and the existing invisible-window path never runs.

Add a `FrameWatchdog` that records each `request_redraw` and reports the windows whose `RedrawRequested` never arrived within 500ms. Those windows run a logic-only egui pass, so the app keeps ticking and can still react (e.g. ask to be shown again), without painting a frame the compositor would not display.

* Closes #5136 

### DIFF
--- a/crates/eframe/src/native/frame_watchdog.rs
+++ b/crates/eframe/src/native/frame_watchdog.rs
@@ -1,0 +1,172 @@
+use core::time::Duration;
+use std::time::Instant;
+
+use ahash::HashMap;
+use winit::window::WindowId;
+
+/// Detects redraws that were requested but never delivered.
+///
+/// On Wayland a compositor may withhold `wl_surface.frame` for a hidden
+/// surface. winit then suppresses `RedrawRequested` until the callback
+/// arrives, and the gate does not re-arm on its own, so `App::update` stops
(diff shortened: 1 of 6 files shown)


42Pupusas:
@emilk keeping as draft for now because I am not sure about the 500ms fix. maybe there is a better way than polling but I have not figured it out

Issue #7140, opened 2025-06-13
Base TextEdit on AtomLayout


lucasmerlin:
To allow adding icons / text into the text edit frame.

PR #8420, opened 2026-08-14
Properly respect `min_size` in `TextEdit`

lucasmerlin:
TextEdit has a min_size field but we completely ignored in the y direciton. (I probably broke it in #7140).


tests/egui_tests/tests/snapshots/visuals/drag_value.png
emilk, line 1: "focussed" typo 😆

emilk (approved):

PR #8427, opened 2026-08-17
Simplify and optimize `Color32::from_rgba_unmultiplied`

Keith-Cancel:
So I first noticed that this function was using a large lookup table behind a OnceLock. I initially thought about just making it a const, but when looking at things further. I realized it could be made much simpler.

If we just treated the numbers as fixed point we can get rid of any of the floating point calculations and especially divisions.  You can see how efficiently this can compile down here: https://llvm.godbolt.org/z/K83jEjvdq

You can see all three versions here: https://godbolt.org/z/nWc1as1nq
* First one is basically the original essentially being called by:  `ColorImage::from_rgba_unmultiplied()`
* Second is the const Lookup table instead of the OnceLock and runtime generation.
* Third is the fixed point implementation.

At least looking at the bytes reported compiler explorer the OnceLock and the const Table results are in similar size, and the const table is surprisingly smaller when I compile to a binary object in compiler explorer. However though the oncelock is producing a lot SIMD instructions for initialization so I guess not too surprised. The fixed point math is much smaller than both. 

The const table is probably faster, but does bloat the binary images, and again when it's this fast to compute: https://llvm.godbolt.org/z/K83jEjvdq I am not sure the extra bytes are worth it.

Next, what I did was merge `from_rgba_unmultiplied` and `from_rgba_unmultiplied_const`. Moreover with the fixed point math the `from_rgba_unmultiplied_const` is probably not necessary anymore, but it's part of the public API so I left it. Lastly, I just added a sanity test to make sure the math checks out which it does. You can even sweep the 2^16 inputs to be sure.

### DIFF
--- a/crates/ecolor/src/color32.rs
+++ b/crates/ecolor/src/color32.rs
@@ -1,4 +1,4 @@
-use crate::{Rgba, fast_round, linear_f32_from_linear_u8};
+use crate::{Rgba, fast_round, mul_frac_round};
 
 /// This format is used for space-efficient color representation (32 bits).
 ///
@@ -131,30 +131,7 @@ impl Color32 {
     /// but for transparent colors what you get back might be slightly different (rounding errors).
     #[inline]
     pub fn from_rgba_unmultiplied(r: u8, g: u8, b: u8, a: u8) -> Self {
-        const LOOKUP_TABLE: &[u8] = &const {
-            const LEN: usize = u16::MAX as usize + 1;
-            let mut tbl = [0u8; LEN];
-            let mut i = 0;
-            while i < LEN {
-                let [value, alpha] = (i as u16).to_ne_bytes();
-                tbl[i] = fast_round(value as f32 * linear_f32_from_linear_u8(alpha));
-                i += 1;
-            }
-            tbl
-        };
-        match a {
-            // common-case optimization:
-            0 => Self::TRANSPARENT,
-
-            // common-case optimization:
-            255 => Self::from_rgb(r, g, b),
-
-            a => {
-                let [r, g, b] = [r, g, b]
-                    .map(|value| LOOKUP_TABLE[usize::from(u16::from_ne_bytes([value, a]))]);
-                Self::from_rgba_premultiplied(r, g, b, a)
-            }
-        }
+        Self::from_rgba_unmultiplied_const(r, g, b, a)
     }
 
     /// Same as [`Self::from_rgba_unmultiplied`], but can be used in a const context.
@@ -166,13 +143,16 @@ impl Color32 {
             // common-case optimization:
             0 => Self::TRANSPARENT,
 
+            // Not sure the extra branch is worth it?
+            //1 => Self::from_rgba_premultiplied(r >> 7, g >> 7, b >> 7, 1),
+
             // common-case optimization:
             255 => Self::from_rgb(r, g, b),
 
             a => {
-                let r = fast_round(r as f32 * linear_f32_from_linear_u8(a));
-                let g = fast_round(g as f32 * linear_f32_from_linear_u8(a));
-                let b = fast_round(b as f32 * linear_f32_from_linear_u8(a));
+                let r = mul_frac_round(r, a);
+                let g = mul_frac_round(g, a);
+                let b = mul_frac_round(b, a);
                 Self::from_rgba_premultiplied(r, g, b, a)
             }
         }
--- a/crates/ecolor/src/color32.rs
+++ b/crates/ecolor/src/color32.rs
@@ -131,7 +131,17 @@ impl Color32 {
     /// but for transparent colors what you get back might be slightly different (rounding errors).
     #[inline]
     pub fn from_rgba_unmultiplied(r: u8, g: u8, b: u8, a: u8) -> Self {
-        use std::sync::OnceLock;
+        const LOOKUP_TABLE: &[u8] = &const {
+            const LEN: usize = u16::MAX as usize + 1;
+            let mut tbl = [0u8; LEN];
+            let mut i = 0;
+            while i < LEN {
+                let [value, alpha] = (i as u16).to_ne_bytes();
+                tbl[i] = fast_round(value as f32 * linear_f32_from_linear_u8(alpha));
+                i += 1;
+            }
+            tbl
+        };
         match a {
             // common-case optimization:
             0 => Self::TRANSPARENT,
@@ -140,18 +150,8 @@ impl Color32 {
             255 => Self::from_rgb(r, g, b),
 
             a => {
-                static LOOKUP_TABLE: OnceLock<Box<[u8]>> = OnceLock::new();
-                let lut = LOOKUP_TABLE.get_or_init(|| {
-                    (0..=u16::MAX)
-                        .map(|i| {
-                            let [value, alpha] = i.to_ne_bytes();
-                            fast_round(value as f32 * linear_f32_from_linear_u8(alpha))
-                        })
-                        .collect()
-                });
-
-                let [r, g, b] =
-                    [r, g, b].map(|value| lut[usize::from(u16::from_ne_bytes([value, a]))]);
+                let [r, g, b] = [r, g, b]
+                    .map(|value| LOOKUP_TABLE[usize::from(u16::from_ne_bytes([value, a]))]);
                 Self::from_rgba_premultiplied(r, g, b, a)
             }
         }
--- a/crates/ecolor/src/color32.rs
+++ b/crates/ecolor/src/color32.rs
@@ -143,9 +143,6 @@ impl Color32 {
             // common-case optimization:
             0 => Self::TRANSPARENT,
 
-            // Not sure the extra branch is worth it?
-            //1 => Self::from_rgba_premultiplied(r >> 7, g >> 7, b >> 7, 1),
-
             // common-case optimization:
             255 => Self::from_rgb(r, g, b),
 
--- a/crates/ecolor/src/color32.rs
+++ b/crates/ecolor/src/color32.rs
@@ -130,15 +130,7 @@ impl Color32 {
     /// You can use [`Self::to_srgba_unmultiplied`] to get back these values,
     /// but for transparent colors what you get back might be slightly different (rounding errors).
     #[inline]
-    pub fn from_rgba_unmultiplied(r: u8, g: u8, b: u8, a: u8) -> Self {
-        Self::from_rgba_unmultiplied_const(r, g, b, a)
-    }
-
-    /// Same as [`Self::from_rgba_unmultiplied`], but can be used in a const context.
-    ///
-    /// It is slightly slower when operating on non-const data.
-    #[inline]
-    pub const fn from_rgba_unmultiplied_const(r: u8, g: u8, b: u8, a: u8) -> Self {
+    pub const fn from_rgba_unmultiplied(r: u8, g: u8, b: u8, a: u8) -> Self {
         match a {
             // common-case optimization:
             0 => Self::TRANSPARENT,
@@ -153,6 +145,13 @@ impl Color32 {
                 Self::from_rgba_premultiplied(r, g, b, a)
             }
         }
+    }
+
+    /// This is the same as [`Self::from_rgba_unmultiplied`] use it instead. 
+    #[deprecated = "Use `Color32::from_rgba_unmultiplied()` instead"]
+    #[inline]
+    pub const fn from_rgba_unmultiplied_const(r: u8, g: u8, b: u8, a: u8) -> Self {
+        Self::from_rgba_unmultiplied(r, g, b, a)
     }
 
     /// Opaque gray.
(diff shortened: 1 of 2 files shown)


crates/ecolor/src/color32.rs
```diff
@@ -166,13 +143,16 @@ impl Color32 {
             // common-case optimization:
             0 => Self::TRANSPARENT,
 
+            // Not sure the extra branch is worth it?
+            //1 => Self::from_rgba_premultiplied(r >> 7, g >> 7, b >> 7, 1),
+
```
emilk: rm

crates/ecolor/src/color32.rs
emilk: We can deprecate this now and just make `from_rgba_unmultiplied` `const`

emilk (commented):
interesting! But please add a benchmark for this (or do we have one?)! I'm in particular interested in not regressing on:

* random semi-transparent colors
* random opaque colors

crates/ecolor/src/color32.rs
Keith-Cancel: No problem.

Keith-Cancel:
> interesting! But please add a benchmark for this (or do we have one?)! I'm in particular interested in not regressing on:
> 
>     * random semi-transparent colors
> 
>     * random opaque colors

let me first see if there is a good existing benchmark. If not I will make one and test various versions. I can tell you that at the https://github.com/emilk/egui/pull/8427/changes/a729f6f87286bf96c0998db43b4f681bb42d3e6a that should not be a regression speed wise. At least x86_64 with optimizations enabled binary is size similar (different archs may vary), memory size is basically the same since both are 64k tables, although you have the extra instructions to generate the table at runtime in the old version.

Keith-Cancel:
Okay I was able to find an existing test in the epaint crate. I posted the bench marks below first is the baseline of the old OnceLock code. Then the fixed point math, and then a compile time const table. Looking at this considering how the fixed point math performs about the same for `from_rgba_unmultiplied_other` and faster for the other two, and the binary and memory space savings I think make the best choice.

## OnceLock Table BaseLine

`cargo bench --package epaint --bench benchmark from_rgba_unmultiplied -- --save-baseline once_lock`

```
     Running benches/benchmark.rs (/home/keith/GitHub/egui/target/release/deps/benchmark-cefd662965a9cd37)
from_rgba_unmultiplied_0
                        time:   [943.13 ns 949.30 ns 956.37 ns]
                        change: [−2.9771% −1.6824% −0.3101%] (p = 0.01 < 0.05)
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  3 (3.00%) high mild
  3 (3.00%) high severe

from_rgba_unmultiplied_other
                        time:   [3.1572 µs 3.1895 µs 3.2298 µs]
                        change: [−2.2236% +0.3742% +3.1906%] (p = 0.79 > 0.05)
Found 11 outliers among 100 measurements (11.00%)
  1 (1.00%) high mild
  10 (10.00%) high severe

from_rgba_unmultiplied_255
                        time:   [1.7178 µs 1.7464 µs 1.7788 µs]
                        change: [−2.5141% −0.4641% +1.6819%] (p = 0.68 > 0.05)
Found 14 outliers among 100 measurements (14.00%)
  1 (1.00%) low mild
  7 (7.00%) high mild
  6 (6.00%) high severe
```

## Fixed Point Math

`cargo bench --package epaint --bench benchmark from_rgba_unmultiplied -- --baseline once_lock`

```
     Running benches/benchmark.rs (/home/keith/GitHub/egui/target/release/deps/benchmark-cefd662965a9cd37)
from_rgba_unmultiplied_0
                        time:   [572.99 ns 573.67 ns 574.50 ns]
                        change: [−39.471% −38.872% −38.221%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  2 (2.00%) high mild
  10 (10.00%) high severe

from_rgba_unmultiplied_other
                        time:   [3.2433 µs 3.2474 µs 3.2523 µs]
                        change: [−2.9816% −0.3287% +1.9602%] (p = 0.80 > 0.05)
Found 12 outliers among 100 measurements (12.00%)
  12 (12.00%) high severe

from_rgba_unmultiplied_255
                        time:   [1.3105 µs 1.3132 µs 1.3164 µs]
                        change: [−24.824% −23.589% −22.390%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 14 outliers among 100 measurements (14.00%)
  4 (4.00%) high mild
  10 (10.00%) high severe
```

## Const Table

`cargo bench --package epaint --bench benchmark from_rgba_unmultiplied -- --baseline once_lock`

```
     Running benches/benchmark.rs (/home/keith/GitHub/egui/target/release/deps/benchmark-cefd662965a9cd37)
from_rgba_unmultiplied_0
                        time:   [575.21 ns 578.83 ns 583.16 ns]
                        change: [−39.355% −38.743% −38.105%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  6 (6.00%) high mild
  9 (9.00%) high severe

from_rgba_unmultiplied_other
                        time:   [2.2373 µs 2.2640 µs 2.2937 µs]
                        change: [−30.556% −27.902% −25.346%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high severe

from_rgba_unmultiplied_255
                        time:   [1.3161 µs 1.3203 µs 1.3251 µs]
                        change: [−24.142% −22.655% −20.833%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  2 (2.00%) high mild
  9 (9.00%) high severe
```
# Floating Point Math
This was more quick test to see how all 3 (e.g OnceLock table, fixed point, and const table) compare to a floating point version like in the const version you talked about deprecating.

All three are definitely faster than using floating point math. 

## Floating Point baseline 

`cargo bench --package epaint --bench benchmark from_rgba_unmultiplied -- --save-baseline floating_point`

```
     Running benches/benchmark.rs (/home/keith/GitHub/egui/target/release/deps/benchmark-cefd662965a9cd37)
from_rgba_unmultiplied_0
                        time:   [667.54 ns 678.01 ns 689.58 ns]
                        change: [−1.2562% +0.7820% +2.7627%] (p = 0.46 > 0.05)
Found 12 outliers among 100 measurements (12.00%)
  9 (9.00%) high mild
  3 (3.00%) high severe

from_rgba_unmultiplied_other
                        time:   [6.0790 µs 6.1437 µs 6.2236 µs]
                        change: [−1.6974% +0.3500% +2.5036%] (p = 0.74 > 0.05)
  6 (6.00%) high mild
  1 (1.00%) high severe

from_rgba_unmultiplied_255
                        time:   [1.3429 µs 1.3476 µs 1.3538 µs]
                        change: [−7.4862% −6.0059% −4.1169%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 14 outliers among 100 measurements (14.00%)
  5 (5.00%) high mild
  9 (9.00%) high severe
```

## Once Lock

`cargo bench --package epaint --bench benchmark from_rgba_unmultiplied -- --baseline floating_point`

```
     Running benches/benchmark.rs (/home/keith/GitHub/egui/target/release/deps/benchmark-cefd662965a9cd37)
from_rgba_unmultiplied_0
                        time:   [885.28 ns 890.53 ns 896.18 ns]
                        change: [+31.103% +34.137% +37.554%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 9 outliers among 100 measurements (9.00%)
  5 (5.00%) high mild
  4 (4.00%) high severe

from_rgba_unmultiplied_other
                        time:   [2.8821 µs 2.9101 µs 2.9430 µs]
                        change: [−54.267% −52.495% −50.710%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe

from_rgba_unmultiplied_255
                        time:   [1.5282 µs 1.5342 µs 1.5427 µs]
                        change: [+10.680% +12.736% +14.431%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 12 outliers among 100 measurements (12.00%)
  3 (3.00%) high mild
  9 (9.00%) high severe
```

## Fixed Point

`cargo bench --package epaint --bench benchmark from_rgba_unmultiplied -- --baseline floating_point`

```
     Running benches/benchmark.rs (/home/keith/GitHub/egui/target/release/deps/benchmark-cefd662965a9cd37)
from_rgba_unmultiplied_0
                        time:   [585.59 ns 590.01 ns 595.26 ns]
                        change: [−12.844% −11.041% −8.7887%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  5 (5.00%) high mild
  1 (1.00%) high severe

from_rgba_unmultiplied_other
                        time:   [3.2787 µs 3.2946 µs 3.3172 µs]
                        change: [−49.066% −47.920% −46.682%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  4 (4.00%) high mild
  11 (11.00%) high severe

from_rgba_unmultiplied_255
                        time:   [1.3453 µs 1.3620 µs 1.3817 µs]
                        change: [−1.5957% +0.3302% +1.9433%] (p = 0.74 > 0.05)
  5 (5.00%) high mild
```

## Const Table

`cargo bench --package epaint --bench benchmark from_rgba_unmultiplied -- --baseline floating_point`

```
     Running benches/benchmark.rs (/home/keith/GitHub/egui/target/release/deps/benchmark-cefd662965a9cd37)
from_rgba_unmultiplied_0
                        time:   [597.52 ns 605.22 ns 613.56 ns]
                        change: [−8.1739% −5.7382% −2.9118%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  3 (3.00%) high severe

from_rgba_unmultiplied_other
                        time:   [2.2106 µs 2.2202 µs 2.2312 µs]
                        change: [−65.862% −65.207% −64.583%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) high mild
  6 (6.00%) high severe

from_rgba_unmultiplied_255
                        time:   [1.3185 µs 1.3228 µs 1.3278 µs]
                        change: [−4.4251% −2.6272% −1.2633%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  6 (6.00%) high mild
  6 (6.00%) high severe
```

crates/ecolor/src/color32.rs
Keith-Cancel: Just added a commit to mark it as deprecated.

emilk (approved):
Very nice - thank you!

crates/ecolor/src/color32.rs
```diff
@@ -140,42 +139,19 @@ impl Color32 {
...
                 Self::from_rgba_premultiplied(r, g, b, a)
             }
         }
     }
 
-    /// Same as [`Self::from_rgba_unmultiplied`], but can be used in a const context.
-    ///
-    /// It is slightly slower when operating on non-const data.
+    /// This is the same as [`Self::from_rgba_unmultiplied`] use it instead. 
+    #[deprecated = "Use `Color32::from_rgba_unmultiplied()` instead"]
```
emilk: on second thought, lets not do this. By having them separate we can still use a const table in the future if we change our minds (again). Sorry for the annoying back-and-forth

Keith-Cancel:
Okay I undid the deprecation. 

Also side question this function definitely could benefit from vectorization: https://github.com/emilk/egui/blob/fd54387eac03f57ca772a8fb590ceaadf780f31c/crates/epaint/src/image.rs#L113-L126

I can easily vectorize it, the problem is portable SIMD is nightly, and still being worked on.

The builtin `#[target_feature(enable = <feature>)]` attribute allows calling safe intrinsics in that context. However, [is_x86_feature_detected](https://doc.rust-lang.org/std/macro.is_x86_feature_detected.html) still would require only and only one unsafe for the initial call into the intrinsic path. (ideally that macro could see the `#[target_feature]` attribute, but does not). However, the only unsafe is located in eframe and winit-related crates. So not sure if there is any where I could put if it I went that route.

The other options is something like this: https://crates.io/crates/safe_arch where it exposes safe versions that are available only at compile time if compiled with a feature flag that allows those intrinsics. 

## Edit

This also is an intersting idea, and looks like they made a crate for it: https://shnatsel.medium.com/safe-simd-in-rust-even-on-the-inside-c6f1ff381828, but it's still relatively new: https://crates.io/crates/archmage

crates/ecolor/src/color32.rs
Keith-Cancel: No Problem easy enough. I also made the main function not-const again as not to increase the API surface/contract  from how it was originally.

emilk:
There is also https://github.com/linebender/fearless_simd - though I am a little hesitant to adding complex simd stuff to epaint :/

Keith-Cancel:
> There is also https://github.com/linebender/fearless_simd - though I am a little hesitant to adding complex simd stuff to epaint :/

@emilk SIMD is not really that complicated, unless you want to support every arch out there vs a small handful. The complexity I have always have found is more in that. I can also try see if maybe I can express the code in a way the compiler finds it really easy to vectorize.

PR #8428, opened 2026-08-17
Support the X11/Wayland PRIMARY selection

### LINKED ISSUE #5852: Add support for copying and pasting to the X11/Wayland PRIMARY clipboard

In most applications using X11/Wayland, it is possible to utilize the primary clipboard by selecting text and pasting this selection using the middle mouse button, but not in egui. Because I want to make an application using egui that I will use myself (a level editor), this would add functionality that I expect from every application and would aid me in the usage of this tool.

This crate uses `arboard`, which has primary clipboard support. It would not be hard to copy selected text into the PRIMARY clipboard, and paste it using a middle mouse button event.

hallyhaa:
* Closes <https://github.com/emilk/egui/issues/5852>

Selecting text in a `Label` or a `TextEdit` now fills the PRIMARY selection, and middle-clicking a `TextEdit` pastes it. This is what all other X11 and Wayland applications do.

Testing: `tests/egui_tests/tests/test_primary_selection.rs`. Publishing from labels and text edits, no publishing of passwords, no re-claiming of an unchanged selection, nothing at all on non-PRIMARY platforms, pasting at the click position rather than at the cursor, and no paste from a click outside the widget.


crates/egui/src/data/output.rs
```diff
@@ -124,6 +124,16 @@ pub enum OutputCommand {
...
 
+    /// Put this text in the X11/Wayland PRIMARY selection.
+    ///
+    /// PRIMARY is the "select here, middle-click paste there" channel, separate
+    /// from the Ctrl+C/Ctrl+V clipboard that [`Self::CopyText`] writes to. egui
+    /// emits this whenever the user finishes selecting text, which is what every
+    /// other application on those platforms does.
+    ///
+    /// Integrations on platforms without a PRIMARY selection should ignore this.
+    CopyTextToPrimary(String),
```
emilk: I think we should generalize this to `TextSelectionChanged`, or even remove it and just have eframe ask egui "What is the current text selection?" and monitor it for change (we should have a `egui_ctx.text_selection()` if we don't already)

hallyhaa: Generalize this: yes, absolutely. But I don't think we should poll. We would then have to build a string every frame a selection exists (for label selections), instead of once when it settles. And "don't re-claim PRIMARY mid-drag" would have to be re-derived by every integration.

hallyhaa: Made some changes. Now use `Options::report_text_selection` (default to _on_ on linux). The cost that the platform check was really guarding is still guarded (only assemble the selected text when someone is listening). Now awaiting your opinion.

crates/egui-winit/src/clipboard.rs
```diff
@@ -120,6 +120,144 @@ impl Clipboard {
...
+    /// Returns `None` on platforms without a PRIMARY selection, and when
+    /// nothing owns it.
+    #[cfg_attr(
+        not(all(
+            any(
+                target_os = "linux",
+                target_os = "dragonfly",
+                target_os = "freebsd",
+                target_os = "netbsd",
+                target_os = "openbsd"
```
emilk: is there any way we can get rid of this repetition? 

My best idea is `cfg_aliases` to create a 'nix' alias or similar, but maybe there is something better

hallyhaa: Found another solution. You might add cfg_aliases later, if you think that's even better ...

crates/egui-winit/src/clipboard.rs
```diff
@@ -144,6 +157,156 @@ impl Clipboard {
...
+))]
+#[cfg_attr(
+    not(any(feature = "arboard", feature = "smithay-clipboard")),
+    expect(
+        clippy::unused_self,
+        clippy::needless_pass_by_ref_mut,
+        clippy::unnecessary_wraps,
+        reason = "these do nothing without a clipboard backend to talk to"
+    )
+)]
```
emilk: I think you can `cfg_select!` a whole `impl` block, so you can get an `else` instead of having another `#[cfg(not(…`

hallyhaa: Hey, that works! The list is down to 3 places now: the field, the constructor and the `cfg_select!` arm.

hallyhaa:
Merged in the main branch. Fixed two collisions (both trivial) with #8472 and an error invisible to git: TextEdit's `text_mutable` was gone, so the middle-click paste asks `text.is_mutable()` now.

PR #8429, opened 2026-08-18
Enable the `clippy::pedantic` lint group

emilk:
Instead of opting in to pedantic lints one by one, enable the whole group and opt out of the noisy ones.

64% of the pedantic lints were already listed individually. This deletes 90 explicit lint lines, enables 51 pedantic lints we never listed, and picks up new pedantic lints for free. Each opt-out carries its hit count, so the cost of turning one back on is visible.

`restriction` and `nursery` stay opt-in per lint.

Stacked on top of #8430, which fixes the one real bug the new lints found.


crates/egui_demo_lib/src/easy_mark/easy_mark_parser.rs
```diff
@@ -123,10 +123,9 @@ impl<'a> Parser<'a> {
                 self.s = &code_start[end + 4..];
                 self.start_of_line = false;
                 return Some(Item::CodeBlock(language, code));
-            } else {
-                self.s = "";
-                return Some(Item::CodeBlock(language, code_start));
             }
+            self.s = "";
+            return Some(Item::CodeBlock(language, code_start));
```
emilk: moving the `return` to before the `if` would be nicer, I think. THen the if/else would be symmetrical. Here and elsewhere

emilk: Done: `return if … { … } else { … };` in both `easy_mark_parser` spots and both `snapshot.rs` spots.

The `tessellator.rs` one is left as-is, since its branches are not symmetrical (`return` vs `break`).

— Claude

PR #8433, opened 2026-08-19
Fix the constrain option not working in the window options demo

limo520:
`constrain_to` overrides `constrain`, so I adjusted it to call `constrain_to` before `constrain`.


emilk (approved):
thanks

PR #8441, opened 2026-08-20
Make kittest's predictable texture filtering honor `TextureOptions::NEAREST`

emilk:
egui-wgpu's `predictable_texture_filtering` path (used by egui_kittest for reproducible snapshots) always did manual 4-tap bilinear filtering, ignoring the texture's sampler. Textures uploaded with `TextureOptions::NEAREST` rendered blurry in kittest snapshots, while real apps rendered them crisp.


lucasmerlin (approved):
nice catch!

PR #8439, opened 2026-08-20
Round the corners of the color picker

emilk:
The color picker's gradients and background checkers now follow the style's corner radius, giving it a softer look.

This required changing the painting from being Mesh-based to being texture-based

it's subtle with the default settings:

Screenshot 2026-08-21 at 15 08 18

But you can [increase it](https://github.com/emilk/egui/pull/8445):

Screenshot 2026-08-21 at 15 08 41

### Before
for reference
Screenshot 2026-08-21 at 19 08 27

PR #8445, opened 2026-08-21
Increase widget corner radius to 4

emilk:
A little more round and friendly!

* See also https://github.com/emilk/egui/pull/8439


lucasmerlin (approved):
I like it!

Issue #8265, opened 2026-06-25
egui-wgpu panics in Renderer::update_buffers when Queue::write_buffer_with returns None on Windows Server + RDP


luo007:
I am seeing recurring native crashes in a long-running desktop application using `egui + egui-wgpu + wgpu`.

The panic happens in `egui-wgpu::Renderer::update_buffers` when uploading index data:

```text
Failed to create staging buffer for index data.
```

The important detail is that the required index buffer size is smaller than the existing buffer size/capacity, so this does not appear to be a normal buffer growth issue. It looks like `Queue::write_buffer_with` returned `None`.

Environment:

```text
OS: Windows Server 2018
GPU: NVIDIA RTX/RXT 4000
Usage: long-running desktop application over Remote Desktop
Trigger pattern: RDP session is disconnected, application keeps running, then after reconnecting and running for some time the panic occurs
egui-wgpu versions observed: 0.34.1, 0.34.2, 0.34.3
```

Example logs:

```text
egui-wgpu-0.34.1 renderer.rs:971:
Failed to create staging buffer for index data. Index count: 79872.
Required index buffer size: 319488.
Actual size 419808 and capacity: 419808.

egui-wgpu-0.34.2 renderer.rs:971:
Failed to create staging buffer for index data. Index count: 114084.
Required index buffer size: 456336.
Actual size 557520 and capacity: 557520.

egui-wgpu-0.34.3 renderer.rs:971:
Failed to create staging buffer for index data. Index count: 116322.
Required index buffer size: 465288.
Actual size 562704 and capacity: 562704.
```

This looks similar to issue #7840, but in this case the application is running on Windows Server with RDP disconnect/reconnect involved.

PR #8452, opened 2026-08-23
Recover rendering instead of panicking after wgpu failures

### LINKED ISSUE #8450: `egui-wgpu` panics with "Failed to create staging buffer for index data" during prolonged execution
`egui-wgpu` panics with "Failed to create staging buffer for index data" during prolonged execution

egui Version: main & All
OS : Windows 10

### Describe the bug

When running an application with the `wgpu` backend for several hours, a validation error occurred and the application panicked inside `egui-wgpu::renderer`.

### Error Log / Stacktrace

```text
WARN  [egui_wgpu] Dropped frame with error: Validation

thread 'main' panicked at egui-wgpu/src/renderer.rs:984:17:
Failed to create staging buffer for index data. Index count: 25200. Required index buffer size: 100800. Actual size 187200 and capacity: 187200 (bytes)
```

### Context & Observations
* **Backend**: `wgpu` (tested on Windows)
* **Trigger condition**: Occurs intermittently after the application runs continuously for a long time (several hours) with active UI updates.
* It looks like an issue related to buffer resizing, alignment, or staging buffer management when the index buffer size/count changes over time.

rustbasic:
Recover rendering instead of panicking after wgpu failures

**Description**

This PR prevents `egui-wgpu` from panicking when `Queue::write_buffer_with` cannot provide a staging buffer for index or vertex uploads. It keeps the existing fast path and falls back to `queue.write_buffer` when needed.

It also avoids uploading GPU resources after surface frame acquisition has already failed.

For `Validation` and `Lost` surface errors, egui-wgpu recreates the invalid GPU render state and reuploads managed textures on the next window setup. This allows rendering to resume instead of leaving the application with a frozen or invalid render path.

* Related #5063
* Related #8265
* Related #8292
* Closes #8450

### DIFF
--- a/crates/egui-wgpu/src/renderer.rs
+++ b/crates/egui-wgpu/src/renderer.rs
@@ -245,12 +245,6 @@ pub struct Renderer {
     uniform_bind_group: wgpu::BindGroup,
     texture_bind_group_layout: wgpu::BindGroupLayout,
 
-    /// Uniform buffers each holding a single `u32`:
-    /// 1 if the texture sampler uses nearest filtering, 0 otherwise.
-    /// Indexed by that flag value.
-    /// Read by the shader when `predictable_texture_filtering` is on.
-    nearest_filtering_flag_buffers: [wgpu::Buffer; 2],
-
     /// Map of egui texture IDs to textures and their associated bindgroups (texture view +
     /// sampler). The texture may be None if the `TextureId` is just a handle to a user-provided
     /// sampler.
@@ -353,27 +347,9 @@ impl Renderer {
                         ty: wgpu::BindingType::Sampler(wgpu::SamplerBindingType::Filtering),
                         count: None,
                     },
-                    wgpu::BindGroupLayoutEntry {
-                        binding: 2,
-                        visibility: wgpu::ShaderStages::FRAGMENT,
-                        ty: wgpu::BindingType::Buffer {
-                            has_dynamic_offset: false,
-                            min_binding_size: NonZeroU64::new(core::mem::size_of::<u32>() as _),
-                            ty: wgpu::BufferBindingType::Uniform,
-                        },
-                        count: None,
-                    },
                 ],
             })
         };
-
-        let nearest_filtering_flag_buffers = [0_u32, 1_u32].map(|flag| {
-            device.create_buffer_init(&wgpu::util::BufferInitDescriptor {
-                label: Some(&format!("egui_nearest_filtering_flag_{flag}")),
-                contents: bytemuck::bytes_of(&flag),
-                usage: wgpu::BufferUsages::UNIFORM,
-            })
-        });
 
         let pipeline_layout = device.create_pipeline_layout(&wgpu::PipelineLayoutDescriptor {
             label: Some("egui_pipeline_layout"),
@@ -482,7 +458,6 @@ impl Renderer {
             previous_uniform_buffer_content: UniformBuffer::zeroed(),
             uniform_bind_group,
             texture_bind_group_layout,
-            nearest_filtering_flag_buffers,
             textures: HashMap::default(),
             next_user_texture_id: 0,
             samplers: HashMap::default(),
@@ -734,8 +709,6 @@ impl Renderer {
         };
 
         let bind_group = bind_group.unwrap_or_else(|| {
-            let nearest =
-                image_delta.options.magnification == epaint::textures::TextureFilter::Nearest;
             let sampler = self
                 .samplers
                 .entry(image_delta.options)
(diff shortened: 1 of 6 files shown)


crates/egui-wgpu/src/winit.rs
```diff
@@ -607,7 +592,23 @@ impl Painter {
                 frame
             }
             other => {
-                match (*self.config.on_surface_status)(&other) {
+                let needs_render_state_recreate = matches!(
+                    &other,
+                    wgpu::CurrentSurfaceTexture::Validation | wgpu::CurrentSurfaceTexture::Lost
+                );
```
Wumpf, line 598: at this point it would be better to fully spell out the match statement - really what this is saying is that we don't do anything for `Timeout` and `Occluded`, but that's not visible here

crates/egui-wgpu/src/renderer.rs
```diff
@@ -1024,28 +1024,44 @@ impl Renderer {
...
+                            let slice = index_offset..(size + index_offset);
+                            index_data.extend_from_slice(bytes);
+                            self.index_buffer.slices.push(slice);
+                            index_offset += size;
+                        }
+                        Primitive::Callback(_) => {}
                     }
-                    Primitive::Callback(_) => {}
                 }
+                queue.write_buffer(&self.index_buffer.buffer, 0, &index_data);
```
Wumpf, line 1064: did you ever observe `write_buffer` to succeed when the index staging buffer worked? That sounds very strange to me and hints at a different problem. I'd much rather figure out why that is instead of adding a second code path here

crates/egui-wgpu/src/renderer.rs
```diff
@@ -1070,28 +1086,44 @@ impl Renderer {
...
+                            let slice = vertex_offset..(size + vertex_offset);
+                            vertex_data.extend_from_slice(bytes);
+                            self.vertex_buffer.slices.push(slice);
+                            vertex_offset += size;
+                        }
+                        Primitive::Callback(_) => {}
                     }
-                    Primitive::Callback(_) => {}
                 }
+                queue.write_buffer(&self.vertex_buffer.buffer, 0, &vertex_data);
```
Wumpf, line 1126: as above

crates/egui-wgpu/src/winit.rs
```diff
@@ -607,7 +592,23 @@ impl Painter {
                 frame
             }
             other => {
-                match (*self.config.on_surface_status)(&other) {
+                let needs_render_state_recreate = matches!(
```
Wumpf, line 595: I think it would be better to express this new state as a new variant of `SurfaceErrorAction`

crates/egui/src/context.rs
```diff
@@ -2452,6 +2452,14 @@ impl Context {
         TextureHandle::new(tex_mngr, tex_id)
     }
 
+    /// Request full uploads of all managed textures after a renderer recovery.
+    ///
+    /// Integrations that recreate their GPU renderer while keeping this context need this
+    /// so live texture handles are restored in the new backend texture map.
+    pub fn request_full_texture_reupload(&self) {
```
Wumpf, line 2459: it seems what you're really after is lost device recovery, i.e. recreation of gpu resources. So more like

```suggestion
    pub fn request_gpu_device_reset(&self) {
```

?

Wumpf: On that note I'm very surprised that you only reach out to the texture manager here. Is that maybe the reason you're getting failures on write buffer?

crates/epaint/src/textures.rs
```diff
@@ -16,6 +16,8 @@ pub struct TextureManager {
     /// Information about currently allocated textures.
     metas: ahash::HashMap<TextureId, TextureMeta>,
 
+    /// CPU-side copies of live managed textures, retained for renderer recovery.
+    images: ahash::HashMap<TextureId, ImageData>,
```
Wumpf, line 20: this is a very heavy thing to do and may not actually be needed on all backends. Not even sure it's needed for all wgpu backends (need to look up which one can run into full device loss)
I think we should make this kind of retaining-for-reveroty wgpu only and a be a knob there that one can opt out of

Wumpf (changes requested):
It looks like you're trying to implement lost device recovery but do so a bit piecemeal by on one hand instruction textures to be re-uploaded and on the other doing odd workarounds for when the buffer objects get poisoned by said device loss. A cleaner approach would be to handle such an error explicitely and recreate all the resources.
It probably also needs some way of crashing or halting the application after a failed attempt in case this just leads to a failure loop (for egui probably best to signal this to the application so it can decide what to do with the fact that rendering is no longer recovering)

PR #8454, opened 2026-08-25
Fix `egui-wgpu` build on `wasm32-unknown-emscripten`

bfeip:
Fix a build issue in the `egui-wgpu` crate when building for the `wasm32-unknown-emscripten` target. WGPU does not support the WebGPU backend on that target.


lucasmerlin (approved):
Thanks!

PR #8455, opened 2026-08-25
StyleProvider for TextEdit and DragValue

lucasmerlin:
Implements `StyleProvider` for `TextEdit`, so you can fully style it as you please

### DIFF
--- a/crates/egui/src/memory/mod.rs
+++ b/crates/egui/src/memory/mod.rs
@@ -617,10 +617,7 @@ impl Focus {
             }
         }
 
-        // AccessKit names the widget by the node id it read from the last tree we sent, so the
-        // widget has been here before and is in the cache. Handing the focus out here, instead of
-        // waiting for the widget to ask for it, means everything that reads the focus before the
-        // widget runs — styling it, for one — already sees the new focus.
+        // Handle accesskit focus requests
         let newly_focused = focus_requested_by_accesskit.and_then(|node_id| {
             self.focus_widgets_cache
                 .keys()
--- a/crates/egui/src/memory/mod.rs
+++ b/crates/egui/src/memory/mod.rs
@@ -618,6 +618,27 @@ impl Focus {
                 self.id_requested_by_accesskit = Some(*target_node);
             }
         }
+
+        // `interested_in_focus` only picks the request up once the widget runs, which is too late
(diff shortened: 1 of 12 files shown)


crates/egui/src/memory/mod.rs
```diff
@@ -645,13 +656,6 @@ impl Focus {
     }
 
     fn interested_in_focus(&mut self, id: Id) {
-        if self.id_requested_by_accesskit == Some(id.accesskit_id()) {
-            self.focused_widget = Some(FocusWidget::new(id));
-            self.id_requested_by_accesskit = None;
-            self.give_to_next = false;
-            self.reset_focus();
-        }
-
```
lucasmerlin, line 654: Having it here meant the focus changed during the frame, and something using `ctx.read_response` might see a different focus than the widget sees later. Moving it up makes this more consistent (which is why the snapshots show cursors now)

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -481,7 +523,7 @@ impl TextEdit<'_> {
         // `min_size` overrides available width
         let allocate_width = desired_width.at_most(available_width).at_least(min_size.x);
 
-        let font_id_clone = font_id.clone();
+        let font_id_clone = font_id;
```
emilk: no longer needed?

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -705,53 +744,19 @@ impl TextEdit<'_> {
                 TextWrapMode::Truncate
             };
 
-            let mut allocated = AtomLayout::new(atoms)
+            let allocated = atom_layout_style
+                .apply(AtomLayout::new(atoms))
+                // The text being edited gets its color from the layouter, so the only atoms
+                // left to color are the prefix and the suffix.
+                .fallback_text_color(prefix_suffix_color)
```
emilk, line 750: make sure we have a test for different prefix/suffix colors

emilk (approved):

crates/egui/src/widgets/text_edit/builder.rs
lucasmerlin: https://github.com/emilk/egui/pull/8477

PR #8463, opened 2026-08-28
Point Rust Analyzer to `target/rust-analyzer`

emilk:
So that `cargo clean`, `cargo sweep` etc clears it for us! And one less annoying folder in root.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,7 +8,7 @@
         }
     },
     "files.exclude": {
-        "target_ra/**": true,
+        "target_ra/**": true, // legacy Rust Analyzer target dir
         "target_wasm/**": true,
         "target/**": true,
     },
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -18,7 +18,7 @@
     "rust-analyzer.cargo.allTargets": true,
     "rust-analyzer.cargo.features": "all",
     // Use a separate target directory for Rust Analyzer so it doesn't prevent cargo/clippy from doing things.
```
Mingun, line 20: So this comment no longer valid?

lucasmerlin (approved):
I don't use rust-analyzer or vscode, so I can't verify, but I guess it's fine

.vscode/settings.json
emilk: It is still valid; the target dir has just moved to be inside `target` 🫠

PR #8464, opened 2026-08-28
Rename classes, more button options via `StyleProvider`

lucasmerlin:
Moves global classnames to e.g. `egui::class::ROOT`.

Widget specific classes live on the widgets struct. 

Adds some more classes for our buttons styling options (`CLASS_SMALL` etc), so that custom `StyleProvider`s can choose to implement them differently or ignore them altogether. 

Also removes some styleproviders that were partly or completely unused (e.g. basestyle / labelstyle). We can add them back once we actually implement styling for those.


crates/egui/src/containers/frame.rs
```diff
@@ -298,6 +298,18 @@ impl Frame {
...
+    /// Expand the frame without affecting layout.
+    ///
+    /// This handles `expansion` by subtracting it from the outer margin and adding it to the
+    /// inner margin. It also corrects for a stroke changing on hover, by subtracting the stroke
+    /// width from `inner_margin`.
+    #[inline]
+    pub fn expand_in_place(mut self, expansion: f32) -> Self {
+        self.outer_margin = self.outer_margin - Margin::from(expansion);
+        self.inner_margin = self.inner_margin + Margin::from(expansion - self.stroke.width);
+        self
```
emilk, line 317: ⚠️ This is wrong!

Stroke width is already part of the Frame dimensions. From the doc-string of `Frame`:

Image


```suggestion
    /// Expand the frame without affecting layout.
    ///
    /// This handles `expansion` by subtracting it from the outer margin and adding it to the
    /// inner margin.
    #[inline]
    pub fn expand_in_place(mut self, expansion: f32) -> Self {
        self.outer_margin = self.outer_margin - Margin::from(expansion);
        self.inner_margin = self.inner_margin + Margin::from(expansion);
        self
```

crates/egui/src/widget_style/mod.rs
```diff
@@ -147,3 +149,15 @@ pub struct StyleArgs<'a> {
...
+
+impl StyleArgs<'_> {
+    /// Does the widget or any of its parents contain this class?
+    ///
+    /// See also:
+    /// - [`Classes::has_class`]
+    /// - [`UiStack::has_class`]
+    pub fn has_class(&self, class: impl Into<ClassName>) -> bool {
+        let class = class.into();
+        self.classes.has_class(class.clone()) || self.stack.has_class(class)
```
emilk: We should be able to avoid this `.clone()` by using the `Borrow` trait, e.g. like [`HashSet::get`](https://doc.rust-lang.org/stable/std/collections/struct.HashSet.html#method.get)

crates/egui/src/widgets/button.rs
```diff
@@ -42,16 +37,29 @@ pub struct Button<'a> {
...
+    pub const CLASS_SMALL: &'static str = "egui::small";
+
+    /// Present on a button that should have no frame at all.
+    pub const CLASS_NO_FRAME: &'static str = "egui::no_frame";
+
+    /// Present on a button that should have a frame, even when the global default is frameless.
+    pub const CLASS_FRAME: &'static str = "egui::frame";
+
+    /// Present on a button that should have no frame while it is inactive.
+    pub const CLASS_HIDE_FRAME_WHEN_INACTIVE: &'static str =
```
emilk: This could use `ClassName` instead of `&'static str`. It is a bit ugly to write `ClassName::Borrowed(…)` though… maybe `ClassName` could/should be a newtype-wrapper around `Cow` instead? It would make it easier to switch it to something else in the future (e.g. something that uses ref-counting in the Owned-case)


```
/// Cheap to clone (unlike `Cow::Owned`)
struct ClassName(ClassNamePrivate)

enum ClassNamePrivate {
    Static(&'static str),
    Owned(Arc<str>),
}

impl ClassName {
    pub fn from_static(&'static str) -> Self { … }
}

impl Borrow<str> for ClassName {
    fn borrow(&self) -> &str {
        …
    }
}
```

crates/egui/src/widgets/button.rs
```diff
@@ -142,41 +152,53 @@ impl<'a> Button<'a> {
...
 
     /// Make this a small button, suitable for embedding into text.
+    ///
+    /// This adds the built-in [`Self::CLASS_SMALL`], which with the default style removes the top and
+    /// bottom margin.
     #[inline]
     pub fn small(mut self) -> Self {
-        self.small = true;
+        self.add_class(Self::CLASS_SMALL);
         self
```
emilk: nicer ergonomics

```suggestion
        self.with_class(Self::CLASS_SMALL)
```

crates/egui/src/widgets/button.rs
```diff
@@ -353,21 +358,12 @@ impl<'a> Button<'a> {
...
 
-        // Retrocompatibility with button settings
-        layout = if has_frame_margin && (state != WidgetState::Inactive || frame_when_inactive) {
-            layout.frame(frame)
-        } else {
-            layout.frame(Frame::new().inner_margin(frame.inner_margin))
-        };
-
-        let mut prepared = layout.min_size(min_size).allocate(ui);
+        let mut prepared = layout.frame(frame).min_size(min_size).allocate(ui);
```
emilk, line 366: really nice to move out this logic out of `Button`!

crates/egui/src/context.rs
```diff
@@ -372,7 +372,7 @@ impl ViewportRepaintInfo {
 // ----------------------------------------------------------------------------
 
 #[derive(Default)]
-struct ContextImpl {
+pub(crate) struct ContextImpl {
```
emilk: I rather now make this `pub(crate)` if we can avoid it. Why is this needed?

crates/egui/src/context.rs
```diff
@@ -414,7 +414,7 @@ struct ContextImpl {
 
     loaders: Arc<Loaders>,
 
-    themes: theme::Themes,
+    pub(crate) themes: theme::Themes,
```
emilk: same here

crates/egui/src/context.rs
```diff
@@ -764,7 +767,7 @@ impl Context {
     }
 
     /// Do read-write (exclusive access) transaction on Context
-    fn write<R>(&self, writer: impl FnOnce(&mut ContextImpl) -> R) -> R {
+    pub(crate) fn write<R>(&self, writer: impl FnOnce(&mut ContextImpl) -> R) -> R {
```
emilk: and here

crates/egui/src/theme/default_style.rs
```diff
@@ -15,97 +15,88 @@ use crate::{
...
-            WidgetState::Inactive => style.visuals.widgets.inactive,
-            WidgetState::Hovered => style.visuals.widgets.hovered,
-            WidgetState::Active => style.visuals.widgets.active,
-        };
+        let mut widget_visuals = *style.visuals.widgets.state(*state);
 
-        let mut ws: BaseStyle = ctx.get_widget_style(modifiers);
-
-        if classes.has(SELECTED_CLASS) {
+        if classes.has_class(Button::CLASS_SELECTED) {
```
emilk: This naming is pretty odd, from a standard Rust API perspective

```suggestion
        if classes.contains(Button::CLASS_SELECTED) {
```

?

crates/egui/src/theme/themes.rs
```diff
@@ -2,64 +2,23 @@ use std::sync::Arc;
...
+/// (e.g. [`crate::widget_style::ButtonStyle`]), and that provider computes the final style from the
 /// widget's classes and state.
 ///
-/// A default provider is registered for every built-in style; register your
-/// own with [`Context::add_widget_theme`](crate::Context::add_widget_theme) or
-/// [`Context::replace_widget_theme`](crate::Context::replace_widget_theme).
+/// A default provider is registered for every built-in style. Register your
+/// own with [`crate::Context::add_widget_theme`] or [`crate::Context::replace_widget_theme`].
+#[derive(Default)]
 pub struct Themes {
```
emilk, line 18: So this changes `Themes::default()` from being the default egui theme into being an empty theme. How does an empty `Themes` work? Please document

emilk (changes requested):

crates/egui/src/containers/frame.rs
lucasmerlin: In widgets you don't want the stroke width changing to affect layout, or else you get weird layout shifts on hover (like seen here https://github.com/emilk/egui/pull/8471)
We already had this stroke-correction logic before, and it was repeated across widget impls so I added it here so we have a single place where the logic is defined.

lucasmerlin: In css the dev can correct for changing stroke widths themselves via margin etc. but in current egui styling (ignoring StyleProvider) there would be no way to correct for it. Since `WidgetVisuals` only defines stroke and not margin, it would be impossible to have a widget have 1px outline by default and 2px on hover (since this would cause layout shifts on hover).

crates/egui/src/theme/default_style.rs
lucasmerlin: These fns come from the `HasClasses` trait, so they have the `_class` suffix to make it more clear when calling them on a widget. I could duplicate them onto `Classes` without the suffix, but not sure if that's worth it

crates/egui/src/widget_style/mod.rs
lucasmerlin: I think Borrow is only needed when things are generic, I've changed it to take &str

crates/egui/src/widgets/button.rs
lucasmerlin: I've added `struct ClassName` and changed the consts. Unfortunately you now need to pass `&` when using the consts with `has_class`: `classes.has_class(&Button::CLASS_SELECTED)`.
If they were still `&'static str` they could be passed as-is.

crates/egui/src/containers/frame.rs
emilk: My point is this one: I would expect `.expand_in_place(0.0)` to be a no-op. Your version is not.

As for the separate discussion:

> it would be impossible to have a widget have 1px outline by default and 2px on hover (since this would cause layout shifts on hover).

You can do this in current egui, if you _also_ change margins to compensate. That is: if the hovered style has a wider stroke, then it needs to also shrink one of the margins.

This might not be obvious or ergonomic, but that's the way it is currently designed, going back to

* https://github.com/emilk/egui/pull/5575
* https://github.com/emilk/egui/issues/4019

If we want to change the definition of `Frame`, then we need to discuss a new design for it, but this PR is not the right place for that

lucasmerlin: > You can do this in current egui, if you also change margins to compensate.

Yes, but you cannot change margins via our current `WidgetVisuals` styling system. There is no way to have a different margin on hover as on a resting button.

But I'll rename the fn to something that makes more clear what it does!

EDIT: Renamed it to `apply_stroke_and_expansion_without_layout_shift` to make it more obvious what it does

crates/egui/src/theme/default_style.rs
emilk: Ah; makes sense

crates/egui/src/theme/default_style.rs
```diff
@@ -15,97 +15,88 @@ use crate::{
...
+        let painted_frame = Frame {
+            fill: widget_visuals.weak_bg_fill,
+            corner_radius: widget_visuals.corner_radius,
+            inner_margin,
+            ..Default::default()
+        }
+        // Ensure changing expansion and stroke don't affect layout:
+        .apply_stroke_and_expansion_without_layout_shift(
+            widget_visuals.bg_stroke,
+            widget_visuals.expansion,
```
emilk, line 65: oh I see, this is all because of `expansion`… which is an annoying feature. Makes sense now

crates/egui/src/widget_style/classes.rs
```diff
@@ -1,33 +1,184 @@
...
+    ///
+    /// A class is never added twice. This never removes a class: use [`Self::set`] for that.
     #[inline]
     fn add_if(&mut self, class: impl Into<ClassName>, condition: bool) {
         if condition {
-            self.classes.push(class.into());
+            let class = class.into();
+            // Always retain and push again, since order of classes can matter.
+            self.classes.retain(|existing| existing != &class);
+            self.classes.push(class);
```
emilk:
```suggestion
            self.set(class, true);
```

crates/egui/src/widget_style/classes.rs
```diff
@@ -67,7 +218,17 @@ pub trait HasClasses {
...
     }
 
-    /// Add the given class by consuming `self` if the condition is true
+    /// Add all the given classes by consuming `self`.
+    #[inline]
+    fn with_classes(mut self, classes: Classes) -> Self
+    where
+        Self: Sized,
+    {
+        self.classes_mut().classes.extend(classes.classes);
```
emilk: here and elsewhere: you are using the private `classes` here leading to a bug: classes rusk being duplicated. We should instead add a `fn extend` on `Classes`.

I guess this is a another motivation for putting each type into its own file

```suggestion
        self.classes_mut().extend(classes);
```

crates/egui/src/widget_style/classes.rs
```diff
@@ -87,7 +248,14 @@ pub trait HasClasses {
         self
     }
 
-    /// Add the given class in-place if the condition is true
+    /// Add all the given classes.
+    #[inline]
+    fn add_classes(&mut self, classes: Classes) -> &mut Self {
+        self.classes_mut().classes.extend(classes.classes);
```
emilk:
```suggestion
        self.classes_mut().extend(classes);
```

crates/egui/src/widget_style/classes.rs
```diff
@@ -87,7 +248,14 @@ pub trait HasClasses {
...
+        self
+    }
+
+    /// Add the given class in-place if the condition is true.
     #[inline]
```
emilk: this should have the same explanation as `Class::add_if`, explaining how it is different from `set_classa

crates/egui/src/widget_style/classes.rs
```diff
@@ -97,13 +265,68 @@ pub trait HasClasses {
...
+        self
+    }
+
+    /// True if the class is present.
+    #[inline]
+    fn has_class(&self, class: &str) -> bool {
+        self.classes()
+            .classes
+            .iter()
+            .any(|existing| existing.as_str() == class)
```
emilk: not a bug, but I still think we should avoid touching the privates

```suggestion
        self.classes().has_class(class)
```

crates/egui/src/widget_style/classes.rs
```diff
@@ -97,13 +265,68 @@ pub trait HasClasses {
...
     }
 
-    /// The list of class
+    /// The list of class.
```
emilk:
```suggestion
    /// The list of classes.
```

crates/egui/src/widget_style/classes.rs
```diff
@@ -1,33 +1,184 @@
...
     classes: SmallVec<[ClassName; 5]>,
 }
 
 impl Classes {
-    /// Add a class to the list if the condition is true
+    /// Add a class to the list if the condition is true.
+    ///
+    /// A class is never added twice. This never removes a class: use [`Self::set`] for that.
     #[inline]
     fn add_if(&mut self, class: impl Into<ClassName>, condition: bool) {
```
emilk: I suggest putting the most complex methods (add_if) first, and the simplest (add, remove, …) first. Remember that this is the order functions show up in the docs!

crates/egui/src/widget_style/mod.rs
```diff
@@ -147,3 +150,14 @@ pub struct StyleArgs<'a> {
...
     pub ctx: &'a Context,
 }
+
+impl StyleArgs<'_> {
+    /// Does the widget or any of its parents contain this class?
+    ///
+    /// See also:
+    /// - [`Classes::has_class`]
+    /// - [`UiStack::has_class`]
+    pub fn has_class(&self, class: &str) -> bool {
```
emilk, line 149: Image

emilk (approved):
looks great!

PR #8465, opened 2026-08-28
Add `AtomLayoutStyle`

lucasmerlin:
Extracts a shared struct with things that once might want to override for any atomlayout-based widget.


crates/egui/src/widget_style/mod.rs
```diff
@@ -48,26 +50,88 @@ impl TextVisuals {
...
+    ///
+    /// A per-widget [`AtomLayout::gap`] wins over [`Self::gap`], so widgets like
+    /// [`crate::DragValue`] can pack their atoms tighter than the theme does.
+    pub fn apply<'a>(&self, mut layout: AtomLayout<'a>) -> AtomLayout<'a> {
+        layout.map_images(|image| {
+            if image.image_options().tint == Color32::WHITE {
+                image.tint(self.image_tint)
+            } else {
+                image
+            }
```
emilk: Won't this cause a weird discontinuity, e.g. if one animates `image_options().tint` smoothly towards `WHITE`?

lucasmerlin: Right... Hmm. Should tint be a `Option<Color32>` instead? It would be very helpful to have the ability to "set tint if no tint was applied yet"

emilk: Option sounds like the right choice, yes. Or `Color32::PLACEHOLDER` if you want to use less space

emilk:
Please be mindful to build/run benchmarks every now and then. I'm worried about bloat

crates/egui/src/widgets/button.rs
```diff
@@ -340,39 +345,29 @@ impl<'a> Button<'a> {
...
+        if let Some(stroke) = stroke {
+            atom_layout_style.frame = atom_layout_style.frame.stroke(stroke);
+        }
         if let Some(fill) = fill {
-            frame = frame.fill(fill);
+            atom_layout_style.frame = atom_layout_style.frame.fill(fill);
         }
         if let Some(corner_radius) = corner_radius {
-            frame = frame.corner_radius(corner_radius);
+            atom_layout_style.frame = atom_layout_style.frame.corner_radius(corner_radius);
```
emilk, line 361: nit: these would maybe be nicer as `frame.set_corner_radius(…);` etc rather than using builder methods

crates/egui/src/widget_style/mod.rs
emilk: As discussed: fix this somehow (e.g. by always multiplying the colors together), then LGTM

emilk (approved):

PR #8469, opened 2026-08-28
Fix zero-duration value animations returning the previous value

cuishuang:
animate_value_with_time returned the previous value for one frame when called with a zero duration because it calculated the current value before updating the animation target.

Handle zero-duration animations before interpolation so they immediately return and store the target value. Add a regression test covering consecutive target changes.


emilk (approved):

cuishuang:
@emilk Thank you for your approval.

PR #8471, opened 2026-08-30
Fix widgets shifting around on hover in interactive container demo

MonterraByte:
This happens because `bg_stroke`'s width isn't constant.

| Before | After |
|--------|--------|
| before | after | 

---


lucasmerlin (approved):
Thanks!

Issue #2108, opened 2022-10-04
Ability to get image paste events


creikey:
Similar to #1167

**Is your feature request related to a problem? Please describe.**
In some applications I want to develop I need to allow the user to paste images into the application.

**Describe the solution you'd like**
There is https://docs.rs/egui/0.19.0/egui/enum.Event.html Paste(String), perhaps a PasteImage(Image) could be added as well

**Describe alternatives you've considered**
n/a

**Additional context**
From what I understand this would need to be added to winit and however egui does its web support, some advice on this would be helpful as I'm interested in implementing this feature myself

creikey:
https://crates.io/crates/clipboard-win it seems like this crate supports image copy paste on windows

emilk:
https://crates.io/crates/arboard also supports image copy/paste, and is cross-platform (except for web).

I think this could be added to `egui` and `eframe` with not too much effort. Unless you really need web support for images you can just not implement it there.

The way I would implement it is to make `arboard` and opt-in feature of `eframe`, and to add `egui::Event::PasteImage(ColorImage)` and `PlatformOutput::copied_image: Option<ColorImage>`.

Its-Just-Nans:
For eframe

Couldn't we get the images from https://github.com/emilk/egui/blob/f6fa74c66578be17c1a2a80eb33b1704f17a3d5f/crates/eframe/src/web/events.rs#L322-L323


```rust
// pseudo code
if data.types() == ["files"] {
   let file_list = data.files();
   let file = file_list.item(0).unwrap(); // to safe
   if file is image {
       let egui_event = egui::Event::PasteImage(file);
   }
}
```

BernardDGS:
Using egui/eframe 0.29.1. We hit this exact gap building a portable app (egui/eframe on Windows):
-  copying a QR-code image with Ctrl+C works fine (Event::Copy), but there is no way to detect a Ctrl+V paste when the clipboard holds an image instead of text 
-  Event::Paste(String) never fires (nothing to extract as text), and neither the raw RawInput.events nor keys_down carry the keypress either, so no application-level workaround exists.

Looked at current main (0.36.1):
- OutputCommand::CopyImage(ColorImage) was added recently (nice!) and egui-winit::Clipboard::set_image() already wraps arboard::Clipboard::set_image() for it — but there's no get_image() counterpart, and Event::Paste is still text-only. 
[...]

PR #8472, opened 2026-08-30
Add clipboard image paste support (`Event::PasteImage`)

BernardDGS:
Adds `egui::Event::PasteImage`, emitted on Ctrl+V/Cmd+V when the clipboard holds an
image with no usable text representation (mirrors `OutputCommand::CopyImage` for
the opposite direction). `egui-winit::Clipboard` gains `get_image()`, backed by
`arboard::Clipboard::get_image()` — the symmetric counterpart to the `set_image()`
already used for copy. Wired in the winit keyboard-shortcut path and in both the
glow and wgpu eframe backends' AccessKit-driven paste action.

Straight (unmultiplied) alpha from the OS clipboard is converted via the existing
`ColorImage::from_rgba_unmultiplied` helper rather than reinterpreting bytes
directly, since the two aren't bit-compatible for translucent pixels.

Not covered here: the web (wasm32) target — browser clipboard image reads are
inherently async (`Blob::array_buffer` returns a Promise), while `Event::PasteImage`
as designed here carries an already-decoded `ColorImage`. Bridging that needs
either an async event payload or a deferred/queued event on `AppRunner` — more of
a design decision than a small mirrored addition, flagging it rather than guessing.

* Related: #2108 (not fixed: web)


crates/egui-winit/src/clipboard.rs
```diff
@@ -144,6 +167,19 @@ impl Clipboard {
     }
 }
 
+/// Converts `arboard`'s straight-alpha `ImageData` into an [`egui::ColorImage`], which stores
+/// _premultiplied_ alpha (see [`egui::Color32`]) — a plain byte-cast would be wrong for any
+/// pixel with partial transparency, hence [`egui::ColorImage::from_rgba_unmultiplied`] (the same
+/// helper the `image` crate loading example in its own docs uses) rather than reinterpreting
+/// the bytes directly, as `Self::set_image` does for the write direction.
```
emilk:
```suggestion
```

emilk (approved):

emilk:
Trying to paste an image now prints a warning:

> [2026-09-01T09:48:49Z ERROR egui_winit::clipboard] arboard paste error: The clipboard contents were not available in the requested format or the clipboard is empty.

This is `clipboard.get_text()` returning an error

BernardDGS:
> This is `clipboard.get_text()` returning an error

Good catch — fixed in f0e4dbee. The paste handler tries get() first, then falls back to get_image() when there's no text — but get() was already logging arboard::Error::ContentNotAvailable at error level on its own, before that fallback even got a chance to run. That's the expected case when the clipboard holds an image instead of text (and symmetrically in get_image(), when it holds neither). Both now just skip the log for that specific error and still return None.

I also pulled the classification into its own is_expected_content_absence() and added a unit test for it, since nothing had exercised this distinction before — that's exactly how it slipped through in the first place.

emilk (approved):
Awesome stuff - thanks!

PR #8475, opened 2026-09-01
Fix `sin_in`, `sin_out`, and `circular_out` easing functions

dylanpulver:
`emath::easing::sin_in`, `sin_out`, and `circular_out` are numerically wrong: the sine cases multiply by `2π` where they should use `π/2`, and `circular_out` groups the `sqrt` incorrectly. The module doc promises every easing maps `[0, 1] -> [0, 1]` with `f(0) == 0`, `f(1) == 1`, and each links its easings.net curve, but on `main`:

- `sin_in(0.0) == 1.0` (should be 0), `sin_in(0.25) == 2.0` — 98 of 101 samples fall outside `[0, 1]`
- `sin_out(1.0) == 0.0` (should be 1), dipping to `-1.0` at `t = 0.75`
- `circular_out(0.25) == 0.33` vs the reference `0.66`

So an animation using these (e.g. via `animate_bool_with_easing`) flickers and overshoots instead of easing.

Root cause: the file is ported from AHEasing, whose sine/circular cases use `M_PI_2` (π/2); that was read as "π × 2". The `sin_in_out` sibling (already `t * PI`) and `circular_in` are correct, which is why it stayed hidden — and `easing.rs` had no tests, while egui itself only calls `linear`/`cubic_out`/`quadratic_out` internally.

Fix: `π/2` for the two sine cases and `((2. - t) * t).sqrt()` for `circular_out`. All three now match their easings.net reference to `< 5e-7` over 1000 samples. Adds a contract test (endpoints + `[0, 1]` range, plus reference-curve match) that fails on the current code and passes after.

Disclosure: diagnosis and patch prepared with AI assistance; opened as a draft for human review per CONTRIBUTING.

### DIFF
--- a/crates/emath/src/easing.rs
+++ b/crates/emath/src/easing.rs
@@ -77,7 +77,7 @@ pub fn cubic_in_out(t: f32) -> f32 {
 /// Modeled after quarter-cycle of sine wave
 #[inline]
 pub fn sin_in(t: f32) -> f32 {
-    ((t - 1.) * 2. * PI).sin() + 1.
+    ((t - 1.) * 0.5 * PI).sin() + 1.
 }
 
 /// <https://easings.net/#easeOuSine>
@@ -85,7 +85,7 @@ pub fn sin_in(t: f32) -> f32 {
 /// Modeled after quarter-cycle of sine wave (different phase)
 #[inline]
 pub fn sin_out(t: f32) -> f32 {
-    (t * 2. * PI).sin()
+    (t * 0.5 * PI).sin()
 }
 
 /// <https://easings.net/#easeInOutSine>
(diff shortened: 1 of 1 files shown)


crates/emath/src/easing.rs
```diff
@@ -77,15 +77,15 @@ pub fn cubic_in_out(t: f32) -> f32 {
 /// Modeled after quarter-cycle of sine wave
 #[inline]
 pub fn sin_in(t: f32) -> f32 {
-    ((t - 1.) * 2. * PI).sin() + 1.
+    ((t - 1.) * 0.5 * PI).sin() + 1.
```
Mingun, line 80: There is already [FRAC_PI_2](https://doc.rust-lang.org/std/f32/consts/constant.FRAC_PI_2.html) constant

Issue #513, opened 2021-06-25
Concave polygon fill and the plot widget


jbuckmccready:
From what I can tell by reading the docs/playing with the code currently only convex polygon filling is supported. Is supporting concave (or possibly even complex polygons) a future goal for egui or is that outside the scope of the project? I think something like [Lyon](https://github.com/nical/lyon) could be used for tessellation, and I saw the [Mesh](https://docs.rs/epaint/0.13.0/epaint/struct.Mesh.html) type so I can probably combine the two, but I didn't see support for the mesh type with the plot widget.

I really like the egui API in general and the recent improvements to the plot widget are very nice. I would like to be able to add concave polygons to plots. I am currently using wasm + html/js to create the [cavalier contours](https://github.com/jbuckmccready/cavalier_contours) web demo (web demo link [here](https://jbuckmccready.github.io/cavalier_contours_web_demo_page/)), but the egui API looks very nice and keeps everything in Rust.

I looked briefly at the code and I see the plot widget structs use the shape types/rendering under the hood. One option I see is just adding generic mesh object support to the plot widget (kind of a low level interface for the widget), another possibility is building out support for concave polygons in general, any thoughts/plans already around this?

EmbersArc:
I touched on this in https://github.com/emilk/egui/pull/471 but haven't gotten a response yet:

> I also have an implementation of the ear-clipping algorithm lying around that can be used to tessellate non-convex polygons. It's relatively few lines of code, but it's not the most efficient algorithm with O(n²) time complexity. I could add this to the plotting library, but maybe it's something for egui in general?

More efficient algorithms are exponentially more difficult to implement, so going with `lyon`'s faster algorithms may be a good option as well.

emilk:
The problem with `lyon` is that it doesn't support feathering (what `epaint` uses to smooth edges) AFAIK. Creating a convex filling algorithm that also supports feathering seems like a big task, but I agree it would be very nice to have.

There is another alternative: turn off feathering and turn on MSAA to get smoother edges. This is already supported on `eframe` native, and it could be added to `egui_web` too. Then we could just use `lyon` directly to produce `epaint::Mesh`es.

TYPEmber:
Maybe we could decompose one concave polygon into some convex polygons, and store them.
And I can implement it if possible.

techninja1008:
Just came across this same issue - has any work towards this been done? My use case in particular is filling the area underneath a line plot.

[...]

PR #8476, opened 2026-09-01
Add `BandShape`

emilk:
Add `BandShape` to epaint, supporting and varying width band along some direction.

This can be useful in egui_plot, e.g. for filling the area under a curve, or for visualizing a audio waveform, for instance.

The band has feathering for the fill, plus an optional stroke that is just drawn on top of the fill.

This solves a small subset of non-convex polygons (since the general solution is quite difficult, at least if you want feathered edges).

Screenshot 2026-09-01 at 20 56 33

Includes an interactive waveform demo and alpha screenshot coverage.

Related work:

- https://github.com/emilk/egui/issues/513
- https://github.com/emilk/egui_plot/pull/260


crates/epaint/src/tessellator.rs
```diff
@@ -752,12 +766,121 @@ fn cw_signed_area(path: &[PathPoint]) -> f64 {
...
+
+    for i in 0..num_points / 2 - 1 {
+        out.add_triangle(vertex(i), vertex(i + 1), vertex(last - i - 1));
+        out.add_triangle(vertex(i), vertex(last - i - 1), vertex(last - i));
+    }
+
+    if 0.0 < min_width {
+        for (i, point) in run.iter().enumerate() {
+            let width = point.y.span();
+            if width < min_width {
```
emilk, line 850:
```suggestion
            if width < min_width {
                // Fade out the fill when the width is below the feathering width
```

crates/egui_demo_lib/tests/snapshots/fixed_width_band_matches_path_stroke.png
lucasmerlin, line 1: I think specific snapshots like this should ideally be self-describing with e.g. a label describing what is what and what is the expected result. Looking at this in kitdiff I'd have no idea what I'm seeing

lucasmerlin (approved):
It's cool but it also feels oddly-specific and niche, especially since it's horizontal only. 

Couldn't the shape live in egui_plot and added to the painter as a Mesh? And make the helpers like `stroke_and_fill_path` public?

emilk:
I agree that it is pretty niche

It is not horizontal-only - it can be rotated

lucasmerlin:
> It is not horizontal-only - it can be rotated

Oh, missed that, nice!

emilk:
Yes, maybe this belongs in egui_plot… I'll give it some thought some other day

emilk:
Interestingly, we just added something very similar to this in Rerun with measurement variance:

image

https://github.com/rerun-io/reality/pull/1992

So maybe not so niche after all

emilk:
I vote we add this for now. It's also a nice fit for rendering a audio envelope

PR #8477, opened 2026-09-02
Add tests for `StyleProvider` and ensure every property works

lucasmerlin:
Follow up to #8455

Based on #8485


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -382,17 +382,26 @@ impl<'t> TextEdit<'t> {
...
+        self
+    }
+
     /// Set the horizontal align of the inner text.
+    #[deprecated = "Use `align` instead"]
     #[inline]
     pub fn horizontal_align(mut self, align: Align) -> Self {
-        self.align.0[0] = align;
+        self.align = Some(self.align.unwrap_or(Align2::LEFT_TOP).with_x(align));
         self
```
lucasmerlin, line 397: The test showed that TextEdit ignored the `align2` field from the new `AtomLayoutStyle`, because it always overwrote it with LEFT_TOP. 

I made it `Option`, so that it can correctly fall back to the styleprovider now. 
But that means `horizontal_align` is kinda unintuitive now since it also overwrites `vertical_align`, so I deprecated them in favor of a `align` that sets both.

We could instead also change the type to `[Option<Align>; 2]`, then we could keep separate `horizontal/vertical_align`

crates/egui/src/widget_style/mod.rs
```diff
@@ -155,6 +155,8 @@ impl WidgetStyle for TextEditStyle {}
 #[derive(Debug, Clone)]
 pub struct CheckboxStyle {
     /// Style of the checkbox's atom layout.
+    ///
+    /// [`AtomLayoutStyle::align2`] vertical align has no effect, the checkbox is always centered.
```
emilk:
```suggestion
    /// [`AtomLayoutStyle::align2`] vertical align has no effect; the checkbox is always centered.
```

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -382,17 +382,26 @@ impl<'t> TextEdit<'t> {
...
     }
 
     /// Set the vertical align of the inner text.
+    #[deprecated = "Use `align` instead"]
     #[inline]
     pub fn vertical_align(mut self, align: Align) -> Self {
-        self.align.0[1] = align;
+        self.align = Some(self.align.unwrap_or(Align2::LEFT_TOP).with_y(align));
```
emilk: simpler

```suggestion
        self.align.get_or_insert(Align2::LEFT_TOP).set_y(align);
```

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -382,17 +382,26 @@ impl<'t> TextEdit<'t> {
...
+        self.align = Some(align);
+        self
+    }
+
     /// Set the horizontal align of the inner text.
+    #[deprecated = "Use `align` instead"]
     #[inline]
     pub fn horizontal_align(mut self, align: Align) -> Self {
-        self.align.0[0] = align;
+        self.align = Some(self.align.unwrap_or(Align2::LEFT_TOP).with_x(align));
```
emilk: same here

emilk (approved):

PR #8485, opened 2026-09-03
Add `Align2` helpers

lucasmerlin:
Add utilities to make it easier to construct and use `Align2`

### DIFF
--- a/crates/emath/src/align.rs
+++ b/crates/emath/src/align.rs
@@ -163,6 +163,37 @@ impl Align2 {
 }
 
 impl Align2 {
+    #[inline(always)]
+    pub fn new(x: Align, y: Align) -> Self {
+        Self([x, y])
+    }
+
+    /// Set align on the horizontal axis.
+    #[inline]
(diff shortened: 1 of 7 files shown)


crates/emath/src/align.rs
```diff
@@ -183,19 +214,19 @@ impl Align2 {
...
     /// Flip on the y-axis
     /// e.g. `TOP_LEFT` -> `BOTTOM_LEFT`
     pub fn flip_y(self) -> Self {
-        Self([self.x(), self.y().flip()])
+        self.with_y(self.y().flip())
     }
 
     /// Flip on both axes
     /// e.g. `TOP_LEFT` -> `BOTTOM_RIGHT`
     pub fn flip(self) -> Self {
```
emilk, line 264: No fault of this PR, but these three are all foot-gun-shaped. `flip` sounds like it takes `&mut`

They should all be marked `#[must_use]`. Even better, they should be renamed `flip` -> `flipped`

emilk (approved):

PR #8487, opened 2026-09-03
Add new `render_every_step` flag to egui_kittest

lucasmerlin:
Some applications / test depend on specific gpu state and may be doing gpu-readback. This adds a `always_render` flag, which calls `render` every frame so that the gpu dependent logic will work correctly.
It also saves the last render result so we don't have to render the same frame twice.


crates/egui_kittest/src/builder.rs
```diff
@@ -127,6 +133,16 @@ impl<State> HarnessBuilder<State> {
         self
     }
 
+    /// Should every pass be rendered?
```
emilk: every pass, or every step? or is that the same thing?

lucasmerlin: Good catch, one step can be multiple passes

crates/egui_kittest/src/builder.rs
```diff
@@ -127,6 +133,16 @@ impl<State> HarnessBuilder<State> {
         self
     }
 
+    /// Should every step be rendered?
+    ///
+    /// Useful when test logic requires some specific gpu logic, e.g. reading data back from the gpu.
+    #[cfg(any(feature = "wgpu", feature = "snapshot"))]
+    #[inline]
+    pub fn with_always_render(mut self, always_render: bool) -> Self {
+        self.always_render = always_render;
```
emilk:
```suggestion
    /// Should every step be rendered?
    ///
    /// Useful when test logic requires some specific gpu logic, e.g. reading data back from the gpu.
    #[cfg(any(feature = "wgpu", feature = "snapshot"))]
    #[inline]
    pub fn with_render_every_step(mut self, render_every_step: bool) -> Self {
        self.render_every_step = render_every_step;
```

crates/egui_kittest/src/lib.rs
```diff
@@ -99,6 +99,19 @@ pub struct Harness<'a, State = ()> {
...
+    ///
+    /// A pass should be rendered at most once. A paint callback can do GPU work of its own —
+    /// schedule a readback, say — and running it a second time for the same pass corrupts that
+    /// work.
+    #[cfg(any(feature = "wgpu", feature = "snapshot"))]
+    last_render: Option<(u64, image::RgbaImage)>,
+
+    /// Render every pass. See [`HarnessBuilder::with_always_render`].
+    #[cfg(any(feature = "wgpu", feature = "snapshot"))]
+    always_render: bool,
```
emilk:
```suggestion
    /// Render every step. See [`HarnessBuilder::with_render_every_step`].
    #[cfg(any(feature = "wgpu", feature = "snapshot"))]
    render_every_step: bool,
```

crates/egui_kittest/src/lib.rs
```diff
@@ -293,6 +316,12 @@ impl<'a, State> Harness<'a, State> {
         self.renderer.handle_delta(&mut output.textures_delta);
         self.output = output;
 
+        #[cfg(any(feature = "wgpu", feature = "snapshot"))]
+        if self.always_render {
+            self.render()
+                .expect("Failed to render during `always_render`");
```
emilk:
```suggestion
                .expect("Failed to render during `render_every_step`");
```

emilk (approved):

PR #8489, opened 2026-09-03
⚠️ Make `ViewportCommand::Screenshot` callback based

lucasmerlin:
Migration guide for the changelog:
`ViewportCommand::Screenshot` was changed to be callback based.
Previously you needed to iter the input events to check for the screenshot (but didn't need interior mutability):
```rs
[derive(Default)]
struct MyApp {
    screenshot: Option<Arc<egui::ColorImage>>,
}
impl eframe::App for MyApp {
        if ui.button("Screenshot").clicked() {
            ui.send_viewport_cmd(ViewportCommand::Screenshot(UserData::default()));
        }

        let image = ui.input(|i| {
            i.events
                .iter()
                .filter_map(|e| {
                    if let egui::Event::Screenshot { image, .. } = e {
                        Some(Arc::clone(image))
                    } else {
                        None
                    }
                })
                .next_back()
        });
        
        if let Some(image) = image {
            self.screenshot = Some(image);
        }
    }
}
```

Now you need to use a mutex, channel, or some other sync primitive to get the result:
```rs
#[derive(Default)]
struct MyApp {
    screenshot: Arc<Mutex<Option<Arc<egui::ColorImage>>>>,
}
impl eframe::App for MyApp {
        if ui.button("Screenshot").clicked() {
            let ctx = ui.ctx().clone();
            let screenshot_mutex = self.screenshot.clone();
            ui.request_screenshot(move |screenshot| {
                screenshot_mutex.lock().replace(screenshot);
                ctx.request_repaint();
            });
        }
    }
}
```
----

This implements a callback-based screenshot api for egui

Upsides over `ViewportCommand::Screenshot`:
- More convenient api: `ctx.request_screenshot(|screnshot| ...)` (and no need to downcast UserData)
- Reading screenshot data doesn't require rendering another frame
  - Should allow us to add recording as a plugin without special hooks in egui_kittest (otherwise capturing the last frame would have required custom hooks in kittest)
  - This was also a problem in [kittest_inspector](https://github.com/rerun-io/kittest_inspector): When streaming a real app, the frame seen was always a couple frames behind when the app was idle


crates/egui/src/data/user_data.rs
```diff
@@ -73,3 +78,69 @@ impl<'de> serde::Deserialize<'de> for UserData {
...
     }
 }
+
+/// A one-shot callback for receiving a screenshot.
+///
+/// Clones share the same callback. Whichever clone is completed first consumes it, so it is
+/// invoked at most once.
+#[derive(Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub struct ScreenshotCallback {
```
emilk, line 88: belongs in its own file

crates/egui/src/data/user_data.rs
```diff
@@ -73,3 +78,69 @@ impl<'de> serde::Deserialize<'de> for UserData {
...
+            })))),
+        }
+    }
+
+    /// Complete this screenshot request.
+    ///
+    /// Callback requests are invoked immediately. Requests created for the legacy event-based
+    /// API instead return the event that the integration should enqueue for the next pass.
+    #[doc(hidden)]
+    pub fn complete(self, viewport_id: ViewportId, image: Arc<ColorImage>) -> Option<Event> {
```
emilk, line 125: we should mark the `#[must_use]`, or produce a warning if the callback has already been consumed (is likely a bug)

crates/egui/src/viewport.rs
```diff
@@ -1198,6 +1198,12 @@ pub enum ViewportCommand {
     /// The results are returned in [`crate::Event::Screenshot`].
     Screenshot(crate::UserData),
```
emilk, line 1199: Link to the new alternative

crates/egui/src/viewport.rs
```diff
@@ -1198,6 +1198,12 @@ pub enum ViewportCommand {
...
 
+    /// Take a screenshot of the next frame after this and pass it to a callback.
+    ///
+    /// Unlike [`Self::Screenshot`], this does not deliver an input event or require another egui
+    /// pass. Use [`crate::Context::request_screenshot`] for a convenient way to send this command.
+    ScreenshotCallback(crate::ScreenshotCallback),
```
emilk, line 1205: Should we still keep the old? 🤔

emilk (approved):

crates/egui/src/viewport.rs
lucasmerlin: I could deprecate it

emilk: but is there a case where the old command is more ergonomic for the user? I haven't thought it through

lucasmerlin: 🤷‍♂️Probably not. 
I guess the benefit of `ViewportCommand::Screenshot` is that it's serializable (Even though `UserData` isn't). But I don't think there are any backends that make use of that? 

I'll remove it

emilk (approved):
let's add a quick migration guide in the changelog

Issue #7325, opened 2025-07-09
Provide a way to reduce memory usage by fonts


kjarosh:
**Is your feature request related to a problem? Please describe.**

Our application ([Ruffle](https://github.com/ruffle-rs/ruffle)) has to support multiple languages in GUI at the same time (so we want to display latin, arabic, hebrew, CJK, etc. text). In order to do that we're loading fonts, like Noto Sans CJK KR, Noto Sans CJK JP, Noto Sans CJK SC, Noto Sans CJK TC, Noto Sans Hebrew, Noto Sans Arabic. Unfortunately they end up taking a lot of memory, because they all have to be provided as `Vec<u8>`.

We're planning to use fontconfig to provide that list of fonts, which can actually make it even worse.

**Describe the solution you'd like**

We'd like egui to support fonts without loading them into memory in their entirety. An example solution would be to accept `File` objects that represent those fonts, and lazily load glyphs from the files when needed, without keeping their contents in memory.

**Describe alternatives you've considered**

We were thinking about mmapping font files into memory, but it's inherently unsafe in Rust.

**Additional context**

White-Rabbit-Scientific:
Wow, Ruffler is a really nice piece of work !

PR #8490, opened 2026-09-03
Render any glyph and colored emoji on web

emilk:
* Closes https://github.com/emilk/egui/pull/7852
* Web-only support for https://github.com/emilk/egui/issues/2551
* Related to https://github.com/emilk/egui/issues/7325

This adds support for rendering _any glyph_ on web, including colored emojis!

Test string: ✨ We support ไทย, 한국어, and 🦀 🚀 🤖/🦾/📷/🎞️

Screenshot 2026-09-03 at 20 29 43
Screenshot 2026-09-03 at 20 29 36

PR chain, in order:
* #8490
* #8500
* #8501
* #8502
* #8503
* #8504
* #8491
* #8492
* #8493
* #8508


crates/epaint/src/tessellator.rs
```diff
@@ -2077,6 +2089,11 @@ impl Tessellator {
                             color = *fallback_color;
                         }
 
+                        if color_glyph_vertices.iter().any(|range| range.contains(&i)) {
```
emilk: This is potentially slow, if there are a lot of emojis spaces out over a large text

emilk: maybe we can binary-search into `row.glyphs` instead to find the matching glyph (if any)? no need for `color_glyph_vertices` at all

MichaelGrupp:
A thing to consider: emoji characters were a cheap simple way to have UI icons (e.g. for buttons) that rendered in the UI style/color scheme. Now they would show up as colored emojis on the web. Breaking visual change for a bunch of UIs? 😬 

For example the egui demo upstream vs. this branch:
grafik
grafik

emilk:
Yes, this is a downside. I added a way (`GlyphSource`) to customize which characters goes to the fallback, so it is fixable in userspace.

I'm not sure we'll keep `GlyphSource` though. Maybe we just remove `NotoEmoji-Regular.ttf` and `emoji-icon-font.ttf` (or make them opt-in) now that they are no longer needed.

crates/eframe/src/web/canvas_glyphs.rs
```diff
@@ -0,0 +1,218 @@
...
+        if self.is_tofu(&white, &font, *subpixel_offset_px) {
+            return None; // Let egui draw its own replacement glyph instead.
+        }
+
+        // Color glyphs (e.g. emoji) ignore the fill style, so they come out
+        // the same in black. Comparing the two is exact, unlike inspecting
+        // the pixel colors: the browser's alpha (un)premultiplication rounds
+        // anti-aliased white text to slightly off-white.
+        let black = self.draw(cluster, &font, "black", *subpixel_offset_px)?;
+        let is_color = white.rgba == black.rgba;
```
lucasmerlin: how is inspecting pixel colors different? I don't get the comment

crates/eframe/src/web/canvas_glyphs.rs
```diff
@@ -0,0 +1,218 @@
...
+            left,
+            ascent,
+            advance,
+        } = white;
+
+        Some(RasterizedGlyph {
+            image: ColorImage::from_rgba_unmultiplied([width as _, height as _], &rgba),
+            // The pen sits at `(PADDING + left, PADDING + ascent)` in the image,
+            // so the image's top-left is this far from the pen:
+            offset_px: vec2(-(left + PADDING) as f32, -(ascent + PADDING) as f32),
```
lucasmerlin, line 114: Its weird that the const is defined below where it's used

crates/eframe/src/web/canvas_glyphs.rs
```diff
@@ -0,0 +1,218 @@
...
+    advance: f64,
+}
+
+thread_local! {
+    // `Context` and `GlyphRasterizer` are `Send + Sync`, but browser Canvas handles are not.
+    // Keeping the handles here lets the callback stay capture-free and usable from `egui::Context`.
+    static CANVAS_GLYPHS: RefCell<Option<CanvasGlyphs>> = const { RefCell::new(None) };
+}
+
+pub(super) fn glyph_rasterizer() -> GlyphRasterizer {
```
lucasmerlin, line 252: It'd be awesome if we could make it public, since I'd like to add custom emojis to my app, and then my custom emoji GlyphRasterizer could fallback to this one for the built in one (alternative: having some way to add multiple fallback glyph rasterizers)

crates/egui/src/context.rs
```diff
@@ -591,7 +594,16 @@ impl ContextImpl {
...
+            let mut fonts = Fonts::new(
+                text_options,
+                self.font_definitions.clone(),
+                self.glyph_rasterizer.clone(),
+            );
+            if let Some(prefer) = &self.glyph_source_preference {
+                let prefer = Arc::clone(prefer);
+                fonts = fonts.with_glyph_source_preference(move |cluster| prefer(cluster));
+            }
+            fonts
```
lucasmerlin, line 605: If this is only read during init, doesn't that make the `set_glyph_rasterizer` and `set_glyph_source_preference` below broken/misleading when called after init?

lucasmerlin (approved):
Awesome! 

I wonder if we could even make the built in font rasterizer optional in a follow-up PR, that should make wasm apps a lot smaller, not having to include vello, skrifa and font definitions

lucasmerlin:
On firefox for android my emojis turn dark sometimes 🤔
Arguably it's a niche browser and egui is half-broken anyways (look at the ugly font rasterization), but still weird. Looks like the color font detection isn't working right and it's tinting them?

https://github.com/user-attachments/assets/3cd75730-4704-4c0f-bbef-85d54dc83afc

emilk:
> I wonder if we could even make the built in font rasterizer optional in a follow-up PR, that should make wasm apps a lot smaller, not having to include vello, skrifa and font definitions 

I suspect the kerning will be quite bad, but easy enough to try!

crates/eframe/src/web/canvas_glyphs.rs
emilk: Multiple fallback rasterizers makes sense to me - that's a great way to hook in custom emojis! But I'll leave that to you for a later PR

crates/egui/src/context.rs
emilk: `set_glyph_rasterizer` and `set_glyph_source_preference` sets `ctx.fonts = None;` which triggers the code above to run next frame

emilk:
> On firefox for android my emojis turn dark sometimes 🤔

I've improved the emoji detector now - please try again!

PR #8491, opened 2026-09-03
Render any glyph on native

### LINKED ISSUE #5233: Automatically load system fonts

emilk:
* Closes <https://github.com/emilk/egui/issues/5233>
* Follows #8490 and the font refactor chain ending in #8504

This PR adds support for any (monochrome) glyph on Native: Korean, Thai, Chinese, Japanese, … as long as there is a font somewhere on the system that supports it (and there likely is).

Tested on macOS (日 → PingFang SC, ع → Geeza Pro, א → Lucida Grande). Untested on Windows, Linux, and Android.

Implemented by Claude.

PR chain, in order:
* #8490
* #8500
* #8501
* #8502
* #8503
* #8504
* #8491
* #8492
* #8493
* #8508
* #8509
* #8510
* #8511

### DIFF
--- a/crates/egui_system_fonts/src/lib.rs
+++ b/crates/egui_system_fonts/src/lib.rs
@@ -247,4 +247,22 @@ mod tests {
             assert!(insert.name.starts_with("system:"));
         }
     }
+
+    #[test]
+    #[ignore = "Depends on which fonts are installed on this machine"]
+    fn renders_cjk_through_egui_context() {
+        let ctx = egui::Context::default();
+        let font_id = egui::FontId::proportional(14.0);
+        let mut output = ctx.run_ui(Default::default(), |ui| {
+            ui.fonts_mut(|fonts| assert!(!fonts.has_glyph(&font_id, '日')));
+        });
+        output.textures_delta.clear();
+
+        ctx.add_font_provider(Arc::new(SystemFontProvider::new()));
+        let mut output = ctx.run_ui(Default::default(), |ui| {
+            ui.fonts_mut(|fonts| assert!(fonts.has_glyph(&font_id, '日')));
+            assert_eq!(ui.fonts(|fonts| fonts.provided_fonts().len()), 1);
+        });
+        output.textures_delta.clear();
+    }
 }
--- a/crates/egui_system_fonts/src/lib.rs
+++ b/crates/egui_system_fonts/src/lib.rs
@@ -9,8 +9,6 @@
 //! egui_ctx.add_font_provider(std::sync::Arc::new(egui_system_fonts::SystemFontProvider::new()));
 //! ```
 
-use std::sync::{Arc, OnceLock};
-
 use epaint::{
     mutex::Mutex,
     text::{
@@ -22,6 +20,7 @@ use fontique::{
     Collection, CollectionOptions, FallbackKey, GenericFamily, Language, QueryFamily, QueryFont,
     QueryStatus, Script, SourceCache, SourceCacheOptions,
 };
+use poll_promise::Promise;
 
 struct SystemFonts {
     collection: Collection,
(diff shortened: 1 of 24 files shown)


crates/egui_system_fonts/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+/// The font files are memory-mapped, not copied.
+///
+/// Only fonts with outline glyphs are returned. System emoji fonts contain
+/// bitmaps or color glyphs, which epaint cannot render yet, so they are skipped.
+///
+/// Enumerating the system fonts can take hundreds of milliseconds,
+/// so [`Self::new`] starts doing that on a background thread.
+/// A lookup before it is done blocks until it is.
+pub struct SystemFontProvider {
+    fonts: Arc<OnceLock<Mutex<SystemFonts>>>,
```
emilk: I think this would be a lot cleaner using poll_promise, or a channel

crates/epaint/src/text/fonts.rs
```diff
@@ -224,6 +224,64 @@ mod has_emoji_presentation_tests {
...
 
+/// Input to a [`FontProvider`].
+pub struct FallbackRequest<'a> {
+    /// A grapheme cluster that no installed font can render.
+    pub cluster: &'a str,
+
+    /// The first character of [`Self::cluster`].
+    ///
+    /// The returned font must have a glyph for it.
+    pub base_char: char,
```
emilk: why include both? the receiver can just do `cluster.chars().next()`?

umajho:
Hello. It seems this is still a work in progress. I wonder if the final version will support changing `locale` at runtime?

Chinese characters have different typographic standards across regions, even when they share the same Unicode code point. The most famous example is the character “骨”:
| Mainland China | Japan |
|-|-|
|  |  |

`日 → PingFang SC` shows that the mechanism for choosing a fallback font work: on macOS, when no preferred language that uses Chinese characters is configured, it chooses a Simplified Chinese font.
On the other hand, this is not a good example for Japanese[^1], because `PingFang SC` is a font for **S**implified **C**hinese, which uses the standard for mainland China.
(A typical Japanese user will have Japanese as their most preferred language among languages that use Chinese Characters, so the fallback should be a Japanese font for them.)

If we can change `locale` at runtime, app developers could simply set the `locale` based on the interface language they provide. This would ensure that the standard for mainland China is used for Simplified Chinese, while other appropriate standards are used for other languages.

Ideally, I think we should be able to set `locale` in `Visuals`, or perhaps through the new style system as well (I'm still not familiar with it). For example, a two-panel translator app could render Chinese characters according to the standard for mainland China for `source: Chinese (Simplified)` in the left panel and according to the standard for Japan for `target: Japanese` in the right panel. Though, I think this could be addressed in a follow-up PR.

Even further, perhaps we could allow developers to specify a font-family list directly in `Visuals` or through the new style system, similar to what CSS allows.

[^1]: I assume this is intended to demonstrate Japanese because “日” is the first character in “日本語”, which means the Japanese language.

emilk:
Runtime locale is already possible on this branch, just not through Visuals:

```rs
ctx.set_font_providers(vec![Arc::new(
    SystemFontProvider::new().with_locale(Some("ja")),
)]);
```

I'm open to making this more ergonomic and findable

crates/epaint/src/text/fonts.rs
```diff
@@ -440,7 +488,24 @@ impl FontsImpl {
...
+    }
+
+    /// Like [`Self::resolve_face`] for the first char of `cluster`,
+    /// but lets the [`FontProvider`]s see the whole grapheme cluster.
+    #[inline]
+    pub fn resolve_cluster_face(
+        &mut self,
+        family: FamilyKey,
+        cluster: &str,
+        base_char: char,
```
emilk: Here and elsewhere: do we need both these arguments?

emilk: No. `base_char` was always the first char of `cluster`, so `resolve_cluster_face` and `Family::resolve_cluster` now take only the cluster and derive it. An empty cluster resolves to the replacement face.

Claude

PR #8492, opened 2026-09-03
Render colored emojis on native

emilk:
* Closes <https://github.com/emilk/egui/issues/2551>
* Closes https://github.com/emilk/egui/issues/7325

System emoji fonts store their glyphs as `sbix`/`CBDT` bitmaps or `COLR` paint graphs, often next to *empty* outlines. epaint could only render outlines, so #8491 skipped them. This PR adds bitmap and `COLR` glyph rendering to `FontFace`, behind the new epaint feature `color_fonts`, using `vello_cpu`'s text support. Glyphs are marked `is_color`, so the tessellator leaves their colors alone.

`egui_system_fonts` enables the feature and now accepts system color fonts, so eframe native renders 🦀 🤖 🦾 in color from Apple Color Emoji (macOS), and should from Segoe UI Emoji and Noto Color Emoji too (untested). Web is unaffected: the feature adds a PNG decoder, and the browser already renders emoji there.

`GlyphSource::Rasterizer` is now `GlyphSource::Platform` (discovered fonts + font providers + glyph rasterizer), and the preference moved from `GlyphRasterizer::prefer` to `Context::set_glyph_source_preference` / `Fonts::with_glyph_source_preference`, so it also works on native. Emoji-presentation clusters resolve platform-first, so 🚀 and 📷 come from the system emoji font even though the bundled NotoEmoji has them, while text-presentation symbols like ⏮ still come from the bundled fonts.

Implemented by Claude.

PR chain, in order:
* #8490
* #8500
* #8501
* #8502
* #8503
* #8504
* #8491
* #8492
* #8493
* #8508
* #8509
* #8510
* #8511

### DIFF
--- a/crates/egui_system_fonts/src/lib.rs
+++ b/crates/egui_system_fonts/src/lib.rs
@@ -33,9 +33,6 @@ struct SystemFonts {
 /// so e.g. 日本語 gets a Japanese font and العربية an Arabic font.
 /// The font files are memory-mapped, not copied.
 ///
-/// Only fonts with outline glyphs are returned. System emoji fonts contain
-/// bitmaps or color glyphs, which epaint cannot render yet, so they are skipped.
-///
 /// Enumerating the system fonts can take hundreds of milliseconds,
 /// so [`Self::new`] starts doing that on a background thread.
 /// A lookup before it is done blocks until it is.
@@ -123,12 +120,13 @@ impl FontProvider for SystemFontProvider {
             query.set_families([QueryFamily::Generic(generic)]);
             query.set_fallbacks(FallbackKey::new(script, self.locale.as_ref()));
         } else {
-            // Punctuation, symbols, etc. belong to no script, so there is no fallback for them.
+            // Punctuation, symbols, emoji, etc. belong to no script, so there is no fallback for them.
             // Try the generic families instead.
             query.set_families(
                 [
                     generic,
                     GenericFamily::SystemUi,
+                    GenericFamily::Emoji,
                     GenericFamily::Math,
                     GenericFamily::Serif,
                     GenericFamily::Monospace,
@@ -139,7 +137,7 @@ impl FontProvider for SystemFontProvider {
 
         let mut found = None;
         query.matches_with(|font| {
-            if has_outline_for(font, base_char) {
+            if can_render(font, base_char) {
                 found = Some(font.clone());
                 QueryStatus::Stop
             } else {
@@ -177,10 +175,10 @@ fn script_of(c: char) -> Option<Script> {
     }
 }
 
-/// Does the font have an outline glyph for the character?
+/// Does the font have a glyph for the character that epaint can render?
 ///
-/// False for bitmap and color glyphs, which epaint cannot render yet.
-fn has_outline_for(font: &QueryFont, c: char) -> bool {
+/// That is an outline, a bitmap (`sbix`, `CBDT`), or a `COLR` glyph.
+fn can_render(font: &QueryFont, c: char) -> bool {
     use skrifa::MetadataProvider as _;
 
     let Some(glyph_id) = font.charmap().and_then(|charmap| charmap.map(c)) else {
@@ -194,6 +192,11 @@ fn has_outline_for(font: &QueryFont, c: char) -> bool {
         return false;
     };
     font_ref.outline_glyphs().get(glyph_id).is_some()
+        || font_ref.color_glyphs().get(glyph_id).is_some()
+        || font_ref
+            .bitmap_strikes()
+            .glyph_for_size(skrifa::instance::Size::unscaled(), glyph_id)
+            .is_some()
 }
 
 #[cfg(test)]
@@ -236,6 +239,26 @@ mod tests {
             println!("{script} {c:?}: {}", insert.name);
             assert!(insert.name.starts_with("system:"));
         }
+    }
+
+    #[test]
+    #[ignore = "Depends on which fonts are installed on this machine"]
+    fn renders_color_emoji_through_egui_context() {
+        let ctx = egui::Context::default();
+        ctx.add_font_provider(Arc::new(SystemFontProvider::new()));
+        let mut output = ctx.run_ui(Default::default(), |ui| {
+            let galley = ui.fonts_mut(|fonts| {
+                fonts.layout_no_wrap(
+                    "🦀".to_owned(),
+                    egui::FontId::proportional(32.0),
+                    egui::Color32::WHITE,
+                )
+            });
+            let glyph = &galley.rows[0].row.glyphs[0];
+            assert!(glyph.is_color, "Should come from a system color emoji font");
+            assert!(!glyph.uv_rect.is_nothing());
+        });
+        output.textures_delta.clear();
     }
 
     #[test]
--- a/crates/egui_system_fonts/src/lib.rs
+++ b/crates/egui_system_fonts/src/lib.rs
@@ -243,26 +243,6 @@ mod tests {
 
     #[test]
     #[ignore = "Depends on which fonts are installed on this machine"]
-    fn renders_color_emoji_through_egui_context() {
-        let ctx = egui::Context::default();
-        ctx.add_font_provider(Arc::new(SystemFontProvider::new()));
-        let mut output = ctx.run_ui(Default::default(), |ui| {
-            let galley = ui.fonts_mut(|fonts| {
-                fonts.layout_no_wrap(
-                    "🦀".to_owned(),
-                    egui::FontId::proportional(32.0),
-                    egui::Color32::WHITE,
-                )
-            });
-            let glyph = &galley.rows[0].row.glyphs[0];
-            assert!(glyph.is_color, "Should come from a system color emoji font");
-            assert!(!glyph.uv_rect.is_nothing());
-        });
-        output.textures_delta.clear();
-    }
-
-    #[test]
-    #[ignore = "Depends on which fonts are installed on this machine"]
     fn renders_cjk_through_egui_context() {
         let ctx = egui::Context::default();
         let font_id = egui::FontId::proportional(14.0);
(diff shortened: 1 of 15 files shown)


crates/egui/src/context.rs
```diff
@@ -788,6 +788,24 @@ impl Context {
...
 
+    /// Decide where to look first for the glyphs of each grapheme cluster:
+    /// the fonts in the [`FontDefinitions`], or what the platform offers
+    /// (system fonts from a [`FontProvider`], or the browser on web).
+    ///
+    /// By default, color emoji come from the platform, and everything else from the fonts.
+    /// Use `|_| GlyphSource::Fonts` to only use the platform for characters no installed font has.
+    ///
+    /// See [`GlyphSourcePreference`].
+    pub fn set_glyph_source_preference(
```
emilk: revert churn

crates/epaint/src/text/font.rs
```diff
@@ -849,32 +999,57 @@ impl Font<'_> {
...
+                    self.fonts_by_id,
+                )
+                .or_else(|| self.provided_face_for(cluster, c))
+                .or_else(|| {
+                    CachedFamily::find_face_for_char_in(
+                        &self.cached_family.fonts[..num_definition_fonts],
+                        c,
+                        self.fonts_by_id,
+                    )
+                })
```
emilk: needs some comments

```suggestion
                // First search the fallback fonts we've already loaded…
                CachedFamily::find_face_for_char_in(
                    &self.cached_family.fonts[num_definition_fonts..],
                    c,
                    self.fonts_by_id,
                )
                // …then XXX…
                .or_else(|| self.provided_face_for(cluster, c))
                // …finally YYY
                .or_else(|| {
                    CachedFamily::find_face_for_char_in(
                        &self.cached_family.fonts[..num_definition_fonts],
                        c,
                        self.fonts_by_id,
                    )
                })
```

crates/epaint/src/text/fonts.rs
```diff
@@ -972,7 +980,16 @@ impl CachedFamily {
...
     ) -> Option<FontFaceKey> {
-        for font_key in &self.fonts {
+        Self::find_face_for_char_in(&self.fonts, c, fonts_by_id)
+    }
+
+    /// The first of `fonts` whose charmap supports `c`.
+    pub(crate) fn find_face_for_char_in(
+        fonts: &[FontFaceKey],
+        c: char,
+        fonts_by_id: &mut nohash_hasher::IntMap<FontFaceKey, FontFace>,
```
emilk: `use nohash_hasher::IntMap;`

crates/epaint/src/text/fonts.rs
```diff
@@ -1398,16 +1425,6 @@ impl FontsImpl {
...
-    /// Decide where to look first for the glyphs of each grapheme cluster.
-    ///
-    /// See [`GlyphSourcePreference`].
-    pub fn set_glyph_source_preference(
-        &mut self,
-        prefer: impl Fn(&str) -> GlyphSource + Send + Sync + 'static,
-    ) {
-        self.glyph_source_preference = Arc::new(prefer);
-    }
-
```
emilk, line 446: more churn here and above

crates/epaint/src/text/text_layout.rs
```diff
@@ -184,7 +183,7 @@ impl ShapingContext {
         physical_x: i32,
         advance_width_px: f32,
         face_metrics: &StyledMetrics,
-        uv_rect: UvRect,
+        alloc: GlyphAllocation,
     ) -> Glyph {
         Glyph {
```
emilk, line 189: `let GlyphAllocation  { uv_rect, is_color } = alloc;`

crates/epaint/src/text/text_layout.rs
```diff
@@ -195,8 +194,8 @@ impl ShapingContext {
             font_face_ascent: face_metrics.ascent,
             font_height: self.font_metrics.row_height,
             font_ascent: self.font_metrics.ascent,
-            uv_rect,
-            is_color: false,
+            uv_rect: alloc.uv_rect,
+            is_color: alloc.is_color,
```
emilk: or alternatively: replace uv_rect/is_color pair with a single `GlyphAlloc` (inside of `Glyph`)

crates/epaint/Cargo.toml
```diff
@@ -35,6 +35,11 @@ cint = ["ecolor/cint"]
 ## Enable the [`hex_color`] macro.
 color-hex = ["ecolor/color-hex"]
 
+## Render bitmap (`sbix`, `CBDT`) and `COLR` glyphs, i.e. color emoji fonts like Apple Color Emoji and Segoe UI Emoji.
```
emilk:
```suggestion
## Support bitmap (`sbix`, `CBDT`) and `COLR` glyphs, i.e. color emoji fonts like Apple Color Emoji and Segoe UI Emoji.
```

crates/epaint/src/text/text_layout.rs
emilk: I went with the destructuring in the other thread and kept the `uv_rect` / `is_color` pair on `Glyph`. `Glyph` is public, and both fields are read by the tessellator and by tests in egui and egui_kittest, so folding them into one `GlyphAllocation` field is an API change that I would rather do as its own PR. Leaving this thread open in case you want it in this PR anyway.

Claude

crates/epaint/src/text/font_face.rs
```diff
@@ -56,6 +56,14 @@ pub(super) enum GlyphIdResolution {
 /// The parsed `skrifa` views into a font file, borrowing from the file's [`Blob`].
 struct DependentFontData<'a> {
     skrifa: skrifa::FontRef<'a>,
+
+    #[cfg(feature = "color_fonts")]
+    index: u32,
```
emilk, line 65: docstring: index into what??

crates/epaint/src/text/font_face.rs
```diff
@@ -152,6 +173,107 @@ impl FontCell {
...
+        let font_size_px = metrics.scale;
+        let size = skrifa::instance::Size::new(font_size_px);
+
+        let has_color_glyph = font_data.skrifa.color_glyphs().get(glyph_id).is_some()
+            || font_data
+                .skrifa
+                .bitmap_strikes()
+                .glyph_for_size(size, glyph_id)
+                .is_some();
+        if !(has_color_glyph && 0.0 < font_size_px && font_size_px.is_finite()) {
```
emilk:
```suggestion
        if !has_color_glyph || font_size_px == 0.0 || !font_size_px.is_finite() {
```

crates/epaint/src/text/font_face.rs
```diff
@@ -152,6 +173,107 @@ impl FontCell {
...
+
+    /// Rasterize a bitmap (`sbix`, `CBDT`) or `COLR` glyph, e.g. a color emoji.
+    ///
+    /// Returns `None` if the glyph has neither.
+    #[cfg(feature = "color_fonts")]
+    fn rasterize_color_glyph(
+        &self,
+        metrics: &StyledMetrics,
+        glyph_id: GlyphId,
+        bin: SubpixelBin,
```
emilk: we should turn off subpixel binning for all colored glyphs (like we do for cjk) - that's just a waste. Same on web backend, if that's not already the case

crates/epaint/src/text/font_face.rs
```diff
@@ -152,6 +173,107 @@ impl FontCell {
...
+        // We don't know the bounds of the glyph up front, so we render it onto a
+        // canvas with room for one em on each side of the em box, then crop.
+        let canvas_size = (3.0 * font_size_px)
+            .ceil()
+            .clamp(1.0, crate::text::MAX_GLYPH_SIZE as f32) as u16;
+        let origin_x = font_size_px as f64 + bin.as_float() as f64;
+        let origin_y = 2.0 * font_size_px as f64;
+
+        let font = peniko::FontData::new(
+            peniko::Blob::new(std::sync::Arc::clone(self.borrow_owner())),
```
emilk:
```suggestion
            peniko::Blob::new(Arc::clone(self.borrow_owner())),
```

crates/epaint/src/text/font_face.rs
```diff
@@ -152,6 +173,107 @@ impl FontCell {
...
+            return None; // Nothing painted
+        }
+        let width = max[0] - min[0];
+        let height = max[1] - min[1];
+
+        let cropped: Vec<Color32> = (min[1]..max[1])
+            .flat_map(|y| (min[0]..max[0]).map(move |x| (x, y)))
+            .map(|(x, y)| {
+                let i = 4 * (y * canvas_size + x);
+                Color32::from_rgba_premultiplied(
```
emilk, line 274: please double-check that premultiplied is correct here

emilk: Confirmed: `vello_common::Pixmap` documents its buffer as "premultiplied RGBA8", and `data_as_u8_slice` returns that buffer as-is, so `Color32::from_rgba_premultiplied` is correct. Added a comment saying so.

-- Claude

crates/epaint/src/text/font_face.rs
emilk: Done: `allocate_outline` now skips sub-pixel binning when `FontFace::is_color_glyph` is true (per-glyph `COLR`/bitmap-strike lookup, only for faces that have a color table). The web backend already gets `subpixel_offset_px: 0.0` from `allocate_raster`, so no binning there either.

-- Claude

crates/egui_system_fonts/src/lib.rs
```diff
@@ -238,6 +241,26 @@ mod tests {
         }
     }
 
+    #[test]
+    #[ignore = "Depends on which fonts are installed on this machine"]
```
emilk: remove test

emilk (commented):
LGTM

PR #8493, opened 2026-09-03
Make the bundled monochrome emoji fonts opt-in

emilk:
* Closes <https://github.com/emilk/egui/issues/8328>

Stacked on #8492.

The bundled `NotoEmoji-Regular` and `emoji-icon-font` are 725 kB of the binary, and now that the platform provides emoji (the browser on web via #8490, the system emoji fonts on native via #8491 and #8492) they are only needed if you want the same monochrome emoji everywhere. So they are opt-in from now on, behind a new `monochrome_emoji_fonts` feature in `epaint_default_fonts`, `epaint`, and `egui`.

With the bundled fonts no longer claiming the emoji, the per-cluster source preference has nothing left to decide: a grapheme cluster goes to the platform exactly when no installed font has it. So the second commit removes `GlyphSource`, `GlyphSourcePreference`, `default_glyph_source`, `Context::set_glyph_source_preference`, `TextRun::source`, and the per-cluster raster run in the layout code. `has_emoji_presentation` stays, since eframe's web glyph rasterizer uses it to guess whether a browser-drawn glyph is color. The `NOTDEF` path already rasterizes whole grapheme clusters, so ZWJ sequences and flags still work.

The demo, `egui_kittest`, and `egui_tests` enable the feature in their dev-dependencies, so the snapshots keep their emoji.

Tested on macOS: 🦀 🚀 📷 all come from Apple Color Emoji in color. Note that text-presentation symbols like ⏮ (no U+FE0F) now come from the platform too, and therefore render as color emoji there.

Implemented by Claude.

PR chain, in order:
* #8490
* #8500
* #8501
* #8502
* #8503
* #8504
* #8491
* #8492
* #8493
* #8508
* #8509
* #8510
* #8511

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1639,6 +1639,7 @@ dependencies = [
  "skrifa",
  "smallvec",
  "unicode-general-category",
+ "unicode-properties",
  "unicode-segmentation",
  "vello_cpu",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1639,7 +1639,6 @@ dependencies = [
  "skrifa",
  "smallvec",
  "unicode-general-category",
- "unicode-properties",
  "unicode-segmentation",
  "vello_cpu",
 ]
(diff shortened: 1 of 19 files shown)


emilk (commented):
LGTM

PR #8501, opened 2026-09-05
Move glyph bitmap caches out of `FontFace` into a new `GlyphAtlas`

emilk:
`GlyphAtlas` owns the `TextureAtlas` plus every cache that points into it (outline glyphs keyed by face, and platform-rasterized glyphs). `FontFace` no longer knows about the atlas: it just renders an outline to a bitmap.

When the atlas gets full, `Fonts::begin_pass` now clears the bitmaps instead of reparsing and re-hinting every font face. That also removes the duplicated `glyph_rasterizer` field on `Fonts`, which only existed to survive the rebuild. Fonts found by font providers (#8491) will survive an atlas clear the same way.

PR chain, in order:
* #8490
* #8500
* #8501
* #8502
* #8503
* #8504
* #8491
* #8492
* #8493
* #8508

### DIFF
--- a/crates/epaint/src/text/glyph_atlas.rs
+++ b/crates/epaint/src/text/glyph_atlas.rs
@@ -1,4 +1,5 @@
 use emath::{NumExt as _, Vec2, vec2};
+use nohash_hasher::IntMap;
 use skrifa::GlyphId;
 
 use crate::{
@@ -42,6 +43,19 @@ impl UvRect {
 pub struct GlyphAllocation {
     /// UV rectangle for drawing.
     pub uv_rect: UvRect,
+}
+
+/// An outline glyph in the atlas, positioned for one [`ShapedGlyph`].
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub(crate) struct OutlineGlyph {
+    pub allocation: GlyphAllocation,
+
+    /// Left edge of the glyph's pixel grid, in physical pixels.
+    ///
+    /// This is [`ShapedGlyph::h_pos`] rounded to the pixel (or sub-pixel bin)
+    /// the bitmap was rendered for. Draw the bitmap here, not at the
+    /// un-rounded `h_pos`, or it will be blurry.
+    pub x_px: i32,
 }
 
 /// A glyph from the [`GlyphRasterizer`], allocated in the atlas.
@@ -192,12 +206,12 @@ pub(crate) struct GlyphAtlas {
     atlas: TextureAtlas,
 
     /// Glyphs rendered from font outlines.
-    outline_glyphs: nohash_hasher::IntMap<OutlineGlyphKey, GlyphAllocation>,
+    outline_glyphs: IntMap<OutlineGlyphKey, GlyphAllocation>,
 
     /// Glyphs from the [`GlyphRasterizer`].
     ///
     /// `None` means the rasterizer could not handle the cluster.
-    raster_glyphs: nohash_hasher::IntMap<RasterGlyphKey, Option<RasterGlyphAllocation>>,
+    raster_glyphs: IntMap<RasterGlyphKey, Option<RasterGlyphAllocation>>,
 }
 
 impl GlyphAtlas {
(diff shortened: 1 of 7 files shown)


crates/epaint/src/text/text_layout.rs
```diff
@@ -934,8 +937,9 @@ fn replace_last_glyph_with_overflow_character(
             let (replacement_glyph_alloc, physical_x) = font_face
                 .as_mut()
                 .map(|f| {
-                    f.allocate_glyph(
-                        font.atlas,
+                    font.glyphs.allocate_outline(
+                        font_id,
+                        f,
```
emilk: Can we rename `f` to `face`?

crates/epaint/src/text/text_layout.rs
```diff
@@ -347,8 +348,9 @@ fn layout_shaped_run(
                     glyph_info.advance_width_unscaled.0 * fallback_metrics.px_scale_factor;
                 let (glyph_alloc, physical_x) =
                     if let Some(ff) = font.fonts_by_id.get_mut(&fallback_key) {
-                        ff.allocate_glyph(
-                            font.atlas,
+                        font.glyphs.allocate_outline(
+                            fallback_key,
+                            ff,
```
emilk: `ff` -> `face`, here and elsewhere

crates/epaint/src/text/glyph_atlas.rs
```diff
@@ -0,0 +1,369 @@
...
+                is_color: glyph.is_color,
+            })
+        });
+        self.raster_glyphs.insert(key, allocation);
+        allocation
+    }
+
+    /// Copy a bitmap into the atlas.
+    ///
+    /// Returns `None` for empty and oversized bitmaps.
```
emilk, line 370: document what happens if the atlas is full!

crates/epaint/src/text/glyph_atlas.rs
```diff
@@ -0,0 +1,372 @@
...
+
+/// The font atlas, plus every cache that points into it.
+///
+/// Cleared as a unit when the atlas is getting full.
+/// Knows nothing about families or fallback chains.
+pub(crate) struct GlyphAtlas {
+    atlas: TextureAtlas,
+
+    /// Glyphs rendered from font outlines.
+    outline_glyphs: nohash_hasher::IntMap<OutlineGlyphKey, GlyphAllocation>,
```
emilk: Add `use nohash_hasher::IntMap;` at the top of this file, and other files using `IntMap`

crates/epaint/src/text/glyph_atlas.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    /// Forget what the [`GlyphRasterizer`] produced, e.g. because it was replaced.
+    pub fn clear_raster_glyphs(&mut self) {
+        self.raster_glyphs.clear();
+    }
+
+    /// Get or render the glyph of `face` for `shaped`.
+    ///
+    /// Returns the allocation and the rounded horizontal position (in physical pixels).
+    pub fn allocate_outline(
```
emilk, line 268: what happens if the atlas is full?

crates/epaint/src/text/glyph_atlas.rs
```diff
@@ -0,0 +1,372 @@
...
+        &mut self,
+        face_key: FontFaceKey,
+        face: &mut FontFace,
+        metrics: &StyledMetrics,
+        shaped: &ShapedGlyph,
+    ) -> (GlyphAllocation, i32) {
```
emilk: Lets change this tuple to a named type, so we can docstring the `i32`

Why doe we only return the rounded horizontal position? And why don't we return it un-rounded? so many questions!

crates/epaint/src/text/glyph_atlas.rs
```diff
@@ -0,0 +1,372 @@
...
+        let ShapedGlyph {
+            glyph_id,
+            h_pos,
```
emilk, line 277: maybe `h_pos` should be replaced by `SubpixelBin` instead, and we don't need to return an `i32`?

crates/epaint/src/text/glyph_atlas.rs
emilk: Done: it now returns `OutlineGlyph { allocation, x_px }` with a docstring on `x_px`.

Why rounded: the bitmap is rasterized at a sub-pixel bin relative to an integer pixel, so the caller must draw it at exactly that integer pixel or the glyph blurs. The un-rounded position is what the caller passed in as `h_pos`, so returning it again would be redundant.

Claude

crates/epaint/src/text/glyph_atlas.rs
emilk: I kept `h_pos`. Whether to bin at all depends on the face (`FontTweak::subpixel_binning`) and on the glyph (CJK glyphs skip binning to save atlas space), so the bin is decided inside `allocate_outline`. If the caller passed a `SubpixelBin`, that decision would have to be duplicated at every call site.

Claude

emilk (commented):
LGTM

PR #8502, opened 2026-09-05
Add `FaceStore`: one place that parses and stores font faces

emilk:
`FaceStore` wraps the old `fonts_by_id` and `fonts_by_name` maps and owns `FontFaceKey`. It has a single `install(name, &FontData)` so fonts from `FontDefinitions` and fonts found later by a font provider (#8491) go through the same code.

* [x] I have followed the instructions in the PR template

PR chain, in order:
* #8490
* #8500
* #8501
* #8502
* #8503
* #8504
* #8491
* #8492
* #8493
* #8508

PR #8503, opened 2026-09-05
Replace `CachedFamily` with `Family`, which owns the fallback walk

emilk:
`Family::resolve` is now the only place that decides which face renders a char: it walks the chain, consults the per-char cache, and falls back to the replacement face. `Font` no longer reaches into the family's fields. Asking font providers (#8491) for a missing char becomes one more branch in `Family::resolve`.

PR chain, in order:
* #8490
* #8500
* #8501
* #8502
* #8503
* #8504
* #8491
* #8492
* #8493
* #8508

### DIFF
--- a/crates/epaint/src/text/font.rs
+++ b/crates/epaint/src/text/font.rs
@@ -13,8 +13,8 @@ use crate::{
     text::{
         FontTweak, GlyphSourcePreference, VariationCoords,
         face_store::{FaceStore, FontFaceKey},
+        family::Family,
         font_data::Blob,
-        fonts::CachedFamily,
         glyph_atlas::{GlyphAtlas, GlyphBitmap, RasterGlyphAllocation, SubpixelBin},
         styled_metrics::{LocationHash, StyledMetrics},
         unicode::invisible_char,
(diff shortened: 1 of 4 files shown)


crates/epaint/src/text/font.rs
```diff
@@ -302,8 +302,13 @@ impl FontFace {
         )
     }
 
+    #[inline]
+    pub(crate) fn name(&self) -> &str {
+        &self.name
```
emilk, line 308: make pub, and give it a docstring

crates/epaint/src/text/font.rs
```diff
@@ -302,8 +302,13 @@ impl FontFace {
...
+    }
+
     /// An un-ordered iterator over all supported characters.
-    fn characters(&self) -> impl Iterator<Item = char> + '_ {
+    pub(crate) fn characters(&self) -> impl Iterator<Item = char> + '_ {
```
emilk:
```suggestion
    pub fn characters(&self) -> impl Iterator<Item = char> + '_ {
```

emilk (commented):
LGTM

PR #8504, opened 2026-09-05
Remove the `Font` borrow-bag; put its methods on `FontsImpl`

emilk:
`Font<'a>` was a hand-rolled split borrow of every field of `FontsImpl`, and it grew a new `&'a mut` for each new feature. `text_layout` now calls `FontsImpl` directly (`resolve_face`, `glyph_info`, `allocate_glyph`, `rasterize_cluster`, …) with a cheap `FamilyKey` index instead of re-hashing the `FontFamily` per glyph.

`font.rs` is renamed to `font_face.rs`, since only `FontFace` is left in it. `FontsView::characters` replaces `f.fonts.font(family).characters()` in the font book demo.

`FontsImpl` and `epaint::text::layout` are now crate-private: nothing outside `epaint` used them, and `Fonts` / `FontsView` cover the public API. The `fonts` fields of `Fonts` and `FontsView` are private too. The benchmark that used `layout` directly now calls the new `Fonts::layout_uncached`.

PR chain, in order:
* #8490
* #8500
* #8501
* #8502
* #8503
* #8504
* #8491
* #8492
* #8493
* #8508

### DIFF
--- a/crates/epaint/src/text/fonts.rs
+++ b/crates/epaint/src/text/fonts.rs
@@ -374,6 +374,7 @@ impl FontsImpl {
 
     /// Use this platform glyph rasterizer, e.g. the browser on web.
     #[cfg(test)]
+    #[inline]
     pub fn with_glyph_rasterizer(mut self, glyph_rasterizer: GlyphRasterizer) -> Self {
         self.set_glyph_rasterizer(Some(glyph_rasterizer));
         self
--- a/crates/epaint/src/text/fonts.rs
+++ b/crates/epaint/src/text/fonts.rs
@@ -357,7 +357,7 @@ pub(crate) struct FontsImpl {
 impl FontsImpl {
     /// Create a new [`FontsImpl`] for text layout.
     /// This call is expensive, so only create one [`FontsImpl`] and then reuse it.
-    pub(crate) fn new(options: TextOptions, definitions: FontDefinitions) -> Self {
+    pub fn new(options: TextOptions, definitions: FontDefinitions) -> Self {
         let faces = FaceStore::new(options, &definitions);
 
         Self {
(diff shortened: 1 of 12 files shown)


crates/epaint/src/text/fonts.rs
```diff
@@ -334,25 +357,24 @@ pub struct FontsImpl {
 impl FontsImpl {
     /// Create a new [`FontsImpl`] for text layout.
     /// This call is expensive, so only create one [`FontsImpl`] and then reuse it.
-    pub fn new(options: TextOptions, definitions: FontDefinitions) -> Self {
+    pub(crate) fn new(options: TextOptions, definitions: FontDefinitions) -> Self {
```
emilk: if the type is `pub(crate)`, then the methods on it don't have to be

emilk: here and everywhere: only mark the type with `pub(crate)`, not every method on it!

crates/epaint/src/text/fonts.rs
```diff
@@ -362,48 +384,201 @@ impl FontsImpl {
...
+    /// fallback-face advances, resolve the face first with [`Self::resolve_face`]
+    /// and build metrics for that face.
+    pub(crate) fn glyph_info(
+        &mut self,
+        family: FamilyKey,
+        c: char,
+        metrics: &StyledMetrics,
+    ) -> (FontFaceKey, GlyphInfo) {
+        let face_key = self.resolve_face(family, c);
+        let replacement_char = self.family(family).replacement_char();
```
emilk: can we avoid calling this on the happy-path?

emilk (commented):
LGTM

PR #8505, opened 2026-09-05
Deprecate `global_theme_preference_switch` in favor of `global_theme_preference_buttons`

emilk:
`global_theme_preference_switch` can only toggle between dark and light mode, with no way back to following the system theme.

`global_theme_preference_buttons` covers all three, and now uses compact icon-only buttons (💻 🌙 ☀), so it fits in a top bar too. The demo app now uses it.

* Rename `ThemePreference::radio_buttons` to `ThemePreference::buttons` (old name deprecated)
* Add `ThemePreference::icon` and `ThemePreference::name`


umajho:
Is there a reason ☀ is used instead of ☀️? The other two icons are colorful, while only the icon for light mode is monochrome, which feels inconsistent.

Here is a screenshot in Chrome (it looks the same in Safari and Firefox; I have not checked the native version.)
image

The three emojis in the PR description also shows the same inconsistency:
image

I'm still on macOS 15, so perhaps apple changed how these characters are rendered in a later version of macOS, and the monochrome version is rendered as colorful on your end…?

Edit: It turns out this has always been the case in the code. it was just that emojis were previously always rendered using monochrome fonts, so whether the code used the colorful or monochrome variant didn't matter.

emilk:
Good catch! This is just because colored emoji in egui is _extremely new_. I'll take a pass over this and all other things emoji after https://github.com/emilk/egui/pull/8492 is merged

PR #8507, opened 2026-09-05
CI: run once per PR, and stop PR caches from evicting the `main` cache

emilk:
The `Run tests` job takes 9-12 minutes because it finds no cargo cache and rebuilds all 390 crates; the last run that hit the cache took 4m20s for the whole job. The repository holds 30 caches totalling 10.6 GB, over the 10 GB limit, so entries are evicted constantly. Each PR saves its own ~2.5 GB copy of every job cache, which pushes out the `main` cache that every new PR starts from.

* `on: [push, pull_request]` ran the whole workflow twice for a PR from a branch in this repo, doubling both runner time and cache writes. The push half now only runs for `main`.
* Only `main` writes to the cargo cache (`save-if`). PRs still read it.
* New workflow deletes the caches of a PR when it is closed.

### DIFF
--- a/.github/workflows/cache_cleanup.yml
+++ b/.github/workflows/cache_cleanup.yml
@@ -1,0 +1,33 @@
+# Delete the GitHub Actions caches of a pull request once it is closed.
+#
+# A repository can only hold 10 GB of caches, and the oldest ones are evicted when it is full.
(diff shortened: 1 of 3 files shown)


lucasmerlin (approved):
Nice!

PR #8508, opened 2026-09-05
Build font families from `FontProvider`s only: `FontDefinitions` becomes the `ConfiguredFonts` provider

emilk:
Experiment on top of #8493: one mechanism for where fonts come from.

* `FontProvider` gains `fonts_for_family(&FontFamily) -> Vec<FontInsert>`, with `font_for` (on-demand discovery) now optional too. A provider can do either or both.
* New `ConfiguredFonts` wraps `FontDefinitions` and is always the first provider. `Family::new` no longer reads the definitions; it installs whatever the providers return, in provider order, and on-demand discovery appends to the same chain.
* Faces survive a `TextOptions` change: hinting and sub-pixel binning are reconfigured in place (`FontFace::set_options`), and only the atlas starts over. That removes the carry-over machinery (`install_provided`, `provided_keys_for`, the `misses` set) since nothing is rebuilt anymore. Per-char misses are already remembered by `Family::face_cache`.
* An unknown `FontFamily` now logs an error and yields no glyphs instead of panicking, so a later provider can serve it.

Line count is about even (+211/-163, mostly docs). The simplification is structural: one resolution path, no special cases for definitions vs. providers, no state to carry across rebuilds. The natural next step, not in this PR, is letting a provider supply the family's primary face so that an app can run with zero configured fonts on native. Web still needs configured fonts, since the browser rasterizer returns bitmaps, not font files.

PR chain, in order:
* #8490
* #8500
* #8501
* #8502
* #8503
* #8504
* #8491
* #8492
* #8493
* #8508
* #8509
* #8510
* #8511


crates/epaint/src/text/configured_fonts.rs
```diff
@@ -0,0 +1,48 @@
...
+///
+/// Wraps [`FontDefinitions`]: for each family it hands out the fonts listed there,
+/// in priority order. It never discovers anything on demand,
+/// so the same text looks the same on every machine.
+///
+/// This is always the first provider, so configured fonts win over discovered ones.
+#[derive(Clone, Debug)]
+pub struct ConfiguredFonts {
+    definitions: FontDefinitions,
+}
```
emilk: do we really want/need this wrapper?

crates/epaint/src/text/font_provider.rs
```diff
@@ -49,28 +66,22 @@ where
...
 #[derive(Default)]
 pub(crate) struct FontProviders {
-    /// Asked in order; the first font found wins.
     providers: Vec<Arc<dyn FontProvider>>,
 
     /// Fonts discovered so far, in the order they were found.
     discovered: Vec<FontInsert>,
-
-    /// No provider had a font for these.
-    misses: ahash::HashSet<(FontFamily, char)>,
```
emilk, line 65: isn't it good to keep misses, so we don't re-lookup the same characters each frame?

emilk: Misses are still cached, just in one place instead of two.

`Family::resolve_slow` writes the result of every lookup into `face_cache`, including a miss: when no face has the char and no provider offers one, it caches `replacement_face_key` for that char. `resolve_cluster` checks `face_cache` first, so a missing char costs one hash lookup per frame and the providers are asked at most once per (family, char), exactly as the old `misses` set guaranteed.

The separate set was redundant once the fallback chain moved into `Family`: it was a second cache keyed by (family, char) in front of a per-family cache keyed by char.

`discovered_fonts_and_misses_survive_an_options_change` covers it: it asks for `HANGUL` (which every provider misses) before and after a `TextOptions` change and asserts the provider saw only 2 requests total.

— Claude

PR #8509, opened 2026-09-05
Add emoji presentation selectors to the emoji used in the UI

emilk:
Emoji like ☀, ⚙, 🖱, 🖼, and ⬇ have *text* presentation by default, so browsers and most platforms draw them monochrome, next to color emoji like 🌙 and 💻. Reported in https://github.com/emilk/egui/pull/8505#issuecomment-5553069797.

This appends U+FE0F (the emoji presentation selector) wherever such an emoji is used as an icon in egui, egui_extras, and the demos. The selector is zero-width and has no advance, so builds with the bundled monochrome emoji fonts render exactly as before (snapshots unchanged).

Not touched: emoji in doc comments, and the character list in `egui/src/lib.rs`.

PR chain, in order:
* #8490
* #8500
* #8501
* #8502
* #8503
* #8504
* #8491
* #8492
* #8493
* #8508
* #8509
* #8510
* #8511

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -3563,13 +3563,13 @@ impl Context {
                 paint_stats.ui(ui);
             });
 
-        CollapsingHeader::new("🖼 Textures")
+        CollapsingHeader::new("🖼️ Textures")
             .default_open(false)
             .show(ui, |ui| {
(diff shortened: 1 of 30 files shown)


emilk (commented):
LGTM

PR #8510, opened 2026-09-05
Log at debug level when no font has a char

emilk:
Logs once per character (the result is cached in `Family::face_cache`, and the lookup is `#[cold]`):

```
No font for '🖹' (U+1F5B9) in Proportional: rendering '?' instead
```

Makes it easy to see which characters your font set is missing, e.g. after turning off the bundled emoji fonts.

PR chain, in order:
* #8490
* #8500
* #8501
* #8502
* #8503
* #8504
* #8491
* #8492
* #8493
* #8508
* #8509
* #8510
* #8511


emilk (commented):
LGTM

PR #8511, opened 2026-09-05
Render the font's `.notdef` glyph (tofu) for characters no font has

emilk:
Characters that no font in the family has were rendered as the *character* ◻ (U+25FB), or as `?` when no font in the chain had ◻. Only the bundled emoji fonts have ◻, so after #8493 made those opt-in, every missing character showed up as `?`.

Now we render glyph 0, the `.notdef` glyph, of the family's primary face. Every font has one (both Hack and Ubuntu-Light draw a box), so the tofu is always there and always matches the text font.

* `Family` loses `replacement_face_key` and `replacement_char`
* `GlyphAtlas::allocate_outline` renders `NOTDEF` instead of skipping it; invisible glyphs are skipped by the caller instead
* The CJK layout tests no longer depend on which fonts are bundled, so they run in both feature configurations. Their expected line breaks changed, since the tofu is narrower than ◻ was.

PR chain, in order:
* #8490
* #8500
* #8501
* #8502
* #8503
* #8504
* #8491
* #8492
* #8493
* #8508
* #8509
* #8510
* #8511

### DIFF
--- a/crates/epaint/src/text/family.rs
+++ b/crates/epaint/src/text/family.rs
@@ -16,7 +16,7 @@ pub(crate) struct FamilyKey(pub(crate) usize);
 ///
 /// This is the only place that decides which face renders a given char:
 /// first the fonts the [`FontProviders`] install up front (the configured fonts),
-/// then fonts they discover on demand, then the replacement glyph.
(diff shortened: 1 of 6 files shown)


emilk (commented):
LGTM

PR #8516, opened 2026-09-06
Fix egui-winit compilation on wasm32

### LINKED ISSUE #8436: egui-winit 0.36.1 fails to compile for wasm32 due to DroppedFile trait mismatch
## Description

egui-winit 0.36.1 does not compile for wasm32-unknown-unknown when used directly.

On WASM, egui::DroppedFile requires bytes_async(), but egui-winit’s NativeFile implementation unconditionally provides the native-only bytes() method.

## Reproduction

```
[package]
name = "egui-winit-wasm-repro"
version = "0.1.0"
edition = "2024"

[dependencies]
egui-winit = { version = "0.36.1", default-features = false }
```

```
rustup target add wasm32-unknown-unknown
cargo check --target wasm32-unknown-unknown
```

## Error

error[E0407]: method `bytes` is not a member of trait `egui::DroppedFile`
  --> egui-winit-0.36.1/src/dropped_file.rs:19:5

error[E0046]: not all trait items implemented, missing: `bytes_async`
  --> egui-winit-0.36.1/src/dropped_file.rs:14:1

leotimmins1974:
Disclosure - AI written PR with GPT 6
I was unable to compile for wasm on my project because I use egui-winnit instead of eframe

See AI explanation below:

## Summary

* Fixes #8436

`egui-winit` unconditionally compiled `NativeFile`, which implements `DroppedFile::bytes()`. On wasm32, the trait requires `bytes_async()` instead, causing compilation to fail.

This change gates the native file implementation and its usage to native targets. An explicit wasm fallback ignores dropped-file events, which winit’s web backend does not emit.

Adds a direct wasm compilation check to CI and the local check script, since eframe’s wasm builds do not compile egui-winit.

## Validation

- wasm32 compilation with default features disabled passes.
- Native Linux/X11 compilation passes.
- Formatting and shell syntax checks pass.
- Strict wasm Clippy encounters six existing, unrelated warnings in clipboard.rs.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -91,6 +91,10 @@ jobs:
       - name: clippy wasm32 eframe
         run: cargo clippy -p eframe --lib --no-default-features --features wgpu,persistence --target wasm32-unknown-unknown
 
+      # eframe only depends on egui-winit on native targets, so check it directly on wasm.
+      - name: check wasm32 egui-winit
+        run: cargo check -p egui-winit --no-default-features --target wasm32-unknown-unknown --locked
+
       - name: wasm-bindgen
         uses: jetli/wasm-bindgen-action@v0.1.0
         with:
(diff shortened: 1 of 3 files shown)


emilk (approved):
thanks!

PR #8521, opened 2026-09-07
Fix compile error with `web_sys_unstable_apis`

lucasmerlin:
Web sys changes the signature of `get_image_data` based on the flag. In rerun we need `web_sys_unstable_apis` so we need to handle this in egui

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -183,8 +183,7 @@ unused_lifetimes = "warn"
 trivial_casts = "allow"
 unused_qualifications = "allow"
 
-# `web-sys` gates some APIs (and even some signatures) behind this cfg,
-# which downstream crates may set globally via `.cargo/config.toml`.
+# `web_sys_unstable_apis` changes some signatures, so we need to check for presence of that flag
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(web_sys_unstable_apis)'] }
 
 [workspace.lints.rustdoc]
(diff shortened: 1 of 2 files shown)


crates/eframe/src/web/canvas_glyphs.rs
```diff
@@ -183,12 +183,17 @@ impl CanvasGlyphs {
...
+        #[cfg(not(web_sys_unstable_apis))]
+        let image_data = self
             .context
-            .get_image_data(0.0, 0.0, width as f64, height as f64)
-            .ok()?
-            .data()
-            .0;
+            .get_image_data(0.0, 0.0, width as f64, height as f64);
+        #[cfg(web_sys_unstable_apis)]
+        let image_data = self
```
emilk: use `cfg_select!`

Cargo.toml
```diff
@@ -183,6 +183,10 @@ unused_lifetimes = "warn"
 trivial_casts = "allow"
 unused_qualifications = "allow"
 
+# `web-sys` gates some APIs (and even some signatures) behind this cfg,
+# which downstream crates may set globally via `.cargo/config.toml`.
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(web_sys_unstable_apis)'] }
```
emilk, line 188: wait, what does this do?

lucasmerlin: This whitelists checking `web_sys_unstable_apis` which is otherwise flagged by the lint. Made the comment better

emilk (approved):

PR #8525, opened 2026-09-08
Paint `SubMenuButton` arrow as a triangle instead of a glyph

lucasmerlin:
Eguis built in widgets should work right even if there are no default/system fonts. This renders the submenubutton triangle with epaint so it will always look right and isn't system-font dependent.


crates/egui/src/containers/menu.rs
```diff
@@ -340,17 +342,44 @@ pub struct SubMenuButton<'a> {
 }
 
 impl<'a> SubMenuButton<'a> {
-    /// The default right arrow symbol: `"⏵"`
-    pub const RIGHT_ARROW: &'static str = "⏵";
+    /// The submenu arrow triangle shape.
+    pub fn arrow_shape(rect: Rect, color: impl Into<Color32>) -> Shape {
+        let rect = Rect::from_center_size(
+            rect.center(),
+            vec2(rect.width() * 0.55, rect.height() * 0.35),
```
lucasmerlin, line 349: Factors chosen to match the previous font based arrows size

crates/epaint/src/shapes/shape.rs
```diff
@@ -256,6 +256,21 @@ impl Shape {
         Self::Path(PathShape::convex_polygon(points, fill, stroke))
     }
 
+    /// A filled triangle inscribed in `rect`, pointing down.
+    ///
+    /// `rotation` is in radians, and rotates the triangle around the center of `rect`,
+    /// e.g. `-TAU / 4.0` makes it point to the right.
```
emilk: Maybe spell out all four cardinal directions for convenience (0 = …, tau/4 = … )

emilk (approved):

Issue #122, opened 2021-01-18
Autocomplete / suggestions popup for TextEdit


Weasy666:
Hello! Is it possible to something like this
![grafik](https://user-images.githubusercontent.com/11967542/104949247-f12f4f80-59be-11eb-8d0e-2b08f2c1bd07.png)
with `TextEdit`?

I would like to show some matching suggestions when a user enters text into the `TextEdit`.

emilk:
There is nothing like this for `TextEdit` in particular, but I've added a helper function for you: `egui::popup::popup_below_widget`. Good luck!

Weasy666:
Oh! That is really nice of you. Thank you very much!

crumblingstatue:
I think it would be worth revisiting this. The current support for autocompletion is very crude.
For example, you can't override the behavior of TextEdit to ignore up/down key when the user is navigating your list.
You also can't set the cursor position after the autocomplete succeeds.
This basically means there is no way to make a good working autocompletable text buffer with TextEdit to my knowledge.

gabrieldechichi:
Seconding @crumblingstatue  comments, I'm trying to implement a game console lib, and feeling blocked by not being able to set the cursor position.

PR #8529, opened 2026-09-08
Add `CompletionPopup`: a code completion popup over `TextEdit`

emilk:
New widget `egui::CompletionPopup`: a code completion popup over a `TextEdit`. ↑/↓ selects, Enter or Tab accepts, Escape closes, click accepts. Keyboard focus never leaves the text field. The caller supplies suggestions for the word under the cursor:

```rust
CompletionPopup::new(id).show(ui, TextEdit::singleline(&mut text), |query| suggest(query));
```

New example `code_completion` shows slash-command completion, with headless `egui_kittest` tests for the arrow/Enter/Tab/Escape behavior, including a multiline chat composer.

Builds on:
* https://github.com/emilk/egui/pull/8530

Note on Escape: egui surrenders focus on Escape in `Memory::begin_pass`, before any widget runs. So an app that wants "Escape does X while the text field is focused" must set `TextEdit::event_filter(EventFilter { escape: true, .. })`, or the handler never sees the field focused. The popup does this for you while it is open.

screenshot

* Related: <https://github.com/emilk/egui/issues/122>


emilk (commented):
Tried this against the chat composer in `re_agent_ui` (rerun-io/reality#3548), which today has a hand-rolled inline slash-command list and just hit the "popup steals focus" bug this widget avoids by construction. Fit is good: the `make_text_edit` closure still lets us set `hint_text`, `desired_rows`, a custom frame, and `return_key(Shift+Enter)`, and a post-`show` `consume_key(Enter)` to send still works because the popup only eats Enter while open.

Two things a chat UI runs into:

* **Id uniqueness.** The example uses `Id::new("prompt")`. With several conversations in tabs each needs `ui.make_persistent_id(…)`, or they share `TextEditState` and popup state. Worth a sentence in the `new()` docs, like `TextEdit::id` has.
* **`suggest` only sees the word, not where it is.** Slash commands are only valid at the start of the prompt, so `fix /he` should not complete. The callback can't check, since `text` is mutably borrowed during `show`. Passing a small context (`text: &str`, `word_range: Range<usize>`) would cover this and things like "only after `@`".

On the implementation:

* **Selection index survives filtering.** Type `/c`, press ↓ (selects `/compact`), type `o`: the list is now `/compact`, `/config` and index 1 highlights `/config`. Completion UIs reset to the first item when the word changes. Remember the word the selection belongs to and reset when it differs. A test for this would be good.
* **Stale cursor on key-accept.** The Enter path runs before the `TextEdit` and combines last frame's cursor with this frame's text. A typed char plus Enter in the same frame (kittest, fast typing, IME commit) makes `accept` delete the wrong range. Accepting after `show` using `output.cursor_range` would fix it.
* **One-frame open delay.** Keys are only intercepted when `was_open`, so Enter on the very first frame the popup appears goes to the `TextEdit`. Fine, but worth documenting.
* **API shape.** `show(ui, &mut text, |text| TextEdit::singleline(text), suggest)` passes the buffer twice. `completion.rs` lives inside the `text_edit` module, so `TextEdit::text` could be `pub(super)` and the popup could take the built `TextEdit` by value: `CompletionPopup::new(id).show(ui, TextEdit::multiline(&mut text).hint_text(…), suggest)`.
* **`word_boundary: fn(char) -> bool`** rules out closures that capture config. `impl Fn(char) -> bool` or `Box<dyn Fn>` is the usual pattern elsewhere in egui.
* **Escape with the popup closed** still surrenders focus (default filter). Our composer uses Esc to cancel the running turn and wants to keep focus. Letting the caller supply the base `EventFilter` that the popup ORs `tab`/`escape` into would avoid hard-coding one.
* **Hover doesn't move the selection.** Minor. Setting `state.selected = i` on `response.hovered()` matches what every editor does.
* **Tests:** the four cases are good. Consider adding one with `TextEdit::multiline` + `return_key`, since that is the shape chat UIs use.

`TextEdit::event_filter` is a useful addition on its own.

— Claude

emilk (commented):
Re-read at e7205e8. Everything from the first pass is addressed: `CompletionQuery` with `is_at_start()`, per-word selection reset, keys applied after `show` with the fresh cursor, `TextEdit` taken by value via `show_returning_text`, boxed `word_boundary`, the caller's `EventFilter` respected and only OR-ed with tab/escape, hover moves the selection (only while the pointer moves, nice), the one-frame delay documented, and multiline + same-frame + start-of-prompt tests added. Hiding the popup on the accept frame is a good catch too.

Two small things left:

* `consume_unmodified_key` duplicates `InputState::count_and_consume_key(Modifiers::NONE, key)`. That one already uses `matches_logically`, which requires shift/alt to match exactly, so Shift+Enter is left alone by it as well. The helper can go, and its docstring is currently wrong about `consume_key`.
* `selection_delta += consume_unmodified_key(…) as isize` and the `as usize` back: fine in practice, but egui's clippy config may flag the casts (`cast_possible_wrap`). `isize::try_from(..).unwrap_or(0)` or keeping the delta as two `usize` counts sidesteps it. CI will tell.

For rerun's chat composer this is now a straight swap for the hand-rolled list. One consequence of `TextEdit::event_filter` worth noting in the PR description: without it, Escape surrenders focus in `Memory::begin_pass` before any widget runs, so an "Escape does X while the field is focused" handler after `show()` never sees the field focused. We hit exactly that in our composer today; the new `event_filter(EventFilter { escape: true, .. })` is the proper fix once this is released.

— Claude

emilk:
Thanks for the second pass.

* Casts: gone in 57f8a4b. The arrow presses are now two `usize` counts and the selection moves with modular arithmetic.
* `consume_unmodified_key`: kept. `matches_logically` allows *extra* Shift/Alt in the pressed modifiers when the pattern does not require them, see its own doctest: `(Modifiers::CTRL | Modifiers::SHIFT).matches_logically(Modifiers::CTRL)` is `true`. So `count_and_consume_key(Modifiers::NONE, Key::Enter)` does consume Shift+Enter. The multiline chat composer test in the example hit exactly that before the helper existed: Shift+Enter got eaten as an accept. The helper docstring stands.

— Claude

crates/egui/src/widgets/text_edit/completion.rs
```diff
@@ -0,0 +1,443 @@
...
+#[derive(Clone)]
+pub struct Suggestion {
+    /// The text that replaces the current word when this suggestion is accepted.
+    pub insert: String,
+
+    /// What to show in the popup. Defaults to [`Self::insert`].
+    pub label: WidgetText,
+
+    /// Optional extra text shown weakly to the right of the label.
+    pub description: Option<WidgetText>,
```
emilk: maybe we should just combine this into an atom-layout (with a friendly helper ctor)

crates/egui/src/widgets/text_edit/completion.rs
```diff
@@ -0,0 +1,443 @@
...
+}
+
+/// State stored between frames.
+#[derive(Clone, Default)]
+struct CompletionState {
+    /// Index of the selected suggestion.
+    selected: usize,
+
+    /// The word the selection belongs to. The selection resets when the word changes.
+    selected_word: String,
```
emilk, line 94: I dont get this

crates/egui/src/widgets/text_edit/completion.rs
```diff
@@ -0,0 +1,443 @@
...
+
+    /// Was the popup open at the end of the last frame?
+    was_open: bool,
```
emilk:
```suggestion
    open: bool,
```

examples/code_completion/src/main.rs
lucasmerlin, line 1: Why not as an egui.rs demo? It's an official widget after all

crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -476,6 +476,20 @@ impl TextEdit<'_> {
...
+        self.event_filter
+    }
+
+    /// Like [`Self::show`], but also gives back the text buffer,
+    /// so the caller can keep editing it after the text edit has been shown.
+    pub(super) fn show_returning_text(
+        self,
+        ui: &mut Ui,
+    ) -> (TextEditOutput, &'t mut dyn TextBuffer) {
         let TextEdit {
```
lucasmerlin, line 493: Why does the completeion popup need a private api? 🤔 Someone should be able to implement their own

lucasmerlin (approved):
It feels nice to use! I think if it was generic so that you can define your own `Suggestion` struct and popup contents (e.g. for tagging via @ username someone might want to use their existing `Username` component) it would be more useful, but its nice as is.

crates/egui/src/widgets/text_edit/completion.rs
```diff
@@ -0,0 +1,441 @@
...
+                    .max_height(max_height)
+                    .show(ui, |ui| {
+                        let pointer_moved = ui.input(|input| input.pointer.is_moving());
+                        let mut clicked = None;
+                        for (i, suggestion) in suggestions.iter().enumerate() {
+                            let is_selected = i == state.selected;
+                            let response = ui.add(
+                                Button::selectable(is_selected, suggestion.content.clone())
+                                    .min_size(vec2(ui.available_width(), 0.0)),
+                            );
```
lucasmerlin, line 336: Add a `egui::button::completion` class so users can style it 😁

examples/code_completion/src/main.rs
lucasmerlin: Maybe as a egui demo it would be nice to have a emoji selection demo (:emojiname: autocomlete)? I think thats a more generally useful demo (comletions not only at the start, and more apps have emojis than chat commands). And it would be a nice showcase of the new emojis!

crates/egui/src/widgets/text_edit/builder.rs
emilk: agree - lets fix that

examples/code_completion/src/main.rs
emilk: Thats a good idea for a demo!

PR #8534, opened 2026-09-09
Deprecate `Id::new` and `From<&str> for Id`

emilk:
`Id::new` and `From<&str> for Id` etc are foot-guns: they hide the fact that an `Id` MUST be globally unique (well, at least unique within your app).

So this PR removes both and replaces them with `Id::unique`, which is very explicit.
In places where non-uniqueness is desired, you can use `IdSalt`, which has the convenient `From<&str>`.

Most changes are in tests.

### DIFF
--- a/crates/egui/src/containers/panel.rs
+++ b/crates/egui/src/containers/panel.rs
@@ -245,21 +245,21 @@ pub struct Panel {
 impl Panel {
     /// Create a left panel.
     ///
-    /// The id should be globally unique, e.g. `Id::new("my_left_panel")`.
+    /// The id should be globally unique, e.g. `Id::unique("my_left_panel")`.
     pub fn left(id: impl Into<Id>) -> Self {
         Self::new(PanelSide::Left, id)
     }
 
     /// Create a right panel.
     ///
-    /// The id should be globally unique, e.g. `Id::new("my_right_panel")`.
+    /// The id should be globally unique, e.g. `Id::unique("my_right_panel")`.
     pub fn right(id: impl Into<Id>) -> Self {
         Self::new(PanelSide::Right, id)
     }
 
     /// Create a top panel.
     ///
-    /// The id should be globally unique, e.g. `Id::new("my_top_panel")`.
+    /// The id should be globally unique, e.g. `Id::unique("my_top_panel")`.
     ///
     /// By default this is NOT resizable.
     pub fn top(id: impl Into<Id>) -> Self {
@@ -268,7 +268,7 @@ impl Panel {
 
     /// Create a bottom panel.
     ///
-    /// The id should be globally unique, e.g. `Id::new("my_bottom_panel")`.
+    /// The id should be globally unique, e.g. `Id::unique("my_bottom_panel")`.
     ///
     /// By default this is NOT resizable.
     pub fn bottom(id: impl Into<Id>) -> Self {
@@ -277,7 +277,7 @@ impl Panel {
 
     /// Create a panel.
     ///
-    /// The id should be globally unique, e.g. `Id::new("my_panel")`.
+    /// The id should be globally unique, e.g. `Id::unique("my_panel")`.
     fn new(side: PanelSide, id: impl Into<Id>) -> Self {
         let default_outer_size: Option<f32> = match side {
             PanelSide::Left | PanelSide::Right => Some(200.0),
--- a/crates/egui/src/containers/panel.rs
+++ b/crates/egui/src/containers/panel.rs
@@ -18,8 +18,9 @@
 use emath::GuiRounding as _;
 
 use crate::{
-    Align, Context, CursorIcon, Frame, Id, InnerResponse, LayerId, Layout, Margin, NumExt as _,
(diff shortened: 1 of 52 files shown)


tests/egui_tests/tests/test_panel_drag.rs
```diff
@@ -24,6 +24,21 @@ struct State {
     /// deliberately doesn't persist its size while its resize handle is being
     /// dragged — which is exactly when these tests need to observe it.
     panel_width: Option<f32>,
+
+    /// [`egui::Ui::id`] of the harness root ui, recorded each pass.
+    ///
+    /// Panel ids are salted with this, so we need it to look up [`egui::PanelState`].
+    ui_id: Option<egui::Id>,
```
emilk: maybe call this `root_ui_id`

emilk: or maybe we should add `viewport_id.root_ui_id()` and use that in `run_ui_dyn`. That way we can get the default root_ui_id with `ViewportId::root().root_ui_id()`

emilk: Added `ViewportId::root_ui_id()` and used it in `run_ui_dyn`. But the tests can't use `ViewportId::ROOT.root_ui_id()` directly: kittest wraps the closure in a `scope_builder` and a `CentralPanel` frame, and every unsalted child `Ui` gets `parent.id().with("child")`. So the ui the panels are shown in has a different id than the viewport root ui. Kept recording `ui.id()` in the harness state, renamed to `parent_ui_id`.

Claude

emilk: Follow-up: added `Harness::ui_id()` to egui_kittest, which returns the `Ui::id` of the ui passed to the closure. The panel tests now use `harness.ui_id().with("…")` instead of recording the id in their state.

Claude

crates/egui/src/id.rs
```diff
@@ -70,8 +72,13 @@ impl Id {
         }
     }
 
-    /// Generate a new root [`Id`] by hashing some source (e.g. a string or integer).
-    pub fn new(source: impl AsId) -> Self {
+    /// Generate a new, globally unique, root [`Id`] by hashing some source (e.g. a string or integer).
```
lucasmerlin:
```suggestion
    /// Creates a new root [`Id`] from a globally unique source (e.g. a string or integer) by hashing it.
```
I wouldn't trust people to read past the first sentence and the current wording might make it sounds like something within `unique` ensures that the id is globally unique...

crates/egui/src/ui.rs
lucasmerlin, line 50: It's a bit of a shame that we still have these two `Id`s for `Ui` now, meaning our core struct breaks the rules of `Id`...

crates/egui/src/id_salt.rs
lucasmerlin, line 26: I feel like `LocalId` would be a better name now

crates/egui/src/containers/panel.rs
```diff
@@ -205,7 +206,9 @@ impl PanelSide {
 #[must_use = "You should call .show()"]
 pub struct Panel {
     side: PanelSide,
-    id: Id,
+
+    /// Combined with the parent [`Ui::id`] to form the [`Id`] of the panel.
+    id_salt: IdSalt,
```
lucasmerlin, line 211: It can be really useful to give a widget a specific `Id`, e.g. when it moves between differnt parent Uis but you want to keep its state, or as seen in the test. 
Maybe we could add a `MaybeUniqueId` enum that can hold either a `IdSalt` or `Id`? We already have something like this in `UiBuilder`, would be nice to have a nice pattern for this

lucasmerlin (approved):
Nice!

crates/egui/src/id_salt.rs
emilk: yeah, probably… next time!

crates/egui/src/containers/panel.rs
emilk: I agree it could be useful, but the current approach has another benefit: a strict tree hieararchy… not that we make use of it right now, but it's there 🤷

crates/egui/src/ui.rs
emilk: I think it is the naming here that is wrong. `id` above is really for a _group_ of Uis, where a new group can be created by salting a child Ui. Each _group_ has a globally unique and stable `id`. So we could rename `id` here to `stable_group_id` or something

emilk: I'll improve the naming and docstring of these next!

lucasmerlin: Maybe `scope_id` and then rename `push_id` to `id_scope`, and it would mean it's the `Id` of the nearest parent scope that had a expilcit `Id` set, or the `Window` id if there was no other scope?

emilk: "scope" is a great word for it 💡

emilk: * https://github.com/emilk/egui/pull/8542

PR #8535, opened 2026-09-09
Add `Context::root_ui` and `Context::run_pass`

### LINKED ISSUE #8415: 0.34 containers API change makes them more difficult to use in `bevy_egui`
# Context

From 0.34 release highlights:

> * Use `Ui` as the main entrypoint

# Problem

Some `bevy_egui` users (me included) stumbled upon the change where `Panel` now expects `&mut Ui` in its `show` method.

The only way `bevy_egui` users can access egui's `&mut Ui` is through containers that accept `egui::Context` in their `show` method. Due to the nature of Bevy's ECS, we have to keep contexts as entities, and the only way to access them is through a query:
```rust
fn ui_system(context: Single<&EguiContext>) {
  egui::Window::new("window").show(context.get_mut(), |ui| { 
    // ...
  });
}
```

Previously, it was possible to make `Panel` a "root" widget by passing `egui::Context`. Now, the only workaround I found is to first wrap a `Panel` into an `Area`, but that seems to break some things still (in my app, panels have some margins between their own and native window's borders, panel endlessly extends vertically, etc).

# Considered workarounds on the `bevy_egui` end

## 1. Single entry-point system (like in Eframe)

```rust
fn egui_system(world: &mut World) {
  let (ctx, input) = ... ; // a bit of boilerplate to query both egui context and input,
  // this can be theoretically wrapped into a prettier abstraction to hide the `run_ui` details
  // but this is roughly the idea still
  ctx.run_ui(input, |ui| {
    // ...
  });
}
```

Unfortunately, it would be a huge and unwelcome breaking change for the ecosystem, as a lot of `bevy_egui` users split their UI into multiple systems for separating concerns and overall code organisation, where it allows them for each system to have their own dependencies (queries, resources), etc.

## 2. Custom UI systems registration to support `&mut Ui` as an input parameter

```rust
fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_plugins(EguiPlugin::default())
        .add_egui_system(EguiPrimaryContextPass, ui_system)
        .run();
}

fn ui_system(ui: In<&mut Ui>) {
  // ...
}
```

This could be a better alternative API-wise, but it's not trivial to implement. Passing custom input parameters requires manual calls of `run_system_with`, and that bypasses Bevy's scheduler, so we'd lose most of the Bevy schedules functionality such as system ordering, run conditions.

I looked into whether Bevy could expose a schedule system graph, and it does, but I didn't find any way how we can iterate through existing systems and call them with custom parameters. All the returned types force systems to have `()` as `Input`, unless I've missed something.

Most likely, this requires contributions to Bevy itself.

# Proposed solution

Return the previous flavour of the `show` method that accepts `&egui::Context` for containers.

emilk:
* Closes <https://github.com/emilk/egui/issues/8415>

Adds `Context::root_ui` to access the root `Ui` of the current pass from anywhere, and `Context::run_pass` (like `run_ui`, but the callback gets `&Context`). The root `Ui` is stored in the viewport state during the pass and created on demand, so `root_ui` also works with `begin_pass`/`end_pass`.

This lets integrations like `bevy_egui`, where ui code is spread over many systems sharing only a `Context`, add panels again:

```rust
ctx.root_ui(|ui| {
    egui::Panel::left("left").show(ui, |ui| { /* … */ });
});
```

Panels added this way share one root `Ui`, so they stack correctly and `Context::is_pointer_over_egui` stays correct. Calling `root_ui` while the root `Ui` is already borrowed (e.g. inside the `run_ui` callback) panics with a clear message.

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -187,6 +187,25 @@ impl ContextImpl {
 
 // ----------------------------------------------------------------------------
 
+/// The root [`Ui`] of the current pass of a viewport.
+///
+/// See [`Context::root_ui`].
+#[derive(Default)]
+#[expect(clippy::large_enum_variant, reason = "There is only one per viewport")]
+enum RootUi {
+    /// Not yet created this pass.
+    ///
+    /// It is created on demand by [`Context::root_ui`].
+    #[default]
+    Pending,
+
+    /// Created, and ready to be borrowed by [`Context::root_ui`].
+    Available(Ui),
+
+    /// Currently borrowed by a call to [`Context::root_ui`].
+    Borrowed,
+}
+
 /// State stored per viewport.
 ///
 /// Mostly for internal use.
@@ -251,6 +270,11 @@ pub struct ViewportState {
     ///
     /// See [`crate::Options::sync_window_theme`].
     pub(crate) last_sent_window_theme: Option<crate::SystemTheme>,
+
+    /// The root [`Ui`] of the current pass.
+    ///
+    /// Only set between [`Context::begin_pass`] and [`Context::end_pass`].
+    root_ui: RootUi,
 }
 
 /// What called [`Context::request_repaint`] or [`Context::request_discard`]?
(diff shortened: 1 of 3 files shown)


vladbat00:
Thank you for the change!

The new nesting level in the user code is a bit unfortunate, but it works! I guess it's not a bad compromise, considering it'll have to be used only where panels are constructed.

PR #8542, opened 2026-09-09
Rename `Ui::id` to `Ui::scope_id` and clarify its docstring

emilk:
`Ui::id` was easy to confuse with `Ui::unique_id`. It is the id of the _scope_ the `Ui` belongs to: stable across frames, shared by sibling `Ui`s, and the base for widget state ids. `Ui::unique_id` is unique per `Ui` but unstable.

This renames `Ui::id` to `Ui::scope_id` (old name deprecated) and rewrites both docstrings to say so.

Follow-up to discussion in:
* https://github.com/emilk/egui/pull/8534#discussion_r3969041538

### DIFF
--- a/crates/egui/src/ui.rs
+++ b/crates/egui/src/ui.rs
@@ -124,7 +124,7 @@ impl Ui {
 
         let placer = Placer::new(max_rect, layout);
         let ui_stack = UiStack {
-            id: scope_id,
+            unique_id: scope_id,
             layout_direction: layout.main_dir,
             info: ui_stack_info,
             parent: None,
@@ -248,7 +248,7 @@ impl Ui {
 
         let placer = Placer::new(max_rect, layout);
         let ui_stack = UiStack {
-            id: unique_id,
+            unique_id,
             layout_direction: layout.main_dir,
             info: ui_stack_info,
             parent: Some(Arc::clone(&self.stack)),
--- a/crates/egui/src/ui.rs
+++ b/crates/egui/src/ui.rs
@@ -316,7 +316,7 @@ impl Ui {
 
     // -------------------------------------------------
 
-    /// The [`Id`] scope of this `Ui`.
+    /// The stable [`Id`] scope of this `Ui`.
     ///
     /// This is _stable_ from one frame to the next,
     /// so it should be used as the base for the [`Id`]s of widgets that store state
@@ -342,7 +342,7 @@ impl Ui {
         self.scope_id
     }
 
-    /// A globally unique [`Id`] of this `Ui`.
+    /// A globally unique, but unstable, [`Id`] of this `Ui`.
     ///
     /// This is NOT _stable_: it is based on where in the widget hierarchy this `Ui` is,
     /// so it changes if widgets are added or removed before it.
--- a/crates/egui/src/ui.rs
+++ b/crates/egui/src/ui.rs
@@ -238,12 +238,12 @@ impl Ui {
         debug_assert!(!max_rect.any_nan(), "max_rect is NaN: {max_rect:?}");
 
         let id_source = id_source.unwrap_or_else(|| IdSource::Child(IdSalt::new("child")));
-        let (stable_id, unique_id) = match id_source {
+        let (scope_id, unique_id) = match id_source {
             IdSource::Explicit(id) => (id, id),
             IdSource::Child(id_salt) => {
-                let stable_id = self.scope_id.with(id_salt);
-                let unique_id = stable_id.with(self.next_auto_id_salt);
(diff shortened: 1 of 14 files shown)


crates/egui/src/ui.rs
emilk:
```suggestion
    pub fn new(ctx: Context, scope_id: Id, ui_builder: UiBuilder) -> Self {
```

crates/egui/src/ui_builder.rs
emilk: we should rename this too

crates/egui/src/ui.rs
```diff
@@ -139,7 +124,7 @@ impl Ui {
 
         let placer = Placer::new(max_rect, layout);
         let ui_stack = UiStack {
-            id,
+            id: scope_id,
```
emilk: let's rename UiStack::id to scope_id

```suggestion
            scope_id,
```

emilk: `UiStack::id` held `Ui::unique_id` (set in `new_child`, since #5192), not the scope id, and `menu.rs` keys `MenuState` on it. Renamed it to `UiStack::unique_id` instead, as agreed.

— Claude

crates/egui/src/ui.rs
```diff
@@ -139,7 +124,7 @@ impl Ui {
 
         let placer = Placer::new(max_rect, layout);
         let ui_stack = UiStack {
-            id,
+            unique_id: scope_id,
```
emilk: this looks like a bug!

emilk: I believe this is intended, not a bug: a root `Ui` has no parent to derive a unique id from, so the id passed to `Ui::new` (a panel or area id, which must be globally unique) serves as both `scope_id` and `unique_id`. This was already the case before this PR (`id` and `unique_id: id`); the rename just made it stand out. Added a comment in 7364afbd8 explaining it. Left this thread open in case you meant something else.

— Claude

crates/egui/src/ui.rs
```diff
@@ -273,7 +262,7 @@ impl Ui {
         };
 
         let mut child_ui = Ui {
-            id: stable_id,
+            scope_id: stable_id,
```
emilk: let's avoid `stable_id` here and everywhere

crates/egui/src/ui.rs
```diff
@@ -880,22 +881,33 @@ impl Ui {
 
 /// # [`Id`] creation
 impl Ui {
-    /// Use this to generate widget ids for widgets that have persistent state in [`Memory`].
+    /// Generate an [`Id`] for a widget that has persistent state in [`Memory`].
+    ///
+    /// This is the same as `ui.scope_id().with(id_salt)`.
+    /// Since it is based on the stable [`Self::scope_id`], it is stable from frame to frame,
```
emilk: let's replace `from frame to frame` with `over time` here and everywhere

```suggestion
    /// Since it is based on the stable [`Self::scope_id`], it is stable over time,
```

crates/egui/src/ui_builder.rs
```diff
@@ -58,21 +58,28 @@ impl UiBuilder {
...
+    /// You should set the same unique `scope_id` at every place in the ui tree where you want the
     /// child widgets to share state.
     /// If the child widgets are not moved in the ui tree, use [`UiBuilder::id_salt`] instead.
     ///
-    /// This is a shortcut for `.id_salt(my_id).global_scope(true)`.
+    /// The `scope_id` is also used as the [`Ui::unique_id`] of the new `Ui`, so it must be globally unique.
     #[inline]
-    pub fn id(mut self, id: Id) -> Self {
-        self.id_source = Some(IdSource::Explicit(id));
+    pub fn scope_id(mut self, scope_id: Id) -> Self {
```
emilk, line 71: Ah, so this sets _both_ ids. How do we signal that with a name? 🤔 
Whatever we pick it should be the same as the `Id` parameter name of `Ui::new`.

Maybe `id` is good enough?

crates/egui/src/ui_stack.rs
```diff
@@ -207,7 +207,11 @@ impl UiTags {
 #[derive(Debug)]
 pub struct UiStack {
     // stuff that `Ui::child_ui` can deal with directly
-    pub id: Id,
+    /// The [`crate::Ui::unique_id`] of this [`crate::Ui`].
+    ///
+    /// Globally unique, but NOT stable over time.
+    /// For the stable id scope, see [`crate::Ui::scope_id`].
+    pub unique_id: Id,
```
emilk, line 214: Let's add scope_id here too

PR #8543, opened 2026-09-09
Use `Ui::unique_id` for `WidgetRect::parent_id`

emilk:
`WidgetRect::parent_id` used `Ui::id`, the stable id _scope_ shared by sibling `Ui`s, while accesskit and `WidgetRect::id` use the per-`Ui` `unique_id`. The `warn_if_rect_changes_id` debug check relies on `parent_id` changing during a cascading id shift, but cascades propagate through `unique_id`, not the scope id. So a parent's auto-id shifting produced a false positive for every widget inside it.

Now `parent_id` is the parent's `unique_id`. Adds a regression test.

Known limitation: the child `Ui`'s own `WidgetRect` still warns in that case, since its parent's id is genuinely unchanged. The heuristic cannot tell that apart from a bug.

Found while working on:
* https://github.com/emilk/egui/pull/8542

### DIFF
--- a/crates/egui/src/response.rs
+++ b/crates/egui/src/response.rs
@@ -152,7 +152,7 @@ bitflags::bitflags! {
 }
 
 impl Response {
-    /// The [`Id`] of the parent [`crate::Ui`] that hosts this widget.
+    /// The [`crate::Ui::unique_id`] of the parent [`crate::Ui`] that hosts this widget.
     ///
     /// Looks up the [`WidgetRect`] from the current (or previous) pass.
     pub fn parent_id(&self) -> Id {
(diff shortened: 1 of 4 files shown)


crates/egui/src/response.rs
emilk, line 165: make sure we document the fields in `WidgetRect` too

crates/egui/src/widget_rect.rs
```diff
@@ -17,7 +17,7 @@ pub struct WidgetRect {
     /// You can ensure globally unique ids using [`crate::Ui::push_id`].
     pub id: Id,
```
emilk, line 21: this should maybe also be renamed, or at least link to `crate::Ui::unique_id`

PR #8512, opened 2026-09-05
Render any glyph and colored emoji on native 🚀

emilk:
Add full emoji support to eframe on native! Also supports any glyph that your system fonts support.

Test string: ✨ We support ไทย, 한국어, and 🦀 🚀 🤖/🦾/📷/🎞️

Chain, in order, all merged into `emilk/emojis`:

* https://github.com/emilk/egui/pull/8491
* https://github.com/emilk/egui/pull/8492
* https://github.com/emilk/egui/pull/8493
* https://github.com/emilk/egui/pull/8508
* https://github.com/emilk/egui/pull/8509
* https://github.com/emilk/egui/pull/8510
* https://github.com/emilk/egui/pull/8511

PR #8545, opened 2026-09-09
Allow multiple `GlyphRasterizer`s, with priority over installed fonts

emilk:
`GlyphRasterizer` now has a `FontPriority`:

* `FontPriority::Highest`: asked about every cluster, before any font. Use it to override how specific glyphs looks
* `FontPriority::Lowest` (default): a fallback, only asked about clusters no installed font (or `FontProvider`) has. Same as before.

Several rasterizers can be installed; within a tier they are asked in order, and the first `Some` wins.

```rust
ctx.add_glyph_rasterizer(
    egui::GlyphRasterizer::new(|request| (request.cluster == "…").then(|| my_ellipsis(request)))
        .with_priority(egui::FontPriority::Highest),
);
```

API changes (all unreleased): `Context::set_glyph_rasterizer(Option)` → `add_glyph_rasterizer` / `set_glyph_rasterizers(Vec)`, same on `epaint::Fonts`.

Follow-up to:
* https://github.com/emilk/egui/pull/8512

### DIFF
--- a/crates/egui/src/context.rs
+++ b/crates/egui/src/context.rs
@@ -375,7 +375,7 @@ impl ViewportRepaintInfo {
 struct ContextImpl {
     fonts: Option<Fonts>,
     font_definitions: FontDefinitions,
-    glyph_rasterizer: Option<GlyphRasterizer>,
+    glyph_rasterizers: Vec<GlyphRasterizer>,
     font_providers: Vec<Arc<dyn FontProvider>>,
 
     memory: Memory,
@@ -596,9 +596,7 @@ impl ContextImpl {
             profiling::scope!("Fonts::new");
             let mut fonts = Fonts::new(text_options, self.font_definitions.clone())
                 .with_font_providers(self.font_providers.clone());
-            if let Some(glyph_rasterizer) = &self.glyph_rasterizer {
-                fonts = fonts.with_glyph_rasterizer(glyph_rasterizer.clone());
-            }
+            fonts.set_glyph_rasterizers(self.glyph_rasterizers.clone());
             fonts
         });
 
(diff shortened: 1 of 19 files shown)


crates/egui/src/context.rs
```diff
@@ -769,15 +767,29 @@ impl Default for Context {
...
 
 impl Context {
-    /// Set the platform glyph rasterizer, e.g. the browser on web.
+    /// Add a [`GlyphRasterizer`], e.g. the browser on web, or for custom glyphs.
     ///
-    /// It is used for grapheme clusters that no installed font can render,
+    /// By default it is a fallback, used for grapheme clusters that no installed font can render,
     /// and that no [`FontProvider`] has a font for.
+    /// With [`crate::FontPriority::Highest`] it instead overrides the installed fonts.
+    /// See [`GlyphRasterizer`] for details.
```
```suggestion
    /// Depending on [`FontPriority`] this is either a fallback or an override
```

crates/egui_extras/src/svg_glyph.rs
```diff
@@ -0,0 +1,280 @@
...
+    MAX_GLYPH_SIZE, RasterizedGlyph, Vec2, vec2,
+};
+use resvg::{
+    tiny_skia::Pixmap,
+    usvg::{Transform, Tree},
+};
+
+/// A glyph rendered from an SVG, for a [`GlyphRasterizer`].
+///
+/// Use it to override how a character looks (e.g. `…`),
```
emilk:
```suggestion
/// Use it to override how a character looks
```

crates/egui_extras/src/svg_glyph.rs
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    /// Use an already parsed SVG.
+    pub fn from_tree(tree: Tree) -> Self {
+        Self {
+            tree: Arc::new(tree),
+            height_em: 1.0,
+            baseline_offset_em: 0.0,
+            advance_em: None,
+            is_color: false,
```
emilk: can we by default look for non-grayscale colors in the tree? Maybe a bad idea 🤔

emilk: Done: `from_tree` now walks the tree and sets `is_color` if any fill, stroke, gradient stop, pattern, or embedded image uses a non-gray color. `with_color(bool)` still overrides it.

Downside: a gray icon becomes tinted (probably wanted), a subtly tinted one keeps its color (probably also wanted). Ambiguous only for SVGs mixing both.

Claude

crates/egui_extras/src/svg_glyph.rs
```diff
@@ -0,0 +1,360 @@
...
+/// Use [`Self::with_color`] to decide yourself.
+///
+/// ```no_run
+/// # let ctx = egui::Context::default();
+/// let svg = std::fs::read("ellipsis.svg").unwrap();
+/// let ellipsis = egui_extras::SvgGlyph::from_bytes(&svg).unwrap();
+/// ctx.add_glyph_rasterizer(ellipsis.into_rasterizer("…"));
+/// ```
+#[derive(Clone, Debug)]
+pub struct SvgGlyph {
```
lucasmerlin, line 33: I feel like this deserves a example and changelog mention. 
And maybe a screenshot test so we can verify the rendering actually works?

lucasmerlin (commented):
Nice!

lucasmerlin (approved):

crates/egui_extras/src/svg_glyph.rs
emilk: Yeah, good call-out

PR #8555, opened 2026-09-11
Remove `WidgetType` in favor of `accesskit::Role`

### LINKED ISSUE #8474: Add `WidgetType::Tab`
There is no suitable `WidgetType` for tabs while I was implementing AccessKit support for egui_dock. AccessKit provides [`accesskit::Role::Tab`](https://docs.rs/accesskit/latest/accesskit/enum.Role.html#variant.Tab) and other tab-related widgets that will be described by assistive technologies accurately.

Add a `WidgetType::Tab` variant mapping to `accesskit::Role::Tab`, with `WidgetInfo::selected` triggering set_selected instead of set_toggled for it. Possibly also some way to mark a container `Ui` as `Role::TabList` or `Role::TabPanel`.

Currently I had to override the role after `widget_info` using `accesskit_node_builder`:
```rust

response.widget_info(|| WidgetInfo::selected(WidgetType::Button, enabled, selected, title));
ui.ctx().accesskit_node_builder(response.id, |node| {
    node.set_role(accesskit::Role::Tab);
    node.set_selected(selected);
    node.clear_toggled();
});
```

Also tried using other existing `WidgetTypes` like selectable label, but assistive technologies will interpret it as a selectable label instead of a tab, which can be confusing.

**Additional context**
See https://github.com/anhosh/egui_dock/pull/345.

**Edit:** the following has been opened to #8557.

During that PR, I also found out that there is no way to describe the direction (horizontal/vertical) on `WidgetType::ResizeHandle`, and all splitters are interpreted as "horizontal" in VoiceOver, and also window resizing doesn't work using VoiceOver (it always reports the window resizing handle as a horizontal splitter stuck at 0%). The PR works around this by

```rust
let ak_steps = ui.input(|input| {
    input.num_accesskit_action_requests(response.id, accesskit::Action::Increment) as f32
        - input.num_accesskit_action_requests(response.id, accesskit::Action::Decrement) as f32
});
let ak_set_value = ui.input(|input| {
    input
        .accesskit_action_requests(response.id, accesskit::Action::SetValue)
        .find_map(|request| match request.data {
            Some(accesskit::ActionData::NumericValue(value)) => Some(value as f32),
            _ => None,
        })
});
let mut delta = arrow_key_offset.unwrap_or(response.drag_delta());
delta.dim_point += ak_steps * style.separator.arrow_key_step_distance;
if let Some(value) = ak_set_value {
    delta.dim_point += (value.clamp(0.0, 1.0) - split.fraction) * rect.dim_size();
}
apply_separator_delta(split, SeparatorAxis::sep_axis, delta, &style.separator);

if response.double_clicked() {
    split.fraction = 0.5;
}

// Allow assistive technologies to adjust the separator fraction.
let fraction = split.fraction;
let step = style.separator.arrow_key_step_distance / rect.dim_size().max(1.0);
ui.ctx().accesskit_node_builder(response.id, |node| {
    node.set_orientation(accesskit::Orientation::ak_orientation);
    node.set_numeric_value(fraction.into());
    node.set_min_numeric_value(0.0);
    node.set_max_numeric_value(1.0);
    node.set_numeric_value_step(step.into());
    node.add_action(accesskit::Action::Increment);
    node.add_action(accesskit::Action::Decrement);
    node.add_action(accesskit::Action::SetValue);
});
```

lucasmerlin:
- closes https://github.com/emilk/egui/issues/8474

Its a required dependency now so we might as well their type for best compatibility

### DIFF
--- a/crates/egui/src/containers/collapsing_header.rs
+++ b/crates/egui/src/containers/collapsing_header.rs
@@ -1,6 +1,6 @@
 use crate::{
-    AsIdSalt, Context, Id, IdSalt, InnerResponse, NumExt as _, Rect, Response, Sense, TextStyle,
-    TextWrapMode, Ui, UiBuilder, UiKind, UiStackInfo, WidgetInfo, WidgetText, WidgetType, emath,
+    AsIdSalt, Context, Id, IdSalt, InnerResponse, NumExt as _, Rect, Response, Role, Sense,
+    TextStyle, TextWrapMode, Ui, UiBuilder, UiKind, UiStackInfo, WidgetInfo, WidgetText, emath,
     epaint, pos2, remap, remap_clamp, vec2,
 };
 use emath::GuiRounding as _;
@@ -546,7 +546,11 @@ impl CollapsingHeader {
         }
 
         header_response.widget_info(|| {
-            WidgetInfo::labeled(WidgetType::CollapsingHeader, ui.is_enabled(), galley.text())
+            WidgetInfo::labeled(Role::DisclosureTriangle, ui.is_enabled(), galley.text())
+        });
+
+        ui.ctx().accesskit_node_builder(header_response.id, |node| {
+            node.set_expanded(state.is_open());
         });
 
         let openness = state.openness(ui.ctx());
--- a/crates/egui/src/containers/window.rs
+++ b/crates/egui/src/containers/window.rs
@@ -679,11 +679,7 @@ impl Window<'_> {
         }
 
         area.with_widget_info(|| {
-            WidgetInfo::labeled(
-                WidgetType::Window,
-                true,
-                title.text().as_deref().unwrap_or(""),
-            )
+            WidgetInfo::labeled(Role::Window, true, title.text().as_deref().unwrap_or(""))
         });
 
         {
@@ -1082,7 +1078,7 @@ fn do_resize_interaction(
             },
         );
 
-        response.widget_info(|| WidgetInfo::new(crate::WidgetType::ResizeHandle));
+        response.widget_info(|| WidgetInfo::new(Role::Splitter));
 
         SideResponse {
             hover: response.hovered(),
@@ -1381,7 +1377,7 @@ fn title_ui(
         );
         icon_response.widget_info(|| {
             WidgetInfo::labeled(
-                WidgetType::Button,
+                Role::Button,
                 child_ui.is_enabled(),
                 if collapsing.is_open() { "Hide" } else { "Show" },
             )
@@ -1466,8 +1462,7 @@ fn title_ui(
 fn close_button(ui: &mut Ui, rect: Rect) -> Response {
     let close_id = ui.auto_id_with("window_close_button");
     let response = ui.interact(rect, close_id, Sense::click());
-    response
-        .widget_info(|| WidgetInfo::labeled(WidgetType::Button, ui.is_enabled(), "Close window"));
+    response.widget_info(|| WidgetInfo::labeled(Role::Button, ui.is_enabled(), "Close window"));
 
     ui.expand_to_include_rect(response.rect);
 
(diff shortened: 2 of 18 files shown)


crates/egui/src/containers/collapsing_header.rs
```diff
@@ -546,7 +546,11 @@ impl CollapsingHeader {
         }
 
         header_response.widget_info(|| {
-            WidgetInfo::labeled(WidgetType::CollapsingHeader, ui.is_enabled(), galley.text())
+            WidgetInfo::labeled(Role::DisclosureTriangle, ui.is_enabled(), galley.text())
+        });
```
lucasmerlin, line 550: This should be the only breaking change from output (was `Button` before). I wish the role had a better name though...

crates/egui/src/lib.rs
```diff
@@ -437,6 +437,7 @@ mod callstack;
 pub mod class;
 
 pub use accesskit;
+pub use accesskit::Role;
```
emilk, line 440: Please also add a `#[deprecated(…` alias for the old `WidgetType` so we don't break compatibility (as much)

crates/egui/src/data/output.rs
```diff
@@ -745,33 +745,16 @@ impl WidgetInfo {
...
-        };
-
-        let mut description = widget_type.to_owned();
+        let mut description = String::new();
+        for ch in format!("{role:?}").chars() {
+            if ch.is_ascii_uppercase() && !description.is_empty() {
+                description.push(' ');
+            }
+            description.push(ch.to_ascii_lowercase());
+        }
```
emilk, line 754: break this out into a function with a self-describing name and a couple of doctests

crates/egui/src/containers/window.rs
```diff
@@ -1082,7 +1078,7 @@ fn do_resize_interaction(
             },
         );
 
-        response.widget_info(|| WidgetInfo::new(crate::WidgetType::ResizeHandle));
+        response.widget_info(|| WidgetInfo::new(Role::Splitter));
```
emilk, line 1081: I guess this is the closest we have

crates/egui/src/containers/collapsing_header.rs
```diff
@@ -546,7 +546,11 @@ impl CollapsingHeader {
...
+
+        ui.ctx().accesskit_node_builder(header_response.id, |node| {
+            node.set_expanded(state.is_open());
         });
```
emilk, line 554: is this a bug fix?

emilk (approved):

crates/egui/src/lib.rs
emilk: Added a `#[deprecated]` `WidgetType` type alias for `Role`, with a table of the variants that were renamed (e.g. `TextEdit` -> `TextInput`, `ColorButton` -> `ColorWell`).

crates/egui/src/data/output.rs
emilk: Broken out into a public `role_description(role)` function with doctests. It also returns an empty string for `Role::Unknown`, matching what `WidgetType::Other` used to do.

crates/egui/src/containers/collapsing_header.rs
emilk: Yes. `set_expanded` appears nowhere on main, so collapsing headers never reported their open/closed state to the accessibility tree. `Role::DisclosureTriangle` is only useful with it, so it comes along with the role change.

crates/egui/src/containers/window.rs
emilk: Agreed, `Splitter` is the nearest accesskit role for a resize handle. Left as is.

emilk (commented):
LGTM

PR #8558, opened 2026-09-11
Fix deep indents blowing up memory in debug mode

IsseW:
Including the whole Id in the Id chain for indents made memory blow up exponentially. Causing freezes for me in some instances.


emilk:
I would like to fix this in the Id/IdSalt debug-tracker instead

EDIT: had a clanker try something:
* https://github.com/emilk/egui/pull/8619

PR #8566, opened 2026-09-15
Bring window to front if an `egui_inspection` screenshot is requested

emilk:
A minimized or occluded window stops painting on most platforms, and every inspection request is served from a frame. An inspector attached to a backgrounded app could therefore only time out — `screenshot` hit this the moment another window covered the app:

```
egui_inspect rpc failed: code: 'The operation was cancelled', message: "Timeout expired"
```

A screenshot now restores and focuses the window when it reports itself hidden. That is deliverable in this state: eframe keeps ticking app logic while hidden precisely so the app can ask to be shown again.

Only screenshots do this. The other requests need a painting app too, but they read and drive an app the user may be watching, and a window that takes focus whenever something reads its widget tree is worse than one that answers late. They raise it when they time out instead, so the next request finds a painting app.

A window that does not report itself hidden is left alone.

Written by Claude (LLM), from a session driving a Rerun Viewer over `egui_mcp`.


emilk:
Let's make a repro example and look for better solutions - we should be able to support screenshots for hidden windows too (maybe not minimized, but hidden)

lucasmerlin:
Fixed via #8571

PR #8571, opened 2026-09-15
Handle `egui_inspection` requests even while a window is hidden

emilk:
An inspection request is served from a frame, and a minimized or occluded window runs none: eframe skips the whole pass, and wgpu hands out no surface texture. An inspector attached to a backgrounded app could only time out.

* `ViewportCommand::RequestPaintWhileHidden` asks for a painted frame from a hidden window. It rides the command channel, which already reaches the integration while hidden (`run_logic` hands it the pending commands), so `Context` needs nothing new.
* eframe paints when a pending action wants paint: that command, or a screenshot, which asks for the same thing.
* egui-wgpu captures without a surface texture: the capture renders to its own texture anyway, and only the present needs the surface. Pending screenshot callbacks used to be dropped silently whenever the frame could not be acquired, which also hit visible-but-occluded windows.

`egui_inspection` sends the command, so nothing raises or focuses the window.

Repro and manual test: `cargo run -p test_inspection_hidden_window`. Verified on macOS: screenshots and trees are served while the window is minimized.

Replaces:
* https://github.com/emilk/egui/pull/8566

Written by Claude (LLM).


crates/eframe/src/epi.rs
```diff
@@ -161,6 +161,9 @@ pub trait App {
     /// [`egui::InputState::viewport`], but the rest of [`egui::Context::input`]
     /// (events, time, …) is that of the last shown frame.
     ///
+    /// Send [`egui::ViewportCommand::RequestPaintWhileHidden`] when you want a full pass,
+    /// painted and all, out of a hidden window — e.g. to screenshot a backgrounded app.
+    ///
```
emilk:
```suggestion
    /// Send [`egui::ViewportCommand::RequestPaintWhileHidden`] if you want `App::ui` to be called
    /// even if the application is hidden.
```

crates/egui-winit/src/lib.rs
```diff
@@ -1735,6 +1735,26 @@ pub enum ActionRequested {
     Cut,
     Copy,
     Paste,
+
+    /// Paint a frame even though the window is hidden (minimized or occluded).
```
emilk:
```suggestion
    /// Run `App::ui` and paint a frame even though the window is hidden (minimized or occluded).
```

crates/egui/src/viewport.rs
```diff
@@ -1218,6 +1218,25 @@ pub enum ViewportCommand {
     ///
     /// This is equivalent to the system keyboard shortcut for paste (e.g. CTRL + V).
     RequestPaste,
+
+    /// Request a painted frame, even though this window is hidden.
```
emilk:
```suggestion
    /// Run `eframe::App::ui` and paint a frame, even though this window is hidden.
```

crates/egui-winit/src/lib.rs
```diff
@@ -1735,6 +1735,26 @@ pub enum ActionRequested {
...
     Paste,
+
+    /// Run `App::ui` and paint a frame even though the window is hidden (minimized or occluded).
+    ///
+    /// Someone wants the ui to run with nothing on screen to show for it. This is how a tool
+    /// drives an app that sits in the background: `egui_inspection` sends it with every
+    /// request, so that a hidden app still runs its ui — painting the screenshot, rebuilding
+    /// the widget tree, and applying injected input.
+    ///
+    /// See [`egui::ViewportCommand::RequestPaintWhileHidden`].
```
lucasmerlin:
```suggestion
    /// Run App::ui and paint a frame even while the window is hidden (minimized or occluded).
    ///
    /// This is useful when the UI needs to keep running without anything currently visible on
    /// screen. For example, a tool can use this to drive an app in the background:
    /// egui_inspection includes it with every request so that a hidden app still runs its UI,
    /// paints the screenshot, rebuilds the widget tree, and applies injected input.
    ///
    /// See [egui::ViewportCommand::RequestPaintWhileHidden].
```

lucasmerlin (approved):
Nice!

PR #8572, opened 2026-09-16
Respect the menu style config in sub menus

lucasmerlin:
Previously it was ignored since menu_style was applied within `Popup` and the frame was created on the previous, unchanged outer ui.


crates/egui/src/containers/menu.rs
```diff
@@ -458,7 +458,7 @@ impl SubMenu {
         button_response: &Response,
         content: impl FnOnce(&mut Ui) -> R,
     ) -> Option<InnerResponse<R>> {
-        let frame = Frame::menu(ui.style());
+        let measurement_frame = Frame::menu(ui.style());
```
emilk, line 461: this is a bit of a weird name - maybe add a comment to it?

crates/egui_kittest/tests/snapshots/submenu_style.png
emilk, line 1: I would like to also see what style a sub-sub menu gets now. Does it inherit it from the closest ancestor, i.e. also become red with green outline?

emilk (approved):

PR #8574, opened 2026-09-16
Keep the AccessKit role a widget declared when its text is selectable

emilk:
`update_accesskit_for_text_widget` unconditionally called `builder.set_role`, so the selectable-label pass overwrote the role a widget had already reported through its `WidgetInfo`. A `Link` came out as `Role::Label`, indistinguishable from static text to a screen reader.


crates/egui/src/widgets/text_edit/builder.rs
```diff
@@ -1028,19 +1028,23 @@ impl<'t> TextEdit<'t> {
...
         let role = if password {
             accesskit::Role::PasswordInput
         } else if multiline {
             accesskit::Role::MultilineTextInput
         } else {
             accesskit::Role::TextInput
         };
+        ui.ctx().accesskit_node_builder(id, |builder| {
+            builder.set_role(role);
+        });
```
emilk: nit

```suggestion
        ui.ctx().accesskit_node_builder(id, |builder| {
	        // `WidgetInfo` only knows about `WidgetType::TextEdit`, which maps to
	        // `Role::TextInput`, so refine the role here:
	let role = if password {
	accesskit::Role::PasswordInput
	} else if multiline {
	accesskit::Role::MultilineTextInput
	} else {
	accesskit::Role::TextInput
	};
            builder.set_role(role);
        });
```

lucasmerlin (approved):
Oof, good catch!

PR #8580, opened 2026-09-17
Add RangeSlider, a two-handled slider sharing Slider's core

### LINKED ISSUE #2744: (Feature Request) Double-handed slider
### Discussed in https://github.com/emilk/egui/discussions/2006

<div type='discussions-op-text'>

<sup>Originally posted by **majorsopa** September  5, 2022</sup>
image

The image attached is a google image roughly showing what I am talking about. Instead of its value being an `f64`, its value would be a `RangeInclusive<f64>` or similar. How would I go about creating this slider with 2 handles? </div>

gavrelina:
* Closes #2744.

Adds `RangeSlider`: two handles selecting a sub-range, an editable value at each end, and the accent fill between them.

Three commits:

**1. Move `Slider`'s geometry, painting and interaction into a private `slider_core`.** No new `Visuals` field. The one rendering change: the trailing fill is clamped to the rail, which moves it by under a pixel for narrow rect handles.

**2. Add `RangeSlider` on top of it.** Rail, handles, hover growth, `HandleShape` and the fill are the same code `Slider` runs, so the two do not look different.

** 3. Added per-handle keyboard and screen-reader support. Each handle is its own focus stop, so Tab reaches either end and the arrow keys move the focused one. Each handle reports itself to AccessKit bounded by its neighbour rather than by the end of the rail.

A separate widget rather than a mode on `Slider`, because none of this means anything with one handle:

- the grabbed handle is remembered for the whole drag, so dragging one handle past the other does not hand the pointer to its neighbour
- when the handles coincide, the side the pointer is on decides which one moves
- `min_separation` sets how close they may come

One change: `paint_fill` takes a `Rangef` instead of a handle centre.

API changes from the review follow-ups:

- `Slider::prefix`/`suffix` take `impl IntoAtoms`, like `DragValue`, so images and styled text work. Plain `char` or number arguments no longer do.
- A prefix goes in front of any prefix already set, and a suffix after any suffix already set, on `DragValue`, `Slider`, `RangeSlider`, and `TextEdit`. `Slider` and `TextEdit` used to replace.
- `ValueFormat`, `SliderSpec`, and `DragValueSettings` are public, with `format()`, `spec()`, and `drag_value()` builders on both sliders.
- `DragValue::atoms()` returns `Atoms` by value.

Snapshots: the new gallery row changes `widget_gallery_*` and `demos/Scene` (the Scene demo embeds a Widget Gallery). `RangeSlider` is also added to `test_widgets.rs`, which adds two new snapshots.

Written with Claude (llm), I (gavrelina) have reviewed the code myself (I did!).


crates/egui_demo_lib/src/demo/widget_gallery.rs
```diff
@@ -15,6 +15,7 @@ pub struct WidgetGallery {
     opacity: f32,
     radio: Enum,
     scalar: f32,
+    range: (f32, f32),
```
emilk: style: could use Rangef here

crates/egui/src/widgets/slider.rs
emilk: While we're at it, I think we should group together in a struct all the things that relate to the drag-value display on the side, so that `ValueOptions` can just point to that:

prefix, suffix, min_decimals, max_decimals, custom_formatter, custom_parser, update_while_editing

crates/egui/src/widgets/drag_value.rs
```diff
@@ -777,6 +678,17 @@ impl Widget for DragValue<'_> {
     }
 }
 
+/// The text of the atoms, for screen readers.
+fn atoms_text(atoms: &Atoms<'_>) -> String {
```
emilk: this should be a member of `Atoms`

crates/egui/src/widgets/value_format.rs
```diff
@@ -0,0 +1,240 @@
+//! How a number is written, read back, and applied while being edited.
```
emilk, line 1:
```suggestion
//! How a number is written, read back, and applied while being edited.
//!
//! Used by …
```

crates/egui/src/widgets/range_slider.rs
```diff
@@ -0,0 +1,639 @@
...
+}
+
+impl<'a> RangeSlider<'a> {
+    /// Creates a new horizontal range slider.
+    ///
+    /// The values are clamped to `range`, and `low` is clamped to `high`.
+    pub fn new<Num: emath::Numeric>(
+        low: &'a mut Num,
+        high: &'a mut Num,
+        range: RangeInclusive<Num>,
```
emilk:
```suggestion
        range: impl Into<RangeInclusive<Num>>,
```

crates/egui/src/widgets/range_slider.rs
```diff
@@ -0,0 +1,651 @@
...
+};
+
+use super::drag_value::{GetSetValue, clamp_value_to_range, get, set};
+use super::slider::{SliderClamping, SliderOrientation};
+use super::slider_core::{self, SliderGeometry, SliderSpec, StepOptions, ValueOptions};
+use super::value_format::ValueFormat;
+
+/// Which handle a drag gesture grabbed, remembered for the whole gesture.
+#[derive(Clone, Copy, Default)]
+struct GrabbedLow(bool);
```
emilk: an enum would be nicer (`enum Handle { Low, High }` )

crates/egui/src/widgets/range_slider.rs
```diff
@@ -0,0 +1,651 @@
...
+
+/// Select a range of numbers with a two-handled slider.
+///
+/// The handles may meet but never cross. Looks and behaves like [`crate::Slider`], whose
+/// geometry, painting and pointer handling it shares.
```
emilk:
```suggestion
/// The handles may meet but never cross.
/// Looks and behaves like [`crate::Slider`], but with two handles.
```

crates/egui/src/widgets/range_slider.rs
```diff
@@ -0,0 +1,651 @@
...
+
+    /// The closest the two handles may come to each other, in values.
+    min_separation: f64,
+
+    /// Sets the minimal step of the widget value
+    step: Option<f64>,
+
+    drag_value_speed: Option<f64>,
+    format: ValueFormat<'a>,
+    handle_shape: Option<HandleShape>,
```
emilk: take all the things here that are also in `Slider` and put them in a new `struct SliderCore`, just to further emphesize what is similar and what is different

crates/egui/src/widgets/range_slider.rs
```diff
@@ -0,0 +1,667 @@
...
+    ) -> Self {
+        Self {
+            get_set_low: Box::new(get_set_low),
+            get_set_high: Box::new(get_set_high),
+            range,
+            spec: SliderSpec {
+                logarithmic: false,
+                smallest_positive: 1e-6,
+                largest_finite: f64::INFINITY,
+            },
```
emilk:
```suggestion
            spec: SliderSpec::default()
```

…assuming it is the same between the two types of sliders

emilk (commented):
LGTM

crates/egui/src/widgets/range_slider.rs
```diff
@@ -100,6 +100,12 @@ impl<'a> RangeSlider<'a> {
         self
     }
 
+    #[inline]
+    pub fn text_color(mut self, text_color: Color32) -> Self {
```
emilk, line 105: docstring for all public methods on all public widgets

PR #8581, opened 2026-09-17
Improve accessibility of popups, menus, tooltips, modals and panels

emilk:
Found while making the Rerun viewer navigable by accessible name: floating containers were anonymous `Unknown`/`GenericContainer` nodes at the root of the tree, and icon-only buttons with a tooltip had no name.

* `Ui`s get an accesskit role from their `UiKind`: panels are `Pane`, modals and popups `Dialog`, menus `Menu`, tooltips `Tooltip`.
* Popups and tooltips hang under the widget that opened them (`Area::accessibility_parent`, set through `Popup::anchor_widget`) instead of floating at the root.
* Combo box popups are `ListBox`.
* `on_hover_text` sets the widget's accessible description, and its name if it has none.

Drafted with Claude Code (an LLM), tests included.


crates/egui/src/ui_stack.rs
```diff
@@ -62,6 +62,24 @@ pub enum UiKind {
...
+        match self {
+            Self::LeftPanel
+            | Self::RightPanel
+            | Self::TopPanel
+            | Self::BottomPanel
+            | Self::CentralPanel => Role::Pane,
+            Self::Modal | Self::Popup => Role::Dialog,
+            Self::Menu => Role::Menu,
+            Self::Tooltip => Role::Tooltip,
+            _ => Role::GenericContainer,
```
emilk: do exhaustive matching here

lucasmerlin (approved):
Nice!

PR #8584, opened 2026-09-17
Remove the `AsId` trait

emilk:
We want to be explicit when constructing `Id`s as `Id::unique`. `AsId` was the old implicit way. For implicit parameters, use `impl AsIdSalt`


crates/egui/src/id_salt.rs
```diff
@@ -29,6 +29,17 @@ impl nohash_hasher::IsEnabled for IdSalt {}
 
 impl IdSalt {
     /// Create a new [`IdSalt`] by hashing some source (e.g. a string or integer).
+    ///
+    /// The source is anything that implements [`Hash`](core::hash::Hash),
+    /// including tuples. Prefer a tuple over formatting a string:
```
emilk:
```suggestion
    /// including string, integers, and tuples. Prefer a tuple over formatting a string:
```

crates/egui/src/id.rs
```diff
@@ -89,13 +92,24 @@ impl Id {
...
+    pub fn new(source: impl core::hash::Hash + core::fmt::Debug) -> Self {
         Self::unique(source)
     }
 
     /// Generate a child [`Id`] by salting the parent [`Id`] with the given argument.
     ///
     /// `id.with(salt)` is the same as `id.with_salt(IdSalt::new(salt))`.
+    ///
+    /// The salt is anything that implements [`Hash`](core::hash::Hash),
+    /// including tuples. Prefer a single tuple over a chain of [`Self::with`]:
```
emilk:
```suggestion
    /// including strings, integers, and tuples.
    /// Prefer a single tuple over a chain of [`Self::with`]:
```

PR #8586, opened 2026-09-17
Bring back `AsId`, deprecated

emilk:
Removing `AsId` broke downstream crates that name the trait in their own bounds or re-exports. It is back, but `#[deprecated]`

* https://github.com/emilk/egui/pull/8584


crates/egui/src/id.rs
```diff
@@ -4,6 +4,16 @@ use core::num::NonZeroU64;
 
 use crate::{AsIdSalt, IdSalt};
 
+/// Types that can be converted to an [`Id`].
+///
+/// This is all types implementing `Hash` and `Debug`,
+/// which includes things like strings, integers, tuples of those, etc.
+#[deprecated = "Use `impl core::hash::Hash + core::fmt::Debug` instead"]
```
emilk:
```suggestion
/// Deprecated
#[deprecated = "Use AsIdSalt or Id instead"]
```

PR #8587, opened 2026-09-17
Allow transforming after shapes were tessellated and rounded

lucasmerlin:
Via a new `TransformedShape`. When the transform is applied before rounding during an animation, it introduces a ugly snap once the transform ends:

https://github.com/user-attachments/assets/1e75cd21-4bc4-4869-b3fa-cde749991518

When the transform is applied after rounding it looks way smoother:

https://github.com/user-attachments/assets/ccba1562-66dc-4b93-b187-53928510c0b8


crates/egui/src/layers.rs
```diff
@@ -182,6 +182,31 @@ impl PaintList {
...
+    pub fn transform_after_rounding_range(
+        &mut self,
+        start: ShapeIdx,
+        end: ShapeIdx,
+        transform: TSTransform,
+    ) {
+        for ClippedShape { clip_rect, shape } in &mut self.0[start.0..end.0] {
+            *clip_rect = transform.mul_rect(*clip_rect);
+            shape.transform_after_rounding(transform);
+        }
```
emilk, line 207: This will now cause an allocation per shape.

Two alternatives:
* Extend `ClippedShape` to hold the transform instead of adding a `TransfromShape`
* Join neighboring `ClippedShape`s into a single `ClippedShape { clip_rect, shape: Shape::Transform(…Shape::Vec(…) )  }`… well, that still causes allocations, so nevermind

probably leave it as a TODO; flagging the allocation. Don't want to force us into a huge refactor if we don't need to

crates/epaint/src/shapes/transformed_shape.rs
```diff
@@ -0,0 +1,41 @@
...
+///
+/// For a lasting transform, such as a pan/zoom canvas, transform the shapes instead
+/// ([`Shape::transform`]): the contents then stay snapped to the pixel grid at any transform.
+#[derive(Clone, Debug, PartialEq)]
+pub struct TransformedShape {
+    /// Applied to the tessellated rendering of `shape`.
+    pub transform: TSTransform,
+
+    pub shape: Shape,
+}
```
emilk, line 24: I wonder if we should instead use a unified `TransformedShape` design, with a bool flag to set if the transform is before or after snap-to-pixel. That way we can remove most per-shape transforming, and just wrap shapes that needs transforming in a `TransformedShape`.

emilk, line 24: Usually one transforms more than one shape at once, so we could make this take a vec of shapes… and maybe combine it with `Shape::Vec`

emilk (approved):
feels a bit hacky and tacked-on… but that's mostly because I really want to redesign our current `Shape` system, but now is not the right time. And it is also a clever way to solve the issue-at-hand. So LGTM :)

emilk:
Please add a regression test of some sort!

PR #8588, opened 2026-09-17
Add `DragValue::clip_text` to hold field's width while editing

ljkarel:
A DragValue cannot be held to a width while it is text-edited: the TextEdit it builds for that state hardcodes clip_text(false), so the widget grows to fit whatever has been typed and carries its layout with it. A long entry can stretch a containing popup off-screen. range() does not help, since it only applies once the edit commits.

clip_text(true) forwards to that TextEdit, so the widget keeps its width and the text scrolls inside it, following the cursor. Defaults to false, leaving existing behavior unchanged.

Example on egui.rs:

* Small fix; no corresponding issue created


emilk (approved):
makes sense

PR #8595, opened 2026-09-19
Fix burst of extra repaints before request_repaint_after deadlines

xmakro:
`request_repaint_after` subtracts `predicted_dt` from the delay (added in #4787). But everything that
uses a delayed repaint waits for a deadline (cursor blink, tooltip delay, next GIF frame…). Woken one
frame early, it sees the deadline hasn't passed and asks for the remainder, which is < `predicted_dt`
and so becomes zero: we repaint in a loop until the deadline.

Focused idle `TextEdit`, eframe on Linux/X11, 8 s: 115 frames before (burst of ~7 per blink), 17 after
(one per blink). The blink doesn't happen any earlier with the early wake-up.

`egui_kittest` sets `predicted_dt = step_dt` (250 ms), so there any delay ≤ 250 ms counted as an
immediate repaint (`request_repaint_after(100ms)` made `Harness::run` exceed `max_steps`).


emilk (approved):
LGTM, but please remove the tests

emilk (approved):

PR #8600, opened 2026-09-21
Make `Context::add_glyph_rasterizer` idempotent

emilk:
`Context::add_glyph_rasterizer` appended a rasterizer and requested a repaint on every call. Calling it from a setup path that runs each frame piled up duplicates and kept the UI repainting forever (we hit this in rerun's tests, where `apply_style_and_install_loaders` runs every frame and the harness never settled).

* `GlyphRasterizer::new` now takes a mandatory `key`: adding a rasterizer whose key is already installed is a no-op.
* `Context`/`Fonts`/`FontsImpl::add_glyph_rasterizer` only reload fonts, clear caches and request a repaint when something actually changed. `Fonts::add_glyph_rasterizer` returns whether it did.
* `SvgGlyph::into_rasterizer` is keyed by cluster, the web canvas rasterizer by a fixed name.

Written by an LLM (Claude) together with @emilk.


crates/egui_extras/src/svg_glyph.rs
```diff
@@ -188,6 +192,7 @@ impl SvgGlyph {
             }
         })
         .with_priority(FontPriority::Highest)
+        .with_key(key)
```
emilk: make the key part of the `GlyphRasterizer::new` ctor instead so it is mandatory

emilk: Done in c354eecc1: `GlyphRasterizer::new(key, rasterize)`, `with_key` removed, `key: Arc<str>`.

crates/epaint/src/text/glyph_rasterizer.rs
```diff
@@ -77,13 +81,23 @@ pub struct GlyphRasterizer {
     /// Is this a fallback ([`FontPriority::Lowest`], the default),
     /// or does it override the installed fonts ([`FontPriority::Highest`])?
     pub priority: FontPriority,
+
+    /// Identifies this rasterizer: two rasterizers with the same key are the same rasterizer.
+    ///
+    /// Adding one when a rasterizer with this key is already installed
+    /// replaces that one in place, so it is safe to add it every frame.
+    pub key: Arc<str>,
```
emilk: put first

crates/epaint/src/text/glyph_rasterizer.rs
```diff
@@ -102,12 +117,37 @@ impl GlyphRasterizer {
...
+    #[doc(hidden)]
+    pub fn insert_into(self, rasterizers: &mut Vec<Self>) -> bool {
+        let existing = rasterizers.iter().position(|r| r.key == self.key);
+        if let Some(index) = existing {
+            if rasterizers[index].is_same_as(&self) {
+                false
+            } else {
+                rasterizers[index] = self;
+                true
+            }
```
emilk: I think we should ONLY go by key, not any `is_same_as` stuff

crates/epaint/src/text/glyph_rasterizer.rs
emilk: Done in c0aa5402d.

crates/epaint/src/text/glyph_rasterizer.rs
emilk: Done in c0aa5402d: `is_same_as` removed. Same key already installed = no-op (nothing replaced, no repaint). This also covers a closure recreated every frame, which `Arc::ptr_eq` never would have.

crates/epaint/src/text/glyph_rasterizer.rs
```diff
@@ -149,6 +179,42 @@ pub fn has_emoji_presentation(cluster: &str) -> bool {
...
+    }
+
+    #[test]
+    fn different_keys_append() {
+        let mut list = Vec::new();
+        assert!(rasterizer("a").insert_into(&mut list));
+        assert!(rasterizer("b").insert_into(&mut list));
+        assert_eq!(list.len(), 2);
+    }
+}
```
emilk: rm tests

emilk (commented):
LGTM

crates/epaint/src/text/glyph_rasterizer.rs
emilk: Done in 271da6705.

PR #8601, opened 2026-09-21
Draw tofu for families without fonts, and add `MissingGlyphPolicy` (panics by default in `egui_kittest`)

emilk:
A font family with no font at all used to lay out to an *empty galley*: the run had no face to shape with, so it was dropped. `Galley::clamp_cursor` on that empty galley returns 0, so a `TextEdit` lost its cursor the frame after typing — text stayed, cursor jumped to 0, and e.g. completion popups never opened. Found in rerun's tests after dropping `default_fonts`, where nothing installed a font for the bare kittest harness.

### Changes
* epaint: a family without fonts now lays out **one synthetic tofu box per grapheme cluster**, with metrics derived from the font size. `glyphs.len() == char_count` and non-zero row heights hold even with no font.
* epaint: new `MissingGlyphPolicy { Tofu, Panic }` on `Fonts`, consulted only after every font, `FontProvider` and `GlyphRasterizer` had its turn. Control characters (`\t`, `\n`, …) never panic, since fonts have no glyphs for them. The panic names the character, the family and the installed fonts.
* egui: `Context::set_missing_glyph_policy` / `missing_glyph_policy`.
* egui_kittest: **panics on a missing glyph by default**, since tofu in a snapshot is easy to miss. Opt out with `HarnessBuilder::allow_missing_glyphs()`. One demo test needed it (the emoji completion popup lists emoji the bundled fonts lack).

Top: with fonts, U+10FFFD drawn with Ubuntu's `.notdef`. Bottom: the same UI with `FontDefinitions::empty()`, all synthetic tofu (3× zoom):

<picture>
  
  <source media="(max-width: 480px)" srcset="https://static.rerun.io/tofu_cmp/1c2a567a1f53c301897e716452974a0b6fbc8e12/480w.png">
  <source media="(max-width: 768px)" srcset="https://static.rerun.io/tofu_cmp/1c2a567a1f53c301897e716452974a0b6fbc8e12/768w.png">
</picture>

Written by Claude (LLM) together with @emilk.


crates/egui_kittest/tests/missing_glyphs.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+/// Whitespace and control characters never panic, whatever the fonts have.
+#[test]
+fn whitespace_is_not_a_missing_glyph() {
+    let mut harness = Harness::new_ui(|ui| {
+        ui.label(" \t\n\u{00A0}\u{200B}");
+    });
+    harness.run();
+}
```
emilk: why is this important to test?

crates/epaint/src/text/text_layout.rs
```diff
@@ -468,6 +470,79 @@ fn layout_raster_run(
     );
 }
 
+/// Emit the glyphs of a run in a family that has no font at all.
+///
+/// Each grapheme cluster still gets a glyph: from a fallback
+/// [`GlyphRasterizer`](crate::text::GlyphRasterizer) if one handles it,
+/// else a synthetic tofu box (or nothing but an advance, for whitespace).
+/// Cursors and selections need `glyphs.len() == char_count` to hold even here.
+fn layout_fontless_run(
```
emilk: do we really need another layout function in this file? :/

crates/egui_kittest/tests/missing_glyphs.rs
```diff
@@ -0,0 +1,30 @@
...
```
emilk: why?

crates/epaint/src/text/text_layout.rs
emilk: is there anything we can do to simplify the code in this file?

emilk (commented):
LGTM

crates/epaint/src/text/text_layout.rs
emilk: Removed in 651cafc05. `segment_into_runs` now turns each cluster of a fontless family into a raster run (fallback rasterizer, else synthetic tofu, else an advance for whitespace) via `FontsImpl::fontless_cluster`, so the existing `layout_raster_run` handles it.

crates/egui_kittest/tests/missing_glyphs.rs
emilk: It was not. With the bundled fonts those characters never reach the missing-glyph path: disabling the exemption still let the test pass. Removed in 651cafc05. The exemption itself is covered by epaint's `missing_glyph_policy_panic_exempts_whitespace`, which uses no fonts.

crates/epaint/src/text/fonts.rs
```diff
@@ -629,6 +682,122 @@ impl FontsImpl {
...
+        self.on_missing_glyph(family, cluster);
+
+        let chr = cluster.chars().next();
+        let advance_only = |advance_px: f32| RasterGlyphAllocation {
+            allocation: GlyphAllocation::default(),
+            advance_px,
+        };
+
+        if chr.is_some_and(char::is_whitespace) {
+            advance_only((0.3 * font_size * pixels_per_point).round())
```
emilk: this 0.3 is very arbitrary. Why shouldn't this become a tofu/failure instead?

crates/epaint/src/text/text_layout.rs
```diff
@@ -571,7 +574,7 @@ fn layout_section(
                 continue;
             }
             let Some(font_face) = fonts.face(run.font_key) else {
-                continue;
+                continue; // Cannot happen: `segment_into_runs` turns fontless clusters into raster runs.
```
emilk: add a debug-only-panic then

crates/epaint/src/text/fonts.rs
emilk: No good reason. Done in b20c80971: whitespace and control characters in a fontless family now get a tofu box, or the panic, like everything else. The exemption in `on_missing_glyph` is gone too.

crates/epaint/src/text/text_layout.rs
emilk: Done in b20c80971 (`debug_assert!(false, ..)`, as elsewhere in epaint).

PR #8604, opened 2026-09-21
Add `Response::accessible_name`

emilk:
Useful for adding an accessibility name to a widget


crates/egui/src/context.rs
```diff
@@ -4084,7 +4061,11 @@ impl Context {
     }
 
     /// Does the widget already have an accessibility node this pass?
-    pub(crate) fn has_accesskit_node(&self, id: Id) -> bool {
+    /// Does the widget with this id have a node in the accessibility tree this pass?
```
emilk: weird almost-duplicate docstring line?

emilk: Removed the stale line.

PR #8605, opened 2026-09-21
Name a popup after the widget that opened it

emilk:
Follow-up to #8581: a popup already hangs under the widget that opened it; now it is also *labelled by* that widget (unless its content named it), so a menu, context menu or combo box popup can be found by that widget's name: `get_by_role_and_label(Role::Menu, "File")`, `get_by_role_and_label(Role::ListBox, "Fruit")`.

Rerun did this by hand for each menu; this lets it stop.


emilk:
Closing: labelling a popup by its anchor makes every `get_by_label(button)` ambiguous while the popup is open, and kittest then picks the rect-less popup node (`reopened_popup_resizes_for_wider_items` fails with "Every egui node should have a rect"). The same would hit every downstream test that clicks a button by label while its menu is open. Not worth it; explicit names where they matter instead.

crates/egui_kittest/tests/accesskit.rs
```diff
@@ -223,6 +223,52 @@ fn menu_hangs_under_its_button() {
...
+    let mut harness = Harness::new_ui(|ui| {
+        MenuButton::new("File").ui(ui, |ui| {
+            let _ = ui.button("Open");
+        });
+    });
+
+    harness.get_by_label("File").click();
+    harness.run();
+
+    harness.get_by_role_and_label(Role::Menu, "File");
```
emilk, line 239: for explicitness

```suggestion
    harness.get_by_role_and_label(Role::Button, "File");
    harness.get_by_role_and_label(Role::Menu, "File");
```

crates/egui_kittest/tests/accesskit.rs
```diff
@@ -223,6 +223,52 @@ fn menu_hangs_under_its_button() {
...
+    let mut harness = Harness::new_ui(|ui| {
+        ui.label("Item").context_menu(|ui| {
+            let _ = ui.button("Delete");
+        });
+    });
+
+    harness.get_by_label("Item").click_secondary();
+    harness.run();
+
+    harness.get_by_role_and_label(Role::Menu, "Item");
```
emilk, line 253: same here and below

PR #8606, opened 2026-09-21
Fix invisible ui elements being added to access kit tree

Wumpf:
Previously, invisible UI widgets were still exposed to the accesskit. Not only can this cause issues in tests (for example by seeing duplicate elements during a sizing pass... which happens to be not a far fetched example ;)), it also is just plain wrong for accessibility itself since we only want to express things that are actually on screen in the accesskit tree!

A lot of complexity arises from the fact that we also don't want to _reference_ any hidden ui nodes [...].


crates/egui/src/pass_state.rs
```diff
@@ -71,6 +73,97 @@ impl ScrollTarget {
 pub struct AccessKitPassState {
     pub nodes: IdMap<accesskit::Node>,
     pub parent_map: IdMap<Id>,
+
+    /// Nodes belonging to explicitly invisible UI (not merely clipped or discarded).
+    pub excluded: IdSet,
```
emilk, line 78: We should define `invisible` very clearly: Is this exactly `UI::is_visible`, or something else?

Because I think accesskit stores the nodes of everything that has ever been shown, like a menu that used to be open but is no longer. What do we call such a widget? "An ex-widget"?

crates/egui/src/ui.rs
```diff
@@ -147,12 +150,16 @@ impl Ui {
...
             enabled: true,
+            explicitly_invisible: invisible,
             sizing_pass,
             stack: Arc::new(ui_stack),
             sense,
             min_rect_already_remembered: false,
         };
 
+        if invisible {
+            ui.ctx().exclude_from_accesskit(ui.unique_id);
```
emilk, line 161: why register the widget to accesskit at all then? Seems very weird to say "here is a widget, now forget about it"

Wumpf: hmmm yeah maybe I've let myself be pushed too much by the agent here. My understanding so far was that this is needed because the way things are set up we add things sometimes too eagerly to accesskit and only later notice that a widget or reference to it is invisible.
But writing this I realize that I don't have a great understanding of all that and have to verify this

PR #8589, opened 2026-09-18
Fail kittest tests when an input widget has no accessible name

emilk:
* Stacked on #8581

**⚠️ Breaking ⚠️**

`egui_kittest` that every input widget (button, checkbox, slider, combo box, text field, …) has an accessible name.

Opt out with `HarnessBuilder::with_accessibility_check(false)`, or run it by hand with `Harness::check_accessibility`.

Running it over egui's own tests found real gaps, fixed here:
* An empty label (`checkbox(&mut b, "")`) took precedence over `labelled_by`; now it is not set at all.
* `DragValue` is named after its prefix or suffix (it used to clear the name).
* The color button has a name, with the color as its value.
* `TextEdit` lost its placeholder on frames where the selection changed.
* `Margin`, `CornerRadius` and the font-size slider name their inputs.
* The demos and kittest tests name their remaining inputs.

PR #8608, opened 2026-09-21
Add `Ui::label_inputs_by` and `Ui::name_inputs`

emilk:
Two ways to name every input a closure adds that has no accessible name yet:

* `ui.label_inputs_by(label.id, |ui| …)`: after a row label, for property rows and forms whose label is painted apart from its value widgets, so the editors need not know the row label.
* `ui.name_inputs("Code block", |ui| …)`: a fixed name, for widgets that cannot be named where they are built, e.g. the read-only `TextEdit` `egui_commonmark` renders for a code block.

Only the widgets registered on the `Ui`'s layer during the closure are inspected (the per-layer widget list is in creation order), so the cost is one node lookup per widget added, not a walk over the whole tree per call. Both are what #8589's check asks for and what Rerun does by poking at `accesskit_state` from outside egui today.

`egui::accessibility::is_input(role)` decides which roles count, as an exhaustive match over `accesskit::Role`; `egui_kittest` uses the same function.


crates/egui/src/accessibility.rs
```diff
@@ -0,0 +1,21 @@
+//! Helpers for the accessibility tree.
+
+use crate::accesskit::Role;
+
+/// Roles a user reaches by name: everything you can click, toggle, pick from, drag, or type into.
+///
+/// A widget with one of these roles and no accessible name cannot be found by a screen reader,
+/// nor by a test that looks it up by label.
+pub const INPUT_ROLES: &[Role] = &[
```
emilk: should DisclosureArrow be part of this? if not, why not?

emilk: Yes, good catch, and it was a live hole. Since #8555 a CollapsingHeader reports Role::DisclosureTriangle instead of Role::Button, so it silently dropped out of the check: an unnamed one gave unnamed_widgets().len() == 0 on current main. Added it, plus two tests (an unnamed collapsing header is now flagged, and name_unnamed_inputs can name one).

Still left out, of the roles egui emits: ScrollBar and Splitter. Those are chrome a user does not reach by name, and you said #8578 covers the splitter. Everything else egui emits (Label, Image, Window, Pane, Dialog, Menu, Tooltip, TextRun, ListBox, ProgressIndicator, GenericContainer, Unknown) is not an input.

crates/egui/src/accessibility.rs
```diff
@@ -0,0 +1,22 @@
...
+    Role::CheckBox,
+    Role::ColorWell,
+    Role::ComboBox,
+    Role::DisclosureTriangle,
+    Role::Link,
+    Role::MultilineTextInput,
+    Role::PasswordInput,
+    Role::RadioButton,
+    Role::Slider,
+    Role::SpinButton,
```
emilk: Should include `Role::Splitter`

crates/egui/src/ui.rs
```diff
@@ -963,6 +963,59 @@ impl Ui {
...
+    /// Call it after adding the widgets. Costs nothing when accessibility is off.
+    pub fn name_unnamed_inputs(&self, name: impl Into<String>) {
+        let name = name.into();
+        for id in self.unnamed_inputs() {
+            self.ctx()
+                .accesskit_node_builder(id, |node| node.set_label(name.clone()));
+        }
+    }
+
+    /// Every input widget under this `Ui` this pass that has neither a label nor a `labelled_by`.
```
emilk: define "input widget" by linking to `INPUT_ROLES`

crates/egui/src/accessibility.rs
```diff
@@ -0,0 +1,23 @@
...
+    Role::Splitter,
+    Role::TextInput,
+];
```
emilk: I think this would make more sense as a `pub fn is_input(role: Role) -> bool` - that way it can be an exhaustive match

crates/egui/src/accessibility.rs
emilk: In since 439df518, and `is_input` now lists it.

crates/egui/src/ui.rs
emilk: Done: both docs now say an input is a widget whose role passes `accessibility::is_input`, and link it.

crates/egui/src/accessibility.rs
emilk: Done: `pub fn is_input(role: Role) -> bool`, an exhaustive match over all 182 `Role`s, with the true arm listing what one clicks, toggles, picks, drags or types into. `egui_kittest` uses it too.

crates/egui/src/accessibility.rs
```diff
@@ -0,0 +1,197 @@
...
+        | Role::DateTimeInput
+        | Role::WeekInput
+        | Role::MonthInput
+        | Role::TimeInput
+        | Role::EmailInput
+        | Role::PhoneNumberInput
+        | Role::UrlInput
+        | Role::Slider
+        | Role::SpinButton
+        | Role::Splitter => true,
```
emilk: Sort alphabetically

crates/egui/src/accessibility.rs
```diff
@@ -0,0 +1,197 @@
...
+        | Role::DocPart
+        | Role::DocPreface
+        | Role::DocPrologue
+        | Role::DocPullquote
+        | Role::DocQna
+        | Role::DocSubtitle
+        | Role::DocTip
+        | Role::DocToc
+        | Role::ListGrid
+        | Role::Terminal => false,
```
emilk: Sort alphabetically

emilk (commented):
LGTM

crates/egui/src/accessibility.rs
emilk: Sorted.

crates/egui/src/accessibility.rs
emilk: Sorted.

crates/egui/src/accessibility.rs
```diff
@@ -0,0 +1,197 @@
...
+        | Role::Navigation
+        | Role::Note
+        | Role::PluginObject
+        | Role::ProgressIndicator
+        | Role::RadioGroup
+        | Role::Region
+        | Role::RootWebArea
+        | Role::Ruby
+        | Role::RubyAnnotation
+        | Role::ScrollBar
```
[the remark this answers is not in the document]
emilk: Left out on purpose: a scroll bar is chrome nobody reaches by name (no screen reader announces "the scroll bar named X"), and egui gives it none. Same call @emilk made in the earlier thread.

crates/egui/src/ui.rs
```diff
@@ -963,6 +963,81 @@ impl Ui {
...
+        &mut self,
+        label_id: Id,
+        add_contents: impl FnOnce(&mut Self) -> R,
+    ) -> R {
+        let first = self.widget_count_in_layer();
+        let result = add_contents(self);
+        for id in self.unnamed_inputs_added_since(first) {
+            self.ctx().accesskit_node_builder(id, |node| {
+                node.push_labelled_by(label_id.accesskit_id());
+            });
```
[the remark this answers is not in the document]
emilk: Right. `label_inputs_by` now clears the label before `push_labelled_by`, as `Response::labelled_by` does.

crates/egui/src/ui.rs
```diff
@@ -963,6 +963,81 @@ impl Ui {
...
+        let layer_id = self.layer_id();
+        self.ctx().viewport(|viewport| {
+            let Some(state) = &viewport.this_pass.accesskit_state else {
+                return Vec::new();
+            };
+            viewport
+                .this_pass
+                .widgets
+                .get_layer(layer_id)
+                .skip(first)
```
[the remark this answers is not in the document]
emilk: The only `move_to_top: true` is the window title-bar drag, and a window lives on its own layer, so the list of the layer the closure runs on only grows. Noted in the doc of `unnamed_inputs_added_since`.

crates/egui/src/ui.rs
```diff
@@ -963,6 +963,81 @@ impl Ui {
...
+                .filter_map(|rect| {
+                    let node = state.nodes.get(&rect.id)?;
+                    let unnamed = crate::accessibility::is_input(node.role())
+                        && node.label().is_none_or(|label| label.trim().is_empty())
+                        && node.labelled_by().is_empty();
```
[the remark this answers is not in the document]
emilk: Fixed: a non-blank placeholder now counts as a name, same as in `egui_kittest`, with a test.

PR #8609, opened 2026-09-21
Apply font changes right away while no text has been laid out this pass

emilk:
Generalizes #8602 from "no font loaded yet" to "nothing laid out yet this pass": any font, provider, rasterizer or glyph-policy change made before the first text of a pass takes effect in that pass, since nothing is laid out with the old fonts.

#8602 missed the case where `default_fonts` is on: an app that installs its own fonts during its first pass then laid that pass out with the bundled fonts, and under `MissingGlyphPolicy::Panic` any glyph they lack (Rerun's ⏎, on a Linux CI box with no system fonts) panics. New `Fonts::used_since_begin_pass` is what the check reads.


emilk (commented):
It aint pretty, but LGTM

PR #8613, opened 2026-09-21
A `TextEdit` over an immutable buffer is read-only too

emilk:
Follow-up to #8611. Passing `&mut &str` is how `egui_commonmark` (and the docs) show selectable text; it cannot be typed into any more than `interactive(false)` can, so it now gets the same read-only mark, and `egui_kittest` asks no name of it.


emilk (commented):
LGTM

